PR #2750, opened 2020-06-03
Setup - FileSystem - call parent constructor in ilFileSystemComponent…

daniwe4:
…DataDirectoryCreatedObjective to initialize path property

avoid attempts in setup to create empty directories

https://mantis.ilias.de/view.php?id=28358


klees (approved):

klees:
Hi @chfsx,

I hand this over to you because this needs to be fixed in the FileSystem component that you maintain. I had a look into the code and it looks plausible to me.

Best regards!

chfsx:
@daniwe4 thank you very much for the PR, I'll CP is asap to trunk as well

chfsx:
as I see it's already there :-)

PR #2739, opened 2020-05-29
CSS: Table actions, fix #28414

Amstutz:
This is a try fixing: https://mantis.ilias.de/view.php?id=28414 and related issues. However, this feels like playing mikado. You take one, and break 3 others. I would be very glad if a second dev. could have a look at this. For me and my local installations it seemed to work. Tables in responsive view remain scrollable. Table in other view are not using a scrollable area.

Thx for feedback.

PR #2751, opened 2020-06-04
CSS: Table actions, fix #28414 Revised

Amstutz:
Thx for the feedback of @alex40724 in https://github.com/ILIAS-eLearning/ILIAS/pull/2739. Kudos for spotting the error.

There seemed to be some rebasing issue in the old PR, so I opened this new. Hope that is oke.

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -15848,6 +15848,11 @@ td > img[src$="icon_custom.svg"] {
     max-width: 94vw;
   }
 }
+@media (min-width: 769px) {
+  .table-responsive {
+    overflow: visible;
+  }
+}
 /* Services/Notes */
 div.ilNotes {
   background-color: #f9f9f9;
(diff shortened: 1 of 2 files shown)


Amstutz:
@alex40724 , fine with you if I merge this? This would resolve several rather annoying related issues.

alex40724 (approved):
Yes, thanks.

catkrahl:
@Amstutz: there is something wrong in the stylecode for the responsive-table-class, why this fix is not working. the math-operation in the media-query should stand in brackets and only operate with numbers not with px. so the right expression should be: 
(@grid-float-breakpoint-max + 1)).
only this gives back the intended result of 769px after compiling on my local system! could be that the operation without brackets and  "+1px" only fails in my version of lessc?!
all the best CK

Amstutz:
Thx @catkrahl for the feedback. Yes, the issue has been fixed here (see also discussion referenced in the mantis ticket referenced there): https://github.com/ILIAS-eLearning/ILIAS/commit/f6ee992fbde575e7f7b949a57e48d13964c2a0da#diff-b7e54a2a697227a83d63170b3971b007 .

However, this is also a lessc version issue. It those not seem an issue on the latest lessc version.

PR #2445, opened 2020-01-20
Offline Concept

alex40724:
This is the result of the a first requirements VC of the ILIAS Offline Group [1]. We are planning to work on solutions for the different issues outlined in this document in the next weeks/months. If you are interested, you may join the group any time.

We will create a number of PRs for adressing the different identified requirements:

- https://github.com/ILIAS-eLearning/ILIAS/pull/2457 Javascript General Coding Guidelines
- https://github.com/ILIAS-eLearning/ILIAS/pull/2477 Javascript Modules Guideline
- https://github.com/ILIAS-eLearning/ILIAS/pull/2478 Typescript Guideline

[1] https://docu.ilias.de/goto_docu_grp_8190.html

Relevant after 2nd VC:

@ JourFixe: Please decide if this document can become part of the trunk. I am not sure if this format is ok, or if it should be transformed into ROADMAP.md files (if yes, where to put them?). If it becomes part of the trunk we would agree to share these requirements for a substantial number of different scenarios. It oulines a way to move forward, without including concrete guidelines yet. These should be handed in with separate PRs.

The main benefit would be that it splits a bigger problem into digestible chunks and enables to collaboratively and incrementally work on these different sub-issues.

PR #2752, opened 2020-06-04
JS Unit Testing

alex40724:
This PR is another follow-up of #2445.

The proposal suggests to introduce a unit testing framework for Javascript code in ILIAS.

I had a look Jasmine, Mocha and Jest. This PR goes with Mocha.

[js-unit-test.md](https://github.com/ILIAS-eLearning/ILIAS/pull/2752/files#diff-c007fbbd2fc1d4b458a7795cfce25c04)

[Jasmine](https://jasmine.github.io/), [Mocha](https://mochajs.org/) and [Jest](https://jestjs.io/) are the three most popular unit testing frameworks I found. Jest is the most popular one, originally based on Jasmine (which is the oldest one), but improving things like performance and adding features like snapshots, see [1].

I started with Jest. One of the features I was looking for was ES6 module support. Unfortunately this is not fully implemented in Jest, see [2] and [3]. So I gave Mocha a try and was quickly able so setup a simple [example](https://github.com/ILIAS-eLearning/ILIAS/pull/2752/files#diff-ea8ebfa875f0be429d786271bf78c248).

Mocha's architecture is modular and adds assertion or mock/spy features through additional libraries like [Chai](https://www.chaijs.com/) or [Sinon](https://sinonjs.org/). So the main reasons to suggest Mocha were ES6 module support, flexibility and popularity.

However this choice is open for discussion. Also which other aspects must be covered in a first basic version of such a guideline. 

Please add your comments and thoughts in the PR. You are also invited to join the ILIAS Offline Group [4].

[1] https://medium.com/welldone-software/an-overview-of-javascript-testing-7ce7298b9870
[2] https://github.com/facebook/jest/issues/4842
[3] https://github.com/facebook/jest/issues/9430
[4] https://docu.ilias.de/goto_docu_grp_8190.html

### DIFF
--- a/docs/development/js/js-unit-test-example/src/component.js
+++ b/docs/development/js/js-unit-test-example/src/component.js
@@ -1,0 +1,16 @@
+import math_mod from "./math.js";
+
+export default class Component {
+
+  /**
+   * Pass dependency to math module in constructor
+   * @param math
+   */
+  constructor(math) {
+    this.math = math || math_mod;
+  }
+
+  calculate(a,b) {
+    return this.math.sum(a,b);
+  }
+}
--- a/docs/development/js/js-unit-test-example/src/math.js
+++ b/docs/development/js/js-unit-test-example/src/math.js
@@ -1,0 +1,5 @@
+export default {
+  sum: function(a,b) {
+    return a + b;
+  }
+}
--- a/docs/development/js/js-unit-test-example/test/component.test.js
+++ b/docs/development/js/js-unit-test-example/test/component.test.js
@@ -1,0 +1,30 @@
+import Component from '../src/component.js';
+import { expect } from 'chai';
+
+// test 1
+describe('component', function() {
+
+  let c = new Component();
+
+  it('calculate without mock', function () {
+    expect(c.calculate(1, 2)).to.equal(3);
+  });
+});
+
+// test 2
+describe('component', function() {
+
+  // pass mock as dependency
+  const mathMock = {
+    sum: (a, b) => {
+      return a + b;
+    }
+  };
+
+  let c = new Component(mathMock);
+
+  it('calculate with mock', function () {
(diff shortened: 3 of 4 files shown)


Amstutz:
Finally, somebody proposing a JS-Testing framework. This is desperately needed. So far I do not have any preference. I worked with jasmine in the past and thought it to be fine. I guess mocha will be fine as well. Is it not just mostly a matter of "pick one, stick with it"? However, I do look forward to the reasoning.

Also very interesting will be the question, how we can write testable JS code. In PHP we did get some experience in the past. In js, not so much. Especially if logic is tightly coupled with async commands etc.. Is there any good reading anybody is aware of?

alex40724:
>  Especially if logic is tightly coupled with async commands etc.. Is there any good reading anybody is aware of?

Mocha claims to support testing async code easily, [see their docu](https://mochajs.org/#asynchronous-code). However I am also lacking experience with this.

mjansenDatabay:
@alex40724 Thanks for this PR and the framework comparison. Mocha (and Sinon for mocks) is already used for the (few) chat server tests: https://github.com/ILIAS-eLearning/ILIAS/tree/trunk/Modules/Chatroom/chat/test . I can confirm that setting up Mocha was very easy, also the PHPStorm integration is quite good.

alex40724:
I put this on the JF agenda just to make a short announcement.

matthiaskunkel:
Jour Fixe, 08 JUN 2020 : Alexander notied about his investigation on frameworks for JS unit testing and recommends to use Mocha. If you have comments and questions, please join the VC of the Offline group (https://docu.ilias.de/goto_docu_grp_8190.html). For date and place see https://doodle.com/poll/ct6iqdidcmknxsfh

matthiaskunkel:
Jour Fixe, 22 JUN 2020 : We highly appreciate this suggestion and accept the PR for trunk. Please merge.

PR #2755, opened 2020-06-15
Docs: Change reference system for ILIAS 7

mjansenDatabay:
This PR suggest software versions for the upcoming ILIAS 7 test installations.


matthiaskunkel:
Jour Fixe, 22 JUN 2020 : We highly appreciate this suggestion and accept the defined software versions for the test reference system for ILIAS 7 (test7).

PR #2756, opened 2020-06-17
Trim each entry of valid email domains

shecken:
There is an option to prevent the registration with invalid email domains. It can be a problem, if the administrator uses space after each separating semicolon. In this case it will be nice if we trim each entry to get rid of these spaces.


smeyer-ilias (approved):

smeyer-ilias:
Thanks

PR #2757, opened 2020-06-17
#28495 - MainMenu Subitems

nhaagen:
Use storage-item's parent over gs-item's.
This fixes https://mantis.ilias.de/view.php?id=28495 to the extend that tests are still passing and the described behavior is no longer observable.

However, especially the modifications in class.ilMMAbstractItemFacade.php do not feel too right - so please correct me and maybe point out better ways?
Thanks a lot.


Amstutz:
Thx, probably also related to mantis issue: https://mantis.ilias.de/view.php?id=28426

chfsx:
Hi @nhaagen thank you very much for the PR! @Amstutz I check the other report, too

PR #2758, opened 2020-06-17
setup: add folder Customizing/global/plugins to classmap to include p…

daniwe4:
…lugins

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -75,6 +75,7 @@
 			"ILIAS\\" : "./src"
 		},
 		"classmap": [
+			"./Customizing/global/plugins",
 			"./Services/",
 			"./Modules",
 			"./libs/ilias"


mstuder:
Dear Richard

Can you please explain why this change has been made and how this change is related to https://github.com/ILIAS-eLearning/ILIAS/pull/2782?

Thank you very much for your time.

Best regards
Martin

klees:
Dear @mstuder,

no problem at all =) This is the implementation of that "feature" here:

https://docu.ilias.de/goto.php?target=wiki_1357_Include_Plugins_into_composer_autoloader

The relation to #2782 is as such: The components of ILIAS include themselves into the setup by implementing an [`Agent`](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/Setup/Agent.php). To find these `Agent`s, we use the [`ImplementationOfInterfaceFinder`](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/Setup/ImplementationOfInterfaceFinder.php), which basically uses the composer classmap to, well, find implementations of a given interface. By including the plugins to the classmap `Agent`s in the plugins will be discovered as well.

I hope this and the FW page helps, feel free to ask further questions, though.

Best regards!

theodortruffer:
Hey @klees (and anyone else concerned)

I would like to propose a little adjustment to this change, namely exluding the plugin's vendor folders in the autoloader classmap (should be possible with wildcards: "exclude-from-classmap": [ ..., "./Customizing/**/vendor"]).

The reason for this is that we are using composer libraries in plugins. If multiple plugins are using the same libraries (which is often the case), composer will have an "ambiguous class resolution" warning - which, with composer v3, will not only be a warning but an error. So this PR may render many ILIAS setups uninstallable with our plugins, or if they are installable, wrong library versions may be used.

The above proposition would be an approach to omit these errors/warnings while still being able to build artifacts and find interfaces etc. from the plugin code. It doesn't have to be solved this way, but we need some solution to avoid serious problems with ILIAS 7. I will gladly make a PR if this here solution seems reasonable.

I hope you see this issue as well and I'm open for different approaches and ideas!

Best regards
Theo

klees:
Hi @theodortruffer,

sounds absolutely fine for me. I think we will need some more iterations on that general issue (plugins + dependencies...), but most certainly this won't happen for ILIAS 7. Please create a PR and push it on the JF agenda (via that label), so that others will be informed about this as well.

Thanks a lot!

theodortruffer:
Done: https://github.com/ILIAS-eLearning/ILIAS/pull/2929
 
Thank you for your work on this!

PR #2759, opened 2020-06-17
#28500 - init Explorer2-JS for category selection

nhaagen:
fix for https://mantis.ilias.de/view.php?id=28500;
ilExplorerBaseGUI::init() adds JS to the template that is needed for the tree/category selection.
This is commonly done by components in the MainBar, but if those are deactivated, the JS is missing.

### DIFF
--- a/Modules/StudyProgramme/classes/class.ilObjStudyProgrammeGUI.php
+++ b/Modules/StudyProgramme/classes/class.ilObjStudyProgrammeGUI.php
@@ -226,6 +226,7 @@ class ilObjStudyProgrammeGUI extends ilContainerGUI
                 $this->tabs_gui->activateTab(self::TAB_SETTINGS);
                 $this->tabs_gui->activateSubTab('auto_content');
                 $this->autocategories_gui->setRefId($this->ref_id);
+                $this->initTreeJS();
                 $this->ctrl->forwardCommand($this->autocategories_gui);
                 break;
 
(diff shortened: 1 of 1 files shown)


Modules/StudyProgramme/classes/class.ilObjStudyProgrammeGUI.php
```diff
@@ -863,4 +864,10 @@ protected function deliverCertificateObject()
         );
         $pdf_action->downloadPdf($user_id, $obj_id);
     }
+
+    protected function initTreeJS()
+    {
+        include_once("./Services/UIComponent/Explorer2/classes/class.ilExplorerBaseGUI.php");
```
klees: Please remove the include. We have autoloading via classmap now.

Modules/StudyProgramme/classes/class.ilObjStudyProgrammeGUI.php
```diff
@@ -863,4 +864,10 @@ protected function deliverCertificateObject()
         );
         $pdf_action->downloadPdf($user_id, $obj_id);
     }
+
+    protected function initTreeJS()
```
klees: Please declare `void` return type.

klees (changes requested):
Hi @nhaagen,

I positive that this will work, but have some minor change requests.

Thanks!

PR #2760, opened 2020-06-17
BT 0027895: Fixed individual assessment always having 0 accesses in usage intensity statistics

M-Barz:
I added a call to to recordreadevent on the info screen of individual assessments, which is the only page a normal user is able to see.
So now it shows up correctly in the statistics when checking these.

### DIFF
--- a/Modules/IndividualAssessment/classes/class.ilObjIndividualAssessmentGUI.php
+++ b/Modules/IndividualAssessment/classes/class.ilObjIndividualAssessmentGUI.php
@@ -156,6 +156,13 @@ class ilObjIndividualAssessmentGUI extends ilObjectGUI
         $this->ctrl->setCmdClass('ilinfoscreengui');
         $info = $this->buildInfoScreen();
         $this->ctrl->forwardCommand($info);
+
+        ilChangeEvent::_recordReadEvent(
+            $this->object->getType(),
+            $this->object->getRefId(),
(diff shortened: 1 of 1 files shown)


klees (changes requested):
Hi @M-Barz,

thanks a lot for looking into this. I have one question and one request. The question: Is it really enough to only put that call into the `viewObject`-method? Or put differently: For which actions of a user would we want to count an access? What about, e.g., accessing the settings? The request: Please move the call to `ilChangeEvent` into a `recordRead` (maybe you'll find a more appropriate name) which we can then potentially use for other circumstances where we want to record an access.

Thanks!

M-Barz:
Hi @klees,
I think it's a good idea to move the call to `ilChangeEvent` to it's own function, too. If there is going to be a wish to record the calls of other individual-Assessment tabs, we can now easily call the record function in one of the cases of the `executeCommand` Function.

To answer your question:
As far as I understood it, we are mostly interested to see how often actual users open the assessment. These users can only see the info-Tab with their data however.
The same usage of the `recordReadEvent` function can also be seen when looking at Learning Modules for example, where it's only called at places a normal user can reach.

If I am missing something, please let me know! Thank you!

klees:
Hi @M-Barz,

I have picked this to trunk and release_5-4 as well.

Best regards!

PR #2761, opened 2020-06-18
BT-0028182: Removed delete button in registration code list if write-rights are missing

M-Barz:
Also replaced rights check in listcodes function with checkAccess and checkAccessBool


smeyer-ilias (approved):
Thanks,
looks good.

PR #2764, opened 2020-06-19
fix #28501 - do not position DDs for negative width-values

nhaagen:
https://mantis.ilias.de/view.php?id=28501


Amstutz:
Did some basic testing, seems to work.

Amstutz (approved):

klees:
Picked to release_6 as well.

PR #2767, opened 2020-06-22
UI, Revision of Input-Rendering

nhaagen:
The inputs' renderer is currently quite nested, with a lot of sequenced method-calls and repeated switches/conditions on classes in between.
This is a little more straight forward, and  has a rendering-function for every component.
Please kindly consider :)

### DIFF
--- a/src/UI/Implementation/Component/Input/Field/Renderer.php
+++ b/src/UI/Implementation/Component/Input/Field/Renderer.php
@@ -6,13 +6,12 @@ namespace ILIAS\UI\Implementation\Component\Input\Field;
 
 use ILIAS\Data\DateFormat as DateFormat;
 use ILIAS\UI\Component;
-use ILIAS\UI\Component\Input\Field\MultiSelect;
-use ILIAS\UI\Component\Input\Field\Password;
-use ILIAS\UI\Component\Input\Field\Select;
+use ILIAS\UI\Component\Input\Field as F;
+
 use ILIAS\UI\Implementation\Render\AbstractComponentRenderer;
 use ILIAS\UI\Implementation\Render\ResourceRegistry;
-use ILIAS\UI\Implementation\Render\Template;
 use ILIAS\UI\Renderer as RendererInterface;
+use ILIAS\UI\Implementation\Render\Template;
 
 /**
  * Class Renderer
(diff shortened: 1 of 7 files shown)


Amstutz (changes requested):
Hi @nhaagen 

Thank you so much for cleaning this up. This was overdue. 

One question:
- [ ] Public Interface Usage in implementation: For descends of Field Input, you use the public interface for type hinting in the implementation for the renderer. This leads to some issues when trying to use auto complete or using code inspection, since in some cases the implementation of those fields offer more functions than the public interface. Could we not switch to the implementations in the implementation of the renderer?

Maybe change or give some reasoning why not:
- [ ] Unused variables: I do not like lines such as `$id = $this->bindJSandApplyId($component, $tpl);` when $id is not used afterwards. Could we not just leave out $id?


There is one issue we need to resolve before merging:
- [ ] Section Input: They are currently rendered as group inputs. Also, if rendered as section, I believe to would through an error, due to some bad naming of template variables. I wondered, why this was not caught by our unit tests. Seems section is lacking one for rendering, this is clearly not your mistake, so if you want me to, Ill provide one for trunk which then should pass for your refactorings. I am somewhat unsure if we have more such cases, so we will need to double check the completeness of our test suite here. Note that the switch statements is not independent from the order in which the items to be rendered. E.g. since Section is also a Group, in most appear before the group statements. Seems somewhat implicit to me, but I do not have a good proposal on enhancing this other than working with a more complete set of testst.

Several things remain to do here. However, I am almost sure, they have been there before. I would be fine with doing those things in a next cleanup session, since it seems somewhat crude for asking the guy taking the broom an cleaning mostly somebody elses mess of also taking on the brush for cleaning the toilet ;-).

However, if you still ressources to smash into this, here a short washlist:
- [ ] Inline JS: getOptionalGroupOnLoadCode and getSwitchableGroupOnLoadCode have to much JS inline php code. Goal is to have one line with maybe 3 lines top in some exceptions and all the rest in properly scoped external js files with an interface to be called from the php side. 
- [ ] Double Check the completeness of our Input tests: It seems like we (probably I) missed to complete the set of tests for some of the field. We should close those gaps.

However, most importantly, thx A LOT for this work. Much appreciated. @klees maybe ping me to shortly discuss on how to proceed.

nhaagen:
Hi @Amstutz ,
thank you for looking into this. When you said "completeness of Input tests", you probably had more in mind than the Section?
Do you still remember and would you then give me a hint? There are so many places where we could improve or extend tests...
Thanks, and cu.

Amstutz:
@nhaagen 

sry, I missed out on your reply here. Just stumbled over this, when wondering, while this is still open. I observed, that several Inputs are missing tests for rendering, while analyzing why one issue was not caught by our tests. IMO we should have one for each input element. However, to get this here done, I would just add one for section here otherwise we might overload this important task.

Ping me, if this is ready for a final review. Thx you very much.

nhaagen:
Hi @Amstutz ,
I finally understood - and fixed - the matter with sections ;( 
I'll try and write a few more rendering tests, but for now I just added a roadmap-entry.
Hope, that's OK for you...

Amstutz (approved):
AFAIS may points have been taken care of. @klees add your review and maybe merge afterwards after resolving conflicts

PR #2770, opened 2020-06-25
ScormAicc: fix 28363

klees:
With Chrome 80+ it became deprecated to send async requests during beforeunload:

* https://www.chromestatus.com/feature/4664843055398912
* https://mantis.ilias.de/view.php?id=28363

This is why we now send and async request on beforeunload on chrome and just pretend that everything is "ok".

This is a hack, but worked at least for me. Maybe someone could check if it works for them too. I'll look into a similar solution for SCORM 2004 in the meantime.

### DIFF
--- a/Modules/ScormAicc/scripts/basisAPI.js
+++ b/Modules/ScormAicc/scripts/basisAPI.js
@@ -1,6 +1,7 @@
-/* Copyright (c) 1998-2010 ILIAS open source, Extended GPL, see docs/LICENSE */
+/* Copyright (c) 1998-2020 ILIAS open source, Extended GPL, see docs/LICENSE */
 /**
 * @author  Uwe Kohnle <kohnle@internetlehrer-gmbh.de>
+* @author  Richard Klees <richard.klees@concepts-and-training.de>
 * @version $Id$
 */
 
@@ -291,6 +292,26 @@ function status4tree(i_sco,s_status,s_time){
 	}
 }
 
+// Find out if we are in a beforeunload event
+function isBeforeUnloadEvent() {
+	if (window.event && window.event.type === "beforeunload") {
+		return true;
+	}
+	var frame;
+	for (var i = 0; i < window.frames.length; i++) {
+		frame = window.frames[i];
+		if (frame.event && frame.event.type === "beforeunload") {
+			return true;
+		}
+	}
+	return false;
+}
+
+// Find out if we are using Chrome 80+
+function isChromium80plus() {
+	return navigator.userAgent.match(/Chrom(e|ium)\/8\d+/) !== null;
+}
+
 // store data
 function IliasCommit() {
 	if (a_toStore.length==0){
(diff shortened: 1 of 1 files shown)


Modules/ScormAicc/scripts/basisAPI.js
```diff
@@ -337,7 +358,21 @@ function IliasCommit() {
...
+				// ATTENTION: This is a mean hack!
+				//
+				// With Chrome 80+ it became deprecated to send async requests
+				// during beforeunload:
+				// 	- https://www.chromestatus.com/feature/4664843055398912
+				// 	- https://mantis.ilias.de/view.php?id=28363
+				// This is why we send and async request here and just pretend
+				// that everything was "ok".
+				sendRequest ("./storeScorm.php?package_id="+iv.objId+"&ref_id="+iv.refId+"&client_id="+iv.clientId+"&do=store", s_s, function () {});
+				ret="ok";
```
chfsx, line 371: That means if something went wrong, wouldn't you notice? I don't think there's anything else we could do, is there?

chfsx (approved):
Thank you very much for your suggestion. From my point of view it looks good! I would support a fast merge and in my opinion new releases of ILIAS.

Modules/ScormAicc/scripts/basisAPI.js
klees: As far as I understand the state of affairs, the SCORM-API is synchronous here, so an immediate answer is expected by the client. I cannot see how we might get that answer with an async request. So yes, the client won't notice if something went wrong. But: since we are just in the process of closing the browser there is not much he could do anyway, even if we would detect a problem. Also: the current situation for Chrome 80+ is complete loss of data, this should at least be better =)

chfsx: absolutely, everything is better than the current situation :-)

Modules/ScormAicc/scripts/basisAPI.js
```diff
@@ -291,6 +292,26 @@ function status4tree(i_sco,s_status,s_time){
 	}
 }
 
+// Find out if we are in a beforeunload event
+function isBeforeUnloadEvent() {
```
klees, line 296: For some reasons I don't understand, this does not seem to work for SingleSCO, at least not for the one I am testing with. `event` is not set, neither on the window nor on any of the frames. Did someone test this with a SingleSCO-WBT? Is this something specific for the WBT I have tested with or is this a general problem with SingleSCO?

Uwe-Kohnle: Please check https://github.com/ILIAS-eLearning/ILIAS/commit/f9bec0286ece536059a02b674b59699044723ece 
My solution uses sendBeacon and works with singleSCO- and multiSCO-WBTs. Still an issue: Start last visited SCO at relaunch of a multiSCO-WBT.

Modules/ScormAicc/scripts/basisAPI.js
```diff
@@ -291,6 +292,26 @@ function status4tree(i_sco,s_status,s_time){
...
+		if (frame.event && frame.event.type === "beforeunload") {
+			return true;
+		}
+	}
+	return false;
+}
+
+// Find out if we are using Chrome 80+
+function isChromium80plus() {
+	return navigator.userAgent.match(/Chrom(e|ium)\/8\d+/) !== null;
```
corro, line 312: This regex only matches for Chrome 80-89, is this intended? AFAIK there are no plans to allow these kinds of async requests in the future again and we may have to live with this workaround for quite a while. Depending on the release schedule of Chrome this might break in <1 year.

klees:
Thanks y'all. Uwe's fix works for 3 of 4 WBTs that we tested after #2797 is applied.

PR #2771, opened 2020-06-25
#28543: Add Permission Check in User Administration

shecken:
Check permission to create user when someone tries to import a user. Same when rendering the according button.


mjansenDatabay:
@smeyer-ilias We stumbled upon this old PR in our TB meeting. Do you have any objections to merge this?

smeyer-ilias:
Thanks for providing a PR but due to conflicts, I fixed this without merging it.

PR #2772, opened 2020-06-26
Database/Setup: trigger custom updates in db update

klees:
Currently the setup from CLI does not trigger custom updates. This should do the trick.


chfsx (approved):
Thanks a lot!

chfsx:
I just cherry-picked it to trunk as well

PR #2724, opened 2020-05-26
[WIP] UI, DataTable - Actions

nhaagen:
Based on https://github.com/ILIAS-eLearning/ILIAS/pull/2679, this is an idea on Actions in tables.

Please refer to the [docs](https://github.com/ILIAS-eLearning/ILIAS/pull/2724/files#diff-488841fa74a3b3894b10afa09f10f507R209-R246) and the [example](https://github.com/ILIAS-eLearning/ILIAS/pull/2724/files#diff-da67b86bfcde546a13848a188b8970a2).

Please ask and comment ;)

PR #2780, opened 2020-06-29
UI/mainbar.js - split and roll up 

nhaagen:
This splits up JS for mainbar according to its internal parts and uses rollup (Thanks @alex40724 !) to re-assemble.
It also includes fixes for 
https://mantis.ilias.de/view.php?id=28282
and
https://mantis.ilias.de/view.php?id=28237
https://mantis.ilias.de/view.php?id=28488


Amstutz:
Hi @nhaagen 

Thx a lot for this. Looking oke, I did some basic checks on the frontend, AND even better, drumroll, ran the first productive js tests of yours with the new mocha environment. 

What cost me some time was spotting the hint of introducing, the scripts line documented here: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js/js-unit-test.md into package json. @alex40724 Why do we ship the package.json not by default with this line?

@klees I am not yet really fond of how js tests are organized for the UI Comonents. Here @nhaagen is proposing of linking the tests from the ../src/UI/templates/js/MainControls/test/mainbar.test.js file to tests/mainbar.test.js . Do we really want to go this way? Also now that I see it, it feels funky having tests AND test on top level. I would like to have the php tests and the js tests much closer togater (e.g. in tests/UI/...) since the question of what logic is on the js and what on the PHP side seems like an implementation detail to me.

However, for the moment this seems in line with the currently accepted way of organizing js tests. So this can be shipped and we can discuss the detailed organization in a separate thread.

nhaagen:
Hey folks,
any update on this? 
I see the tests/test and linking uncleanliness, and fully agree that this is more than a bit sub optimal,
but also I feel that that is another issue altogether which can be changed easily enough.
Should we not merge this one in the meantime? 
I'm already adding another JS-test with https://github.com/ILIAS-eLearning/ILIAS/pull/2724, btw...
Thanks and Greetings,
Nils

Amstutz (approved):
I am fine with merging, thx a lot @nhaagen . Only trunk, right? @klees merge when ready.

PR #2781, opened 2020-06-29
BT-28539: Added missing back tab to role administration

M-Barz:


Services/AccessControl/classes/class.ilObjRoleGUI.php
```diff
@@ -1420,16 +1420,10 @@ public function getTabs()
 
         // not so nice (workaround for using tabs in repository)
         $this->tabs_gui->clearTargets();
+        $this->tabs_gui->setBackTarget($this->lng->txt('previous'), $this->ctrl->getLinkTargetByClass("ilobjrolefoldergui", "view"));
```
smeyer-ilias: Is this really sufficient? Is the back tab also shown when editing role templates? Please can you check this.

Please can you use the language variable "btn_back" here.

smeyer-ilias (changes requested):

Services/AccessControl/classes/class.ilObjRoleGUI.php
M-Barz: I checked and it turned out to not be sufficient, sorry about that. It's now fixed, fine and dandy however. :)

smeyer-ilias:
Thanks @M-Barz

PR #2782, opened 2020-06-30
Setup: implement ability to handle plugins during cli setup

daniwe4:
Add new options and arguments for the commands install and update:

- install                          -> will setup an ILIAS-Instance including plugins
- install --no_plugins     -> will setup an ILIAS-Instance without plugins
- install plugin_name    -> will setup the plugin plugin_name

- update                          -> will update an ILIAS-Instance including plugins
- update --no_plugins     -> will update an ILIAS-Instance without plugins
- update plugin_name    -> will update the plugin plugin_name

Add option and functionality depending on [JourFix Discusion Point 9](https://docu.ilias.de/goto.php?target=wiki_1357_Setup_-_Update_Plugins_via_Setup):

- install --skip plugin_name    -> will setup an ILIAS-Instance excluding selected plugin/s
- update --skip plugin_name  -> will update an ILIAS-Instance excluding selected plugin/s

- add a flag for plugin.php  'ignore_cli_setup'. Default or not set in plugin.php results in value false.

### DIFF
--- a/Services/Component/classes/Setup/class.ilComponentActivatePluginsObjective.php
+++ b/Services/Component/classes/Setup/class.ilComponentActivatePluginsObjective.php
@@ -1,0 +1,150 @@
+<?php
+/* Copyright (c) 2020 Daniel Weise <daniel.weise@concepts-and-training.de> Extended GPL, see docs/LICENSE */
+
+declare(strict_types=1);
+
+use ILIAS\Setup;
+use ILIAS\DI;
+
+class ilComponentActivatePluginsObjective implements Setup\Objective
+{
+    /**
+     * @var string
+     */
+    protected $plugin_name;
+
+    public function __construct(string $plugin_name)
+    {
+        $this->plugin_name = $plugin_name;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getHash() : string
+    {
+        return hash("sha256", self::class . $this->plugin_name);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getLabel() : string
+    {
+        return "Activate plugin $this->plugin_name.";
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function isNotable() : bool
+    {
+        return true;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getPreconditions(Setup\Environment $environment) : array
+    {
+        $setup_config = $environment->getConfigFor('common');
+        $db_config = $environment->getConfigFor('database');
+
+        return [
+            new \ilIniFilesPopulatedObjective($setup_config),
+            new \ilDatabasePopulatedObjective($db_config),
+            new \ilComponentPluginAdminInitObjective()
+        ];
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function achieve(Setup\Environment $environment) : Setup\Environment
+    {
+        $ORIG_DIC = $this->initEnvironment($environment);
+
+        $plugin = $GLOBALS["DIC"]["ilPluginAdmin"]->getRawPluginDataFor($this->plugin_name);
+
+        if (!is_null($plugin) && $plugin['activation_possible'] && $plugin['supports_cli_setup']) {
+            $pl = ilPlugin::getPluginObject(
+                $plugin['component_type'],
+                $plugin['component_name'],
+                $plugin['slot_id'],
+                $plugin['name']
+            );
+
+            $pl->activate();
+        }
+
+        $GLOBALS["DIC"] = $ORIG_DIC;
+
+        return $environment;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function isApplicable(Setup\Environment $environment) : bool
+    {
+        $ORIG_DIC = $this->initEnvironment($environment);
+
+        $plugin = $GLOBALS["DIC"]["ilPluginAdmin"]->getRawPluginDataFor($this->plugin_name);
+
+        if (is_null($plugin) || !$plugin['supports_cli_setup']) {
+            return false;
+        }
+
+        $GLOBALS["DIC"] = $ORIG_DIC;
+
+        return $plugin['activation_possible'];
+    }
+
+    protected function initEnvironment(Setup\Environment $environment) : ILIAS\DI\Container
+    {
+        $db = $environment->getResource(Setup\Environment::RESOURCE_DATABASE);
+        $plugin_admin = $environment->getResource(Setup\Environment::RESOURCE_PLUGIN_ADMIN);
+        $ini = $environment->getResource(Setup\Environment::RESOURCE_ILIAS_INI);
+        $client_ini = $environment->getResource(Setup\Environment::RESOURCE_CLIENT_INI);
+
+
+        // ATTENTION: This is a total abomination. It only exists to allow various
+        // sub components of the various readers to run. This is a memento to the
+        // fact, that dependency injection is something we want. Currently, every
+        // component could just service locate the whole world via the global $DIC.
+        $DIC = $GLOBALS["DIC"];
+        $GLOBALS["DIC"] = new DI\Container();
+        $GLOBALS["DIC"]["ilDB"] = $db;
+        $GLOBALS["DIC"]["ilIliasIniFile"] = $ini;
+        $GLOBALS["DIC"]["ilClientIniFile"] = $client_ini;
+        $GLOBALS["DIC"]["ilLoggerFactory"] = new class() {
+            public function getLogger()
+            {
+            }
+        };
+        $GLOBALS["DIC"]["ilBench"] = null;
+        $GLOBALS["DIC"]["lng"] = new ilLanguage('en');
+        $GLOBALS["DIC"]["ilPluginAdmin"] = $plugin_admin;
+        $GLOBALS["DIC"]["ilCtrl"] = new ilCtrl();
+        $GLOBALS["DIC"]["ilias"] = null;
+        $GLOBALS["DIC"]["ilLog"] = null;
+        $GLOBALS["DIC"]["ilErr"] = null;
+        $GLOBALS["DIC"]["tree"] = null;
+        $GLOBALS["DIC"]["ilAppEventHandler"] = null;
+        $GLOBALS["DIC"]["ilSetting"] = new ilSetting();
+        $GLOBALS["DIC"]["objDefinition"] = new ilObjectDefinition();
+        $GLOBALS["DIC"]["ilUser"] = new class() {
+            public $prefs = [];
+            public function __construct()
+            {
+                $this->prefs['language'] = 'en';
+            }
+        };
+
+        if (!defined('DEBUG')) {
+            define('DEBUG', false);
+        }
+
+        return $DIC;
+    }
+}
--- a/Services/Component/classes/Setup/class.ilComponentInstallPluginObjective.php
+++ b/Services/Component/classes/Setup/class.ilComponentInstallPluginObjective.php
@@ -1,0 +1,150 @@
+<?php
+/* Copyright (c) 2020 Daniel Weise <daniel.weise@concepts-and-training.de> Extended GPL, see docs/LICENSE */
+
+declare(strict_types=1);
+
+use ILIAS\Setup;
+use ILIAS\DI;
+
+class ilComponentInstallPluginObjective implements Setup\Objective
+{
+    /**
+     * @var string
+     */
+    protected $plugin_name;
+
+    public function __construct(string $plugin_name)
+    {
+        $this->plugin_name = $plugin_name;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getHash() : string
+    {
+        return hash("sha256", self::class . $this->plugin_name);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getLabel() : string
+    {
+        return "Install plugin $this->plugin_name.";
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function isNotable() : bool
+    {
+        return true;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getPreconditions(Setup\Environment $environment) : array
+    {
+        $setup_config = $environment->getConfigFor('common');
+        $db_config = $environment->getConfigFor('database');
+
+        return [
+            new \ilIniFilesPopulatedObjective($setup_config),
+            new \ilDatabasePopulatedObjective($db_config),
+            new \ilComponentPluginAdminInitObjective()
+        ];
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function achieve(Setup\Environment $environment) : Setup\Environment
+    {
+        $ORIG_DIC = $this->initEnvironment($environment);
+
+        $plugin = $GLOBALS["DIC"]["ilPluginAdmin"]->getRawPluginDataFor($this->plugin_name);
+
+        if (!is_null($plugin) && $plugin['must_install'] && $plugin['supports_cli_setup']) {
+            $pl = ilPlugin::getPluginObject(
+                $plugin['component_type'],
+                $plugin['component_name'],
+                $plugin['slot_id'],
+                $plugin['name']
+            );
+
+            $pl->install();
+        }
+
+        $GLOBALS["DIC"] = $ORIG_DIC;
+
+        return $environment;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function isApplicable(Setup\Environment $environment) : bool
+    {
+        $ORIG_DIC = $this->initEnvironment($environment);
+
+        $plugin = $GLOBALS["DIC"]["ilPluginAdmin"]->getRawPluginDataFor($this->plugin_name);
+
+        if (is_null($plugin) || !$plugin['supports_cli_setup']) {
+            return false;
+        }
+
+        $GLOBALS["DIC"] = $ORIG_DIC;
+
+        return $plugin['must_install'];
+    }
+
+    protected function initEnvironment(Setup\Environment $environment) : ILIAS\DI\Container
+    {
+        $db = $environment->getResource(Setup\Environment::RESOURCE_DATABASE);
+        $plugin_admin = $environment->getResource(Setup\Environment::RESOURCE_PLUGIN_ADMIN);
+        $ini = $environment->getResource(Setup\Environment::RESOURCE_ILIAS_INI);
+        $client_ini = $environment->getResource(Setup\Environment::RESOURCE_CLIENT_INI);
+
+
+        // ATTENTION: This is a total abomination. It only exists to allow various
+        // sub components of the various readers to run. This is a memento to the
+        // fact, that dependency injection is something we want. Currently, every
+        // component could just service locate the whole world via the global $DIC.
+        $DIC = $GLOBALS["DIC"];
+        $GLOBALS["DIC"] = new DI\Container();
+        $GLOBALS["DIC"]["ilDB"] = $db;
+        $GLOBALS["DIC"]["ilIliasIniFile"] = $ini;
+        $GLOBALS["DIC"]["ilClientIniFile"] = $client_ini;
+        $GLOBALS["DIC"]["ilLoggerFactory"] = new class() {
+            public function getLogger()
+            {
+            }
+        };
+        $GLOBALS["DIC"]["ilBench"] = null;
+        $GLOBALS["DIC"]["lng"] = new ilLanguage('en');
+        $GLOBALS["DIC"]["ilPluginAdmin"] = $plugin_admin;
+        $GLOBALS["DIC"]["ilCtrl"] = new ilCtrl();
+        $GLOBALS["DIC"]["ilias"] = null;
+        $GLOBALS["DIC"]["ilLog"] = null;
+        $GLOBALS["DIC"]["ilErr"] = null;
+        $GLOBALS["DIC"]["tree"] = null;
+        $GLOBALS["DIC"]["ilAppEventHandler"] = null;
+        $GLOBALS["DIC"]["ilSetting"] = new ilSetting();
+        $GLOBALS["DIC"]["objDefinition"] = new ilObjectDefinition();
+        $GLOBALS["DIC"]["ilUser"] = new class() {
+            public $prefs = [];
+            public function __construct()
+            {
+                $this->prefs['language'] = 'en';
+            }
+        };
+
+        if (!defined('DEBUG')) {
+            define('DEBUG', false);
+        }
+
+        return $DIC;
+    }
+}
--- a/Services/Component/classes/class.ilPlugin.php
+++ b/Services/Component/classes/class.ilPlugin.php
@@ -54,6 +54,10 @@ abstract class ilPlugin
      * @var ProviderCollection
      */
     protected $provider_collection;
+    /**
+     * @var string
+     */
+    protected $message;
 
 
     public function __construct()
--- a/Services/Component/classes/class.ilPluginAdmin.php
+++ b/Services/Component/classes/class.ilPluginAdmin.php
@@ -74,7 +74,6 @@ class ilPluginAdmin
             if (!is_file($plugin_php_file)) {
                 throw new ilPluginException("No plugin.php file found for Plugin :" . $a_pname . ".");
             }
-
             $plugin_db_data = ilPlugin::getPluginRecord($a_ctype, $a_cname, $a_slot_id, $a_pname);
             $plugin_data = $this->parsePluginPhp($plugin_php_file);
 
@@ -243,8 +242,9 @@ class ilPluginAdmin
             "ilias_max_version" => $ilias_max_version,
             "responsible" => $responsible,
             "responsible_mail" => $responsible_mail,
-            "learning_progress" => (bool) $learning_progress,
-            "supports_export" => (bool) $supports_export,
+            "learning_progress" => (bool) ($learning_progress ?? false),
+            "supports_export" => (bool) ($supports_export ?? false),
+            "supports_cli_setup" => (bool) ($supports_cli_setup ?? true)
         ];
 
         return $values;
@@ -513,7 +513,6 @@ class ilPluginAdmin
         return $this->data[$a_ctype][$a_cname][$a_slot_id][$a_pname]["supports_export"];
     }
 
-
     /**
      * Get info for all plugins.
      *
@@ -633,5 +632,56 @@ class ilPluginAdmin
                 yield $pl->getGlobalScreenProviderCollection();
             }
         }
+    }
+
+    public function getRawPluginDataFor(string $name) : ?array
+    {
+        $this->clearCachedData();
+
+        $modules = ilModule::getAvailableCoreModules();
+        $services = ilService::getAvailableCoreServices();
+
+        foreach ($this->getPluginSlots($modules, IL_COMP_MODULE) as $plugin_slot) {
+            $plugin = $plugin_slot->getPluginInformationFor($name);
+            if (!is_null($plugin)) {
+                return $plugin;
+            }
+        }
+
+        foreach ($this->getPluginSlots($services, IL_COMP_SERVICE) as $plugin_slot) {
+            $plugin = $plugin_slot->getPluginInformationFor($name);
+            if (!is_null($plugin)) {
+                return $plugin;
+            }
+        }
+
+        return null;
+    }
+
+    protected function getPluginSlots(array $components, string $type) : \Iterator
+    {
+        foreach ($this->getComponentSlotsByType($components, $type) as $slot) {
+            $subdir = (explode('/', $slot['component']))[1];
+            yield new ilPluginSlot($type, $subdir, $slot['id']);
+        }
+    }
+
+    protected function getComponentSlotsByType(array $components, string $type) : \Iterator
+    {
+        foreach ($components as $component) {
+            $component_slots = ilComponent::lookupPluginSlots($type, $component["subdir"]);
+            foreach ($component_slots as $component_slot) {
+                if (count($component_slot) > 0) {
+                    yield $component_slot;
+                }
+            }
+        }
+    }
+
+    final public function clearCachedData()
+    {
+        unset($this->data);
+        unset($this->got_data);
+        ilCachedComponentData::flush();
     }
 }
--- a/Services/Component/classes/class.ilPluginRawReader.php
+++ b/Services/Component/classes/class.ilPluginRawReader.php
@@ -8,7 +8,7 @@ class ilPluginRawReader
     const BASE_PLUGIN_PATH = 'Customizing/global/plugins';
     const SEARCH_PATTERN = 'plugin.php';
 
-    public function getPluginNames() : \Iterator
+    public function getPluginNames() : ?\Iterator
     {
         if (!@is_dir(self::BASE_PLUGIN_PATH)) {
             throw new LogicException('Path not found: ' . self::BASE_PLUGIN_PATH);
--- a/Services/Component/classes/class.ilPluginRawReader.php
+++ b/Services/Component/classes/class.ilPluginRawReader.php
@@ -1,0 +1,25 @@
+<?php
+/* Copyright (c) 2020 Daniel Weise <daniel.weise@concepts-and-training.de> Extended GPL, see docs/LICENSE */
+
+declare(strict_types=1);
+
+class ilPluginRawReader
+{
+    const BASE_PLUGIN_PATH = 'Customizing/global/plugins';
+    const SEARCH_PATTERN = 'plugin.php';
+
+    public function getPluginNames() : \Iterator
+    {
+        if (!@is_dir(self::BASE_PLUGIN_PATH)) {
+            throw new LogicException('Path not found: ' . self::BASE_PLUGIN_PATH);
+        }
+
+        $it = new RecursiveDirectoryIterator(self::BASE_PLUGIN_PATH);
+        foreach (new RecursiveIteratorIterator($it) as $file) {
+            $path = $file->getPathName();
+            if (is_file($path) && basename($path) === self::SEARCH_PATTERN) {
+                yield basename(dirname($path));
+            }
+        }
+    }
+}
--- a/Services/Database/classes/Setup/class.ilDatabaseSetupConfig.php
+++ b/Services/Database/classes/Setup/class.ilDatabaseSetupConfig.php
@@ -36,7 +36,7 @@ class ilDatabaseSetupConfig implements Setup\Config
     protected $create_database;
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $collation;
 
@@ -51,7 +51,7 @@ class ilDatabaseSetupConfig implements Setup\Config
     protected $password;
 
     /**
-     * @var	string
+     * @var	string|null
      */
     protected $path_to_db_dump;
 
--- a/Services/Database/classes/class.ilDBUpdate.php
+++ b/Services/Database/classes/class.ilDBUpdate.php
@@ -32,6 +32,14 @@ class ilDBUpdate
      * @var string
      */
     public $updateMsg;
+    /**
+     * @var ilIniFile|null
+     */
+    protected $client_ini;
+
+    protected $custom_updates_current_version;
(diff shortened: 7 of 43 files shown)


setup/README.md
```diff
@@ -1,21 +1,28 @@
...
-* `install` will [set an installation up](#install-ilias)
-* `update` will [update an installation](#update-ilias)
+* `install` will [set an installation up including plugins](#install-ilias)
+* `update` will [update an installation including plugins](#update-ilias)
 * `build-artifacts` [recreates static assets](#build-ilias-artifacts) of an installation
 * `reload-control-structure` [rebuilds structure information](#build-ilias-artifacts) of an installation
 
+`install` and `update` also supply switches and options to granular control what to install or update.
+
+* `install --no_plugins` will [set an installation up without plugins](#install-ilias-without-plugins)
```
chfsx: we decided on `--no-plugin` with a "-" during JF, there are several places here which have a "_" in the subcommand.

daniwe4: Ty Fabian, i fixed it.

Services/Component/classes/Setup/class.ilComponentsSetupAgent.php
```diff
@@ -35,7 +35,12 @@ public function getArrayToConfigTransformation() : Transformation
      */
     public function getInstallObjective(Setup\Config $config = null) : Setup\Objective
     {
-        return new \ilComponentDefinitionsStoredObjective();
+        return new Setup\ObjectiveCollection(
+            'Complete objetives from Services/Component',
```
klees:
```suggestion
            'Complete objectives from Services/Component',
```

Services/Component/classes/class.ilPluginAdmin.php
```diff
@@ -243,8 +242,9 @@ protected function parsePluginPhp($plugin_php_file)
             "ilias_max_version" => $ilias_max_version,
             "responsible" => $responsible,
             "responsible_mail" => $responsible_mail,
-            "learning_progress" => (bool) $learning_progress,
-            "supports_export" => (bool) $supports_export,
+            "learning_progress" => (bool) ($learning_progress ?? false),
+            "supports_export" => (bool) ($supports_export ?? false),
+            "ignore_cli_setup" => (bool) ($ignore_cli_setup ?? false)
```
klees: Let's rename this to `supports_cli_setup` (and default to `true`). This needs to be changed in various places, then.

Services/Component/classes/class.ilPluginAdmin.php
```diff
@@ -634,4 +633,46 @@ public static function getGlobalScreenProviderCollections() : Generator
...
+            $plugin = $plugin_slot->getPluginInformationFor($name);
+            if (!is_null($plugin)) {
+                return $plugin;
+            }
+        }
+
+        return null;
+    }
+
+    protected function getPluginSlots(array &$slots, array $components, $type) : void
```
klees: Is there a reason we want to return the results this via a reference? If not, please just make this return its result as an array or even `yield` the result. References are hard to read, see the location where we call this method. We can only know that this indeed modifies `$plugin_slots` when looking into the header of the method, and there only a little `&` hints on the fact that we might actually modify the first array. Returning results via a reference is a pattern from C that has no purpose (that I can currently see) in PHP.

Services/Component/classes/class.ilPluginAdmin.php
```diff
@@ -634,4 +633,46 @@ public static function getGlobalScreenProviderCollections() : Generator
...
+    {
+        foreach ($components as $component) {
+            $component_slots = ilComponent::lookupPluginSlots($type, $component["subdir"]);
+            $component_slots = array_filter($component_slots, function ($slot) {
+                return (bool) count($slot);
```
klees: Why do we count the slots? This is a genuin question, I just want to understand what `$component_slots` contains after the call to `lookupSlots`.

Services/Component/classes/class.ilPlugin.php
```diff
@@ -54,6 +54,10 @@ abstract class ilPlugin
      * @var ProviderCollection
      */
     protected $provider_collection;
+    /**
+     * @var string
+     */
+    protected $message;
```
klees, line 60: What's this for?

Services/Component/classes/class.ilPluginRawReader.php
```diff
@@ -0,0 +1,24 @@
...
+
+class ilPluginRawReader
+{
+    const BASE_PLUGIN_PATH = 'Customizing/global/plugins';
+    const SEARCH_PATTERN = 'plugin.php';
+
+    public function getPluginNames() : ?\Iterator
+    {
+        if (!@is_dir(self::BASE_PLUGIN_PATH)) {
+            return null;
```
klees: Maybe we would want to throw an exception here? Why should the `BASE_PLUGIN_PATH` be inexistent and ain't it something we should notify our user about?

Services/Database/classes/Setup/class.ilDatabaseSetupConfig.php
```diff
@@ -36,12 +36,12 @@ class ilDatabaseSetupConfig implements Setup\Config
...
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $collation;
 
     /**
-     * @var	string
+     * @var	string|null
```
klees: I don't think that we can actually allow the user to be `null`. We need a user to connect to the database, right? Why did you change this?

Services/Database/classes/class.ilDBUpdate.php
```diff
@@ -32,6 +32,14 @@ class ilDBUpdate
...
      */
     public $updateMsg;
+    /**
+     * @var ilIniFile|null
+     */
+    protected $client_ini;
+
+    protected $custom_updates_current_version;
+    protected $custom_updates_file_version;
+    protected $custom_updates_info_read;
```
klees, line 42: Why do we want to add this?

Services/FileSystem/classes/Setup/class.ilFileSystemComponentDataDirectoryCreatedObjective.php
```diff
@@ -70,4 +72,12 @@ public function achieve(Setup\Environment $environment) : Setup\Environment
...
         return parent::achieve($environment);
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function isApplicable(Setup\Environment $environment) : bool
+    {
+        return !file_exists($this->buildPath($environment));
+    }
```
klees, line 83: How is this change related to the goal to implement that plugin-functionality?

Services/FileSystem/classes/Setup/class.ilFileSystemSetupConfig.php
```diff
@@ -17,13 +17,13 @@ public function __construct(
...
 
-    protected function normalizePath(string $p) : string
+    protected function normalizePath(string $p) : ?string
     {
         $p = preg_replace("/\\\\/", "/", $p);
         return preg_replace("%/+$%", "", $p);
     }
 
-    public function getDataDir() : string
+    public function getDataDir() : ?string
```
klees: How can the data directory be inexistent?

Services/FileSystem/classes/Setup/class.ilFileSystemSetupConfig.php
```diff
@@ -17,13 +17,13 @@ public function __construct(
         $this->data_dir = $this->normalizePath($data_dir);
     }
 
-    protected function normalizePath(string $p) : string
+    protected function normalizePath(string $p) : ?string
```
klees: Why would this ever return null?

Services/Setup/classes/Setup/class.ilCommonSetupAgent.php
```diff
@@ -12,7 +12,7 @@
  * general all this stuff here is supposed to go elsewhere once we find out
  * which service it really belongs to.
  */
-class ilSetupAgent implements Setup\Agent
+class ilCommonSetupAgent implements Setup\Agent
```
klees: How is this related to the changes for the plugins?

Services/Setup/classes/Setup/class.ilIniFilesPopulatedObjective.php
```diff
@@ -28,23 +28,23 @@ public function getPreconditions(Setup\Environment $environment) : array
...
         ];
     }
 
     public function achieve(Setup\Environment $environment) : Setup\Environment
     {
-        $path = dirname(__DIR__, 2) . "/ilias.ini.php";
+        $path = dirname(__DIR__, 4) . "/ilias.ini.php";
         $ini = new ilIniFile($path);
-        $ini->GROUPS = parse_ini_file(__DIR__ . "/../ilias.master.ini.php", true);
+        $ini->GROUPS = parse_ini_file( "setup/ilias.master.ini.php", true);
```
klees: Please keep the path relative to the location of the file.

Services/Setup/classes/Setup/class.ilIniFilesPopulatedObjective.php
```diff
@@ -28,23 +28,23 @@ public function getPreconditions(Setup\Environment $environment) : array
...
         $ini->write();
 
         $path = $this->getClientDir() . "/client.ini.php";
         $client_ini = new ilIniFile($path);
-        $client_ini->GROUPS = parse_ini_file(__DIR__ . "/../client.master.ini.php", true);
+        $client_ini->GROUPS = parse_ini_file("setup/client.master.ini.php", true);
```
klees: Please keep the path relative to the location of the file.

Services/UICore/classes/Setup/class.ilCtrlStructureReader.php
```diff
@@ -73,11 +73,16 @@ public function readStructure(
         // plugin path
         $this->plugin_path = $a_plugin_path;
 
+        $read_plugins = false;
+        if ($this->plugin_path != "" && $this->comp_prefix != "") {
+            $read_plugins = true;
+        }
+
```
klees, line 80: Why do we need this changes in `ilCtrlStructureReader`?

setup/README.md
```diff
@@ -1,21 +1,30 @@
...
 
+`install` and `update` also supply switches and options to granular control what to install or update.
+
+* `install --skip` will [set an installation up including plugins except](#install-ilias-except-plugin)
+* `install --no-plugins` will [set an installation up without plugins](#install-ilias-without-plugins)
+* `install <plugin name>` will [install a specific plugin](#install-specific-plugin)
+* `update --skip` will [update an installation including plugins except](#update-ilias-except-plugin)
+* `update --no-plugins` will [update an installation without plugins](#update-ilias-without-plugins)
+* `update <plugin name>` will [update a specific plugin](#update-specific-plugin)
+
```
klees:
```suggestion
`install` and `update` also supply switches and options for a granular control of the inclusion of plugins:

* `--skip <plugin name>` will exclude the named plugin from the command
* `--no-plugins` will exclude all plugins from the command
* `install <plugin name>` (or `update <plugin name>` respectively) will update or install the specified plugin
```

setup/README.md
```diff
@@ -73,6 +122,10 @@ in the database. Sometimes it might be necessary to refresh that information.
 Please do not invoke this function unless it is explicitly stated in update
 or patch instructions or you know what you are doing.
 
+## Exclude a plugin permanently from Command Line Setup
+
+To permanently exclude a plugin from the command line setup add the following flag to
+the plugin.php file `$ignore_cli_setup = true;`. Default value for this flag is false.
```
klees: I would not add this here, this rather belongs to some plugin-documentation. The `README.md` here is for users and they should not exclude plugins like this IMO, so this is an information for developers. Maybe @mstuder has an idea where to put this?

setup/cli.php
```diff
@@ -14,18 +14,18 @@
...
+require_once("Services/Setup/classes/Setup/class.ilSetupConfig.php");
+require_once("Services/Setup/classes/Setup/class.ilMakeInstallationAccessibleObjective.php");
+require_once("Services/Setup/classes/Setup/class.ilUseRootConfirmed.php");
+require_once("Services/Setup/classes/Setup/class.ilOwnRiskConfirmedObjective.php");
+require_once("Services/Setup/classes/Setup/class.ilOverwritesExistingInstallationConfirmed.php");
+require_once("Services/Setup/classes/Setup/class.ilIniFilesPopulatedObjective.php");
+require_once("Services/Setup/classes/Setup/class.ilIniFilesLoadedObjective.php");
+require_once("Services/Setup/classes/Setup/class.ilNICKeyRegisteredObjective.php");
+require_once("Services/Setup/classes/Setup/class.ilNICKeyStoredObjective.php");
+require_once("Services/Setup/classes/Setup/class.ilSetupConfigStoredObjective.php");
```
klees: I think we can remove these, as the autoloader should discover that classes now. They weren't included before, because the setup-folder wasn't scanned for the classmap.

src/Setup/AgentCollection.php
```diff
@@ -136,7 +136,9 @@ public function getBuildArtifactObjective() : Objective
 
     protected function getXObjective(string $which, Config $config = null) : Objective
     {
-        $this->checkConfig($config);
+        if (!is_null($config)) {
```
klees: I think we should check if we actually need a config here instead of just dropping the check if a user did not supply one. If any agent in this collection needs a config, the collection needs one as well.

src/Setup/AgentFinder.php
```diff
@@ -0,0 +1,30 @@
...
+/* Copyright (c) 2020 Danie Weise <daniel.weise@concepts-and-training.de> Extended GPL, see docs/LICENSE */
+
+declare(strict_types=1);
+
+namespace ILIAS\Setup;
+
+interface AgentFinder
+{
+    /**
+     * Collect all files exclude plugin directories ending with SetupAgent.php
```
klees: I think this is just wrong. We do not take names of files into account. Also we do not return files, the return type is inaccurate AFAIK. This should be `Agent[]`, right? Instead we are looking for classes implementing the `Agent`-interface and return instances of these.

src/Setup/AgentFinder.php
```diff
@@ -0,0 +1,30 @@
...
+     *
+     * @return array
+     */
+    public function getSystemAgents() : array;
+
+    /**
+     * Collect all files in plugin directories ending with SetupAgent.php
```
klees: Same as for `getSystemAgents`.

src/Setup/AgentFinder.php
```diff
@@ -0,0 +1,30 @@
...
+     *
+     * @return array
+     */
+    public function getPluginAgents(array $skip) : array;
+
+    /**
+     * Get a specific plugin agent.
```
klees: What would be a name here? Is it the name of the class?

src/Setup/CLI/BaseCommand.php
```diff
@@ -45,27 +40,20 @@ abstract class BaseCommand extends Command
     protected $preconditions;
 
     /**
-     * @var callable $lazy_agent must return a Setup\Agent
+     * @var callable $lazy_system_agent must return a Setup\Agent
```
klees:
```suggestion
```

src/Setup/CLI/BaseCommand.php
```diff
@@ -74,6 +62,8 @@ public function configure()
             ->addOption("yes", "y", InputOption::VALUE_NONE, "Confirm every message of the setup.");
     }
 
+    abstract protected function getRelevantAgent(InputInterface $input) : Agent;
```
klees: Please move to the other abstracts. For extra credits, add some lines for each that tells what they are supposed to do.

src/Setup/CLI/BuildArtifactsCommand.php
```diff
@@ -49,4 +50,9 @@ protected function getObjective(Agent $agent, ?Config $config) : Objective
     {
         return $agent->getBuildArtifactObjective();
     }
+
+    protected function getRelevantAgent(InputInterface $input) : Agent
+    {
+        return $this->agent_finder->buildAgentCollection($this->agent_finder->getSystemAgents());
```
klees: Why don't we build artifacts for plugins?

src/Setup/CLI/InstallCommand.php
```diff
@@ -24,6 +25,9 @@ public function configure()
     {
         parent::configure();
         $this->setDescription("Creates a fresh ILIAS installation based on the config");
+        $this->addArgument('plugin_name', InputArgument::OPTIONAL, 'Name of the plugin to install.');
```
klees:
```suggestion
        $this->addArgument('plugin-name', InputArgument::OPTIONAL, 'Name of the plugin to install.');
```

src/Setup/CLI/InstallCommand.php
```diff
@@ -36,14 +40,6 @@ protected function printOutroMessage(IOWrapper $io)
         $io->success("Installation complete. Thanks and have fun!");
     }
 
-    public function execute(InputInterface $input, OutputInterface $output)
-    {
-        // ATTENTION: This is a hack to get around the usage of the echo/exit pattern in
-        // the setup for the command line version of the setup. Do not use this.
-        define("ILIAS_SETUP_IGNORE_DB_UPDATE_STEP_MESSAGES", true);
-        return parent::execute($input, $output);
-    }
```
klees: Why did you remove this?

src/Setup/AgentFinder.php
```diff
@@ -0,0 +1,30 @@
...
+     * @return array
+     */
+    public function getSystemAgents() : array;
+
+    /**
+     * Collect all files in plugin directories ending with SetupAgent.php
+     *
+     * @return array
+     */
+    public function getPluginAgents(array $skip) : array;
```
klees: I'm not happy with that `$skip`-Parameter here. For one, you do not document what this is supposed to be. Also, this seems to be some detail on the interface (that `--skip`-parameter) that is handed all the way down to the implementation. Let's talk options.

src/Setup/CLI/ReloadControlStructureCommand.php
```diff
@@ -65,4 +65,9 @@ protected function getObjective(Agent $agent, ?Config $config) : Objective
             new \ilComponentDefinitionsStoredObjective(false)
         );
     }
+
+    protected function getRelevantAgent(InputInterface $input) : Agent
+    {
+        return $this->agent_finder->buildAgentCollection($this->agent_finder->getSystemAgents());
```
klees: This is odd. We don't need an agent here, right? `getObjective` always returns the same objective.

src/Setup/CLI/UpdateCommand.php
```diff
@@ -27,6 +28,9 @@ public function configure()
         $this
             ->setDescription("Updates an existing ILIAS installation")
             ->addOption("ignore-db-update-messages", null, InputOption::VALUE_NONE, "Ignore messages from the database update steps.");
+        $this->addArgument('plugin_name', InputArgument::OPTIONAL, 'Name of the plugin to update.');
```
klees: Please also apply this style to the rest of `configure` (instead of that fluent interface thingy).

src/Setup/CLI/UpdateCommand.php
```diff
@@ -54,16 +58,6 @@ protected function buildEnvironment(Agent $agent, ?Config $config, IOWrapper $io
...
 
-    public function execute(InputInterface $input, OutputInterface $output)
-    {
-        // ATTENTION: This is a hack to get around the usage of the echo/exit pattern in
-        // the setup for the command line version of the setup. Do not use this.
-        if ($input->hasOption("ignore-db-update-messages") && $input->getOption("ignore-db-update-messages")) {
-            define("ILIAS_SETUP_IGNORE_DB_UPDATE_STEP_MESSAGES", true);
-        }
-        return parent::execute($input, $output);
-    }
```
klees: Why did you remove this?

src/Setup/CLI/UpdateCommand.php
```diff
@@ -72,4 +66,23 @@ protected function getObjective(Agent $agent, ?Config $config) : Objective
...
+            }
+            $agents = array_merge($agents, $this->agent_finder->getPluginAgents($skip));
+        }
+
+        if ($input->hasArgument('plugin_name') && !is_null($input->getArgument('plugin_name'))) {
+            $agents = array_merge($agents, [$this->agent_finder->getPluginAgent($input->getArgument('plugin_name'))]);
+        }
+
+        return $this->agent_finder->buildAgentCollection($agents);
+    }
```
klees: Maybe this could go into a trait for `InstallCommand` and `UpdateCommand`?

src/Setup/ImplementationOfInterfaceFinder.php
```diff
@@ -9,36 +9,55 @@
...
         $this->interface = $interface;
-        $this->getAllClassNames();
     }
 
+    public function getIgnoreList() : array
+    {
+        return $this->ignore;
+    }
+
+    public function getMatchingClassNames(array $ignore = []) : \Iterator
```
klees: Please use `array $ignore = null` and add some documentation on what this parameter is doing. Using `[]` as "not set" will make it impossible to ignore nothing. Also, the interface to this class feels somehow odd now. Let's talk about how this should work. I'm not even sure that this actually finds agents in plugins, as `Customizing` is ignored.

klees (commented):
Hi @daniwe4,

thanks a lot for this impressive work. This looks functional overall, but I think there are some rough edges we should polish and also some things I just don't understand. It feels as if most of them have been introduced by the latest changes regarding the various options that the JF demanded. I think there are some things we need to discuss and added according markers. Also, this seems to contain changes that are not inherently related to the feature we want to implement but rather seem to be applications of the boyscouting rule. Is that correct?

Still, this looks good mostly and I'm optimistic that we may merge this soon.

Best regards!

klees:
Hi @mjansenDatabay,

there are two rather minor changes in your components that are about making the possibility of `null` explicit:

https://github.com/ILIAS-eLearning/ILIAS/pull/2782/files#diff-7b9bbb9e92568aa650501b87d359c219
https://github.com/ILIAS-eLearning/ILIAS/pull/2782/files#diff-c30ab9fa0625fb421b2b13f202d35130

Please have a quick look!

Thanks!

klees:
HI @chfsx,

there are some rather minor changes in your components in this PR that mostly are about making things a little more explict:

https://github.com/ILIAS-eLearning/ILIAS/pull/2782/files#diff-c9682631c4e05afc9d19a7e578971dff
https://github.com/ILIAS-eLearning/ILIAS/pull/2782/files#diff-521588d3c5034f71f766d82e42ec2809
https://github.com/ILIAS-eLearning/ILIAS/pull/2782/files#diff-d80b6ad080d0156cf243c2cbfbd32cbd
https://github.com/ILIAS-eLearning/ILIAS/pull/2782/files#diff-c97cba5216df64309adf35304e26db45
https://github.com/ILIAS-eLearning/ILIAS/pull/2782/files#diff-4dcd69097cb5d596a32ea88be5e449a7

Please have a quick look!

Thanks!

klees:
Hi @fneumann,

there are some minor changes in your component in this PR that are supposed to make things a little more explicit:

https://github.com/ILIAS-eLearning/ILIAS/pull/2782/files#diff-a8eb1b95d60d1350e9985c733a24378c
https://github.com/ILIAS-eLearning/ILIAS/pull/2782/files#diff-e31a63d94576f8fc823014630e14046d

Please have a quick look!

Thanks!

klees:
Hi @alex40724,

there are some minor changes in your components in this PR that are supposed to make things a little more explicit:

https://github.com/ILIAS-eLearning/ILIAS/pull/2782/files#diff-3e06480ec8ff9a02f838e7304c9eb91c
https://github.com/ILIAS-eLearning/ILIAS/pull/2782/files#diff-aa502a16c082092fc31e747f2026bb77

Please have a quick look!

Thanks!

klees:
Hi @mstuder,

this PR contains some refactorings and additions to the component `Services/Component` where you are the first maintainer. Please review the changes.

Thanks!

mjansenDatabay (approved):

mjansenDatabay:
LGTM:

- ilMailTemplateContextService
- ilCronManager

fneumann (approved):
Fine for me

Services/UICore/classes/Setup/class.ilCtrlStructureReader.php
alex40724: How will this be set for the case "install --no_plugins"? You won't have a plugin_path or comp_prefix, but you do not want to "read_plugins" in this case, right?

In general it would be ok for me so set a $this->read_plugins instead of passing it, even if this not settled in the constructor, which would be a better place to configure the reader instead doing it in readStructure().

alex40724 (approved):
Changes are ok in general, I only have one question.

Services/Component/classes/class.ilPluginAdmin.php
daniwe4: done

Services/Component/classes/Setup/class.ilComponentsSetupAgent.php
daniwe4: done

Services/Component/classes/class.ilPlugin.php
daniwe4: missing property declaration

Services/Component/classes/class.ilPluginAdmin.php
daniwe4: refactored

Services/Component/classes/class.ilPluginAdmin.php
daniwe4: slot holds an empty array if there is no plugin for this slot or some plugin data

Services/Component/classes/class.ilPluginRawReader.php
daniwe4: done

Services/Database/classes/Setup/class.ilDatabaseSetupConfig.php
daniwe4: you are right, change user to string only

Services/Database/classes/class.ilDBUpdate.php
daniwe4: missing properties

Services/FileSystem/classes/Setup/class.ilFileSystemComponentDataDirectoryCreatedObjective.php
daniwe4: this should be an extra commit for setup. hope we can nevertheless hold it here.

Services/FileSystem/classes/Setup/class.ilFileSystemSetupConfig.php
daniwe4: put it back

Services/FileSystem/classes/Setup/class.ilFileSystemSetupConfig.php
daniwe4: put it back

Services/Setup/classes/Setup/class.ilCommonSetupAgent.php
daniwe4: this should be an extra commit for setup. hope we can nevertheless hold it here.

Services/Setup/classes/Setup/class.ilIniFilesPopulatedObjective.php
daniwe4: done

Services/Setup/classes/Setup/class.ilIniFilesPopulatedObjective.php
daniwe4: done

Services/UICore/classes/Setup/class.ilCtrlStructureReader.php
daniwe4: done

setup/README.md
daniwe4: remove it

setup/README.md
daniwe4: done

src/Setup/AgentCollection.php
daniwe4: done

src/Setup/AgentFinder.php
daniwe4: done

src/Setup/AgentFinder.php
daniwe4: done

src/Setup/AgentFinder.php
daniwe4: done

src/Setup/AgentFinder.php
daniwe4: done

src/Setup/CLI/BaseCommand.php
daniwe4: done

setup/cli.php
daniwe4: done

src/Setup/CLI/InstallCommand.php
daniwe4: done, also for update command

src/Setup/CLI/BuildArtifactsCommand.php
daniwe4: We do it later.

src/Setup/CLI/InstallCommand.php
daniwe4: replace it

src/Setup/CLI/ReloadControlStructureCommand.php
daniwe4: We do it later

src/Setup/CLI/UpdateCommand.php
daniwe4: done

src/Setup/CLI/UpdateCommand.php
daniwe4: done

src/Setup/CLI/UpdateCommand.php
daniwe4: done

src/Setup/ImplementationOfInterfaceFinder.php
daniwe4: done
For the second part of this comment we have to do a closer look to ImplementationOfInterfaceFinder and ImplementationOfAgentFinder especially to the ignore list. Think we have do do some refactoring.

daniwe4:
Hi @alex40724, to your comment https://github.com/ILIAS-eLearning/ILIAS/pull/2782#discussion_r492537238. I think we should do this in an extra ticket. So we won't furthermore blow up this PR. If this is ok for you i'll open a new ticket.

alex40724:
> Hi @alex40724, to your comment [#2782 (comment)](https://github.com/ILIAS-eLearning/ILIAS/pull/2782#discussion_r492537238). I think we should do this in an extra ticket. So we won't furthermore blow up this PR. If this is ok for you i'll open a new ticket.

Hi @daniwe4, all fine for me, I was just thinking that "install --no_plugins" might not work as expected due to these lines. But I did not test it.

klees (approved):
Hi @daniwe4,

thanks a lot for this impressive amount of work. I'm fine to tackle the open issues in another PR to keep this manageable.

Best regards!

Services/Component/classes/Setup/class.ilComponentActivatePluginsObjective.php
```diff
@@ -0,0 +1,150 @@
...
+
+    protected function initEnvironment(Setup\Environment $environment) : ILIAS\DI\Container
+    {
+        $db = $environment->getResource(Setup\Environment::RESOURCE_DATABASE);
+        $plugin_admin = $environment->getResource(Setup\Environment::RESOURCE_PLUGIN_ADMIN);
+        $ini = $environment->getResource(Setup\Environment::RESOURCE_ILIAS_INI);
+        $client_ini = $environment->getResource(Setup\Environment::RESOURCE_CLIENT_INI);
+
+
+        // ATTENTION: This is a total abomination. It only exists to allow various
```
chfsx, line 111: @klees Thanks for the note here, this is absolutely correct and definitely something we need to tackle in the future.

Services/Component/classes/Setup/class.ilComponentActivatePluginsObjective.php
```diff
@@ -0,0 +1,150 @@
...
+
+        return $environment;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function isApplicable(Setup\Environment $environment) : bool
+    {
+        $ORIG_DIC = $this->initEnvironment($environment);
```
chfsx, line 90: here i was confused at first because i did not expect the original DIC to be returned. to name the method correctly would be a bit cumbersome like `initTemporaryEnvironmentAndPreserveOrdiginal`... Let's leave it like this ;-)

Services/Component/classes/Setup/class.ilComponentInstallPluginObjective.php
```diff
@@ -0,0 +1,150 @@
...
+        if (is_null($plugin) || !$plugin['supports_cli_setup']) {
+            return false;
+        }
+
+        $GLOBALS["DIC"] = $ORIG_DIC;
+
+        return $plugin['must_install'];
+    }
+
+    protected function initEnvironment(Setup\Environment $environment) : ILIAS\DI\Container
```
chfsx, line 103: Could we outsource this, since it exists at least three times? For example, as a trait?

Services/Component/classes/class.ilPluginAdmin.php
```diff
@@ -243,8 +242,9 @@ protected function parsePluginPhp($plugin_php_file)
             "ilias_max_version" => $ilias_max_version,
             "responsible" => $responsible,
             "responsible_mail" => $responsible_mail,
-            "learning_progress" => (bool) $learning_progress,
-            "supports_export" => (bool) $supports_export,
+            "learning_progress" => (bool) ($learning_progress ?? false),
+            "supports_export" => (bool) ($supports_export ?? false),
+            "supports_cli_setup" => (bool) ($supports_cli_setup ?? true)
```
chfsx, line 247: I know we had discussed this and probably the JF decision is the same. but i currently tend to do opt-in instead of opt-out. as default 'false'.

i am afraid that otherwise a lot of plugins will appear relatively quickly where this is not possible and so updates to ILIAS 7 can quickly become a problem. developers of plugins could first check with opt-in that everything works as desired before this is offered. what do you think?

Services/Component/classes/Setup/class.ilComponentInstallPluginObjective.php
klees: I think we could, but I don't think we should =)

I mean: From my point of view this is something we want to get rid of by using dependency injection a lot more. This snippet shows very much why we would want to do so. I would not bother to "improve" here by abstracting this away into some trait or something else.

Services/Component/classes/class.ilPluginAdmin.php
klees: I disagree. Maybe we would want to discuss this on the JF with other developers. For one, I wonder why and how many plugins would really not be installable with this mechanism. I understood that you are using GUIs in the `afterInstallation`-hook of the plugins, but is that really common for many plugins and are there other failure modes? After all, the CLI setup uses the same mechanism as the GUI for plugins. Second, developers need to test the plugins and bump the version in the `plugin.php` anyway. There could be some problem regarding the general information about updates for plugin developers. I don't think, we have a strategy here, right? Finally I think this is just one step on the way to move plugins and the other components closer together, at least on a technical base. So this (defaulting this to `true`) is something that will happen eventually anyway. I don't see enough reasons to defer this.

Since this is something we can change very cheaply, I think we could move on with this PR anyway (right?) and, if you want, bring this up on the JF. I won't attend on 20-11-09, but I'll be there the one afterwards.

chfsx:
On behalf of @mstuder : OK for Services/Component, thanks a lot!

chfsx (approved):

Services/Component/classes/class.ilPluginAdmin.php
iszmais: @klees the here given fallback would be also very usefull ot apply to ILIAS6 to prevent notices inside the log on unaccessable variables.

PR #2787, opened 2020-07-06
Missing plural in german Lang-File

kergomard:
Small language error. Would need to be fixed in all supported Versions.


matthiaskunkel:
Thanks for the PR. I will cherry-pick it into release_6 as well.

PR #2789, opened 2020-07-07
fix/28550/6/hostfix_update

mstuder:
As discussed yesterday, could you please take a quick look at it?

### DIFF
--- a/Services/Database/classes/class.ilDBUpdate.php
+++ b/Services/Database/classes/class.ilDBUpdate.php
@@ -431,12 +431,12 @@ class ilDBUpdate
     /**
      * apply an update
      *
-     * @param int nr number what patch to apply
+     * @param int nr number what patch to apply (Reference: Patch for https://mantis.ilias.de/view.php?id=28550)
      *
      * @return bool
      * @access private
      */
-    public function applyUpdateNr($nr, $hotfix = false, $custom_update = false)
+    public function applyUpdateNr(&$nr, $hotfix = false, $custom_update = false)
     {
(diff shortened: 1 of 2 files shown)


mjansenDatabay (approved):
LGTM!

I also tested this and it worked like a charm.

mstuder:
Thx Michael!!

Services/Database/classes/class.ilDBUpdate.php
```diff
@@ -431,12 +431,12 @@ public function loadXMLInfo()
...
      * apply an update
      *
-     * @param int nr number what patch to apply
+     * @param int nr number what patch to apply (Reference: Patch for https://mantis.ilias.de/view.php?id=28550)
      *
      * @return bool
      * @access private
      */
-    public function applyUpdateNr($nr, $hotfix = false, $custom_update = false)
+    public function applyUpdateNr(&$nr, $hotfix = false, $custom_update = false)
```
klees, line 439: Hi @mstuder,

why did you introduce that reference on `$nr` here?

Best regards!

mjansenDatabay: Because if step 1 of the file is executed, it should immediatly switch to the next `$nr` determined by the changes in https://github.com/ILIAS-eLearning/ILIAS/pull/2789/files#diff-539334eb8c92e96b15929c1f4086297aR7 , so the `for` loop in ` \ilDBUpdate::applyHotfix` will continue with the next step (which could be `> 2`).

PR #2790, opened 2020-07-07
Changed ILIAS version number

tfamula:
Version number should be changed from 6.0 to 7.0 in `inc.ilias_version.php` for trunk


matthiaskunkel:
Thanks for the PR!

PR #2801, opened 2020-07-17
UI/MainMenu: 28608 split preparation for ui presentation and filter a…

shecken:
…fter only prepairing

The closure to filter items by visibility was called only once, so the repository link was never filtered by the defined closure within the preparation for ui presentation.

So i splitted the preparation function in three steps.
1. Preparation: Still configures the items with ref id and new isVisible Closure.
2. Cleanup: Removes all parent items with empty children (Same as now)
3. Sorting: Sorts the items as it was (Same as now)

Complementary i moved the filter action after the preparation to execute all correctly defined Filter closures.


shecken:
Only moving filterItemsByVisibilty after the old preparation didn't work. I think this was because the cleanup of the items.

chfsx:
Hi @shecken 
Thanks you very much for the PR, I'll have a look at it asap!

chfsx (approved):
Thanks a lot for investigating and fixing this!

PR #2802, opened 2020-07-20
Components/Plugins: Fix version check of plugins

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=28642


mstuder:
THX Michael!

PR #2747, opened 2020-06-02
CSS Doc: Updating guideline on compiling by using lessc

Amstutz:
This is a follow up on #2734 trying to handle the switching of css files due to different lessc versions. I really hope this will be not as frequent in future as they currently seem to appear.

PR #2803, opened 2020-07-20
UI/Footer: Add interface changes to support shy buttons with modals

mjansenDatabay:
This PR suggests interface changes for the `UI Footer` component to support `Shy Buttons` with `Modals` being opened when a button is clicked.

Feature Request: https://docu.ilias.de/goto_docu_wiki_wpage_5120_1357.html

Alternatives:
* Enable `Shy Buttons` to receive a `Modal`, which is rendered when the shy button is rendered:
  1. How many optional `Modal` elements should a shy button support? 1, more than 1?
  2. What about the binding mechanisms of `Button` and `Modal`, which is currently always solved/implemented in the consuming code. `Modal` elements support multiple signals (`Show`, `Close`), so an implicit binding when adding the `Modal` to the `Button` is so easy and maybe not desired by the consumer.
  3. Should the passed `Modal` element(s) be rendered if the `Button` is rendered? I am not sure if rendering the modal in the `Button` renderer is a good idea, because `Modal` elements sometimes include `<form>` elements, and `Buttons` are sometimes rendered within `<form>` elements as well. Nested form elements are not supported by HTML standard. This is already/currently an issue with `Modal` elements because a correct rendering (to avoid forms in forms) totally relies on the consuming code.

### DIFF
--- a/src/UI/examples/Modal/RoundTrip/show_a_modal_which_cannot_be_closed_with_the_keyboard.php
+++ b/src/UI/examples/Modal/RoundTrip/show_a_modal_which_cannot_be_closed_with_the_keyboard.php
@@ -7,7 +7,7 @@ function show_a_modal_which_cannot_be_closed_with_the_keyboard()
 
     $modal = $factory->modal()->roundtrip('My Modal 1', $factory->legacy('You cannot close this modal with the ESC key'));
     $modal = $modal->withCloseWithKeyboard(false);
-    $button1 = $factory->button()->standard('Open Modal 1', '#')
+    $button1 = $factory->button()->shy('Open Modal 1', '#')
         ->withOnClick($modal->getShowSignal());
 
     return $renderer->render([$button1, $modal]);
--- a/src/UI/examples/Modal/RoundTrip/show_a_modal_which_cannot_be_closed_with_the_keyboard.php
+++ b/src/UI/examples/Modal/RoundTrip/show_a_modal_which_cannot_be_closed_with_the_keyboard.php
@@ -7,7 +7,7 @@ function show_a_modal_which_cannot_be_closed_with_the_keyboard()
 
     $modal = $factory->modal()->roundtrip('My Modal 1', $factory->legacy('You cannot close this modal with the ESC key'));
     $modal = $modal->withCloseWithKeyboard(false);
-    $button1 = $factory->button()->shy('Open Modal 1', '#')
+    $button1 = $factory->button()->standard('Open Modal 1', '#')
         ->withOnClick($modal->getShowSignal());
 
     return $renderer->render([$button1, $modal]);
--- a/src/UI/Component/MainControls/Factory.php
+++ b/src/UI/Component/MainControls/Factory.php
@@ -312,12 +312,13 @@ interface Factory
      * ---
      * description:
      *   purpose: >
-     *     The Footer is a unique page section to accommodate links and shy buttons that
-     *     are not being used on a regular basis, such as links to the pages's
-     *     imprint or a privacy policy document.
+     *     The Footer is a unique page section to accommodate links and shy buttons
+     *     triggering Round Trip Modals that are not being used on a regular basis,
+     *     such as links to the pages's imprint or a privacy policy document.
      *
      *   composition: >
-     *     The Footer is composed of a list of links or shy buttons and an optional text-part.
+     *     The Footer is composed of a list of Links or Shy Buttons triggering
+     *     Round Trip Modals and an optional text-part.
      *
      * context:
      *   - The Footer is used with the Standard Page.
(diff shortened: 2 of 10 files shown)


matthiaskunkel:
Jour Fixe, 20 JUL 2020 : We highly appreciate this suggestion and accept the PR for trunk. Timon will make a review of the PR and merge it when accepted.

Amstutz:
Note: @alex40724 brought up today in the discussion the very valid point that it seems somewhat strange from a semantic point of view to attach modals to the footer that are triggered by shy buttons. He points out, that the modals could be understood as part of the shy buttons as well. I will take that into consideration when performing the review.

Amstutz:
Hi @mjansenDatabay 

Sry that this hit me somewhat unprepared in the JF. Note it is possible to fast-line urgent PRs, however @klees or @Amstutz we need to approve the PR or at least give green light for JF, especially if opened on such short notice.

However, thx very much for this necessary addition to the footer, thx for the PR!

IMO the point of @alex40724 is valid, this is not a request for change, but a proposal meant to be discussed. What do you think of the following, move the binding to the renderer of the footer, which would/could imply:
- [x] Purpose description: Change  "The Footer is a unique page section to accommodate links and shy buttons that" to " The Footer is a unique page section to accommodate Links and Shy Buttons triggering Round Trip Modals that..."
- [x] Purpose description: "The Footer is composed of a list of links or shy buttons and an optional text-part." to "The Footer is composed of a list of Links or Shy Buttons triggering Round Trip Modals and an optional text-part."
- [x] Shy Buttons: Remove the Shy Buttons PHP Doc from the factory and getLinks method on the interface.
- [x] withModals Method: Change to withAdditionalModalAndTrigger(RoundTripModal $modal, ShyButton $button). In the implementation bind the onclick of the button implicit to showModal. If the buttons has additional bindings (for closing etc.) made by the consumer, it should work nevertheless. 
- [x] getModals Method: Return an array containing doubles of the triggering button and the modals.

What do you think of that @mjansenDatabay ?

Thx for the feedback
@Amstutz

mjansenDatabay:
Hi @Amstutz 

thx for you detailed feedback. I am totally fine with you suggestions. Maybe we should wait so @klees can also provide some feedback. Afterwards I could apply the changes add the `JourFixe` label again.

Best regards,
@mjansenDatabay

Amstutz:
@mjansenDatabay if you are fine with waiting to august, that would be better. Afaik @klees is out of office for the rest of juli. If it this is urgent and you would like to start working on the FR related, I would be fine with merging as soon as revised, since this is not an earth shaking change...

mjansenDatabay:
@Amstutz I just pushed the changes

Amstutz:
Hi @mjansenDatabay 

Thx a lot. The interface is ready to be shipped, for the implementation, the only thing left to do is IMO:

- [x] Unit Tests: Add a test for Adding, Getting and rendering Shy Buttons Triggering Modals.

thx.

mjansenDatabay:
@Amstutz Thansk, of course I will add the missing tests.

mjansenDatabay:
@Amstutz I added the tests and an example.

tfamula:
Just my two cents: I noticed the changes in `delos.css` which are not related to this PR. Please keep in mind the decision ([here](https://github.com/ILIAS-eLearning/ILIAS/pull/2747)) we made a few weeks ago to deal with this case. Please check if you use the latest lessc version and if so, please inform @Amstutz and only push the changes related to your PR.

src/UI/examples/Modal/RoundTrip/show_a_modal_which_cannot_be_closed_with_the_keyboard.php
```diff
@@ -7,7 +7,7 @@ function show_a_modal_which_cannot_be_closed_with_the_keyboard()
 
     $modal = $factory->modal()->roundtrip('My Modal 1', $factory->legacy('You cannot close this modal with the ESC key'));
     $modal = $modal->withCloseWithKeyboard(false);
-    $button1 = $factory->button()->standard('Open Modal 1', '#')
+    $button1 = $factory->button()->shy('Open Modal 1', '#')
```
tfamula: Is this change really intended? Because it affects an example from the Modals.

mjansenDatabay: Good catch, I will remove this.

mjansenDatabay:
@tfamula The CSS changes are related, otherwise shy buttons do not have the same font color as the links. I will check my `lessc` version and maybe re-transpile again.

tfamula:
@mjansenDatabay You are totally right regarding the changes for the shy buttons [here](https://github.com/ILIAS-eLearning/ILIAS/pull/2803/files#diff-b7e54a2a697227a83d63170b3971b007L9891-R9806). I just mean all the other changes in `delos.css`.

mjansenDatabay:
@tfamula Fixed ;-)

tfamula:
Oh yes, this looks much better! :sunglasses:

Amstutz:
hi 

@tfamula thank you for your catch concerning examples and CSS. I am sure I would have missed, kudos. Love to see multiple people contributing here!

@mjansenDatabay merging to trunk since all my points are taken care of, so you can continue the work on the linked feature.

PR #2805, opened 2020-07-21
BT-28034: Fixed sorting of languages in Loginscreen Editor

M-Barz:
Also removed sorting of actions because the only action is "edit"


smeyer-ilias:
Thanks!

PR #2806, opened 2020-07-21
Awareness: Don't determine users in certain cases

mjansenDatabay:
This PR "filters" the user collections if the actor "does not match the expectation". 

If approved, please merge to `release_5-3`, `release_5-4` and `release_6`.


mjansenDatabay:
@corro FYI

mjansenDatabay:
Already fixed by @alex40724

PR #2812, opened 2020-07-23
Improvement/headline schema

playamongthestars:
According to Mantis issue https://mantis.ilias.de/view.php?id=27327 this pull request applies the headline schema changes (see: https://docu.ilias.de/goto_docu_pg_118822_459.html) **as accepted** by the JourFix (see: https://docu.ilias.de/goto_docu_wiki_wpage_6475_1357.html)

As already discussed and outlined in the JourFixe discussions, this headline schema does **not** claim to be semantically correct, but is instead a more 'technical' approach which should be a real improvement in a first of probably multiple future steps to achieve a impeccable headline schema.

These changes do **not** cover content created by users (e.g. with the page editor)

If accepted, we will merge this into `trunk` first (see #2820) and leave this PR open for a while, so we have the option to fix `visual issues` (which might potentially arise, but we do not expect any).

# Changes for Headline Schema:
General changes depending to ilheader 
* templates/default/delos.less

@alex40724 
Accordions:
* Services/Accordion/templates/default/tpl.accordion.html

Block: 
* templates/default/less/Services/Block/delos.less

Blogs:
* Modules/Blog/templates/default/tpl.blog_list.html 

Container:
* Services/Container/templates/default/tpl.container_list_item.html 
* Services/Container/templates/default/tpl.container_list_block.html

COPage:
* Services/COPage/css/content.css
* Services/COPage/css/content.less
* Services/COPage/templates/default/tpl.glossary_term_output.html
* Services/COPage/xsl/page.xsl 

Exercise: 
* Modules/Exercise/templates/default/tpl.assignment_head.html

Form:
* Services/Form/templates/default/tpl.property_form.html 

Glossary:
* Modules/Glossary/templates/default/tpl.glossary_definition_delete.html
* Modules/Glossary/templates/default/tpl.glossary_definition_list.html

LearningModule:
* Modules/LearningModule/templates/default/tpl.glossary_term_output.html

InfoSreen:
* Services/InfoScreen/templates/default/tpl.infoscreen.html
* templates/default/less/Services/InfoScreen/delos.less

Notes:
* Services/Notes/templates/default/tpl.notes_list.html 
* Services/Notes/templates/default/tpl.notes_list.html

Object:
* Services/Object/templates/default/tpl.creation_acc_head.html

UIComponent:

* Services/UIComponent/GroupedList/templates/default/tpl.grouped_list.html
* Services/UIComponent/Panel/templates/default/tpl.panel.html

---------------

@Amstutz @klees 

Card:
* src/UI/templates/default/Card/tpl.card.html

Divider:
* src/UI/templates/default/Divider/divider.css  
* src/UI/templates/default/Divider/divider.less 
* src/UI/templates/default/Divider/tpl.horizontal.html 

Dropdown:
* src/UI/templates/default/Dropdown/dropdown.less 

Input: 
* src/UI/templates/default/Input/tpl.section.html

Item:
* src/UI/templates/default/Item/tpl.item_standard.html
* src/UI/templates/default/Item/tpl.item_notification.html 
* src/UI/templates/default/Item/tpl.group.html
* src/UI/templates/default/Item/item.less 

Panel: 
* src/UI/templates/default/Panel/panel.less
* src/UI/templates/default/Panel/tpl.listing_standard.html
* src/UI/templates/default/Panel/tpl.secondary.html
* src/UI/templates/default/Panel/tpl.standard.html
* src/UI/templates/default/Panel/tpl.sub.html

---------------

@bheyser  @mbecker-databay 
Test:
* Modules/Test/templates/default/tpl.il_as_tst_correct_solution_output.html
* Modules/Test/templates/default/tpl.il_as_tst_output.html
* Modules/TestQuestionPool/templates/default/tpl.il_as_qpl_content.html
* Modules/TestQuestionPool/templates/default/tpl.qpl_question_preview.html 
* Modules/Test/templates/default/tpl.il_as_tst_list_of_questions_short.html

### DIFF
--- a/Modules/Blog/templates/default/tpl.blog_list.html
+++ b/Modules/Blog/templates/default/tpl.blog_list.html
@@ -3,7 +3,7 @@
 <br class="ilClearFloat" />
 <!-- END prtf_edit_bl -->
 <!-- BEGIN month_bl -->
-<h3 class="ilHeader">{TXT_CURRENT_MONTH}</h3>
+<h2 class="ilHeader">{TXT_CURRENT_MONTH}</h2>
 <div class="ilBlogList">
 	<!-- BEGIN posting -->
 	<div class="ilBlogListItem{DRAFT_CLASS}">
--- a/Modules/DataCollection/classes/class.ilDataCollectionGlobalTemplate.php
+++ b/Modules/DataCollection/classes/class.ilDataCollectionGlobalTemplate.php
@@ -775,7 +775,7 @@ class ilDataCollectionGlobalTemplate implements ilGlobalTemplateInterface
      *
      * Will override the header_page_title.
      */
-    public function setTitle($a_title)
+    public function setTitle($a_title, $hidden = false)
     {
         $this->title = $a_title;
         $this->header_page_title = $a_title;
--- a/Modules/Exercise/templates/default/tpl.assignment_head.html
+++ b/Modules/Exercise/templates/default/tpl.assignment_head.html
@@ -1,6 +1,6 @@
 <div class="ilExcAssignmentHead">
 	<img src="{IMG_STATUS}" alt="{ALT_STATUS}" title="{ALT_STATUS}" />
-	<h3 class="ilAssignmentHeader">{TITLE}</h3>
+	<h2 class="ilAssignmentHeader">{TITLE}</h2>
 	<div>
 	<!-- BEGIN prop --><span>{PROP}: {PROP_VAL}</span>&nbsp;&nbsp;<!-- END prop -->
 	</div>
--- a/Modules/Glossary/templates/default/tpl.glossary_definition_delete.html
+++ b/Modules/Glossary/templates/default/tpl.glossary_definition_delete.html
@@ -1,6 +1,6 @@
 <table class="ilc_Page" width="100%" cellspacing="0" cellpadding="0"><tr><td class="ilc_Page">
 <div class="ilc_Page">
-<div class="ilc_PageTitle">{TXT_TERM}</div>
+    <div class="ilc_PageTitle"><h2>{TXT_TERM}</h2></div>
 <!-- BEGIN definition -->
 <!-- BEGIN definition_header -->
 <div class="ilc_DefinitionHeader">{TXT_DEFINITION}</div>
--- a/Modules/Glossary/templates/default/tpl.glossary_definition_list.html
+++ b/Modules/Glossary/templates/default/tpl.glossary_definition_list.html
@@ -1,5 +1,5 @@
 <div id="ilGloContent" class="ilc_page_cont_PageContainer">
-<h1 class="ilc_page_title_PageTitle">{TXT_TERM}</h1>
+<h2 class="ilc_page_title_PageTitle">{TXT_TERM}</h2>
 	<!-- BEGIN toc -->
 	<div class="ilc_page_toc_PageTOC">
 		<!-- BEGIN toc_item -->
--- a/Modules/LearningModule/templates/default/tpl.glossary_term_output.html
+++ b/Modules/LearningModule/templates/default/tpl.glossary_term_output.html
@@ -23,7 +23,7 @@
 <div class="ilInvisibleBorder">
 	<div class="xxxilc_page_frame_PageFrame">
 		<div class="xxxilc_page_cont_PageContainer">
-			<h1 class="ilc_page_title_PageTitle">{TXT_TERM}</h1>
+			<h2 class="ilc_page_title_PageTitle">{TXT_TERM}</h2>
 			<!-- BEGIN definition -->
 			<!-- BEGIN definition_header -->
 			<div class="ilc_DefinitionHeader">{TXT_DEFINITION}</div>
--- a/Modules/Test/templates/default/tpl.il_as_tst_correct_solution_output.html
+++ b/Modules/Test/templates/default/tpl.il_as_tst_correct_solution_output.html
@@ -5,7 +5,7 @@
 				<table border="0" cellpadding="0" cellspacing="0" width="100%">
 					<tr>
 						<td>
-							<h1 class="ilc_PageTitle">{QUESTION_TITLE}</h1>
+							<h2 class="ilc_PageTitle">{QUESTION_TITLE}</h2>
 <!-- BEGIN objectives -->
 							<div>{OBJECTIVES}</div>
 <!-- END objectives -->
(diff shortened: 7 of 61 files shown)


mjansenDatabay (changes requested):
Thx @nmatuschek

Please ...

- [x] ... fix the failing unit tests
- [x] ... add the code to keep the h1 heading even if the object title is disabled in container objects, according to https://mantis.ilias.de/view.php?id=27327#c68462
- [x]  ... add the headline schema (linked in your description above) to https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/accessibility.md#headings-and-labels-and-language

alex40724 (changes requested):
@nmatuschek Wow, thanks for the great work! Everything in my code is ok, except the page editor content related ones.

Please revert the changes in ...
Services/COPage/xsl/page.xsl
Services/COPage/css/content.less
Services/COPage/css/content.css
... for now.

The reason is, that the page editor allows to add h1/h2/h3 headlines which will mess up the structure anyway. We need a separate migration strategy for the content, since all existing content styles refer to these headlines.

mjansenDatabay:
@alex40724 Are you sure about the changes in `Services/COPage/[css/xsl]`? This only changes the page header from`<h1>` to `<h2>` (because there MUST be no concurrent `<h1>`), the changed LESS/CSS should only affect this headline. Does this really affect manually created headlines/user content?

alex40724:
> @alex40724 Are you sure about the changes in `Services/COPage/[css/xsl]`? This only changes the page header from`<h1>` to `<h2>` (because there MUST be no concurrent `<h1>`), the changed LESS/CSS should only affect this headline. Does this really affect manually created headlines/user content?

No, it does not affect the manually created headlines, which is part of the problem. The manually created headlines are h1,h2,h3. And it does not make sense, that they follow a h2 with the page title. So we need a concept for the whole page content area. We could decide to move the content h1,h2,h3 to h3,h4,h5 for example. But this would need migration of the content styles and we might rename the heading classes as well, since it may be strange that Heading1 renders as h3.

Or is there any existing strategy for the page editor content headings already?

mjansenDatabay:
@alex40724 Okay, I understand. No, there is no existing strategy regarding the user generated content. 

I reverted the changes you mentioned above.

mjansenDatabay (approved):

alex40724 (approved):
Thanks @mjansenDatabay.

Btw. I discussed the heading issues in the content area with the Page Editor Revision Group. Most probable there will be a solution for this. Currently the preference is to not match the class labels "Heading1/2/3" to the h1/h2/h3 anymore. But this needs a final decision, so I would take it from there.

klees (approved):
Dear @nmatuschek, dear @mjansenDatabay,

this is what we already talked about some time ago and agreed on as a temporary fix, so **we do not object to the changes in the UI-framework**. Also, the general change wasn't your idea and you are only executing it, so you are not the main addressee of this comment.

Still we feel the requirement to clearly and unambiguously point out our discontent with the proposed solution. In short: **We do not see that the changes are in fact a solution to the problem, as they do not attack the root cause**. They are merely some patch, that might fix the symptoms of the problem for the moment, but will break soon. We know that you two are well aware of this fact =)

The general problem to be solved is, that we need to present an unbroken hierarchy of headlines, where <hx> is only nested into <h(x-1)> but not <h(x-2)> (and so forth). The approach implemented in this PR tries to accomplish this via a semantic schema that provides a list of headline levels for various components in ILIAS. This totally neglects the idea of reuse of components in different contexts and will therefore break when the same component is used in two different places.

For example: The schema assigns `<h2>` to headlines in forms, presumably because they are currently used in the content section only (having a `<h1>`). Once forms will be used in some other section of the page, e.g. the slates of the Main Bar or some modal, this won't work anymore when that section is missing an upper `<h1>`. This problem isn't theoretical at all, considering the fact that the Page Editor Revision currently considers using forms in its tool slate.

Every next instance of that problem might be fixable by another ad-hoc solution (e.g. adding another headline somewhere), but we will need to adjust the semantic schema accordingly and, at some point, will inexorably run into a conflict we cannot solve via another headline or adjusting the schema anymore. The last option then will be to just remove the headline for the conflicting component, which is a questionable win for accessibility considering that headlines provide a fast way to navigate the page. This also isn't theoretical at all, this PR already removes headlines for some UI components.

We further predict that codifying the insufficient schema in a guideline (as proposed here) will lead to the behaviour to just not use h-tags anymore, to prevent potential fallout from the guidelines. Please also note, that the change in the guideline is in an odd place (in the "Cheat Sheet") and also needs approval by the JF, so we propose to remove it here and open a single PR that just contains the guideline changes.

So as a wrap up, we can determine that **a semantic headline schema is no viable way to provide an unbroken hierarchy of headlines for a flexible UI system** like ILIAS has and requires. This assertion is also backed by the the [interview with an external accessibility expert](https://docu.ilias.de/goto_docu_xpdl_8771.html) that was hired by the SIG Accessibility and the fact, that the semantic solution already breaks with pages build in the Page Editor.

This, however, does not mean that there is no solution to the problem of an unbroken hierarchy of headlines in our UI system. It is just not as simple as writing down an incomplete semantic list, but rather has a technical component that will need some time and resources to be build. Just like @alex40724 proposes for the Page Editor, the headline nesting level will need to adjust dynamically according to the placement of given component. The UI framework is well prepared to implement such a mechanism since it deals with an abstract component tree most of the time and renders to HTML very late. This will allow us to assign correct nesting levels to headlines to present an unbroken hierarchy and have reusable components at the same time.

**We ask anyone who is interested in a lasting solution for an unbroken headline schema to contact us to figure out how we could get there.**

Best regards!
@Amstutz and @klees

Amstutz (approved):
Thx @nmatuschek and @mjansenDatabay.

mjansenDatabay:
Thx for your review and remarks @Amstutz / @klees 

We totally agree with (almost all of) your points. Furthermore I am of course interested in future discussions about a sustainable mechanism.

Regarding the `Cheat Sheet`. AFAIK I presented the identical schema https://docu.ilias.de/goto_docu_pg_118822_459.html when we discussed this on: https://docu.ilias.de/goto.php?target=wiki_1357_JourFixe-2020-06-22 . So I am not sure if we have to decide upon it once again.

alex40724:
@klees I do not see big issues with the page editor.

Having h2 without leading h1 should not be a problem (forms in the slates in the editor), see e.g. Example 3 at https://www.w3.org/WAI/tutorials/page-structure/headings/.

For the page editor content, starting with h2 should be ok in all cases. But in general this is a special case, since the editor does not enforce a valid heading structure within the content (h3 could follow h1). The editor could do so in the future though.

PR #2813, opened 2020-07-23
BT-28641: Removed updateTranslation for cat type in ilSoapObjectAdministration, because it was outdated

M-Barz:
After testing the functionality with SOAP while removing the call for updateTranslation and still having the changed title in the Multilanguage subtab, I do believe that it's outdated.
I also could not find a translate call or similar in the save function of ilObjCategoryGui, which further makes me believe that it's outdated at this point.


smeyer-ilias:
Thanks Marvin!

PR #2814, opened 2020-07-24
#28667: Fix background download for normal users

corro:
See https://mantis.ilias.de/view.php?id=28667


chfsx:
Thanks @corro ! Fixed for 5.4 using your commit, in ILIAS 6 this is already handled a bit different

PR #2815, opened 2020-07-29
UI/Layout/Footer: Remove z-index from footer

mjansenDatabay:
This PR removes the `z-index` property (currently set to `z-index: 2`) from the `<footer>` HTML element.

Reasons: Because of the set `z-index`, modals rendered in the footer area (see #2803) are hidden **behind** the `Modal Backdrop`.

I did not notice any negative site effect after removing the `z-index` (neither in the desktop view, nor in the mobile view). I also talked to @nhaagen and he has no objections as well.


klees (approved):

Amstutz:
Thx a lot

PR #2816, opened 2020-07-30
28566 -LearningSequence: Fixed bug regarding changes on position if more objects exist than hits/page is set

lukastocker:
https://mantis.ilias.de/view.php?id=28566

### DIFF
--- a/Modules/LearningSequence/classes/Content/class.ilObjLearningSequenceContentTableGUI.php
+++ b/Modules/LearningSequence/classes/Content/class.ilObjLearningSequenceContentTableGUI.php
@@ -43,6 +43,8 @@ class ilObjLearningSequenceContentTableGUI extends ilTable2GUI
         $this->addColumn($this->lng->txt("table_online"));
         $this->addColumn($this->lng->txt("table_may_proceed"));
         $this->addColumn($this->lng->txt("table_actions"));
+
+        $this->setLimit(9999);
     }
 
     protected function fillRow($set)


klees (changes requested):
Hi @lukastocker,

thanks a lot for looking into this. We should not change Services/Container/classes/class.ilContainerSorting.php. This is a class that is used in quite a lot different places, so we could break stuff somewhere. Also I do not understand the purpose of the change. Why do we need to change that class to fix the problem with pagination? Please contact me via Skype to discuss that change.

Best regards!

lukastocker:
> Hi @lukastocker,
> 
> thanks a lot for looking into this. We should not change Services/Container/classes/class.ilContainerSorting.php. This is a class that is used in quite a lot different places, so we could break stuff somewhere. Also I do not understand the purpose of the change. Why do we need to change that class to fix the problem with pagination? Please contact me via Skype to discuss that change.
> 
> Best regards!

Thank you for your feedback. As discussed I have included a query to dismiss the pagination output for the module LearningSequence.

Best regards!

klees (changes requested):
Hi @lukastocker,

thanks a lot for the changed PR. Currently I do not understand how this is supposed to work at all, to be honest. `checkTemplate` just outputs the provided input verbatim, so neither will [this line](https://github.com/ILIAS-eLearning/ILIAS/pull/2816/files#diff-0098be4dfda62a75688745b9dbb5a63aR38) do anything interesting, nor will [that else](https://github.com/ILIAS-eLearning/ILIAS/pull/2816/files#diff-f922bd33a858bb04250c51183fcefd69R245) be reached ever.

In general we should be hesitant to change central classes (like `ilTable2GUI`) when we have problems in the periphery (like `ilObjLearningSequenceGUI`). I recently found an blog post that demands to [prefer to push fixes upstream instead of working around problems downstream](http://www.haskellforall.com/2020/07/the-golden-rule-of-software-quality.html) which to me seems to be solid advice. But this only applies if the problem really is an upstream problem, which I doubt here.

For this case there should be a simple way to tell the table to use no pagination by using some method on the table object, without needing any modification of some general class. Please have another look here and try to find the correct method on the table class.

As always, feel free to get back to me with anything =)

Best regards!

lukastocker:
> Hi @lukastocker,
> 
> thanks a lot for the changed PR. Currently I do not understand how this is supposed to work at all, to be honest. `checkTemplate` just outputs the provided input verbatim, so neither will [this line](https://github.com/ILIAS-eLearning/ILIAS/pull/2816/files#diff-0098be4dfda62a75688745b9dbb5a63aR38) do anything interesting, nor will [that else](https://github.com/ILIAS-eLearning/ILIAS/pull/2816/files#diff-f922bd33a858bb04250c51183fcefd69R245) be reached ever.
> 
> In general we should be hesitant to change central classes (like `ilTable2GUI`) when we have problems in the periphery (like `ilObjLearningSequenceGUI`). I recently found an blog post that demands to [prefer to push fixes upstream instead of working around problems downstream](http://www.haskellforall.com/2020/07/the-golden-rule-of-software-quality.html) which to me seems to be solid advice. But this only applies if the problem really is an upstream problem, which I doubt here.
> 
> For this case there should be a simple way to tell the table to use no pagination by using some method on the table object, without needing any modification of some general class. Please have another look here and try to find the correct method on the table class.
> 
> As always, feel free to get back to me with anything =)
> 
> Best regards!

Hi @klees , 

thank you for your detailed feedback. I set off the pagination via setting the limit (a method from the ilTable2GUI) to 9999 in the ilObjLearningSequenceContentTableGUI. I hope it's what you were asking for. Let me know what you think about the new solution.

Best regards

klees:
I have cherry-picked this to trunk and release_6 as well.

PR #2818, opened 2020-07-31
28380 - Possible fix for File- and User-Links in LearningModule

rsheer:
Probably found a fix for the mantis issue 28380. We had a similar problem as described in this bug report. But instead of the problem with links to user profiles, we had a problem with links to files.


Amstutz:
We run in to the same issue when performing the update. Since the feature of adding a username into some LM is rather common on our installation, this was major. Thx to @rsheer for the patch, we have that on production now.

alex40724:
Thanks, fixed a little bit differently.

PR #2822, opened 2020-08-03
Performance in Learning Progress see #28718

Amstutz:
Note, this is not well tested yet. Please review carefully before merging.

**Background**: On our Production Env. Opening Learning Progress on files with enabled "Ermittlung des Lernfortschritts" is extremely slow (up to several minutes). On the test Env with a weaker DB server, it crashes completely with a timeout. Loggin Queries shows 5000+ queries to the DB by just opening e.g. a file with enabled "Ermittlung des Lernfortschritts".  Even if LP is currently inactive for the file. See: https://mantis.ilias.de/view.php?id=28718

**Impact**: This change reduces the loading time from minutes to < 1 sec as well as a drastic reduction in the queries used.

**Explanation**: Issue is a large part of the function getLPPathInfo, which loads all LP information of all parents which can be a huge amount of data. Interestingly, this data seems to be used nowhere since the respective code is outcommented in handleLPUsageInfo (see removed lines, IMO it is better to delete ouctommented code. It is not lost and stays in the history of the Version Control system but makes the productive source look cleaner). So I would propose to remove the code which looks like not in use anymore in getLPPathInfo as well. Please check if this is really not needed anywhere anymore.

### DIFF
--- a/Services/Tracking/classes/repository_statistics/class.ilLPListOfSettingsGUI.php
+++ b/Services/Tracking/classes/repository_statistics/class.ilLPListOfSettingsGUI.php
@@ -403,56 +403,6 @@ class ilLPListOfSettingsGUI extends ilLearningProgressBaseGUI
                 $node["child"] != $a_ref_id) {
                 $a_res[$node["child"]]["node"]["lp"] = true;
                 $has_lp_parents = true;
-                
-                $parent_obj_id = $node["obj_id"];
-                $parent_obj_lp = ilObjectLP::getInstance($parent_obj_id);
-                
-                // parse LP collection
-                $parent_collection = $parent_obj_lp->getCollectionInstance();
-                if ($parent_collection &&
-                    $parent_collection->hasSelectableItems() &&
-                    $parent_collection->isAssignedEntry($a_ref_id)) {
-                    $a_res[$node["child"]]["node"]["active"] = true;
-
-                    if ($parent_collection instanceof ilLPCollectionOfRepositoryObjects) {
-                        $group = array();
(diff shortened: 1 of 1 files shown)


Services/Tracking/classes/repository_statistics/class.ilLPListOfSettingsGUI.php
```diff
@@ -403,56 +403,6 @@ protected function getLPPathInfo($a_ref_id, array &$a_res)
...
-                
-                $parent_obj_id = $node["obj_id"];
-                $parent_obj_lp = ilObjectLP::getInstance($parent_obj_id);
-                
-                // parse LP collection
-                $parent_collection = $parent_obj_lp->getCollectionInstance();
-                if ($parent_collection &&
-                    $parent_collection->hasSelectableItems() &&
-                    $parent_collection->isAssignedEntry($a_ref_id)) {
-                    $a_res[$node["child"]]["node"]["active"] = true;
```
smeyer-ilias, line 415: Without this line all parent objects have an inactive LP

smeyer-ilias (changes requested):
Thanks, Timon
I highly appreciate the analysis and will try to improve the implementation.

smeyer-ilias:
I commited a new version of the parent node LP status determination. 
Please can you check whether commit 850d9fa5f86bd7bab0b89fec47bdbfc47b496578 leads to similar performance improvements.

Amstutz:
Hi @smeyer-ilias ,

Thank you very much for looking into that and for the quick response. Good catch concerning the parents LP, that slipped my attention. Thansk for the review and the new commit.

I just run several tests, and it look like the perf. improvements remain in the corrected version. Nice.

Thx.

PR #2823, opened 2020-08-03
WAC/Random: Fix Mantis issue 28719

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=28719

### DIFF
--- a/Services/Randomization/classes/class.ilRandom.php
+++ b/Services/Randomization/classes/class.ilRandom.php
@@ -7,18 +7,22 @@
 class ilRandom
 {
     /**
-     * @var ilLogger|null
-     */
-    private $logger;
-
-    /**
-     * constructor.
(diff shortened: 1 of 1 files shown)


chfsx:
@mjansenDatabay  LGTM from my side, @smeyer-ilias up to you to merge this

smeyer-ilias:
Thanks Michael, this fixes the issue, but makes the logStack() calls quite useless since the original call stack is not shown anymore.
Nevertheless I merged it to the stable releases.

mjansenDatabay:
@smeyer-ilias 

> ... this fixes the issue, but makes the logStack() calls quite uselesse ...

That's true. This was a hotfix (because many WAC requests failed because of the issue) and by far not the best solution. But IMO having a few more call stack log lines for the added closure are not a big issue.

PR #2824, opened 2020-08-04
UI/Dropzone: Add fix for type check of 'lastRedirect' property

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=28726

If merged, this should be also cherry-picked to `release_5-4` and `trunk`.


Amstutz:
Looks correct, thx!

Amstutz:
Promoted to 5.4, 6 and trunk as well.

PR #2825, opened 2020-08-04
[Mantis 28727] Remove reference to deleted block_top.png

Rillke:
`block_top.png` has been deleted in commit f1f472c798f. This commit removes the rules from less and css because the reference is causing 404s in server logs which may trigger false alerts to app admins.


Amstutz:
Hi @Rillke ,

Good catch thx a lot and Kudos. ill promote that down to 6 as well.

thx.

PR #2828, opened 2020-08-06
fix_28737: use il-footer-bg-color in default layout

fsesterh:
In case #28737 is classified as a bug and wasn't an intentional change for ILIAS 6.


mjansenDatabay:
https://mantis.ilias.de/view.php?id=28737

Amstutz:
Hi @fsesterh 

Thanks a lot for the PR, nice catch @fsesterh and @mjansenDatabay thx a lot. Kudos.

Ill promote this down to 6 as well.

PR #2829, opened 2020-08-06
Modules - StudyProgramme: set permissions for auto content references 

lukastocker:
If permissions are not disabled and the course is online, auto content references can now be seen in the StudyProgramme overview.

https://mantis.ilias.de/view.php?id=28562


klees:
Also picked to trunk.

PR #2830, opened 2020-08-06
Fixed typo in Survey Question Pool

rsheer:
This commit just fixes a very small typo. For a call to `ilObjSurveyQUestionPoolAccess`, the U is written uppercase instead of lowercase. Even though class names in PHP seem to be case insensitive, our installation throws an exception for this.


mjansenDatabay:
@alex40724 We stumbled upon this old PR in our TB meeting. Do you have any objections to merge this?

PR #2834, opened 2020-08-10
[Mantis 15366] Test: Reliably calc. time remaining

Rillke:
image

- The implementation before this fix used an incremented counter based
  approach for calculating time remaining: Every 1000ms add one to
  your counter and assume the counter is the time passed since page
  load in s.
- Modern browsers de-prioritize inactive tabs in order to safe CPU
  and energy.
- This causes `setInterval()` fire less frequently than expected.
- The clock is slow.
- Instead, we use the browser's performance API to calculate the time
  passed since test start. The performance API is supported by all
  modern browsers and does not suffer from above issues, not even from
  time zone or clock changes.
  https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
  https://www.w3.org/TR/hr-time-2/#dom-performance-now
- This is similar to how frames in (3D) games are calculated.
- Additional changes:
    - prevent pollution of global scope wrapping the code in an IIFE
    - begin sort the code units into functions
    - use more strict comparison operators
    - make jshint happy declaring global dependencies

### DIFF
--- a/Modules/Test/classes/class.ilTestPlayerAbstractGUI.php
+++ b/Modules/Test/classes/class.ilTestPlayerAbstractGUI.php
@@ -1483,6 +1483,9 @@ abstract class ilTestPlayerAbstractGUI extends ilTestServiceGUI
         $this->tpl->setVariable("USER_REMAINING_TIME", sprintf($this->lng->txt("tst_time_already_spent_left"), $str_time_left));
         $this->tpl->parseCurrentBlock();
 
+        // jQuery is required by tpl.workingtime.js
+        require_once "./Services/jQuery/classes/class.iljQueryUtil.php";
+        iljQueryUtil::initjQuery();
         $template = new ilTemplate("tpl.workingtime.js", true, true, 'Modules/Test');
         $template->setVariable("STRING_MINUTE", $this->lng->txt("minute"));
         $template->setVariable("STRING_MINUTES", $this->lng->txt("minutes"));
--- a/Modules/Test/templates/default/tpl.workingtime.js
+++ b/Modules/Test/templates/default/tpl.workingtime.js
@@ -1,45 +1,104 @@
-	var serverdate = -1;
-	var unsaved = true;
-	
-	function setWorkingTime()
-	{
-		if (serverdate == -1) 
-		{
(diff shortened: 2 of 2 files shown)


Rillke:
Should I undo the white space changes in `class.ilTestPlayerAbstractGUI.php`?

Rillke:
Additional note about why we use an offset to an initial `performance.now()`:

> The current high resolution time is the high resolution time from the time origin to the present time (typically called "now").

> The time origin is the time value from which time is measured:
>    If the global object is a Window object, the time origin MUST be equal to:
>        the time when the browsing context is first created if there is no previous document;
>        otherwise, the time of the user confirming the navigation during the previous document's prompt to unload algorithm, if a previous document exists and if the confirmation dialog was displayed;
>        otherwise, the time of starting the navigation responsible for loading the Window object's newest Document object.

The issue here is that the performance timer begins when navigation starts, i.e. when the request is sent to the server, not when the server calculates all the time values filled into this template script. When the server is very busy, or the request is transmitted slowly, we would assume too much time to be passed. That's an issue because auto-save/-submission, which is triggered when the time reaches the test's end, would be before the test actually ends. ILIAS will just display "autosave successful" in this case instead of the expected "The test ended and your last solution has been stored and you will be redirected in a few moments"

mjansenDatabay:
@Rillke Thx for your contribution. Indeed, the whitespace changes are a bit distracting when reviewing the changes :-).

@mbecker-databay Please review this PR and merge it accordingly (to other stable branches as well) if you don't have any objections.

Rillke:
> [Required and Supported Software ILIAS 6](https://docu.ilias.de/goto_docu_wiki_wpage_5724_1357.html): ILIAS supports current versions of Firefox and Chrome. The browser support for particular views or features can be limited due to browser CSS/JavaScript limitations.

Is this statement final?

Considering [the compatibility matrix](https://caniuse.com/#feat=mdn-api_performance_now), means we can drop the counter-based approach altogether. I am asking because in the past I noticed some commits restoring compatibility in ILIAS with older software.

Rillke:
@mjansenDatabay The white space changes complicating review have been removed. Thanks for your quick response.

gvollbach (approved):
Looks good, thanks for the contribution @Rillke 

@mbecker-databay Review is done.

PR #2837, opened 2020-08-12
Forum: Hotfix for Mantis issue 28765

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=28765

### DIFF
--- a/setup/sql/6_hotfixes.php
+++ b/setup/sql/6_hotfixes.php
@@ -382,3 +382,474 @@ $remove = ['ILIAS\Administration\AdministrationMainBarProvider|adm',
 $ilDB->manipulate("DELETE FROM il_mm_items WHERE ".$ilDB->in('identification', $remove, false, 'text'));
 
 ?>
+<#31>
+<?php
+/* Determine threads which cannot be migrated because they are not valid (just to make the script robust) */
+$query = "
+        SELECT frmpt.thr_fk
+        FROM frm_posts_tree frmpt
+        INNER JOIN frm_posts fp ON fp.pos_pk = frmpt.pos_fk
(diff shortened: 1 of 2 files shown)


mjansenDatabay:
@Amstutz / @yvseiler FYI

Amstutz:
Hi @mjansenDatabay 

Thx a lot. I just ran this script on a testinstallation with around 65k threads. Everything went smoothly, I did around 10 random checks of well populated threads. The tree is shown nicely afterwards posts as well. 

However I observed two things:

- [ ] answering on previously existing post does not seem to work, new posts are always attached to the very bottom.
- [ ] There seems to be an issue with the following query:
```
        $query = '
            UPDATE frm_posts_tree
            SET 
                parent_pos = CASE WHEN parent_pos = %s THEN %s ELSE parent_pos END, 
                rgt = rgt + %s,
                lft = lft + %s,
            WHERE lft >= %s AND rgt <= %s AND thr_fk = %s 
        ';
```
Is the "," before WHERE really needed? Further it seemst to be possible that $source_parent filling the first %s can be null and therefore making the query invalid. I had to slightly adapt with:

```
        if(!$source_parent){
            $source_parent = "0";
            $GLOBALS['ilLog']->warning("source parent has been 0, manualy set to 0!");
        }

```

However, I have no clue why this was possible, I only had one such case.

mjansenDatabay:
@Amstutz : Argh, you were to fast 😁. I fixed some issues in the meantime. Maybe your testinstallion thread trees are now even more corrupted than before 🤣. You should restore it from a backup. I will contact you if it our internal tests are successful tomorrow.

Amstutz:
No issue, just give me a nudge as soon as I should perform another go.

PR #2838, opened 2020-08-13
[FIX] removed ilMainMenuGUI

chfsx:
Hi there!
In https://mantis.ilias.de/view.php?id=28772 I just wanted to remove the template file, but it is still used in ilMainMenuGUI which is itself used in a lot of components. May we remove those places in ILIAS 6 or should we remove this in trunk?

my preference would be to remove it in trunk only


Modules/MediaPool/classes/class.ilObjMediaPoolGUI.php
```diff
@@ -581,27 +581,27 @@ public function listMedia()
...
-    
+
             $ilToolbar->addButton(
                 $lng->txt("mep_create_folder"),
                 $ilCtrl->getLinkTarget($this, "createFolderForm")
             );
-                        
-            if (trim($mset->get("upload_dir")) != "" && ilMainMenuGUI::_checkAdministrationPermission()) {
+
+            if (trim($mset->get("upload_dir")) != "") {
```
alex40724, line 604: This will remove the admin check for this feature.

Please change this to
```(trim($mset->get("upload_dir")) != "" && false)```
instead. I will fix this afterwards.

File: Modules/MediaPool/classes/class.ilObjMediaPoolGUI.php [hunk 2]
```diff
@@ -1825,7 +1825,7 @@ public function createMediaFromUploadDir()
 
         include_once("./Services/MediaObjects/classes/class.ilObjMediaObject.php");
 
-        if (is_array($_POST["file"]) && ilMainMenuGUI::_checkAdministrationPermission()) {
+        if (is_array($_POST["file"])) {
```
alex40724, line 1828: same as above, please add && false

alex40724 (changes requested):
Beside the two changes in the comments, everything is ok from my side. Thanks!

matthiaskunkel:
Jour Fixe, 17 AUG 2020: We accept the removal of ilMainMenuGUI for trunk but keep it in release_6 to avoid side effects. @chfsx : Please make a short notice in the code of ILIAS 6.

chfsx:
This will be done using a refactoring for ILIAS 8

PR #2842, opened 2020-08-15
Composer: Delete directories by PHP code, not by OS specific code

mjansenDatabay:
This PR changes the approach to delete directories from the composer dependency folders on `post-install-cmd` or `post-update-cmd`. Instead of using OS specific code I use PHP to recursively delete folders.

Currently ILIAS does not work on Microsoft Windows (I am aware that MS will drop the official support for PHP on Windows).

I guess this PR makes @mbecker-databay happy ;-).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,16 +12,16 @@
 			"php setup/cli.php build-artifacts --yes"
 		],
 		"post-install-cmd": [
-		  "rm -rf libs/composer/vendor/geshi/geshi/contrib"
+			"php libs/composer/rmdirs.php"
 		],
 		"post-update-cmd": [
-		  "rm -rf libs/composer/vendor/geshi/geshi/contrib"
+			"php libs/composer/rmdirs.php"
 		],
 		"test-php-all": [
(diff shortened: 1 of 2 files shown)


mbecker-databay:
It certainly does!

mjansenDatabay:
@alex40724 You introduced the `rm -rf libs/composer/vendor/geshi/geshi/contrib`. Are the proposed changes okay for you?

alex40724 (approved):
Thanks @mjansenDatabay!

This is fine for me. Ee might want to move the code to a mini cli php script being called here instead of copy/pasting this quite long line of code more often.

mjansenDatabay:
> We might want to move the code to a mini cli php script being called here instead of copy/pasting this quite long line of code more often.

@alex40724 I totally agree. Maybe such a script (or command) could be part of the CLI setup currently implemented by @klees , because it is a final step to setup a `safe` ILIAS installation.

mjansenDatabay:
@klees Any objections or opinions here? I request a review because this has (IMO) some correlations to the setup.

klees:
No, it's fine for me.

PR #2500, opened 2020-02-13
Aria (Landmark) Roles + some more accessibility fixes

tfamula:
Hello all,
this PR improves/introduces [Aria Landmark Roles](https://docu.ilias.de/goto_docu_wiki_wpage_5879_1357.html) to make ILIAS more accessible. 

> Aria landmark roles allow identifying the organization and structure of ILIAS. They classify and label different sections of ILIAS, allowing screen readers to read out visually conveyed structural information. Screen readers provide keyboard navigation to landmark roles.

To make these changes more comprehensible, you find some explanations for each landmark role listed below. You will probably also need these references:

- [Landmark Roles by W3C](https://www.w3.org/TR/wai-aria-practices-1.1/#landmark-roles)

- [Landmark Roles by MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles) (less official, but more comprehensible)

Banner:
- `role="banner"` is added to the page template
- `<header>` tag cannot be used here because its context is not the `<body>` element and it is descendant of `<section>` element

Footer:
- `role="contentinfo"` is added to the page template
- `<footer>` tag cannot be used here because its context is not the `<body>` element and it is descendant of `<section>` element

Navigation:
1. The `<nav>` tag for main controls (Mainbar + Slates) is removed because Slates are not always used for navigation and can contain other content. Only the Mainbar gets the `<nav>` tag instead.
2. The UI Tree gets a `<nav>` tag.
3. The Legacy Tree gets a `<nav>` tag.
4. For Breadcrumbs, only the second level in the HTML gets the `<nav>` tag. In the first level, the tag is removed because the screen reader should not identify empty breadcrumbs as a landmark role and also not read it twice when breadcrumbs are shown.

- Because there are multiple navigation, each of them gets an `aria-label` to distinguish them.

Main:
- `<main>` tag is already used in page template, no changes needed here

Complementary:
- `<aside>` tag is added to the right page content

Form:
- Forms in ILIAS do not have `aria-labelledby`,` aria-label` or `title` element. That is why they must get `role="form"`. This is already given and no changes are needed.

Search:
- The Bulky Button for Search in the Metabar gets `role="search"`

Dialog (no landmark role):
- Slates in the Metabar get `role="dialog"`

Tab (no landmark role):
- First-level tabs already have `role="tablist"` respectively `role="tab"`, no changes needed here

[...]

PR #2844, opened 2020-08-17
Aria-label for UI Tree

tfamula:
This is a follow-up of one open task, which was discussed in https://github.com/ILIAS-eLearning/ILIAS/pull/2500. 
It tackles the UI Tree, which should get a new (aria-)label to improve accessibility. Because the UI Tree is already used in a few places, this PR includes also suggestions for existing Trees, where the labels have to be added retroactively.

### DIFF
--- a/Modules/Forum/classes/class.ilForumExplorerGUI.php
+++ b/Modules/Forum/classes/class.ilForumExplorerGUI.php
@@ -134,6 +134,14 @@ class ilForumExplorerGUI extends ilTreeExplorerGUI
     }
 
     /**
+     * @return string
+     */
+    public function getTreeLabel()
+    {
+        return $this->lng->txt("frm_posts");
+    }
+
+    /**
      * @inheritDoc
      */
     public function getTreeComponent() : Tree
@@ -152,7 +160,7 @@ class ilForumExplorerGUI extends ilTreeExplorerGUI
         ];
 
         $tree = $this->ui->factory()->tree()
-            ->expandable($this)
+            ->expandable($this->getTreeLabel(), $this)
             ->withData($rootNode)
             ->withHighlightOnNodeClick(false);
 
--- a/Modules/OrgUnit/classes/Provider/OrgUnitToolProvider.php
+++ b/Modules/OrgUnit/classes/Provider/OrgUnitToolProvider.php
@@ -57,11 +57,13 @@ class OrgUnitToolProvider extends AbstractDynamicToolProvider
     private function getTree() : Tree
     {
         global $DIC;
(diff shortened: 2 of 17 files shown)


mjansenDatabay (approved):
Thx @tfamula , the changes in my consuming components LGTM!

IMO the `label` should be the 2nd parameter when retrieving the tree from the factory, but tha's just my personal taste (maybe because it is a primitive type).

alex40724 (approved):
+1 Thanks @tfamula !

Amstutz:
Hi @tfamula 

Thanks a lot for this PR. As far as I understand the linked guideline in the PR above, I believe this to be a step in the right direction of making ILIAS more accessible. Kudos!

This is just a review on the interface, as soon as accepted by the JF, I will also provide one on the implementation:

Improve or argue why not:
- [x] Some more detail in Rule: The proposed rule:    
`*The "aria-label" attribute MUST be set for Expandable Trees, which MUST be language-dependant.`

Is good, but to me it should also provide help in finding a good label. We should make clear here, the simply providing e.g. "Tree" is not good enough. I propose to rephrase to:
```
*The "aria-label" attribute MUST be set for Expandable Trees
*The "aria-label" MUST be language-dependant.
*The "aria-label" MUST describe the content of the tree as precisely as possible. "Tree" is not a good label, "Forum Posts" is much more helpful. Note that the "Tree" is already set by the aria-role attribute.
```

Since the interface change is not large and therefore not too complex to discuss, I put this on the JF today. This is on very short notice, if anybody needs more time to review the interface change here, feel free to ask for it.

klees:
Hi @tfamula.

what @Amstutz says + Please keep the order of the parameters of the constructor (other then what @mjansenDatabay suggests). There are many components where the label comes first in the parameter list, I think this is a good convention to maintain.

Thanks!

matthiaskunkel:
Jour Fixe, 17 AUG 2020: We highly appreciate this PR and accept it for trunk and ILIAS 6 (bug fix). Please merge.

tfamula:
Hi @Amstutz and @klees 
Thanks for your feedback! I agree with everything and updated the rules for the Tree according to Timon's suggestion. Please have a quick look, because I adjusted the wording a little bit.

Amstutz:
Hi @tfamula 

Nice, thx a lot. looking good. Is the complete implementation ready for the final review?

Thx.

tfamula:
@Amstutz Yes, it is. The tests are added / modified, too.

klees (approved):

Amstutz:
Thx a lot, I just cherry-picked this down to 6 as well.

PR #2845, opened 2020-08-21
Service/Component: purge plugin ctrl-structure before reloading it

klees:
This is suppposed to fix [#28284](https://mantis.ilias.de/view.php?id=28284) by making the control structure loading for plugins more robust. Without the change, updating the plugin just writes ctrl-structure into the database without purging existing ctrl-structure entries for the plugin. This leads to the problem described in the bug, when the structure of an non-installed plugin is loaded. Since a ctrl-structure load for the plugin should load all entries for that plugin anyway, we could purge existing entries before.


klees:
@daniwe4 Could you please try this out? I.e. check, if normal install and update of plugins works and the bug is also fixed? Get back to me if you have any questions.

klees:
There seem to be some more issues with this, see bug report.

PR #2849, opened 2020-08-24
fixed 0028471: removed grid in print-styles

catkrahl:
This fixes the print styles for longer page content.

https://mantis.ilias.de/view.php?id=28713

### DIFF
--- a/src/UI/templates/default/Layout/standardpage.less
+++ b/src/UI/templates/default/Layout/standardpage.less
@@ -562,9 +562,14 @@ footer {
 			header,
 			.il-maincontrols,
 			.breadcrumbs,
-			footer {
-				display: none;
+			footer,
+			.ilFileDropTargetOverlay {
+				display: none !important;
 			}
+		}
+		.il-layout-page-content,
+		.il-layout-page-content div	{
+			display: block;
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


klees:
@Amstutz Looks plausible to me. Want to have a look?

Amstutz:
Hi @catkrahl and @klees 

Thx a lot for this. The proposed solution resolves the issue with printing only the first page in FF. However, it introduces a new one by displaying dropzones in print view in the content pages of courses for admin. I look into it. Ill get back to you here.

Thx.

Amstutz:
Seems like .ilFileDropTargetOverlay is suddenly shown, due to the new display block type for print view. Display Block is added inline per JS, so this will certainly solve the issue (in media print):

```
    .ilFileDropTargetOverlay{
      display: none !important;
    }
```

However, you might finde a way working around the `!important ` statement.

catkrahl:
i found out the same solution and will update this PR... thx @Amstutz

catkrahl:
@Amstutz: i removed all the relevant container-classes in the print-query. please have a look... now printing seem to work fine in FF without any droptargetoverlays...

klees:
@Amstutz Are you on this? IMO we could merge this and apply it to release_6 as well. I do not like the `!important` but currently do not see a realistic and fast way to solve this more generally. Print style seems to be an angle we need to look at when thinking about SASS and KS elements, currently we are not very good in this department.

Amstutz:
Yes on it, only thing left is compile to CSS, then this is ready for 6 and trunk.

klees:
Also picked to release_6: https://github.com/ILIAS-eLearning/ILIAS/commit/569358bfcb473866a727a293e0404a9a3ac4f473

PR #2850, opened 2020-08-25
Database/Setup: password in config is optional

klees:
Hi @chfsx,

I discovered this bug when debugging some other problem: the `ilDatabaseSetupConfig` says that `password` might be null, but these two locations did not respect that and explode when people do in fact set no database password.

This should fix the issue and would be required in the trunk as well.

Best regards!


chfsx (approved):
Thanks a lot @klees !

PR #2853, opened 2020-08-27
LSO: space between lines in Kiosk-navbar; fixes #28783

nhaagen:
https://mantis.ilias.de/view.php?id=28783
"As soon as there are two lines due to too many custom menu entries there is no padding between the lines."

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -19022,6 +19022,9 @@ span.ilProfileBadge .modal .img-responsive {
   margin-bottom: 10px;
   background-color: #e0e0e0;
 }
+.ilLSOKioskModeNavigation .navbar-form {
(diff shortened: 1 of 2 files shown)


klees (changes requested):
I hate this, because we would use absolute px-values =) Not your fault, though, but some habit we need to change soon. Please get in touch with @catkrahl and/or @Amstutz (?) to find out if there is some less-variable we could use instead.

Thanks!

nhaagen:
I know, I hate it, too. But line-height, computed line-height, etc. is too low.
I saw other absolutes here, so I went for this, rather than adding more less-vars.

klees:
Picked to trunk as well.

PR #2855, opened 2020-08-27
Making maps more robust for async requests

Amstutz:
We use maps in some extension async in a modal. I had to make the changes put in this PR to make it work. They just make the script somewhat more robust against unitialized variables.

However, not sure if this is needed, since there is currently some refactoring on the way anyways. So take or leave.


klees:
Thanks @Amstutz!

klees:
@daniwe4 Please have a look how this should be incorporated in the refactored maps-implementation in the trunk.

PR #2858, opened 2020-09-01
PHP 8 Support (Travis/Composer)

mjansenDatabay:


mjansenDatabay:
IMO updating single `Composer` dependencies step by step will **NOT** be possible. IMO the best way is to rebuild the `composer.lock`. With the state I proposed here we are at least able to execute a `composer install` without getting any errors.

I need some cross-checks regarding the `dev dependencies` I removed here. IMO all these were dependencies (of dependencies ...) of our dependencies.
- `phpunit/php-code-coverage`
- `phpunit/php-token-stream`
- `phpdocumentor/reflection-docblock`
- `sebastian/environment`
- `phpunit/php-timer`
- `phpspec/prophecy`
- `theseer/tokenizer`
- `myclabs/deep-copy`

Does anyone know if we have an explicit dependency on one of these libraries?

Current Issues:
- In this PR, I removed `zendframework/zend-httphandlerrunner`. This has to be replaced by an alternative, as suggested in https://github.com/ILIAS-eLearning/ILIAS/pull/3039

alex40724 (approved):
Looks good to me. I am not aware of a direct dependency to the mentioned libs.

mjansenDatabay:
Closed because of #3084 and the pending issue with `zendframework/zend-httphandlerrunner` and `laminas/laminas-httphandlerrunner`.

PR #2856, opened 2020-08-31
Error handling: Modify strings

mjansenDatabay:
https://mantis.ilias.de/view.php?id=28821

PR #2861, opened 2020-09-02
Remove POST from SESSION

mjansenDatabay:
https://mantis.ilias.de/view.php?id=28821

Follow up of #2856

### DIFF
--- a/Services/Init/classes/class.ilInitialisation.php
+++ b/Services/Init/classes/class.ilInitialisation.php
@@ -1145,15 +1145,6 @@ class ilInitialisation
         );
         $ilErr->setErrorHandling(PEAR_ERROR_CALLBACK, array($ilErr, 'errorHandler'));
 
-        // :TODO: obsolete?
-        // PEAR::setErrorHandling(PEAR_ERROR_CALLBACK, array($ilErr, "errorHandler"));
-
-        // workaround: load old post variables if error handler 'message' was called
-        include_once "Services/Authentication/classes/class.ilSession.php";
-        if (ilSession::get("message")) {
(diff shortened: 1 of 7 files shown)


Services/Init/classes/class.ilInitialisation.php
```diff
@@ -1145,15 +1145,6 @@ protected static function initCore()
         );
         $ilErr->setErrorHandling(PEAR_ERROR_CALLBACK, array($ilErr, 'errorHandler'));
 
-        // :TODO: obsolete?
-        // PEAR::setErrorHandling(PEAR_ERROR_CALLBACK, array($ilErr, "errorHandler"));
-
-        // workaround: load old post variables if error handler 'message' was called
-        include_once "Services/Authentication/classes/class.ilSession.php";
-        if (ilSession::get("message")) {
```
mjansenDatabay, line 1153: @alex40724 The only usage of `$_SESSION["message"]`, `$_SESSION['message']`, `ilSession:set('message'` or `ilSession:set("message"` is: https://github.com/ILIAS-eLearning/ILIAS/commit/6c1e193c4181ab6af5da2e3454c76a17658c35a5#diff-19d773d6a143f355b2b5edecc3f7b893R495

Any objections?

alex40724: Very interesting. Ancient code. `$_POST = $_SESSION["post_vars"];` exists since the first commit of ilInitialisation in May 2006. I created this class, but most probably copied this part from some other code file. So something may break here. Hard to tell what exactly. :-) For the trunk this change is ok for me. We will see.

alex40724 (approved):

PR #2862, opened 2020-09-03
LSO, Legacy LP checks, fix #28840

nhaagen:
Constantly pulling LP of legacy objects takes its toll on resources; 
this relays the current object's obj_id to the RPC and thus avoids looping over all learner items.
That should reduce load quite a bit, especially in LSOs with many objects.
Also, the default interval is now set to 4000.

https://mantis.ilias.de/view.php?id=28840

### DIFF
--- a/Modules/LearningSequence/classes/class.ilObjLearningSequenceGUI.php
+++ b/Modules/LearningSequence/classes/class.ilObjLearningSequenceGUI.php
@@ -156,6 +156,19 @@ class ilObjLearningSequenceGUI extends ilContainerGUI
 
         //exit real early for LP-checking.
         if ($cmd === LSControlBuilder::CMD_CHECK_CURRENT_ITEM_LP) {
+            if (isset($_GET[LSControlBuilder::PARAM_LP_CURRENT_ITEM_OBJID])) {
+                $current_obj_obj_id = (int) $_GET[LSControlBuilder::PARAM_LP_CURRENT_ITEM_OBJID];
+
+                print (int) ilLPStatus::_lookupStatus(
+                    $current_obj_obj_id,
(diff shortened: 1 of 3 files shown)


Modules/LearningSequence/classes/class.ilObjLearningSequenceGUI.php
```diff
@@ -156,6 +156,19 @@ public function executeCommand()
...
         if ($cmd === LSControlBuilder::CMD_CHECK_CURRENT_ITEM_LP) {
+            if (isset($_GET[LSControlBuilder::PARAM_LP_CURRENT_ITEM_OBJID])) {
+                $current_obj_obj_id = (int) $_GET[LSControlBuilder::PARAM_LP_CURRENT_ITEM_OBJID];
+
+                print (int) ilLPStatus::_lookupStatus(
+                    $current_obj_obj_id,
+                    $this->user->getId(),
+                    true
+                );
+                exit;
```
mjansenDatabay, line 167: This introduces a new dicto violation.

For ILIAS >= 6 I suggest to use `\ILIAS\HTTP\GlobalHttpState::saveResponse` and `\ILIAS\HTTP\GlobalHttpState::close`

klees: Good catch @mjansenDatabay, thx a lot!!

nhaagen: Good spotting; it's not that new, though, it rather perpetuates the old one ;)
I'll have a go at \ILIAS\HTTP\GlobalHttpState with trunk/6, thank you for pointing that out.

PR #2864, opened 2020-09-03
Data/URI: add some convenience functions

nhaagen:
With a URI, I often want to add/change a parameter in the query and then use the URI as string.
This PR adds those functions to the data-type.

### DIFF
--- a/src/Data/URI.php
+++ b/src/Data/URI.php
@@ -384,9 +384,14 @@ class URI
 
     public function __toString() : string
     {
-        $uri = $this->getBaseURI() . '?' . $this->getQuery();
-        if ($this->fragment) {
-            $uri .= '#' . $this->getFragment();
+        $uri = $this->getBaseURI();
+        $query = $this->getQuery();
+        if ($query) {
+            $uri .= '?' . $query;
+        }
+        $fragment = $this->getFragment();
+        if ($fragment) {
+            $uri .= '#' . $fragment;
         }
         return $uri;
     }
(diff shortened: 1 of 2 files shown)


klees (changes requested):
Hi @nhaagen,

thanks a lot! This LGTM in general but as usual I have some questions/suggestions:

* [ ] Is there a reason that we use `toString` as method name instead of the magic `__toString`?
* [ ] Let's explicitely declare `$params = []` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/2864/files#diff-a331b8ead29f905062d29eb549e7dee1R402)

Please fix and then ping me immediately, this should make the table code simpler...

Best regards!

src/Data/URI.php
```diff
@@ -381,4 +381,56 @@ public function getBaseURI() : string
...
+        $params = [];
+        parse_str($this->getQuery(), $params);
+        return $params;
+    }
+
+    /**
+     * Get the value of the given parameter (or null)
+     * @return  mixed|null
+     */
+    public function getParameter(string $param)
```
mjansenDatabay, line 416: You could use `:? string` as explicit return type here.

nhaagen:
Hi @klees ,
thanks for your comments - I changed accordingly.
Best regards

src/Data/URI.php
nhaagen: Hi @mjansenDatabay , 
I did, at first, but arrays do work :)
please see [here](https://github.com/ILIAS-eLearning/ILIAS/pull/2864/files#diff-71a136a2582520363a4a814f20782245R945-R948).

PR #2867, opened 2020-09-04
Fix missing target attribute in iltable

ghost:
https://mantis.ilias.de/view.php?id=28850

Readd missing target attribute in tpl.table2.html, so result of delete cloud plugin items works again


theodortruffer:
@alex40724 could you have a look at that? This messes with the delete workflow in every cloud plugin.
Thanks!

alex40724:
Thanks for the PR. Its merged to the upper branches as well.

theodortruffer:
@alex40724 thanks!

PR #2870, opened 2020-09-04
mysql innodb adapter make fulltext index available

christianheimann:
Make fulltext index with innoDB adapter available.
Even with recommended mysql 5.6+, innoDB was already capable to handle fulltext index.
Or was there a particular reason?


mjansenDatabay:
I think we can close this one, because the JourFixe decided to abandon the SQL Fulltext Search, see https://docu.ilias.de/goto_docu_wiki_wpage_6157_1357.html

Nevertheless, thanks a lot for your contribution @christianheimann

PR #2871, opened 2020-09-04
28853: Delete assigned users and roles when competence profile is deleted

tfamula:
Fix for [28853](https://mantis.ilias.de/view.php?id=28853)


alex40724 (approved):
Looks good, thanks!

PR #2872, opened 2020-09-07
Auth. Administration: Fix captcha settings not being displayed

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=28858


smeyer-ilias:
Thanks.

PR #2875, opened 2020-09-08
BT-25530: Added labels to calendar selection

M-Barz:
Also set label bottom-margin to 0 in template, so the standard css for labels won't affect the Calendar Selection visuals.


Amstutz:
Hi @M-Barz and hi @smeyer-ilias 

I missed this PR so I am somewhat late to the party. I am partly responsible for CSS as second maintainer. I just observed, that we have new Inline CSS in ILIAS 6:

https://github.com/ILIAS-eLearning/ILIAS/pull/2875/files#diff-919a084c5c6722c90bb88327462a5feaR20

Please remove that. Note that it makes it impossible to overwrite without using !important. Note that it also violates the CSS guideline: https://github.com/ILIAS-eLearning/ILIAS/tree/release_6/templates#general

There are situations, where there seems no way around it, if this is one of those, please note shorty why.

Thx

PR #2876, opened 2020-09-08
LSO, AdminGUI (see #28840)

nhaagen:
- add LSOs to Administration
- Customize LP polling-interval

https://mantis.ilias.de/view.php?id=28840

### DIFF
--- a/Modules/LearningSequence/classes/class.ilObjLearningSequenceAdminAccess.php
+++ b/Modules/LearningSequence/classes/class.ilObjLearningSequenceAdminAccess.php
@@ -1,0 +1,8 @@
+<?php declare(strict_types=1);
+
+/**
+ * Class ilObjStudyProgrammeAdminAccess
+ */
+class ilObjLearningSequenceAdminAccess extends ilObjectAccess
+{
+}
--- a/Modules/LearningSequence/classes/class.ilObjLearningSequenceAdminGUI.php
+++ b/Modules/LearningSequence/classes/class.ilObjLearningSequenceAdminGUI.php
@@ -1,0 +1,133 @@
+<?php declare(strict_types=1);
+
+use ILIAS\UI\Component\Input;
+
+/**
+ * LearningSequence Administration Settings
+ *
+ * @ilCtrl_Calls ilObjLearningSequenceAdminGUI: ilPermissionGUI
+ */
+class ilObjLearningSequenceAdminGUI extends ilObjectGUI
+{
+    const CMD_VIEW = 'view';
+    const CMD_EDIT = 'edit';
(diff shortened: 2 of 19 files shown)


Modules/LearningSequence/classes/class.ilObjLearningSequenceAdminGUI.php
```diff
@@ -0,0 +1,133 @@
...
+        $form = $this->getForm($values);
+        $this->show($form);
+    }
+
+    protected function save() : void
+    {
+        $form = $this->getForm()->withRequest($this->request);
+        $data = $form->getData();
+        if ($data) {
+            $data = array_shift($data);
```
klees: I think this rather belongs to the form, because it is a detail on how the form is structured internally. So we should attach it to the form via an according transformation.

Modules/LearningSequence/classes/class.ilObjLearningSequenceAdminAccess.php
```diff
@@ -0,0 +1,8 @@
+<?php declare(strict_types=1);
+
+/**
+ * Class ilObjStudyProgrammeAdminAccess
+ */
+class ilObjLearningSequenceAdminAccess extends ilObjectAccess
```
klees, line 6: Is this enough? This somehow freaks me out. Don't we need to check any permissions?

klees (changes requested):
Hi Nils,

this LGTM in general, I will take the mantis ticket to the JF before we can merge this, though.

Please introduce a `ilLSOGlobalSettings` class and an according repository, implemented over `ilSettings` as you do currently, in the meantime. Let's just stop using arrays with keys when we actually have structured data. I think this is worth the effort here because we will port this solution with only little changes to release_6+, right?

The new settings class also would be a great place to document if we store the interval in seconds or milliseconds. I wasn't quite sure which is used where, because of that `*1000` operation after reading the value.

One questions: The icon is just a copy of the current LSO-icon, right?

Thanks for looking into this!

nhaagen:
Hi @klees ,
I wrapped the settings for LSOs and of course, the icon is just a copy.
I must confess that I just copied class.ilObjLearningSequenceAdminAccess.php (i.e. mimicked the pattern).
Same - empty - classes can be found in Group, Course, StudyProgramm and others.
This  is prob. due to ilAccess::doStatusCheck,  $full_class = "ilObj" . $class . "Access"; && include_once($location . "/class." . $full_class . ".php");
I'd not tackle this here and just go with the flow in the meantime...
Cheers,
Nils

klees (approved):
Works for me. I'll be waiting for the opinion of the JF with the merge, though.

PR #2879, opened 2020-09-09
allow uuid creation from strings

Hadrian-II:
Allow uuid creation from strings, as it is needed to load Uuids from the Database or to process Uuid query-parameters

### DIFF
--- a/tests/Data/UuidTest.php
+++ b/tests/Data/UuidTest.php
@@ -29,7 +29,6 @@ class UuidTest extends TestCase
         $factory = new Factory();
         $uuid = $factory->uuid4();
 
-        $this->assertTrue(get_class($uuid) === RamseyUuidWrapper::class);
         $this->assertEquals(1, preg_match(self::VALID_UUID4, $uuid->toString()));
     }
 
@@ -53,9 +52,8 @@ class UuidTest extends TestCase
         $factory = new Factory();
         $uuid = $factory->fromString(self::UUID4);
 
-        $this->assertTrue(get_class($uuid) === RamseyUuidWrapper::class);
         $this->assertEquals(1, preg_match(self::VALID_UUID4, $uuid->toString()));
-        $this->assertTrue($uuid->toString() === self::UUID4);
+        $this->assertEquals(self::UUID4, $uuid->toString());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


klees (changes requested):
Hi @Hadrian-II,

thanks a lot for this improvement. Please add an according unit test that shows that the desired serialisation round trip UUID -> string -> UUID works as intended.

Best regards!

tests/Data/UuidTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @depends test_init
+     */
+    public function test_uuid4()
+    {
+        $factory = new Factory();
+        $uuid = $factory->uuid4();
+
+        $this->assertTrue(get_class($uuid) === RamseyUuidWrapper::class);
```
klees:
```suggestion
```

We don't need to test this. The fact that this is indeed a `Uuid` is already guaranteed via the return type hint on the factory (right?). The fact, that this is a `RamseyUuidWrapper` is an implementation detail we don't need to test IMO.

tests/Data/UuidTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @depends test_init
+     */
+    public function test_from_string()
+    {
+        $factory = new Factory();
+        $uuid = $factory->fromString(self::UUID4);
+
+        $this->assertTrue(get_class($uuid) === RamseyUuidWrapper::class);
```
klees:
```suggestion
```

See above.

tests/Data/UuidTest.php
```diff
@@ -0,0 +1,71 @@
...
+        $this->assertEquals(1, preg_match(self::VALID_UUID4, $uuid->toString()));
+        $this->assertTrue($uuid->toString() === self::UUID4);
```
klees:
```suggestion
        $this->assertEquals(self::UUID4, $uuid->toString());
```

This will allow PHPUnit to print an error message saying what we expected and what we got instead.

klees (changes requested):
Hi @Hadrian-II,

Thanks for adding missing tests for the other methods as well. I added some small change requests inline.

Best regards!

Hadrian-II:
added you changes to the tests

PR #2884, opened 2020-09-10
Use of jQuery

alex40724:
This PR fixe some links and guidelines to avoid jQuery code and libs in the future.


matthiaskunkel:
Jour Fixe, 14 SEP 2020 : We highly appreciate this PR and accept it for trunk. In case someone has a lot of dependencies in his component and needs funding to get rid of them, please contact @Amstutz.

PR #2887, opened 2020-09-11
UI/Input/Field/Numeric - allow empty strings in optional numeric inputs

nhaagen:
validation OKs empty string, isNull has no effect;
add tests.

fixes https://mantis.ilias.de/view.php?id=28861


Amstutz (approved):
Hi @nhaagen 

Thx a lot for the quick fix, adaption of examples and added tests. 

I guess the fix for 6 will look slightly different. right?  @klees feel free to merge.

PR #2888, opened 2020-09-11
UI/Input/Field/Numeric - allow empty strings in optional numeric inputs

nhaagen:
fixes https://mantis.ilias.de/view.php?id=28861 for ILIAS 6,
see https://github.com/ILIAS-eLearning/ILIAS/pull/2887

### DIFF
--- a/src/UI/Implementation/Component/Input/Field/Numeric.php
+++ b/src/UI/Implementation/Component/Input/Field/Numeric.php
@@ -14,7 +14,6 @@ use ILIAS\UI\Component\Signal;
  */
 class Numeric extends Input implements C\Input\Field\Numeric
 {
-
     /**
      * Numeric constructor.
      *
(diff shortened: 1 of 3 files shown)


Amstutz (approved):
@nhaagen 

Hero of the day. Kudos. Like to see the logical parallel refinery at work here ;-).

@klees feel free to merge.

klees (changes requested):
Hi @nhaagen,

thanks for looking into this. I really like seeing the benefit of the Refinery =)

I have two change requests:

* [ ] [This](https://github.com/ILIAS-eLearning/ILIAS/pull/2888/files#diff-ceff3bc00ba7f87a3fbaced148586879R162) is not good, as we now have a numeric input that might contain a string. This will make people trip and also will most certainly need rectification by our consumers anyway. So, if `""` is indeed what we receive from the client when the user did not enter anything in the field (which would be plausible, imo) the field should contain a `null` as this is what this means ("user did not set any value"). Please change the test and the implementation accordingly.
* [ ] [This](https://github.com/ILIAS-eLearning/ILIAS/pull/2888/files#diff-d7be0d485a86061c394456f913e3627aR60) does not do what is intended most certainly. `Logical\Parallel` applies two transformations `a` and `b` to a value `v` to produce a tuple `[a(v), b(v)]` and I don't think we want that there. I think the check for `isNumeric` would be still sufficient to check for required fields.

Get back to me if you have any questions (as always).

Best regards!

nhaagen:
Hi @klees ,
thank you, of course you are right.
Please see the last changes.

As a side-note: checks will fail at unexpected positions (checking for a result that should be error, but isn't), when language is a Mock with Custom\Constraint::getErrorMessage not returning a string. This is quite nested....

Best regards!

klees (approved):

PR #2896, opened 2020-09-14
UI/Input/Radio - numeric values, fix #28862

nhaagen:
Fix 28862 and cover with tests.
This is https://github.com/ILIAS-eLearning/ILIAS/pull/2892 for 6


klees:
Please promote to trunk as well.

PR #2901, opened 2020-09-16
0028831: Learning Progress Performance -> SCORM

smeyer-ilias:
This improves the performance of the LP matrix view, if SCORM objects are part of a collection.
Without this static cache the method checkScormPreconditions() performs NUMBER_OF_USERS * NUMBER_SCORM_ITEMS database queries.


Uwe-Kohnle:
Thanks.

PR #2859, opened 2020-09-01
[WIP] Update accessibility.md

atoedt:
changes on cheat sheet: 
* changes are supposed to reflect discussion in SIG accessibility on process (i.e. title) 
* changes supposed to reduce number of things to check to the utmost minimum (leaving out second rank issues) 
* changes supposed to reduce ambivalences and improve decidability

This is still work in progress and not finished yet, sorry.

PR #2903, opened 2020-09-16
Aria Landmark Roles for accessibility.md

tfamula:
Aria Landmark Roles for #2859

### DIFF
--- a/docs/development/accessibility.md
+++ b/docs/development/accessibility.md
@@ -796,6 +796,24 @@ Check how your suggested element fits into our semantic structures.
 8.	Always use the language attribute on the HTML element. Screen readers change pronunciation automatically. (3.1.1)
 9.	Indicate the language for each chunk of text (3.1.2). Using the language attribute on the HTML element. Screen readers change pronunciation automatically. 
 10.	Functional and informative components, i.e. labels and icons, must have consistent labels, aria-labels and alt-texts. 
+##Landmark Roles
+* Banner
+    * For the Header of the UI Component [*Standard Page*](../../src/UI/Component/Layout/Page/Factory.php), where Logo and Title are placed, the HTML tag < header > is used to be identified as the ARIA Landmark Role *Banner*.
+* Complementary
+    * In the template [*tpl.page_content.html*](../../templates/default/tpl.page_content.html), the HTML tag < aside > is used for the right column of the page to be identified as the ARIA Landmark Role *Complementary*.
+* Contentinfo
+    * For the Footer of the UI Component [*Standard Page*](../../src/UI/Component/Layout/Page/Factory.php), the HTML tag < footer > is used to be identified as the ARIA Landmark Role *Contentinfo*. As long as the Footer is nested in the HTML element < main >, the HTML element of the Footer is additionally declared with the ARIA role *contentinfo*.
+* Form
+    * currently missing
+* Main
+    * For the content area of the UI Component [*Standard Page*](../../src/UI/Component/Layout/Page/Factory.php), the HTML tag < main > is used to be identified as the ARIA Landmark Role *Main*.
+* Navigation
+    * For the UI Component [*Main Bar*](../../src/UI/Component/MainControls/Factory.php), the HTML tag < nav > is used to be identified as the ARIA Landmark Role *Navigation*.
+    * For the UI Component [*Breadcrumbs*](../../src/UI/Factory.php), the HTML tag < nav > is used to be identified as the ARIA Landmark Role *Navigation*.
+* Search
+    * currently missing
+* Application
+    * not needed
 ## Alt-Texts and aria labels
 11.	All non-text elements (icons, glyphs, images, graphs and the like) must have alt-texts (1.1.1). Do they? For all their statuses? 
 12.	The aria-label for an element must contain the text the element presents (2.5.3). This is really important for speech input. 


klees:
Hey @tfamula,

thx for your suggestions. We believe it to be a good idea to use aria landmarks and, even better, semantic HTML in general.

Please be so kind and help us to understand your motivation and general direction by answering some questions:

- [ ] Why do we need *both*, the aria landmarks and the semantic HTML-elements? The [W3C-Recommendation states that WAI-ARIA "clarifies semantics to assistive technologies when authors create new types of objects [...] that are not yet directly supported by the language of the page"](https://www.w3.org/TR/wai-aria/#co-evolution), which does not seem to be the case for the HTML-elements you mention. Or do you just propose to use the according HTML elements, without additional aria-attributes? Why do we need to mention the aria landmarks then at all?
- [ ] Why do you add this in the "Cheat Sheet" section of the file? If this is supposed to help developers in the future, we wonder when someone will ever need info about the header or the footer again. Maybe this is supposed to fulfill some purpose we are just not seeing currently.

Some requests for improvement:

- [ ] Stay more general in this doc here by mentioning semantic HTML (maybe with a link to a list of elements) and aria for additional semantics for custom objects. Maybe the items 1. and 2. in the Cheat Sheet are already sufficient and just need some refinement.
- [ ] These new items are very concrete for specific existing UI Components. Please place them at those respective places (accessibility rules section of the respective components). This is important for us as UI Coordinators.
- [ ] „Currently Missing“ is not helpful for us, are we planning to use them in the future? Are we not using them at all for some reason? We do not believe that those items belong to the Accessibility Cheat Sheet. Maybe flesh out somewhat more to make them actual helpful. This could also be that we just note, that we just do not know if we want to use them in future.

In general we wonder, if it wouldn't be simpler and faster to just create a PR for the UI-framework that implements the requirements mentioned here and maybe enhances the rules accordingly if they are not sufficient. Since the changes are not big and at one single place, we should just implement them along with the requested guidelines.

As always, feel free to contact us.

Best regards!
@Amstutz and @klees

docs/development/accessibility.md
```diff
@@ -796,6 +796,24 @@ Check how your suggested element fits into our semantic structures.
 8.	Always use the language attribute on the HTML element. Screen readers change pronunciation automatically. (3.1.1)
 9.	Indicate the language for each chunk of text (3.1.2). Using the language attribute on the HTML element. Screen readers change pronunciation automatically. 
 10.	Functional and informative components, i.e. labels and icons, must have consistent labels, aria-labels and alt-texts. 
+##Landmark Roles
+* Banner
+    * For the Header of the UI Component [*Standard Page*](../../src/UI/Component/Layout/Page/Factory.php), where Logo and Title are placed, the HTML tag < header > is used to be identified as the ARIA Landmark Role *Banner*.
```
kergomard, line 801: This goes for all the following bullet points:
I read the passive tense in this sentence as "it is enough to put a ``<header>`` tag around it to make sure that automatically a rolle *Banner* is added". Is this really what you want to say, or should it be "...is used and a Landmark Role *Banner* is added."?

klees:
Hi @tfamula,

I assigned you because I think you are the one who will need to carry this on. There is no "owner" of the guidelines AFAIK. Feel free to disagree, though. Please ping @ILIAS-eLearning/technical-board then...

Thanks!

tfamula:
Hi @klees,
thanks for your constructive feedback and very sorry for the late answer. This one had not the highest priority in the last weeks.

On 15 October, a pared-down group of the SIG Accessibility was discussing the draft of the Accessibility Checklist (former: Cheat Sheet). We came to the conclusion that you are right and the Landmark Roles should not be mentioned in the Checklist at all. Mentioning the usage of semantic HTML is the proper way here.
But the group also agreed to keep the text for the Landmark Roles in another place. It should be placed in the ILIAS Docu in the section [Usability and Accessibility Guidelines](https://docu.ilias.de/goto_docu_pg_124778_459.html). It should give kind of an overview of the current status of Landmark Roles we use (and also of the missing ones).

> These new items are very concrete for specific existing UI Components. Please place them at those respective places (accessibility rules section of the respective components). This is important for us as UI Coordinators.

This was already done with #2500. The accessibility rules I added for the UI Components were used as templates for this current PR. I totally agree, that the UI Framework is the right place to store the information about the Landmark Roles. But spreading the information in different UI Components makes it difficult to keep the overview of all Landmark Roles as a whole. That's why we proposed a collected list in the Usability and Accessibility Guidelines (see above). This could be helpful for someone who wants to know which Landmark Roles we use and where (in which UI Components) we use them. I see the problem that the information in the Guidelines could be outdated in the future because there is not really a responsive person for it. If this would be the case and the list becomes needless, we can still remove it. Anyhow it is more important to pay attention to the rules in the UI Framework.

I hope you can agree on this "tradeoff". If so, I would close this PR.

Greetings
@tfamula

docs/development/accessibility.md
tfamula: The former is correct. It is "enough" to use semantic HTML here. The Screen Reader reads the Landmark Role automatically. Saying that the Landmark Role is _added_ would be wrong IMO because the Landmark Role is not added in the HTML, but the Screenreader itself identifies it.
But I am open to change the wording if it can be done more comprehensible.

klees:
Hi @tfamula!

I totally understand that you had different things to do.

I'm absolutely fine if anyone compiles lists, so of course you (or anybody else) are free to collect information about the landmark roles from the UI framework as well. I can't see why I would need to agree here =)

Best regards!

PR #2904, opened 2020-09-17
(fix #28907) some translations missing in module meta

utesche:
see bugreport: https://mantis.ilias.de/view.php?id=28907


matthiaskunkel:
Thanks for the fix.

PR #2905, opened 2020-09-17
Remove NO_AUTO_CREATE_USER which not works with MySQL 8

ghost:
Possible fix for https://mantis.ilias.de/view.php?id=28894

I don't know for what `NO_AUTO_CREATE_USER` is needed


chfsx:
Hi @sragfw and thanks for this PR! I'll have to investigate what happens when removing this and using a MySQl-Version which does not have NO_AUTO_CREATE_USER as default. Your changes will definitely work with MySQL 8 but I'm not sure what happens in e.g. MySQL 5.7.

In general I do not think it will be a problem in older MySQL-Releases, https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_auto_create_user

ghost:
@chfsx 
Ok, I add a version check and only remove `NO_AUTO_CREATE_USER` if MySQL >= 8

ghost:
@chfsx Current status?

chfsx:
Hi @sragfw 
Thank you for your patience! I have now fixed the 0028894 directly and removed the NO_AUTO_CREATE_USER version-independent, since it is already default.

I therefore close the PR, thank you for looking at it!

PR #2906, opened 2020-09-17
`ilObjDefReader`: Quote `system` which a keyword in MySQL 8

ghost:
Possible fix for https://mantis.ilias.de/view.php?id=28908


alex40724:
Thanks for the PR.

I accepted this and merged this (since this is my old code), but basically Services/Component is under Martins first maintainership.

PR #2907, opened 2020-09-17
`ilCachedComponentData`: Quote `system` which a keyword in MySQL 8

ghost:
Possible fix for https://mantis.ilias.de/view.php?id=28909


chfsx (approved):
LGTM, thanks @sragfw ! I'll cherry-pick that to release_6 as well

PR #2908, opened 2020-09-21
Refinery: kindlyToNull and byTrying

nhaagen:
As a successor of https://github.com/ILIAS-eLearning/ILIAS/pull/2899,
this still
 - fixes https://mantis.ilias.de/view.php?id=28861 for trunk
but also
 - adds a kindlyToNull-Transform
 - uses it with the ByTrying-Transform
and 
- moves the ProblemHandler-callable into a trait.


src/Refinery/Factory.php
```diff
@@ -175,4 +176,13 @@ public function uri() : URI\Group
     {
         return new URI\Group();
     }
+
+    /**
+     * Accepts Transformations and uses first successful one.
+     * @param Transformation[]
```
klees:
```suggestion
     * @param Transformation[] $transformations
```

src/Refinery/KindlyTo/Transformation/NullTransformation.php
```diff
@@ -0,0 +1,45 @@
...
+        );
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function __invoke($from)
+    {
+        return $this->transform($from);
+    }
```
klees: Use `DeriveInvokeFromTransform` instead.

src/Refinery/KindlyTo/Transformation/NullTransformation.php
```diff
@@ -0,0 +1,45 @@
...
+
+    public function accepts($value) : bool
+    {
+        return is_null($value) || (is_string($value) && trim($value) === '');
+    }
```
klees: We don't need that IMO. This is something only `Constraint`s need but not defined on the `Transformation`-interface.

src/Refinery/To/Transformation/StringTransformation.php
```diff
@@ -38,4 +38,9 @@ public function __invoke($from)
     {
         return $this->transform($from);
     }
+
+    public function accepts($value) : bool
+    {
+        return is_string($value);
+    }
```
klees: I don't think we need that. Am I wrong?

src/Refinery/ByTrying.php
```diff
@@ -0,0 +1,61 @@
...
+        };
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function __invoke($from)
+    {
+        return $this->transform($from);
+    }
```
klees, line 53: Use `DeriveInvokeFromTransform`.

tests/Refinery/ByTrying/ByTryingTransformTest.php
```diff
@@ -0,0 +1,111 @@
...
+            'empty string3' => ['  ', null],
+            'null' => [null, null],
+            'string' => ['str', self::ERROR],
+            'int' => [1, 1],
+            'negative int' => [-1, -1],
+            'zero' => [0, 0],
+            'array' => [[], self::ERROR],
+            'bool (false)' => [false, self::ERROR],
+            'bool (true)' => [true, self::ERROR]
+        ];
```
klees, line 40: Pleasuse PHPUnits dataproviders here (and also in the other test in this case).

src/Refinery/ByTrying.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    public function transform($from)
+    {
+        foreach ($this->transformations as $transformation) {
+            if ($transformation->accepts($from)) {
```
klees: `accepts` is no method on the `Transformation`-interface, so we cannot use it here. I think we could just `applyTo` instead and use the first result that is `ok`.

src/Refinery/ProblemBuilder.php
```diff
@@ -0,0 +1,76 @@
...
+namespace ILIAS\Refinery;
+
+use ILIAS\Refinery\ConstraintViolationException;
+
+trait ProblemBuilder
+{
+    /**
+     * @var string|callable
+     */
+    protected $error;
```
klees: Maybe we should add this as an abstract method here, to force our users to always give an error. Just having a property makes it possible to just forget to set this.

klees (changes requested):
Hi @nhaagen,

this looks mostly fine, special kudos for factoring out that `ProblemBuilder`-trait. Looks useful.

I think there are some confusions regarding `Transformation` and `Constraint` that show in your code. The relation is like this: every `Constraint` is a `Transformation` but not the other way round. A `Constraint` is a special kind of `Transformation` that does not actually change values (i.e. transforms them) but rather only looks into the values and throws (or creates `Error`s) if the values do not conform to some rule. This allows to have a more specific base class for constraints where we can define more methods than we would want for generic `Transformation`s.

Thanks for working on this so carefully!!!

Best regards!

mjansenDatabay:
@nhaagen LGTM (except the tiny changes @klees already requested). Indeed kudos for the `ProblemBuilder` trait.

src/Refinery/ProblemBuilder.php
mjansenDatabay: Something like `abstract protected getError() : string|callable;` or `abstract protected getDefaultError() : string|callable;`  (I know these return types do not work, yet)? I would appreciate this.

nhaagen:
Thank you both very much for your comments and hints.
I think, I got them all.

Best regards.

klees:
Thanks @nhaagen!

PR #2909, opened 2020-09-21
SPRP/Permissions: filter available users by OrgU

nhaagen:
https://mantis.ilias.de/view.php?id=28860

filter shown users by orgu


klees (commented):
Hi @nhaagen,

I think this will make the members table empty for users that have the permission "Manage Members" at the Study Programme but do not have any position based access to individual members. Am I right?

The thing is: If a user has "Manage Members" on the Study Programme, he is supposed to see all members in the table, regardless of any position. If a user hasn' got "Manage Members" he is supposed to see the users where he has the according permission via his position. I think the latter now works, but the former doesn't.

Please check...

Best regards!

klees:
Picked to trunk and release_7 as well.

PR #2910, opened 2020-09-22
LSO: check availability, update LP (fixes #28936, #28934)

nhaagen:
https://mantis.ilias.de/view.php?id=28936
https://mantis.ilias.de/view.php?id=28934

### DIFF
--- a/Modules/LearningSequence/classes/Player/class.ilLSPlayer.php
+++ b/Modules/LearningSequence/classes/Player/class.ilLSPlayer.php
@@ -19,6 +19,7 @@ class ilLSPlayer
     const LSO_CMD_GOTO = 'lsogoto'; //with param ref_id
     const LSO_CMD_SUSPEND = 'lsosuspend';
     const LSO_CMD_FINISH = 'lsofinish';
+    const LSO_CMD_CONTINUE = 'lsostay';
 
 
     public function __construct(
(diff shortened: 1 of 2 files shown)


Modules/LearningSequence/classes/Player/class.ilLSPlayer.php
```diff
@@ -92,11 +100,15 @@ function ($item) {
...
                 $next_item = $this->getNextItem($current_item, $param);
+                if ($next_item->getAvailability() !== \ILIAS\UI\Component\Listing\Workflow\Step::AVAILABLE) {
+                    $next_item = $current_item;
+                }
                 break;
             case self::LSO_CMD_GOTO:
                 list($position, $next_item) = $this->findItemByRefId($param);
                 break;
-            default: //view-internal / unknown command
+            case self::LSO_CMD_CONTINUE:
```
klees: Where does this new command originate from? I do not see that we set this somewhere...

klees (commented):
Hi @nhaagen,

frankly, this is hard to understand. I cannot really relate this to the bugs you linked. Please use the extended commit message to elaborate what happens and how this fixes the problem. Is this something we will want to pick to release_6 as well?

Thanks!

Modules/LearningSequence/classes/Player/class.ilLSPlayer.php
nhaagen: sorry, this was a fragment I forgot to remove.

nhaagen:
Hi @klees ,
I understand, those are tricky ones...
28934:
	retrieve LP immediately, fall into interval afterwards;
	if a LP change occurs BEFORE first javascript-call, the page is not being updated
	for the change is not recognized as such.
	Calling "lso_checkLPOfObject" directly initializes the correct progress at JS-level.

28936:
	If an LSO-Item is reached via the next-command (as it most often is), a page-reload
	will tell the LSO to go to the next item.
	If the item is not available, e.g. by conditions, the current (or last accessible) item should be shown.

Since there are quite general differences between 5.4 and 6 concerning the LSO, I'll check (and fix) seperatedly.

PR #2911, opened 2020-09-23
Fix: #28888 - decouple focussing of main-container 

nhaagen:
Fixes https://mantis.ilias.de/view.php?id=28888,
mainbar adjustments broken in Chrome.

Should also go down to 6.

### DIFF
--- a/src/UI/templates/js/Page/stdpage.js
+++ b/src/UI/templates/js/Page/stdpage.js
@@ -71,5 +71,9 @@ il.UI = il.UI || {};
 	})($);
 })($, il.UI);
 il.Util.addOnLoad(function() {
-    $("main").attr("tabindex", -1).focus();
+	window.setTimeout(
+		function(){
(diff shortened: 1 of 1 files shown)


Amstutz (approved):
Hi @nhaagen 

setTimeout is almost as horrible as !important. However I guess this is needed here, but am still interested in the specifics. Is some other element claiming focus *after* the page has been completely loaded? If so, why? If not, why is this needed then?

@klees feel free to merge, since I atm I see no better option.

klees:
Hi @Amstutz,

frankly, I do not understand why this fixes the problem. Nils basically found this by bisecting the codebase. I expect suspect this is some implementation detail of chrome kicking in here, so we did not bother to investigate this thoroughly. We introduced that focussing change not long ago to fix an a11y keyboard scroll ticket. This again reinforces that a11y is indeed a complicated topic which cannot simply be implemented retroactively and needs to be tested in context.

Best regards!

PR #2912, opened 2020-09-23
[Mantis:28923]: remove events without start-time from ical export

okaufman:
Sometimes there is no starting date set for a course. The constructor of ilCalendarEntry will still create an instance of ilDateTime.
This would solve the problem by checking if getStart() returns nothing, an empty string or "<br>"
https://mantis.ilias.de/view.php?id=28923


mjansenDatabay:
Hi @smeyer-ilias ,

we stumbled over this PR when we checked for old PRs at the meeting of the Technical Board. Could you please have a look at it and merge or give your reasons for not merging it. Thx!

smeyer-ilias:
Thanks for fixing this. I commited a slightly different fix in the  Release >= 5.4

PR #2913, opened 2020-09-23
FIX #28876: SPRG, role-based Permissions

nhaagen:
accept role-based permissions as well as OrgU-Permissions for members
https://mantis.ilias.de/view.php?id=28876

### DIFF
--- a/Modules/StudyProgramme/classes/class.ilObjStudyProgrammeMembersGUI.php
+++ b/Modules/StudyProgramme/classes/class.ilObjStudyProgrammeMembersGUI.php
@@ -950,18 +950,16 @@ class ilObjStudyProgrammeMembersGUI
         );
     }
 
-    public function getLocalMembers(): array
+    public function getLocalMembers() : array
     {
         return $this->object->getMembers();
     }
 
-    public function isOperationAllowedForUser(int $usr_id, string $operation): bool
+    public function isOperationAllowedForUser(int $usr_id, string $operation) : bool
     {
-        $ret = $this->position_based_access->isUserAccessibleForOperationAtPrg(
-            $usr_id,
-            $this->object,
-            $operation
-        );
+        $ret = $this->position_based_access->isUserAccessibleForOperationAtPrg($usr_id, $this->object, $operation)
+            || $this->access->checkAccess("manage_members", "", $this->object->getRefId());
+
         return $ret;
     }
 }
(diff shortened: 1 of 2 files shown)


Modules/StudyProgramme/classes/class.ilObjStudyProgrammeMembersGUI.php
```diff
@@ -950,18 +950,16 @@ public function mayManageMembers() : bool
...
-    public function isOperationAllowedForUser(int $usr_id, string $operation): bool
+    public function isOperationAllowedForUser(int $usr_id, string $operation) : bool
     {
-        $ret = $this->position_based_access->isUserAccessibleForOperationAtPrg(
-            $usr_id,
-            $this->object,
-            $operation
-        );
+        $ret = $this->position_based_access->isUserAccessibleForOperationAtPrg($usr_id, $this->object, $operation)
+            || $this->access->checkAccess("manage_members", "", $this->object->getRefId());
```
klees: Let's switch the operators of the `||` and do the check via the RBAC first. `||` has short circuit behaviour in PHP. That means, if for `a || b` the `a` already evaluates to true, `b` is never evaluated. Since RBAC checks are simpler, this will save time here. Also: Please just return the result instead directly, instead of assigning it to `$ret` before returning this then afterwards,

Modules/StudyProgramme/classes/class.ilObjStudyProgrammeMembersGUI.php
```diff
@@ -950,18 +950,16 @@ public function mayManageMembers() : bool
...
 
-    public function isOperationAllowedForUser(int $usr_id, string $operation): bool
+    public function isOperationAllowedForUser(int $usr_id, string $operation) : bool
     {
-        $ret = $this->position_based_access->isUserAccessibleForOperationAtPrg(
-            $usr_id,
-            $this->object,
-            $operation
-        );
+        $ret = $this->position_based_access->isUserAccessibleForOperationAtPrg($usr_id, $this->object, $operation)
```
klees: Does `isUserAccessibleForOperation` check if the the position system is activated for the Study Programme already? If not, we would lose this check, right?

klees (commented):
Hi @nhaagen,

thanks for looking into this. This is hard stuff, so I have questions =)

Best regards!

Modules/StudyProgramme/classes/class.ilObjStudyProgrammeMembersGUI.php
nhaagen: I think it  does; 
ilOrgUnitPositionAccess::filterUserIdsByPositionOfCurrentUser

        // If context is not activated, return same array of $user_ids
        if (!$this->set->getObjectPositionSettingsByType($this->getTypeForRefId($ref_id))->isActive()) {
            return $user_ids;
        }

PR #2916, opened 2020-09-24
Fix #28873, SPRG, Dash-overview does not count crss on root level

nhaagen:
https://mantis.ilias.de/view.php?id=28873


klees (changes requested):
Hi Nils,

could you just write some simple tests for this method? This should be simple if you factor out `ilObjStudyProgramme::getAllChildren` and `ilObjStudyProgramme::getInstanceByRefId` to methods on the class. I think this will be worth it, regarding the various branches in that algorithm here.

Best regards!

klees:
Also picked to trunk.

PR #2917, opened 2020-09-24
fix 28814 - no language selector, if only one language is available

nhaagen:
https://mantis.ilias.de/view.php?id=28814


chfsx:
@nhaagen thanks a lot! I'm not sure who to assign this, the rpovider comes from "Services/Language" which has @matthiaskunkel in the maintenance.json (but I think he is maintainer of the language-files and not the implementation itself)

matthiaskunkel:
I have assigned this PR to @fneumann who takes care of the technical backend of the language support in ILIAS.

fneumann:
The change looks fine to me but the he changed class belongs to the Init service which is maintained by @ smeyer-ilias .
Stefan, could you please habe a look and merge it?

PR #2919, opened 2020-09-25
Forum: Fix 28972 and partly 28010

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=28972 and https://mantis.ilias.de/view.php?id=28010 (partly)


mjansenDatabay:
@yvseiler FYI

PR #2924, opened 2020-10-05
Fix for 0028884 in ILIAS 6

rsheer:
Additional to the bugfix, I removed a variable which was unused anyway. This small cleanup was needed to make the bugfix possible.

This bugfix is in git not compatible with the branch for ILIAS 5.4. I will create a separate PR for the ILIAS 5.4 fix. The 2 commits in this PR should however be compatible to the trunk branch.


kergomard:
Thx a lot @rsheer !

PR #2925, opened 2020-10-05
Fix for 0028884 in ILIAS 5.4

rsheer:
This bugfix is in git not compatible with the branch for ILIAS 6 and trunk. I will create a separate PR which is meant for ILIAS 6 and the trunk-branch.


kergomard:
Thx a lot @rsheer !

PR #2926, opened 2020-10-06
UI/Form - Improvements: Cancel-Button 

nhaagen:
Dear All,
this PR  features 

- **a custom-label for submit**
   As mentioned in the rules already, it is OK to re-label the submit button. 
   There was no way to do that, though.
- **optional cancel-button**
   With a form embedded in some workflow, or the explicit rule to avoid top-bars, a "cancel" gets handy to get back 
   without submitting
- **optional top-buttons**
   For very short forms (such as login, accept ToC,..), it looks sort of exaggerated to have a submit at the top as well...

We adjusted the [rules](https://github.com/ILIAS-eLearning/ILIAS/pull/2926/files#diff-239223547575f8840330e9efc4ead806) accordingly.
This is a standard form with all mentioned additions:

Screenshot from 2020-10-06 16-22-19

Please kindly consider.
Best regards,
Nils and Oli

### DIFF
--- a/src/UI/Component/Input/Container/Form/Factory.php
+++ b/src/UI/Component/Input/Container/Form/Factory.php
@@ -9,7 +9,6 @@ namespace ILIAS\UI\Component\Input\Container\Form;
  */
 interface Factory
 {
-
     /**
      * ---
      * description:
@@ -18,10 +17,13 @@ interface Factory
      *      configuring objects or services.
      *   composition: >
      *      Standard forms provide a submit-button.
+     *      Additionally, they MAY provide a cancel-button.
      *   effect: >
      *      The users manipulates input-values and saves the form to apply the
      *      settings to the object or service or create new entities in the
      *      system.
+     *      Operating the cancel-button will take the user back to the situation
+     *      before calling the form.
      *
      * rules:
      *   usage:
@@ -39,10 +41,16 @@ interface Factory
      *        changed, then you MUST propose it to the JF.
      *     4: >
      *        On top and bottom of a standard form there SHOULD be the “Save” button for the form.
-     *     5: >
-     *        In some rare exceptions the Buttons MAY be named differently: if “Save” is
-     *        clearly a misleading since the action is more than storing
+     *        When the form is very small (e.g. accepting Terms of Condition, or username/password only,
+     *        the upper button MAY be omitted.
+     *   interaction:
+     *     1: Operating the cancel button MUST NOT change the system in any way.
+     *   wording:
+     *     1: >
+     *        In some rare exceptions the Buttons MAY be named differently, e.g.
+     *        if “Save” is clearly a misleading since the action is more than storing
      *        the data into the database. “Send Mail” would be an example of this.
+     *     2: The cancel-button, however, MUST NOT be labelled other than “Cancel”.
      *
      * ---
      *
--- a/src/UI/Component/Input/Container/Form/Standard.php
+++ b/src/UI/Component/Input/Container/Form/Standard.php
@@ -1,19 +1,56 @@
-<?php
+<?php declare(strict_types=1);
 
 /* Copyright (c) 2017 Richard Klees <richard.klees@concepts-and-training.de> Extended GPL, see docs/LICENSE */
 
 namespace ILIAS\UI\Component\Input\Container\Form;
+
+use ILIAS\Data\URI;
 
 /**
  * This describes a standard form.
  */
 interface Standard extends Form
 {
-
     /**
      * Get the URL this form posts its result to.
      *
      * @return    string
      */
     public function getPostURL();
+
+    /**
+     * Set the URL that is called when the user clicks 'cancel'.
+     * Setting an cancel-url will enable an accroding button in the form.
+     *
+     * @return    Standard
+     */
+    public function withCancelURL(URI $url) : Standard;
+
+    /**
+     * Get the URL that is called when the user clicks 'cancel'.
+     */
+    public function getCancelURL() : ?URI;
+
+    /**
+     * Labels the submit-button of the form.
+     *
+     * @return    Standard
+     */
+    public function withSubmitLabel(string $label) : Standard;
+
+    /**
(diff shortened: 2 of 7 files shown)


klees:
Hi @nhaagen and Oliver!

Did you try to find a rule for the cancel button that tries to pin a little more when to use it (and when not to use it)? Currently, you are stating this in the description part of the KS entry, and IMO weakly so. I think, the "Cancel"-Button only makes sense in situations, where I cannot go back to where I was before via a reload or similar "standard"-means, like closing a modal, deselecting something, etc.. I mean: In the settings of the course, what would I want to "cancel" there? So I think we somehow need to include the notion of something like a workflow here, and the "Cancel"-Button only makes sense when I am in some workflow and want to cancel it. Think, maybe, about the self-registration on the login page. You start some sort of self registration workflow where cancel drops you where you started that workflow (on the front page). Does this make sense somehow? Could you try to derive a rule here?

Something similar IMO goes for the "bottom only buttons". I do not think that this really about the length of the form but rather a question of: do I need the user to fill all fields and do I expect him to look through the complete form? This also feels related to that workflow notion we could well use for the cancel button. I feels logical that something like the self registration form only has a submit button below the form, I will need to go through the form anyway. This form is long, but a button on the top won't be very useful.

Please put some more thought into the rules regarding the cancel-button and the bottom only buttons. As always I'll be glad to be a sparring partner =)

Best regards!

kergomard:
Hi you all and thank you very much tackling this!

> I think, the "Cancel"-Button only makes sense in situations, where I cannot go back to where I was before via a reload or similar "standard"-means, like closing a modal, deselecting something

Sorry to weigh in, but I think we are touching an important discussion here. I agree that the rules should be sharpened, but I believe the cancel button should be available in all cases where the immediate website context of the form doesn't offer a way to leave the form without saving. I don't think we should rely on browser means to do this (e.g. reloading) and I don't think we should rely on menus for leaving the form. I probably also don't think the closing button of a modal is enough (I hate having to click on the finicky little bugger of a cross, when modals can't be closed by clicking outside and clicking outside is a not so easily to learn pattern). I have less of a clear opinion though.

In any case I would have a strong preference for having a MUST or SHOULD statement for the cancel-button. Where the cancel button brings you back to the context you very in before opening the form without saving, thus providing a way out without fearing to save anything accidentally.

nhaagen:
Hi,
first of all, thank you for your comments; and you are right, this is worth digging deeper.
For now, I'd just like to add some more usecases to the discussion, since I do not have
a proper solution/suggestion yet.

- (Course) Settings / property forms
   Those forms are used for both showing and editing settings;
   IMO it must have top-buttons, and cancel is useless.

- Login
   There are only mandatory fields.
   Top-Button should be ommited, since the user will have to go through the entire form.
   There cannot be a back-button, since there is no workflow to cancel.
   No properties are being changed.

- Plugin- or Cron-Config
   Top- and bottom-buttons should be there, and the forms are about settings/properties.
   However, I think that the forms would profit from a back-button:
   If there is but one tab, a "cancel" in the form could serve the same purpose as the "back"-link in the tab-bar.
   If there is more than one tab, the link could go back to the Plugin's overview.
   
- Page Editor
   The page editor wants to have some very small forms in the 
   tools-slate (like "give number of cols/rows for a table").
   When used as a "settings-form", the fields are prefilled, and 
   there is not really a process to cancel; however, _inserting_ a
   table will use the same form, and then "cancel" would abort
   inserting a table.
   One line of buttons (cancel/insert) is enough though, as not to clutter the slate.

kergomard:
@nhaagen : Fine with me, ...with one little change: For property forms we should IMHO keep current behavior with a cancel button. This way we can postpone the whole discussion. I agree with you that they need top-buttons (current behavior).

alex40724:
Please do not change the order, or current guidlines make Cancel the button on the right, see for example https://docu.ilias.de/goto_docu_wiki_wpage_978_1357.html

Imo there is no point in opening this discussion again, see e.g. https://www.nngroup.com/articles/ok-cancel-or-cancel-ok/

nhaagen:
Hi @alex40724 , sorry, I was not aware of that discussion. I will probably never get used to "cancel" on the right...
I will update the code, but meanwhile leave the example.

alex40724:
>  and the "Cancel"-Button only makes sense when I am in some workflow and want to cancel it

Yes.

> In any case I would have a strong preference for having a MUST or SHOULD statement for the cancel-button. Where the cancel button brings you back to the context you very in before opening the form without saving, thus providing a way out without fearing to save anything accidentally.

Imo all forms available under "tabs" (settings screens) SHOULD NOT get a cancel button. What should happen if I click through the tabs and then start to hit "Cancel" subsequently? Should ILIAS back-track the "click stack"? I know that there are still several settings screens showing a "Cancel" button, but I always wonder where they go.

alex40724:
> Hi @alex40724 , sorry, I was not aware of that discussion. I will probably never get used to "cancel" on the right...
> I will update the code, but meanwhile leave the example.

This is just my personal opinion. No need to change things immediately. :-)

klees:
Hi everybody!

Seems as if this deserves some discussion. Thanks for your input so far! If I understood Oliver correctly, there is no need to hurry here, so I prepared a doodle:

https://doodle.com/poll/hsbawb4zig7ytfcu

Please try to participate until monday noon, so I can announce the appointment on the JF.

Thanks and best regards!

klees:
Hi everyone,

I closed the Doodle and chose **2020-11-05, 10:00 - 11:30**. Please join [this group on skype](https://join.skype.com/ewstln67Acho) if you want to participate in the discussion!

Best regards!

klees:
Minutes of our discussion in October are to be found here: https://docu.ilias.de/goto_docu_file_9803.html

Currently this PR does not reflect what we agreed on in our meeting.

Amstutz:
I see no progress here for some time now. Further the PR seems out of date. Since we try to keep the list of open PRs small and up to the current state of the discussion, I close this down ftm. Feel free to open this PR as soon as there is any update in this matter.

PR #2928, opened 2020-10-07
Fix regex for PCT encoding in Data/URI

kergomard:
Following [rfc3986](https://tools.ietf.org/html/rfc3986#section-2.1) we should be more tolerant here. I would see this as a bug fix and ask to cherry pick to 5.4 and 6.

### DIFF
--- a/src/Data/URI.php
+++ b/src/Data/URI.php
@@ -49,7 +49,7 @@ class URI
     const ALPHA = '[A-Za-z]';
     const DIGIT = '[0-9]';
     const ALPHA_DIGIT = '[A-Za-z0-9]';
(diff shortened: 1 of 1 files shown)


klees:
Thanks @swiniker, I will pick this to the other active branches.

klees:
Sorry, a bit fast here. Could you back this with an automated test, please? trunk will be enough imo, and no new PR needed for me. Just push =)

kergomard:
Sorry for missing this! Something like this: https://github.com/ILIAS-eLearning/ILIAS/commit/111feb802a180c2cd89c08ceff3f901a2087d2b8 ?

PR #2929, opened 2020-10-07
exclude plugin vendor directories from autoload

theodortruffer:
Change to PR #2758 to prevent conflicts in duplicate composer libraries (see https://github.com/ILIAS-eLearning/ILIAS/pull/2758#issuecomment-704917682).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -83,7 +83,8 @@
 		"exclude-from-classmap": [
 			"./Services/Migration",
 			"./*/*/lib", "./*/*/test",
-			"./Services/CAS/phpcas"
+			"./Services/CAS/phpcas",
+			"./Customizing/**/vendor"
 		]
 	},
 	"extra": {


matthiaskunkel:
Jour Fixe, 12 OCT 2020 : We highly appreciate this suggestion and accept the PR for ILIAS 7 / trunk. Please merge.

composer.json
```diff
@@ -83,7 +83,8 @@
 		"exclude-from-classmap": [
 			"./Services/Migration",
 			"./*/*/lib", "./*/*/test",
-			"./Services/CAS/phpcas"
+			"./Services/CAS/phpcas",
+			"./Customizing/**/vendor"
```
klees, line 87: Hi @theodortruffer,

just for curiosity: Is there any reason that there are two `**` here?

Best regards!

theodortruffer: Hey Richard

Single Asterisk matches everything but a slash and double asterisk everything, including slashes. So * is just one directory, ** can be a whole subpath. See https://getcomposer.org/doc/04-schema.md#exclude-files-from-classmaps

PR #2932, opened 2020-10-12
Fix #28809: Enable/disable Accessibility Control Concept

tfamula:
Fix for [28809](https://mantis.ilias.de/view.php?id=28809)

### DIFF
--- a/Services/Accessibility/classes/class.ilAccessibilityControlConceptGUI.php
+++ b/Services/Accessibility/classes/class.ilAccessibilityControlConceptGUI.php
@@ -45,6 +45,7 @@ class ilAccessibilityControlConceptGUI
         $lng = $DIC->language();
         $http = $DIC->http();
         $user = $DIC->user();
+        $settings = $DIC->settings();
         $accessibilityEvaluation = $DIC['acc.document.evaluator'];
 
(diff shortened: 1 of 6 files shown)


Services/Accessibility/classes/class.ilAccessibilityControlConceptGUI.php
```diff
@@ -112,11 +114,14 @@ protected function showControlConcept()
             $document = $this->accessibilityEvaluation->document();
             $tpl->setVariable('ACCESSIBILITY_CONTROL_CONCEPT_CONTENT', $document->content());
         } else {
+            $mails = (ilAccessibilitySupportContacts::getMailsToAddress() != "")
+                ? ilUtil::prepareFormOutput(ilAccessibilitySupportContacts::getMailsToAddress())
+                : $this->settings->get("admin_email");
```
tfamula, line 119: Additionally, I added a fallback here if link for Accessibility Control Concept is enabled, but no Accessibility Point of Contact is defined, so the user does not run into a mailto without recipient.

alex40724 (approved):

klees:
Thx y'all.

PR #2939, opened 2020-10-14
Refinery/ReadingTime Trafo: Ignore comment nodes in XHTML strings

mjansenDatabay:
When the reading time is determined by the given
text string, XHTML comments MUST not be considered.
Therefore these document nodes have to be skipped
when counting words of a (XHTML) string.

https://mantis.ilias.de/view.php?id=29038

### DIFF
--- a/tests/Refinery/String/Transformation/EstimatedReadingTimeTest.php
+++ b/tests/Refinery/String/Transformation/EstimatedReadingTimeTest.php
@@ -4,7 +4,6 @@ namespace ILIAS\Tests\Refinery\String;
 
 use ILIAS\Refinery\Factory;
 use PHPUnit\Framework\TestCase;
-require_once 'libs/composer/vendor/autoload.php';
 
 class EstimatedReadingTimeTest extends TestCase
 {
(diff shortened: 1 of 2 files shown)


mjansenDatabay:
@klees A short review is highly appreciated.

tests/Refinery/String/Transformation/EstimatedReadingTimeTest.php
```diff
@@ -1,10 +1,10 @@
 <?php declare(strict_types=1);
 /* Copyright (c) 1998-2019 ILIAS open source, Extended GPL, see docs/LICENSE */
-
 namespace ILIAS\Tests\Refinery\String;
 
 use ILIAS\Refinery\Factory;
 use PHPUnit\Framework\TestCase;
+require_once 'libs/composer/vendor/autoload.php';
```
klees: I wonder why we need this. What command do you use to run the tests? This looks like something we should solve in general.

klees (approved):
Hi @mjansenDatabay,

LGTM in general. There's one thing I wonder about, though. I marked it below. This is no blocker to merge this for me, but seems to hint at something we would want to look into. Any opinion.

Best regards!

tests/Refinery/String/Transformation/EstimatedReadingTimeTest.php
mjansenDatabay: Argggghh

mjansenDatabay: Thx @klees , good catch. I just removed the line.

PR #2942, opened 2020-10-15
#29047 - fix for copy of content page object

fneumann:
When you copy a content page object that has for example H5P content, then the copied object refers to the same H5P content. If you modify it on the copied object, then the original gets modified, too. If you delete it on the copied page, then the editor in the original shows that that content does not exist anymore. This is fixed by the PR.


mjansenDatabay:
Thx @fneumann

PR #2943, opened 2020-10-16
BUGFIX[29052]

iszmais:
https://mantis.ilias.de/view.php?id=29052

Dieser PR verbessert die Ansicht der Favoriten im Hauptmenü und steigert die User-Satisfactory und Barrierefreiheit.

Vorher:
image

Nacher:
image


iszmais:
https://mantis.ilias.de/view.php?id=29052

alex40724:
Thanks for the PR. The use of the listing panels has been a decision of the page layout revision project. You may present your suggestion into the current page layout revision "part ii" meetings. There are other related discussions (e.g. the icon discussion).

Amstutz:
Hi @iszmais 

Thx for your PR. Note that we migrate all of the rendering of HTML towards the UI Components: https://github.com/ILIAS-eLearning/ILIAS/tree/trunk/src . The Listing panel you are Looking at is also rendered as Listing Panel. 

It would help if you could point out clearly, which A11y issue you tackle with your PR. Feel free to suggest adjustement towards the respective UI Component, however note that you need to follow the above guidelines. E.g. if you propose an interface change we need to discuss this in JF, since various devs will be affected. 

As @akill pointed out, it might also be a good approach to present your idea/change in the PLR III group. Further information here:  https://docu.ilias.de/goto_docu_grp_8739.html

Thx @akill for fixing the empty list issue. From my perspective this PR should be closed. However, I will not do it, since it is not pushed towards any of my code (except maybe the less part).

Thx

alex40724:
I reassign it to @iszmais to take over the discussion part. If the groups accepts this idea, it may be reassigned to me as a maintainer. Personally I hope that the group will have a look at
a) the mixing of action and object item (configure as first element) and
b) the use of different icons, which is a general ongoing discussion.

I will not accept anything that looks like "another intermediate step" to me.

@iszmais thanks for pushing this forward. I definitely agree that the current look and feel is not consistent in the menu. But there are different ways to tackle this.

klees:
Hi @iszmais,

I hereby offer my help to find out how this could be discussed in the community.

Feel free to ask me anytime!

iszmais:
Thx!

Due to some time issues i will not be able to look further into this before 4. annual quarter.
I will get in touch with you when in can find the time.

Greeting,

Ingmar

kergomard:
@iszmais 

The forth quarter has come and gone. Any chance you can take a look at this?

Thanks you very much,
Stepan in the name of the Technical Board

Amstutz:
I still believe the suggestion to be of value. In the meanwhile a a11y, probably one that @iszmais was initially refering to showed up in an a11y audit. We have an empty heading in the list.

I will push this to the JF to reach a decision on how to push forward. I would propose this or a related version of it as usability fix for: https://mantis.ilias.de/view.php?id=32356

Edit: Note that the non-interactive titles between the section could be implemented as Horizontal Deviders inside the combined slates.

matthiaskunkel:
Jour Fixe, 02 MAY 2022 : We discussed the PR and its functional and visual impact. We would like to keep the subheadlines (for place / type of the favourites) and have to get rid of the empty headline tag due to accessibility. Timon will create a new PR and discuss it with the product manager. 
On the long run, we should improve the UI component (combined slate) and display a kind of 'settings icon' for Favourites (and have one line less in the slate).

Amstutz:
Still open todo, however this is tracked through mantis ticket. The PR will not be merged as is. A new suggestions needs to be made here by @Amstutz .

PR #2945, opened 2020-10-16
UI/Symbol/Icon/Standard - Introduction of PageEditorIcons

ezenzen:
This PR provides the ILIAS page editor with new icons. These serve as placeholders for page elements if they are added by the user.
These icons will be inserted into the KitchenSink as standard icons. The KS description of the standard icons should be adapted so that the Milos Outlined Icon version is defined as the standard icon. The "filled version" is available as an additional version. This allows (for the ILIAS page editor icons as well as for future ILIAS icons) to provide only one outlined version at a time.

With this PR the following adjustments are made:
- Description of the KS article Symbol > Icon > Standard
- 27 new icons as outlined version
- 27 new language variables for de and en

Feedback is warmly welcome.


yvseiler:
**Page Editor Icons**
Bildschirmfoto 2020-10-16 um 15 55 08

**Page Editor Mockup (example insert media)**
Mockup_PageEditor_Placeholder_Media

ezenzen:
**KS Entries for Page Editor Icons**

Screenshot_2020-10-16 ILIAS Documentation
Screenshot_2020-10-16 ILIAS Documentation(1)

klees (changes requested):
Hi @ezenzen and @yvseiler

thx a lot for this PR! We are amazed and astonished that you managed to provide a full implementation, while claiming not to have any particular PHP experience. Big kudos!

We trust your judgement of changing to the outlined Icons as default version as well as the selection of the particular Icons for the Page Editor along with the provided description. 

Nevertheless, there are questions:
- [ ] Outlined as Default: We guess that this does have no further implications for existing usages of Icons for ILIAS 7, correct? If so, we believe that it is better to omit the change in the Description of the Standard Icon. The description should be reflected in the code. There, „Filled is still the default“. Changing this default would require further refactoring/discussion on usages for existing Icon usages, for which the time falls short. However, feel free to add an entry to our roadmap for changing this for 8. The only thing that would need change for the moment in your PR is the part: „The Standard icon is reduced to its minimal form, expressing essential characteristics.“ to -> „The outlined icon is reduced to its minimal form, expressing essential characteristics.“. The proposed rule seems to be at work already and can stay.
- [ ] New Icons vs Existing Icons: You provide special versions of existing Icons, e.g. for Test in the context Page Editor. Why? Are those really different things? E.g., we would now have a Course Icon, and a Course Page Editor Icon. Couldn't we just use the same Icon in both contexts?

There is one thing we need to fix before merging:
- [ ] Currently the code requires BOTH, the outlined and the filled version. Otherwise, tests will fail and one can cause fatals by asking for the filled version of an Icon, where there is none (currently, everytime such an icon is rendered, without setting it explicitly to outlined). We see the point you are making and that it does not make sense to provide another version of the icon, if there is none required. We would fix this by just providing the filled version for those here as well with the same svg as the outlined, but adding in the description, under composition: „Note that in some cases, the outlined and the filled version might look the same“.

Thx
@klees and @amstutz

matthiaskunkel:
26 OCT 2020 : We highly appreciate this suggestion and accept the PR for ILIAS 7. Please add alt texts for all icons to support accessibility. We might change the icons for inserting test questions and list of courses in the future to improve their message (question is not a test). But these both icons won't be used before ILIAS 8. For ILIAS 7, we keep the "filled" Milos icons as default icons. Outline icons as defaults should be added to the roadmap.

Amstutz:
After discussing with @ezenzen on how to finish this up, we decided to close this one here and create a new PR were @Amstutz will resolve the mentioned issues. See: https://github.com/ILIAS-eLearning/ILIAS/pull/2972

PR #2946, opened 2020-10-16
BUGFIX[29054]

iszmais:
https://mantis.ilias.de/view.php?id=29054

Dieser PR verbessert die Ansicht der Tag-Ergebnisse im Hauptmenü und steigert die User-Satisfactory.

Vorher:
image

Nachher:
image

### DIFF
--- a/Services/Tagging/classes/class.ilTaggingSlateContentMenuGUI.php
+++ b/Services/Tagging/classes/class.ilTaggingSlateContentMenuGUI.php
@@ -1,0 +1,58 @@
+<?php
+
+/* Copyright (c) 1998-2019 ILIAS open source, Extended GPL, see docs/LICENSE */
+
+use ILIAS\Data\URI;
+
+/**
+ * Tagging slate UI
+ *
+ * @author killing@leifos.de
+ */
+class ilTaggingSlateContentMenuGUI extends ilTaggingSlateContentGUI
+{
+    /**
+     * show resources
+     *
+     * @param
+     * @return string
+     */
+    protected function showResourcesForTag()
+    {
+        // back button
+        $tag_cmd = $this->ctrl->getLinkTarget($this, "showTagCloud", "", true, false);
+        $items[] = $this->ui->factory()->button()->bulky(
+            $this->ui->factory()->symbol()->glyph()->back(),
+            $this->lng->txt("back"), "#")->withOnLoadCode(function ($id) use ($tag_cmd) {
+            return "$(\"#$id\").click(function() { 
+                il.Util.ajaxReplaceInner('$tag_cmd', 'il-tag-slate-container'); return false;
+            });";
+        });
+
+        $items[] = $this->ui->factory()->item()->standard(sprintf(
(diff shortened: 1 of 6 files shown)


Services/Tagging/classes/class.ilTaggingSlateContentMenuGUI.php
```diff
@@ -0,0 +1,58 @@
...
+            $this->ui->factory()->symbol()->glyph()->back(),
+            $this->lng->txt("back"), "#")->withOnLoadCode(function ($id) use ($tag_cmd) {
+            return "$(\"#$id\").click(function() { 
+                il.Util.ajaxReplaceInner('$tag_cmd', 'il-tag-slate-container'); return false;
+            });";
+        });
+
+        $items[] = $this->ui->factory()->item()->standard(sprintf(
+            $this->lng->txt("tagging_resources_for_tag"),
+            "<i>" . $_GET["tag"] . "</i>"
```
mjansenDatabay: Please check if this is vulnerable against any XSS attacks. If so, please use `ilUtil::stripSlashes()`.

iszmais: Thx, this was indeed not XSS safe.
Within fixing this i encountered that the tagging has problems with quotings in general (Tags with quote cant be resolved in the Menu). This fix does not fix this problems completly (tag will now be resolved but always with an empty result) but solves at least the security issues.

alex40724:
I fixed the exit issue. Please present the presentation at the PLR II meetings.

klees:
Hi @iszmais,

I hereby offer my help to find out how this could be discussed in the community.

Feel free to ask me anytime!

iszmais:
Thx!

Due to some time issues i will not be able to look further into this before 4. annual quarter.
I will get in touch with you when in can find the time.

Greeting,

Ingmar

chfsx:
Please excuse my late feedback! From my point of view, it makes sense to carry out this conversion. Visually it looks right, on the code side I would have the following application/wish:

Until now the concerned `Providers` delivered so called `Complex` items, which can contain any HTML. With your conversion, however, corresponding `Items` would have to be used, which are also intended for this purpose. For example, the link list can be used as a `LinkList` item, regular links as `Link` items. The possible `Items` from the GlobalScreen can be found here: src/GlobalScreen/Scope/MainMenu/Factory/MainMenuItemFactory.php

kergomard:
@iszmais

We stumbled upon this PR in our TB meeting. Could you please take a look at this?

Thanks you very much,
Stephan

Amstutz:
Hi @iszmais 

Thx for this proposal. I believe it to be a step in the right direction. However, there is no open related issue around this anymore. If you want to proceed further with this, shortly explain, why this is important from a UI or Technical perspective and why this does not need an FR. 

In my role as TB Member checking requests that are open for a long amount of time, I close this for the moment. You are welcome to re-open if you wish to continue the discussion here.

Thx.

PR #2950, opened 2020-10-20
Bugfix: Remove include to invalid path

rsheer:
The cron-job "survey_notification" crashes on our instance (and others probably as well). The reason is an `include_once` to a class which was moved to another folder for ILIAS6. `ilSurveyEvaluationGUI` is now in the folder "Evaluation" instead of "classes".

Since we use autoloading and the class is included in the class map, I just removed the `include_once`-Statement. If you want me to do some more cleanup, I will remove the other includes in this method as well.


alex40724:
Thanks for the PR. This line has been removed in the meantime. However the PDF generation still suffers some issues...

PR #2952, opened 2020-10-21
column layout flex box grid support for row-flex class

catenglaender:
### Expected behavior
When 12 units of the Bootstrap column layout grid are exceeded, a column element is supposed to flow beneath and start a new row from the left.

### Issue:
Because Bootstrap 3 works with floats, a column in a subsequent row can get "stuck" on a long column in the previous row. This causes the new row to start to the right of the tall column, rather than from the leftmost edge where it should be. This throws off layouts and can cause unexpected results.

image

### Previous attempts

There have been previous attempts to fix this in 2017 by turning the float grid into a flex box grid: https://docu.ilias.de/goto_docu_wiki_wpage_4292_1357.html

The fix in this PR is based on this commented out code that was still left in delos.less. It requires the "row-flex" class to be added to the row container (which is still the case for all page editor column layouts). Otherwise the layout falls back onto the old float-method.

An issue that columns from nested grids would ignore their width settings stopped the previous implementation attempt. This has now been fixed. For this, it was necessary to redefine the bootstrap mixin building the "col-*" classes. Assigning the width at any later point caused issues with nested grids or would require to call many Bootstrap mixins inside of delos.less generating an unnecessarily large amount of CSS. All Bootstrap classes, mixins and variables still work within both contexts (only "row" class = float / "row-flex" class = flex box grid).

### Modernizing the grid system

While this approach only builds on top of Bootstrap 3, potentially, this could be a step towards building our own grid framework for Ilias. Updating to Bootstrap 4 or 5 would have major implications for many other areas and comes with a lot of bloat. Instead, it might be interesting to isolate the entire Bootstrap 3 grid code into our own grid library and start changing outdated und undesired parts like this one. New versions of Bootstrap and other grid frameworks could serve as inspiration, while we should be able to keep compatibility with all currently in-use mixins and classes.

### CSS not yet compiled

This PR does not include the compiled css, as the css.map is very likely to cause a merge conflict and needs to be replaced in its entirty. If the PR is accepted I can compile delos.less, so we have a most up to date delos.css.

### Screenshots

Before and after screenshots from cherry-picking this fix into release_6 after compiling delos.less:

**Before - problematic float behavior**

Ilias6 Column Flex Fix Inactive

**After - a clean new row with flex box**

Ilias6 Column Flex Fix Active


Amstutz:
@catenglaender : thx for the PR and the good documentation around it. However, I am not sure if we can handle this as "Improvement" PR without some related issue or feature request. I expect that to need some testing and discussion. See also the guidelines for PRs here: https://github.com/ILIAS-eLearning/ILIAS/blob/release_6/docs/development/contributing.md. 

@alex40724 : I also assigned you next to me, since you have rather tight coupling with the Page Editor to the grid. What do you think, do we need an FR for this?

klees:
Hi @Amstutz,

as far as we understand, the linked FR basically requests the behaviour that @catenglaender now implemented, this is why we are proposing this as PR. Last comment in the FR (to me) suggests, that there was merely a problem implementing this (with Bootstrap 3) before. This does not answer how we would want to handle this change, though.

Best regards!

Amstutz:
Hi @klees 

Thanks for the clarification. My bad, I was to quickly in browsing the linked FR above. After re-reading the FR and **all** the related comments bellow I agree. @catenglaender thx for this neat contribution, this seems to tackle an issue we carry around for some time. Your proposal seems much closer to what we have decided some years ago. I agree with the statement concerning **Modernizing the grid system** and the way forward with the grid system. 

However, if I understand things right, merging this would break with already productively styled pages created with the page editor in wikis/learning modules/repository etc.. That could create some havok. IMO we should discuss this in a mantis ticket. Probably we can all agree, that this is an mantis issue, since things do not work as specified / agreed upon.

So it would be great to:
- [x] Have a linked mantis issue to discuss the consequences for productive environments for this. I would then also push this to JF to give this change some echo chamber. 

Thx
@amstutz

klees:
Hi @Amstutz,

I created the ticket and set in on the JF agenda:

https://mantis.ilias.de/view.php?id=29071

I try to make @catenglaender to participate on monday, because, honestly, I'm not very much into the details here.

I want to point out, though, that we do not propose this for a productive version of ILIAS, but for the trunk, and intentionally so.

Best regards!

matthiaskunkel:
Jour Fixe, 26 OCT 2020 : We highly appreciate this suggestion and accept the PR for ILIAS 7. We do not want to merge this to already published stable releases to prevent unexpected side effects in the page layout.

catenglaender:
I rebased on top of current trunk and compiled delos.less as discussed with @klees and @Amstutz 
From my side this is ready for merging and testing. (edit: Never mind. My LESS compiler output seems to be different. I need to investigate that.)

All the best

catenglaender:
Oh nevermind, just noticed, it seems like my less compiler is giving a slightly different output throughout the entire CSS. Are these the correct best practices for compiling delos? (Node-less?) https://docu.ilias.de/goto_docu_pg_75028_367.html?lang=de

catenglaender:
@klees fixed the less compile.

Now the PR can be merged and tested, @Amstutz.

Thanks to all for the help with my first PR! :)

PR #2954, opened 2020-10-22
UI Head Info

chfsx:
This pull request suggests the announced UI component "Head Info".

A "Head Info" is currently being developed for the feature "Introduction of Administrative Notification", see https://docu.ilias.de/goto_docu_wiki_wpage_6359_1357.html

Summary:
- The rules can be found at https://github.com/ILIAS-eLearning/ILIAS/compare/trunk...studer-raimann:feature/7/ui-head-info-interfaces#diff-c8e66b8889d5aadf79a14acee5586c8c204a4629ed1a79f457dcba7dc3c91eb5R375
- New read variables can be found at https://github.com/ILIAS-eLearning/ILIAS/compare/trunk...studer-raimann:feature/7/ui-head-info-interfaces#diff-c8e66b8889d5aadf79a14acee5586c8c204a4629ed1a79f457dcba7dc3c91eb5R375
- Several adjustments to the standard page are needed, because the Head Infos are added in the upper part of the page. This PR contains the adjustments to the interface, less and the HTML file

If you want to see more or less, I can adapt the PR. An implementation is already done, so here are some screenshots. I would be happy to receive your feedback and work it into this PR.

### DIFF
--- a/src/UI/Component/Layout/Page/Standard.php
+++ b/src/UI/Component/Layout/Page/Standard.php
@@ -6,6 +6,7 @@ namespace ILIAS\UI\Component\Layout\Page;
 use ILIAS\UI\Component\Breadcrumbs\Breadcrumbs;
 use ILIAS\UI\Component\Image\Image;
 use ILIAS\UI\Component\JavaScriptBindable;
+use ILIAS\UI\Component\MainControls\SystemInfo;
 use ILIAS\UI\Component\MainControls\Mainbar;
 use ILIAS\UI\Component\MainControls\Metabar;
 use ILIAS\UI\Component\MainControls\ModeInfo;
@@ -97,4 +98,18 @@ interface Standard extends Page, JavaScriptBindable
 
 
     public function hasModeInfo() : bool;
+
+    /**
+     * @param SystemInfo[] $system_infos
+     * @return Standard
+     */
+    public function withSystemInfos(array $system_infos) : Standard;
+
+    /**
+     * @return SystemInfo[]
+     */
+    public function getSystemInfos() : array;
+
+
+    public function hasSystemInfos() : bool;
 }
--- a/src/UI/Component/MainControls/Factory.php
+++ b/src/UI/Component/MainControls/Factory.php
@@ -425,10 +425,10 @@ interface Factory
      *        provides such an interaction. In this case the user MUST be able to
      *        close the info in its context by clicking on the Close Glyph.
      *   accessibility:
-     *     1: >
-     *         The System Info informs about an important circumstance, which must be
-     *         recognizable in particular also for persons with a handicap. A
-     *         role="alert" MUST be added to a System Info.
+     *     1: Breaking System Infos MUST have a role="alert".
+     *     2: Important and neutral System Infos MUST have an aria-live="polite".
+     *     3: The headline MUST be referenced by aria-labelledby
+     *     4: The information MUST be referenced by aria-describedby
      * ----
      * @return \ILIAS\UI\Component\MainControls\SystemInfo
      */
--- a/src/UI/Component/MainControls/Factory.php
+++ b/src/UI/Component/MainControls/Factory.php
@@ -347,7 +347,6 @@ interface Factory
      *   purpose: >
      *     The Mode Info is a section on a page that informs the user that he is
      *     in a certain mode (e.g. in the preview as a member of a course).
-     *
      *   composition: >
      *     The Mode Info MUST contain a title explaining the mode.
      *     The Mode Info MUST contain a Close Button to leave the
@@ -355,10 +354,11 @@ interface Factory
      *   effect: >
      *      By clicking the Close Button, the user leaves the current
      *      (application wide) mode.
-     *
+     *   rivals:
+     *      System Info: >
+     *         use ModeInfo to indicate a certain state in a user context. The SystemInfo on the other hand informs about system-wide information.
      * context:
      *   - The Mode Info is used with the Standard Page.
-     *
      * rules:
      *   usage:
      *     1: The Mode Info is unique for the page - there MUST be not more than one.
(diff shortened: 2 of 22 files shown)


klees:
I cannot look into this in detail right now, but the screenshots look absolutely awesome!!

Amstutz (changes requested):
Hi @chfsx,

thanks a lot for this contribution. Looks great! Special thanks for taking some rule

- [x] For curiositie sake: There will be an accompanying implementation in the `GlobalScreen`, correct? It will take care about shuffeling the infos from components to the Page Layout, right? Will it also take care about storing the content and the state of the infos?
- [x] Close Button: If [there is no close button](https://github.com/ILIAS-eLearning/ILIAS/pull/2954/files#diff-c8e66b8889d5aadf79a14acee5586c8c204a4629ed1a79f457dcba7dc3c91eb5R386), how could the user dismiss the information? Or is this for cases where the info goes away after some time? What would be the use case for this and do we really want to provide no option to remove the message at all? Maybe this could be solved in the implementation of the persistence in some way, e.g. by having configurations for how users should be reminded liḱe `once`, `per login`, ...
- [x] There are some typos in the text. We omit those in this review. Do you like us to create a PR for those or can we push them directly to your repo?


- [x] SystemInfo: We are trying to find more semantically meaningful names for our components recently. We would suggest to do just that here and rename the component to `SystemInfo`. 
- [ ] Long Info: [Currently you propose replace text that is to long with a `More Glyph` entirely.](https://github.com/ILIAS-eLearning/ILIAS/pull/2954/files#diff-c8e66b8889d5aadf79a14acee5586c8c204a4629ed1a79f457dcba7dc3c91eb5R391) We think it would be nice to display the first line of the text as a sneak peak and then add the `More Glyph`. The length of the line is a moving target anyway, as we have to deal with different screen sizes, so we would need to have an according mechanism to dynamically add or remove the `More Glyph`.

- [x] RFC 2119: Please use [RFC 2119 language (MUST, ...)](https://tools.ietf.org/html/rfc2119) for the rules and not for the informative part of the Kitchen Sink Entry. It's just the other way round =)
- [x] Denotation: There seems to be some concept of "denotation" that we cannot find an explanation for. From the implementation and screenshot we gather, that this is used to give the message some kind of importance by color. Please explain what this is in the description of the component, and the according function or maybe just remove it. The description of the component says, that it should be used for "relevant and important information". This would mean we only need that "important" denotation (if we understand correctly) and would not want to bother our users with less important stuff.

Note, the above only concerns the interface. We had a quick look at the implementation as well, however this is incomplete. Maybe some of it helps when finishing up for the final review:
- [x] Less: Thanks for providing a set of variables for the new component as well as using the grid system. Do not introduce values for margin, font-size, height and padding directly in the less file. Try to use existing values or create variations of those in the variables.less file if necessary.
- [x] JS Log: Remove the lines logging tho the console.
- [x] JS Scoping Pattern: Use the scoping pattern to define the public interface of your js code. In your case, this will probably only be one function called by il.Util.addOnLoad (see e.g. . This is not perfect, but atm the minimal standard we have for js code. 
- [x] Example: At least one is needed.
- [x] Tests: Are needed.

Best regards!
@klees and @Amstutz

chfsx:
> For curiositie sake: There will be an accompanying implementation in the GlobalScreen, correct? It will take care about shuffeling the infos from components to the Page Layout, right? Will it also take care about storing the content and the state of the infos?

Yes there will be a GlobalScreen equivalent. But storing and providing items is a task of the components which provide items (as in any other cases of the GlobalScreen). The will be a service "Administrative Notifications" which is part tof the feature. this service allow to create and modify messages which are delivered via the GlobalScreen. Other components can provide own items (e.g. the Chat-Service could provide a HeadInfo for "Chat server is currently not available").

> Close Button: If there is no close button, how could the user dismiss the information? Or is this for cases where the info goes away after some time? What would be the use case for this and do we really want to provide no option to remove the message at all? Maybe this could be solved in the implementation of the persistence in some way, e.g. by having configurations for how users should be reminded liḱe once, per login, ...

Exaclty, there can be HeadIfos without  Close Button. In this case the delivering Component has to decide how long this HeadInfo is deplayed (e.g. as soon as the Chat-Server is available again, the message coud disappear). I will add a note to this in the rules!

> There are some typos in the text. We omit those in this review. Do you like us to create a PR for those or can we push them directly to your repo?

You can push them directly or as suggested changes using the github feature.

SystemInfo: We are trying to find more semantically meaningful names for our components recently. We would suggest to do just that here and rename the component to SystemInfo.

> absolutely fine with this, I'll change that ASAP 

Long Info: Currently you propose replace text that is to long with a More Glyph entirely. We think it would be nice to display the first line of the text as a sneak peak and then add the More Glyph. The length of the line is a moving target anyway, as we have to deal with different screen sizes, so we would need to have an according mechanism to dynamically add or remove the More Glyph.

> I already tried to do that, but I did not work. I could explain it better in a skype call, why I switched to the proposed solution (shortly: to have only a bit of text to be displayed we would need to pack this in a span which is inline (durrently, inline-block). only inline spans can have overflow:hidden with the desired behaviour, but inline-spand do not have a (computed-) hight and therefore im unable to locate items which have overlowing text..).

> RFC 2119: Please use RFC 2119 language (MUST, ...) for the rules and not for the informative part of the Kitchen Sink Entry. It's just the other way round =)

Done :-) 

> Denotation: There seems to be some concept of "denotation" that we cannot find an explanation for. From the implementation and screenshot we gather, that this is used to give the message some kind of importance by color. Please explain what this is in the description of the component, and the according function or maybe just remove it. The description of the component says, that it should be used for "relevant and important information". This would mean we only need that "important" denotation (if we understand correctly) and would not want to bother our users with less important stuff.

I added an explenation to the description of the component (with examples).

Note, the above only concerns the interface. We had a quick look at the implementation as well, however this is incomplete. Maybe some of it helps when finishing up for the final review:

> Less: Thanks for providing a set of variables for the new component as well as using the grid system. Do not introduce values for margin, font-size, height and padding directly in the less file. Try to use existing values or create variations of those in the variables.less file if necessary.
 
I rewrote the less and removed explicit margins etc. 
 
> JS Log: Remove the lines logging tho the console.

Done

> JS Scoping Pattern: Use the scoping pattern to define the public interface of your js code. In your case, this will probably only be one function called by il.Util.addOnLoad (see e.g. . This is not perfect, but atm the minimal standard we have for js code.
What if there is no plublic interface? the code is only started once, like e.g. in src/UI/templates/js/MainControls/footer.js

> Example: At least one is needed.

Done

> Tests: Are needed.

Done

In genreal: This was a bit unclear, I did not pushed the whole implementation to this PR since it would have been a bit to much. But now I merged the whole implementation as well, I hope this still works for review. Feel free to cantact me at any time. thanks in advance and best regards!

chfsx:
@Amstutz and @klees : Another update, I just implemented Triggarable in SystemInfo to have a proper Close-Trigger to combine with the Close-Button.

FMPOV you can start with the review, thanks in advance!

chfsx:
You can see the System Info in action here: http://feature-7-administrative-notifications.ilias-core.review.studer-raimann.ch (root/review)

Amstutz:
Thanks for the changes. If fine with you, we can discuss this in the JF. 

Concerning, small PRs are always better. Since we browsed over the parts of the implementation, we just thought it good to note what we saw anyways. As soon as accepted, we provide the detailed review on that as well.

Thx

chfsx:
I can open a second PR with just the Interface and rules?  NP

klees:
@chfsx No need to open a new PR afaic.

matthiaskunkel:
Jour Fixe, 26 OCT 2020: We highly appreciate this suggestion and accept the PR for ILIAS 7. Please modify the current position of the System Info and place it above the breadcrumb to keep breadcrumb and content together. In addition, it would be great if the shortened text is not only represented by the 'show more' glyph … but would consists at least of some characters plus … to clarify that there is more to read here. Fabian is looking for a possible solution.
Concerning accessibility, please make sure that tabbing to show more and cancel is supported properly. And please use the suggestion role 'alert' only for breaking notifications (as it always catches the focus after page load) and probably use role 'polite' for the other two types of notifications. And please add this information to the accessibility rules, too.

chfsx:
Hi @klees  and @Amstutz 
I implemented all discussed changes, the PR is now complete (with tests, examples, implementation, less, ...) and ready for review:

Bildschirmfoto am 2020-10-29 um 21 23 25-fullpage
Bildschirmfoto am 2020-10-29 um 21 17 14-fullpage

chfsx:
Didn't push my last commit, now it's complete ;)

klees:
I polished this a little and committed as 1a88f6ba5658c669d97463b27cb3c636ff4a52ba.

chfsx:
Thanks a lot!

PR #2956, opened 2020-10-23
Certificate: Introduce new event when certificate was achieved

mjansenDatabay:
This PR suggests to introduce a new event which is raised when an ILIAS certificate is issued to a user.


mbecker-databay:
Great help for the many requirements related to certificates!

matthiaskunkel:
Jour Fixe 26 OCT 2020 : We highly appreciate this suggestion and accept the PR for all maintained versions and trunk.

PR #2961, opened 2020-10-26
TravisCI/Composer: Try to force usage of composer 1

mjansenDatabay:


matthiaskunkel:
We highly appreciate this suggestion and accept it for release_54. Pleaese merge.

PR #2963, opened 2020-10-28
Mail - extend allowed image types for mail attachments by jpeg and gif

daniwe4:


mjansenDatabay:
Thx @daniwe4 , I already used something similar in a project. Thanks for adding this to `trunk`.

PR #2964, opened 2020-10-28
Composer/libris: Use lowercase package to support Composer 2

mjansenDatabay:


chfsx (approved):
Thanks a lot! I'll CP this to 6 and trunk as well

chfsx:
I see this has already been updated in trunk and release_6, thanks!

PR #2967, opened 2020-10-29
Global Screen, Fixing Meta Bar Labels: #28946

Amstutz:
Fixing: https://mantis.ilias.de/view.php?id=28946

Note: While checking for usages of AbstractStaticMetaBarProvider I stumbled over Services/GlobalScreen/classes/Provider/GSMetaBarProvider.php which seems not to be in use anymore. I therefore removed it, I hope this is fine.

If all good, I would promote this down to 6 after merging.


chfsx:
Thanks a lot @Amstutz! yes the class is definitely not needed anymore, thanks for deleting it.

the other changes look correct to me! actually the individual maintainers of the components should look at this, but since i made this initial change and the changes are trivial from my point of view, i can accept them collectively.

chfsx (approved):

chfsx:
And I cherry-picked to release_6 as well

PR #2969, opened 2020-10-30
Update class.ilObjMailGUI.php

aGeNET:
Added autocomplete-disabling on mail_smtp_user and mail_smtp_password to avoid a pitfall.
If an user has saved his credientials for Ilias and those fields are supposed to be empty, the browser autocompletion will fill in the saved user-credentials and will set this data as mail-credentials on saving the settings. This will result in smtp connection errors.


mjansenDatabay:
Thx for your contribution @aGeNET

PR #2955, opened 2020-10-23
UI/Symbol/Icon/Standard - Introduction of missing Glyphs

ezenzen:
This PR offers new glyphs that are needed in the page editor and for removing the ilGlyphGUI.

These Glyphs will be inserted into the KitchenSink. The glyphs bulletlist, numberedlist, listindent and listoutdent are needed for the page editor. With the introduction of the filter glyph, even the last glyph from the ilGlyphGUI can be replaced by a KS element and the ilGlyphGUI can be removed soon.

With this PR the following adjustments are made:

- Description of the KS article Symbol > Glyph
- 5 new Glyphs

Feedback is warmly welcome.

PR #2971, opened 2020-11-03
UI Components, Improving Glyph Tests

Amstutz:
This is a short follow up of: https://github.com/ILIAS-eLearning/ILIAS/pull/2955 .

In there we observed that tests are missing. I was somewhat bothered, that no test failed, eventhough, there are new Glyphs added. I changed the tests somewhat, reading alls the available glyphs and test all by default. Note that some incomplete tests have been uncovered with this change.

Further I removed some Glyphs from the list of Glyphs not having context information, since it seems like there is context now available for them.


klees:
And even removed 24 lines of code =)

Nice job. Thanks!

PR #2973, opened 2020-11-05
UI Components, Buttons: Establishing Engaged Buttons

Amstutz:
In it’s core this PR takes care of the "Engaged Buttons“ entry in the UI Components Roadmap.

@matthiaskunkel Thx you very much for initiating this and proposing the changes in CSS.

In more detail in contains the following changes:
- Change in ilCalendarHeaderNavigation to reflect the changes proposed here. Thx @smeyer for looking into this.
- Move some rule for engaged state from the engageable interface to the set of rules in the factory.
- Adapt the Buttons rules and description to follow the idea of engaged buttons
- Make all deriviates of the Button interface engageable. Note, not all Buttons extend that interface, e.g. Close Button.
- Some word smithing in the View Control Factory using the „engaged“ term to make it easier to connect to the descriptions in the buttons. Same in the Interface Descriptions.
- Adding Engageable Examples for Buttons, to feature and explain this property better.
- Completely rework the View Control Section and Mode examples, to have them working closer to some real case example. Including some engaged state.
- Adding some less along with variables to have visuals matching the style rules for engaged buttons.

Note, Unit Tests have been gold for this one, did catch some errors I added on the first try!


Amstutz:
In Standard Button:
image

Revised View Control Example:
image

Revised Section View Control Example:
image

Use Case in Calendar:
image

mkunkel:
While it's nice to get a shout out, I literally deserve no credit here. I think you tagged the wrong person.

Amstutz:
@mkunkel corrected thx for the heads up ;-).

src/UI/Component/Button/Engageable.php
```diff
@@ -4,14 +4,11 @@
...
 namespace ILIAS\UI\Component\Button;
 
+use ILIAS\UI\Component\Component;
 /**
- * A Button can be stateful; when engaged, the state MUST be refelected
- * in the "aria-pressed"-, respectively the "aria-checked"-attribute.
- * If the Button is not stateful (which is the default), the aria-attribute can
- * be omitted.
- * https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role
+ * Engegaeble Components can cary the engaged state.
```
klees:
```suggestion
 * Engageable Components have an "engaged" state and will be displayed accordingly.
```

src/UI/Component/ViewControl/Factory.php
```diff
@@ -20,25 +20,21 @@ interface Factory
      *      Mode View Controls enable the switching between different aspects of some data. The different modes are mutually
      *      exclusive and can therefore not be activated at once.
      *   composition: >
-     *      Mode View Controls are composed of Buttons switching between active and inactive states.
+     *      Mode View Controls are composed of Buttons switching between active/engaged and inactive states.
```
klees, line 23: Why not switch to that "engaged/disengaged" wording completely?

src/UI/Component/ViewControl/Mode.php
```diff
@@ -12,14 +12,14 @@
 interface Mode extends Component
 {
     /**
-     * set the currently active Button by label.
+     * set the currently active/engaged Button by label.
      *
      * @param string $label. The label of the button to activate
      */
     public function withActive($label);
```
klees, line 19: Maybe we should switch here to the "Engaged"-wording as well.

src/UI/Factory.php
```diff
@@ -210,6 +211,10 @@ public function link();
      *      4: >
      *           All Buttons visible in a view MUST be accessible by keyboard by using the
      *           ‘Tab’-Key.
+     *      5: >
+     *          The engaged state MUST be reflected in the "aria-pressed" -, respectively
+     *          the "aria-checked"-attribute if active.
```
klees, line 216: When would we use the one or the other?

matthiaskunkel:
@Amstutz Thanks for the PR. It's a real good improvement. Just one tiny suggestion to modify the proposal: how about adding two pixel space between the different buttons (I guess it was already in the proposed CSS). Yvonne also liked it, also because we have also a bit space between buttons in the Selection View Control. With this bit of space they still look as group but especially the engaged button is more obvious because its left and right border is visible (and not part of the neighbour buttons).

Amstutz:
Hi @matthiaskunkel 

Thx for the Feedback. I explained the reasoning behind my decision to stick with the 1px border by mail since I initially thought this is only of interest to you Enrico and Yvonne. In short: Finding a good clean solution without side effects is tricky and would require some more work on the button CSS. Simply providing 2px border with is not cutting it, since the buttons would not align with existing order buttons. Changing the border for all Buttons can lead to other unintended side effects (I quickly tested that and several jumped to my head).

I believe that the improvement made here is important and we should do it. If somebody likes to look into the width of the border in a further improvement, feel free to do so. However, lets first get to more important discussion on the semantics of the engaged buttons out of the way.

matthiaskunkel:
Thanks @Amstutz for clarification. I fully agree to go on with your suggestion. One pixel more or less is not that important. The general layout change is the important one.

klees (approved):
Hi @Amstutz,

sorry for the inline comments, time was short today. This is fine for me and I could not find any real issues. The one thing I noticed, and which I think we should fix or at least write a ROADMAP-entry for, is the duplicate wording which I marked several times. I think we could get rid of that "active" wording in favour of "engaged" completely, in the prose as well as in the code. This is not blocker though, just some room for further improvement.

Best regards!

src/UI/Component/ViewControl/Factory.php
Amstutz: I thought I could be easier, but maybe completely switch to engaged.

matthiaskunkel:
Jour Fixe, 09 NOV 2020 : We highly appreciate this suggestion and accept it for ILIAS 7. We will continue the discussion if the current visualisation by CSS is the best or if another background colour might improve the understanding of the button's semantic. Nevertheless, we need a general concept of a colour scheme in ILIAS to avoid creating "Insellösungen" in the UI. Please merge to trunk.

PR #2974, opened 2020-11-06
Feature/7/splitting up file object/moving webdav

chfsx:
Hi @swiniker 
As discussed in the Feature Request ["Splitting up File Object into Object and Service"](Splitting up File Object into Object and Service), we moved the WebDAV-setting to a new administration node "WebDAV" in "Extending ILIAS".

There might be conflicts in the DB-Update, we can fix them when merging locally


kergomard:
Hi @chfsx 

Thank you very much for moving the settings! @rsheer and I have had a look at it and have only found three details we would suggest to change:
- [x] The changes in Services/Repository/classes/class.ilObjRepositorySettings.php shouldn't be part of this change, right?
- [x] We would suggest to also move the settings from the 'file_access'-module to a new 'webdav'-module [here](https://github.com/ILIAS-eLearning/ILIAS/pull/2974/files#diff-73877208e61c4576196960c8c8e77dab69edf4fc05dde5344fc275a98a8c09bdR184) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/2974/files#diff-73877208e61c4576196960c8c8e77dab69edf4fc05dde5344fc275a98a8c09bdR212). Adding an corresponding step to the dbupdate.
- [ ] We are missing an icon (I'm a little fuzzy on the process to get one, but it is also not the most urgent of questions).

kindly,
@swiniker

chfsx:
Hi @swiniker 
- yes the changes in ilObjRepositorySettings did not belong to this PR, I removed them
- I introduces a new setting smodule 'webdav? and moved the entries
- as far as I know you can contact Yvonne concerning the Icons

kergomard:
Hi @chfsx 
Thank you for the changes! One more thing:
You should also move "webdav_versioning_enabled" to the "webdav" module.
You can go ahead and merge, once this is done.

We will figure out the icon issue.

Thanks again and best,
@swiniker

chfsx:
Thanks for the feedback, I moved the webdav_versioning_enabled setting as well and merged directly.

PR #2979, opened 2020-11-09
release_6 PRG:29140 add ilFormPropertyDispatchGUI to ctrl structur ca…

daniwe4:
…ll into ilObjStudyProgrammeMembersGUI

https://mantis.ilias.de/view.php?id=29140


klees:
Picked to trunk and release_7.

PR #2981, opened 2020-11-10
Added course-specific userdata to certificate text placeholders

M-Barz:
As discussed, version 1 of the feature for review

### DIFF
--- a/Services/Certificate/classes/Gui/class.ilCertificateGUIFactory.php
+++ b/Services/Certificate/classes/Gui/class.ilCertificateGUIFactory.php
@@ -69,7 +69,7 @@ class ilCertificateGUIFactory
             case 'crs':
                 $hasAdditionalElements = true;
 
-                $placeholderDescriptionObject = new ilCoursePlaceholderDescription();
+                $placeholderDescriptionObject = new ilCoursePlaceholderDescription($objectId);
                 $placeholderValuesObject = new ilCoursePlaceholderValues();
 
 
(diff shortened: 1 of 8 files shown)


mjansenDatabay:
@M-Barz / @smeyer-ilias Although a review has not beend requested, yet :): This looks good to me, especially the changes in the actual `Certificate` service classes:
* ilCertificateGUIFactory
* ilUserDefinedFieldsPlaceholderDescription
* ilUserDefinedFieldsPlaceholderValues

The changes in the other classes are course specific and could be moved to`Modules/Course` during a future refactoring (e.g. when a new feature will be implemented in `Services/Certificate`).

Feel free to merge this (after fixing the unit tests).

M-Barz:
First of all, thanks for the quick look into the pull request @mjansenDatabay ! :)
The Unit Tests are fixed now. We are however currently waiting for the OK from the Customer, which we should be getting very soon, before we merge.

mjansenDatabay:
@M-Barz Okay. But the unit tests still are still failing :-).

M-Barz:
Yeah, that was a case of "It ran on my machine". I'm on it!

M-Barz:
Some information for the "bandaid fix":
ilObjectCustomUserFieldsPlaceholderDescription accesses a static method that needs a DB connection, since that can't really be given like that, I decided to mock the class and give some value the getPlaceholderDescriptions method of the class would return.
However, for some reason that I am too nooby to understand, the method in the Unittest won't for the life of me be actually initialised and I get a "tried to access method of null" error.
So after many hours of googling, I found a solution by calling setmethodsexcept in the MockBuilder. This works, but I don't think it's good code and should probably be refactored when there's more time.

For now, it does it's job though so I pushed it like that.

PR #2983, opened 2020-11-11
Bugfix 28971 Sorting of forum drafts

playamongthestars:
Mantis Issue: https://mantis.ilias.de/view.php?id=28971

### DIFF
--- a/Modules/Forum/classes/GUI/class.ilForumThreadFormGUI.php
+++ b/Modules/Forum/classes/GUI/class.ilForumThreadFormGUI.php
@@ -155,7 +155,7 @@ class ilForumThreadFormGUI extends \ilPropertyFormGUI
             $this->addItem($captcha);
         }
 
-        if (\ilForumPostDraft::isSavePostDraftAllowed()) {
+        if (\ilForumPostDraft::isSavePostDraftAllowed() && !$this->user->isAnonymous()) {
             $this->ctrl->setParameter($this->delegatingGui, 'draft_id', $this->draftId);
             if (in_array($this->ctrl->getCmd(), ['publishThreadDraft', 'editThreadDraft', 'updateThreadDraft'])) {
                 $this->addCommandButton('publishThreadDraft', $this->lng->txt('publish'));
--- a/Modules/Forum/classes/class.ilForumPostDraft.php
+++ b/Modules/Forum/classes/class.ilForumPostDraft.php
@@ -60,6 +60,9 @@ class ilForumPostDraft
      * @var int
      */
     protected $notify = 0;
+    /**
+     * @var int
+     */
     protected $post_notify = 0;
     
     /**
@@ -384,7 +387,52 @@ class ilForumPostDraft
             self::$instances[$user_id]['draft_ids'][$tmp_obj->getDraftId()] = $tmp_obj;
         }
     }
+
+    /**
+     * @param int $user_id
+     * @param int $thread_id
+     * @param int $sorting
+     * @return ilForumPostDraft[]
+     */
+    public static function readSortedDrafts(int $user_id, int $thread_id, int $sorting = ilForumProperties::VIEW_DATE_ASC) : array
+    {
+        global $DIC;
+        $ilDB = $DIC->database();
     
+        $drafts = [];
+    
+        $order_statement = ' ';
+        $order_direction = ' ';
+    
+        if ($sorting != ilForumProperties::VIEW_TREE) {
+            $order_statement = ' ORDER BY post_date ';
+            $order_direction = 'ASC';
+            if ($sorting == ilForumProperties::VIEW_DATE_DESC) {
+                $order_direction = 'DESC';
+            }
+        }
+        $res = $ilDB->queryF(
+            'SELECT * FROM frm_posts_drafts WHERE post_author_id = %s AND thread_id = %s' .
+            $order_statement . $order_direction,
+            ['integer', 'integer'],
+            [$user_id, $thread_id]
+        );
+    
+        while ($row = $ilDB->fetchAssoc($res)) {
+            $draft = new ilForumPostDraft();
+            self::populateWithDatabaseRecord($draft, $row);
+            $drafts[] = $draft;
+            self::$instances[$user_id][$thread_id][$draft->getPostId()][] = $draft;
+        }
+    
+        if ($sorting == ilForumProperties::VIEW_TREE) {
+            return self::$instances[$user_id][$thread_id];
+        } else {
+            return $drafts;
+        }
+        return [];
+    }
+
     /**
      * @param int $user_id
      * @return \ilForumPostDraft[]
--- a/Modules/Forum/classes/class.ilObjForumGUI.php
+++ b/Modules/Forum/classes/class.ilObjForumGUI.php
@@ -703,7 +703,7 @@ class ilObjForumGUI extends \ilObjectGUI implements \ilDesktopItemHandling
         }
 
         // Mark all topics as read button
-        if ($this->user->getId() != ANONYMOUS_USER_ID && !(int) strlen($this->confirmation_gui_html)) {
+        if (!$this->user->isAnonymous() && !(int) strlen($this->confirmation_gui_html)) {
             $this->toolbar->addButton(
                 $this->lng->txt('forums_mark_read'),
                 $this->ctrl->getLinkTarget($this, 'markAllRead'),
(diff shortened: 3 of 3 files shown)


Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -3078,22 +3097,49 @@ public function viewThreadObject()
...
                     $this->renderPostingForm($threadContentTemplate, $frm, $firstNodeInThread, $this->requestAction);
                 }
-                $this->renderDraftContent(
-                    $threadContentTemplate,
-                    $this->requestAction,
-                    $render_drafts,
-                    $firstNodeInThread,
-                    $selected_draft_id
-                );
+                if (ilForumPostDraft::isSavePostDraftAllowed() && $currentSortation < ilForumProperties::VIEW_DATE) {
```
mjansenDatabay: Why do you use the `<` operator here? I'd expect something like `===` or `!=='.

Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -777,11 +777,16 @@ protected function renderDraftContent(ilTemplate $tpl, string $action, bool $ren
 
         $frm = $this->object->Forum;
 
-        $draftsObjects = ilForumPostDraft::getInstancesByUserIdAndThreadId(
-            $this->user->getId(),
-            $this->objCurrentTopic->getId()
-        );
-        $drafts = $draftsObjects[$node->getId()];
+        if($_SESSION['viewmode'] >= ilForumProperties::VIEW_DATE) {
```
mjansenDatabay: Why do you use the `<` operator here? I'd expect something like `===` or `!=='.

Can't we use `$currentViewMode` instead of `$_SESSION['viewmode']`?

Modules/Forum/classes/class.ilForumPostDraft.php
```diff
@@ -384,6 +384,31 @@ protected static function readDrafts($user_id)
             self::$instances[$user_id]['draft_ids'][$tmp_obj->getDraftId()] = $tmp_obj;
         }
     }
+
+    public static function readSortedDrafts($user_id, $thread_id, $sorting = ilForumProperties::VIEW_DATE_ASC)
```
mjansenDatabay: Please add PHP 7 typehints and return type declarations. Try to use https://psalm.dev/docs/annotating_code/type_syntax/array_types/ if a method returns an arrray.

Is there any way to avoid this `static` method here? Is it possible to use an object instead and call an instance-method?

Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -3078,22 +3097,49 @@ public function viewThreadObject()
                     }
                 }
                 $this->renderPostContent($threadContentTemplate, $node, $this->requestAction, $pageIndex, $postIndex);
-                $this->renderDraftContent($threadContentTemplate, $this->requestAction, $render_drafts, $node, $selected_draft_id);
+                if ( ilForumPostDraft::isSavePostDraftAllowed() && $_SESSION['viewmode'] < ilForumProperties::VIEW_DATE) {
```
mjansenDatabay: Why do you use the `<` operator here? I'd expect something like `===` or `!=='.

mjansenDatabay (changes requested):
Thanks for this bugfix @nmatuschek . If only left some minor remarks.

Please check the `PSR-2 + addions` coding style.

Modules/Forum/classes/class.ilObjForumGUI.php
playamongthestars: No. Because the "$currentSortation === ilForumProperties::VIEW_TREE" condition does not match in every case when it is expected. 
I am still not sure why this happens, but the $currentSortation variable isn't even set properly in every changed viewmode case. 

The tests showed, that the session var is set properly instead.

Modules/Forum/classes/class.ilObjForumGUI.php
playamongthestars: No. Because the "$currentSortation === ilForumProperties::VIEW_TREE" condition does not match in every case when it is expected. 
I am still not sure why this happens, but the $currentSortation variable isn't even set properly in every changed viewmode case. 

The tests showed, that the session var is set properly instead.

Modules/Forum/classes/class.ilObjForumGUI.php
playamongthestars: No. Because the "$currentSortation === ilForumProperties::VIEW_TREE" condition does not match in every case when it is expected. 
I am still not sure why this happens, but the $currentSortation variable isn't even set properly in every changed viewmode case. 

The tests showed, that the session var is set properly instead.

mjansenDatabay:
@nmatuschek Yesterday I pushed some changes to the PR branch. Please verifiy (cross-check) these changes regarding to the desired functionality.

Please have a look at the static method calls to `ilForumPostDraft`. We MUST ensure that only drafts of the current thread (determined by the request) are fetched (and preloaded). There MUST NOT be any further database queries if we already have all relevant drafts for the current view.

We furthermore MUST do some refactoring (let's begin with `Extract Method` followed by `Extract Class`) of the `Draft` related code based on `trunk` if new features are implemented for the forum.

playamongthestars:
Strange. I don't know why this happens. I checked and compared the code manually. Every thing is fine there. Only this github-view seems to mess up the code.

mjansenDatabay:
@nmatuschek You are right, this seems to be a visual diff issue on `GitHub`.

mjansenDatabay:
@nmatuschek For now, I'd propose the following: 

- [ ] Fix `\ilForumPostDraft::getInstancesByUserIdAndThreadId` by respecting the passed `$thread_id` when reading the drafts from database.
- [ ] Remove `ilForumPostDraft::readSortedDrafts` from the code base
- [ ] Create a `SortedDraftsCollection` (maybe you'll find a better name) and pass the preloaded drafts to it's constructor. Whereever you'll need the drafts to be sorted `ASC` or `DESC`, call a corresponding method of this class which returns the drafts in the desired order.

Modules/Forum/classes/GUI/class.ilForumThreadFormGUI.php
```diff
@@ -155,7 +155,7 @@ protected function initForm()
             $this->addItem($captcha);
         }
 
-        if (\ilForumPostDraft::isSavePostDraftAllowed()) {
+        if (\ilForumPostDraft::isSavePostDraftAllowed() && $this->user->getId() != ANONYMOUS_USER_ID) {
```
mjansenDatabay: @nmatuschek Please use `!$this->user->isAnonymous()`

Modules/Forum/classes/class.ilForumPostDraft.php
```diff
@@ -384,7 +387,52 @@ protected static function readDrafts($user_id)
...
+    {
+        global $DIC;
+        $ilDB = $DIC->database();
+    
+        $drafts = [];
+    
+        $order_statement = ' ';
+        $order_direction = ' ';
     
+        if ($sorting == ilForumProperties::VIEW_DATE_ASC || $sorting == ilForumProperties::VIEW_DATE_DESC) {
```
mjansenDatabay: @nmatuschek Can't we use a default sorting and use only one `if` statement to adust the sorting in this case?

Modules/Forum/classes/class.ilForumPostDraft.php
```diff
@@ -384,7 +387,52 @@ protected static function readDrafts($user_id)
...
+        );
+    
+        while ($row = $ilDB->fetchAssoc($res)) {
+            $draft = new ilForumPostDraft();
+            self::populateWithDatabaseRecord($draft, $row);
+            $drafts[] = $draft;
+            self::$instances[$user_id][$thread_id][$draft->getPostId()][] = $draft;
+        }
+    
+        if ($sorting == ilForumProperties::VIEW_TREE) {
```
mjansenDatabay, line 428: @nmatuschek Why do we have this different return statements here?

Modules/Forum/classes/GUI/class.ilForumThreadFormGUI.php
playamongthestars: done.

Modules/Forum/classes/class.ilForumPostDraft.php
playamongthestars: Yes,  we could use if($sorting != ilForumProperties::VIEW_TREE)

mjansenDatabay: Okay, please try to improve this. Thx in advance.

Modules/Forum/classes/class.ilForumPostDraft.php
playamongthestars: The rendering process is handled differently in the GUI dependent to the viewmode. While the rendering by date does not need a parent node, the hierarchical rendering by tree does. The parent post node can have multiple drafts beneath, that's why the return of an array with multiple drafts is needed here.   
For the rendering by date we know, the it is also possible, that in some cases there is no parent post id needed i.e. the dummy node. For now, I did not found a practicable solution to unify this process without the need to change the whole rendering process itself. :-( not happy with it. In fact we need a thoughtful concept for this.

mjansenDatabay: Okay, I see. It's okay for me for this bugfix, but as you said: We'll have to consider this (refactoring) for (possible) future features.

mjansenDatabay:
@nmatuschek Thanks a lot for your work. Please stash all commits, so we could easily cherry-pick this to all relevant branches.

mjansenDatabay (approved):

PR #2984, opened 2020-11-11
Migrate WebDAV Settings to DB for ILIAS7

rsheer:
With this PR, I moved the settings if WebDAV is enabled from the client.ini file to the database. (1. commit)

Additionally, there is some cleanup of unused settings and methods in the new ilObjWebDAV file. (2. commit)

@klees I'm not sure about the DB-Update part. My idea was to migrate the existing setting `webdav_enabled` from the client.ini file to the database. But the reading of the client.ini file looks kind of hacky to me in the database migrations. What is your opinion about it?

### DIFF
--- a/Services/WebDAV/classes/class.ilObjWebDAV.php
+++ b/Services/WebDAV/classes/class.ilObjWebDAV.php
@@ -92,28 +92,6 @@ class ilObjWebDAV extends ilObject
     }
 
     /**
-     * Sets the webdavActionsVisible property.
-     *
-     * @param boolean    new value
-     *
-     * @return    void
-     */
-    public function setWebdavActionsVisible($newValue)
-    {
-        $this->webdavActionsVisible = $newValue;
-    }
-
-    /**
-     * Gets the webdavActionsVisible property.
-     *
-     * @return    boolean    value
-     */
-    public function isWebdavActionsVisible()
-    {
-        return $this->webdavActionsVisible;
-    }
-
-    /**
      * Gets the defaultWebfolderInstructions property.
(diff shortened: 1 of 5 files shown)


Services/WebDAV/classes/class.ilObjWebDAV.php
```diff
@@ -210,30 +173,8 @@ public function read()
...
-    /**
-     * Gets instructions for the usage of webfolders.
-     *
-     * The instructions consist of HTML text with placeholders.
-     * See _getWebfolderInstructionsFor for a description of the supported
-     * placeholders.
-     *
-     * @return String HTML text with placeholders.
-     */
-    public static function _getDefaultWebfolderInstructions()
```
kergomard, line 230: Thanks for removing this, but you will need to adapt the function [`public function getDefaultWebfolderInstructions()`](https://github.com/ILIAS-eLearning/ILIAS/pull/2984/files#diff-73877208e61c4576196960c8c8e77dab69edf4fc05dde5344fc275a98a8c09bdL122), too.

kergomard (commented):
Thanks @rsheer 
Let's stick with the funky implementation of the update step for the time being, if we find a better solution, we can still change it.
Other than that I've just one little change request.

Services/WebDAV/classes/class.ilObjWebDAV.php
rsheer: Thanks for having an eye over the changes (and even the forgotten ones)! I removed the mentioned function an the additional other ones which are also unused.

kergomard (approved):
Thank you very much @rsheer

PR #2985, opened 2020-11-12
Various ToS Withdrawal Refactorings

mbecker-databay:


mjansenDatabay (changes requested):
Thx @maxbecker Can we use `$this->lng` instead of `$GLOBALS['lng']`?

mjansenDatabay (approved):

PR #2988, opened 2020-11-12
Release_6 Survey:28739 remove unacessary and wrong include statement

daniwe4:
https://mantis.ilias.de/view.php?id=28739

This patch will only avoid the whoops message. The functionality to download a pic isn't given yet, due to the missing library phantom.js in composer.json.
I create a new ticket for that issue. https://mantis.ilias.de/view.php?id=29164


alex40724:
These lines have been removed in the meantime. There are still remaining issues in this feature.

alex40724:
Thanks for the PR though.

PR #2991, opened 2020-11-12
ilCtrl: Add cheap option to enable CSRF token verification for HTTP GE…

mjansenDatabay:
…T request commands

Maybe these changes could be a 'cheap' approach to support CSRF tokens for command strings passed in HTTP GET requests.

I did a regex based search for `->getCmd\([^\)]+?,(.+?)\)` and only found 2 usages where `Safe POST commands` were passed, with this PR we instead introduced some `configuration object`.

### Link Creation

In this example case, we offer a link to the user in a listing of forum postings. If the user clicks on the link, the system state will be modified (here: mark a posting as read):

```php
$this->ctrl->setParameterByClass(ilObjForumGUI::class, 'pos_pk', 4711);

$link = $this->ctrl->getLinkTarget(
    ilObjForumGUI::class,
    'markPostRead',
    'posting-4711',
    false,
    false,
    true // Yes, this is (unfortunately) another parameter beingt introduced with this PR ...
);
```
The CSRF token is appended to this URL only, not to other URLs contained in the HTML response being send to the browser.

### Command Retrieval from `ilCtrl`

If we wan't to process this request, we'll have to retrieve the command from `ilCtrl`. We'll have to tell `ilCtrl` that `markPostUnread` is a state modifying command where a CSRF token has to be checked in consequence.

```php
        $cmd = $this->ctrl->getCmd('showContent', new class implements ilCtrlCommandSecurity {
            public function getSafePostCommands() : array
            {
                return [];
            }

            public function getUnsafeGetCommands() : array
            {
                return ['markPostUnread'];
            }

        });
```

Result: If the command `markPostUnread` is given in a HTTP GET request, the CSRF token will be verified. If the verification fails, the default command will be returned by `ilCtrl`. This is the same behaviour already implemented for failed CSRF token verifications in HTTP POST requests.

----

This approach could be a way to handle security issues being reported for the internal views (user is logged in), where system state is currently modified by HTTP GET requests.

### DIFF
--- a/Modules/LearningModule/Presentation/classes/class.ilLMPresentationGUI.php
+++ b/Modules/LearningModule/Presentation/classes/class.ilLMPresentationGUI.php
@@ -376,8 +376,17 @@ class ilLMPresentationGUI
         }
         
         $next_class = $this->ctrl->getNextClass($this);
-        $cmd = $this->ctrl->getCmd("layout", array("showPrintView"));
+        $cmd = $this->ctrl->getCmd("layout", new class implements ilCtrlCommandSecurity {
+            public function getSafePostCommands() : array
+            {
+                return ['showPrintView'];
+            }
 
+            public function getUnsafeGetCommands() : array
+            {
+                return [];
+            }
+        });
 
         $obj_id = $this->requested_obj_id;
         $this->ctrl->setParameter($this, "obj_id", $this->requested_obj_id);
--- a/Services/Init/classes/class.ilStartUpGUI.php
+++ b/Services/Init/classes/class.ilStartUpGUI.php
@@ -94,7 +94,17 @@ class ilStartUpGUI
      */
     public function executeCommand()
     {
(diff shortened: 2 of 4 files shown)


alex40724:
@mjansenDatabay Thanks for the PR, please discuss this in the JF or maybe in the TB. If accepted, this can be merged from my side.

Amstutz:
The TB discussed this today while monitoring old PRs. @mjansenDatabay will look into this and discuss this with the JF.

thibsy:
Hi @mjansenDatabay, thank you for the contribution! 

As I have mentioned this tuesday, I am working on refactoring the ilCtrl class and exchange the storage of the call structure in the database to the storage in an artifact. 

This means a lot has already changed in the codebase so it wouldn't make much sense to accept your PR right now, therefore I am closing this PR. I do like the option to enable CSRF validation for GET requests though, and I will implement this in the new version as well.  

Thanks for your understanding!

PR #2993, opened 2020-11-16
Component/Setup: use generalized objectives for plugin updates (#29169)

klees:
This fixes [#29169](https://mantis.ilias.de/view.php?id=29169) by using objectives that do not depend on configs.


klees:
Please also pick to trunk!

PR #2996, opened 2020-11-17
TermsOfService: Fix Mantis issue 29166

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=29166

If merge, please cherry-pick this to `release_7`.

### DIFF
--- a/Services/Init/classes/class.ilStartUpGUI.php
+++ b/Services/Init/classes/class.ilStartUpGUI.php
@@ -240,7 +240,7 @@ class ilStartUpGUI
             $tpl->setVariable('LPE', $page_editor_html);
         }
 
-        $tosWithdrawalGui = new ilTermsOfServiceWithdrawalGUIHelper();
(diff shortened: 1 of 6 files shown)


mjansenDatabay:
@mbecker-databay 

With this PR, I ...

- ... injected the `Subject User` as a constructor dependency in the `ilTermsOfServiceWithdrawalGUIHelper` you introduced, so it does NOT depend on a global state user object anymore.
- ... reordered the public methods you introduced there, according to the process.
- ... fixed the argument when calling `deleteAcceptanceHistoryByUser` in https://github.com/ILIAS-eLearning/ILIAS/pull/2996/files#diff-c93fe13352074acc0cf3bb986fa1b0f7e159bd53be8f326083fd2de529d06a1dR1558 . You passed an instance of `ilObjUser`, where an `int` is expected.
- ... introduced some checks regarding the subject user, so only users with certain criteria (whitelist, or to be polical correct: inclusion) are considered for withdrawl operations.
- ... got rid of some static calls used to check if the ToS service is globally enabled.

PR #3001, opened 2020-11-17
Booking Pool: Add deletion of reservations

iszmais:
https://mantis.ilias.de/view.php?id=28973

### DIFF
--- a/Modules/BookingManager/classes/class.ilObjBookingPoolGUI.php
+++ b/Modules/BookingManager/classes/class.ilObjBookingPoolGUI.php
@@ -1695,15 +1695,7 @@ class ilObjBookingPoolGUI extends ilObjectGUI
             foreach (explode(',', $get) as $id) {
                 $res = new ilBookingReservation($id);
                 $obj = new ilBookingObject($res->getObjectId());
-                $ref_ids = ilObject::_getAllReferences($obj->getPoolId());
-                $access = false;
-                foreach ($ref_ids as $ref_id) {
-                    if ($this->access->checkAccess("write", "", $ref_id)) {
-                        $access = true;
-                        break;
-                    }
-                }
-                if (!$access) {
+                if ($obj->getPoolId() != $this->object->getId() || !$this->checkPermissionBool("write")) {
                     ilUtil::sendFailure($this->lng->txt('permission_denied'), true);
                     $this->ctrl->redirect($this, 'log');
                 }
--- a/Modules/BookingManager/classes/class.ilObjBookingPoolGUI.php
+++ b/Modules/BookingManager/classes/class.ilObjBookingPoolGUI.php
@@ -1648,6 +1648,70 @@ class ilObjBookingPoolGUI extends ilObjectGUI
         $this->logObject();
     }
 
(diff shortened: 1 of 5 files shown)


Modules/BookingManager/classes/class.ilObjBookingPoolGUI.php
```diff
@@ -1648,6 +1648,71 @@ public function rsvCancelObject()
...
+        $conf->addItem('rsv_ids', implode(',', $ids), $details);
+        $this->tpl->setContent($conf->getHTML());
+    }
+
+    public function rsvDeleteObject()
+    {
+        global $DIC;
+        if ($DIC->http()->request()->getParsedBody()['rsv_ids']) {
+            include_once 'Modules/BookingManager/classes/class.ilBookingReservation.php';
+            foreach (explode(',', $DIC->http()->request()->getParsedBody()['rsv_ids']) as $id) {
```
mjansenDatabay: @iszmais Do we need an additional check, if each `id` does belong to the current object?

iszmais: Hmm... good question. But i would say no since even if they dont belong to the current object the user should be still able to delete them if he has the right permission to do so. So the function can be keeped abstract to provide a higher usability without security issues.

mjansenDatabay: I am not sure if `rsvDeleteObject` has to fulfill an abstract purpose. It is bound to a very concrete controller/command. If you (as an attacker) are able to inject reservation ids to `$_POST['rsv_ids']`, and these ids belong to another repository object where you did **not** have any higher priviledges, you successfully exploited an `Broken Access Control` issue.

iszmais: In this case i will expand the Security check below.

mjansenDatabay (changes requested):
@iszmais LGTM, except the missing context check for the reservation ids.

Modules/BookingManager/classes/class.ilObjBookingPoolGUI.php
iszmais: The new implementation should fullfill all security guidlines.

Modules/BookingManager/classes/class.ilObjBookingPoolGUI.php
```diff
@@ -1689,23 +1689,33 @@ public function rsvConfirmDeleteObject()
...
+                $res = new ilBookingReservation($id);
+                $obj = new ilBookingObject($res->getObjectId());
+                $ref_ids = ilObject::_getAllReferences($obj->getPoolId());
+                $access = false;
+                foreach ($ref_ids as $ref_id) {
+                    if ($this->access->checkAccess("write", "", $ref_id)) {
+                        $access = true;
+                        break;
+                    }
+                }
```
alex40724: You only need to check
- write permission for the current ref id and
- does the $obj->getPoolId() match the current pool id. (ensure that the reservations belong to the current pool)

No need to get all the references.

mjansenDatabay (approved):

PR #3009, opened 2020-11-20
release_6 prg:29141 redirect after modal

daniwe4:


klees:
I have manually picked this to release_6, release_7 and trunk.

PR #3011, opened 2020-11-20
Chatroom/Setup: make config object more robust for update

klees:
The objectives `ilIniFilesInitializedObjective` and `ilDatabaseInitializedObjective` have been introduced to make it simpler to build objectives that work during setup and update. For timing reasons in the development, they have been forgotten for the chatroom.


mjansenDatabay:
Thx @klees , I'll pick this to `trunk`.

PR #3016, opened 2020-11-23
Mantis 28534

ILIASLM:
https://mantis.ilias.de/view.php?id=28534 Role is not mentioned anymore.


smeyer-ilias (approved):
Thanks for fixing this.

PR #3020, opened 2020-11-25
Fix Mantis #29224 

seul-qualitus:
… course details (0029224)

 Changes to be committed:
	modified:   Services/Tracking/classes/repository_statistics/class.ilTrObjectUsersPropsTableGUI.php


Services/Tracking/classes/repository_statistics/class.ilTrObjectUsersPropsTableGUI.php
```diff
@@ -296,7 +296,7 @@ public function initFilter()
                         ilLPStatus::LP_STATUS_COMPLETED_NUM + 1 => $lng->txt(ilLPStatus::LP_STATUS_COMPLETED),
                         ilLPStatus::LP_STATUS_FAILED_NUM + 1 => $lng->txt(ilLPStatus::LP_STATUS_FAILED)));
                     $this->filter["status"] = $item->getValue();
-                    if ($this->filter["status"]) {
+                    if (is_int($this->filter["status"])) {
```
mjansenDatabay: @seul-qualitus Are you sure `is_int` is **not** to strict for this check? If this is a `(string) 1`, this will evaluate in a boolean `false`. Maybe `is_numeric` is sufficient here?

Another option: `if (is_int((int) $this->filter["status"])) {`

smeyer-ilias:
Thanks, @seul-qualitus  and @mjansenDatabay

PR #3023, opened 2020-11-25
Metadata: Possible fix for Mantis issue 29241

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=29241

If approved, please cherry-pick this to `trunk` and alle release branches.


smeyer-ilias:
Thanks, that was the right approach, but there was a similar problem with the element "Annotation", so I added a slightly different implementation to >=5.4

PR #3024, opened 2020-11-26
add lang pd_personal_workspace

fsesterh:
pd_personal_workspace is missing here in ILIAS 6:
- Personal Workspace -> Portfolio -> Add Portfolio
Screenshot_2020-11-26 DEV ILIAS Next Portfolio


matthiaskunkel:
As language maintainer, I would like to reject this PR. We already have a lang var for "Personal Workspace", see common#:#mm_personal_workspace# that could be used. I do not see a reason to introduce a second, similar lang var. The responsible maintainer just need to replace the outdated lang var (due to the abandonment of the Personal Desktop) by common#:#mm_personal_workspace#

PR #3027, opened 2020-11-26
Use boolean value for a boolean parameter

nicoroeser:
Being more strict should improve code readability.


nicoroeser:
This is https://mantis.ilias.de/view.php?id=29260 in the ILIAS issue tracker.

PR #3035, opened 2020-12-02
Fixed forum viewmode

playamongthestars:


Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -97,6 +97,8 @@ class ilObjForumGUI extends \ilObjectGUI implements \ilDesktopItemHandling
     public $rbac;
     public $locator;
     public $ilHelp;
+    /** @var int */
+    public $selectedSorting = ilForumProperties::VIEW_TREE;
```
mjansenDatabay: Why did you choose a `public` access modifier here?

Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -2706,26 +2694,28 @@ private function getForumObjects()
 
     public function checkUsersViewMode() : void
     {
-        if (!isset($_SESSION['viewmode'])) {
-            $_SESSION['viewmode'] = $this->objProperties->getDefaultView();
+        if (!isset($this->selectedSorting)) {
```
mjansenDatabay: You initialize this member variable with `public $selectedSorting = ilForumProperties::VIEW_TREE;` So IMO it is **not** possible, that this variable is **not** set.

Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -2706,26 +2694,28 @@ private function getForumObjects()
...
 
-        if (isset($_GET['viewmode']) && (int) $_GET['viewmode'] !== (int) $_SESSION['viewmode']) {
-            $_SESSION['viewmode'] = (int) $_GET['viewmode'];
+        if (isset($this->httpRequest->getQueryParams()['viewmode'])
+            && (int) $this->httpRequest->getQueryParams()['viewmode']  !== (int) $this->selectedSorting) {
+            $this->selectedSorting = (int) $this->httpRequest->getQueryParams()['viewmode'];
         }
 
-        if (isset($_GET['action']) && !in_array($_SESSION['viewmode'], [
+        if (isset($_GET['action']) && !in_array($this->selectedSorting, [
```
mjansenDatabay: Please use `$this->httpRequest->getQueryParams()` instead of `$_GET`, if possible.

Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -2706,26 +2694,28 @@ private function getForumObjects()
...
             ilForumProperties::VIEW_TREE,
             ilForumProperties::VIEW_DATE_ASC,
             ilForumProperties::VIEW_DATE_DESC
         ])) {
-            $_SESSION['viewmode'] = $this->objProperties->getDefaultView();
+            $this->selectedSorting = $this->objProperties->getDefaultView();
         }
+        \ilSession::set('viewmode',$this->selectedSorting);
```
mjansenDatabay: Where is the `viewmode` fetched by calling `ilSession::set('viewmode')`? I did not find any usage.

* The `viewmode` **MUST** be persisted in session (if changed) and **always** read from session when determining the `effective viewmode` for the presentation of postings.
* If the `viewmode` is changed by an according UI action (by choosing a another **valid** `viewmode`), it **MUST** be modified in the session and used for the presentation of postings.

mjansenDatabay (changes requested):
Thx @nmatuschek for proposing these changes. Please have a look at my remarks.

Modules/Forum/classes/class.ilObjForumGUI.php
playamongthestars: We decided, to change the $_GET['viewmode']  to $this->httpRequest->getQueryParams() only. We could change all GET parameters, If needed.

Modules/Forum/classes/class.ilObjForumGUI.php
playamongthestars: true ^^

Modules/Forum/classes/class.ilObjForumGUI.php
playamongthestars: I had no special reason for this

mjansenDatabay: Then please make it `private`.

Modules/Forum/classes/class.ilObjForumGUI.php
mjansenDatabay: We did not, maybe we misunderstood. We **MUST NOT** rely only on HTTP query parameters.

Scenario:

1. Given a user enters the thread view of a forum (list of postings)
2. ... and the user selects a viewmode
3. When the user navigates to (e.g.) a course and then back to the thread view of the forum (list of postings)
4. Then the previous selected viewmode **MUST** be applied.

Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -2706,27 +2694,34 @@ private function getForumObjects()
...
             ilForumProperties::VIEW_DATE_ASC,
             ilForumProperties::VIEW_DATE_DESC
         ])) {
-            $_SESSION['viewmode'] = $this->objProperties->getDefaultView();
+            $this->selectedSorting = \ilSession::get('viewmode');
         }
+
+        if (isset($this->httpRequest->getQueryParams()['viewmode'])
+            && (int) $this->httpRequest->getQueryParams()['viewmode']  !== (int) $this->selectedSorting) {
+            $this->selectedSorting = (int) $this->httpRequest->getQueryParams()['viewmode'];
```
```suggestion
        if (
            isset($this->httpRequest->getQueryParams()['viewmode']) &&
            (int) $this->httpRequest->getQueryParams()['viewmode']  !== (int) $this->selectedSorting
        ) {
            $this->selectedSorting = (int) $this->httpRequest->getQueryParams()['viewmode'];
```

mjansenDatabay (approved):

PR #3039, opened 2020-12-06
replace abandoned zend-httphandlerrunner with laminas-httphandlerrunner

eqsoft:
Hi Uwe,
Replaced abandoned zend-httphandlerrunner with laminas-httphandlerrunner. Additional Tests with xapiproxy are required. Should also be commited as bugfix into release_6.
Please take a look at it.
Stefan


matthiaskunkel:
Jour Fixe, 07 DEC 2020: Thanks for the PR. But before deciding on the replacement of the Zend library we would like to have some more details about the reason to take the Laminas library instead and if it couldn't be an alternative to extend our HTTP service by a PR and to get rid of an external dependency. We keep the JourFixe label and ask for comments and additional info until the next JF at 21 DEC 20.

klees:
Hi @eqsoft,

thanks for looking into replacing that abandoned library in our core. Please understand that using code from somewhere else is a liability we accept for the future. It comes with risks we should understand and consciously accept. I currently cannot find when we accepted the zend-httprunnerhandler (I expected a date for JF acceptance in the `extra`-section in the `composer.json`), but if I remember correctly you only included it about a year ago. So I hope you understand that I try to perform due diligence here. Please answer some questions to help me understand the risks:

* What is the role of the library in the XAPI-implementation? How important is it for the functionality of the component? How deeply is it integrated to the component?
* Are there notable alternatives to the library? Maybe even one we already use for other purposes or in other components?
* What is the story behind that zend-to-laminas move and why do we expect laminas to be existent and reliable in the future? Some links and a maybe a short comment will be absolutely enough, no need for a history lesson.

Thanks a lot!

eqsoft:
Hi @klees 

----
* What is the role of the library in the XAPI-implementation? How important is it for the functionality of the component? How deeply is it integrated to the component?

We are using the "emit()" function of the laminas/laminas-httphandlerrunner (formally zend) in the xapiproxy which consumes and emits a psr-7 conformant response implementation. The Guzzle HttpClient in our proxy creates a new request to a learninglocker backend and send it back to the client by using the Emitter->emit() function

----

* Are there notable alternatives to the library? Maybe even one we already use for other purposes or in other components?

As i know there is no existing alternative for us. But we might discuss a further Emitter implementation into our own HTTP Service in ILIAS 8, to get rid of this library. But this would produce a new FR with little extra funding. I don't think this would be a big thing to implement.

-----

* What is the story behind that zend-to-laminas move and why do we expect laminas to be existent and reliable in the future? Some links and a maybe a short comment will be absolutely enough, no need for a history lesson.

Laminas is the official open source continuation of the Zend Framework (https://docs.laminas.dev/migration/). A diff of the source code changes in the Emitter class shows no functional differences between the old zend and new laminas Emitter class.

best regards! 
Stefan

matthiaskunkel:
Jour Fixe, 21 DEC 2020 : We highly appreciate this suggestion and accept the PR for ILIAS 7 and trunk. For ILIAS 8 we would like to reduce dependicies from external libraries and ask the maintainer to get rid of one of the HTTP frameworks (Laminar or Guzzle) or - even better - to implement related functions for the ILIAS HTTP service to get rid of both libs.

mjansenDatabay:
Hi @Uwe-Kohnle ,

we stumbled over this PR when we checked for old PRs at the meeting of the Technical Board. Could you please have a look at it and process this PR according to the JF decision? Thanks!

**Important:** The `zend-httphandlerrunner` / `laminas-httphandlerrunner` topic will be relevant for the upcoming PHP 8 project, because all of our `Composer` dependencies MUST support PHP 8 as well, which is not given for `zend-httphandlerrunner`.

PR #3040, opened 2020-12-09
User/CustomFields: Handle empty values

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=29369


smeyer-ilias:
Thanks for fixing this.

PR #3041, opened 2020-12-09
fixed 29372: PDF Generation currently results in a grey background

alex40724:
See https://mantis.ilias.de/view.php?id=29372


Amstutz:
Thx a lot, promoted to 7 and trunk as well.

PR #3043, opened 2020-12-09
GlobalScreen: Fix Mantis isue 29375 (PHP 8 final private method)

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=29375


chfsx (approved):
Thanks a lot!

PR #3044, opened 2020-12-09
Object/Container: Fix Mantis issues 29379 and 29380 (PHP 8 final private method)

mjansenDatabay:
Mantis Issues:
- https://mantis.ilias.de/view.php?id=29379
- https://mantis.ilias.de/view.php?id=29380


mjansenDatabay:
@alex40724 Any objections to merge this? Since we merged the huge 'PHP 8 PR' yesterday, it would be great if this is merged (and cherry-picked) as well.

PR #3048, opened 2020-12-09
Setup: Fix Mantis issue 29383 (PHP 8 Declaration not compatible with …

mjansenDatabay:
…base class)

Mantis Issue: https://mantis.ilias.de/view.php?id=29383


mjansenDatabay:
The solved issue also occurs with PHP 7.4.

klees:
We already changed this in the meantime:

https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/tests/Setup/ImplementationOfInterfaceFinderTest.php#L14
https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/tests/Setup/ImplementationOfInterfaceFinderTest.php#L14

Thanks anyway!

PR #3051, opened 2020-12-18
Fix 29469 deck card sizes

alex40724:
Fixes https://mantis.ilias.de/view.php?id=29469

### DIFF
--- a/src/UI/Implementation/Component/Deck/Deck.php
+++ b/src/UI/Implementation/Component/Deck/Deck.php
@@ -132,20 +132,70 @@ class Deck implements D\Deck
      */
     public function getCardsSizeSmallDisplays()
     {
-        switch ($this->getCardsSize()) {
-            case self::SIZE_XS:
-                return 2;
-            case self::SIZE_S:
-                return 4;
-            case self::SIZE_M:
-                return 6;
-            case self::SIZE_L:
-                // no-break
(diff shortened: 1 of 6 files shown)


Amstutz (approved):
Hi @alex40724 

Thank you for this. To make the implementation of the cards more flexible in the number of cards for individual screen sizes seems necessary. Your approach makes sense. I would definitely push this also to 7, as for 6 I am somewhat more catious, due to 6 already being productive. 

Thanks for taking care of the test as well. 

- [x] only thing missing is also adapting the comments in src/UI/Component/Deck/Deck.php . 

I see that you put a JF label to the linked ticket, so I guess, we can ask the JF while discussing the ticket, on how to proceed with merging. Thx a lot.

Amstutz:
Thx a lot @alex40724 . I adapted the comments in Deck.php after merging.

chfsx:
@Amstutz und @alex40724 thanks a lot for this fix! as discussed in JF this solution may be ported to ILIAS 6 as well, we would highly appreciate that!

Amstutz:
Hi @chfsx , as voiced last time, I have no strong opinion on whether we should port this back. If I remember, we decided on gathering some feedback on then make the final decision. If you already did run some test and feel ready to discuss the 6 port, drop me a message and ill tag it again for JF.

PR #3053, opened 2020-12-21
NotificationCenter: Fix Mantis issue 29500

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=29500

If approved, this has to be cherry-picked to `release_7` as well.


Amstutz (approved):
thx a lot!

PR #2833, opened 2020-08-10
Implemented new method to check the special chars

catlauraherzog:
I have implemented a new faster way to check the files for special chars. Before that the tool checked the whole project. Now it just checks the changed files from the base branch to the pull-request. The base branch is setted in the .travis.yml

PR #3056, opened 2020-12-27
CI: Added cheaper, faster and better way to check for special chars

catlauraherzog:
I've had some problems with my local GitHub repository due to several tests within some CI stuff for ILIAS so I lost the reference to a certain branch which was mentioned here: https://github.com/ILIAS-eLearning/ILIAS/pull/2833. So I had to redo the coding in a new fresh branch (release_6 as base) and push it to a new pull request. The old one is closed with this one and its todos are done.

---

Before this patch the special char checker walked over every single file within the pull request or standard tests to check for hidden chars. With this patch we only use the changed files of the current branch. With that the check is cheaper, faster and therefor better because it costs less minutes.


klees (changes requested):
Please rebase to trunk, everything else looks fine. Thx!

catlauraherzog:
Moved here: #3065

PR #3057, opened 2020-12-31
Smoke Tests for the UI Components Examples

Amstutz:
Hi @klees and everybody else taking interest in this

To finish this screwed 2020 year of on a positiv note, I thought it to be a good idea to tackle one of our todos on the roadmap of the UI Components. Glaring at the threateningly low unit test coverage in ILIAS in general, especially in regard to the upcoming task of migrating to PHP 8, I thought it to be a good idea to chose "Smoke-Tests for Example". The 4h estimate however, seemed to be rather on the optimistic side, however the change is really not as big as the almost 300 files changed might suggest you. So despair not while thinking about reviewing. I believe the amount canges of changes are still manageable. 

Here a list of the necessary changes:
- Since we have multiple examples using the same name, we have numerous redeclaration issues if loading them all in one run. To avoid that, I decided to introduce namespaces for the examples (note this is where to large part of the change comes from).
- Since the example rely on a completely initialized factory of UI Components, I decided to not create one only for tests (this seems error prone for the future) but use the one already in place in ilInitalization. However, to make this logic easier to re-use I factored it out. 
- Same for HttpServices
- For some examples helpers (e.g. providing data to be used) have been moved into the same file as the examples. Otherwise they would have been detected as examples and run (which would have caused in error, since nothing would have been rendered as string). See e.g. src/UI/examples/Table/Presentation/environment.php

And here some benefits of the change to maybe sugar the task of the review somewhat:
- All Examples are „Smoke Tested“, meaning, they all render some output without producing an error.
- For each Component we test now, that there exist at least one example show-casing it’s usage.
- The test coverage of the UI Components jumped from somewhat bellow 50% to almost 90% (271 examples are now all „smoke unit tested").
- The dependencies used in the examples are now explicitly listed in one point (setUpMockDependencies in tests/UI/Examples/ExamplesTest.php)
- InitUIFramework and InitHttpServices are now unit tested and easier to re-use since factored out of the horribly large ilInitialisation class. Especially InitUIFramework might come in handy for other tests having components of the UI Components as dependancy.
- The examples now all enforce strict_type=1 and use namespaces 
- Since the examples are now unit tested, they are much easier and quicker to debug (no browser needed). 
- Some broke examples have been detected and fixed or partially fixed (see comment in filter example, this seems to be a tough one and I am not happy with the fix or workaround, however completely fixing this issue is beyond the scope of this PR, see also the newly created ticker for this issue: https://mantis.ilias.de/view.php?id=29543)
- Several Undefined variable or Undefined index issues have been detected and fixed.

Further I observed that test might give us some valuable feedback on the performance of the various UI Components, since we can easily order the tests of the example by running time and so easily see which take long and maybe take a closer look at those to find an answer. Already clear seem to be, the template engine needs to change or be improved. Examples which have a large amount of different templates to be rendered seem to taker rather long. Most of the running time (almost 95%) seems to be rendering by ilTemplate. The template engine seems to be definitely one of the first points of focus if we try to improve the performance at some point.

This is implemented for trunk. Since there are not breaking changes in here, we could also think on porting this back to 7. However, if so, we should discuss it with the JF.

Feel free give feedback or suggest changes. You might also contact me via skype to dicuss questions.

Thx.


src/UI/Implementation/Crawler/Entry/ComponentEntry.php
```diff
@@ -497,14 +506,41 @@ protected function getCaseInsensitiveExampleFolder() : string
     public function getExamplesPath()
     {
         if (!$this->examples_path) {
-            $path_componants = str_replace("Component", "examples", $this->getPath())
+            $path_componants = str_replace("/Factory","",
```
klees:
```suggestion
            $path_components = str_replace("/Factory","",
```

src/UI/Implementation/Crawler/Entry/ComponentEntry.php
```diff
@@ -497,14 +506,41 @@ protected function getCaseInsensitiveExampleFolder() : string
...
+            $this->examples_namespace = str_replace("/", "\\",
+                str_replace("src/UI","\ILIAS\UI", $this->getExamplesPath()));
+        }
+        return $this->examples_namespace;
+    }
+
+    /**
+     * @return string
+     */
+    public function getNamesapce() : string
```
klees:
```suggestion
    public function getNamespace() : string
```

src/UI/Implementation/Crawler/Entry/ComponentEntry.php
```diff
@@ -497,14 +506,41 @@ protected function getCaseInsensitiveExampleFolder() : string
...
+    {
+        return $this->namesapce;
```
klees:
```suggestion
        return $this->namespace;
```

src/UI/Implementation/Crawler/Entry/ComponentEntry.php
```diff
@@ -497,14 +506,41 @@ protected function getCaseInsensitiveExampleFolder() : string
...
+    }
+
+    /**
+     * @param string $namesapce
+     */
+    public function setNamesapce(string $namesapce) : void
+    {
+        $this->namesapce = $namesapce;
```
klees:
```suggestion
        $this->namespace = $namespace;
```

src/UI/Implementation/Crawler/Entry/ComponentEntry.php
```diff
@@ -497,14 +506,41 @@ protected function getCaseInsensitiveExampleFolder() : string
...
+     */
+    public function getNamesapce() : string
+    {
+        return $this->namesapce;
+    }
+
+    /**
+     * @param string $namesapce
+     */
+    public function setNamesapce(string $namesapce) : void
```
klees:
```suggestion
    public function setNamespace(string $namespace) : void
```

src/UI/Implementation/Crawler/Entry/ComponentEntry.php
```diff
@@ -552,7 +588,8 @@ public function jsonSerialize()
             'parent' => $this->getParent(),
             'children' => $this->getChildren(),
             'less_variables' => $this->getLessVariables(),
-            'path' => $this->getPath()
+            'path' => $this->getPath(),
+            'namespace' => $this->getNamesapce()
```
klees:
```suggestion
            'namespace' => $this->getNamespace()
```

klees (changes requested):
Hi @Amstutz,

thanks a lot for taking care of this. Smoketesting our examples is a huge step IMO, and it seems as if there were certain other improvements to be made to get there. Kudos!

Besides the various, rather minor, inline comments I have one strong suggestion:

* [ ] Mockery: Replace Mockery by functionality native to PHPUnit where possible. Our unit test documentation recommends to [only use Mockery if strictly required](https://github.com/ILIAS-eLearning/ILIAS/tree/release_6/tests#mockery) and as far as I can see we could remove that additional dependency to our code completely.

I also have two comments that do not require any action (imo), but instead are remarks on things I have noticed  when looking into this PR:

* This makes, once again, very obvious how dependency injection will improve unit testing and make code simpler. Many changes in this PR are just rather editorial, but another large group of code is concerned with initializing the service locator `$DIC` correctly. I would be very surprised if this will prove to be very stable in the future. Still, we cannot do any better currently IMO.
* This also shows that, we will require some more structure in the initialisation process, as witnessed by the two new InitXYZ classes you have introduced. I think what you did is a good step forward, but we should look into the general pattern on how components should initialize their part of the DIC some time. Maybe this would be an interesting topic for the Dev Conf or some other workshop. I have started to work on an according paper, but did not manage to complete it until now. Maybe this also is some discussion that @smeyer-ilias as maintainer of Services/Init wants to initiate.

Best regards!

Amstutz:
Hi @klees 

Thx for the feedback. Fixed the typos. Note, the new hooks running the tests seem tow work fine, caught several issues with the corrected naming of the namespace function :-)

I removed mockery where possible, with one exception, ilPluginAdmin is still mocked with mockery due to static call of getActivePluginsForSlot. I hope this is fine.

I would like to merge this down to 7 if possible, so we do not run into conflicts between 7 and trunk when fixing examples. I can do the picking, since I caused multiple commits due to the typos issue mentioned above. Just ping me.

Thx

klees (approved):

Amstutz:
@smeyer-ilias 

There is also a small change in ilInitialisation.php. Can you have a short look at it an green light this before we merge. Thx a lot!

Lg

Amstutz:
@smeyer-ilias , we are running into conflicts here. The change in ilInitialisation are small, oke if we merge this?

Amstutz:
@smeyer-ilias ping.

smeyer-ilias (approved):
Thanks for that and moving the initialisation in a subdirectory.

PR #3058, opened 2021-01-07
Increase user login field to 255 characters

ghost:
https://mantis.ilias.de/view.php?id=29502

### DIFF
--- a/Services/User/classes/class.ilUserProfile.php
+++ b/Services/User/classes/class.ilUserProfile.php
@@ -47,7 +47,7 @@ class ilUserProfile
         "username" => array(
                         "input" => "login",
                         "maxlength" => 64,
-                        "size" => 40,
+                        "size" => 255,
                         "method" => "getLogin",
                         "course_export_fix_value" => 1,
                         "group_export_fix_value" => 1,
@@ -516,7 +516,7 @@ class ilUserProfile
                             $val->setValue($a_user->getLogin());
                         }
                         $val->setMaxLength($p['maxlength']);
-                        $val->setSize(40);
+                        $val->setSize(255);
                         $val->setRequired(true);
                     } else {
                         // user account name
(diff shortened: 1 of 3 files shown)


setup/sql/dbupdate_04.php
```diff
@@ -24790,4 +24790,14 @@ function writeCtrlClassEntry(ilPluginSlot $slot, array $plugin_data)
     }
     $setting->set('sty_media_cont_mig', 2);
 }
+?>
+<#5428>
+<?php
+global $DIC;
+$DIC->database()->modifyTableColumn("usr_data", "login", [
```
mjansenDatabay: @sragfw We MUST NOT add additional database update steps to `dbupdate_04.php`. Please add this schema change to the corresponding hotfix file for ILIA 5.4.x

ghost: Right, thanks for your hint

smeyer-ilias:
Thanks, will merged to other hotfix files.

Services/User/classes/class.ilUserProfile.php
```diff
@@ -47,7 +47,7 @@ class ilUserProfile
         "username" => array(
                         "input" => "login",
                         "maxlength" => 64,
```
ghost, line 49: @smeyer-ilias
Sorry, I found a missing needed change, the max length need also be increased
Could you add this too?

smeyer-ilias:
Yep, done

smeyer-ilias:
We have to reduce the column length due to https://mantis.ilias.de/view.php?id=29625
I 've added this report to the next jour fix agenda since the maximum column size for indexed table columns is currently not defined.

smeyer-ilias:
@sragfw : I reduced the login column to 190 chars, which should support our minimum index length (innodb engine with utf8mb4 and innodb_large_index = 0).

ghost:
190 chars are ok for the client

PR #3059, opened 2021-01-07
SPRG 29358, 29529, 29522: fixes writing/update of individual plan from program

nhaagen:
https://mantis.ilias.de/view.php?id=29358
https://mantis.ilias.de/view.php?id=29522
https://mantis.ilias.de/view.php?id=29529

There were several issues with updating individual plans based on program-settings.
I did not want to hook into ilStudyProgrammeUserAssignment::updateFromProgram because this is used quite often throughout the prg-code, so I separated updateValidityFromProgram/updateDeadlineFromProgram.


nhaagen:
threre are more issues; we'll wait.

PR #3060, opened 2021-01-07
OpenIdConnect: add maintainer in maintenance.json 

klees:
Hi @smeyer-ilias,

looking into various files and `git blame` I came to the conclusion that you are the maintainer here. Hence the change in maintenance.json. If this is not the case, please just pass back.

Thanks!


smeyer-ilias:
Thanks, merged.

PR #3073, opened 2021-01-13
UI/MainBar: give tool-button a more striking color

klees:
In the [ongoing discussion about the tools in the Main Bar](https://docu.ilias.de/goto_docu_grp_8739.html), the group (or at least parts of the group) came to the conclusion that we could alleviate problems with the tools easily if we would give a more striking color to the tools button. This, by no means, is the final conclusion how we would want to treat tools in ILIAS, but could help people to notice tools (when they are there) with simple means. The change is indeed simple, but IMO should help people already to notice the tools.

tools_colored

This does only help sighted people indeed, but does not make things worse for people that depend on screenreaders. The tool button is already marked with a color, we simply change that color. The contrast ratio from background to foreground should be enough to satisfy WCAG-criteria.

I currently propose this for the trunk only, but we might want to backport it to release_7 or even release_6.

### DIFF
--- a/src/UI/templates/default/MainControls/mainbar.less
+++ b/src/UI/templates/default/MainControls/mainbar.less
@@ -136,19 +136,19 @@
 .il-mainbar-tools-button {
 	.btn-bulky,
 	.btn-bulky.engaged {
-		background-color: lighten(@brand-primary, 10%);
+		background-color: @il-mainbar-tools-btn-bg;
 		border: 0;
-		color: @il-mainbar-btn-color;
+		color: darken(@il-mainbar-btn-color-engaged, 10%);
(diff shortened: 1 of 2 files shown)


Amstutz (approved):
Thx a lot for the work. Highly appreciated. Thx for making use of the existing variable brand-danger. This PR once again shows to me, that we still are not in a really good place concerning the color-system for text and backgrounds. However, this is nothing this PR can solve atm. Nothing to add here.

Amstutz:
I compiled the CSS and merged to trunk. Since this is labeled as improvement I only pushed this change to trunk and ported back to 7.

klees:
Thx!

PR #3074, opened 2021-01-14
0026267 - nachträgliche Parallelgruppen werden nicht übertragen

heimruf:
Parallel groups are not created if they are created after the first synchronisation. Maybe it would help to look up the last contentid instead of the first one?


smeyer-ilias:
Thanks for providing a PR for this issue,  but I have to close this report due to the reasons mentioned in https://mantis.ilias.de/view.php?id=26267

PR #3076, opened 2021-01-14
Exercise import fix

smeyer-ilias:
This is a fix for https://mantis.ilias.de/view.php?id=29713
Since R6 a protected  member variable $request is defined in ilObjectGUI

This PR renames the exercise request member variable.


smeyer-ilias:
Wrong base

PR #3078, opened 2021-01-14
BT-13407: Added Aria Label to search dropdown

M-Barz:
Mantis Report:
https://mantis.ilias.de/view.php?id=13407


smeyer-ilias (approved):
Thanks Marvin!

PR #3081, opened 2021-01-15
CI: Implemented github actions as replacement for travis

catlauraherzog:
Recently travis-ci changed their billing plan and basically made it impossible for open source projects to maintain unit tests or simple continuous integration features. With github actions we have a competitive feature-rich service directly from github - fully integrated within the standard github workflows.

This PR is here to start the switch from travis-ci to github actions. It needs some time to get every single script moved because the behavior and environment of github actions is slightly different from travis-ci. So I concentrated on the most basic things so that we can automatically check the Pull Requests again.

Following basic checks are implemented:
* PHP Unit Tests
* PHP CS Fixer
* Special Char Checker

The environment itself is configured that the PHP versions 7.2, 7.3 and 7.4 are tested.

About the reporting: I removed it in this step. It's more important to get the unit tests running again. The reporting needs more time to evaluate and will be implemented in future again.

@gvollbach I'd like to talk to you about the architecture of github actions and what's possible in the future. So first, you might look at the implementation here and then we talk through it.


.github/workflows/checks.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: composer install --no-interaction --no-progress
+
+      - name: PHP Unit Test 
+        run: CI/PHPUnit/run_tests.sh
+
+      - name: Install PHP CS Fixer
+        run: composer require friendsofphp/php-cs-fixer --no-interaction --no-progress
+
+      - name: PHP CS Fixer
+        run: CI/PHP-CS-Fixer/run_check.sh
```
catlauraherzog, line 32: I implemented the PHP CS Fixer directly here. Its runtime is around 10 to 12 minutes which is okay, if a pull request is opened which for me is okay.

CI/Import/Variables.sh
```diff
@@ -1,10 +1,13 @@
...
 TMP_DIR="/tmp"
 
+#CI_RESULTS_REPO="https://github.com/ILIAS-eLearning/CI-Results"
+CI_RESULTS_REPO="https://github.com/lauraquellmalz/CI-Results"
+
 PHP_CS_FIXER_RESULTS_PATH="$TMP_DIR/phpfix_results"
 PHP_CS_FIXER="libs/composer/vendor/friendsofphp/php-cs-fixer/php-cs-fixer"
 
-PHP_VERSION=$(phpenv version-name)
+PHP_VERSION=$(php -v | tac | tail -n 1 | cut -d " " -f 2 | cut -c 1-3)
```
catlauraherzog, line 9: Don't wonder please, I couldn't find an easier way to get the php version. This one is based on pure bash tools without any other software like phpenv in between. So we can move this script in the future somewhere else if needed.

klees:
Hi Jour Fixe!

There are two things I want to discuss with all of you in the meeting:

* I want to make this move and just want to check if anyone has objections. This will disable reporting for the moment, but we are looking to rebuild this later. First goal was to reinstate the checks for the PRs.
* This would be adding two dependencies to our project, which I think we should discuss publicly and decide on together: [actions/checkout@v2](https://github.com/actions/checkout) and [shivammathur/setup-php@v2](https://github.com/shivammathur/setup-php). Both dependencies are no dependencies in our production code but dependencies in our CI process. Still, the code provided by third parties will have significant access to our repository and thus should be subject to due diligence, just as other dependencies. IMO it is fine to add the two actions in question: One is provided by GitHub itself, which has access to our repos anyway. The other action is the (current) de-facto standard in the PHP-community (AFAIK), has recent commits and various contributors, where lokalheinz is the most popular one. So this is fine IMO.

Best regards!

matthiaskunkel:
Jour Fixe, 25 JAN 2021: We highly appreciate Laura's PR and accept it for versions 6, 7 and trunk. We ask the TB if it wants to call for an investigation on an alternative project to shivammathur/setup-php@v2 to reduce the risk of dependency in case this one-man-project comes suddenly to an end.

mjansenDatabay (approved):

PR #3084, opened 2021-01-15
Add Composer 2 Support for trunk (with rebuilt  composer.lock)

mjansenDatabay:
This PR adds Composer 2 support for our code base. To achieve this, I rebuilt the composer.lock file because of the existing dependency hell issues and corrupted versions of dependencies (and their dependencies) being upgraded explicitly from time to time in the past.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ addons:
     packages:
     - unzip
 install:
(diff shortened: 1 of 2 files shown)


matthiaskunkel:
We highly appreciate this PR and accept it for trunk.

PR #3086, opened 2021-01-18
bugfix for #29733 (allow image format 'svg' for tiles)

utesche:
Added image format 'svg' for uploading a tile image.
See bugreport:  https://mantis.ilias.de/view.php?id=29733


alex40724:
Thanks for the PR. Unfortunately it is not that easy (it does not work at all on my installation, due to convert processing). Please open a FR. There has also been some discussion of sympols/icons vs. images (bitmaps) during the original FR discussion.

utesche:
I wasn't aware of a 'convert issue' on permitting svg in general.  More and more we are using svg's as tile images due to its better scaling qualities, so I presume a patch coming for our own branch to limit 'svg uploads' to tiles only...

PR #2820, opened 2020-08-03
Apply headline schema

playamongthestars:
Same as #2812 

According to Mantis issue https://mantis.ilias.de/view.php?id=27327 this pull request applies the headline schema changes (see: https://docu.ilias.de/goto_docu_pg_118822_459.html) **as accepted** by the JourFix (see: https://docu.ilias.de/goto_docu_wiki_wpage_6475_1357.html)

As already discussed and outlined in the JourFixe discussions, this headline schema does **not** claim to be semantically correct, but is instead a more 'technical' approach which should be a real improvement in a first of probably multiple future steps to achieve a impeccable headline schema.

These changes do **not** cover content created by users (e.g. with the page editor)

If accepted, we will merge this into `trunk` first (see #2820) and leave this PR open for a while, so we have the option to fix `visual issues` (which might potentially arise, but we do not expect any).

# Changes for Headline Schema:
General changes depending to ilheader 
* templates/default/delos.less

Accordions:
* Services/Accordion/templates/default/tpl.accordion.html

Block: 
* templates/default/less/Services/Block/delos.less

Blogs:
* Modules/Blog/templates/default/tpl.blog_list.html 

Container:
* Services/Container/templates/default/tpl.container_list_item.html 
* Services/Container/templates/default/tpl.container_list_block.html

COPage:
* Services/COPage/css/content.css
* Services/COPage/css/content.less
* Services/COPage/templates/default/tpl.glossary_term_output.html
* Services/COPage/xsl/page.xsl 

Exercise: 
* Modules/Exercise/templates/default/tpl.assignment_head.html

Form:
* Services/Form/templates/default/tpl.property_form.html 

Glossary:
* Modules/Glossary/templates/default/tpl.glossary_definition_delete.html
* Modules/Glossary/templates/default/tpl.glossary_definition_list.html

LearningModule:
* Modules/LearningModule/templates/default/tpl.glossary_term_output.html

InfoSreen:
* Services/InfoScreen/templates/default/tpl.infoscreen.html
* templates/default/less/Services/InfoScreen/delos.less

Notes:
* Services/Notes/templates/default/tpl.notes_list.html 
* Services/Notes/templates/default/tpl.notes_list.html

Object:
* Services/Object/templates/default/tpl.creation_acc_head.html

UIComponent:

* Services/UIComponent/GroupedList/templates/default/tpl.grouped_list.html
* Services/UIComponent/Panel/templates/default/tpl.panel.html

---------------

Card:
[...]

PR #3087, opened 2021-01-19
UI Component Item, Listing Panel: Fixing Title, see #29616

Amstutz:
Proposal to fix: https://mantis.ilias.de/view.php?id=29616 .

This was a casualty of: https://github.com/ILIAS-eLearning/ILIAS/commit/c182f875521fa7ccca24e280cb59ea11dad0d60b , removing, changing the headline schema. This had impact on the CSS. To still identify the title as title by CSS, I surrounded it with a div.

I propose to push this to 6,7 and trunk. However, note the slight change in the tpl, skins could be affected.

### DIFF
--- a/src/UI/templates/default/Item/item.less
+++ b/src/UI/templates/default/Item/item.less
@@ -8,12 +8,8 @@
 	padding: @il-item-padding;
 	.clearfix();
 
-	h4 {
-		padding-top: 0;
-		margin: 0;
+	.il-item-title {
 		font-size: @font-size-base;
-		float: left;
-
 		.btn-link, a {
 			font-size: inherit;
 			line-height: @line-height-base;
--- a/src/UI/templates/default/Item/tpl.item_standard.html
+++ b/src/UI/templates/default/Item/tpl.item_standard.html
@@ -14,7 +14,7 @@
 		</div>
 		<div class="media-body">
 			<!-- END lead_start_icon -->
-			<!-- BEGIN title_link_start --><a href="{TITLE_HREF}"><!-- END title_link_start -->{TITLE}<!-- BEGIN title_link_end --></a><!-- END title_link_end -->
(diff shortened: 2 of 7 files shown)


klees (commented):
Hi @Amstutz,

I'm quite confident that this fixes the bug in question. However, I'm a bit irritated on the general direction this is taking. To make sure that I understand this correctly:

* We had a h5-element surrounding the title of the item before #2820 was merged.
* The element was simply removed, which caused the bug because CSS is not triggering as before.
* This now proposes to introduce a semantically weaker div-element instead of the former headline-element to fix the visual bug.

If this is correct so far, this irritates me on various levels:

* By virtue of a  "headline scheme" that was known to be problematic at the time of its introduction, we end up having a semantically weaker DOM than we had before the introduction of the schema.
* We now are forced to add a mere visual appearance to something that IMO clearly also has semantic meaning. Is this really where we want to be heading regarding accessibility?
* Instead of fixing the root cause (the problematic "headline scheme") we now tinker with a local fix for the Item. Will this end here? Or will we get new tickets for other components with the same problem or for new a11y problems at the same location?
* We are now both asked to deal with a problem that we did neither introduce nor cause, in a way that I cannot really support. Despite our reservations with the move that caused the problem, without funding or mere acknowledgement. Why should I deal with this?
 
I know that you are not the one who has caused this, but this feels wrong to me. Maybe we should at least bring this to the attention of the JF to prevent similar problems in the future.

Best regards!

Amstutz:
I fully agree with your analysis of the current situation. This was the pattern I could come up with to ship around the issue. I just pushed this to JF to discuss the issue there with our colleges.

matthiaskunkel:
Jour Fixe, 25 JAN 2021: We highly appreciate this PR and accept it for versions 6 and 7 and trunk. The PR reminds us to pick-up again the improvement of the current headline scheme.

PR #3088, opened 2021-01-19
update tinymce to 5.x

jeph864:


alex40724 (changes requested):
Hi,

thanks for the PR. First of all, thanks for taking care of this. I know that this is a messy thing to do. It's a lot of files, so I hope I did not lost overview. Just a quick question: The new sources are now located in  `Services/RTE/tiny_mce_5_6_0`, right? My assumption was that the lib would completely move to the `node_modules/` directory and `package.json` would be updated accordingly, see https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js/third-party-libs-npm.md

jeph864:
Hi,
I will update the code! 
However I have a question, In the past the RTE service provided  a function to choose versions and depending on versions different tinymce versions would be loaded(recently it was either 3.4.7 or 3.5.11). In the new version, all of the 3.x version code should be removed. Does it make sense if I remove this functionality and refactor the other components accordingly?

mjansenDatabay:
Hi @jeph864,

thanks for your question. In our opinion it makes sense to remove this functionality. There should only be one version of the TinyMCE, so feel free to refactor the components.

Best regards,
Michael

on behalf of the ILIAS e.V. and the Technical Board

jeph864:
Hi @alex40724 ,
The lib has been moved to node_modules/. Additionally the files 'package.json and package-lock.json' have been updated

alex40724 (commented):
Thanks! I just checked the page editor and everything seems to be fine with the 5.6 tiny. Please note that this is not a full review, just my personal comment related to my main dependent component.

kergomard:
Hi @jeph864

Thank you very much for your PR to update the TinyMCE. This is not a review in the proper sense of the word, but rather the result of our tests of the functionality. Please understand that the list below is probably not yet complete, but we would like to give you a feedback now, so that you can start bugfixing.

* The editor isn't shown, when answering essay questions in the test.
* Please remove the option to set captions of images or make sure they show up in the right place (for us they showed up on the right side of an image after saving).
* We are missing the source-view when creating an Assignment (ne to an "overflow: hidden" on the html and body tags. If ever you find a quick fix for this, this would be kind.

Thanks again and best,

The Technical Board of the ILIAS e.V

jeph864:
@swiniker 
After fixing the bugs, here are some remarks:

- We removed the option to set captions in images
- when answering essay questions in the test, the editor shows(it was showing even before, only the wordcounter was not working)!
- We were unable to reproduce the issue about source-view:  "**view-source**" is possible in 4 of the browsers used to reproduce this : Chrome, Safari, Microsoft Edge and Mozilla Firefox. It would help if you could tell us a way to reproduce this!

kergomard:
@jeph864 

First: Sorry, I just realized that I truncated the bullet points in the original feedback somehow. The full version would have been:
* The editor isn't shown, when answering essay questions in the test.
* Please remove the option to set captions of images or make sure they show up in the right place (for us they showed up on the right side of an image after saving).
* We are missing the source-view when creating an Assignment (not when answering it).
* We have an issue that the scrollbar is not shown in the editor field. This is not your fault, but is due to an "overflow: hidden" on the html and body tags. If ever you find a quick fix for this, this would be kind.

And then: Thank you very much for removing the captions and adding the word counter.

And finally for your feedback:
* Interesting: I can still reproduce the problem with the TinyMCE when answering essay questions on our original testing instance, but if I create a new instance with the current code it works. Good and Thank you!
* I still have the problem in Excercise Assignments. Please find a printscreen from Chromium below and a second one from FF. All of them made from the ilias7 branch. As last printscreen I addedd the current view, when creating an exercise assignment, ...and when adding this I also realized that actually also the link-controls are missing.
image
image
image
Sorry again form my mistake.
Thanks and best,
Stephan

jeph864:
@swiniker 

thanks to the pictures  in the previous comment, I have fixed the controls' issue (a few were missing and there were other unwanted ones):

- as you have shown in the two first pictures, there are alignment controls which weren't there in the previous versions! the current tinymce has renamed them and the function removing the alignment was still referring to the old names. I have corrected this.
- I have enabled the link, code and list controls! 


I'd like to make a remark about the pasteword-control: probably in the tests you have remarked that  there is no corresponding button in the toolbars as in the older versions. The pasteword-button  in the toolbar is no longer necessary! with tinymce 5.x you can paste the word contents directly in the editor and the editor will take care of the contents' sanitization!

Kind regards,
Jephte

alex40724 (approved):

alex40724:
Please note that I only checked my part (COPage use). This should be reviewd by TB members.

kergomard:
Thx @jeph864 !

sunweaver:
@swiniker, @jeph864: this PR introduces a regression for people that don't use TinyMCE. Here is a patch to amend that:

```
diff --git a/Services/Form/classes/class.ilTextAreaInputGUI.php b/Services/Form/classes/class.ilTextAreaInputGUI.php
index 7f6f276ad7..196fd0c81d 100755
--- a/Services/Form/classes/class.ilTextAreaInputGUI.php
+++ b/Services/Form/classes/class.ilTextAreaInputGUI.php
@@ -532,7 +532,9 @@ class ilTextAreaInputGUI extends ilSubEnabledFormPropertyGUI
                         $rte->addPlugin("link");
                         $rte->addPlugin("code");
 
-                        $rte->removeAllContextMenuItems(); //https://github.com/ILIAS-eLearning/ILIAS/pull/3088#issuecomment-805830050
+                        if (method_exists($rte, 'removeAllContextMenuItems')) {
+                            $rte->removeAllContextMenuItems(); //https://github.com/ILIAS-eLearning/ILIAS/pull/3088#issuecomment-805830050
+                        }
                         
                         // #11980 - p-tag is mandatory but we do not want the icons it comes with
                         $rte->disableButtons(array("anchor", "alignleft", "aligncenter",
```

PR #3089, opened 2021-01-19
update Tinymce to  5x 

jeph864:


kergomard:
Thank you very much @jeph864 . I found one more issue: When uploading images they are only shown after one safes the form containing the TinyMCE.
The problem is that the subfolder of an installation is not included when generating the preview img-url as the URL is an absolut path:
"/data/ilias7/mobs/mm_1180/blobid1617270152126.png"
grafik
As far as I can see it should be either:
"data/ilias7/mobs/mm_1180/blobid1617270152126.png" or "subfolder/data/ilias7/mobs/mm_1180/blobid1617270152126.png"

And one more thing: Could you please also resolve the conflicts. This way we can merge after this change.

Thanks and best!

kergomard:
Thx @jeph864 !

PR #3092, opened 2021-01-21
Correctly write CC attribution

habi:
As per https://creativecommons.org/licenses/by-sa/2.0/

Related to https://mantis.ilias.de/view.php?id=29366, e.g. the attribution of OpenStreetMap might *not* be correct anyways.


klees:
Thanks a bunch!

klees:
I also merged this to release_5-4, release_7 and the trunk.

PR #3108, opened 2021-01-25
29782: Content Snippets are not searchable

alex40724:
At least a partial fix for
https://mantis.ilias.de/view.php?id=29782


alex40724:
I will commit this with slight changes after discussion with Stefan.

PR #3115, opened 2021-01-26
Fix condition for .htaccess restriction

nicoroeser:
RedirectMatch is provided by mod_alias, not by mod_rewrite, see
<https://httpd.apache.org/docs/2.4/en/mod/mod_alias.html#redirectmatch>.

This properly blocks access to version-control files in deployments
where the web server has mod_alias enabled, but not mod_redirect.

This is the fix for trunk.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -12,7 +12,8 @@
 	RewriteCond %{HTTP_USER_AGENT} ^(DavClnt)$
 	RewriteCond %{REQUEST_METHOD} ^(OPTIONS)$
 	RewriteRule .* "-" [R=401,L]
-
+</IfModule>
+<IfModule mod_alias.c>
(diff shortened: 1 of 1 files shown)


Amstutz:
Thx looking good, since we do not have a maintainer, we shortly discuss that in an upcoming tb meeting. Thx for contributing.

klees:
In yesterdays TB-meeting we discussed this changes as an example for simple and obvious changes in files that are not clearly maintained by anyone. We decided, that we (the TB) will trust each other to merge these changes without further ado to safe each others precious time, if we are sufficiently sure that the change is fine.

I checked back with my admin, he found this change to be correct. Thus, I just merge this. Thx @nicoroeser, I'm looking forward to more contributions from you =)

PR #3116, opened 2021-01-26
Fix condition for .htaccess restriction

nicoroeser:
RedirectMatch is provided by mod_alias, not by mod_rewrite, see
<https://httpd.apache.org/docs/2.4/en/mod/mod_alias.html#redirectmatch>.

This properly blocks access to version-control files in deployments
where the web server has mod_alias enabled, but not mod_redirect.

This is the fix for release_7.


klees:
I checked back with my admin, he found this change to be correct. Thus, I just merge this. Thx @nicoroeser, I'm looking forward to more contributions from you =)

PR #3120, opened 2021-01-27
Setup 29552: extend GUI_CLASS_FILE_REGEXP to also handle window pathes

daniwe4:
https://mantis.ilias.de/view.php?id=29552

### DIFF
--- a/Services/UICore/classes/Setup/class.ilCtrlStructureReader.php
+++ b/Services/UICore/classes/Setup/class.ilCtrlStructureReader.php
@@ -343,7 +343,7 @@ class ilCtrlStructureReader
     // GUI CLASS FINDING
     // ----------------------
 
-    const GUI_CLASS_FILE_REGEXP = "~^.*/class\.(.*GUI)\.php$~i";
+    const GUI_CLASS_FILE_REGEXP = "~^.*[/\\\\]class\.(.*GUI)\.php$~i";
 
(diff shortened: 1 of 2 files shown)


klees:
Also picked to trunk.

klees:
Hi everybody!

This is what we came up with to show the status of the setup in the administration, as [requested by the JF](https://docu.ilias.de/goto_docu_wiki_wpage_6652_1357.html) and documented in [#29644](https://mantis.ilias.de/view.php?id=29644):

ilias-setup-status-administration

This is a very simple implementation that just uses the YAML output as also displayed in the console and puts in a legacy ui component with `<pre>`, wrapped by a report panel. We tried to use other UI-components (Characteristic Value Listings, some more elaborate structure of sub panels), but could not come up with a more convincing way to display that nested tree of status information. Seems we are missing an appropriate UI component at this point.

Currently I cannot see how our schedule at CaT and the ILIAS release schedule would allow to implement that missing UI component. From my POV it would still be ok to merge this change for two reasons: The page won't be viewed by many users often, but instead is something that some users (admins...) will look at sometimes (after updates, maybe?). Also, it is kind of similar to the page showing the PHP-info, which also isn't styled. Instead of being an important ILIAS page, this just shows some very technical output to some technical person. If we come up with better ways to show that nested tree, we will be happy to implement this for that view as well, but currently I do not feel that we would want to spend much time with this page.

Best regards!

PR #3122, opened 2021-01-27
Setup: 29644 show setup status in administration

daniwe4:
- add new subtab 'Status' for Administration/System Settings and
Maintenance/General Settings/Server
- gather status information and render as report

### DIFF
--- a/Modules/SystemFolder/classes/class.ilObjSystemFolderGUI.php
+++ b/Modules/SystemFolder/classes/class.ilObjSystemFolderGUI.php
@@ -4,6 +4,12 @@
 require_once "./Services/Object/classes/class.ilObjectGUI.php";
 require_once('./Services/Repository/classes/class.ilObjectPlugin.php');
 
+use ILIAS\Setup\Metrics;
+use ILIAS\Setup\ImplementationOfInterfaceFinder;
+use ILIAS\Setup\ImplementationOfAgentFinder;
+use ILIAS\Data\Factory;
+use ILIAS\Setup\CLI\StatusCommand;
+
 /**
  * Class ilObjSystemFolderGUI
  *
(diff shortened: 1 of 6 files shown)


klees:
Hi everybody!

This is what we came up with to show the status of the setup in the administration, as [requested by the JF](https://docu.ilias.de/goto_docu_wiki_wpage_6652_1357.html) and documented in [#29644](https://mantis.ilias.de/view.php?id=29644):

ilias-setup-status-administration

This is a very simple implementation that just uses the YAML output as also displayed in the console and puts in a legacy ui component with `<pre>`, wrapped by a report panel. We tried to use other UI-components (Characteristic Value Listings, some more elaborate structure of sub panels), but could not come up with a more convincing way to display that nested tree of status information. Seems we are missing an appropriate UI component at this point.

Currently I cannot see how our schedule at CaT and the ILIAS release schedule would allow to implement that missing UI component. From my POV it would still be ok to merge this change for two reasons: The page won't be viewed by many users often, but instead is something that some users (admins...) will look at sometimes (after updates, maybe?). Also, it is kind of similar to the page showing the PHP-info, which also isn't styled. Instead of being an important ILIAS page, this just shows some very technical output to some technical person. If we come up with better ways to show that nested list of status information, we will be happy to implement this for that view as well, but currently I do not feel that we would want to spend much time with this page.

Best regards!

klees (approved):

klees:
Hi @alex40724,

I assigned you as well because this introduces some changes in `Modules/SystemFolder`. Please have a look!

Best regards!

Amstutz:
HI @klees, hi @daniwe4 

Concerning the UI Components, IMO what we are missing is a component to display code in a formated way (pre) and maybe display some syntax highlighting (for json probably not neceassary). I wrestled myself with that in the UI Components example. Since I introduced in this step various other UI Components, I used the existing pattern to display code. IMO we should fix this gap soon. However, to not block this, IMO what you do is re-using an existing pattern, so this should be fine FTM. 

Greetings.

matthiaskunkel:
Jour Fixe, 08 FEB 2021: We highly appreciate this suggestion and accept the PR for 7 and trunk. Please label the new subtab "Installation Status" and make it the first entry in the row (presented when clicking tab "Server"). For the future, it would be great to have a UI component that could present such information or any other type of code as mentioned above by Timon. Once this UI component is available, it would be great if this page could re-use it. Thanks in advance.

klees:
The requested changes are included. Thanks @daniwe4!

alex40724 (approved):
Thanks!

klees:
I have picked this to trunk as well.

PR #3123, opened 2021-01-28
Setup: 29269 update installed lng files without having config

daniwe4:
- implement new Objective 'ilLanguagesUpdatedObjective'
- if no config is given for update command use ilLanguagesUpdateedObjective
- without config, the update command updates the lng files in any case, i couldn't find a way to check if an update is necessary at all (if smb. knows a way please contact me)

### DIFF
--- a/Services/Language/classes/Setup/class.ilLanguageSetupAgent.php
+++ b/Services/Language/classes/Setup/class.ilLanguageSetupAgent.php
@@ -80,7 +80,12 @@ class ilLanguageSetupAgent implements Setup\Agent
                 new ilDefaultLanguageSetObjective($config)
             );
         }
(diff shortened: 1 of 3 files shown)


klees:
Hi @katringross,

this is supposed to fix [#29269](https://mantis.ilias.de/view.php?id=29269). @daniwe4 did test the change on his local machine, the code LGTM. Still, this change is in the language-service where you are the new maintainer =)

Feel free to contact me if there is anything we can do to help you.

I'm very happy that you are on board!

fneumann:
No objection from my side. @katringross:  Richard Klees is much more familiar with the new setup and its support in the components. So I would trust his estimation.

katringross:
No objection from my side too. @daniwe4 thanks for your help.

PR #3124, opened 2021-01-28
Setup: 29220 getContactCity now returns corect property

daniwe4:
https://mantis.ilias.de/view.php?id=29220


klees:
Hi @alex40724,

this LGTM, thanks for looking into this!

Best regards!

PR #3126, opened 2021-01-28
fixed #27776: Missing Form Label in Permalink in Footer

alex40724:
https://mantis.ilias.de/view.php?id=27776


Amstutz:
Thx a lot for the fix!

Amstutz:
promoted to trunk and 7 as well.

PR #3127, opened 2021-01-29
Fix plugin filepath in ctrl structure (ILIAS 7)

ghost:
https://mantis.ilias.de/view.php?id=28906


ghost:
I close this because it should be fixed in https://github.com/ILIAS-eLearning/ILIAS/commit/1e05e1330fdde2abec0d21c3fd66009654402607

PR #3128, opened 2021-01-29
Fix plugin filepath in ctrl structure (ILIAS 8)

ghost:
https://mantis.ilias.de/view.php?id=28906


ghost:
I close this because it should be fixed in https://github.com/ILIAS-eLearning/ILIAS/commit/72009a884140800657a72c571bc4ca8dd0cb67af

PR #3129, opened 2021-01-29
fix plugins with namespaces in ctrl classes (ILIAS 7)

ghost:
https://mantis.ilias.de/view.php?id=29902

### DIFF
--- a/Services/UICore/classes/Setup/class.ilCtrlStructureReader.php
+++ b/Services/UICore/classes/Setup/class.ilCtrlStructureReader.php
@@ -365,7 +365,7 @@ class ilCtrlStructureReader
     // ILCTRL DECLARATION FINDING
     // ----------------------
 
-    const IL_CTRL_DECLARATION_REGEXP = '~^.*@{WHICH}\s+(\w+)\s*:\s*(\w+(\s*,\s*\w+)*)\s*$~mi';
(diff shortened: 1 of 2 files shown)


klees (changes requested):
Hi @sragfw,

LGTM in general, but please add testcases that cover your changes.

Thanks for looking into this!

ghost:
Another problem, there could be, at least for plugins

In https://github.com/ILIAS-eLearning/ILIAS/blob/43c2dac71205a5500f8d6bc3be28485d4539df37/Services/UICore/classes/Setup/class.ilCtrlStructureReader.php#L399 is checking, if there are multiple classes left from the `:`, this should work for `@ilCtrl_isCalls`

But not for `@ilCtrl_isCalledBy`, because they can be the opposite way, the current class chould be right from the `:`
So I think, for this, it should check, there are no multiple classes right from the `:`

But even it should check this so, there is a third problem, because currently it only takes the children from left, but before it was also possible to define the children right

What do you think?

klees:
Hi @sragfw,

thanks for your feedback. I don't think we need to worry about the problem you are mentioning. As far as I can see, the ilCtrl-declarations always are above the classes they declare something about. We even could think about removing the front part (before `:`) altogether, because the name the ilCtrl-declaration is about will always be repeated later when the class is declared in PHP. I don't see why we would want to support making ilCtrl-declarations about a class `A` on top of a class `B`.

Best regards!

PR #3131, opened 2021-02-01
Setup 27859: add lng var cache_activate_global_screen

daniwe4:
https://mantis.ilias.de/view.php?id=27859


klees:
Cherry-picked to release_6 as well.

PR #3133, opened 2021-02-02
CI: Implement github actions as replacement for travis (ILIAS 7)

catlauraherzog:
Same as in https://github.com/ILIAS-eLearning/ILIAS/pull/3081 just for release_7


klees:
Hi @chfsx,

we have moved some file in the GlobalScreen to make it comply to PSR-4. We assumed that the namespace is right, not the location.

Best regards!

klees:
Picked to trunk as well.

PR #3134, opened 2021-02-02
BT-29921: Added functionality for max_upload_file to filewizardinput, so the user can't go above it

M-Barz:
Link to Mantis-Report:
https://mantis.ilias.de/view.php?id=29921


smeyer-ilias:
Thanks Marvin
I opened a new one based on your implementation

PR #3136, opened 2021-02-02
fix access of protected variable from outside

Hadrian-II:
ilObjComponentSettingsGUI tries to access protected variable in ilPlugin, greated getter for variable and use that in ilObjComponentSettingsGUI


ghost:
Duplicates https://github.com/ILIAS-eLearning/ILIAS/pull/3099

ghost:
Can be closed - https://github.com/ILIAS-eLearning/ILIAS/pull/3099 was merged

Github can't detect the conflict if this should merged too (2 methods with same name), because changed in differend places of code

PR #3137, opened 2021-02-02
UI Item: withProgress Mutator

tfamula:
Feature Request: https://docu.ilias.de/goto_docu_wiki_wpage_6133_1357.html

- [ ] CSS needs some more love as you can see in the screenshot below

Item_Chart


Amstutz:
Hi @tfamula

Thx for this proposals. We definitely see the use of the Progress Meter in this view. Thx for providing an example. This helped quickly understanding the changes. Further, thx a lot for the added unit test!

However, there seem to be some things which we should wrap our head around first. Some questions:
- [ ] Progress Meter as Property I, Amount: You add the Progress Meter as property. This allows to add multiple. Do we really need/what multiple Progress Meters? When would we need those? If we have a view displaying the same content as cards, how would those multiple property meters be shown?
- [ ] Progress Meter as Property II, Law of Proximity: As you say, the CSS needs some work. However, we are not sure if the Progress Meter is placed right. If placed as property it looks like we are displaying the progress on some property of the Item (e.g. the Meta Data of Learning Module is 75% percent completed) and not the Progress of the Complete Component (e.g. the Learning Module is 75% completed). Does the Progress Meter not belong closer to the title?
- [ ] Progress Meter as Property III, Semantic: There is quite a semantical cap between the RepositoryObjectCard and the Item here. In the Repository Object card we have the withProgress and the getProgress Mutators. Do we not need them here? Again, this probably also comes down to the question, do we need multiple Progress Meter or just one and is the Progress Meter showing the progress of the whole Item or just some property of the item.

Please Change (after discussion of the above):
- [ ] Composition Description: No matter which way we go, the composition description needs some update hinting on the possibility of adding a Progress Meter into the update.
- [ ] Testrail: The added example will trigger some adaptations of the cases in Testrail. Please adapt those.

Since there are several non-trivial open questions, we advice to not discuss this in the JF at this stage. Maybe discuss the question here or even better, let's make an appointment in the UI Clinic, so we can discuss those via skype. Maybe @yseiler has also some thought to contribute to the points above.

Thx a lot for pushing this along.
@amstutz and @klees

tfamula:
Hi @Amstutz and @klees 
Thanks for your constructive feedback! I try to answer your questions:

> Progress Meter as Property I, Amount: You add the Progress Meter as property. This allows to add multiple. Do we really need/what multiple Progress Meters? When would we need those? If we have a view displaying the same content as cards, how would those multiple property meters be shown?

IMO, the use of multiple Progress Meters could give us even more overvalue. One possibility was mentioned by Alex in the Jour Fixe: For example we could use this in the "My Staff" view while showing the status of multiple Competence Profiles for one user. Each user would be represented in an Item and each Competence Profile of the user would be represented in one property. The key of the Property would contain the title of the Competence Profile and the value would contain the Progress Meter. The superior would have the opportunity to have a nice overview of the statuses of the employee's Competence Profiles. Maybe there would be even more use cases where we could benefit from multiple Progress Meters in one Item.
To be honest, I don't understand the last question here. If there would be a problem showing multiple Properties, why is it already possible to create such Items with Strings and Shy Buttons as Properties? Why would just adding the Progress Meter as a new possible Property make any problems here?

> Progress Meter as Property II, Law of Proximity: As you say, the CSS needs some work. However, we are not sure if the Progress Meter is placed right. If placed as property it looks like we are displaying the progress on some property of the Item (e.g. the Meta Data of Learning Module is 75% percent completed) and not the Progress of the Complete Component (e.g. the Learning Module is 75% completed). Does the Progress Meter not belong closer to the title?

Actually, I see the point here. If using just a single Progress Meter as Property for the Item, the title becomes a little bit needless. But I do not consider it as a big problem. What would be the alternative? Would you render the template differently, depending on the number of Properties (1 Property = with no title; >1 Property = with titles)? IMO, when the title is just named "Progress" or the like, it is quite clear that the whole Item is meant when using only one Property. 
In my eyes, both use cases should be possible when extending the Item with Progress Meter(s): Showing the progress of the whole Item or showing the progress of the Properties of the Item. It depends on whether you use 1 or more than 1 Progress Meter.

> Progress Meter as Property III, Semantic: There is quite a semantical cap between the RepositoryObjectCard and the Item here. In the Repository Object card we have the withProgress and the getProgress Mutators. Do we not need them here? Again, this probably also comes down to the question, do we need multiple Progress Meter or just one and is the Progress Meter showing the progress of the whole Item or just some property of the item.

I do not really understand why the composition of the RepositoryObjectCard has relevance here. As you correctly said, it has been decided to make the RepositoryObjectCard extendable by only one (Mini) Progress Meter. But why does we have to do the same in the Item? IMO, it would be even more complicated to have two mutators then, i.e. `withProgress` and `withProperties`, because mixing these two would maybe get more complicated. Mixing different Properties (Strings, Shy Buttons, Progress Meter) with just one mutator is much easier, for the technical side of the UI Framework as well as the consumer's side. And last but not least, this decision with the extra mutator would completely prevent the possibility to use more than one Progress Meter in an Item, like you already said.

> Please Change (after discussion of the above):
> - [ ] Composition Description: No matter which way we go, the composition description needs some update hinting on the possibility of adding a Progress Meter into the update.
> - [ ] Testrail: The added example will trigger some adaptations of the cases in Testrail. Please adapt those.

I will do.

> Since there are several non-trivial open questions, we advice to not discuss this in the JF at this stage. Maybe discuss the question here or even better, let's make an appointment in the UI Clinic, so we can discuss those via skype. Maybe @yseiler has also some thought to contribute to the points above.

That's a good idea! How can I propose it to the UI Clinic? I could not find enough information how I do this. Do i simply create a request [here](https://docu.ilias.de/goto_docu_dcl_8186_166.html) and wait for your answer? What should I put in as Date?


Greetings
Thomas

Amstutz:
> That's a good idea! How can I propose it to the UI Clinic? I could not find enough information how I do this. Do i simply create a request here and wait for your answer? What should I put in as Date?

Cool. Yes, pretty much. Fill in a date, so we get a rough idea for the timeframe you need answers, and then we will get back to you. It is probably best to discuss the above questions there and write up a summary then here for whoever might be interested.

Amstutz:
We discussed the PR in the UI Clinic today with @tfamula (https://docu.ilias.de/goto_docu_fold_8199.html ). We agreed that FTM it would be better to add just one Progress Meter with some withProgress Mutator. This gives a lot more liberties concerning the size and styling of the Progress Meter. We agree that the screenshot in the FR listed above in section 3.2 with the Progress Meter on the right side might be a good start. 

For the use case brought up by @tfamula, where one user (represented as an item) has progress in various competence profiles there would be another possible solution besides having various progress meters in properties. We somehow need the concept of "sub items" for other purpose, like preconditions or session material anyway. This would be something like a secondary listing panel, attached to an item. We then could look for a convincing solution to display progress meters in secondary listing panels and just reuse these panels as "sub items" for various use cases.

tfamula:
Hi @klees and @Amstutz !
I updated this PR according to our decision in the UI Clinic. This is ready for final review.

You find some screenshots listed below to show the rendering of the Progress Chart in different scenarios of the Item.

- Item with Progress:

progress_only

- Item with Progress and Lead Image:

progress_with_lead_image


- Item with Progress and Lead Icon:

progress_with_lead_icon

Amstutz (approved):
Hi @tfamula 

Thx a lot for the contribution. This got somewhat buried by workload for the Dev.Conf. and MV.

I do not like the media-right class, we should switch to flex sometime in the future with this component. However, this is definitely beyond the scope of this PR. Her media-right is in line with the rest of the structure currently used here. So, probably your only real option here.

IMO this is ready to be merged. @klees Feel free to do so.

Thx to everybody (@tfamula , @yvseiler , @klees ) involved!

Amstutz:
Hi @tfamula 

I merged this, since @klees is in holidays, so you can soldier on with your work. 

thx.

PR #3141, opened 2021-02-03
UI Components, Bulky Buttons: Fragments for Button, see: #28849

Amstutz:
We ignore fragments for Bulky Buttons, see: https://mantis.ilias.de/view.php?id=28849

This fixes the missing fragments for Bulky Buttons and adds some tests to make sure we process the Action as URI properly. However, IMO we should think about centralizing actions from URIs somewhat more. Having this in Bulkys seems wrong in the long run. If we use URI more widely we have such issues all over the place. Maybe add an according helper on the URI itself? 

However, maybe merge this before implementing the general fix. This needs to be merged down to 6,7 and trunk.

### DIFF
--- a/src/UI/Implementation/Component/Link/Bulky.php
+++ b/src/UI/Implementation/Component/Link/Bulky.php
@@ -30,6 +30,9 @@ class Bulky extends Link implements C\Link\Bulky
         if ($target->getQuery()) {
             $action .= '?' . $target->getQuery();
         }
+        if($target->getFragment()){
+            $action .= '#' .$target->getFragment();
+        }
         parent::__construct($action);
         $this->label = $label;
         $this->symbol = $symbol;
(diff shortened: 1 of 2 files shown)


klees (changes requested):
Hi @Amstutz,

LGTM in general, but I think you just want to use [`URI::__toString`](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/Data/URI.php#L385), which should also simplify the test you have written. This is already the helper you are looking for, right?

Best regards!

Amstutz:
lol, thx. I missed out on that one. Using to string now. Tests are somewhat redundant here I guess, however, I leave them as safeguard just if someone messes with the constructor in future and tries a more elaborate way of parsing the url ;-). May I hit rebase and merge?

klees (approved):
Go on!

PR #3142, opened 2021-02-03
IndividualAssessment 29867: change lng var in action menu

daniwe4:
- chande lng var 'edit' to 'settings' for action menu command edit


klees:
Picked to trunk as well.

PR #2639, opened 2020-04-07
UI, Icon: Render all Icons as img-tag

nhaagen:
This removes the surrounding div of icons and renders them, standard and custom, as img.
The aria-label is substituted by the alt-attribute according to https://github.com/ILIAS-eLearning/ILIAS/pull/2518.
The abbreviation-functions are still in the code, but do not show in the rendering besides a  data-attribute. In the future, the text might be amended to the default-svg.

PR #3143, opened 2021-02-03
UI, Icon: use img-tag for all icons

nhaagen:
- remove div-block around icon, do not show abbreviation anymore
 - move path-constants to renderer
 - ImagePathResolver for ComponentRenderers
 
rebased https://github.com/ILIAS-eLearning/ILIAS/pull/2639 to 7.


Amstutz:
Hi @nhaagen

Thx you very much for this important step. 287 Lines added, 1326 removed already gains my love and hints towards the much needed simplification, thx you. It simplifies the structure of Icons, resolves 2 A11y issues (by making use of alt and removing the invalid DOM structure of DIV in Button in Bulkies with Icons).

To me, there is one important question we should definitely discuss on the JF. @chfsx will probably also be interested in this:
- [x] Abbreviations: You removed the Abbreviations. As far as I know, the Global Screen still makes use for the for the Main Menu, if no Icon for an entry is provided. Can those really be removed? Note, the avatar is not affected by this change.

Further a question to anybody with expertise in A11Y
- [x] Aria-Label vs alt: Is it correct to completely remove the aria-label in favor of alt? It would be very frustrating if we have to change this back some weeks from now.

Things need to be done before merging:
- [x] Test Rail: we will need to check the test cases in testrail to see whether those changes affect them. 
- [ ] Exception Message: The exception Message in hasSymbolTrait is now slightly of, probably just remove the aria term completely, since an aria label is still a label, but a label used as tag is no aria label.

Thx Big Times.
@amstutz

chfsx:
Hi @nhaagen

As @Amstutz already mentioned, the abbreviations are used e.g. for self-created MainBar items, if no icon was uploaded. M.E. you can not simply remove them or then we would have to set on another fallback if a MainBar item has no icon. this happens in \ILIAS\GlobalScreen\Scope\MainMenu\Collector\Renderer\BaseTypeRenderer::getStandardSymbol

kergomard:
Just an idea, maybe a dumb one, sorry: Couldn't we solve the abbreviation by generating inline, base64 encoded svgs? What am I missing?

klees:
@nhaagen: Could you look into @swiniker's suggestion?

nhaagen:
Quite naive string-handling of xml, but it does the trick (preserving custom-icons)...

matthiaskunkel:
Jour Fixe, 08 FEB 2021: We highly appreciate this suggestion and accept the PR for 7 and trunk. In addition, it would be great to get a second PR that removes the image path resolver from ilUtil and just call it from system styles.

Amstutz:
Thx a lot @nhaagen and everybody joining into the discussion making this better.

PR #3144, opened 2021-02-03
(fix #29939) naming of startpage targets in personal settings

utesche:
Fixes naming of startpage targets in personal settings,
see mantis: https://mantis.ilias.de/view.php?id=29939


Services/User/classes/class.ilUserUtil.php
```diff
@@ -227,7 +227,7 @@ public static function getPossibleStartingPoints($a_force_all = false)
         
         $all = array();
         
-        $all[self::START_PD_OVERVIEW] = 'overview';
+        $all[self::START_PD_OVERVIEW] = 'obj_dshs';
```
smeyer-ilias, line 230: "obj_dshs" is the translation of the "Dashboard Administration Node". 
I suggest to use the language variable "dash_dashboard" instead from language module 'dash'

Please also add $lng->loadLanguageModule('dash');

Services/User/classes/class.ilUserUtil.php
```diff
@@ -248,7 +248,7 @@ public static function getPossibleStartingPoints($a_force_all = false)
             $all[self::START_PD_CALENDAR] = 'calendar';
         }
 
-        $all[self::START_REPOSITORY] = 'repository';
+        $all[self::START_REPOSITORY] = 'obj_root';
```
smeyer-ilias, line 251: Yes, that's better

smeyer-ilias (changes requested):
Thanks for fixing this and creating a PR!

Services/User/classes/class.ilUserUtil.php
utesche: ok,  I orientated my research to use a 'common' langvar, if possible.

alex40724:
Thanks for the PR. Fixed in git.

PR #3145, opened 2021-02-04
StudyProgramme 29856: add redirect and success message on adding auto…

daniwe4:
…matism


klees:
I have picked this to trunk.

PR #3146, opened 2021-02-04
StudyProgramme 29855: unify lng var using for automatism

daniwe4:


klees:
I have picked this to the trunk.

PR #3148, opened 2021-02-04
Maps 29366: add correct credit byline

daniwe4:
- found the correct notation here https://wiki.openstreetmap.org/wiki/Legal_FAQ/CC-BY-SA_Archive


klees:
Picked to trunk.

PR #3149, opened 2021-02-04
BT-28935: Replaced td with th in time column of week view in calendar

M-Barz:
Link to Mantis-Report:
https://mantis.ilias.de/view.php?id=28935


smeyer-ilias (approved):
Thanks, please merge to related branches.

PR #3151, opened 2021-02-05
UI/ModeInfo: extend ModeInfo (Maincontrol) to surround the entire page

nhaagen:
This suggests the ModeInfo maincontrol to surround the entire page,
taking into account that some modes may also modify other parts than the content.

ModeInfoAroundPage

### DIFF
--- a/src/UI/Implementation/Component/Layout/Page/Renderer.php
+++ b/src/UI/Implementation/Component/Layout/Page/Renderer.php
@@ -148,6 +148,7 @@ class Renderer extends AbstractComponentRenderer
             include_once("./Services/jQuery/classes/class.iljQueryUtil.php");
             array_unshift($js_files, './libs/bower/bower_components/jquery-migrate/jquery-migrate.min.js');
             array_unshift($js_files, \iljQueryUtil::getLocaljQueryPath());
+            $css_files[] = ['file' => './templates/default/delos.css'];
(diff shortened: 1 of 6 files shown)


klees (approved):
Hi @nhaagen,

it will come as no surprise that I like this =)

Your picture shows, though, that we might want to think about the coloring of the Tools and the Mode Info because they look as if they are related to each other. This already is a problem now, so from my perspective this would not block this PR here.

Best regards!

Amstutz (changes requested):
Hi @nhaagen 

Thx for tackling the issue. I approve the general approach, some minor less stuff should be refined though:

- [ ] Z-Index: You chose 10 000 this seems somewhat random and very high. E.g. the one from bootstrap for modals is only 1050. Probably introduce a less variable and reduce it. also check up on the "Z-index master list" in ilias/libs/bower/bower_components/bootstrap/less/variables.less .
- [ ] il-mode-info-height variable: IMO you removed all instances of the variable from less. Double check on that. If so, remove from variables.less as well.

@klees feel free to merge if those changes are applied.

Thx a lot!

nhaagen:
Hi @Amstutz ,
thanks for pointing that out. 
z-index is now a var, and I used il-mode-info-height to shove the logo a little lower for small screens.

PR #3154, opened 2021-02-05
StudyProgramme 29762: change lng vars

daniwe4:
- remove unacessary 'RefId' from lng var 'prg#:#category'
- rename 'Search' into 'Add'


klees:
Picked to trunk.

PR #3156, opened 2021-02-08
[FIX] 0029997: Wrongly placed DB-Step in release_7

chfsx:
As mentioned in https://mantis.ilias.de/view.php?id=29997 this PR should fix the wrongly placed DB step in release_7


chfsx:
merged locally

PR #3157, opened 2021-02-09
UI Components, Input Fields: Fix for issue, see e.g. #29826

Amstutz:
This fixes various a11y issues, all following the base template of: https://mantis.ilias.de/view.php?id=29826 .

We currently handle the "<Label for='...' " wrong by setting the for attribute to the name of the element and not the ID. This PR fixes this.

Note I included some further improvements, since the rendering tests but in some cases also the tpl files for inputs needed love:
-> We have some extremely long lines as rendering string outputs. Those are hard to adapt for a change like this. Adapting would have been a lot easier, if we had properly formated the html string output in tests to begin with. In various cases I changed to a well formatted multi line output (therefore we have several new lines).
-> There have been slight inconsistencies with ' vs " for attributes, where I observed those, I fixed.
-> We had some invalid xml by not closing <input properly with \>, I fixed where I observed one.


klees (changes requested):
Hi @Amstutz,

thanks a lot for your work and dedication. Without further ado:

- [x] Id Block: Why is the id a block in [this template](https://github.com/ILIAS-eLearning/ILIAS/pull/3157/files#diff-2926fea19e1429455978113bbbde49a60e40da725c947df79da13b08a95b34f3R1) but not the others?


- [x] Force Id: Please consider removing the option to pass an empty `$id` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3157/files#diff-f3892e7d92b0895c6b153d2f1754b755fc888a4b5199b1d4f8784a7344b81d5fR109). Since we now need the id to connect the fields to a label, I cannot think of a case where we would not need one.

- [x] Typehints: Add one on `string` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3157/files#diff-f3892e7d92b0895c6b153d2f1754b755fc888a4b5199b1d4f8784a7344b81d5fR109).

Best regards!

Amstutz:
Hi @klees 

Thx for looking into this.

Concerning your question, why non-consistent use of blocks: 

No good excuse here. Some had ID's mandatory before. Some in some cases (e.g. if used in filters), some not at all. I changed only if necessary. However, you are right, this led to an inconsistent state. We now require IDs everywhere in every case. Not exceptions there, no blocks needed anymore. Note that this led to some minor work in filters as well.

Note during making IDs more consistent and adapting tests, I observed some minor issue with multiple fields, this is now fixed as well.

Thx.

klees (approved):

klees:
I have picked this to release_7 as well.

PR #3082, opened 2021-01-15
Install.md fixes for 7, see #29725

Amstutz:
As reported in https://mantis.ilias.de/view.php?id=29725, the install guide is not up-to-date. This fixes some of the most important issues for updating to 7. I am sure there is further work needed.

PR #2978, opened 2020-11-06
Update install.md according to new release

fwolf-ilias:
Hi @ILIAS-eLearning/technical-board !

I would like to use this Pull request as platform to discuss some changes, which could be made to the installation page. The first commit is just about changing facts like release names and test server configuration.

1. Recommended Setup for Running ILIAS
 - Shouldn't we recommend any more recent OS here? 
 - Set OpenJDK to 8

2. PHP Installation/Configuration
 - This chapter should be moved before chapter "Build PHP-Dependencies and Artifacts"

3. WebDAV Configuration (OPTIONAL)
 - There is an configuration available to use php-fpm aside with webDAV as i remember correctly. An apache configuration could looke like this:
```
<FilesMatch  \.php$>
                        SSLOptions +StdEnvVars
                        SetEnvIf Authorization "(.*)" HTTP_AUTHORIZATION=$1
                        SetHandler "proxy:unix:/var/run/php/php7.3-fpm.sock|fcgi://localhost/"
</FilesMatch>
```
I don't have any experience with nginx so far. We should ask some nginx experienced admin for an example here.

4. Install other Depedencies
 - Remove Ubuntu 14.04 as it is outdated.

5. Installation Wizard
 - Has to be rewritten completely, as the setup page is not available anymore. My thoughts would be an explanation with an minimal config.json and a reference to the setup pages.

6. Upgrading ILIAS
 - I would like to change the link from github releases to the release page on docu.ilias.de as there are more informations avaiable.

7. Database Update
 - This needs also an rewrite to the new setup CLI. Should I do this or @klees ?

These are my few cents which our install.md could benefit from. If you have some ideas too, feel free to comment and discuss here and I will merge all approved ideas.

Regards,
Fabian Wolf

Systemadministrator
ILIAS open source e-Learning e.V.

PR #3158, opened 2021-02-09
Update and Improve install.md

klees:
Hi folks,

I had an intense look into the install.md and this is what I came up with. I restructured a lot of stuff, removed some outdated or duplicate information and incorporated information about the command line setup. This also includes the changes from #2978 and the problems adressed in #3082, at least I tried. I also removed the RFC-2119 language, because this is no RFC but a howto.

I would appreciate if someone other then me could have a long look into this as well.

Best regards!

### DIFF
--- a/docs/configuration/install.md
+++ b/docs/configuration/install.md
@@ -16,7 +16,7 @@
       1. [PHP Installation and Configuration](#php-installation)
       1. [Database Installation/Configuration](#database-installation)
          1. [MySQL Strict Mode \(5.6+\)](#mysql-strict-mode-56)
-         1. [MySQL Perfomance Tuning \(optional\)](#mysql-perfomance-tuning-optional)
+         1. [MySQL Performance Tuning \(optional\)](#mysql-performance-tuning-optional)
       1. [Install other Depedencies](#install-other-depedencies)
          1. [Optional Dependencies](#optional-dependencies)
    1. [Get the code](#get-code)
@@ -51,12 +51,17 @@ The necessary hardware to run an ILIAS installation is always dependent from the
 <a name="hardware"></a>
 ## Hardware
 
-We recommend a **common dual core server CPU**. Memory requirements are greatly
-variable, depending on the number of users and server activity. We recommend a
-minimum of **4096 MB of RAM**. We recommend **250 GB of harddrive** for usual
-ILIAS deployments. 25 GB would be used by the operating system and ILIAS itself.
-225 GB would remain for the database and files. For the web server WAN connection
-we recommend at least **100 Mbit/sec**.
+The hardware requirements for ILIAS vary widely, depending on the number of concurrent
+users you expect and the features you want to enable. Please be aware that ILIAS
+is not a webpage, but a highly interactive application, thus requirements will be
+higher than in case of the former. Snappiness of the system will highly depend on
+deploying enough resources and tailoring the system to your needs. In any case we
+recommend an absolute minimum of a common dual core server CPU, **4GB of RAM**,
+and a 100 Mbit/s internet connection. Roughly estimated the operating system and
+ILIAS itself will use around **25GB**. From there you can calculate your storage
+needs based on the amount of files and media content you expect to upload plus a
+few GBs for the database.
+
 
 <a name="recommended-setup-for-running-ilias"></a>
 ## Recommended Setup for Running ILIAS
@@ -125,7 +130,7 @@ The ILIAS Testserver (https://test7.ilias.de) is currently configured as follows
 ## Other Platforms or Configurations
 
 Please note that other platforms and configurations should be possible, but it
-might be harder to find someone who can help when things go south. You shall not
+might be harder to find someone who can help when things go south. You should not
 use a different configuration unless you are an experienced system administrator.
 
 
@@ -203,15 +208,8 @@ systemctl restart httpd.service
 <a name="php-installation"></a>
 ### PHP Installation and Configuration
 
-On Debian/Ubuntu execute:
-```
-apt-get install libapache2-mod-php7.3 php7.3-gd php7.3-mysql php7.3-mbstring php7.3-curl php7.3-dom php7.3-zip php7.3-xml
-```
-
-On RHEL/CentOS execute:
-```
-yum install php
-```
+Refer to the to documentation of your installation to install either PHP 7.3 to
+PHP 7.4 including packages for gd, mysq, mbstring, curl, dom, zip and xml.
 
 To check if the installation was successfull create the file `/var/www/html/phpinfo.php`
 with the following contents:
@@ -225,7 +223,7 @@ Then point your browser to ```http://yourservername.org/phpinfo.php```. If you s
 the content of the file as shown above your configuration is **not** working. If
 you can see details of your PHP Configuration everything works fine. Search for
 the entry ```Loaded configuration file``` as we now made some changes to it (e.g.
-`/etc/php5/apache2/php.ini`).
+`/etc/php5/apache2/php.ini`). Delete the file `phpinfo.php` afterwards.
 
 We recommend the following settings for your php.ini:
 
@@ -375,11 +373,12 @@ commands:
 
 ```
 cd /var/www/html/
-git clone https://github.com/ILIAS-eLearning/ILIAS.git .
-git checkout release_7
+git clone https://github.com/ILIAS-eLearning/ILIAS.git . --single-branch
+git checkout release_X
 ```
 
-or unpack the downloaded archieve to the docroot
+or unpack the downloaded archieve to the docroot. Replace `release_X` with the
+branch or tag you actually want to install.
 
 The repository of ILIAS doesn't contain all code that is required to run. To
 download the required PHP-dependencies and to create static artifacts from
@@ -401,9 +400,7 @@ chown www-data:www-data `/var/www/html
 chown www-data:www-data `/var/www/files
 ```
 
-If you want to server other applications from your webserver, you might want to
-use subdirectories in the docroot and the files-directory for ILIAS. The commands
-above will directly serve ILIAS from the docroot.
+The commands above will directly serve ILIAS from the docroot.
 
 <a name="install-ilias"></a>
 ## Install ILIAS
@@ -456,10 +453,11 @@ configuration might look like this afterwards:
 }
 ```
 
-Run the ILIAS command line setup with your configuration file from your ILIAS folder:
+Run the ILIAS command line setup from within your ILIAS folder with your configuration
+file (located outside your doc-root!) as a parameter:
 
 ```
-php setup/setup.php install my-configuration.json
+php setup/setup.php install /foo/bar/my-configuration.json
 ```
 
 The installation will display what currently happens and might prompt you with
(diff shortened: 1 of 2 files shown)


Amstutz (approved):
Hi @rklees

Thx very much for this much needed work-over of the install docs. Much appreciated. I am no sysadmin, so I can not give much feedback on large parts of the content. Hover, I like the general direction towards which you pulled the install docs. IMO they seem a bit more manageable this way.

Consider the following suggestions:
- [x] Release numbers: Refrain from using exact release numbers anywhere. They tend to get outdated. Instead of release_7, use release_X (maybe with the comment the X should be replaced for the targeted number.
- [x] System Styles: On old line 573 you removed an old pointer to some section which was not helpful in terms of system styles. Thx. However, it would be neat to replace it with the really helpful docs at: [Custom
Styles](/templates/Readme.md#custom-styles) . They tend to be overlooked. So an additional pointer here would be neat.
- [x] Current DB step: I personally found the tip concerning checking up on the current DB step really neat:To check which update step gets currently executed run the following SQL-Statement on your ILIAS database: `SELECT * FROM settings WHERE keyword = "db_update_running"`. As long as the script is not spitting out those numbers, I would keep it here. Can really save the day.

Take or leave, but mainly thx very very much.

docs/configuration/install.md
```diff
@@ -1,39 +1,47 @@
 # ILIAS Installation
 
-ILIAS is a powerful Open Source Learning Management System for developing and realising web-based e-learning. The software was developed to reduce the costs of using new media in education and further training and to ensure the maximum level of customer influence in the implementation of the software. ILIAS is published by ILIAS open source e-Learning e.V. under the General Public Licence and free of charge.
+Around since 1998, ILIAS is a powerful learning management system that fulfils
```
kergomard:
```suggestion
Since around 1998, ILIAS is a powerful learning management system that fulfils
```

docs/configuration/install.md
```diff
@@ -56,141 +62,99 @@ ILIAS is a powerful Open Source Learning Management System for developing and re
...
-<a name="harddrive"></a>
-## Harddrive
-
-We RECOMMEND 250 GB for usual ILIAS deployments. 25 GB would be used by the operating system and ILIAS itself. 225 GB would remain for the database and files.
-
-<a name="bandwidth"></a>
-## Bandwidth
-
-We RECOMMEND at least 100 Mbit/sec. for the web server WAN connection.
+We recommend a **common dual core server CPU**. Memory requirements are greatly
```
kergomard: Do we really want this block? Personally I would never run ILIAS on a setup like this unless I'm running a very small instance with all goodies switched off. Our instance is not huge, but I'm still counting around 50 Cores, multiple Terabytes of data (without counting RAIDs and failovers) and 100's of MB of RAM. Additionally running ILIAS on a hard drive will be painful and clearly not comply with today's demands for snappiness. 100 Mbits does also sound slightly 1990's ;-).

Just a suggestion from where I stand we can also simply throw it out:
"The hardware requirements for ILIAS vary widely, depending on the number of concurrent users you expect and the features you want to enable. Please be aware that ILIAS is not a webpage, but a highly interactive application, thus requirements will be higher than in case of the former. Snappiness of the system will highly depend on deploying enough resources and tailoring the system to your needs. In any case we recommend an absolute minimum of a **common dual core server CPU**, **4GB of RAM**, and a **100 Mbit/s** internet connection. Roughly estimated the operating system and ILIAS itself will use around 25GB. From there you can calculate your storage needs based on the amount of files and media content you expect to upload plus a few GBs for the database.

docs/configuration/install.md
```diff
@@ -56,141 +62,99 @@ ILIAS is a powerful Open Source Learning Management System for developing and re
...
-download the required PHP-dependencies and to create static artifacts from
-the source, run the following in your ILIAS folder:
+<a name="other-platforms"></a>
+## Other Platforms or Configurations
 
-```
-composer install --no-dev
-```
+Please note that other platforms and configurations should be possible, but it
+might be harder to find someone who can help when things go south. You shall not
```
kergomard: Sounds a little bit like Gandalf :-)
```suggestion
might be harder to find someone who can help when things go south. You should not
```

docs/configuration/install.md
```diff
@@ -249,31 +214,34 @@ On RHEL/CentOS:
...
-<a name="php-installationconfiguration"></a>
-## PHP Installation/Configuration
+<a name="php-installation"></a>
+### PHP Installation and Configuration
 
-On Debian/Ubuntu 14.04 or 16.04 execute:
+On Debian/Ubuntu execute:
 ```
-apt-get install libapache2-mod-php7.1 php7.1-gd php7.1-mysql php7.1-mbstring php7.1-curl php7.1-dom php7.1-zip php-xml
+apt-get install libapache2-mod-php7.3 php7.3-gd php7.3-mysql php7.3-mbstring php7.3-curl php7.3-dom php7.3-zip php7.3-xml
```
kergomard: I would recommend to get rid of this. It is hard to keep track of and changes can be sudden. Additionally, if you need this instructions and follow them by the letter, you may end up with a system including eol packages really quickly.

My suggestion:
```suggestion
Please refere to the documentation of your installation to install either php7.3 or php7.4 including the gd, mysql, mbstring, curl, dom, zip, and xml modules. 
```

docs/configuration/install.md
```diff
@@ -502,93 +541,223 @@ ExecStop=/usr/bin/java $JAVA_OPTS -jar $ILSERVER_JAR $ILSERVER_INI stop
...
+```
+
+<a name="webdav-mac-finder"></a>
+### WebDAV with Mac Finder
+To upload files, the WebDAV Client *Finder* on Mac uses chunked transfer encoding.
+Some webservers can't handle this way of uploading files and are serving ILIAS
+an empty files, which results in an empty file object on ILIAS. Due to a bug in
+apache, the configuration of **Apache with PHP-FPM** does not work with the *Mac
+Finder*. If you use WebDAV on your ILIAS installation, we recommend to either
+use **Apache with mod_php** or **Nginx with PHP-FPM (> 1.3.8)**.
```
kergomard:
```suggestion
use **Apache with mod_php** or **Nginx with PHP-FPM (> 1.3.8)**. Please find more information in the corresponding [README](../../Services/WebDAV/README.md)
```
...and then we get rid of the rest of the information in here concerning WebDAV. We might want to do this for all optional elements to keep this file short and hand responsibility over to maintainers. What do you think?

docs/configuration/install.md
```diff
@@ -361,108 +333,169 @@ mysql -i -BN -e 'SELECT @@sql_mode' | grep -E 'ONLY_FULL_GROUP_BY|STRICT_TRANS_T
...
+Create a directory outside the web servers docroot (e.g. `/var/www/files`). Make
+sure that the web server is the owner of the files and directories that were created
+by changing the group and owner to www-data (on Debian/Ubuntu) or apache (on RHEL).
+
+```
+chown www-data:www-data `/var/www/html
+chown www-data:www-data `/var/www/files
+```
+
+If you want to server other applications from your webserver, you might want to
```
kergomard:
```suggestion
If you want to serve other applications from your webserver, you might want to
```
...but I would actually suggest to kick that part out and only keep "The commands above will directly serve ILIAS from the docroot." I don't see many good reasons to want to serve ILIAS from subdirectories. Certificates are free (or at least getting cheaper if you want e.Vs.) and subdomains seem by far more elegant to me.

docs/configuration/install.md
```diff
@@ -361,108 +333,169 @@ mysql -i -BN -e 'SELECT @@sql_mode' | grep -E 'ONLY_FULL_GROUP_BY|STRICT_TRANS_T
...
-To prevent this behaviour, add the following rewrite rules to a .htaccess file in your webroot or to the corresponding section of the configuration of your webserver:
+Clone the code to the web servers docroot (e.g. `/var/www/html`) with the following
+commands:
 
 ```
-RewriteCond %{HTTP_USER_AGENT} ^(DavClnt)$
-RewriteCond %{REQUEST_METHOD} ^(OPTIONS)$
-RewriteRule .* "-" [R=401,L]
+cd /var/www/html/
+git clone https://github.com/ILIAS-eLearning/ILIAS.git .
```
kergomard: We could suggest to use --single-branch here. For me the difference when testing was 500MB of dead waste.

kergomard (commented):
Thank you very much @klees for all the work! I added comments inline, as this is not really code, but if you prefer them in our standardized form here at the end. I will do so, just let me know.

docs/configuration/install.md
```diff
@@ -361,108 +333,169 @@ mysql -i -BN -e 'SELECT @@sql_mode' | grep -E 'ONLY_FULL_GROUP_BY|STRICT_TRANS_T
...
+        "path_to_zip" : "/usr/bin/zip",
+        "path_to_zip" : "/usr/bin/unzip"
+    }
+}
+```
+
+Run the ILIAS command line setup with your configuration file from your ILIAS folder:
+
+```
+php setup/setup.php install my-configuration.json
```
kergomard:
```suggestion
php setup/setup.php install /foo/bar/my-configuration.json
```

docs/configuration/install.md
```diff
@@ -361,108 +333,169 @@ mysql -i -BN -e 'SELECT @@sql_mode' | grep -E 'ONLY_FULL_GROUP_BY|STRICT_TRANS_T
...
+	},
+    "utilities" : {
+        "path_to_convert" : "/usr/bin/convert",
+        "path_to_zip" : "/usr/bin/zip",
+        "path_to_zip" : "/usr/bin/unzip"
+    }
+}
+```
+
+Run the ILIAS command line setup with your configuration file from your ILIAS folder:
```
kergomard:
```suggestion
Run the ILIAS command line setup from within your ILIAS folder passing your configuration file as a parameter:
```

kergomard (commented):
Making it clearer that the config.json has to be outside the webroot. I landed on /foo/bar as a folder, as I was unsure what else to pick: /opt/, /etc/ilias/, /home/ilias ?

docs/configuration/install.md
klees: I'll simply remove this, probably does not belong in installation instructions anyway.

docs/configuration/install.md
klees: Something like this would be great.

klees:
Hi @Amstutz,

the doc for the custom styles is (and was) linked here:

https://github.com/ILIAS-eLearning/ILIAS/pull/3158/files#diff-703d97b1d89d3339a2aab21d2d33190b3b9fa8b044efd785a17ff91ba1cc63e6R594

Was it something else you meant?

Best regards!

Amstutz:
You are right concerning the custom styles, got confused with the deleted headline above ;-).

I believe this is an important step. It is sad that there has not been more feedback on this important work. However, I guess that silence means approval. I would support merging this important contribution.

Thx a lot for doing that.

klees:
Hi Jour Fixe,

there is nothing to decide for you here. I still want to bring this to the attention of the JF: We have a problem with our documentation. Nobody seems to really feel responsible for such an important document like the installation instructions. We need to do better here.

Best regards!

matthiaskunkel:
Jour Fixe, 08 MAR 2021: Thanks for pointing us at this issue. A feasible and sustainable solution for maintaining the installation documentation is highly appreciated and important to extend the user base of ILIAS. We are looking forward to the suggestions made by the TB at the upcoming DevConf.

klees:
I have picked this to trunk.

PR #3159, opened 2021-02-09
add max-length to text inputs

Hadrian-II:
Add Max Length property to text inputs, so input can be limited to a fixed amount of characters. 
Sets maxlength property in input to selected value.


klees (changes requested):
Hi @Hadrian-II,

thanks a lot for contributing to the UI-framework of ILIAS.

This addition looks very plausible and we will be happy to take it to the Jour Fixe.

- [ ] Validation: Currently the code does not protect from incoming strings that are longer than indicated. This will either be surprising for the consumers of the API, or they will need to add an according validation from the Refinery manually. We cannot think of a case where someone would want to indicate a max-length on the frontend but does not care about the length of the received string. Please add the according check when a max-length is set.

- [x] Check: Remove [that check](https://github.com/ILIAS-eLearning/ILIAS/pull/3159/files#diff-b4c6a2119d210e5d7648fae314d964818affab15a8ec8830011c8633d26c2083R38), the `checkXY`-methods are from a time without typehints on primitive types. Since typehints on `int` etc. were introduced, we don't need the checks anymore.
- [x] Docstrings: Remove the docstrings about types [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3159/files#diff-e76641f9ef3f91856e7953fcfa3fa5309f706f64162c664f29e8c5d303ff2538R15) and [there](https://github.com/ILIAS-eLearning/ILIAS/pull/3159/files#diff-e76641f9ef3f91856e7953fcfa3fa5309f706f64162c664f29e8c5d303ff2538R22). Thanks to the typehints on primitives and return type hints, they do not add any information and instead repeat something that is already visible in code. We decided to only use docstrings if they add information. Otherwise they would just be something that potentially goes out of sync with the code but has not benefit.
- [x] Test: Add an automated test for the property and a test for the rendering. This is not necessary before we take this to the JF, but we will need these test before merging.

Since all these things are not relevant for the JF, we will already set the label. If you do not want to present the PR yourself, we will be happy to do that for you.

Best regards!
@Amstutz and @klees

Hadrian-II:
Added test and validation, i hope it is now ok.
I gladly skip the honor to be at the JF so that i dont have to wait for hours to say a sentence :)

matthiaskunkel:
Jour Fixe, 22 FEB 2021: We highly appreciate this suggestion and accept the PR for trunk.

klees:
Hi @Hadrian-II,

please rebase this and insert the missing validation, so we can merge move forward.

I see, that you wanted to include the validation, but the location (`isClientSideValueOk`) is not the one. [`isClientSideValueOk`](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/UI/Implementation/Component/Input/Field/Input.php#L239) checks, "if the value is good to be displayed client side". That is, it is a hard error, maybe even a sign that someone tried to temper with the system, when that method fails. It is good as it was before. Even if someone uses a string longer than required, the value she set in the field can (and should) still be displayed again, so the problem can be examined.

The requested change instead should be done added as a transformation via [`withAdditionalTransformation`](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/UI/Implementation/Component/Input/Field/Input.php#L295) or [`setAdditionalTransformation`](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/UI/Implementation/Component/Input/Field/Input.php#L314). This will allow the machinery to produce a proper error to be displayed to the user. You could even use the prebuild transformation [`hasMaxLength`](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/Refinery/String/Group.php#L54) then.

Feel free to ask for further support or advice, here or via skype, as always.

Best regards!

Hadrian-II:
Hm, did rebase create carnage, or is this how it is supposed to look?

Hadrian-II:
see https://github.com/ILIAS-eLearning/ILIAS/pull/3232 for clean pull request

klees:
Let's use the clean one.

PR #3160, opened 2021-02-10
partly fixed 29439: Failed test: Massenkategorie mit Standardfilter f…

alex40724:
…iltern, see https://mantis.ilias.de/view.php?id=29439

### DIFF
--- a/src/UI/templates/default/Input/filter.less
+++ b/src/UI/templates/default/Input/filter.less
@@ -66,6 +66,10 @@
 	}
 }
 
+.il-filter.disabled .il-filter-bar-controls .btn-bulky {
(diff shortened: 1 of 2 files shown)


Amstutz:
Hi @alex40724 

Sry for letting this sit for some time. Currently there is a rather large set of open filter issues. @enrico is currently sorting through them as part of the PLR III project that he is managing. @yvseiler and Oliver are also involved in refining and simplifying the concept. Therefore I checked back, with @yvseiler and @ezenzen on this here. They both approve this and think it to be a step in the right direction.

So thx a lot, Ill merge and push to 7 and trunk.

PR #3164, opened 2021-02-10
Remove lng duplicates + wrongly placed translation

tfamula:
Hi @matthiaskunkel,
I noticed some discrepancies (e.g. duplicates) in the language files when updating the languages. The Git History tells me that you introduced these variables. Please check and merge if the changes are correct.

The wrong translation for `mob_url_info` also occurs in ILIAS 7. I created a separate PR for this: #3165

PR #3165, opened 2021-02-10
Fix wrongly placed translation in language files

tfamula:
See #3164


matthiaskunkel:
Merged PR by @tfamula to improve language files

PR #3167, opened 2021-02-11
implemented new filesystem for node-modules

chfsx:
Hi @alex40724 

As needed in https://mantis.ilias.de/view.php?id=29332 I implemented a new FileSystem for node_modules. 
of course we should be careful not to introduce too many file systems, but in this case there is no way around it. you may merge whenever you want.

best regards

Fabian


alex40724:
Thanks a lot, Fabian!

klees:
This looks really nice and simple!

PR #3168, opened 2021-02-11
UI/Standard Page: clarify that BASE only exists for examples (#29414)

klees:
Hi @Amstutz,

this is supposed to clarify affairs regarding [#29414](https://mantis.ilias.de/view.php?id=29414).

Best regards!


Amstutz:
Thx a lot! picked to 7 and 6 as well.

PR #3172, opened 2021-02-12
IndividualAssessment 29284: allow null as return value

daniwe4:
- allow null return value for method 'getEventTime'


klees:
Picked to trunk.

PR #3173, opened 2021-02-12
IndividualAssessment 29287: set title, description during import

daniwe4:
- remove wrong casting for getTitle and getDescription in method
'importRecord'
- move setTitle and setDescription before creating settings object


klees (approved):

klees:
Picked to release_6 and trunk.

PR #3175, opened 2021-02-12
IndividualAssessment 29250: send success after saving settings

daniwe4:


klees (approved):

klees:
Picked to trunk and release_6.

PR #3177, opened 2021-02-12
UI Components, Button and Links: Aria Roles and Attributes

Amstutz:
There are several issues with the aria-role=„menuitem“ for bulky-links and buttons in the meta and mainbar:

- They are missing completely for all bulky links, which is wrong, since they should be listed as individual menuitem. Note that the interface currently does not allow links to carry an aria role.
- Buttons which are not triggering a slate currently lack the role Menuitem in Main and Metabar.
- Menuitems can NOT cary the the aria-pressed or aria-checked attribute, see inherited states and properties: https://www.w3.org/TR/wai-aria-1.1/#menuitem . Therefore those have been removed, if BulkyButton acts as Menu Item.
- Attribute aria-haspopup should only be set, if there really is some slate opening. This should be the case if the Button is an engeagable Menu Item, otherwise, if it acts only as Button (as e.g. the Case for the Trigger of the Online help) it should not carry this attribute.

The issues are fixed by:

- Adding the option to add aria-roles for links implicitly. Note that they are not provided on the public interface. Outside the Main/Metabar I currently see no use case. So they are only offered internally (as with bulky buttons already the case).
- Buttons without triggering Slates and Links now also cary the role menuitem.
- Some logic for properly setting the aria-checked and aria-haspopup attributes.
- Adapt/add the according tests

Related issues:

- https://mantis.ilias.de/view.php?id=29841
- https://mantis.ilias.de/view.php?id=29848
- https://mantis.ilias.de/view.php?id=29849

Glad to have Unit tests, would not have stumbled over the aria-haspopup issue otherwise.


klees (changes requested):
Hi @Amstutz,

I'm still not convinced that "menuitem" really is the correct label here, as the Main Bar is not "the menu". This clearly is something that somehow interferes with the improvements of the Main Bar we currently are looking into. So we should be attentive here.  I still think we can proceed with this change for now, as this seems to be an improvement, even if it might not be the final state.

- [x] Line Length: Please adjust [line lengths](https://github.com/ILIAS-eLearning/ILIAS/pull/3177/files#diff-e2754f40b1f9abb18eec5508fc2dc8e13d28e59b23a549d8be81c4de696529adR283) for poor CLI users like me.

- [x] Aria Pressed: I think we are missing a `"false"`-option [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3177/files#diff-a076d60b9a7852496094f6098709d398cc37829b13caa8aba18f3a811126cb41R303), the Bulky-Button is not always engaged at that point, right? Something like [that](https://github.com/ILIAS-eLearning/ILIAS/pull/3177/files#diff-a076d60b9a7852496094f6098709d398cc37829b13caa8aba18f3a811126cb41R104).
- [x] Remove Typehints: They don't give extra information [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3177/files#diff-7a0fbe96bced7f5db9277d260e22f6a61ed61b708f517fa37167bd6de5738369R69) and [there](https://github.com/ILIAS-eLearning/ILIAS/pull/3177/files#diff-7a0fbe96bced7f5db9277d260e22f6a61ed61b708f517fa37167bd6de5738369R87). 

Best regards!

Amstutz:
Hi @klees 

Thx a lot for your feedback! The requested changes should have been taken care of. I perform a merge commit due to a conflict, hope this is fine. 

Special thanks for pointing out the missing false case. Added a unit test to cover this, so I will not happen in future. 

Thx a lot.

PR #3178, opened 2021-02-15
IndividualAssessment 28999: decouple notify checkbox from grading rec…

daniwe4:
…ord in info tab


klees:
Hi @daniwe4,

I decided to go with the change for the label, as the behaviour is as intended.

Best regards!

PR #3179, opened 2021-02-16
(bugfix#30083) fixing comparison of LDAP attribute values

utesche:
mantis:  https://mantis.ilias.de/view.php?id=30083


mjansenDatabay:
Hi @smeyer-ilias  

we stumbled upon this PR in our TB meeting. If you don't have any objections, please merge this PR or provide feedback otherwise.

Best regards!

chfsx:
Hi @smeyer-ilias

we stumbled upon this PR in our TB meeting. If you don't have any objections, please merge this PR or provide feedback otherwise.

Best regards!

smeyer-ilias:
Thanks for your contribution.
I close this report, since it breaks configurations which use special regular expression characters (e.g ?, [, ...) intentionally. 
Thus an attribute value like: "[st][0-9]+" wouldn't match as before.

utesche:
Well, I'm curious for your fix of the given issue now.

Actually the dialog does not inform in any way, that attribut values are treated as regular expressions - and a required 'escaping' of a normal text including some non-alphanumeric character(s) is not an acceptable demand to a common admin, imho.

PR #3186, opened 2021-02-19
BT-29921: Added functionality for max_upload_file to filewizardinput, so

smeyer-ilias:
Link to Mantis-Report:
https://mantis.ilias.de/view.php?id=29921


alex40724 (approved):
Thanks!

PR #3187, opened 2021-02-19
Bt files lucene

smeyer-ilias:
This adds support for the new storage of file objects with the following assumptions:

- file_data.rid is null => no migration is done
- a part of the path to the file can constructed by replacing "_" by "/" or "\"
- version directory is always 1 or higher
- filename is "data"

A switch for the old and new structure is done in Module/File/LuceneObjectDefinition.xml by querying rid against NULL


chfsx:
Thank you very much @smeyer-ilias for your work!

chfsx:
I'll cherry-pick it to trunk asap

PR #3188, opened 2021-02-22
Maps 29411: fix problem with reverse lookup

daniwe4:
- use user entry for reverse lookup server including http or https
- add an alert for reverse lookups with non reachable server
- fix method 'moveToUserMarkerAndOpen'
- update lng var 'maps_custom_geolocation_server_info'


klees:
Picked to trunk.

PR #3196, opened 2021-02-22
30066 Fix: COPage/MediaObjects: Validation of url and title while creating or editing a media object

lukastocker:
HTML tags, quotes or single quote marks will be removed from title, if those exist in title.
If the url is invalid an alert message is thrown. A media object won't be saved until the url is valid. 

https://mantis.ilias.de/view.php?id=30066

### DIFF
--- a/Services/COPage/classes/class.ilPCMediaObjectGUI.php
+++ b/Services/COPage/classes/class.ilPCMediaObjectGUI.php
@@ -575,6 +575,21 @@ class ilPCMediaObjectGUI extends ilPageContentGUI
             $this->insert("edpost", "create_mob", true);
             return;
         }
+
+        if ($_POST['standard_reference'] != "") {
+            try {
+                new \ILIAS\Data\URI($_POST['standard_reference']);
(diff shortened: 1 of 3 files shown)


lukastocker:
Hello @alex40724 ,

I just updated this pull request, because our recent changes led to not being able to upload media images. That should not be a problem anymore, but it occurs to me that this fix could be handled better. Could you look through this PR and tell me if there is a better way to implement our changes? I am not fully aware of the possibilities the MediaObjects (PageEditor) Element brings along as I just recently started working on it's code. I would appreciate your advise and thereafter will gladly fully resolve the mantis ticket 30066.

Best Regards
Luka Stocker

alex40724:
Hi @lukastocker,

thanks for the PR. The best way would be to move this legacy form to the new KS forms, however this is out of scope of this bugfix. So my best solution is to wrap the new URI data element in a legacy input class and use this:

https://github.com/ILIAS-eLearning/ILIAS/commit/b4d1f713718058c85cefec4d8a84ef3f5eb18ecf

PR #3199, opened 2021-02-23
Setup FileSystem: 30122 use data_dir from ilias.ini

daniwe4:
https://mantis.ilias.de/view.php?id=30122


chfsx:
Wow @daniwe4 thanks a lot for this fleet PR! looks good to me! I'll cherry-pick it to trunk as well

PR #3201, opened 2021-02-23
Ignore config.json to make updates more easy

PurHur:
To make the update process more easy i added the config.json to the .gitignore.

The installation process describes that the user creates the file in the docroot of ilias so this is constantly blocking a user from pulling a new version (or patch) if its not committed in a fork.

But there is a flaw in the docs. If the json file is public reachable an attacker would be able to read the DB access data from it. Probably it shouldnt even allowed to use a config.json inside the public directory. You decide.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@
 # Generated Files
 /ilias.ini.php
 /ilias.log
+/config.json
 /db_template_writer.php
 /setup/sql/dbupdate_custom.php
 /install_client.php


klees:
It is not on me to decide on the `.gitignore`, but I'm slightly opposed to this move here because it sets a bad example IMO. One should not have the config.json somewhere in the doc-root. Everyone could just download it then and most possibly would get valuable insights for possible attacks.

Since the `.gitignore` does not belong to any specific person, I assign the TB members to decide. Feel free to ask for a discussion on the JF, though.

Thanks for pointing out the problem anyway. Maybe this is even some security property (not having config.json in the docroot) we would want to enforce via the setup. I will add it as advice to our installation documentation.

PurHur:
I did open this PR to trigger a discussion in your technical board. But since i am not a real ilias developer i let you do the real work :) 
Thanks for acknowledging the problem. This should not be merged due to security issues. It was just bothering me hacking around.

klees:
Yeah, probably should have read the complete comment =) Sry.

kergomard:
Thank you very much for starting this discussion. I would tend to agree that this would be the wrong solution for a very real problem. I support the proposal of @klees to adapt the installation instructions (I proposed a change in the corresponding PR), but also to enforce it in the setup: It would be neat if it would not run if provided a config.json within the webroot.

klees:
@swiniker I agree. I will make the change in the install.md and would kindly ask you to create an according ticket regarding your second suggestion.

kergomard:
@klees : Thx! Issue ticket created [#30236](https://mantis.ilias.de/view.php?id=30236).

mjansenDatabay:
Dear @PurHur 

we discussed this PR in our technical board meeting. We decided to **not** include the `config.json` in our `.gitignore` file. But: Thanks a lot for pointing towards the actual problem. We'll discuss the next steps and actions here https://mantis.ilias.de/view.php?id=30236 . You're invited to participate in these discussions and we're looking forward to hear your input on our ideas as we move forward.

PR #3203, opened 2021-02-23
Proposal/missing commits in UI

chfsx:
Hi @klees 
Sorry again for committing directly in UI-Service und breaking tests with it :/ and thanks for fixing those tests. here are both commits for trunk as well (fixing the UI component and fixing the tests aka your commit)


klees (approved):

Amstutz:
Thx a lot!

PR #3205, opened 2021-02-24
[FIX] 0030122: Can't create missing directory during update command

chfsx:
The PR #3199 almost did the thing, there still some change missing to create directories during the update process. may you have a look at this please? thanks a lot!

https://mantis.ilias.de/view.php?id=30122


klees (approved):
Hi @chfsx,

looks absolutely plausible!

As a general remark: I intentionally designed the API to the Setup to be used without my knowledge as "the Setup Maintainer". This is also why I moved the various usages of the Setup API to the components, so that maintainers can take care of them independently. I do not feel that you need my approval or even inform me when changing or introducing stuff in your components that use that API. There might be some rough edges where we need to communicate and coordinate (or new requirements pop up), but the general idea is that the Setup is just a service like others (Database, RBAC, ...).

Of course I try to be available for questions, reviews, discussion etc. as much as possible, regarding the Setup as well as other topics =)

Best regards!

PR #3206, opened 2021-02-25
UI: Secondary Panel in Sub Panel

tfamula:
This PR adds the possibility to create Sub Panels not only with a Card on the right side, but also with a Secondary Panel.

- [x] There is one technical problem I could not found a solution for. With these changes, it would be possible to use withCard() and withSecondaryPanel() at the same time. This must not be possible. How can we prevent mixing two mutators in the UI Framework?

Related FR: https://docu.ilias.de/goto_docu_wiki_wpage_5999_1357.html

### DIFF
--- a/Modules/Exercise/classes/class.ilExerciseManagementGUI.php
+++ b/Modules/Exercise/classes/class.ilExerciseManagementGUI.php
@@ -659,7 +659,7 @@ class ilExerciseManagementGUI
         }
 
         $main_panel = $this->ui_factory->panel()->sub($a_data['uname'], $this->ui_factory->legacy($a_data['utext']))
-            ->withCard($this->ui_factory->card()->standard($this->lng->txt('text_assignment'))->withSections(array($this->ui_factory->legacy($card_tpl->get()))))->withActions($actions_dropdown);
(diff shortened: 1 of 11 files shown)


Amstutz:
We discussed the Question of Thomas in the UI Clinic Today. We agreed that we unify the two Methods withCards and withSecondaryPanel to withFurtherInformation. 

Other than that we believe that the general approach is fine. Thomas will implement this change, we will review it, and might fast-track it if everything looks fine.

tfamula:
Hi @Amstutz and @klees 
This is ready for final review!

Please note that I had to substitute `withCard()` with `withFurtherInformation()` in some Modules, which already used it. But I think a review of the corresponding maintainers in not needed here, right?

Amstutz:
@tfamula ,

Thx a lot for tackling this. I like the solution. To not slow your progress IMO we can safely fast-track this minor change on the interface and merge before the announce in the next JF. 

Special thx for the updated example and tests. Kudos.

@klees Feel free to hit merge as soon as reviewed.

Amstutz (approved):

klees:
Thx @tfamula!

PR #3207, opened 2021-02-25
implement __toString() on UUID

Hadrian-II:
implement __toString() on UUID as this method is needed by php for some methods like array_diff

### DIFF
--- a/src/Data/UUID/RamseyUuidWrapper.php
+++ b/src/Data/UUID/RamseyUuidWrapper.php
@@ -61,4 +61,12 @@ class RamseyUuidWrapper implements Uuid
     {
         return $this->wrapped_uuid->toString();
     }
+
+    /**
+     * @return string
+     */
+    public function __toString() : string
+    {
+        return $this->toString();
+    }
 }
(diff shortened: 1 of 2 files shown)


klees (changes requested):
Hi @Hadrian-II,

LGTM in general, thanks a lot. I have one change request, though: Since virtually all code should depend on the interface `Uuid` rather then the implementation you are modifying, please add that method on the interface too.

Best regards!

Hadrian-II:
i added the method to the interface, im not sure if adding magic methods to interfaces feels right

klees:
Hi @Hadrian-II,

sounds interesting =) Why's that?

Best regards!

mjansenDatabay:
@Hadrian-II / @klees `__toString` can be ensured by the `Stringable` interface in PHP 8, so it is not sooooo magic anymore. The `Uuid` interface could then be slightly changed by using `extends Stringable` and the added method could be removed, if we require `PHP >= 8.0`.

klees (approved):

PR #3209, opened 2021-02-26
Setup UICore: 30234 use temp container for read structure

daniwe4:
https://mantis.ilias.de/view.php?id=30234

### DIFF
--- a/Services/UICore/classes/Setup/class.ilCtrlStructureStoredObjective.php
+++ b/Services/UICore/classes/Setup/class.ilCtrlStructureStoredObjective.php
@@ -1,6 +1,7 @@
 <?php declare(strict_types=1);
 
 use ILIAS\Setup;
+use ILIAS\DI;
 
 class ilCtrlStructureStoredObjective implements Setup\Objective
 {
@@ -71,9 +72,20 @@ class ilCtrlStructureStoredObjective implements Setup\Objective
             define("ILIAS_ABSOLUTE_PATH", dirname(__FILE__, 5));
         }
 
+        // ATTENTION: This is a total abomination. It only exists to allow various
+        // sub components of the various readers to run. This is a memento to the
+        // fact, that dependency injection is something we want. Currently, every
+        // component could just service locate the whole world via the global $DIC.
+        $DIC = $GLOBALS["DIC"];
+        $GLOBALS["DIC"] = new DI\Container();
+        $GLOBALS["DIC"]["ilDB"] = $db;
+
         $reader = $this->ctrl_reader->withDB($db);
         $reader->executed = false;
         $reader->readStructure(true);
+
+        $GLOBALS["DIC"] = $DIC;
+
         return $environment;
     }
 


klees (approved):

klees:
Hi @alex40724,

as UICore is a common good and the change was made by us (CaT), I assign this to you. Hope that this is ok.

Best regards!

alex40724:
Hi @daniwe4,

thanks for the PR. I do not think that this is the proper place for this fix.

@klees: What is the role of the DIC object in the setup? I think you introduced this 2016 in the setup.

UICore should not be responsible to init the setup DIC, right? And if there is nothing like this anymore, the Cache/AR component could work around this instead. Or retrieve the DB from ilCtrlStructureStoredObjective.

klees:
Hi @alex40724,

for classes from the setup we use proper dependency injection. There is an instance of `Pimple\Container` initialized in `setup\cli.php` but it is really only a dependency injection container there, not a service locator, like for the most of the rest of ILIAS. The ILIAS service locator DIC was never introduced in the setup. From an architectural perspective I consider dependency injection to be strictly better then the service location pattern. But that even isn't the main concern regarding the setup. We simply cannot have the ILIAS DIC there, because most of the stuff that is inside the container won't even exist in the setup. Some of the stuff just gets created there, some of the stuff will just never belong into the setup.

However, this does not make other classes, like `ilGlobalCache` follow dependency injection instantly. I did some work to make some classes (like `ilCtrlStructureReader`) compatibel with dependency injection, but solving that problem for every class that gets touched in the setup just was too much. In these cases we just initialize the service locator DIC with the services that the called classes locate later on in the respective Objective. Other instances of this patterns can be found e.g. in `ilDatabaseUpdatedObjective`.

This pattern is ugly indeed and also not the way it should be, this is why I added that comment. I still think it is a good way forward for two reasons: For one, the pattern exists on the very boundary between "my" setup code and "other peoples" other code. This allowed me to work with the setup code at my speed and discretion, while other code can stay as it is for the time being. So this is just a form of decoupling between components. Second, the pattern very explicitely clarifies which component actually requires which other services (at least during setup) and moves the actual initialisation close to the component. This can be a hook for further clean up work in the component themself and also, if that work is not done, will make the initialisation of the service locator the work of the component that actually requires it (and not the setup, as it was before). So again a form of decoupling.

This is especially true for the `ilCtrlStructureReader` and its dependency on `ilGlobalCache` (and the transitive dependency on `ilDB`). I think that we will be able to actually just remove that dependency by moving `ilCtrl` to an artifact based implementation. This will be a lot simpler, require no database or caching and still will be cached automatically by PHP (if configured appropriately). I plan on looking into this for ILIAS 8, but due to tight schedules this is not a promise. Still, this is why I would not bother to much with that ugliness in atm.

I hope this makes sense. Still, this is not the one way to solve the problem at hand, so feel free to outline others, so we can weigh pros and cons.

Best regards!

alex40724:
Yes, this makes sense.

I just had a look at ilCachedCtrl/ilGlobalCache.

@chfsx : There is little chance to pass db as a dependency to this service right now, right? There are lots of static calls involved.

@klees, @chfsx: If this is not an option you may simple merge this.

klees:
Picked to trunk.

PR #3210, opened 2021-02-26
Fix wrong flex usage and html structure in panel headers

iszmais:
https://mantis.ilias.de/view.php?id=30237

### DIFF
--- a/Modules/BookingManager/BookingProcess/templates/default/tpl.booking_reservation_post.html
+++ b/Modules/BookingManager/BookingProcess/templates/default/tpl.booking_reservation_post.html
@@ -1,5 +1,7 @@
 <div class="panel panel-primary">
-	<div class="panel-heading ilHeader clearfix"><h3 class="ilHeader">{TITLE}</h3></div>
+	<div class="panel-heading ilHeader">
+		<h3>{TITLE}</h3>
+	</div>
 	<div class="panel-body">
 		<p>
 		{POST_TEXT}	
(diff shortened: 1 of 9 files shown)


Amstutz:
Hi @iszmais 

Thx you very very much for the fix! Highly appreciated. Placement of the dropdowns has been an issue hunting panels for some time now on various screen sizes. Your solution is certainly a step in the right direction. I tested it, and observed that, some issues on smaller screens seem to remain, however, I believe to have a solution ready for those: 

image

In your ticket you mention: 

_Dies resultiert aus einer falschen Implementierung der Flexboxen bzw. wurden nicht alle Artefakte der alten Row/Floating Struktur entfernt. Dies ist auch bei vielen anderen Inhaltselementen der Fall, weswegen ich ein generelles Refactoring der HTML/CSS Struktur empfehlen würde._

I completely agree! Some elements like panels and forms have been implemented in UI Components to have a as little a diff as possible to the original ones in order to ILIAS to look consistent. However, this was 5.4, in the meanwhile we have made big steps towards more coverage of ILIAS in terms of UI Components, so we have also more liberties in the was of how to implement the DOM. Note, there is always always more work than we can chew. So if you have some concept for some component of what could be improved. I would really look forward to discuss this or look into some PR. Especially during corona, I believe it always nice to discuss some concept or idea through skype and a cup of coffee ;-). Just ping me if you have proposals to share.

To move this forward, I will squash merge the requested change to 6 and also 7 and trunk, since I believe them to be good (special thx for not forgetting about the tests ;-) ). I will shortly after push the fix for small screens as well. 

Thx, and looking forward to further improvements.

Amstutz (approved):

PR #3212, opened 2021-02-27
Less, Cleanup work from the CSS Squad, see: #30120

Amstutz:
hi @nhaagen and @klees 

We perform some less/css cleanup work atm. In this process we observed, that some Less logic seems not to be in use anymore. See also: https://mantis.ilias.de/view.php?id=30120

- [ ] Can you look over the change and merge to trunk and 7 if really not used anymore?

Note: If possible also try to get a rid of the !important statements in the less. Further, note that we will remove the custom color assignment in here in the process of unifying the varying shades of grey used in ILIAS. This will be done in an upcoming PR of the CSS_Squad. If possible for 7, if not hopefully sometime soon ;)

Thx.

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -18551,22 +18551,6 @@ span.ilProfileBadge .modal .img-responsive {
   cursor: default !important;
   text-decoration: none !important;
 }
-.ilLSOKioskMode #mainspacekeeper {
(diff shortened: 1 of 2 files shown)


klees:
Thx a bunch!

klees:
Picked to release_7 as well.

PR #3213, opened 2021-03-01
BT-29998: Fixed critical error, that occured when trying to import users while ignoring local role

M-Barz:
Link to Mantis-Report:
https://mantis.ilias.de/view.php?id=29998


smeyer-ilias (approved):
Thanks, please also merge to affected  branches (r6?).

PR #3214, opened 2021-03-03
Setup FileSystem: 30122 precondition respects update and install

daniwe4:
Hello @chfsx,

after the first PR for [#30122](https://mantis.ilias.de/view.php?id=30122) we now noticed that a fresh installation would run into an error 'Path is empty'. This depends on a missing environment resource at this point, that is used in a precondition of ilFileSystemComponentDataDirectoryCreatedObjective. So we have to distinguish between a fresh ILIAS installation and an
already existing ILIAS installation at this precondition.

In case of a fresh installation we have to ensure all file system directories are created. We use the config for file system from the
environment.

In case of an already installed ILIAS we can use the ini files.

If you have questions or need to discuss anything, feel free to contact us.

Best regards.


chfsx:
Thanks a lot!

PR #3218, opened 2021-03-03
Update install.md

lorenzgith:
- Update user php7.4 instead of php7.1 otherwise composer returns an error.
- Remove ubuntu 14&16 because they are deprecated


klees:
Hi @lorenzgith,

thanks for looking into the installation instructions. This certainly is a piece of text that gets to little attention.

Still, I will close this PR for the following reasons:

* [Ubuntu 14.04 and 16.04 are still under maintenance](https://ubuntu.com/about/release-cycle)
* We did not test ILIAS 6 with PHP 7.4 and thus (IMO) should not recommend it in the installation instructions.
* The PR removes the `-Indexes` option from the Apache config. As far as I understand, this would allow to create indexes from the base directory, which is not something we want.

Please note, that I'm not closing this PR in any specific capacity. Currently no one really is responsible for the install.md and it seems as someone needs to move this PR forward. This is what I try to do here. If anyone disagrees with my approach, feel free to reopen the PR and maybe bring it to discussion on the Jour Fixe.

If you are interested in the install.md, please also have a look into [PR 3158]( https://github.com/ILIAS-eLearning/ILIAS/pull/3158).

Best regards!

lorenzgith:
Hey @klees 
Thanks a lot for the feedback.
I understand your attached points.

I have opened a new pull request (#3220). In which the changes are smaller. So that the installation can be done without errors.

PR #3219, opened 2021-03-04
IndividiualAssessment 30249: fix error with mandatory fields

daniwe4:
- save action will send ids now also if mandatory fields are empty
- date field use correct format now
- select a file and then return to overview and hit again edit will not
  return an error anymore

https://mantis.ilias.de/view.php?id=30249


klees:
I picked this to trunk.

PR #3220, opened 2021-03-05
Update install.md

lorenzgith:
- Apache configuration adjusted, because the service can not be started otherwise. 
- Moved from php7.1 to php7.3 because ILIAS 6 no longer supports php7.1.


klees:
Thanks a lot!

PR #3221, opened 2021-03-05
Setup File: 30235 exit migration successfuly if source dir does't exists

daniwe4:
https://mantis.ilias.de/view.php?id=30235


chfsx:
Thanks a lot, I cherry picked to trunks as well...

PR #3222, opened 2021-03-07
UI Components, Inspection Cleanup

Amstutz:
Hi

Again a rather large PR. This is the result of an initiative of @mjansenDatabay who contacted me with an first draft of a possible Insspetion Template for PHP Storms that could be used for the ease of the migration to PHP 8. To get an idea of what this template would mean concerning necessary changes in the code and amount of work and problems we will run into, I tested it with the UI Components (src/UI). Here the results, could be interesting for others planing the migration and usage of some Inspection in PHP storms:

Issues after running the template in the first draft version:
image

After around 6h (let's say 1PD) of work I was down to 5 warnings, that are somewhat hard to get rid of (deprecated methods, still in use etc.). Some thoughts:
- Inspections get more valuable if typing is strict. We should aim towards strict typing. The typing in the UI Framework is moderately strict, however quite strict compared with most other code in ILIAS.
- The inspection will not uncover all issues for migrations. E.g. the undefined array index issue will mostly go unnoticed be the inspections. Those need to be handled by unit tests or manual tests.
- Unit tests proved again to be of major value while performing those changes. Some triggered errors in the code, that went unnoticed by the inspection, but have been catched by the unit tests immediatly.
- It is hard to say, if other code will prove to be more or less works to handle the inspections. Inspections will detect more if the typing ist stricter, however there are less issues with things like undefined variables to begin with.
- The UI Components without tests (src/UI) contain 47284 loc (php), the folders. Modules containts 645746 loc, Services 835062 loc. Meaning, src/UI only contains 2-3% of the ILIAS Code base. IF the amount of work would scale linearly over the whole project (that is a huge if) we could estimate roughly the work needed for the whole project. However, note we need more samples like this to make a good estimate. Since the state of the code base varies heavily from component to component.

@klees I believe almost all of the changes are slight improvements, lots of cleanup work in there. I believe we should aim in merging this case study. No logic should have been changed. We should keep on working on further pushing strict typing. It is definitely worth it and makes inspections more valuable.


src/UI/Component/Symbol/Glyph/Glyph.php
```diff
@@ -13,51 +13,51 @@
...
-    const DISCLOSURE = "disclosure";
-    const LANGUAGE = "language";
-    const LOGIN = "login";
-    const LOGOUT = "logout";
-    const BULLETLIST = "bulletlist";
-    const NUMBEREDLIST = "numberedlist";
-    const LISTINDENT = "listindent";
-    const LISTOUTDENT = "listoutdent";
-    const FILTER = "filter";
+    public const  SETTINGS = "settings";
```
klees: Too much of spaces here.

src/UI/Implementation/Component/Button/Close.php
```diff
@@ -21,7 +21,6 @@ class Close implements C\Button\Close
      */
     public function withOnClick(Signal $signal)
     {
-        $this->action = null;
```
klees, line 24: This looks suspicious. For Buttons we have an "either action (= URL) or signal"-behaviour, as far as I know. This would leave us with a button that has a signal and an action, it seems. Or is this solved somewhere else?

src/UI/Implementation/Component/Input/Field/Checkbox.php
```diff
@@ -53,7 +53,11 @@ public function withValue($value)
             );
         }
 
-        return parent::withValue($value);
+        /**
+         * @var $clone Checkbox
+         */
+        $clone = parent::withValue($value);
+        return $clone;
```
klees, line 60: Why the extra assignment to a variable we do not use anyway?

src/UI/Implementation/Component/Input/Field/Duration.php
```diff
@@ -7,8 +7,7 @@
 use ILIAS\UI\Component as C;
 use ILIAS\Data\Factory as DataFactory;
 use ILIAS\Data\DateFormat as DateFormat;
-use ILIAS\Refinery\Transformation\Factory as TransformationFactory;
-use ILIAS\Refinery\Validation\Factory as ValidationFactory;
+use ILIAS\Refinery as Refinary;
```
klees:
```suggestion
use ILIAS\Refinery as Refinery;
```

The same problem repeats down below.

src/UI/Implementation/Crawler/Exception/CrawlerException.php
```diff
@@ -13,37 +13,37 @@
...
-    const FOLDER_DELETION_FAILED = 3004;
-    const FILE_DELETION_FAILED = 3005;
-
-    const INVALID_INDEX = 4000;
-    const MISSING_INDEX = 4001;
-
-    const CRAWL_MAX_NESTING_REACHED = 5000;
+    public const UNKNOWN_EXCEPTION = -1;
+
+    public const  ARRAY_EXPECTED = 1000;
```
klees: Hmm, theses spaces again...

src/UI/examples/Panel/Secondary/Legacy/with_viewcontrollers.php
```diff
@@ -27,7 +27,7 @@ function with_viewcontrollers()
     $url = $DIC->http()->request()->getRequestTarget();
 
     $parameter_name = 'page';
-    $current_page = (int) @$_GET[$parameter_name];
+    $current_page = (int) (array_key_exists($parameter_name, $_GET) ? $_GET[$parameter_name] : 0);
```
klees, line 30: Would we want to use this construct [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3222/files#diff-284fa4639eb987095447333cc1447d1326c33b9bb0c963017bc78006975c0a05R39) and [there](https://github.com/ILIAS-eLearning/ILIAS/pull/3222/files#diff-bde9e13b5cad94ebe56ee2ac3bc34ce532b0a831ee455d228a472cb01061d442R27)? Or could we just use `$_GET[$parameter_name] ?? 0`? This would then also need to be changed [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3222/files#diff-063e35698485577d1146c3bf20dddb1bd6ccbb669ab6118b7c435accf5f66942R51), [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3222/files#diff-a56c7e4d9ce3f7427a3a5533c02d563ac446d4d55832da3f727e581c45418d4cR30), [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3222/files#diff-671784fa572ee3182b855bcd815191f10e2f69bf586819c8905a6692e2c14699R13), [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3222/files#diff-26d1e677c6db219f15bda44f9dd4e51f9c9286757f75d896c2c55e87178ee468R13) and [there](https://github.com/ILIAS-eLearning/ILIAS/pull/3222/files#diff-5b772ef5bd99a0d70d86b6a949a6d381a4ec0597bd3ca59738579d7446397c99R14).

klees (commented):
Hi @Amstutz,

this LGTM in general, thanks a lot for looking into this. I left some requests for changes, suggestions and questions inline.

Feel free to apply (and answer) them at your own discretion and ping me to merge this afterwards. This is a huge improvement already.

Best regards!

src/UI/Implementation/Component/Button/Close.php
Amstutz: Please give me a failing unit test to prove your point ;-). Ill fix it asap, then.

src/UI/Component/Symbol/Glyph/Glyph.php
Amstutz: sry, hmm, capitain hook did not catch that. Ill fix.

src/UI/Implementation/Component/Input/Field/Checkbox.php
Amstutz: to tell the inspection (and the interested reader), that we passe back a checkbox in this context and that this can safely returned as checkbox (otherwise the injector and maybe also the reader will panic, since parent::withValue might return other things).

klees: Makes sense.

src/UI/examples/Panel/Secondary/Legacy/with_viewcontrollers.php
Amstutz: I used the array_key_exists pattern, with the other one, we would IMO still run into the issue of not defined array key indices. I changed the linked place as well.

src/UI/Implementation/Component/Button/Close.php
Amstutz: I double checked. Action is a standalone Button, does not inherit anything from buttons. withOnclick is only implemented there and only accepts a signal. Action has no use at all. In this case.

Amstutz:
Change requests should have been taken care of. Thx a lot. Note: Captain hook just save my ass :-)

src/UI/examples/Panel/Secondary/Legacy/with_viewcontrollers.php
klees: @Amstutz and me almost agreed to move to the new `??` sometimes soon =)

klees:
I have cherry-picked this to release_7.

PR #3224, opened 2021-03-09
Development: Add PHP inspection profile for PhpStorm

mjansenDatabay:
This PR suggest a PHP inspection profile for the PhpStorm IDE ILIAS developers COULD use.

Some reasons why I chose particular rules:

* PHP | Attributes
  * Reason: We don't use attributes, yet. But if we want to use this profile on a PHP >= 8 language level for ILIAS >= 9 in the future, this inspections could be useful.
* PHP | Code style | Multiple classes declarations in one file
  * Reason: We should enforce PSR-4 compatibility, otherwise this could lead to issues with `Composer 2`.
* PHP | Code style
  * PHP | Code style | Old style constructor
    * Reason: The usage of PHP 4 constructors MUST be reported because they will not be called anymore with PHP >= 8.
  * PHP | Code style | Short open tag usage
    * Reason: Short tags are not supported aynmore with PHP >= 8, see: https://wiki.php.net/rfc/deprecate_php_short_tags
  * PHP | Code style | Usage of a variable variable
    * Reason: `$a = 'Hello'; $$a = 'World';`  Not a real PHP 8 issue, but this causes maintainability issues.
* PHP | Control flow | Typed Property might be uninitialized
  * Reason: Since we support PHP 7.4 and thus could use `Typed Properties`, these issues should be reported.
* PHP | General (Reason: Most of these issues are serious problems):
  * PHP | General | Class can't implement Traversable directly
  * PHP | General | Class hierarchy checks
  * PHP | General | Curly brace access syntax usage
  * PHP | General | Deprecated cast
  * PHP | General | Deprecated implode/join usage
  * PHP | General | Element is not available in configured PHP version
  * PHP | General | Ignored class alias declaration
  * PHP | General | Incorrect magic method signature
  * PHP | General | Invalid magic method modifiers
  * PHP | General | Language level
  * PHP | General | Method declaration of super class is incompatible with implemented interface
  * PHP | General | Named argument may be unresolved
  * PHP | General | Nested ternary operator usage
  * PHP | General | Parameter's name changed during inheritance
  * PHP | General | Promoted property usage
  * PHP | General | Property can be promoted
* PHP | PHPUnit
  *  Reason: `PHPUnit` is an important tool for us, so we should report all problems, even issues of type `Deprecated`.
* PHP | Probable bugs (Reason: Most of these issues are serious problems):
  * PHP | Probable bugs | Concatenation with arithmetic usage
  * PHP | Probable bugs | Constant reassignment
  * PHP | Probable bugs | Division by zero
  * PHP | Probable bugs | Duplicate arm in 'match' expression
  * PHP | Probable bugs | Duplicate array keys
  * PHP | Probable bugs | Duplicate branch in switch statement
  * PHP | Probable bugs | Duplicate case in switch statement
  * PHP | Probable bugs | Duplicate condition
  * PHP | Probable bugs | Expression result unused
  * PHP | Probable bugs | Foreach array is used as value
  * PHP | Probable bugs | Format function parameters mismatch
  * PHP | Probable bugs | Goto into loop statement
  * PHP | Probable bugs | Invalid type of unpacked argument
  * PHP | Probable bugs | Method __toString implementation
  * PHP | Probable bugs | Method __toString return type
  * PHP | Probable bugs | Nested vs outer 'foreach' variables conflict
  * PHP | Probable bugs | Optional before required parameter
  * PHP | Probable bugs | Pass parameter by reference
  * PHP | Probable bugs | Unused 'match' condition
  * PHP | Probable bugs | Void function result used
  * PHP | Probable bugs | Wrong string concatenation
* PHP |Quality tools
  * PHP | Quality tools | PHP CS Fixer validation (with the ILIAS configuration file)
    * Reason: Because we decided to use `PSR-2 + X`.
* PHP | Type compatibility (Reason: Type issues should be reported. A more type safe system means less bugs)
  * PHP | Type compatibility | Illegal array key type
  * PHP | Type compatibility | Illegal string offset
  * PHP | Type compatibility | Incompatible return type
  * PHP | Type compatibility | Invalid argument supplied for foreach()
  * PHP | Type compatibility | Parameter type
  * PHP | Type compatibility | PHP 8 TypeError on arithmetic operations
  * PHP | Type compatibility | Strict type checking rules violation
  * PHP | Type compatibility | Type declaration is redundant and could be simplified
  * PHP | Type compatibility | Type mismatch in property assignment
* PHP | Undefined (Reason: These inspection could result in false positives, but serious issues could be found by those inspections)
  * PHP | Undefined | Undefined callback
  * PHP | Undefined | Undefined class
  * PHP | Undefined | Undefined class constant
  * PHP | Undefined | Undefined constant
  * PHP | Undefined | Undefined function
  * PHP | Undefined | Undefined goto label
  * PHP | Undefined | Undefined method
  * PHP | Undefined | Undefined namespace
  * PHP | Undefined | Undefined property
  * PHP | Undefined | Undefined variable

### DIFF
--- a/docs/development/contributing.md
+++ b/docs/development/contributing.md
@@ -69,6 +69,9 @@ code please make sure:
 * that your commit follows the [ILIAS coding
   guidelines](http://www.ilias.de/docu/goto_docu_pg_202_42.html) - this is a
   bare minimun of style we want to maintain for new code
+* you don't introduce new code violations which could have been easyly found by
+  importing and running our
+  [PhpStorm PHP Inspection Profile](./inspection-configs/php-storm-php-inspections.xml)
 * that your are approachable for questions of reviewers
 
 If your PR contains a bugfix please reference the number of the mantis ticket
--- a/docs/development/contributing.md
+++ b/docs/development/contributing.md
@@ -69,7 +69,7 @@ code please make sure:
 * that your commit follows the [ILIAS coding
   guidelines](http://www.ilias.de/docu/goto_docu_pg_202_42.html) - this is a
   bare minimun of style we want to maintain for new code
-* you don't introduce new code violations which could have been easyly found by
+* you don't introduce new code violations which could have been easily found by
   importing and running our
   [PhpStorm PHP Inspection Profile](./inspection-configs/php-storm-php-inspections.xml)
 * that your are approachable for questions of reviewers
--- a/docs/development/coding-style.md
+++ b/docs/development/coding-style.md
@@ -58,6 +58,11 @@ When working with the `PhpStorm` IDE developers can import the
 Furthermore multiple [Git Hooks](./git-hooks.md#code-style-hooks) are provided
 to check or fix the code style of changed files in a Git commit.
 
+Another possibility to apply the code style checks is to import and run
+the [PhpStorm PHP Code Inspection Profile](./inspection-configs/php-storm-php-inspections.xml).
+This does not only check the ILIAS Coding Style but applies other inspection
+rules to the PHP code base (or parts of it) as well. 
+
 Developers can additionally use the [PHP Coding Standards Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer)
 to check or fix one or multiple files.
 
--- a/docs/development/inspection-configs/php-storm-php-inspections.xml
+++ b/docs/development/inspection-configs/php-storm-php-inspections.xml
@@ -1,0 +1,525 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="ILIAS PhpStorm PHP Inspections" />
+    <inspection_tool class="AngularAmbiguousComponentTag" enabled="false" level="ERROR" enabled_by_default="false" />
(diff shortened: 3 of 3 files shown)


Amstutz:
Hi @mjansenDatabay 

Thx 1000! for this great work. It already proved very useful for me while working on the UI Components. Further, thx for the explanation. I would heavily recommend all devs working with PHP Storms to work with the recommended template from @mjansenDatabay thx A LOT.

@Amstutz

matthiaskunkel:
Jour Fixe, 17 MAY 2021 : We highly appreciate this improvement and accept all changes to our docs.

docs/development/contributing.md
```diff
@@ -69,6 +69,9 @@ code please make sure:
 * that your commit follows the [ILIAS coding
   guidelines](http://www.ilias.de/docu/goto_docu_pg_202_42.html) - this is a
   bare minimun of style we want to maintain for new code
+* you don't introduce new code violations which could have been easyly found by
```
nicoroeser: Nitpick: “easyly” → “easily”.

mjansenDatabay: Thx @nicoroeser

PR #3227, opened 2021-03-12
Setup Database: 30367 read correctly port from client_ini

daniwe4:
https://mantis.ilias.de/view.php?id=30367


chfsx:
Thanks! I cherry picked to trunk as well

PR #3229, opened 2021-03-12
0030107: No access to reservations-tab after chosing certain columns

fneumann:
This is a fix for https://mantis.ilias.de/view.php?id=30107 (sorry for the mind the misleading commit message)

Course/group defined user fields are removed from user query and added later with a query of course user data, similar to the way it is done in the list of course members.


alex40724:
Thanks for the PR. Bug is fixed, code is being cherry-picked.

PR #2344, opened 2019-11-21
Feature/6 0/refinery in http service

chfsx:
Dear Technical Board

This PR proposes using the refinery in the HTTP service. It includes the following changes, some of which are still under discussion:

- Separation of the global HTTp service : A distinction is made between raw() and wrapper(). In the future, only the wrapper should be used, but currently many services still access the entire `RequestInterface`. 
- The wrappers offer easy ways to access $_GET, $_POST and $_COOKIE using `transformations` of the refinery (see https://github.com/ILIAS-eLearning/ILIAS/pull/2344/files#diff-eda6070f33899930ba976bfa6b85334cR21 )

This is the part we should discuss:
- The PR also proposes drop-in replacements for the superglobal $_GET, $_POST, $_REQUEST and $_COOKIE. These classes enclose the superglobals and use the refinery internally when accessing content (see https://github.com/ILIAS-eLearning/ILIAS/pull/2344/files#diff-2f6f2df4c9612331e495a008236e315bR12 ). This implementation is certainly worth discussing, which also needs some adjustments in ILIAS because they expect native arrays (e.g. https://github.com/ILIAS-eLearning/ILIAS/pull/2344/files#diff-411e64d40db5eeb3d6a82486fdc5e607R526 ).
- In some places the superglobals in ILIAS were simply overwritten, this must not be from my point of view, I commented out these places (otherwise also the DropIn Replacements do not work) (e.g. https://github.com/ILIAS-eLearning/ILIAS/pull/2344/files#diff-c96313dda4795745e3af54ee394b156eL181 ).

Please also contact me with questions, I will be happy to give you a walk-through if you wish.

PR #3234, opened 2021-03-15
Feature/trunk/refinery http

chfsx:
This is the followup PR for #2344, now based on trunk

### DIFF
--- a/Modules/Chatroom/classes/class.ilChatroomAuthInputGUI.php
+++ b/Modules/Chatroom/classes/class.ilChatroomAuthInputGUI.php
@@ -12,7 +12,7 @@ class ilChatroomAuthInputGUI extends ilSubEnabledFormPropertyGUI
     private const NAME_AUTH_PROP_1 = 'key';
     private const NAME_AUTH_PROP_2 = 'secret';
 
-    private \ILIAS\DI\HTTPServices $http;
+    private \ILIAS\HTTP\Services $http;
     /** @var string[]  */
     protected array $ctrl_path = [];
     protected int $size = 10;
@@ -22,7 +22,7 @@ class ilChatroomAuthInputGUI extends ilSubEnabledFormPropertyGUI
     ];
     protected bool $isReadOnly = false;
 
-    public function __construct(string $title, string $httpPostVariableName, \ILIAS\DI\HTTPServices $http)
+    public function __construct(string $title, string $httpPostVariableName, \ILIAS\HTTP\Services $http)
     {
         parent::__construct($title, $httpPostVariableName);
         $this->http = $http;
--- a/Modules/Chatroom/classes/class.ilChatroomFormFactory.php
+++ b/Modules/Chatroom/classes/class.ilChatroomFormFactory.php
@@ -11,7 +11,7 @@ class ilChatroomFormFactory
 {
     protected ilLanguage $lng;
     protected ilObjUser $user;
-    protected \ILIAS\DI\HTTPServices $http;
+    protected \ILIAS\HTTP\Services $http;
 
     /**
      * Constructor
--- a/Modules/File/classes/Versions/class.ilFileVersionsGUI.php
+++ b/Modules/File/classes/Versions/class.ilFileVersionsGUI.php
@@ -1,6 +1,6 @@
 <?php
 
-use ILIAS\DI\HTTPServices;
+use ILIAS\HTTP\Services;
 use ILIAS\Filesystem\Exception\FileNotFoundException;
 
 /**
@@ -43,7 +43,7 @@ class ilFileVersionsGUI
      */
     private $lng;
     /**
-     * @var HTTPServices
+     * @var Services
      */
     private $http;
     /**
--- a/Modules/Forum/classes/class.ilObjForumGUI.php
+++ b/Modules/Forum/classes/class.ilObjForumGUI.php
@@ -60,7 +60,7 @@ class ilObjForumGUI extends \ilObjectGUI implements \ilDesktopItemHandling
 
     /** @var \Psr\Http\Message\ServerRequestInterface */
     private $httpRequest;
-    /** @var \ILIAS\HTTP\GlobalHttpState */
+    /** @var \ILIAS\HTTP\Services */
     private $http;
 
     /** @var Factory */
--- a/Modules/OrgUnit/classes/ARHelper/DIC.php
+++ b/Modules/OrgUnit/classes/ARHelper/DIC.php
@@ -85,7 +85,7 @@ trait DIC
 
 
     /**
-     * @return \ILIAS\DI\HTTPServices
+     * @return ILIAS\
      */
     protected function http()
     {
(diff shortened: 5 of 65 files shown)


chfsx:
Copied from the old PR:

- [x] ilPropertyFormGUI: Can we really remove [this](https://github.com/ILIAS-eLearning/ILIAS/pull/2344/files#diff-411e64d40db5eeb3d6a82486fdc5e607L503) chunk of code? Why don't we need it anymore?
- [x] ReflectionClass: What's the purpose of [`HTTP\Services::getWithMultipleArguments`](https://github.com/ILIAS-eLearning/ILIAS/pull/2344/files#diff-e50f175337eeff55968c5b4b2650f8c8R153)? Using the `ReflectionClass` is (deemed to be) a performance impact. Why do we take it here? We know the classes that we want to instantiate.
- [x] DirectArrayAccessDropInReplacement: What is the purpose or added benefit of `DirectArrayAccessDropInReplacement`? Using the `Refinery` internally won't do anything to make this more secure, at least regarding the current implementation. Transforming the contents to string basically is a non-operation from the perspective of the result, the contents in the request super globals are strings anyway.

- [x] DI: Please do not remove the dependency injection of `GlobalHttpState`. This will make testing and reuse of the class harder. What's the benefit of creating dependencies inline?

- [x] Wrapper/Raw: In current proposal you suggest to add a `wrapper` method to the `GlobalHttpState`, which will give access to the methods that use the `Refinery` for data retrieval. These should become the new normal for almost all of the ILIAS code. The other methods, like `request`, on the `GlobalHttpState` on the other hand provide access to the raw request objects, which mostly should not be to used anymore. Thus we suggest to change the `GlobalHttpState` so that `request` and `cookieJar` return the objects that use the `Refinery` for access and only make the raw objects accessible behind `raw`. Maybe we could even give a name that hints at the unsafeness even more, like  `getUnsafeRawObjects`. An update of the existing usage should be straight forward.
- [x] HTTPService: With the changes here, `HTTPServices` seems to have become an odd name. Please rename to `RawHTTPObjects`.
- [x] RequestWrapper: This should be pretty generic for all kinds of inputs. Rename it to something like `KeyValueInput` and move it to the top-level of the Refinery. The same goes for `ArrayBasedRequestWrapper`, which could be renamed to `ArrayInput` and also be moved to the `Refinery`.
- [ ] Dicto: Add some Dicto rules that flag usages of the request super globals and `getUnsafeRawObject`.
- [ ] Tests: Please add automated tests that at least cover the core features of the HTTP library.

chfsx:
also copied from the old PR: 

Thanks for the review and no problem that it takes a little longer (it took me a long time now, too :-) )

with pleasure I will go into your points:

>  ilPropertyFormGUI: Can we really remove this chunk of code? Why don't we need it anymore?

this was to check if post_max_size exceeded, but this leads to an error anyway.

> ReflectionClass: What's the purpose of HTTP\Services::getWithMultipleArguments? Using the ReflectionClass is (deemed to be) a performance impact. Why do we take it here? We know the classes that we want to instantiate.

This was to handle the internal instances a bit easier, but I can remove this as well

> DirectArrayAccessDropInReplacement: What is the purpose or added benefit of DirectArrayAccessDropInReplacement? Using the Refinery internally won't do anything to make this more secure, at least regarding the current implementation. Transforming the contents to string basically is a non-operation from the perspective of the result, the contents in the request super globals are strings anyway.

Its correct that this doesn't bring a direct improvement of security. But it's a way to handle code which accesses $_GET directly in the future. With this implementation all accesses to $_GET go through this implementation, it will be easy to prevent from doing this in the future (which should be done IMO). I highly recommend to use this.

> DI: Please do not remove the dependency injection of GlobalHttpState. This will make testing and reuse of the class harder. What's the benefit of creating dependencies inline?

chfsx:
and now for the new PR:

> Wrapper/Raw: In current proposal you suggest to add a wrapper method to the GlobalHttpState, which will give access to the methods that use the Refinery for data retrieval. These should become the new normal for almost all of the ILIAS code. The other methods, like request, on the GlobalHttpState on the other hand provide access to the raw request objects, which mostly should not be to used anymore. Thus we suggest to change the GlobalHttpState so that request and cookieJar return the objects that use the Refinery for access and only make the raw objects accessible behind raw. Maybe we could even give a name that hints at the unsafeness even more, like getUnsafeRawObjects. An update of the existing usage should be straight forward.

I agree that this should be the new normal. however, we probably have to work with deprecated (and perhaps dicto), otherwise many places in ILIAS would already have to be adapted. of course, this could also be done, but it is difficult, because plugins, for example, could already rely on it.

> HTTPService: With the changes here, HTTPServices seems to have become an odd name. Please rename to RawHTTPObjects.

absolutely, I'll be happy to adjust that.

> RequestWrapper: This should be pretty generic for all kinds of inputs. Rename it to something like KeyValueInput and move it to the top-level of the Refinery. The same goes for ArrayBasedRequestWrapper, which could be renamed to ArrayInput and also be moved to the Refinery.

I'm not adapting that either. Unfortunately, I lost some of the groundwork by re-implementing the PR.

> Dicto: Add some Dicto rules that flag usages of the request super globals and getUnsafeRawObject.

at the moment i don't have a working dicto environment to test this, but i am still working on it.

> Tests: Please add automated tests that at least cover the core features of the HTTP library.

I would then supply the tests as a final adjustment.

src/HTTP/Wrapper/DirectArrayAccessDropInReplacement.php
```diff
@@ -0,0 +1,90 @@
...
+     * @inheritDoc
+     */
+    public function offsetGet($offset)
+    {
+        if (!$this->offsetExists($offset)) {
+            return null;
+        }
+
+        if (is_array($this->raw_values[$offset])) {
+            return $this->factory->to()->dictOf($this->factory->kindlyTo()->string())->transform($this->raw_values[$offset]);
```
mjansenDatabay: @chfsx Maybe I am wrong, but IMO this does not work with multi level POST array.

chfsx:
@mjansenDatabay II implemented the topic we discussed:
- deprecation of "raw" methods
- renaming of Services
- KeyValueAccess in refinery AND injection of the transformation to use. please note that I extend that for a new SuperGlobalDropInReplacement to have the possibility to throw a exception if data is modified
- removal of statics and reflection

currently missing are tests and dicto-rules which I will deliver after your feedback to the current state. thanks a lot!

chfsx:
@mjansenDatabay I implemented some test for the new part of the HTTP Service but I just fail to implement (or run) dictor-tests, not because they're broken but I just don't get a running environment for this.

klees:
Hi @chfsx,

could you just post a ticket that demands the dicto rule and hand it to me?

Thanks!

chfsx:
OC, see in https://mantis.ilias.de/view.php?id=31013

kergomard:
Thank you for the work on the refinery! We discussed this PR in the Technical Board today and decided to merge without waiting for the dicto rules to not slow things down additionally.

PR #3235, opened 2021-03-16
UI: fix some namespaces to make composer happy

mjansenDatabay:
This PR fixes PSR-4 namespaces issues for ILIAS 6 (similar to ILIAS >= 7).


Amstutz:
Thx a lot! We do need those inspections up and running and always showing all files being green ;-).

PR #3237, opened 2021-03-16
Setup GlobalCache: 30376 show memcached nodes in setup status

daniwe4:
https://mantis.ilias.de/view.php?id=30376


chfsx:
@daniwe4 Thank you very much!

PR #3238, opened 2021-03-16
Setup 30373: store default inst_id during installation

daniwe4:
- 'update lucence index' fails if inst_id equals to '', so set inst_id
  to '0'

https://mantis.ilias.de/view.php?id=30373


klees:
Picked to trunk.

PR #3239, opened 2021-03-16
Setup Chatroom: 30370 add chatroom to metric system

daniwe4:
- add ilChatroomMetricsCollectionObjective
- change string 'http://' to 'http', same for https
- log now represents the corect value
- show ilias_url and client_url if proxies are set

https://mantis.ilias.de/view.php?id=30370


mjansenDatabay:
Thx @daniwe4  and @klees

PR #3246, opened 2021-03-18
WAC: Fix ternary operator

mjansenDatabay:
This PR fixes an issue with a ternary operator which will result in an error in recent PHP versions.

There is no need to pick this to other versions.


chfsx (approved):

chfsx:
Thanks @mjansenDatabay !

PR #3248, opened 2021-03-20
WorkflowEngine: Fix unit tests and productive code for PHP 8

mjansenDatabay:
This PR will fix the unit tests for the `WorkflowEngine` component with PHP 8. It looks like some of the tests only passed (in the past) because you adjusted the expectations (gold samples) based on the generated output ;-).

Furthermore it removes  some @ (silence) operators used to suppress errors when parsing XML, handling regex results, opening files etc..


mbecker-databay:
Thank you so much!

PR #3249, opened 2021-03-20
Init/ErrorReporting: Use -1 in DEVMODE if PHP 8 or SHOWNOTICES = true

mjansenDatabay:
With this PR I suggest to set the error_reporting to -1 in DEVMODE when using PHP 8 or if SHOWNOTICES evaluates to true.

Reason: Because the PHP policy regarding notices and deprecations has changed in the last years, I highly recommend to report as much errors as possible.

### DIFF
--- a/Services/Init/classes/class.ilInitialisation.php
+++ b/Services/Init/classes/class.ilInitialisation.php
@@ -1063,12 +1063,8 @@ class ilInitialisation
      */
     protected static function handleDevMode()
     {
-        if (defined(SHOWNOTICES) && SHOWNOTICES) {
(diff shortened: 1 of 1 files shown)


matthiaskunkel:
Jour Fixe, 22 MAR 2021 : We highly appreciate this PR and accept it for trunk. As suggested in the PR this change affects only PHP 8. After having started the PHP 8 migration project and implemented several fixes to support PHP 8, a second PR can be offered to extend this behaviour to PHP 7 as well. This will allow us now to continue development and prevent a "unusable trunk" due to the stricter error handling.

PR #3250, opened 2021-03-22
LSO 29944: fix type 'extor' to 'extro'

daniwe4:
- also remove return statement that returns an non existing object


klees:
I have picked this to all maintained versions.

PR #3252, opened 2021-03-22
UI/Tree: Rectify Aria-Labels and According Rules

klees:
Hi everyone!

This is supposed to improve the accessibility of the UI-component `Tree`, as documented in [#29851](https://mantis.ilias.de/view.php?id=29851). Thanks Kristina from HSLU for looking into this and pointing out the fix. I also improved the accessibility-rules according to [the WAI aria practices](https://www.w3.org/TR/wai-aria-practices/#tree_roles_states_props).

Best regards!

### DIFF
--- a/src/UI/Component/Tree/Factory.php
+++ b/src/UI/Component/Tree/Factory.php
@@ -29,12 +29,6 @@ interface Factory
      *       Nodes MUST restrict themselves to a minimal presentation, i.e.
      *       they MUST solely display information supportive and relevant for
      *       the intended task.
-     *   accessibility:
-     *     1: Nodes with further subnodes MUST bear the "aria-expanded" attribute.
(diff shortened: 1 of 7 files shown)


Amstutz (approved):

Amstutz:
LGTM. However, would be good if somebody with somewhat more insight into a11y might look into that here. @atoedt , any feedback? 

Note due to the change of interface we should announce the change in the next JF. however, since we are closing in on the release, and the next JF is in one month, I believe we could fast track this. However if so, I would some 3rd opinion on the correctness of our solution.

Amstutz:
Since the release is approaching fastly, I merged this so that the test case can be re-tested. I also Pick this to trunk. We will announce it in the next JF.

PR #3254, opened 2021-03-23
[FIX] 0029727: Failed test: 1.1.1 Meta Bar bietet Alt-Texte für grafische Bedienelemente

chfsx:
See https://mantis.ilias.de/view.php?id=29727

### DIFF
--- a/Services/User/classes/class.ilUserAvatarResolver.php
+++ b/Services/User/classes/class.ilUserAvatarResolver.php
@@ -1,6 +1,7 @@
 <?php
 
 use ILIAS\UI\Component\Symbol\Avatar\Avatar;
+use ILIAS\UI\Factory;
 
 /**
  * Class ilUserAvatarResolver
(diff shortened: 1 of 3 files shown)


Amstutz (changes requested):
Hi @fschmid

Thx for fixing this. as noted in the Ticket, 

- [ ] plz adapt:

> accessibility:
>     * 1: Avatars MUST bear an aria-label with the username.

Note that for alt, the wcag states that:

> "When an image contains words that are important to understanding the content, the alt text should include those words. This will allow the alt text to play the same function on the page as the image. Note that it does not necessarily describe the visual characteristics of the image itself but must convey the same meaning as the image."

 See: https://www.w3.org/TR/WCAG20-TECHS/H37.html

I am not sure, whether the user name will cut it. Also, not sure if we need the aria label and the alt-text. Sry, to nag about those thing, but I like to have clarity, since this issue will/does appear all over the place. I am aware, that maybe you are no more expert in this that I am, therefore I asked for some more guidance in the ticket.

Amstutz:
Thx for the work, based on this I created an adapted PR according the discussion in Mantis. https://github.com/ILIAS-eLearning/ILIAS/pull/3312

PR #3255, opened 2021-03-23
LSO 29944: encode data from tiny to prevent lost text areas

daniwe4:
https://mantis.ilias.de/view.php?id=29944


klees:
I have cherry-picked this to release_6, release_7 and trunk.

PR #3256, opened 2021-03-23
Changed poll object to use standard offline handling. Also added section to object readme about offline handling

M-Barz:


setup/sql/dbupdate_05.php
```diff
@@ -6365,4 +6365,22 @@
         'value' => ['text', '30']
     ]
 );
-?>
\ No newline at end of file
+?>
+<#5778>
+<?php
+$ilCtrlStructureReader->getStructure();
+?>
```
klees: Hi @M-Barz,

you don't need to trigger that structure reload manually anymore. This will be done by the setup automatically everytime `update` is called.

Best regards!

smeyer-ilias (approved):
Thanks Marvin. Please merge this in the trunk.

PR #3257, opened 2021-03-24
Setup Component: 30035 add parent class for anonymous classes, so they

daniwe4:
can pass typehints

https://mantis.ilias.de/view.php?id=30035


klees:
I have merged this manually.

PR #3258, opened 2021-03-24
Fixed forward to ilPropertyFormGUI

smeyer-ilias:
This is a fix https://mantis.ilias.de/view.php?id=30233


smeyer-ilias:
R6 is not effected

mjansenDatabay:
Thx @smeyer-ilias

PR #3259, opened 2021-03-24
Setup GlobalCache/Chatroom: 30397 provide a ilDB while collecting GlobalCache metrics

daniwe4:
- use correct storage type for port in chatroom metrics


klees:
@daniwe4 Please provide separate PRs for the distinct changes, even if both are reported in one ticket. Thanks!

PR #3260, opened 2021-03-24
Setup Chatroom: 30397 use correct storage type for port in chatroom metrics

daniwe4:
https://mantis.ilias.de/view.php?id=30397


mjansenDatabay:
Thx @daniwe4

PR #3261, opened 2021-03-24
Setup GlobalCache: 30397_1 provide a ilDB while collecting GlobalCache metrics

daniwe4:
https://mantis.ilias.de/view.php?id=30397

### DIFF
--- a/Services/GlobalCache/classes/Setup/class.ilGlobalCacheMetricsCollectedObjective.php
+++ b/Services/GlobalCache/classes/Setup/class.ilGlobalCacheMetricsCollectedObjective.php
@@ -3,6 +3,7 @@
 /* Copyright (c) 2020 Richard Klees <richard.klees@concepts-and-training.de> Extended GPL, see docs/LICENSE */
 
 use ILIAS\Setup;
+use ILIAS\DI;
 
 class ilGlobalCacheMetricsCollectedObjective extends Setup\Metrics\CollectedObjective
 {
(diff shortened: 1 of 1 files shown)


Services/GlobalCache/classes/Setup/class.ilGlobalCacheMetricsCollectedObjective.php
```diff
@@ -15,11 +16,21 @@ public function getTentativePreconditions(Setup\Environment $environment) : arra
...
         }
 
+        // ATTENTION: This is a total abomination. It only exists to allow various
+        // sub components of the various readers to run. This is a memento to the
+        // fact, that dependency injection is something we want. Currently, every
+        // component could just service locate the whole world via the global $DIC.
+        $DIC = $GLOBALS["DIC"];
+        $GLOBALS["DIC"] = new DI\Container();
+        $GLOBALS["DIC"]["ilDB"] = $db;
+
```
chfsx, line 33: Do we need to "destroy" this after running this Objective?

Services/GlobalCache/classes/Setup/class.ilGlobalCacheMetricsCollectedObjective.php
```diff
@@ -98,5 +109,7 @@ public function collectFrom(Setup\Environment $environment, Setup\Metrics\Storag
             "component_activation",
             $component_activation
         );
+
+        $GLOBALS["DIC"] = $DIC;
```
chfsx, line 113: ah sorry, I think this does the "destroy" thing, right?

klees: Yes. We try to restore the environment to the previous state to be as non-invasive as possible =)

chfsx:
Thanks for the fix! I'll CP to trunk as well

PR #3263, opened 2021-03-25
fix 29897

alex40724:
Fix for 29897

### DIFF
--- a/Services/MediaObjects/classes/class.ilFFmpeg.php
+++ b/Services/MediaObjects/classes/class.ilFFmpeg.php
@@ -268,7 +268,7 @@ class ilFFmpeg
         $sec = (int) $a_sec;
         $cmd = "-y -i " . ilUtil::escapeShellArg($a_file) . " -r 1 -f image2 -vframes 1 -ss " . $sec . " " . ilUtil::escapeShellArg($target_file);
         //echo "-$cmd-"; exit;
-        $ret = self::exec($cmd . " 2>&1");
+        $ret = self::exec($cmd );
         self::$last_return = $ret;
         
         if (is_file($target_file)) {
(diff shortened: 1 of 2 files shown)


mjansenDatabay (approved):

mjansenDatabay:
@gvollbach Can you please check the `Virus Scan` with these changes?

Amstutz:
Hi @alex40724 

This change is already merged into 7 and trunk and is breaking my unit tests there. This is an issue, since I installed the git pre-push hook running the unit tests before pushing (since I broke a test as well 2 days ago ;-) ). Currently I can not push to 7 und trunk to to the active pre-push hook.

This seems to break the $compress_content $flag if true in zip of ilUtil, which I use in testImportSkin. The shell command is set from:
`"/usr/bin/zip -r '../skin1.zip' *"`
to:
`"/usr/bin/zip -r '../skin1.zip' \*"`

Is there some workaround for this issue?

Thx.

alex40724:
> Hi @alex40724
> 
> This change is already merged into 7 and trunk and is breaking my unit tests there. This is an issue, since I installed the git pre-push hook running the unit tests before pushing (since I broke a test as well 2 days ago ;-) ). Currently I can not push to 7 und trunk to to the active pre-push hook.

@Amstutz  Why is this merged to 7 and trunk? This is an open PR, I did not merge anything. I can't see any commit that did the merge, see e.g.
https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/Services/Utilities/classes/class.ilUtil.php
Last commit on Feb 12.

Amstutz:
Hi @alex40724 

Thx for the feedback.

You are correct, sry. Something else seems to be blocking my push. I just checked back on recently failing tests and I stumbled over this PR, did look close enough. Problem is fixed for me in 7 and trunk and had nothing to do with this one here.

Still not the failing unit tests. So this will break 6 and most certainly 7 and trunk if merged. The suspected reason is listed above.

PR #3266, opened 2021-03-26
Refactored random element provider and fixed unit test.

mBeym:
Changed test_getItemswithShuffle_shouldReturnShuffledItems to check if ilRandomArrayElementProvider::shuffle is called instead of checking if two arrays are different (which can fail because there is a chance that arrays get shuffled like the expected order)


mjansenDatabay:
@mbecker-databay This PR fixes a unit test in the `T&A Test Suite` which tried to test `Random`. IMO this test made no sense, because, well, testing random will randomly work, or not.

In  the last weeks  I already noticed PR  builds which failed because of  that.

PR #3268, opened 2021-03-29
add submit caption to standard form

Hadrian-II:
Adding the option to set a custom caption for the submit button in a KitchenSink UI Standard form. I had a Form that Creates something new and 'Create' was much better fitting than save, and i think there is possibility for much diversity in forms that should be supportet

### DIFF
--- a/src/UI/Component/Input/Container/Form/Standard.php
+++ b/src/UI/Component/Input/Container/Form/Standard.php
@@ -16,4 +16,14 @@ interface Standard extends Form
      * @return    string
      */
     public function getPostURL();
+
+    /**
+     * Sets the caption of the submit button of the form
+     */
+    public function withSubmitCaption(string $caption) : Standard;
+
+    /**
+     * Gets the submit caption of the form
+     */
+    public function getSubmitCaption() : ?string;
 }
--- a/src/UI/Implementation/Component/Input/Container/Form/Renderer.php
+++ b/src/UI/Implementation/Component/Input/Container/Form/Renderer.php
@@ -38,7 +38,7 @@ class Renderer extends AbstractComponentRenderer
         }
 
         $f = $this->getUIFactory();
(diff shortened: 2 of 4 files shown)


Amstutz:
Hi @Hadrian-II 

Thx a lot for this. We will need some cohesive set of rules for using of labels for certain contexts. IMO they should be well aligned to ensure application wide consistency. However, I completely agree, that "Save" will not always cut it. This is also already covered by one of the rules for forms:

```
     *     5: >
     *        In some rare exceptions the Buttons MAY be named differently: if “Save” is
     *        clearly a misleading since the action is more than storing
     *        the data into the database. “Send Mail” would be an example of this.
```

So you change not is not only good, but necessary for the component to comply with the rules. THX!

Also thx for the provided rendering tests. However the tests are somewhat incomplete, please add the following:

- [ ] Test Default: Add a test making sure, the label is "Save" if nothing has been set.
- [ ] Test Mutator: Add add test just testing if the mutator works as expected without rendering. I am aware that this is a subset of your rendering test, but also adding the trivial version is cheap an might help to find the cause of some defect quicker.

Thx a lot, kudos.

Hadrian-II:
Added the test for the Mutator.

I did not add a test for the default value "save" as the form definition does not know this value before rendering, and to make it aware of it would make the form definition depend on ilLanguage. but the standard render test renders the default save caption, so i don't see a problem here.

Also i have no idea why the tests are failing now, the errors are thrown by assFormulaQuestionTest which i have not touched, and as my commits are adding a new feature no existing code can be impaired by it.

Amstutz:
Hi, @Hadrian-II 

Sry for letting this sit somewhat. I got irritated by the failing tests, and did not check into it, since I thought you are still working on it. However, I just double checked locally, your added tests run fine. I fail to see any connections with the ones failing and yours.

Thx for your contribution. Kudos
Timon

PR #3270, opened 2021-03-29
0030460 - Implement Redirection to userprofile from latest article

enterfusion:


enterfusion:
https://mantis.ilias.de/view.php?id=30460

mjansenDatabay (changes requested):
Thx a lot for this PR @enterfusion!

IMO we don't need the `fragment` part with a posting id for an URL pointing to a user's profile, so please remove the `#` etc.

mjansenDatabay (approved):

PR #3273, opened 2021-03-30
[FIX] 0029643: UI Modal: Change in Modal renderer breaks existing usa…

chfsx:
…ges of interruptive modals

@mjansenDatabay / @Amstutz 
This is basically a revert of the commit https://github.com/ILIAS-eLearning/ILIAS/commit/a63a4a07b68bf6e8d8c4c51942c97c96e9f2f26e which I made and which is quite nonsense...

The interruptive Modal can be used with the "fallback" command of ilCtrl.

see https://mantis.ilias.de/view.php?id=29643


mjansenDatabay (approved):

Amstutz:
Thx a lot, cherry picked to trunk as well.

PR #3274, opened 2021-03-30
Less Ordering Question, Fix of: #25806

Amstutz:
Hi @alex40724 

I got the following issue for CSS of ordering Questions assigned: https://mantis.ilias.de/view.php?id=25806 .

Since this is the first time I am fixing CSS in content styles (content.less), I would like to have your approval before merging. Maybe you also like to merge, since the less is in your domain (COPage). 

Also, I observed, that content.less does not compile atm, due to usage of less variables in there. However, I think we should use the variables there as well, to get a more consistent look of the default UI. Is there a specific (maybe technical) reason, not to use the variables in content.less?

I propose the following: for 7 and trunk, introduce and use the less variables as proposed in this PR. In future we can shift more values to variables with this.

For 5.4 and 6, stick with using raw values. If you like to merge this PR, it would be great if you can set those manually when merging and picking this request for 7 and trunk.

### DIFF
--- a/Services/COPage/css/content.css
+++ b/Services/COPage/css/content.css
@@ -223,7 +223,7 @@ div.ilc_iim_ContentPopup {
   border-style: solid;
   border-width: 2px;
   background-color: #FFFFFF;
-  border-color: #A0A0A0;
+  border-color: #A8A8A8;
 }
 div.ilc_qover_Correct {
   background-repeat: no-repeat;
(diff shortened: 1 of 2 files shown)


alex40724:
Hi Timon,

thanks for the PR. We cannot use system style variables in content styles (see bug report).

I will take care of the bug.

Best regards
Alex

Amstutz:
Thx for the feedback. 

Not yet sold on the not using variables here, see comment in report.

PR #3278, opened 2021-03-30
UI/CombinedSlate: Add optional entry id for menu sub entries (ILIAS 6)

iszmais:
https://mantis.ilias.de/view.php?id=30469

### DIFF
--- a/src/GlobalScreen/Scope/MainMenu/Collector/Renderer/TopParentItemRenderer.php
+++ b/src/GlobalScreen/Scope/MainMenu/Collector/Renderer/TopParentItemRenderer.php
@@ -31,7 +31,7 @@ class TopParentItemRenderer extends BaseTypeRenderer
         foreach ($item->getChildren() as $child) {
             $component = $child->getTypeInformation()->getRenderer()->getComponentForItem($child, false);
             if ($this->isComponentSupportedForCombinedSlate($component)) {
-                $slate = $slate->withAdditionalEntry($component);
+                $slate = $slate->withAdditionalEntry($component, $this->hash($child->getProviderIdentification()->serialize()));
             }
         }
 
--- a/src/UI/Component/MainControls/Slate/Combined.php
+++ b/src/UI/Component/MainControls/Slate/Combined.php
@@ -14,9 +14,6 @@ interface Combined extends Slate
 {
     /**
      * @param Slate|Bulky|Horizontal $entry
-     * @param string|null            $id
-     *
-     * @return Combined
      */
     public function withAdditionalEntry($entry, ?string $id = null) : Combined;
 }
--- a/src/UI/Component/MainControls/Slate/Combined.php
+++ b/src/UI/Component/MainControls/Slate/Combined.php
@@ -5,6 +5,7 @@
 namespace ILIAS\UI\Component\MainControls\Slate;
 
 use ILIAS\UI\Component\Divider\Horizontal;
+use ILIAS\UI\Component\Link\Bulky;
 
 /**
  * This describes the Combined Slate
@@ -13,6 +14,9 @@ interface Combined extends Slate
 {
     /**
      * @param Slate|Bulky|Horizontal $entry
+     * @param string|null            $id
+     *
+     * @return Combined
      */
-    public function withAdditionalEntry($entry) : Combined;
+    public function withAdditionalEntry($entry, ?string $id = null) : Combined;
 }
--- a/src/UI/Implementation/Component/MainControls/Slate/Combined.php
+++ b/src/UI/Implementation/Component/MainControls/Slate/Combined.php
@@ -20,14 +20,14 @@ class Combined extends Slate implements ISlate\Combined
     const ENTRY_ACTION_TRIGGER = 'trigger';
 
     /**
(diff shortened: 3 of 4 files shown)


chfsx:
@Amstutz and @klees this change is mainly in UI, adapting it in the GlobalScreen MainBar-"Renderer" is fine for me

chfsx (approved):
Fine for the GlobalScreen part

klees (changes requested):
Hi @iszmais!

Thanks a lot for looking into this.

Technically the code looks trivially and is fine. From a functional perspective, we are not convinced that, in general, using GUI modifications like this is a good way to modify ILIAS. We should be looking into more powerful ways to adapt ILIAS to certain use cases, e.g. by directly replacing or modifying the components you are looking to change, instead of changing their appearance after they are turned into UI components. This goal is reflected in the strategy of the TB and we kindly ask you to help here: As a maintainer of a plugin (or plugins) that want to change components in the Main Bar, please contact the respective maintainer of the component or the TB to find appropriate mechanisms to modify the component in question. Still, this is not reality yet and we understand that you have an urgent requirement to just change these components. We are fine to introduce the mechanism you are proposing here, but, as often, have some requests for improvement to make this more robust. Also, we will need to run this by the Jour Fixe as we would normally not make interface changes in already released versions of ILIAS. Please make sure to follow the development of our interfaces in the future to voice your requirements early in the development process.

- [x] Doc-Strings: Only use docstrings where they strictly provide more information than the typehints.
- [x] Unit Test 1: Write a Unit Test that proves that the parameter has indeed the effect you expect.
- [x] Unit Test 2: Write a Unit Test that proves that each id can only be used once per combined slate.

Since the proposed change is breaking, we feel that it would be ok to make the `$id` optional for ILIAS 6. But, in general, that interface is not where we would want to go. Open a PR for `release_7` that, on top, adds these two changes:
- [ ] Require Id: Why is the id optional only? This, from an interface perspective, would mean that you could end up with entries that still have no id after all. This would defeat your purpose. Make the id required.
- [ ] Streamlining: Make the id the first parameter to `withAdditionalEntry`, just like it is on `MetaBar` and `MainBar`.

We would be very pleased if you would join the next JF to present your change. We are sure that @mjansenDatabay could help with advice. Feel free to contact us again for questions or further advice!

Best regards!
@Amstutz and @klees

iszmais:
Hi @klees ,
I implemented the requested changes for ILIAS 6 and also added an Exception for doubled entries to verify the process.

Im concerned about the requested changes for ILIAS 7 since the withAdditionalEntry() Function is used in at least 10 places and im assuming even more additional polymorphic calls.
To implement the requirement i assume it would be necessary to apply all calls and some content based switches.

I would appreciate if you could gave a brief statement about if this is your implied intention.

Regards!
@iszmais

klees:
Hi @iszmais,

we should definitely raise these concerns regarding `release_7` on the Jour Fixe. I would definitely expect you to change the calling locations as well. From my inspection of the current state of `release_7`, most of these locations are in the examples, only four of these locations would be actually live code. Feel free to convince me otherwise =)

Maybe it would be simpler to judge this based on an actual PR?

Best regards!

klees (approved):

iszmais:
Hi @klees ,

I agree. I created an new PR for ILIAS 7 : https://github.com/ILIAS-eLearning/ILIAS/pull/3315

Regards,
@iszmais

matthiaskunkel:
Jour Fixe, 19 APR 2021: We highly appreciate this suggestion and accept the PR for ILIAS 6. Implementation for ILIAS 7 and trunk is tackled in PR #3315. Changes to 6 will be adapted similar for 7 due to upcoming release date of 7.0. Changes for trunk can be more extended.

iszmais:
@Amstutz 
Could you merge this PR, as @klees  will probably not be able to take action here for the time being?

Amstutz:
@iszmais sure. However with great power comes great responsibility :-). Since I will be the one merging, I will have to lock at it in a level of detail, I did not up to this point. This takes some time. 

I hope to be able to do so today. If you did not get reply from me by tomorrow evening, ping me.

Amstutz:
@iszmais : Merged to 6, 7 and trunk. As discussed, for 6 and 7 the current status is fine due to the fast approaching release. 

@iszmais : For trunk, sind the current state is inconsistent with the meta bar as @klees points out, It would be neat to have a PR soon, fixing this for trunk. Can you provide this?

iszmais:
@Amstutz Thanks alot! If all conceptual questions of #3315 are answered, i could provide a PR for this. But this of course also depends on the resulting expenses of this questions.

Amstutz:
Hi @iszmais Unfortunately this led to issues, due to the length of the used IDs, see: https://mantis.ilias.de/view.php?id=30638#c73867 . For this to be merged, we would need to do some work to cut down the length of the ID. However, this has not much to do with the UI Component, this is task of the Global Screen. You might want to involve @chfsx to look for ways on how to cut down the length of the IDs. However, I am not sure if this is really the way to tackle, maybe we should make a short workshop along with @chfsx and @nhaagen to check if we can solve this somewhat better.

PR #3284, opened 2021-04-01
Fix Scrolling in TinyMCE

kergomard:
This is a possible fix for: https://mantis.ilias.de/view.php?id=30183
I would not call this positively beautiful, but as far as I can see, it works.

I will provide a separate PR for ILIAS 7 and ILIAS Trunk as soon as https://github.com/ILIAS-eLearning/ILIAS/pull/3089 is merged. With the new version we won't need the additional css-file, but can solve this through the "content_style"-Setting of TinyMC.


mjansenDatabay (approved):
LGTM! As far as I understand this new CSS file will only be loaded/applied in the TinyMCE popup window, so I consider these changes as minimally invasive.

kergomard:
Thx @mjansenDatabay . Yes, this was the least invasive approach I could find.
I will merge this, now. I think the changes in css are so small, that this shouldn't bother you @Amstutz .

PR #3285, opened 2021-04-02
PublicArea: Add redirect if not enabled

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=30108


smeyer-ilias:
Thanks for fixing this

PR #3289, opened 2021-04-07
Fix 30513: Fatal error because of access level

tfamula:
https://mantis.ilias.de/view.php?id=30513

### DIFF
--- a/Services/Certificate/classes/class.ilObjCertificateSettingsGUI.php
+++ b/Services/Certificate/classes/class.ilObjCertificateSettingsGUI.php
@@ -38,7 +38,7 @@ class ilObjCertificateSettingsGUI extends ilObjectGUI
     /**
      * @var string
      */
-    private $type;
+    protected $type;
 
     /**
      * @var ilAccessHandler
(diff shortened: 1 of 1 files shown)


tfamula:
@alex40724: FYI

Services/Certificate/classes/class.ilObjCertificateSettingsGUI.php
```diff
@@ -38,7 +38,7 @@ class ilObjCertificateSettingsGUI extends ilObjectGUI
     /**
      * @var string
      */
-    private $type;
+    protected $type;
```
mjansenDatabay: @tfamula I guess we can remove this property completely. IMO it makes no sense that this is declared redundantly in the certificate class.

mjansenDatabay (changes requested):
Thx @tfamula , but we should delete the property instead.

Services/Certificate/classes/class.ilObjCertificateSettingsGUI.php
tfamula: @mjansenDatabay This was actually my first thought, but I was not completely sure about it :+1:

tfamula: @mjansenDatabay Updated!

mjansenDatabay (approved):

PR #3291, opened 2021-04-08
 UI Components, Tree, integrate Keyboard support, see: #30505

Amstutz:
This PR mainly fixes [#30505](https://mantis.ilias.de/view.php?id=30505), thx to @akill for reporting. However, while implementing keyboard support for the UI Components tree, I observed some further detail needing some love. 

Please note, that I tried to implement this as closely to the wcag specs as possible. Noteworthy:

> only one treeitem in the tree has tabindex=„0“

see: https://www.w3.org/TR/wai-aria-practices/examples/treeview/treeview-2/treeview-2a.html . Meaning there is only one element at once accessible by tabbing. The rest can be optioned by arrows, but not tabbing (similar to the current legacy tree implementation). Note that currently the complete set Keyboard Support has been implemented, including the recommended typehead by letters and multiple opening of nodes by asterisk (see: https://www.w3.org/TR/wai-aria-practices/examples/treeview/treeview-2/treeview-2a.html for a complete list).
 

Here an Overview of the other relevant changes:
- Changed rules and descriptions of the Tree slightly to match implementation
- Improved rendering for parts of tree that are loaded async. See withIsSubTree mutator. Before those sub tree have been rendered as individual trees inside a tree, which is semantically incorrect.
- Setting aria role to „none“ if a node features a link inside (which then has then aria-role treeitem according to the specs) and has no subnodes and is not loaded async.
- The „expanded“ class is redundant since with have „aria-expanded“ set. I removed it and use only area expanded.
- Added some test for async rendering and expanded setting
- We have a race condition in ilTreeExplorerGUI which uses onclick in competition with tree.js. I am not sure, we this has not led to unexpected behavior before know, but while adapting for the removal of the „expanded“ class I run into issues with this. I do not like the timeout solution, but it was the best I could think of. @akill, any idea, why this has not been an issue before this change?


src/UI/Component/Tree/Tree.php
```diff
@@ -54,4 +54,14 @@ public function withHighlightOnNodeClick(bool $highlight) : Tree;
...
      */
     public function getHighlightOnNodeClick() : bool;
+
+    /**
+     * Is this only a part of a tree? Needed if parts are loaded async
+     */
+    public function isSubTree() : bool;
+
+    /**
+     * Set this tree to be a part of a tre. Needed if parts are loaded async.
```
klees:
```suggestion
     * Set this tree to be a part of a tree. Needed if parts are loaded async.
```

alex40724:
@mjansenDatabay Could you have a look at Timons issue in the ilTreeExplorerGUI javascript code, please? This is mostly related to your changes in 2019. This just sends the ajax request, right? I am not sure if the timeout leads to issues. It would be fine for me.

mjansenDatabay:
@Amstutz / @alex40724 For me the changes are fine as well.

Maybe you could furher eleborate on the `Race Condition`. IMO/AFAIR the two event handles for the `click` event are registered by purpose.

Of course there is a temporal coupling of the listeners, which means:
* The `callback` for `tree.find('.il-tree-node .node-line')` in src/UI/templates/js/Tree/tree.js is expected to be called first.
* The callcack registered in `\ilTreeExplorerGUI::build` is expected to be registered (and thus called) after the first one.

klees (changes requested):
Hi everyone,

I discussed this with @Amstutz today. There were several pros and cons that we discussed:

* This is mostly a copy from a WAI Aria example. This is **good** because we hope that we can skip some discussion this way, because the tree will now just behave as in the example.
* This is also **bad**, because the code looks like a stranger. Variables are named differently then we would do it, we don't use `prototype` (often) etc. 
* In general we think that the "redux-pattern" that @nhaagen introduced for the Main Bar should be a good fit for many UI components. At least the Tree would be a good target for that pattern. Also, this would allow to add unit tests!

We both agreed that we still would want to add the code as is, since it allows us to have a keyboard controllable tree soon and also come to an agreement regarding the behavior we want. Regarding the implementation (and unit tests), @Amstutz will add an according entry to the roadmap.

Best regards!

Amstutz:
The entry had been added, thx for the Feedback.

Amstutz:
There have been some js improvements to remove the race condition completely. Further, this makes sure the aync call back to save the state of the tree is not only fired for click events, but for any action toggling the expanded state of a node. AFAIK, this could be ready for merging. Thx to everybody involved.

mjansenDatabay (approved):

klees (approved):

klees:
Rebased, squashed and merged as: https://github.com/ILIAS-eLearning/ILIAS/commit/7b33f68bade95a3b4ae1a45e8bc4179eddbb8749

PR #3292, opened 2021-04-09
fixed #30527 wrong ldap server instert query

chfsx:
fixed an issue where column-list did not match column-values, added missing '%s' query binding. (see https://mantis.ilias.de/view.php?id=30527)


mjansenDatabay:
Fixed with https://github.com/ILIAS-eLearning/ILIAS/commit/ef5955db37eee28ceb96470d021334791197edd8

PR #3293, opened 2021-04-09
Standard Layout, Introduce Text Direction and Rendering Tests, see: #30532 and #29426

Amstutz:
This PR is the first part in tackling the issue reported in: https://mantis.ilias.de/view.php?id=29426 by adding mutators for the text direction. A second part will be the correct usage of those mutators by global screen. I did not want to bundle the two to keep the complexity manageable. 

By providing the necessary tests, I got reminded that we did not have rendering tests for the standard layout. This PR also contains a fix for this, by adding an option to render parts of a component as "Stub". E.g. one can now render the as Mock Provided Main Bar in the Page Layout as simple string just displaying the Canonical Name of the Component. This could be useful for other components as well, featuring many sub components in their default version. Since I believe the missing of rendering tests for a component to be an issue, I created a ticket for this as well, which is hereby adressed: https://mantis.ilias.de/view.php?id=30532

### DIFF
--- a/src/UI/Component/Layout/Page/Standard.php
+++ b/src/UI/Component/Layout/Page/Standard.php
@@ -17,6 +17,9 @@ use ILIAS\UI\Component\MainControls\Footer;
  */
 interface Standard extends Page, JavaScriptBindable
 {
+    //Possible Text Directions
+    public const LTR = 'ltr';
+    public const RTL = 'rtl';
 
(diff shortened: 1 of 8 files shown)


klees (approved):
LGTM, thanks a bunch for the drive by improvement of the testing infrastructure!

PR #3294, opened 2021-04-09
Global Screen, Text Direction Fix: see #29426

Amstutz:
This is the second Part of the Fix for: https://mantis.ilias.de/view.php?id=29426, Part I is here: https://github.com/ILIAS-eLearning/ILIAS/pull/3293. 

@chfsx : Those changes are targetting mostly the Global Screen.  Please check the changes, I am not 100% that they are in the right places. The system works, I checked, but I am somewhat unsure if this properly follows the logic of the components in there. Further, not that in Global Screen are some unrelated changes about the coding format. I had to fix the coding violations, otherwise the pre-commit hook did not allow me to commit my changes. Note that there are a few more in Global Screen. Maybe best run  ./CI/PHP-CS-Fixer/run_check.sh only for Global Screen without the Dry Run Option in run_check.sh.

@akill, @chfsx , @katringross: Since I needed to feed the global screen with the information on whether the current language ist ltr of rtl, I checked into the previous implementations. Turns out, there are 4 all in Templating classes where they imo are not perfectly placed. So I moved them to Language, where this logic IMO belongs. I hope this is fine. If so, plz approve this change.

Edit: This last part has been removed.


katringross:
@Amstutz our developer @iomidev found to points, where we are not sure if it is 100% correct. 
1. Dropdown menu is still on the right. It should move to left instead or what do you think?
2. On Layout and Navigation, the labels should also be right justified.

rtl_screenshot

Amstutz:
Hi @katringross 

Thx for the feedback. Valid points. 

Note: This PR does not influence or change the RTL on a css level in any way. This is something that would be needed to be done in a follow up. This PR simply makes it possible to even use RTL (so e.g. one could use RTL in a custom skin or similar). 

For this PR I would need some thumbs up (or down) from you as maintainer for the changes made in the changes I propose for the language component.

Thx.

katringross (approved):

chfsx:
@Amstutz I made some adjustements for the GlobalScreen part as discussed via skype: https://github.com/Amstutz/ILIAS/pull/10

Amstutz:
Thx a lot @chfsx I cherry picked your changes into this one here. Thx for your help. Should be ready for merging now.

PR #3295, opened 2021-04-09
Fix paths to learning module and mob upload direcotries

fwolf-ilias:
I changed the paths based on issue https://mantis.ilias.de/view.php?id=30137 . Also i would suggest to increase the visibility of this documentation because i can't imagine that anyone will find this feature whos is not involed in its developement.


smeyer-ilias (approved):
Thanks for fixing this

PR #3296, opened 2021-04-09
Switching repository chat message field to textarea and also switched…

dmundra:
… keypress to keydown.

This is a suggested solution to the feature request: [Repository Chat: Message box supports multiple lines](https://docu.ilias.de/goto_docu_wiki_wpage_6782_1357.html)

Keypress is deprecated according to https://developer.mozilla.org/en-US/docs/Web/API/Document/keypress_event

### DIFF
--- a/Modules/Chatroom/js/iliaschat.jquery.js
+++ b/Modules/Chatroom/js/iliaschat.jquery.js
@@ -1556,8 +1556,9 @@ var ServerConnector = function ServerConnector(url, scope, user, userManager, gu
 		});
 
 		// when the client hits ENTER on their keyboard
-		$('#submit_message_text').keypress(function(e) {
-			if(e.which == 13) {
+		$('#submit_message_text').keydown(function(e) {
+			var keycode = e.keyCode || e.which;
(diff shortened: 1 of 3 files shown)


mjansenDatabay:
Thx for your contribution @dmundra . There is one issue with the textarea which should be addressed before I take this PR to the ILIAS JourFixe to decide whether or not we integrate this also in our stable releases:

- [x] Linebreaks should be possible when pressing SHIFT + ENTER


Best regards,
@mjansenDatabay

dmundra:
@mjansenDatabay easy enough to check for SHIFT + ENTER (added that check from https://stackoverflow.com/a/6015906) but the chat display does not retain the line breaks. ~~Looking to see if can adjust that next.~~

Went with a CSS fix in the next commit.

matthiaskunkel:
Jour Fixe, 19 APR 2021: We highly appreciate this suggestion and accept the PR as usability fix for all maintained versions.

dmundra:
Awesome news. Thanks, @matthiaskunkel!

PR #3302, opened 2021-04-13
added date/time input issues to roadmap

alex40724:
See https://mantis.ilias.de/view.php?id=29816


Amstutz:
Hi @alex40724 

I definitely agree that we need accessible solutions. 

Question: Since it is possible, to enter the date by keyboard only, not using the picker. Is this really an issue? AFAIK this is already accessible in a WCAG sense, since the complete set of features offered by this picker is available by keyboard as well. If this is not the case, we should indeed add this to the roadmap as suggested.

Thx

Amstutz:
Note, we certainly would need a solution for the buttons, other than for the inputs, there is simply no alternative...

PR #3303, opened 2021-04-13
BT-29425: Fixed wrong language variable in mobile search view

M-Barz:
Link to Mantis-Report:
https://mantis.ilias.de/view.php?id=29425


smeyer-ilias (approved):
Thanks

PR #3304, opened 2021-04-13
BT-29427: Changed language variable in mobile view of meta bar to refer to user instead of avatar

M-Barz:
Link to Mantis-Report:
https://mantis.ilias.de/view.php?id=29427


smeyer-ilias (approved):
Thanks

PR #3314, opened 2021-04-15
Marked ilUtil as deprecated

chfsx:
The 2021 Technical Board has decided to mark the ilUtil class as deprecated. The ilUtil is a historically grown helper class with many different UseCases and functions. The class is not under direct maintainership and the responsibilities are unclear. In this context, the class should no longer be used in the code and existing uses should be converted to their own service in the medium term. If you need ilUtil for the implementation of a new function in ILIAS > 7, please contact the Technical Board.

### DIFF
--- a/Services/Utilities/classes/class.ilUtil.php
+++ b/Services/Utilities/classes/class.ilUtil.php
@@ -10,9 +10,14 @@ use ILIAS\FileUpload\DTO\UploadResult;
 /**
  * Util class
  * various functions, usage as namespace
+ * @author     Sascha Hofmann <saschahofmann@gmx.de>
+ * @author     Alex Killing <alex.killing@gmx.de>
  *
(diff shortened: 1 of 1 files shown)


alex40724:
Great news. But I think the whole Services/Utilities folders suffers from this and should be "abandoned".

I removed lot of unused code in this directory in the last weeks and moved some Dom-related classes to COPage, since they have not been used anywhere else.

In ilUtil, we have some widely used functions:

- ilUtil::stripSlashes
- ilUtil::makeClickable
- ilUtil::getImagePath

How should "consumer" code be migrated for these?

matthiaskunkel:
Jour Fixe, 19 APR 2021: We highly appreciate this suggestion and accept that ilUtil is set to deprecated. Please contact Fabian or Timon in case of problems in your components that rely to this decision.

PR #3317, opened 2021-04-16
Declare DIC for autocomplete in IDEs

Amstutz:
Hi @alex40724 

Currently I have the issue, that autocomplete often fails me for $DIC in PHPStorms. This change here improves this, by declaring the nature of $DIC in a central place. Maybe others suffer from the same Issues or other IDE's might profit as well from this declaration. Not sure if this is the ideal place. If there are strong opinions on some alternative, I am happy to adapt.

Thx, for considering.

### DIFF
--- a/ilias.php
+++ b/ilias.php
@@ -15,7 +15,10 @@
 require_once("Services/Init/classes/class.ilInitialisation.php");
 ilInitialisation::initILIAS();
 
-global $ilCtrl, $ilBench;
+/**
+ * @var $DIC \ILIAS\DI\Container
+ */
+global $DIC, $ilBench;
 
-$ilCtrl->callBaseClass();
+$DIC->ctrl()->callBaseClass();
 $ilBench->save();


alex40724:
Hi @Amstutz,

thanks for the PR. @klees introduced the DIC, so I leave it to him to decide where the best place for this declaration should be. Furthermore this script is part of the initialization process imo. So I put @smeyer-ilias here as a responsible, too. I will remove me as an "@author" from the script in the trunk, I think it should be maintained together with Services/Initialization.

klees:
Hi @alex40724,

I'm fine with this change. But: I don't see why I would be responsible for a typehint in `ilias.php`, just becaused I introduced the hinted class and concept. Things would get interesting if we applied that to other classes and concepts =)

If you ask me: if used as intended, no one would even need that typehint on `DI\Container`.

Best regards!

alex40724:
> I don't see why I would be responsible for a typehint in ilias.php

Me neither. I wrote "I leave it to him to decide **where the best place for this declaration should be**".

smeyer-ilias:
Thanks for removing one global

PR #3319, opened 2021-04-18
#0030242: Failed test: TeX im Forum anzeigen

fneumann:
This a fix for Mantis bug 30242.
If RTE ist not enabled for the forum, the authors still can add TeX code with the delimiters [tex] and [/tex]. This is correctly displayed in the forum if the servier-side rendering of mathjax is enabled for the browser. The detail view of generated news items however looks strange because a post-processing in ilNewsDefaultRendererGUI::getDetailContent corrupts the svg code. This should not be done if ilNewsItem::getContentHtml indicates that the content is already HTML. The creation of the news item in the forum code, however checked that with the original message but not with the content after processing by ilMathJax.
Additionally ilNewsDefaultRendererGUI::makeClickable should not replace URLs in SVG code.

### DIFF
--- a/Modules/Forum/classes/class.ilForum.php
+++ b/Modules/Forum/classes/class.ilForum.php
@@ -499,7 +499,7 @@ class ilForum
             $news_item->setPriority(NEWS_NOTICE);
             $news_item->setTitle($objNewPost->getSubject());
             $news_item->setContent(ilRTE::_replaceMediaObjectImageSrc($this->prepareText($objNewPost->getMessage(), 0), 1));
-            if ($objNewPost->getMessage() != strip_tags($objNewPost->getMessage())) {
+            if ($news_item->getContent() != strip_tags($news_item->getContent())) {
                 $news_item->setContentHtml(true);
             }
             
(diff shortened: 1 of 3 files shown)


mjansenDatabay (approved):
The forum changes LGTM!

alex40724 (changes requested):
@mjansenDatabay But this will make usual URLs in the news text not "clickable" right?

I think this should be solved in makeClickable somehow. This method has been introduced by Wolfgang Merkens (Databay) ages ago. I had to fix security issues in it once, but I don't maintain this code any longer.

Imo this needs a decent and secure transformation logic.

fneumann:
The makeClickable() is called independently from whether the news item content is declared as HTML or not. It can definitly be improved in a way that no HTML tag with a URL attribute is corrupted. The PR can't solve that problem generally. It treats the case if an svg tag is included in the news item. This will only be the case if the MathJax server is activated for the browser and if the forum article contains TeX code. In this case the links will not be made clickable.

kergomard:
Hi @mjansenDatabay 
This PR has been stuck for quite a while now. Could it be closed or merged?

Best regards!

mjansenDatabay:
@alex40724 / @fneumann 

Regarding `makeClickable`: For `trunk` you could try https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/Refinery/String/Group.php#L100 . This `Transformation` contains unit tests for the `TeX` scenario, see: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/tests/Refinery/String/MakeClickableTest.php#L60 .

Although it is not perfectly safe (still relying on `RegEx`) when operating on DOM trees (see ROADMAP: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/Refinery/ROADMAP.md#mid-term), you could maybe give it a try.

PR #3322, opened 2021-04-21
Fix use news container setting in course import xml

ghost:
Set `cont_use_news` `ContainerSetting` on course import xml (For instance on soap `addCourse`) works per se in https://github.com/ILIAS-eLearning/ILIAS/blob/96c6f65ccad6fc39680e44412009b29de6bf8cdf/Modules/Course/classes/class.ilCourseXMLParser.php#L603, but later on https://github.com/ILIAS-eLearning/ILIAS/blob/96c6f65ccad6fc39680e44412009b29de6bf8cdf/Modules/Course/classes/class.ilCourseXMLParser.php#L502 > https://github.com/ILIAS-eLearning/ILIAS/blob/96c6f65ccad6fc39680e44412009b29de6bf8cdf/Modules/Course/classes/class.ilObjCourse.php#L1341 > https://github.com/ILIAS-eLearning/ILIAS/blob/96c6f65ccad6fc39680e44412009b29de6bf8cdf/Services/Container/classes/class.ilContainer.php#L1046 it will be overridden with the previous value

This PR proposes to use `setUseNews` from `ilObjCouse` instead of `_writeContainerSetting` from `ilContainer` directly, so the `update` works with the new value

### DIFF
--- a/Modules/Course/classes/class.ilCourseXMLParser.php
+++ b/Modules/Course/classes/class.ilCourseXMLParser.php
@@ -600,11 +600,15 @@ class ilCourseXMLParser extends ilMDSaxParser implements ilSaxSubsetParser
             
             case 'ContainerSetting':
                 if ($this->current_container_setting) {
+                    if ($this->current_container_setting === ilObjectServiceSettingsGUI::USE_NEWS) {
+                        $this->course_obj->setUseNews($this->cdata);
+                    } else {
                     ilContainer::_writeContainerSetting(
                         $this->course_obj->getId(),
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
Thx for your PR @fluxfw , could you please create a Mantis issue according https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/contributing.md#list-of-labels ?

ghost:
Part of https://mantis.ilias.de/view.php?id=25388

PR #3325, opened 2021-04-22
Fix disabled course xml import

ghost:
Set `subscriptionLimitationType` to deactivated on course import xml (For instance on soap `addCourse`) not works because it uses the getter not the setter


ghost:
May part of https://mantis.ilias.de/view.php?id=25388

PR #3327, opened 2021-04-22
Fix double view mode course xml export

ghost:
`ViewMode` on course xml export is double set


ghost:
May part of https://mantis.ilias.de/view.php?id=25388

PR #3329, opened 2021-04-22
Add whitespaces instead of line brakes for latest article view in forum

kergomard:
This is a very small usability improvement fixing an edge case in the list view of a forum. It ensures a white space is shown in the "last posting" line when a new line is in the posting part to be shown.

If accepted this should probably be cherry-picked to all maintained versions (I can do this, if you want).

### DIFF
--- a/Modules/Forum/classes/class.ilObjForumAccess.php
+++ b/Modules/Forum/classes/class.ilObjForumAccess.php
@@ -135,6 +135,7 @@ class ilObjForumAccess extends ilObjectAccess
      */
     public static function prepareMessageForLists($text)
     {
+        $text = str_replace('<br />', ' ', $text);
         $text = strip_tags($text);
         $text = preg_replace('/\[(\/)?quote\]/', '', $text);
         if (ilStr::strLen($text) > 40) {


mjansenDatabay (changes requested):
Thx @swiniker for this PR. I highly appreciate this fix.

Please ask the following question:

- [ ] Is it sufficient to to replace `<br />`? What about `<br>` and `<br/>`? Does the input processing (`TinyMCE` on the client, `HtmlPurifier` on the PHP backend) ensure that there is only one characteristic of a HTML line break?

Best regards,
@mjansenDatabay

kergomard:
Thx @mjansenDatabay for the fast feedback!
I had checked through the interface and I always got `<br />` back from the TinyMCE.
Here a more formal answer after some research in the code:
We are not setting the element_format option when initializing tinymce thus it falls back to rendering as xhtml (https://www.tiny.cloud/docs/configure/content-filtering/#element_format). I think this should already ensure that only `<br />` reaches the php-side of things (unless somebody tampers with javascript).
Now, for the PHP-Side:
- You set the HTMLPurifier to "'HTML.Doctype', 'XHTML 1.0 Strict'" (https://github.com/ILIAS-eLearning/ILIAS/blob/cb9e5b8231b5b92bbf7d86e4e2336f65be091b66/Services/Html/classes/class.ilHtmlForumPostPurifier.php#L28) -> should lead to `<br />`or `<br/>` (from my understanding)
- And then the HTMLPurifier does change `<br/>` to `<br />` on line 51 of /ilias6/libs/composer/vendor/ezyang/htmlpurifier/extras/ConfigDoc/HTMLXSLTProcessor.php, if I'm getting this right.

If you would like to be more flexible when it comes to changes in tinymce we could use a preg_replace instead and cover all cases. A little bit more expensive, but that should not matter much.

mjansenDatabay:
Thx as well for the fast and elaborated feedback @swiniker 

I am fine with the current solution for now based on your analysis. If the forum behaves different with the new `TinyMCE 5`, we simply have to adapt the solution to the new input handling of the `TinyMCE`. For me this is a feasible way because this only affects the rendering, not the persistence (and thus the fix could easily be extended).

mjansenDatabay (approved):

mjansenDatabay:
I'll pick this to the maintained branches.

PR #3330, opened 2021-04-22
UI Components, Meta Bar Dom Structure, see #29717 and #28735

Amstutz:
Currently we have issues with the Tabbing order in the Meta Bar, see: https://mantis.ilias.de/view.php?id=29717 and https://mantis.ilias.de/view.php?id=28735 . They are caused by the order of the html elements in the DOM. A similar thing is going on in the Main Bar. However, while the solution for the Main Bar is not clear and not easily changeable (e.g. due to the responsive view). A nested solution is possible for the Meta Bar, which is provided here.

I tried to implement as closely to the WCAG example as possible, see: https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html . Maybe I missed some point, if so point me to them ;-). Not that I did NOT change the content of the Slate. The structure now looks as follows (some attributes are left out, see the provided rendering test for more detail):

```
<nav class="il-maincontrols-metabar" aria-label="Metaleiste">
   <ul class="il-metabar-entries" role="menubar">
      <li role="none">
        <button class="btn btn-bulky" role="menuitem" ></button>
        SlateWithContent
      </li>
      <li role="none">
        <button class="btn btn-bulky" role="menuitem" ></button>
        SlateWithContent
      </li>
     ....
   </ul>
</nav>
```

This means, that the tabbing order is now as requested in the ticket (jumping inside the slates when opened).

There should be no visual change or any change in behavior while operating with the mouse on large and small displays.

### DIFF
--- a/src/UI/templates/js/MainControls/metabar.js
+++ b/src/UI/templates/js/MainControls/metabar.js
@@ -86,7 +86,7 @@ il.UI.maincontrols = il.UI.maincontrols || {};
 		};
 		var _disengageAllButtons = function() {
 			$('#' + id +' .' + _cls_entries)
-			.find('li > .btn.' + _cls_btn_engaged)
+			.children('li').children('.btn.' + _cls_btn_engaged)
 			.each(
 				function(i, btn) {
 					_disengageButton($(btn));
@@ -164,7 +164,7 @@ il.UI.maincontrols = il.UI.maincontrols || {};
 
 		var _getMetabarEntries = function() {
 			return $('#' + id +' .' + _cls_entries)
-				.find('li > .btn')
+				.children('li').children('.btn')
 				.not('.' + _cls_more_btn);
 		}
 
--- a/src/UI/templates/js/MainControls/metabar.js
+++ b/src/UI/templates/js/MainControls/metabar.js
@@ -86,7 +86,7 @@ il.UI.maincontrols = il.UI.maincontrols || {};
 		};
 		var _disengageAllButtons = function() {
 			$('#' + id +' .' + _cls_entries)
-			.children('.btn.' + _cls_btn_engaged)
+			.find('li > .btn.' + _cls_btn_engaged)
 			.each(
 				function(i, btn) {
 					_disengageButton($(btn));
@@ -164,7 +164,7 @@ il.UI.maincontrols = il.UI.maincontrols || {};
 
 		var _getMetabarEntries = function() {
 			return $('#' + id +' .' + _cls_entries)
-				.children('.btn')
+				.find('li > .btn')
 				.not('.' + _cls_more_btn);
 		}
 
--- a/src/UI/templates/js/MainControls/metabar.js
+++ b/src/UI/templates/js/MainControls/metabar.js
@@ -7,7 +7,7 @@ il.UI.maincontrols = il.UI.maincontrols || {};
 
 		var id
 			,_cls_btn_engaged = 'engaged'
-			,_cls_entries = 'il-metabar-entries'
+			,_cls_entries = 'il-maincontrols-metabar'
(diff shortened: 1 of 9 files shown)


src/UI/templates/js/MainControls/metabar.js
```diff
@@ -86,7 +86,7 @@ il.UI.maincontrols = il.UI.maincontrols || {};
 		};
 		var _disengageAllButtons = function() {
 			$('#' + id +' .' + _cls_entries)
-			.children('.btn.' + _cls_btn_engaged)
+			.find('li > .btn.' + _cls_btn_engaged)
```
nhaagen: This probably is sufficient, but may have effects on further contents to come.
Maybe better explicitly stick to children.children or '> li >'  (does that work?)

tests/UI/Component/MainControls/MetaBarTest.php
```diff
@@ -119,6 +119,7 @@ public function brutallyTrimHTML($html)
         $html = str_replace(["\n", "\r", "\t"], "", $html);
         $html = preg_replace('# {2,}#', " ", $html);
         $html = preg_replace('/<!--(.|\s)*?-->/', '', $html);
+        $html = str_replace('> <', '><', $html);
```
nhaagen, line 122: :) I like that.

nhaagen (approved):
Hi Timon,
super cool addressing that issue.
I was a bit worried about dom-structure and changing to small screens, but this still seems to work.
Hooray and cudos, thank you.

Amstutz:
@nhaagen 

Thx for looking over and the feedback. You are correct, concerning the error prone selector used. I revised to children().children() which will also be a lot quicker. 

Thx.

klees (changes requested):
Hi @Amstutz!

Thanks a lot for taking actual care about various accessibility issue.

The solution looks fine and functional in general.

There is one thing where I am not convinced that the change is indeed an improvement. This is the introduction of the `<nav>` element around the various entries in the Meta Bar. In a nutshell: I'm not convinced that this is actually the correct element for the purpose semantically.

The current [living standard spec](https://html.spec.whatwg.org/multipage/sections.html#the-nav-element) and the [doc from mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav) both state that "The nav element represents a section of a page that links to other pages or to parts within the page: a section with navigation links.". This is not what the Meta Bar actually does. None of the entries (with the exception of the "User Avatar") does simply provide links to other pages or inside the document. Instead, they provide information about and interactions with various entities in the system (like the "Notifications" slate and the "Contacts" slate) or the search. So most certainly, the Meta Bar is not a `<nav>`.

Why would we want to have that additional `<nav>`(or `<div>`) wrapper anyway? Wouldn't it be enough to have that `<ul>` as the top node? I would thus suggest to remove that `<nav>` and adjust the selectors accordingly.

Feel free to contact me for further discussion!

Best regards!

Amstutz:
Hi @klees 

Thx for the feedback. I agree with your reasoning. I added the nav mainly because it was listed in the menubar example. But you are right. Seems no good fit. I removed it along with the dispensable div. The DOM looks somewhat cleaner, nice.

Note that I expect multiple iterations to be necessary until we land on a semantically reasonable version. However, since the release is coming soon, IMO this could be an acceptable version for 7. 

Thx.

Amstutz:
Added some CSS cleanup

klees:
Hi @Amstutz,

please remove that `il-metabar-entries`-class, we don't seem to need it anymore.

Best regards!

Amstutz:
@klees , done.

klees (approved):

klees:
Picked to trunk.

PR #3331, opened 2021-04-23
Revert "UI/CombinedSlate: Add optional entry id for menu sub entries (ILIAS 6)"

nhaagen:
Reverts ILIAS-eLearning/ILIAS#3278

The changes seem to have major impact on the length of mainbar-cookie, since enumaration is now supecseeded by longish globalscreen-ids. We reverted changes for now, an will have to think about better/other/bypassing solutions.

please see: https://mantis.ilias.de/view.php?id=30638

### DIFF
--- a/src/GlobalScreen/Scope/MainMenu/Collector/Renderer/TopParentItemRenderer.php
+++ b/src/GlobalScreen/Scope/MainMenu/Collector/Renderer/TopParentItemRenderer.php
@@ -31,7 +31,7 @@ class TopParentItemRenderer extends BaseTypeRenderer
         foreach ($item->getChildren() as $child) {
             $component = $child->getTypeInformation()->getRenderer()->getComponentForItem($child, false);
             if ($this->isComponentSupportedForCombinedSlate($component)) {
-                $slate = $slate->withAdditionalEntry($component, $this->hash($child->getProviderIdentification()->serialize()));
+                $slate = $slate->withAdditionalEntry($component);
             }
         }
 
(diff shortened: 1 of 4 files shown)


iszmais:
@nhaagen i **highly** recommend to solve the initial problem! These informations should not be stored inside a cookie, or at least should be reduced to an essential level!

The reverted changes are an improvement to the core structure validated by the JourFixe and shouldn't be reverted without the same level of validation.

Amstutz:
@iszmais : The change led to major issues. The problem is not for @nhaagen to solve. Note that the ID management including the cookies is not even part of the UI Components, but the Global Screen. Any issues with the global Screen would needed to be discussed with @chfsx . Note that the existing code seems just not to have been laid out for the new scenario. This was not on my radar when merging on monday. So I take responsibility for the merge. However, now, the issues are on my radar. With this information, the PR should not have been merged and clearly needs improvement. If your change makes further refacotrings necessary for it to work, you need to include them in your request, thx.

@nhaagen Thx for quickly helping out here, even if the problem was not in your component, nor caused by you. thx.

iszmais:
@Amstutz / @nhaagen i created a solution for the cookie compression which will secure the whole system against long id names (since this problem could also appear in other cases) and increases the performance to allow the user to create even more menu entries before cookie size transgression.

https://github.com/ILIAS-eLearning/ILIAS/pull/3332

I still would recommend changing the storage in general to localStorage or other less size-sensitive storages than cookie. But, as i understands, since its an matter of time this should provide an stable solution for all cases and would allow the Re-revert of the target feature.

PR #3332, opened 2021-04-23
Increase cookie compression by use of session storage

iszmais:
Decreases the risk too exceed the cookie size, by persisting the main content inside the local storage.

### DIFF
--- a/src/UI/templates/js/MainControls/mainbar.js
+++ b/src/UI/templates/js/MainControls/mainbar.js
@@ -517,6 +517,10 @@ il.UI.maincontrols = il.UI.maincontrols || {};
 
 (function($, mainbar) {
 	mainbar.persistence = (function($) {
+		if (sessionStorage.getItem('entry_id_map') === null) {
+			sessionStorage.setItem('entry_id_map', JSON.stringify([]));
+		}
+		let id_map = JSON.parse(sessionStorage.getItem('entry_id_map'));
 		var cs,
 			storage = function() {
 				if(cs) { return cs; }
@@ -549,8 +553,21 @@ il.UI.maincontrols = il.UI.maincontrols || {};
 			compressEntries = function(entries) {
 				var k, v, ret = {};
 				for(k in entries) {
+					let id = null;
+					let map_id;
+					for(map_id in id_map) {
+						if (id_map[map_id] === k) {
+							id = map_id;
+							break;
+						}
+					}
+					if(id === null) {
+						id_map.push(k);
+						sessionStorage.setItem('entry_id_map', JSON.stringify(id_map));
+						id = id_map.length - 1;
+					}
 					v = entries[k];
-					ret[k] = [
+					ret[id] = [
 						v.removeable ? 1:0,
 						v.engaged ? 1:0,
 						v.hidden ? 1:0
@@ -562,8 +579,8 @@ il.UI.maincontrols = il.UI.maincontrols || {};
 				var k, v, ret = {};
 				for(k in entries) {
 					v = entries[k];
-					ret[k] = {
-						"id": k,
+					ret[id_map[k]] = {
+						"id": id_map[k],
 						"removeable": !!v[0],
 						"engaged": !!v[1],
 						"hidden": !!v[2]
--- a/src/UI/templates/js/MainControls/mainbar.js
+++ b/src/UI/templates/js/MainControls/mainbar.js
@@ -517,10 +517,10 @@ il.UI.maincontrols = il.UI.maincontrols || {};
 
 (function($, mainbar) {
 	mainbar.persistence = (function($) {
-		if (sessionStorage.getItem('entry_id_map') === null) {
-			sessionStorage.setItem('entry_id_map', JSON.stringify([]));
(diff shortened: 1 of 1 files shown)


Amstutz:
Hi @Iszmais

Thx a lot for your PR to increase to cookie compression by the use of session. We believe this to be a way to make https://github.com/ILIAS-eLearning/ILIAS/pull/3278 possible, right?

In our understanding, there are two ways to handle the issue in https://github.com/ILIAS-eLearning/ILIAS/pull/3278:

1. Decrease the size of our IDs
2. Use a different storage for our IDs
----
1) Is certainly easier, however it only handles the issue partly. If you want to decrease the size of the IDs, probably contact @chfsx , since he handles the creation of the IDs in global screen. Maybe this could be a solution for the already released 6 and 7.
2) Is certainly the preferable solution. However, the way is a bit longer. Issue here is, that we do not have a global solution on how to work with localStorage. In PHP we see how painful it is to deal with legacy code that is infected with global deps everywhere. We do not want to reproduce this on the JS side. We are fully aware that option 2 has the tendency of morphing into a feature creep, however, if we really want to introduce local storage into the UI Framework, we do not see a way around it. So if you really are willing to proceed with Option 2, local storage, then this would need to meet the following minimal requirements:
    - [ ] 	A storage Component in src/UI/templates/js/Core/storage.js providing a scope and documented public interface of the local Unit tests for the localStorage to be used in ILIAS
    - [ ] 	Tests for the Public Interface (see: ilias/docs/development/js/js-unit-test.md, e.g. ilias/src/UI/templates/js/MainControls/test/mainbar.test.js as example)
    - [ ] 	Documentation of this interface in: ilias/docs/development/js/
    - [ ] 	Inject to mainbar.js

Thx for feedback on how best to proceed with this request, thx.
@Amstutz and @klees

Amstutz (changes requested):
see above

iszmais:
Hi @Amstutz ,

thanks for your response! I agree that this shouldn't be the final solution for ILIAS. I would also recommend using this solution for already released ILIAS Versions (6 and 7) since it needs no migration and is minimal inversive.
The persistence of dynamic occupancy of limited storage like the cookie, is a technical bug and could therefore be fixed with a Hotfix like this one.
The main Problem of misuse of the cookie storage is a conceptual issue and should fixed as a feature/improvement by the maintainer and therefore follow the appointed workflow of such a request inside ILIAS.

klees:
Hi @iszmais,

your argumentation regarding the nature of the problem might or might not be correct. But it misses the point.

The current implementation for the Main Bar was discussed and approved publicly over a year ago. The same goes for the implementation of the persistency layer for the state of the Main Bar. Problems with that implementation or missing requirements have been brought up and have been adressed in the development process of the community. As almost always when developing software, we became smarter in the meantime and now know more regarding requirements (the one which you brought up in #3278) and possible problems (the one uncovered by the change in #3278).

Although your requirement haven't been brought up at the appropriate time in the development process, we tried to make a solution possible for release 6. For reasons of stability we try to avoid implementation of new requirements in published release in general, but due to the nature of the problem and the minimal change we tried to include the new requirement anyway. As you know, the solution triggered the cookie problem, which you now try to solve here. The initial change in #3278, which was supposed to implement the requirement brought up late, now needs to be backed by this even larger change.

We now have a situation, where you ask us to include a solution to a problem which is caused by a requirement that we never promised to implement in a release we have already published. The solution has no tests and is not easy to understand. It also is not the solution we would want to finally want to implement. If we accept this PR, we have the risk to brake things again in a stable release and also have no final solution for the next release. Please understand that this does not look good from our perspective currently.

@Amstutz has outlined what we would need to accept a change like this. We propose to make this change in the trunk first but in general are open to patch this down to other releases if we can gain enough confidence in the change. We would be happy if you would look into implementing a solid and lasting solution to the problem at hand. As always, feel free to ask for further support or advice.

Best regards!
@Amstutz and @klees

iszmais:
Hi @klees ,
i understand the reason and appreciate the dialog. Actually this was why i tried to present a solution wich is minimal inversive.
I may have misunderstood @Amstutz i tought he meant to proclaim his approval for this solution for the already stable releases 6 and 7, too.
I approve the concept of the trunk solution and would like too implement it, however im fully booked for the next few month.
That is why i would recommend to contact the maintainer to implement those concept and maybe further approve the changes made here.

In Summary: I consent the Change requests, but I am not able to implement this in time.

klees:
Hi @iszmais,

thanks for the sympathy!

> I consent the Change requests, but I am not able to implement this in time.

This is our daily bread :-)

We are working under the [coordinator model](docs/development/maintenance-coordinator.md) here. This means that you basically have two options:

* look for other devs that have time/are willing to work on your problem or
* make time yourself

This is really a collaborative section in ILIAS where everybody has to push and help develop the UI framework to meet individual requirements. Of course we both have requirements ourselves, but as coordinators we just try our best to keep the quality standards on a good level and guide people when pushing their requirements and implementations. This of course implies that you can always ask us for input and advice how to achieve things in and with the UI framework.

So looks as if we could close this PR for the time being, right?

Best regards!

@Amstutz and @klees

PR #3333, opened 2021-04-26
UI: Filter Revision - Rules and Interface

tfamula:
Hi @Amstutz & @klees,
@ezenzen and me did the first steps for the Filter Revision. This PR includes:

- a big revision for the description and rules of the Filter to make clear the new/changed appearance and behavior of the component
- the introduction of a new method `isComplex` in the `FilterInput` Interface to differentiate between Input Fields which can be edited directly and Input Fields which need to be rendered in a Popover

At the moment, I have not spotted other necessary changes for the public interface, which have to be discussed in the Jour Fixe.

We are looking forward to your feedback!

Greetings

### DIFF
--- a/src/UI/Component/Input/Container/Factory.php
+++ b/src/UI/Component/Input/Container/Factory.php
@@ -39,33 +39,63 @@ interface Factory
      *      presented on the screen.
      *   composition: >
      *      Filters are composed of two visually separated areas:
-     *      First, there is the Filter Bar at the top. It contains an Expand/Collapse Glyph on the left side.
-     *      On the right, an "Apply" Bulky Button, a "Reset" Bulky Button and a Toggle Button for activating/deactivating
-     *      the Filter are placed. An additional row with set filter settings will be shown at the bottom of the Filter Bar,
-     *      when the Filter Bar is collapsed and the Filter is activated at the same time.
-     *      Second, there is an area where the Input Fields are displayed. Every Input Field is rendered with a Label
-     *      on the left and a "Remove" Glyph on the right. The values, that are currently entered in the Input Fields,
-     *      are displayed as non-editable text and will get editable when the user focuses an Input Field (see effect).
-     *      After the last Input Field, an "Add" Bulky Button is shown if additional Input Fields can be added.
+     *      First, there is the Filter Bar at the top. It contains an Expand/Collapse Bulky Button with the label
+     *      "Filter" on the left side. On the right, a Toggle Button for activating/deactivating the Filter is placed.
+     *      Second, there is an area which is called the Filter Content Area. It is placed below the Filter Bar. This
+     *      Filter Content Area is available in two versions depending on the view (expanded = editable,
+     *      collapsed = non-editable) of the Filter:
+     *      If the Filter is expanded, editable Input Fields are shown in this Filter Content Area, starting from the
+     *      top left. Input Fields within a Filter are rendered with its label on the left and a "Remove" Glyph on the
+     *      right. Values can be entered in the fields. After the last Input Field, an "Add" Bulky Button is shown if
+     *      currently hidden Input Fields can be added. To hide Input Fields, the "Remove" Glyph of the corresponding
+     *      field can be used. Two Bulky Buttons, one called "Apply" and another called "Reset", are placed on the left
+     *      side below the Input Fields.
+     *      If the Filter is collapsed, the labels and values of the applied Input Fields are visible as non-editable
+     *      text in the Filter Content Area. The texts of the Input Fields have a reduced size and are rendered close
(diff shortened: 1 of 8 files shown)


tfamula:
BTW: Due to the amount of text in the "composition" and "effect" sections, these segments are hard to read in the Kitchen Sink Documentation. Unfortunately, it is not possible to do some formatting (or I don't know how), e.g. with paragraphs and bullet points. Is it maybe planned to implement this in the future?

Amstutz:
Thx a lot to everybody involved. This is a big step forwards for the filter in the UI Components, thx.

@klees and @amstutz discussed the proposal of the interface today. We also asked @yvseiler for an opinion on the final wording on the concept. She might have some minor feedback on parts on the content. If so, she will propose those them until monday. 

We like the isComplex change on the interface, since it promotes the interface from a pure marker to an interface with some meaningful semantics. Note: "isComplex" might even be an aspect that could be interesting for input fields independently from filters in the long run.

We are looking forward to discuss the change in the upcoming JF.

Amstutz (approved):

yvseiler (commented):
Hi @tfamula 

It is great to see how the filter concept already discussed earlier by the JF is now being implemented. Thank you very much for your cooperation and thinking through the whole revision! The filters will thus receive a great upgrade in use.

Please consider the following suggestions:
- [x] **Resolve duplications**: To shorten the text a little more, I suggest deleting the two sentences [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3333/files#diff-132413157cba32aac68b271318a8ddc8affc411ef496624b745aa3ca29da5351R56-R58); "Only the Input Fields... are not visible in this area." because this is already introduced in the sentence line 53.
- [ ] **Always clickable buttons**: It is written [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3333/files#diff-132413157cba32aac68b271318a8ddc8affc411ef496624b745aa3ca29da5351R115-R117) that the buttons (Apply and Reset) should be clickable, whether it has an effect on the filter or the item collection or not. The consequence of this is that it is not made visible to a user whether his new filter values are already filtered or not. What was the reason for deciding that the buttons should be clickable in any case? Possibly certain decisions led to this decision, which I cannot (any more) reproduce at the moment. I think it would be good to think about whether it might not make sense to show the user whether his last changes to the filter have already been applied or not (f.e. in which the buttons are only clickable if they also result in a change).

Best regards,
@yvseiler

src/UI/Component/Input/Container/Factory.php
```diff
@@ -39,57 +39,122 @@ public function form();
...
+     *      out. If the Filter is collapsed and the values for all Input Fields are removed, the Filter Content Area
+     *      will disappear. If the Filter is expanded, the Input Fields stay editable in the deactivated state of the
+     *      Filter.
+     *      The Expand/Collapse Bulky Button in the Filter Bar can be used to show/hide the Filter Content Area if the
+     *      Filter is empty. If there are applied values, it can be used to change the Filter Content Area between the
+     *      editable and not-editable version. The varying rendering of this area is described in "composition".
+     *      In the Filter Content Area, the Apply Button and Reset Button can be used. Clicking on the Apply Button
+     *      has multiple effects: The appearance of the Input Fields, i.e. if an Input Field is shown or hidden in the
+     *      Filter Content Area, will be saved. The values, which are entered into the Input Fields, are applied. The
+     *      content of the item colletion is reloaded immediately according to the Filter settings. The status of the
```
yvseiler:
```suggestion
     *      content of the item collection is reloaded immediately according to the Filter settings. The status of the
```

matthiaskunkel:
Jour Fixe, 03 MAY 2021 : We highly appreciate the suggested rules and interface and accept the PR for trunk.

tfamula:
Hello all!

I adjusted the rules according to Yvonne's comments.
I also implemented the `isComplex` method for the Filter Inputs, so that the Unit Tests will not fail anymore (It seems the Tag Input was accidentally a Filter Input, because there is no rendering for it in the `FilterContextRenderer` and you get an error when using it in the Filter).

As we already mentioned in the Jour Fixe, the concept of the "not always clickable buttons" will not be considered in this step of the Filter Revision. It needs an extra conceptual discussion and should definitively be implemented/discussed in a separate PR. IMO, the concept can also be liable for bugs/issues. For now, we do not see real disadvantages when the Apply Button and Reset Button are always clickable, since this is usual behavior for filters on other platforms, too.

I would appreciate it if this PR could be merged. 
Next week, I will create a separate PR for the implementation part of the the improved Filter.


Greetings
Thomas

Amstutz:
Thx a lot for this first part towards better filters! Thx to everybody involved. Looking forward to the implementation.

tfamula:
Implementation part: #3390

PR #3334, opened 2021-04-26
UI Components, Custom Aria Label for Glyphs: See #28930

Amstutz:
This PR proposes a fix for: https://mantis.ilias.de/view.php?id=28930

Main insight from the report is, that glyph need to be a bit more flexible in the way devs can set aria-labels. There are situation, where the label needs to have a more context driven label than the one that was derived as default from the purpose description.


klees (commented):
Hi everybody,

@Amstutz and I discussed this and we came to the conclusion that we would want to add that `withAriaLabel` only internally. For now we assume, that the correct label for a glyph may always be derived from the context in the UI components. That means that there would be now need for any user of the UI framework to set said label. If cases come up that challenge this assumption we are happy to hear from them.

In the UI framework we try to encode as much rules as possible in code (instead of in prose guidelines) to allow the users of the framework to think about their use case instead of the UI. Hopefully we continue that program for the accessibility requirements =)

Best regards!

PR #3335, opened 2021-04-26
Less, fix less as development dependency issue, see: #28609

Amstutz:
I was asked from various sides to add the used lessc version as npm development dependencies. Note that I also slightly updated the bower readme.

If I understood the readme right, then we currently "Commit the changes in packages.json and package-lock.json and node_modules." also for deps dependencies, (see /docs/development/js/third-party-libs-npm.md). That results in lots of files added. I hope this is valid, if not, ping me, and ill change.


kergomard (approved):
Thank you very much @Amstutz for adding this! We might want to find a cleaner solution for this kind of dependencies in the future, but right now this complies with our guidelines.

PR #3338, opened 2021-04-28
Add Icon for WebDAV

kergomard:
We have a new icon for WebDAV in the administration.


Amstutz:
thx a lot! pushed to trunk as well.

PR #3347, opened 2021-04-30
about refactorings

chfsx:
The Technical Board proposes to include the following rules and tips on refactorings as documentation.


matthiaskunkel:
Jour Fixe, 03 MAY 2021: We discussed the original suggestion, made a slight change concerning the manual steps a developer should do after refactorings and accept the current version for trunk.

PR #3354, opened 2021-05-03
Auth/Shib.: Fix Mantis issue 28913 (error after selecting Shib. as de…

mjansenDatabay:
…fault auth. mode)

Mantis Issue: https://mantis.ilias.de/view.php?id=28913


smeyer-ilias:
Thanks for fixing this.

PR #3359, opened 2021-05-03
Auth/SOAP: Remove dead code for user selection in SOAP auth. (Mantis …

mjansenDatabay:
…issue 13832)

Mantis Issue https://mantis.ilias.de/view.php?id=13832


smeyer-ilias:
Thanks for fixing this.

PR #3362, opened 2021-05-03
Fix pass null to ilOrgUnitUser (ILIAS 8)

ghost:
https://mantis.ilias.de/view.php?id=28485


ghost:
This PR is not the reason the tests are fail (Something with composer and PHP 8)

PR #3363, opened 2021-05-03
UI/Mainbar, Accessibility: render as list, focus sub-slates

nhaagen:
As discussed in workshops earlier this year, the mainbar will heavily profit from auto-focussing certain elements.
Also, in terms of a11y, the "main menu",  need to be rendered as list/items.

This PR will:
- render mainbar top entries as list-items within an ordered list. (There's a roadmap-entry for removig the div-wrappers, too)
- add roles menubar/menuitem to those entries
- automativally focus subslates of top-entries
- re-focus the top-entry after closing all slates using the according button (on the bottom of slates)
- include the examples from the workshops

### DIFF
--- a/Modules/StudyProgramme/classes/class.ilStudyProgrammeDashboardViewGUI.php
+++ b/Modules/StudyProgramme/classes/class.ilStudyProgrammeDashboardViewGUI.php
@@ -84,7 +84,7 @@ class ilStudyProgrammeDashboardViewGUI
             krsort($assignments);
             /** @var ilStudyProgrammeAssignment $current */
             $current = current($assignments);
-            if (!$this->isReadable($current)) {
+            if (true || !$this->isReadable($current)) {
                 continue;
             }
 
--- a/Modules/StudyProgramme/classes/class.ilStudyProgrammeDashboardViewGUI.php
+++ b/Modules/StudyProgramme/classes/class.ilStudyProgrammeDashboardViewGUI.php
@@ -84,7 +84,7 @@ class ilStudyProgrammeDashboardViewGUI
             krsort($assignments);
             /** @var ilStudyProgrammeAssignment $current */
             $current = current($assignments);
-            if (true || !$this->isReadable($current)) {
+            if (!$this->isReadable($current)) {
                 continue;
             }
 
--- a/src/UI/Component/MainControls/Factory.php
+++ b/src/UI/Component/MainControls/Factory.php
@@ -206,6 +206,7 @@ interface Factory
      *     5: Bulky Buttons in the Main Bar MUST bear the "aria-haspopup" attribute.
      *     6: Bulky Buttons in the Main Bar MUST bear the ARIA role "menuitem".
      *     7: Slates in the Main Bar MUST bear the ARIA role "menu".
+     *     8: Top-Level entries of the Main Bar MUST be rendered as a listitems.
      * ----
      * @return  \ILIAS\UI\Component\MainControls\MainBar
      */
--- a/src/UI/ROADMAP.md
+++ b/src/UI/ROADMAP.md
@@ -234,8 +234,13 @@ The UI Inputs do not all have a rendering test.
 Add, where missing, and refine existing.
 
 ### Make date/time input accessible (advanced)
-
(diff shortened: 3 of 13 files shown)


Amstutz:
Hi @nhaagen

Thx a lot for the implementation of the prototype!

There are some questions:

- [ ] Region Role: Why do you set and unset the region role per JS? Is this not always a region? Important: This is really just a question, not suggestion for change ;-).

- [ ] Ordere List: Why did you use an ordered list, and not an unordered? I used unordered for the Meta Bar, just because ul is used in the examples menubar in the wcag examples (I can see no specific reason for that and did not much reasoning for this, see: https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html). If there are not strong reason for ol, I vote just be safe an use ul. Not to run in some validation issues for this detail.
- [ ] Position of MenuItem: In https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html it is noted that the menuitem role „none“ should be set for the li element since "the parent ul is serving as a menu so the li elements are not in their required list context.“ The Role menuitem seems to need to got into the bulky buttons. 
- [ ] Tools: Tools seem not to be tackled by the change to set focus inside the Slate on activation. Is this intended, error of mine while testing or just an issue?
- [ ] Examples: Thx a lot for the added examples! IMO a great asset. However, the old example in Main Controls -> Main Bar example are not completely workable. In my case, I can not open slates. Since we have the nice example included in Standard Page Layout, we could imo just point to those in the Main Bar component.

Please Change

- [ ] Tests: Fix PHP and js Unit tests.


Thx a lot for this important step!

Kindly
Timon

Amstutz (changes requested):
See above

nhaagen:
Hi @Amstutz ,
thanks for looking at this so thouroughly.

- [x]  Region Role: Why do you set and unset the region role per JS? 
Looking at the accordion, I understood the entry for role=region to state that only the "currently expanded" panel should bear the attribute. Maybe I'm wrong, though, and it should rather be switched than removed?
https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html

- [x] Ordered List: Why did you use an ordered list, and not an unordered? 
We undertake quite an effort to sort and order menuitems; I thought the ordered list to correspond to this the most.
However, I have no issues with that at all - ul for anticipated compatibility it is.

- [x] Position of MenuItem: the menuitem role „none“ should be set for the li element , the Role menuitem seems to need to got into the bulky buttons.
Thank you for pointing that out.

- [x] Tools: Tools seem not to be tackled by the change to set focus inside the Slate on activation. 
Yes. I fixed that.

- [x] Examples: the old example in Main Controls -> Main Bar example are not completely workable. 
done. Good idea!

Amstutz:
> Looking at the accordion, I understood the entry for role=region to state that only the "currently expanded" panel should bear the attribute. 

Thx for the Info! 

Tests seem to run as well. There seems some issue unrelated to you changes. Is this ready for some final testing and if fine merging?

klees:
Hi @Amstutz, hi @nhaagen,

we had some issues with the cs-fixer, they should be solved with #3381. If @nhaagen would rebase this to the current trunk, at least the tests for PHP 7.4 should run. The tests for PHP 8 should show some unrelated fails.

Best regards!

Amstutz:
Hi @nhaagen 

Thx a lot! We are currently performing some final tests (on a GUI) level, to make sure this all runs fine also on 7 (since this will be picked down to 7 as well). If all is good, ill merge. 

One minor thing popped up after the tests run again; The example fails the PHP 8 tests (Undefined array key "ui_mainbar" in the smoke tests, see here: https://github.com/ILIAS-eLearning/ILIAS/pull/3363/files#diff-337fc47f53023509a82bbac420adb1d91f56bd66ab36f4db3cf30236305a9ac5R186). There are other tests failing 8, however since this would be the only test failing from the UI Framework for 8, a fix would be neat. However, I can also contribute the fix immediately after merge. 

In the near future we will have a passing set of PHP 8 Unit tests, after that those will be a lot easier to spot.

Thx!

PR #3379, opened 2021-05-04
iass: Fix PHP 8 issues for (Unit Tests)

mjansenDatabay:
This PR fixes required parameters after optional parameters.
Furthermore passing a function result directly as an
argument (by ref) is not valid, so this is fixed as well.


klees:
Hi @mjansenDatabay,

thanks a lot, I have fixed this a little differently:

https://github.com/ILIAS-eLearning/ILIAS/commit/efb41480d0d951abb305c5a6bd2034477ae7555d

Best regards!

PR #3381, opened 2021-05-04
GitHubActions: Add possible fix for php-cs-fixer version when executi…

mjansenDatabay:
…ng GitHub actions

This PR fixes an issue with the latest v. 3.0 release of `php-cs-fixer`, which is installed by the `GitHub Actions` step `Install PHP CS Fixer` and causes our builds to fail (even for PHP 7.4).

```bash
Run composer require friendsofphp/php-cs-fixer --no-interaction --no-progress
Using version ^3.0 for friendsofphp/php-cs-fixer
./composer.json has been updated
Running composer update friendsofphp/php-cs-fixer
Gathering patches for root package.
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires friendsofphp/php-cs-fixer ^3.0 -> satisfiable by friendsofphp/php-cs-fixer[v3.0.0].
    - friendsofphp/php-cs-fixer v3.0.0 requires php-cs-fixer/diff ^2.0 -> found php-cs-fixer/diff[v2.0.0, v2.0.1, v2.0.2] but the package is fixed to v1.3.1 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.
Error: Your requirements could not be resolved to an installable set of packages.
```

CC @lauraquellmalz

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -30,7 +30,7 @@ jobs:
           GHRUN: "yes"
 
       - name: Install PHP CS Fixer
-        run: composer require friendsofphp/php-cs-fixer --no-interaction --no-progress
+        run: composer require "friendsofphp/php-cs-fixer:^2.19" --no-interaction --no-progress
 
       - name: PHP CS Fixer
         run: CI/PHP-CS-Fixer/run_check.sh
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -29,9 +29,6 @@ jobs:
         env:
           GHRUN: "yes"
 
-      - name: Install PHP CS Fixer
-        run: composer require "friendsofphp/php-cs-fixer:^2.19" --no-interaction --no-progress
-
       - name: PHP CS Fixer
         run: CI/PHP-CS-Fixer/run_check.sh
         env:


klees:
@mjansenDatabay Thanks for looking into this.

IMO we don't need that "Install PHP CS Fixer" step at all. We require the cs-fixer ourselves and should have already installed it in "Install dependencies". This will also make local results and CI results more consistent because we use the same software everywhere.

@lauraquellmalz @mjansenDatabay Any opinions?

mjansenDatabay:
> @mjansenDatabay Thanks for looking into this.
> 
> IMO we don't need that "Install PHP CS Fixer" step at all. We require the cs-fixer ourselves and should have already installed it in "Install dependencies". This will also make local results and CI results more consistent because we use the same software everywhere.
> 
> @lauraquellmalz @mjansenDatabay Any opinions?

@klees I agree. `php-cs-fixer` should be already installed by the previous step `Install dependencies`.

kergomard:
Just broke the tests on ilias6 and ilias7 by merging the composer update PRs, too. We would need to cherry-pick to 6 and 7, right. ...otherwise I will revert the breaking changes in the composer PRs.
Sorry for not making the connection!

catlauraherzog:
I checked and it's okay to remove the explicit requirement for php-cs-fixer in github actions. I can approve this PR.

klees:
I'll pick this down to the other branches.

kergomard:
Thx very much @klees , @lauraquellmalz !

klees:
This were @mjansenDatabay|s doings mostly! =)

PR #3384, opened 2021-05-07
Who is online, fix of #30702

Amstutz:
See mantis issue https://mantis.ilias.de/view.php?id=30702


Amstutz:
Sry, just discussed this with @oliversamoila, I got the setting wrong. sry.

PR #3385, opened 2021-05-07
Mail: Improve background task handling

mjansenDatabay:
The `Mail` component uses `Background Tasks` internally to process the incoming email delivery tasks (according to the JourFixe decision in 2019). But there could be issues when using `Background Tasks`  with many parameters, like documented in:

- https://mantis.ilias.de/view.php?id=28515
- https://mantis.ilias.de/view.php?id=26888
- https://mantis.ilias.de/view.php?id=24961

These issues are hard (read: impossible) to reproduce and the reason could not be found until now.

Currently the `Mail` component passes all it's parameters (to, cc, bcc, ...) to the `Bucket`/`Task`.

This PR introduces a new  `Background Task` used by the `Mail` component which only passes a `Uuid`. All related data is stored by the `Mail` component in a new database table before running the `Task Manager`. When the concrete `Background Task` gets executed, the mail properties are fetched from database and processed. Afterwards, the processed record will be deleted.

### DIFF
--- a/Modules/Forum/classes/Notification/class.ilForumMailEventNotificationSender.php
+++ b/Modules/Forum/classes/Notification/class.ilForumMailEventNotificationSender.php
@@ -417,7 +417,6 @@ class ilForumMailEventNotificationSender extends ilMailNotification
         $bodyText .= $attachmentText;
 
         $mailObject = new ilMailValueObject(
-            '',
             ilObjUser::_lookupLogin($recipientUserId),
             '',
             '',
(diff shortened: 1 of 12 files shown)


chfsx:
The PR is marked as Draft, are you still working on it? thanks for clarifying the state of this PR.

Best regards!

mjansenDatabay:
@chfsx I guess this is not needed anymore since you latest `sorting/ordering` fixes in the `BackgroundTasks` component.

PR #3389, opened 2021-05-11
DIC: Fix wrong return type for global page template

mjansenDatabay:


matthiaskunkel:
@ILIAS-eLearning/technical-board Just a check...

klees:
Thx!

PR #3390, opened 2021-05-11
UI: Filter Revision - Implementation

tfamula:
Hello everybody!
This is the implementation part for the Filter Revision! Feedback is very appreciated.

There a few few things, which have to be done during/after discussing this PR:

- [x] Check and fix unit tests
- [x] ~~Create a separate PR for the ON/OFF Label for the Toggle Button and add the Label to the Filter~~ (see #3392)
- [x] Check code where the Filter is consumed and adjust it if needed, especially in view of the `getData()` function (the usages are manageable, I detected 4 usages)

Here are some screenshots from the KS example for a first impression:
Filter_activated_expanded
Filter_deactivated_collapsed


klees:
I'm really looking forward to review, merge and then use this =) Yiah!

yvseiler (commented):
Thanks a lot @tfamula  the adjustments and the clarification. I am looking forward to seeing the filter in action!

tfamula:
Hello!
I have fixed the unit tests now. By doing this, I stumbled over strange errors in the Filter render tests which I could not assign to any of my changes in the implementation. After a while, I figured out that the real issue was in the Select Input. That's why I made small changes for this Input to match HTML semantics and also added rendering tests for it which were missing before, especially to avoid such problems in the future.
Is it okay if these changes are in the same PR as the Filter to understand the context? Or should I make an extra PR?

Services/UI/classes/class.ilUIFilterService.php
```diff
@@ -162,12 +173,13 @@ protected function writeFilterStatusToSession($filter_id, $inputs)
         }
 
         if ($this->request->getFilterCmd() == self::CMD_COLLAPSE) {
-            $this->handleRendering($filter_id, $inputs);
             $this->session->writeExpanded($filter_id, false);
         }
 
         if ($this->request->getFilterCmd() == self::CMD_APPLY) {
             $this->handleRendering($filter_id, $inputs);
+            // always acitvate the filter when it is applied
```
klees:
```suggestion
            // always activate the filter when it is applied
```

klees (changes requested):
Hi @tfamula,

thanks a lot for improving this important UI component. Please note, that we did not review the snippets in `Services/UI`, as we do not feel that we are responsible there.

Regarding the changes in the UI framework we found some room for improvements. The most of the stuff looks fine, though. Special thanks for making this accessible by keyboard. We performed some basic checks in this regard and it seem to work. Also kudos for throwing in strict typing and other improvements concerning the code style. We like that a lot. 

- [x] Check 1: Why did you remove [this check](https://github.com/ILIAS-eLearning/ILIAS/pull/3390/files#diff-b446ce9273435c5606ce7fc83a0ffe0cbeaec335c0a3f8ca3854005d284776bbL135)?
- [x] A lot of added lines: We hoped a little hat we would lose more lines than add, since the concept seems a simplification. However, we could not point the finger to unnecessary lines ourselves. Do you have some hints, on why the new concept needs more and not less lines? Is this still a simplification?


- [x] Less 1: Can we derive the [border-color here](https://github.com/ILIAS-eLearning/ILIAS/pull/3390/files#diff-596f8b5902f856b7544d5a57d929e0d7a80dc52ef671ec11f225fccb82392150R445) from another color? Maybe check this with Enrico. Same for the Font Color below. 

- [x] Less 2: There are many changes in the filter.less. Most of it look like formatting changes, it is really hard to find the real changes here.  Do you use some auto-formatting for less? Please undo the whitespace changes so the PR contains the real changes only.
- [x] filter.js 1: Simplify filter.js. This PR is supposed to add a huge if-else block, separated by comments. The statements in the blocks look partially similar. This looks like a really good target for some cleanup: introduce methods and brake that huge block apart.
- [x] filter.js 2: Introduce some unit testing for the code. The methods from the previous change should be a good target for some tests. At least try to introduce an infrastructure for the tests, so we can add tests incrementally afterwards.
- [x] a11y, Labels: Wave gives us errors for the inputs, since they are missing a label. Note that all inputs need to carry a label, so the screenreader can properly communicate what an input is operating on. Add the labels. Best, checkout on how context_form.html is doing this (you can just reuse the Label field there and connect to the input by ID).
- [x] Multi Select Padding: There is not much padding on left and right for the multi select when displayed in the dropdown. This looks somewhat broken. Can you add some padding:

Regarding the formatting of the less-file, we might want to have a look into some shared guideline for less-format, if you are interested. Feel free to get back to us.

Best regards!
@Amstutz and @klees

klees:
Hi @tfamula,

is it correct that this PR is currently on your turf or are you waiting for any action by @Amstutz or me?

Best regards!

tfamula:
Hi @klees,
yes, this is on my radar. Probably, I will get round to it at the beginning of next month. No action of you or Timon needed at the moment, thanks!

Amstutz:
Hi @tfamula 

Any update on this? We anxiously wait for those improvements ;-).

Thx

tfamula:
Hi @Amstutz and @klees!

Thank you for your patience. I finally found time to get back to the Filter. And thank you very much for your feedback which is, as always, very constructive.

> Check 1: Why did you remove this check?Check 1: Why did you remove this check?

I found out that this check was misplaced and therefore had no effect. I moved it to the right place in [`ilUIFilterService`](https://github.com/ILIAS-eLearning/ILIAS/pull/3390/files#diff-46250b460210ff5c1c95a86b83bf5de22c3a26bce1633b51c5cbcd890e69943bR95).

> A lot of added lines: We hoped a little hat we would lose more lines than add, since the concept seems a simplification...

In my understanding, the concept should be a simplification for the user. The Filter should be easier to understand and to use. This has been achieved IMO. Just like you, I checked the whole component if there is maybe some code which becomes superfluous, but I have found almost nothing. The added lines come mainly from some bugfixes I detected in connection with realizing the concept.

> Less 1: Can we derive the border-color here from another color? Maybe check this with Enrico. Same for the Font Color below.

Enrico and I checked the two colors. We replaced the border-color. For the font-color, that was not possible because we do not have a color we can derive from. In our eyes, we need a color set for things in a disabled state (@il-font-color-disabled, @il-background-color-disabled, @il-border-color-disabled). Something like that exist for inactive buttons, but we need it more general. Enrico will discuss this in the next meeting of the CSS Revision. Until we have a decission here, I will leave the font-color in the Filter as it is.

> Less 2: There are many changes in the filter.less. Most of it look like formatting changes, it is really hard to find the...

I reverted the formatting changes. The format looked wrong to me, so I reformatted the less file by PhpStorm using the `php-storm.xml` (ILIAS Coding Style 4/2020). I thought it also covers formatting for less, because it is different to the Default coding style from PhpStorm. But having a closer look into the xml file shows me that this is not the case. Do we even have a guideline for less-format? What do you mean by having "a look into some shared guideline for less-format"? You've got me stumped here... However, this should not prevent us merging the PR.

> filter.js 1: Simplify filter.js. This PR is supposed to add a huge if-else block, separated by comments. The statements in the...

I moved parts of the huge block into one separate method and simplified the code a little bit. However, there still exists if-else-statements, because some Inputs must be handled in a different way than other Inputs. I have not found a more elegant way to check for the type of the Input.

> filter.js 2: Introduce some unit testing for the code. The methods from the previous change should be a good target for some...

I introduced the desired infrastructure for JS tests. However, if we want to add JS rendering tests in the future (and the whole JS file for the Filter is about rendering), we maybe need an extra library like [jsdom](https://github.com/jsdom/jsdom) to implement that.

> a11y, Labels: Wave gives us errors for the inputs, since they are missing a label. Note that all inputs need to carry a label,...

I decided to completely refactor the `FilterContextRenderer` based on the `Renderer` for the Form Inputs, which was revised by Nils. That makes it much easier to understand the differences between these two files and also to make changes in the future. In this way, I added the labels for the inputs, too. 
But for `ProxyFilterField`s, which are rendered for "complex" inputs (`Multiselect`), I have not added a connection to the label. This would be wrong in my eyes. Visually, it looks like an input, but semantically, it is not. I also thought about connecting the label with the input in the Popover. But I am not sure about this because the input is not displayed until the user performs a click event on the `ProxyFilterField`. Additionally, I figured out that our `Multiselect` Input has no connection to the label in Forms. I guess the reason is that we handle it semantically as a list. So maybe the question becomes superfluous for the Filter, too.

> Multi Select Padding: There is not much padding on left and right for the multi select when displayed in the dropdown. This...

Done!


Greetings
Thomas

Amstutz:
Hi @tfamula,

thanks a lot for iterating on this and providing some answers to our questions. The filter work a lot better now. Kudos.

* Regarding the filter.js: This looks a lot cleaner now, thanks. There is no need to make more changes to get this merged. Still, we want to comment on some of your remarks. You mention, that you did not find a more elegant approach to deal with the many `if`s to discriminate behaviour regarding specific entities. A typical strategy would be to use polymorphism and implement the varying behaviour as an interface with different implementations on different objects. This is not something we use in ILIAS' javascript often (or at all) currently, though. Also you mention that we may need some library to test js-rendering. This seems to be correct in general. Still, we can greatly reduce the requirement for such a library if the js implementation is structured in a certain way. This is done in the js of the Main Bar to the best of our knowledge. The general strategy is to break that typical js-dom cycle apart. A typical js implementation would be to implement event handlers, which then make the appropriate changes to the dom. This can be stretched to a cycle like this: event handlers in dom trigger internal events (1). The internal events cause changes on an internal representation of the component structure (2). That representation is then mapped to the DOM (3). The component logic then is contained in (2) and the DOM-integration layers (1) and (3) can be rather slim then. More importantly, (2) can be tested without any reference to DOM. This is not something we invented ourselves though, but rather is the gist of models like react or FRP.
* Regarding the less-formatting: There is a [less guideline](https://github.com/ILIAS-eLearning/ILIAS/tree/release_7/templates#less-guideline) but it does not say anything about formatting. Formatting for less is something where, unfortunately, there currently is no guidline or (better) tooling for the community. Feel free to pick up that slack and propose something. We will be happy to support you.
* Regarding the labels for complex inputs: We are not sure if we follow your argumentation there, but this isn't something to be discussed in this PR. Good move to stick to the general implementation of the input more closely, but currently the copy is copied, not reused, right?
* We observed some minor issue with the length of the span in the preview if there is a huge amount of content placed inside (as the case e.g. in the Plugin Table in the administration). If you can fix this, after merge, would be appreciated. We do not require it here, since it is not strictly related with your change.

One final question, we seem to recall, that this is also supposed to be merged down to 7, correct? If so, since this is rather large and 7 is already in production, can you create a second PR for 7 and check if we are running into any issues with the filters already in place?

Best regards!
@klees and @Amstutz

PR #3391, opened 2021-05-12
Authentication 30728: add OpenID-Connect as possible valid mode

daniwe4:
https://mantis.ilias.de/view.php?id=30728


smeyer-ilias:
Thanks for fixing this.

PR #3392, opened 2021-05-12
UI: On/Off Label for Toggle Button

tfamula:
Hello,
this PR adds an optional On/Off Label inside the Toggle Button, which is needed for the improved Filter, see #3390.

We have to discuss about the question if the Label has to be translated or not. Because if the answer is yes, it is IMO impossible to render the label inside the button due to the different length of "on" and "off" in other languages. If found a [Mantis ticket](https://mantis.ilias.de/view.php?id=27551) where the question about multilingualism already came up one year ago, but no answer was found.
Toggle_on
Toggle_off


Amstutz:
Hi @tfamula,

thanks a lot for looking into this! The change looks good in general.

From our perspective, we can use on/off as label in any language, the wording seems to be international by now. However, finally this is up to @mkunkel to decide, since he maintains the translations. In German, it would not be an issue "An"/"Aus". However, sure there are languages that might need to fall back on "on"/"off". Lets decide this on the JF. We have no strong opinion on this.

- [x] Non-Optional: Make the on/off-label non-optional. Just use it for every toggle button.

That's it =)

Best regards!

@Amstutz and @klees

matthiaskunkel:
I fully agree to keep ON | OFF for all languages but would like to add both labels to the lang files nevertheless. @tfamula : you could add a comment to the related lang file entries like "### Don't translate this label to prevent rendering problems of the related toggle buttons!" This helps to avoid a translation by accident.

matthiaskunkel:
Jour Fixe, 31 MAY 2021 : We highly appreciate this suggestion and accept the PR for trunk. Please change the rules according to Timon's comment (The Toggle Button MUST contain a label inside, representing its current status (ON/OFF).) and introduce two language entries for both status according to Matthias' suggestion.

tfamula:
I made the label non-optional and added the language variables.
Please note, that no ON/OFF label will be displayed for a Toggle Button `withUnavailableAction()`. It seemed wrong to me to show the label here, because the label would give the impression that the button has an effect although it has not, especially in the ON-state.
If you agree with everything, I would be happy about a click on "Merge" ;)

Amstutz:
thx a lot!

PR #3397, opened 2021-05-18
AllComponents/Database: Use 'integer' with length '8' (results in MyS…

mjansenDatabay:
…QL BIGINT(20)) for unix timestamps because of https://en.wikipedia.org/wiki/Year_2038_problem / https://konstantinmb.medium.com/the-unix-timestamp-ticking-time-bomb-navigating-the-2038-challenge-99879dca47a1

Mantis Issue: https://mantis.ilias.de/view.php?id=28758


matthiaskunkel:
Jour Fixe, 31 MAY 2021 : We highly appreciate this suggestion in general. But to make a final decision about this PR we need some more information about the impact and the reason for this PR. Is there already an installation where this change has been introduced successfully? How could a review be organised and made by whom?

PR #3400, opened 2021-05-19
UI/DrillDown - Replace numbers with titles in example

nhaagen:
For better readability and referencabilty I aligned the titles with the example from Page/Mainbar.


klees (approved):

klees:
I would support merging this to release_7, as suggested by @nhaagen. We will directly benefit current development work (because people can use the examples on test7.ilias.de better) and the risk is minimal. Would then pick to trunk, of course.

Amstutz:
Thx a lot. Picking to trunk as well.

PR #3401, opened 2021-05-20
LearningSequence: #30446 - hide curriculum-headline if empty

nhaagen:
https://mantis.ilias.de/view.php?id=30446


klees:
Cherry-picked to trunk and release_6.

PR #3403, opened 2021-05-20
BT-30716: Added title to passwordassistance

M-Barz:
Link to Mantis-Report:
https://mantis.ilias.de/view.php?id=30716


smeyer-ilias (approved):
Thanks for fixing this.

PR #3408, opened 2021-05-25
Modules forum double underscore function name

mBeym:
Removed double underscore from functions inside **Modules/Forum/** to conform with PHP reserving the double underscore prefix.

- Also changed name of the **checkPath()** functions to prevent them from overwriting the parent function (and requiring the **$a_path** parameter to be added).
- Made previously prefixed functions private.


mjansenDatabay:
Thx @mBeym

PR #3409, opened 2021-05-25
UI-Framework: load JavaScript at the end of the page

lukastocker:
by @klees: This will load the javascript after the actual content of the page instead of before. This currently seems to be _the way_ (tm) how things are done in javascript land. It has the advantage, that content will be rendered by the browser earlier, as far as I understand. This also will allow @Hadrian-II to introduce a certain markdown editing library, that insisted on being loaded after the content. Also, this was the problem that actually spawned this change.


Amstutz:
As far as I can see, there are no side effects. Merged to trunk only.

PR #3410, opened 2021-05-25
CSS squad Contrast of the focus outline

Amstutz:
Fix for: https://mantis.ilias.de/view.php?id=30632

All credits to @ezenzen. Kudos.


Amstutz:
Looks good to me, thx @enrico!

PR #3411, opened 2021-05-26
UI Components: Interface for Toasts

iszmais:
This PR proposes to introduce the UI-Component for implementing [Toasts](https://docu.ilias.de/goto_docu_wiki_wpage_6494_1357.html)
**Toast:**
The Toast is an unpersistent, temporary UI Component to trigger the users awareness without affecting the UI permanently.

A Toast can appear in the top right corner, overlaying the ILIAS UI and contains a title and a close glyph. It can contain a description, which is shown below the title, and a number of actions as links, which are shown below the description or, if its not given, below the title. Further it can contain 2 other kinds of actions: an action to trigger when the user clicks the title of the Toast and an action which triggers when the toast vanished per default. IT vanishes after an user-interaction with the title or the close glyph of the toast. If no interaction take place, the toast vanishes after a certain amount of time, which can be global configured for all toasts. There are 4 interactions with the Toast:
- If the user clicks the close glyph of a toast, the Toast **MUST** vanish and no other actions of the Toast are called.
- If the user clicks the title of the toast, the Toast **MUST** execute its title action, if it is set.
- If the user clicks one of the optional actions inside, this action **MUST** be executed.
- If the user does not interact with the Toast and the Toast vanishes it **MUST** execute its vanishing action, if it is set.

_Node that the toast is consciously decided to be a independent component and not to be an heritage of the Item component, since this component expects a special kind of optional actions (Standard/Dropdown) which would violate the concept of the toast._


klees (changes requested):
Hi @iszmais,

thanks a lot for this elaborate proposal. One can clearly see the effort you have put into this.

 - [ ] OSD Notification: We are not sure, if we understand the rivals section. Will OSD Notifications be removed once this PR is finished and the usages transferred to the new UI Components? Can you sketch out somewhat, of how we could proceed with the OSD Notifications?
- [ ] Usage Rule: The first usage rule as rather harsh and a bit unspecific. Can you give a few examples of when to use and when not to use them? We might want to reduce the requirement level to SHOULD. Also, have a look into the rivals, mentioned below in the change requests. Will this rule really lead to the desired differentiation between the various messaging or notification components?
- [ ] Interaction Rule: Why is the first only SHOULD and not MUST. Does this even need to be a rule?
- [ ] Accessibility I: How would screenreaders be informed about a `Toast`?
- [ ] Accessibility II: How and when do `Toast`s get focus by keyboard?
- [ ] Timer: Is the timer set globally and will be the same for all toasts? Where and how will it be set?
- [ ] Vanish Action: What is the intended use case for that action?
- [ ] Link Items: Your descriptions reference "number of link items". Are these "Links" in the sense of the UI framework? Maybe rephrase accordingly.
- [ ] Non-Standard Toast: Do you expect another type of `Toast` (a non-standard one) to be necessary some day?

- [ ] Composition Rule: Are there other Notifications? What would they be? Are these the rivals, mentioned below? What does this rule mean? Probably, it should be rephrased.
- [ ] Design Detail I: Probably remove "The item will be displayed overlaying the main content in the top right corner.“ This is a rather high level of detail in this section and might change if the design is changed.
- [ ] Design Detail II: "The Toast SHOULD be limited in space to less than width of 500px, no matter the size of its content.“ is also very specific. Maybe just leave out or reference some other width to get rid of the fixed amount of pixels at least.
- [ ] withoutActions: Maybe replace with `withActions`. To remove actions one could simply pass an empty array.
- [ ] Close Action: It seems as if we will require a "close" action technically, to notify the system when the user pressed the close button.
- [ ] Rule Contradiction: It seems as if accessibility.1 and responsiveness.2 contradict each other. Probably one of the rules (responsiveness.2 probably) need to be changed or removed.
- [ ] Title: Consider using `Shy Button`, `Shy Link` or `string` as title. The action on clicking any of these would be implied, since these already are standard components with the intended behavior. We could get rid of two methods (regarding title actions) and probably also some text in your descriptions.

- [ ] Rivals Section I: Please use the format where the rivals are listed and then the rivalry is described. Have a look into the [Sub Panel](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/UI/Component/Panel/Factory.php#L53), for example.
- [ ] Rivals Section II: Add `Message Box` and `System Info` as rivals and point out the differences to the `Toast`. For extra points, add the `Toast` at the two components as rival as well.
- [ ] Typo I: … affecting the user experience permanent*ly*
- [ ] Typo II: … which is trigger*ed* when the user clicks 
- [ ] Typo III: … which is trigger*ed when the the Toast vanishe*s
- [ ] Typo IV: … Therefore it MUS*T*
- [ ] Wording: Unify "Toast" und "Toast Item" to "Toast" in your descriptions.
- [ ] Return Types I: Return types need to be completely specified with full namespace to work properly. This is only the case in the factory of the components.
- [ ] Return Types II: Remove union types `T1|T2|...`, as they are not supported in PHP 7.4 which we will want to support with ILIAS 8.
- [ ] Capitalization: Only MUST, SHOULD, MAY and NOT should be capitalized according to https://datatracker.ietf.org/doc/html/rfc2119. COULD is not one of them.
- [ ] Doc Strings: Please remove doc strings where they do not provide additional information over type hints etc.
- [ ] withAdditionalLinks: Rename `withAdditionalAction` (and according methods) to `withAdditionalLinks`, if these should really be `Links` only.

Thx a lot again, we are looking forward to your reply. Feel free to contact us for a meeting via Skype if you feel that there are things that are easier to be discussed in person.

Best regards!

@rklees and @amstutz

iszmais:
Hi @klees,

thanks for your feedback!

**To the questions:**
- Yeah These component should replace the OSD Notifications, at least their presentation. The handling of OSDNs will be reworked withing a combination of the called Feature Wiki request and https://docu.ilias.de/goto_docu_wiki_wpage_6251_1357.html in long term. As long as no new OSDNs are added this shouldn't lead to a problem.
- It specified the usage. The relevant criteria for a toast should by that the information is useful for a user (not admin or else) and has to depend on the time it was created _(like: "someone wants to chat with you" wich is loosing its relevance  when decaying, and not like "background task finished" wich keeps its relevance for an unlimited (or at least very long) time)_.
- I see the point. I guess that  some criterias are not necessary for toast in general, Therefore i move some of them to the  general factory and some to the standard item of the factory, to differentiate. This one i would keep inside the  standard item, since it may be no requirement for upcoming new kinds of  toast, and there should be mentioned if it is required.
- I haven't looked into this, but i assume  that this item may be not relevant for screen readers either. Since visual observance isnt time passed due to its parallel character, auditive observance is  due to the fact of it serial approach. Therefore a time depending  Component may don't have a use there. And since toast have a interface to handle actions on their vanishing i would assume this would better place to persist them an make them  accessible for screen reader.
- The toast is not supposed to become focused by the keyboard, since could lead to  an interrupt of the users workflow.
- As mentioned the toast time will be set globally withing a new Administration node "Notification" https://docu.ilias.de/goto_docu_wiki_wpage_6494_1357.html
- The Vanish action can be used to persist the toast after its vanishing in  a more decent place like e.g. the notification center. Further it can be used to handle specific ignore actions for  the user _(e.g. the user receives a chat invitation, but ignores it until vanishing, the action then  could inform the chat partner that the user does not noticed the invitation and could therefore close the chat)_
- Changed it
- Other Toast could change the arrangement of the toast or may allow other content, like media previews. As an specific example i you imagine using it to display most of the ilUtil messages (Success, Failure, Info) since most of these informations are temporal relevant an sometimes are missarranged due to their persistence.

**To the suggestions:**
- The expected workflow in the implementation  of the toast should always be that it is displayed to the user  for a short amount of time, and, when the user doesn't interact with it to be filed away in a more decent area (which may no be directly accessible). This seems logic for temporal relevant information.  Since there  are also non-temporal relevant information (e.g. a cookie notification) this specific  precede may not be necessary for their implementation. Therefore this would need another notification  component.
-  I removed the positioning "top right" but would like to keep to mention that its overlaying all other content, since this may be  an exclusion criteria in design  for a developer who  considers using the component.
- Changed it
- I would hold on to that to not loose sturdiness inside the object since the single addition of types is capable of strict types.
- Since the close should abort any persistence of the toasts information i do not see the use of an action here. Do you have an specific example for me of  a necessary  close action?
- Changed it
- Changed it

**To the Requirements:**
- Changed it
- Changed it _(I  intentionally left out to add a rivalry to the system info since i assume it to be one-sided. So the system info is a rival to  the toast but not otherwise since it only offers options with are favorable for the system info )_
- Changed it
- Changed it
- Changed it _(But kept **vanished** instead of **vanishes**, since it more precisely)_
- Changed it
- Changed it
- Changed it
- Changed it
- Changed it
- Changed it
- Changed it

I would like to get your feedback to the open points!

Greetings,
@iszmais

klees:
Hi @iszmais,

I think we could take this PR to the JF, if you are ready to present it on monday. I have set the according label. If you disagree or need some more preperation, feel free to itervene.

I found several little things on revisiting this, and also some feedback.

* Please remove the PHPUnit cache file.
* According to your description, I think you are missing a not [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3411/files#diff-1553aa024cb63fc5c043630c7da136bae9b46dd335b1628dcf58a0b00c580eacR831): "The Toast SHOULD precede all Notifications which are not temporal relevant for the user.".
* Regarding the close action: Currently, the Close Button on the Toast would be completely internal to the component. I.e., clicking that button would basically remove the HTML element of the toast from the DOM on client side, without any chance for the according ILIAS component to notice that click. It looks as if the "vanish" action you propose is basically a device for the same functionality that I proposed the "close" action for. For a "vanish" action the logic in a component would be "if vanish action is triggered, store the notification permanently elsewhere". For a close action the logic would be "if the close action is triggered, remove the notification from some permanent storage". I would guess (but it's just a hunch), that using a "close" action will turn out to deliver more robust results. Consider the case of a lost internet connection: In case of the "vanish" action, the system will never have the chance to know, if a user actually received or noticed the notification. In case of doubt, the message will be lost, because we'll need a specific action to actually store the notification permanently. For the close action, things would be the other way round: In case of doubt or a lost internet connection, the notification will never be removed from the permanent storage and notifications will be available. I would opt for the "Close" action instead of the "Vanish" action.
* Regarding accessibility: We definitely need a way to present the same possibilities to all our users, sighted or not. So just stipulating "not relevant" doesn't cut it =) Getting a notification for, e.g., some chat request, is just as relevant for non-sightes users as it is for any other user, same goes for interacting with the notification. I'm positive that there is a solution to be found here, we are far from the only application that wants to feature notifications. I would propose to make an appointment with the [UI Clinic](https://docu.ilias.de/goto_docu_fold_8199.html) to look into solutions.

Is there anything you are missing a feedback for? Feel free to contact me via Skype, as always.

Best regards!

klees (approved):

matthiaskunkel:
Jour Fixe, 28 JUN 2021: We highly appreciate this PR and see it as an important component for ILIAS. Before accepting it for trunk we would like to have some information in the rules about the following aspects:
- Defining a place where all toasts appear to prevent that toasts are shown no matter where on the screen
- Describing how screensreaders present the information in toasts
- Using getShowSignal
- And please consider the latest comments by Richard (see above).

iszmais:
Hi @klees,

thanks again for the feedback.
* Changed it
* I dont get why a NOT would be neccessary here. This point is recommend to preced normal notification with a toast if its an information wich could be helpfull for the user atm implying to only use normal notifications (without toasts preceeding) when the information isnt helpfull in moment of her apperance. Why would i not show time relevant informations and, therefor implied, show not time relevant informations inside of toasts?
* I agree on that point but i guess this also should cause a change in all other interactions and may also requires more strict rules for the toasts content etc. Maybe we could get in contact to talk about that within the next days.
* Changed it

iszmais:
Results of the meeting 14.07.2021:
- Change the sentence "The Toast SHOULD precede all Notifications which are not temporal relevant for the user." to "If a notification has temporal relevance for a user, it SHOULD be preceded by a Toast." for better understanding.
- Dismiss the vanish action. The using component handles the persistence of a temporary toast if needed. The toast provides an action on every interaction which may triggers e.g. a removal of a persistence.
- Provide a js interface to wich can call the serverside creation of toasts from the client.

iszmais:
@klees all Changes from ther meeting 14.07.2021 where implimented, except the js interface which wasnt mandatory. Could you give you final confirmation as said?

klees:
@iszmais Yes, of course. This week will be tough, but next week should work well. Are you looking to discuss this next week on JF?

iszmais:
@klees As we concluded in the last meeting there should be no need for another appointment in JF. :thinking: 
Or do you have anything major to add since the last meeting?

klees:
@iszmais Seems to be right, I just not very into this atm. Good that we have that documentation here =)

src/UI/Factory.php
```diff
@@ -800,4 +802,47 @@ public function menu() : Component\Menu\Factory;
...
+     *     1: The Toast SHOULD be used for all Notifications which include temporal relevant information for a user.
+     *     2: The Toast SHOULD NOT be used for Notifications which are not time relevant to the point of their creation.
+     *   composition:
+     *     1: If a notification has temporal relevance for a user, it SHOULD be preceded by a Toast.
+     *   interaction:
+     *     1: Click interactions with the Toast MUST remove it permanently.
+     *   style:
+     *     1: The Toast MUST be visible on the top layer of the page, Therefore it MUST cover up all other UI Items in
+     *        its space.
+     *     2: The Toast ä disappear after a certain amount of time or earlier by user interaction. No interaction can
```
klees: `ä` here =)

klees (approved):
LGTM, please remove the `ä` then I'll merge =)

iszmais:
@klees Typo fixed!

PR #3419, opened 2021-05-31
Introduction of PHP Rector

chfsx:
With Rector, some code can be migrated very easily for new PHP versions and generally improved. I therefore propose Rector as a dev-dependency to support us in the PHP8 project as well.


matthiaskunkel:
Jour Fixe, 31 MAY 2021: We highly appreciate this suggestion and accept the PR for trunk. We hope it will make the PHP8 project much easier.

PR #3422, opened 2021-06-02
Rule for new files in root

kergomard:
The Technical Board believes we should try to minimize the amount of files and directories we hold in the root directory of ILIAS. In the middle run, we believe we should move most of the files out of the web-root altogether. To avoid creating more files and directories in the root directory of ILIAS, if not absolutely necessary, we propose the rule in this PR.


matthiaskunkel:
Jour Fixe, 14 JUN 2021: We highly appreciate this suggestion and accept the PR for all maintained versions.

PR #3427, opened 2021-06-04
Make reset administrative notification work from table (trunk)

nicoroeser:
Fix a copy and paste mistake from 1334c90afb8fcf06f960afeaa145b27702e82a8c (forward-ported as be0bcf957bc406d48a751ce15e1d7e3944f50159) so that the _reset_ function (and not the delete function) for administrative notifications is called when users select the corresponding entry from the actions button menu.

This is the pull request for trunk.


chfsx:
Nice catch, thanks a lot! I'll CP this to release_7 as well

PR #3428, opened 2021-06-04
Make reset administrative notification work from table (release_7)

nicoroeser:
Fix a copy and paste mistake from 1334c90afb8fcf06f960afeaa145b27702e82a8c so that the _reset_ function (and not the delete function) for administrative notifications is called when users select the corresponding entry from the actions button menu.

This is the pull request for the release_7 branch.


nicoroeser:
Has been handled as 35721e7eb10365bf74f66109f74442c1d18fa34b by @chfsx, see <https://github.com/ILIAS-eLearning/ILIAS/pull/3427#issuecomment-860470717>. Thanks!

PR #3429, opened 2021-06-08
Module interaction fixes for php 8 compatibility

M-Barz:
When trying to create an object that extends from ilObject2GUI instead of ilObjectGUI (e.g. a Poll or a File), an error occurs because the class path cannot be found. This is due to the requested_ref_id variable not being set when calling the __getHTML function in ilObjectGUI over ilObject2GUI.
Setting it in the constructor of ilObject2GUI fixes that issue.

There was a critical error when trying to delete any object in ILIAS, since $target['cmd'] sometimes isn't set, but still checked for an in_array call. This is because of the new check for isset that happens in line 524.
Setting it to an empty array if $target['cmd'] isn't set fixes the issue.

https://mantis.ilias.de/view.php?id=30895


mjansenDatabay:
Thanks @M-Barz this is one of the 'Blocking Issues' of the current `trunk`.

mjansenDatabay:
@alex40724 Could we please merge this, because otherwise creating objects is (partly) not possible (no matter which PHP version you use) and we'll have to work with local patches until this is fixed.

smeyer-ilias (approved):
LGTM, thanks.

PR #3430, opened 2021-06-09
BT-29650: After registration user can now click on a link to get to the Login page instead of landing on another registration form

M-Barz:
Link to Mantis-Report:
https://mantis.ilias.de/view.php?id=29650

Additional Change: I removed the password parameter from the Login-Function and it's call, since the variable wasn't used anywhere in the function anymore.


smeyer-ilias (commented):
Thanks Marvin, LGTM in general. 
You removed the line ilSession::set('registered_user'...
Please can search for other occurrences of this session variable and also refactor/delete possible ilSession::get('registered_user') occurrences.

PR #3431, opened 2021-06-09
Adding a check for empty condition types when saving a learning sequence

dmundra:
Fixes the bug I created in Mantis: https://mantis.ilias.de/view.php?id=30854. It doesn't show a message as I suggested but maybe it does not need to.


dmundra:
Ah I see that this is not a bug in release_6, will submit a separate pull request for 5.

PR #3432, opened 2021-06-09
Adding a check for empty condition types when saving a learning sequence

dmundra:
Fixes the bug I created in Mantis: https://mantis.ilias.de/view.php?id=30854. It doesn't show a message as I suggested but maybe it does not need to.

Not an issue in version 6.9

### DIFF
--- a/Modules/LearningSequence/classes/Content/class.ilObjLearningSequenceContentGUI.php
+++ b/Modules/LearningSequence/classes/Content/class.ilObjLearningSequenceContentGUI.php
@@ -151,12 +151,14 @@ class ilObjLearningSequenceContentGUI
             $order = $this->getFieldName(self::FIELD_ORDER, $ref_id);
             $condition_type = $this->getFieldName(self::FIELD_POSTCONDITION_TYPE, $ref_id);
 
-            $condition = $lsitem->getPostCondition()
+            if (!empty($post[$condition_type])) {
+              $condition = $lsitem->getPostCondition()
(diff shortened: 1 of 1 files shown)


klees:
Hi @dmundra,

thanks for looking into this. @nhaagen will come here with a feedback soon. Please make sure to apply our [coding style](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/docs/development/coding-style.md) to the code you propose.

Thanks!

dmundra:
Thanks @klees. Fixed the indentation.

nhaagen:
Hi @dmundra,
thanks for the PR. 
I had difficulties reproducing the issue, and found that I could not due to https://github.com/ILIAS-eLearning/ILIAS/commit/7aa847c98b83f7517295f00a3326204d4e612c0e

For now, I'd go with the already accepted/merged version.
If pagination is a must, the update-routine should check not only for the erroring condition-field, but for the "online"-option as well, as not to produce strange side-effects.
Best regards,
Nils

@klees: I'd suggest closing this.

klees:
Hi @dmundra,

this seems to already have a solution, you may want to have a look into the reasoning in the [ticket](https://mantis.ilias.de/view.php?id=28566) as well.

So I close this here.

Still, thanks for caring and looking into this. Feel free to get back to us if we can be of further help.

Best regards!

dmundra:
Thanks, @nhaagen, and @klees. That fix and link make sense.

PR #3434, opened 2021-06-10
ILIAS 7 forum html export overflow added to template

mBeym:
If multiple forum threads are exported as html.
The html can't be viewed correctly as the body won't scroll.
This adds **overflow: auto;** to the **<body>** tag in the **tpl.forums_export_html.html** template

https://mantis.ilias.de/view.php?id=30853


mjansenDatabay (changes requested):
Thx for your PR @mBeym 

Could we try to add a new CSS class in `templates/default/less/Modules/Forum/delos.less` instead of using an inline style?

mjansenDatabay (approved):

PR #3435, opened 2021-06-10
ILIAS trunk forum html export overflow added to template

mBeym:
If multiple forum threads are exported as html.
The html can't be viewed correctly as the body won't scroll.
This adds overflow: auto; to the tag in the tpl.forums_export_html.html template

https://mantis.ilias.de/view.php?id=30853

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -7091,7 +7091,7 @@ button.close {
   position: relative;
   background-color: white;
   border: none;
-  box-shadow: 0 1px 2px rgb(0 0 0%), 0 0px 0px rgb(0 0 0%);
+  box-shadow: 0 1px 2px rgba(0, 0, 0, 0.15), 0 0px 0px rgba(0, 0, 0, 0.15);
   /* see bug #24947 */
 }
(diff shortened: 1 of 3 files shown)


mjansenDatabay (changes requested):
Thx for your PR @mBeym 

Could we try to add a new CSS class in `templates/default/less/Modules/Forum/delos.less` instead of using an inline style?

templates/default/delos.css
```diff
@@ -7091,7 +7091,7 @@ button.close {
   position: relative;
   background-color: white;
   border: none;
-  box-shadow: 0 1px 2px rgb(0 0 0%), 0 0px 0px rgb(0 0 0%);
+  box-shadow: 0 1px 2px rgba(0, 0, 0, 0.15), 0 0px 0px rgba(0, 0, 0, 0.15);
```
mjansenDatabay: @mBeym Please remove this change.

mBeym: @mjansenDatabay done.

mjansenDatabay (approved):

PR #3436, opened 2021-06-11
Service/PHPUnit: remove component

klees:
Hi everybody,

this removes `Services/PHPUnit` (but not PHPUnit-tests of course). What this indeed does is:

* first and foremost: remove that component, which is without maintenance
* this means: removing `ilUnitUtils`, which is located in that folder
* this implies: commenting out `ilUnitUtils::performInitialisation` in the according locations, which did not work for a long time anyway (to my knowledge)
* this allows to also remove `needsInstalledILIAS` and according code
* which means that according tests won't be working `TestCase`s anymore, so we can outcomment that `extends` for the moment to stop the tests from being picked up by the global suite
* finally, the last bit of code that we actually use from that folder is the global test suite, which gets a new life as `ILIASSuite` in the common `tests` folder.

I assume that the test that I modified did not run for a long time anyway, so this should not remove any actual benefit. I did not look into specifics of the tests that have been marked with `needsInstalledILIAS` but from a very cursory look I assume that at least some of them don't actually need an installed ILIAS. Some others seem to actually require an installed ILIAS, on the other hand. So I left all the code to allow maintainers to look for themselves.

To proceed here, I ask every affected maintainer to approve this PR, which just renders some `TestCase`s that require an installed ILIAS disfunctional in another way as before. Then everyone could look into their affected tests to find out if or how the tests can be made functional again.

Best regards!


mjansenDatabay:
@klees Thx for this PR!

Most of the test cases in `Modules/Chatroom` also called/relied on `ilUnitUtils::performInitialisation`. I just removed the broken tests (which were simply nonsense or never worked) of this component and integrated all working tests (after I made some adjustments to not rely on an initialized ILIAS anymore) into the global test suite (+44 tests, +114 assertions).

chfsx (approved):

mbecker-databay (approved):
Looks good to me, thank you, Richard

theodortruffer (approved):

alex40724:
All fine for me. Thanks.

alex40724 (approved):

matthiaskunkel:
Jour Fixe, 28 JUN 2021: We highly appreciate this suggestion and accept to abandon the PHPUnit directory and related code with ILIAS 8 - in case Stefan Meyer is not giving a veto.

klees:
This is open for a month now. Since I did not receive any veto from Stefan until now, I decided to merge this so we can go on.

PR #3441, opened 2021-06-14
CI: Checks if file exists before testing it

catlauraherzog:
There was a problem here: https://github.com/ILIAS-eLearning/ILIAS/runs/2804371985?check_suite_focus=true#step:6:24

With this PR its gone.


klees:
Cherry-picked to ` release_7` and `release_6`.

PR #3443, opened 2021-06-14
UI: remove script-tag in template-files

klees:
This deals with the script-tags in the template files of the UI framework as far as currently possible. The general need to remove the script tags now is implied by [Ticket 30833](https://mantis.ilias.de/view.php?id=30833)/[PR 3409](https://github.com/ILIAS-eLearning/ILIAS/pull/3409). The script-tag in the `Month Button` has been removed completely. The Lightbox Modal (the other according location) has been investigated. The Lightbox Modal seems to be in dire need of some love anyway and does not work properly currently. So I did not remove that script-tag, but made it complient to the changes in the aforementioned PR by simply wrapping a `setTimeout` around. Also, I did add an according entry to the roadmap.

### DIFF
--- a/src/UI/Implementation/Component/Button/Renderer.php
+++ b/src/UI/Implementation/Component/Button/Renderer.php
@@ -256,14 +256,10 @@ class Renderer extends AbstractComponentRenderer
         }
         $tpl->setVariable("LANG", $lang_key);
 
+        $component = $component->withAdditionalOnLoadCode(fn($id) => "il.UI.button.initMonth('$id');");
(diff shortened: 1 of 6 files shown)


Amstutz:
Hi @klees 

Thx a lot for taking care of this. I dislike the timeout hack. Together with the entry in the roadmap, I can live with it however. The code concerning the Month Button looks clear now, so kudos for that.

I tested both, examples for the Lightbox and Month Button, they seem to work and the Unit Tests run.

Thx a lot.

PR #3444, opened 2021-06-15
Services/Form: make it work with PR 3409

klees:
Hi everyone,

as [discussed on the JF](https://docu.ilias.de/goto.php?target=wiki_1357_JourFixe-2021-06-14) and documented in [Ticket 30833](https://mantis.ilias.de/view.php?id=30833), moving the javascript includes to the footer of the page ([PR 3409](https://github.com/ILIAS-eLearning/ILIAS/pull/3403)) caused problems in various components that violate the rule `IliasTemplateFiles cannot contain text: "<script"`.

This PR is supposed to make inputs in the legacy `ilPropertyFormGUI` work again.

As already suggested in the discussion, there are four strategies to deal with the problem that I could identify. Three of them can be seen in this PR:

* if the code in script does not depend on any placeholders or sections processed during templating, it can simply be moved to its own js-file
* if the code is not used anymore, it can be removed
* if the code is not worth a refactoring (or time is short), it can be simply wrapped into a `window.setTimeout` to defer execution to after the inclusion of the script tags in the footer of the page

The fourth strategy that I could think of would be to refactor the code in a way that removes dependencies on templating, e.g. by using data attributes in HTML elements to pass data instead of directly pasting it into js-code. Since the legacy forms are supposed to be replaced by the form implementation of the UI framework, it did not consider the code worth a refactoring in this case here.

Finally: This PR removes the `ilFlashFileInputGUI` and its dependants in the T&A. This will finalize the move to abandon that question type, started with [ILIAS 5.4](https://docu.ilias.de/goto.php?target=wiki_1357_Remove_Question_Types_Flash_and_Java_Applet). This is also, why I added the JF tag to this PR.

Best regards!


dsstrassner:
Thanks, @klees, for removing the last code snippets of the flash question type!

theodortruffer (approved):

alex40724 (approved):
Changes in my code are fine. Thanks.

matthiaskunkel:
Jour Fixe, 28 JUN 2021: We highly appreciate this PR and accept it for trunk. We fully support the decision to abandon the test question type Flash due to this PR.

mbecker-databay (approved):

PR #3455, opened 2021-06-18
Setup: trunk7_30865 implement command achieve

daniwe4:
- remove command reload-control-structure
- add class AchieveCommand
- add tests for command achieve
- add method reloadCtrlStructure to ilUICoreSetupAgent
- adjust Readme.md for setup


klees:
Hi Jour Fixe,

I want to ask for approval to merge this to the trunk, even so some might consider this change a feature that was brought up to late. To solve [#30940](https://mantis.ilias.de/view.php?id=30940) we might even consider to pull this down to release_7.

When working on the setup we found that some facility to directly achieve objectives greatly improves debugging and is also of general value for the setup. We thus propose to introduce an `achieve` command, that allows to target single objectives directly. The command will superseed the current `reload-control-structure`-command, which will then be `achieve ilUICoreSetupAgent::reloadCtrlStructure`. A similar call could be used to request a new NIC: `achieve ilSetupAgent::registerNIC`.

Best regards!

matthiaskunkel:
Jour Fixe, 28 JUN 2021: We highly appreciate this PR and accept it for trunk and ILIAS 7 (last but not least to support NIC registration in already installed ILIAS installations also with 7).

PR #3464, opened 2021-06-24
Setup: Throw exception if NIC registration objective could not be ach…

mjansenDatabay:
…ieved due to unexpected response

Related to Mantis Issue: https://mantis.ilias.de/view.php?id=30907


klees:
Picked to trunk. Thanks a bunch!

PR #3465, opened 2021-06-24
GlobalCache: streamline config and status output.

klees:
Hi @chfsx,

this streamlines the output of the global cache via status and the input as a config, as found in: https://mantis.ilias.de/view.php?id=30727

I think this should be pulled down to previous versions as well.

Best regards!


chfsx:
Thanks a lot! I'll cherry pick this to the relevant branches as well

PR #3466, opened 2021-06-24
UI, TagInput: add dependency 'tagify', remove 'typeahead' and 'bootstrap-tagsinput'

nhaagen:
Dear colleagues,
this replaces the bootstrap-taginput with a a tagify-based component.
tags_new
Besides dropping the dependencies, this version behaves better with special chars in tags.
Please kindly consider, best regards,
Nils


Amstutz (changes requested):
Hi @nhaagen

Thx a lot for this important contribution!

Some Questions:
- [ ]  Tagfiy CSS: Do we really need the complete Tagify CSS? Could we not just cherry pick the lines important to use and not load the rest? If it is too much, forget about it, but maybe there is hope ;-).

Please do, or reason why not:
- [x]  a11y I: The title in the Tag Field is set redundantly. Once in the attribute of the title tag (<tag title=„…“) and once in the content. I am not sure if you can avoid this with this library. If possible try not to have such duplicates. This is annoying for screenreaders. If not possible, a ticket to the author of the library reporting the issue would be neat.
- [ ]  Design I: The strike-through animation looks a bit strange in the context of delos. Maybe remove. Any solution @yvseiler or @ckrahl appove, are fine with me here.
PastedGraphic-2
- [ ]  Design II: Same as above for the disabled look. 
PastedGraphic-3
 
Please do:
- [x]  a11y II: The Input field does not provide a label. Probably this is best fixed, by adding setting the label-for field in the context_form, e.g. like the numeric field does it. Screenreaders need this to identify the purpose of the input.
- [x]  Compile CSS: Compile the final CSS output from less. Currently the compiled version is lacking and the inputs looks therefore somewhat funky. 
- [x]  Remove Outcommented Code: The code for adding a value via template in the renderer is outcommented. Remove it. It will not work with the new template anyways.
- [x]  Remove Remove Debugging Output: Currently you output some data in js into the dev console. Remove those lines. However note, that by looking at you code, I just learned a really neat way of nicely grouping output. Kudos for that, thx ;-). 
- [x]  New Less Color: You introduce a new Less color in your less code (f3fafb). All new colors used must be defined in the variables.less Main Colors section. Please chose one of those and make sure the contrast is high enough with the used background.
- [x]  Bootstrap Less Color Variable: You use the bootstrap brand-primary color variable from bootstrap in your new less. We try to get rid of as much of the bootstrap deps as we can. Use `@il-main-color` instead (value is the same as brand-primary).

Thx a lot for all this work.
@amstutz

nhaagen:
Hi @Amstutz ,
thanks for looking into this, and sorry for the delay.
I moved/mentioned the less-vars in variables.less and defaulted them there.
Other points are taken care of as well, with the exception of the design issues you documented with screenshots.
I should not even try and hand this over to @yvseiler or @catkrahl directly ;)
Well, and then there is the tagify-package; I don't see a good solution to only checkout the dist-files and still use npm - but this is a general problem, right?

Best regards,
Nils

Amstutz:
@nhaagen Thx for taking care of the mentioned points. I can live with the design ftm, IMO this is overall a big step into the right direction. thx.

PR #3469, opened 2021-06-25
Object: move cache from component to object

klees:
Hi @smeyer-ilias and @alex40724,

I moved caching related to ilObject from `Services/Component` to `Services/Object`. I was surprised to find, that we already have an according entry (`obj_def`) in the configuration of the global cache, which, in fact, does not do anything. This PR makes this config effective and modularizes the system a little more.

Best regards!

### DIFF
--- a/Services/Object/classes/class.ilCachedObjectDefinition.php
+++ b/Services/Object/classes/class.ilCachedObjectDefinition.php
@@ -1,0 +1,168 @@
+<?php declare(strict_types=1);
+
+/**
+ * Cache for object definitions, based on ilGlobalCache.
+ */
+class ilCachedObjectDefinition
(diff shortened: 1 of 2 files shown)


alex40724:
@chfsx I think this is mostly your code. Could you check this, please?

chfsx (approved):
Thanks a lot, @klees! LGTM and it's an improvement. beside this (and after merging), we could have a look at moving some of those infos to an artifact, which could improve performance much more. let's have a skype meeting for this

klees:
@alex40724 @smeyer-ilias Any objections to merge this?

chfsx:
Thanks @klees for this PR, I merged this to trunk

PR #3470, opened 2021-06-25
fixed #26448: ViewControl Pagination nicht mehr shy

alex40724:
https://mantis.ilias.de/view.php?id=26448


Amstutz:
Hi @akill

Thx a lot for the fix! I resolved the conflict and cherry picked to 6, 7 and trunk.

PR #3474, opened 2021-06-28
Setup: fix results on inspection code run

daniwe4:
src/Data
src/DI
src/KioskMode
src/Refinery


klees:
@daniwe4: Please provide separate PRs for the various components you change here.

PR #3476, opened 2021-06-28
Setup: Process possible NIC registration error to build exception mes…

mjansenDatabay:
…sage

This commit adds further processing of the NIC registration HTTP response to build a more useful exception message if the registration failed.


klees:
Thanks Michael!

PR #3477, opened 2021-06-28
usability fix: addition to https://mantis.ilias.de/view.php?id=30136

eqsoft:
Hi Max,
i guess it should also be fixed for results overview like in patch-30316. Can you check this?
regrard, Stefan


mbecker-databay:
Thank you so much! 👍

PR #3483, opened 2021-06-29
src/Setup: add declare strict types and use type hints

daniwe4:


klees:
I have merged this manually!

PR #3487, opened 2021-06-30
src/Refinery: add declare strict types and use type hints

daniwe4:


mjansenDatabay:
Thx @klees / @daniwe4

PR #3488, opened 2021-06-30
Refinery: Provide fixes based on inspection results

mjansenDatabay:
This PR ...

- Adds access modifiers to class constants
- Adds types where missing (and applicable)
- Use strict comparisons  (where applicable)
- Replaces `intval`, `floatval` and `strval` with respective cast (because faster and less cognitive load)
- Adds `u` modifier to RegEx
- Removes redundant interface implementations of constraints
- Improves loop and `sprintf` call in `ProblemBuilder`
- Optimizes conditions in if statements (less expensive first)


klees (approved):

klees:
Thx a bunch!

PR #3489, opened 2021-06-30
Fixed #30691: Skipping between Learning Modules inside a LearningSequence throws LogicException

skullbox305:
Fix for the following issue: https://mantis.ilias.de/view.php?id=30691

### DIFF
--- a/src/GlobalScreen/ScreenContext/BasicScreenContext.php
+++ b/src/GlobalScreen/ScreenContext/BasicScreenContext.php
@@ -104,7 +104,7 @@ class BasicScreenContext implements ScreenContext
      */
     public function addAdditionalDataIfNotExist(string $key, $value) : ScreenContext
     {
-        if($this->additional_data->exists($key)) {
+        if ($this->additional_data->exists($key)) {
             $this->additional_data->replace($key, $value);
         } else {
             $this->additional_data->add($key, $value);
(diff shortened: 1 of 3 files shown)


src/GlobalScreen/ScreenContext/BasicScreenContext.php
```diff
@@ -95,10 +95,23 @@ public function getAdditionalData() : Collection
...
+   
         return $this;
     }
 
+    /**
+     * @inheritDoc
+     */
+    public function addAdditionalDataIfNotExist(string $key, $value) : ScreenContext
+    {
+        if($this->additional_data->exists($key)) {
```
mjansenDatabay: Hi @skullbox305,

thanks for your contribution.

Please use the `php-cs-fixer` to apply our coding style guideline (see: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/coding-style.md).

A detailed review will be provided by the assignees.

Best regards,
@mjansenDatabay

skullbox305:
Hi @mjansenDatabay ,
I applied the coding style with `php-cs-fixer`. Thank you for the tip!

Kind regards

chfsx:
@skullbox305 as mentioned in https://mantis.ilias.de/view.php?id=30691, I agree with richard's statement that the problem should be solved outside in the calling instance. in the same request, the same "additionalData" should only be called once. The GlobalScreen context provides for this and therefore only accepts additional data once. If there had been a legitimate reason to pass the same data several times during a request, your solution would have been the right one. but since this should be fixed in the learning module, i'll close the PR. thanks anyway for your work and the proposed solution!

PR #3492, opened 2021-07-05
Updated base.php to demonstrate issue #30933

thibsy:
This adjustment demonstrates an inconsistent behaviour of "withValue()":
Steps to reproduce:
1) load the snippet, a form with a multi-select- and text-input appears.
2) note that "Pick 2" is selected and "some string" is the text-input's value.
3) uncheck "Pick 2" and remove the value "some string" from the text-input (so it has no value).
4) click save to submit the form and check $result that appears above the form.
You will notice that 'text' has a nullish entry but 'multi' does still have "Pick 2" (index '2') as the selected option.

### DIFF
--- a/src/UI/examples/Input/Field/MultiSelect/base.php
+++ b/src/UI/examples/Input/Field/MultiSelect/base.php
@@ -22,11 +22,11 @@ function base()
     );
 
     //Step 1: define the select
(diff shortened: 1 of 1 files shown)


klees:
Thx for this demo, @thibsy!

PR #3493, opened 2021-07-06
Data: add value type for versions

klees:
Hi everyone,

this PR will add a more or less naive implementation of a value type for versions. I'm looking to use this for plugins in `Services/Component` and to implement the [strict update paths](https://docu.ilias.de/goto_docu_wiki_wpage_6911_1357.html).

I announce this here because this might be interesting to some more people and also to get some feedback. Maybe someone could review this before I merge.

Best regards!

### DIFF
--- a/src/Data/Version.php
+++ b/src/Data/Version.php
@@ -1,0 +1,88 @@
+<?php declare(strict_types=1);
+
+namespace ILIAS\Data;
+
+/**
+ * A version number that consists of three numbers (major, minor, patch).
+ */
+class Version
+{
+    private const REGEXP = '(?<major>\d+)([.](?<minor>\d+)([.](?<patch>\d+))?)?';
+
+    protected int $major;
+    protected int $minor;
+    protected int $patch;
+
+    public function __construct(string $version)
+    {
+        $match = [];
+        if (!preg_match("/" . self::REGEXP . "/", $version, $match)) {
+            throw new \InvalidArgumentException(
+                "Expected version string to match this regular expression: " . self::REGEXP
+            );
+        }
+        $this->major = (int) $match["major"];
+        $this->minor = (int) ($match["minor"] ?? 0);
+        $this->patch = (int) ($match["patch"] ?? 0); 
+    }
+
+    public function getMajor() : int
+    {
+        return $this->major;
+    }
+
+    public function getMinor() : int
+    {
+        return $this->minor;
+    }
+
+    public function getPatch() : int
+    {
+        return $this->patch;
+    }
+
+    public function equals(Version $other) : bool
+    {
+        return
+            $this->major === $other->major
+            && $this->minor === $other->minor
+            && $this->patch === $other->patch;
+    }
+
+    public function isGreaterThan(Version $other) : bool
+    {
+        if ($this->major > $other->major) {
+            return true;
+        }
+        if ($this->major < $other->major) {
+            return false;
+        }
+        if ($this->minor > $other->minor) {
+            return true;
+        }
+        if ($this->minor < $other->major) {
+            return false;
+        }
+        return $this->patch > $other->patch;
+    }
+
+    public function isGreaterThanOrEquals(Version $other) : bool
+    {
+        return $this->equals($other) || $this->isGreaterThan($other);
+    }
+
+    public function isSmallerThan(Version $other) : bool
+    {
+        return $other->isGreaterThan($this);
+    }
+
+    public function isSmallerThanOrEquals(Version $other) : bool
+    {
+        return $other->isGreaterThan($this) || $this->equals($other);
+    }
+
+    public function __toString()
+    {
+        return "$this->major.$this->minor.$this->patch";
+    }
+}
--- a/tests/Data/VersionTest.php
+++ b/tests/Data/VersionTest.php
@@ -1,0 +1,94 @@
+<?php declare(strict_types=1);
+
+use ILIAS\Data;
+use PHPUnit\Framework\TestCase;
+
+class VersionTest extends TestCase
+{
+    protected function setUp() : void
+    {
+        $this->f = new Data\Factory();
+    }
+
+    public function testSmoke() : void
+    {
+        $v = $this->f->version("0.1.0");
+
+        $this->assertEquals("0.1.0", (string) $v);
+    }
+
+    public function testMajorOnly() : void
+    {
+        $v = $this->f->version("1");
+
+        $this->assertEquals("1.0.0", (string) $v);
+    }
+
+    public function testNoPatchVersion() : void
+    {
+        $v = $this->f->version("1.2");
+
+        $this->assertEquals("1.2.0", (string) $v);
+    }
+
+    public function testSubVersions() : void
+    {
+        $v = $this->f->version("1.2.3");
+
+        $this->assertEquals(1, $v->getMajor());
+        $this->assertEquals(2, $v->getMinor());
+        $this->assertEquals(3, $v->getPatch());
+    }
+
+    /**
+     * @dataProvider greaterThanProvider
+     */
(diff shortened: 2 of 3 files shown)


src/Data/Version.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        return
+            $this->major === $other->major
+            && $this->minor === $other->minor
+            && $this->patch === $other->patch;
+    }
+
+    public function isGreaterThan(Version $other) : bool
+    {
+        if ($this->major > $other->major) {
```
mjansenDatabay, line 54: You could use the `__toString` method (or `(string` cast) of both values and use `version_compare()` instead.

tests/Data/VersionTest.php
```diff
@@ -0,0 +1,92 @@
...
+    {
+        $this->assertFalse($l->isGreaterThan($r));
+        $this->assertFalse($r->isSmallerThan($l));
+        $this->assertTrue($l->isGreaterThanOrEquals($r));
+        $this->assertTrue($r->isSmallerThanOrEquals($l));
+        $this->assertTrue($l->equals($r));
+        $this->assertTrue($r->equals($l));
+    }
+
+    public function equalsProvider()
```
mjansenDatabay:
```suggestion
    public function equalsProvider() : array
```

tests/Data/VersionTest.php
```diff
@@ -0,0 +1,92 @@
...
+    {
+        $this->assertTrue($l->isGreaterThan($r));
+        $this->assertTrue($r->isSmallerThan($l));
+        $this->assertTrue($l->isGreaterThanOrEquals($r));
+        $this->assertTrue($r->isSmallerThanOrEquals($l));
+        $this->assertFalse($l->equals($r));
+        $this->assertFalse($r->equals($l));
+    }
+
+    public function greaterThanProvider()
```
mjansenDatabay:
```suggestion
    public function greaterThanProvider() : array
```

tests/Data/VersionTest.php
```diff
@@ -0,0 +1,92 @@
...
+            [$f->version("1.2.3"), $f->version("1.2.2")],
+            [$f->version("1.2.3"), $f->version("1.1.3")],
+            [$f->version("1.2.3"), $f->version("0.2.3")]
+        ];
+    }
+
+    /**
+     * @dataProvider equalsProvider
+     */
+    public function testEquals($l, $r)
```
mjansenDatabay:
```suggestion
    public function testEquals(\ILIAS\Data\Version $l, \ILIAS\Data\Version $r) : void
```

tests/Data/VersionTest.php
```diff
@@ -0,0 +1,92 @@
...
+
+        $this->assertEquals(1, $v->getMajor());
+        $this->assertEquals(2, $v->getMinor());
+        $this->assertEquals(3, $v->getPatch());
+    }
+
+    /**
+     * @dataProvider greaterThanProvider
+     */
+    public function testGreaterThan($l, $r)
```
mjansenDatabay:
```suggestion
    public function testGreaterThan(\ILIAS\Data\Version $l, \ILIAS\Data\Version $r) : void
```

tests/Data/VersionTest.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public function testNoPatchVersion()
+    {
+        $v = $this->f->version("1.2");
+
+        $this->assertEquals("1.2.0", (string) $v);
+    }
+
+    public function testSubVersions()
```
mjansenDatabay:
```suggestion
    public function testSubVersions() : void
```

tests/Data/VersionTest.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public function testMajorOnly()
+    {
+        $v = $this->f->version("1");
+
+        $this->assertEquals("1.0.0", (string) $v);
+    }
+
+    public function testNoPatchVersion()
```
mjansenDatabay:
```suggestion
    public function testNoPatchVersion() : void
```

tests/Data/VersionTest.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public function testSmoke()
+    {
+        $v = $this->f->version("0.1.0");
+
+        $this->assertEquals("0.1.0", (string) $v);
+    }
+
+    public function testMajorOnly()
```
mjansenDatabay:
```suggestion
    public function testMajorOnly() : void
```

tests/Data/VersionTest.php
```diff
@@ -0,0 +1,92 @@
...
+use PHPUnit\Framework\TestCase;
+
+class VersionTest extends TestCase
+{
+    protected function setUp() : void
+    {
+        $this->f = new Data\Factory();
+    }
+
+    public function testSmoke()
```
mjansenDatabay:
```suggestion
    public function testSmoke() : void
```

mjansenDatabay (approved):
Thanks a lot for this PR @klees 

I like the idea of having some kind of `Value Object` or higher data type for a `Version`. This could be specifically interesting in `Services/Component` when checking the state of plugins.

I approve merging these changes and have only some minor remarks.

src/Data/Version.php
klees: I'll leave it like this to be more explicit about the algorithm. I somehow expect that we will want to add more parts to the version at some point. Thanks anyway for the suggestion!

klees:
Thx for the feedback @mjansenDatabay. That plugin use case is exactly what I'm looking to use this for.

src/Data/Version.php
```diff
@@ -0,0 +1,94 @@
...
+        $match = [];
+        if (!preg_match("/" . self::REGEXP . "/", $version, $match)) {
+            throw new \InvalidArgumentException(
+                "Expected version string to match this regular expression: " . self::REGEXP
+            );
+        }
+        $this->major = (int) $match[1];
+        $this->minor = 0;
+        $this->patch = 0;
+        if (count($match) > 3) {
```
chfsx: Just as an idea, you could use named groups in regex here. this would allow to use isset/empty. 
https://www.regular-expressions.info/refext.html

chfsx (approved):
LGTM, thanks a lot!

tests/Data/VersionTest.php
```diff
@@ -0,0 +1,92 @@
...
+
+    public function equalsProvider() : array
+    {
+        $f = new Data\Factory();
+        return [
+            [$f->version("0.0.1"), $f->version("0.0.1")],
+            [$f->version("0.1.0"), $f->version("0.1.0")],
+            [$f->version("1.0.0"), $f->version("1.0.0")],
+            [$f->version("1.2.3"), $f->version("1.2.3")]
+        ];
```
tfamula, line 92: Two entries are missing here if I'm not mistaken

tfamula (approved):

tests/Data/VersionTest.php
klees: Hi @tfamula!

You mean entries with zeros in second and third position, right?

Thanks!

src/Data/Version.php
klees: Nice idea, thanks!

tests/Data/VersionTest.php
tfamula: I don't see anymore how it looked before (you did a forced push, right?). But it looks correct now :D

klees:
Thanks y'all!

rsheer:
Sorry, I'm a little bit late to this since the PR has already been merged...

Nevertheless, I would like to thank you for this PR! I totally approve it post-merge.

klees:
Hi @rsheer!

No need to apologize, just added you as potential reviewer =) Since this received various reviews quickly, I did not wait for your reply.

Thanks anyway!

PR #3500, opened 2021-07-13
T&A: explain inavailability of corrections (release_5-4)

nicoroeser:
Introduce a change to the test&assessment code which should help users understand ILIAS behavior in certain situations: when corrections are enabled, and users open the corrections view of a test which is not using a fixed question set, tell them that corrections are not supported in this configuration.

I simply exchange the table for a message. The code has been tested with current ILIAS 5.4, 6, and 7 branches (couldn’t test with trunk due to a problem during setup), and works properly, as far as I can tell.

Two questions which I am not sure about:
1. Is the code correct which sets up and retrieves the localized string? I looked at other parts of the code and this seems to be OK. But I do not know whether the initialization put into the constructor is necessary and correct like this.
2. I tried to find out how a message like the new one should best be displayed in the ILIAS web UI, but I haven’t been successful. If there is a dedicated component which is suitable, this one could of course be used. I don’t see a big problem with the current solution, but thought about using one of these alert boxes (with informative/warning level perhaps?). What do you think and recommend? (I also do not like for this seemingly simple change to become overly complex in the end, as my main intent is to signal to users that what they see is intended behavior.)

[Mantis #17901](https://mantis.ilias.de/view.php?id=17901 "Mantis 0017901: Bei einem Test mit zufälliger Fragenauswahl funktioniert die Nachkorrektur nicht") is the issue report about the underlying problem for ILIAS users.

This is the pull request for the `release_5-4` branch.

### DIFF
--- a/Modules/Test/classes/class.ilTestCorrectionsGUI.php
+++ b/Modules/Test/classes/class.ilTestCorrectionsGUI.php
@@ -74,19 +74,32 @@ class ilTestCorrectionsGUI
     {
         $this->DIC->tabs()->activateTab(ilTestTabsManager::TAB_ID_CORRECTION);
         
-        $table_gui = new ilTestQuestionsTableGUI(
-            $this,
-            'showQuestionList',
-            $this->testOBJ->getRefId()
-        );
-        
-        $table_gui->setQuestionTitleLinksEnabled(true);
-        $table_gui->setQuestionRemoveRowButtonEnabled(true);
-        $table_gui->init();
-        
-        $table_gui->setData($this->getQuestions());
-        
-        $this->DIC->ui()->mainTemplate()->setContent($table_gui->getHTML());
+        $ui = $this->DIC->ui();
+
+        if ($this->testOBJ->isFixedTest()) {
+            $table_gui = new ilTestQuestionsTableGUI(
+                $this,
+                'showQuestionList',
+                $this->testOBJ->getRefId()
+            );
+
+            $table_gui->setQuestionTitleLinksEnabled(true);
+            $table_gui->setQuestionRemoveRowButtonEnabled(true);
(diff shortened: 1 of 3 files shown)


klees:
Hi @nicoroeser,

I don't know much about the content of your change, but I think I can answer your two questions, or at least help:

**at 1.:** The initialisation of `ilLanguage` (`loadLanguageModule`) looks fine, although I would expect that this module is already loaded at the point where the GUI is called because other GUIs will have needed it already. But no harm in this AFAIK.

**at 2.:** This is not the way to do it. Inline HTML is a bad idea, because it breaks the modularization between GUI and templating and can't be modified via the skin. In general we are looking to move ILIAS to a place, where all GUI output is created via the UI framework. I think an [Info Message Box](https://test54.ilias.de/goto_test54_stys_21_MessageBoxMessageBoxInfo_default_delos.html) is what you are looking for.

Hope this helps, feel free to ping me for further questions.

Best regards!

dsstrassner:
Hi @nicoroeser, 

I think the message text is good, and I also think that the "message box" is the right way to point out the fact to the user. 

If you could still include it, I have nothing against the merge in terms of content. @mbecker-databay would have to give his OK technically ;-)

nicoroeser:
Many thanks for your helpful comments, @klees and @dsstrassner!

I’m going to change the pull request to use an Info Message Box instead.

nicoroeser:
I have integrated the comments from @klees and updated all four pull requests. Thanks again for your feedback! :smiley:

matthiaskunkel:
Jour Fixe, 09 AUG 2021 : PR is highly appreciated. Please merge (see decision in https://mantis.ilias.de/view.php?id=17901).

PR #3508, opened 2021-07-16
UI Components, Rule change, see: #28954

Amstutz:
Proposed change of Dropdown Rule, see: https://mantis.ilias.de/view.php?id=https://mantis.ilias.de/view.php?id=28954

### DIFF
--- a/src/UI/Factory.php
+++ b/src/UI/Factory.php
@@ -294,8 +294,9 @@ interface Factory
      *           The focus MAY leave the Dropdown if tab is pressed while focusing the last
      *           element. This differs from the behaviour in Popovers and Modals.
      *      8: >
-     *           Dropdowns MUST set the aria-label describing the contained options, if the
-     *           term "Actions" does not describe the contained options well.
+     *           If the description of the contained options are not already given by the component containing the
+     *           dropdown or the Button triggering it, then it MUST be set with the aria-label.
+     *           If the aria-label is just a the Buttons title, it MUST be omitted.
      * ---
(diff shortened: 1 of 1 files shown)


src/UI/Factory.php
```diff
@@ -294,8 +294,9 @@ public function button();
      *           The focus MAY leave the Dropdown if tab is pressed while focusing the last
      *           element. This differs from the behaviour in Popovers and Modals.
      *      8: >
-     *           Dropdowns MUST set the aria-label describing the contained options, if the
-     *           term "Actions" does not describe the contained options well.
+     *           If the description of the contained options are not already given by the component containing the
+     *           dropdown or the Button triggering it, then it MUST be set with the aria-label.
+     *           If the aria-label is just a the Buttons title, it MUST be omitted.
```
kergomard: Thank you very much! There is a little grammatical error in the last sentence, I assume you meant:
```suggestion
     *           If the aria-label is just the title of the corresponding Button, it MUST be omitted.
```

matthiaskunkel: Just another tiny comment: "button" in this context is used as a general term and therefore written in lower-case letters. If it would be a proper name (like "Repository" in our UI), using a uppercase "B" would be fine. 

P.S.: I have to confess that our English language file needs a streamlining in the use of capital letters... It's on my list...

matthiaskunkel:
Jour Fixe, 26 JUL 2021 : We highly appreciate this suggestion and accept the PR for trunk and 7. In the future we will try to support such changes automatically by the UI service (in case the context is known for the specific problem).

PR #3509, opened 2021-07-19
Proposal to fix creation_timestamp

kergomard:
There would be a very short version to fix this: `strtotime($history_data[$version]['date'] ?? '01-01-1970')`. I think this is a bad idea though, so I chose the long way around. The ternary operator looked unreadable to me in this context, too.


chfsx (approved):
Thanks a lot, makes absolutely sense to go the longer way here.

chfsx:
Cherry-picked to release_7 as well

PR #3513, opened 2021-07-20
Feature/php8 compatibility/services copage verification

mBeym:
Makes the php classes 
**Services/COPage/classes/class.ilPCVerification.php**
and
**Services/COPage/classes/class.ilPCVerificationGUI.php**

php 8.0 compatible.

### DIFF
--- a/Services/COPage/classes/class.ilPCVerification.php
+++ b/Services/COPage/classes/class.ilPCVerification.php
@@ -1,27 +1,18 @@
-<?php
+<?php declare(strict_types=1);
 
 /* Copyright (c) 1998-2021 ILIAS open source, GPLv3, see LICENSE */
 
 /**
  * Class ilPCVerification
- *
  * Verification content object (see ILIAS DTD)
- *
  * @author Jörg Lützenkirchen <luetzenkirchen@leifos.com>
  */
 class ilPCVerification extends ilPageContent
 {
-    /**
-     * @var ilObjUser
-     */
-    protected $user;
+    private php4DOMElement $vrfc_node;
+    protected ilObjUser $user;
 
-    public $dom;
-
-    /**
-    * Init page content component.
-    */
-    public function init()
+    public function init() : void
     {
         global $DIC;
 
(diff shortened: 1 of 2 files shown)


Services/COPage/classes/class.ilPCVerification.php
```diff
@@ -11,17 +11,17 @@
  */
 class ilPCVerification extends ilPageContent
 {
+    private php4DOMElement $vrfc_node;
+    protected ilObjUser$user;
     /**
-     * @var ilObjUser
+     * @var php4DOMDocument
```
mjansenDatabay: IMO this property can be completely deleted because it is derive from `ilPageContent`.

Services/COPage/classes/class.ilPCVerification.php
```diff
@@ -44,7 +45,7 @@ public function setNode($a_node)
     * @param	object	$a_pg_obj		Page Object
     * @param	string	$a_hier_id		Hierarchical ID
     */
-    public function create(&$a_pg_obj, $a_hier_id, $a_pc_id = "")
+    public function create(object $a_pg_obj, string $a_hier_id, $a_pc_id = "") : void
```
mjansenDatabay: The 3rd parameter still has no typehint.

mjansenDatabay: Could you specifiy the first type more precisely? Maybe you get an idea if you execute this code while testing it in the user interface. You could use `Xdebug`.

Services/COPage/classes/class.ilPCVerification.php
```diff
@@ -11,17 +11,17 @@
  */
 class ilPCVerification extends ilPageContent
 {
+    private php4DOMElement $vrfc_node;
+    protected ilObjUser$user;
```
mjansenDatabay: A space is missing here.

mjansenDatabay (changes requested):
Thx @mBeym  for this PR!

`ilPCVerificationGUI`:

- [x] Please use/add strict comparisons for `in_array`.

In General:

- [x] Please replace the `array()` syntax with `[]`.
- [x] Please apply the PhpStorm code style formatting (https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/docs/development/code-style-configs/php-storm.xml) first and the PHP-CS-FIXER formatting afterwards
- [x] Please remove the `PHPDoc` sections if they do not provide any added value


Please ask the following questions:

- [x] Have you tested this by inserting an achieved certificate (no matter for which repository object, course, test, etc.) into a portfolio page (has to be globally enabled)? If you have any issues with testing this with PHP 8 (because the code of other maintainers is not ready, yet) please test it at least with PHP 7.4.

Best regards,
@mjansenDatabay

mjansenDatabay (approved):

PR #3518, opened 2021-07-26
fixed bug #31118: Permanent Link to Data Collection fails if only vis…

alex40724:
See
https://mantis.ilias.de/view.php?id=31118


theodortruffer:
Solid fix, thanks a lot! I'll cherry-pick it for 7 and trunk.

PR #3529, opened 2021-07-29
[FIX] 0031033: Update fails from ILIAS 5.3 to 5.4

chfsx:
This fixes Mantis https://mantis.ilias.de/view.php?id=31033


chfsx:
Hi @mstuder 
This Bug (31033) blocks updates and also new installations of trunk (and other releases), which we need for development and especially the PHP8 project. Could you please merge and downpatch to older releases as well the next days? we (TB) will merge this on friday if there is no feedback from you, but without any further responsibility. Thanks for your feedback and action!

chfsx:
unfortunately we didn't get any feedback, therefore we merge this without taking any further responsibility.

PR #3531, opened 2021-07-29
Global Template, double head Action

Amstutz:
While working on a11y issues with automated HTML Validators (W3c Online in Google Chrome in this case) I observed, that one ID is given twice: il_head_action, which causes the tool to display an error on almost any page. This is caused due to rendering the respective section in the template twice. I am not sure why this is done, however, IMO we can just remove the touch block. 

**Important**: I am not sure if this is the way to go. It is somewhat hard for me to completely understand the rendering logic in tpl.amd_content.html. E.g. I guess it could be possible, that this touch block statement is needed in some (edge) cases. So please check this rather throughly. 

If possible I would like to merge this down to 7 as well, if fine, give me a short +1. 

See: https://mantis.ilias.de/view.php?id=31156

### DIFF
--- a/Modules/DataCollection/classes/class.ilDataCollectionGlobalTemplate.php
+++ b/Modules/DataCollection/classes/class.ilDataCollectionGlobalTemplate.php
@@ -885,7 +885,6 @@ class ilDataCollectionGlobalTemplate implements ilGlobalTemplateInterface
             $this->setCurrentBlock("head_action_inner");
             $this->setVariable("HEAD_ACTION", $header);
             $this->parseCurrentBlock();
-            $this->touchBlock("head_action");
         }
 
         if (count((array) $this->title_alerts)) {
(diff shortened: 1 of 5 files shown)


klees:
I don't know why this is there or if it will cause any breakage to remove this, but this seems to be the commit that introduced that line:

https://github.com/ILIAS-eLearning/ILIAS/commit/64acf2de080ee933b7cd7714aea4bf9bce675041

chfsx (approved):

Amstutz:
@alex40724 : any feedback on this. The resulting issue is minor, however it shows up in every a11y test/automated test/audit we perform.

Amstutz:
@alex40724 : any feedback on this? This again popped up in an audit. The resulting issue is minor, however it shows up in every a11y test/automated test/audit we perform. If your fine with the approach I would resolve the conflicts and merge into 7 and trunk. However, I still have a hard time understanding possible side effects.

Amstutz:
The Materna Audit is on the horizon for November, this really needs to be fixed in core until then, @alex40724 any update?

alex40724:
The touchBlock line is 10 years old, however the error appears since ILIAS 6. At least I was not able to reproduce this on ILIAS 5.4.x.

Actually I lost track on the changes in the global template since 6. If the head actions still work for you, you can commit these changes. This might be even the cause for the issues with comments/notes modals, that "sometimes" fail to load content.

Amstutz:
@akill: Thx for the Feedback and for the time you took looking into this. I will do some re-testing and resolve all conflicts. If all is fine, Ill merge.

Amstutz:
Merged this one, same content: https://github.com/ILIAS-eLearning/ILIAS/pull/3731

PR #3532, opened 2021-07-30
Added additional check if style doesn't exist for the user which will then revert to the default style.

dmundra:
This fixes the bug https://mantis.ilias.de/view.php?id=31078 where we discovered an error if the style and skin is deleted for user and the default is not 'delos'


klees:
Hi @dmundra,

thanks a lot for looking into this. This problem is really annoying.

Best regards!

Amstutz:
Kudos and thx for the fix @dmundra !

Amstutz:
Since this mostly concerns the System Styles I took the liberty to merge this and push to 7 and trunk as well. I hope this is fine.

dmundra:
Thanks, @Amstutz!

PR #3533, opened 2021-08-02
[FIX] recursion in KeyValueAcess

chfsx:
There is an infinite recursion in KeyValueAccess introduced by me. This PR fixes the recursion and prevents it with a test in future ;)


klees:
Hi @chfsx,

thanks for looking into this. This fixes the infinite recursion for sure but tbh I'm still confused why we want or need that `KeyValueAccess`-class at all. You will need to explain this (again) sometimes =)

Best regards!

PR #3535, opened 2021-08-03
UI, Drilldown: refactor drilldown menu

nhaagen:
This refactors the (up to now unused) UI Drilldown.

Screenshot from 2021-08-03 15-35-44
Screenshot from 2021-08-03 15-36-13
Screenshot from 2021-08-03 15-36-45

- use strictly ul-trees
- add header-section with title and backlink
- simplify rendering and example, remove node-copies in JS
- fix tests for new rendering
- LESS backnav and hr display states
- remove labeledMenu-interface
- remove mutator for label
- typehint label to string


klees (changes requested):
Hi @nhaagen,

my 2 cents below, since I was involved here I ask @Amstutz for a review as well:

* [ ] Please don't [uncomment code](https://github.com/ILIAS-eLearning/ILIAS/pull/3535/files#diff-6a3ae0cff1d5107152b49cc1bba5cc6b29686b9cc5c7f0c731021f5a224f0e98R68). Just remove it. Git will keep it for us.
* [ ] Can we remove that [backlink signal](https://github.com/ILIAS-eLearning/ILIAS/pull/3535/files#diff-f44539e7d12afb4817f8f40dcd7e2ed485f62c3aaf6279f3796781fdc05d5eabR19) in favour of an js only mechanism or at least internalize it to the UI framework? I don't think we want our consumers to hook into that signal. Or do we want that?
* [ ]  Finally: It would be awesome to have some tests for the JS.

As discussed with Ferdinand, this PR shows opportunities to improve the less code:
* extract some comonalities between buttons in slates and the drilldown
* rectify that situation with the back-glyph, where we hide that glyph via css and replace it with another

I would prefer to not do this here, since this change is big enough IMO.

Best regards!

nhaagen:
Hi @klees ,
I removed the signal from the interface, but kept using it in the implementation.
We should think about pulling adhoc-signals from the renderer - equally to the UIFactory - for those internal bindings; there is a similar occasion in filters, I think.

With the JS, there probably is still room for improvement, especially on the renderer, but the general direction is what you imagined, right?

Best regards!

nhaagen:
Last update shows outline around currently focused element to improve visibility; also, aria-expanded is used to tell menulevels and item-buttons/links apart. Finally, focus is set explicitely to the first element of the current list.

Screenshot from 2021-08-17 11-13-40
Screenshot from 2021-08-17 11-14-29

Amstutz:
Hi @nhaagen

Thx a lot for refactoring this component. 

At everybody, note that we checked this component with our accessibility consultant, @nhaagen took care of the main issues found there already, namely the visibility of the focus and the auto focus to the firs element of the list. 

Special thx for using the new redux-like js pattern along with rollup and tests. 

Answer the following questions:
- [ ] You introduce code for rendering the chevron-left glyph in the Bulky Button, why is this necessary? Same Question for chevron right.

Please do:
- [x] Translate all content in the example to english
- [x] Avoid introducing new CSS numeric values for padding or margin hardcoded into your less. Check if you can use a global variable. If not introduce a global one or at least a local one. E.g. 48px is used multiple times, use a variable there.

The interface to be discussed in the next JF is good to go from my point of view.

Thx

matthiaskunkel:
Jour Fixe, 23 AUG 2021 : We highly appreciate this suggestion and accept the PR for trunk. Please merge.

src/UI/templates/js/Menu/dist/drilldown.js
```diff
@@ -0,0 +1,202 @@
+var drilldown = function(model, mapping) {
```
mjansenDatabay: Doesn't this polute the global namespace?

src/UI/templates/js/Menu/dist/drilldown.js
```diff
@@ -0,0 +1,202 @@
...
+		mapping.setHeaderBacknav(current.parent != null);
+	},
+
+	public_interface = {
+		init: init
+    };
+    return public_interface;
+};
+
+var ddmodel = function() {
```
mjansenDatabay: Doesn't this polute the global namespace?

src/UI/templates/js/Menu/dist/drilldown.js
```diff
@@ -0,0 +1,202 @@
...
+        }
+    },
+
+    public_interface = {
+        actions : actions
+    };
+    return public_interface;
+};
+
+var ddmapping = function() {
```
mjansenDatabay: Doesn't this polute the global namespace?

nhaagen:
Hi @mjansenDatabay and @Amstutz ,

I switched the JS-packing to IIFE; this looks like the much cleaner solution. 
Probably this might also improve the Mainbar's JS - I'll have a look there, too.

Another commonality with the Mainbar is the CSS substituting left-glyph with the chevron char, or adding right-arrows to the buttons. 
There are similar to equal tweaks in mainbar.less and slate.less, and I agree very much that those look awkward and somewhat out of place.

The solution, however, should be on a much lower level, I think, i.e. the glyph itself (do we need a new one?) and maybe the bulky-button with an additional option to render a right-/down-arrow.
I promise to digg a little deeper here, but also suggest to separate this issue from the current PR. Is this OK with you?

As for numerics in the less, I tried to reduce fixed settings as well as to use variables.

Best regards, 
Nils

Amstutz:
Thx @nhaagen for this final enhancement. Also thx @mjansenDatabay for checking into this and pointing out the scoping issue in js. I completely agree with the left-gyph cause. Thx for offering support here.

IMO this is ready to be merged. Thx a lot to everybody involved!

PR #3536, opened 2021-08-04
User/Style: Fix accessing undefined array key

mjansenDatabay:
This PR fixes a trivial PHP 8 issue caused by https://github.com/ILIAS-eLearning/ILIAS/commit/aa3fbdf86e9f632729c98aa0d82b86a9da44c0bd


Amstutz:
Thx a lot for the catch and sry, for not catching this myself.

PR #3538, opened 2021-08-05
0031173: Too many open files: close existing curl handles

smeyer-ilias:
This fixes an issue when trying to read the ecs event queue. 
See mantis #000031173 for further information


pascalseeland:
Thanks

PR #3542, opened 2021-08-09
[FIX] 0031174: Lucene Suche indiziert keine PDF Dateien

chfsx:
Hi @smeyer-ilias 
as mentioned in https://mantis.ilias.de/view.php?id=31174 I tried to fix the issue that files in storage v2 are not indexed (only files in fsv1). I implemented the creation of the path to the file in the query, don't know if this is a good idea (performance?). otherwise this would need another `FileObjectPathCreator`. Could you have a look at this and merge at will? Thanks a lot!


smeyer-ilias:
Thank @chfsx 
I will merge this to the trunk  and compile the sources.

PR #3547, opened 2021-08-11
#0031207: Group assignments after initial login

smeyer-ilias:
The description in mantis https://mantis.ilias.de/view.php?id=31207


smeyer-ilias:
Wrong base

PR #3552, opened 2021-08-12
$this->ilClientIniFile need a fallback

padvincenzo:
if $ilClientIniFile !== null then $this->ilClientIniFile will be undefined, throwing errors later (in functions ``read`` and ``update``)

I don't really know if this is the best solution, another one can be to throw a specific exception.

### DIFF
--- a/Services/Logging/classes/error/class.ilLoggingErrorSettings.php
+++ b/Services/Logging/classes/error/class.ilLoggingErrorSettings.php
@@ -38,9 +38,7 @@ class ilLoggingErrorSettings
             $this->ilias_ini = $ilIliasIniFile;
         }
 
-        if ($ilClientIniFile !== null) {
-            $this->gClientIniFile = $ilClientIniFile;
-        }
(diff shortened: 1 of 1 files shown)


klees:
Hi @padvincenzo,

thanks for caring about ILIAS!

Did you already have a look into our [guideline for contributing](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/contributing.md)?

I'll close this PR, because I'm almost 100% sure that this won't lead to the desired results. What you are writing there seems to be just a contrieved way to say `$this->gClientIniFile = $ilClientIniFile`. Because: if `$ilClientFile` is not null, `$this->gClientFile` will be set to `$ilClientFile`. If `$ilClientFile` is null, `$this->gClientIniFile` will be set to null as well, which just was the value of `$ilClientFile` in the first place.

Either way, if later code assumes that `$this->gClientIniFile` indeed is not null and call methods on the property, this change won't help.

Best regards!

PR #3556, opened 2021-08-17
 UI: Filter Revision - Merge down to release 7

tfamula:
The Jour Fixe decided to merge down the Filter Revision to ILIAS 7 as an usability fix (see Mantis [30290](https://mantis.ilias.de/view.php?id=30290)). This PR includes all PRs which were made for the revision in trunk: #3333, #3390, #3392, #3557

PR #3557, opened 2021-08-17
Fix existing UI Filters after Filter Revision

tfamula:
Hello @alex40724 and @mjansenDatabay,
in this PR I fixed some usages of the UI Filter in `Services/Container` and `Services/Cron`, so that the Filter can work properly at these places after the revision was done in #3390.
Please note that I have done the same changes also for the down merge to release 7 which can be found in #3556.


mjansenDatabay (approved):
LGTM @tfamula !

alex40724 (approved):
Great! Thanks.

PR #3558, opened 2021-08-18
fixed missing method when throwing exception

theodortruffer:
Just a small syntax fix to show the correct error message.


klees (changes requested):
Hi @theodortruffer,

- [x] Use the `$previous` parameter of [`Exception`](https://www.php.net/manual/de/class.exception.php) when throwing the new exception. We would then keep the stack trace of the previous exception.

Best regards!

klees:
Merged to trunk.

PR #3565, opened 2021-08-22
Null array offset fixed

padvincenzo:
When ``soap_user_administration`` and/or ``soap_connect_timeout`` and/or ``rpc_server_port`` are not set in the config setup file, PHP prints a warning:

```php
PHP Notice:  Trying to access array offset on value of type null in /var/www/html/ilias/Services/WebServices/classes/Setup/class.ilWebServicesSetupAgent.php on line 46
PHP Notice:  Trying to access array offset on value of type null in /var/www/html/ilias/Services/WebServices/classes/Setup/class.ilWebServicesSetupAgent.php on line 48
PHP Notice:  Trying to access array offset on value of type null in /var/www/html/ilias/Services/WebServices/classes/Setup/class.ilWebServicesSetupAgent.php on line 50
```

The problem is that casting has the precedence, not the ``??`` operator. Solved by adding parentheses.


smeyer-ilias:
Thanks for fixing this.

PR #3566, opened 2021-08-23
LearningSequence: 30836 refactor _goto function based on hotfix from mjansen

daniwe4:
Thanks @mjansen for your preliminary work.


klees:
Picked to release_6 and release_7.

PR #3567, opened 2021-08-24
Refinery: refactor the to()->Transformations into Constraints

daniwe4:


klees:
Hi @mjansenDatabay,

maybe you can have a look. I checked this as @daniwe4's teamlead but as a coordinator I don't want to greenlight stuff from my colleagues.

Thanks!

mjansenDatabay:
Thx for this PR @klees / @daniwe4 

Technically this looks good to me, but I am missing the `Why?` :-). What is the intention for this change, why only the `to` group? Could you please add some explanations?

klees:
Hi @mjansenDatabay!

Yes of course we can =)

We stumble upon this in the context of the PHP 8 improvements in the UI framework. The Input Fields might have a constraint that is used when a field is "required". We hint for `Constraint` there. Recall that a `Constraint` is a special transformation, that does not touch the actual value, but might create an error upon inspection. Which is exactly what man of the `to`-transformations in the `Refinery` do. Unfortunately, we were missing the appropriate interface implementation there, which is just what this PR adds.

Hope this helps!

mjansenDatabay:
Thx @klees , this helped a lot to understand the purpose :-).

PR #3572, opened 2021-08-25
BT-31095: Added legend and css rules for it to search fieldset

M-Barz:
Link to Mantis-Report:
https://mantis.ilias.de/view.php?id=31095


smeyer-ilias (commented):
Are the changes templates/default/delos.css required for this fix. 
If not, please remove them from the PR.

smeyer-ilias (approved):
My comment was nonsense. 
Obviously they are required.

Amstutz:
Note, there is an alternative solution with using the sr-only class, making this only visible for screen-readers. For me, both solutions are fine. 

@M-Barz : Please clean up the css here to only include the changes related to your changes in less. maybe a new lessc version is required...

M-Barz:
@Amstutz Yeah, it turns out I was quite a bit behind on lessc versions. This should be a lot cleaner now.

Since Stefan is on vacation for around 3 weeks, I could merge this myself if that's alright. Only if no further changes are requested of course.

Amstutz:
@M-Barz and @Amstutz had a short skype session today. We decided to pick this version over the other one, since here we have an additional benefit for all users with vision.

@M-Barz thx a lot for the quick fix and the PR. I will pick this to trunk as well.

PR #3573, opened 2021-08-25
Draft bugfix for mantis 31277 file inline download

rsheer:
This is a short draft on how mantis issue [31277 ](https://mantis.ilias.de/view.php?id=31277) could be fixed.

I assume there is a way better / cleaner solution since the `isInline`-Method is marked as deprecated in `ilObjFileImplementationLegacy`. It was just the nearest solution before the condition if consume should use `->download()` or `->inline()`.


chfsx:
Hi @rsheer and thanks for the PR, I modified this a little bit and committed today in release_7 and trunk

PR #3580, opened 2021-08-30
UI/Input: 31271, render duration group before generic group

nhaagen:
https://mantis.ilias.de/view.php?id=31271

group derivates must accounted for before rendering generic group.


Amstutz:
Thx for fixing. Note I added the missing duration rendering unit test to prevent this from happen in the future. plz, when fixing issues, always try to add/adapt a unit test to prevent the error from happening again.

PR #3582, opened 2021-08-30
UI/Input: 31268, relay transformation to group-inputs

nhaagen:
instead of setting a new/separate to self

https://mantis.ilias.de/view.php?id=31268


Amstutz:
Thx a lot!

Amstutz:
picked to 7 as well.

PR #3584, opened 2021-08-30
Fix TinyMCE active controll and TinyMCE5 styling

iszmais:
https://mantis.ilias.de/view.php?id=31151


mjansenDatabay:
@maxbecker This has to be cherry-picked to `trunk` if approved.

PR #3586, opened 2021-08-31
Setup: List achievable objectives for 'achieve' command

mBeym:
Adds a list command to the cli `achieve` command.

The command lists all available objectives.

How to use/call:
`php setup/setup.php achieve --list`
OR
`php setup/setup.php achieve`

If no options or arguments are passed to the **achieve** command, all available objectives will be listed.

ilias-cli-achieve-list-command

Discussion: https://mantis.ilias.de/view.php?id=30940#c75718


klees:
Hi @mBeym,

thanks a lot for looking into this, this feature was definitely missing. I'm positive that your changes should work as is, but I think that we should do a little cleanup as well to keep the code tidy. Most of my change requests are concerned with this.

- [x] Clean Up: Please clean up `AchieveCommand::execute` a little. The code you added in lines [77 - 101](https://github.com/ILIAS-eLearning/ILIAS/pull/3586/files#diff-da8b12ed92ce4f5a6ee374ef4000027833a580dd8eb2fdabcf927fd8a4e4e9e4R77) could go into a method `AchieveCommand::shouldListAvailableObjectives` and most likely will be shorter with early return pattern then. Move the code for achieving a single objective to a single method `AchieveCommand::achieveObjective`, just like you introduced a new method for listing the objectives.
- [x] Sub Title: Please remove the subtitles for the components in the output ("common", "uicore"). Having this might be a good idea some time in the future, but for now, there is more noise (the titles) than signal (the named objectives). Just list the objectives with the complete name ("common.registerNICKey"), ordered alphanumerically.

- [x] Unrelated Changes: Please do not include unrelated changes (like [this one](https://github.com/ILIAS-eLearning/ILIAS/pull/3586/files#diff-7b01c8b4c320cb9b967c0a32cee61c2cb46d45c702870bf582f1061d62124853L31) or [this one](https://github.com/ILIAS-eLearning/ILIAS/pull/3586/files#diff-d9ea208916e9319dcce9286abaea57b271784e58a37d55d8b6a0ff5aa5da25a5L94)) in PRs. Looks a little as if some auto-formatter with stronger configs then ILIAS' defaults fired... Please just use the [ILIAS coding style](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/coding-style.md) for auto formatting other peoples ILIAS code, nothing more, nothing less.
- [x] Wording: The initial wording for objectives that can be achieved explicitly was "named objectives". You now call them "achievable objectives" in the new code. For one, we should not use two different wordings for the same thing. Second, while we might want to improve the wording "named objectives", "achievable objectives" most likely is not the way to go. After all, many other objectives might be achievable as well, but just have no name to trigger them via the `achieve` command. Also, not every command that gets listed via the new command would be actually achievable always, e.g. the NIC server might not be reachable from the current installation. This won't be checked in advance. So the wording promises something it cannot keep. We should stick to "named objectives" (or propose something better). This item here most likely will be obsolete when you implemented the next one, but one concrete location to be changed is the [output to the user](https://github.com/ILIAS-eLearning/ILIAS/pull/3586/files#diff-da8b12ed92ce4f5a6ee374ef4000027833a580dd8eb2fdabcf927fd8a4e4e9e4R144). Please use "List available objectives:" instead.
- [x] Interface: Please change the interface to the "named objectives" functionality. Currently, there is one method "Agent::getNamedObjective" that can be called by giving an objective name. You added a method to list the available names. Please switch to an interface like we use for migrations: `function getNamedObjectives(Config $config = null) : array`. This method is supposed to return all named objectives for a certain agent, where the names are the keys in the array.
- [ ] HasNoNamedObjective Trait: There is a trait `ILIAS\Setup\Agent\HasNoNamedObjective` that all the classes use that have no named objectives. Use that trait for an empty implementation of the new `getNamedObjectives` and you will lose a lot of added lines in this PR.

Just contact me via Skype if you feel that we should discuss the change requests, I'm not sure if the level of detail and explanations is sufficient in my written items. @mjansenDatabay has my handle.

Best regards!

mBeym:
@klees Should be all done.

(I couldn't check the boxes)

I had to name the functions handling the commands (in this case the --list and objective) differently   
as there was already a function called 'achieveObjective' from ObjectiveHelper.
So i named them both executeXYZ.

mjansenDatabay:
@mBeym There are still a lot of obsolete changes in many files.

e.g.

- `ILIAS\Setup\Config;`
- Removed spaces / Some of them like here `array_walk($data["memcached_nodes"], function (array $node) use ($settings) : void {` might be required because of the PSR-2 checks
- Removed `\` from namespaces like `use \ILIAS\Refinery\Transformation;`

mBeym:
@mjansenDatabay 

Should be done.

Services/UICore/classes/Setup/class.ilUICoreSetupAgent.php
```diff
@@ -20,16 +20,19 @@ public function __construct()
...
-            "reloadCtrlStructure" => new ObjectiveCollection(
-                "Reload Control Structure of ILIAS",
-                false,
-                new \ilCtrlStructureStoredObjective(
-                    new \ilCtrlStructureReader()
-                ),
-                new \ilComponentDefinitionsStoredObjective(false)
-            ),
+            new Setup\NamedObjective("reloadCtrlStructure", "Reload Control Structure of ILIAS",
+                static function () {
```
mjansenDatabay: @mBeym You can define a return type for the closure.

setup/classes/class.ilSetupAgent.php
```diff
@@ -155,11 +153,20 @@ public function getNamedObjective(string $name, Setup\Config $config = null) : S
     public function getNamedObjectives(?Config $config = null) : array
     {
         return [
-            "registerNICKey" => new ObjectiveCollection(
-                "Register NIC key",
-                false,
-                new ilNICKeyRegisteredObjective($config),
-            )
+            new Setup\NamedObjective("registerNICKey", "Register NIC key",
+                static function () use ($config) {
```
mjansenDatabay: @mBeym You can define a return type for the closure.

src/Setup/NamedObjective.php
```diff
@@ -0,0 +1,50 @@
...
+    private Closure $objectiveCollectionClosure;
+
+    public function __construct(string $cmd, string $description, Closure $objectiveCollectionClosure)
+    {
+        $this->cmd = $cmd;
+        $this->description = $description;
+        $this->objectiveCollectionClosure = $objectiveCollectionClosure;
+    }
+
+    /**
```
mjansenDatabay: Please remove obsolete `PhpDoc` comments.

src/Setup/NamedObjective.php
```diff
@@ -0,0 +1,50 @@
...
+     * @return Closure
+     */
+    public function getObjectiveCollectionClosure() : Closure
+    {
+        return $this->objectiveCollectionClosure;
+    }
+}
```
mjansenDatabay: Please add a blank line here.

mjansenDatabay (changes requested):

klees:
@mBeym @mjansenDatabay Please ping me when this is ready. Thx!

mjansenDatabay (approved):

mjansenDatabay:
@klees I think this is ready.

@mBeym introduced a new class `Setup\NamedObjective` to encapsulate the `command`, `label` and `closure` (for lazy execution) for the `ObjectiveCollection`.

Reason: In our last call we decided to get rid of redundant objective instantiation we used in `getNamedObjective` and `getAchievableObjective` (former state of this PR). We then (after our call) figured out that we might run into problems if `NULL` is passed as `$config`.

klees:
Hi @mBeym, @mjansenDatabay,

just want to check in quickly about the status here. My current status is: I had a talk with @mjansenDatabay about some problems this still has, he took notes and wanted to check back in with @mBeym. After some changes you wanted to check back in with me.

Did I miss anything or is this your current state as well?
Best regards!

mjansenDatabay:
@klees IMO @mBeym addressed all the things we discussed:

- [x] The command name should be defined as an `Array Key`
- [x] `getNamedObjective` (singular) should be removed
- [x] Further unit tests should be added, especially to verify sorting works in `AgentCollection::getNamedObjectives`
- [x] The class `NamedObjective` should be replaced, we decided to use `ObjectiveConstructor` with a `create` method.

Is there anything we did no address?

klees:
Hi @mBeym, hi @mjansenDatabay,

I fixed the remaining problems in this PR by myself. Please have a look into (some of) the additional commits that I have made:

* the type hint on `ObjectiveConstructor::create` was too tight: https://github.com/ILIAS-eLearning/ILIAS/pull/3586/commits/e9db687423291c1f84fca6dab8012fc9f977d313
* tests didn't run: https://github.com/ILIAS-eLearning/ILIAS/pull/3586/commits/071385b0ec6acf328c513491645b6836a42c4844
* test for UICore was in the wrong location: https://github.com/ILIAS-eLearning/ILIAS/pull/3586/commits/247593efd5042ddaabfaa84e920982c516943c75
* nested config for `AgentCollection` was not respected: https://github.com/ILIAS-eLearning/ILIAS/pull/3586/commits/005dc615b479fa0b72c0b4538b19ffdac6c0d182

I removed some of the tests in `AchieveCommandTest` because I don't have any time left (now) to fix them.

Thanks for figuring out how this could work properly, this looks good now IMO.

Best regards!

mjansenDatabay:
@klees Thx for the fixes and your efforts.

But: Now, the command does not work at all :-). Maybe we'll have to check for `NullConfig` in `\ILIAS\Setup\AgentCollection::getNamedObjectives`.

klees:
Should be better now =)

PR #3589, opened 2021-09-02
UI: Add CollapseHorizontalGlyph for use in Mainbar and Drilldown.

nhaagen:
also add a glyph-like span to menu-buttons.

Screenshot from 2021-09-02 10-13-52

Screenshot from 2021-09-02 10-14-57

Dear colleagues,
this suggests a collapse horizontal glyph. 
We already have "back" (which actually triggers navigation - it is used incorrectly in the mainbar slate)
and "collapse" (which points downwards - in a drilldown you would spontaneously expect something pointing 
upwards or left).

This PR also covers implementation in Mainbar and Drilldown.

There still is some difference in the way Buttons are styled in the Mainbar Slates to the Drilldown; my colleague @catenglaender suggests to adapt the button-styling from the Drilldown, since it offers easier ways to customize vertical allignment of the symbol and (multiline-)text, but this is another round.

Please kindly consider, 
cheers, 
Nils


Amstutz (approved):
Thx for the proposal

I am definitely see the use cases for this new Glyph Component. Note, there is a further one in the notification slate, when BG Tasks are expanded. Kudos and big thx. 

I did propase the name "Collapse Horizontal" in lack of a better idea. However, I am not completely sold on it. So if anybody has a better suggestion, would be warmly welcome.

Nevertheless, IMO this is good to go for a JF decision on the interface. @klees feel free to attach the JF sticker if fine with you as soon as you are back.

looking forward to the discussion.

matthiaskunkel:
Jour Fixe, 04 OCT 2021: We highly appreciate this suggestion and accept the PR for trunk. Please check if this new glyph is clearly distinguishable from the glyph Next / Back to prevent problems of interpretation.

PR #3591, opened 2021-09-02
Bugfix: Missing direct notification for new posts

playamongthestars:


mjansenDatabay:
@nmatuschek Could you please explain (Skype call) a bit what you changed here?

Furthermore I noticed this class during review:

```php
class ilForumNotificationEvents
{
    public const UPDATED = 1;
    public const CENSORED = 2;
    public const UNCENSORED = 4;
    public const POST_DELETED = 8;
    public const THREAD_DELETED = 16;
    
    protected bool $is_notify_updated_enabled = false;
    protected bool $is_notify_censored_enabled = false;
    protected bool $is_notify_post_deleted_enabled = false;
    protected bool $is_notify_thread_deleted_enabled = false;
    
    protected $interested_events = 0;
    
    public function __construct()
    {
    }
}
```

If we do not need the properties here and only the constants:
- [x] Please remove all properties and transform the `class` into an `interface`. Maybe we could migrate this interface and use an `Enum` approach with PHP 8.1.

PR #3592, opened 2021-09-02
Added missing property in ObjectDataCache

PurHur:
Removes a undefined property notice.

The lines are just copied from the trunk: (i removed the type for extra compatiblity ;) )
https://github.com/ILIAS-eLearning/ILIAS/commit/4e72c3d6aee5c0b99be3953a46ee753adae1924a


alex40724:
Thanks!

PR #3593, opened 2021-09-02
LSO: add successmessage in admingui

nhaagen:
There was no message after saving the settings.


klees:
Cherry-picked to trunk.

PR #3594, opened 2021-09-02
Bugfix 31290: regression in AreaInput introduced with tinymce5

jeph864:
According to https://github.com/ILIAS-eLearning/ILIAS/pull/3088#issuecomment-907789238, the update  tinymce to v5 introduced a regression! This PR introduces a patch to solve this

https://mantis.ilias.de/view.php?id=31290#c75945


kergomard:
Thank you very much @jeph864 , @mjansenDatabay , and @sunweaver !

PR #3598, opened 2021-09-06
Use bool false instead of empty string

pascalseeland:
This fixes an issue with the typed ilPDO


mjansenDatabay:
Thx @pascalseeland

PR #3599, opened 2021-09-07
GlobalCache/Setup: Expose status of global cache service

mjansenDatabay:
Currently the `status` does not provide any explicit information whether the global cache service is active.

This PR adds an explicit `active` property to the status of the `ilGlobalCacheMetricsCollectedObjective`.


chfsx:
Thanks, I cherry-picked to trunk as well!

PR #3600, opened 2021-09-07
News/MembershipNotifications: Fix accessing array for ref_ids where a…

mjansenDatabay:
…ll children are trashed

Mantis Issue: https://mantis.ilias.de/view.php?id=31322


alex40724:
Thanks!

PR #3606, opened 2021-09-07
MainBar/Performance: Prevent DDoS-like HTTP requests when resizing wi…

mjansenDatabay:
... ndow

Mantis Issue: https://mantis.ilias.de/view.php?id=30610

### DIFF
--- a/src/UI/templates/js/MainControls/mainbar.js
+++ b/src/UI/templates/js/MainControls/mainbar.js
@@ -150,7 +150,7 @@ il.UI.maincontrols = il.UI.maincontrols || {};
 					mb.model.actions.disengageAll();
 				}
 				mb.model.actions.initMoreButton(amount);
-				mb.renderer.render(mb.model.getState());
+				mb.renderer.renderDeferred(mb.model.getState());
 			},
 			init_desktop = function(initially_active) {
 				var mb = il.UI.maincontrols.mainbar,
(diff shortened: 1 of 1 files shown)


Amstutz:
Nice one, looks good to me. @nhaagen can you look at this as well and give some feedback, if all lights green, we could merge this and push to 7 and trunk as well.

nhaagen:
Thank you,  good finding. However, I do not really agree with the solution.
While it's certainly ok to be less exited about rendering, the actual issue is reloading of slate-content, triggered by renderer.engage via this.getElement().trigger('in_view');
Maybe better wrap this one into something like

if(!this.getElement().hasClass('gs_triggered')){
    this.getElement().trigger('in_view');
    this.getElement().addClass('gs_triggered');
}

mjansenDatabay:
Even better if there is an easy fix like the one proposed by @nhaagen

klees:
I think we could do this (for trunk at least) anyway, because, as @nhaagen says, we don't need to be that excited on rendering when resizing the window. But we should only add that complexity if there actually is any perfomance hit from the actual rendering (besides the requests).

I agree with @nhaagen that this does not seem to fix the root cause of the initial problem, though.

Amstutz:
Thx a lot. See comment in #3610 for the explanation in the decision on this one.

PR #3607, opened 2021-09-07
LSO: 31326, tracking events 

nhaagen:
- use LP update instead of refresh 
- cache results from tree and references

https://mantis.ilias.de/view.php?id=31326

### DIFF
--- a/Modules/LearningSequence/classes/LearnerProgress/class.ilLSLPEventHandler.php
+++ b/Modules/LearningSequence/classes/LearnerProgress/class.ilLSLPEventHandler.php
@@ -19,6 +19,15 @@ class ilLSLPEventHandler
      */
     protected $lpstatus;
 
+    /**
+     * @var array
+     */
+    protected $cached_parent_lso = [];
+    /**
+     * @var array
+     */
+    protected $cached_refs_for_obj = [];
+
 
     public function __construct(
         ilTree $tree,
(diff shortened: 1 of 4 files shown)


Modules/LearningSequence/classes/LearnerProgress/class.ilLSLPEventHandler.php
```diff
@@ -46,12 +55,19 @@ public function updateLPForChildEvent(array $parameter)
...
      */
     protected function getParentLSO(int $child_ref_id)
+    {
+        if (!array_key_exists($child_ref_id, $this->cached_parent_lso)) {
+            $this->cached_parent_lso[$child_ref_id] = $this->getParentLSOByFullPath($child_ref_id);
+        }
+        return $this->cached_parent_lso[$child_ref_id];
+    }
+
+    private function getParentLSOByFullPath(int $child_ref_id)
```
mjansenDatabay: @nhaagen One more thought: AFAIK the LSO can only contain leaf nodes (meaning: non-containers)

So if the learning progress state of a user for an arbitrary object is updated, only the direct parent should be relevant when checking for a LSO. So maybe `\ilTree::getParentNodeData` is more appropriate as `\ilTree::getPathFull`.

klees:
Picked to release_7 and trunk.

PR #3608, opened 2021-09-08
Throw exception if file does not exist instead of returning null

pascalseeland:


Services/WebServices/RPC/lib/src/de/ilias/services/lucene/index/file/path/FileObjectPathCreator7.java
```diff
@@ -96,17 +82,17 @@ public File buildFile(CommandQueueElement el, ResultSet res)
 			if(file.exists() && file.canRead()) {
 				return file;
 			}
+			if(!file.exists()) {
+			  throw new PathCreatorException("Cannot find file: " + fullPath.toString());
+			}
+			if(!file.canRead()) {
+			  throw new PathCreatorException("Cannot read file: " + fullPath.toString());
+			}
 			return null;
```
mjansenDatabay:
```suggestion
			if (!file.exists()) {
			  throw new PathCreatorException("Cannot find file: " + fullPath.toString());
			}

			if (!file.canRead()) {
			  throw new PathCreatorException("Cannot read file: " + fullPath.toString());
			}

			return file;
```

mjansenDatabay:
If approved, we need a new *.jar file

smeyer-ilias:
Thanks for fixing this

PR #3609, opened 2021-09-08
UI: Mainbar, 30610 fix gs-call on resize

nhaagen:
https://mantis.ilias.de/view.php?id=30610,
followup to https://github.com/ILIAS-eLearning/ILIAS/pull/3606


nhaagen:
As @klees correctly said, the former commit actually supressed the in_view event; there might be components (now/plugins/future) relying on that. 
I kindly ask for comment on this one, introducing another event that is thrown only the first time the element comes into view.
If OK, I'll gladly elaborate and update 7/trunk.

Amstutz:
See comment in #3610

PR #3610, opened 2021-09-08
UI: Mainbar, check for state before gs-reloading contents; fixes 30610

nhaagen:
https://mantis.ilias.de/view.php?id=30610,
followup to https://github.com/ILIAS-eLearning/ILIAS/pull/3606

This one is also valid for 7, for 6 please see https://github.com/ILIAS-eLearning/ILIAS/pull/3609

### DIFF
--- a/src/UI/templates/js/MainControls/dist/mainbar.js
+++ b/src/UI/templates/js/MainControls/dist/mainbar.js
@@ -691,10 +691,16 @@ var renderer = function($) {
             return $('#' + this.html_id);
         },
         engage: function() {
-            this.getElement().addClass(css.engaged);
-            this.getElement().removeClass(css.disengaged);
-            this.getElement().trigger('in_view'); //this is most important for async loading of slates,
-                                                  //it triggers the GlobalScreen-Service.
+            var element = this.getElement(),
+                loaded = element.hasClass(css.engaged);
+
+            element.addClass(css.engaged);
+            element.removeClass(css.disengaged);
+
+            if(! loaded) {
+                element.trigger('in_view'); //this is most important for async loading of slates,
+                                            //it triggers the GlobalScreen-Service.
+            }
             if(il.UI.page.isSmallScreen() && il.UI.maincontrols.metabar) {
                 il.UI.maincontrols.metabar.disengageAll();
             }
--- a/src/UI/templates/js/MainControls/src/mainbar.renderer.js
+++ b/src/UI/templates/js/MainControls/src/mainbar.renderer.js
@@ -23,10 +23,16 @@ var renderer = function($) {
             return $('#' + this.html_id);
         },
         engage: function() {
-            this.getElement().addClass(css.engaged);
-            this.getElement().removeClass(css.disengaged);
-            this.getElement().trigger('in_view'); //this is most important for async loading of slates,
-                                                  //it triggers the GlobalScreen-Service.
+            var element = this.getElement(),
+                loaded = element.hasClass(css.engaged);
+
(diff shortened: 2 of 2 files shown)


klees:
Hi everyone,

I understand how and why this would fix some reloading issue, but I'm not very happy with the implementation of the patch. From my POV it looks as such:

Currently, the code of the UI framework provides two facilities, that are used here:

* an event "in_view" which is "is triggered when the slate 'comes into view', i.e. is being engaged." (according to the documentation of `Slate::appendOnInView`).
* an signal "Replace" which causes the the slate to be replaced by something returned from some server side URL

The Global Screen then uses these facilities to say: If the slate comes into view, replace it with the HTML returned by `gs_content.php?item=something`.

What now happens is that the according "in_view" event is triggered by the rendering code of the mainbar, which is called on resize events (to make buttons move from the main bar to [More], if applicable). This, in turn, triggers the replace signal, which then causes the url to be loaded again (and again, and again).

The previous fix (as still available here) attempted to fix this issue by fireing the "in_view" event only once. The problem here is that this effectively makes the naming of the event lie: instead being fired when the slate comes into the view, it only gets fired when the slate comes into the view *for the first time*. This is what I discussed with @nhaagen. This is something we should not do: lie with the naming of our stuff. Also, this patch could have broken code (in the core or plugins) already, because the interface promises something other than we would have delivered.

This insight led to the changes in [PR 3609](https://github.com/ILIAS-eLearning/ILIAS/pull/3609) (thanks @nhaagen), which fixes the problem by introducing a second event with the desired semantics: firing on the first time the slate comes into view.

This is much better, IMO, because we are not lying in the naming now and get the desired semantics.

But looking into the code again I'm not even sure that everything is as announced currently. Because: Why does this "in_view"-event even trigger when the window resizes only? This should not happen (mostly) because resizing a window does not open or close slates (only once we switch from desktop width to mobile width and back, maybe). The problem here IMHO is that the event is triggered in the rendering of the Main Bar, i.e. in the place where we apply the changes in the model to the DOM. This, probably, is the reason why this problem appears in the first place: When the window is resized, the complete current model status is rendered to the DOM.

This makes me think: The code for triggering the event is just in the wrong place. This is a matter of changes to the model, not changes to the rendered DOM. The event should then be called "OnEngage" and "OnFirstEngage" and we should check this fact and implement the triggering of the event accordingly.

So, probably, we should just remove the "InView"-event(s) completely in favour of the "Engage"-events, because, really, it is hard to find out if some DOM element actually "is in the view". What would this even mean for screenreaders? Second, we should adjust the triggering conditions in a way that actually make sure these events get triggered on changes in the model (and not the view).

I think this should lead to the desired performance improvements and improve the semantics of the component (and hence decrease probability of future bugs) at the same time.

What do you think? If anyone interested in discussing this in person, feel free to ask. This looks as if this is some interesting problem, because, basically, this seems to be caused by some naming on the interface of the UI component.

Best regards!

Amstutz:
Hi @nhaagen 
@klees and I discussed this morning this PR and https://mantis.ilias.de/view.php?id=30610 with #3606. We believe that this is somewhat better than option #3606, however not fully resolves the issue. See the explanation of @klees above. @klees might provide a solution to improve the situation for trunk in the near future. However, it resolves the issue in the ticket, which we both believe to be important. Therefore we merge this and close #3606. Nevertheless, thx a lot @mjansenDatabay for the contribution and bringing this to the table in the first place!

I just tested it on trunk. Work as expected and I did not see any negative side effect. Will be merged down to 7 and via #3609
to 6 as well.

Thx.

PR #3611, opened 2021-09-08
SETUP: fix 31311, increase field size for updatesteps start/finished

nhaagen:
https://mantis.ilias.de/view.php?id=31311


klees (changes requested):
Hi @nhaagen,

please move this to a db-update file in the new format and the according component. This should definitely be possible... (right?)

Best regards!

klees (approved):

klees:
Hi @chfsx,

this LGTM, but since this is in `Services/Database` i cannot really tell if this matches your idea of a proper organisation and naming of steps, agents, etc. in the service. Please have a look!

Thanks!

chfsx:
Thanks a lot!

PR #3615, opened 2021-09-10
UI/Inputs: tagify and errormessage for datetime (merges from trunk)

nhaagen:
The included commits are already in trunk, but they did not make it to 7.


klees:
This would allow us to get rid of bootstrap-tagsinput.

klees (approved):

Amstutz:
Normally we do not merge back mere improvements to released versions. If PRs should be ported back, plz create a ticket in mantis describing the issue or give some reasoning in the initial PR of why it should be back ported (e.g. also perf. or security could be a reason).

In this case, the improvement clearly tackles one of the above (contact me for more info), so I porting back is i.O.. Thx again @nhaagen for the effort.

PR #3618, opened 2021-09-10
Correct error in presentation of sc questions in survey

kergomard:
Fix for https://mantis.ilias.de/view.php?id=31350

This avoids that the readio-buttons of a second sc question are hidden when not in compact view.


alex40724:
Thanks for the PR! Merge to trunk.

PR #3620, opened 2021-09-12
fix misformatted language files de/en

Bigwern:
Saw this when I updated a installation to latest 7 release. Prevents Ilias from updating the languages from the files.


klees:
Thanks a lot.

PR #3622, opened 2021-09-15
LSO: constant LP polling, bypass player GUIs

nhaagen:
There are two relevant changes in this PR:
1st: polling learning progress is started instantly when reaching the (legacy-)view of an object, w/o the need to open the object's window by clicking "start". This also fixes interrupted polling for subsequent status changes, as e.g. in starting a survey (->in progress) and finishing it (->completed) - the latter is not found due to an reload caused by the former.

2nd: constructing the player GUI (class.ilObjLearningSequenceLearnerGUI.php) needs quite some information about contained items; this is not necessary for simply polling a LPstatus. The new ilObjLearningSequenceLPPollingGUI bypasses the player GUI. I also do not see a need for security checks, since the LP-check depends on the current user which is not injected via external parameter.

### DIFF
--- a/Modules/LearningSequence/classes/Player/class.ilObjLearningSequenceLPPollingGUI.php
+++ b/Modules/LearningSequence/classes/Player/class.ilObjLearningSequenceLPPollingGUI.php
@@ -1,0 +1,43 @@
+<?php declare(strict_types=1);
+
+class ilObjLearningSequenceLPPollingGUI
+{
+    const PARAM_LSO_LP_OBJID = LSUrlBuilder::PARAM_LSO_PARAMETER;
+
+    protected ilCtrl $ctrl;
+    protected int $current_user_id;
+    protected ilObjectDataCache $obj_data_cache;
+
+    public function __construct(
+        ilCtrl $ctrl,
+        int $current_user_id,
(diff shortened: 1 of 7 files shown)


Modules/LearningSequence/classes/Player/class.ilObjLearningSequenceLPPollingGUI.php
```diff
@@ -0,0 +1,43 @@
...
+                $this->getCurrentItemLearningProgress();
+                // no break
+            default:
+                throw new ilException("Command not supported: $cmd");
+        }
+    }
+    
+    protected function getCurrentItemLearningProgress()
+    {
+        $obj_id = (int) $_GET[self::PARAM_LSO_LP_OBJID];
```
klees, line 35: Please do not use $_GET but `retrieve` from the new HTTP wrapper instead. @daniwe4 may give hints.

klees (changes requested):
LGTM in general, please fix the issue mentioned in the comment.

Thanks!

daniwe4:
Please close this PR and use this https://github.com/ILIAS-eLearning/ILIAS/pull/3788 instead.

PR #3625, opened 2021-09-16
Search Form, a11y fix for #31095

Amstutz:
In our audits, issue https://mantis.ilias.de/view.php?id=31095 comes up every single time, the fix is rather small. Here a proposal on how to fix. 

If accepted plz also pick to trunk.


Amstutz:
Sry, I did not see that there is already another fix for this one: https://github.com/ILIAS-eLearning/ILIAS/pull/3572

3572 takes a slightly different approach by making the legend visible for everybody. I do not have a special preference. @smeyer-ilias the choice is yours.

PR #3627, opened 2021-09-16
Database: 31311 increase field size in correct script

klees:
Hi @chfsx,

this fixes [PR 3611](https://github.com/ILIAS-eLearning/ILIAS/pull/3611) by moving the change into the db_update-steps file. One cannot execute changes in the new step-files before this db-change was made. So the changes in the other PR can never be executed without manual intervention. Looks as if @nhaagen was right all along in the original PR by putting the db-updates in the old update step files. Sry.

Best regards!


klees:
I don't understand why the fix in the tests of the Test suddenly is required to make the tests run, but here we are.

mbecker-databay:
I have no explanation, either. But the fix looks good, tysm for it.

mbecker-databay (approved):

chfsx (approved):

PR #3630, opened 2021-09-17
Setup: add metric collector for db update steps

daniwe4:
Collect db update steps from different modules and services and add them to status command output.

### DIFF
--- a/Modules/StudyProgramme/classes/Setup/ilStudyProgrammeUpdateAgent.php
+++ b/Modules/StudyProgramme/classes/Setup/ilStudyProgrammeUpdateAgent.php
@@ -1,6 +1,8 @@
 <?php declare(strict_types=1);
 
 use ILIAS\Setup;
+use ILIAS\Setup\Objective;
+use ILIAS\Setup\Metrics;
 
 class ilStudyProgrammeUpdateAgent extends Setup\Agent\NullAgent
 {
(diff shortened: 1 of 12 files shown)


klees (changes requested):
Hi @daniwe4,

this looks fine already, but there is one change I still like to request. Let's use "Favor composition over inheritance" here and turn the `trait DatabaseStepHelper` into a full fledged `class DatabaseStepReader` that then is used by the two respective objectives. You can then write tests for the `DatabaseStepReader` once and simplify the tests for the objectives as well. This will turn out to be a lot simpler, I would expect.

Give me a call to discuss, as always.

Thanks!

klees:
Hi @chfsx,

this LGTM. Please pick this up as maintainer of the database =)

Thanks!

chfsx:
Hi @daniwe4 
Sorry for the late response!
I just checked your PR (and rebased it locally fist): while running a `php setup/cli.php update -y` I get errors like ` [ERROR] Call to undefined method ilMailDatabaseUpdateSteps::step_2()`. the class `ilMailDatabaseUpdateSteps` currently only has one step and the `step_2`does not exist. could you please have a look at this? thanks a lot!

daniwe4:
Hi @chfsx,
thank you for the hint. I hopefully fixed the problem.

chfsx:
Thanks a lot, works now like a charm :-)

PR #3631, opened 2021-09-17
UI Components, fix for Invalid href in Close Button Glyph, see #31369

Amstutz:
Glyphs in the Context Bulky Button currently render a href in span, which is not valid. 

See: https://mantis.ilias.de/view.php?id=31369

Note, we already agreed, that it in the longrun, it might be better, that glyphs should not get an action at all. This is a further indication, that this seems correct.

### DIFF
--- a/src/UI/Implementation/Component/Symbol/Glyph/ButtonContextRenderer.php
+++ b/src/UI/Implementation/Component/Symbol/Glyph/ButtonContextRenderer.php
@@ -4,10 +4,18 @@
 
 namespace ILIAS\UI\Implementation\Component\Symbol\Glyph;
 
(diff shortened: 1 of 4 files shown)


klees (approved):
Hi @Amstutz,

changes LGTM. Thanks for the little cleanup in the code.

Regarding the alleged root cause of the problem: The plot thickens =) Should we add that item to the roadmap? Basically, I think, the current usages of glyphs with action should be replaced by shy buttons with glyphs. We will need improvements in the `Button` hierarchy then, though. Shall we have a chat some time about this?

Best regards!

Amstutz:
Yes we should tackle this. ping me on skype as soon as you have some spare time.

PR #3632, opened 2021-09-17
Search, Duplicate attribute type in Search Form Button

Amstutz:
Attribute Type is listed twice, which leads to invalid html, see https://mantis.ilias.de/view.php?id=31371


M-Barz (approved):
LGTM!

PR #3633, opened 2021-09-17
Container: fix typehint problem in ContainerRenderer

daniwe4:


alex40724:
Thanks, seems to be fixed already.

PR #3637, opened 2021-09-20
RTE: PHP 8

mjansenDatabay:


mjansenDatabay:
The template class is used within the `node_modules` folder, which is not indexed by the `PhpStorm` search. I'll have to fix this class as well ...

PR #3639, opened 2021-09-21
Fixes: 0031377: Verständlichkeit der 'Top-Thema-Sortierung'

playamongthestars:


Modules/Forum/classes/class.ilForumProperties.php
```diff
@@ -614,4 +614,21 @@ public function setLpReqNumPostings(?int $lp_req_num_postings) : void
     {
         $this->lp_req_num_postings = $lp_req_num_postings;
     }
+    
+    public function countStickyThreads() : int
```
mjansenDatabay: - Please rename this to `getNumberOfStrickyThreads()`
- Maybe we should move this to `ilForum` or `ilObjForum`
- Can you please check the SQL with an `EXPLAIN` and verify, that this is **not** a query of type `ALL`.

mjansenDatabay (changes requested):

Modules/Forum/classes/class.ilForumProperties.php
playamongthestars: done

mjansenDatabay (approved):

PR #3640, opened 2021-09-21
Enabled Strict Mode per Default and fixed field order_nr of copg_pc_def

chfsx:
See: https://mantis.ilias.de/view.php?id=31292


klees (approved):

alex40724 (approved):
Thanks!

PR #3644, opened 2021-09-23
UI Components: Interface for Button/Minimize

iszmais:
This PR proposes to introduce the UI-Component for implementing Minimize Buttons

The Minimize Button is a visual, interactive area wich is capable of minimizing its related, surrounding content if clicked. 

A minimize Button can appear inside the top right corner of its related content but always left to the close button of the same content if it exists. It contains a square area with a simple graphic of minimisation. Otherwise its completly undercorated.
Further it contains an action wich can be triggered on left clicking the button. This interaction causes the related surrounding content of the minimize button to free its occupying space partialy or complete. Further in can cause logical reactions to this action like moving the minimized content into another area for later reaccess.

image

### DIFF
--- a/src/UI/Component/Button/Factory.php
+++ b/src/UI/Component/Button/Factory.php
@@ -160,7 +160,7 @@ interface Factory
      * ---
      * @return  \ILIAS\UI\Component\Button\Minimize
      */
-    public function minimize();
+    public function minimize() : Minimize;
 
     /**
      * ---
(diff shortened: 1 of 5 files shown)


iszmais:
@klees i would like present this on the upcoming JF if possible.
Therefore i would aprreciate if you could give me some feedback in advance to reduce the overhead here.

klees:
I just checked this with @Amstutz, we'll be happy to see you present this to the JF. Thanks!

matthiaskunkel:
Jour Fixe, 04 OCT 2021: We highly appreciate this suggestion and accept the PR for trunk.

iszmais:
@klees could you pls merge the approved PR? Thx.

PR #3651, opened 2021-09-25
0031404: ILServer does not finish createIndex

pascalseeland:
This PR contains a set of commits, which I did while trying to get createIndex to work. While the final issue was that the POI library does does not work as expect, which lead to the option to disable indexing doc file. While debugging, server other issues have been fixed
- The log4j lib was upgraded and the configuration was put in a file, making it adjustable
- The way objects indexed marked as finished was changed, as the current implementation was not thread safe
- Instead of querying the objects and inserting them one by one into the search_command_queue_table, this is now done using one query. This reduced the time of this step for file on our system from 21min to 17sec.
- A shutdown-hook was added, so there is no more need to call ilServer stop. 
- The leftovers from oracle have been removed
- pdfbox and poi have been upgrade to the latest version
- The ant build system has been replaced by maven. The main reason for this was, that there where already threads getting stuck as there was a missing dependency (commons collection 4) and that adding new binary deps to the git repo does not seam like a good idea

The best way to review the PR is probably the looking at the individual commits.


Amstutz:
Wow, awesome. Thx so much @pascalseeland . Very important and some love is definitely needed in this component. We are not able to get it up and running since roughly ILIAS 5.2 or 5.3. Looking forward to test it again with your changes. I am sure this will be highly appreciated, since it seems very hard to get dev. resources for the search.

Thanks so much and kudos.

klees:
Hi @smeyer-ilias,

we had a look into this PR in the Technical Board today and found that it will be open for a year in about a month, without any feedback from you yet. Please try to react to PRs within 21 days as [described in our guidelines](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/contributing.md#rules-for-maintainers-and-coordinators-assigned-to-prs).

Thanks!

smeyer-ilias:
As discussed, I'm closing this PR, but I'll still be taking over large parts for a new version of the Java server in the near future.
Many thanks for sharing your work!

Hufschmidt:
I know this was closed in favor of beeing implemented again later on, but it would be really nice if most of this could be incorporated into ILIAS7 or at least going forward be improved by ILIAS8/9 as we are encountering a few of the issues mentioned/fixed here.

Specifically e32938c34c166352284a62cd1a335da0f8f7c706, a23f9804d8973c22da3a9d7e5bb4ef82dc16ca12 and most likely b1d103709474815094b7dacb57a0841e1cc2b4bd

---

PS: Danke an @pascalseeland für deine Arbeit, sieht echt gut aus. :heart:

PR #3652, opened 2021-09-25
Tests/PHPUnit: Use XML config (instead of arguments) / Be more strict

mjansenDatabay:
This PR improves the unit test executions by ...

- ... using the XML configuration file instead of the arguments on `PHPunit` command execution
- ... ignoring the `PHPUnit` result cache file, which is useful for local test execution
- ... being more strict for:
  - warnings
  - deprecations
  - undesired output on test execution
  - risky tests  (which lead to https://github.com/ILIAS-eLearning/ILIAS/pull/3653)

Passing the `args` to the `PHPUnit` call is still possible when executing `CI\PHPUnit\runtests.sh`.

Nevertheless, the tests for this PR fail because of the `risky` test mentioned above, which is fixed by https://github.com/ILIAS-eLearning/ILIAS/pull/3653

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,4 @@ virtual-data
 override.php
 captainhook.config.json
 captainhook.local.json
+.phpunit.result.cache
--- a/CI/PHPUnit/phpunit.xml
+++ b/CI/PHPUnit/phpunit.xml
@@ -5,6 +5,10 @@
          cacheResultFile="../../.phpunit.result.cache"
          verbose="true"
          colors="true"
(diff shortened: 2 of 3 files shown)


klees:
Hi @mjansenDatabay!

Can I still use `--filter` for `CI\PHPUnit\runtests.sh` if we do this?

Best regards!

mjansenDatabay:
@klees Yes, you can still use it like `./CI/PHPUnit/run_tests.sh --filter 'ilBcryptPhpPasswordEncoderTest'`

klees (approved):

PR #3655, opened 2021-09-26
Error Handler: Remove PHP < 7.0.x version handling

mjansenDatabay:


mjansenDatabay:
@smeyer-ilias Any objections to merge this?

mjansenDatabay:
@smeyer-ilias 

I'll take the responsibility for that (which is easy in this case because it removes code never executed again :-)) and merge this.

PR #3656, opened 2021-09-26
Session Handler: Implement interface / Remove PHP version conditional…

mjansenDatabay:
…s / Remove obsolete register_shutdown_function

This PR ...

- ... removes several obsolete `version_compare` conditions for PHP 7.2
- ... adds an implementation of the `SessionHandlerInterface` interface to `ilSessionDBHandler`
- ... removes an explicit `register_shutdown_function('session_write_close');` call (**please double/cross check on review/testing**)
- ... uses `strict_types`
- ... formats `PHPDoc` comments

### DIFF
--- a/Services/Authentication/classes/class.ilSessionDBHandler.php
+++ b/Services/Authentication/classes/class.ilSessionDBHandler.php
@@ -1,119 +1,90 @@
-<?php
-
+<?php declare(strict_types=1);
 /* Copyright (c) 1998-2009 ILIAS open source, Extended GPL, see docs/LICENSE */
 
-include_once("./Services/Authentication/classes/class.ilSession.php");
-
 /**
-* Database Session Handling
-*
-* @module		inc.db_session_handler.php
-* @modulegroup	iliascore
-* @version		$Id: inc.db_session_handler.php 18894 2009-02-06 15:24:04Z akill $
-*/
-class ilSessionDBHandler
+ * Database Session Handling
+ * @module inc.db_session_handler.php
+ * @modulegroup iliascore
+ */
+class ilSessionDBHandler implements SessionHandlerInterface
 {
-    /*
-    * register callback functions
-    * session.save_handler must be 'user' or safe mode off to succeed
-    */
-    public function setSaveHandler()
+    /**
+     * Register callback functions
+     * session.save_handler must be 'user' or safe mode off to succeed
+     */
+    public function setSaveHandler() : bool
     {
-        // register save handler functions
         if (session_status() === PHP_SESSION_ACTIVE) {
             return true;
         }
 
-        if (ini_get("session.save_handler") == "user" || version_compare(PHP_VERSION, '7.2.0', '>=')) {
-            session_set_save_handler(
-                array($this, "open"),
-                array($this, "close"),
-                array($this, "read"),
-                array($this, "write"),
-                array($this, "destroy"),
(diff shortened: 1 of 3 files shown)


mjansenDatabay:
@smeyer-ilias Any objections to merge this?

Services/Authentication/classes/class.ilSessionDBHandler.php
```diff
@@ -1,119 +1,90 @@
...
-                array($this, "close"),
-                array($this, "read"),
-                array($this, "write"),
-                array($this, "destroy"),
-                array($this, "gc")
-            );
-
-            return true;
-        }
+        session_set_save_handler(
```
smeyer-ilias: I suggest to use return session_set_save_handler... here and to not hide the return status of the call.
Nevertheless I don't know what kind of error could occur here.

The return value is evaluated in ilInitialisation.

Services/Authentication/classes/class.ilSessionDBHandler.php
```diff
@@ -1,119 +1,90 @@
...
     /**
-    * removes sessions that weren't updated for more than gc_maxlifetime seconds
-    *
-    * @param	integer		$gc_maxlifetime			max lifetime in seconds
-    */
-    public function gc($gc_maxlifetime)
+     * Removes sessions that weren't updated for more than gc_maxlifetime seconds
+     * @param int $max_lifetime Maximum lifetime in seconds
+     */
+    public function gc($max_lifetime) : bool
```
smeyer-ilias: The interface demands  the return type: int | bool

Services/Init/classes/class.ilInitialisation.php
```diff
@@ -572,14 +572,9 @@ protected static function initDatabase()
...
     {
-        if (ini_get('session.save_handler') != 'user' && version_compare(PHP_VERSION, '7.2.0', '<')) {
-            ini_set("session.save_handler", "user");
-        }
-
-        require_once "Services/Authentication/classes/class.ilSessionDBHandler.php";
         $db_session_handler = new ilSessionDBHandler();
         if (!$db_session_handler->setSaveHandler()) {
-            self::abortAndDie("Please turn off Safe mode OR set session.save_handler to \"user\" in your php.ini");
+            self::abortAndDie("Please set session.save_handler to \"user\" in your php.ini");
```
smeyer-ilias: Maybe a more generic error message: something like "Cannot start session handling."

smeyer-ilias (changes requested):
Thanks for these improvements. 
Please can you check the comments.

Services/Authentication/classes/class.ilSessionDBHandler.php
mjansenDatabay: Changed with the latest commit

Services/Init/classes/class.ilInitialisation.php
mjansenDatabay: Changed with the latest commit

Services/Authentication/classes/class.ilSessionDBHandler.php
mjansenDatabay: Changed with the latest commit

smeyer-ilias (approved):

smeyer-ilias:
Thanks for that!

PR #3658, opened 2021-09-27
Add pull request stats GitHub action

mBeym:
This adds the GitHub action **Pull Request Stats**.

**Pull Request Stats** adds information review time, total comments and total reviews to a pull request when it get's created by any user.

See: https://github.com/marketplace/actions/pull-request-stats

### DIFF
--- a/.github/workflows/stats.yml
+++ b/.github/workflows/stats.yml
@@ -1,5 +1,5 @@
 name: Pull Request Stats
-on: pull_request_target
+on: pull_request
 jobs:
   stats:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
Hmmm, maybe somebody with more permissions has to create the PR. I'll check this.

-- Update --

This did not solve the issue

mjansenDatabay:
@ILIAS-eLearning/technical-board From our understanding it is not possible anymore to use the `pull_request` event for this workflow. We could try to change the event to `pull_request_target`.

I already pushed the workflow (as documented) into our trunk. It still fails for new PRs similar to the failure here in this PR.

See:
* https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/
* https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target

PR #3659, opened 2021-09-27
Feature/php8 compatibility/services/certificate

mBeym:
Adds missing type declarations in **Services/Certificate/classes** and **Services/Certificate/test**

### DIFF
--- a/Services/Certificate/classes/Helper/ilCertificateExerciseMembersHelper.php
+++ b/Services/Certificate/classes/Helper/ilCertificateExerciseMembersHelper.php
@@ -6,12 +6,7 @@
  */
 class ilCertificateExerciseMembersHelper
 {
-    /**
-     * @param int $objectId
-     * @param int $userId
-     * @return mixed
-     */
-    public function lookUpStatus(int $objectId, int $userId)
+    public function lookUpStatus(int $objectId, int $userId) : ?string
     {
         return ilExerciseMembers::_lookupStatus($objectId, $userId);
     }
--- a/Services/Certificate/classes/Helper/ilCertificateLPStatusHelper.php
+++ b/Services/Certificate/classes/Helper/ilCertificateLPStatusHelper.php
@@ -6,18 +6,13 @@
  */
 class ilCertificateLPStatusHelper
 {
-    /**
-     * @param int $objId
-     * @param int $userId
-     * @return mixed
-     */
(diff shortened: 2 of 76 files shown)


Services/Certificate/classes/Helper/ilCertificateExerciseMembersHelper.php
```diff
@@ -6,12 +6,7 @@
  */
 class ilCertificateExerciseMembersHelper
 {
-    /**
-     * @param int $objectId
-     * @param int $userId
-     * @return mixed
-     */
-    public function lookUpStatus(int $objectId, int $userId)
+    public function lookUpStatus(int $objectId, int $userId) : ?string
```
mjansenDatabay, line 9: @mBeym Can you please check the `Consumers` of this call? Maybe we have to handle the possible `null` value.

Services/Certificate/classes/Helper/ilCertificateLPStatusHelper.php
```diff
@@ -6,23 +6,13 @@
...
         return ilLPStatus::_lookupStatusChanged($objId, $userId);
     }
 
-    /**
-     * @param $objectId
-     * @param $userId
-     * @return mixed
-     */
-    public function lookUpStatus($objectId, $userId)
+    public function lookUpStatus(int $objectId, int $userId) : ?int
```
mjansenDatabay:
```suggestion
    public function lookUpStatus(int $objectId, int $userId) : int
```

Services/Certificate/classes/Helper/ilCertificateExerciseMembersHelper.php
mBeym: value returned by lookUpStatus is already checked in the consumers.

mjansenDatabay: Hmmm, that's not true IMO:

See: \ilExercisePlaceholderValues::getPlaceholderValues If `$status` is `NULL`, this will cause an error or lead to unexpected behaviour because of `$this->language->txt('exc_' . $status)`.

mBeym: Should be fixed now.

ilExercisePlaceholderValues::99

PR #3660, opened 2021-09-27
Fix: https://mantis.ilias.de/view.php?id=31380

kergomard:
Fix for: https://mantis.ilias.de/view.php?id=31380
I don't think we need this.


alex40724:
Thanks, but I am not able to reproduce the issue, see Mantis report.

kergomard:
You are right, me neither now. Ok, this still seems to be unneeded, but as no error appears anymore I will close it.

PR #3663, opened 2021-09-28
UI/ViewControl: add possibility of return value null to Pagination::getRange

lukastocker:
Now Pagination::getRange can return null as value. Thus a user can react appropriate if the list is empty.

### DIFF
--- a/src/UI/Component/ViewControl/Pagination.php
+++ b/src/UI/Component/ViewControl/Pagination.php
@@ -121,5 +121,5 @@ interface Pagination extends C\Component, JavaScriptBindable, Triggerer
     /**
      * Get the current number of entries on this page.
      */
-    public function getRange() : Range;
+    public function getRange() : ?Range;
 }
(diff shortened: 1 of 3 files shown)


Amstutz (changes requested):
Hi @lukastocker 

Makes sense, thx for the proposal. 

- [ ] One small addition, can you provide a small unit test, making sure, that getRange returns null and not causes any errors in case getPageLength < 1?

Thx.

lukastocker:
> Hi @lukastocker
> 
> Makes sense, thx for the proposal.
> 
>     * [ ]  One small addition, can you provide a small unit test, making sure, that getRange returns null and not causes any errors in case getPageLength < 1?
> 
> 
> Thx.

Hi @Amstutz,

glad to hear that. 
Sure, I can do that and will update this PR as soon as possible.

lukastocker:
Hi @Amstutz ,
I just pushed the test to this PR. 

Kind regards
Luka Stocker

Amstutz:
Thx a lot! Sry for the merge delay! Note, I only merged to trunk, since this is labeled as improvement, not as an issue (with some issue ticket). If you also like this for 7, this should not be a problem. Best just make a PR for 7 as well.

Again, thx a lot for improving the UI Components!

PR #3667, opened 2021-09-29
Remove unused debug file with debug functions

mjansenDatabay:


smeyer-ilias:
Thanks for removing this.

PR #3671, opened 2021-09-30
Feature/toast implementation

iszmais:
Implementation of the given Concept of [ILIAS Toast](https://docu.ilias.de/goto_docu_wiki_wpage_6494_1357.html)
Depends on https://github.com/ILIAS-eLearning/ILIAS/pull/3411 since its his implementation

### DIFF
--- a/src/UI/Implementation/Component/Toast/Renderer.php
+++ b/src/UI/Implementation/Component/Toast/Renderer.php
@@ -1,0 +1,77 @@
+<?php declare(strict_types=1);
+
+namespace ILIAS\UI\Implementation\Component\Toast;
+
+use ilException;
+use ILIAS\UI\Component\Button\Shy;
+use ILIAS\UI\Component\Link\Link;
+use ILIAS\UI\Implementation\Render\AbstractComponentRenderer;
+use ILIAS\UI\Implementation\Render\ResourceRegistry;
+use ILIAS\UI\Renderer as RendererInterface;
+use ILIAS\UI\Component;
+use LogicException;
+
+class Renderer extends AbstractComponentRenderer
+{
+    /**
+     * @inheritdoc
+     */
+    public function render(Component\Component $component, RendererInterface $default_renderer) : string
+    {
+        $this->checkComponent($component);
+
+        if ($component instanceof Component\Toast\Toast) {
+            return $this->renderToast($component, $default_renderer);
+        }
+
+        throw new LogicException("Cannot render: " . get_class($component));
+    }
+
+    protected function renderToast(Component\Toast\Toast $component, RendererInterface $default_renderer) : string
+    {
+        $tpl = $this->getTemplate("tpl.toast.html", true, true);
+
+        $title = $component->getTitle();
+        if ($title instanceof Shy || $title instanceof Link) {
+            $title = $default_renderer->render($title);
+        } else {
+            $title = htmlentities($title);
+        }
+        $tpl->setVariable("TITLE", $title);
+
+        $tpl->setVariable("TOAST_DELAY", $component->getDelayTime());
+        $tpl->setVariable("TOAST_VANISH", $component->getVanishTime());
+        $tpl->setVariable("VANISH_ASYNC", $component->getAction());
+
+        $desc = htmlentities($component->getDescription());
+        if (trim($desc) != "") {
+            $tpl->setCurrentBlock("desc");
+            $tpl->setVariable("DESC", $desc);
+            $tpl->parseCurrentBlock();
+        }
+
+        $actions = $component->getLinks();
+        if (!empty($actions)) {
+            foreach ($actions as $action) {
+                $tpl->setCurrentBlock("action");
+                $tpl->setVariable("ACTION", $default_renderer->render($action));
+                $tpl->parseCurrentBlock();
+            }
+        }
+
+        $tpl->setVariable("ICON", $default_renderer->render($component->getIcon()));
+        $tpl->setVariable("CLOSE", $default_renderer->render($this->getUIFactory()->button()->close()));
+
+        return $tpl->get();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function getComponentInterfaceName() : array
+    {
+        return [
+            Component\Toast\Toast::class
+        ];
+    }
+}
--- a/Services/Init/classes/Dependencies/InitUIFramework.php
+++ b/Services/Init/classes/Dependencies/InitUIFramework.php
@@ -33,6 +33,7 @@ class InitUIFramework
                 $c["ui.factory.tree"],
                 $c["ui.factory.menu"],
                 $c["ui.factory.symbol"],
+                $c["ui.factory.toast"],
                 $c["ui.factory.legacy"]
             );
         };
@@ -74,6 +75,9 @@ class InitUIFramework
         };
         $c["ui.factory.item"] = function ($c) {
             return new ILIAS\UI\Implementation\Component\Item\Factory();
+        };
+        $c["ui.factory.toast"] = function ($c) {
+            return new ILIAS\UI\Implementation\Component\Toast\Factory($c["ui.signal_generator"]);
         };
         $c["ui.factory.viewcontrol"] = function ($c) {
             return new ILIAS\UI\Implementation\Component\ViewControl\Factory(
--- a/src/UI/Component/Toast/Toast.php
+++ b/src/UI/Component/Toast/Toast.php
@@ -45,7 +45,7 @@ interface Toast extends Component
     /**
      * Init the default signals
      */
-    public function initSignals();
+    public function initSignals() : void;
 
     /**
      * Get the signal to show this toast in the frontend
--- a/src/UI/Implementation/Component/Toast/Factory.php
+++ b/src/UI/Implementation/Component/Toast/Factory.php
@@ -1,0 +1,24 @@
+<?php declare(strict_types=1);
+
+namespace ILIAS\UI\Implementation\Component\Toast;
+
+use ILIAS\UI\Component\Symbol\Icon\Icon;
+use ILIAS\UI\Implementation\Component\SignalGeneratorInterface;
+
+class Factory implements \ILIAS\UI\Component\Toast\Factory
+{
+    protected SignalGeneratorInterface $signal_generator;
+
+    public function __construct(SignalGeneratorInterface $signal_generator)
+    {
+        $this->signal_generator = $signal_generator;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function standard($title, Icon $icon) : Toast
+    {
+        return new Toast($title, $icon, $this->signal_generator);
+    }
+}
--- a/src/UI/Implementation/Component/Toast/Toast.php
+++ b/src/UI/Implementation/Component/Toast/Toast.php
@@ -1,0 +1,129 @@
+<?php declare(strict_types=1);
+
+namespace ILIAS\UI\Implementation\Component\Toast;
+
+use ILIAS\UI\Component\Button\Shy;
+use ILIAS\UI\Component\Link\Link;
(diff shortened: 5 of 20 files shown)


klees:
Hi @iszmais,

please ping @Amstutz and me once we should review this.

Best regards!

iszmais:
Hi @Amstutz @klees ,

this is now ready to review.
Feel free to leave feedback!

Greetings,
@iszmais

iszmais:
Hi @Amstutz @klees ,

the Pipeline for 8.1 seems broken right now, but as far as i can see this is a core problem and not caused by this PR.
Do you have any Change requests beside that?

Greetings,
@iszmais

klees (changes requested):
Hi @iszmais,

thanks a lot for working on the UI framework with us. Could you rebase this on trunk?

The changes look mostly fine. There are some matters where we, most likely, will just need to find some alignment regarding coding style and approaches. This might take some time, but will have huge long time benefits for sure.


- [ ] Inclusion: How would one include the toast currently? I think we have discussed this, but I can't remember and can't find it. Also: Where will the toast be displayed?

- [ ] New Methods: There are four new methods on the Toast-interface, right? For one, we would need to [take them to the JF](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/UI/docu/rules.md#modification-of-existing-ui-components). We could fast track the changes, but we would need to announce it at least. In general, I wonder if even want these methods. The "vanish"-time controls, how long the toast is visible until it vanishes, right? Why would we want this to be different for different toasts? What is the "delay"-time? I would suggest to move the getters from the public interface to the internal interface and remove the mutators all together in this PR.
- [ ] striptags: Why do you use `strip_tags` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3671/files#diff-2a4216b4c1da0207de3d8eae951b56a492ea8d8ebdfd15b47c8e15807f2259b1R53) instead of `htmlentities` [there](https://github.com/ILIAS-eLearning/ILIAS/pull/3671/files#diff-d08c76b5cf5210b2e917f668ef5c1fdcbd87f3e3172e01c2685a616e511911e9R45)? Using `htmlentities` would also be applicable for the title, why currently [is unchecked](https://github.com/ILIAS-eLearning/ILIAS/pull/3671/files#diff-2a4216b4c1da0207de3d8eae951b56a492ea8d8ebdfd15b47c8e15807f2259b1R35).
- [ ] JavaScript: The JS is missing tests (see example [here](https://github.com/ILIAS-eLearning/ILIAS/tree/trunk/tests/UI/Client)):  and, also, is very different from the other js we currently have. Most notably, in other JS we include java script in the header and bind it to html with an init. You load your client side code via an inline script. Is there a reason why we would implement the client side code so differently here? If not: please align with the other JS. Feel free to contact @nhaagen for support, also regarding tests.

- [ ] Unknown Component: Please throw a `LogicException` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3671/files#diff-d08c76b5cf5210b2e917f668ef5c1fdcbd87f3e3172e01c2685a616e511911e9R25). Returning an empty string in that case was common for some time, but I came to the understanding that this implementation just masks errors, so throwing will help our future selfs =)
- [ ] No Magic Numbers: Please use constants instead [of simple numbers](https://github.com/ILIAS-eLearning/ILIAS/pull/3671/files#diff-2a4216b4c1da0207de3d8eae951b56a492ea8d8ebdfd15b47c8e15807f2259b1R29). Put them on top of the class, just under `use`. Give them descriptive names (`DEFAULT_VANISH_TIME_IN_MS`, e.g.).
- [ ] HTML-Layers: Please get rid of the outer html layer ("wrapper") in [the template](https://github.com/ILIAS-eLearning/ILIAS/pull/3671/files#diff-01fa5c4ab2e013d3ce28b4589a26bc39c96d808e72771dbdb529ee020f7ec54dR1).
- [ ] Strict-Typing: Use strict typing, wherever possible. Please use strict typing wherever possible, e.g. see: https://github.com/ILIAS-eLearning/ILIAS/pull/3671/files#diff-d136f7a1d79770bc6a10e380b9991006e7f0d51507253548aac85aabc49e4685R58 or https://github.com/ILIAS-eLearning/ILIAS/pull/3671/files#diff-d136f7a1d79770bc6a10e380b9991006e7f0d51507253548aac85aabc49e4685R65 or https://github.com/ILIAS-eLearning/ILIAS/pull/3671/files#diff-2a4216b4c1da0207de3d8eae951b56a492ea8d8ebdfd15b47c8e15807f2259b1R42 or https://github.com/ILIAS-eLearning/ILIAS/pull/3671/files#diff-2a4216b4c1da0207de3d8eae951b56a492ea8d8ebdfd15b47c8e15807f2259b1R111 or https://github.com/ILIAS-eLearning/ILIAS/pull/3671/files#diff-2a4216b4c1da0207de3d8eae951b56a492ea8d8ebdfd15b47c8e15807f2259b1R111 

Important note, we do not see any major roadblocks concerning a11y atm. However, we are no experts in this field. We would test this again by an external expert before or early in beta.

Thx a lot, 
@klees and @amstutz

iszmais:
Hi @Amstutz @klees ,

I updated the branch and gave feedback to all points feel free to rereview.

**To the Questions:**
- The Toast will not have itsself is not bound to any context so it can provide a dynamic usage (e.g. like the modal). We came to the point that the location of it is in the responsebility of the using context. So e.g. for the initial usage inside the NotificationCenter (see https://github.com/ILIAS-eLearning/ILIAS/pull/3740) it will use the same container as the the uninified current popups.

**To the Recommendations:**
- The origin feature request of the toast (https://docu.ilias.de/goto_docu_wiki_wpage_6494_1357.html) implies that each kind of future toast has its own individual vanish time wich may be similar but dont have to. Therefore i intended too connect it to the toast itsself to guarantee expandability for future. This maybe could be centralized, but i couldnt find any demarcation inside the request wich would allow such centralisation. The delay time is just resulting addition to the vanishing time to delay the toast apperance when shown to increas the visibility o the page (to make it less alarming) wich seemed reasonable to me. But this is indeed optional and could be removed if requested. Also this might comin handy for toast wich more text in future.
- changed it
- changed it

**To the change requests:**
- changed it
- changed it
- This layer is neccessary since the since the Toast come with an animation and therefore needs a static ancor on which the animation is bound to prevent side-effects.
- changed it (except getTitle() since its return type is mixed)

Greetings,
@iszmais

iszmais:
Hi @nhaagen ,

could you provide any documentation how to setup js testing for ilias?
I figured out it uses mocha but seems to miss some modules (even in dev installation).

Thx in forward,
@iszmais

nhaagen:
Hi @iszmais ,
here is the ultra-short version:

you will need mocha, chai, esm and jsdom.
```
npm i --save-dev mocha chai esm jsdom
```
place your tests in tests/UI/Client/Toast
(best have a look at existing tests to see how they work)

run 
```
npm test
```

Amstutz:
Also, checkout the docs at: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js/js-unit-test.md

iszmais:
Thx @Amstutz @nhaagen that helped alot!

There seems to be an issue with a sup-dependency of the newest jsdom version (whatwg-url).
But nothing serious, just FYI.

Greetings,
@iszmais

Amstutz:
cool, thx for the heads-up.

iszmais:
Hi @klees ,

i implemented all feedback or gave a statement to the disscusable points.
Feel free to review!

Greetings,
@iszmais

iszmais:
@klees @Amstutz   and Feel free to call me on skype if you see some major dicussable points.

iszmais:
Conclusion of meeting 10.02.2022 with @mjansenDatabay in representation of @klees /  @Amstutz  : 

- [x] The `Toast` interface should be kept untouched by implementation PRs. While it might be okay to add the missing type hint here, all interface changes in general must be discussed in dedicated PRs with the `ILIAS JourFixe`.
- [x] The `Toast` implementation class should not be aware of non-elemental validations like HTML/XML syntax and therefore should not interfere on those. It must be `output-agnostic`. Escaping should be placed inside the renderer, complex validation in the depending data types/handlers/repositories etc..
- [ ] The `Toast` should have a centralised container to be rendered in.
- [ ] The `JavaScript` of the `Toast` should **not** be loaded inline but registered inside the head like common in ILIAS.

iszmais:
Hi @klees @Amstutz ,

to the given Feedback:
- I changed the 2 first minor points (Toast Interface and html validation).
- I would appreciate to have a talk about the centralisation of the toast itsself and its JS. I understand the basic problem of that but would request another solution. I would highly appreciate if we could have a talk about that. Feel free to call me on skype or  suggest a time for a meeting.

Greetings,
@iszmais

mjansenDatabay:
@Amstutz, @klees, @iszmais and @mjansenDatabay had a meeting to discuss the open issues:

- [ ] We need a new UI Component `Toast Container`, which should be retrievable by the `\ILIAS\UI\Component\Toast\Factory`. A standard `Toast Container` MUST be implemented by the corresponding concrete `Toast Factory`. This new interface has to be provided by a separate PR and discussed by the `JourFixe`. An implementation will be provided afterwards. Advantage: The container can have its own rules, like how to address A11Y or other future requirements (e.g. how many `Toasts` can be displayed simultaneously, etc.).
- [ ] The `JavaScript` ressource(s) for the `Toast` component MUST be registered by the `Toast Container`. This solves the current issue of including the `JavaScript` file in the asychronously rendered `HTML` template. Furthermore the `Toast Container` can be used as `Labeled/Name Location` where `Toast` components are rendered in. A11Y requirements like the option to introduce an `aria-live` region during initial render time could be addressed in this way, instead of dynamically adding such regions when `Toast` components are retrieved and rendered asynchrously.
- [ ] The `Toast Container` should be used in the `\ILIAS\UI\Component\Layout\Page\Factory::standard`. The interface change requires another PR as well.

@klees / @Amstutz : Did I forget something?

src/UI/Implementation/Component/Toast/Renderer.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        $tpl = $this->getTemplate("tpl.toast.html", true, true);
+
+        $title = $component->getTitle();
+        if ($title instanceof Shy || $title instanceof Link) {
+            $title = $default_renderer->render($title);
+        } elseif ($title !== htmlentities($title)) {
+            throw new ilException('Title should not contain HTML!');
+        }
+        $tpl->setVariable("TITLE", $title);
```
klees, line 40:
```suggestion
        $tpl->setVariable("TITLE", htmlentities$title));
```

Please simply render the escaped string instead of throwing an exception here. Same for `$desc`.

src/UI/examples/Toast/Standard/with_vanish_time.php
```diff
@@ -0,0 +1,13 @@
+<?php declare(strict_types=1);
+
+namespace ILIAS\UI\examples\Toast\Standard;
+
+function with_vanish_time() : string
```
klees: Please remove this example, we will not need it anymore since consumers can't set that property anymore.

src/UI/examples/Toast/Standard/with_delay_time.php
```diff
@@ -0,0 +1,13 @@
+<?php declare(strict_types=1);
+
+namespace ILIAS\UI\examples\Toast\Standard;
+
+function with_delay_time() : string
```
klees: Please remove this example, we will not need it anymore since consumers can't set that property anymore.

klees (changes requested):
Hi @iszmais,

I found three minor issues and kindly ask you to rebase this to trunk  and squash the commits. I'll click merge immediately after...

Best regards!

iszmais:
@klees @Amstutz ,
updated and squashed. Feel free to merge (the test seem to fail due to an external cause).

Greeting,
@iszmais

PR #3674, opened 2021-10-02
BackgroundTasks: Delete obsolete classes and interfaces

mjansenDatabay:
This PR removes classes and interfaces I could not find any usage for.

Maybe you would like to keep ` src/BackgroundTasks/Implementation/Tasks/SlowPlusJob.php` as an example, but the interfaces are definitely not used.


chfsx:
Thanks @mjansenDatabay !

PR #3685, opened 2021-10-03
JS Unit tests, some proposals and refactoring JS Unit Tests

Amstutz:
I am currently working on a new UI Component involving some js parts, which I would like to unit tests. While implementing the tests according to the [js-unit-test docs](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js/js-unit-test.md) , I observed some aspects, which I believe need some love, before continuing adding new logic.

This PR looks rather large, but no small part is just moving parts around. However, to make this easier to process, here the main parts of it:

- We have currently two folders involving the word *test* on top level, on called *test* the other *tests*. There is no telling from the outside which does what. Also, this further pollutes our already crowded top level. After a short ping with @klees, I hereby propose to have a main [mocha.config.js](https://github.com/ILIAS-eLearning/ILIAS/pull/3685/files#diff-da6caa1bafcadfbad112f260d18d149fb95bc32a59f088e92cfda1398ae83612) file in which we collect all places to look for js tests. The main one being the place, where also the mocha.config.js file is located, tests/UI/Client . The doc is updated accordingly.
- I propose to also test the [example tests](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js/) in the docs directory if npm test is fired, to make sure those updates are kept up-to-date.
- I propose to move from symlinking the files to the test files in this folder, to simply having them stored directly there. As we do with the php unit tests.
- I propose to recursively loop through all folders in tests/UI/Client to make more structure possible. I expect there to be lots of files in future.
- Since in we had already two components in tests/UI/Client already being unit tested, but through the browser and not yet with mocha, I propose hereby to migrate those to mocha testing as well. This made some further changes/additions necessary:
  - I moved both components to the bundling as documented [here](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js/js-bundling.md). This made some slight refactoring necessary.
  - Since some tests really on jQuery and jQuery is used inside the component, I needed to inject jQuery properly as dependency. Also I could not really anymore on il.UI.SomeComponent inside the Components, therefore I needed e.g. to inject counter into NotificationItem. IMO, this is good practice and a nice side effect from unit testing those components through cli with mocha.
  - I needed to load some DOM in the unit tests, to work on with jQuery.
  - To make sure that this DOM stays up to date with DOM rendered by PHP, I added some php tests, explicitly testing, if the DOM loaded in the js Tests is actually DOM we render in php. IMO this was also necessary, since we had very recently a bug which had been caught, if we had made sure that the html used to test js, had been html rendered in php (there was a change in a template, leading js to fail). It might be good practice to do this, whenever we use example html to test js code upon.
  - I added a one liner to .gitIgnore `/node_modules/**/.cache`, I always have lots of changes in those, when running npm test.
  - Another [one liner](https://github.com/ILIAS-eLearning/ILIAS/compare/trunk...Amstutz:improvement/trunk/js_unit_tests_mocha#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) in package.json, pointing to the npm test to the proper config.

There are some things we should further do soon. I would not like to include them here, since this is big enough already. They are:
- ~~I do not like the way bundling is currently shown in the docs by using the „es“ format. It tends to pollute the global scope of the application. Maybe we should move to iife, however afaik, this makes it hard to expose the public interface to other components with il.UI~~. --> See the discussion with @nhaagen bellow. However, we need to make sure, to initialize il, before loading the UI.
- We should bundle all UI Components js into one single file. This is something, we already agreed upon but I think the time is here to do it.
- We should make further use of the redux-like model proposed in the Mainbar. In the new component I work with it, it makes js a lot easier to test. 
- Finally and most importantly, we should require unit tests for all new js code (at least in the UI Components). I am sure it improves the modularization and overall quality of the JS code a lot. At least it already did IMO for Counter and Notification items.


Amstutz:
@mjansenDatabay and @nhaagen I added you two as reviewers as well since you have contributed to unit js unit tests or have experience/proposed aspects of bundling js. If you do not find time to look at it, all good. I understand.

@chfsx , @tfamula and @alex40724 : This might be of interest to you as well, since you all contributed to js in the recent past. Opinions/objections/feedback/questions are always welcome.

src/UI/templates/js/Item/rollup.config.js
```diff
@@ -0,0 +1,7 @@
+export default {
+  input: './src/notification.js',
+  output: {
+    file: './dist/notification.js',
+    format: 'es'
```
nhaagen: format: 'iife'

nhaagen:
Hey @Amstutz ,
thank you very much for this, the location of the tests and linking files really bugged me, but then I simply did not find a proper way to move them. If the "--recursive --config tests/UI/..." does the trick -  hooray, I'm all for it. 
One minor note: if you render the js in IIFE instead of ES, the top namespace will be less cluttered. Give it a try ;)
Otherwise: thumps up, and thanks again.

nhaagen (approved):

Amstutz:
Hi @nhaagen 

Thx very much for the feedback. Highly appreciated. I tested before with iife, there is one (up to this point in time) minor issue, we will run into, we will depend on some external piece of code to globally initialize il for us, right? So this piece of code does not make sense anymore:

```
var il = il || {};
```

Because with IIFE we would only define il internally, if not initialized yet. I do not think this is an issue. But at some point, we will need to make sure, il is initialized before we load UI. However, the benefit is bigger than the risk. I agree. I changed the code and also updated the docs, proposing to make iife the default.

Thx.

alex40724:
1. Bundling

Edit: After reading through all the changes I have the feeling that there is some kind of misinterpretation with the bundling concept.

The bundling.md file currently promotes ES modules. "The single files SHOULD contain export ES6 modules and the overall package SHOULD also be an ES6 module." The change in this file (using "iife" instead) would be strange.

I do not understand how ES modules pollute the top namespace. Is this an UI framework related or a general issue? It should e.g. be possible to bundle the whole UI framework js code in one module which imports the single components (like the component1.js part of the bundling example). Everything would be self contained. If this is currently not possible, we still should not change the general bundling guideline imo.

2. Testing

Using `mocha.config.js` is a good idea, but `tests/UI/Client/mocha.config.js` is a strange place for a global config being referenced in `package.json` imo. The `docs` tests do not belong to the UI framework and we want tests in `Services` and `Modules` also being included. I would suggest to include paths to `./Services/**/test/*.js` and `./Modules/**/test/*.js`.

The `mocha.config.js` could e.g. go directly to the test directory `tests/mocha.config.js` imo.

Amstutz:
Hi @alex40724 

Thx a lot for the Feedback and for your time looking into this! I am no expert on this, so take everything with a grain of salt. Thx for helping forstering the discussion here.

1. Bundling

> Edit: After reading through all the changes I have the feeling that there is some kind of misinterpretation with the bundling concept.

That is possible. Therefore, I have no strong opinion as long as we try to be as consistent as possible. 

> "The single files SHOULD contain export ES6 modules and the overall package SHOULD also be an ES6 module." The change in this file (using "iife" instead) would be strange.

Yeah, that would need to be consistent, I will improve this further as soon as we have a decision on the way forwart with bundling.
 
Scoping in ILIAS is performed by putting everything into the il scope. From there we devide further into e.g. il.UI, or il.Foo etc. This prevents as from having naming naming collisions or from having variables, with a scope much to large. We could discuss improving this, but this would probably involve more changes we could handle atm. 

Further I think it is good and common practise to export functionality of modules to other modules using to "export" statement, like so:

```
var X = function(){
...
}

export {X};
```

The problem I did run into, that it was only possible for me to export named functions, e.g. the following is not possible afaik:

```
function(il){
  var X = function(){
  ...
  }
  
  il.X = X;
}(il);

export {il.X};
```

So I did the first option. Leaving X in the global namespace, where it does not belong as pointed out by @nhaagen. So I followed his recommendation and changed to iife, which wraps everything inside a self executing function after compiling. Leaving us with the issue of when we initially define `il`.

> Is this an UI framework related or a general issue?

I do not think so. To me, this will be an issue, wherever we have multiple modules tested individually being exported individually. However, I am sure there coud be other, better solution in place or possible. Do you have some productive JS using ES which you think could be a good modle of how we bundle/modularize/export/import/scope and in future also unit test our js code? Then we can compare the options possible.

> It should e.g. be possible to bundle the whole UI framework js code in one module which imports the single components (like the component1.js part of the bundling example). Everything would be self contained.

Yes this is the plan and will be the immediate next step, once we reach a decision here. See in the initial description, last paragraph:

> We should bundle all UI Components js into one single file. This is something, we already agreed upon but I think the time is here to do it.

So, TL;DR: I am open to other options, as long as we have a pattern that a) does not pollute to global name space and b) makes it easy/possible to unit test small parts of large components.

2. Testing

> Using mocha.config.js is a good idea, but tests/UI/Client/mocha.config.js is a strange place for a global config being referenced in package.json imo.

+1, I definitely see that. This was just in lack of a better idea. I want to have it out of the root dir. This dir is just crumbled with things, which gives a very messy bad first impression. `tests/mocha.config.js` works for me.

> I would suggest to include paths to `./Services/**/test/*.js` and `./Modules/**/test/*.js`.

+1, definitely agree as soon as we have js tests there. Not sure, if there are only the docs and the UI Components featuring such. If there are, we should immediately add the directories. If there are none, I wonder, whether it is not a bid misleading adding those dir at this moment. I do not have strong opinion here. 

Thx for the feedback, looking forward to more discussion on this.

docs/development/js/js-unit-test.md
```diff
@@ -14,18 +15,26 @@ ILIAS is using [Mocha](https://mochajs.org/) as test framework for Javascript un
...
   ...
   "scripts": {
-    "test": "mocha --require esm"
+    "test": "mocha tests/UI/Client/* --require esm"
   },
   ...
 }
 ```
 
+Note that it might be good practise to keep the tests cloze to the source an only symlink the files to the above
```
klees:
```suggestion
Note that it might be good practise to keep the tests close to the source an only symlink the files to the above
```

klees (approved):
Hi @Amstutz,

thanks a lot for iterating on this and cleaning up stuff. I don't feel I have much to add, besides one typo (which I commented on inline).

Regarding `iife` I tried to come up with an opinion by reading a little. As far as I understand, the affairs are as such:

* We ask people to write ES6 modules to modularize their stuff, which gives a clear path to export and import functionality.
* Current state of the art is that components import functionality from other components via `il` and its "namespaces". They do not import functionality from other components via ES6 modules (at least I'm not aware of this).
* Futhermore, current state of the art is that components (such as the ones edited here) use ES6 modules to modularize internally, and then use rollup to bundle their internal modules into one file.
* We ask people to bundle the js of their component as an ES6 module as well.
* Using ES6 modules for a components js (as we currently do) could be another way to import functionality from other, but currently `il` is the way we use.
* @Amstutz changes try to fix this by breaking the ES6 pattern for the bundling of a components js by adding a top level file that shoves the functionality to a sub namespace of `il` and bundling this as `iife` to minimize surprises from changes in the global namespace.

So basically, this seems to be a matter of interoperability between the old `il`-based modularization and the new ES6 based modularization. We would keep using the old model for export/import functionality among components but use ES6 modules in these components internally.

From my perspective this feels plausible. I cannot see how we would switch to an ES6 based model for inter component modularization currently. I would be happy to be told differently, though. So we would keep using the `il` model while improving components internally to use ES6. Some day, when components are in an according shape, the switch to use ES6 throughout the system would become a lot smaller and we could get rid of `il`.

Is this something we want to discuss on Jour Fixe, though?

Best regards!

Amstutz:
Thx for the feedback. Especially for your opinion on the handling of modules. Yes, JF might be a good point to bring this up. This will have some impact for a somewhat extended group of devs.

matthiaskunkel:
Jour Fixe, 15 NOV 2021 :  Timon made a short presentation of his PR and its current discussion. We do not have any objections against using iife for creating the bundles. The mocha.config.js should go directly to the test directory tests/mocha.config.js. Please merge PR to trunk accordingly.

Amstutz:
I added the changes as requested by jf. However, I will manually pick the changes to trunk. I seem to have messed something up while rebasing my changes onto trunk.

PR #3686, opened 2021-10-05
UI: add on-first-view trigger to slate (#31434)

klees:
This fixes [#31434](https://mantis.ilias.de/view.php?id=31434) by adding the missing trigger for "on_first_view" to the slate as well (and not the mainbar only).

This is a witness for the observation, that we really want some good js-tests that also cover PHP/JS-integration.


Amstutz:
Thx a lot for the quick fix. @mjansenDatabay thx for the report.

PR #3691, opened 2021-10-06
ILIAS 6: Reverted the admin slate to combined and saved performance by pre-applied visible check on the administration

iszmais:
Partially reverts https://github.com/ILIAS-eLearning/ILIAS/commit/f1bf5e5926aae81f0977662c17c1db00c81c788a to fix the problems inside the Administration Node without decreasing the performance:
https://mantis.ilias.de/view.php?id=31451
https://mantis.ilias.de/view.php?id=31421
https://mantis.ilias.de/view.php?id=31449
https://mantis.ilias.de/view.php?id=31447

### DIFF
--- a/Services/Administration/GlobalScreen/classes/AdministrationMainBarProvider.php
+++ b/Services/Administration/GlobalScreen/classes/AdministrationMainBarProvider.php
@@ -30,7 +30,7 @@ class AdministrationMainBarProvider extends AbstractStaticMainMenuProvider
         $access_helper = BasicAccessCheckClosures::getInstance();
         $top = StandardTopItemsProvider::getInstance()->getAdministrationIdentification();
         $logged_in = $access_helper->isUserLoggedIn();
-        if (!$logged_in()) {
+        if (!$logged_in() || !$this->dic->rbac()->system()->checkAccess("visible", SYSTEM_FOLDER_ID)) {
             return [];
         }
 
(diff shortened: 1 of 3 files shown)


mjansenDatabay:
This solves the `Performance Issues` for people without `visibile` permission to the administration (`ref_id`: 9), but of course the permission checks for all sub nodes will be still executed for users **with** `visible` permission.
The permission handling for the administration item in your PR is similar to 5.4.x, so IMO this is a good approach.

And: Of course this PRs solves the regressions caused by the latest changes.

IMO the perfect solution would be: This PR + `Asynchronous Loading` + `No Legacy Slates but Combined Slates` :-).

Amstutz:
I agree with the statement, of @mjansenDatabay above:

> IMO the perfect solution would be: This PR + Asynchronous Loading + No Legacy Slates but Combined Slates :-).

iszmais:
I copied that PR for the other affected and maintained versions: 
https://github.com/ILIAS-eLearning/ILIAS/pull/3692
https://github.com/ILIAS-eLearning/ILIAS/pull/3693

chfsx:
Hi @iszmais and thank you very much for looking into this. from my point of view it is not necessary to undo most of this. the most important points are the following:
- the access checks for the individual admin menu entries must be made within the callable, which is already the case.
- the provider for the admin mainmenu can also return an empty array if someone has no access to the administration at all (as suggested by you and adopted by me). 
- TopParentItems could still load their content async, but in the case of the existing core menu items, it no longer does so.

I have just pushed the adjustments together with a few optimisations in the rights checks for ILIAS 6-trunk and therefore I am running the PRs here. Thanks anyway!

iszmais:
@chfsx Thanks for acknowledging my solution!
The final solution looks good to me, too.
Could you link me your PR in wich you solved this for self education?
I would highly appreciate!

chfsx:
Hi @iszmais ! most changes are here in this commit: https://github.com/ILIAS-eLearning/ILIAS/commit/cfe1a1321009bf148fa21f1f3d62e2ddb581c239
but the desired behaviour will be implemented after merging the PR #3698.

PR #3698, opened 2021-10-08
[FIX] 0031418: bad performance due to unnecessary background requests

chfsx:
Hi @klees !

Mantis: https://mantis.ilias.de/view.php?id=31418

As discussed before I had a look at the mainbar.js to prevent the mainbar from pulling to much async content. The state model now saves:
- the current active top-level (it did before, but not in all cases)
- the last "clicked" entry
- and an entry always knows it's toplevel entry

with this info, the event to reload the content of a slate only fires when
- the active top-level changed and an entry is within the active toplevel
- the entry was actively clicked

why this both cases?
- we could also trigger the event for every entry which is in the active top-level, but this leads to unnecessary requests for siblings within the same toplevel, therefore we check for a "top level changed"
- but if we only fire the event on a "top level changed", a new sibling actively clicked nested in the same top-level would not fire, therefore the 'fore on actively clicked'

see: https://github.com/ILIAS-eLearning/ILIAS/compare/release_7...studer-raimann:fix/0031418/7/unnecessary-requests#diff-84d572049e14e3939184146d40735f1ab00485ec64184969aa6992f1a39c7850R926

### DIFF
--- a/src/UI/templates/js/MainControls/dist/mainbar.js
+++ b/src/UI/templates/js/MainControls/dist/mainbar.js
@@ -335,8 +335,7 @@ var model = function() {
             tools: {}, //"moving" parts, current tools
             known_tools: [], //gs-ids; a tool is "new", if not listed here
             current_active_top: null,
-            last_actively_engaged: null,
-            top_level_changed: null
+            last_actively_engaged: null
         },
         entry: {
             id: null,
@@ -452,9 +451,7 @@ var model = function() {
             state.entries = reducers.entries.engageEntryPath(state.entries, entry_id);
             state = reducers.bar.disengageTools(state);
             state = reducers.bar.anySlates(state);
-            var last_active_top = state.current_active_top;
             state.current_active_top = helpers.getEngagedTopLevelEntryId();
-            state.top_level_changed = state.current_active_top !== last_active_top;
         },
         disengageEntry: function (entry_id) {
             state.last_actively_engaged = null;
@@ -645,6 +642,7 @@ var persistence = function() {
     storeStates = function(state) {
         state.entries = compressEntries(state.entries);
         state.tools = compressTools(state.tools);
+        state.last_actively_engaged = null;
         cs = storage();
         for(idx in state) {
             cs.add(idx, state[idx]);
@@ -693,6 +691,7 @@ var renderer = function($) {
         ,mainbar_buttons: '.il-mainbar .il-mainbar-entries .btn-bulky, .il-mainbar .il-mainbar-entries .link-bulky'
         ,mainbar_entries: 'il-mainbar-entries'
     },
+      events_fired_for = [],
 
     dom_references = {},
     dom_element = {
@@ -847,7 +846,7 @@ var renderer = function($) {
 
             var triggerer = parts.triggerer.withHtmlId(dom_references[entry.id].triggerer),
                 slate = parts.slate.withHtmlId(dom_references[entry.id].slate);
-                
+
                 //a11y
                 triggerer.getElement().attr('aria-controls', slate.html_id);
                 triggerer.getElement().attr('aria-labelledby', triggerer.html_id);
@@ -863,7 +862,8 @@ var renderer = function($) {
                 triggerer.engage();
                 slate.engage();
                 // fire event
-                if(fire_event) {
+                if(fire_event && !events_fired_for.includes(entry.id)) {
+                    events_fired_for.push(entry.id);
                     slate.getElement().trigger('in_view');
                 }
 
@@ -924,7 +924,7 @@ var renderer = function($) {
             for (idx in model_state.entries) {
                 actions.renderEntry(model_state.entries[idx], false,
                   model_state.last_actively_engaged === idx
-                  || (model_state.top_level_changed && model_state.current_active_top === model_state.entries[idx].getTopLevel()));
+                  || model_state.current_active_top === model_state.entries[idx].getTopLevel());
             }
             for(idx in model_state.tools) {
                 actions.renderEntry(model_state.tools[idx], true);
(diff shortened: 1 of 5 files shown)


chfsx:
I found some "strange" behaviour when having tool (in my case the Online Help) active, there were some places where the content of async-slates did not reload properly

chfsx:
@klees I fixed the behaviour, additionally these two changes:
- storing the state of top_item_changed is no longer needed
- events are only fired once (to prevent even more async requests)

chfsx:
and I used the generated code of mainbar.js with ILIAS 6 wich seems to work as well

mjansenDatabay:
Will also fix: https://mantis.ilias.de/view.php?id=31480

bogenschlag:
Hi, @klees and @chfsx, are you still debating about this behind the scenes? And/or would it be reasonably safe for me to cherrypick this? (Thanks!) 🙂

5qfjxx

klees:
Hi @bogenschlag,

I couldn't check this until now. Merging this could lead to some client side errors but data loss or other permanent damage seems to be unlikely.

Best regards!

bogenschlag:
@klees Thank you—I think I’ll just wait then. 👍🏼😌

chfsx:
@bogenschlag I rolled this changes out on a ILIAS 6 installation, most of the clients had the old mainbar.js cached and this could lead to some strange bevahiour. But in most cases it worked well...

bogenschlag:
Thanks, @chfsx! 🙂

klees (changes requested):
Hi @chfsx,

thanks for looking into this!

- [ ] main.js: The changes there are only naming changes, right? Do I understand correctly, that this is just the clarification that this really is the top node that is currently active, not the one that was active before that?
- [ ] Tests: Did you run them?

I'm not very happy with [this change](https://github.com/ILIAS-eLearning/ILIAS/pull/3698/files#diff-a0da3d31e9ff3e426b823867ccd90586a38c0980bc3abe3029fd7be771699ad8R186) (and [this related change](https://github.com/ILIAS-eLearning/ILIAS/pull/3698/files#diff-a0da3d31e9ff3e426b823867ccd90586a38c0980bc3abe3029fd7be771699ad8L32), because the meaning of the "in_view"-event will deteriorate then (see [this comment](https://github.com/ILIAS-eLearning/ILIAS/pull/3610)). If I understand correctly, the event will only fire when the element comes into view for the first time. But that part of the change is not essential, right? Can we just remove it to keep promises? Maybe the "only load once"-behaviour rather can move to the Global Screen?

Best regards!

chfsx:
Hi @klees 

Thanks for looking into this!

main.js: this changes the behaviour internally to store the current active top-item `current_active_top`. this is needed to dedice whether a sub-item is 'in view' or not. IMO the behaviour and especially the event `in_view` before this changes was completely wrong and not a in view. With these changes only really `in_view` get triggered. one could discuss the term `in_view` but i don't understand exactly what makes you unhappy with the change. It's exactly the part needed to avoid the unnecessary requests (currently nearly EVERY open slate triggers a async request to get it's content, even all the slates which are not visible since another slate is open.). 

I don't see the "only load once"-behaviour in the GS. The GS transports information from the components to the UI but is not responsible for the behaviour of the UI or the whole MainBar.

Tests: unfortunately no, I didn't even know there are some.

klees:
Hi @chfsx!

Thanks for the feedback.

**Regarding `current_active_top`**:

Where is the actual change of behaviour for that property in the model? I could only find naming changes.


**Regarding `in_view`**:

The event in the JS is the implementation of [`Slate::appendOnInView`](`https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/src/UI/Component/MainControls/Slate/Slate.php#L60). The documentation says `A Signal that is triggered when the slate "comes into view", i.e. is being engaged.`.

This signal is used by the GlobalScreen to implement [`MakeSlateAsync`](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/src/GlobalScreen/Scope/MainMenu/Collector/Renderer/MakeSlateAsync.php#L31).

As far as I understand the implementation currently mostly implements the announced behaviour. [That `if` here](https://github.com/ILIAS-eLearning/ILIAS/pull/3698/files#diff-a0da3d31e9ff3e426b823867ccd90586a38c0980bc3abe3029fd7be771699ad8L32) makes the event only trigger if the status was switched. There might be circumstances where this fails for some reason, though.

[This change](https://github.com/ILIAS-eLearning/ILIAS/pull/3698/files#diff-a0da3d31e9ff3e426b823867ccd90586a38c0980bc3abe3029fd7be771699ad8R186), as far as I understand it, would make the signal only trigger when it is engaged for the first time. This is desireable from the perspective of the ticket this is trying to solve but would in fact make the documentation lie.

So, this the reason why I asked whether this "only load once" behaviour could be moved to the Global Screen. The UI framework currently does not know anything about loading (only a variable name suggests otherwise). That change would effectively break the announced interface. Also, I cannot see that this change is announced in the description of the PR (or I am missing something...). This is the part of the change that makes me unhappy, because I would not break interface in a released version.

Regarding the [other change](https://github.com/ILIAS-eLearning/ILIAS/pull/3698/files#diff-a0da3d31e9ff3e426b823867ccd90586a38c0980bc3abe3029fd7be771699ad8R248), it feels as if it might move the implementation a little closer to the name of the signal ("in view") but away from the documentation ("being engaged"). I think this is acceptable, even for a released version, as we move the implementation nearer to the actual name. The "i.e. being engaged" IMO mostly is caused by a underdeveloped understanding of what "engaged" means when this signal was created, so I would propose to just remove that part in the documentation.

I see that these changes solve the issue they are supposed to solve, but (partly) in the wrong place, at the cost of creating a source for possible new bugs and at the cost of creating divergences between docu and implementation. From my perspective, this not worth it, because software development is a marathon, not a sprint.

I hope this clarifies my concerns.

I now will ask my colleague @nhaagen to implement the announced "in_view"-behaviour, a little differently than you propose here but to the same effect. I will not implement the "only fire once" behaviour on top of "in view", because this is what would effectively break the interface.

Best regards!

PR #3699, opened 2021-10-09
Color picker access of undefined

Amstutz:
Just a tiny php8 fix, which currently throws an error in system styles, making refactoring there somewhat cumbersome. Would love to have this fix in trunk, even if this component is not fully refactored for php8.


alex40724:
Thanks!

PR #3576, opened 2021-08-26
ILIAS 8: Abandon MySQL Fulltext Search (Part I/II of Abandon MyISAM)

mjansenDatabay:
Abandoned Feature Request: https://docu.ilias.de/goto_docu_wiki_wpage_6157_1357.html

A PR to migrate the given MySQL table engines will be provided by @gvollbach (and has to be reviewed by @chfsx afterwards). This PR can be reviewed and merged separately.

PR #3704, opened 2021-10-12
ILIAS 8: Remove MyISAM (Part II/II) artifacts from core and migrate tables

gvollbach:
This pull request removes some artifacts from the core, which still use MyISAM, since it seems @chfsx had already removed the most part of the code, there was not much left.

Important: Since there is no documentation about the ilDBTemplate creation process, or I'm to blind to find it, this has to be also tested, after the ilDBTemplate creation. We will contact @fwolf-ilias regarding this.

Follow up of: https://github.com/ILIAS-eLearning/ILIAS/pull/3576


setup/sql/dbupdate_05.php
```diff
@@ -7071,6 +7071,10 @@
 }
 
 ?>
+<#5809>
+<?php
+    $ilDB->migrateAllTablesToEngine();
```
mjansenDatabay: @gvollbach 

1. Actually database update steps MUST NOT be appended to `dbupdate_05.php` anymore, see: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/setup/sql/dbupdate_05.php#L7076 . There may be an exception, and of course this migration could be such an exception (IMO).
    *  Maybe @chfsx as the maintainer of `Services/Database` has a strong opinion on this.
    * @gvollbach : If this step has to be moved, you could create a class like this: https://github.com/ILIAS-eLearning/ILIAS/blob/a22b793b6389699cbece1d0492cc0b338078d09d/Services/Database/classes/Setup/ilDatabaseSetupStepsUpdateSteps.php
2. This method returns an array of errors. I personally suggest to compare this array against `!== []`. If there are errors, they should be displayed, and an exception should be thrown. The step MUST not be completed in this case.

mjansenDatabay (changes requested):

mjansenDatabay:
@gvollbach Could you please resolve the conflict :)

setup/sql/dbupdate_05.php
chfsx: Thanks @gvollbach! I think the part of the „auto migration“ of all tables ist the most tricky one. not sure, if we can do this using „ILIAS“ because there can be a lot of reasons this might fail. At least we should follow the suggestion of @mjansenDatabay to collect and show the failed tables.

@corro you did „a lot“ of MyISAM to InnoDB migrations using cli-tools, what were the problems which occured? Is it the time needed since this can last a while on larger databases? or were there any other problems? thanks for you hints! This helps to decide if this could be an UpdateStep, a Migration or nothing ILIAS should do itself.

corro: @chfsx I usually don't encounter any issues when converting databases to InnoDB. A simple statement per table usually does the trick:
```
ALTER TABLE xyz ENGINE=InnoDB
```

It does take some time to convert the tables though. In my experience it can take up to 30min to convert a large ILIAS database. Apart from that, I don't have any concerns doing such a conversion as part of an update step.

A migration task seems less fitting to me, since ILIAS will not be accessible while important tables are being converted. The way I understand it, migrations are meant to be running in the background without impacting access to the application.

mjansenDatabay: Thx @corro, that matches the experiences of our admins when migrating complete ILIAS databases to `InnoDB`.

gvollbach: Done, also conflicts from other merge are resolved

gvollbach: @chfsx Thanks for your input, the errors get now shown in the cli setup. 
@corro Thanks for the information :)

mjansenDatabay: @gvollbach / @chfsx How can we proceed from here?

Suggestion: If you @chfsx don't have a strong opinion here, let's discuss/decide this on the next JourFixe, if the `InnoDB` migration could be one the exceptions where a new database step may be added to `dbupdate_05.php` (which is my personal preference).

chfsx:
Hi @gvollbach I'd like discuss some points concerning this PR in a virtual meeting and will send you some possible dates for a meeting via mail. thanks in advance!

chfsx:
@all: we discussed the party of "where should this update-step be located". there are 3 options we discussed:
- in the dbupdate-Script (where it's located atm)
- as a "new" updatestep
- as a migration

The switch of the db-engine should not be done during productive usage of the installation which excludes a migration. there is no need to perform this step before all other updatesteps which excludes the dbupdate-Script. Therefore, a "new" update-step it is :-) @gvollbach will adopt the PR, thanks a lot!

chfsx (changes requested):
please move the updatestep out of the dbupdate to a separate step, thanks in advance!

chfsx:
Hi @gvollbach ! Is there anything I can do/help with the changes? thanks in advance!

gvollbach:
@chfsx No, thx. I think you will get the new pull request at the beginning of next week, I would like to test some things before the request is ready for review.

gvollbach:
New pull request: https://github.com/ILIAS-eLearning/ILIAS/pull/3946

PR #3710, opened 2021-10-15
Static ilCtrl

thibsy:
This pull request changes mainly these things:

- the storage of ilCtrl's call structure moved from the database to an artifact, which is built when running `composer install` or `composer dump`.
- the identification of base classes no longer depends on service.xml or module.xml, or on the database tables service_class and module_class. I introduced a new interface `ilCtrlBaseClassInterface` which is gathered by the setup's interface collector and stored in an artifact as well. Similar to the call structure, the artifact will be generated by composer.
- `ilCtrl::getCmd()` no longer accepts an array of safe commands, that implementation has moved to a GUI class itself. For that functionality I introduced a new interface `ilCtrlSecurityInterface` as well. The usage and behaviour is described below. 
- ilCtrl's method arguments as well as their return values don't use empty strings anymore. Instead, null will be returned or accepted. This breaks a lot of stuff in regards to `ilCtrl::getNextClass()`, `ilCtrl::getCmd()`, whose return values are often directly used within PHPs `strtolower` function. If wished, I can add a null coalescing operator for a defined period, but in the end, I won't change it entirely.
- Multiple methods have been deprecated, which I intend to remove entirely for the next ILIAS release. If there are any edge cases that will not work anymore without these, please let me know.

`ilCtrlSecurityInterface` is intended to be implemented by GUI classes within ilCtrl's call structure. At the moment only one method is prescribed, but there might be added some more in the future. The one method that must be implemented right now, is `getSafeCommands`. This method should return a list of commands (string-array) that are considered safe. Safe in an ilCtrl context means, the command can be performed without checking for cross-site-request-forgery (CSRF). If a command is not returned by that method, ilCtrl automatically appends a cryptographically secure CSRF token, when `ilCtrl::getFormActionByClass()` or `ilCtrl::getLinkTargetByClass` is called. That same process can be found again when retrieving a command with `ilCtrl::getCmd()`, which will automatically perform CSRF validation if the command is not returned by `getSafeCommands()`. If the validation passes the command is returned, but if the validation fails, the provided fallback command or NULL will be returned. If a GUI class does not yet implement this interface, the old behaviour applies and the CSRF-token and -validation are only applied to POST link targets, generated with `ilCtrl::getFormActionByClass()`.


Services/Container/classes/class.ilContainerGUI.php
```diff
@@ -918,7 +918,7 @@ protected function getTabs()
         if ($rbacsystem->checkAccess('edit_permission', $this->ref_id)) {
             $this->tabs_gui->addTarget(
                 "perm_settings",
-                $this->ctrl->getLinkTargetByClass(array(get_class($this), 'ilpermissiongui'), "perm"),
+                $this->ctrl->getLinkTargetByClass(array($_GET['baseClass'], get_class($this), 'ilpermissiongui'), "perm"),
```
iszmais: It might be better to call the Query params here with the given dic functions as long as the surrounding enviroment allows it.

Services/Init/classes/class.ilInitialisation.php
```diff
@@ -1631,7 +1631,7 @@ protected static function blockedAuthentication($a_current_script)
 
         $requestBaseClass = strtolower((string) ($_REQUEST['baseClass'] ?? ''));
         if ($requestBaseClass == strtolower(ilStartUpGUI::class)) {
-            $requestCmdClass = strtolower((string) ($_REQUEST['cmdClass']) ?? '');
+            $requestCmdClass = strtolower((string) ($_REQUEST['cmdClass'] ?? ''));
```
iszmais, line 1651: This could also be called by using the DIC

Services/Container/classes/class.ilContainerGUI.php
iszmais: I Also woudl recommend to switch to square bracket array implementations to increase the readability compared to the function wise implementation.

Services/Search/classes/class.ilMainMenuSearchGUI.php
```diff
@@ -100,12 +96,21 @@ public function getHTML()
...
+
+        // temporary fix until $ilCtrl functions flawlessly.
+        $token = new ilCtrlToken(
+            $DIC->database(),
+            $DIC->user()
+        );
+
+        // @TODO: user $ilCtrl->getFormAction() instead.
+        $this->tpl->setVariable('FORMACTION', 'ilias.php?base_class=ilSearchController&cmd=post' .
+            '&csrf_token=' . $token->getToken() . '&cmd_fallback=remoteSearch');
```
iszmais: I would recommend using the http_build_query function to increase fututre readability and scalabitlity

Services/Search/classes/class.ilMainMenuSearchGUI.php
```diff
@@ -100,12 +96,21 @@ public function getHTML()
...
+
         $this->tpl->setVariable('BTN_SEARCH', $this->lng->txt('search'));
         $this->tpl->setVariable('SEARCH_INPUT_LABEL', $this->lng->txt('search_field'));
-        $this->tpl->setVariable('AC_DATASOURCE', "ilias.php?baseClass=ilSearchController&cmd=autoComplete");
+        // @TODO: user $ilCtrl->getFormAction() instead.
+        $this->tpl->setVariable('AC_DATASOURCE', "ilias.php?base_class=ilSearchController&cmd=autoComplete");
```
iszmais: I would recommend using the http_build_query function to increase fututre readability and scalabitlity

Services/UIComponent/CharSelector/classes/class.ilCharSelectorGUI.php
```diff
@@ -164,8 +164,8 @@ public function addFormProperties(ilPropertyFormGUI $a_form) : void
                 break;
         }
     }
-    
-    
+
+
```
iszmais: While you are at it to correct the indentation i would recommend to also correct the spacings lines in numbers (decrease all "normal" spacings to 1 line).

(For all following occurences, too)

Services/UICore/classes/Paths/class.ilCtrlAbstractPath.php
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * @var ilCtrlException|null
+     */
+    protected ?ilCtrlException $exception = null;
+
+    /**
+     * @var string|null
+     */
+    protected ?string $cid_path;
```
iszmais: If this should be an optional string, it may also set initially as null (just like the one above) for clearification.

Services/UICore/classes/Paths/class.ilCtrlAbstractPath.php
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Class ilCtrlAbstractPath
+ *
+ * @author Thibeau Fuhrer <thf@studer-raimann.ch>
+ */
+abstract class ilCtrlAbstractPath implements ilCtrlPathInterface
+{
+    /**
+     * @var ilCtrlStructureInterface
```
iszmais: I would recommend removing unnecessary doc param and var declarations that are already given throught the newly implemented strict types.

()

Services/UICore/classes/Paths/class.ilCtrlAbstractPath.php
```diff
@@ -0,0 +1,173 @@
+<?php
```
iszmais: This is missing a strict type declaration and License

Services/UICore/classes/Paths/class.ilCtrlArrayClassPath.php
```diff
@@ -0,0 +1,102 @@
+<?php
```
iszmais: This is missing a strict type declaration and License

Services/UICore/classes/Paths/class.ilCtrlExistingPath.php
```diff
@@ -0,0 +1,22 @@
+<?php
```
iszmais: This is missing a strict type declaration and License

Services/UICore/classes/Paths/class.ilCtrlNullPath.php
```diff
@@ -0,0 +1,65 @@
+<?php
```
iszmais: This is missing a strict type declaration and License

Services/UICore/classes/Paths/class.ilCtrlPathFactory.php
```diff
@@ -0,0 +1,59 @@
+<?php
```
iszmais: This is missing a strict type declaration and License

Services/UICore/classes/Paths/class.ilCtrlSingleClassPath.php
```diff
@@ -0,0 +1,85 @@
+<?php
```
iszmais: This is missing a strict type declaration and License

Services/UICore/classes/Setup/class.ilCtrlBaseClassArtifactObjective.php
```diff
@@ -0,0 +1,40 @@
+<?php
```
iszmais: This is missing a strict type declaration and License

Services/UICore/classes/Setup/class.ilCtrlBaseClassArtifactObjective.php
```diff
@@ -0,0 +1,40 @@
...
+    public function build() : Setup\Artifact
+    {
+        $finder = new Setup\ImplementationOfInterfaceFinder();
+
+        $base_classes = [];
+        foreach ($finder->getMatchingClassNames(ilCtrlBaseClassInterface::class) as $base_class) {
+            $base_classes[] = strtolower($base_class);
+        }
+
+        return new Setup\Artifact\ArrayArtifact($base_classes);
```
iszmais: You could cleanup the paths so all of the are fully covered inside the use blog. This is just a personal preference and no hard convention but it clearifies the usage of a class more strict and is implemented in many IDEs cleanup functions

Services/UICore/classes/Setup/class.ilCtrlStructureArtifactObjective.php
```diff
@@ -0,0 +1,38 @@
+<?php
```
iszmais: This is missing a strict type declaration

Services/UICore/classes/Setup/class.ilCtrlStructureReader.php
```diff
@@ -1,443 +1,294 @@
...
-
     /**
-     * @return null|(string,string[])
+     * Helper function that returns all children references.
+     *
+     * @param ReflectionClass $reflection
+     * @return array
      */
-    protected function getIlCtrlCalls(string $content) : ?array
+    private function getChildren(ReflectionClass $reflection) : array
```
iszmais: I dont see the neccesarity of using privates instead of protected here, since this function shoudl also be available in an heritage

Services/UICore/classes/Setup/class.ilCtrlStructureReader.php
```diff
@@ -1,443 +1,294 @@
...
 
     /**
-     * @return null|(string,string[])
+     * Helper function that returns all parent references.
+     *
+     * @param ReflectionClass $reflection
+     * @return array
      */
-    protected function getIlCtrlIsCalledBy(string $content) : ?array
+    private function getParents(ReflectionClass $reflection) : array
```
iszmais: I dont see the neccesarity of using privates instead of preotected here, since this function shoudl also be available in an heritage

Services/UICore/classes/Setup/class.ilCtrlStructureReader.php
```diff
@@ -1,443 +1,294 @@
 <?php
```
iszmais: This is missing a strict type declaration and License

Services/UICore/classes/Setup/class.ilUICoreSetupAgent.php
```diff
@@ -1,21 +1,17 @@
-<?php declare(strict_types=1);
+<?php
```
iszmais: This is missing a strict type declaration and License

Services/UICore/classes/class.ilCtrl.php
```diff
@@ -1,1890 +1,1116 @@
 <?php
```
iszmais: This is missing a strict type declaration and License

Services/UICore/classes/class.ilCtrlContext.php
```diff
@@ -0,0 +1,286 @@
+<?php
```
iszmais: This is missing a strict type declaration and License

Services/UICore/classes/class.ilCtrlStructure.php
```diff
@@ -1,106 +1,318 @@
...
+
+    /**
+     * Returns a stored structure value of the given key from the
+     * corresponding class mapped by CID.
+     *
+     * @param string $identifier_key
+     * @param string $cid
+     * @return array|string|null
+     */
+    private function getValueForKeyByCid(string $identifier_key, string $cid)
```
iszmais: It would be nice if this would also had a return type. You could maybe define an object for that wich wrapps an array or string return, since this seems to occure very common

Services/UICore/classes/class.ilCtrlToken.php
```diff
@@ -0,0 +1,156 @@
+<?php
```
iszmais: This is missing a strict type declaration and License

Services/UICore/interfaces/interface.ilCtrlBaseClassInterface.php
```diff
@@ -0,0 +1,24 @@
+<?php
```
iszmais, line 1: This is missing a strict type declaration and License

Services/UICore/interfaces/interface.ilCtrlContextInterface.php
```diff
@@ -0,0 +1,143 @@
+<?php
```
iszmais, line 1: This is missing a strict type declaration and License

Services/UICore/interfaces/interface.ilCtrlInterface.php
```diff
@@ -0,0 +1,486 @@
+<?php
```
iszmais, line 1: This is missing a strict type declaration and License

Services/UICore/interfaces/interface.ilCtrlPathInterface.php
```diff
@@ -0,0 +1,86 @@
+<?php
```
iszmais, line 1: This is missing a strict type declaration and License

Services/UICore/interfaces/interface.ilCtrlSecurityInterface.php
```diff
@@ -0,0 +1,47 @@
+<?php
```
iszmais, line 1: This is missing a strict type declaration and License

Services/UICore/interfaces/interface.ilCtrlStructureInterface.php
```diff
@@ -0,0 +1,185 @@
+<?php
```
iszmais, line 1: This is missing a strict type declaration and License

Services/UICore/interfaces/interface.ilCtrlTokenInterface.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
iszmais, line 1: This is missing a strict type declaration and License

studip_referrer.php
```diff
@@ -68,7 +68,7 @@
             $_POST['new_type'] = $_GET['type'];
             $_POST['cmd']['create'] = 'add';
             $_GET['cmd'] = 'post';
-            $_GET[ilCtrl::IL_RTOKEN_NAME] = $ilCtrl->getRequestToken();
+            $_GET[ilCtrl::PARAM_CSRF_TOKEN] = $ilCtrl->getRequestToken();
```
iszmais: It might be better to call the Query params here with the given dic functions as long as the surrounding enviroment allows it.

iszmais (commented):
Just basic feedback to code style. Nothing conceptional.

Feel free to correct me if im wrong

Services/UICore/classes/class.ilCtrl.php
```diff
@@ -1,1890 +1,1116 @@
...
     }
 
     /**
-     * Save classes respective to the class id's of a node and store corresponding
-     * class calls for future reference.
+     * This helper function wraps the deprecated UI functionality that
+     * modifies a URL target and "hacks into" existing HTML.
      *
-     * @param	string	$a_node
+     * (Tbh I don't really get that mechanism, but it stays for now.)
```
mjansenDatabay: With this mechanism UI-Hook plugins could intercept the intended HTTP redirect and overwrite the actual 302 URL. So instead of routing the user to URL A, a plugin could decide to route the user to URL B instead.

For me it looks like the refactoring here changes the behaviour of this method.

If a plugin does **not** want to`ilUIHookPluginGUI::KEEP` the input parameters, the `$target_url` MUST be overwritten. With your changes you only overwrite `$target_url` if a plugin whould like to `KEEP` the input parameters. This is wrong IMO.

thibsy: I think you're right, will take another look at that mechanism, thanks!

PurHur: Please dont remove this hook. Feel free to change it but i need the feature. Its quiet important to a lot of hacky things i do in some projects. I am using this as a after init hook.

Services/UICore/classes/class.ilCtrlStructure.php
```diff
@@ -1,106 +1,318 @@
...
 /**
- * The CtrlStructure knows how GUIs call each other and where the code of the
- * guis is located.
+ * Class ilCtrlStructure holds the currently read control
+ * structure.
+ *
+ * @author Thibeau Fuhrer <thf@studer-raimann.ch>
  */
-class ilCtrlStructure
+final class ilCtrlStructure implements ilCtrlStructureInterface
```
PurHur: I actually dont know if this class is the new 'ilCtrl'. In the past i needed to extend this. I know the implications of extending such base classes a lot. I promise i dont write bug reports for extending this :) Does it really need to be a final class?

Services/UICore/classes/class.ilCtrl.php
thibsy: @PurHur I don't think that this method will stay forever, as the method `ilUIHookPluginGUI::getHTML()` is already marked as deprecated and should not be used in the future. The mechanism depends on this method and I'm not sure if there will be methods to replace this behaviour.

Services/UICore/classes/class.ilCtrlStructure.php
thibsy: May I ask why you needed to extend `ilCtrlStructure`? IMO you don't have to manipulate the call-structure, as it's read properly if all the necessary `@ilCtrl_Calls` or `@ilCtrl_isCalledBy` statements were made.

PurHur: I did that to make a fancy routing plugin. See: https://training.mtm.org/kursangebot/firmenspezifisch-11.html

So you are able to overwrite all ilias routes to make them static. This requires a lot of chained hacks to make it even work. Its nothing you want in the Core =)

I talked with @mjansenDatabay about this thuesday and he also showed me a plugin where he used a hack to manipulate some Tabs in a Ilias Object. 

If this class would be final, what would be the possibility to make such things happen? I actually would change the final on runtime or even copy the whole class into a plugin. 

Bute note: I could have the wrong class in mind. But still i am not a fan of final. Another example is that i am extending ilInitialisation. I know that i shouldnt do that =)

Services/UICore/classes/class.ilCtrl.php
PurHur: Feel free to replace this method with another hook system. I am totally on your side that this should not be named "getHTML". It took me ages to find this hook in the first place. Change it as you like but remain the feature. All plugins now can listen on all events. That could be a direction

PurHur: Also how is deprecation handled in Ilias? I did not got a deprecation warning yet. 

@mjansenDatabay 
We use the typo3 system to deprecate methods. I did not see that in ilias. I think you are the right developer to push this topic in the right direction. Please see this link: 
https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/Deprecation/Index.html#deprecate-functions-in-extensions

Services/UICore/classes/class.ilCtrlStructure.php
mjansenDatabay: @PurHur What I did (in a plugin) was:

- Create an anonymous subclass of `ilCtrl` (which is possible because it is currently not final), which overrides all `get(LinkTaget|FormAction)(byClass)?` methods
- These methods don't call the parent method, but put all arguments in a map by using a hash as key, where this hash string is also returned to the caller
- Inject this into the `DIC`
- Call a method of `ilObj*GUI` to populate `ilTabsGUI` with all tabs of a repository object (e.g. a course or test) from an `UserInterfaceHook` plugin context
- Because I exchanged the `ilCtrl` instance, retrieving the tabs did not result in an error
- I could display the tabs in the user interface of my plugin after I processed the stacked `ilCtrl` information, without being in a real (e.g.) course context

Reason for this: Repository objects have no mechanism/interface to expose their user interface tabs

Services/UICore/classes/class.ilCtrl.php
mjansenDatabay: @PurHur

> Also how is deprecation handled in Ilias? I did not got a deprecation warning yet. 


We mark the class/interface/method/you name it as `@deprecated`.

PurHur: Is there a possibility to trigger a real deprecation error?
Since i am managing a lot of code i cant read all of it all the time. This is an impossible task for me.  A deprecation notice would be sended via https://sentry.io/ to me.

PurHur: I would really like that. And i would join the TA for that discussion if you like the idea behind that. This is not related to the headline of this PR.

matthiaskunkel:
Jour Fixe, 15 NOV 2021: We highly appreciate this suggestion and accept the PR for trunk.

Services/UICore/classes/class.ilCtrlStructure.php
thibsy: BTW: I removed the final keyword in the latest version

PurHur: Thank u

PurHur: There should be a discussion in the jourfix/technical board about the general usage of final. But i am not the person who decides that.

klees: Hi @PurHur,

this sounds like an interesting topic. May I ping you for a presentation weighing pros and cons (or something similar) for the next Dev-Conf in ~March?

Best regards!

colin-kiegel:
@thibsy Hey Thibeau, I understand that this moves the Control-Structure to the filesystem - out of the database.

But does this also address the fact that _local_ code changes lead to _global_ changes of all cmdNodes, invalidating a lot of URLs?

I am asking, because I always felt this behaviour is very painful and completely unnecessary. It could easily be avoided if regenerated cmdNodes would default to their previous value, IF the "expanded" routing information is identical.

Thanks for clarification / your thoughts!

chfsx:
@colin-kiegel this is just a refactoring of ilCtrl and it has the same behaviour as before (new nodeIds possible after reloading structure). persisting ids would be a nice feature in some cases but there are better options for persisting links (new request handling/routing)

PR #3711, opened 2021-10-18
UI/Mainbar: 31492, move entire li-tag instead of button only (on resize)

nhaagen:
https://mantis.ilias.de/view.php?id=31492


klees (approved):
Hi @nhaagen,

thanks for looking into this. This won't play well with templating, but the Main Bar doesn't play well there already AFAIK, so this is fine for me. I still ask @Amstutz for his opinion.

Best regards!

Amstutz:
thx a lot! looks good. 

note: for future fixes, we will start asking to provide also a js test making sure, this will not happen again. since the js testing env, especially rendering tests are currently in discussion in an other PR, I think we are fine here. But we really need more coverage for js logic.

PR #3712, opened 2021-10-18
D&D Fileupload

thibsy:
This pull request introduces a new sort of Kitchensink inputs, described by the interface `AdditionalInputsAware`.

An `AdditionalInputsAware`-input is able to be provided with any other sort of `Input`, which serves as a template for subordinate inputs generated by this input. Subordinate in this context means, that the template is used to generate additional-inputs when `AdditionalInputsAware::withValue()` or `AdditionalInputsAware::withInput()` is called. Sort of like the `Group`-input, the input holds a list of subordinate Inputs - with the difference, that they can also be generated from client-side as well. For this reason a new `SubordinateNameSource` was introduced, that generates inputs stacked to an already existing one. So for example, an input name could look like `form_input_1[{INDEX}][sub_input_1]`, whereas `{INDEX}` stands for a placeholder that needs to be set when rendering subordinate Inputs on both, client- and server-side. This leads to POST results structured like this: 
```
array(
    'form_input_1 => [
        0 => [
            'sub_input_1' => value,
             ...
        ],
    ],
);
```
Now this means that subordinate inputs can be generated on client-side and they are stacked on the parent Input.

The use-case of client-side generated Inputs came along with the feature-request of refactoring the D&D dropzones and Module/File classes, where `File`-inputs must be able to hold further information about a file that has been uploaded from the computer, for example new title and additional description or whether it should be extracted or not etc. 

When implementing this I stumbled across several points where Kitchensink would be very limited to allow that sort of use-case, so I had to "patch things up a bit". I have commented below what code-snippets I mean with that.


chfsx:
@thibsy if possible I would suggest to separate the changes in UI Inputs from the needed changes in the fileupload/fileobject to have a sleeker overview of the changes needed in UI-Inputs

src/UI/Implementation/Component/Input/Field/File.php
```diff
@@ -2,114 +2,439 @@
...
+     * @param InputFactory  $factory
+     * @param DataFactory   $data_factory
+     * @param ilLanguage    $lang
+     * @param Refinery      $refinery
+     * @param UploadHandler $upload_handler
+     * @param string        $label
+     * @param string|null   $byline
+     * @param bool          $with_zip_options
+     */
     public function __construct(
```
thibsy, line 83: Instead of providing an input-template by `withTemplateForAdditionalInputs()`, it should be passed in the constructor, otherwise `withValue()` depends on whether the method is called or not. That was the same reason for creating the file-inputs zip-options already in the constructor, so if both, template and zip-options are enabled the options can be merged with the template. Passing the template to the constructor would solve that problematic agreed?

src/UI/Implementation/Component/Input/Field/File.php
```diff
@@ -2,114 +2,439 @@
...
+                        $values[$file_id][$key] = $value;
+                    }
+                } else {
+                    $values[$file_id][] = $template_value;
+                }
+
+                // file-info is appended as last array entry, in
+                // order to "consistently" fetch it with PHPs
+                // array_key_last() function.
+                $values[$file_id][] = $file_info;
```
thibsy, line 278: `File::getValue()` consists of a mixture of integer and string array-key-indexes, because if a `Group` input was provided as a template, the input's array might be associative. The `File`-inputs zip-options can certainly not be, because then they might collide with other keys when merging them into the template-group (`File::mergeInputWithZipOptions()`). Therefore the actual file-info is now appended with the `[]` operator.

src/UI/Implementation/Component/Input/Field/File.php
```diff
@@ -2,114 +2,439 @@
...
+                $data = [];
+                foreach ($file_data as $key => $value) {
+                    if ($key !== $file_identifier) {
+                        $input_name = "{$clone->getName()}[" . SubordinateNameSource::INDEX_PLACEHOLDER . "][$key]";
+                        $data[$input_name] = $value;
+                    }
+                }
+
+                $template = ($template) ?: $clone->getZipExtractOptionsTemplate();
+                $template = $template->withInput(new ArrayInputData($data));
```
thibsy, line 321: I needed to introduce a new sort of `InputData`, because the client-side generated input names are now split up into an array mapped to the parent input. Therefore I had to "rebuild" their original name and map their values to it in a separate array, which I then passed along the newly introduced `ArrayInputData`-object. This way it was possible to delegate the handling of operations and validation to the inputs themselves.

klees (changes requested):
Hi @thibsy,

thanks for improving the implementation of the file input capabilities of the ILIAS UI framework. Please note that this is only a review from me, not from @Amstutz and me as we mostly do for PRs to the UI framework.

I am a little sorry that it took me so long, but this is a lot to digest. Maybe we could make faster progress if we would concentrate on the `AdditionalFormInputAware`-part first and move it to a single PR.

Feel free to contact me for another discussion, as always. I hope this helps.

- [ ] FileDropzone::getData: What is it that we will return on this method? Could we hint on (or at least document) something better than just `array`?
- [ ] FileDropzone::withUploadButton: Are we sure we can remove this? As far as I remember, this was introduced to customize the appearance and title of the button. Did we get rid of the use cases?
- [ ] SubordinateNameSource: Why do we need this [const](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-8fa5adc903b33bfeb00464fa82427a909c114077aac9a3ba4f1a41ddcd3de9f2R17) if it is always the same? Couldn't we just drop this level in the nested array?
- [ ] Selectors: Do we really need that much selectors in the JS? This will make the connection between the template and the js code brittle. Could we use selectors that are distinct from the selectors that we use for styling to introduce some separation of concerns here?
- [ ] JavaScript 1: Would this JS work for multiple inputs on the same page?
- [ ] A11y: Did you do any accessibilitiy assessment with the new implementation? If so: What was the outcome?
- [ ] FormInput: Why did you remove the `extends FormInput` on `Input\Field\File`?
- [ ] UploadHandler::getInfoResults: Why do we need that new method?
- [ ] Required: Will it really be correct to assume, that sub inputs are required if the parent is required (see [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-ecc5cce787348f8b383528c542fba119f75a5fca27d9b3ba14495f1de4a819deR74)). This feels wrong. This seems to be correct for disabled inputs, though.
- [ ] Sub Content 1: Nothing seems to happen if sub content of a `AdditionalFormInputAwareInput` [is not ok](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-ecc5cce787348f8b383528c542fba119f75a5fca27d9b3ba14495f1de4a819deR136). Is this correct?
- [ ] Sub Content 2: [This](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-ecc5cce787348f8b383528c542fba119f75a5fca27d9b3ba14495f1de4a819deR138) looks as if you unwrap sub content in an array. Why would we want to do this? Can't we keep it wrapped in that array? This feels like a surprising special case for fields that return an array.
- [ ] Sub Content 3: Do we really need to allow `null` for [`AdditionalFormInputAwareInput::$additional_inputs`](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-ecc5cce787348f8b383528c542fba119f75a5fca27d9b3ba14495f1de4a819deR35)? Using an empty array instead would make the code simpler [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-ecc5cce787348f8b383528c542fba119f75a5fca27d9b3ba14495f1de4a819deR188). 


- [ ] File(Dropzone): Why did you rename `Dropzone/File/File` to `Dropzone/File/FileDropzone` (hence, repeating the Dropzone)? Stick to the existing name...
- [ ] Removed Constructors: Why did you remove `Field\Factory::url` and `::link` from the factory of the input fields? Just an oversight, I guess?
- [ ] Checkbox::withValue: I think the change there is problematic. The `$value` should strictly be `bool` there, not anything else. Why did you remove this?
- [ ] WTF: is [this](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-11283614eb1a80aab33a9d2c520f66aa468f09f15caa5ee7518123e2091df16fR18)?!? =D
- [ ] Static: Please don't use [`static`](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-f8dba1dd818b533395780a15f6d1243674aaac83caa54a7bdd826b3b4b39ae70R15). Inject dependencies instead (see "DI" below).
- [ ] Rendering 0: Please try to modularize the rendering code a little by introducing methods. The code is extremely hard to read like this...
- [ ] Rendering 1: I think the file input won't show a disabled version. At least I don't understand how this would actually happen =)
- [ ] Rendering 2: There is HTML in the `Renderer` which should be moved to the respective template. I would expect that the code will look cleaner afterwards as well.
- [ ] Javascript 2: It looks as if we could get rid of jquery and just use standard browser APIs. Please have a look into this.
- [ ] Responsiveness/Bootstrap: Only judging from the code it looks as if the responsiveness of the new implementation is worse than before, because the Bootstrap grid classes are used and we will likely see grid in grid effects somewhere. Also, we are looking to get rid of Bootstraps. So I would rather like you to remove the grid and Bootstrap classes and move to another HTML structure. Feel free to ask for support again (if you need any) for appropriate style code.
- [ ] Template: Why do we need that doubly wrapped div in `tpl.wrapper.html`? Could we get away with one div to get a little more semantic html?
- [ ] Naming: Would `HasDynamicSubFields` be a name that would be also correct for `AdditionalFormInputAware`? I deem this to be understandable more easily.
- [ ] Inheritance: Could we use something else then a base class to implement `AdditionalFormInputAware(Input)`? Preferably a standalone object that could be used by an Input Field that `HasDynamicSubFields`. If that feels odd, maybe a trait would do. Only if these two are not feasible we should use the current subclassing approach. The `use composition over inheritance` rule of thumb is especially strong in PHP, where we only can have one base class. The Input Fields (and the UI components in general) are things where we very likely will have various properties like `HasDynamicSubFields` on one component and we will not be able to implement them all via inheritance. Hence `interface`s for declaring the properties (we can have multiple there) and composition for implementing them.
- [ ] Null Object: Maybe the problem you mention in your comment ("Instead of providing an input-template by withTemplateForAdditionalInputs(), [...]") could be solved by introducing a [null object](https://www.tutorialspoint.com/design_pattern/null_object_pattern.htm) for an Input Field. This should make your code in [`AdditionalFormInputAwareInput::withInput`](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-ecc5cce787348f8b383528c542fba119f75a5fca27d9b3ba14495f1de4a819deR115) simpler and would come in handy in various other cases as well (not related to this PR).
- [ ] Merge Inputs: [`File::mergeInputWithZipOptions`](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-11283614eb1a80aab33a9d2c520f66aa468f09f15caa5ee7518123e2091df16fR428) seems to be overly complicated. Can't we just wrap the two inputs in a group and use that? If you would just pass this new group on to `parent::withTemplateForAdditionalInputs` in `File::withTemplateForAdditionalInputs`, the code in `File::withValue` should become considerably simpler.
- [ ] Data Structures: It looks as if some more datastrucutures could make your code simple. E.g.: Introduce a `ZipOption` class that bundles your zip options. You won't confuse them with other fields from the additional inputs anymore. Or: Build a structure `FileResult` that has the fields `$file_info`, `$zip_options` and `$additional_info`. You won't be confusing array keys anymore. This is mostly a reaction to [this comment](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#r731807157), but also an idea for some intricate array-merging code snippets. 


- [ ] Docstrings: Remove `@package` and the meaningless `Interface...` from the docstrings of the interfaces and classes. Remove `@return` and `@param` if they do not add anything to the PHP typehints.
- [ ] `Data\DataSize`: Use `Data\DataSize` to encode data sizes.
- [ ] Mimetypes: No super strict requirement, but would be really really appreciated: Could you introduce a type in `src/Data` for the mimetypes as well?
- [ ] Dependency Injection: Please inject the dependency [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-7da0060b76695771e40e879022750775741454de17057dfe585acc16a4887174R47). Objects should either use or create objects, not both.
- [ ] If: [This](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-236ec82c3b735183d86d2fd3f68c5c9c40d0dc87796317ec3703a6b71b1e3bc6R197) is superfluous. If the array is empty, we just won't iterate on the next line.
- [ ] Coding Style: Please use the ILIAS coding style for autoformatting, not less not more. E.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-f3892e7d92b0895c6b153d2f1754b755fc888a4b5199b1d4f8784a7344b81d5fR604) or [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-516004f06bdf85d36882a8ab2c09ac0936f360dfa9f88ec51523f7dfb2b54968R109).
- [ ] FileInputTest1: The new tests should not be working, I guess. The fourth parameter to the `file`-constructor seems to be [`$with_zip_options`](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-f857fabc59021b54422647cb7c8c77b1cb280a7b70024c1df86571f02863b2eeR612), not max-files.
- [ ] FileInputTest2: We should [reject nonsensical arguments](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-63b53dda562adbd47190b76d2dd456d70b59eb88052cb3001c1043d7ccde6c01R376) instead of defaulting them to something else.
- [ ] JavaScript 3: Please use [rollup.js](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/docs/development/js/js-bundling.md) and [our coding style](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/docs/development/js/js-coding-style.md) and write [unit tests](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/docs/development/js/js-unit-test.md) for the js-code. Feel free to contact @nhaagen for some insights in how js code could be structured to improve modularization and unit testing. The JS currently is hard to follow, maybe this could be improved via some modularization and testing as well.
- [ ] Todo: There is one [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-6e5d62d076afb2a11175ad22e23f9c622ad0b95a648a70f5a0be210258c302c5R48) =)
- [ ] Check: [This check](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-ecc5cce787348f8b383528c542fba119f75a5fca27d9b3ba14495f1de4a819deR42) is already done by PHP.
- [ ] withInput 1: Use [the "return early pattern"](https://medium.com/swlh/return-early-pattern-3d18a41bba8) in [`AdditionalFormInputAwareInput::withInput`](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-ecc5cce787348f8b383528c542fba119f75a5fca27d9b3ba14495f1de4a819deR115) and [`File::withInput`](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-11283614eb1a80aab33a9d2c520f66aa468f09f15caa5ee7518123e2091df16fR213).
- [ ] withInput 2: [`AdditionalFormInputAwareInput::withInput`](https://github.com/ILIAS-eLearning/ILIAS/pull/3712/files#diff-ecc5cce787348f8b383528c542fba119f75a5fca27d9b3ba14495f1de4a819deR115) doesn't seem to respect the "required" property. If there is no input, it will happily return `Ok(null)` as content, even if the input is required. It should report an `Error` instead.

I really look forward to see this moving on.

Best regards!

Amstutz:
Just a minor additon: 

- [ ] Unit Tests: I am somewhat skeptical that the 27 lines changed in the tests really reflect the changes of the implementation. Please try to cover all changes. 

Just a note, as  of today, we do not strictly require js tests. But this might change soon, see: https://github.com/ILIAS-eLearning/ILIAS/pull/3685 . You do not have to, but some js tests would be nice as well.

PR #3715, opened 2021-10-21
made file migration compatible with Windows

simoor:
cheers to @chfsx 🍻

### DIFF
--- a/Modules/File/classes/Setup/class.ilFileObjectToStorageDirectory.php
+++ b/Modules/File/classes/Setup/class.ilFileObjectToStorageDirectory.php
@@ -45,15 +45,15 @@ class ilFileObjectToStorageDirectory
                 ),
                 RecursiveIteratorIterator::LEAVES_ONLY
             ),
-            '/.*\/file_[\d]*\/([\d]*)\/(.*)/',
+            '/.*(\/|\\\)file_[\d]*(\/|\\\)([\d]*)(\/|\\\)(.*)/',
             RegexIterator::GET_MATCH
         );
 
         $this->versions = [];
 
         foreach ($g as $item) {
-            $version = (int) $item[1];
-            $title = $history_data[$version]['filename'] ?? $item[2];
+            $version = (int) $item[3];
+            $title = $history_data[$version]['filename'] ?? $item[5];
             $action = $history_data[$version]['action'] ?? 'create';
             $owner = $history_data[$version]['owner_id'] ?? 13;
             $creation_date_timestamp = strtotime($history_data[$version]['date'] ?? '0');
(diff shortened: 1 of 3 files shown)


mjansenDatabay:
Hi @simoor,

thanks a lot for your contribution. Could you please create a Mantis issue (according to https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/contributing.md#pull-request-to-the-repositories) for that and describe the actual problem. Thanks in advance.

Best regards
@mjansenDatabay

simoor:
@mjansenDatabay
The regex only matches forward slash, and on windows environment backslashes are used. So it doesn't detect any files on windows environment and it's not possible to do a migration. @chfsx knows already about the issue.

PurHur:
I still wonder after years is there a contribution rule that you have to make a mantis report. Last time i checked its not a rule of the contribution guidelines. 
This makes the contribution even more less to happen. The first post in all the issues of new developer are often "Do a mantis report" but its stated nowhere. I also dont want that rule to be real.....

mjansenDatabay:
@PurHur 

> I still wonder after years is there a contribution rule that you have to make a mantis report. Last time i checked its not a rule of the contribution guidelines. This makes the contribution even more less to happen. The first post in all the issues of new developer are often "Do a mantis report" but its stated nowhere. I also dont want that rule to be real.....

https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/contributing.md#rules-for-contributors

> If your PR contains a bugfix please reference the number of the mantis ticket in the title `12345 - To many spaces`, link the ticket in the description and label the ticket with bugfix. You may make one PR per affected branch.

Amstutz:
Hi @PurHur 

Thx for the feedback on the process. Note, that not for every contribution, a mantis issue is needed. There is also the category "improvement", which is described in the above linked guideline as: "The label improvement is used for PRs which propose a general improvement of code or documentation which is not related to a bug."

This is not strictly black and white, many cases borderline between bug and improvement. The reason for the need to have an issue related to the "bugfix" label, is that we (whoever is interested) can then track properly why a change has been made.  IMO it's a good why of seperating the issue from the solution. E.g. the issue might be valid, but the solution is for some reason not applicable. This would allow to close the solution, while having still an open issue at hand to track and generate further progress on. 

At the end of the day, it's just one of multiple way of handling things. When it comes to process definitions, I believe best is to "pick one, stick with it" and follow it consistently to make it easier for others to follow. 

Hope this helps somewhat in understanding the reasoning, and again, thx for sharing feedback. always appreciated.

simoor:
@chfsx kannst du schauen dass die Korrektur reinkommt? Zuviel Diskussion hier...

chfsx:
i'll take care of it, sorry for the delay

chfsx:
@simoor FYI, Windows support has been implemented in the latest 7.x release

PR #3716, opened 2021-10-24
deleted composer lib zendframework/zend-httphandlerrunner and patch

eqsoft:
Hi,
this PR removes zendframework/zend-httphandlerrunner and the patch. I am unsure about the changes in composer.lock after execution of "composer remove zendframework/zend-httphandlerrunner". Can you please give this a short review and commit this if ok?
thx, Stefan Schneider

PR #3717, opened 2021-10-24
deleted composer lib zendframework/zend-httphandlerrunner and patch

eqsoft:
see https://github.com/ILIAS-eLearning/ILIAS/pull/3716

same for release_6


mjansenDatabay:
LGTM

kergomard:
@Uwe-Kohnle : Could you please see if this is ok and if so merge soon, thanks.

PR #3718, opened 2021-10-25
	[FIX] 0031386: TinyMCE 5.x: Wrong HTTP path set to template

jeph864:
https://mantis.ilias.de/view.php?id=31386 

Due to the  HTTP path in node_modules/tinymce/plugins/latex/latex.php being incorrect, the cross-origin communication between the main window objects and the latex pop-up window was throwing errors. This PR fixes the issue.


mjansenDatabay:
Thx a lot @jeph864

PR #3719, opened 2021-10-25
PRIVACY.md: format according to guidelines and streamlining

klees:
Hi @ezenzen,

AFAIK you took the lead in working out these PRIVACY.mds on some JF some time ago. So I address you here. I assign @alex40724, since I cannot assign you directly.

So, this formats the new documentation files according to our [guideline for docs](docs/development/docs-guidelines.md). I also streamlined the header of the files and the formatting a little. I did not change anything in the content, except for the explicit mention of pull requests in the header.

You might want to have a look into the [guideline for docs](docs/development/docs-guidelines.md), I think we would want to mention the PRIVACY.md there explicitely. Currently the naming of the file isn't covered by the guideline, I think we would want to adjust the guideline there. Maybe we could also include or reference the template and content for the PRIVACY.md that you have been using.

Best regards!


alex40724:
Thanks!

PR #3720, opened 2021-10-27
Added workaround for 0031515: initIlias fails silently

smeyer-ilias:
See comments in https://mantis.ilias.de/view.php?id=31515

This is more a workaround than good fix. Please check carefully whether unsetting the DIC offset is correct here.


smeyer-ilias:
This is work in progress and produces other errors. 
To be continued...

mjansenDatabay:
@smeyer-ilias Maybe this is not necessary anymore because of the latest changes in `release_7`.

PR #3722, opened 2021-10-28
update repository pattern: ilGeoLocationDBRepository.php

utesche:
For this pattern shall act as a general guideline, it should be as 'good as possible' - not at least concerning performance.

The changes by this update:

(A) in functions: getGeoLocationById()  and  getGeoLocationsByCoordinates()
One never should (even implicit) select column values to be returned in the result which are literally provided in the where clause.

Even more: if the further processing of a query's result is NOT totally generic 
(in the meaning of: the processing does NOT expect any certain column to be present in the resultset, but determines itself what kind of columns are returned by the query),
then a "SELECT * " is inappropriate at all.
BTW, I did not notice any generic pattern in ILIAS so far, my provocative conclusion: every "SELECT * " in ILIAS' code should be revised! Matches in trunk on 2021-10-28: 1618 matches in '*.php'.

(B) in functions: ifGeoLocationExistsById()  and  ifAnyGeoLocationExistsByCoordinates()
Both functions determine the existance of rows according to some criteria (id or lat/long) and return their findings as a boolean.
So the exact count is not relevant and a "SELECT COUNT(*)" should be avoided due to performance considerations:

A "SELECT COUNT(*)" always physically (!) fetches **all** rows from the table matching the where clause:

	sql> select count(*) as count from lng_data where module = 'common';
	+-------+
	| count |
	+-------+
	|  5138 |
	+-------+
	1 row in set (0.003 sec)
	
	sql> analyze select count(*) as count from lng_data where module = 'common';
	+------+-------------+----------+------+----------------+--------+---------+-------+------+---------+----------+------------+--------------------------+
	| id   | select_type | table    | type | possible_keys  | key    | key_len | ref   | rows | r_rows  | filtered | r_filtered | Extra                    |
	+------+-------------+----------+------+----------------+--------+---------+-------+------+---------+----------+------------+--------------------------+
	|    1 | SIMPLE      | lng_data | ref  | PRIMARY,i1_idx | i1_idx | 92      | const | 9570 | 5138.00 |   100.00 |     100.00 | Using where; Using index |
	+------+-------------+----------+------+----------------+--------+---------+-------+------+---------+----------+------------+--------------------------+

In this case column 'r_rows' reports 5138 rows are fetched from the table (returned as 'count' in the resultset).

A "SELECT EXITS(SELECT 1 FROM ...)" reads **only** the first matching row and then immediately ends processing the table:

	sql> select exists(select 1 from lng_data where module = 'common') as count;
	+-------+
	| count |
	+-------+
	|     1 |
	+-------+
	1 row in set (0.001 sec)
	
	sql> analyze select exists(select 1 from lng_data where module = 'common') as count;
	+------+-------------+----------+------+----------------+--------+---------+-------+------+--------+----------+------------+--------------------------+
	| id   | select_type | table    | type | possible_keys  | key    | key_len | ref   | rows | r_rows | filtered | r_filtered | Extra                    |
	+------+-------------+----------+------+----------------+--------+---------+-------+------+--------+----------+------------+--------------------------+
	|    1 | PRIMARY     | NULL     | NULL | NULL           | NULL   | NULL    | NULL  | NULL |   NULL |     NULL |       NULL | No tables used           |
	|    2 | SUBQUERY    | lng_data | ref  | PRIMARY,i1_idx | i1_idx | 92      | const | 9570 |   1.00 |   100.00 |     100.00 | Using where; Using index |
	+------+-------------+----------+------+----------------+--------+---------+-------+------+--------+----------+------------+--------------------------+
	1 row in set (0.008 sec)

... column 'r_rows' reports just **one** row are actually fetched from table: 'EXISTS' turns out to be TRUE ==> 1 is returned as query result.

Otherwise, if **no** row matches the where clause ==> 0 is returned (and no rows are read, of course):

	sql> select exists(select 1 from lng_data where module = 'XYZ') as count;
	+-------+
	| count |
	+-------+
	|     0 |
	+-------+
	1 row in set (0.000 sec)
	
	sql> analyze select exists(select 1 from lng_data where module = 'XYZ') as count;
	+------+-------------+----------+------+----------------+---------+---------+-------+------+--------+----------+------------+--------------------------+
	| id   | select_type | table    | type | possible_keys  | key     | key_len | ref   | rows | r_rows | filtered | r_filtered | Extra                    |
	+------+-------------+----------+------+----------------+---------+---------+-------+------+--------+----------+------------+--------------------------+
	|    1 | PRIMARY     | NULL     | NULL | NULL           | NULL    | NULL    | NULL  | NULL |   NULL |     NULL |       NULL | No tables used           |
	|    2 | SUBQUERY    | lng_data | ref  | PRIMARY,i1_idx | PRIMARY | 92      | const |    1 |   0.00 |   100.00 |     100.00 | Using where; Using index |
	+------+-------------+----------+------+----------------+---------+---------+-------+------+--------+----------+------------+--------------------------+
	2 rows in set (0.000 sec)

### DIFF
--- a/docs/development/code-examples/repository-pattern/01_simple_example/ilGeoLocationDBRepository.php
+++ b/docs/development/code-examples/repository-pattern/01_simple_example/ilGeoLocationDBRepository.php
@@ -4,7 +4,7 @@ class ilGeoLocationDBRepository implements ilGeoLocationRepository {
 
 	public const TABLE_NAME = 'geo_location';
 
-	public function __construct(\ilDBInterfacee $a_db)
+	public function __construct(\ilDBInterface $a_db)
 	{
 		$this->db = $a_db;
 	}
@@ -41,8 +41,9 @@ class ilGeoLocationDBRepository implements ilGeoLocationRepository {
 	public function getGeoLocationById(int $a_id) : ilGeoLocation
 	{
 		// Set up SQL-Statement
-		$query = 'Select * FROM ' . $this->db->quoteIdentifier(self::TABLE_NAME) .
-				 ' WHERE id = ' . $this->db->quote($a_id, 'integer');
+		$query = 'SELECT title, latitude, longitude, expiration_timestamp' .
+				' FROM ' . $this->db->quoteIdentifier(self::TABLE_NAME) .
+				' WHERE id = ' . $this->db->quote($a_id, 'integer');
 
 		// Execute query
 		$result = $this->db->query($query);
@@ -52,7 +53,7 @@ class ilGeoLocationDBRepository implements ilGeoLocationRepository {
 		{
 			// Create object out of fetched data and return it
 			return new ilGeoLocation(
-				(int)$row['id'],
+				$a_id,
 				$row['title'],
 				(float)$row['latitude'],
 				(float)$row['longitude'],
@@ -66,9 +67,10 @@ class ilGeoLocationDBRepository implements ilGeoLocationRepository {
 	public function getGeoLocationsByCoordinates(float $a_latitude, float $a_longitude) : array
 	{
 		// Set up SQL-Statement
-		$query = 'Select * FROM ' . $this->db->quoteIdentifier(self::TABLE_NAME) .
-				 ' WHERE latitude = ' . $this->db->quote($a_latitude, 'float') .
-				 ' AND longitude = ' . $this->db->quote($a_longitude, 'float');
+		$query = 'SELECT title, expiration_timestamp' .
+				' FROM ' . $this->db->quoteIdentifier(self::TABLE_NAME) .
+				' WHERE latitude = ' . $this->db->quote($a_latitude, 'float') .
+				' AND longitude = ' . $this->db->quote($a_longitude, 'float');
 
 		// Execute query
 		$result = $this->db->query($query);
@@ -81,8 +83,8 @@ class ilGeoLocationDBRepository implements ilGeoLocationRepository {
 			$locations[] = new ilGeoLocation(
 				(int)$row['id'],
 				$row['title'],
-				(float)$row['latitude'],
-				(float)$row['longitude'],
+				$a_latitude,
+				$a_longitude,
 				new DateTimeImmutable($row['expiration_timestamp'])
 			);
 		}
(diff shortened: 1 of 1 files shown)


rsheer:
Hey @utesche

Thanks a lot for taking the time to improve my SQL-Statements and commenting them with best practises! There are quite a few tricks in SQL which I was not taught at school or which I had forgotten already.

I'm more than happy with this PR and will have a deeper look in the SQL-Changes for my code as well. :-)

But I have a question about the `SELECT * FROM`. Is there a big performance difference to selecting the columns? For the cases where I use the repository pattern for one specific table, I always have a `buildObjectFormRow`-Method which needs this generic `SELECT * FROM` approach to work. Following an example of what I mean:
```php
$query = "SELECT * FROM " . self::TABLE_NAME . " WHERE col_xyz = " . $this->db->quote($xyz, \ilDBConstants::T_INTEGER);
$result = $this->db->query($result);
if($row = $this->db->fetchAssoc($result) { // or use "while($row = this->db->fetchAssoc($result){ ... }" for multiple objects)
    return $this->buildObjectFromRow($row);
}
return null; // or throw Exception
```
For me, this is more error-proof during development since only the `buildObjectFormRow`-Method has to know the columns of a table and the arguments of a model-object. I know, this method doesn't exist in the given code example. But it is a case for which I think the `SELECT * FROM` could be appropriate.

utesche:
Hi Raphael,

_Is there a big performance difference to selecting the columns?_

To be a little subtle, I would like to _transform_ your question to a relevant one ...... by removing the "big" from the question ;-)  .
EVERY effort on code-side to reduce impact on DBMS processing and/or to the physical size of the resultset is worth its prize (prize in terms of some more invest in coding a SQL query).

_Is there a performance difference to selecting the columns?_
Yes. 
The most important difference makes the reduction of the resultset's size by avoiding redundant data.
When the where clause uses **explicit** values, then this values should not be fetched by DBMS nor transferred to the querying process in the resultset.  Typically these are where clauses like  "where col1 = 'value' ".  
In contrast to this are where clauses with 'like' predicates, of course.  In this cases the columns must be selected, fetched and transfered if (!) their individual values are used by further processing and not neccessary only for a proper selection.

I don't get your point of 'error-proof'.
In my notion a SELECT should **always** explicit name the desired columns.  So when you insist you want to use SELECT * _during development_ ,  this is fully contrary to my claim _a programmer should **always** be aware of what he needs to select_ .
To use * as a convenience is just a lazy _I-do-not-want-to-switch-on-brain-just-now_ attitude, in my opinion - harsh but hits the point.

Additionally: expressively nominating the desired columns is an _active act_ of inline documentation.

wbr.

rsheer:
Well, I agree in most parts with you. I'm all for improving performance and I see a lot of potential in it for ILIAS. Even though  the repository pattern is not about performance according to our [documentation](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/docs/development/repository-pattern.md#disadvantages). But if performance was its top priority, the whole model-object part could be replaced with the selection of the needed primitiv values.

The point I'm trying to make with the `SELECT *` is really just about the *repository-returns-model-object part in the Repository-Pattern*. If it is about queries where just certain columns are needed, e.g. `selectTitleForObjId(int $id) : string {...}`, I don't see a reason for the `SELECT *` either.

The use of * is in my case indeed an act about convenience and lazyness. With writing around 4 differenct select-functions (or more), which nearly look the same (build query, fetch, object construction, return object), there is an urge in either copy-pasting the code or writing generic functions to reduce some code repetitions.

By changing the table schema or the object-constructor, either in development or during a refactoring, every query and model-object constructor in every function has to be changed accordingly. I know, this is not something that takes days to rewrite. But in my opinion it is still something that bothers me (replace the same code snippets at 4 functions...) and a case where one could overlook a part in the code which leads to an error (thats what I mean by 'error-proof').

Don't get me wrong. **I'm not completely against or for** using `SELECT *` at all. In the specific case described above, this is just a trade off that I think could be made.

Thanks for going with me in this discussion. I appreciate it very much :-)

kergomard:
Thank you very much @utesche for contributing to the ILIAS documentation. I will merge this.

PR #3724, opened 2021-10-29
CI: activate captain hook and fix codestyle before commit

daniwe4:
This PR will activate the captain hooks tool.
To activate the tool, a new composer install/update run must take place.
Be aware, this will override all existing hooks.
In addition, CS fixes are executed immediately during a commit instead of just being displayed.

Thanks to @mjansenDatabay. Your video from the last DevConf about CaptainHook helpes a lot.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,8 @@
 	},
 	"scripts": {
 		"post-autoload-dump": [
-			"php setup/cli.php build-artifacts --yes"
+			"php setup/cli.php build-artifacts --yes",
+			"libs/composer/vendor/bin/captainhook install -f"
 		],
 		"post-install-cmd": [
 			"php libs/composer/rmdirs.php"
--- a/captainhook.json
+++ b/captainhook.json
@@ -36,7 +36,7 @@
         "conditions": []
       },
       {
-        "action": "libs/composer/vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix --config=./CI/PHP-CS-Fixer/code-format.php_cs --using-cache=no --dry-run -vvv {$STAGED_FILES|of-type:php}",
+        "action": "libs/composer/vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix --config=./CI/PHP-CS-Fixer/code-format.php_cs --using-cache=no -vvv {$STAGED_FILES|of-type:php}",
         "options": [],
         "conditions": [
           {


mjansenDatabay:
I appreciate these changes. In the past some ILIAS developers had objections regarding automatic code manipulation by `PHP-CS-FIXER`, so I am curious to follow the discussions.

composer.json
```diff
@@ -9,7 +9,8 @@
 	},
 	"scripts": {
 		"post-autoload-dump": [
-			"php setup/cli.php build-artifacts --yes"
+			"php setup/cli.php build-artifacts --yes",
+			"libs/composer/vendor/bin/captainhook install -f"
```
mjansenDatabay, line 13: One concern: `CaptainHook` will most probably not be installed on productive environments, so `libs/composer/vendor/bin/captainhook` might not exist. Does this command result in any error if you execute `composer install --no-dev` or `composer dump-autoload` (without dev dependencies being installed)?

PurHur: I found a good explanation about how to run this script only in dev mode here: https://stackoverflow.com/a/53022623/3655829

daniwe4: Thanks for your hints.
 
You are right @mjansenDatabay, CaptainHook isn't present on live systems. I didn't think about that. So we have to find some workaround to avoid errors on composer runs with --no-dev flag. 
After reading the link posted by @PurHur i think we have a couple of potential solutions.

**Defining a condition in the command**
`"[ $COMPOSER_DEV_MODE -eq 1 ] && libs/composer/vendor/bin/captainhook install -f || echo 0"`
But that can cause problems on windows platforms, because $COMPOSER_DEV_MODE isn't available. Is that relevant?

**Install scriptsdev for a script-dev section in composer.json**
https://github.com/neronmoon/scriptsdev
Much cleaner way but one more dependency.

I think one of them could do the job for us. How should we proceed? Did i forgot something?

klees:
Hi @mjansenDatabay,

I think you are best suited to handle this, so assigned you. Hope this is ok =)

Best regards!

composer.json
mjansenDatabay: I know there are some Windows users out there, e.g. @mbecker-databay  :-).

I am sure we will find a solution for them. I would like to discuss this change request in general on the next JourFixe. If approved, we can discuss possible solutions for Windows users.

Maybe we'll need a PHP helper script to check for `$_SERVER['COMPOSER_DEV_MODE']` and interprete the exit code of this helper script in the command

mjansenDatabay:
> I think you are best suited to handle this, so assigned you. Hope this is ok =)

Sure @klees

matthiaskunkel:
Jour Fixe, 15 NOV 2021 : We highly appreciate the intention of the PR and accept it for trunk. Before merging the PR we need to clarify how Windows user could use Captain Hook without problems. Please notify us if you observce double runs with Captain Hook so we could investigate what are the reasons for this and hopefully improve Captain's behaviour.

mjansenDatabay:
We'll provide a new PR today.

mjansenDatabay:
New PR: https://github.com/ILIAS-eLearning/ILIAS/pull/3839

PR #3725, opened 2021-10-29
UI, Mainbar: 30644 render menuitems as list, focus contents

nhaagen:
https://mantis.ilias.de/view.php?id=30644
This renders mainbar's toolentries as ul/li lists and sets focus to their slates contents when triggered.
Screenreaders will now mention the number of tools, and tool-contents can be accessed by tabbing.
I'd be glad, though, if somebody would doublecheck on the anticipated a11y-improvement.

### DIFF
--- a/src/UI/examples/Layout/Page/Standard/ui.php
+++ b/src/UI/examples/Layout/Page/Standard/ui.php
@@ -399,7 +399,17 @@ function getDemoEntryTools($f)
     $slate = $f->maincontrols()->slate()->legacy(
         'Help',
         $symbol,
-        $f->legacy('<h2>tool 1</h2><p>Some Text for Tool 1 entry</p>')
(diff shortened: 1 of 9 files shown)


Amstutz:
Hi @nhaagen 

Thx a lot for the a11y improvement! I am not 100% if this resolves the issue completely, but it very much improves the situation. My main concern is, that it could be, that we would need to label the Button itself again also as "menuitem". However, I belive this step here to be big enough to be merged, and thrown into the next audit. 

Thx

Amstutz:
Forget my comment on the "menuitem" above, I checked this again and it looks fine. From my limited expertise, I would say, this looks good. However, we will see what future audits tell us. Thx a lot for tackling this. I just picked this down to 7 as well.

PR #3726, opened 2021-11-03
UI: Mainbar/Drilldown - use DD for Mainmenu Administration

nhaagen:
This renders the mainbar's administration menu as drilldown.

- UI/Menu Drilldown with multiple instances (was buggy before, sorry)
- make instances and engageLevel available externally
- use CookieStorage for DD-persistence
- TopParentItemDrilldownRenderer
- introduces [DrilldownSlate](https://github.com/ILIAS-eLearning/ILIAS/pull/3726/files#diff-f8a1e6c4212102c1cae91939d22ba208e085092c371d020a8d1af3878e265a62R131-R166)

dd1
dd2

### DIFF
--- a/Services/MainMenu/classes/Provider/StandardTopItemsProvider.php
+++ b/Services/MainMenu/classes/Provider/StandardTopItemsProvider.php
@@ -6,6 +6,8 @@ use ILIAS\GlobalScreen\Identification\IdentificationInterface;
 use ILIAS\GlobalScreen\Scope\MainMenu\Provider\AbstractStaticMainMenuProvider;
 use ILIAS\MyStaff\ilMyStaffAccess;
 use ILIAS\UI\Component\Symbol\Icon\Standard;
+use ILIAS\GlobalScreen\Scope\MainMenu\Collector\Renderer\TopParentItemDrilldownRenderer;
+use ILIAS\GlobalScreen\Scope\MainMenu\Collector\Information\TypeInformation;
 
 /**
  * Class StandardTopItemsProvider
@@ -158,7 +160,8 @@ class StandardTopItemsProvider extends AbstractStaticMainMenuProvider
             ->withAvailableCallable(
                 static function () : bool {
                     return (bool) ilMyStaffAccess::getInstance()->hasCurrentUserAccessToMyStaff();
-                });
+                }
+            );
 
         $title = $f("mm_administration");
         $icon = $this->dic->ui()->factory()->symbol()->icon()->standard("adm", $title)->withIsOutlined(true);
@@ -171,6 +174,12 @@ class StandardTopItemsProvider extends AbstractStaticMainMenuProvider
             ->withPosition(70)
             ->withVisibilityCallable($this->basic_access_helper->hasAdministrationAccess());
 
+        $dd_renderer = new TopParentItemDrilldownRenderer();
+        $ti = new TypeInformation(get_class($administration), get_class($administration));
+        $ti->setRenderer($dd_renderer);
+        $administration = $administration->setTypeInformation($ti);
+
+
         return [
             $dashboard,
             $repository,
@@ -178,7 +187,7 @@ class StandardTopItemsProvider extends AbstractStaticMainMenuProvider
             $achievements,
             $communication,
             $organisation,
-            $administration,
+            $administration
         ];
     }
 
--- a/src/GlobalScreen/Scope/MainMenu/Collector/Renderer/BaseTypeRenderer.php
+++ b/src/GlobalScreen/Scope/MainMenu/Collector/Renderer/BaseTypeRenderer.php
@@ -31,6 +31,12 @@ class BaseTypeRenderer implements TypeRenderer
     protected $ui_factory;
 
     /**
+     * @var Renderer
+     */
+    protected $ui_renderer;
+
+
+    /**
      * BaseTypeRenderer constructor.
      */
     public function __construct()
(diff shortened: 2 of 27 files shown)


Services/MainMenu/classes/Provider/StandardTopItemsProvider.php
```diff
@@ -171,14 +174,20 @@ static function () : bool {
             ->withPosition(70)
             ->withVisibilityCallable($this->basic_access_helper->hasAdministrationAccess());
 
+        $dd_renderer = new TopParentItemDrilldownRenderer();
```
chfsx, line 177: This is because the administration menu should be the only-one which will be rendered as a Drilldown, right? I'm currently not sure if we should go this way OR whether we should introduce a dedicated `Item` in the Globalscreen-Scope `MainMenu`. Currently one type of `Item` results in one type of `UI\Component`. This is a predictable behaviour, every `Provider` which provides some type if `Items` will result in the same rendering of this type. One can replace the "renderer" – a.k.a "GS-item to UI-Component translator" – as a whole and all `Items` of the same type will be translated like this.

Both ways have their advantages and disadvantages and I'm currently not sure which way to go ;-)

src/GlobalScreen/Scope/MainMenu/Collector/Renderer/BaseTypeRenderer.php
```diff
@@ -30,13 +30,20 @@ class BaseTypeRenderer implements TypeRenderer
...
+
+
     /**
      * BaseTypeRenderer constructor.
      */
     public function __construct()
     {
         global $DIC;
         $this->ui_factory = $DIC->ui()->factory();
+        $this->ui_renderer = $DIC->ui()->renderer();
```
chfsx: where is this needed?

chfsx (commented):
Thanks @nhaagen for this PR! see my comments in code, as I mentioned I'm currently not sure which way to go...

src/GlobalScreen/Scope/MainMenu/Collector/Renderer/BaseTypeRenderer.php
nhaagen: I currently need a legacy slate to inject a DD in the menu: https://github.com/ILIAS-eLearning/ILIAS/pull/3726/files#diff-7deee1ec3fe029b2adca88f0f292dea0f8599c3715a6f394cc4da916b3dd1665R27

Hoever, with a (yet to build) DD-Slate, this surely could and should go away again.

Services/MainMenu/classes/Provider/StandardTopItemsProvider.php
nhaagen: I'm not sure if adm will remain the only entry, but it certainly is a promising candidate :)
What I thought was, that with the DD for Administration, I'm actually not changing the nature of adm as topParentItem, but only display in a different way. 
The item-solution would be additional, right ? i.e. we would move the renderer from provider to collector?

klees:
Hi @nhaagen,

LGTM mostly from my perspective. I did only have a cursory look into the `GlobalScreen` part. Could you get in touch with @chfsx to work out details in that part? I had a closer look into the UI framework stuff.

I have the following suggestions there:

* [x] withPersistanceId: Could we keep that one internal and use the position in the main bar structure to set it? I'm most certain that we will experience naming clashes if we make this public to our consumers and we also did not announce it via the public interface also.
* [x] Debugging: Please remove debugging statements in `src/UI/templates/js/MainControls/dist/mainbar.js`.
* [x] Test: Please add tests for the new js-modules.

Since this is created by my team, I ask @Amstutz to have a look as well.

Thanks and best regards!

klees (changes requested):
:point_up:

Amstutz:
Hi @nhaagen,

I did not check up on the review of @klees before review, but my main points have turned out to be similar. Also, I can not contribute much to the Global Screen part.

Please do:

- [x] Remove the console.log statements
- [x] Resolve conflicts
- [x] Add the Unit Tests for the new Slate Type. There seems not much in terms of rendering, but check nevertheless, make sure they are created by factory as expected and render in case of some future change.

Consider doing:
- [x] Expand on the JS Unit tests. Note that they might be mandatory in the near future.

Since there is no major issue with the Interface appart from the question around withPersistanceId, ill attach the JF label.

Thanks and best regards!

matthiaskunkel:
Jour Fixe, 15 NOV 2021 : We highly appreciate the suggested UI element Drilldown Slate and accept it for trunk.

nhaagen:
Hi @chfsx ,
I added a strong warning and JF reservation for changing the renderer for a singular/specific item: 
https://github.com/ILIAS-eLearning/ILIAS/pull/3726/files#diff-eefb28c171af42d3199d2dd2114a2bf4b24eaf0a0bf28376817d5cc8e83f5bb6R114

Will that meet up with concerns regading uniformity and item->ui mapping in the GS?
Thanks for caring

chfsx:
@nhaagen this is perfectly fine, thanks a lot!

chfsx (approved):
This is fine for the GS part, thank you very much for your work @nhaagen !

nhaagen:
Hi @Amstutz ,
I rebased and added some test. Would you mind having another look, please?
Thank you

klees (approved):

Amstutz:
Hi @nhaagen 

I just made a final test of your change on the fronten. I really like it. Thx a lot for the contribution. From what I can see, this also works well by just using the keyboard.

Ill merge. 

There is just one thing that catched my eyes, but it has nothing todo with the PR, Org Units are really strangely placed in the admin. It never catched my eye as much as it does with the new Drilldown adaption. IMO, this does make a good point for the new layout:

Amstutz:
image

oliversamoila:
That's cool. Thank you!
I’m on fire to try it out live.

PR #3727, opened 2021-11-03
Mantis 31346: Add missing $

PurHur:
This PR fixes:

https://mantis.ilias.de/view.php?id=31346

but actually this variable is not needed in the first place, but i didnt want to make huge changes.

### DIFF
--- a/Services/Language/classes/class.ilObjLanguageExtGUI.php
+++ b/Services/Language/classes/class.ilObjLanguageExtGUI.php
@@ -408,7 +408,7 @@ class ilObjLanguageExtGUI extends ilObjectGUI
 	    $changesSuccessBool = 1;
 
         // view the list
-        $this->viewObject(changesSuccessBool);
+        $this->viewObject($changesSuccessBool);
     }
 
 


PurHur:
/assign @mjansenDatabay 

Could we merge this before we setup new instances so we have to update them tomorrow?

PurHur:
Thanks for reacting but this simple error is blocking a lot of stuff. 

Somebody with the merge rights should simple press on merge. I dont want to wait another 4 months for merging my PR here.  This is a critical error for all my projects and it would be resolved by a simple click.

mjansenDatabay:
> Thanks for reacting but this simple error is blocking a lot of stuff.
> 
> Somebody with the merge rights should simple press on merge. I dont want to wait another 4 months for merging my PR here. This is a critical error for all my projects and it would be resolved by a simple click.

As far as I understand this only affects the advanced language maintenance in ILIAS, not every request. In this case, the responsible maintainer should review the PR and merge it accordingly. Of course here the current code is obviously wrong and the fix is trivial.
Because ILIAS 7.4 is already released, I don't see the need for an intervention of me (personally) or the @ILIAS-eLearning/technical-board .

PurHur:
Thanks! This has also be mergend into trunk.

PR #3728, opened 2021-11-03
Invalid attribute in table of exercise

kergomard:
Fixes: https://mantis.ilias.de/view.php?id=31540


alex40724:
This is fixed in all supported branches. Thanks!

PR #3730, opened 2021-11-03
Fix aria labels in action menu

kergomard:
Fixes: https://mantis.ilias.de/view.php?id=31541


Amstutz:
Thx a lot! Ill pick to trunk as well.

PR #3736, opened 2021-11-04
Refinery Randomization - Functional Approach

lscharmer:
On behalf of the @ILIAS-eLearning/technical-board  , this PR adds a new group `random()` to the ILIAS `Refinery`. This should replace the strategies of `Services/Randomization` to shuffle or keep the order of array elements (primarily consumed in `Services/TestQuestionPool`).

We implemented this in consideration of:

- This implementation uses the builtin `shuffle(...)` function instead of the `ilArrayElementShuffler::mtShuffle` because since PHP 7.1 `shuffle(...)` already uses the Mersenne Twister algorithm (_see https://www.php.net/manual/en/function.shuffle.php_).
- The existence check for `mt_srand(...)` and `mt_rand(...)` is also removed, because `mt_srand(...)` replaced `srand(...)` in PHP 7.1 (_see https://www.php.net/manual/en/function.mt-srand.php_).
- The added `ILIAS\Refinery\Random\Transformation\ShuffleTransformation::class` does not shuffle an array by itself but returns a `ILIAS\Refinery\Effect\Effect::class` which will then return the shuffled array. The array can not be shuffled in a Transformation itself because the `Transformation`'s MUST be side effect free and therefore MUST NOT call `mt_srand(...)`. The usage of an Effect class as a return value ensures that no side effect is done in the `ShuffleTransformation::class`.

We did not move `ilRandom` to the `Refinery`, because we do not consider it being a `Transformation`. `ilRandom` is widely used within the ILIAS code base, so we only see one option: Leave it where it is.


klees:
I just discussed this with @mjansenDatabay. I think the solution is theoretically sound. Nice seeing another functionalist in the community @lscharmer =)

I still would not support to merge this, basically for two reasons:
* Introducing a monad-like mechanism just to implement an array shuffler seems to be a lot when we have a very simple `shuffle`-function on the other side of the spectrum of possibilities. I cannot really see what justifies this complexity...
* The new code, interfaces and mechanisms will be hard to understand for many, hence be a constant source of confusion, bugs and questions. Much stuff to be understood, just to use an array shuffler.

This isn't a hard "no", from my side, though, just a "rather not". Thanks for working this out anyway, always interesting to see functional concepts implemented in PHP.

klees:
Since no objections for #3765 came up, I assume that we go on with the approach there. Hence, this can be closed.

PR #3737, opened 2021-11-04
This changes the role-attribute in BulkyLinks for Menu-Items

kergomard:
Fix for: https://mantis.ilias.de/view.php?id=31552


klees (approved):
Thx, LGTM

Amstutz:
Thx! Special kudo, for the test. Maybe we should move in future in general for the "withAriaRole" to just "withRole". Using aria is not the preferred solution, if role is also an option. However, IMO this is beyond the scope of this.

PR #3743, opened 2021-11-09
[FIX] 0031576: add missing st in Selbstständig

floen-de:


matthiaskunkel:
Thanks for the fix. I merge it to 5.4 and adapt it in the other branches.

PR #3748, opened 2021-11-10
Proposal/8/headers in files

chfsx:
We looked for a mechanism to add a standard-license-header to every file of our code-base (and a mechanism to update those comments if needed). With PHPRector (already in trunk) we developed a Rector which can do that with some restrictions:
- only existing comments above or beneath reuires/use statements and/or above a class definition are removed with string containing (copyright|Copyright|GPL-3\.0|GPLv3|LICENSE)
- if there are directories which must be excluded, the list in (lib|vendor|CI|data|Customizing) must be adopted
- if a file has two class-definitions in it, two license headers will be added (such files should not exists but there are some...)
- there can be false positives, use with care

with the current rector.php in CI/Rector you can run it for a directory with:

./libs/composer/vendor/bin/rector process --config ./CI/Rector/rector.php Modules/LTIConsumer

### DIFF
--- a/CI/Rector/ChangeLicenseHeader.php
+++ b/CI/Rector/ChangeLicenseHeader.php
@@ -1,0 +1,96 @@
+<?php
+
+namespace ILIAS\CI\Rector;
+
+use PhpParser\Node;
+use Rector\Core\Rector\AbstractRector;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+use PhpParser\Comment;
+
+final class ChangeLicenseHeader extends AbstractRector
+{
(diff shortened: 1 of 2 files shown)


CI/Rector/ChangeLicenseHeader.php
```diff
@@ -0,0 +1,96 @@
...
+            }
+            $default[] = $comment;
+        }
+        return $default;
+    }
+
+    public function getRuleDefinition() : RuleDefinition
+    {
+        return new RuleDefinition(
+            'Remove requires and includes', [
```
chfsx: The docu is currently wrong, copy/paste...

matthiaskunkel:
Jour Fixe, 13 DEC 2021 : We highly appreciate this suggestion and accept the PR for trunk. Please let us know if this tool does not work for your component

PR #3750, opened 2021-11-10
Improve performance of ilAwarenessUserProviderMemberships

corro:
See https://mantis.ilias.de/view.php?id=31584


alex40724:
Thanks for the investigation! I merged it manually, due to conflicts and downpatched 6+7.

PR #3751, opened 2021-11-11
JS, MainmenuSearch: #31567, add JS separately

nhaagen:
Moved the JS parts from template to separate file; fixes https://mantis.ilias.de/view.php?id=31567


Amstutz:
Sooo much love for that one.

Services/Search/classes/class.ilMainMenuSearchGUI.php
```diff
@@ -54,13 +54,12 @@ protected function initRefIdFromQuery() : void
...
     }
 
-
-
     public function getHTML() : string
     {
         iljQueryUtil::initjQuery();
         iljQueryUtil::initjQueryUI();
         $this->tpl = new ilTemplate('tpl.main_menu_search.html', true, true, 'Services/Search');
+        $tpl->addJavascript('Services/Search/js/SearchMainMenu.js');
```
chfsx: isn't `$tpl` undefined here?

nhaagen: you're right

Amstutz:
@smeyer-ilias : could you look into this, the error shows up constantly in the js console and seems very annoying. Thx.

alex40724:
@smeyer-ilias I merged this one, however the background request fails with a ilCtrlException: ilCtrl::callBaseClass was not given a baseclass and the request doesn&#039;t include one either. in file /usr/local/var/www/ilias/Services/UICore/classes/class.ilCtrl.php on line 139

I will follow up in the Mantis report.

PR #3752, opened 2021-11-11
PHP8: some fixes

nhaagen:
Dear maintainers, 
please check and OK/comment on those changes.
These are patches on whoopses I ran into when creating new objects in the repository,
and I have to cherry-pick them every time I want to use the GUI :)
Thank you.

### DIFF
--- a/Modules/Group/classes/class.ilGroupParticipantsTableGUI.php
+++ b/Modules/Group/classes/class.ilGroupParticipantsTableGUI.php
@@ -348,7 +348,7 @@ class ilGroupParticipantsTableGUI extends ilParticipantTableGUI
                     continue;
                 }
             }
-            if ($this->current_filter['org_units']) {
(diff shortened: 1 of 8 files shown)


klees (approved):

alex40724 (approved):
Thanks for the fixes!

Personally I would prefer
`return $this->selected_column[$col] ?? false;`
instead of
`return array_key_exists($col, $this->selected_column) && $this->selected_column[$col];`

But it works boths ways.

klees:
@alex40724 @nhaagen I totally agree. ``??`` is way easier to read IMO.

Amstutz:
@klees : I opted for the contrary some month ago. However, I changed my mind / have been convinced otherwise and it seems to be common practice by now in ILIAS.

nhaagen:
Wow, you are fast  - thank you. 
Added ?? and squashed the previous two commits.

smeyer-ilias (approved):
Thanks for fixing this.

PR #3754, opened 2021-11-11
UI: Bar Charts

tfamula:
Hello everybody!
This PR adds a new UI Component to the Kitchensink: The Bar Chart. It's available as Vertical Bar Chart on the one hand, and as Horizontal Bar Chart on the other hand.
The interface is ready for review. As soon as the implementation part is also ready, I will push it and let you know.

For your information (because it is not identifiable in the interface): For creating the charts, I was looking for a chart library which fits my (and hopefully our all) needs best. I decided upon [Chart.js](https://www.chartjs.org/).

Examples:

Vertical
Horizontal

Greetings and thanks!
Thomas


klees:
Hi @tfamula,

we're on this. This is a lot to digest though, so please bear with us.

Best regards!

tfamula:
Hi @klees,
thanks for the information. I also have one: Tomorrow I will push the Implementation part of the component. So maybe it will help you to understand some things of the Interface.

Greetings!

klees:
Hello @tfamula,

Thanks a lot for improving the charts in the UI-Framework. That section could surely use some more content...

We are aware that drawing charts is not a simple problem, hence thanks a lot that you accepted that challenge.

We are afraid that presenting a formal review with actionable points will lead to the false impression that there simply are some boxes to check until we deem this JF ready. Instead we ask you to join us in a meeting and discuss these two problems before we move on with this PR.
 
For one, it feels as if there are some concepts related to measurement (like `datum`, `dataset`, `point`, `dimension`) hidden in here, that want to be expressed. E.g.: We will be adding labels [in the constructor](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-209872a47fa5a295dee922a3a86854f0e7ec3bf25c8ef5c1761e2da6f6b79f3bR82), that will somehow relate to the [size of `$bars` in `withData`](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-7a7443f52ed0016b0e5eceb8653922b7d20d57716cd2af6cebfce1887701eed4R84), right? How? How does the [`$y_axis_id`](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-7a7443f52ed0016b0e5eceb8653922b7d20d57716cd2af6cebfce1887701eed4R88) relate to the labels? What is the [relative width of this dataset](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-7a7443f52ed0016b0e5eceb8653922b7d20d57716cd2af6cebfce1887701eed4R86)? [`withData`](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-7a7443f52ed0016b0e5eceb8653922b7d20d57716cd2af6cebfce1887701eed4R61) feels as if it mixes drawing directions with data, which also feels odd. We strongly suggest to start by defining some of these measurement concepts (maybe in `src/Data`) and then incorporate these in the Bar Chart. This most probably will clearify and simplify the interface of the UI component a lot. This is the central thing that we want to discuss. We already have some ideas how this could be expressed.

Second, [the Strategy 2019 of the Technical Board](https://docu.ilias.de/goto_docu_file_8005_download.html) states the goal that "ILIAS ships with a web interface that provides a satisfying user experience on screens ranging from a mobile phone to a high end display". Currently there are various methods in this proposal that fix sizes and remove responsiveness. Allowing developers to explicitly make something unresponsive does not seem to align with the strategic goal. We need to find a way how a possibly huge chart could work on a small screen. Maybe we could open the chart in an extra screen, so user could pinch? Maybe we want to display less information to keep the size of the chart manageable? This is why we want decisions on sizes and responsiveness to be removed from the developer that uses the component and make them in the renderings of the components instead.

This may sound like a lot of work and a long way to go, but we believe that we will find ways to make this component and the PR simpler that are easy to implement as well. We expect that you also might be able to keep a lot of the internal code.

Please get in touch via mail so we look for an appointment.

Best regards!
@Amstutz & @klees

tfamula:
Hi @klees,
thank you for the feedback already. Just one thing regarding the section with the measurement concepts: The links you made in your comment are all the same, I always end up [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-66d09e944a623930045dc1f82c0d02741693260722a8115c2a234789d36f5573R23). I guess this is not intended? Could you please fix the links? I think this will make this section more understandable.
I will get in touch with you by mail soon!

Greetings!

Amstutz:
Hi @tfamula 

Last Tuesday Dec. 7th 2021, we had a really good discussion about possible ways to type the dataset and dimensions somewhat better. Do you have any update on how you would like to proceed with this PR?

Please let us know, if we can assist any further with this. We believe this to be a very important contribution.

Thx

tfamula:
Hi @Amstutz 

I will tackle the new concept in the new year in January and will let you know when I have a result!
Thanks for your offered help, I really appreciate it.

tfamula:
Hi @Amstutz and @klees,
I refactored the Bar Charts by adding the new datatypes for it. As you predicted, the interface on the KS side is much clearer now and it is easier too use. I hope this implementation meets your expectations. Feel free to contact me via mail or skype if there are any questions.

Greetings
Thomas

klees (changes requested):
Hi @tfamula!

Thanks a lot for looking into this again. The new implementations look very nice already. Let's apply some polish to make it even better!

The interface looks good now and is, besides some minor issues outlined below, ready to be taken to the JF.

- [x] Dimension::getType: We're a little puzzled by everything that has to do with `Dimension::TYPE`. For one, we wonder why we even need to add a property like this to the class. Wouldn't the original class name work in the same way? Is it allowed to have another class implementing `Dimension` that has the same `type` as an existing class? Why? `Dimension::getAllTypes`, in turn, is only used [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-a14a4e68896543d4d2afbfc302bd74b9a21e06ee3efdecc539a05ea6bbff6b39R289). We had a hard time understanding how this works, but as far as we understand now, this make the axis, that are defined [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-a14a4e68896543d4d2afbfc302bd74b9a21e06ee3efdecc539a05ea6bbff6b39R362) disappear. This location, in turn, won't work if we really want `yAxisID` to be unique. There could very well be two dimensions with the same type in the dataset, so the id won't be unique. This feels as we would need to find another way to hide the axis and could then remove everything that has to do with `Dimension::TYPE`.
- [x] Id: We know, we discussed this already, but please tell us again: Why do we need a user provided `$id` for the charts? If this just exist to bind js to the charts: Please use the id provided by `withAdditionalOnLoadCode`. For user provided ids, we simply cannot guarantee that they are unique, which is an important property of any id.

- [x] Checks: You check for [`instanceof Dimension` in the constructor of `Dataset`](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-3fa671d4fdee6442b9e87c3e020b17a7395e4259f7e4af137ad0fffbe08a90dfR44) but expect [certain `Dimension`s in a method](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-3fa671d4fdee6442b9e87c3e020b17a7395e4259f7e4af137ad0fffbe08a90dfR77). This is odd, because either the set of `Dimension`s is open for extension or it is not. If it is open, the approach in the constructor is the way to go, but the [checks in the methods](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-3fa671d4fdee6442b9e87c3e020b17a7395e4259f7e4af137ad0fffbe08a90dfR77) need to be implement with a polymorphism approach (i.e. adding a `checkValue`-method to `Dimension`). If the set of dimension is closed, we already need to check them in the constructor, and [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-3fa671d4fdee6442b9e87c3e020b17a7395e4259f7e4af137ad0fffbe08a90dfR104) is a `\LogicException` at most.
- [x] checkValue: [These checks here](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-3fa671d4fdee6442b9e87c3e020b17a7395e4259f7e4af137ad0fffbe08a90dfR76) should be moved to a dedicated method `Dimension::checkValue`. This is true even if we consider the set of available dimension to be closed. For one, it will help with overview and maintenance. Second, writing unit tests will become a lot simpler. Suggestions like these become more obvious if you use TDD to develop new stuff like `Dataset`.
- [x] Open/Closed `Dimension`s: This came up in the previous two items. We would suggest to build everyhing in `src/Data` in way that allows the set of dimensions to be extended. You could then check for dimensions that are supported in the `Bar Chart` code in `src/UI`, most probably in the constructor of `Chart\Bar` already.

- [x] Naming 1: [`withPoints`](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-3fa671d4fdee6442b9e87c3e020b17a7395e4259f7e4af137ad0fffbe08a90dfR63) only adds one point (right?). So `withPoint` would be better.
- [x] Naming 2: The [`Tooltips`](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-3fa671d4fdee6442b9e87c3e020b17a7395e4259f7e4af137ad0fffbe08a90dfR162) should get another name on the layer they are currently located on. When talking about datasets, there is no such thing as a "tooltip". A tooltip is something that only becomes relevant at the GUI, as a specific way to present some information. This also becomes obvious in the accessibility-feature. The "tooltips" are rendered as simple text there. So, for the `Dataset`, calling them `Hint` or `Detail` (or maybe something else) would be a better fit, depending on what these thingies should really represent.
- [x] Outdated 1: [This example](src/UI/examples/Chart/Bar/Horizontal/base.php) will not work anymore, right?
- [x] Outdated 2: [This example](src/UI/examples/Chart/Bar/Vertical/base.php) will not work anymore either, correct?
- [ ] Tests: Tests for PHP-code are missing completely. You might want to consider using TDD for new stuff like `Dataset` in the future, as this helps to understand the intended behaviour and model edge cases. These tend to be forgotten when tests are added later on. Also, tests help when reviewing because they clarify the intended behaviour a lot. The JS-tests should be extended to cover some computation heavy functions like `determineToolTipYLabels`. 
- [x] Conversion: Instead of [iterating over arrays to convert them to `\StdClass`-objects](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-a14a4e68896543d4d2afbfc302bd74b9a21e06ee3efdecc539a05ea6bbff6b39R325) just [cast the array to `object`](https://stackoverflow.com/questions/19272011/how-to-convert-an-array-into-an-object-using-stdclass). This will make the Renderer more readable.
- [x] Location: Move that [`Bar` class](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-507615051c1eb686dabeb417a980971bafb7164169a315f7d97fc812b054bbaeR27) to the UI-framework part of your implementation. It seems to be just some config for the bar charts. Consider naming it to something like `BarConfig`, because it does not represent "the bar", but rather some configuration to repeatably create similar bars (right?).
- [x] Comment: Please add a comment for [Bar::withSize](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-507615051c1eb686dabeb417a980971bafb7164169a315f7d97fc812b054bbaeR48) and/or give it a more descriptive, probably longer name, and/or change the parameter name. We know we have discussed this, but we are not able to recall what this will be doing, what "size" this is or how one could determine a proper value for the parameter. This will be a lot worse for others =)
- [x] Typing 1: Use [`Color` for internal storage and return types in `Bar`](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-507615051c1eb686dabeb417a980971bafb7164169a315f7d97fc812b054bbaeR43) and turn it into a string as late as possible. If one sees a `Color`, one can always turn it into a string, but if one sees a string it could just contain anything, not only a color.
- [x] Typing 2: Use [`float`](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-3fa671d4fdee6442b9e87c3e020b17a7395e4259f7e4af137ad0fffbe08a90dfR190) as return value here and throw a `\RuntimeException` instead [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-3fa671d4fdee6442b9e87c3e020b17a7395e4259f7e4af137ad0fffbe08a90dfR193). Same for `getMaxValue`. You never check for the `null`-case when using these methods anyway, nor would anyone expect these methods to be used when there is nothing in the Dataset. You might want to add a simple `Dataset::isEmpty` and use it when constructing the `Chart` to prevent that case altogether when inside the UI framework code. Make invalid state irrepresentable.
- [x] DRY: Turn [this check](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-3fa671d4fdee6442b9e87c3e020b17a7395e4259f7e4af137ad0fffbe08a90dfR65) into a method. It is repeated [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-3fa671d4fdee6442b9e87c3e020b17a7395e4259f7e4af137ad0fffbe08a90dfR140). This also would have been more obvious when using TDD.
- [x] Primitive Value: Please do not [use arrays for structured data](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-f027d7d3247f2a8f57ef74ccdd9c5482b058dfb0618aece962b5156f8c478a2aR26). Just introduce some class to hold that data, even if it is just some smallish value class located directly next to the class(es) that use them. This allows to add typehints, prevents typos and, in general, helps to maintain code. That way you might get rid of some of [these constants](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-66d09e944a623930045dc1f82c0d02741693260722a8115c2a234789d36f5573R31) as well. Also, it looks as if the specific structure more belongs to the renderer and the specific library used for rendering the charts. So maybe it should be moved into the rendering anyway.
- [x] Dimensions: Please have a look into how dimension are structure. "Ordinal" does not seem to be what one would [expect from the name](https://en.wikipedia.org/wiki/Ordinal_data). Since a `Dataset` and its `Dimension`s describe measurement, we should stick to [according nomenclature](https://en.wikipedia.org/wiki/Level_of_measurement). We don't know which of these you really need for your usecase currently, so we don't ask to implement each of these, but at least nomenclature should fit.
- [x] `getMinValue`: We understand why you implemented [`getMinValue`](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-3fa671d4fdee6442b9e87c3e020b17a7395e4259f7e4af137ad0fffbe08a90dfR190) to cover all dimensions of a `Dataset` (to render axis correctly). But for the `Dataset` it does not really make sense to compare values from different dimensions. You would be comparing apples to pears. Please implement `getMinValue` and `getMaxValue` to take the name of a dimension and look only among values in the defined dimension to find a minimum for that dimension. The comparison over mins and maxs from different dimensions should then be moved to the point where it starts to make sense: The rendering code in the UI framework (where we turn measurments to pixels). Depending on the meaning of `Bar::withSize` (see above) you might need to introduce that scale factor there into the calculation as well.

We really appreciate the work you are doing here and totally recognize that this is neither easy nor simple, and, most probably, far out of what one would expect to be the comfort zone of a typical software developer. So if there is anything we can help you with, feel free to get back to us again on short notice.

Best regards!

tfamula:
Hi @klees, Hi @Amstutz!
Thank you very much for the appreciation and the (as always) constructive feedback!


> Dimension::getType

- Good point. I got rid of the vague property and check for the class name now. But in case of the `yAxisID`, I don't want it to be unique. I want to find out how much Dimensions are used in the chart to know how many pseudo-axes I need. This will make sure that bars of the same Dimension are rendered one after another, but bars of different Dimensions are rendered overlapping (see RangeDimension "Target" in example). The question is if this overlapping is desired in every use case. If not, we maybe need an additional mutator, something like `->withOverlappingDimensionBars(false)` (I hope you find a better name for it).

> Id

- As we discussed in our sykpe call, I made some adjustments for the template and use only one unique id now. The user provided id is not needed anymore.

> Checks
> checkValue
> Open/Closed `Dimension`s

- I decided for open Dimensions and the polymorphism approach. Thanks for the detailed elaboration.

> Naming 1

- I disagree. You can define multiple Dimensions for a Dataset and the method adds a point for each. So in fact, we still add multiple points with this method like [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3754/files#diff-4db0aa9096ff6441d24e34986a8325ab84525ee89430fa107edb3dc1caf5ef32R30) and plural still fits, right?

> Naming 2

- Thanks for the very good clarification. I changed it to `AlternativeInformation`. It fits most IMO.

> Outdated 1

- Corrected, thanks!

> Outdated 2

- Corrected, thanks!

> Tests

- Work in progress!

> Conversion

- Done. Thanks for the tip!

> Location

- Done. Hope it comes up to your expectations now.

> Comment

- Done. I hope it's clearer now.

> Typing 1

- Totally agree! Done.

> Typing 2

- Done. Thanks for the explanation.

> DRY

- Done.

> Primitive Value

- Everything done. I like the improvement!

> Dimensions

- As we discussed in our sykpe call, I changed `OrdinalDimension` to `CardinalDimension` and also changed documentation accordingly.

> `getMinValue`

- Gotcha. Done. `Bar::withSize` is not relevant here in my eyes.



Greetings
Thomas

matthiaskunkel:
Jour Fixe, 07 MAR 2022 : We highly appreciate this suggestion and accept this PR for trunk.

tfamula:
I created unit tests in `src/UI` and `src/Data`. The JS unit tests are still pending.

There is still one unit test failing, but I do not understand why. It says "_Non abstract Component Vertical does not provide any example. Please provide at least one in Vertical_". But there are even two examples, like in Horizontal.

klees:
@tfamula Cannot look currently, but maybe something wrong with the location or the spelling?

tfamula:
@klees I checked multiple times. Guess I'm overlooking something...

klees:
@tfamula For some reason this seems to be in some draft status, checks are not running here. Can you remove that status? I can't find the according switch...

klees:
Hi @tfamula,

two insights: For one, the test for the README.md in `Data` fails. We are checking code in the README to make sure that the code works. The tests for the examples did not work because the crawler currently seems to use the complete string provided in `@return` for the factory and does not complete the namespace. It would have failed for both components, but due to the way the test is written, it just fails for `Vertical` first and then never gets to look into `Horizontal`.

If you pick [this commit](https://github.com/klees/ILIAS/commit/cc5ea4bab52f9ddc082e520dffe1258e12d322de) the tests will work. Still: Do you understand why they don't run here?

Best regards!

tfamula:
Hi @klees 
sorry for the late reply, I was OOO the last two weeks.

Thanks for clarifying the two problems with the tests and also providing a fix. I understand what was going on there!

Regaring the draft status: For me, everything looks fine in Github. No draft status and all checks run and also pass!?

klees:
@tfamula Maybe some hickup related to the series of outages on GitHub lately, donno. It's fine now...

klees:
Hi @tfamula (and everybody else, of course)!

It would be nice if you'd accustom yourself to `rebase` stuff onto trunk (and then force push it) instead of merging trunk for PRs. This makes it a lot easier to pull changes into the trunk later on and gives a cleaner commit history. Thanks.

I will do it for these changes here myself, now.

Best regards!

klees:
Hi @tfamula,

once again: Thanks for this amazing work. I'm really looking forward to build on and with this.

I have pushed the changes to trunk as https://github.com/ILIAS-eLearning/ILIAS/commit/1fe1275752a4cd87d7956cee9a7c08ea7a882208

I made some rather minor improvements in the process:

* fixed a spelling mistake
* renamed `DataSet::withPoints` to `DataSet::withPoint` (maybe we can discuss this sometime in person)
* check values against the underlying `CardinalDimension` in `RangeDimension`
* moved `BarConfig`, `Axis`, `XAxis` and `YAxis` from the internal `Implementation` namespace to the public namespace

I'm looking forward to the PR with JS unit tests.

Have a nice weekend!

tfamula:
Hi @klees 

> - moved BarConfig, Axis, XAxis and YAxis from the internal Implementation namespace to the public namespace

I am little bit confused by this change. Because normally, the public namespace only contains interfaces. Now, the Bar Chart is the first component which breaks this "rule". On the other side, I do understand why these classes should be placed there. Because they are **not** internal and the consumer needs to create them when working with Bar Charts, right? So the public namespace is not only limited to interfaces as I assumed?

Greetings
Thomas

klees:
@tfamula Yes, this is exactly the reason why I moved the classes into the public namespace there.

I pondered a little if I should introduce interfaces for theses classes and hide the implementing classes behind a factory. I could not see much benefit there because all three classes are simple value classes (currently). This might be something to do in the future, though.

In general, the main idea to differentiate between a public and an implementation namespace is that this structure allows us to hide implementation details from our consumers. We most often achieve this by using interfaces, but in this case I think three value object classes are just fine. More importantly, we need to signal customers what they can safely use and rely on, which we still do.

Anyway, thanks for asking =)

PR #3756, opened 2021-11-12
UIComponent: Fix 'Access key function ids are integers not strings'

mjansenDatabay:
`Access Keys` are constants of type `integer`, see: `ilAccessKey`

Currently (latest trunk revision) these constants are passed to `Legacy Buttons` or `Toolbars Buttons` by consumers, but these classes expect arguments of type `string`.


alex40724:
Yes, I took the wrong type here. Thanks for the fix!

PR #3755, opened 2021-11-11
JS, Forms: move JS to global template

nhaagen:
This removes more JS loading errors.

PR #3758, opened 2021-11-12
JS, Forms: 'softer' use of  global; fix broken tests

nhaagen:
https://github.com/ILIAS-eLearning/ILIAS/pull/3755 broke a lot of tests (sorry for that).
This allows for "tpl" to be not available in global $DIC and checks before adding JS - that would be useless anyways in a scenario w/o template, right?


klees (approved):
I'm not convinced that this is the "right" solution, but at least it is consistent and fixes the tests.

Best regards!

PR #3761, opened 2021-11-12
Tree: Fix wrong tree root id fallback in constructor

mjansenDatabay:
This PR fixes a major (and silent) issue of setting the `root` ID of main tree to `0`, which leads to a lot of problems (e.g. with RBAC).


smeyer-ilias:
Thanks, merged.

PR #3762, opened 2021-11-12
Administration: Fix creation of ilObjectGUI derivatives

mjansenDatabay:
This PR fixes the creation of `ilObjectGUI` derivatives, which expect an array as first argument.


mjansenDatabay:
@alex40724 Any objections to merge this?

PR #3765, opened 2021-11-15
Refinery Randomization - Approach with Side Effects

lscharmer:
On behalf of the @ILIAS-eLearning/technical-board, this PR adds a new group `random()` to the ILIAS `Refinery`. This should replace the strategies of `Services/Randomization` to shuffle or keep the order of array elements (primarily consumed in `Services/TestQuestionPool`).

We implemented this in consideration of:

- This implementation uses the builtin `shuffle(...)` function instead of the `ilArrayElementShuffler::mtShuffle` because since PHP 7.1 `shuffle(...)` already uses the Mersenne Twister algorithm (_see https://www.php.net/manual/en/function.shuffle.php_).
- The existence check for `mt_srand(...)` and `mt_rand(...)` is also removed, because `mt_srand(...)` replaced `srand(...)` in PHP 7.1 (_see https://www.php.net/manual/en/function.mt-srand.php_).
- The added `ILIAS\Refinery\Random\Transformation\ShuffleTransformation::class` DOES seed & shuffle an array by itself. This violates the purity of Transformations. This exception is agreed upon by the coordinators of the refinery (Klees & Jansen) because the previously created PR https://github.com/ILIAS-eLearning/ILIAS/pull/3736 is to complicated and this approach offers a viable alternative.

We did not move `ilRandom` to the `Refinery`, because we do not consider it being a `Transformation`. `ilRandom` is widely used within the ILIAS code base, so we only see one option: Leave it where it is.

### DIFF
--- a/Modules/Test/classes/class.ilObjTestGUI.php
+++ b/Modules/Test/classes/class.ilObjTestGUI.php
@@ -156,6 +156,7 @@ class ilObjTestGUI extends ilObjectGUI
         $tree = $DIC['tree'];
         $ilias = $DIC['ilias'];
         $ilUser = $DIC['ilUser'];
+        $randomGroup = $DIC->refinery()->random();
 
         $cmd = $this->ctrl->getCmd("infoScreen");
 
@@ -659,7 +660,7 @@ class ilObjTestGUI extends ilObjectGUI
                 $this->ctrl->saveParameter($this, "q_id");
 
                 require_once 'Modules/TestQuestionPool/classes/class.ilAssQuestionPreviewGUI.php';
-                $gui = new ilAssQuestionPreviewGUI($this->ctrl, $this->tabs_gui, $this->tpl, $this->lng, $ilDB, $ilUser);
+                $gui = new ilAssQuestionPreviewGUI($this->ctrl, $this->tabs_gui, $this->tpl, $this->lng, $ilDB, $ilUser, $randomGroup);
 
                 $gui->initQuestion($this->fetchAuthoringQuestionIdParameter(), $this->object->getId());
                 $gui->initPreviewSettings($this->object->getRefId());
--- a/Modules/Test/classes/class.ilTestPlayerAbstractGUI.php
+++ b/Modules/Test/classes/class.ilTestPlayerAbstractGUI.php
@@ -1,6 +1,11 @@
 <?php
 /* Copyright (c) 1998-2013 ILIAS open source, Extended GPL, see docs/LICENSE */
 
+use ILIAS\Refinery\Transformation;
+use ILIAS\Refinery\Random\Seed\RandomSeed;
+use ILIAS\Refinery\Random\Seed\GivenSeed;
+use ILIAS\Refinery\Random\Group as RandomGroup;
+
 require_once './Modules/Test/classes/inc.AssessmentConstants.php';
 require_once './Modules/Test/classes/class.ilTestPlayerCommands.php';
 require_once './Modules/Test/classes/class.ilTestServiceGUI.php';
@@ -61,6 +66,8 @@ abstract class ilTestPlayerAbstractGUI extends ilTestServiceGUI
      */
     protected $testSequence = null;
 
+    private RandomGroup $randomGroup;
+
     /**
     * ilTestOutputGUI constructor
     *
@@ -81,6 +88,7 @@ abstract class ilTestPlayerAbstractGUI extends ilTestServiceGUI
         $this->processLocker = null;
         $this->testSession = null;
         $this->assSettings = null;
+        $this->randomGroup = $DIC->refinery()->random();
     }
 
     protected function checkReadAccess()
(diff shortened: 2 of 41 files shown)


klees (changes requested):
Hi @lscharmer,

thanks a lot for looking into this (again) and providing this alternative PR.

Although we would (consciously) introduce a violation to the "no side effect" rule with this PR, this feels much simpler then the other PR. We might want to look again into more principled ways to handle the PRNG and its seeds sometimes, but for now this approach here is fine for me, although not theoretically super sound.

- [x] Identity 1: I honestly don't understand what `IdentityTransformation::transformResult` (and the related methods) are trying to solve. Why is it not enough, to just return `$from` in `IdentityTransformation::transform` and `use DeriveApplyFromTransform` and call it a day?
- [x] ShuffleTransformation 1: The current implementation would reset the seed on every call to `transform`, right? This, in turn, would mean that the transformation creates the same sortation every time, correct? Is this the intended behaviour?

- [x] Side Effect: Let's name the side effect explicitly [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3765/files#diff-5388732cca6e19496deea7e60cc2480e539f0f06569e8abc7e8666767ec1f595R21): The internal state of the PRNG will be advanced on every usage.
- [x] Identity 2: Add it on top level of the refinery as outlined in [the initial paper](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/input-processing.md#improve-transformation-and-validation). Thanks for adding that transformation and tests for it!
- [x] RandomSeed: I assume that you use inheritance [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3765/files#diff-9970aff3152ee214ca9bded197a8b9c5b7322e728356b35885a1e37baf011e9bR8) to share code between `GivenSeed` and `RandomSeed`. But `extends` also establishes an "is a"-relation between the two classes. `RandomSeed extends GivenSeed` also implies, that any `RandomSeed` also "is a" `GivenSeed`, since a `RandomSeed` would pass typehints for `GivenSeed`. I don't think that this is what you want to express. So I would suggest to give up on the code sharing here and keep the model intact, although this might feel super minor.
- [x] ShuffleTransformation 2: Maybe this is the answer to **Identity 1**? Similar to the point before, I would not agree that a `ShuffleTransformation` "is a" `IdentityTransformation`. This could go wrong [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3765/files#diff-e9b86baf62ad653fbbe9bb9eb7f07aeffb3d488ad3f0a42236ebdc4e918c95a5R38) already...
- [ ] Test: Could we test [here](https://github.com/ILIAS-eLearning/ILIAS/pull/3765/files#diff-b5117400c4c352ee750b5ce1c155b32370ed4a50ca29fda9f64b4ad6644a83d5R17) if `ShuffleTransformation` indeed shuffles?

- [x] Typo 1: [expectation -> exception](https://github.com/ILIAS-eLearning/ILIAS/pull/3765/files#diff-aa3a495aa62d97b46004ce0a1bd06631105988b29657e253881fc5a95ba88870R16). Maybe link the PRs?

My colleagues from the @ILIAS-eLearning/technical-board might look into this as well until the end of the week. I would merge this when my change requests are incorporated, if no veto or anything new come up.

Feel free to contact me on short notice (e.g. via Skype) if you have any question or anything else.

Best regards!

lscharmer:
Hi @klees,
thank you for your feedback.

- Identity 1: I did not use the `DeriveApplyFromTransform` trait, because the implementation of `DeriveApplyFromTransform::applyTo` is incorrect (in my opinion).
- ShuffleTransformation 1: `ilArrayElementShuffler::shuffle(...)` initializes the PRNG also on every call.
- Side Effect: will be added
- Identity 2: will be changed
- RandomSeed: This is a good point, I did not thought about that, but I would still disagree. The `Given`- and `RandomSeed` classes are implementation details that should not be type hinted on. For type hinting the `Seed` Interface must be used. Alternatively I would implement an abstract class where GivenSeed is just an empty class that extends AbstractSeed.
- ShuffleTransformation 2: Yes, that is why implemented Identity 1 this way. Same as RandomSeed, I do not think that one should depend of the `IdentityTransformation` type but only on `Transformation` and I would propose the same alternative.
- Test: will be changed
- Type 1: will be fixed

If anything remains unclear, please contact me.

Best regards.

klees:
Hi @lscharmer!

Could you explain a little why you consider `DeriveApplyFromTransform::applyTo` to be incorrect?

Best regards!

lscharmer:
Hi @klees,

the `Result` interface provides a `then(...)` method which binds a new calculation to an existing one. Because the calculation is provided as a callback it may be never evaluated or be evaluated at another point in time.
When the `value(...)` method of the `Result` is called (as it is in `DeriveApplyToFromTransform`) the whole execution chain MUST be executed, which it would not be when using `then(...)` instead.

Currently this does not cause any issues, because there is only one implementation of the `Result` Interface, but if future releases would provide for example a lazy implementation of `Ok` and `Error` all results would be always evaluated which would render the lazy implementation pointless.

If `DeriveApplyToFromTransform::applyTo(...)` is the intended usage the `map(...)` and `then(...)` methods are merely an alternative to an if statement or exceptions.

Maybe we should discuss this in detail with you and @mjansen via Skype.

Best regards.

lscharmer:
> Hi @lscharmer!
> 
> Could you explain a little why you consider `DeriveApplyFromTransform::applyTo` to be incorrect?
> 
> Best regards!

Hi @klees,
I hope I  was able to explain this issue sufficiently, if it remains unclear please contact me.

Best regards.

lscharmer:
All points discussed with @klees and @mjansenDatabay are now commited:
* `ShuffleTransformation` uses `DeriveApplyFromTransform`
* `IdentitiyTransformation` uses `DeriveApplyFromTransform`
* `DeriveApplyFromTransform` uses `Result::then(...)` instead of `Result::value(...)`

PR #3766, opened 2021-11-15
Updated color codes (from 3-digits to 6-digits)

padvincenzo:
Hi, I've just replaced all svgs rgb color codes, from 3-digits to 6-digits, in order to solve a [problem mentioned in mantis](https://mantis.ilias.de/view.php?id=31499).

All previous ``#000`` are now ``#000000`` and ``#fff`` are now ``#ffffff``


Amstutz:
Hi @padvincenzo

Your PR is highly appreciated. Thx a lot! How did you change the colors, did you write some script?

Your work and the sheer amount of icons needing change got me thinking; probably we will deal with this issue in future as well. I therefore provided and pushed a fix, enabling the feature to be able to deal with color values of 3 digits only. The PR here is therefore not needed anymore. Nevertheless, it was important to show the scale of the issue we are dealing with. So thx a lot.

padvincenzo:
Hi @Amstutz
I totally agree with you. It is always better to fix the root of the problem, not the leafs.

Here is the command in bash I used to convert all 3-digits color codes to 6-digits:

```sh
find * -type f -exec sed -E -i 's/#([a-f0-9])([a-f0-9])([a-f0-9])([^a-f0-9])/#\1\1\2\2\3\3\4;/gi' {} \;
```

I currently didn't find a way to filter svg files (without loosing the recursive feature)... the best I could do was moving to the images folder first, and then execute the command.

PR #3769, opened 2021-11-15
Fix Glossary Print Presentation

kergomard:
This fixes the issue with missing titles in the Glossary Print View:

https://mantis.ilias.de/view.php?id=31593


alex40724:
Thanks!

PR #3776, opened 2021-11-17
UI/Mainbar: trigger in_view for fully engaged path only, store loaded…

nhaagen:
In reference to https://github.com/ILIAS-eLearning/ILIAS/pull/3698, this changes the mainbar's loading-behavior;
'in_view' will be triggered for engaged slates, but only if _all_ upper levels are engaged as well.

### DIFF
--- a/src/UI/templates/js/MainControls/dist/mainbar.js
+++ b/src/UI/templates/js/MainControls/dist/mainbar.js
@@ -424,6 +424,21 @@ var model = function() {
                 }
             }
             return null;
+        },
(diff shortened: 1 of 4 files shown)


klees (approved):
LGTM

klees:
@Amstutz: Wanna have a look?

Amstutz:
I did not review the code line by line in this case, however, I did check this out and tested locally. Works as intended.

Amstutz:
@klees will pick to trunk, we will not push down to 6.

PR #3778, opened 2021-11-18
Refactoring of Services/Component, Part 1

klees:
Dear colleagues,

since I took responsibility for `Services/Component` as a maintainer, I started to refactor the component to make it fit the requirements from the PHP 8 project and move it closer to contemporary practices of ILIAS development. I am not there yet, but I want to bring this first part of the refactoring to the core asap so that people start to notice and work with it.

This basically accomplishes three things:

* The information about components and plugins won't be stored in the database anymore (mostly), but instead static artifacts that are generated on `composer install` (or `setup\setup.php build-artifacts` respectively).
* The functionality from `ilPluginAdmin` is moved to `ilComponentRepository` and according model classes that bring a clarified API and automated tests.
* A lot of duplicate functionality, superfluous information passing and questionable call patterns are removed.

These changes, of course, imply a lot of changes in the consumers of `ilPluginAdmin`. There is one thing I could not get rid of due to the intricate nature of the T&As dependency pattern: `ilPluginAdmin::isActive`. I will keep it for the time being, but I hereby publicly (and hopefully noticeably) announce: **I will remove `ilPluginAdmin::isActive` with ILIAS 9. Please make sure to switch your usages to the new patterns until then.** I marked the class and the method accordingly. If there are any questions or concerns (or you need help), feel free to contact me, as always.

When implementing changes in the consumers of `Services/Component`, I observed that a lot of consumers store the names of plugins as handles to them. I mostly consider the name of a plugin as a string to be displayed, not as a means of identification. Thus, **I advice everyone to use and store the id as a handle to a plugin**, which is the expected access pattern for the API of `ilComponentRepository` and friends.

This does not change the interface that components or plugins must use, so everything should keep working as before after this change. I'm looking to make improvements here as well, which I will address on next Jour Fixe for big projects for ILIAS 9.

I will present this PR on the next JF and plan to merge this soon after. If there are any questions, concerns or feedbacks regarding this PR or the general course of `Services/Component`, feel free to post them here, address them at the next JF or contact me via skype, mail etc.

Thanks for looking into this and best regards!


klees:
I added almost all maintainers (hope I did not forget anyone...) as reviewers since this contains changes in a lot of places. Most of them are not very interesting, though, but just replace calls to `ilPluginAdmin` by calls to the new API.

Services/Component/classes/Setup/class.ilComponentBuildPluginInfoObjective.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        $path = static::BASE_PATH . "$type/$component/$slot/$plugin/" . static::PLUGIN_PHP;
+        $file = $this->readFile($path);
+        if (is_null($file)) {
+            throw new \RuntimeException(
+                "Cannot read $path."
+            );
+        }
+
+        eval("?>" . $file);
```
mjansenDatabay: Just a short inline feedback: Do we really want to rely on `eval` here?

klees: I did not like this too. But: Any better idea? Also: Is this really a problem?

mjansenDatabay: >  Any better idea?

ATM not, therefore I had to competely understand the change (mental model) ;-).

> Is this really a problem?

I don't think so. In the past some ILIAS institutions blocked the `eval` function (because of security concerns) and only allowed the execution during the setup. So with your database update step revision we got finally rid of `eval`, and now we introduce it once again :-).

klees: Yeah, I'm really looking to get rid of this completely soon =) This is what I want to talk about (among other things) on the Project JF.

PurHur: Why dont you use an `include`? Like it was used in `ilPluginAdmin`:

```
 protected function parsePluginPhp($plugin_php_file)
    {
        include($plugin_php_file);

        $values = [
            "version" => $version,
            "id" => $id,
            "ilias_min_version" => $ilias_min_version,
            "ilias_max_version" => $ilias_max_version,
            "responsible" => $responsible,
            "responsible_mail" => $responsible_mail,
            "learning_progress" => (bool) ($learning_progress ?? false),
            "supports_export" => (bool) ($supports_export ?? false),
            "supports_cli_setup" => (bool) ($supports_cli_setup ?? true)
        ];

        return $values;
    }
```

Is this faster?  Or because of the tests?

klees: @PurHur: The tests indeed.

matthiaskunkel:
Jour Fixe, 29 NOV 2021 : Richard asks all maintainers and developers to read this PR carefully and consider its impact for the future development of ILIAS. In case of questions and feedback, please contact Richard directly.

mjansenDatabay (approved):
LGTM!

chfsx (approved):
Thanks @klees for this work, LGTM!

PR #3779, opened 2021-11-18
E-Mail: clickable mixed content type

lscharmer:
Bugfix for https://mantis.ilias.de/view.php?id=31597.

Depends on #3782

### DIFF
--- a/Services/Mail/classes/class.ilMimeMail.php
+++ b/Services/Mail/classes/class.ilMimeMail.php
@@ -1,6 +1,8 @@
 <?php declare(strict_types=1);
 /* Copyright (c) 1998-2021 ILIAS open source, Extended GPL, see docs/LICENSE */
 
+use ILIAS\Refinery\Factory as Refinery;
+
 class ilMimeMail
 {
     public const MAIL_SUBJECT_PREFIX = '[ILIAS]';
@@ -28,6 +30,7 @@ class ilMimeMail
     protected array $adispo = [];
     /** @var string[] */
     protected array $adisplay = [];
+    private Refinery $refinery;
 
     public function __construct()
     {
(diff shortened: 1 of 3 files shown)


tests/Refinery/String/MakeClickableTest.php
```diff
@@ -0,0 +1,53 @@
...
+namespace ILIAS\Tests\Refinery\String;
+
+use PHPUnit\Framework\TestCase;
+use ILIAS\Refinery\String\MakeClickable;
+
+class MakeClickableTest extends TestCase
+{
+    public function testConstruct() : void
+    {
+        new MakeClickable();
```
mjansenDatabay: I prefer the `instanceof` assertion instead of a static `true` ;)

tests/Refinery/String/MakeClickableTest.php
```diff
@@ -0,0 +1,53 @@
...
+    public function testTransformSuccess(string $expected, string $input) : void
+    {
+        $clickable = new MakeClickable();
+
+        $this->assertEquals($expected, $clickable->transform($input));
+    }
+
+    public function provideInput() : array
+    {
+        return [
```
mjansenDatabay: I'd prefer some `Lorem Ipsum` texts instead of random (cat ran over the keyboard) characters :).

mjansenDatabay (changes requested):
Thx for the PR @lscharmer

Could you please create a separate PR which only includes the `Refinery Transformation`?

mjansenDatabay:
@lscharmer Could you drop the non-related commits and rebase this on `trunk` after #3782 has been merged?

mjansenDatabay (approved):

PR #3780, opened 2021-11-18
Setup: add composer dependency Seld\JsonLint and use it in ConfigReader

daniwe4:
Using the Seld\JsonLint lib will help to locate errors in the setup config file. Its output contains line number and error type.


matthiaskunkel:
Jour Fixe, 29 NOV 2021 : We highly appreciate this suggestion and accept the PR and library for trunk.

PR #3782, opened 2021-11-18
Refinery: MakeClickable Transformation

lscharmer:
New Transformation to convert plain URL's to anchor tags.
This PR will be used to fix the Clickable Bug of https://mantis.ilias.de/view.php?id=31597.

This may replace `ilUtil::makeClickable(...)` in the future.

Tested for multibytes: All functions operate on bytes and multibyte tests are added.


mjansenDatabay:
Hi @lscharmer,

thanks for this contribution to the ILIAS `Refinery`.

- [ ] Add Tests for Multibyte Characters: Please provide more tests which cover multibyte string scenarios. All string functions you used here a not multibyte safe. Please consider using `ilStr` instead.

Best regards,
@mjansenDatabay

mjansenDatabay (changes requested):
Hi @lscharmer,

thanks for this contribution to the ILIAS `Refinery`.

- [x] Add Tests for Multibyte Characters: Please provide more tests which cover multibyte string scenarios. All string functions you used here a not multibyte safe. Please consider using `ilStr` instead.
- [x] Please add a reference to the comments of `\ilUtil::makeClickable', so that developers are aware of the new method. Please also add the following code:
```php
trigger_error(
   'Use the respective `Refinery` transformation `$refinery->string()->makeClickable("foo bar")` to convert URL-like string parts to an HTML anchor (`<a>`) element ',
   E_USER_DEPRECATED
);
```
- [x] Please update the `ROADMAP.md` that addresses a revision of the internal implementation of `MakeClicktable` to a (for instance) DOM parsing/modification approach

Best regards,
@mjansenDatabay

mjansenDatabay (approved):

src/Refinery/String/MakeClickable.php
```diff
@@ -0,0 +1,58 @@
...
+            return \strpos($string, $matches[0]);
+        }
+
+        return \strlen($string);
+    }
+
+    private function requireString($maybeHTML) : void
+    {
+        if (!\is_string($maybeHTML)) {
+            throw new \InvalidArgumentException('not a string');
```
mjansenDatabay: Can you use a `ConstraintViolationException` instead?

mjansenDatabay:
@klees Do you have any objections? If not, I will merge this after @lscharmer tackled the requested changes.

tests/Refinery/String/MakeClickableTest.php
```diff
@@ -0,0 +1,58 @@
...
+            'test without protocol to URL with protocol' => ['<a href="https://www.ilias.de">www.ilias.de</a>', 'www.ilias.de'],
+            'test link with parameters' => ['<a href="http://ilias.de/ilias.php?ref_id=29&admin_mode=settings&cmd=view&cmdClass=ilobjdashboardsettingsgui&cmdNode=1c:lg&baseClass=ilAdministrationGUI">http://ilias.de/ilias.php?ref_id=29&admin_mode=settings&cmd=view&cmdClass=ilobjdashboardsettingsgui&cmdNode=1c:lg&baseClass=ilAdministrationGUI</a>', 'http://ilias.de/ilias.php?ref_id=29&admin_mode=settings&cmd=view&cmdClass=ilobjdashboardsettingsgui&cmdNode=1c:lg&baseClass=ilAdministrationGUI'],
+            'test multibyte' => ['I♥ILIAS', 'I♥ILIAS'],
```
mjansenDatabay, line 53: Could you change this test case (or add another one) to something like `I♥ILIAS' <a href ...>...</a> I♥ILIAS`.

PR #3785, opened 2021-11-21
error to import xml file with open_basedir set

rapmue:
Fixed an error when open_basedir restriction is in place and the $_SESSION variables are not set.

Not set $_SESSION variables result in a filename like //manifest.xml which is not in the permitted paths and the execution of the script crashes.

To reproduce, restrict the execution of the php script to /tmp and the upload and main folder of ILIAS.

### DIFF
--- a/Modules/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php
+++ b/Modules/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php
@@ -697,7 +697,7 @@ class ilObjQuestionPoolGUI extends ilObjectGUI
             $newObj->setPermissions($_GET["ref_id"]);
         }
 
-        if (is_file($_SESSION["qpl_import_dir"] . '/' . $_SESSION["qpl_import_subdir"] . "/manifest.xml")) {
+        if (@is_file($_SESSION["qpl_import_dir"] . '/' . $_SESSION["qpl_import_subdir"] . "/manifest.xml")) {
             $_SESSION["qpl_import_idents"] = $_POST["ident"];
             
             $fileName = $_SESSION["qpl_import_subdir"] . '.zip';


PurHur:
I stubled on this PR and now i wonder if Ilias is able (or should) to run properly without $_SESSION? Just curious.

rapmue:
Not all the $_Session variables are empty. At least the two which are used to combine the path to check dir the manifest.

I'm not an ILiAS specialist (I just made one Installation for a friend of mine). In my opinion this session variables used for the check are only set if a zip file ist uploaded (not tested).

PurHur:
I am not a real Ilias developer but i feel like that PR is not needed at all. 

Sorry for the inconvenience. But the open_basedir (and disable_functions) setting is somewhat personal to me because i was the one who removed the comment that this is a security feature from the php codebase. If you use that configuration, and you understand what it does, configurate it well enough to get the software running.  If you use that as security feature (i heavily think so) remove it. You got a false sense of security which doesnt do anything at all to prevent attacks. 

The `@` here just silence a error for all other people running the code.

rapmue:
I agree on the open_basedir topic about security feature. Let's say open_basedir is just there to detect wrong "configured" code which want to access unwanted ressources, like a unit test.
I also agree about the silencing with @ and that there should be something else in place.

So maybe we should focus on why exactly the path is pointing to //manifest.xml ?

PurHur:
The code runs wrong because the session path is configured wrong. If you dont allow $_SESSION to work a whole lot of other code also dont work as expected.

PurHur:
Ohhhhhhh. Now i get it. Shame on me not seeing the problems =) Thanks kek

mbecker-databay:
Here's no movement since a long time and the actual change as per the PR was discussed.
I close this PR for now, if you have a result on the open question as per the remark of 22 NOV 21, please feel free to open a new PR. Thanks a lot!

PR #3787, opened 2021-11-22
UI Components: Interface for Item/Shy

iszmais:
This PR proposes to introduce the UI-Component for implementing Shy Items
This item will initially be used for the new presentation of minimized chat inside the menu.

The Shy Item is a minimalistic visual item wich shows only a title, description, some properties and a close action and a lead icon.

It occupies a rectangled space where the description is shown in the middle, the title above and the properties below. It may also show a leading icon preffixed on the left side of the Shy and/or a close button in the top right corner of the Shy.

image

As explained in https://docu.ilias.de/goto_docu_wiki_wpage_6251_1357.html a separation of notifications an chat is sought. Therefore the new chat presentation shouldnt consist notification items since these are only allowed to be used inside the metabar.


src/UI/examples/Item/Contribution/base.php
```diff
@@ -0,0 +1,16 @@
+<?php declare(strict_types=1);
+
+namespace ILIAS\UI\examples\Item\Contribution;
+
+function base()
```
PurHur: Why do you use a function here instead of a class? This makes extending/overwriting it very very hard. This applies to all the other functions.

mjansenDatabay: @PurHur This is an example, this code MUST not be used in productive code, nor SHOULD it be derived.

PurHur: Wouldnt an example be better if it shows how to use in a real code environment? As outside developer i have no clue what todo with that function.

Amstutz: Why would you want to extend the example? What is a real "code environment"? 

The idea here is to show a few short examples of how the component can be generated and rendered, next to the actual output that is rendered by the code. For me, this is rather useful. It does not matter if I would wrap a class around it. If I would use the code, I would but it again into a function (probably inside a class or trait), not?

Further it is handy for testers, to have one central place where they can check if each UI Component looks like it is supposed to.

iszmais: @PurHur Thx for the review!
As said these functions arent for production or to use inside actual content structure, but to give developers exaples about the usage of new components. There is no usecase for that, its more like a documentation

iszmais:
@Amstutz @klees 
Could you take i quick overlook on this (since i would like to present it on the next jour fixe)?
Im aware and apologize that this is very short term. This was an missed out required component for the new changes on the Onscreenchat presentation (https://docu.ilias.de/goto_docu_wiki_wpage_6251_1357.html) so i would highly appreciate it so i can move forward with the supordinate project.

Amstutz (changes requested):
Hi @iszmais 

Thx a lot for the proposal. Special thx for adding the examples already. This helped us to better understand, how this component can/will be used.

Some questions:

- [x] Low-Fidelity Mockup / Sketch: For completely new components, or new existing once in a new context it often helps, to provide a small sketch or low-fidelity mockup into the description. Pen and paper is perfectly fine. Note that you can easily provide those by dragging a picture here into the description field or comments field on github. I did see the link to the FW page and guess that this refers to the chat contribution moving to the Main Bar, right? We would recommend adding one for the JF in this case / or maybe you will use the FW page which also works.
- [x] Context: Can you give some examples on where you will use this? From the JF article I guess this will be used for the chat. If so, you can list this directly in the context section.
- [x] withIdentifier: What will this identifier by used for? Why do we need it? Why does the id that you can generate and later refer to not suffice?  We are looking to provide composable components in the UI framework. Ids are problematic in that regard because on a page that is composed of components from multiple sources there simply is no means to be sure that an id is actually unique. There is an [existing solution for that problem](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/src/UI/Component/JavaScriptBindable.php#L14). Can we reuse it for the this component?
- [x] withLeadIcon: Will this really be an icon, or can it also be the users avatar? Maybe it should be just an `Avatar`?
- [ ] `$content`: Is `$content` really only a string or will there be more to it? Do you, e.g., expect that some rendered version of other UI components will be ending up in `$content`? This feels as if there is some specification missing here. This seems to be related to the questions regarding the `Context` and the `Purpose`.
- [ ] Naming: Would `Message` be an equally good name for the component? `Contribution` seems to be rather broad... (see `Purpose` and `Context`).

Some thing you should extend on or give some reason why you prefer not to:
- [ ] Purpose: You write "Contributions are content that can be credited to one explicit user and time." That would be probably all of the content. Can you drop some words on when such content could/should be highlighted with "Contribution Items". E.g. if it is important for others to see who exactly contributed what at which point in time (e.g. in some chat-like context).
- [ ] withDateTime/getDateTime: Look for some better name to clearly indicate, what this DateTime refers to. E.g. the creation date of the contribution? Its last change? E.g. rename to "getContributionCreateDateTime".
- [ ] DateFormat: We should expect that the developer using this component wants to specify a formatting for the date time. [`DateFormat`](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/src/Data/DateFormat/DateFormat.php#L10) would come in handy...
- [ ] Item: This most probably is not an [`Item`](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/src/UI/Factory.php#L489) in the current sense of the UI framework and thus should not be located in that section. We might want to style it rather similarly, but semantically it seems to be different. Currently we have a hard time to see how this new component could be located in a section. We would propose to locate it in the root for now.

Things to change:
- [ ] withUser, getUser, Constructor: Complete User Objects are injected into this component. User objects have a huge surface to work on. This surface is far to big for the UI Component. Also, it breaks with the paradigm that UI Components are mostly agnostic of ILIAS objects. Clearly state what is needed of the User object and pass this (e.g. username or name representing author of contribution or something like this). The way in which the user name will be presented, should and is not done by this component anyways. 
- [ ]  Spelling; Change "Contribution items" to Contribution Items (both words capitalized)
- [ ] Constructor: PHP might allow adding [constructors to interfaces](https://github.com/ILIAS-eLearning/ILIAS/pull/3787/files#diff-a2d20b679901f7ae0c0917de6995624613792c89d5ac10ef4a6bb23e4558aa14R21). But it just does not make any sense to have a constructor on an interface. You neither won't be able to construct an instance of an interface (only classes can be constructed) nor need it on already constructed implementations of the interface. On the flipside, you will commit implementors to an implementation detail, that you will not be able to rely on anymore. Maybe I just want to provide a constant `$content`? Please remove the constructor from the interface.
- [ ] DateTime: Please use `DateTimeImmutable` from PHP for datetimes.

We are looking forward to discuss this, but we currently do not feel that this is ready for the JF. Feel free to contact us to ponder on our questions and find some solutions so we can get this to the JF real quick.

Thx
@Amstutz and @klees

iszmais:
Hi @Amstutz ,

thanks for your feedback!

**To the questions:**

- Added a screenshoot.
- Added a description inside the PR. But i would like to keep it out of the interfaces itsself since , in my opinion, the component can have multiple other purposes for other quotable represantions in all areas.
- The identifier isnt primary a technical tool but more for verification in an conceptual way. This will give the developer who use it a option to leave a content, wich proves the authenticity of a quote/contribution whithout using visual space for it (since this prove may not be neccessary for most of its viewers). For example this could be a ISBN of a quotet source or another analog used Identifier. Since, as you can see below, i removed the user from the component this type of authenticity seems even more ipmortant for quotations.
- The leadIcon should only provide a classification inside a context to increase the readability of mutliple items listed. Its not to extend the presentation of the user/contributor itsself (even more since its not rewuired to be a ilias user now).
- This should only be textual and will be trimemd that way. I changed the keyword to quote to enhance clarity here.
- We had the same discussion here localy :D .As result we went from "message" or "comment" to "contribution" since its more general to fit the context of the component. A message implies to be directed to a person and a comment implies to be directed to an object (like another comment, image, ...). A contribution would fit both of these purposes.

**To the extends:**

- changed it. In general the contribution is not only content which is created by a person but whats value depends on that contributor liek a rating or a critic.
- changed it
- changed it
- In my opinion the item scope fits it since it follows the same purposes like the standard item with the different that its focus lies on its creator not on it content itself. But im not familiar with the practical usage of the standard item and its soft convetions in the existing ilias context. Could you further explain where you see the mismatch here?

**To the changes**

- changed it
- changed it
- changed it
- changed it

Feel free to further give feedback to the disscussable entries.

Greetings,
@iszmais

klees:
Hi @iszmais,

thanks for the feedback, we won't be able to look into it in detail today.

Thanks for the clarifications, anyway. As some food for thought: Could "quote" or "testimonial" work as a name. I feel that finding a good name is crucial to understand what this component is, "contribution" feels strange, but I see how "message" or "comment" does not fit as well. Will think about this...

Best regards and have a nice weekend!

Amstutz:
Thx, also from my part for the provided feedback. I see nothing that would prevent a discussion on monday in the jf. 

Looking forward to it. Things left to decide upon:

- [ ] Item Scope, Does it fit in there?
- [ ] Name: Is "contribution" right.

until then.

matthiaskunkel:
Jour Fixe, 29 NOV 2021 : We highly appreciate this suggestion but ask for a final discussion about the name of this UI component and where it is placed within the KS (sub- or top element). Please come back to the JF once these decisions have been taken.

alex40724:
Hi,

thanks for the proposal. The current concepts seems to target also comments. The current comments presentations in ILIAS would definitely need an avatar instead of a leading icon (since it would always be the same icon).

I think we need to either
a) decide to not use this for comments in ILIAS
b) change it to support avatars or
c) remove avatars from comments (which I would not like)

Best
Alex

iszmais:
Result of the final Feedback from 03 DEC 2021 : 
Ths component will be renamed to a more absract, but also more fitting for the item section, Shy. These will present a simplyfied Standard Item. (With some minor changes on the other Item Components).

The concept of a Contribution as a own component for authentic user priorised Content will be kept open for Discuss inside https://github.com/ILIAS-eLearning/ILIAS/pull/3816. This will have no dependencies to the Item component.

iszmais:
@klees 
Could you rereview the renamed component?

klees (approved):
Hi @iszmais!

Thank you for the great discussion the other day and also for reworking this. Looks as if we are there, mostly.

This feels a lot more in line with the current intent and structure of item and also with the use case you are looking to implement with this. Really nice that we could reuse that "shy" qualifier from other components.

- [x] Uses: We don't need [these `use` statements](https://github.com/ILIAS-eLearning/ILIAS/pull/3787/files#diff-b84dddcb4040f0309648a00e52da9ede0b8c05dd33ac7aa7ba4cc02e9ab607a5R5).

Best regards!

iszmais:
Hi @klees ,

thx for the quick feedback!
I fixed the uses.

Greetings

matthiaskunkel:
Jour Fixe, 13 DEC 2021 : We highly appreciate this suggestion and accept the PR for trunk.

PR #3791, opened 2021-11-24
Composer: `composer update` + Upgrade simplesamlphp and symfony/yaml

mjansenDatabay:


Amstutz (approved):
thx @mjansenDatabay for the version udpate! If any issues arise in trunk in the listed component of mine due to the change, I will take care of them.

klees:
Thx @mjansenDatabay. Same here.

mjansenDatabay:
@Amstutz / @klees Thanks for your feedback.

I hate the dependency hell :-).

I'll have to update all dependencies in `trunk` to make the deps. work in `PHP 7.4` and `PHP 8.0`

mjansenDatabay:
@Amstutz I had to change the `YAML Parser` call in `src/UI/Implementation/Crawler/EntriesYamlParser.php`.

This static call was the recommended usage I found in the documentation. If you don't agree, feel free to push changes to the PR branch or contact me.

Amstutz (approved):
Thx for taking care of this @mjansenDatabay . 

I dislike to usage of static methods a lot. I currently try to get rid of as many of them as possibe in the system styles. However, I guess there is not much we can do, since this is an external lib, right?

If so, this is imo good for my part.

thx!

mjansenDatabay:
> However, I guess there is not much we can do, since this is an external lib, right?

According to the [docs](https://symfony.com/doc/current/components/yaml.html#reading-yaml-contents) this seems to be the way to go, unfortunately ...

But this static convenience methods only wrap the known methods of the `Parser`, so I reverted the usage of the static call.

PR #3802, opened 2021-11-30
Setup: Fix error when duplicate setup agent is detected

mjansenDatabay:


klees:
Thx! Seems to be ok in trunk already.

PR #3809, opened 2021-12-02
UI: Introduction of `DynamicInputsAware` inputs

thibsy:
Hi @klees,

As promised, here the first piece of the d&d feature request (#3712) which covers the `DynamicInputsAware`, or previously `AdditionalInputsAware` inputs. I took some of your advice and implemented some suggestions too.

This pull-request only introduces **the possibility** of such inputs, I did not implement a standalone input yet. Developers can extend the abstract base class `DynamicInputsAwareInput` and use the methods `addDynamicInputsAwareClientsideRenderer()` and `renderDynamicInputsAwareInputs()` in `ILIAS\UI\Implementation\Component\Input\Field\Renderer` in order to implement their own versions of this input-type. The script `dynamic_inputs_renderer.js` or accessible via `il.UI.Input.DynamicInputsRenderer` is responsible for the clientside rendering of the dynamic input template. Developers can trigger or render a dynamic input by calling its `render()` function, that must be supplied with the original or "parent" input ID.

Best regards!


klees:
Hi @thibsy,

thanks a lot for working on this basic functionality for processing forms, this will come in handy for more than the file inputs that you made this for. Much appreciated! I know, that this is not easy and I'm sorry that I need to send you into another iteration. I'm 100% sure this is worth the effort, though, since this will become basic tech for other functionality.

- [x] initClientsideRenderer: This method is not invoked anywhere. What am I missing?   

- [x] withTemplateForDynamicInputs: I wonder why this is a mutator and not just a parameter to the constructor (for the abstract implementation) or even not defined at all (for the abstract implementation). Is this really something that we would want our users to exchange dynamically? I thought a little about other use cases for the class, and the "Multi Input" (where you can add new fields via [+] or remove them via [-]) from the legacy form implementation certainly is such a case. Would we really want to be able to exchange the contained fields after that kind of input is created? To iterate one further: I could think of something like a predefined "Multi User Input", where zero to many usernames could be selected. This would certainly be an instance of `DynamicInputsAware` (right?) but we would not want our consumers to be able to exchange the contained input. What would I do when someone calls `withTemplateForDynamicInputs`? To me this feels as if we would want to remove `DynamicInputsAware::withTemplateForDynamicInputs` from the interface. We could then, also, get rid of some edge-case handling in `withInput`.
- [x] template: Use a [`<template>`](https://developer.mozilla.org/de/docs/Web/HTML/Element/template) to render the template input fields. You should be able to get rid of some [renaming stuff](https://github.com/ILIAS-eLearning/ILIAS/pull/3809/files#diff-f3892e7d92b0895c6b153d2f1754b755fc888a4b5199b1d4f8784a7344b81d5fR842), it should be simple to identify the template fields client side and it should be a path to abstract some more rendering functionality, client and server side.
- [x] Docu: Please add a few words about purpose and behaviour for the classes `DynamicInputsNameSource` and `DynamicInputsTemplateNameSource` (if we really need both...).
- [x] overridePostInputNames: `DynamicInputsAwareInput::overridePostInputNames` seems to be something that rather belongs to `DynamicInputsTemplateNameSource` (if we really need it...).

- [x] Test 1: `ArrayInputData` is missing tests. Please add some.
- [x] Test 2: `FormInputNameSource` is missing tests. Please add some.
- [x] Test 3: `DynamicInputsAwareInput` is missing tests. Please add some.
- [ ] Test 4: Clientside code is missing tests. Please add some.
- [x] Cleanup: [This](https://github.com/ILIAS-eLearning/ILIAS/pull/3809/files#diff-8ce13a0cc864734cd3dea59929cab00a9efc124714a9267e21c29f815fced5e7R167) is not used anywhere.

Overall, `DynamicInputsAwareInput::withInput` does not feel quite right at the moment, but I'm not sure I can really put my finger on it. Some hints:

* [This artificial input](https://github.com/ILIAS-eLearning/ILIAS/pull/3809/files#diff-8ce13a0cc864734cd3dea59929cab00a9efc124714a9267e21c29f815fced5e7R159) feels strange. Why do we pretend to have a `null` here? Wouldn't we just fail to pull input from POST (and show errors) when iterating over inputs, just as we do in the `Group`?
* Why do we need that complicated process to override POST-names in `withInput`? Can't we just give names to the template fields as usual and then enumerate the occurences of the templates? Let's say we give the name "dynamic" to the combined field, and "template_1" to "template_n" to the template fields. What's stopping us to use "dynamic[0][template_1]" to "dynamic[0][template_n]" for the fields in the first dynamic input group, "dynamic[1][template_1]" to "dynamic[1][template_n]" for the fields in the second, and so on. This would get rid of almost all the renaming logic, the additional name sources (server and client side) and should also simplify `withInput`.

I might be missing some complexity which might just make my points moot. If you want to discuss my remarks or figure out together if the proposed simplifications would work, feel free to contact me. I'm absolutely available as co-driver in a pair programming session =)

Best regards!

klees (dismissed):
:point_up: :grin:

thibsy:
Hello @klees 

I'd really like to figure out some issues together and would welcome a pair programming session :).

Then I could also explain some of the questions to you in person, which might be easier than to describe it. I'll write you an email to figure out an appointment.

Kind regards!

thibsy:
Hello @klees

I was finally able to push the changes we made during our pair programming session =). You can find one small adjustment in the `DynamicInputDataIterator`, where `current()` uses the `DynamicInputsNameSource` in order to map the value to the original input name. This had to be done so I could delegate the retrieved `InputData` to the dynamic inputs template. 

I think now everything noted in your review should be covered, except for the javascript tests. The simple reason for this is that I don't feel quite sure about how to implement them and don't have the capacity to look into this at the moment. Other than that I think this feature should ready to be merged now, right? 

Kind regards!

klees:
@thibsy Right! I will look on monday.

src/UI/Implementation/Component/Input/DynamicInputDataIterator.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        $this->post_data = $data->getOr($parent_input_name, []);
+        $this->parent_input_name = $parent_input_name;
+    }
+
+    public function current() : ?InputData
+    {
+        if ($this->valid()) {
+            $entry = [];
+            $name_source = new DynamicInputsNameSource($this->parent_input_name);
```
thibsy: I just noticed, that this gets rather stupid if the InputData should be created for a Group input, because then we would have to call `getNewName()` once to skip the name of the Group itself. For all other inputs that wouldn't be necessary, so an optional Parameter when constructing the Iterator would be necessary to do something like this here:
`if ($this->is_input_group) { $name_source->getNewName(); }`

thibsy: `Group::withNameFrom()` calls the parent method which set's the groups name, though it is never used except I am missing something?. The problem would resolve if we just change this method not to call the parent's method.

thibsy: All tests pass even if the Group input does never have a name. However I had do adjust the Optional- and Switchable Group to use `Input::withNameFrom()` because it seems they do need it and extended it from Group. But other than this I think the issue should be resolved.

klees:
I'm on this, but I will need a little more time to wrap my head around...

klees:
Please have a look here: https://github.com/srsolutionsag/ILIAS/pull/12

klees:
I have rebased, squashed and pushed this: https://github.com/ILIAS-eLearning/ILIAS/commit/ac937d81705a97c93584d71f10361a691953b7a7

klees:
@thibsy: We (@Amstutz and me) would much appreciate if you would look into tests for the JS some time. Feel free to contact @nhaagen for advice!

thibsy:
Thank you very much! We are planing on an internal workshop to this topic, so I can take another look into this afterwards =).

PR #3810, opened 2021-12-03
[PHP8] ilTree: Add type casts to match return type

tfamula:
Hi @smeyer-ilias 
I added two type casts to ilTree. Otherwise, I get erros because of incorrect return types.
Additionally, the coding style fixer did some work here.


smeyer-ilias:
Thanks for fixing this

PR #3813, opened 2021-12-03
fixed bug #31511: Detail level of block News is not the highest as de…

alex40724:
…cided

https://mantis.ilias.de/view.php?id=31511


Amstutz:
Thx a lot @alex40724 , I checked if we need to cherry pick to 7 and/or trunk, they seem already fixed.

PR #3814, opened 2021-12-06
Fix for mantis #31695

okaufman:
Fixes https://mantis.ilias.de/view.php?id=31695 
by checking if a filename is saved in DB


mbecker-databay:
Awesome, thank you so much!

PR #3816, opened 2021-12-06
UI Components: Interface for Item/Contribution

iszmais:
This PR proposes to introduce the UI-Component for implementing Contribution
This item will initially be used for the new presentation of minimized chat inside the menu but offers multiple opportunities for other contexts with quotable contents.

The Contribution is a visual area wich shows a content created by a user at a specific time or to credit an unknown user if no further informations are given or public.

A Contribution can be used for any single user created content such as reactions, comments or messages.
It occupies a rectangled space where the created content is shown in the middle, the creator as a userpresentation above and the created time as a datetime presentation below. It may also show a leading icon preffixed on the left side of the Contribution and/or a close button in the top right corner of the Contribution. A Contribution may also contain a non visable identifier wich can used for further authetification or verification of the the Contributions authenticity.
Further the goal should be to create a more abstract, reusable component so that this need of seperation can be prevented in future.

Its origin is defined here https://github.com/ILIAS-eLearning/ILIAS/pull/3787


Amstutz:
Hi @iszmais 

Thx for the PR. As far as I can see, there is no further changes in the pipeline here, correct? Feel free to open the PR again, if there is anything further we can/should look at.

Thx for all the work. Note, Review for https://github.com/ILIAS-eLearning/ILIAS/pull/3816 will be published any minute ;-).

PR #3818, opened 2021-12-07
UI Form: 31701, display error-message in form

nhaagen:
A user my add contraints on an entire form; however, there will not be any messages. 
See https://mantis.ilias.de/view.php?id=31701

This will display an (additional) error-message on forms, not only on inputs.

### DIFF
--- a/tests/UI/Component/Input/Container/Form/StandardFormTest.php
+++ b/tests/UI/Component/Input/Container/Form/StandardFormTest.php
@@ -8,6 +8,8 @@ require_once(__DIR__ . "/FormTest.php");
 
 use ILIAS\UI\Implementation\Component\SignalGenerator;
 use \ILIAS\Data;
+use Psr\Http\Message\ServerRequestInterface;
+use ILIAS\UI\Implementation\Component\Input\Container\Form;
 
 class WithButtonNoUIFactory extends NoUIFactory
 {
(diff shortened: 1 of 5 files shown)


Amstutz:
Hi @nhaagen

Your proposal looks GTM. Thx for adding test, to prevent this issue in future. Special thx for adding the role to the role alert to the output. Note that we have a change on the public interface, meaning, you or @rklees will have to quickly announce this to the next JF.

Thx a lot!
Timon

tests/UI/Component/Input/Container/Form/StandardFormTest.php
```diff
@@ -136,4 +138,127 @@ public function test_render_no_url()
         ');
         $this->assertHTMLEquals($expected, $html);
     }
+
+
+    public function testRenderWithErrorOnFieled()
```
tfamula:
```suggestion
    public function testRenderWithErrorOnField()
```

Just a little typo

matthiaskunkel:
Jour Fixe, 13 DEC 2021 : RIchard notified us about a change in a public UI interface. We highly appreciate this change for ILIAS 7 and trunk.

Amstutz:
@nhaagen thx for contributing. LGTM. @tfamula thx for also looking into this!

@klees all lights green for a merge. feel free to do so or ping me if I should.

PR #3830, opened 2021-12-13
UI/Input/Duration - customize labels for start/end

nhaagen:
There was no way to customize (translate) labels for the datetime inputs used in the duration group.
I'd consider that a bug, actually.

### DIFF
--- a/src/UI/Implementation/Component/Input/Field/Duration.php
+++ b/src/UI/Implementation/Component/Input/Field/Duration.php
@@ -35,7 +35,7 @@ class Duration extends Group implements C\Input\Field\Duration
         ilLanguage $lng,
         Factory $field_factory,
         string $label,
-        string $byline
+        ?string $byline
     ) {
         $inputs = [
             $field_factory->dateTime('start'),
--- a/src/UI/Implementation/Component/Input/Field/Duration.php
+++ b/src/UI/Implementation/Component/Input/Field/Duration.php
@@ -35,11 +35,13 @@ class Duration extends Group implements C\Input\Field\Duration
         ilLanguage $lng,
         Factory $field_factory,
         string $label,
-        string $byline
+        string $byline,
+        string $label_start,
+        string $label_end
     ) {
         $inputs = [
-            $field_factory->dateTime('start'),
-            $field_factory->dateTime('end')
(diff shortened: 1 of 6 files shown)


klees:
Since there is no ticket for this issue, I change the tag to `improvement`. Should not make much of a difference, though.

klees (changes requested):
Hi @nhaagen,

one minor change request: Please use something like `$OTHER_START_LABEL` (and for end label accordingly), to set the labels and include them in the HTML in the test. This would make it a lot easier to check that the test actually checks what it says =)

I pass this on to @Amstutz and ask for his approval to take this to the JF.

Thanks!

Amstutz (approved):
Thx a lot, does this need to be pushed down to 7 as well?

If not, IMO we could fast lane this and merge, as soon as the requested change is implemented. Since this is really only a minor not a controversial and not a breaking change, we can announce this to the JF after merging.

klees:
No need to hurry here. I currently don't see any cause to pull this down to 7. I'll just set the JF label.

matthiaskunkel:
Jour Fixe, 10 JAN 2022 : We highly appreciate this suggestion and accept the PR for trunk.

Amstutz:
Thx a lot for this contribution!

src/UI/Implementation/Component/Input/Field/Duration.php
```diff
@@ -35,11 +38,11 @@ public function __construct(
         ilLanguage $lng,
         Factory $field_factory,
         string $label,
-        string $byline
+        ?string $byline
     ) {
         $inputs = [
-            $field_factory->dateTime('start'),
-            $field_factory->dateTime('end')
+            $field_factory->dateTime(self::DEFAULT_START_LABEL),
```
chfsx, line 44: @Amstutz is it correct, that the fields have a label 'start', 'end' if we do not pass any "custom-label"? without any translation? I once had a bugreport concernig this (still searching, didn't find it yet)

klees: Hmm, no, this certainly isn't. @nhaagen Can you have a look and, in case, provide another PR?

PR #3835, opened 2021-12-15
Do not show TOC if it is disabled

padvincenzo:
Hi, I've added this condition before the method `isTocNecessary()` so that the TOC won't be displayed if it is deactivated from the Menu settings of the module. Also, I think that that method, `isTocNecessary()`, is somehow messed up, I didn't get it.


padvincenzo:
This pull solves the bug on mantis: https://mantis.ilias.de/view.php?id=31325

padvincenzo:
Sorry for messing up, I'll close this and open a new one

PR #3836, opened 2021-12-15
fixing 'comma in filenames' issue:  #31730

utesche:
see issue: https://mantis.ilias.de/view.php?id=31730#c77112

### DIFF
--- a/Modules/File/classes/Implementation/class.ilObjFileImplementationLegacy.php
+++ b/Modules/File/classes/Implementation/class.ilObjFileImplementationLegacy.php
@@ -291,7 +291,7 @@ class ilObjFileImplementationLegacy extends ilObjFileImplementationAbstract impl
 
           // trying to distinguish the next-to-last being a 'last part of the filename'
           // or a 'version information',  based on having a dot included or not
-          if (str_contains($data[$last - 1], "."))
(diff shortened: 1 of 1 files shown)


Modules/File/classes/Implementation/class.ilObjFileImplementationLegacy.php
```diff
@@ -279,6 +279,32 @@ public static function parseInfoParams($entry) : array
...
+        {
+          $last = sizeof($data) - 1;
+          for ($n = 1; $n < $last - 1; $n++)
+          {
+            $data[0] .= "," . $data[$n];
+          }
+
+          // trying to distinguish the next-to-last being a 'last part of the filename'
+          // or a 'version information',  based on having a dot included or not
+          if (str_contains($data[$last - 1], "."))
```
mjansenDatabay: Just a note: This is a PHP 8 function, which does not exist in the PHP 7.4 core. I know, this might work, but just because of `composer` (which define this function if it does not exists).

utesche: check,  nice pitfall for a PHP wannabe like me..  ;-)
changed 'str_contains()' to 'strpos()' now.

chfsx:
Thanks @utesche !

PR #3839, opened 2021-12-16
CI: activate captain hook and fix codestyle before commitook + Windows approved

lscharmer:
As #4930 already adds the git hooks from captainhook this PR just removes the `--dry-run` option from the `captainhook.json` to automatically fix the codestyle before commiting.

As a side note:
The captainhook git hooks are executed with `/usr/bin/env php`. Therefore `$ which php` must point to `php8.0` or `php8.1`. Linting will fail under `php7.4`.

### DIFF
--- a/captainhook.json
+++ b/captainhook.json
@@ -36,7 +36,7 @@
         "conditions": []
       },
       {
-        "action": "libs/composer/vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix --config=./CI/PHP-CS-Fixer/code-format.php_cs --using-cache=no --dry-run -vvv {$STAGED_FILES|of-type:php}",
+        "action": "libs/composer/vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix --config=./CI/PHP-CS-Fixer/code-format.php_cs --using-cache=no -vvv {$STAGED_FILES|of-type:php}",
         "options": [],
         "conditions": [
           {


matthiaskunkel:
Jour Fixe, 10 JAN 2022 : We highly appreciate this suggestion and accept the PR for trunk.

libs/composer/only-in-dev.php
```diff
@@ -0,0 +1,14 @@
+<?php declare(strict_types=1);
+/* Copyright (c) 2020 ILIAS open source, Extended GPL, see docs/LICENSE */
```
mjansenDatabay: @lscharmer Can you please replace the license header (see: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/CI/Rector/ChangeLicenseHeader.php#L15)?

mjansenDatabay:
@lscharmer / @daniwe4 I stumbled upon an issue today (actually I thought about this some weeks ago and today I finally found time to check this): According to our documentation, a **custom** configuration file can be specified when installing `CaptainHook`. This unversioned file in turn includes the `captainhook.json` file provided by ILIAS.

If we force the installation with the specified file (like we do with this PR), the GIT hooks will be overwritten each time `post-autoload-dump` is fired and the custom file is not respected anymore.

lscharmer:
Added `libs/composer/captainhook.php` to check for the captainhook.local.json.

libs/composer/captainhook.php
```diff
@@ -0,0 +1,32 @@
...
+ * @author Lukas Scharmer <lscharmer@databay.de>
+ */
+
+if (php_sapi_name() !== 'cli') {
+    exit(1);
+}
+
+$root = realpath(__DIR__ . '/../..');
+$configFile = $root . '/captainhook.json';
+$localConfigFile = $root . '/captainhook.local.json';
```
mjansenDatabay: The problem is: This file could be located anywhere in the file system :(

lscharmer: Which file: `libs/composer/captainhook.php` or `captainhook.local.json`?

mjansenDatabay: `captainhook.local.json`

lscharmer: So the file should be searched from the root directory?

mjansenDatabay: No. If there is no option to retrieve the current configuration from `CaptainHook` (to check wether it is already installed, e.g. with a custom file), we could not achieve our goal to force the activation.

libs/composer/captainhook.php
```diff
@@ -22,11 +22,23 @@
...
+    'pre-commit',
+    'post-change',
+    'post-merge',
+    'commit-msg',
+    'post-commit',
+    'prepare-commit-msg',
+];
+
+foreach ($hooks as $hook) {
+    if (file_exists("$root/.git/hooks/$hook")) {
```
mjansenDatabay: I suggest to check the file content for `captainhook`.

mjansenDatabay:
My plan is to merge this if the `trunk` is open for ILIAS 9.

mjansenDatabay:
@lscharmer Could you please "fix" the license header :-)?

mjansenDatabay:
Thx @lscharmer I'll merge this if the `trunk` is open for ILIAS 9 changes.

mjansenDatabay:
The installation part is solved with https://github.com/ILIAS-eLearning/ILIAS/pull/4930 . This PR still provides an added value because of the changed hooks.

mjansenDatabay:
@lscharmer Could you please rebase this PR? Because https://github.com/ILIAS-eLearning/ILIAS/pull/4930 is now part of the `trunk`, we don't need the custom scripts anymore. So the only change relevant for the `trunk` integration is the change in the `captainhook.json` file.

mjansenDatabay (changes requested):
@lscharmer Could you please rebase this PR? Because https://github.com/ILIAS-eLearning/ILIAS/pull/4930 is now part of the `trunk`, we don't need the custom scripts anymore. So the only change relevant for the `trunk` integration is the change in the `captainhook.json` file.

mjansenDatabay (approved):

mjansenDatabay:
I added the `JF` label to create awareness for this chang (, again).

matthiaskunkel:
Jour Fixe, 19 SEP 2022 : Michael wants to notify all developers about this change.

chfsx:
Hi @mjansenDatabay  

Best regards!

klees:
Hi @mjansenDatabay

Best regards!

kergomard:
Hi @mjansenDatabay

This here seems to have gotten stuck on the way ;-).

Best regards!
@kergomard

mjansenDatabay:
I withdraw this for the time being.

klees:
This makes me sad, I would have really liked this.

mjansenDatabay:
@klees You closed a similar or even better PR here :-) https://github.com/ILIAS-eLearning/ILIAS/pull/5500

klees:
Sometimes I surprise myself =) I seem to be somehow productive at times...

PR #3841, opened 2021-12-16
Improve error propagation in ilServer (release_6)

nicoroeser:
With the recent changes to ilServer that upgraded Log4j to 2.15 (and later, 2.16) there has been a small oversight: if logging initialization fails, the initialization method returns success. This pull request fixes that.

While at it, let’s make ilServer return a proper exit code to the operating system so that callers can handle it appropriately.

I have verified that the code compiles, but nothing further.

This is the pull request for the `release_6` branch.


smeyer-ilias:
Already merged.

PR #3845, opened 2021-12-16
[FIX] Plugin installation per CLI

thibsy:
Hey @klees 

I think I found a bug when installing plugins per cli `php setup/setup.php install --plugin=<NAME>`, where an illegal offset was trying to be accessed. I fixed it by by expanding the abomination in `initEnvironment()`.

Cheers!


klees:
Hi @thibsy,

thanks for looking into this.

The fix seems to be correct. Could you add the same lines in `ilComponentUpdatePluginObjective`, please?

In general: We will not be able to provide everything that every plugin may pull from the void (aka `$DIC`), so we will need to draw a line at some point. `ilLog` is certainly not beyond that line, though. If there is a service that a plugin would need and that we don't or won't provide via the setup, the plugin can opt out from the CLI via `$supports_cli_setup = false;` in the `plugin.php`.

Best regards!

thibsy:
@klees `ilComponentUpdatePluginObjective` seems to already contain them 👍 .

klees:
I have merged this manually: https://github.com/ILIAS-eLearning/ILIAS/commit/436844076543c00ff67b1fad2e990e67e0c9514d

PR #3846, opened 2021-12-16
fixed #31747: ILIAS Install/Update broken Identifier ilAppEventHandle…

alex40724:
…r is not defined.

https://mantis.ilias.de/view.php?id=31747

There might be nicer ways to do this. But this fixes the setup install procedure.


smeyer-ilias:
Thanks, merged the workaround.

PR #3847, opened 2021-12-16
Setup/Component: add objective to ensure plugin language refresh on cli update command

daniwe4:
This fixes [#31768](https://mantis.ilias.de/view.php?id=31768) by making the plugins update the language files when `update` is called for the plugin on the CLI.


klees:
Cherry-picked to trunk.

Thx @daniwe4!

PR #3848, opened 2021-12-16
UI: Implementation of `HasDynamicInputs` file-input

thibsy:
Hi @klees,

As promised, here a further piece of the d&d feature request (#3712) which covers the implementation of `DynamicInputsAware` for the file-input (`ILIAS\UI\Implementation\Component\Input\Field\File`), and is therefore based on the pull request #3809. 

I also introduced a minimal implementation of a `Hidden` input, which accepts all kinds of values and is not yet published in the Factory. Maybe we could though. The file input uses this newly introduced input as a default template for dynamic inputs in order to store the files IRSS ID. 

The file input will (probably) distinguish itself from further `DynamicInputsAware` inputs, because the file input is a dynamic input per default. That's because the template for dynamic inputs consists of at least a `Hidden` input. Further, as I call them, metadata inputs can still be added by calling `withTemplateForDynamicInputs()`. Metadata inputs will be merged into a `Group` with the default template, whereas the `Hidden` input will be mapped to the file-identifier-key provided by the given `UploadHandler`. This simplifies passing values to the file input because the array doesn't have to be unnecessarily nested.  

Kind regards!


thibsy:
Hi @klees and @Amstutz,

Since #3809 (Introduction of `DynamicInputsAware`) has been merged into trunk I'd like to move forward with this PR. As mentioned above I refactored the file input to now implement `DynamicInputsAware`, which uses by default a newly introduced `Hidden` input. Right now this input is not yet accessible via factory, because it is merely a minimal implementation of it. Though, maybe this already suffices for it to be accepted as an actual input?

One notable change is that `UploadHandler::getFileInfoResult()` has been made public and included in the interface. The reason for this was so the Renderer could retrieve file-data from the IRSS one-by-one. There isn't really an upside nor a downside to this IMO. 

Unfortunately in this PR too, the JS unit tests for `file.js` are not yet implemented, but I plan to do this after our internal workshop =).

Kind regards!

klees:
Hi @thibsy,

could you rebase this to trunk, please? It feels as if there are many changes in here that don't actually belong into this PR.

Thanks!

thibsy:
Hello @klees, I rebased trunk again and properly introduced the hidden input. There was a change in the UploadHandler interface, which resulted in some changes that might not seem related to this PR. 

This PR will be needed for a feature (https://docu.ilias.de/goto_docu_wiki_wpage_6317_1357.html) and should therefore be reviewed before the begin of the ILIAS 8 beta-phase (29.04.22 I think).

Thx, and best regards!

klees:
@thibsy I'm on it...

klees:
Hi @thibsy,

Thanks a lot for contributing this impressive work!

- [x] Disabled File: Will the renderer `File Input Field` deal with deactivated fields? What will happen there?
- [x] Revert: The changes in [`Container\Form\Factory`](src/UI/Implementation/Component/Input/Container/Form/Factory.php) (and some related changes) seem to revert the changes from the previous PR (related to the Name Source). Why is this?

Please consider the following suggestions. You do not need to follow those, but, please indicate shortly why you prefer to do otherwise:
- [x] JS-Tests: Write tests for the [JS powering the file input](https://github.com/ILIAS-eLearning/ILIAS/pull/3848/files?diff=unified&w=1#diff-2d9abcacf08b7e7e7a931e40d78ee734af912ab8af6659fa4b7a570dd349f632R17). @nhaagen will be happy to give hints how to make that happen. 
- [x] Wording: I'm a little put off by this "Aware"-wording. I mean, how is `FileUploadAware` aware of a file input? This just is something that all file uploads share (right?), so we would lose nothing, if we just rename to `FileUpload`. Similar thing for `DynamicInputsAware`. This is just some input that has dynamic inputs, so something like `HasDynamicInputs` would be a lot easier to understand. Wording is super important (to me, at least 🙂 ) when developing software.

- [x] White spaces: Please don't commit [whitespace changes](https://github.com/ILIAS-eLearning/ILIAS/pull/3848/files#diff-f857fabc59021b54422647cb7c8c77b1cb280a7b70024c1df86571f02863b2eeL25) or [other changes](https://github.com/ILIAS-eLearning/ILIAS/pull/3848/files#diff-f3892e7d92b0895c6b153d2f1754b755fc888a4b5199b1d4f8784a7344b81d5fR247) that are not warranted by our coding style. If you are using the (and only the) [official ILIAS/PHPStorm code style config](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/code-style-configs/php-storm.xml), please contact me to find out what's going on. If you use another coding style: don't use it in other peoples code. We will keep on changing things back and forth based on personal preferences or tool preferences... If there is something to be codified (such as formatting of docstrings), feel free to propose it as a change request to our coding style.

Note that we will introduce/change the less variables for paddings and margins. From this point on, please use those in your code, not fixed values in numbers. For the moment, this is oke.

If there's anything we could help you with: Feel free to contact as on short notice, as always.

Best regards!

klees (dismissed):
:point_up:

thibsy:
Hello @klees and thanks for the review!

Here is just a brief feedback:

- [x] Disabled File: The input will be disabled on client-side, where the file.js won't initialize the dropzone and disable the "upload button" instead.
- [x] Revert: Tbh I forgot what the error was, but when multiple forms were rendered on the same page and they shared the same name source, only the first form could be submitted and processed properly, all others had errors with `withInput()` I think. 
- [x] JS-Tests: Unfortunately we don't have capacity to implement JS unit-tests, we'd probably need some funding if it were absolutely necessary. 
- [x] Wording: The wording is just a personal preference, I'll change it in an upcoming commit. 
- [x] Whitespaces: I actually do use this configuration for PHPStorm but I'll check again and try to revert those changes.

Kind regards

klees:
Hi @thibsy,

we need to look into this namesource problem.

Best regards!

thibsy:
More Infos about the NameSource issue as discussed with @klees yesterday: 
Using the same name source across multiple forms on the same page would work, but that also means if the forms are submitted to different actions and therefore different methods, ALL forms must be built again in the same order for the naming to work properly. Because we don't want that unnecessary burden, I now create a new name source each time a form is constructed.

Other than that, I removed some unwanted whitespace changes that occurred due to PHPStorms code style config, where I fixed an Issue for blank lines in PHPDoc comments, see #4236.

I also changed the name from `DynamicInputsAware` to `HasDynamicInputs` and `FileUploadAware` to `FileUpload` and adjusted the names of classes implementing those interfaces. 

Cheers!

src/UI/Implementation/Component/Input/Container/Form/Form.php
```diff
@@ -19,7 +19,6 @@ abstract class Form implements C\Input\Container\Form\Form
 {
     use ComponentHelper;
 
-    protected Input\NameSource $name_source;
```
klees, line 22: This needs to be retained, right?

src/UI/Implementation/Component/Input/Field/File.php
```diff
@@ -2,102 +2,161 @@
...
+ * @author  Thibeau Fuhrer <thf@studer-raimann.ch>
  */
-class File extends Input implements C\Input\Field\File
+class File extends HasDynamicInputsBase implements C\Input\Field\File
 {
-    private array $accepted_mime_types = [];
-    private int $max_file_size;
-    private C\Input\Field\UploadHandler $upload_handler;
+    // ===============================================
+    // BEGIN IMPLEMENTATION OF FileUploadAware
```
klees:
```suggestion
    // BEGIN IMPLEMENTATION OF FileUpload
```

src/UI/Implementation/Component/Input/Field/File.php
```diff
@@ -2,102 +2,161 @@
...
+            $byline
+        );
     }
 
-    protected function getConstraintForRequirement() : ?Constraint
-    {
-        return $this->refinery->to()->string();
-    }
+    // ===============================================
+    // END IMPLEMENTATION OF FileUploadAware
```
klees:
```suggestion
    // END IMPLEMENTATION OF FileUpload
```

src/UI/Implementation/Component/Input/Field/File.php
```diff
@@ -2,102 +2,161 @@
...
+    // ===============================================
 
-    protected function isClientSideValueOk($value) : bool
+    // ===============================================
+    // BEGIN OVERWRITTEN METHODS OF DynamicInputsAware
```
klees:
```suggestion
    // BEGIN OVERWRITTEN METHODS OF HasDynamicInputs
```

src/UI/Implementation/Component/Input/Field/File.php
```diff
@@ -2,102 +2,161 @@
...
+            $clone->dynamic_inputs[$file_id] = $clone->dynamic_input_template->withValue($data);
         }
 
-        return true;
+        return $clone;
     }
 
-    public function getUpdateOnLoadCode() : Closure
+    // ===============================================
+    // END OVERWRITTEN METHODS OF DynamicInputsAware
```
klees:
```suggestion
    // END OVERWRITTEN METHODS OF HasDynamicInputs
```

klees (approved):
Just some minor things I noticed when looking through the stuff again. LGTM in general, thanks a bunch!

src/UI/Implementation/Component/Input/Container/Form/Form.php
thibsy: It's not necessary ATM because the naming already happens in the constructor, the property was never used.

alex40724:
@thibsy Thanks for this contribution. Just one note: You are using jQuery, which should be prevented according to our guidelines. Most of it should be achievable by using vanilla JS, see 
https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js/third-party-libs-npm.md#use-of-jquery
and
https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js/js-coding-style.md

matthiaskunkel:
Jour Fixe, 04 APR 2022 : We highly appreciate this PR and accept it for trunk. Please merge.

klees:
Hi @thibsy,

I rebased this onto `trunk` and squashed the various commits. Could you please look into https://github.com/klees/ILIAS/commit/55b08a4a500de6aada589680510b95e6915ed6d7 and check if everything is where it is supposed to be?

I'll then move that commit to `trunk`.

Thanks!

thibsy:
Thanks @klees!

I cherry-picked your commit and checked my example, looks all good to me.
You can move the commit into trunk.

Cheers =).

klees:
- [x] done: https://github.com/ILIAS-eLearning/ILIAS/commit/5a4c6385d830ad79130d6e29ca41ebd95917754f

PR #3849, opened 2021-12-16
MediaObject: remove duplicate file and folder

klees:
Hi @alex40724,

I just found this duplicate, the file is identical to the one in the subfolder `SubTitles` (with capital T).

Best regards!


alex40724:
Thanks, this is fixed in the meantime.

PR #3850, opened 2021-12-16
fix contraint for numeric inputs

iszmais:
The constraint to check the given request value from numeric input checks falsefully if those are int.
It should check if those are numerics (including numeric strings).

### DIFF
--- a/src/UI/Implementation/Component/Input/Field/Numeric.php
+++ b/src/UI/Implementation/Component/Input/Field/Numeric.php
@@ -53,7 +53,7 @@ class Numeric extends Input implements C\Input\Field\Numeric
      */
     protected function getConstraintForRequirement() : ?Constraint
     {
-        return $this->refinery->to()->int();
+        return $this->refinery->numeric()->isNumeric();
     }
 
     /**


klees (changes requested):
Hi @iszmais,

thanks a lot for taking care. There currently is no test covering this, right? Would you be so kind and add a regression test for the problem?

Best regards!

iszmais:
Hi @klees thx for the feedback.

Unfortunatly I wont be able to provide test for that this year.

Amstutz:
would be highly appreciated, still need the test though. Can we assist in some way with this?

iszmais:
Hi @Amstutz ,

with not more or less than time :D
I guess i will be able to look into it mid february. feel free to take my solution attempt for an own PR if you want that ASAP.

Greetings,
@iszmais

Amstutz:
Hi @iszmais 

Thx again. I just pushed your change along with a test, see: https://github.com/ILIAS-eLearning/ILIAS/commit/daa3963e1d78b1561db615f199a6b702c7eb4c8c

Note to however might be interested: We implicitly did cast float to int if no "required" was set. When required, this failed in the required constraint. Now it passes, so its more inline with the default behavior.

PR #3851, opened 2021-12-17
PHP 8 Language Service

katringross:
Must package
excluding unit tests


mjansenDatabay:
@katringross Hmmmm, this PR had no visible changes (see: https://github.com/ILIAS-eLearning/ILIAS/pull/3851/files). This is always an indicator that the PR branch should be rebased of merged with the latest changes of the target branch, to make those changes visibile.

Since this has been merged, the unit tests in `trunk` fail.

And, you used `PHP 8 only` syntax. We MUST support `PHP 7.4`:

- https://github.com/ILIAS-eLearning/ILIAS/commit/7a723e83bcc776a30c2657db28d00add9b8dc611#diff-6fefeab7c72a7323d4e1126e0bdb95125838bd78aadd4e7d513d06f3f188ca65R327
-  https://github.com/ILIAS-eLearning/ILIAS/commit/7a723e83bcc776a30c2657db28d00add9b8dc611#diff-6206446334c56b71a4e6cc09591fb2a64736207e8697cd6f40ec1f9bf2432e42R432

Furthermore https://github.com/ILIAS-eLearning/ILIAS/commit/7a723e83bcc776a30c2657db28d00add9b8dc611#diff-998b47a2bd87cdbc9bd11e0687e2147eb9ec709a2349fc71b4a2948af814cf50R408 seems to break the trunk in `PHP 7.4`.

katringross:
hey @mjansenDatabay, I contacted you on Skype. Maybe you can help me a bit. This PR should merge the php 8 refactoring for the language service into trunk. has not worked out...

iszmais:
Further to @mjansenDatabay commented issues this also seems to have some major semantic problems since most variables dont seem to be that specific as documented in earlier versions.
(So its not always possible to make `protected string $var;` out of `/** @var string */`)

This happens to be a problem in class.ilLanguage.php:69 as only 1 of many conflicts i would assume.

I would suggest to revert the changes of the initial commit (wich seem not to be shown here weirdly) and ensure their semantic correctness and maybe consult an external review for that purpose.

PR #3854, opened 2021-12-17
UI Components legacy, ilSubmitButton fix

Amstutz:
Just a minor fix, to prevent ILIAS8 to crash on some few screens featuring ilSubmitButtons.


alex40724:
Thx, this made it already to the trunk in other ways...

PR #3857, opened 2021-12-21
UI: Refactored File Dropzones

thibsy:
Hi @klees 

As promised, here the last piece of the d&d feature request (#3712) which covers the refactoring of the File dropzones, that use the improved `DynamicInputsAware` file input. This PR is therefore based on #3848.

The PR mainly changes how the dropzones are rendered, because they can use functionalities of the file input now and share an interface called `FileUploadAware`.

Cheers!


Modules/File/classes/class.ilObjFileUploadHandler.php
```diff
@@ -0,0 +1,142 @@
...
+ * @author Thibeau Fuhrer <thf@studer-raimann.ch>
+ *
+ * @ilCtrl_isCalledBy ilObjFileUploadHandler : ilObjFileGUI
+ */
+class ilObjFileUploadHandler extends AbstractCtrlAwareUploadHandler
+{
+    /**
+     * @var \ILIAS\ResourceStorage\Services
+     */
+    private $storage;
```
chfsx:
```suggestion
    private \ILIAS\ResourceStorage\Services $storage;
```

Modules/File/classes/class.ilObjFileUploadHandler.php
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * @var ilObjFileStakeholder
+     */
+    private $stakeholder;
```
chfsx:
```suggestion
    private ilObjFileStakeholder  $stakeholder;
```

src/UI/Component/Dropzone/File/Factory.php
```diff
@@ -1,18 +1,17 @@
-<?php declare(strict_types=1);
```
chfsx, line 1: was this intended to remove?

src/UI/Component/Dropzone/File/Wrapper.php
```diff
@@ -1,39 +1,19 @@
-<?php declare(strict_types=1);
+<?php
```
chfsx: was it intended to remove this?

chfsx (dismissed):
@thibsy just some early catches from my side, didn't had the time to review the whole request but will have a look at it later again.

src/UI/Component/Dropzone/File/Wrapper.php
thibsy: Yes I think I did, because strict_types does nothing in interfaces.

thibsy:
Hi @klees, I rebased trunk and adopted changes from #3848. Once the other PR is merged I'll rebase again to show the actual changes of this PR.

This PR, too, will be needed for a feature (https://docu.ilias.de/goto_docu_wiki_wpage_6317_1357.html) and should therefore be reviewed before the begin of the ILIAS 8 beta-phase (29.04.22 I think).

Thx, and best regards!

klees:
@thibsy: Could you rebase this onto trunk? Is this ready for review afterwards?

thibsy:
Hello @klees, I've rebased as best I could, there should only be changes concerning the dropzones now.

The PR would be ready to be reviewed then =).

Thanks!

Amstutz (changes requested):
Hi @thibsy!

Nice to see this huge effort come to an end. And also nice to see, how the groundwork leads to removed code in this PR...

- [ ] ID/DI? Could you elaborate [this](https://github.com/ILIAS-eLearning/ILIAS/pull/3857/files?diff=split&w=0#diff-5cc978ff4520ee01dd94dc4688717cbcbab81abb8a1f4d4ce5c02a66ed09f0a8R100) a little?
- [ ] `File::$form`: How will `File` ever have a [`$form`-member](https://github.com/ILIAS-eLearning/ILIAS/pull/3857/files?diff=split&w=0#diff-8062c00b69c4b472783ef0caf3e1ef0ca91774051835002f70ac2d03529bbb61R35). Am I missing something, or can we just remove this (and the according code in the getter)?

Change or shortly indicate, why not:
- [ ] Less Variables: Thx for using them! Less overall seems good. Only some exceptions with margins and paddings, where you used fixed values. Check if some padding or newly introduced padding variables could do the trick (e.g. ui-dropzone or ui-dropzone-message ...). 

- [ ] strict_types: Please add `declare(strict_types=1);` in all files that can use them, even if they might do nothing for interfaces. I have no doubt you are correct there, but I think it will be simpler to just demand to "add declare strict_types" than "add declare strict_types in interfaces" because we will need to explain the point again and again in the latter case.
- [ ] constants: The [constants in `File`](https://github.com/ILIAS-eLearning/ILIAS/pull/3857/files?diff=split&w=0#diff-8062c00b69c4b472783ef0caf3e1ef0ca91774051835002f70ac2d03529bbb61R27) can be private IMO.
- [ ] `File::$form`: I first wondered: How will `File` ever have a [`$form`-member](https://github.com/ILIAS-eLearning/ILIAS/pull/3857/files?diff=split&w=0#diff-8062c00b69c4b472783ef0caf3e1ef0ca91774051835002f70ac2d03529bbb61R35). Am I missing something? Then I found [this clone method](https://github.com/ILIAS-eLearning/ILIAS/pull/3857/files?diff=split&w=0#diff-8062c00b69c4b472783ef0caf3e1ef0ca91774051835002f70ac2d03529bbb61R96). I think this implementation will make `FileUpload::withMaxFileSize` and friends useless on clones `withRequest`. I think this is odd... IMO, the solution would be to remove the methods on `File` that are basically copied from `Form` (`getData`, `withRequest`, `withAdditionalTransformation`). This would allow to always create fresh forms on `getForm` and the consumer could perform the `Form`-stuff on the freshly created `Form` objects.
- [ ] Default renderer: I know you already found the `Renderer` like this, but: Please pass the [`$default_renderer`](https://github.com/ILIAS-eLearning/ILIAS/pull/3857/files?diff=split&w=0#diff-5cc978ff4520ee01dd94dc4688717cbcbab81abb8a1f4d4ce5c02a66ed09f0a8R22) to the various method explicitly instead of storing it as some kind of state.

There are some minor changes in interfaces, that we will want to announce at the JF. Please add the according label and present the changes. I would agree to merge this before, to speed up your development.

Best regards!
@klees and @Amstutz

thibsy:
Hello @Amstutz and thanks for the review!

I've committed a few changes and rebased once again onto trunk. I'll elaborate quickly:

- [x] ID/DI: I think I meant dependency injection with DI, but that would not be entirely correct I assume. What I meant there was, that it would be nice if we could pass the JS-ID (applied when rendering) of the newly used file-input as well. But since the rendering of the input happens in a different renderer it would be hacky to retrieve that ID. Thats also the reason why [this](https://github.com/ILIAS-eLearning/ILIAS/pull/3857/files#diff-77af12ee6c21d5cf66dd814c033f37514d9b5ca665802edf4f33a0a168045ddbR61) is necessary.
- [ ] `File::$form`: I think it might not have been entirely obvious, but a dropzone in this state basically IS a form. I changed the `File` interface to extend `Form` now which makes it more clearly. `File::getForm()` should IMO not be used by the consumer (that's why I didn't include it in the interface). The method cannot always return a fresh form, because it must be rendered by the dropzone-renderer within the modal content. Due to `withRequest()` there might be data or messages that must be displayed if a submission failed for example. The member-variable is therefore necessary, because the `Form` interface methods should all be applied to the same instance. 
- [x] Less variables: I've found some variables that did the trick.
- [x] strict_types: I unterstand the reasoning, I added the missing declarations.
- [x] constants: The `FILE_INPUT_KEY` constant is used within tests, the other one I protected.
- [ ] `File::$form`: It's correct that once `withRequest()` or any other `Form` method is called, `FileUpload` methods become useless. But I only see a problem with `withAdditionalTransformation`, since that method might be called before the ones from `FileUpload`. All other form-methods will most likely be called after `withRequest` and the form (or dropzone) should not be changed afterwards anyways. 
- [x] Default renderer: Implemented the change.

I leave those `File::$form` bullets open, since I'm open for feedback about that "order of operations" problem with `Form` methods.

Kind regards!

thibsy:
As discussed with @klees I now store `Form` information in their own member variables (just like with `FileUpload`) in order to always create a fresh form with `getForm()`. This solves the Issue I've been having with the order of operations, now you can e.g. call `withAdditionalTransformation` before calling `withMaxFiles` and `getForm` will return still return a form where the file-input has X amount of max-files.

Do you guys think the PR would be ready for merge then?

Kind regards =).

klees:
@thibsy I'll look into this today.

src/UI/Implementation/Component/Input/Field/Renderer.php
```diff
@@ -899,7 +899,7 @@ protected function initClientsideRenderer(
     protected function replaceTemplateIds(string $template_html) : string
     {
         // regex matches anything between 'id="' and '"', hence the js_id.
-        preg_match_all('/(?<=id=")(.*?)(?=\s*")/', $template_html, $matches);
+        preg_match_all('/(?<=id=")(.*?)(<?>=\s*")/', $template_html, $matches);
```
klees, line 902: I'll remove this on merge... =)

klees:
Merged as: https://github.com/ILIAS-eLearning/ILIAS/commit/62ae601531688e90b296c935bf16370cdda7f1ff

PR #3858, opened 2021-12-22
Set return-type to what was mentioned in the original comment.

pascalseeland:
The original comment states that this function returns a bool. This PR makes setup work again.

### DIFF
--- a/Services/Language/classes/Setup/class.ilSetupLanguage.php
+++ b/Services/Language/classes/Setup/class.ilSetupLanguage.php
@@ -46,7 +46,6 @@
  */
 class ilSetupLanguage extends ilLanguage
 {
-
     public array $text;
     public string $lang_default = "en";
     public string $lang_path;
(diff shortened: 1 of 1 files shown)


klees:
It's more contrived than that, I fear. This might return an array of errors:

```php
public function installLanguages(array $a_lang_keys, array $a_local_keys) : array
{
    // [...]

    $err_lang = array();

    // [...]

    return ($err_lang) ? $err_lang : true;
}
```

pascalseeland:
Thanks for pointing it out. After having a second look at the code( it is as far as I can tell only called in ilLanguagesUpdatedObjective and ilLanguagesInstalledAndUpdatedObjective), I think there is no need to return anything from this function and in case of an error throw a RuntimeException. Even if there would be a failure, the Objective would still be achieved, as it never tested, if it failed.

I can provide an updated PR for this, if we agree on this way to solve the issue.

klees:
> I can provide an updated PR for this, if we agree on this way to solve the issue.

This would be up to the maintainer of the service (@katringross) to decide. To me this sounds fine.

Best regards!

katringross:
Hi @pascalseeland and @klees,

Thanks for following up this issue! @pascalseeland for me your suggestion makes sense too. Would be great If you provide a PR. 

Thanks Katrin

katringross:
@pascalseeland und @klees for fixing https://mantis.ilias.de/view.php?id=31763#c77166 I merged this PR. As second step we will remove the return for the function.

PR #3860, opened 2021-12-23
UI Components, Unit Tests, add some helper Class for Dependency Injection

Amstutz:
While working on the sustainability classes for PHP8, I got rid of some legacy GUI components by replacing them with UI Components. The refactored classes got much easier to test, which is done. However, I observed, that even though it is not that much work to mock the deps to the framework, it still work that generates code duplicates. 

It would be helpful, to have some helper at hand, which creates a working UI Factory and Renderer to be injected in the classes to be tested without any hassle. This is what I hereby propose.


klees:
Thx @Amstutz, definitely helpful!

PR #3862, opened 2021-12-28
add language default to php.ini (#31759)

katringross:


klees (changes requested):
Hi @katringross,

I would really appreciate if you made the following changes:

* rebase on trunk to get rid of the unnecessary merge commits
* add the `=` on the last line, which you removed
* make this PR include only one commit by amending that `=` to your initial commit

Regarding the change in general: I quickly checked if the master ini ever contained that field: it didn't. I then looked into the code in `ilLanguage`. The setting from the ini is only ever used if there is no `$ilSetting` object present in the dependency injection container and if that object does not contain a `language`-field. Which, afaik, is virtually never.

On the flip side, if we would keep this change, every existing ILIAS installation would need to add an according field in the `ilias.ini.php` to keep running. Which some people would miss inevitably.

I would thus prefer to change the behaviour of `ilLanguage` instead of adding a field in the master ini which barely has any effect. It's up to you as language maintainer to decide, though.

Best regards!

katringross:
@klees instead I merged the PR #3870 from @chfsx. So I will close this PR.

PR #3863, opened 2021-12-28
Services/Cron: refresh job data for plugin jobs as well (#18411)

klees:
Hi @mjansenDatabay,

for some (unknown, to me) reason, the fix for [#18411](https://mantis.ilias.de/view.php?id=18411) has not been extended to cronjobs of plugins. This changes this situation.

Best regards!


mjansenDatabay:
Thx @klees

PR #3871, opened 2022-01-03
Fix type of maxlength from bool to int

pascalseeland:
The type should be int. Otherwise the maxlength will always be 1.


alex40724:
Thanks

PR #3873, opened 2022-01-05
[FIX] 0031812: Fix color of selected area in imagemap questions

corro:
Fixes https://mantis.ilias.de/view.php?id=31812


dsstrassner:
Hmm, right now, I'm not convinced that this is the right way to approach this. getConvertCmd is marked as deprecated. I consult my code consultants on this, and then we provide feedback. Another approach is maybe shown in [#31016](https://mantis.ilias.de/view.php?id=31016#c76918)

corro:
Fixed with #4016.

PR #3875, opened 2022-01-05
Fix wrong namespace in Refinery Tests

lscharmer:
Fix wrong namespace for Refinery Tests


mjansenDatabay (approved):
Thanks a lot @lscharmer for fixing the PSR-4 issues.

PR #3876, opened 2022-01-05
BT-31667: Added missing descriptor of hidden breadcrumb headline

M-Barz:
Link to PR:
https://mantis.ilias.de/view.php?id=31667

I believe that adding content to the invisible headline is the least invasive way of fixing this accessibility issue and also helps describing the breadcrumb navigation below if it gets read by a screenreader.


smeyer-ilias (approved):
Thanks, please merge.

PR #3881, opened 2022-01-06
BT-30506: Added aria label to tag input

M-Barz:
Link to Mantis-Report:
https://mantis.ilias.de/view.php?id=30506


alex40724:
Thanks!

PR #3882, opened 2022-01-06
Block: Escape special characters when adopting block title to aria la…

mjansenDatabay:
…bel of actions dropdown

Mantis Issue: https://mantis.ilias.de/view.php?id=31832


alex40724:
Original bug has been fixed differently, but I take this one, too. Thanks.

PR #3883, opened 2022-01-06
BT-31110: Fix show filesize in filelist

rsheer:
A fix for mantis issue https://mantis.ilias.de/view.php?id=31110


alex40724:
Great, thanks for the PR!

PR #3885, opened 2022-01-10
GlobalCache/APC: Fix iterating over invalid iterator state

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=31725


chfsx:
Thanks a lot!

PR #3888, opened 2022-01-11
UI/Input/Tag: update tagify-version

nhaagen:
This updates tagify to 4.9.4 or higher; there seems to be a bug in 4.8.1 that caused faulty rendering of existing tag-values.
It does not occur in 4.9.4, though.


Amstutz:
Thx a lot.

PR #3889, opened 2022-01-11
PHP(: Service/HTTPS

chfsx:
Hi @mjansenDatabay and @smeyer-ilias 

I refactored Service/HTTPS for PHP8 and made several changes:
- made ilHTTPS only accessible via DIC and removed the gestInstance
- License Header
- Typing
- Removal of $_GET/$_POST

maybe it's not 100% but it's quite a good state. There is no real maintainer for Service/HTTPS, you both have classes there and we should thin about moving those things to other locations I think.

Feel free to merge at any time


mjansenDatabay (approved):
LGTM, thanks @chfsx I created a PR with small improvements: https://github.com/srsolutionsag/ILIAS/pull/11/files

PR #3890, opened 2022-01-11
Implemented copyright checker for new and modified files

catlauraherzog:
I implemented a check if the copyright lines are at a correct place and contain the exact wording. For every modified and new file the lines are defined as:

```
/**
 * published by ILIAS open source e-Learning e.V.
 *
 * ILIAS is licensed with the GPL-3.0,
 * see https://www.gnu.org/licenses/gpl-3.0.en.html
 * source code, too.
 *
 * us at:
 * https://www.ilias.de
 */
```

They need to start at line 2 or 3.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -12,6 +12,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
@@ -33,10 +35,19 @@ jobs:
         run: CI/PHP-CS-Fixer/run_check.sh
         env:
           GHRUN: "yes"
-          PR_NUMBER: ${{ github.event.pull_request.number }}
+          PR_NUMBER: ${{ github.event.number }}
+          GH_SHA: ${{ github.sha }}
 
       - name: Special Char Checker
         run: CI/Special-Char-Checker/special-char-checker.sh
         env:
           GHRUN: "yes"
-          PR_NUMBER: ${{ github.event.pull_request.number }}
+          PR_NUMBER: ${{ github.event.number }}
+          GH_SHA: ${{ github.sha }}
+
+      - name: Copyright Check
+        run: CI/Copyright-Checker/copyright-checker.sh
+        env:
+          GHRUN: "yes"
+          PR_NUMBER: ${{ github.event.number }}
+          GH_SHA: ${{ github.sha }}
--- a/CI/Copyright-Checker/copyright-checker.sh
+++ b/CI/Copyright-Checker/copyright-checker.sh
@@ -1,0 +1,152 @@
+#!/bin/bash
+
+# Copyright Checker
+#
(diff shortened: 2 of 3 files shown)


chfsx:
Hi @lauraherzog 
I tried the checker in the following setting:
- used your branch lauraherzog:copyright
- adopted the script a bit to just scan Service/ActiveRecord (which has no license header in the state of your branch)
- added the license header to Service/ActiveRecord-Files with `./libs/composer/vendor/bin/rector process --clear-cache --xdebug --no-diffs --config ./CI/Rector/rector.php Services/ActiveRecord`
- run the (modified) checker

as far I can see all files of `Services/ActiveRecord` get listed as `Copyright not as expected in ...`. Maybe I'm doint something wrong :-)

chfsx (changes requested):
see above

catlauraherzog:
@chfsx I don't think you are doing anything wrong. I just think that my script is not bulletproof so we need to find the error :)

I see that you modified Services/ActiveRecord with this Rectorthing (which I don't know). Could you give me an example of two or three modified files (just the header with the copyright statement)?

Best, Laura

chfsx:
Hi @lauraherzog 
I still fail to run the check for files, wich have a license-header introduces by rector. E.G. Services/ActiveRecord has automatically added License-header (currently in trunk without whitespaces before the two https://... like in your check). could not find a difference between the header you check against and the headers in the files... I assume the check is too hard and should ignore some things like whitespaces. 

i changes the line 20 in copyright-checker.sh to just check Services/ActiveRecord like this:
`CHANGED_FILES=$(find Services/ActiveRecord -path ./libs -prune -o -type f -name '*.php')`

catlauraherzog:
Hey @chfsx 

I reworked the checkmodel by removing whitespace, stars and slashes. My tests worked with the copyright notice from above. Additionally I tested with notices where I added stars and slashes. So the test is a bit weaker.

Does that satisfy our needs?

chfsx:
we are almost there :-) I made a branch with your checker and changes a lot of files with the license header. the action runs und succeeds, but have a look at the step "Copyright Check" here
https://github.com/srsolutionsag/ILIAS/runs/5157170888?check_suite_focus=true

it doesn't seem to run: 
```

Run CI/Copyright-Checker/copyright-checker.sh
jq: error (at <stdin>:4): Cannot index string with string "filename"
Scanning changed files for copyright notice ...
Scan complete. Found 0 incidents.
```
no Idea where this comes from, maybe just because it's not yet part of the repo?

klees:
@lauraherzog @chfsx What's the status here?

chfsx:
@klees from my point of view the same as in [11 Feb](https://github.com/ILIAS-eLearning/ILIAS/pull/3890#issuecomment-1036252947), but did not check it again.

catlauraherzog:
> @lauraherzog @chfsx What's the status here?

@klees @chfsx Fixed the implementation. Could you verify pls?

chfsx (approved):
Hi @lauraherzog, thanks a lot for the implementation, seems to work as expected now!

chfsx:
Thank you very much @lauraherzog !

chfsx:
Ok the new check is quite hard :) I have my first failed test now

PR #3893, opened 2022-01-11
DataCollection: Fix file upload for new entry with enabled confirmati…

mjansenDatabay:
…on (Mantis 31829)

Mantis Issue: https://mantis.ilias.de/view.php?id=31829


theodortruffer:
@mjansenDatabay Thanks a lot!

PR #3894, opened 2022-01-11
StudyProgramme: fix whoops for adding zero users

nhaagen:
When a user adds participants to the PRG via "search users", but does not select any and hits submit, an exception is thrown.


klees:
Picked to trunk.

PR #3896, opened 2022-01-12
A11Y/UI/Tree: Fix duplicate role attribute for tree node list items (Manti…

mjansenDatabay:
…s 31826)

Mantis Issue: https://mantis.ilias.de/view.php?id=31826


Amstutz:
Thx a lot!

PR #3897, opened 2022-01-12
This allows for correct naming of files from php://input streams.

kergomard:
Right now if you hand a php://input wrapped in a stream to the IRSS the filename will be resolved to "input". This will in this case set it to the given name, same as for memory. Sabre/DAV hands over the data as a handle on php://input.


chfsx:
Thanks @swiniker ! I'll cherry-pick to release_7 as well

PR #3898, opened 2022-01-13
TermsOfService: wrong lang var name ('...no_consent_yet')

utesche:
The language var name 'withdrawal_sure_account_no_consent_yet' was falsely typed as 'withdrawal_sure_account_no_consent_ye' (without the trailing 't' in all language files.

The var is being referenced from PHP code with this suffix:
https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/Services/TermsOfService/classes/class.ilTermsOfServiceWithdrawalGUIHelper.php#L175


mjansenDatabay:
Thx @utesche , I'll pick this into `trunk`.

PR #3899, opened 2022-01-13
ilTree: Add type cast

tfamula:
Without this cast, I get a type error (string instead of int) when using `deleteTree()`


smeyer-ilias (approved):
Sorry, saw this to late.
Thanks for fixing.

PR #3900, opened 2022-01-13
BT-30502/BT-30498: Assignments: Added alt text to magnifier image, made image focusable and interactable by tabbing

M-Barz:
Link to Mantis Reports:
https://mantis.ilias.de/view.php?id=30502
https://mantis.ilias.de/view.php?id=30498


alex40724:
Thanks!

PR #3901, opened 2022-01-13
Cleanup of Services/Component

klees:
Hi everyone,

I made some more progress in cleaning up Services/Component. Besides some minor removals, I mostly moved code to locations that are more appropriate imo:

* I introduced an `ilPluginLanguage` in `Services/Language` that takes care about handling language variables. The class is now called from `ilPlugin` instead of having code that is language related almost exclusively in `ilPlugin`.
* I move `_getImagePath` from `ilPlugin` to `ilRepositoryObjectPlugin`, since the method is only used in the context of Repository Plugins.
* `ilClaimingPermissionHelper` is now located in Services/AccessControl, as it contains functionality more related to access control instead of plugins.

Also, i removed some stuff that is not required anymore:

* `ilPlugin::includeClass` and code that depends on it, since plugin classes are part of the classmap as of ILIAS 7. In this process, I removed methods that depended on that code.
* the tables `service_class` and `module_class` and code that fills these, since UICore (that was the exclusive user of that tables) is now artifact based as well.

I consider these changes to be of general interest, but I specifically like to get approval from the following maintainers:

* @katringross for the introduction of `ilPluginLanguage` in `Services/Language`
* @alex40724 for the move of the `_getImagePath` to `Services/Repository`
* @smeyer-ilias for the move of `ilClaimingPermissionHelper` to `Services/AccessControl`
* @thibsy because I used his UICore-DB-updates to finally remove the tables `service_class` and `module_class`. Also: we really don't need these anymore, right?

If there are any questions or concerns, feel free to ask or voice, as always.

Also, this fixes [#31905](https://mantis.ilias.de/view.php?id=31905).

Thanks for your cooperation!

PR #3902, opened 2022-01-14
Update PHP-CS-Fixer

klees:
Hi everyone,

I only wanted to changes stuff in [Services/Component](https://github.com/ILIAS-eLearning/ILIAS/pull/3901), turns out that there is some yak shaving hidden there.

[PR 3901](https://github.com/ILIAS-eLearning/ILIAS/pull/3901) fails, because PHP-CS-Fixer does not support PHP 8.1. An update of PHP-CS-Fixer required a major update of `symfony/console`, which in turn required an update of the dependencies of simplesamlphp.

This is all contained in here.

@mjansenDatabay Could you have a look here and merge this if ok?

Thank!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,7 +70,7 @@
 		"mikey179/vfsstream": "^1.6",
 		"phpunit/phpunit": "^9.5",
 		"mockery/mockery": "^1.4",
-		"friendsofphp/php-cs-fixer": "^2.15",
+		"friendsofphp/php-cs-fixer": "^3.5",
 		"sebastian/environment": "^5.0",
 		"captainhook/captainhook": "^5.3",
 		"rector/rector": "^0.11.7"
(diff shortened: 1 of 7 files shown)


klees:
The yak seems to be naked now.

mjansenDatabay:
Hmmm, I wonder why the builds pass in `trunk`, but they fail for PRs.

klees:
image

this would be the explanation...

composer.json
```diff
@@ -5,7 +5,11 @@
 	"authors": [],
 	"config": {
 		"optimize-autoloader": true,
-		"vendor-dir": "./libs/composer/vendor"
+		"vendor-dir": "./libs/composer/vendor",
+		"allow-plugins": {
+			"simplesamlphp/composer-module-installer": true,
```
mjansenDatabay, line 10: I like! +1 for the usage of additional security layers.

mjansenDatabay (approved):

PR #3903, opened 2022-01-14
Add Unit Tests for Services/Mail

lscharmer:
Add Unit Tests for ilMail.

Remove unnecessary $usrId Parameter from `ilMail::getNewDraftId(...)` and adjust the only usage of this method (in `ilMailFormGUI`).

### DIFF
--- a/Services/Mail/classes/class.ilMail.php
+++ b/Services/Mail/classes/class.ilMail.php
@@ -400,14 +400,14 @@ class ilMail
         return $row;
     }
 
-    public function getNewDraftId(int $usrId, int $folderId) : int
+    public function getNewDraftId(int $folderId) : int
     {
         $nextId = $this->db->nextId($this->table_mail);
         $this->db->insert($this->table_mail, [
             'mail_id' => ['integer', $nextId],
-            'user_id' => ['integer', $usrId],
+            'user_id' => ['integer', $this->user_id],
             'folder_id' => ['integer', $folderId],
-            'sender_id' => ['integer', $usrId],
+            'sender_id' => ['integer', $this->user_id],
         ]);
 
         return $nextId;
--- a/Services/Mail/test/ilMailTest.php
+++ b/Services/Mail/test/ilMailTest.php
@@ -3,6 +3,7 @@
 /* Copyright (c) 1998-2021 ILIAS open source, Extended GPL, see docs/LICENSE */
 
 use ILIAS\Refinery\Factory;
+use \PHPUnit\Framework\MockObject\MockObject;
 
 /**
  * Class ilMailMimeTest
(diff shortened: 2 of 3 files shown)


Services/Mail/test/ilMailTest.php
```diff
@@ -177,4 +191,369 @@ static function (string $login) use ($loginToIdMap) : int {
...
+                'attachments' => '',
+                'folder_id' => '',
+                'mail_id' => '',
+                'sender_id' => '',
+                'tpl_ctx_params' => '[]',
+                'use_placeholders' => '',
+                'user_id' => ''
+            ]],
+            [[
+                'allowed_classes' => '',
```
mjansenDatabay: `allowed_classes` is only used as an option parameter when unserializing data. Thy did you include this key here?

Services/Mail/test/ilMailTest.php
```diff
@@ -177,4 +191,369 @@ static function (string $login) use ($loginToIdMap) : int {
...
+        $this->mockDatabase->expects(self::once())->method('setLimit')->with(1, 0);
+        $instance->getPreviousMail($mailId);
+    }
+
+    public function provideGetPreviousMail() : array
+    {
+        return [
+            [[]],
+            [[
+                'allowed_classes' => '',
```
mjansenDatabay: `allowed_classes` is only used as an option parameter when unserializing data. Thy did you include this key here?

Services/Mail/classes/class.ilMail.php
```diff
@@ -400,14 +400,14 @@ protected function fetchMailData(?array $row) : ?array
         return $row;
     }
 
-    public function getNewDraftId(int $usrId, int $folderId) : int
+    public function getNewDraftId(int $folderId) : int
```
mjansenDatabay, line 403: Thanks a lot for removing this!

Services/Mail/test/ilMailTest.php
```diff
@@ -177,4 +191,369 @@ static function (string $login) use ($loginToIdMap) : int {
...
+        $instance = $this->create($refId);
+
+        $this->assertEquals($refId, $instance->getMailObjectReferenceId());
+    }
+
+    public function testFormatNamesForOutput() : void
+    {
+        $instance = $this->create();
+
+        $this->assertEquals('', $instance->formatNamesForOutput(''));
```
mjansenDatabay: Hmmm, according to `\ilMail::formatNamesForOutput` a translation of `not_available` should be returned if the input string is empty. I wonder why `''` is expected here (and why this test works).

Services/Mail/test/ilMailTest.php
lscharmer: Oh right, this was auto generated. Will be removed.

Services/Mail/test/ilMailTest.php
lscharmer: Because the language service is mocked but the method `txt` is not configured. So the empty value of the method's return type is returned (in this case it is the empty string).

mjansenDatabay: Hmmm, IMO the test does not make sense in this case :-). I suggest either to test real scenarios oder just remove this test :-).

lscharmer: Is now adjusted

PR #3905, opened 2022-01-14
UI/Layout: Breadcrumbs in Content

nhaagen:
This moves the breadcrumbs into the content-space of the main template.
Results look like this:
Screenshot from 2022-01-14 11-55-30
Screenshot from 2022-01-14 11-56-07

### DIFF
--- a/src/UI/templates/default/Breadcrumbs/breadcrumbs.less
+++ b/src/UI/templates/default/Breadcrumbs/breadcrumbs.less
@@ -12,8 +12,8 @@
 		margin: 0;
 		font-weight: @il-standard-page-breadcrumbs-font-weight;
 		font-size: @font-size-small;
-		padding: @il-padding-small-vertical;
-		padding-left: @il-mainbar-btn-width + @il-standard-page-content-padding;
+		padding: (@il-padding-small-vertical * 2);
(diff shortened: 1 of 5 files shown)


matthiaskunkel:
Does this change will remove the empty breadcrump row from screens where no tree-based content (like in Dashboard or Calendar) is shown as well?

yvseiler:
> Does this change will remove the empty breadcrump row from screens where no tree-based content (like in Dashboard or Calendar) is shown as well?

 With this PR only the position of the breadcrumb is changed for the moment. In case of an empty breadcrumb row, this has the advantage that it no longer takes the prominent place in the header (and there is a "gap" between header and mainbar), however, the empty breadcrumb row remains in the content area.

matthiaskunkel:
Jour Fixe, 07 FEB 2022 : We highly appreciate this suggestion and accept the PR for trunk. It would be great if empty breadcrumbs would be invisible.

Amstutz:
@nhaagen , @klees : is it possible, to only render the breadcrumbs if there are entries? If not, Ill merge as is. If so, adding this along with a unit test making sure, empty breadcrumbs do not render would be great.

nhaagen:
I'll have a look.

PR #3907, opened 2022-01-14
ilObject: Add missing $ilias property.

lscharmer:
Add missing property `$ilias` in `ilObject` (property is set in the constructor).

Fix inconsistent indentation of comments.

### DIFF
--- a/Services/Object/classes/class.ilObject.php
+++ b/Services/Object/classes/class.ilObject.php
@@ -60,17 +60,17 @@ class ilObject
 
 
     /**
-    * lng object
-    * @var		object language
-    * @access	private
-    */
+     * lng object
+     * @var		object language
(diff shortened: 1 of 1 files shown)


klees:
I'm tasked to look into `Services/Object` for the PHP8 update. So I will take a look here as well if no one objects.

klees:
Hi @lscharmer,

thanks a bunch for taking care!! We have [incorporated a similar change](https://github.com/ILIAS-eLearning/ILIAS/blob/553b0fa093b7feb3314084810f00cc2ecc7ebb93/Services/Object/classes/class.ilObject.php#L19) during our efforts regarding `Services/Object` and PHP 8, so I'll close this one here.

Still: Keep up the good work!

Best regards!

PR #3908, opened 2022-01-17
Scorm/CmiXapi: Fix PHP types / Apply PSR-2 + X Code Style

mjansenDatabay:
With the latest changes for `SCORM` and `CmdXapi` several issues have been introduced.

- Mixed types, which are only allowed in PHP 8. We'll have to support PHP 7.4 as well.
- Missing PHP return type declarations, which were defined base classes.
- Wrong PHP type hints, which were NOT defined in base classes.

This PR fixes all these issues. Furthermore I applied the `PSR-2 + X` code style to the changed files.

**Important:** Please always check the code style and run our unit test suite before you push to branches of the official GitHub repository.


alex40724:
@Uwe-Kohnle This is kind of a blocker for me currently. Could you please accept / fix this. At least "composer dump-autoload" should run, otherwise issues may pile up (I am currently committing withouth being able to check if I add additional issues).

chfsx:
@Uwe-Kohnle same here, I'm currently flying blind while developing in trunk, could you please fix the classmap generation,  artifact building and tests (if any) asap? thanks in advance

PR #3914, opened 2022-01-19
UI/Tree: Fix PSR-4 issue is expandable2 example

mjansenDatabay:
This commit fixes a PSR-4 issue reported when the autoloader files are build by composer.
Instead of using a class definition, a factory function is introduced to retrieve anonymous classes (equivalent to the former `DataNode`).

`Class ILIAS\UI\examples\Tree\Expandable\DataNode located in ./src/UI/examples/Tree/Expandable/expandable2.php does not comply with psr-4 autoloading standard. Skipping.`


klees:
Thx @mjansenDatabay!

PR #3915, opened 2022-01-19
User/Registration: Fix determination of access limit timestamp (Manti…

mjansenDatabay:
…s 31756)

Mantis Issue: https://mantis.ilias.de/view.php?id=31756

If approved, this has to be cherry-picked to `trunk` as well.


mjansenDatabay:
@smeyer-ilias Any update on this?

PR #3916, opened 2022-01-19
Data URI Support

thibsy:
Ignored concatenation of content version in case of data URIs


thibsy:
The same issue also occurs in release_6, might we also deploy this fix there?

PR #3920, opened 2022-01-20
PHP8: src/DI

chfsx:
Hi!
I made a first step of refactoring the src/DI. There were some things I had to change:
- Adopting Tests in tests/DI
- Tests of Service/Style made some problems, not absolutely sure why... they had influence to other tests, maybe because of the missing `clone` while backing up global $DIC. I removed some Mockery-things and introduces the backing up of DIC using a clone

### DIFF
--- a/Services/FileDelivery/test/FileDeliveryTypes/FileDeliveryTypeFactoryTest.php
+++ b/Services/FileDelivery/test/FileDeliveryTypes/FileDeliveryTypeFactoryTest.php
@@ -34,7 +34,6 @@ use PHPUnit\Framework\TestCase;
  */
 class FileDeliveryTypeFactoryTest extends TestCase
 {
-    use Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration;
     private \ILIAS\FileDelivery\FileDeliveryTypes\FileDeliveryTypeFactory $subject;
     /**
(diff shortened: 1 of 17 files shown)


klees (approved):
LGTM, thx!

chfsx:
Hi @Amstutz 
I see we have several conflicts now in the system-styles-checks, could we have a skype or discord session to have a look at these things?

Amstutz:
Hi @chfsx ,

The DIC needs mocks (or used to) since other deps directly use mocks. Interesting, that it just now starts to cause issues. Clone should definitely resolve it, since DIC should be written back using tear-down of the unit tests to reset the initial case.

However, having to use global DIC is bad, not matter which we look at it. If possible, I inject everything through constructor. 

However, it might be possible, that I do not use to hijeck into global DIC anymore, since other components are doing proper DIC now as well.

Yeah, discord is possible for me. Sadly, I did catch you few minutes ago. Just ping me, when  ready. 

Note, just checked your changes for none-affected lines, could it be, that you have a different max line length setup, then e.g. I do?

Greets, looking forward to resolve this.

Amstutz:
As discussed via discord with @chfsx , I fixed the System Style tests and merged this change with trunk directly after. Thx @chfsx .

PR #3922, opened 2022-01-20
[FIX] 0031817: ilCtrl: PHP Warning when building artifacts - preg_match_all()

chfsx:
@thibsy as discussed earlier today, this should fix https://mantis.ilias.de/view.php?id=31817


thibsy (approved):
LGTM, thanks!

PR #3923, opened 2022-01-20
#31796: Consideration of ilTable commands.

thibsy:
Hi all,

As reported in https://mantis.ilias.de/view.php?id=31796 `ilCtrl` didn't consider table GUI commands anymore. This PR solves the issue for the command retrieval via `ilCtrl::getCmd()` but does not execute the direct `$_POST` manipulations as before. Instead a new trait has been introduced, that delivers "fake" `$_POST` arrays which contain the encapsulated data for the same keys as before. So for example, instead of `$_POST['select_cmd_all']` you'd do something like `$this->getSelectAllPostArray['select_cmd_all']`, as done in `ilObjUserFolderGUI::getActionUserIds()`.


matthiaskunkel:
Jour Fixe, 24 JAN 2022 : We highly appreciate this PR and accept it for trunk. @thibsy : please send a mail to developers@ to notify all developers about this change.

PR #3924, opened 2022-01-20
#31893 fixed undefined method calls to ilCtrl::initBaseClass().

thibsy:
Closes https://mantis.ilias.de/view.php?id=31893.

### DIFF
--- a/saml.php
+++ b/saml.php
@@ -17,7 +17,10 @@ ilContext::init(ilContext::CONTEXT_SAML);
 require_once 'Services/Init/classes/class.ilInitialisation.php';
 ilInitialisation::initILIAS();
 
-$DIC->ctrl()->initBaseClass('ilStartUpGUI');
-$DIC->ctrl()->setCmd('doSamlAuthentication');
 $DIC->ctrl()->setTargetScript('ilias.php');
-$DIC->ctrl()->callBaseClass();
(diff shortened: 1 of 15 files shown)


saml.php
```diff
@@ -17,7 +17,10 @@
 require_once 'Services/Init/classes/class.ilInitialisation.php';
 ilInitialisation::initILIAS();
 
-$DIC->ctrl()->initBaseClass('ilStartUpGUI');
-$DIC->ctrl()->setCmd('doSamlAuthentication');
 $DIC->ctrl()->setTargetScript('ilias.php');
-$DIC->ctrl()->callBaseClass();
+$DIC->ctrl()->redirectToURL(
```
mjansenDatabay, line 21: @thibsy Are you sure this works? `saml.php` could be requested with some query parameters. Are these preserved with this redirect? Furthermore I think (but am not 100% sure) that there might be cases where this script is requested by HTTP POST. In this case, a redirect will result in a failed SAML flow.

I don't think it is a good idea to remove `callBaseClass` and use a HTTP redirect as an alternative at all. I call this `Redirect Abuse` and IMO this is a code smell we find often in ILIAS ;).

thibsy: Agreed, will change this in a new commit to trunk.

PR #3928, opened 2022-01-20
ilCtrl: Don't use PHP 8.0 functions

mjansenDatabay:
See: https://stitcher.io/blog/new-in-php-8#new-str_contains()-function-rfc and https://stitcher.io/blog/new-in-php-8#new-str_starts_with()-and-str_ends_with()-functions-rfc


thibsy:
Thanks a lot! My IDE didn't yell at me for using these methods and I assumed they are available for 7.4 :-).

mjansenDatabay:
@thibsy AFAIK composer already provides these functions (as manually create functions) if they are not defined, even for PHP 7.4. But IMO we should stay clean and should not rely on functions actually not suppoted in the PHP versions we have to support in ILIAS.

PR #3933, opened 2022-01-21
GlobalScreen: Initialize variables with null do prevent issues when a…

mjansenDatabay:
…ccessing them


mjansenDatabay:
Arghhhh, I reverted the commit, I'll create a new PR.

https://github.com/ILIAS-eLearning/ILIAS/pull/3934

PR #3934, opened 2022-01-21
GlobalScreen: Initialize variables with null do prevent issues when a…

mjansenDatabay:
…ccessing them


chfsx:
makes sense, thanks a lot!

PR #3935, opened 2022-01-21
NPM/TinyMCE: Use correct version/brach of TinyMCE for ILIAS 7

mjansenDatabay:


kergomard:
I would propose to move the patched TinyMCE to the an internal Repo and would propose a PR next week, if you agree.

mjansenDatabay:
Closed because of #3945

PR #3936, opened 2022-01-21
NPM/TinyMCE: TinyMCE PHP 8

mjansenDatabay:


mjansenDatabay:
Closes because of https://github.com/ILIAS-eLearning/ILIAS/pull/3953

PR #3942, opened 2022-01-25
improved submission of unchecked checkboxes.

thibsy:
It's a know issue that unchecked checkboxes aren't submitted and can therefore not be found in `$_POST`. This [solution](https://stackoverflow.com/a/1992745) on stack overflow seemed appropriate to me, so I implemented it for the UI component `Checkbox`. There was really no downside to this until now, but I stumbled upon a problem that was solved by this solution when implementing `DynamicInputsAware` (#3809).


thibsy:
Note that the failed test was already in trunk...

klees:
@thibsy We don't need this anymore since we found a solution for #3809, right?

thibsy:
Yes, that isn't necessary anymore.

PR #3943, opened 2022-01-25
BookingPool: Fix iterating over non array

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=31924

### DIFF
--- a/Modules/BookingManager/Objects/classes/class.ilBookingObjectsTableGUI.php
+++ b/Modules/BookingManager/Objects/classes/class.ilBookingObjectsTableGUI.php
@@ -317,18 +317,20 @@ class ilBookingObjectsTableGUI extends ilTable2GUI
         
         if (!$this->has_schedule) {
             $cnt = 0;
-            foreach ($this->reservations[$a_set["booking_object_id"]] as $item) {
-                if ($item["status"] != ilBookingReservation::STATUS_CANCELLED) {
-                    $cnt++;
-                
-                    if ($item["user_id"] == $ilUser->getId()) {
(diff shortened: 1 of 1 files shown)


Modules/BookingManager/Objects/classes/class.ilBookingObjectsTableGUI.php
```diff
@@ -342,10 +344,11 @@ protected function fillRow($a_set)
                 $assign_possible = false;
             }
         } elseif (!$this->may_edit) {
-            foreach ($this->reservations[$a_set["booking_object_id"]] as $item) {
-                if ($item["status"] != ilBookingReservation::STATUS_CANCELLED &&
```
alex40724, line 346: Why has this line be removed? The status can be cancelled, right?

alex40724:
@mjansenDatabay Thanks for the PR. I am unsure if the fix is correct, I will need more time to reproduce / test this.

Modules/BookingManager/Objects/classes/class.ilBookingObjectsTableGUI.php
mjansenDatabay: Looks like copy/paste, I'll check this.

mjansenDatabay: Fixed

alex40724:
Thanks again. This should be fixed by https://github.com/ILIAS-eLearning/ILIAS/commit/6a813f003ada1cc4a692bcd789650547a8d4dcd8

PR #3944, opened 2022-01-25
BookingPool: Fix missing $idx when filtering by period

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=31932


alex40724:
Thanks for the PR!

alex40724:
Issue is fixed.

PR #3947, opened 2022-01-26
setup: StatusCommand::execute() return type fixed

utesche:
@klees,

(executed with PHP 8.0.14)
PHP Fatal error:  Uncaught TypeError: Return value of "ILIAS\Setup\CLI\StatusCommand::execute()" must be of the type int, "null" returned. in /srv/www/ilias-t.hsu-hh.de/ilias/8/trunk8/libs/composer/vendor/symfony/console/Command/Command.php:301


klees:
Thx!

PR #3952, opened 2022-01-26
UI/Input/Duration - translate default labels for start/end

nhaagen:
The default labels of the inner inputs were not translated. This is the fix. 
(thanks, @chfsx, for pointing that out)


Amstutz:
Thx a lot!

Amstutz:
Note, I only merged to trunk. There are several conflicts to 7. Even though the previous defaults where rather bad, I am not sure if this is worth merging it down, since all cases seem to translate anyways. If not, feel free to create a PR for 7 as well.

PR #3956, opened 2022-01-26
Cron: Get rid of static methods (almost all) / Use `Dependency Injection`

mjansenDatabay:
This PR changes the nature of most `ilCronManager` methods from `static` to `instance`, which allows `Dependency Injection` and more `Unit Tests` in the future.

Depends on: (#3957 or #3954) and #3966

@klees With this PR I'll get rid of almost all static method calls in `Services/Cron` and used a lot (more) dependency injection. This should prevent issues like mentioned here: https://github.com/ILIAS-eLearning/ILIAS/commit/97b50e9c31a3ed21ab0ef8db24df9683c649194f .
To achieve this, I'll had to slightly modify [`ilComponentDefinitionsStoredObjective`](https://github.com/ILIAS-eLearning/ILIAS/pull/3956/files#diff-efcbdd1c8b2ea080a6b6dc4a6a09ccbc124be47fc6fe127a3a3c0d6d60d1236c). Furtheremore these code changes depend on #3966 .

### DIFF
--- a/Services/Component/classes/Setup/class.ilComponentDefinitionsStoredObjective.php
+++ b/Services/Component/classes/Setup/class.ilComponentDefinitionsStoredObjective.php
@@ -47,7 +47,10 @@ class ilComponentDefinitionsStoredObjective implements Setup\Objective
     {
         return [
             new \ilDatabaseUpdatedObjective(),
-            new \ilDatabaseUpdateStepsExecutedObjective(new ilCOPageDBUpdateSteps())
+            new \ilDatabaseUpdateStepsExecutedObjective(new ilCOPageDBUpdateSteps()),
+            new \ilSettingsFactoryExistsObjective(),
+            new \ilComponentRepositoryExistsObjective(),
+            new \ilComponentFactoryExistsObjective(),
         ];
     }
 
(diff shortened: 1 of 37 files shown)


Services/Component/classes/Setup/class.ilComponentDefinitionsStoredObjective.php
```diff
@@ -47,7 +47,9 @@ public function getPreconditions(Setup\Environment $environment) : array
     {
         return [
             new \ilDatabaseUpdatedObjective(),
-            new \ilDatabaseUpdateStepsExecutedObjective(new ilCOPageDBUpdateSteps())
+            new \ilDatabaseUpdateStepsExecutedObjective(new ilCOPageDBUpdateSteps()),
```
mjansenDatabay, line 50: @klees Could you please have a look at the changes in this file?

klees: Please add the `Objective` for the component repository explicitly too. Currently, `RESOURCE_COMPONENT_REPOSITORY` (attention, name change!) only exists by accident, as `RESOURCE_COMPONENT_FACTORY` depends on it. This could just change and this code would break. It's best too be super explicit here, IMO.

Services/Component/classes/Setup/class.ilComponentDefinitionsStoredObjective.php
```diff
@@ -61,6 +63,12 @@ public function achieve(Setup\Environment $environment) : Setup\Environment
         $db = $environment->getResource(Setup\Environment::RESOURCE_DATABASE);
         $ini = $environment->getResource(Setup\Environment::RESOURCE_ILIAS_INI);
         $client_ini = $environment->getResource(Setup\Environment::RESOURCE_CLIENT_INI);
+        /** @var ilComponentRepository $component_repository  */
+        $component_repository = $environment->getResource(Setup\Environment::RESOURCE_COMPONENT_DATABASE);
```
klees: I [changed the name to `RESOURCE_COMPONENT_REPOSITORY`](https://github.com/ILIAS-eLearning/ILIAS/commit/c88e42946608716a7fc6bc92cabcc2330a1435a9) to implement our [repository naming policy](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/repository-pattern.md#conventions).

klees (approved):
Hi @mjansenDatabay,

I hope inline comments are ok here. This LTGM in general, I left two remarks.

Feel free to change and merge afterwards without further review or notice.

Best regards!

Services/Component/classes/Setup/class.ilComponentDefinitionsStoredObjective.php
mjansenDatabay: Thanks a lot for your review @klees . I'll tackle your remarks,

PR #3958, opened 2022-01-27
Forum: Bugfixes of PHP 8 Regressions

playamongthestars:


Modules/Forum/classes/class.ilForumTopicTableGUI.php
```diff
@@ -290,8 +290,6 @@ protected function fillRow(array $a_set) : void
         $this->tpl->setVariable('CSS_ROW', $css_row);
 
         $this->ctrl->setParameter($this->getParentObject(), 'thr_pk', '');
-        $this->ctrl->setParameter($this->getParentObject(), 'user', '');
```
mjansenDatabay, line 293: Why did you remove this?

Modules/Forum/classes/class.ilForumTopic.php
```diff
@@ -688,9 +688,6 @@ public function getNestedSetPostChildren(?int $pos_id = null, ?int $num_levels =
             $casted_row['post_id'] = (int) $row['post_id'];
             $casted_row['post_read'] = (int) $row['post_read'];
             $casted_row['children'] = (int) $row['children'];
-            $casted_row['rgt'] = (int) $row['rgt'];
```
mjansenDatabay, line 691: Are you sure these are never used?

Modules/Forum/classes/class.ilForumSettingsGUI.php
```diff
@@ -762,30 +762,19 @@ private function initNotificationSettingsForm() : bool
 
             $cb_grp = new ilCheckboxGroupInputGUI($this->lng->txt('notification_settings'), 'notification_events');
 
-            $notify_modified = new ilCheckboxInputGUI($this->lng->txt('notify_modified'), 'notify_modified');
-            $notify_modified->setValue((string) ilForumNotificationEvents::UPDATED);
+            $notify_modified = new ilCheckboxOption($this->lng->txt('notify_modified'), (string) ilForumNotificationEvents::UPDATED);
```
mjansenDatabay, line 765: Dind't affect this the storage of the form?

Modules/Forum/classes/class.ilForumTopicTableGUI.php
playamongthestars: Because, it adds empty parameters to the url which leads to crashes

Modules/Forum/classes/class.ilForumTopic.php
playamongthestars: yes

Modules/Forum/classes/class.ilForumSettingsGUI.php
playamongthestars: In fact, I can test this, after the getLinkTarget-issue is fixed.

PR #3959, opened 2022-01-27
WIP: Use Roave Better reflection to not include all classes while composer install

PurHur:
Hello,

sorry that i open this issue again. I would really love to see a change here and i am open for discussions to have make some progress. The PR is not in a final state, i will clean it up, do code style changes if this topic has some chance to get merged. 

Let me describe my (and for others) problem again:

After running composer install a post-install script build the artifacts for ilias. But while doing this a class named `ImplementationOfInterfaceFinder` is iterating over **every** class inside the ilias core and the customizing folder to find classes which implements different Interfaces.
Actually this method is called 7 times. Here is a list of callers:
/var/www/ilias/src/Setup/ImplementationOfAgentFinder.php:70->getCoreAgents
/var/www/ilias/src/Setup/ImplementationOfAgentFinder.php:137->iterator_to_array
/var/www/ilias/Services/GlobalScreen/classes/Setup/class.ilGlobalScreenBuildProviderMapObjective.php:37->iterator_to_array
/var/www/ilias/Services/GlobalScreen/classes/Setup/class.ilGlobalScreenBuildProviderMapObjective.php:37->iterator_to_array
/var/www/ilias/Services/GlobalScreen/classes/Setup/class.ilGlobalScreenBuildProviderMapObjective.php:37->iterator_to_array
/var/www/ilias/Services/GlobalScreen/classes/Setup/class.ilGlobalScreenBuildProviderMapObjective.php:37->iterator_to_array
/var/www/ilias/Services/GlobalScreen/classes/Setup/class.ilGlobalScreenBuildProviderMapObjective.php:37->iterator_to_array

I think the interfaces of each class should be cached, even in the current implementation.

While calling ReflectionClass the class is obviously included. This means that every class in the whole project is included in the same process. This needs a lot of memory. Also if you have ILIAS installations with 20 plugins (of different vendors not known to each other) the chance that something is duplicated or just dont expects to be included breaks the whole composer install process.
Currently me and other people therefore move the Customizing folder outside ILIAS for an update, but this removes the features introduced by this mechanic. This cant be a solution for the future.
https://github.com/ILIAS-eLearning/ILIAS/pull/3901 This PR of @klees looks like the `composer install` will be required even if you want to install a plugin, so we need a viable solution to our problem. 

So here is my solution. I introduced a package https://github.com/Roave/BetterReflection which is a drop in replacement for the ReflectionClass but reads the source without loading it with the php-interpreter but in php itself. So these classes are not included but still reflected to get the interfaces and other stuff of a class. The quality of this package is very good and we can trust the maintainer(s) todo a good job even in the future.

The downside of using this package is that its quite slow. With cached reflection results, this needs about 4 minutes on my laptop. 
To mitigate this, we could introduce a cache file, which lives inside the repository which includes the reflections data to increase the speed to almost nothing.
A composer (or build-artifacts) argument could be used to refresh that cache file, which then can be committed into a release. 

I hope someone is interested in this. I hope i will be able to join your next JourFix to discuss this in a bigger round, but feel free to start the discussion now. 

More Information:
I tried out multiple solutions like described here: https://mantis.ilias.de/view.php?id=28848
The solution of Richard looked nice for me, but is still slower. (I think due to the thousand new processes)
The solution of Michael is basically what this package does. (I think a lot more optimized)

I had to upgrade `"phpdocumentor/reflection-docblock": "^5"` from 3 to 5. This needs testing. 

Please note that this PR is just a starter for a longer discussion. I see that this cant be merged quick, but it works if you want to try it. 

Greetings
Andre

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,11 @@
 	"authors": [],
 	"config": {
 		"optimize-autoloader": true,
-		"vendor-dir": "./libs/composer/vendor"
+		"vendor-dir": "./libs/composer/vendor",
+		"allow-plugins": {
+			"simplesamlphp/composer-module-installer": true,
+			"cweagans/composer-patches": true
+		}
 	},
 	"scripts": {
 		"post-autoload-dump": [
@@ -51,7 +55,8 @@
 		"ramsey/uuid": "^3.8",
 		"enshrined/svg-sanitize": "^0.13.0",
 		"guzzlehttp/guzzle": "^6.3",
-		"simplesamlphp/simplesamlphp": "^1.18"
+		"simplesamlphp/simplesamlphp": "^1.18",
+		"roave/better-reflection": "^4.12"
 	},
 	"require-dev": {
 		"mikey179/vfsstream": "^1.6",
@@ -59,7 +64,7 @@
 		"phpunit/php-code-coverage": "^8.0",
 		"mockery/mockery": "^1.2",
 		"phpunit/php-token-stream": "^4.0",
-		"phpdocumentor/reflection-docblock": "^3.3",
+		"phpdocumentor/reflection-docblock": "^5",
 		"friendsofphp/php-cs-fixer": "^2.15",
 		"sebastian/environment": "^5.0",
 		"phpunit/php-timer": "^3.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "9bf3c5a26523a2791f771ce70a5546b2",
+    "content-hash": "11dc0b98ad01e6d756ede05d1dcdd170",
     "packages": [
         {
             "name": "cweagans/composer-patches",
@@ -211,16 +211,16 @@
         },
         {
             "name": "filp/whoops",
-            "version": "2.14.4",
+            "version": "2.14.5",
             "source": {
                 "type": "git",
                 "url": "https://github.com/filp/whoops.git",
-                "reference": "f056f1fe935d9ed86e698905a957334029899895"
+                "reference": "a63e5e8f26ebbebf8ed3c5c691637325512eb0dc"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/filp/whoops/zipball/f056f1fe935d9ed86e698905a957334029899895",
-                "reference": "f056f1fe935d9ed86e698905a957334029899895",
+                "url": "https://api.github.com/repos/filp/whoops/zipball/a63e5e8f26ebbebf8ed3c5c691637325512eb0dc",
+                "reference": "a63e5e8f26ebbebf8ed3c5c691637325512eb0dc",
                 "shasum": ""
             },
             "require": {
@@ -270,7 +270,7 @@
             ],
             "support": {
                 "issues": "https://github.com/filp/whoops/issues",
-                "source": "https://github.com/filp/whoops/tree/2.14.4"
+                "source": "https://github.com/filp/whoops/tree/2.14.5"
             },
             "funding": [
                 {
@@ -278,7 +278,7 @@
                     "type": "github"
                 }
             ],
-            "time": "2021-10-03T12:00:00+00:00"
+            "time": "2022-01-07T12:00:00+00:00"
         },
         {
             "name": "geshi/geshi",
@@ -859,6 +859,53 @@
             "time": "2021-09-22T16:34:51+00:00"
         },
         {
+            "name": "jetbrains/phpstorm-stubs",
+            "version": "v2019.3",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/JetBrains/phpstorm-stubs.git",
+                "reference": "883b6facd78e01c0743b554af86fa590c2573f40"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/JetBrains/phpstorm-stubs/zipball/883b6facd78e01c0743b554af86fa590c2573f40",
+                "reference": "883b6facd78e01c0743b554af86fa590c2573f40",
+                "shasum": ""
+            },
+            "require-dev": {
+                "nikic/php-parser": "^4",
+                "php": "^7.1",
+                "phpdocumentor/reflection-docblock": "^4.3",
+                "phpunit/phpunit": "^7"
+            },
+            "type": "library",
+            "autoload": {
+                "files": [
+                    "PhpStormStubsMap.php"
+                ]
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "Apache-2.0"
+            ],
+            "description": "PHP runtime & extensions header files for PhpStorm",
+            "homepage": "https://www.jetbrains.com/phpstorm",
+            "keywords": [
+                "autocomplete",
+                "code",
+                "inference",
+                "inspection",
+                "jetbrains",
+                "phpstorm",
+                "stubs",
+                "type"
+            ],
+            "support": {
+                "source": "https://github.com/JetBrains/phpstorm-stubs/tree/master"
+            },
+            "time": "2019-12-05T16:56:26+00:00"
+        },
+        {
             "name": "jumbojett/openid-connect-php",
             "version": "v0.9.5",
             "source": {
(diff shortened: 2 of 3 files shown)


PurHur:
Tests are broken, but its about the change. I will have a look into these tomorrow or so.

PurHur:
I wondered why so many people told me that they see no errors. Now i know. Only real fatal errors not catched by the `catch(\Throwable $e)` are shown. Most of the errors are just silenced. So even if someone tries a composer update the developer will probably not see that code is running and throwing errors all over the place.

There are two errors in the ILIAS Core itself. If they are removed atleast the error message could be printed so devs are aware that something is wrong. 


```
Call to undefined function wfUrlProtocols()
#0 /var/www/ilias/libs/composer/vendor/composer/ClassLoader.php(571): include()
#1 /var/www/ilias/libs/composer/vendor/composer/ClassLoader.php(428): Composer\Autoload\includeFile('/var/www/ilias/...')
#2 [internal function]: Composer\Autoload\ClassLoader->loadClass('OnlyIncludeRepl...')
#3 [internal function]: spl_autoload_call('OnlyIncludeRepl...')
#4 /var/www/ilias/src/Setup/ImplementationOfInterfaceFinder.php(60): ReflectionClass->__construct('OnlyIncludeRepl...')
#5 /var/www/ilias/src/Setup/ImplementationOfAgentFinder.php(102): ILIAS\Setup\ImplementationOfInterfaceFinder->getMatchingClassNames('ILIAS\\Setup\\Age...', Array)
#6 /var/www/ilias/src/Setup/ImplementationOfAgentFinder.php(70): ILIAS\Setup\ImplementationOfAgentFinder->getCoreAgents()
#7 /var/www/ilias/src/Setup/CLI/HasAgent.php(48): ILIAS\Setup\ImplementationOfAgentFinder->getAgents()
#8 /var/www/ilias/src/Setup/CLI/BuildArtifactsCommand.php(45): ILIAS\Setup\CLI\BuildArtifactsCommand->getRelevantAgent(Object(Symfony\Component\Console\Input\ArgvInput))
#9 /var/www/ilias/libs/composer/vendor/symfony/console/Command/Command.php(255): ILIAS\Setup\CLI\BuildArtifactsCommand->execute(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#10 /var/www/ilias/libs/composer/vendor/symfony/console/Application.php(1009): Symfony\Component\Console\Command\Command->run(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#11 /var/www/ilias/libs/composer/vendor/symfony/console/Application.php(273): Symfony\Component\Console\Application->doRunCommand(Object(ILIAS\Setup\CLI\BuildArtifactsCommand), Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#12 /var/www/ilias/libs/composer/vendor/symfony/console/Application.php(149): Symfony\Component\Console\Application->doRun(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#13 /var/www/ilias/setup/cli.php(39): Symfony\Component\Console\Application->run()
#14 {main}

------

syntax error, unexpected 'new' (T_NEW)
#0 /var/www/ilias/libs/composer/vendor/composer/ClassLoader.php(428): Composer\Autoload\includeFile('/var/www/ilias/...')
#1 [internal function]: Composer\Autoload\ClassLoader->loadClass('soap_client')
#2 [internal function]: spl_autoload_call('soap_client')
#3 /var/www/ilias/src/Setup/ImplementationOfInterfaceFinder.php(60): ReflectionClass->__construct('soap_client')
#4 /var/www/ilias/src/Setup/ImplementationOfAgentFinder.php(102): ILIAS\Setup\ImplementationOfInterfaceFinder->getMatchingClassNames('ILIAS\\Setup\\Age...', Array)
#5 /var/www/ilias/src/Setup/ImplementationOfAgentFinder.php(70): ILIAS\Setup\ImplementationOfAgentFinder->getCoreAgents()
#6 /var/www/ilias/src/Setup/CLI/HasAgent.php(48): ILIAS\Setup\ImplementationOfAgentFinder->getAgents()
#7 /var/www/ilias/src/Setup/CLI/BuildArtifactsCommand.php(45): ILIAS\Setup\CLI\BuildArtifactsCommand->getRelevantAgent(Object(Symfony\Component\Console\Input\ArgvInput))
#8 /var/www/ilias/libs/composer/vendor/symfony/console/Command/Command.php(255): ILIAS\Setup\CLI\BuildArtifactsCommand->execute(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#9 /var/www/ilias/libs/composer/vendor/symfony/console/Application.php(1009): Symfony\Component\Console\Command\Command->run(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#10 /var/www/ilias/libs/composer/vendor/symfony/console/Application.php(273): Symfony\Component\Console\Application->doRunCommand(Object(ILIAS\Setup\CLI\BuildArtifactsCommand), Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#11 /var/www/ilias/libs/composer/vendor/symfony/console/Application.php(149): Symfony\Component\Console\Application->doRun(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#12 /var/www/ilias/setup/cli.php(39): Symfony\Component\Console\Application->run()
#13 {main}
```

`soap_client` should probably ignored. 
`OnlyIncludeReplacer` is the second class in `Parser.php`. With the classmap and the Reflection PHP Files cant have more that one class, because it would be included twice.


@klees What do you think? Should we add these two files to the global ignore on the top of the InterfaceFinder so we could print (or with a logger) these errors to the devs? Else it could be that more errors occure which are not detected.

klees:
@PurHur Nice to hear that there are only two problems as of now =) This was worse when I started the implementation...

We would definitely require some kind of abstraction to print these errors, `print_r` is not something I want to see there.

Maybe we could ask callers of `getMatchingClassNames` for an optional callabel of the structure `function($class, $error)`, so we could simply print something like `Error when reading $class` for the default case and the error only when `--verbose` is used?

In general, I don't expect `build-artifacts` to be the tool to detect errors, so I would not put too much effort (and code) into that problem. Something along the line of the aforementioned closure would be slim enough, though.

PurHur:
I will into it how it could be printed better, probably the IOWarpper or the OutputInterface implementation has to be forwarded somehow. I just wanted to know how much errors are appearing. 

If build-artifacts is not the right place, at wich point in the dev cycle should you be able too see that the class file is included in such a context? 
E.g. A dev is trying to detect a specific interface and it wont work (because of a typo or something else). Now he has to use a verbose argument to see an error?

klees:
Hi @PurHur,

regarding the change that is announced in the title of the PR: We discussed the proposal in the Technical Board and decided to not follow your idea. The two main reasons are:

* We try to be super prudent when including new dependencies. We often feel, that we are already attached too much to other projects and code. The vendor of the code seems to be solid in general, but we get the same functionality from vanilla php (almost) and hence don't see that it is worth to include that dependency.
* For the general problem, there seems to be some kind of inherent time/space tradeoff. Currently we are at the point of lowest time (and highest space?) that is known. Other solutions, such as this here, might require less space but considerably more time. Since this code (building artifacts) is run by developers a lot, we feel that the bias towards less time is the one to choose currently.

Thanks a lot for making us think and reason about this area in our application by proposing this change. @mjansenDatabay will be happy to unroll our discussion for you in more detail, if required.

Regarding your proposal to treat errors in the script differently: I would be happy to get a PR to improve the situation there. The gist of my idea regarding output just is: Let's try to be consistent in the setup, which would mean to give full stacktraces only if users use `verbose` as an option and just print the error in the other case.

Regarding your question how to know when your classes are included: When moving your code into a framework (which ILIAS basically is at this point) you simply give up much control about how your code is handled, especially when it comes to autoloading of classes. As far as I understand, this is the reason for this [rule from PSR-1](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-1-basic-coding-standard.md#1-overview): "Files SHOULD either declare symbols (classes, functions, constants, etc.) or cause side-effects (e.g. generate output, change .ini settings, etc.) but SHOULD NOT do both." Frameworks need to be able to inspect symbols without causing side effects, which basically is what happens here. So from my POV, we should not have too much expectations when or how our classes are loaded (and when they are not).

Best regards!

PR #3962, opened 2022-01-27
Fix Mime Type Detection for stream

kergomard:
When using getContents() on a stream the position is moved forward, thus
a consecutive call on getContents() will receive an empty string back.
As we are calling this twice here (in getMimeType and in getFileSize)
this will leed to undetected mime-types (second call). To avoid any
troubles I rewind the position both times and check first if strings are
seekable to avoid throwing an exception.

See: https://mantis.ilias.de/view.php?id=31948


chfsx:
Thanks a lot @swiniker !

PR #3965, opened 2022-01-27
UI: allow Bulky Links in Meta Bar

klees:
Hi everyone,

this allows Bulky Links to be added in the Meta Bar. I wonder, why this isn't possible now...

Anyway, this will help to fix [#31808](https://mantis.ilias.de/view.php?id=31808).

Best regards!


Amstutz:
Thx a lot. I suggest, that we fix this completely and then pick to 7 and test along with the fix in global screen in one go, i.O.?

PR #3968, opened 2022-01-28
Component: move ilTable to presentation table

daniwe4:
Move all ComponentSettings files into a Settings folder.
Change GUI-Table to TablePresentation.


klees:
I added this to #4008.

PR #3970, opened 2022-01-28
fixed item group expansion error

thibsy:
In an item group a clientside error occurred due to an invalid target URL to ilContainerBlockPropertiesStorage. This was because a) the class did not match ilCtrl's GUI class requirements and b) the link target was not built properly.


thibsy:
Hi @alex40724, also wanted to ask here again to move forward with this minor change.

alex40724:
@thibsy Thanks for the PR. And sorry for the delay, I mostly handle bugs according to votings.

There are definitely some backward compatibility breaks here. GUI ending has been a convention, but not been enforced in ILIAS 6 and ILIAS 7 (this is why the other proposed fixes by @mjansenDatabay work, too). As mentioned in the other PR we have two remaining issues in ILIAS 8

- ilAccordionPropertiesStorage (has been fixed the same way as ilTablePropertiesStorage in ILIAS 7, broken in trunk)
- ilTablePropertiesStorage (fixed in 7, broken in trunk)

PR #3971, opened 2022-01-28
fixed DynamicInputDataIterator::current().

thibsy:
The solution introduced by @klees in https://github.com/srsolutionsag/ILIAS/pull/12 was almost right. I forgot that the key when iterating is the submitted name of the input, but in order to delegate it with `withInput()` it must be the same as the rendered name. Therefore the key must be used similarly as in `DynamicInputsNameSource::getNewName()`.


thibsy:
Thanks!

PR #3972, opened 2022-01-28
Remove type bug in STR Utils

PurHur:
I saw that other fix and this is the same as the other one.

https://www.php.net/manual/de/function.strpos.php

The function could return `int|false`. In PHP 8 unified types are possible but i dont know if ilias would/want to support these so i just used the same fix.

### DIFF
--- a/Services/Utilities/classes/class.ilStr.php
+++ b/Services/Utilities/classes/class.ilStr.php
@@ -34,7 +34,7 @@ class ilStr
         }
     }
     
-    public static function strPos(string $a_haystack, string $a_needle, ?int $a_offset = null) : int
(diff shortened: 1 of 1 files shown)


PurHur:
Some tests are failing but i guess this is because some error in the usage of a function which was not used correctly or a CI problem. If its first all the failed tests are actually bugs.

PurHur:
I guess the tests are notfailing because of this change.

I looked up all usages of that function. There are 7 usages which all are using the function correctly. 
If the type is not removed 3 additional bugs in the T&A are introduced :) @mbecker-databay

chfsx:
Thanks @PurHur

PR #3985, opened 2022-02-01
remove external feeds; migration part

alex40724:
This removes the main feed object data. Unfortunately there does not seem to be a better way to do this (use ilObject->delete, incl. events).

### DIFF
--- a/Modules/ExternalFeed/classes/Setup/class.ilExternalFeedRemoveMigration.php
+++ b/Modules/ExternalFeed/classes/Setup/class.ilExternalFeedRemoveMigration.php
@@ -136,6 +136,20 @@ final class ilExternalFeedRemoveMigration implements Setup\Migration
             " obj_id = " . $db->quote($typ_id, "integer")
         );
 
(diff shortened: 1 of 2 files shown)


Modules/ExternalFeed/classes/Setup/class.ilExternalFeedRemoveMigration.php
```diff
@@ -0,0 +1,218 @@
...
+        );
+
+        // ...object_data (feed objects)
+        $this->log("Delete object_data entries of feed objects.");
+        $this->manipulate(
+            "DELETE FROM object_data WHERE " .
+            " " . $db->in("obj_id", $obj_ids, false, "integer")
+        );
+        // ...object_data (roles)
+        $this->log("Delete object_data entries of local roles of feed objects.");
```
smeyer-ilias, line 127: Please also delete the entries in role_data:
DELETE from role_data WHERE  $db->in("role_id", $local_roles, false, "integer")

smeyer-ilias (approved):

PR #3986, opened 2022-02-01
LS: PHP8 improvements 2

daniwe4:


Modules/LearningSequence/classes/class.ilObjLearningSequenceGUI.php
```diff
@@ -785,4 +801,9 @@ public function addCustomData(array $a_data) : array
 
         return $res_data;
     }
+
+    public function showPossibleSubObjects()
+    {
+        parent::showPossibleSubObjects();
+    }
```
klees, line 811: Why add this if it only call the parent anyway?

Modules/LearningSequence/classes/class.ilObjLearningSequence.php
```diff
@@ -210,7 +210,7 @@ protected function getDIC() : ArrayAccess
         return $this->dic;
     }
 
-    public function getDI() : ArrayAccess
+    public function getDI() : ilLSDI
```
klees: Only hint on `ArrayAccess` here, so we will be able to overwrite this with a simple `array` for testing purpose. Always hint on the weakest possible type, as long as there is no really really good reason to do otherwise.

Modules/LearningSequence/classes/class.ilObjLearningSequence.php
```diff
@@ -417,12 +417,15 @@ public static function _goto(int $target, string $add = "")
         $ilErr = $DIC['ilErr'];
         $lng = $DIC['lng'];
         $ilUser = $DIC['ilUser'];
+        $request_wrapper = $DIC->http()->wrapper()->query();
+        $refinery = $DIC->refinery();
 
         if (substr($add, 0, 5) == 'rcode') {
             if ($ilUser->getId() == ANONYMOUS_USER_ID) {
+                $tar = $request_wrapper->retrieve("target", $refinery->kindlyTo()->string());
```
klees: Don't be lazy, just write `$target`. This will help our future selves.

klees (changes requested):
Hi @daniwe4,

LGTM in general, but some minor things to be improved.

Thanks!

Modules/LearningSequence/classes/class.ilObjLearningSequenceGUI.php
daniwe4: this changes the modifier from protected to puplic. It is called by ilObjLearningSequenceContentGUI.

Modules/LearningSequence/classes/class.ilObjLearningSequence.php
daniwe4: done
I have to change the typehints of ilLSLocalDI::init
`public function init(
        ArrayAccess $dic,
        ilLSDI $lsdic,
        DataFactory $data_factory,
        ilObjLearningSequence $object
    )`

`public function init(
        ArrayAccess $dic,
        ArrayAccess $lsdic,
        DataFactory $data_factory,
        ilObjLearningSequence $object
    )`

Modules/LearningSequence/classes/class.ilObjLearningSequence.php
daniwe4: I'm not lazy ;)
$target won't work here, because we got a parameter called $target.
rename it to $request_target
done

klees: Thanks, I just missed that =)

PR #3991, opened 2022-02-01
Curl fix parseHeader

pascalseeland:
To my understanding, the parseHeader-callback is called with one headerline each


smeyer-ilias (approved):
Thanks, LGTM

PR #3992, opened 2022-02-02
Session: fix error giving an int as setting-default

nhaagen:
There is typechecking for settings-defaults now, and int-defaults should be handled differently.


pascalseeland:
thx

PR #3993, opened 2022-02-02
Fix escaping of \\

kergomard:
This fixes the escaping of \\ in MathJax as reported in https://mantis.ilias.de/view.php?id=31788. I tested this with server- and client-side rendering. With client-side rendering this also fixes the rendering of newlines (\\). When using server-side rendering this does still not produce an output when a newline is inserted, but instead of showing an Error it leads to an empty image (at least on my setup).

### DIFF
--- a/Services/MathJax/classes/class.ilMathJax.php
+++ b/Services/MathJax/classes/class.ilMathJax.php
@@ -341,18 +341,15 @@ class ilMathJax
                 }
 
                 // omit the html newlines added by the ILIAS page editor
-                // handle custom newlines in JSMath (still needed?)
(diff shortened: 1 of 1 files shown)


fneumann:
The PR removes a problematic replacement of '\\\\' by '\\cr'.

I did some historic search and found the earliest implementation here:
https://github.com/ILIAS-eLearning/ILIAS/blob/release_4-3/Services/Utilities/classes/class.ilUtil.php#L4402

As far as I understand that pattern, the current replacement is wrong anyway. The original comment says its is related to jsMath which is not supported anymore.
For me it would be ok, to remove it competely, but I'm not an expert in Latex expressions to oversee the consequences.

PR #3996, opened 2022-02-03
Authentication/SoapAuth: Fix SOAP authentication single sign on

mjansenDatabay:
This PR re-adds a feature which got lost in
a refactoring some years ago.
Currently, the relevant query parameters get
lost if the request targets to goto.php.

Mantis Issue: https://mantis.ilias.de/view.php?id=31990

If approved this should also be picked to `release_6` and `trunk`.


mjansenDatabay:
Hi @pascalseeland,

Best regards!

PR #3997, opened 2022-02-03
didactic templates in create form

mhilbert:
Selecting and applying didactic templates missing in the create dialogue, only possible via the settings tab.
see discussion: https://docu.ilias.de/goto_docu_frm_3454_8030.html


alex40724:
https://mantis.ilias.de/view.php?id=31995

alex40724:
Thanks for the PR!

PR #3999, opened 2022-02-03
Add client ip column in user list

padvincenzo:
This PR adds a selectable column for the client IP.

### DIFF
--- a/Services/User/classes/class.ilUserTableGUI.php
+++ b/Services/User/classes/class.ilUserTableGUI.php
@@ -256,6 +256,11 @@ class ilUserTableGUI extends ilTable2GUI
 
         // fields that are always shown
         unset($cols["username"]);
+
+        $cols["client_ip"] = array(
+            "txt" => $lng->txt("client_ip"),
+            "default" => false
+        );
         
         return $cols;
     }


alex40724:
Thanks for the PR.

I think this might need a FR and a privacy discussion.

mjansenDatabay:
Dear @padvincenzo,

thanks a lot for your pull request. We very much appreciate your contributions and welcome all developers providing pull requests to improve ILIAS.

Even though the amount of code changes in this PR is very small, there are two issues we need to bring up about it:

1. Changing the behavior of the application this way could potentially introduce issues in a stable release. Unfortunately those changes won't be covered by our TestRail test cases anymore, which are manually processed in our beta phase only. This is why changes which do not address an actual issue (in Mantis tickets) should be best provided as a feature request ([feature wiki page](https://docu.ilias.de/goto_docu_wiki_wpage_788_1357.html)) and discussed with the respective `Component Maintainer`, which leads to the next problem:
2. We did have a look at the PR and find the change non-trivial in light of security and privacy implications, so we would rely a lot on a capable maintainers informed opinion on the change. Your PR addresses a component which actually has no current maintainer - which of course is not your fault.

Therefore we would like to close this PR for now. Nevertheless we strongly like to encourage you to contribute to ILIAS again. While this PR can't be taken in, your other work is highly regarded.

If you have any questions regarding the contribution process (see also: 
https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/contributing.md)
or other topics, don't hesitate to contact me or the @ILIAS-eLearning/technical-board .

Best regards,
@mjansenDatabay

PR #4003, opened 2022-02-04
Proposal/rector ilutils

chfsx:
As announced in the PHP8 meeting, I used Rector to replace the ilUtil:sendXY methods throughout the codebase.

For this purpose a specific Rector was implemented: https://github.com/ILIAS-eLearning/ILIAS/compare/trunk...srsolutionsag:proposal/rector-ilutils?expand=1#diff-721f6d9434d1660bbb895c63e1b63ec7073d98f809a919b8ee22f1fcd170c2d5
Furthermore, I outsourced the logic how to find a DIC dependency. We can certainly use this in further refactorings.

The logic is the following:
- if a ClassPropery exists that can provide ilGlobalTemplateInterface, it will be used.
- if not, it will be introduced.
- in static methods a fallback is used:
- the same fallback is used if ilUtil::sendXY is used in the constructor. Unfortunately, I haven't found a better variant there yet. Rector would introduce the ClassPropery, but unfortunately would not do an assingment. Maybe I can improve this, but it is only a few places.

I have applied this to Modules and Services, with a few exceptions: https://github.com/ILIAS-eLearning/ILIAS/compare/trunk...srsolutionsag:proposal/rector-ilutils?expand=1#diff-0eceb7c15ebfd2a2294dbc94676ad078018fe598aba3182f3ac03966c38fe90bR14
there are sometimes files with multiple classes, or wrong derivations, that break rector and the code analysis.

As mentioned, take a quick look at your services/modules. If you prefer to refactor your classes manually, feel free to contact me those in the PR. I will then take them out of the PR.

A Readme, how to use this Rector, is here: https://github.com/ILIAS-eLearning/ILIAS/compare/trunk...srsolutionsag:proposal/rector-ilutils?expand=1#diff-1323e745b4e3d755d247a77a3b99cf6ddcb060ced46ab294ba385a2556e6a85f

I also had to adjust some tests in the T&A, because they now fail due to the "visible" dependency.


chfsx:
As an addition: The static methods ilUtil::sendInfo and the other three will be removed either way

chfsx:
Here some examples:
- Rector introduced a new ClassMember: https://github.com/ILIAS-eLearning/ILIAS/pull/4003/files#diff-311446b8583e92bf482deac59fe0ec1effd9f5dc391786b9321907543a159d81
- An existing ClassMember has been used here: https://github.com/ILIAS-eLearning/ILIAS/pull/4003/files#diff-ad330bc0cde322cf891725746e5bdea8311115b4911e182dc5f0c7ac36474dda
- Rector could not introduce the ClassMember (method call was in constructor), used a fallback: https://github.com/ILIAS-eLearning/ILIAS/pull/4003/files#diff-c11aaf8dd444a7f9cc9b0046dfe56c0a59fce6ca122dfb8b188baab014b16e04
- Method call is in a static method, Rector used Fallback: https://github.com/ILIAS-eLearning/ILIAS/pull/4003/files#diff-6e8b110cb0b24c9f135dd1d2f1a7b2d3ffa804d5d6fd3ac624fb634149fdeccdR513

Modules/BookingManager/BookingProcess/class.ilBookingAssignParticipantsTableGUI.php
```diff
@@ -79,12 +79,10 @@ public function __construct(
 
 
         if ($this->pool->getScheduleType() == ilObjBookingPool::TYPE_NO_SCHEDULE) {
-            ilUtil::sendInfo(
-                sprintf(
-                    $this->lng->txt("book_objects_available"),
-                    ilBookingReservation::numAvailableFromObjectNoSchedule($a_booking_obj_id)
-                )
-            );
+            $GLOBALS['DIC']->ui()->mainTemplate()->setOnScreenMessage('info', sprintf(
```
alex40724: Can we avoid those $GLOBALS? There is a $DIC already available in this constructor.

Modules/Exercise/Submission/class.ilExSubmission.php
```diff
@@ -29,6 +29,7 @@ class ilExSubmission
     protected ilExAssignmentTypeInterface $ass_type;
     protected ilExAssignmentTypes $ass_types;
     protected ilExcAssMemberState $state;
+				private \ilGlobalTemplateInterface $main_tpl;
```
alex40724: Not sure about the indentation here.

Modules/Exercise/Submission/class.ilExSubmission.php
```diff
@@ -38,6 +39,7 @@ public function __construct(
         bool $a_public_submissions = false
     ) {
         global $DIC;
+								$this->main_tpl = $DIC->ui()->mainTemplate();
```
alex40724: Indentation looks strange here, too.

Modules/Exercise/classes/class.ilObjExerciseGUI.php
```diff
@@ -901,13 +901,10 @@ public static function _goto(
...
             );
         } elseif ($ilAccess->checkAccess("read", "", ROOT_FOLDER_ID)) {
-            ilUtil::sendFailure(
-                sprintf(
-                    $lng->txt("msg_no_perm_read_item"),
-                    ilObject::_lookupTitle(ilObject::_lookupObjId($a_target))
-                ),
-                true
-            );
+            $GLOBALS['DIC']->ui()->mainTemplate()->setOnScreenMessage('failure', sprintf(
```
alex40724: Again a $GLOBALS introduction.

Modules/LearningModule/classes/class.ilObjContentObject.php
```diff
@@ -56,12 +56,14 @@ class ilObjContentObject extends ilObject
     public array $auto_glossaries = array();
     private string $import_dir = '';
     protected ilObjLearningModule $lm;
+				private \ilGlobalTemplateInterface $main_tpl;
```
alex40724: ...indentation

Modules/LearningModule/classes/class.ilObjContentObject.php
```diff
@@ -56,12 +56,14 @@ class ilObjContentObject extends ilObject
...
 
     public function __construct(
         int $a_id = 0,
         bool $a_call_by_reference = true
     ) {
         global $DIC;
+								$this->main_tpl = $DIC->ui()->mainTemplate();
```
alex40724: ...indentation

Modules/LearningModule/classes/class.ilStructureObjectGUI.php
```diff
@@ -583,7 +583,7 @@ public static function _goto(
         }
         
         if ($ilAccess->checkAccess("read", "", ROOT_FOLDER_ID)) {
-            ilUtil::sendFailure(sprintf(
+            $GLOBALS['DIC']->ui()->mainTemplate()->setOnScreenMessage('failure', sprintf(
```
alex40724: Another $GLOBALS introduction. This is a static method, but it already includes a global $DIC; at the beginning.

Modules/Blog/Exercise/class.ilBlogExerciseGUI.php
```diff
@@ -30,10 +30,12 @@ class ilBlogExerciseGUI
     protected int $ass_id;
     protected string $file;
     protected \ILIAS\DI\UIServices $ui;
+    private \ilGlobalTemplateInterface $main_tpl;
```
alex40724, line 33: This is avoidable, since the class already depends on \ILIAS\DI\UIServices.

Modules/Blog/Exercise/class.ilBlogExerciseGUI.php
```diff
@@ -240,7 +242,7 @@ protected function finalize() : void
         $exc_gui = ilExSubmissionObjectGUI::initGUIForSubmit($this->ass_id);
         $exc_gui->submitBlog($this->node_id);
 
-        ilUtil::sendSuccess($lng->txt("blog_finalized"), true);
+        $this->main_tpl->setOnScreenMessage('success', $lng->txt("blog_finalized"), true);
```
alex40724, line 245: Could be $this->ui->mainTemplate()->setOnScreen...
instead.

alex40724 (approved):
Thanks for the PR.

I did not check all changes ;-) this is just too much. However I added some comments. 99,5% of the cases are trivial, since the main template is already a member var. So overall it is ok for me to commit this. However it would be nice to avoid the $GLOBALS addition.

chfsx:
Thanks @alex40724 for the feedback! I try to change the fallback to $GLOBAL and prevent some places from using this and use $DIC->ui()->mainTemplate() instead. I will update the PR accordingly.

concerning the strange indentation, did not find the reason for this. but I will try to run the cs-fixer for the changed files before committing them.

chfsx:
@alex40724 I updated the PR, no $GLOBAL['DIC'] are introduced now, I use a local variable assignment $main_tpl instead. Additionally I could fix some other locations and did run the php-cs-fixer for all changed files (which leads to a lot more of changes in this PR...)

chfsx:
To all others: any other feedback? I like to proceed with this as soon as possible

chfsx:
concerning tests I'll have a look at...

mjansenDatabay (approved):
I checked all my components: LGTM!

PR #4004, opened 2022-02-05
Clock: Introduce `PSR-20 Clock Service` as `src/Data` type

mjansenDatabay:
With this PR I would like to introduce a [`PSR-20` compliant `Clock Service`](https://github.com/php-fig/fig-standards/blob/master/proposed/clock.md). While I refactored `Services/Cron` I once again stumbled upon `time()` and `date()`, which caused issues due to wrong usage/wrong assumptions when dealing with `now` (`time()` relies on `January 1 1970 00:00:00 GMT` which is more or less `UTC`, `date()` always relies on `date_default_timezone_set()`, this MUST NOT be mixed when storing `now` information or doing arithmetics when retrieving a `Y-m-d H:i:s` back from the persistence layer).
In addition the usage of these functions posseses the same issues like accessing the global scope or other infrastructure without relying on `Dependency Injection`. Accessing the `time` or `date` is just another subtle way of “reaching out”, which makes it hard/impossible to test those methods.
I would like to change this with this PR.

`User Land` Perspective: See https://github.com/ILIAS-eLearning/ILIAS/pull/4004/files#diff-19ff14f8652abd51b667e794b4258fa8cc58e7845c95af91fbc8ac15f0a41f8aR411

### DIFF
--- a/src/DI/Container.php
+++ b/src/DI/Container.php
@@ -3,13 +3,10 @@
 
 namespace ILIAS\DI;
 
-use ILIAS\Filesystem\Filesystems;
-use ILIAS\FileUpload\FileUpload;
-use ILIAS\GlobalScreen\Services;
-use ILIAS\Refinery\Factory;
-use ILIAS\Skill\Service\SkillService;
-use ILIAS\Repository;
 use ILIAS\BackgroundTasks\BackgroundTaskServices;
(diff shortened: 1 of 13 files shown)


matthiaskunkel:
Jour Fixe, 07 MAR 2022 : We highly appreciate this suggestion and accept this PR for trunk.

PR #4005, opened 2022-02-07
Fix Mime Type Detection for stream

kergomard:
This is a backport of PR 3962 to ILIAS 7. This would allow us to get rid of a very resource consuming copy process in WebDAV.


chfsx:
Thanks @swiniker I wasn't aware we need the same fix in 7 as well, thanks for porting it back!

PR #4006, opened 2022-02-07
BT-31816: Changed h3 to h2 for table titles from table2gui

M-Barz:
Link to Mantis-Report:
https://mantis.ilias.de/view.php?id=31816


alex40724:
Thanks for the PR!

PR #4009, opened 2022-02-08
Object: improvements to PHP8

daniwe4:


Modules/BookingManager/classes/class.ilObjBookingPool.php
```diff
@@ -200,7 +200,7 @@ public function delete() : bool
         return true;
     }
     
-    public function cloneObject($a_target_id, $a_copy_id = 0, $a_omit_tree = false)
+    public function cloneObject(int $a_target_id, int $a_copy_id = 0, $a_omit_tree = false) : ?ilObject
```
utesche, line 203: ...no datatype 'bool' for '$a_omit_tree' yet

daniwe4: Hello @utesche , that's probably lost in the amount of files. Thank you very much for your hint. I'll fix that in an extra PR. Greetings Daniel.

daniwe4: You will find the PR with the fix here: https://github.com/ILIAS-eLearning/ILIAS/pull/4113

Modules/Course/classes/class.ilObjCourse.php
```diff
@@ -593,16 +593,21 @@ public static function _registrationEnabled(int $a_obj_id) : bool
...
-    public function read()
+    public function read() : void
     {
         parent::read();
         $this->setOrderType(ilContainerSortingSettings::_lookupSortMode($this->getId()));
         $this->__readSettings();
     }
 
-    public function create($a_upload = false)
+    public function create($a_upload = false) : int
```
utesche, line 603: stumbled upon another one:  argument $a_upload without type (bool) in function class.ilObjCourse::create()..

daniwe4: Hi @utesche ,
the parent classes of ilObjCourse doesn't use this parameter. So this is only used by ilObjCourse and doesn't fit into this PR.
But thanks again for the detailed look.

PR #4010, opened 2022-02-08
fix for 31592: check if value for divider is too big

okaufman:
Fixes https://mantis.ilias.de/view.php?id=31592
When trying to save a value for divider that is to big, a message will show up: 
"This number has to be smaller than the maximum range."

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assFormulaQuestionGUI.php
+++ b/Modules/TestQuestionPool/classes/class.assFormulaQuestionGUI.php
@@ -689,6 +689,11 @@ class assFormulaQuestionGUI extends assQuestionGUI
                         $max_range->setAlert($this->lng->txt('err_range'));
                         $custom_errors = true;
                     }
+                    $intPrecision = $form->getItemByPostVar('intprecision_' . $variable->getVariable());
+                    if ( $intPrecision >= $max_range) {
+                        $intPrecision->setAlert($this->lng->txt('err_division'));
+                        $custom_errors = true;
+                    }
                 }
             }
 
--- a/Modules/TestQuestionPool/classes/class.assFormulaQuestionGUI.php
+++ b/Modules/TestQuestionPool/classes/class.assFormulaQuestionGUI.php
@@ -690,7 +690,7 @@ class assFormulaQuestionGUI extends assQuestionGUI
                         $custom_errors = true;
                     }
                     $intPrecision = $form->getItemByPostVar('intprecision_' . $variable->getVariable());
-                    if ( $intPrecision >= $max_range) {
+                    if ($intPrecision->getValue() > $max_range->getValue()) {
                         $intPrecision->setAlert($this->lng->txt('err_division'));
                         $custom_errors = true;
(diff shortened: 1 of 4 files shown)


kergomard:
Hi @okaufmann!
As you were the first one of the group to create a PR (so cool, thank you very much!) I had a look atit and have three little questions/comments:
- I think it would be good to change the comparison from `if ( $intPrecision >= $max_range) {` to `if ( $intPrecision->getValue() >= $max_range->getValue()) {`, right? Otherwise, I think, but really correct me if I'm wrong, you are relying on $value being the first variable in the object.
- As the range is inclusive: Wouldn't it be better to use ">" than ">="?  It seems a little random to not include the upper limit.
- Finally: This only addresses the first of the two cases (ie. it prevents errors for new questions), but it doesn't address the problem for existing questions. Is this on purpose?

Thanks again!
Stephan

mbecker-databay:
I think we can skip the existing questions: Due to the error they seem to be totally unusable, so not hidden from anyones sight.
But that is a decision I would pass to @dsstrassner 
Regarding the language file additions: "Nummer" -> "Number" or "Wert" -> "Value" I see there.

Great job, thank you so much!

Modules/TestQuestionPool/classes/class.assFormulaQuestionGUI.php
```diff
@@ -689,6 +689,11 @@ public function editQuestion($checkonly = false)
                         $max_range->setAlert($this->lng->txt('err_range'));
                         $custom_errors = true;
                     }
+                    $intPrecision = $form->getItemByPostVar('intprecision_' . $variable->getVariable());
+                    if ( $intPrecision >= $max_range) {
```
mjansenDatabay:
```suggestion
                    if ($intPrecision >= $max_range) {
```

dsstrassner:
> I think we can skip the existing questions: Due to the error they seem to be totally unusable, so not hidden from anyones sight. But that is a decision I would pass to @dsstrassner Regarding the language file additions: "Nummer" -> "Number" or "Wert" -> "Value" I see there.
> 
> Great job, thank you so much!

Existing questions in the database we really do not need to touch. Super. Many thanks, @okaufman

Modules/TestQuestionPool/classes/class.assFormulaQuestionVariable.php
```diff
@@ -47,6 +47,12 @@ public function getRandomValue()
     {
         
 //		@todo check this
+        if ($this->getIntprecision() > $this->getRangeMax())
```
mjansenDatabay: This is not valid `PSR-2 + X` code style.

okaufman:
@swiniker  thank you for looking at my pull-request. I made the changes you suggested. 
I also added another if-statement in  getRandomValue() to check for the problem in already existing questions.

@mbecker-databay Thank you for your comment. I changed the the lang file to value/Wert

@mjansenDatabay I removed the extra space. Thank you for your sugestion.

PR #4024, opened 2022-02-10
TA Fix, Table Reset Filter, see: 32026

Amstutz:
Fix of https://mantis.ilias.de/view.php?id=32026 along with some gentle refactoring.


Amstutz:
Has this been promoted to trunk as well?

PR #4027, opened 2022-02-11
TA: 21997, do not (auto)save over max passes.

nhaagen:
Please see: https://mantis.ilias.de/view.php?id=21097


dsstrassner:
I discussed the approach with @nhaagen and his fix covers both scenarios, which are provided in the mantis report by ctunjic and nils_brockmann. Thanks a lot @nhaagen.

mbecker-databay:
Thank you for tackling this one! I really appreciate it!

PR #4028, opened 2022-02-11
Reduce Memory usage of StreamInfoResolver

kergomard:
This change radically reduces the memory usage of the StreamInfoResolver and makes it possible to deal with big files, close to the php memory_limit that would otherwise lead to an error.


chfsx:
Thanks a lot, I CP to trunk as well

PR #4030, opened 2022-02-11
Fix logic if no clientId parameter is available

pascalseeland:
In https://github.com/ILIAS-eLearning/ILIAS/commit/a8203bb3cb2f2488c4a2cdabc324321755a4c8d7 the behaviour when no clientId is available in the request changed. It ends up being an empty string instead of the default client id. This PR fixes this behaviour.


chfsx:
Thanks a lot!

PR #4031, opened 2022-02-11
Fix issue is no HTTP_COOKIE key is available in $_SERVER

pascalseeland:
The $server["HTTP_COOKIE"] is not guaranteed to be available in PHP.  On my system, this is not the case and fails the script.


klees (approved):
LGTM, thx!

smeyer-ilias:
Thanks, LGTM.
This is only a PHP8 issue, so no merge to older branches is required. If not please write a comment.

PR #4033, opened 2022-02-13
KS Audio Element

alex40724:
KS Audio Element

Should replace the legacy implementation of audio player rendering, e.g. in mediacasts.

Related FW entry:
https://docu.ilias.de/goto_docu_wiki_wpage_6318_1357.html

Rules:
https://github.com/ILIAS-eLearning/ILIAS/pull/4033/files#diff-1553aa024cb63fc5c043630c7da136bae9b46dd335b1628dcf58a0b00c580eac

Bildschirmfoto 2022-02-14 um 15 33 10

Bildschirmfoto 2022-02-14 um 15 33 47


Amstutz:
Hi @alex40724 ,

Thx for your patience,
@klees and @Amstutz

alex40724:

matthiaskunkel:

klees (changes requested):
Hi @akill,

thanks a lot for adding this component to the UI framework. Besides some minor remarks (see below) the interface to the component looks fine and can be presented on the JF. Note that some comments tackle the interface, others the implementation. However, some things about the implementation are yet missing (such as tests). As far as we understand, this is intended and a further review of the complete implementation will be needed.

- [ ] IRSS: Any idea how we would connect this component to the ILIAS Resource Storage Service in the future?
- [ ] Mutators: Why do we need a `withSource`-mutator? Would we expect the source to be exchanged for a given audio-element? Same for `withTranscript`.
- [ ] A11y: Is there any best practice to provide transcriptions for screenreaders (etc)? Can we write it down as a rule and implement it?
- [ ] Binary File: In the example is a binary file. Is this licensed correctly? Could this be subjects to change? Would it be possible to maybe link an external file to keep this out of our source? It is not big, so we are not sure if this is really necessary.

- [ ] Lang-Vars: Please move them to the `ui`-namespace instead of using `common`.
- [ ] Player: Let's group `Video` and `Audio` into a `Player` category. They even might be eligible for a shared interface (`withSource`...) or shared descriptions ("have play\pause-buttons"...).
- [ ] Modal/Button: Currently you neither use the modal nor the button from the UI framework but instead add them yourself in the template. This comes with various potential problems. Changing styling or template of the button or the modal will not work as expected with this component. If we understand correctly, the modal will work because the according code will be bound to it via the class attribute. What if we decide to implement the modal differently? There might be something we are missing here, but please just use the according UI components to create the button and the modal.
- [ ] Less Color: You introduce a new less color. Is this necessary? Is there none, that could be re-used for this. Note that Each new color brings a new set of possible contrast issues which again will lead to failing a11y tests.

- [ ] License: Please format the license info in the headers of the file according to [the template defined in the Jour Fixe](https://docu.ilias.de/goto_docu_wiki_wpage_7116_1357.html).
- [ ] Docu 1: Please clarify: What is the expected "path" to the "source"? Is it just a path on the filesystem? Or is it an URL? Will it need to be a path in the webfolder? Maybe an additional `on the filesystem` will already do.
- [ ] Transcript: The transcript is said to [be optional here](https://github.com/ILIAS-eLearning/ILIAS/pull/4033/files#diff-1553aa024cb63fc5c043630c7da136bae9b46dd335b1628dcf58a0b00c580eacR82), but [required there](https://github.com/ILIAS-eLearning/ILIAS/pull/4033/files#diff-1553aa024cb63fc5c043630c7da136bae9b46dd335b1628dcf58a0b00c580eacR107). Please make it optional in the constructor. We first thought about removing it from the constructor completely and make it a mutator only. But: The source and the transcript have an intimate relation, the transcript belongs to a certain audio. So it makes sense to set them together. The mutators are either not necessary (see "Mutators" above) or we would want to bundle the source and the transcript in the same manner, like `withSource(string $path, ?string $transcription = null)`.
- [ ] Docstrings: Remove docstrings that do not provide additional information but merely repeat stuff already told by the code.
- [ ] checkStringArg: This method has become obsolete mostly, thanks to typehints. Just remove it where it does not do more then typehints.
- [ ] Tests: are missing. Please add them.

Interesting transcript in the example, by the way 🙂

If there is anything you want to discuss or ask, feel free to contact us on short notice.

Best regards!
@Amstutz and @klees

alex40724:
@Amstutz @klees 

Thanks for the feedback. I have

- removed the mutators,
- added a link to the relevant accessibility guideline for the transcriptions
- move the lang var to the ui namespace
- added a parent Player element for the Audio component
- switched the rendering to the existing Modal/Button implementation
- switched the color to an existing variable
- modified the documentation about the source parameter to be hopefully clear enough
- made the transcript optional
- removed docstrings
- removed checkStringArg
- added tests

I consulted the current IRSS documentation and don’t have a clear idea how the integration should look like. Either the resource returns the path passed to the KS or the KS accepts a resource as parameter in the future. There are other alternatives. It should be streamlined with the Image component.

The example binary files will be removed. I asked the PM to provide example files on ilias.de that can be referenced without adding them to the code repo.

matthiaskunkel:
Jour Fixe, 07 MAR 2022 : We highly appreciate this suggestion and accept the PR for trunk.

yvseiler:
Following Stephan Winiker's contribution to the discussion at the last JourFixe (07 March 2022) on the question of whether the modal is the appropriate place for the transcript, I have looked into this question from a UI/UX point of view and I like to provide hints and suggestions here. 

I am aware that the discussions at the JF have already taken place and that the commentary on this is rather late. Sorry for that. At best, this could be addressed in a next adaptation step, if the inputs for the current implementation no longer fit in terms of time.

**1. The good news: The modal allows access to the transcript as such**. This makes it accessible for people who only want to access the transcript (e.g. for people who are deaf, are hard of hearing, have difficulty processing auditory information, and others). [W3C](https://www.w3.org/WAI/media/av/transcripts/) does not give a direct rule on where the transcript must be placed. Their comments are: "Make it is easy for users to know that a transcript is available and to get to the transcript." and "For media on your website, usually it's best to include the transcript on the same page.". This is given with the modal.
**2. However: The decoupling from the actual audio file makes it impossible to use the audio track at the same time as reading along in the transcript.** For people who find it difficult to record what they hear with audio only, it can be helpful to use audio and transcript in parallel. For example, to control the audio track (pause, play, fast forward or rewind) and read along in the text at the same time. If the transcript were presented in parallel with the audio file, it would also be possible to help people for whom parallel use offers advantages.
**3. What happens if the audio has already started?** With the current solution with the modal, it is currently not clear to me whether the audio track continues to run in the background while I am viewing the transcript or whether the audio track pauses at that moment. If the audio track is running in the background, it can take a relatively long time until the audio can be paused (first close the modal, then click the pause button in the audio track).
**4. An alternative to the modal** that would allow simultaneous use of the audio track and the transcript could be a kind of accordion that unfolds when you click on "Transcript" and displays the transcript below the audio track. Or that the audio track is also made accessible in the modal.

These are my few points on the open question from the last JourFixe. As I said, I am aware that these comments come rather late and that the PR would have been ready for such inputs for some time. Nevertheless, I hope that one or the other hint may still help with the implementation.

Best regards
Yvonne

kergomard:
Hi @alex40724 , hi all

As promised we discussed the transcript modal in our consultancy meeting with Amanda from ADWebKey:
* From a usability point of view the solution is far from ideal (see also Yvonne's comment above). In addition to the points listed in her comment, scrolling can become a huge issue. The text can get really long and for people with some dexterity issue it might become impossible to move to the right place.
* On the other hand: The solution may pass a WCAG testing, ...provided it allows to have titles in the text that provide a convenient way to jump in the text, if for example one needs to close the modal.

Best,
Stephan

alex40724:
@yvseiler @swiniker Thanks for the feedback. I am not sure, if I understand the reasoning.

Why is scrolling in a modal a bigger issue than in an accordion like solution?

The modal decouples the text presentation from the context of the audio element, e.g.:

Imagine a multi-column layout, either by using the page editor, or by opening the menu slate and having a right column.

Now an audio element is within this structure. The layout will heavily limit the space of the audio element
| menu | slate | main content with audio | right column |
or even in a page with three columns:
| menu | slate | colum 1 | column 2 with audio | column 3 |

If the transcript is now presented underneath the audio element it will suffer from the same space limitations. Scrolling will be even longer. If users increase the font size things get even worse.

And scrolling will not only scroll the transcript, but also the other columns (if you don't nest scollable areas, wich is usually mayhem).

On the other hand, the modal allows us to use the full space (we could even define a full width modal). Font size could be increased to a wider degree, the transcript does not compete agains the content of other columns.

Could you elaborate how the accordion in content alternative would improve scrolling?

Maybe we should implement both options in the future. But I definitely see use cases, where we cannot assume a wide space context (like page editor content).

Best, Alex

kergomard:
Dear @alex40724

Where do your read this? As far as I can see @yvseiler doesn't say anything about scrolling and I don't say anything about an accordion.

The scrolling issue is an additional issue, ...but yes, even if the accordion would not solve it by itself, it looses a little bit of importance, if you don't have to constantly jump from the player to the transcript in a modal back to the player.

The base issue is described by @yvseiler:
* It is a usability and accessibility issue if the player cannot be controlled during reading.

And, though @yvseiler doesn't state that:
This solution also makes it hard to use the rest of the site. If e.g. your reading/listening reminds you of something that came up a few sessions ago, you can't simply navigate through ILIAS (other then by opening a new window, entering the url and so on) and have a look at it, while you keep your position in the text and maybe stop the video.

The scrolling issue would be solved, by pushing for anchored chapters in the text, as stated in my previous comment. You state that "we could even create a full screen modal". Yes we could also do that. This would only be a good approach though if we believed it to be the right path to move forward on. After the different talks we had on this, I believe it would be more fruitful to solve the problems you state for the accordion or to look for yet another solution. Why not e.g. introduce a way to expand the player to use the whole of the content area, thus also giving the transcript in the accordion the place it needs? The problems with the modal are from where I stand conceptual and fundamental.

Best,
Stephan

alex40724:
Yes, my points relate to both, your criteria (scrollability) and Yvonnes alternatives.

We have three conrete suggestions:

a) Modal as given in the PR
b) Modal with integrated player
c) accordion in content area.

> Why not e.g. introduce a way to expand the player to use the whole of the content area, thus also giving the transcript in the accordion the place it needs?

But how? We would need a concrete proposal how to do this in c). The player already uses the full width of the available space. The transcript would too. However the content area competes e.g. with other columns (see example). I don't see any way to solve this. Should the other columns suddenly be hidden (strange and almost a modal like behaviour). Or should they be pushed aside (horizantal scrolling)? See, https://test6.ilias.de/goto_test6_pg_165713_66809.html

Bildschirmfoto 2022-03-21 um 19 55 25

I don't see how c) could improve the space situation.

b) seems to be the best option I can see, especially for page editor cases. So a) seems to be acceptable as a first step, since it can be easily extended to b) in the future. The modal could also easily hold heading/anchor/text structures in the future. But we don't have these structures in the KS yet. And they most probably should be an independent concept imo.

alex40724:
@klees @Amstutz The binary files has been removed. As far as I can see I adressed all points mentioned by @klees above.

Is this ready to be merged? I could merge and take care of the conflicts. I would not like trying to rebase this, since I messed up PRs this way in the past and a new PR would lose the discussion.

I would suggest to put the alternative b) discussed above (integrate the player on top of the modal) to the ROADMAP.

The issue "long texts need headers/anchors" should be moved to a separate discussion, since this is nothing that should be solved for audio transcripts only. E.g. the user agreements or other longer texts in ILIAS may suffer from similar issues. This might need a separate KS component that can be later integrated here, too.

yvseiler:
I would like to respond to your [questions from above](https://github.com/ILIAS-eLearning/ILIAS/pull/4033#issuecomment-1074226593) @alex40724. Thank you for your example showing a possible use of an audio file within a page!
Even though this is not a very readable example because of the small line length on smaller screens, I know that users create something like this as long as they have the opportunity to do so.

> Could you elaborate how the accordion in content alternative would improve scrolling?

An accordion like we have in the editor now may not be the right solution for the transcript yet. By "a kind of accordion" I meant something that is similar to the behaviour of an accordion, something like "on click it shows more of what was hidden before". 

Maybe it's more along the lines of "click on button and a previously hidden panel slides in" (as seen on [TED](https://help.ted.com/hc/en-us/articles/360018572954-How-do-I-find-transcripts-for-TED-and-TEDx-talks-) or [accessibleweb.com](https://accessibleweb.com/tips-tools-tutorials/assistive-technology-focus-captioning-transcripts-and-audio-descriptions/#:~:text=off%2Dscreen%20voices.-,Examples,-of%20Transcripts%20in).
Prototypical experiments would probably be helpful to find a suitable container inside the content page.

I see that we do not yet have a concrete solution here for c). b) is certainly better than a).

> Why is scrolling in a modal a bigger issue than in an accordion like solution?

We need to avoid difficult scrolling behaviour where possible, I agree with both of you. 
Scrolling in and of itself we cannot prevent, I assume transcribed text is long. Other things should help here in the future, such as the linked chapters mentioned by Stephan Winiker or perhaps an autoscroll.

> On the other hand, the modal allows us to use the full space (we could even define a full width modal). 

For readability, in principle a non-page-filling line length is even better. For desktop, depending on the source, approx. 80 characters per line is recommended. Even in a modal that gives enough space to the audio and the transcript, care should be taken that the line length of the transcript is within this range to ensure good readability.

Best,
Yvonne

alex40724:
@yvseiler Thanks for the TED player example. I think this is definitely a possible way for a next step in the mediacast view, especially for the video version. The WebVTT format pretty much has the structure of the transcript view in the TED player. I also think that the audio view could move in a similar direction. At least we could make use of WebVTT file support here in the future, too.

The advantage of the mediacast is that we have full control of the whole view. This is not the case in page editor content. So personally I think b) is something we need. You are right about the Modal content width. My main point is, that there is no competition with other content in the Modal, so we can give it any responsive widths we want.

Amstutz (approved):
Hi @akill

Thx a lot for working in all the feedback above. Also, thx for checking in with @matthiaskunkel to find a place for the MP3 outside the repo. Further, nice job with the rendering test by only checking if key-parts are available. This makes this much less prone to issue not related to this component.

We see that there has been a lot of discussion around further improving this component. Thx at @yvseiler , @swiniker for the feedback and @alex40724 for the replies and questions back. The discussion seems to make clear, that there is room for improvements, from a UX/A11Y perspective. Thx for the offer to put a short summary of that into the roadmap. We believe this to be a good way forward. Since we do believe that this first iteration is a major improvement to the current version, we agree, that this can be moved to the next iteration with this component. 

So, things left to-do:
- [ ] Roadmap: Create the entry.
- [ ] Conflicts: Feel free to resolve them while merging. We see your point on not losing the discussion.

Thx a lot to everybody involved.  
@klees and @Amstutz

alex40724:
@Amstutz @klees Thanks for the feedback. I added an entry to the roadmap, see https://github.com/ILIAS-eLearning/ILIAS/pull/4033/files#diff-c90cf75d39f6e34d3d80e97eca92d37018d67e3c5220d81e61887e645718ee4d

If there are no objections, I will merge this to the trunk next week.

Amstutz:
No objections, thx. Can be merged as soon as conflicts are resolved.

alex40724:
See https://github.com/ILIAS-eLearning/ILIAS/commit/2918bae7e769602f44f19ad7f2553bc1ddf4c91c

PR #4034, opened 2022-02-14
KS Item with Audio Element

alex40724:
This PR depends on https://github.com/ILIAS-eLearning/ILIAS/pull/4033

It contains all changes of the base PR. Relevant change in the item element:
https://github.com/ILIAS-eLearning/ILIAS/pull/4034/files#diff-b84dddcb4040f0309648a00e52da9ede0b8c05dd33ac7aa7ba4cc02e9ab607a5

This change allows to add an audio element to an item, see discussion in the related FW article:

Related FW entry:
https://docu.ilias.de/goto_docu_wiki_wpage_6318_1357.html

Bildschirmfoto 2022-02-14 um 15 02 14


Amstutz:
Hi @alex40724 ,

Thx for your patience,
@klees and @Amstutz

alex40724:

matthiaskunkel:

Amstutz:
Hi @akill,

thanks for the effort to move the Media Cast to components from the UI framework.

We certainly agree that this solution is better than the one that the [JF initialy appreciated](https://docu.ilias.de/goto_docu_wiki_wpage_6318_1357.html). Still we are not very happy with this solution for two main reasons:

* The `Item` is further cluttered with additional properties. It already feels like a loose bag of various properties, with an use case that is hardly defined and a structure that is barely perceptible. This is not something that this change will introduce, but it certainly will get worse with this change. This state of affair is bad because it makes it nearly impossible to style the item appropriately for different screen sizes or process the information contained in the item for other purposes. Questions like "how important is the audio, do we need to display it directly or could we hide it under an accordion" simply cannot be answered. The way the `Item` and the `Audio Player` are used together exemplifies this: The `"Length"` that is added to the `Item` in the example is a property of the mp3 referenced in the `Audio Player`. Still we have no way to connect these two together once they are poured into the `Item`. How will we every know that the `"Length"` is more important to a user then another random property that is slapped onto the `Item` as well? How will we ever be able to consistently style different `Items->withAudio` in the same list to, e.g., highlight the length?
* The way the `Item->withAudio` is used in the feature requests invokes further questions. Can or should I run multiple audios on the page? How will I be able to grasp quickly which audio is playing on a page with many of these elements? Also, users might expect that the next audio starts playing when one is finished. This won't happen, right? It looks as if the thing we would really want is something similar to Spotify: An integrated view, where we have one audio playing and a list of other tracks that the user can pick from. Along with a UI that is optimized for this use case. This would simplify the problems outlined in the previous paragraph as well.

We understand, though, that this is something that is a little harder to achieve than the change proposed here. For the sake of introducing UI components to the Media Cast at all, we will accept this change (with the change requests below).

In general, this PR shows why we urge everyone to work on UI components before working on the feature requests using them. The initial idea (`withLeadAudio`) was discarded and the new idea (`withAudio`) is also problematic. The new solution will have a questionable usability, but the short time to the release date does not give any room to improve the state of affairs, even if someone would have funding. Hence, we will most likely need to stick to this imperfect solution for some time, because there will be new requests to work on in the next release. The Media Cast will continue to look and feel somewhat odd with this view not much optimized for audio and video playing for an indefinite amount of time. Maybe we can improve this by starting out with taking a deep dive into the necessary new UI composes and/adjustments to existing ones before finalizing and deciding on the feature itself in future.

Please implement the following change:
* [ ] withAudio: change it to `withAudioPlayer`.

Best regards!
@Amstutz and @klees

matthiaskunkel:
Jour Fixe, 07 MAR 2022 : We highly appreciate this suggestion and accept it for trunk.

Amstutz:
We wait here for the final review of the implementation, until https://github.com/ILIAS-eLearning/ILIAS/pull/4033 is merged, since those are depended, plz ping us, as soon as you believe this to be ready for the final review.

alex40724:
@Amstutz @klees Follow-up PR is here: https://github.com/ILIAS-eLearning/ILIAS/pull/4305

PR #4041, opened 2022-02-15
KS Video Element

alex40724:
KS Video Element

Should replace the legacy implementation of video player rendering, e.g. in mediacasts.

Related FW entry:
https://docu.ilias.de/goto_docu_wiki_wpage_6210_1357.html

Rules:
https://github.com/ILIAS-eLearning/ILIAS/pull/4041/files#diff-1553aa024cb63fc5c043630c7da136bae9b46dd335b1628dcf58a0b00c580eac

Bildschirmfoto 2022-02-15 um 14 51 29


mjansenDatabay:
Just my two cents: The example files are quite large (~ 8,5 MB) and when merged, these will be part of every clone (https://robinwinslow.uk/dont-ever-commit-binary-files-to-git / https://vegeta2102.medium.com/why-you-should-not-push-binary-files-to-git-and-what-to-do-if-you-do-29f34692cef0). I guess there is a low probability that these will change in the future, but maybe we should use smaller binary files here.

Amstutz:
Hi @alex40724 ,

Thx for your patience,
@klees and @Amstutz

alex40724:

matthiaskunkel:

klees (changes requested):
Hi @akill,

thanks a lot for adding this component to the UI framework. Besides some minor remarks (see below) the interface to the component looks fine and can be presented on the JF. Note that some comments tackle the interface, others the implementation. However, some things about the implementation are yet missing (such as tests). As far as we understand, this is intended and a further review of the complete implementation will be needed.

- [ ] IRSS: Any idea how we would connect this component to the ILIAS Resource Storage Service in the future?
- [ ] Mutators: Why do we need a `withSource`-mutator? Would we expect the source to be exchanged for a given `Video Player`?
- [ ] Binary File: In the example is a binary file. Is this licensed correctly? Could this be subjects to change? Would it be possible to maybe link an external file to keep this out of our source? Other than the mp3 in the `Audio Player`, this file is rather large... Could we use a smaller one if we can't link an external one?
- [ ] Container: Why do we need the additional `div` around the `video`-element?

- [ ] Player: Let's group `Video` and `Audio` into a `Player` category. They even might be eligible for a shared interface (`withSource`...) or shared descriptions ("have play\pause-buttons"...).
- [ ] A11y: Consider adding an alternative audio source as additional property, so consumers have a way to implement [WCAG success criterion "Audio-only and Video-only (Prerecorded)"](https://www.w3.org/WAI/WCAG21/quickref/#audio-only-and-video-only-prerecorded). This also should be another rule then.

- [ ] License: Please format the license info in the headers of the file according to [the template defined in the Jour Fixe](https://docu.ilias.de/goto_docu_wiki_wpage_7116_1357.html).
- [ ] Docu 1: Please clarify: What is the expected "path" to the "source"? Is it just a path on the filesystem? Or is it an URL like for videos hosted on YouTube or Vimeo?
- [ ] Docu 2: Please clarify: What is the expected value for "$poster" in [Video::withPoster](https://github.com/ILIAS-eLearning/ILIAS/pull/4041/files#diff-a48c4c12fcbae827bf18f413afef4b4f10ea9aac2a238cc46944b45d694e2c20R51)? Is it just a path on the filesystem? Or is it an URL?
- [ ] Docstrings: Remove docstrings that do not provide additional information but merely repeat stuff already told by the code.
- [ ] checkStringArg: This method has become obsolete mostly, thanks to typehints. Just remove it where it does not do more then typehints.
- [ ] Tests: are missing. Please add them.

If there is anything you want to discuss or ask, feel free to contact us on short notice.

Best regards!
@Amstutz and @klees

matthiaskunkel:
Jour Fixe, 07 MAR 2022 : We highly appreciate this suggestion and accept the PR for trunk. Please adapt the suggested changes by the UI coordinators as already done in the PR for KS Audio Element.

matthiaskunkel:
I have created a new subdomain files.ilias.de for the purpose of reference files and added two files:
• ILIAS-Video.mp4
• ILIAS-Audio.mp3
They can be used for the Kitchen Sink elements. If more file types are necessary, please let me know.

alex40724:
Follow-up https://github.com/ILIAS-eLearning/ILIAS/pull/4306

PR #4046, opened 2022-02-15
MyStaff: Fix query when determining courses of user(s)

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=32073


mstuder:
thank you - i will merge this, this week also for ilias7 and trunk!

PR #4050, opened 2022-02-16
User/AccessControl: Fix wrong type in permanentlink (Mantis 32075)

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=32075


smeyer-ilias:
Thanks, for fixing this

PR #4051, opened 2022-02-16
HTTP: Change `Drop-In-Replacement` handling

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=32081

This PR changes the behaviour of the `Drop-In-Replacement` used for `$_GET`, `$_POST`, `$_COOKIE` and `$_REQUEST`. We noticed issues with this when relying on external libraries not expecting an `ArrayAccess` implementation, but a real array.

We suggest to ...

- throw an exception in the `Drop-In-Replacement` if a value of the respective super global variable is manipulated in `DEVMODE`
- add an additional `INI` setting (see comments) to disable the `Drop-In-Replacement`

### DIFF
--- a/src/HTTP/Wrapper/SuperGlobalDropInReplacement.php
+++ b/src/HTTP/Wrapper/SuperGlobalDropInReplacement.php
@@ -41,6 +41,8 @@ class SuperGlobalDropInReplacement extends KeyValueAccess
         if ($this->throwOnValueAssignment) {
             throw new \OutOfBoundsException("Modifying global Request-Array such as \$_GET is not allowed!");
         }
+
+        parent::offsetSet($offset, $value);
     }
 
     /**
--- a/src/HTTP/Wrapper/SuperGlobalDropInReplacement.php
+++ b/src/HTTP/Wrapper/SuperGlobalDropInReplacement.php
@@ -35,14 +35,11 @@ class SuperGlobalDropInReplacement extends KeyValueAccess
     }
 
     /**
-     * @deprecated Please note that this will throw an exception in a future version
      * @inheritDoc
      */
     public function offsetSet($offset, $value) : void
     {
-        /** @noRector */
-        // this is currently possible, throw new \OutOfBoundsException("Modifying global Request-Array such as \$_GET is not allowed!"); if you want to prevent from overriding a value here
-        parent::offsetSet($offset, $value);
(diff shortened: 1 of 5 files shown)


mjansenDatabay:
@klees / @chfsx What did we say regarding the suggested 'Ini' setting? Should the setting disable or enable the `Drop-In-Replacement` (especially in conjunction with the `DEVMODE`, and what is the default behaviour? I can't remember (by far to little sleep ;-)).

I'll add a corresponding commit afterwards and present this on the next `JourFixe`.

chfsx:
@mjansenDatabay We had discussed that we would introduce the following three variants:

- with DEVMODE enabled: exception if a value is changed. presumably we would have to take this via the constructor into the SuperGlobalDropInReplacement.
- Default behavior: As before (therefore the new parameter on SuperGlobalDropInReplacement)
- With an ini-setting the replacing of the SuperGlobals by SuperGlobalDropInReplacement can be deactivated, then remains for example $_GET as it is

src/HTTP/Wrapper/SuperGlobalDropInReplacement.php
```diff
@@ -25,24 +25,22 @@
...
      * @inheritDoc
      */
     public function offsetSet($offset, $value) : void
     {
-        /** @noRector */
-        // this is currently possible, throw new \OutOfBoundsException("Modifying global Request-Array such as \$_GET is not allowed!"); if you want to prevent from overriding a value here
-        parent::offsetSet($offset, $value);
+        if ($this->throwOnValueAssignment) {
+            throw new \OutOfBoundsException("Modifying global Request-Array such as \$_GET is not allowed!");
+        }
```
chfsx, line 43: I think the `parent::offsetSet($offset, $value);` is needed if `$this->throwOnValueAssignment === false`

mjansenDatabay: Yep

chfsx (approved):

matthiaskunkel:
Jour Fixe, 21 FEB 2022 : We highly appreciate this suggestion and accept the PR for ILIAS 8.

mjansenDatabay:
@chfsx I suggest to add an additional information to the `src/HTTP/README.md` file, okay?

chfsx:
@mjansenDatabay see https://github.com/mjansenDatabay/ILIAS/pull/13

chfsx:
Thanks @mjansenDatabay !

PR #4063, opened 2022-02-18
GlobalScreen/Layout: Add named location for displaying `Toasts`

iszmais:
Depends on the implementation of [Toast Container](https://github.com/ILIAS-eLearning/ILIAS/pull/4059) since its the first overlay content.

### DIFF
--- a/src/GlobalScreen/Scope/Layout/Builder/StandardPageBuilder.php
+++ b/src/GlobalScreen/Scope/Layout/Builder/StandardPageBuilder.php
@@ -41,10 +41,10 @@ class StandardPageBuilder implements PageBuilder
      */
     public function build(PagePartProvider $parts) : Page
     {
-        $header_image = $parts->getLogo();
-        $main_bar = $parts->getMainBar();
         $meta_bar = $parts->getMetaBar();
+        $main_bar = $parts->getMainBar();
         $bread_crumbs = $parts->getBreadCrumbs();
+        $header_image = $parts->getLogo();
         $footer = $parts->getFooter();
         $title = $parts->getTitle();
         $short_title = $parts->getShortTitle();
@@ -56,6 +56,7 @@ class StandardPageBuilder implements PageBuilder
             $main_bar,
             $bread_crumbs,
             $header_image,
+            $this->ui->factory()->toast()->container(),
             $footer,
             $title,
             $short_title,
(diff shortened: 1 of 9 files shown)


chfsx (changes requested):
Hi @iszmais and @klees 
Thank you for this contribution, I know the concept and the feature only superficially but think that this will be a valuable extension of the `Page`.

What is currently unclear to me is the function of the GlobalScreen in this scenario. i see that the `PagePartProvider` has been extended that it can also return a `container` (or `null`). However, the `StandardPagePartProvider` always returns a `container`.

Are there cases where
a) there is no `Container`?
b) any `LayoutProvider` of a component wants to provide another `Container` than the default one?

M.e. there is no need for the extension, but as I said, I may not know the requirements well enough. I think there are the following options:
1. if a `page` always contains a `container` anyway, this could be inserted directly when rendering the page
2. alternatively the `StandardPageBuilder` in `build()` could simply add a container to the page without getting it from the PagePartProvider.
3. if really the `container` can be contributed by the GlobalScreen, I would have to understand better the idea and the future behind the `container`: Will there be other `Containers` besides the `ToastContainer`, which should possibly also work in parallel side by side? In the case of toasts, who would now have the task of providing a container? What about multiple containers of the same kind?

I suppose that at the moment we should just go variant 2.

chfsx:
BTW: Tests seem broken ATM

iszmais:
@chfsx 

Yeah this should still be draft. Seems i had a littl misscommunication with @mjansenDatabay here :sweat_smile: 

But i will integrate your feature before i finish this. 

Greetings,
@iszmais

iszmais:
Hi @chfsx ,

Thanks fpr your feedback!
I agree in you major point. Guess this is better inside the builder as the provider in view of future expandability.
I changed that and implemented/epanded the test cases.
I additionally fixed the spacing inside of the html-blocks of the tests to be more similar to the real template, since else this caused and may cause minor testing troubles in the future (even if it is with brutallyTrimmed because this doesnt seem to match all spacing matches).

Fell free to give me another feedback or merge (after JF approval).

Greetings,
@iszmais

matthiaskunkel:
Jour Fixe, 07 MAR 2022 : We highly appreciate this suggestion and accept the PR for trunk.

chfsx (approved):
Thanks a lot for the changes and your explanation in JF. from the GS-side everything is fine, I'll let the merge up to you since this PR has dependencies and the changes in GS are trivial.

iszmais:
All dependencies are merged

PR #4066, opened 2022-02-21
Fix logging. Check if classname is set. Fixes #32107

pascalseeland:
The classname is not set when php internal class are contained in the stacktrace. This happens if logging is used in callback functions.


smeyer-ilias:
Thanks for fixing this.
Inspired by your PR I just commited a fix which also tackles other potential issues

PR #4067, opened 2022-02-22
[FIX] 0031865: Online Help Main Menu Tooltips are missing (was: Place…

chfsx:
… holders are missing)

The UI `Icon` is `JavaScriptBindable` but did not implement an ID in its template, this causes e.g. https://mantis.ilias.de/view.php?id=31865


Amstutz:
Hi @chfsx ,

Thx for reporting and providing a fix for this. We are not sure, if the issue is not rather the JavascriptBindable on the public interface for ALL symbols. However it is there, therefore it should work. 

- [x] One minor thing, can you provide a short test for the standard icon making sure, that the rendering of the ID will not be lost again in future. Thx!

Kudos.

Amstutz:
Thx a lot! picked to trunk as well.

PR #4069, opened 2022-02-22
Fix issue that headers must contain 2 parts

pascalseeland:
The previous PR contained a wrong check for the splitting of the header. It will never be more than 2 elements from explode, but one also gets passed the HTTP/1.1 200 OK as header, which cannot be split and is not a proper header.


smeyer-ilias:
I will not merge this to 6 and 7, since no concrete bug was produced by the old implementation, correct?
The only consumer of the header array is class ilECSConnector  (calls of getResponseHeaderArray())

PR #4075, opened 2022-02-22
Bugfix 0032114

ezenzen:
Hi @all,

to increase the contrast, I have adjusted the border colors of inputs. Since the number of colors used in ILIAS should not diffuse apart, I have chosen in consultation with Yvonne already used and low barrier color combinations.

il-main-border-dark-color is changed to #757575 . The inputs on the login page (red) use the variable @il-danger-color. 

0032114: Nicht-Text-Kontrast, Various Inputs - https://mantis.ilias.de/view.php?id=32114

Greetings,
Enrico

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -7223,7 +7223,7 @@ button.close {
 }
 .il-upload-file-item.standard {
   background: white;
-  border: 1px solid #A8A8A8;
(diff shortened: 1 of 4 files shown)


Amstutz:
Thx a lot!

Amstutz:
picked to trunk as well.

PR #4077, opened 2022-02-23
Fix for 31918: remove the function archiveParticipantSubmission() 

okaufman:
This removes the function archiveParticipantSubmission because it is no longer called anywhere in ilias.
Fixes https://mantis.ilias.de/view.php?id=31918


mbecker-databay:
I am happy this was found. But ... strictly speaking it's a non-issue and solely cleanup/refactoring request.
While we're all trying to leave the code around bugfixes cleaner than we found it, we try to not do such cleanups in already released versions.

@okaufman, please re-do this for the trunk version.

Amstutz:
As @mbecker-databay said, this should target trunk, since this is refactoring only.

PR #4079, opened 2022-02-23
Centralized lastAdmin validation inside the membership deletion

iszmais:
When you unsunscribe a user from multiple objects, the objects are not checked on if the user is the last admin of the object since these validation is only accessable for the single unsubscription.

This PR resolves the issue by centralizing the validation inside the deletion process instead of the confirm process.

Mantis Issue: https://mantis.ilias.de/view.php?id=29899

### DIFF
--- a/Services/Membership/classes/class.ilParticipants.php
+++ b/Services/Membership/classes/class.ilParticipants.php
@@ -9,6 +9,8 @@
 * @ingroup ServicesMembership
 */
 
+use ILIAS\Data\Result;
+
 define("IL_CRS_ADMIN", 1);
 define("IL_CRS_TUTOR", 3);
 define("IL_CRS_MEMBER", 2);
@@ -28,6 +30,8 @@ define("IL_ROLE_POSITION_MEMBER", 3);
 
 abstract class ilParticipants
 {
+    public const ERROR_LAST_ADMIN = "last admin";
+
     protected $component = '';
     
     protected $obj_id = 0;
@@ -806,15 +810,13 @@ abstract class ilParticipants
         }
         return false;
     }
-    
+
     /**
      * Drop user from all roles
      *
-     * @access public
-     * @param int usr_id
-     *
+     * @param int $a_usr_id
      */
-    public function delete($a_usr_id)
+    public function delete($a_usr_id): bool
     {
         global $DIC;
 
@@ -844,6 +846,18 @@ abstract class ilParticipants
         );
         
         return true;
+    }
+
+    private function unsubscribe(int $usr_id): Result
+    {
+        if (!$this->checkLastAdmin([$usr_id])) {
+            return new Result\Error(self::ERROR_LAST_ADMIN);
+        }
+        try{
+            return new Result\Ok($this->delete($usr_id));
+        } catch (Exception $e) {
+            return new Result\Error($e);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Services/Membership/classes/class.ilParticipants.php
```diff
@@ -823,6 +823,12 @@ public function delete($a_usr_id)
         $rbacadmin = $DIC['rbacadmin'];
         $ilDB = $DIC['ilDB'];
 
+        if (!$this->checkLastAdmin([$a_usr_id])) {
+            $DIC->language()->loadLanguageModule($this->getType());
+            ilUtil::sendFailure($this->lng->txt($this->getType() . '_at_least_one_admin'), true);
```
smeyer-ilias: This introduces a dependency to the global template interface (trunk ilUtil::sendInfo replacement) in a non GUI-class.
Throwing an exception would be more appropiate here. 

Nevertheless even if an exception would be thrown here, I wouldn't introduce it in R7, because all consumers of ilParticipants:delete would have to be adapted/checked whether they handle the exception correctly.  E.g:

- SOAP calls with ilCourse|GroupXMLParser
- ECS member de/assignments have to be adapted. Otherwise they will fail. 
- maybe more

smeyer-ilias (changes requested):
Thanks for the PR. 
For the reasons mentioned in ilParticipants, I would be prefer a solution, which checks this contraint in the dashboard context.

Services/Membership/classes/class.ilParticipants.php
iszmais: Totally agree on that one, i missed that sideeffect in my concept.

Therefore i abstracted the response so the Error can be handled inside the GUI due to a return value of the progress

iszmais:
Thx for the quick response @smeyer-ilias !

I totally agree that that the Error handling and visualizing itself shouldnt be handled inside the ilParticipant. But i dont quite understand why the constraint shouldnt be checked there. Is there an exception to the concept of ILIAS participations, where admins are not required?

A updated the solution to handle the constraint more abstract. Would like to hear your opinion on this.

Greetings,
@iszmais

chfsx:
Hi @smeyer-ilias  

Best regards!

alex40724:
```
if (!$this->getMembersObject()->deleteParticipants($participants)) {
    ilUtil::sendFailure($this->lng->txt($this->getParentObject()->getType() . '_at_least_one_admin'), true);
```
The returned "false" should not represent this particular issue imo.

Final course/group deletion might be a process that needs deletion of all members regardless of this check, there might be other similar cases.

iszmais:
> ```
> if (!$this->getMembersObject()->deleteParticipants($participants)) {
>     ilUtil::sendFailure($this->lng->txt($this->getParentObject()->getType() . '_at_least_one_admin'), true);
> ```
> 
> The returned "false" should not represent this particular issue imo.
> 
> Final course/group deletion might be a process that needs deletion of all members regardless of this check, there might be other similar cases.

@alex40724 Thx for the hint!
I changed the process to be result based for better identification withing the target occurences.
I would still like this to be catnralized but totally agree that therefore it should be more detailed

(PS: in not fixed to the naming "unsibcribe" feel free to propose another, better fitting one).

Greetings,
@iszmais

iszmais:
This can be close since it was already altarnativöy solved

PR #4082, opened 2022-02-23
UI Link/MetaBar Language Selection: Support language for links (A11Y)

mjansenDatabay:
Depends on https://github.com/ILIAS-eLearning/ILIAS/pull/4653

This PR consists of two parts (and yes, maybe they could have been provided in two separate PRs).

1. The `UI Link` components have been extended to accept optional language attributes for the `Label` and for the content provided when the link `Action` is executed.
2. The `MetaBar` `Item` components of the `GlobalScreen` have also been been extended to accept optional language attributes for the `Title` (not named `Label` like in the UI components)  and for the content provided when the link `Action` is executed.

Concrete Usage: As a first usage I provided these language attributes for the `Language Selection` in the `Public Area`.

Why?

https://www.w3.org/WAI/GL/WCAG20/translation/instructions_html:
> The hreflang attribute indicates the language of content targeted by links. It is helpful though not required to add this information to links for which the target will not be translated in this process. If the link text also is not translated (e.g., because it is a formal title that should be kept in the original language), you should also add the language attributes to the anchor element.


klees (changes requested):
Hi @mjansenDatabay,

thanks a lot for investigating this issue about language and accessibility and proposing changes that are well grounded in standards.

- [x] Inheritance: Why do we need to repeat the methods `getLabelLanguage` and friends on `Bulky` and `Standard`? Can't we just add them to `Link`?

- [x] ActionContentLanguage: This 
- [x] LanguageTag: Add a data type `LanguageTag` (name debatable) that implements [BCP 47](https://www.ietf.org/rfc/bcp/bcp47.txt). Use this instead of plain strings in the interface.
- [x] Language interface: Please consider introducing an interface `Language` that holds the according methods and a trait that can be used to easily implement it for components. The [lang attribute](https://developer.mozilla.org/de/docs/Web/HTML/Global_attributes/lang) can be used on any HTML attribute, so we will introduce it for many components as well.
- [x] withoutLabelLanguage: Do we really need this?


- [x] Typo: `withouLabelLanguage` -> `withoutLabelLanguage` (if we really need this).
- [x] Naming 1: Please use `withLanguage` (or maybe, if this is too generic, `withContentLanguage`) for the methods on the interface that we could generalize (see "Language Interface" above). If I understand correctly, [lang](https://developer.mozilla.org/de/docs/Web/HTML/Global_attributes/lang) tells, which language we can expect from all stuff contained in the element we tag it on. It is a special case for the link that this is about the "label".
- [x] Naming 2: Please use `getLanguageOfReferencedRessource` (and friends) instead of `getActionContentLanguage` (and friends). If I understand the [docu](https://developer.mozilla.org/de/docs/Web/HTML/Element/a#attr-hreflang) correctly, this name fits better.

Feel free to contact us on short notice, if there is anything to discuss.

Best regards!
@Amstutz and @klees

mjansenDatabay:
Thanks a lot for your feedback @klees / @Amstutz . I was totally aware that there might be many remarks :-). It was just an idea that popped up yesterday evening :-).


@chfsx Anything to add related to the `GS`?

src/GlobalScreen/Scope/MetaBar/Factory/LinkItem.php
```diff
@@ -36,11 +38,12 @@ public function __construct(IdentificationInterface $provider_identification)
         $this->renderer = new LinkItemRenderer();
     }
     
-    public function withAction(string $action) : self
+    public function withAction(string $action, ?string $action_content_language = null) : self
```
chfsx: I'd like to have this as a separate method `withActionContentLanguage` since it's optional anyway. I see this is quite near to the actions itself but I do not like optional parameters and `Items` should be as simple as possible.

src/GlobalScreen/Scope/MetaBar/Factory/LinkItem.php
```diff
@@ -82,19 +90,28 @@ public function hasSymbol() : bool
     /**
      * @inheritDoc
      */
-    public function withTitle(string $title) : hasTitle
+    public function withTitle(string $title, ?string $title_language = null) : hasTitle
```
chfsx: Same here, I would rather have a  `withTitleLanguage(string $title_language)`

src/GlobalScreen/Scope/MetaBar/Factory/TopLegacyItem.php
```diff
@@ -65,11 +66,12 @@ public function hasSymbol() : bool
     /**
      * @inheritDoc
      */
-    public function withTitle(string $title) : hasTitle
+    public function withTitle(string $title, ?string $title_language = null) : hasTitle
```
chfsx: Some think here... I currently think of an Interface which defines those Methods at once. A trait implementeting them would reduce the duplicate code. I do not think all `hasTitle` items has the title_language?

src/GlobalScreen/Scope/MetaBar/Factory/hasTitle.php
```diff
@@ -6,8 +6,9 @@
  */
 interface hasTitle extends isItem
 {
-    public function withTitle(string $title) : hasTitle;
-    
+    public function withTitle(string $title, ?string $title_language = null) : hasTitle;
```
chfsx: I'm not sure: do all items `hasTitle` need the $title_langauge? if so, then define the new methods here, if not a new Interface would be helpful

chfsx: additionally, I think there is an Interface `hasAction` as well. the same question there.

klees:
@mjansenDatabay The idea is nice indeed =)

The sheer number of remarks does not relate to the validity of the idea, though. For bad ideas, the number of remarks might be just one: "Hell no!". I hope our numerous feedbacks here just help to make a good idea even better...

chfsx (changes requested):
Hi @mjansenDatabay 

Thank you very much! I made several inline-comments, here is a summary:

- [x] Do all `hasTitle` items in GS need the $title_language feature?
- [x] Do all `hasAction` items in GS need the $action_content_language feature?

- [x] After implementing new Interfaces (if needed), you could implement the new methods with a trait

- [x] Move those options to new methods (e.g. `withTitleLanguage`)
- [x] Move those new method to new interfaces, if they do not apply to all `hasTitle` or `hasAction`

Feel free to contact me at any time if there are questions

kindly,
@chfsx

mjansenDatabay:
@klees 

> Inheritance: Why do we need to repeat the methods getLabelLanguage and friends on Bulky and Standard? Can't we just add them to Link

Because I initially thought that this is strongly coupled to the `Label`, and the label related methods are duplicated in the link types. But of course the `Language` can be moved to the base interface (which I already did).

mjansenDatabay:
@chfsx

> Do all hasTitle items in GS need the $title_language feature?

I'll provide separate interfaces/traits.

> Do all hasAction items in GS need the $action_content_language feature?

I'll provide separate interfaces/traits.

mjansenDatabay:
TODO:

- [ ] `LanguageTag` data type
- [ ] Interfaces/Traits for the `UI Framework`

mjansenDatabay:
TODO:

- [x] `LanguageTag` data type (which will take a while, because the RFC has to be studied and implemented / **Current Status: We'll maybe have an option for funding**)
- [x] Interface/Trait for the `UI Framework`

Amstutz:
Just to make sure, that we are not stranded in a deadlock here, you are working on the LanguageTag, @mjansenDatabay , right? Feel free us to look over this if ready.

mjansenDatabay:
> Just to make sure, that we are not stranded in a deadlock here, you are working on the LanguageTag, @mjansenDatabay , right? Feel free us to look over this if ready.

Yes, I added the current status to my latest comment :-). We'll maybe have funding for the new data type, Oliver has to get in contact with the funding party.

-----------

Update: We scheduled the impl. of a `LanguageTag`.

mjansenDatabay:
The language tag will be added with this PR: https://github.com/ILIAS-eLearning/ILIAS/pull/4653

Once it got approved I will rebase this PR here and integrate the new data type.

lscharmer:
Updated branch to use the new `LanguageTag` data type.
@klees, @Amstutz, @chfsx As all open issues are resolved with this, please add your review for this PR.

src/GlobalScreen/Scope/MetaBar/Collector/Renderer/LinkItemRenderer.php
```diff
@@ -37,10 +38,20 @@ protected function getSpecificComponentForItem(isItem $item): Component
...
          * @var $item LinkItem
          */
-        return $this->ui->factory()->link()->bulky(
+        $link = $this->ui->factory()->link()->bulky(
             $this->getStandardSymbol($item),
             $item->getTitle(),
             $this->getURI($item->getAction())
         );
+
+        if ($item->getContentLanguage()) {
```
chfsx: Could we use a `!== null` comparisation here?

src/GlobalScreen/Scope/MetaBar/Collector/Renderer/LinkItemRenderer.php
```diff
@@ -37,10 +38,20 @@ protected function getSpecificComponentForItem(isItem $item): Component
...
+            $link = $link->withContentLanguage($item->getContentLanguage());
+        }
+
+        if ($item->getLanguageForTargetedResource()) {
```
chfsx: and here as well: Could we use a `!== null` comparisation?

src/UI/Implementation/Component/Link/Renderer.php
```diff
@@ -54,6 +54,12 @@ protected function setStandardVars(
         if ($component->getOpenInNewViewport()) {
             $tpl->touchBlock("open_in_new_viewport");
         }
+        if ($component->getContentLanguage()) {
```
chfsx: Not my component here, but I suggest a `!== null` condition as well

src/UI/Implementation/Component/Link/Renderer.php
```diff
@@ -54,6 +54,12 @@ protected function setStandardVars(
...
+            $tpl->setVariable("CONTENT_LANGUAGE", $component->getContentLanguage());
+        }
+        if ($component->getLanguageOfReferencedResource()) {
```
chfsx: and same here, `!== null`

chfsx (changes requested):
Hi! I made some comments concerning conditions, the whole rest seems fine to me

lscharmer:
@chfsx I changed the comparison.

chfsx (approved):
Thanks for adopting the comments!

mjansenDatabay:
@klees / @Amstutz Any objections left? @chfsx is happy (at least I hope so ;-)) and all review remarks from @klees have been addressed.

klees (approved):

klees:
@Amstutz volunteered to add an example: https://mantis.ilias.de/view.php?id=36252 Thx, y'all!

mjansenDatabay:
@klees If you did not begin, yet, you could ask @lscharmer.

PR #4083, opened 2022-02-25
WIP: Fix for Mantis #16075

PurHur:
Hello,

i think i have found the solution for a very old bug. https://mantis.ilias.de/view.php?id=16075
If this solution is correct the bug still exists in 7. I think even in the trunk. But i only have fixed it for a 5.4 installation now.

My test setup was a test with 2 allowed passes and a limited time of 60 minutes for each pass.

The learner didnt pass the first try. The tst_active table looks like this:
tst_active

`_getPass` returns the col `tries` which is then increased by one. That cant be right?! Then the check fails because `2 > 2` is false

Greetings
Purhur


klees:
"Off by one", classic... =D

PurHur:
You really dont want to get into that if branch... this makes things even worse. We handled the errors via DB fixes.

PR #4088, opened 2022-02-28
UI/Mainbar: 30738, dispatch resize-event on slate active/inactive

nhaagen:
https://mantis.ilias.de/view.php?id=30738


Amstutz (approved):
Hi @nhaagen 

Thx a lot for providing this fix! I had to check up on the meaning of "induced", learning something new today :-)

@klees, IMO this can be merged as soon as this is good to go from your pov.

PR #4092, opened 2022-03-01
il-padding-xlarge

ezenzen:
Hi @all,

to unify the paddings in ILIAS we need two new LESS variables. With this PR only the variables @il-padding-xlarge-vertical: 9px and @il-padding-xlarge-horizontal: 15px are introduced.

Since I compiled the less and color variables seem to have changed in the meantime, you will also find some color changes in this PR. But these are not part of my customizations. 

King regards, 
Enrico

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -2758,14 +2758,14 @@ output {
   box-shadow: inset 0 1px 1px rgba(0, 0, 0, .075), 0 0 8px rgba(129, 152, 183, 0.6);
 }
 .form-control::-moz-placeholder {
-  color: #7c7c7c;
+  color: #626262;
   opacity: 1;
 }
 .form-control:-ms-input-placeholder {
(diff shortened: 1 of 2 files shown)


Amstutz (changes requested):
Hi @ezenzen 

Looks good, thx a lot. However there is one change in there, that does not belong in this PR: https://github.com/ILIAS-eLearning/ILIAS/pull/4092/files#diff-596f8b5902f856b7544d5a57d929e0d7a80dc52ef671ec11f225fccb82392150R49 (@il-text-light-color: lighten(@il-text-color, 30%); ), best remove that. Note that this is probably also the cause of the conflict.

ezenzen:
Hi @Amstutz,

thanks for the tip. I have adjusted the PR.

Amstutz:
thx a lot!

PR #4093, opened 2022-03-01
UI Components, Slates in Main- and Metabar wrong aria attributes

Amstutz:
In the Materna report we failed the part "4.9.4.1.1 Syntaxanalyse". Various Issues have been found by using the Nu-Checker, see: https://mantis.ilias.de/view.php?id=32121 . Two of them have todo with the wrong use the aria-pressed (see: https://mantis.ilias.de/view.php?id=32150, should only be used for toggle buttons) and aria-expanded (see: https://mantis.ilias.de/view.php?id=32149, should be used on the trigger, not the element being triggered) attributes. 

Both issues are fixed with this PR, if I understood the this correctly. Note that this will be audited again after fixing. It might be possible, that further improvements will be needed.

### DIFF
--- a/src/UI/templates/js/MainControls/dist/mainbar.js
+++ b/src/UI/templates/js/MainControls/dist/mainbar.js
@@ -752,10 +752,10 @@ var renderer = function($) {
         triggerer: Object.assign({}, dom_element, {
             remove: function() {},
             additional_engage: function(){
-                this.getElement().attr('aria-pressed', true);
+                this.getElement().attr('aria-expanded', true);
             },
             additional_disengage: function(){
-                this.getElement().attr('aria-pressed', false);
+                this.getElement().attr('aria-expanded', false);
             }
         }),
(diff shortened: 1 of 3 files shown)


klees:
Hi @Amstutz,

I just had a look into [Mozillas according documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded). It definitely confirms that `aria-expanded` should be used on the controlling element. The documentation also hints at the related attributes `aria-owns` and `aria-controls`. This seems to be something that we want to use to connect the triggers to the slates they control. Can you have a look into that or pass it on to someone?

I'll merge this in the meantime...

Best regards!

Amstutz:
Thx, I will definitely stay on topic here. For the time being, it is good, that this does not show it the automated tests, since it makes other remaining issues much easier to spot (there are still some in the content area, e.g. the mini calendar).

PR #4098, opened 2022-03-02
QTI PHP 8 compatibility

lscharmer:
##### Add compatibility for PHP8.0's change from a XML resource to the XmlParser class:
Changed: `ilQTIParser::depth` from `array<int, int>` to `SplObjectStorage<XmlParser|resource<XML>, int>` so keys can be objects.
Previously the XML resource handle was used as the key (the resource was automatically casted into a file descriptor / integer).

##### Fix bug in `ilQTIResponseLabel::setRange(...)`:
Change `$this->rshuffle = ...` to `$this->rrange = ...`.

##### Replace `strcmp(...)` with `===`
In many places `strcmp($string2, $string2) == 0` was used to compare if 2 strings are equal.

##### Remove unnecessary `is_array(...)` checks in `ilQTIParser::handlerParseBeginTag(...)`:
`xml_set_element_handler(...)` will call the given callback with a `XmlParser|resource<XML>`, a string and an array.
Because of that all related methods are now type hinted on the string and array and all `is_array(...)` checks on the given array parameter are deleted.

##### Add doc types and type hints
Only a few type hints are added, to be sure not to brake any code (because all properties are public).
Instead doc types are added.

##### Add Unit Test suite and tests


Services/QTI/test/ilServicesQTISuite.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+class ilServicesQTISuite extends TestSuite
+{
+    public static function suite() : self
+    {
+        $suite = new self();
+
+        $dir = __DIR__;
+        $a = [];
+        foreach (array_filter(explode("\n", `find "$dir" -name \*.php -and -not -name ilServicesQTISuite.php -print`)) as $file) {
```
mjansenDatabay: ???^^

Services/QTI/classes/class.ilQTIResponseLabel.php
```diff
@@ -94,52 +122,73 @@ public function setRarea($a_rarea)
...
+    /**
+     * @param string $a_rrange
+     */
+    public function setRrange($a_rrange) : void
     {
         switch (strtolower($a_rrange)) {
             case "1":
             case "excact":
-                $this->rshuffle = RRANGE_EXACT;
+                $this->rshuffle = RRANGE_EXACT; // @todo Ask if this should't be $this->rrange.
```
mjansenDatabay: I'm pretty sure this has to be `$this->rrange`

mbecker-databay:
Awesome, thank you so much!

PR #4099, opened 2022-03-02
WIP: Add double border to focus

kergomard:
This is the first of two pull requests to remove a visibility issue for the focus on tabbing (see: https://mantis.ilias.de/view.php?id=32137). It adds a double outline to the element having :focus-visible. I believe this one to be the inferior of the two as it only shields the outline to one side from its environment.

A few important notes:
- This does not touch the entries in the main- and metabar, as there the focus is shown as a border. A few changes where added to keep it this way.
- I had to move the open/close glyph in the slate to avoid removing it. I don't think this has serious consequences for accessibility, but we could also change the approach: I would then move the border to before and keep after. This would require more changes to CSS then the approach I chose here.
- I introduced new less variables and renamed existing ones. I can a
- There is still a issue on FF: When tabbing into the breadcrumb it gets an ellipse. Will try to find a workaround.

### DIFF
--- a/src/UI/templates/default/Button/button.less
+++ b/src/UI/templates/default/Button/button.less
@@ -17,15 +17,12 @@ input.btn{
 .btn-default {
   .button-variant(@il-btn-standard-color; @il-btn-standard-bg; @il-btn-standard-border);
   &:focus-visible {
-    outline-offset: @il-focus-outline-offset;
+    position: relative;
   }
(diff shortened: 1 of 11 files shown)


kergomard:
I close this again. The correct PR is here: https://github.com/ILIAS-eLearning/ILIAS/pull/4100

PR #4100, opened 2022-03-02
Add background to focus

kergomard:
This is the second of two pull requests to remove a visibility issue for the focus on tabbing (see: https://mantis.ilias.de/view.php?id=32137). It adds a double outline to the element having :focus-visible. I believe this one to be the superior of the two as it shields the outline on both sides from its environment.

A few important notes:
- This does not touch the entries in the main- and metabar, as there the focus is shown as a border. A few changes where added to keep it this way.
- I had to move the open/close glyph in the slate to avoid removing it. I don't think this has serious consequences for accessibility, but we could also change the approach: I would then move the border to before and keep after. This would require more changes to CSS then the approach I chose here.
- I introduced new less variables and renamed existing ones.
- There is still a issue on FF: When tabbing into the breadcrumb it gets an ellipse. Will try to find a workaround.

### DIFF
--- a/Services/WebDAV/classes/mount_instructions/class.ilWebDAVMountInstructionsGUI.php
+++ b/Services/WebDAV/classes/mount_instructions/class.ilWebDAVMountInstructionsGUI.php
@@ -119,7 +119,7 @@ class ilWebDAVMountInstructionsGUI
 
         // Add view control and legacy add the beginning of the array (so they will be rendered first)
         $header_comps = array(
-            $f->legacy("<div style='text-align: center'>"),
+            $f->legacy("<div class='webdav-view-control'>"),
             $view_control,
             $f->legacy("</div>"),
             $js_function_legacy);
--- a/src/UI/templates/default/Breadcrumbs/breadcrumbs.less
+++ b/src/UI/templates/default/Breadcrumbs/breadcrumbs.less
@@ -20,6 +20,13 @@
 			&:hover {
 				color: @il-standard-page-breadcrumbs-active-color;
 			}
+			&:focus {
+				border: @il-focus-outline-inner;
+				
+				&::after {
+					content: none;
+				}
+			}
 		}
 
 		&>span+span:before {
--- a/src/UI/templates/default/Button/button.less
+++ b/src/UI/templates/default/Button/button.less
@@ -17,15 +17,12 @@ input.btn{
 .btn-default {
   .button-variant(@il-btn-standard-color; @il-btn-standard-bg; @il-btn-standard-border);
   &:focus-visible {
-    outline-offset: @il-focus-outline-offset;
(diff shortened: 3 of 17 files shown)


kergomard:
I think this is ready for review:
- I also adapted the perma-link. If we change the footer, the change is trivial
- Focus on breadcrumbs, Mainbar and Metabar is reset to use unprotected borders to avoid side-effects. The focus on input-elements is also implemented separately as inputs don't have children and ::after would be a child.
- We have a pre-existing issue with long titles in focused items with a long text in the mainbar on small screens. I see two fixes: We either increase the size of the menu-item or we decrease the size of the image. I chose the latter and put it into a separate commit. The variable @il-icon-size-xsmall seems to only be used in the mainbar.
- There is one more separate commit fixing an issue arising from a not too orthodox use of the view control in WebDAV.

I will close the other PR.

Amstutz (changes requested):
Hi @swiniker 

Thx a lot for this great contribution. I believe this will resolve many of the issue we are currently dealing with. Since this is a productive release we must merge with care, therefore, please fix:
- [x] Radio and Checkboxes in Forms: They seem to have double outlines:
image
image
- [x] Checkboxes in Tables: Seem to also outline the full table:
image
- [x] Checkboxes in Side Panel: Same behaviour in Side Panel:
image

- [x] Remove everything not related to the focus change, especially new lines or reformatting. PR for those are highly welcome.

If possible fix, or indicate, why not possible:
- [x] Jumping Button: Some controls, especially buttons, seem to shift positions slightly. This is best visible e.g. on the Date Time Pagination Control on the Calendar on the Right-hand side of the dashboard. Also, the Subscribe Button there seems affected. We have such cases already in the current solution, but much less. Any hope we can improve this somewhat? Some controls seem not impacted. Notable, button in card for Files is impacted, links in other cards not.
- [x] Calendar Again: This looked bad before, but now it looks worse, can we do something here:
image


I am sure there are more, but those have been the most obvious to me. If nobody interferes, I would be willing to merge this with this state.

Thx!

kergomard:
Thank you @Amstutz 
Ok, The issue with the double borders is Blink specific. They decided to be kind with us web developers and to show us ::after inside inputs. Not very kind of them ;-).
Please have a closer look at the calendar changes: The solution I would choose to always make it look good, is to set "display: block" and "width: 100%" on the a-element inside the calmini cells. For me this is the right way to show it anyways, but it is a change with a bigger scope than only the focus.

Amstutz:
Hi @swiniker 

Thx a lot. IMO this is a big improvement and a great contribution for the a11y in ILIAS. THX! Very nice, IMO this change improved some other details like size of clickable area in mini-calendar.

Thx and Kudos
@Amstutz

PR #4101, opened 2022-03-03
Services/Object: PHP8 Improvements

daniwe4:


Services/Object/classes/class.ilObjectGUI.php
```diff
@@ -63,7 +63,7 @@ class ilObjectGUI
     protected string $tmp_import_dir;	// directory used during import
     protected string $sub_objects = "";
     protected bool $omit_locator = false;
-    protected string $type = "";
+    protected ?string $type = "";
```
klees: Why do we need the `?` here when we set this to an empty string anyway?

daniwe4: We don't. It's removed.

klees (approved):
This LGTM, I'll merge on monday so I'll be able to look into problems and not leave this broken during the weekend.

PR #4103, opened 2022-03-04
28790, 26497, 25648 Rename showMoveIntoObjectTree to showPasteTree

padvincenzo:
Hi everyone,
This PR solves old bug (mantis codes 28790, 26497 and 25648). I used mantis code 13258 to find this solution.
This bug is present in all ILIAS versions. If this is the best solution to solve the bug, should I create PRs for release 6, 7 and trunk as well?


smeyer-ilias:
Wow, thanks for analyzing these issue and providing a PR. 
I will merge to the other releases as well.

PR #4104, opened 2022-03-05
Setup: Use `ClientId` type in setup

mjansenDatabay:
The `Client Id` validation (to be more precisely: the `RegEx`) in the setup differs from that used in our `ClientId` data type. This PR applies a fix by using the `ClientId` data type in the `ilSetupConfig` instead. 

This should be cherry-picked to `release_7`. The `Type Property` MUST be removed afterwards.

Mantis Issue: https://mantis.ilias.de/view.php?id=32301

### DIFF
--- a/setup/classes/class.ilSetupAgent.php
+++ b/setup/classes/class.ilSetupAgent.php
@@ -44,7 +44,7 @@ class ilSetupAgent implements Setup\Agent
         return $this->refinery->custom()->transformation(function ($data) {
             $datetimezone = $this->refinery->to()->toNew(\DateTimeZone::class);
             return new \ilSetupConfig(
-                $data["client_id"],
+                $this->data->clientId($data["client_id"] ?? ''),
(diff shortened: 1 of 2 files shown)


klees (changes requested):
Hi @mjansenDatabay,

I'm about 100% sure that this will work =)

I still think we should push this code up one layer:

Make the `ilSetupConfig` accept a `ClientId` instead of the `string` in the constructor and move the creation of said `ClientId` to the according `Agent`. We will also have a `Data\Factory` available there...

Thanks for looking into this and best regards!

mjansenDatabay:
@klees I totally agree, I'll provide some changes in next minutes.

mjansenDatabay:
@klees Done. If possible, please cherry-pick this to `release_7` afterwards. https://github.com/ILIAS-eLearning/ILIAS/pull/4128 should be cherry-picked before merging THIS PR.

Of course we could debate on whether `\ilSetupConfig::getClientId` should return the data type now :-).

PR #4105, opened 2022-03-05
fix/services/php8/orgunits

mstuder:

PR #4106, opened 2022-03-05
php8/orgunits/typehinting

mstuder:
wait until https://github.com/ILIAS-eLearning/ILIAS/pull/4105 is merged


mstuder:
@see https://github.com/ILIAS-eLearning/ILIAS/pull/4105#issuecomment-1059997622

PR #4109, opened 2022-03-07
Remove trailing space

corro:
Remove trailing space from options `key` and `dhparam`. The space leads to the options not being recognized by the setup CLI and therefore missing in the configuration file server.cfg.


chfsx:
Thanks a lot @corro !

PR #4110, opened 2022-03-07
Test Assessment Player View Refactoring

catenglaender:
This PR is meant for trunk as soon as it's repaired and has been applied to release_7 for testing purposes.

# Scope of the PR

This PR for the Test Player of the Test and Assessment Center includes:
* refactoring of the style code and the way how it's loaded
* slight visual tweaks for broken and outdated interface elements to make them more coherent with the rest of ILIAS

These changes were implemented in our ILIAS fork and we want to offer them to the community.

#### Issue

* CSS files ta.css and ta_split.css were loaded in an unconventional way by putting a style tag into the DOMs head instead of importing it through the global delos.less.
* There was no LESS file from which the CSS was generated, so the style code specific to the Test Assessment had no access to general LESS variables like colors or paddings, which had changed since its creation.

#### Changes

* To follow the structure of other modules, the content of ta.css and ta_split.css have been moved to templates/default/less/Modules/Test/ into a delos.less, which is imported by the global delos.less. Global LESS variables can now be used in this new file.
* Legacy.less includes the code from those CSS files that hasn't been refactored yet.
* Overwriting Test Assessment style code should now be done through the ILIAS wide customizing solution via a delos.less/css in the Customizing folder.

#### Issue

* Split view with question list wasn't responsive.

#### Changes

* A flexbox solution has been implemented to support dynamic growing and shrinking of columns and a reverse order for the mobile breakpoint.
* This solution will be obsolete when the question shortlist is moved to the slate with this [feature request](https://docu.ilias.de/goto.php?target=wiki_1357_Revamping_Exam_View_with_List_of_Questions) and should be removed then.

image

### Fix of broken bottom navigation

* an existing bottom navigation inside the HTML template wasn't shown because the same variable was used for top and bottom navigation
* the variable has been fixed, so the bottom navigation appears now as it seems to have been intended
* a bottom navigation makes it easier to submit an answer to a long question because the user doesn't have to scroll up

image

#### Issue

* Time remaining message floats loosely above the test and doesn't seem coherent with the contained panel principles that are applied throughout the ILIAS interface

#### Change

* Time remaining message is now placed inside an info alert and uses an Ilias variable for the big font size

image

#### Issue

* Marked star icon was placed below marked question title in shortlist, which makes it hard to see which question is marked and wastes a lot of space
* h3 tags in a menu list are not good accessibility practice for screen readers, because no actual content follows there.

#### Changes

* Marked star icon in shortlist now appears next to the question title, which makes reading the list a lot easier.
* Shortlist uses just li tags instead of h3 tags, which was a problem for screen reader accessibility
* Font size of shortlist now comes from core delos.less

image


dsstrassner:
Thank you for your offer and work. I spoke with @catenglaender, and we discussed the single aspects of this PR. We agreed that every aspect he brought up is also broken in ILIAS 7. So, I close this PR, and he creates single PRs for every aspect with the related Mantis Issue report.

PR #4111, opened 2022-03-08
Fix #31111: missing custom metadata after upgrade.

thibsy:
Hello @smeyer-ilias, as described in https://mantis.ilias.de/view.php?id=31111 an update step is missing that migrates data from the database-table `adv_md_values_text` to `adv_md_values_ltext`. This PR introduces that missing step.

I made an assumption though, because the new table `adv_md_values_ltext` contains another column called `value_index`, where I just adpoted the default lang-code from `adv_md_field_int` because that information previously didn't exist.

Is it also necessary to create an entry with and without value_index? because that is the default behaviour when adding custom-metadata values per GUI in ILIAS 7 and I didn't know if that's intended.

### DIFF
--- a/setup/sql/7_hotfixes.php
+++ b/setup/sql/7_hotfixes.php
@@ -1378,11 +1378,39 @@ if(!$ilDB->indexExistsByFields('il_bt_bucket', ['user_id'])) {
 <#73>
 <?php
 $ilDB->replace("settings", [        // pk
-                           "module" => ["text", "common"],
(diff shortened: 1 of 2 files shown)


chfsx:
Hi @smeyer-ilias  

Best regards!

smeyer-ilias:
Thanks for fixing this and sorry for the delay. 
This is implemented in R7 since last week.

PR #4112, opened 2022-03-09
Fix erratic scroll behaviour when editing gap question

kergomard:
This fixes the scroll behavior in gap questions. See https://mantis.ilias.de/view.php?id=32191 .


mbecker-databay:
Thank you!

PR #4117, opened 2022-03-10
Search: Revise translation

mjansenDatabay:
This PR suggests new labels/bylines (DE/EN) for the `Global User Search` option in the `Search Administration`.

See: https://mantis.ilias.de/view.php?id=32039

If approved, this has to be cherry-picked to `trunk` (and depending on the datet/time of merge to `release_8`) as well.


matthiaskunkel:
Thanks @mjansen. It would be great if you could avoid the "If enabled..." form. We currently try to get rid of it and just write: "Users can be found in the global search. Furthermore, data …" and "Benutzerkonten können über die globale Suche gefunden werden...." Thanks!

mjansenDatabay:
@matthiaskunkel Changed

matthiaskunkel:
Thanks @mjansen for this PR. Please commit it to trunk, too.

PR #4128, opened 2022-03-14
Data/ClientId: Throw exception if argument is empty

mjansenDatabay:
This PR adds an additional constraint to the validation of the passed `$cliendId` string: A `Client ID` MUST NOT be an empty string, therefore a dedicated exception should be thrown.

If approved, this should be cherry-picked to `release_7` and `trunk`.

### DIFF
--- a/src/Data/ClientId.php
+++ b/src/Data/ClientId.php
@@ -21,6 +21,10 @@ class ClientId
      */
     public function __construct(string $clientId)
     {
+        if ($clientId === '') {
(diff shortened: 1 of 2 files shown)


mjansenDatabay:
@klees Anything missing here? I hope it was okay to add some more tests and don't created a separate PR for this.

klees:
@mjansenDatabay No, I consider "more tests" to be a valid application of the boy scout rule: Always leave a place a little nicer than you have found it.

PR #4029, opened 2022-02-11
UI Components, fix of responsive Logo Issue, see: #32016

Amstutz:
Currently, the responsive Header Logo is not displayed on small screen sizes, the standard one is. See: https://mantis.ilias.de/view.php?id=32016

Fix intended for ILIAS 7 and Trunk, JF approval needed due to interface change on Standard Page Layout.

PR #4131, opened 2022-03-15
UI Components, fix of responsive Logo Issue, see: #32016

thibsy:
Currently, the responsive Header Logo is not displayed on small screen sizes, the standard one is. See: https://mantis.ilias.de/view.php?id=32016

This PR is based on https://github.com/ILIAS-eLearning/ILIAS/pull/4029 introduced by @Amstutz with ILIAS 8 compatibility.

PR #4129, opened 2022-03-15
ILIAS7: UI Components, fix of responsive Logo Issue, see: #32016

thibsy:
Currently, the responsive Header Logo is not displayed on small screen sizes, the standard one is. See: https://mantis.ilias.de/view.php?id=32016

This PR is based on #4029 introduced by @Amstutz with the addition of a new `ReplaceResponsiveLogoCommand`. The command will replace the ILIAS default responsive header icon in skins, which have a custom `HeaderIcon.svg` but still the default ILIAS `HeaderIconResponsive.svg`. This solves the issue of displaying the wrong or default responsive icon when the site enters a certain width, for ppl who haven't used the responsive-icons yet.

PR #4132, opened 2022-03-15
ILIAS6: UI Components, fix of responsive Logo Issue, see: #32016

thibsy:
Currently, the responsive Header Logo is not displayed on small screen sizes, the standard one is. See: https://mantis.ilias.de/view.php?id=32016

This PR is based on https://github.com/ILIAS-eLearning/ILIAS/pull/4029 introduced by @Amstutz with ILIAS 6 compatibility.


Amstutz (approved):
See comment in https://github.com/ILIAS-eLearning/ILIAS/pull/4129. @thibsy shortly talked about the introduced strict types. We agree that this would be better for only trunk normally, however since they have been manually tested by @thibsy I am fine with them in this case. @klees do you want to check the cli part?

setup/cli.php
```diff
@@ -67,9 +67,13 @@ function build_container_for_setup(string $executed_in_directory)
             $c["command.install"],
             $c["command.update"],
             $c["command.build-artifacts"],
-            $c["command.reload-control-structure"]
+            $c["command.reload-control-structure"],
+            $c["command.replace-logo"]
```
chfsx: @thibsy not sure if this need an own command like this, maybe only a Objective would be sufficient?

thibsy: Would be sufficient, I implemented a command to simplify the composer.json entry. However, something like this should also be possible: `php -r "(new ReplaceResponsiveLogoObjective())->execute(new ArrayEnvironment([]))"`.

thibsy:
As per @chfsx's suggestion I moved the implementation as a command to an objective that has been added to the ILIAS update process. This makes more sense because the replacement doesn't need to take place on every composer dump.

klees (approved):
I didn't check the algorithm of the `Objective` in detail, but the form and inclusion looks just fine.

Remind yourself that you can remove it with ILIAS 9 because of our [new update policy](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/supported-versions.md#supported-versions).

thibsy:
@klees yes, I already removed the objective for the ILIAS 8 PR (#4131). 

I'll merge the PRs then =).

PR #4134, opened 2022-03-16
Introduction of further less variables

ezenzen:
Hi @all,

while working in the "css squad" group, I identified more hex color values and replaced them with less variables. The PR further reduces the number of different color tones.

I also replaced some bootstrap variables with ILIAS less variables. 

@alex40724 pointed out to me that less-variables should not be used in content.less due to dependencies on the content style. I have removed all variables from this file and replaced them with the correct hex values.

Many greetings,
Enrico

### DIFF
--- a/Modules/Cloud/templates/css/delos.less
+++ b/Modules/Cloud/templates/css/delos.less
@@ -4,7 +4,7 @@
   z-index:    9000;
   top:        0;
   left:       0;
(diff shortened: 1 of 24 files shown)


ezenzen:
As discussed with Timon Amstutz, I close this PR and create it again for trunk.

PR #4118, opened 2022-03-10
Implement Renderer for plugins to prevent collision

iszmais:
~~This PR prpose a solution to https://mantis.ilias.de/view.php?id=32196 by offering the plugin developers a base renderer wich can handle multiple replacements/eschanges by `exchangeUIRendererAfterInitialization() `.~~

~~Additional this also adds a wrapping Renderer wich allows to prepend or append conent to ILIAS components without having to manipulate them and therefore allows multiple plugins to take an effect on the same component.
This provides a new way to use the ILIAS5.4 concept of APPEND and PREPEND in UIHookPlugins without take along their problems.~~

This PR will centralize the basic function of chaining renderers inside a abstract renderer to simplify the usage of renderer exchanges without the need to implment a chaining yourself

PR #4137, opened 2022-03-16
Make Chaining of `Renderer`s possible

klees:
As pitched by @iszmais in #4118, this makes it possible to write decorators for `Renderer` in the UI-framework, by passing the root of a chain of decoratred `Renderer`s through the `render`-function.

The change is non-breaking from the perspective of the core, since existing usages of `Renderer::render` will continue to work as expected. This is breaking from the perspective of plugins, however, since the interface changed and implementing classes will need to change accordingly.

We might want to take this to the JF for this reason or even decide to only make the change in the trunk. I would be happy if @Amstutz takes an according decision as I'm the one who proposes this.

### DIFF
--- a/src/UI/Implementation/DefaultRenderer.php
+++ b/src/UI/Implementation/DefaultRenderer.php
@@ -32,17 +32,18 @@ class DefaultRenderer implements Renderer
     /**
      * @inheritdoc
      */
-    public function render($component)
+    public function render($component, ?Renderer $root = null)
     {
+        $root = $root ?? $this;
+
         $out = '';
         if (is_array($component)) {
             foreach ($component as $_component) {
-                $renderer = $this->getRendererFor($_component);
-                $out .= $renderer->render($_component, $this);
+                $out .= $root->render($_component);
             }
         } else {
             $renderer = $this->getRendererFor($component);
-            $out = $renderer->render($component, $this);
+            $out = $renderer->render($component, $root);
         }
 
         return $out;
@@ -51,24 +52,18 @@ class DefaultRenderer implements Renderer
     /**
      * @inheritdoc
      */
-    public function renderAsync($component)
+    public function renderAsync($component, ?Renderer $root = null)
     {
-        $out = '';
+        $root = $root ?? $this;
 
+        $out = '';
         if (is_array($component)) {
             foreach ($component as $_component) {
-                $out .= $this->render($_component) .
-                $this->component_renderer_loader
-                        ->getRendererFactoryFor($_component)
-                        ->getJSBinding()
-                        ->getOnLoadCodeAsync();
+                $out .= $root->renderAsync($_component);
             }
         } else {
-            $out = $this->render($component) .
-            $this->component_renderer_loader
-                    ->getRendererFactoryFor($component)
-                    ->getJSBinding()
-                    ->getOnLoadCodeAsync();
+            $out = $this->render($component, $root) .
+            $this->getJSCodeForAsyncRenderingFor($component);
         }
         return $out;
     }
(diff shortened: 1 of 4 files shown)


iszmais:
Thx for the contribution @klees ,

LGTM for me codewise, but i agree that this should go to the JF.

Greetings,
@iszmais

iszmais (approved):

iszmais:
Just a proposal:
Maybe it would be usefull for all stages below trunk to provide this changes excluding the interface changes. This would make them non-breaking for already existing plugins but still provides the option to use the chaining feature since addition dont violate the origin interface.

klees:
@iszmais You mean: making the change in the `DefaultRenderer` but not the `Renderer`-interface? Sounds plausible...

Amstutz (changes requested):
Hi @klees 

I am fine with the changes on the interface and agree that we should get the opinion of the JF here.

Concerning the implementation, a two questions:
- [x] Arrays of Components: IMO, [this line](https://github.com/ILIAS-eLearning/ILIAS/pull/4137/files#diff-c56a548132be9f8c8ccf6cde8287558f810c7a7fc0c8c97f18d4d67e4049b3f7R42) is wrong, right? It should be the same logic as if rendering a single component, `$renderer = $this->getRendererFor($component); $out = $renderer->render($component, $root)`. Maybe factor out does two duplicate line of code.
- [x] Async Rendering: IMO, this does not work for Async Rendering atm. The $root parameter needs to be passed in the function for sync rendering as well, right?

Please Change:
- [x] Add Tests: Maybe I am wrong with the above. If so, prove it by adding tests for passing arrays, passing arrays on async and rendering single async components. If not, make sure those issues will not hunt us in the future. 

Thx!

klees:
Hi @Amstutz,

thanks for looking into this:

* Arrays of Components: If we would not include this line we would need every decorator for `Renderer` to deal with arrays of components as well. If we include this line, they would only need to deal with single components because the `DefaultRenderer` would start the chain again for the case of an array of components. Both would be valid, but I expect the current version to cause less surprises.
* Async Rendering: I simplied missed this. Will add...

Best regards!

klees:
@Amstutz Done.

Amstutz (approved):
Hi @klees 

Thx for adding the answer and the tests. This is good to be merged, if accepted by JF.

Thx.

iszmais:
Hi @klees ,

This may be a little late, but the problem occured to me just now. I tried some testing with this an dchaining renderers together and came to some difficulties. E.g. this is still bound to some kinde of order or deep level in its rendering. Like when i have to make some wrapping modifications inside other renderers (e.g. Render A manipulates a section and Render B manipulates the page outside the section and a button inside the section).
I came to the conclusion that this is cause conceptionally by the fact the the added parameter "root" i required to serve 2 purposes here, as a fallback (wich should always be the complete UI-Renderer) and an chaining (wich should only be the next chain node) wich it cant solve both. (I may be wrong here, pls correct me if needed)
To clsoe the circle: I think to solve this problem we neverless need a addition to the getRendererFor function, but much smaller as my first approach.
This may still be a good improvement, just not the one solving the mentioned problem.

Greetings,
@iszmais

matthiaskunkel:
Jour Fixe, 21 MAR 2022 : Timon pointed to this PR that will be discussed in more detail on the next JF and that is a pre-requisite for a following PR for a plugin rendererer.
In addition, please don't forget to add information about changes and implications of a PR to the readme files to make it easier for others to understand the PR and its aim.

matthiaskunkel:
Jour Fixe, 04 APR 2022: We highly appreciate this suggestion and accept the PR for all maintained versions and trunk.

Amstutz:
Thx a lot!

PR #4141, opened 2022-03-16
GlobalCache: add named objective to flush all caches

klees:
Hi @chfsx,

this adds a named objective that makes it possible to flush all caches via the setup. The GUI setup had that functionality, but we did not recreate it for the CLI setup until now.

One can call this via `php setup/setup.php achieve globalcache.flushAll`.

Best regards!


chfsx:
Thanks a lot!

chfsx:
I will cherry-pick/adapt it for trunk as well

PR #4142, opened 2022-03-16
PHP 8: Code Review `UIComponent`

mjansenDatabay:
Hi @alex40724,

I completed the review of the `Services/UIComponent` component.

This component looked very good, I only found one file where the coding style was not compliant. Furthermore I had to add a few types here and there.

Results:
- [x] The unit tests pass

I am not sure whether https://packagist.org/packages/geshi/geshi (used by `Services/UIComponent/SyntaxHighlighter`) is compatible with PHP 8. Please check the list item above if you already checked this.

Best regards,
@mjansenDatabay

### DIFF
--- a/Services/Skill/classes/class.ilSkillSelectorGUI.php
+++ b/Services/Skill/classes/class.ilSkillSelectorGUI.php
@@ -32,7 +32,6 @@ class ilSkillSelectorGUI extends ilVirtualSkillTreeExplorerGUI
     protected $select_gui = "";
     protected string $select_cmd = "";
     protected string $select_par = "";
-    protected bool $select_multi = false;
     protected SkillAdminGUIRequest $admin_gui_request;
     protected array $requested_selected_ids = [];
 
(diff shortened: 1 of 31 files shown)


Services/Skill/classes/class.ilSkillSelectorGUI.php
```diff
@@ -32,7 +32,6 @@ class ilSkillSelectorGUI extends ilVirtualSkillTreeExplorerGUI
     protected $select_gui = "";
     protected string $select_cmd = "";
     protected string $select_par = "";
-    protected bool $select_multi = false;
```
mjansenDatabay, line 35: @alex40724 I had to move this to the base class, because it is also used there, and without my changes this was dynamically declared in the base class.

alex40724:
We are using the latest geshi version now. It does not explicit state to be compatible with PHP8, however I tested the code elements in the page editor and they work flawlessly.

alex40724:
Thanks for the review and changes!

PR #4146, opened 2022-03-17
fix for 26215

okaufman:
https://mantis.ilias.de/view.php?id=26215


mbecker-databay:
Very good! This is the solution I had in mind for this one :)

PR #4153, opened 2022-03-18
Fixed `UIServices::mainTemplate()` type-hint

thibsy:
This annoyed me one time too many now, hence the PR. I don't think there is any case there's not an `ilGlobalTemplateInterface` instance returned right?

### DIFF
--- a/src/DI/UIServices.php
+++ b/src/DI/UIServices.php
@@ -41,7 +41,7 @@ class UIServices
     /**
      * Get the ILIAS main template.
      *
-     * @return	\ilTemplate
+     * @return	\ilGlobalTemplateInterface
      */
     public function mainTemplate()
     {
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
I suggest to use an explicit return type: `: \ilGlobalTemplateInterface`

klees:
Hi @thibsy!

I will not merge this for two reasons:

* I'm not that sure that this really is always the case and I also cannot check. So I'm a little afraid that this will simply break release_7. Would be ok for me doing this in `trunk`, though.
* Nobody should rely on this typehint in the first place. Be a good citizen, pull your dependencies in the constructor and store them in type-hinted properties =) I know that this is not state of the art in many places, but it still makes me even more hesitant to merge this to a productive release.

Still: thanks for taking care about stuff that annoys you. I hope you understand my reasoning (and maybe reopen this for `trunk`).

Thanks!

thibsy:
Hi @klees, thanks nevertheless, I understand. Trunk luckily already has this so no need to reopen 👍

PR #4151, opened 2022-03-18
ItemGroup: Fix ilCtrl issue when toggling item group accordion state

mjansenDatabay:
If https://github.com/ILIAS-eLearning/ILIAS/pull/4154 has been approved, this PR here can be closed  because it only solves a concrete symptom.

Maybe also a duplicate of https://github.com/ILIAS-eLearning/ILIAS/pull/3970

Mantis Issue: https://mantis.ilias.de/view.php?id=31483

If approved, please cherry-pick this to `trunk` as well.

PR #4154, opened 2022-03-18
UICore/ilCtrl: Support classes without parents and children

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=31483

If a PR for ILIAS 8 is needed, I already fixed this in a hotfix branch: https://github.com/mjansenDatabay/ILIAS/commit/8dc514ee78195cbfbba9d4628c802dc9a55d3583

### DIFF
--- a/Services/UICore/classes/Setup/class.ilCtrlStructureReader.php
+++ b/Services/UICore/classes/Setup/class.ilCtrlStructureReader.php
@@ -364,7 +364,7 @@ class ilCtrlStructureReader
     // ILCTRL DECLARATION FINDING
     // ----------------------
 
-    const IL_CTRL_DECLARATION_REGEXP = '~^.*@{WHICH}\s+([\w\\\\]+)\s*:\s*([\w\\\\]+(\s*,\s*[\w\\\\]+)*)\s*$~mi';
+    const IL_CTRL_DECLARATION_REGEXP = '~^.*@{WHICH}\s+([\w\\\\]+)\s*:\s*([\w\\\\]+(\s*,\s*[\w\\\\]+)*)?\s*$~mi';
 
     /**
      * @return null|(string,string[])
@@ -402,7 +402,7 @@ class ilCtrlStructureReader
 
         $declaration = [];
         foreach ($res[2] as $ls) {
-            foreach (explode(",", $ls) as $l) {
+            foreach (array_filter(explode(",", $ls)) as $l) {
                 $declaration[] = strtolower(trim($l));
             }
         }


thibsy:
Hi @mjansenDatabay, I actually fixed the issue in #3970, which hasn't been approved by @alex40724 yet. I don't think this is an ilCtrl issue. Classes without parents or children are considered a base class (for which I created an interface in ILIAS 8) and should either be treated as such or named properly.

mjansenDatabay:
@thibsy The similar issue is also valid for `ilStartingPoint`, `ilTableTemplatesStorage` and `ilNotification`, according to @utesche

thibsy:
@mjansenDatabay but a quick search did not find any references to these classes with `@ilctrl_iscalledby`, so they should not be considered by ilCtrl at all.

thibsy:
I suggest we accept #3970 instead and close this PR (incl. #4151).

alex40724:
@thibsy Please note that e.g. ilTablePropertiesStorage declares itself as base class (ILIAS 6)
```
	<baseclasses>
		<baseclass name="ilTablePropertiesStorage" dir="classes" />
	</baseclasses>
```
and contains a ilCtrl_Calls comment
```
* @ilCtrl_Calls ilTablePropertiesStorage:
```
This was ok to make it work as a base class in ILIAS 6. I did not check the cases mentioned by Uwe yet.

I will just check these cases. If they are all irrelevant, I will close again...

(1) ilTablePropertiesStorage works again in ILIAS 7 because of this change:

```
* @ilCtrl_Calls ilTablePropertiesStorage: ilTablePropertiesStorage
```

(2)
ilTableTemplatesStorage, ilNotification and ilStartingPoint contain ``@ilCtrl_Calls`` comments without having an executeCommand. I will remove these comments.

thibsy:
Hi @alex40724, the class you mention should be called `ilTablePropertiesStorageGUI` instead of `ilTablePropertiesStorage`, otherwise ilCtrl will not treat this class as "necessary" for the control-structure. The convention is `class.il<name>GUI.php`. 

After renaming this class link-targets should work again (similar case as in #3970).

alex40724:
@thibsy As written in (2) above the other cases are non-issues. 

ilTablePropertiesStorage works in ILIAS7 (its being called, if you hide/show a table filter, e.g. in the user administration). The class does not implement a UI hence it was not named this way, still works however.

alex40724:
@thibsy There are 4 base classes in ILIAS 7 not ending on "GUI":

- ilAccordionPropertiesStorage (has been fixed the same way as ilTablePropertiesStorage in ILIAS 7, broken in trunk)
- ilContainerBlockPropertiesStorage (see PR)
- ilSearchController (has children, seems to work in 7, has been renamed to ilSearchControllerGUI in trunk)
- ilTablePropertiesStorage  (fixed in 7, broken in trunk)

thibsy:
@alex40724 could you point me to the mantis-tickets if they exist, or create them if not? Thx!

PR #4156, opened 2022-03-21
Test&Assessment: CSS and LESS location - Mantis #32212

catenglaender:
Mantis Issue: https://mantis.ilias.de/view.php?id=32212

# Issue

* CSS files ta.css and ta_split.css are loaded in an unconventional way by putting a style tag into the DOMs head instead of importing it through the global delos.less. 
* There was no LESS file from which the CSS was generated, so the style code specific to the Test Assessment had no access to general LESS variables like colors or paddings, which had changed since its creation.

# Changes

* To follow the structure of other modules, the content of ta.css has been moved to templates/default/less/Modules/Test/ into a delos.less, which is imported by the global delos.less. Global LESS variables can now be used in this new file.
* later today a PR to fix split layout responsiveness https://mantis.ilias.de/view.php?id=32213 will follow which removes ta_split.css as it will become redundant with the responsive layout.

# Impact

* Overwriting Test Assessment style code should now be done through the ILIAS wide customizing solution via a delos.less/css in the Customizing folder.


Amstutz (changes requested):
Hi @catenglaender 

Thx a lot. Moving CSS to the proper place is a good and important step. Thx a lot @catenglaender. 

Just one question:

- [ ] ta_split.css: Do we still need this file? Could we not move that to the proper location in templates/default? Maybe even merge that with the test's main CSS?

One thing to adjust:
- [ ] Docu: This is a good and significant refactoring. I see  you mentioned the relation to https://mantis.ilias.de/view.php?id=32213, since you require variables to fix this. I see how they could help properly fixing this. Note that this change might break with skins already in production, currently offering an own version of the tests CSS in the Customizing folder. So best drop a line here: https://github.com/ILIAS-eLearning/ILIAS/tree/release_7/templates#ilias-7 and note that the old test.css is no longer in use, and that CSS changes to the test need to go in the main CSS (as almost all other CSS adaptions for modules). 

Thx a lot for this contribution!

@Amstutz

catenglaender:
Hi @Amstutz

thanks for your feedback! 👍 

in this PR ta_split.css is still necessary as it's loaded dynamically to get the split view with the shortlist to open and close. In a later PR https://github.com/ILIAS-eLearning/ILIAS/pull/4157 I made the splitview responsive and use :empty to detect the layout change purely with CSS, so I have removed the file and its php calls there.

Do you know of any place where the previous styling method was documented, so I can choose my wording in the migration recommendation accordingly?

Happy to contribute :)

Amstutz:
Hi @catenglaender 

Thx for making the ta_split redundant and removing it! Always good to have one less CSS file dangling around.

> Do you know of any place where the previous styling method was documented, so I can choose my wording in the migration recommendation accordingly?

If you mean the CSS in the test, unfortunately no. I get myself warmed up currently in CSS/Less for test along with its DOM structure for work on a11y. I have a hard time finding my way around. This PR of yours was the best doc so far in this regard ;-). 

If you mean doc concerning breaking changes in updates, then you will find some here: https://github.com/ILIAS-eLearning/ILIAS/tree/release_7/templates#migration (Section Migration). I would suggest something like: "In Section **ILIAS 7** (add new paragraph); In March 2022 we moved the test.css to less/Modules/Test/delos.less (as all other modules specific CSS) to fix ... . Note that this might affect your CSS in skin if you introduced changes to test.css". Or similar.

catenglaender:
Hi @Amstutz,

I added a note to the readme.
As mentioned ta_split.css will be taken care of with https://github.com/ILIAS-eLearning/ILIAS/pull/4157

Let me know if there is anything else to do :)

Amstutz (approved):
From a Less/CSS perspective, this is good to go. Thx a lot for this refactoring/cleaning up!

Amstutz:
Hi @catenglaender 

Normally, the maintainer picks changes to several versions. However, here, the maintenance is split between TA and CSS. Maybe best in this case, to avoid errors due to improper handling of conflicts, would be a separate PR for trunk. Is it possible for you to provide that? Would be highly appreciated. 

thx

Amstutz:
Hi @catenglaender 

I picked the changes today to trunk. There have been some conflicts, so I hope I resolved everything correctly, see: https://github.com/ILIAS-eLearning/ILIAS/commit/fb7f2b9f94507e4ac04e62e938490100e5b5d8c7 and 3 children.

PR #4157, opened 2022-03-21
Test&Assessment: Responsive Split View - Mantis #32213

catenglaender:
Mantis Issue: https://mantis.ilias.de/view.php?id=32213

Is dependent on PR #4156, which is included in first commit for reference and testing.

# Issue

* Split view with question list wasn't responsive.

# Changes

* A flexbox solution has been implemented to support dynamic growing and shrinking of columns and a reverse order for the mobile breakpoint.
* Shortlist jumps below question
  * on mobile
  * on medium screen size when slate is engaged
* Split layout is activated/deactivated by shortlist being present or not in DOM. Therefor, ta_split.css could be removed as it's no longer needed to activate the split layout. This completes an aspect of https://mantis.ilias.de/view.php?id=32212
* This solution will be obsolete when the question shortlist is moved to the slate with this [feature request](https://docu.ilias.de/goto.php?target=wiki_1357_Revamping_Exam_View_with_List_of_Questions) and should be removed then.

image

PR #4164, opened 2022-03-22
Test&Assessment: missing bottom navigation 32184

catenglaender:
Mantis Issue: https://mantis.ilias.de/view.php?id=32184

Is dependent on PR https://github.com/ILIAS-eLearning/ILIAS/pull/4156 and https://github.com/ILIAS-eLearning/ILIAS/pull/4157 which are included in first commit for reference and testing.

# Fix of broken bottom navigation

* an existing bottom navigation inside the HTML template wasn't shown because the same variable was used for top and bottom navigation
* the variable has been fixed, so the bottom navigation appears now as it was in Ilias 5.4
* a bottom navigation makes it easier to submit an answer to a long question because the user doesn't have to scroll up

image


dsstrassner:
hi @catenglaender!

We discussed this PR today @ TechSquad! 

For me as T&A maintainer, this is the right solution to the problem. This consent was missing in the PR until today.

@nhaagen will communication internally with you about the next steps.

catenglaender:
To solve merge conflicts, this is now part of this PR: https://github.com/ILIAS-eLearning/ILIAS/pull/4792

PR #4165, opened 2022-03-22
TA 31132, kindly transform points to float

nhaagen:
https://mantis.ilias.de/view.php?id=31132


mbecker-databay:
Awesome, thank you!

PR #4166, opened 2022-03-22
TA 32027, re-enable key-checking on numeric input

nhaagen:
re-enable input checking (while typing) with numeric inputs.
fixes https://mantis.ilias.de/view.php?id=32027, but probably there were reasons for taking this out...


alex40724 (approved):
Unfortunately the old issue referenced in the commit outcommenting the JS does not seem to exist anyore in mantis. Quite strange, I thought all issues are kept forever.

So I don't feel very comfortable with this. Input checking should be done on the server side, using checkInput(), see https://docu.ilias.de/goto_docu_pg_213_42.html

Yes, documentation is old, but it's documented. This form in modal thing is not implemented according the documentation imo.

@nhaagen Please merge this to the trunk and ilias 6, if relevant, too. Thanks.

PR #4167, opened 2022-03-22
Update permission's check

padvincenzo:
Shouldn't the course in the Dashboard be shown to the user if he can't read _or_ see it?

### DIFF
--- a/Services/Dashboard/ItemsBlock/classes/class.ilPDSelectedItemsBlockMembershipsProvider.php
+++ b/Services/Dashboard/ItemsBlock/classes/class.ilPDSelectedItemsBlockMembershipsProvider.php
@@ -63,7 +63,7 @@ class ilPDSelectedItemsBlockMembershipsProvider implements ilPDSelectedItemsBloc
             $parentTreeLftValue = $item->getParentLftTree();
             $parentTreeLftValue = sprintf("%010d", $parentTreeLftValue);
 
-            if (!$this->access->checkAccess('read', '', $refId) &&
+            if (!$this->access->checkAccess('read', '', $refId) ||
                 !$this->access->checkAccess('visible', '', $refId)) {
                 continue;
             }


iszmais:
Hi @padvincenzo ,

thank you for your contribution. Your statement is correct (at least for exclusive or), but maybe a little bit confusing due to double negation. To wrap it up: 
A course is shown on the dashboard if the user has at least one of the to permission of 'read' or 'visible'. Therefore the logic syntax for this is **(read || visible)**. Since the target scope is to exclude courses wich shouldnt be shown the case for this is the negation of it which should **!(read || visible)** wich can be converted through de morgans law to **(!read && !visible)** or said in prosa _"A course should not be shown on the dashboard if the user has not a read permission and not a visible permission on it"_.

Hope this was understandable, feel free to ask if you have further question for to the permission itsself.


PS: All semantic and conceptional reports in ILIAS need to have a mantis ticket linked to it. So if you still see a problem here i would appreciate if you could create such a ticket [here](https://mantis.ilias.de/).

Greetings,
@iszmais

padvincenzo:
Hi @iszmais 
Thanks for your reply, but I still do not understand the meaning of "visible" permission.
I'll open a mantis ticket to delve into this topic.

Thank you again,
Vincenzo

iszmais:
Hi @padvincenzo ,
The read persmission allows to enter the object. this does not allow further actions inside the object itself but it makes the object "clickable" on the dashboard e.g.
The visible permission only allows to see the object. You cannot interact with or enter it/see infos/ etc. Its simply to see the object, nothing more.
I hope this helps. Feel free to ask further.

Greetings,
@iszmais

padvincenzo:
Thanks @iszmais for your explanation. Now I understand that if I want to have a course hidden from the dashboard, I have to create a new permission or (better) a new option in the course settings.

Thank you,
Vincenzo

iszmais:
That is correct,

therefore i will close this PR, but feel free to ask further questions.

Greetings,
@iszmais

PR #4169, opened 2022-03-23
WAC: Add workaround for anonymous having id 0

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=32284

### DIFF
--- a/Services/WebAccessChecker/classes/class.ilWebAccessChecker.php
+++ b/Services/WebAccessChecker/classes/class.ilWebAccessChecker.php
@@ -178,6 +178,8 @@ class ilWebAccessChecker
      */
     public function initILIAS()
     {
+        global $DIC;
+
         if ($this->isInitialized()) {
             return;
         }
@@ -212,6 +214,10 @@ class ilWebAccessChecker
             }
         }
         $this->setInitialized(true);
+
+        if ($DIC->user()->getId() === 0) {
+            $DIC->user()->setId(ANONYMOUS_USER_ID);
+        }
     }
 
 


Services/WebAccessChecker/classes/class.ilWebAccessChecker.php
```diff
@@ -212,6 +214,10 @@ public function initILIAS()
             }
         }
         $this->setInitialized(true);
+
+        if ($DIC->user()->getId() === 0) {
+            $DIC->user()->setId(ANONYMOUS_USER_ID);
```
chfsx, line 219: Hi @mjansenDatabay 
Thanks for the proposal! I'm currently not sure if one should set the users id manually to another, this may have some side-effects?

mjansenDatabay: Hi @chfsx ,

of course this is a workaround. The actual issue is: https://mantis.ilias.de/view.php?id=32063

There should not be a global user instance with ID 0, so this workaround simply sets the correct id for the context (user is not authenticated, so it must be an anonymous user). It is not a persistent change, only an 'in memory' state change of the user's id. Possible (I think mostly possitive) side effects are correct permission checks etc..

chfsx:
Hi @lscharmer and @mjansenDatabay, I rebased and merged this locally with an additional comment: https://github.com/ILIAS-eLearning/ILIAS/commit/25abbbab94c5a5d375aae8554a994d55840c0969
I'll cherry-pick this to trunk as well

PR #4172, opened 2022-03-23
PHP 8: Code Review `SOAPAuth`

mjansenDatabay:
Hello @pascalseeland, 

I completed the review of the `Services/SOAPAuth` component.

Results:

- [x] No usage of `$_GET` / `$_POST` / `$_REQUEST` / `$_SESSION`: I found 7 usages of `$_GET`.
- [ ] There is a `Unit Test Suite` with at least one unit test
- [ ] The unit tests pass

--

Actions needed:
- [x] Please check my inline comments. Some of them are just information, some require an action. You can use `TODO PHP8-REVIEW` when searching. The found super globals might be okay if there is a **good** reason to keep them.
- [ ] Please add a unit test suite with at least one passing unit test

Best regards,
@mjansenDatabay


pascalseeland:
I did create a PR against your branch resolving the all issues except for the test suite. Currently I don't see any suitable way how to test this code without doing a major rewrite.

PR #4177, opened 2022-03-23
PHP8: Code Review `ILIAS Page Editor`

lscharmer:
Hi @alex40724,

here are the results of the `Services/COPage` review.

### Results
- [ ] No usage of `$_GET` / `$_POST` / `$_REQUEST` / `$_SESSION`
- [x] The unit tests pass
- [ ] The types are fully documented (PHPDoc) or explict PHP types are used (type hints, return types, typed properties)

### Changes made in this PR:
- Added `// @TODO: PHP8 Review: Missing return type.`
- Added `// @TODO: PHP8 Review: Missing parameter type.`
- Added `// @TODO: PHP8 Review: Missing property type.`
- Added `// @TODO: PHP8 Review: Direct access to $_POST`
- Added `// @TODO: PHP8 Review: Direct access to $_GET`
- Added `// @TODO: PHP8 Review: switch with one case.`
- Fix `PSR 2 + X`

Best regards,
@lscharmer

### DIFF
--- a/Services/COPage/classes/class.ilPageObject.php
+++ b/Services/COPage/classes/class.ilPageObject.php
@@ -178,7 +178,7 @@ abstract class ilPageObject
         return $this->page_config;
     }
     
-    public static function randomhash()
+    public static function randomhash() // @TODO: PHP8 Review: Missing return type.
     {
         $random = new \ilRandom();
         return md5($random->int(1, 9999999) + str_replace(" ", "", (string) microtime()));
@@ -334,7 +334,7 @@ abstract class ilPageObject
     /**
      * @return bool|array
      */
-    public function buildDom(bool $a_force = false)
+    public function buildDom(bool $a_force = false) // @TODO: PHP8 Review: Missing return type.
     {
         if ($this->dom_builded && !$a_force) {
             return true;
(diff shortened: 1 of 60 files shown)


alex40724:
@lscharmer Thanks a lot for the review! I will definitely remove $_GET/$_POST accesses and missing types. Similar to the user component I may keep some simple array declarations (mostly if things are too much nested and should be replaced by data objects anyway). I will not change the "switch with one case". Not only because this is no MUST, but also because I prefer to have methods like executeCommand() in this style, since it is the "common structure" for these methods and quickly extensible, if new subclasses need to be added.

Services/COPage/classes/class.ilPageObject.php
```diff
@@ -2977,7 +2977,7 @@ public function deleteContent(string $a_hid, bool $a_update = true, string $a_pc
      * @return array|bool
      * @throws ilDateTimeException
      */
-    public function deleteContents(
+    public function deleteContents(// @TODO: PHP8 Review: Missing return type.
```
alex40724, line 2980: @lscharmer Just a remark, there are multiple cases where missing return types are listed, but the return type is part of the phpdoc comment.

alex40724:
@lscharmer , @mjansenDatabay There are around 25 "false positive" complaints about missing return types in this PR. In all these cases the return types are declared via phpdoc comment, all union/mixed types that cannot be declared by other means.

mjansenDatabay:
@alex40724 I'll have a chat with @lscharmer to prevent this in future reviews.

alex40724:
Thanks

PR #4178, opened 2022-03-23
Test&Assessment: fix for 27679

okaufman:
Fix for https://mantis.ilias.de/view.php?id=27679
test-statistics no longer show up when only results-permission is given


mbecker-databay:
Very good, thank you! 👍

PR #4189, opened 2022-03-24
Add PluginRender to unify UI Handling

iszmais:
This is the improved Version of #4118 wich will use the changes made to the [default renderer](https://github.com/ILIAS-eLearning/ILIAS/pull/4137) to add the chaining on the render progress for more flexibility.

### DIFF
--- a/src/UI/Implementation/Render/DecoratedRenderer.php
+++ b/src/UI/Implementation/Render/DecoratedRenderer.php
@@ -1,0 +1,89 @@
+<?php declare(strict_types=1);
+
+/******************************************************************************
+ *
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ *     https://www.ilias.de
+ *     https://github.com/ILIAS-eLearning
+ *
+ *****************************************************************************/
+
+namespace ILIAS\UI\Implementation\Render;
+
+use ILIAS\UI\Renderer;
+use ILIAS\UI\Component\Component;
+
+abstract class DecoratedRenderer implements Renderer
+{
+    private $default;
+
+    public function __construct(Renderer $default)
+    {
+        $this->default = $default;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function withAdditionalContext(Component $context) : DecoratedRenderer
+    {
+        $clone = clone $this;
+        $clone->default = $clone->default->withAdditionalContext($context);
+        return $clone;
+    }
+
+    /**
+     * Manipulates the rendering of one or multiple components by appending, prepending or exchanging their rendered
+     * content with custom adjustments.
+     *
+     * @return string|null Return the manipulated rendering of the component or NULL if the component should be
+     * rendered native
+     */
+    abstract protected function manipulateRendering($component, Renderer $root) : ?string;
+
+    /**
+     * Manipulates the async Rendering separately if needed.
+     * @see manipulateRendering
+     */
+    protected function manipulateAsyncRendering($component, Renderer $root) : ?string
+    {
+        return null;
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


Amstutz:
@iszmais , thx! I know this is still draft, so ping as soon as done for the review. One small thing, passing the complete dic into this object gives it for to much power and knowledge than supposed to. only pass waht is absolutely needed (probably only the default renderer stored in dic at this point in time in this case). If another plugin needs more, this can always be extended by this specific plugin.

iszmais:
@Amstutz Thx for the feed back. 

Unfortunatly i can not only pass the default renderer since they are actualy 2 renderers, one from the state while constructing (similar to the partial chain starting with the actual plugin) and one after all constructions are finished (similar to the whole chain) therefore i need to pass the dic so i can call the latest, complete closure to return the renderer later in render().
I could acces the global inside the funtcion if prefered, but i dont see any imporvement inside that.

Greetings,
@iszmais

klees:
Hi @iszmais!

If you give your `PluginRenderer` the [correct signature](https://github.com/ILIAS-eLearning/ILIAS/pull/4137/files#diff-fe419e589d70c46270a98f0acb34ad60d919754b5d24046534d3799cf803ee2eR26) and then copy [that construction from `DefaultRenderer`](https://github.com/ILIAS-eLearning/ILIAS/pull/4137/files#diff-c56a548132be9f8c8ccf6cde8287558f810c7a7fc0c8c97f18d4d67e4049b3f7R37), you won't need the dependency injection container to pass the root of the renderer chain.

Best regards!

iszmais:
@klees thx i forgot about that :smile: Fixed it. 

~~But i change the signature slightly to have the renderer as an requirement and not just optional since this will be the case in the default rendering where `this = root` most of the times, but i guess a plugin mostly should  ensure to set the correct root to get back to ILIAs default rendering process.~~
_Node: nvm this dosnt work in a typestrict scope_

iszmais:
@klees @Amstutz ,

i added 2 small test for the plugin renderer and fixed the typehints (see also https://github.com/ILIAS-eLearning/ILIAS/pull/4440 and https://github.com/ILIAS-eLearning/ILIAS/pull/4442).

Feel free to respond if you see something missing.

Greetings,
@iszmais

matthiaskunkel:
Jour Fixe, 02 MAY 2022 : We like the approach and waiting for a review by @klees and related improvements until we accept the PR for ILIAS.

iszmais:
@klees feel free to leave me a feedback of what is missing for you as said in the JF.

Feel free to call me if you think this will get more complex.

Greetings,
@iszmais

klees (dismissed):
Hi @iszmais,

thanks a lot for improving the infrastructure of the UI framework here!

- [x] `PluginRenderer`: Please rename the class to `RendererDecorator`. This is the pattern we use and we will help people to understand this if we name the class accordingly. `DecoratedRenderer` will work equally well, I guess. Also: The class won't be helpful for plugins only. Maybe the strict notion of a "plugin" vs "core" will vanish some time soon after all =)
- [x] Crucial Code: I think when we discussed the idea of the `RendererDecorator` we agreed that the [crucial ingredient of the `DefaultRenderer`](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/src/UI/Implementation/DefaultRenderer.php#L37) should be duplicated in the `RendererDecorator` and the consumer of the class should be made to implement an `abstract maybeRenderComponent(Component $component, Renderer $root) : ?string` instead. This would make it a lot less possible that people mess with the core mechanism, maybe we could even make `render` and `renderAsync` final.

Feel free to contact me on short notice if you want to discuss anything.	

Best regards!

iszmais:
Thx for the feedback @klees !

I went for the `DecoratedRenderer` naming and changed the added function to `manipulateRendering` instead of `maybeRenderComponent` to be more specific about its purpose (additional i added `manipulateRendering` for the asnyc usecase, if needed).

Feel free to give me further feebakc or to merge if you approve the made changes.

(Also feel free to take a look add the PRs for 7 and 8 wich have some minor changes for newer php versions)

Greetings,
@iszmais

klees (approved):

matthiaskunkel:
Jour Fixe, 16 MAY 2022 : Thanks for updating the PR. We highly appreciate this change and accept the PR for trunk. It would be great to notify plugin developers about the advantage of this change on the plugin page of the DevGuide.

PR #4195, opened 2022-03-25
PHP 8: Code Review `Poll`

mjansenDatabay:
Hi @smeyer-ilias / @tfamula,

I completed the review of the `Modules/Poll` component.

Results:
- [x] The code style is `PSR-2 + X` compliant: 1 File has been changed by the `PHP-CS-FIXER`
- [x] The unit tests pass

--

Actions needed:

- [ ] Please check my inline comments. Some of them are just information, some require an action. You can use `TODO PHP8-REVIEW` when searching.

Best regards,
@mjansenDatabay

### DIFF
--- a/Modules/Poll/classes/class.ilPollUserTableGUI.php
+++ b/Modules/Poll/classes/class.ilPollUserTableGUI.php
@@ -26,7 +26,7 @@ class ilPollUserTableGUI extends ilTable2GUI
 {
     protected array $answer_ids = [];
     
-    public function __construct(?object $a_parent_obj, string $a_parent_cmd)
+    public function __construct(object $a_parent_obj, string $a_parent_cmd)
     {
         global $DIC;
 
@@ -105,9 +105,9 @@ class ilPollUserTableGUI extends ilTable2GUI
         $a_csv->addColumn((string) ($a_set["firstname"] ?? ''));
         foreach ($this->answer_ids as $answer_id) {
             if ($a_set["answer" . $answer_id]) {
-                $a_csv->addColumn(true);
+                $a_csv->addColumn('true');// TODO PHP8-REVIEW I am not sure if my fix is correct
             } else {
-                $a_csv->addColumn(false);
+                $a_csv->addColumn('false');// TODO PHP8-REVIEW I am not sure if my fix is correct
             }
         }
         $a_csv->addRow();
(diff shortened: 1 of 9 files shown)


Modules/Poll/classes/class.ilPollUserTableGUI.php
```diff
@@ -105,9 +105,9 @@ protected function fillRowCSV(ilCSVWriter $a_csv, array $a_set) : void
         $a_csv->addColumn((string) ($a_set["firstname"] ?? ''));
         foreach ($this->answer_ids as $answer_id) {
             if ($a_set["answer" . $answer_id]) {
-                $a_csv->addColumn(true);
+                $a_csv->addColumn('true');// TODO PHP8-REVIEW I am not sure if my fix is correct
```
mjansenDatabay, line 108: I am not sure if my fix is correct

Modules/Poll/classes/class.ilPollUserTableGUI.php
```diff
@@ -105,9 +105,9 @@ protected function fillRowCSV(ilCSVWriter $a_csv, array $a_set) : void
...
             } else {
-                $a_csv->addColumn(false);
+                $a_csv->addColumn('false');// TODO PHP8-REVIEW I am not sure if my fix is correct
```
mjansenDatabay, line 110: I am not sure if my fix is correct

Modules/Poll/classes/class.ilPollUserTableGUI.php
smeyer-ilias: In addColumn() the csv-string is concatenated with the .= operator and thus an implicit string cast is performed. 
So I guess addColumn('1'); should be compatible with previous implementation.

Modules/Poll/classes/class.ilPollUserTableGUI.php
smeyer-ilias: In addColumn() the csv-string is concatenated with the .= operator and thus an implicit string cast is performed.
So I guess addColumn(''); should be compatible with previous implementation.

smeyer-ilias (changes requested):
Thanks, for reviewing and fixing issues.

smeyer-ilias:
I merge it and will fix the csv->addColumn issue.

PR #4212, opened 2022-03-28
PHP 8: Code Review `Wiki` Part III

mjansenDatabay:
Hi @alex40724,

I completed the 3rd part of the review of the `Modules/Wiki` component.

Results:
- [x] The code style is `PSR-2 + X` compliant: 1 File has been changed by the `PHP-CS-FIXER`
- [x] The unit tests pass

--

Actions needed:

- [x] I am not sure whether the libraries `Sanitize.php` and `Title.php` are compatible with PHP 8. Please check the list item if you already checked this.
- [ ] Please check my inline comments. Some of them are just information, some require an action. You can use `TODO PHP8-REVIEW` when searching.

Best regards,
@mjansenDatabay

### DIFF
--- a/Modules/Wiki/classes/class.ilObjWiki.php
+++ b/Modules/Wiki/classes/class.ilObjWiki.php
@@ -206,7 +206,7 @@ class ilObjWiki extends ilObject implements ilAdvancedMetaDataSubItems
             ));
 
         // create start page
-        if ($this->getStartPage() != "" && !$a_prevent_start_page_creation) {
+        if ($this->getStartPage() !== "" && !$a_prevent_start_page_creation) {
             $start_page = new ilWikiPage();
             $start_page->setWikiId($this->getId());
             $start_page->setTitle($this->getStartPage());
--- a/Modules/Wiki/classes/class.ilObjWiki.php
+++ b/Modules/Wiki/classes/class.ilObjWiki.php
@@ -333,7 +333,7 @@ class ilObjWiki extends ilObject implements ilAdvancedMetaDataSubItems
      */
     public static function _lookupRatingOverall(int $a_wiki_id) : bool
     {
-        return (bool) ilObjWiki::_lookup($a_wiki_id, "rating_overall");
+        return (bool) self::_lookup($a_wiki_id, "rating_overall");
     }
     
     /**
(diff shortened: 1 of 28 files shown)


Modules/Wiki/classes/class.ilObjWiki.php
```diff
@@ -769,13 +769,13 @@ public function createWikiPage(
         return $page;
     }
 
-    public static function getAdvMDSubItemTitle($a_obj_id, $a_sub_type, $a_sub_id)
+    public static function getAdvMDSubItemTitle($a_obj_id, $a_sub_type, $a_sub_id) : string // TODO PHP8-REVIEW Type hints are missing here
```
mjansenDatabay, line 772: Type hints are missing here

Modules/Wiki/classes/class.ilWikiStat.php
```diff
@@ -758,15 +758,15 @@ protected static function getWikiAggrSub(
...
+                        $res[$day] = array_sum($values) / count($values);
                         break;
                     
                     case "SUM(%s)":
                         $res[$day] = array_sum($values);
                         break;
                     
                     default:
-                        var_dump("unsupport aggr " . $a_aggr_value);
+                        var_dump("unsupport aggr " . $a_aggr_value);// TODO PHP8-REVIEW We should not use debugging code here
```
mjansenDatabay, line 769: We should not use debugging code here

Modules/Wiki/classes/class.ilWikiUtil.php
```diff
@@ -447,7 +446,7 @@ public static function splitTrail(
 
     public static function sendNotification(
         string $a_action,
-        string $a_type,
+        string $a_type, // TODO PHP8-REVIEW Type is a `string` but is passed to methods expecting an `int`
```
mjansenDatabay, line 449: Type is a `string` but is passed to methods expecting an `int`

Modules/Wiki/classes/class.ilObjWikiGUI.php
```diff
@@ -1631,7 +1629,7 @@ public function exportHTML() : void
      * Get title for wiki page (used in ilNotesGUI)
      */
     public static function lookupSubObjectTitle(
-        string $a_wiki_id,
+        string $a_wiki_id,// TODO PHP8-REVIEW This is a `string`, but it is passed to a constructor expecting and `int`. Furthermore it is compared to an `int` below.
```
mjansenDatabay, line 1632: This is a `string`, but it is passed to a constructor expecting and `int`. Furthermore it is compared to an `int` below.

Modules/Wiki/classes/class.ilObjWiki.php
alex40724: They are missing in the metadata service.

alex40724:
Fixed everything, except the metadata call, which needs to be fixed in Services/Medata...

PR #4219, opened 2022-03-28
PHP8: Code Review `Modules/ItemGroup`

kergomard:
Hi @alex40724 

Here are the results of the `Modules/ItemGroup` review.

### Results

- [x] The code style is PSR-2 + X compliant: _CS-Fixer made changes to 5 files_
- [x] No usage of $_GET / $_POST / $_REQUEST / $_SESSION
- [x] The unit tests pass

### Comments
- The `Services`-Folder is very much a construction site. Most of what is in there doesn't (yet) do anything. How about removing all the classes without function from this version? They can easily be reintroduced after the Split of release_8.
- This type here is rather under specific: https://github.com/ILIAS-eLearning/ILIAS/blob/0c481000999db16b07030815f91c010cacfaa379/Modules/ItemGroup/classes/class.ilItemGroupItemsTableGUI.php#L29 . Can't we make this more specific or at least add a comment to specify what Object we are expecting?

### Changes made in this PR:

- Some small changes in types and a // @TODO: PHP8 Review: Missing parameter type.

Best regards,
@swiniker

### DIFF
--- a/Modules/ItemGroup/classes/class.ilObjItemGroup.php
+++ b/Modules/ItemGroup/classes/class.ilObjItemGroup.php
@@ -100,12 +100,12 @@ class ilObjItemGroup extends ilObject2
         return $this->item_data_ar->getBehaviour();
     }
 
-    protected function doRead()
+    protected function doRead() : void
     {
         $this->item_data_ar = new ilItemGroupAR($this->getId());
     }
 
-    protected function doCreate()
+    protected function doCreate() : void
     {
         if ($this->getId()) {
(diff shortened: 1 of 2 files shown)


alex40724:
- doCloneObject is still missing types in the parent class, so we can't add them here without producing errors.
- I specified a distinct class in the table gui now
- the Service directory marks an important design decision for the component, so I will keep this. The roadmap includes a point to follow up on this.

alex40724:
Thanks for the review PR!

PR #4220, opened 2022-03-28
Restore page scroll on mobile #26915

catenglaender:
# Issue

Mantis Issue: https://mantis.ilias.de/view.php?id=26915

**Mobile browsers usually hide the address bar and other system or browser UI elements after a certain scrolling distanc**e on the webpage. In ILIAS those elements didn't disappear as they do on other websites.

This happens because the main container in ILIAS is affixed in a CSS grid. Instead of scrolling the actual page, the user would scroll inside a container on an unmoving page. This already caused the need for other fixes e.g. https://mantis.ilias.de/view.php?id=27270 to enable scrolling with the keyboard arrows.

# Changes

All of these changes effect only the mobile breakpoint:
* Switched the layout from a CSS grid to a more conventional setup that keeps **as many elements as possible in the document flow**
* The **overlapping header is solved through position: sticky**, which keeps it inside the document flow until scrolling begins. I chose this over position: fixed so we don't need to reserve any space through margin/padding-top "under" the header. While the legacy IE is unsupported, big websites like spiegel.de are using position: sticky for their header.
* Because of the mainbar's middle position in the DOM structure position: sticky cannot be used for it. Therefor, the mainbar is ripped out of the document flow with position: fixed and a padding-bottom creates sufficient space so that the footer can scroll all the way up from under the mainbar. Maybe this can be improved if we also revise the desktop breakpoint.
* The main container now uses a flexbox layout to push the footer to the bottom as the complexitiy of a grid with columns isn't needed for this screen size. This also might be worth considering for the desktop version.
* With a normal document flow restored, the Javascript put in place to fix https://mantis.ilias.de/view.php?id=27270 caused the mobile page to scroll down to the upper edge of the main container, which caused the page to jump and looked like the header was overlapping the page title immediately after loading. This is why the script now checks for screen size and causes the focus switch on desktop only.

# Test Pages

Page with some content to test scrolling: https://dev.cate-tms.de/ilias7mobile/goto.php?target=cat_85&client_id=ilias
Empty page to test fixed footer: https://dev.cate-tms.de/ilias7mobile/goto.php?target=cat_86&client_id=ilias

# Outlook and bigger context

I discussed this issue with @catkrahl who worked on the initial implementation of the CSS grid. We debated if keeping the CSS grid on desktop is even worth it as detecting alternative layouts (e.g. kiosk mode) are tricky to handle. Having two plugins adding elements to the standard page could cause unsolvable problems as the CSS grid layout can only handle one overwrite at a time.

In the long term, we might want to look into restoring a document flow that could handle the re-ordering and inserting elements more dynamically - maybe for the whole page, maybe just for the inner grids.

### DIFF
--- a/src/UI/templates/default/Layout/standardpage.less
+++ b/src/UI/templates/default/Layout/standardpage.less
@@ -327,19 +327,14 @@ footer {
 	// main grid without slates
 	.il-layout-page {
 		background: @body-bg;
-		display: -ms-grid;
-		display: grid;
-		grid-gap: 0px;
-		-ms-grid-columns: 1fr;
-	    grid-template-columns: 1fr;
-		-ms-grid-rows: @il-standard-page-small-header-height auto 1fr;
-	    grid-template-rows: @il-standard-page-small-header-height auto 1fr;
-	    height: 100%;
-		overflow: hidden;
+		display: block;
+		overflow: initial;
 		width: 100%;
 		.nav.il-maincontrols {
-			grid-row: 4;
-			grid-column: 1;
+			position: fixed;
+			bottom: 0;
+			z-index: @il-standard-page-zindex-maincontrols;
+			width: 100%;
 			.il-maincontrols-mainbar {
 				display: grid;
 				display: -ms-grid;
@@ -374,7 +369,7 @@ footer {
 				-ms-grid-row: 3;
 				grid-column: 1;
 				grid-row: 3;
-				height: 100%; // calc(100% - (@il-standard-page-small-header-height + @il-standard-page-breadcrumbs-height);
+				height: calc(100% - @il-standard-page-small-header-height);
 				.il-maincontrols-mainbar {
 					-ms-grid-columns: 1fr;
 				    grid-template-columns: 1fr;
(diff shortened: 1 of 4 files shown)


Amstutz:
Hi @catenglaender 

Thx a lot for this. Special thx for the detailed explanation! For me, those were indeed needed, since I would have had a hard time in understanding everything going on here, due to my limitations in CSS insight ;-).

Are you joining the CSS Squad today? If so, it would be great if you can walk us through the changes. Since this will have some footprint, and should be tested thoroughly before going into production, we could also come up with a plan on how to test this best. Also, I do have some questions, which we could discuss in this session.

One note: The unrelated changes, due to indentations, made this somewhat hard to review. Best, make a separate PR for such changes. They are highly welcome, but should be kept out of other PRs to make them easier to read.

Thx.
@Amstutz

catenglaender:
Hi @Amstutz 

thx for the comment. Yes, I will join todays CSS squad, so we can discuss everything in more detail.

I reverted to the original indentation. It does make the diff a lot easier to read.

All the best,
Ferdinand

Amstutz:
We discussed the issue today in the CSS-Squad. Thx a lot @catenglaender , for this change! I, personally, regret removing the grid system for mobile devices, but your reasoning does absolutely make sense.

Amstutz:
We also picked the change to trunk, thx a lot!

PR #4228, opened 2022-03-29
PHP8: Code Review `ILIAS Page Editor`

lscharmer:
Hi @alex40724,

sorry for the inconvenience, here is the last last part of the `Services/COPage` review.

### Results
- [ ] There are no serious `PHPStorm Code Inspection` issues left
- [ ] External libraries are compatible with PHP 8 (if relevant, see my comment below)

### Changes made in this PR
- Added `// @TODO: PHP8 Review: Array to string conversion.`
- Added `// @TODO: PHP8 Review: Missing argument.`
- Added `// @TODO: PHP8 Review: Invalid argument.`
- Added `// @TODO: PHP8 Review: Void result used.`
- Added `// @TODO: PHP8 Review: Undefined method.`
- Added `// @TODO: PHP8 Review: Missing return statement.`
- Added `// @TODO: PHP8 Review: Undefined class.`
- Added `// @TODO: PHP8 Review: Array is always empty`
- Added `// @TODO: PHP8 Review: Condition is always false`
- Added `// @TODO: PHP8 Review: Condition is always true`
- Added `// @TODO: PHP8 Review: This is either null or an object, but the object does not implement __toString)`

I am not sure whether `Services/COPage/mediawikidiff/class.WordLevelDiff.php` is compatible with PHP 8. Please check the list item above if you already checked this. There are **a lot** of dynamically declared properties in the library, which will result in issues.

Best regards,
@lscharmer

### DIFF
--- a/Services/COPage/classes/class.ilPCParagraphGUI.php
+++ b/Services/COPage/classes/class.ilPCParagraphGUI.php
@@ -13,7 +13,7 @@
  * https://github.com/ILIAS-eLearning
  */
 
-use \ILIAS\Style;
+use ILIAS\Style;
 
 /**
  * Class ilPCParagraphGUI
--- a/Services/COPage/classes/class.ilPCParagraphGUI.php
+++ b/Services/COPage/classes/class.ilPCParagraphGUI.php
@@ -652,7 +652,7 @@ class ilPCParagraphGUI extends ilPageContentGUI
         //$selection->setSelectedValue($a_selected);
         $selection->setUseImages(false);
         $selection->setOnClickMode(ilAdvancedSelectionListGUI::ON_ITEM_CLICK_NOP);
-        if (is_string($a_use_callback)) {
+        if (is_string($a_use_callback)) {// @TODO: PHP8 Review: Thje
             $selection->setSelectCallback($a_use_callback);
         } elseif ($a_use_callback === true) {
             $selection->setSelectCallback("ilCOPage.setCharacterClass");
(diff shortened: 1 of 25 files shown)


Services/COPage/classes/class.ilPCParagraphGUI.php
```diff
@@ -652,7 +652,7 @@ public static function getCharStyleSelector(
         //$selection->setSelectedValue($a_selected);
         $selection->setUseImages(false);
         $selection->setOnClickMode(ilAdvancedSelectionListGUI::ON_ITEM_CLICK_NOP);
-        if (is_string($a_use_callback)) {
+        if (is_string($a_use_callback)) {// @TODO: PHP8 Review: Thje
```
alex40724, line 655: ??

alex40724:
> There are a lot of dynamically declared properties in the library, which will result in issues.

What issues do you fear? I agree that we need to find an alternative for this, but there are no plans I know of to remove support for dynamically declared properties in PHP completely. The opt-in solution seems the way to go: https://wiki.php.net/rfc/deprecate_dynamic_properties

alex40724:
Thanks for the PR.

Services/COPage/classes/class.ilPCParagraphGUI.php
mjansenDatabay: @alex40724 The `is_string` check does not make sense, because `$a_use_callback` it is **always** a `string`.

mjansenDatabay:
@alex40724 

> What issues do you fear? 

You are right regarding the `undeclared properties`, these will not result in an error. Of course we should get rid of this type of issue in OUR code base, we don't need to fix library code.

But:
- There are undefined classes used in this class, e.g. `Revision`
- There are a lot of undefined functions used
- There a (probably) some undefined variables

I don't know if these code branches are really executed in your component.

All we wanted to say: This library looks like there could be some issues, maybe not now, but in the future :-).

alex40724:
@mjansenDatabay Thanks. Yes, the code needs some action in the future. It's on the ROADMAP already. I fixed the other remaining issues.

PR #4229, opened 2022-03-29
UI Components, Required Tag, see #32305

Amstutz:
Empty string are not allowed for required Tags, see: https://mantis.ilias.de/view.php?id=32305 , including prooper translation of error.


klees (approved):
LGTM, thx. Nice simplification of the check =)

We would want to pull this down to 7, right?

Best regards!

Amstutz:
yes, needs to go down to 7.

PR #4232, opened 2022-03-29
Language/CI 32318, determine header length

nhaagen:
Sorting-script messed up for headers other than 16 lines.
https://mantis.ilias.de/view.php?id=32318


nhaagen:
The later version is w/o awk;  I'd really like to use `line=$(head $file -n ${header_length} | tail -n1 | tr -d '\n' | tr -d '\r')` at least, but I am unsure of windows systems.
With the current version, all files are fine, except _hr and _ja. Those have windowsCR line endings (^M), and this does not compare well.
It is enough though, to remove the trailing char from the `<!-- language file start -->`-marker.

nhaagen:
@matthiaskunkel since the bug is closed already - this still belongs to the fix. 
Also, the translators of hr/ja should probably be noticed, right?
The current script accepts any length headers (up to 64 lines, i.e.) individually for single files.

matthiaskunkel:
Thanks a lot, @nhaagen, for providing this PR so quickly. I have already merged it to trunk.

PR #4223, opened 2022-03-29
PHP 8: Code Review `SystemCheck`

mjansenDatabay:
Hello @smeyer-ilias, 

I completed the review of the `Services/SystemCheck` component.

Results:

- [ ] The code style is `PSR-2 + X` compliant
- [ ] No usage of `$_GET` / `$_POST` / `$_REQUEST` / `$_SESSION`
- [ ] There is a `Unit Test Suite` with at least one unit test
- [ ] The unit tests pass
- [x] External libraries are compatible with PHP 8 (if relevant)
- [ ] There are no serious `PhpStorm Code Inspection` issues left
- [ ] The types are fully documented (PhpDoc) or explict PHP types are used (type hints, return types, typed properties)

--

Actions needed:
- [ ] Please add a unit test suite with at least one passing unit test

Best regards,
@mjansenDatabay

PR #4234, opened 2022-03-29
php8 code review statistics and lp

sailorsilv:
php8 code review statistics and learning progress


mjansenDatabay:
@katringross / @sailorsilv :

- There are still many `PSR-2 + X` violations in the code, even after this PR gets applied.
- There are still many usages of super global variables, which MUST be reported.
- Furthermore I found dozens of errors regarding missing type declarations, which are not reported here.
- You used `Union Types` and `mixed` in your PR. The code MUST be compatible with PHP 7.4.

I strongly recommend to use checklists like used here (for instance): https://github.com/ILIAS-eLearning/ILIAS/pull/4223

Please contact me if you need further help/assistance.

chfsx:
From Technical Board: we close this PR, because it must not be merged in this state. this is because of the technically wrong changes like `Union Types` or `mixed` return types, which would break the code immediately.

PR #4236, opened 2022-03-29
Allow blank linkes in PHPDoc comments.

thibsy:
A known issue with PHPStorm when editing the Kitchensink interfaces is that all blank lines will be removed when the PHPDoc string was modified (even if the reformatting is only applied for "changed lines"). This change would "fix" that behaviour by allowing blank lines in PHPDoc comments.

### DIFF
--- a/docs/development/code-style-configs/php-storm.xml
+++ b/docs/development/code-style-configs/php-storm.xml
@@ -1,8 +1,8 @@
-<code_scheme name="ILIAS Coding Style 4/2020" version="174">
+<code_scheme name="ILIAS Coding Style 3/2022" version="175">
   <PHPCodeStyleSettings>
     <option name="ALIGN_PHPDOC_PARAM_NAMES" value="true" />
     <option name="ALIGN_PHPDOC_COMMENTS" value="true" />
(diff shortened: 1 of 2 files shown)


Amstutz (approved):
Thx a lot for looking into this. All good for me. 

Does the autoformatting in PHP-Storms a good job for you by now? I changed some month ago to using our profile for inspetions and formatting this way. (Yellow indication appears, if there is an issue, if multiple issues appear, the whole file can be fixed with one click). Works like a charm.

thibsy:
No problem, @Amstutz 

I wasn't sure about the workflow for this config, should I adjust the version attribute and date (m/Y) in the name of the `code_scheme` tag as well?

The formatting works nicely with PHPStorm (except for this issue) but I didn't know about the inspections profile until a few days ago, so I can't really say much about that yet.

Amstutz (approved):
Yeah, changing date and version would be best. Thx for pointing that out! I do not think we have a strict procedure for those files yet. They are meant to help devs. and everybody is invited to improve them. Exactly what you did here, so Kudos!

Reviews and responsibility for the whole docs folder are provided by members of the tb up to now.

PR #4241, opened 2022-03-30
Group: PHP8 review

daniwe4:
# Must-Package
## Goals
 * [ ] ILIAS can be used with PHP 8 without producing PHP-specific errors.
   * can't be tested until object creation fails
 * [x] Refactored code conforms to current coding styles
 * [ ] Elimination of all warnings and errors of the static code inspection (without weak | incl. undeclared variables)
   * 1 error = missing constant GRP_REGISTRATION_DEACTIVATED in ilObjGroup
   * 11 warnings
   * 72  weak warnings
   * can't fix this on the fly, because there are missing constants
 * [ ] Transfer of GET params into Constructor
   * have a look at ilGroupAddToGroupActionGUI line 299
 * [x] Avoidance/reduction of POST accesses (e.g. through $form->getInput()) and switch to Request classes
 * [x] Use of SESSION (session object, alternative assignment of static session in constructor member variables)
 * [x] Documentation of all existing typifications as typehints (of parameters, return values and properties)
 * [x] Creation of a unit test infrastructure in the component

## Requirements and analysis criteria
* [x] Preserve unit test coverage even in refactoring.
* [ ]  Simple call of all tabs/screens (without e.g. delete and move actions) and fixing of the corresponding bugs
  * can't create Group objects, because of missing constants
* [x]  Checking of external libraries
* [x]  Basis of the refactoring and the reviews is, among others, ILIAS Development Documentation.

# Should-Package
### What do we need to do in the mid-term to catch recurring large code revisions and to refactor more than the minimum?

 * [x] Expansion of the unit test coverage
  * General and structural improvements and optimisations to the code base. For example ...
   * [ ]  ... in static functions and by removing static methods
   * [x]  ... by introducing return type declarations
   * ... by typification of all variables
     * [x] Input/output typing of methods
     * [x] Typification of member variables
 * [ ] Use of UI components from the UI framework "KitchenSink" that have been introduced and established in the meantime. Reduction of legacy UI and legacy code.

# Sustainability-Package
### he developer community has identified topics that would also improve the quality of ILIAS from the point of view of PHP and ILIAS in general. Ideally, these topics could be directly addressed in the process of systematic refactoring.

* [x] Introduction and use of declarations of "strict_types" for data
* [ ]  Replace previous arrays with classes that enable ArrayAccess - in future also getter / setter
* [ ]  Switching previous arrays to generators/data objects

### DIFF
--- a/Modules/Group/UserActions/classes/class.ilGroupActionTargetExplorerGUI.php
+++ b/Modules/Group/UserActions/classes/class.ilGroupActionTargetExplorerGUI.php
@@ -36,7 +36,7 @@ class ilGroupActionTargetExplorerGUI extends ilRepositorySelectorExplorerGUI
         }
     }
 
-    public function setClickableType(string $a_val)
+    public function setClickableType(string $a_val) : void
     {
         $this->clickable_type = $a_val;
     }
--- a/Modules/Group/UserActions/classes/class.ilGroupAddToGroupActionGUI.php
+++ b/Modules/Group/UserActions/classes/class.ilGroupAddToGroupActionGUI.php
@@ -109,14 +109,13 @@ class ilGroupAddToGroupActionGUI
     }
     public function show() : void
     {
-        $tpl = $this->tpl;
         $lng = $this->lng;
         $ctrl = $this->ctrl;
 
         $toolbar = new ilToolbarGUI();
 
         // button use existing group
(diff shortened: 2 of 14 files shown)


smeyer-ilias (approved):

smeyer-ilias:
Thanks for reviewing and fixing issues. 
I will fix the remaining issues directly in the trunk

PR #4246, opened 2022-03-30
Hotfix for missing translations when importing a course object.

thibsy:
Hello @smeyer-ilias

This will probably not be merged, since you'd have to implement this for every other object too, but it would be a hotfix in order to import courses which have multilingualism active.

This is sort of the opposite of https://mantis.ilias.de/view.php?id=32216.

Kind regards.


kergomard:
Hi @smeyer-ilias  

Best regards!

alex40724:
This is a general object related translation issue, see https://github.com/ILIAS-eLearning/ILIAS/pull/5838 which needs much less code changes to fix this.

PR #4249, opened 2022-03-31
TA, minor changes in handling requests (MultipleChoice)

nhaagen:
Hi @mbecker-databay  ,
as announcend, some things I had to knock out of the way to create a MC-Question.
All in the unclean, but hoping it helps anyway ;)


mbecker-databay:
Good with it, thank you so much!

PR #4256, opened 2022-03-31
Correct type for formSelect

kergomard:
This fixes an issue with the 8th parameter sent to formSelect in Duration Input.


alex40724:
Thanks!

PR #4261, opened 2022-04-01
Fix Tab-Focus in Exercise File Upload

kergomard:
This fixes the focus on the file upload in the exercise. I kept it as specific as possible to avoid any side-effects.
This is a follow up to https://github.com/ILIAS-eLearning/ILIAS/pull/4100 and fixes https://mantis.ilias.de/view.php?id=32359.


Amstutz:
Thx a lot for this fix!

Amstutz:
picked to trunk as well.

PR #4267, opened 2022-04-01
Empty Heading on Infoscreen of Text and Survey

Amstutz:
Currently we have a wcag failing issue on the info screen of the Test and the Survey if no description is provided. There will be an empty h2 heading. See: https://mantis.ilias.de/view.php?id=32255 and https://mantis.ilias.de/view.php?id=32365 . 

Easiest (and maybe best) solution would be to just provide a proper heading. Another option would be to provide s sr-only heading or provide no h2 heading at all in those cases. If @alex40724 prefers this, I will assign the issue to him, since there would be some moving around of the InfoScreenGUI be required to check if the heading is required/sr-only.

The change would look as follows:
image


Amstutz:
I put the related mantis report on JF, hope this is fine.

alex40724 (approved):
Survey changes are ok for me.

matthiaskunkel:
Jour Fixe, 04 APR 2022: We highly appreciate the suggested PR. Please merge to release_7 and trunk.

PR #4271, opened 2022-04-01
A11y Actions Menu is not available by keyboard, see #32257

Amstutz:
Issue seems to be, that if the advanced table is rendered as ilAdvancedSelectionListGUI::STYLE_LINK, then we have a <a> tag with no href and therefore no tabbing per default. See issue: https://mantis.ilias.de/view.php?id=32257#bugnotes as an example


Amstutz:
@alex40724 : could you have a look at this? This currently blocks several TA pages from being error free in a11y audits.

alex40724:
Merged to trunk as well.

PR #4272, opened 2022-04-02
Review services `Services/Init`

kergomard:
Hi @smeyer-ilias  

2022-04-11: Sorry, just added one more commit with more fixes and ToDos after rechecking.

Here are the results of the `ServicesInit` review.

### Results

- [ ] The code style is PSR-2 + X compliant: _CS-Fixer made changes to 4 files_
- [ ] No usage of $_GET / $_POST / $_REQUEST / $_SESSION: _There are quite a few superglobals used, it is hard for me in a useful time frame to judge which ones are necessary in this context. I removed two accesses I was pretty sure we could easily get rid of._
- [x] There is a Unit Test Suite with at least one unit test: _Did you not add the tests for the UIFramework to the Suite on purpose?_
- [x] The unit tests pass

### Comments
- There are quite a few globals used. Do you really need to put all this in the global scope? We e.g. have two different variables in global scope for log: $log and $ilLog. Do we really need both of them and can we not take one or both of them out of the global scope? An other example would be the ilUser here: https://github.com/ILIAS-eLearning/ILIAS/pull/4272/files#diff-5c2233ae6af0264344f1d3415e104429899c63359ffb40cd2503cbd7fd44d894L1524
- Please also see the inline comment

### Changes made in this PR:
- Fixed CS
- Added some types, removed some duplicate doc-strings, and moved some declarations from doc-strings.
- Added a few // @TODO PHP8: - Comments

Best regards,
@swiniker

### DIFF
--- a/Services/Init/classes/Dependencies/InitComponentService.php
+++ b/Services/Init/classes/Dependencies/InitComponentService.php
@@ -9,9 +9,9 @@ class InitComponentService
     {
         $int = $this->initInternal($c);
 
-        $c["component.repository"] = fn($c) : \ilComponentRepository => $int["db_write"];
+        $c["component.repository"] = fn ($c) : \ilComponentRepository => $int["db_write"];
 
-        $c["component.factory"] = fn($c) : \ilComponentFactory => new ilComponentFactoryImplementation(
+        $c["component.factory"] = fn ($c) : \ilComponentFactory => new ilComponentFactoryImplementation(
             $int["db_write"],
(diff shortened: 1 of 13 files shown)


kergomard:
@smeyer-ilias : Do you need anything more from me to follow up on this review?

smeyer-ilias (approved):
Thanks for reviewing and fixing issues. 
I will merge this PR and and create a new PR which will be assigned to the new maintainer @pascalseeland

PR #4274, opened 2022-04-04
EventHandler: Improve logging of parameters

mjansenDatabay:
With ILIAS 8 the logging methods (`debug`, `info`, ...) have been changed to only accept strings. Therefore the anonymous class used for `lazy evaluation` of complex data structure logging could not be used anymore.
To comply with the new method signatures, the `__toString()` method was immediately called. Since other code modifications in the past changed the behaviour to only log the amount of parameters, the initial intention of this approach got lost at all anyway.

This PR suggests a change to (again) log the parameter array. Instead of recursive logging of objects, only the classname is exported.


alex40724:
Thanks!

PR #4276, opened 2022-04-04
PHP 8: Code Review `soap`

mjansenDatabay:
Hi @smeyer-ilias,

I completed the review of the `webservice/soap` component.

Results:
- [ ] The code style is `PSR-2 + X` compliant
- [ ] No usage of `$_GET` / `$_POST` / `$_REQUEST` / `$_SESSION`: I found 6 usages of `$_GET`
- [ ] There is a `Unit Test Suite` with at least one unit test
- [ ] The unit tests pass
- [x] External libraries are compatible with PHP 8 (if relevant)
- [ ] There are no serious `PhpStorm Code Inspection` issues left
- [ ] The types are fully documented (PhpDoc) or explict PHP types are used (type hints, return types, typed properties)

I already fixed many trivial inspection profile issues.

--

Actions required:
- [ ] Please eliminate the super global variable usages.  The found super globals might be okay if there is a **good** reason to keep them (if so, please elaborate).
- [ ] Please add a unit test suite with at least one passing unit test
- [ ] I am not sure whether `webservice/soap/lib/nusoap.php` is compatible with PHP 8. Please check the list item above if you already checked this.
- [x] Please check my inline comments. Some of them might be just information, some require an action. You can use `TODO PHP8-REVIEW` when searching. I used the following types:
  - `// TODO PHP8-REVIEW Property dynamically declared`
  - `// TODO PHP8-REVIEW Method undefined`
  - `// TODO PHP8-REVIEW Constant undefined`
  - `// TODO PHP8-REVIEW Return type missing`
  - `// TODO PHP8-REVIEW Type hint missing`
  - `// TODO PHP8-REVIEW Missing argument`
  - `// TODO PHP8-REVIEW Wrong type`
  - `// TODO PHP8-REVIEW Unnecessary operations`
  - `// TODO PHP8-REVIEW Missing return value`

--

Actions suggested:
- [ ] You could remove many `(require|include)_once` directives

--

Summary:

- [ ] webservice/soap/include/inc.soap_functions.php
  - Return type declarations are missing for all methods. If the return value could be of different types, use `PHPDoc` comments instead, e.g. `@return bool|soap_fault|SoapFault`
  - Type hints are missing for all methods below `\ilSoapFunctions::deleteRole`
  - The `PHPDoc` comments are often wrong, e.g. `\ilSoapFunctions::moveObject`
  - Suggestion: `array` types could be described in regards to their shape (e.g. `string[]`, or `array{type: int, child: int}`)

[...]

PR #4282, opened 2022-04-05
Fixing php issues reported by PHP EA Inspection

pascalseeland:
@smeyer-ilias

I did remove some code from the ilSoapUtils which I think is a leftover and was partially already removed. Could you please check an if this fine with you and if yes please merge this pr.


mjansenDatabay:
@smeyer-ilias It would be great if you could review/merge the changes from @pascalseeland , because my review of the `ECS` component in the `PHP 8 Project` depends on it. I would like to prevent redundant reviews.

And: Maybe #4276 should be merged/handled first, because this PR here modifies a SOAP file (https://github.com/ILIAS-eLearning/ILIAS/pull/4282/files#diff-ebdb670209e355cc813aeccb415126c4049ddb93f69d972502f808d66e105ba0).

webservice/soap/classes/class.ilSoapUtils.php
```diff
@@ -537,31 +537,6 @@ private function linkNode(array $node, ilCopyWizardOptions $cp_options) : int
         return $new_ref_id;
     }
 
-    /**
-     * @todo get rid of the soap wrapper
-     */
-    public function handleECSTasks(string $sid, int $a_server_id)
```
smeyer-ilias, line 537: It's not just this method that needs to be removed. In addition, the WSDL generation and other places must also be adjusted. I will create a PR for this.

smeyer-ilias (approved):
The changes in the SOAP subclass are fine for me.

smeyer-ilias:
There was only one remaining part in class.ilNusoapAdministrationAdapter.php
I fixed this in trunk

PR #4284, opened 2022-04-05
PHP 8 Code Review Individual Assessment

katringross:
Hello @klees ,

we completed the review of the Individual Assessment component.

Results:

- [x]  The code style is PSR-2 + X compliant
- [x]  No usage of $_GET / $_POST / $_REQUEST / $_SESSION: We found 7 usages of $_SESSION, 13 usages of $_GET and 40 usages of $_POST.
- [x]  There is a Unit Test Suite with at least one unit test
- [x]  The unit tests pass
- [x]  External libraries are compatible with PHP 8 (if relevant)
- [x]  There are no serious PhpStorm Code Inspection issues left
- [x]  The types are fully documented (PhpDoc) or explict PHP types are used (type hints, return types, typed properties)

--

Please check our inline comments most of them might be just information. You can use PHP8-Review when searching.

Best regards

@katringross


klees:
Thanks a bunch!

Modules/IndividualAssessment/classes/class.ilIndividualAssessmentExporter.php
```diff
@@ -54,6 +54,7 @@ public function getValidSchemaVersions(string $a_entity) : array
     {
         return [
             "5.2.0" => [
+                // PHP8-Review: Link with unencrypted protocol
```
klees, line 57: This is interesting, I don't exactly understand if the protocol is really relevant here. Since this looks like some naming related stuff and does not seem to be an URL someone actually calls, I expect TLS to be irrelevant. On the other hand, switching to `https` might disturb some mechanism. I really don't know. So we might want to discuss this on the PHP 8 meeting.

Modules/IndividualAssessment/classes/class.ilIndividualAssessmentDataSet.php
```diff
@@ -12,14 +12,18 @@ public function getSupportedVersions() : array
         return ['5.2.0', '5.3.0'];
     }
 
+    // PHP8-Review: Method visibility should not be overridden
+    // PHP8-Review: Parameter's name changed during inheritance
```
klees, line 16: This is also interesting. Parameter names didn't matter a long time in PHP, as long as the signature was intact. [Now they might start to matter with PHP 8](https://wiki.php.net/rfc/named_params). I will as the @ILIAS-eLearning/technical-board for some advice if and how we would want to use named params in ILIAS.

Modules/IndividualAssessment/classes/class.ilObjIndividualAssessmentGUI.php
```diff
@@ -381,6 +381,7 @@ protected function getEntryForStatus(int $status) : string
         }
     }
 
+    // PHP8-Review: Parameter's name changed during inheritance
```
klees, line 384: @katringross I might miss something, but to me the parameter in `ilObjectGUI` seems to be named the same.

klees:
According changes in https://github.com/ILIAS-eLearning/ILIAS/commit/d96aa33db85d3dfa1bdbd165c5f43ddd123b2ef3

Modules/IndividualAssessment/classes/class.ilObjIndividualAssessmentGUI.php
katringross: @klees Hmm I think you are right, we used PHPStan, maybe a false positive...

Modules/IndividualAssessment/classes/class.ilIndividualAssessmentExporter.php
klees: As discussed in the latest PHP 8-Meeting: I will not change this =)

PR #4306, opened 2022-04-06
KS Video Player Element (v2)

alex40724:
This is a follow-up of PR https://github.com/ILIAS-eLearning/ILIAS/pull/4041

- The withSource mutator is gone.
- The binary file has been replaced by a reference to the external resource
- Video element is grouped as a Player
- License fixed
- checkStringArg removed
- source and poster parameter details added
- redundant Docstrings removed
- tests added

> Why do we need the additional div around the video-element?

The mediaelement.js library replaces the video element with all kind of new dom nodes. The purpose of the surrounding div is to make the element addressable by ILIAS CSS, even if the default here does not come with any CSS rules. Imagine someone would like to give all video elements a certain border or margin, this may become error-prone if relying on the mediaelement.js DOM nodes, which might change from version to version.

> A11y: Consider adding an alternative audio source as additional property, so consumers have a way to implement [WCAG success criterion "Audio-only and Video-only (Prerecorded)"](https://www.w3.org/WAI/WCAG21/quickref/#audio-only-and-video-only-prerecorded). This also should be another rule then.

I am not sure, how the component itself should tackle this. Please note that the criterion addresses **video-only** content:
https://www.w3.org/WAI/WCAG21/Understanding/audio-only-and-video-only-prerecorded#dfn-video-only
Meaning a video without an audio track. Usually you have an audio track in your mp4 file making it **not** video-only content. Additionally we have the VTT files that enable to present a text alternative. We could add a SHOULD rule that an audio track should be provided and for video-only files the context needs to provide alternatives (e.g. the page editor allows to add an additional audio element). But I wouldn't add additional mp3 files to this component somehow.

### DIFF
--- a/src/UI/Component/Player/Factory.php
+++ b/src/UI/Component/Player/Factory.php
@@ -56,13 +56,6 @@ interface Factory
      * rules:
      *   accessibility:
      *     1: >
-     *        The play/pause button MUST be accessible via tab key and allow to start/stop the video when the
-     *        space/return key is being pressed.
-     *     2: >
-     *        The playing position SHOULD be adjustable by using the cursor left/right keys.
-     *     3: >
-     *        The volume SHOULD be adjustable by using the cursor up/down keys.
-     *     4: >
      *        A subtitle file SHOULD be provided, if the video content contains speech.
      *   style:
      *     1: >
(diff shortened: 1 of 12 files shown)


alex40724:
@Amstutz @klees This is ready for review.

Amstutz (approved):
Hi @alex40724 

Thx a lot for working in the feedback from https://github.com/ILIAS-eLearning/ILIAS/pull/4041 . Also, thx for the examples and unit tests. 

@klees and @Amstutz

PR #4309, opened 2022-04-07
PHP 8 Code Review PersonalWorkspace

katringross:
Hi @alex40724 ,

I completed the review of the Services/PersonalWorkspace component.

Results:

- [x] The code style is PSR-2 + X compliant
- [x] No usage of $_GET / $_POST / $_REQUEST / $_SESSION
- [x] The unit tests pass
- [x] There are no serious PhpStorm Code Inspection issues left
--
We fixed some trivial issues reported by our inspection profile. We marked some issues with comments, please search for PHP8-Review.

Actions needed:
- [ ] Please have a look at our comments.

Best regards,
@katringross

### DIFF
--- a/Services/PersonalWorkspace/Administration/classes/class.ilObjPersonalWorkspaceSettingsGUI.php
+++ b/Services/PersonalWorkspace/Administration/classes/class.ilObjPersonalWorkspaceSettingsGUI.php
@@ -29,7 +29,7 @@ class ilObjPersonalWorkspaceSettingsGUI extends ilObjectGUI
     protected ilGlobalTemplateInterface $main_tpl;
 
     public function __construct(
-        $a_data,
+        $a_data, // PHP8-Review: Method ilObjPersonalWorkspaceSettingsGUI::__construct() has parameter $a_data with no type specified.
         int $a_id,
         bool $a_call_by_reference = true,
         bool $a_prepare_output = true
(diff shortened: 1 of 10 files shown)


mjansenDatabay:
@alex40724

- [ ] There are also some dynamically defined properties like `rbacsystem` in `ilObjPersonalWorkspaceSettingsGUI`. This is caused by the refactoring of `Services/Objects` and I noticed this kind of issue in many classes. The propery in the abstract base class `ilObjectGUI` is named `rbac_system` now.

alex40724:
@mjansenDatabay @daniwe4 If non private member variables in upper classes are renamed, the developer has to rename all the instances in the derived classes. Otherwise the variables must not be renamed. Or am I wrong here? I mean this would be a lot of open ends otherwise.

This is the commit https://github.com/ILIAS-eLearning/ILIAS/commit/6726d26e

When I currently search for ->rbacsystem in ilObj*GUI classes I get 99 occurences. For >rbac_system I get 112 (I did not check review and admin). The ...ListGUI classes kept $rbacsystem, the ObjGUI went to $rbac_system. A number of classes still using $rbacsystem have been "fixed" in the meantime, meaning they have both members vars now. I know that the Object service was messed up to a certain extent before, but... we should agree on one naming here imo.

mjansenDatabay:
@alex40724 / @daniwe4 

> If non private member variables in upper classes are renamed, the developer has to rename all the instances in the derived classes. 

Yes, this should be the case.

> When I currently search for ->rbacsystem in ilObj*GUI classes I get 99 occurences. For >rbac_system I get 112 (I did not check review and admin). The ...ListGUI classes kept $rbacsystem, the ObjGUI went to $rbac_system. A number of classes still using $rbacsystem have been "fixed" in the meantime, meaning they have both members vars now. I know that the Object service was messed up to a certain extent before, but... we should agree on one naming here imo.

Yes, this is a problem, and maybe the respective maintainers are not aware of the problem.

daniwe4:
Hi @alex40724 and @mjansenDatabay,
maybe I went a little overboard with the renaming. I could create an extra PR by bringing the naming together again. In one direction or the other.

mjansenDatabay:
@daniwe4 This would be great. Please be aware, some components already applied changes in regargs to your refactoring in `ilObject(GUI)` and removed 'local' `rbacsystem` member variables and instead use `rbac_system` (etc.) from the base class.

alex40724:
@daniwe4 Thanks!

alex40724:
@katringross Thanks for the PR. I will take care of the commented issues.

alex40724:
Done

daniwe4:
Hi @mjansenDatabay  and @alex40724 ,
i create a PR to hpefully fix the problems. https://github.com/ILIAS-eLearning/ILIAS/pull/4320

PR #4311, opened 2022-04-07
Fix duplicate entries in german lang file

kergomard:
We have two duplicate entries in our german lang file for 
common#:#cat_edit# and
pdfgen#:#overwrite_font#

See: https://mantis.ilias.de/view.php?id=32391

I removed them. By doing this, the lang file also got reordered by the corresponding script.


matthiaskunkel:
Thanks @swiniker for the PR. I took over your fixes and made some more. This is why I couldn't just take your PR. But thanks nevertheless.

PR #4320, opened 2022-04-08
rollback of rbac_.. to rbac.. name change

daniwe4:
When refactoring, I overdid it a bit by renaming rbac_.. variables. So I'm trying to revert to the old naming. Hopefully that fixes the mess.

### DIFF
--- a/Modules/Chatroom/classes/class.ilObjChatroom.php
+++ b/Modules/Chatroom/classes/class.ilObjChatroom.php
@@ -176,7 +176,7 @@ class ilObjChatroom extends ilObject
 
         $room->saveSettings($original_settings);
 
-        $rbac_log_roles = $this->rbac_review->getParentRoleIds($newObj->getRefId(), false);
+        $rbac_log_roles = $this->rbacreview->getParentRoleIds($newObj->getRefId(), false);
         $rbac_log = ilRbacLog::gatherFaPa($newObj->getRefId(), array_keys($rbac_log_roles), true);
         ilRbacLog::add(ilRbacLog::CREATE_OBJECT, $newObj->getRefId(), $rbac_log);
 
--- a/Modules/Chatroom/classes/class.ilObjChatroomGUI.php
+++ b/Modules/Chatroom/classes/class.ilObjChatroomGUI.php
@@ -301,7 +301,7 @@ class ilObjChatroomGUI extends ilChatroomObjectGUI implements ilCtrlSecurityInte
 
         // create permission is already checked in createObject.
         // This check here is done to prevent hacking attempts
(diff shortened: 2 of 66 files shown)


alex40724:
@daniwe4 Thanks for the PR. This is a difficult subject.

I am not sure about all the changes done related to Services/Object. My assumption is that the main commit that did the rbac member var renaming was this one: https://github.com/ILIAS-eLearning/ILIAS/commit/6726d26e

As far as I can see all these renamings have been done in classes that extend ilObjectGUI.

However this PR now contains different classes, that, if I am not wrong, have not been tackled before, like
Modules/Chatroom/classes/class.ilObjChatroom.php	
Modules/DataCollection/classes/Table/class.ilDclTableHelper.php 
Services/Dashboard/Access/class.DashboardAccess.php
... and lots of others not related to ilObjectGUI.

Is this intended?

klees:
I'll put this on next [PHP8-meetings agenda](https://docu.ilias.de/goto_docu_sess_11153.html).

@daniwe4 Please participate.

klees:
@chfsx @mjansenDatabay Any opinion regarding this PR? I would tend to merge it for consistencies sake, but would appreciate other opinions.

chfsx:
@klees Thanks for the proposal! I'm always fried with consistencies but think we should have a look at other ways o push that. With rector we have a tool with can do this automatically and it would be possible to add CI-checks to prevent from differing in the future. this could be spread to other widely used properties. If you like, I can give an insight on how to write such a rector and we would have a look at the possibilities of integrating this in the CI process together.

mjansenDatabay:
@klees As written in `Discord`, I am afraid that this PR makes things even worse. As @alex40724 outlined above, most maintainers have already fixed the existing issues in the meantime by renaming the properties.

IMO there are different things to discuss:

1. Inheritance hiearchies in general (not here, not now).
2. Naming of variables (e.g.`rbac_system` vs. `rbacSystem`): There is (AFAIK) no PSR specification addressing variable/property naming, yet. Maybe we should think about an own guideline to define if variables/properties should be `lowerCamelCase` or if variables/properties with `underscores` are the preferred style. But: Also to be discussed separately.
3. IMO there should be **no** reason why derived classes in this specific `ilObject(GUI)` hierarchy overwrite a property defined in the `ilObject(GUI)` base class and change the access modifier. This should be also valid for all, not only for the `rbac*` properties. There should be corresponding accessors which can be used by derived classes and the code which depends on them. There should be not situation where `ilObject(GUI)` declares a property `rbac_system` and an inherited class declares a property `rbacSystem` (or vice versa).

klees:
Close this for good reasons (thx @mjansenDatabay, @chfsx), but general issue needs to be discussed.

PR #4321, opened 2022-04-08
T&A: Fix for 31579: check mode before displaying a filter for tests/questi…

okaufman:
This fixes https://mantis.ilias.de/view.php?id=31579


mbecker-databay:
Great! Thank you so much!

PR #4323, opened 2022-04-08
Review `Services/Block`

kergomard:
Hi @alex40724 

Here are the results of the `Services/Block` review.

### Results

- [x] The code style is PSR-2 + X compliant:
- [x] No usage of $_GET / $_POST / $_REQUEST / $_SESSION: 
- [x] There is a Unit Test Suite with at least one unit test:
- [x] The unit tests pass

### Comments
- You still rely heavily on automatic type conversion and it is not always obvious, if things will go right (see Todos)
- Array parameters could be specified more closely, defining what types are expected in arrays. In just added one I was fairly sure about.

### Changes made in this PR:
- Fixed CS
- Added some types and one array declaration in Params
- Added a few //Todo Review PHP8:

Best regards,
@swiniker

### DIFF
--- a/Services/Block/classes/class.StandardGUIRequest.php
+++ b/Services/Block/classes/class.StandardGUIRequest.php
@@ -21,6 +21,11 @@ class StandardGUIRequest
 {
     use BaseGUIRequest;
 
+    /**
+     *
+     * @param string[] $passed_query_params
+     * @param string[] $passed_post_data
+     */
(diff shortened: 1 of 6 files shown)


alex40724:
Thanks for the PR! I will look at the open issues.

alex40724:
Done. I added strict types declarations, added several specific array types and fixed the comment issues.

PR #4328, opened 2022-04-09
Review bibliographfic list item

kergomard:
Hi @chfsx  

Here are the results of the `Services/Block` review.

### Results

- [ ] The code style is PSR-2 + X compliant: _CS-Fixer changed 28 Files_
- [ ] No usage of $_GET / $_POST / $_REQUEST / $_SESSION: _You have one direct access to $_REQUEST (marked by a Todo) and one additional access to $_FILE (ilObjBibliographficGUI line 591) that could maybe also be avoided_
- [ ] There is a Unit Test Suite with at least one unit test: _There are no Unit Tests_
- [ ] The unit tests pass
- [x] External libraries are compatible with PHP 8 (if relevant): _The RISReader hasn't been updated since 2012, it seems to work with PHP8, but I haven't tested it (just ran the Rector-Profile to see if it proposed changes)_

### Comments

- Saving Settings fails when not uploading a tile image. The error is in ilObjBibliographic on line 152 as $has_valid_upload is true because the previous $upload->hasUploads() returns true. Didn't look deep enough into it to be sure if the problem is here or in ILIAS\FileUpload.
- Editing or Deleting a Filter fails with "Typed property ilBiblFieldFilter::$field_id must not be accessed before initialization"

### Changes made in this PR:
- Fixed CS
- Added some types and one array declaration in Params
- Fixed a few obvious problems
- Added a few //Todo PHP8 Review:

Best regards,
@swiniker

### DIFF
--- a/Modules/Bibliographic/classes/Admin/Facade/class.ilBiblAdminFactoryFacade.php
+++ b/Modules/Bibliographic/classes/Admin/Facade/class.ilBiblAdminFactoryFacade.php
@@ -20,7 +20,7 @@ class ilBiblAdminFactoryFacade implements ilBiblAdminFactoryFacadeInterface
      *
      * @param \ilObjBibliographicAdmin $ilObjBibliographicAdmin
      */
-    public function __construct(ilObjBibliographicAdmin $ilObjBibliographicAdmin, $type_id)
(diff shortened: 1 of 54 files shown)


chfsx:
Thanks a lot, I merged and processed your review in https://github.com/ILIAS-eLearning/ILIAS/commit/5816669c759a3d257d57e040fbeca476eb14b6a0

PR #4330, opened 2022-04-11
Review `Services/VirusScanner`

kergomard:
Hi @rschenk0 

Here are the results of the `Services/VirusScanner` review.

### Results

- [ ] The code style is PSR-2 + X compliant:_CS-Fixer changed two files_
- [x] No usage of $_GET / $_POST / $_REQUEST / $_SESSION: 
- [ ] There is a Unit Test Suite with at least one unit test:
- [ ] The unit tests pass

### Comments
- ilVirusScannerConfigStoredObjective line 52: It might be a good idea as $config is nullable to add a check for it and throw an explicit exception instead of potentially just letting it crash and burn.
- The component would be a good place to introduce declare strict_types. I see only two failing functions that are easily fixable: ilVirusScannerscanFileFromBuffer() and scanFileFromBuffer()

### Changes made in this PR:
- Fixed CS
- Scoped Constants
- Some additionnal fixes

Best regards,
@swiniker

### DIFF
--- a/Services/VirusScanner/classes/Setup/class.ilVirusScannerConfigStoredObjective.php
+++ b/Services/VirusScanner/classes/Setup/class.ilVirusScannerConfigStoredObjective.php
@@ -1,23 +1,26 @@
-<?php
+<?php declare(strict_types=1);
 
-use ILIAS\Setup;
-
-/******************************************************************************
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
  *
- * This file is part of ILIAS, a powerful learning management system.
- *
- * ILIAS is licensed with the GPL-3.0, you should have received a copy
- * of said license along with the source code.
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
  *
  * If this is not the case or you just want to try ILIAS, you'll find
  * us at:
- *      https://www.ilias.de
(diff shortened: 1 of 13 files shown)


mjansenDatabay:
@swiniker / @gvollbach I added a new test suite for this component here: https://github.com/ILIAS-eLearning/ILIAS/pull/4339

mjansenDatabay:
@swiniker I pushed some changes (https://github.com/ILIAS-eLearning/ILIAS/pull/4330/commits/3e7628a4a4c05b150189b41d75d3f746e3d457d7) to your review branch. IMO most issues are fixed now. I checked other setup agents and objectives (you mentioned `ilVirusScannerConfigStoredObjective`), but I did not find any agent who handles `null`.

@swiniker / @gvollbach / @rschenk0 : I also introduced `strict_types` for all related PHP classes, added the correct license header and improved most trivial PHP language issues my `PhpStorm` IDE suggested to change.

IMO this is ready to merge now (and don't forget to merge the `Test PR` #4339).

kergomard:
Thank you very much @mjansenDatabay .
Concerning `ilVirusScannerConfigStoredObjective` in `ilVirusScannerSetupAgent` (big sorry for my wrong class reference above!): I understand the approach. I pointed it out because I personally get a  little nervous if it is explicitly stated that a parameter might be null, but I then go ahead and ignore that case in the code dealing with it.

On a side note: As I don't use `PhpStorm` these feedbacks on the reviews also help me to see better into the differences between the `PhpStorm ` inspection profile and phpstan.

mjansenDatabay:
According to @rschenk0 I should merge this.

PR #4333, opened 2022-04-11
Refinery: 32418, KindlyToFloat should accept values smaller than 1

nhaagen:
fixes 
https://mantis.ilias.de/view.php?id=32418 and 
https://mantis.ilias.de/view.php?id=31132


klees:
Picked to trunk.

PR #4335, opened 2022-04-11
TA: 13917, option 'passDeletion' dependent on 'resultsAccessEnabled'

nhaagen:
https://mantis.ilias.de/view.php?id=13917

Screenshot from 2022-04-11 14-30-42


mbecker-databay:
Looks good to me! Thank you very much!

PR #4339, opened 2022-04-11
VirusScanner: Add unit test suite

mjansenDatabay:
This PR adds a unit test suite for the `Services/VirusScanner` component. The tests must run in separate processes because the code under test heavily relies on global constants, which cannot be re-defined within a single process.

### DIFF
--- a/Services/VirusScanner/classes/class.ilVirusScanner.php
+++ b/Services/VirusScanner/classes/class.ilVirusScanner.php
@@ -49,7 +49,7 @@ class ilVirusScanner
         global $DIC;
         $ilias = $DIC['ilias'];
         $lng = $DIC->language();
-        $log = $DIC->logger();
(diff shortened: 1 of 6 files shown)


Services/VirusScanner/classes/class.ilVirusScanner.php
```diff
@@ -49,7 +49,7 @@ public function __construct(string $scan_command, string $clean_command)
         global $DIC;
         $ilias = $DIC['ilias'];
         $lng = $DIC->language();
-        $log = $DIC->logger();
+        $log = $DIC->logger()->root();
```
mjansenDatabay, line 52: @rschenk0 / @gvollbach IMO the logging could not work at all without this fix here.

mjansenDatabay:
According to @rschenk0 I should merge this.

PR #4348, opened 2022-04-12
Trunk font sizes

ezenzen:
Hi @all,

according to bug report 0031813 we want to change the font sizes from px to rem. This has some page effects on existing components. In consultation with the CSS maintainer @amstutz we will test the adjustments in this draft and then resolve the conflicts.

0031813: User preferences are not taken into account - https://mantis.ilias.de/view.php?id=31813

Many Greetings, 
Enrico


Amstutz (changes requested):
Hi @ezenzen 

Thx bunch for this huge work. Lots of gold in here. IMO, this is close to be mergeable from my perspective. There will be some side effects, sure. However, this is a necessary transition (see related mantis issue) and will only be pushed to the trunk. Therefore, I am willing to take some risk here.

Special thanks for:

- New il-fonts variables and making use of them, and so removing lots of fixed and somewhat magic values in our code!
- Introducing and overriding code less
- We are again somewhat closer to losing bootstrap.

There are some things left to-do before merging:

- [ ] Code.less Documentation: Describe shortly, what this file is and why we are overriding this, maybe by linking back to this PR. Also adapt the name of the file to bootstrap-code.lesss, so one can see where this comes from.
- [ ] Remove magic numbers: You introduced some values as somewhat magic fixed values. Introduce a variable for them, especially if you use them more than once. In some cases, you can also introduce a variable only on a file level (e.g. for slate.less). 
- [ ] Remove the conflicts in the file.

We will discuss this today in the CSS-Squad as well. My proposal is, to merge this soon. After merging, we will test this and work on resulting issues. 

Thx @Amstutz

ezenzen:
Hi @Amstutz,

thank you for the feedback.
- I have added a short documentation to the code.less. 
- For magic numbers that are used more than once, I introduced variables.  

Now, as discussed in the CSS Squad, you can resolve the conflicts and then test/merch the PR.

Thx

Amstutz:
Thx a lot to everyboedy involved!

PR #4349, opened 2022-04-12
Fix MimeType recognition for Office 2003 and newer

kergomard:
This fixes: https://mantis.ilias.de/view.php?id=32428

There would be an alternative solution if we move all mime-type detection ton a extension-based approach. Personally I prefer the mixed approach proposed here, but I could also provide a PR for the alternative approach.

If accepted this should be cherry-picked to ILIAS 7


chfsx:
Thanks a lot @swiniker ! I'll cherry pick to release_7 as well and will close the mantis-ticket

PR #4358, opened 2022-04-12
Review of Service/DataSet

pascalseeland:
Hello @alex40724,

I completed the first review of Service/DataSet.

Results:
- [x] The code style is PSR-2 + X compliant
- [x] No usage of $_GET / $_POST / $_REQUEST / $_SESSION: I found 18 usages of $_POST and 2 usages of $_GET
- [x] The unit tests pass
- [x] There are no serious PhpStorm Code Inspection issues left

I did fix switch issue in this pr. Otherwise code looks good. I did add some inline comments you can find with TODO-PHP8-REVIEW. Some if this is more like I feel this would help to improve overall code than a current issue.

Best regards,
@pascalseeland

### DIFF
--- a/Services/DataSet/classes/class.ilDataSet.php
+++ b/Services/DataSet/classes/class.ilDataSet.php
@@ -132,7 +132,7 @@ abstract class ilDataSet
 
     public function getDSPrefixString() : string
     {
-        if ($this->getDSPrefix() != "") {
+        if ($this->getDSPrefix() !== "") {
             return $this->getDSPrefix() . ":";
         }
         return "";
@@ -176,7 +176,7 @@ abstract class ilDataSet
     public function convertToLeadingUpper(string $a_str) : string
     {
         $a_str = strtoupper(substr($a_str, 0, 1)) . substr($a_str, 1);
-        while (is_int($pos = strpos($a_str, "_"))) {
+        while (($pos = strpos($a_str, "_")) !== false) {
             $a_str = substr($a_str, 0, $pos) .
                 strtoupper(substr($a_str, $pos + 1, 1)) .
                 substr($a_str, $pos + 2);
@@ -187,6 +187,7 @@ abstract class ilDataSet
             
     /**
      * Get json representation
+     * //TODO-PHP8-REVIEW It would be helpful an example like below for the xml representation would be added here.
      */
     final public function getJsonRepresentation() : string
     {
@@ -204,7 +205,7 @@ abstract class ilDataSet
             $arr["set"][] = $this->getJsonRecord($d);
         }
         
-        return ilJsonUtil::encode($arr);
+        return json_encode($arr, JSON_THROW_ON_ERROR);
     }
 
     /**
@@ -239,7 +240,7 @@ abstract class ilDataSet
         $this->dircnt = 1;
         
         // step 1: check target release and supported versions
-        
+        //TODO-PHP8-REVIEW either fix numbering or add the checks at this point.
         
         
         // step 2: init writer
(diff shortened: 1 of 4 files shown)


alex40724:
Thanks for the review PR. I will check the ToDos...

Services/DataSet/classes/class.ilDataSet.php
```diff
@@ -521,25 +518,26 @@ protected function parseObjectExportId(string $a_id, ?string $a_fallback_id = nu
...
             }
-        }
-        
-        // external id
-        $id = ilObject::_getIdForImportId($a_id);
-        if ($id) {
-            return array("type" => self::EXPORT_ID, "id" => $id);
         } else {
-            return array("type" => self::EXPORT_ID_INVALID, "id" => $a_fallback_id);
+            // external id
```
alex40724, line 530: @pascalseeland @mjansenDatabay  Just a note: It is not necessary to nest this last block into the else part, since all others above will ensure returns. The [EA] inspections even prefer to get rid of all the else branches after ifs which include a return. It says "[EA] Child instructions can be extracted here (clearer intention, lower complexity numbers" for these cases. Actually I am not convinced that this really improves readability of the code. I don't have (not yet) a strong preference either way...

mjansenDatabay: @alex40724 

IMO the `Redundant 'else' keyword` warning is **valid**. Every control structure/keyword and nesting which could be simplified should be simplified (see: https://williamdurand.fr/2013/06/03/object-calisthenics/ / https://medium.com/@tadeubdev/object-calisthenics-with-php-f756baf6693)

Examples (I suggest ot refactor all of them):

```php
    private function convertInput($a_value) : string
    {
        if (is_array($a_value)) {
            return $a_value[0];
        } else {
            return $a_value;
        }
    }
```

```php
    public function trigger($params)
    {
        if ($this->detection_state == false) {
            $this->setDetectorState(true);
            return true;
        } else {
            return false;
        }
    }
```

```php
    private function getPearAtributeArray() : array
    {
        if ($this->enabledSyncOnLogin()) {
            $mapping = ilLDAPAttributeMapping::_getInstanceByServerId($this->getServerId());
            return array_merge(
                array($this->getUserAttribute()),
                $mapping->getFields(),
                array('dn'),
                ilLDAPRoleAssignmentRules::getAttributeNames($this->getServerId())
            );
        } else {
            return array($this->getUserAttribute());
        }
    }
```

I would also strongly recommed to refactor this kind of code ...

```php
    // ... a lot of code before this
 
   if ($condition === true) {
       $translation = $this->lng->txt('my_lang_id_a');
   }  else {
       $translation = $this->lng->txt('my_lang_id_b');
   }

    // ... a lot of code after this
```

... to:

```php
    // ... a lot of code before this
 
   $translation = $this->lng->txt('my_lang_id_b');
   if ($condition === true) {
       $translation = $this->lng->txt('my_lang_id_a');
   }

    // ... a lot of code after this
```

And yes, there might be exceptions: If the default statement is `expensive` (in regards to performance), it should not be refactored it this way.

------

But I **don't** suggest to convert `elseif` to `if` branches (maybe only in rare cases),

```php
    public function attemptTransition() : bool
    {
        $eval_function = function ($detectors) {
            return eval($this->evaluation_expression);
        };

        if ($eval_function($this->detectors) === null) {
            return false;
        }
        
        if ($eval_function($this->detectors) === true) {
            $this->executeTransition();
            return true;
        } elseif (method_exists($this, 'executeElseTransition')) {
            $this->executeElseTransition();
            return true;
        }
        return false;
    }
```

PR #4359, opened 2022-04-13
Fix 32387

kergomard:
This removes the deprecated setting for scoring MC questions from the test (see: https://mantis.ilias.de/view.php?id=32387). I also remove the corresponding table column and used the same pattern for naming the SetupObjective ans in the Pool. I'm not sure this is the right pattern for splitting them up though.


mbecker-databay:
I am good with the change, but we will speak about the objective in tomorrows TechSquad session.

kergomard:
I did the renaming of the update class.

kergomard:
@mbecker-databay : I resolved the merge conflicts in the lang files. We should not let this linger too long otherwise we are going to run into issues again.

PR #4360, opened 2022-04-13
PHP 8: Code Review `QTI`

tfamula:
Hello @lscharmer , 

I completed the review of the `Services/QTI` component.

Results:

- [x] The unit tests pass

I also did some further changes:

- Added the correct license header
- introduced `strict_types` for PHP classes
- moved constants to the related classes and adjusted all occurrences (also in other components)
- moved initialization of variables from constructor to the place they are declared (where possible). IMO, this way is cleaner than the way with the constructor because it reduces complexity / redundancy. And AFAIK, it also prevents the `Typed property must not be accessed before initialization` error
- some more minor fixes / improvements

--

Actions needed:
- [x] Please check my inline comments. You can use `TODO PHP8-REVIEW` when searching for my comments.
- [x] Please check nullable variables: Many variables in this component are nullable. This is no good approach IMO and should only be used rarely, especially for variables with types like `int` or `string`. I would use defaults of the proper type instead, like `$id=0` or `$text=''`. But maybe the nullable variables are due to dependencies in other components, so maybe it would be too risky to make changes here and also too effortful. So see this point more as information.

Greetings
@tfamula

### DIFF
--- a/Modules/LearningModule/classes/class.ilLearningModuleImporter.php
+++ b/Modules/LearningModule/classes/class.ilLearningModuleImporter.php
@@ -92,7 +92,7 @@ class ilLearningModuleImporter extends ilXmlImporter
         if (is_file($qti_file)) {
             $qtiParser = new ilQTIParser(
                 $qti_file,
-                IL_MO_VERIFY_QTI,
+                ilQTIParser::IL_MO_VERIFY_QTI,
                 0,
                 ""
             );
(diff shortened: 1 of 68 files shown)


lscharmer:
Hi @tfamula,
thank you very much for your review :) I will create a PR based upon this.

alex40724 (approved):
Changes done in Modules/LearningModule like switching to ilQTIParser::IL_MO_VERIFY_QTI are fine for me. You may just merge these kind of things into my components.

lscharmer:
PR is created: https://github.com/ILIAS-eLearning/ILIAS/pull/4433

PR #4362, opened 2022-04-13
#32437: Fix submission of unchecked switchable group

thibsy:
Hi @Amstutz and @klees 

I've opened a ticket (https://mantis.ilias.de/view.php?id=32437) but figured it's an easy fix, therefore I opened this PR =).

I assume this change solves two problems at once, because from how I read it, if the switchable group was disabled there would've been an undefined-offset and call on null error on at `if ($clone->inputs[$key]->getContent()->isError()) { ...`.

I adjusted the tests because I changed `InputData::get()` to `InputData::getOr()`.

Cheers!

### DIFF
--- a/Services/Dashboard/classes/class.ilDashboardGUI.php
+++ b/Services/Dashboard/classes/class.ilDashboardGUI.php
@@ -135,35 +135,6 @@ class ilDashboardGUI
     {
         global $DIC;
 
-        $form = $DIC->ui()->factory()->input()->container()->form()->standard(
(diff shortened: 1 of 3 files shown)


klees:
Thanks a lot!

klees (approved):

klees:
merged as https://github.com/ILIAS-eLearning/ILIAS/commit/ae1304c7bd33843677009ae8ed11707a69ed83cf

Slightly different solution already exists in release_7.

PR #4367, opened 2022-04-14
Setup AccessControl: fix values for array_filter

daniwe4:


smeyer-ilias:
Thanks for fixing this.

PR #4368, opened 2022-04-14
MedaData/LOM: Fix editing of metadata

mjansenDatabay:
This PR fixes the editing of LOM metadata, which does not work since `ilTemplate::addBlockFile` did not work as expected anymore with the release of ILIAS 6.x.

This MUST be also cherry-picked to `release_6` and of course `trunk` (which will  most propably result in conflicts).

Mantis Issue: https://mantis.ilias.de/view.php?id=32447


smeyer-ilias:
Thanks for fixing

PR #4373, opened 2022-04-17
Removal of Services/JSON

pascalseeland:
Hi @mjansenDatabay,

instead of usual review, I would propose removing the last leftovers from pre PHP5.2 support.

Kind Regards,
Pascal


mjansenDatabay:
Thanks @pascalseeland , good idea.

PR #4380, opened 2022-04-19
Review of Modules/Portfolio

pascalseeland:
Hi @alex40724 ,

I completed the review of the Modules/Portfolio component.

Results:
- [x] The code is PSR-2 + X compliant. I did reformat some classes.
- [X] No usage of $_GET / $_POST / $_REQUEST / $_SESSION outside GUI-classes
- [X] There is a Unit Test Suite with at least one unit test
- [X] The unit tests pass
- [X] There are no serious PhpStorm Code Inspection issues left.
- [X] The types are fully documented (PhpDoc) or explicit PHP types are used (type hints, return types, typed properties).

Actions:
- [ ] Inline comments where prefixed with TODO-PHP8-REVIEW. Please take a look at them.
- [ ] The EA-Plugin reports several warnings. Please check them if time permits. I only resolved those where I was confident not to break other things. They mark several things that would break of strict typing would be enforced.

I fixed the issues created by renaming variables is the Service/Object component.

Kind regards,
Pascal

### DIFF
--- a/Modules/Portfolio/Access/class.ilPortfolioAccessHandler.php
+++ b/Modules/Portfolio/Access/class.ilPortfolioAccessHandler.php
@@ -76,7 +76,7 @@ class ilPortfolioAccessHandler implements ilWACCheckingClass
         $ilSetting = $this->settings;
 
         // #20310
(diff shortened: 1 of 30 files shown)


alex40724:
Great. Thanks! I will check the comments.

alex40724:
Done.

PR #4386, opened 2022-04-20
Review of Service/Math

pascalseeland:
Hi @mjansenDatabay, hi @gvollbach,

I completed the review of the Service/Math component.

Results:
- [x] The code is PSR-2 + X compliant.
- [X] No usage of $_GET / $_POST / $_REQUEST / $_SESSION outside GUI-classes
- [X] There is a Unit Test Suite with at least one unit test
- [X] The unit tests pass
- [X] There are no serious PhpStorm Code Inspection issues left.
- [ ] The types are fully documented (PhpDoc) or explicit PHP types are used (type hints, return types, typed properties).

Actions:
- [ ] Inline comments where prefixed with TODO-PHP8-REVIEW. Please take a look at them.
- [ ] The EA-Plugin reports several warnings. Please check them if time permits. I only resolved those where I was confident not to break other things. 

Kind regards,
Pascal


mjansenDatabay:
Thx for the review @pascalseeland 

@pascalseeland / @gvollbach 
- I suggest to **not** change class `EvalMath` if it is compatible with PHP 8, because this is a 3rd party library we should not change if not necessary.

pascalseeland:
@mjansenDatabay Thanks for the hint. I was not aware of this. I will update the PR. But we should in any case add a hint somewhere like in a README.md that this is a 3rd party file. With the header it looks like a normal ILIAS file.

PR #4391, opened 2022-04-21
PHP 8: Strict Code Review `Category and Repository`

mjansenDatabay:
Hi @alex40724,

this PR fixes a few issues which occur if `declare(strict_types=1)` would be set in the respective classes.

I left the (mostly temporarily used) `declare(strict_types=1)` directive in classes where I am very sure no issues occur with strict type handling.

In the folllowing folders I only added a **few** `declare(strict_types=1)` because the risk was to high to break something:

- Services/Container
- Services/Reposistory

Nevertheless, I fixed potential type issues.

Best regards,
@mjansenDatabay


alex40724:
Thanks! Didn't seem to have broken major things.

PR #4396, opened 2022-04-21
PHP 8: Code Review `BookingManager`

iszmais:
Hi, @alex40724  

This is an complete review of the `Module/BookingManager` component.

Results:
- [x] No usage of `$_GET` / `$_POST` / `$_REQUEST` / `$_SESSION` _(I encoutered one uses of $GLOBAL wich was fixed)_
- [x] The unit tests pass

I also fixed some trivial inspection profile issues.

### Further actions needed:
- All unsolved Problems are marked with `//PHP8Review: ` Preffix.
- You may add `declare(strict_types=1);` to all php files (or at least most with a few exceptions). Im aware that this might not work due to side effects in every file, but i assume most a allready strict and therfore compatible.

Best regards,
@iszmais

### DIFF
--- a/Modules/BookingManager/Reservations/class.ilBookingReservationsGUI.php
+++ b/Modules/BookingManager/Reservations/class.ilBookingReservationsGUI.php
@@ -1,17 +1,21 @@
 <?php
 
-/**
+/******************************************************************************
+ *
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
+ *
  * ILIAS is licensed with the GPL-3.0,
  * see https://www.gnu.org/licenses/gpl-3.0.en.html
  * You should have received a copy of said license along with the
  * source code, too.
+ *
  * If this is not the case or you just want to try ILIAS, you'll find
  * us at:
- * https://www.ilias.de
- * https://github.com/ILIAS-eLearning
- */
+ *     https://www.ilias.de
+ *     https://github.com/ILIAS-eLearning
+ *
+ *****************************************************************************/
 
 /**
  * Reservations screen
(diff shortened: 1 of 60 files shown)


iszmais:
Hi @alex40724 ,

I finished the Review. Feel free to cantact me if you have question or feedback.
Since there where less duplicatete documentation tags, wich just reference to the typehint, than in the `Administration` i took the time to correct them directly inside the PR instead of writing a recommandation. Feel free to revert the few occurences if you see any inconvinience.

Greetings,
@iszmais

alex40724:
Thanks. I will take care of the comments.

Modules/BookingManager/Reservations/class.ilBookingReservationsGUI.php
```diff
@@ -506,7 +504,7 @@ public function rsvCancel() : void
                         $ids[] = $aggr_ids;
                     } else {
                         $aggr_ids = array_slice($aggr_ids, 0, $max);
-                        $ids = array_merge($ids, $aggr_ids);
+                        $ids = array_merge($ids, $aggr_ids);//PHP8Review: This may cause performance issues inside of a foreach
```
alex40724, line 507: @iszmais I don't think so. Only if I would change $rsv_aggr...

alex40724:
Done.

Modules/BookingManager/Reservations/class.ilBookingReservationsGUI.php
iszmais: @alex40724 Im not fully aware what you mean with that.

Im not concerned about the looping index of the foreach itself but about the dynmacly called array_merge, since array_merge has a very bad perfromance in general, sepacially for the purpose addition because i might reindex the whole $root array (not just the added part) in every call depending if its key a numeric.

Of course this heavily depends on the array composition and size, but since i couldnt find a proof of associative keys or its size i worte this hint in concern. 
Maybe you are more aware of the circumstances and could provide me with that info.

In both ways it would be an improved (or even) solution to stack the array and marge them after the loop, since you dont need them in it (since this ensure only one reindexing):
```
$stack = [];
for($map as $a) {
   //do something with $a
  $stack[] = $a;
}

array_merge(...$stack);
```

PR #4398, opened 2022-04-21
KS: Item with Lead Avatar

alex40724:
This is needed to switch the comments overview to the KS, see FW:

https://docu.ilias.de/goto_docu_wiki_wpage_6964_1357.html

Rule Change: https://github.com/ILIAS-eLearning/ILIAS/pull/4398/files#diff-b84dddcb4040f0309648a00e52da9ede0b8c05dd33ac7aa7ba4cc02e9ab607a5

In this implementation the visual behaviour is identical to the Lead Icons (fixed size), not to the Lead Images (gaining size with more space). I thought this would be more appropriate, the massive sized Avatars in the FW entry look strange imo. Also this way there is less code change needed, no changes done in Avatar implementation or in the complex Item template. The PR includes a full implementation, incl. unit tests and examples.

Bildschirmfoto 2022-04-21 um 17 38 55

### DIFF
--- a/src/UI/Implementation/Component/Item/Renderer.php
+++ b/src/UI/Implementation/Component/Item/Renderer.php
@@ -14,6 +14,7 @@ use ILIAS\UI\Implementation\Render\Template;
 use ILIAS\UI\Component\Button;
 use ILIAS\UI\Component\Link\Link;
 use ILIAS\UI\Implementation\Render\ResourceRegistry;
+use ILIAS\UI\Component\Symbol\Avatar\Avatar;
 
 class Renderer extends AbstractComponentRenderer
 {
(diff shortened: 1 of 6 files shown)


Amstutz:
Hi @alex40724 

Thx a lot for this contribution. I checked it out, and it looks excellent. Nice addition thx.

There is a minor change on the public interface. However, it is non-breaking and just a detail. Since coding complete is approaching quickly, I did fast-track this and merged it already. I still will announce the change in the next JF.

Nice work, I could not find anything to ask to add/change. Special thx for the examples and tests (including the rendering). Always good to see, that no less was required for this. 

Thx

src/UI/Implementation/Component/Item/Renderer.php
```diff
@@ -108,7 +109,7 @@ protected function renderStandard(Item $component, RendererInterface $default_re
                 $tpl->setVariable("LEAD_IMAGE", $default_renderer->render($lead));
                 $tpl->parseCurrentBlock();
             }
-            if ($lead instanceof Icon) {
+            if ($lead instanceof Icon || $lead instanceof Avatar) {
                 $tpl->setCurrentBlock("lead_icon");
                 $tpl->setVariable("LEAD_ICON", $default_renderer->render($lead));
```
tfamula, line 114: lead_icon should be renamed to lead_icon_or_avatar or something more general here

PR #4399, opened 2022-04-22
Layout/Page: Check for new introduced block (Mantis 32455)

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=32455

I don't necessarily suggest to merge this PR, because the changed code will "silence" the actual issue of missing metadata in the HTML document of ILIAS.


Amstutz:
@mjansenDatabay thx a lot. @mjansenDatabay and @Amstutz hat a short exchange on this. We agree that we will not merge this, since the actual issue (missing metadata in the metadata section of the HTML document), which has been fixed with the latest changes, will still persist and nobody will be aware of it. See https://mantis.ilias.de/view.php?id=32455 for details.

PR #4400, opened 2022-04-22
Database: Make offset optional in interface

mjansenDatabay:
This PR changes makes the `$offset` of `ilDBInterface::setLimit` optional, similar to the implementation.


chfsx:
Thanks @mjansenDatabay !

PR #4401, opened 2022-04-22
Global Screen, Empty Mainbar, fix for #31844

Amstutz:
See: https://mantis.ilias.de/view.php?id=31844 for some more context.

Note, that the current hasItems items check seems not sufficient, since there will be filters applied later, to check if those items are really rendered. So we only know AFTER the logic of the function, how many items there really are.

Note II: if the help is active, the mainbar will always be rendered, since there is always one item in there (maybe not visible, but it is in there).


chfsx:
I fixed the issue in https://github.com/ILIAS-eLearning/ILIAS/commit/9e3a8561319847489b5235bbbeaf5f68b28d3a50, thanks for the proposal!

PR #4402, opened 2022-04-22
PHP 8: Code Review `Conditions`

iszmais:
Hi, @smeyer-ilias 

I completed the review of the `Services/Conditions` component.

Results:
- [x] The unit tests pass

I also fixed some weaker inspection profile issues.

### Further actions needed:
- All unsolved Problems are marked with `//PHP8Review: ` Preffix.
- I encountered some cases where you use optional flags (?bool), i marked one as an example. This is very suspicious and i assume it can mostly be changed into just bools with default false. But since im not aware of the full functionality behind it i will just leave a recommandation here.
- Optional: You might take a look and the `ilConditionHandler` since it seems to have a lot of functionality wich would be better located inside a own repository (which would also improve the size of this class).

Best regards,
@iszmais

### DIFF
--- a/Services/Conditions/classes/class.ilCondition.php
+++ b/Services/Conditions/classes/class.ilCondition.php
@@ -1,6 +1,21 @@
 <?php declare(strict_types=1);
 
-/* Copyright (c) 1998-2018 ILIAS open source, Extended GPL, see docs/LICENSE */
+/******************************************************************************
+ *
+ * This file is part of ILIAS, a powerful learning management system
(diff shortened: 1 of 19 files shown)


smeyer-ilias:
Thanks @iszmais for reviewing and fixing issues.
I will check the remaining issues and fix them in the trunk.

mjansenDatabay:
@iszmais / @smeyer-ilias And: Even the new license header is wrong here

PR #4404, opened 2022-04-25
Migration: PHP8 improvements

daniwe4:


Services/Migration/test/ilServicesMigrationSuite.php
```diff
@@ -0,0 +1,27 @@
...
+
+class ilServicesMigrationSuite extends TestSuite
+{
+    public static function suite()
+    {
+        $suite = new ilServicesMigrationSuite();
+
+        return $suite;
+    }
+}
```
mjansenDatabay: `PSR-2 + X` ;-)

daniwe4: Did you mean the return type? In this case i fix it. ;)

mjansenDatabay: I meant the missing `\n` at the end of the file.

daniwe4: Ok, thank you. I added a new line, but github doesn't show this correctly.

mjansenDatabay: Looks good now ^^

Services/Migration/DBUpdate_3136/classes/class.ilDBUpdate3136.php
```diff
@@ -1,5 +1,18 @@
-<?php
-/* Copyright (c) 1998-2010 ILIAS open source, Extended GPL, see docs/LICENSE */
+<?php declare(strict_types=1);
+
+/******************************************************************************
```
mjansenDatabay: According to our `Rector` in 'CI/Rector/ChangeLicenseHeader.php' the license header has the wrong format:

```php
/**
 * published by ILIAS open source e-Learning e.V.
 *
 * ILIAS is licensed with the GPL-3.0,
 * see https://www.gnu.org/licenses/gpl-3.0.en.html
 * source code, too.
 *
 * us at:
 * https://www.ilias.de
 *
 *********************************************************************/
```

Services/Migration/test/ilServicesMigrationSuite.php
```diff
@@ -0,0 +1,27 @@
+<?php declare(strict_types=1);
+
+/******************************************************************************
```
mjansenDatabay: See my comment above, the licese header differs from the definition in our `Rector`.

daniwe4: Thank you for the hint. I didn't know where to pick it. Hopefully now it is correct.

Services/Migration/DBUpdate_3136/classes/class.ilDBUpdate3136.php
daniwe4: Thank you for the hint. I didn't know where to pick it. Hopefully now it is correct.

PR #4405, opened 2022-04-25
BackgroundTasks: Migrate to `__serialize`/`__unserialize` (PHP 8.1)

mjansenDatabay:
This PR migrates the (in PHP 8.1) deprecated `Serializable` interface by **maintaining the public API**.

Changes made:
- The dependeny on the PHP core interface `Serializable` has been removed.
- A new `BackgroundTasks` related interface `Serializable` has been introduced **providing the known API** to `serialize()`/`unserialize()` a value.
- `serialize()`/`unserialize()` will internally call the new/recommended PHP methods `__serialize()`/`__unserialize()`.

See: https://wiki.php.net/rfc/phase_out_serializable


mjansenDatabay:
Closed because after some further tests it seems to be impossible to guarantee the same input/output with the `__` methods.

PR #4406, opened 2022-04-25
Migration: added method addRbacTemplate back in place

daniwe4:
Hi,

i hope this will fix this two mantis tickets:

https://mantis.ilias.de/view.php?id=32461
https://mantis.ilias.de/view.php?id=32450


klees:
This should be obsolete, since the db-template in trunk is update now.

PR #4407, opened 2022-04-25
Remove old usages of notifications from external consumers

iszmais:
Since the major changes inside of then notifications for trunk (see https://docu.ilias.de/goto_docu_wiki_wpage_6494_1357.html) this occurences dont have any functionality within the current state.

This is a proposal to remove this occurences.

### DIFF
--- a/Services/COPage/Editor/js/src/components/paragraph/ui/paragraph-ui.js
+++ b/Services/COPage/Editor/js/src/components/paragraph/ui/paragraph-ui.js
@@ -559,26 +559,6 @@ export default class ParagraphUI {
       $('#il_EditPage').replaceWith(o.responseText);
       this.reInitUI();
       il.IntLink.refresh();
-      if (o.argument.osd_text && o.argument.osd_text != "") {
-        OSDNotifier = OSDNotifications({
-          initialNotifications: [{
-            notification_osd_id: 123,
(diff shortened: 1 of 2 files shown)


iszmais:
Hi @alex40724 ,

I stumbled upon two occurences/usages of the JS code located in `Services/Notifications/templates/default/notifications.js`. Consuming this code only worked 'by accident' and was neither documented as public API nor was it the intention to directly use the JavaScript functions. That being said the creation of those notifications will/should probably not work in `trunk` with the new integrated notifcation due to major conceptional changes (the visiuals of the notifications are now created server side and fetched within a XHR request, instead of beeing created by the JavaScript itsself).
It seems to me those don't fullfill any purpose and are mostly used as a wrokaround. But I'm not fully aware of the scope of this occurences, so this is just a wild guess.

I would appreciate if you could maybe provide me with more information about the usages and would offer to open or change this PR into a refactoring, if we agree on a purpose of those occurences.
Otherwise we may just could delete them with this PR.

Feel free to give me a response to that or lets have a quick call if this is getting more complex than estimated.

Greeting,
@iszmais

alex40724:
@iszmais Thanks for the PR. Indeed this code is not in use anymore, I just removed it.

PR #4410, opened 2022-04-25
PHP8: Code Review `Competence Management`

lscharmer:
Hi @tfamula,

here are the results of the `Services/Skill` review.

### Results
- [x] The unit tests pass
- [ ] There are no serious `PHPStorm Code Inspection` issues left
- [ ] The types are fully documented (PHPDoc) or explict PHP types are used (type hints, return types, typed properties)

### Changes made in this PR:
- One change is made in `Modules/Test/classes/class.ilTestPersonalSkillsGUI.php` (add cast to int), that was the only place I could not ensure that an actual integer is passed to `ilPersonalSkillsGUI::getGapAnalysisHTML`
- Add return types and parameter types for anonymous functions
- Inline single usages of anonymous functions
- Add return types and parameter types for methods
- Add doc bock for the array structure for properties, parameters and return types (not all)
- Add static modifier for anonymous functions without `$this` usage
- Add `(int)` casts to data fetched from the database where the corresponding DB field is an integer. If the field is nullable the cast is placed inside an `inline if`
- Remove unnessesarry `&` (reference) from `foreach` (`ilSkillProfile:573`)
- Change `ilSkillUsageInfo::getUsageInfo` signatur, see below for details
- Add missing return statements
- Simplify boolean statements: `if($x == true)` => `if($x)`. `if($x){return true;} return false;` => `return !!$x;`
- Replace `array_column`, `array_map` and `array_keys` combo to map over an array with preserving the keys. See below
- Replace `$no_transform = $this->refinery->custom()->transformation(function ($v) {return $v;}` with `$no_transform = $this->refinery->identity();`
- Remove duplicated properties from `ilGlobalVirtualSkillTree` (already defined in `ilVirtualSkillTree`)
- Simplify `max(...)` in `ilSkillTree:115`
- Remove `is_array` check for parameter that is typed as an array
- Change `ilVirtualSkillTree::__getSubTreeRec` from references to return values. See below
- Move property `$base_skill_id` from `ilBasicSkillGUI` to it's parent class `ilSkillTreeNodeGUI` because it is used in the parent class.
- Change default parameter value to "levels" as it is defined in the overwritten method (`ilBasicSkillTemplateGUI::setTabs`)
- Add string type to the `$select_gui` property in the `ilPersonalSkillExplorerGUI` class and remove the doc block with string|object, because it is always a string, in the constructor the property is written with `(is_object($a_select_gui)) ? strtolower(get_class($a_select_gui)) : $a_select_gui`
- Same with `ilSkillSelectorGUI::$select_gui`
- Simplify `ilSkillUsage::getAssignedObjectsForSkillTemplate`
- Fix `PSR 2 + X`

### `getUsageInfo*`
I changed
` public static function getUsageInfo(array $a_cskill_ids, array &$a_usages) : void;`
to
`public static function getUsageInfo(array $a_cskill_ids) : array;`.
And the same with `getUsageInfoGeneric`.
Since the reference is only used as a return value I refactored it so that it is an actual return value.
In `ilSkillUsage:171` I merge them together with `array_merge_recursive`.
If you see my change as a performance issue I will remove it from this PR.

### `array_map` with preserved keys
In class `SkillGUIRequest` the pattern `array_column(array_map($callback, array_keys($arr), $arr), 1, 0)` is used to map over an array with preserving the array keys. I replaced these with `array_map($callback, $arr)`, because `array_map` already preserves the keys if only one array is supplied: https://www.php.net/manual/en/function.array-map.php
Furthermore the `bool` and `int` `array_map` callbacks are replaced with `intval` and `boolval` as they can be nicely supplied as a callback function. (e.g. `array_column(array_map(function($k, $v){return [$k, (int)$v];}, array_keys($arr), $arr), 1, 0)` is now `array_map('intval', $arr)`)

### `ilVirtualSkillTree::__getSubTreeRec`
Same reason as with `getUsageInfo` and if my change is a performance issue I will remove it from this PR.

### `dev.txt.php`
Should this file be renamed to `dev.txt`?

Best regards,
@lscharmer

### DIFF
--- a/Modules/Test/classes/class.ilTestPersonalSkillsGUI.php
+++ b/Modules/Test/classes/class.ilTestPersonalSkillsGUI.php
@@ -53,7 +53,7 @@ class ilTestPersonalSkillsGUI
 
         $gui->setProfileId($this->getSelectedSkillProfile());
 
-        $html = $gui->getGapAnalysisHTML($this->getUsrId(), $this->getAvailableSkills());
+        $html = $gui->getGapAnalysisHTML((int) $this->getUsrId(), $this->getAvailableSkills());
 
         return $html;
     }
--- a/Services/Skill/GlobalScreen/classes/class.ilSkillGSToolProvider.php
+++ b/Services/Skill/GlobalScreen/classes/class.ilSkillGSToolProvider.php
@@ -20,6 +20,7 @@
 use ILIAS\GlobalScreen\Scope\Tool\Provider\AbstractDynamicToolProvider;
 use ILIAS\GlobalScreen\ScreenContext\Stack\CalledContexts;
 use ILIAS\GlobalScreen\ScreenContext\Stack\ContextCollection;
+use ILIAS\UI\Component\Legacy\Legacy;
 
 /**
  * Workspace GS tool provider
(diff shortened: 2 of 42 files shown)


tfamula:
Hi @lscharmer 
Thank you very much for this elaborate review! I will have a closer look at all your points the next days and will come back again to this PR. Because of Coding Completed today and some remaining features I have to push into trunk, I already merged your PR. Otherwise, pushing the features beforehand would lead to merge-conflicts in this PR. At first glance, your review does not contain any critical changes and I also have not spotted `//TODO`s from you in the code, so the merge should be quite save :)

Greetings
Thomas

tfamula:
Hi @lscharmer 
I have reviewed your review. Great work, many thanks again! Special thanks for your explanations in this PR, they helped a lot. I added a few additional fixes and improvements here: https://github.com/ILIAS-eLearning/ILIAS/commit/7889869692faf2c1b149fe0d0c99d88972babebd.

> Should this file be renamed to dev.txt?

I am not sure about that. I checked other components and the file is always called `dev.txt.php`. Maybe it is for a certain reason, so I will leave it as it is.

PR #4412, opened 2022-04-25
PHP8: Code Review `Services/Randomization`

lscharmer:
Hi @mjansenDatabay,

here are the results of the `Services/Randomization` review.

### Results
- [ ] There is a `Unit Test Suite` with at least one unit test
- [ ] The unit tests pass
- [x] There are no serious `PHPStorm Code Inspection` issues left
- [x] The types are fully documented (PHPDoc) or explict PHP types are used (type hints, return types, typed properties)

### Changes made in this PR:
- Add Unit Test for class `ilRandom`
- Remove empty `ilRandom` constructor
- Change default values from `null` to `0` and `PHP_INT_MAX` (I checked that there are no usages of `ilRandom::int` where `null` is given as an argument). The max value is changed from `mt_getrandmax` to `PHP_INT_MAX` because `mt_getrandmax` is not relevant for `random_int` only for `mt_rand` and related functions.

Best regards,
@lscharmer

### DIFF
--- a/Services/Randomization/test/ilRandomTest.php
+++ b/Services/Randomization/test/ilRandomTest.php
@@ -1,0 +1,67 @@
+<?php declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+use ILIAS\DI\Container;
+use ILIAS\DI\LoggingServices;
+
+class ilRandomTest extends TestCase
+{
+    public function testConstruct() : void
+    {
+        $this->assertInstanceOf(\ilRandom::class, new ilRandom());
+    }
+
+    /**
+     * @dataProvider intArguments
+     */
+    public function testIntSuccessfully(int ...$arguments) : void
+    {
+        $this->expectNotToPerformAssertions();
+
+        $random = new \ilRandom();
+         try {
+             $random->int(...$arguments);
+        } catch (Error $e) {
+            $this->fail('Expected no exception.');
+        }
+    }
+
+    public function testIntWithInvalidArguments() : void
+    {
+        $this->expectException(Error::class);
+        $random = new \ilRandom();
+
+        $random->int(10, 9);
+    }
+
+    public function testLogIfPossible() : void
+    {
+        $this->expectException(Error::class);
+
+        $logger = $this->getMockBuilder(\ilLogger::class)->disableOriginalConstructor()->getMock();
+        $logger->expects(self::once())->method('logStack')->with(\ilLogLevel::ERROR);
+        $logger->expects(self::once())->method('error');
+
+        $factory = $this->getMockBuilder(ilLoggerFactory::class)->disableOriginalConstructor()->getMock();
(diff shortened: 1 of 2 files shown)


mjansenDatabay:
Thx a lot @lscharmer 

I am not sure about the first two unit tests. Testing `Random` in general is a philosophical matter we could discuss with a beer and some other community members :-).

I suggest to at least add (e.g.) 1000 iterations, similar to the tests here: https://github.com/ILIAS-eLearning/ILIAS/commit/e277371e7938dff12be993fec785cd8131bbfe00

lscharmer:
> Thx a lot @lscharmer
> 
> I am not sure about the first two unit tests. Testing `Random` in general is a philosophical matter we could discuss with a beer and some other community members :-).
> 
> I suggest to at least add (e.g.) 1000 iterations, similar to the tests here: [e277371](https://github.com/ILIAS-eLearning/ILIAS/commit/e277371e7938dff12be993fec785cd8131bbfe00)

The assertions with `>=`, `<=` and `===` are indeed volatile but they also test if the method can be called without and with arguments without throwing an error and the line `$this->assertTrue($random->int(8, 8) === 8)` tests an edge case to assure that no error is thrown, because the range is expected to be inclusive.

lscharmer:
> Thx a lot @lscharmer
> 
> I am not sure about the first two unit tests. Testing `Random` in general is a philosophical matter we could discuss with a beer and some other community members :-).
> 
> I suggest to at least add (e.g.) 1000 iterations, similar to the tests here: [e277371](https://github.com/ILIAS-eLearning/ILIAS/commit/e277371e7938dff12be993fec785cd8131bbfe00)

I removed the comparison assertions.

Services/Randomization/test/ilRandomTest.php
```diff
@@ -0,0 +1,66 @@
...
+        $this->assertInstanceOf(\ilRandom::class, new ilRandom());
+    }
+
+    public function testIntWithoutArguments() : void
+    {
+        $random = new \ilRandom();
+
+        $random->int();
+        $this->assertTrue(true);
+    }
```
mjansenDatabay: Can we used something like ...

```suggestion
    public function testIntWithoutArguments() : void
    {
        $this->expectNotToPerformAssertions();

        $random = new \ilRandom();

        try {
            $random->int();
        } catch (Error $e) {
            $this->fail('There should be not error when a random integer value is generated');
        }
    }
```

... instead?

Services/Randomization/test/ilRandomTest.php
```diff
@@ -0,0 +1,66 @@
...
+        $random->int(-20, 30);
+        $this->assertTrue(true);
+    }
+
+    public function testIntMinimalRange() : void
+    {
+        $random = new \ilRandom();
+
+        $random->int(8, 8);
+        $this->assertTrue(true);
```
mjansenDatabay: See comment above

PR #4420, opened 2022-04-26
ilCtrl: Proposal `@ilCtrlStructureCalls` Annotation

thibsy:
To sumarize: I removed all `@ilCtrl_Calls` and `@ilCtrl_isCalledBy` statements and replaced them with a newly introduced annotation called `@ilCtrlStructureCalls` that accepts `ilCtrl_isCalledBy` as parents and `ilCtrl_Calls` children. I also removed any unnecessary call-statements.

A note for the future: If the minimal supported PHP-version will be increased to 8.0, this annotation can be implemented via the new PHP Attributes and the Doctrine\Annotation package could be dropped. 

Cheers!


thibsy:
Replacement Script can be found at https://gist.github.com/thibsy/b620e4368fa3d1339741cde3d2ad53f5 in case anyone is interested.

matthiaskunkel:
Jour Fixe, 16 MAY 2022 : @thibsy presented his proposed PR and invites all community members to discuss pros and cons. If possible, this PR should still make it in 8 to keep all ilCtrl refactorings together. But this still has to be discussed. More information about the used library Doctrine Annotations here: https://github.com/doctrine/annotations

thibsy:
Hi @Amstutz, @klees and @mjansenDatabay 

I know you guys have a lot more important stuff on your plates, but I requested your reviews to sort of kick-off the discussion about whether this PR could still be merged in ILIAS 8 or not (as mentioned at JF). Of course you don't have to look at 500+ changed files :-). 

I think the most important topics are:
- do you see any risks with using the `Doctrine\Annotation` library? 
- should I introduce backwards compatibility for plugins (so old call-statements are still considered e.g. until ILIAS 9)?

Kind regards!

Amstutz:
Hi @thibsy 

thx for the ping. However, I have no clue, how I made this elaborate list (:-)) and now just removed my self from it. I have far too little insight here to contribute something meaningful:

_- do you see any risks with using the Doctrine\Annotation library?_
--> I have no clue, did not look into it. I can if you need some 2nd opinion. Should I?

_- should I introduce backwards compatibility for plugins (so old call-statements are still considered e.g. until ILIAS 9)?_
--> I would rather say no. IMO we can expect plugins to adapt for new versions. At one point, those steps have to be made anyways. However, just my 5 cents.

Thx

thibsy:
Thanks for the feedback @Amstutz, I think you don't need to look into the library then. I mainly pinged you because I figured your opinion on the backwards-compatibility matters :).

Amstutz:
Hi @thibsy

Maybe most importantly, thx for the time and thought you put into this! highly appreciated.

thx

mjansenDatabay:
Hi @thibsy,

> do you see any risks with using the `Doctrine\Annotation` library?

Using `Doctrine\Annotation` would be OK for me, because IMO the footprint is quite small. And: The used syntax is not complex, so if `Doctrine\Annotation` will not be maintained anymore, we could replace the implementation.

For me it is important to add appropriate unit tests for this, which you already provide with this PR. 

Furthermore I personally appreciate a migration to `PHP Attributes` once we can rely on this feature and `Doctrine\Annotation` supports this, too. IMO there should be a clear statement for this, maybe in a ROADMAP file.

> should I introduce backwards compatibility for plugins (so old call-statements are still considered e.g. until ILIAS 9)?

I think this it not necessary. Plugin developers have to fix/adjust a lot of things when upgrading to ILIAS 8/PHP 8 anyway, so a BC regarding the `ilCtrl` definitions is reasonable.

**But: I highly recommend to write an email to our admin and developer mailing lists**.

Best regards
@mjansenDatabay

mjansenDatabay (approved):

chfsx:
Hi @thibsy 

From my point of view, with the change to ILIAS 9 and thus also with the change to at least PHP 8, we should reconsider whether this information should continue to be provided via PHPDocs.

From my point of view (andboth also possible automated) there are at least these two possibilities:
- The introduction of an interface for "GUI" classes and thus the forcing of methods that return parents and children.
- Using PHP attributes to define the relationships between classes

I am very much leaning towards the attributes. This would also give us the advantage, for example, that GUI classes can be renamed much easier, because we can then work with `ClassnameGUI::class` everywhere.

A simple implementation with attributes could look like this:

```php
/**
 * GUI class for file objects
 */
#[CtrlParents(
    ilRepositoryGUI::class
)]
#[CtrlChildren(
    ilObjectMetaDataGUI::class,
    ilInfoScreenGUI::class,
    ilPermissionGUI::class,
    ilObjectCopyGUI::class,
    ilExportGUI::class,
    ilWorkspaceAccessGUI::class,
    ilPortfolioPageGUI::class,
    ilCommonActionDispatcherGUI::class,
    ilLearningProgressGUI::class,
    ilFileVersionsGUI::class,
)]
class ilObjFileGUI extends ilObject2GUI
{
    public const UPLOAD_MAX_FILES = 100;
...
```

chfsx:
And as an addition (and since we are sitting next to each other and discussed this point some minutes ago :-) ): I'd suggest ti implement an interface for `CommandClasses` anyway. This leads to the fact that we can identify such classes which are addressed by ilCtrl and this is not only done by the convention of `...GUI` in the name. Further we can force thereby a uniform `executeCommand` and if necessary require further methods, should those be needed.
Furthermore, this will provide us with easier ways to provide alternative (classic) routing that runs in parallel to ilCtrl, for example.

thibsy:
As discussed with @chfsx (my desk neighbor =)) I've decided to wait until ILIAS 9 with the change(es) of this PR, since the contribution is far more valuable when PHP attributes are used. Also, I'd have to open another PR as soon as the ILIAS 9 branch is available, since we'd like to replace all the annotiations by attributes again.

PR #4425, opened 2022-04-27
Fix Removal of Options in Long Menu Questions

kergomard:
This fixes https://mantis.ilias.de/view.php?id=32440. It is a pure selector issue.


mbecker-databay:
Great, thank you so much!

kergomard:
For future reference: This actually fixes: https://mantis.ilias.de/view.php?id=30651 and not https://mantis.ilias.de/view.php?id=32440

PR #4426, opened 2022-04-27
fix missing method call

chfsx:
@smeyer-ilias the function `cal_days_in_month` is not available if PHP is not compiled with the "--enable-calendar" option


smeyer-ilias:
Thanks @chfsx

PR #4433, opened 2022-04-28
Apply PHP 8: Code Review `QTI`

lscharmer:
Apply requested changes for the `Services/QTI` review:  https://github.com/ILIAS-eLearning/ILIAS/pull/4360

This PR is based upon the previously mentioned PR and adapts all comments added in the review (Unused methods are removed, nullable method parameters are checked if they can be non nullable and properties are checked if they can be non nullable).


Services/QTI/classes/class.ilQTIParser.php
```diff
@@ -190,10 +190,6 @@ public function __construct(?string $a_xml_file, int $a_mode = self::IL_MO_PARSE
         parent::__construct($a_xml_file);
 
         $this->qpl_id = $a_qpl_id;
-        if (is_array($a_import_idents)) { // TODO PHP8-REVIEW This can not happen anymore because the variable is typed as string. How the line below should be handled?
-            $this->import_idents = &$a_import_idents;
```
tfamula, line 205: Are you sure that this line can be completely removed?

PR #4434, opened 2022-04-28
Fix/8/tests failing

chfsx:
This PR provides some changes in UnitTests to make them succeed even on macOS

### DIFF
--- a/Services/VirusScanner/classes/class.ilVirusScannerFactory.php
+++ b/Services/VirusScanner/classes/class.ilVirusScannerFactory.php
@@ -35,7 +35,7 @@ class ilVirusScannerFactory
                     break;
                 case "AntiVir":
                     global $DIC;
-                    $DIC->logger()->error('AntiVir is deprecated, please install and use a different virus scanner.');
+                    $DIC->logger()->root()->error('AntiVir is deprecated, please install and use a different virus scanner.');
                     $vs = new ilVirusScannerAntiVir(IL_VIRUS_SCAN_COMMAND, IL_VIRUS_CLEAN_COMMAND);
                     break;
                 case "ClamAV":
--- a/src/Refinery/String/MakeClickable.php
+++ b/src/Refinery/String/MakeClickable.php
@@ -27,7 +27,21 @@ class MakeClickable implements Transformation
 {
     use DeriveApplyToFromTransform;
     use DeriveInvokeFromTransform;
-
+    
+    const PATTERN = '(^|[^[:alnum:]])(((https?:\/\/)|(www.))[^[:cntrl:][:space:]<>\'"]+)([^[:alnum:]]|$)';
+    
+    protected function determineParameters() : array
+    {
+        if (function_exists('mb_ereg')) {
+            $matcher = 'mb_ereg';
+            $pattern = self::PATTERN;
+        } else {
+            $matcher = 'preg_match';
+            $pattern = '@' . self::PATTERN . '@';
+        }
+        return [$matcher, $pattern];
+    }
+    
     /**
      * @inheritDoc
      */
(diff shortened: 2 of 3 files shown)


Services/VirusScanner/classes/class.ilVirusScannerFactory.php
```diff
@@ -35,7 +35,7 @@ public static function _getInstance() : ilVirusScanner
                     break;
                 case "AntiVir":
                     global $DIC;
-                    $DIC->logger()->error('AntiVir is deprecated, please install and use a different virus scanner.');
+                    $DIC->logger()->root()->error('AntiVir is deprecated, please install and use a different virus scanner.');
```
chfsx, line 38: @gvollbach FYI, this made this Test always failing on my local machine since I have assertions active and in \ILIAS\DI\LoggingServices::__call the amount of arguments was checked

src/Data/README.md
```diff
@@ -542,8 +542,7 @@ $dataset = $dataset->withPoint(
         "Target" => [0.95, 1.05],
     ]
 );
-
-assert($dataset->getMinValueForDimension("Measurement 1") === -1);
+assert($dataset->getMinValueForDimension("Measurement 1") === (float)-1);
```
chfsx: @klees as I mentioned yesterday, this fixes this test on my local machine. I just assume the assertions have not been made on other machines (while including this README and it's php code)

tests/Data/DataReadmeTest.php
```diff
@@ -15,27 +15,38 @@ class DataReadmeTest extends TestCase
...
     }
-
+    
     public function testReadme() : void
     {
         ob_start();
         require_once(__DIR__ . "/../../src/Data/README.md");
         ob_end_clean();
-        $this->assertTrue(true);
+        /** @noinspection PhpUndefinedVariableInspection */
```
chfsx: @klees beside the parsing of the code in the README I just added the same tests as PHPUnit assertions as well

tests/Data/DataReadmeTest.php
```diff
@@ -15,27 +15,38 @@ class DataReadmeTest extends TestCase
...
+        }
     }
-
+    
     protected function tearDown() : void
     {
-        ini_set("assert.active", $this->old_active);
-        ini_set("assert.bail", $this->old_bail);
-        ini_set("assert.warning", $this->old_warninig);
+        if (version_compare(PHP_VERSION, '8.0.0', '<')) {
```
chfsx: @klees there are some differences on setting the assertions option in PHP 8 and 7.4

src/Refinery/String/MakeClickable.php
```diff
@@ -38,14 +40,15 @@ public function transform($from) : string
         $endOfMatch = 0;
         $stringParts = [];
         $matches = [];
-        while (1 === preg_match('@(^|[^[:alnum:]])(((https?://)|(www.))[^[:cntrl:][:space:]<>\'"]+)([^[:alnum:]]|$)@', substr($from, $endOfMatch), $matches)) {
+        
+        while (mb_ereg(self::PATTERN, substr($from, $endOfMatch), $matches)) {
```
chfsx: @mjansenDatabay and @lscharmer I has to replace preg_match with mb_ereg since all tests with multibytes failed locally on my machine, looked like this:

image

chfsx: It's what @Amstutz mentioned in #4441

mjansenDatabay: Thx @chfsx

src/Refinery/String/MakeClickable.php
```diff
@@ -38,15 +58,22 @@ public function transform($from) : string
         $endOfMatch = 0;
         $stringParts = [];
         $matches = [];
-        while (1 === preg_match('@(^|[^[:alnum:]])(((https?://)|(www.))[^[:cntrl:][:space:]<>\'"]+)([^[:alnum:]]|$)@', substr($from, $endOfMatch), $matches)) {
+        [$matcher, $pattern, $check] = $this->determineParameters();
+        
+        while ($check == call_user_func_array($matcher, [$pattern, substr($from, $endOfMatch), &$matches])) {
```
chfsx: I wrapped this in call_user_func_array to support preg_match if mb_ereg is not available

Modules/StudyProgramme/test/helpers/ilStudyProgrammeGUIMessagesTest.php
```diff
@@ -28,11 +33,10 @@ public function testMessageFactory() : ilPRGMessageCollection
         return $collection;
     }
     
-    /**
-     * @depends testMessageFactory
-     */
-    public function testCollectionDefaults(ilPRGMessageCollection $collection)
```
chfsx: @klees / @nhaagen I removed the dependency of those tests from `testMessageFactory ` since this does not work with random testing

chfsx:
- [x] @mjansenDatabay I'ts strange, locally the tests for MakeClickable now run (in both versions, with preg_match and with mb_ereg) but on GH those tests now fail...

chfsx:
@mjansenDatabay seems to work with https://github.com/ILIAS-eLearning/ILIAS/pull/4434/commits/406f9c60b485521188ac91f771e98b2eb5a22d9e

src/Data/README.md
Amstutz: I have the same issue with this as @chfsx does. Only on macOS. (float)-1 does fix the problem for me as well.

chfsx:
@Amstutz and @klees have a look at https://github.com/ILIAS-eLearning/ILIAS/pull/4434/commits/7f0928b087275288cef966d9bea693773253549c, I'm not sure if this is a good idea. the problem is, that the canonical_name of a component is only set, it at least 1 time getCanonicalName() is called: src/UI/Implementation/Component/ComponentHelper.php:25

tests/Data/DataReadmeTest.php
klees: I do not understand this, tbh. The general idea for this test is: include the readme, which executes the code, and, while doing that, check assertions that are written down in the readme. Why would we add these assertions as tests here?

chfsx:
@Amstutz and @klees it's not the solution in https://github.com/ILIAS-eLearning/ILIAS/commit/7f0928b087275288cef966d9bea693773253549c, the test is now flacky since due to random canonical_name is sometimes `'canonical_name' => null`, sometimes `'canonical_name' => 'Standard Icon Symbol'`

Amstutz:
@chfsx : Ignore this test. @klees and I need to find a better solution on either a) testing this or b) the way we set this name in the first place. Note if need be you can temporarily deactivate it here. However, I am confident, we have this fixed soon.

klees (approved):

mjansenDatabay (approved):

chfsx:
The latest commit suggests to add a possibility to add files/folders via ENV to the list of allowed files/folders in RootFolderTest: https://github.com/ILIAS-eLearning/ILIAS/pull/4434/commits/7a861672126b0ded3145c2c3e196442fb57eb72d

Amstutz (approved):
Should be merged soon, imo. thx a lot.

PR #4435, opened 2022-04-28
Bugfix/repository pattern

iszmais:
Hi @klees ,

While implementing a repository pattern i encountered an failure inside the examples of the repository pattern, which i corrected.


Amstutz:
Thx a lot!

PR #4441, opened 2022-04-28
Unit Tests: Multibyte Assertion Tests Problems on MacOS

Amstutz:
Had a short chat with @chfsx on this. We both use macOS and stumble over this. I have not found a solution so far. This is a pity, since I have always 2 failing tests since quite some time. Therefore, I would kindly ask to remove the lines causing the issue ftm. 

If somebody has a solution for this, other than removing the lines, even better :-)


Amstutz:
image

mjansenDatabay:
@Amstutz I close this because @chfsx provided a nice solution with #4434

PR #4446, opened 2022-04-28
replaced some superglobals in Services/Init

smeyer-ilias:
This fixes some superglobals access  issues. 
I added comment for some places where the refinery is undefined or the sideeffects by the replacement to refinery are difficult to foresee.


Services/Init/classes/class.ilInitialisation.php
```diff
@@ -1578,7 +1580,8 @@ protected static function initHTML() : void
             // or not set at all (then we want the last offset, e.g. being used from a session var).
             // So I added the wrapping if statement. Seems to work (hopefully).
             // Alex April 14th 2006
-            if (isset($_GET['offset']) && $_GET['offset'] != "") {                            // added April 14th 2006
+            // @todo not replaced by refinery due to unknown sideeffects
```
kergomard, line 1591: I wondered what "unknown sideeffect" means. What is unknown about them?

Services/Init/classes/class.ilInitialisation.php
```diff
@@ -1569,6 +1570,7 @@ protected static function initHTML() : void
         if (ilContext::hasUser()) {
 
             // set hits per page for all lists using table module
+            // @todo this is not fixable due to unknown sideeffects.
```
kergomard, line 1581: I wondered what "unknown sideeffect" means. What is unknown about them?

kergomard (commented):
Other then my comments, this looks good to me.

chfsx:
As a technical board, we regularly check whether pull requests have been open for a long time.
Are there any updates on this? Please note that you can also close this request if you are not able or if you do not have the resources to look into it in detail.

At this point we would like to refer to the rules we all agreed on: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/contributing.md#rules-for-maintainers-assigned-to-prs.

With kind regards!

Services/Init/classes/class.ilInitialisation.php
pascalseeland: The side effects in my opinion are that we don't know, if somebody deep down in the calling chain might still need the value added to the $_GET. Thus changing this wo only read it with refinery would break the code. Changing this requires a scan through the entire code base.

PR #4453, opened 2022-04-29
Object: Rename button label in 'copy' creation form (Mantis 32248)

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=32248


Amstutz:
Since this component is atm without maintenance, we decided to merge this to trunk. Note, that we can currently only handle simple PRs and only pick down on request since we do not have the resources to do more without dedicated maintainer.

PR #4458, opened 2022-05-02
Object: Change label of didactic template element (Mantis 32531)

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=32531

### DIFF
--- a/Services/Object/classes/class.ilObjectGUI.php
+++ b/Services/Object/classes/class.ilObjectGUI.php
@@ -748,7 +748,7 @@ class ilObjectGUI
 
         if (sizeof($options) > 1) {
             $type = new ilRadioGroupInputGUI(
-                $this->lng->txt('type'),
+                $this->lng->txt('didactic_template'),
                 'didactic_type'
             );
             // workaround for containers in edit mode


dsstrassner:
Hi JourFixe!

As institution, I am not pleased with this change of code, since we need to patch our system back to the old code stand, if we want to use "type" for this label and other objects. 
While "Type" is well understood by our users, this is not the case with "didactic template". If the change was only in the language variable, one could simply restore this oneself. 

Therefore, I ask to consider this again carefully.

jf:
please fix your reference. "jf" is me, and I am not involved with this project.

jf:
> Hi @jf!
> 
> As institution, I am not pleased with this change of code, since we need to patch our system back to the old code stand, if we want to use "type" for this label and other objects. While "Type" is well understood by our users, this is not the case with "didactic template". If the change was only in the language variable, one could simply restore this oneself.
> 
> Therefore, I ask to consider this again carefully.

please fix your reference. "jf" is me, and I am not involved with this project.

dsstrassner:
> > Hi @jf!
> > As institution, I am not pleased with this change of code, since we need to patch our system back to the old code stand, if we want to use "type" for this label and other objects. While "Type" is well understood by our users, this is not the case with "didactic template". If the change was only in the language variable, one could simply restore this oneself.
> > Therefore, I ask to consider this again carefully.
> 
> please fix your reference. "jf" is me, and I am not involved with this project.

Sorry. :( Done.

matthiaskunkel:
Jour Fixe, 02 MAY 2022 : As there are several complaints about the re-labeling of the current "Type" to "Didactic Templates" we postpone a decision about this PR and ask the maintainer of the language service to organise a 1h workshop to discuss and find a better wording that reflects both aspects - a clear distinction between those "types" that are didactic templates and those that have different functions - e.g. 360° survey vs. usual survey.

Amstutz:
Closed until discussed, feel free to re-open afterwards.

PR #4460, opened 2022-05-02
Security: introduces a configurable delay for login and registration attempts

thibsy:
Hi all,

Security reporters have made a few remarks about the login and registration process in ILIAS. One issue is the gathering of information about the existence of a profile by inspecting the response-times when logging in or registering a new account. This PR now introduces the ability to enable and configure a delay that will be applied whenever a login or registration attempt is made. Optionally the amount of the delay can be increased with each failed attempt by a custom multiplier, which is also a security measure against brute-force attacks. 

I've implemented a new factory within the HTTP service that is reachable with `$DIC->http()->delay()`. The factory can be used within controllers (such as the ilStartUpGUI or ilAccountRegistrationGUI) to register delays that can later be retrieved and consumed. Delays can be supplied with `DelayIncrement`'s which the factory automatically applies after consuming a delay.

Kind regards


thibsy:
I've implemented the configuration of these kinds of delays in Administration > General Settings > Basic Settings as you can see in the screenshot below. I'm not quite sure if that is the best place, let me know if there is a better one.
Screenshot 2022-05-02 at 10 48 45

mjansenDatabay:
Hi @thibsy,

> I've implemented the configuration of these kinds of delays in Administration > General Settings > Basic Settings as you can see in the screenshot below. I'm not quite sure if that is the best place, let me know if there is a better one.

Maybe the `Security` tab in `Administration > User and Roles > Privacy and Security > Security` would be an appropriate alternative.

---

I suggest to add a few sentences to the README.md of `src/HTTP` once this PR gets an approval.


Best regards,
@mjansenDatabay

matthiaskunkel:
Jour Fixe, 02 MAY 2022 : We highly appreciate this suggestion to improve the security of ILIAS. It would be good not to rely on a session-based solution only. Considering the remote-IP or fingerprinting the browser could be options. And concerning presenting the function: please present it in "System Settings and Maintenance » Authentication and Registration : Authentication : Authentication Settings" and present the feature also in "Privacy and Security: Security" with a link to the authentication administration.

thibsy:
I just checked how other frameworks like symfony or laravel handle request throttling during login attempts and they mostly rely on a storage mechanism based on cache and consider the IP-address of the client. I wouldn't recommend using IP-address for determining recurring visitors in ILIAS, because I believe many institutions use a proxy, which leads to the same IP being used for many different visitors. Furthermore, ILIAS doesn't require caching and the `global cache` is optional, so it would not be possible to rely on that either. Therefore I suggest to evade to an alternative approach based on form-tokens:

The login- and registration-form will be supplied with a new token on each request, which will be mapped to a delay. The controller can then check if the attempt failed and (optionally) increase the delay for the next submission e.g. form-token and associate it with the new form-token. The storage of this kind of association must rely on the database, since the cache will most likely not be available and the session can be suppressed.

What do you guys think?

@klees @mjansenDatabay @smeyer-ilias

klees:
@thibsy I have to think about this for a moment... Maybe it is worth scheduling a meeting with interested devs to discuss the matter?

mjansenDatabay:
IMO we should not reinvent the wheel. This is a security bugfix which should address an issue which is reported in almost every penetration test. If popular and industry relevant frameworks already have proven mechanisms, we could and maybe should adopt them somehow. If the actual issue prevents institutions from using ILIAS and the penetration testers are fine with the level of security offered by the approach proposed here (and in other web applications using such industry relevant frameworks), we should provide an MVP approach ASAP, as long as the interfaces are elaborated.
If most `Firewall/Rate Limiter/Throttling` approaches rely on the client IP address, we should offer this **as one** strategy (of maybe more in the future, where we could extend these). Because ILIAS is often used together with (reverse) proxies, the header name to retrieve the forwarded remote address should be configurable, e.g. in the JSON config used by the setup.

As I already mentioned in the JF: We just cannot prevent `User DoS` attacks in its entirety. IMO real `DoS` and `DDoS` attacks must be addressed in other parts of the infrastructure too, not only in the PHP application (ILIAS). This means we should not be ignorant of the issue as such, but we should also emphasize that the PHP application (the way ILIAS is built) alone cannot mitigate the threat completely. So to be good as application we should at least create some hurdles attackers have to jump over.

thibsy:
> IMO we should not reinvent the wheel

True, I think we could just adopt a version of this feature/bugfix that relies on the IP-address and leave the Codebase open for extension. We would have to evade to database storage for now (unlike industry standards), since ILIAS as mentioned doesn't require caching. This could may be tackled in ILIAS 9.

> we should provide an MVP approach ASAP, as long as the interfaces are elaborated

I agree. I also don't want to "blow up" this PR more than it needs to, so I wouldn't necessarily schedule a meeting @klees. I kind of hoped to present a MVP on monday.

klees:
Hi @thibsy,

the feedback will be a little unstructured and not very close to the code, hope it still helps.

I'm definitely with @mjansenDatabay regarding "not reinvent the wheel". I'm not convinced that "ship an MVP fast" is a good approach to security issues. On the contrary: I rather think, that security related code should be considered super carefully, especially if one wants to implement their own solutions instead of reusing an existing and proven one. Give many people lots of time to think about security things before shipping (if possible), to find bugs in concept and code.

The proposed solution with an increasing timeout bound to a hidden token in the form will only slow down legitimate users. Adversaries will just request a fresh token for every try instead of reusing the one with the increased timeout. Basically, the token is a cookie.

I think there are two rather different problems that you try to solve at once: How can we identify an adversary (with enough precision) if they try to avert identification? How can we prevent sensible information ("does this user exist") to leak via a side channel ("time of request"). I think the second one is easier to solve than the first one and I would suggest to solve them one after another.

Regarding the first problem ("identification of an adversary"): The fact, that Symfony and many others use IPs based throttling basically tells me, that we won't find anything better. Relying on any information that the adversary controls (token, cookie, ...) for identification will fail. It would be ok though, IMO, if we also accidentally throttle other users on the same IP. Their are on a potentially malicious network, after all.

For the second problem (side channel, which seems to be the thing that the actual security report mentions) I would propose another API, something along the lines of:

```php

$section = $service->startSectionWithFixedExecutionTime(/* ... some options, e.g. just a fixed amount of seconds to be expected, or something even more elaborate ... */);

code();
code();
code();

$section->waitUntilFixedExecutionTimeIsExpired(); // will make sure that the time between start and wait is like configured.
```

This has various advantages over the API that you have proposed: The section exists explicitly, hence there will be no problem if `startSectionWithFixedExecutionTime` is called twice, even with the same options. How would `delay()->consume` behave when the same `$identifier` is used twice? Also: no hidden state, everything is in the `$section` object. Since the section then needs to be passed explicitly (if at all!), code should be a lot easier to reason about.

Finally: What's the reason to add the code to the HTTP service? I can barely see any relation to the other stuff in there, except for the fact that we might be interested in some IP at some point.

Hope this helps a little, feel free to contact me for further discussion.

Best regards!

thibsy:
Hi @klees and thanks for the feedback!

> I think there are two rather different problems that you try to solve at once

You're right, the primary goal was to make sure attackers aren't able to gather sensible information via side-channel and that's the one I will focus on. I like your approach where the delay exists explicitly, I'll come up with an implementation soon.

> What's the reason to add the code to the HTTP service?

Initially I wanted to couple this mechanism to the HTTP-response-sender, which sadly isn't possible because response objects are barely ever used. I still left the implementation there because I hope that the mechanism can be integrated in the future when e.g. building a response. Otherwise, I could move the implementation into Utility. 

> I'm not convinced that "ship an MVP fast" is a good approach to security issues

Agreed, though I think it would suffice to request a few code-reviews when I'm done and present the solution at JF, right?

Kind regards!

src/HTTP/Throttling/Increment/MultiplierIncrement.php
```diff
@@ -0,0 +1,38 @@
...
+    protected float $multiplier;
+
+    public function __construct(float $multiplier)
+    {
+        $this->multiplier = $multiplier;
+    }
+
+    public function increment(float $delay) : float
+    {
+        return ($this->multiplier * $delay);
```
corro, line 36: Is there an upper limit to the delay? I am a bit concerned that an attacker might abuse this feature for a DoS attack: When all worker threads are sleeping (potentially for a long time), there might not be any available for legitimate requests. Or is this already considered elsewhere?

thibsy: You're right, there isn't an upper limit yet but we'll have to take this into account. Thanks for the feedback!

thibsy:
I close this PR since we have decided to split up the two Issues it tried to solve at one: constant authentication-time (#4543) and increasing the authentication-time on repeated failures (PR yet to come).

PR #4462, opened 2022-05-02
Composer: Add phpstan as development dependency

mjansenDatabay:
This PR suggest to add `PHPStan` as explicit `Composer` development dependency.


Amstutz:
Thx a lot @mjansenDatabay . Approved by TB.

PR #4465, opened 2022-05-02
unified dashboard visibilities with common repository visibilities

iszmais:
https://mantis.ilias.de/view.php?id=32312

This PR unifies the visibility of repository items on the dashboard to be similar to the apperance inside of the repository.
This does not effect the scopes (Favourites, Groups and Courses, Recommendet Content) itself, only the content so this may cause empty scopes. This may lead to 3 minor inconsitencies:
- The user is told that he has not favourites while these may just be not visible for him.
- The recommended content may just display as a heading without context.
- The user is told that he has not courses and groups while these may just be not visible for him.

The first and last can possibly occur already, but are more likely to appear with this changes. Therefore a general, more satisfying solution could be discussed.


iszmais:
~~@smeyer-ilias @alex40724  i would like to hear your opinion on that.~~

~~I Think its not really "controversial" enough to get its own time scope in a JF, since its mostly just a bugfix, but feel free to correct me there if you see any argument why the accesses of your scopes should be treaded differently or are already treated differently in there specified context.~~

@lscharmer this is still in disscuss i will notice you if all cases are cleared, but feel free to already do a partial code review if you can find the time.

Greetings,
@iszmais

lscharmer (approved):

PR #4468, opened 2022-05-03
moved notification node to communication

iszmais:
https://mantis.ilias.de/view.php?id=32541

As clarified inside the comments of https://docu.ilias.de/goto.php?target=wiki_1357_Integrate_On-Screen_Notifications_into_Notification_Center the notification administration is moved outside the administration into the communication section


mjansenDatabay:
@iszmais Do we have to add this object type (`nota`) to the list in `AdministrationMainBarProvider`, according to the comment of https://mantis.ilias.de/view.php?id=32541#c82088

iszmais:
@mjansenDatabay Im really confused right now.

This PR move the node out of the administration and into the communication, as it is requested in the comment of the FR.
Why would this need a mention inside the Administration?
If any this would need to have an occurence like that removed.
Additional, as far as i know all menu entries are now controlled by the local provider (or at least this is the only requirement to be fully functional). Why would these need a doubled maintainance?

I hope i could explain my confusion, feel free to call me if clearification is needed.

Greetings,
@iszmais

mjansenDatabay:
IMO the provider is not needed anymore if the node is put below the `Communication` entry.

As far as I understand this `nota` has only to be added to the `$layout` array in `\ILIAS\Administration\AdministrationMainBarProvider::getGroups`, similar to (for instance) `mail`. The mail component does not have a provider for the administration node.

iszmais:
@mjansenDatabay Ah ok, so the JF decision refers to the sub-node **administration >> communication** and not the main menu entry **communication**!

Had a missunderstanding there. its now fixed.

Greetings,
@iszmais

PR #4475, opened 2022-05-04
Fix presentation of the root node inside the favourites

iszmais:
https://mantis.ilias.de/view.php?id=32352

This creates an empty Location to place the root node in for the view "Sort by location" of the favourites.

and

creates a new Type **Repository** to place the root node in for the view "Sort by Type".

Wich fixes errors and invisibilities of the root node when its put into favourites.

### DIFF
--- a/Services/Component/classes/class.ilCachedComponentData.php
+++ b/Services/Component/classes/class.ilCachedComponentData.php
@@ -138,9 +138,9 @@ class ilCachedComponentData
             $parent = $rec['parent'];
             $this->subobj_for_parent[$parent][] = $rec;
         }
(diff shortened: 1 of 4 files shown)


iszmais:
@Amstutz This would resolve the error mentioned inside the mantis ticket and also fix the other sorting of favourites. But i not quite sure if its necessary to have the option to put the root node inside the favourites at all. I would like to hear your opinion on that. Feel free to place a recommendation if you think this is JF material.

@lscharmer as always: feel free to review the code, but await a final decision before merge.

Amstutz:
Thx a lot for the deep dive you did here! Deactivating favorites for root node would be completely fine with me as well. Your call. See also reply in the ticket.

iszmais:
@Amstutz Thx for the response!
I Agree that this will be the better solution. Therefore i created an new PR wich targets the initial problem.
https://github.com/ILIAS-eLearning/ILIAS/pull/4480

I will close this PR and the depending.

Greetings,
@iszmais

PR #4479, opened 2022-05-04
Replaced BulkUpload in MediaPool with new inputs

chfsx:
Hi @alex40724 

Currently the `ilFileStandardDropzoneInputGUI` in trunk is broken and we would like to replace it anyway with new Inputs. I therefore refactored the Bulk-Upload in MediaPool.

There would be the possibility to use a "nice" dropzone instead of the here proposed Input\File, but this would need a HTML-snippet for the Dropzone to be created (have a look at Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php:96, you can change the line Modules/MediaPool/classes/class.ilObjMediaPoolGUI.php:1684 to ...->getAsDropZone() to have a look at it). 

the current implementation in the PR uses Input\File.

beside this it would be possible to have the "Title" and "Description" thing done with DynamicFields as well, in PR this has been disabled. Feel free to contact me at any time.

### DIFF
--- a/Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php
+++ b/Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php
@@ -1,0 +1,230 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ */
+
+use ILIAS\FileUpload\Handler\AbstractCtrlAwareUploadHandler;
+use ILIAS\FileUpload\Handler\FileInfoResult;
+use ILIAS\FileUpload\Handler\HandlerResult;
+use ILIAS\UI\Component\Dropzone\File\Wrapper;
+use ILIAS\FileUpload\Location;
+use ILIAS\FileUpload\Handler\BasicFileInfoResult;
+use ILIAS\FileUpload\Handler\BasicHandlerResult;
+use ILIAS\UI\Component\Input\Field\Group;
+
+/**
+ * @author            Fabian Schmid <fabian@sr.solutions>
+ *
+ * @ilCtrl_IsCalledBy ilMediaPoolBulkUploadGUI: ilObjMediaPoolGUI
+ */
+class ilMediaPoolBulkUploadGUI extends AbstractCtrlAwareUploadHandler
+{
+    const MAX_FILE_AMOUNT = 20;
+    const MEP_ID = 'mep_id';
+    private \ILIAS\UI\Factory $ui_factory;
+    private ilObjMediaPool $media_pool;
+    private ilLanguage $lng;
+    
+    private string $post_url;
+    private int $mep_item_id;
+    private int $max_upload_size = 2048;
+    private string $upload_hash;
+    
+    public function __construct(
+        ilObjMediaPool $media_ppol,
+        int $mep_item_id,
+        string $post_url,
+        string $upload_hash
+    ) {
+        global $DIC;
+        parent::__construct();
+        $this->post_url = $post_url;
+        $this->upload_hash = $upload_hash;
+        $this->mep_item_id = $mep_item_id;
+        $this->media_pool = $media_ppol;
+        $this->ui_factory = $DIC->ui()->factory();
+        $this->lng = $DIC->language();
+        $this->max_upload_size = (int) ilFileUtils::getUploadSizeLimitBytes();
+    }
+    
+    protected function getMetadataInputs() : Group
+    {
+        return $this->ui_factory->input()->field()->group([
+            $this->ui_factory->input()->field()->text(
+                $this->lng->txt('title')
+            ),
+            $this->ui_factory->input()->field()->textarea(
+                $this->lng->txt('description')
+            )
+        ]);
+    }
+    
+    public function getAsForm() : ILIAS\UI\Component\Input\Container\Form\Standard
+    {
+        $metadata_input = null; // $this->getMetadataInputs(); the additional fields could be used directly in the upload form
+        
+        $inputs = [
+            $this->ui_factory->input()->field()->group(
+                [
+                    $this->ui_factory->input()->field()->file(
+                        $this,
+                        $this->lng->txt('mep_media_files'),
+                        null,
+                        $metadata_input
+                    )->withMaxFiles(self::MAX_FILE_AMOUNT)
+                                     ->withMaxFileSize($this->max_upload_size)
+                ],
+                $this->lng->txt('mep_bulk_upload')
+            )
+        
+        ];
+        return $this->ui_factory->input()->container()->form()->standard($this->post_url, $inputs);
+    }
+    
+    // TODO check if you want this to keep: it would be possible to have a nice (TBD) dropzone to make the uploads
+    public function getAsDropZone() : Wrapper
+    {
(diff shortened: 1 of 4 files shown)


chfsx:
@alex40724 there is another PR (#4490) for larger dropzones, if accepted I'd suggest to replace the bulkUpload in MediaPool with a large Dropzone instead of e UI\Form

chfsx:
I think this could be adapted in `ilMediaCreationGUI` the same way, after this no other usages of `ilFileStandardDropzoneInputGUI` are in trunk and we will delete it.

Services/MediaObjects/classes/class.ilWidthHeightInputGUI.php
```diff
@@ -21,9 +21,9 @@
 class ilWidthHeightInputGUI extends ilFormPropertyGUI
 {
     protected bool $constrainproportions;
-    protected ?int $height;
-    protected ?int $width;
-    protected array $dirs;
+    protected ?int $height = null;
```
chfsx, line 24: @alex40724 default to null must be set here. Not related to the PR here but a bug anyway

Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php
```diff
@@ -0,0 +1,230 @@
...
+use ILIAS\FileUpload\Handler\BasicFileInfoResult;
+use ILIAS\FileUpload\Handler\BasicHandlerResult;
+use ILIAS\UI\Component\Input\Field\Group;
+
+/**
+ * @author            Fabian Schmid <fabian@sr.solutions>
+ *
+ * @ilCtrl_IsCalledBy ilMediaPoolBulkUploadGUI: ilObjMediaPoolGUI
+ */
+class ilMediaPoolBulkUploadGUI extends AbstractCtrlAwareUploadHandler
```
alex40724, line 30: I never used/extended a AbstractCtrlAwareUploadHandler. Is there any documentation on this? When and how to use this kind of class?

Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php
```diff
@@ -0,0 +1,230 @@
...
+{
+    const MAX_FILE_AMOUNT = 20;
+    const MEP_ID = 'mep_id';
+    private \ILIAS\UI\Factory $ui_factory;
+    private ilObjMediaPool $media_pool;
+    private ilLanguage $lng;
+    
+    private string $post_url;
+    private int $mep_item_id;
+    private int $max_upload_size = 2048;
```
alex40724, line 40: Why do these kind of classes set this limit? Isn't this supposed something configured in ini files and/or global settings?

Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php
```diff
@@ -0,0 +1,230 @@
...
+                ],
+                $this->lng->txt('mep_bulk_upload')
+            )
+        
+        ];
+        return $this->ui_factory->input()->container()->form()->standard($this->post_url, $inputs);
+    }
+    
+    // TODO check if you want this to keep: it would be possible to have a nice (TBD) dropzone to make the uploads
+    public function getAsDropZone() : Wrapper
```
alex40724, line 95: I understand that it's either a form or a dropzone, right? I prefer the form alternative then.

Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php
```diff
@@ -0,0 +1,230 @@
...
+                        ''
+                    );
+                }
+            } catch (Exception $e) {
+                $result = new BasicHandlerResult(
+                    $this->getFileIdentifierParameterName(),
+                    BasicHandlerResult::STATUS_FAILED,
+                    '',
+                    $e->getMessage()
+                );
```
alex40724, line 195: Also this BasicHandlerResult and whole logic that needs to be implemented to handle uploads. Is this part of any README file? I did not find anything.

alex40724 (changes requested):
@chfsx Thanks for the PR. Of course I would like to get this working, too. My main questions are about documentation of the current "API" of these kind of uploads. Is there any? I found similar code but I would not have been able to come to this solution myself without just copying it from elsewhere.

Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php
chfsx: In general one could use it's own endpoints of URL-building for an uploadhandler, but in most cases you just want to use ilCtrl. I added a docu here: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Services/FileServices/README.md

Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php
chfsx: see https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Services/FileServices/README.md

Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php
chfsx: the proposal gets it's limit from `$this->max_upload_size = (int) ilFileUtils::getUploadSizeLimitBytes();` which is the 'global setting'. but we will see the need for custom limits in the future, e.g. see 
https://docu.ilias.de/goto_docu_wiki_wpage_6387_1357.html

but you can change this as you want. I just proposed a fix for the issue, that the ilFileStandardDropzoneInputGUI does no longer work.

Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php
chfsx: yes, I proposed both versions and personally find the dropzone much sexier, in particular if you could use this: #4490. but it's up to you

chfsx:
Hi @alex40724 what change exactly do you request? if it's the docu, i pushed that here: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Services/FileServices/README.md

you can see this PR as a proposal how to do it, no need to merge it. E.g. I will not extend the PR to the other location where a ilFileStandardDropzoneInputGUI is currently used since you maybe want to refactor this anyway (seems to be mostly duplicate code in ilMediaCreationGUI:performBulkUpload)

Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php
alex40724: Great, thanks!

Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php
alex40724: Ok.

Modules/MediaPool/classes/Upload/class.ilMediaPoolBulkUploadGUI.php
alex40724: Thanks.

alex40724:
@chfsx  Yes, these came from different branches. I will check how to proceed with this.

Thanks for the documentation. This was the main reason why I did no set this to "Approved" (and the hardcoded upload size).

Why closing the PR? Do we need the change this? Will ilFileStandardDropzoneInputGUI be fixed or is the migration a "must" for ILIAS 8?

chfsx:
> Do we need the change this? Will ilFileStandardDropzoneInputGUI be fixed or is the migration a "must" for ILIAS 8?

we won't be able to fix `ilFileStandardDropzoneInputGUI`, so therefore a change is needed but you may implement it the way you like

alex40724:
@chfsx  I switched the implementation. ilFileStandardDropzoneInputGUI can be removed from the code now.

chfsx:
Thanks for the info, @alex40724

PR #4480, opened 2022-05-04
Prevent root node from beeing recommended or favourite

iszmais:
https://mantis.ilias.de/view.php?id=32352

Prevents root node from beeing drafted as recommended content and therfore from beein a favourite


Amstutz (approved):
Thx a lot!

alex40724:
Thanks for the PR. Merged manually.

PR #4122, opened 2022-03-14
WIP: Fix/php8/modules/orgunit

mstuder:

PR #4486, opened 2022-05-04
Change/php 8/orgunits

mstuder:


mstuder:
feedback from @daniwe4 to the old pullrequest
https://github.com/ILIAS-eLearning/ILIAS/pull/4122

Hi @mstuder ,
a first check against https://docu.ilias.de/ilias.php?ref_id=10290&cmd=view&cmdClass=ilobjcontentpagegui&cmdNode=j1:j&baseClass=ilrepositorygui results in the following output. 
Please let us know if we should check again.

# Must-Package
### What must be the agreed minimum goal from the community?  - Whether developers, operators or users.

## Goals
 * [x] ILIAS can be used with PHP 8 without producing PHP-specific errors.
 * [ ] Refactored code conforms to current coding styles
 * [ ] Elimination of all warnings and errors of the static code inspection (without weak | incl. undeclared variables)
   * 24 errors
   * 132 warnings
   * 140 weak warnings
 * [ ] Transfer of GET params into Constructor
   * 85 places where $_GET is used
 * [ ] Avoidance/reduction of POST accesses (e.g. through $form->getInput()) and switch to Request classes
   * 63 places where $_POST is used
 * [ ] Use of SESSION (session object, alternative assignment of static session in constructor member variables)
   * 5 places where $_SESSION is used 
 * [ ] Documentation of all existing typifications as typehints (of parameters, return values and properties)
   * There are some errors, because of using PHP8 features that not supported by PHP7.4, that belongs to type declarations.
 * [x] Creation of a unit test infrastructure in the component


## Requirements and analysis criteria
* [ ] Preserve unit test coverage even in refactoring.
* [ ]  Simple call of all tabs/screens (without e.g. delete and move actions) and fixing of the corresponding bugs
* [ ]  Checking of external libraries
* [ ]  Basis of the refactoring and the reviews is, among others, ILIAS Development Documentation.



# Should-Package
### What do we need to do in the mid-term to catch recurring large code revisions and to refactor more than the minimum?

 * [ ] Expansion of the unit test coverage
  * General and structural improvements and optimisations to the code base. For example ...
   * [ ]  ... in static functions and by removing static methods
   * [ ]  ... by introducing return type declarations
   * ... by typification of all variables
     * [ ] Input/output typing of methods
     * [ ] Typification of member variables
 * [ ]   Use of UI components from the UI framework "KitchenSink" that have been introduced and established in the meantime. Reduction of legacy UI and legacy code.

# Sustainability-Package
### he developer community has identified topics that would also improve the quality of ILIAS from the point of view of PHP and ILIAS in general. Ideally, these topics could be directly addressed in the process of systematic refactoring.

* [ ] Introduction and use of declarations of "strict_types" for data
* [ ]  Replace previous arrays with classes that enable ArrayAccess - in future also getter / setter
* [ ]  Switching previous arrays to generators/data objects

mstuder:
@daniwe4 is it sufficient for the static code inspections, when the github checks are OK?
https://github.com/ILIAS-eLearning/ILIAS/actions/runs/2274198564

daniwe4:
Hi @mstuder ,
i use the code inspection files for PHP-Storm under docs/development/inspection-configs. I think there is a way for those don't using PHP-Storm but i don't know where to find it.

klees:
Hi @mstuder,

please also refer to [the criteria that have been agreed on in the PHP 8-Working group](https://docu.ilias.de/goto_docu_copa_10290.html).

Best regards!

PR #4487, opened 2022-05-04
bottom safe area inset for mainbar on mobile 32458

catenglaender:
This PR is addressing this mantis issue: https://mantis.ilias.de/view.php?id=32458

# Issue
Some iOS devices display UI elements on top of the mainbar.

# Change
A mobile device's **bottom safe area inset is now added to the mainbar's height.** This makes room to fit UI Overlays below the mainbar buttons. Read more on user agent environment variables here: https://developer.mozilla.org/en-US/docs/Web/CSS/env

Browsers not supporting env variables should skip this line of CSS code and fall back to the previous line where a fixed height is set preventing an unwanted collapse of the mainbar. Therefore, browser compatibility should remain the same.

# Test Environment
You can see this in action on the test instance previously created for the scroll bug: https://dev.cate-tms.de/ilias7mobile/

### DIFF
--- a/src/UI/templates/default/Layout/standardpage.less
+++ b/src/UI/templates/default/Layout/standardpage.less
@@ -342,7 +342,8 @@ footer {
 				-ms-grid-columns: 1fr;
 				grid-template-columns: 1fr;
 				-ms-grid-rows: 0 @il-standard-page-small-mainbar-height;
-				grid-template-rows: 0 @il-standard-page-small-mainbar-height;
+				grid-template-rows: 0 @il-standard-page-small-mainbar-height; // for browsers not supporting env()
+				grid-template-rows: 0 @il-standard-page-small-mainbar-height-w-bottom-safe-area;
 
 				.il-mainbar-close-slates {
 					display: none;
@@ -375,7 +376,8 @@ footer {
 					-ms-grid-columns: 1fr;
 				    grid-template-columns: 1fr;
 					-ms-grid-rows: 1fr @il-standard-page-small-mainbar-height;
-				    grid-template-rows: 1fr @il-standard-page-small-mainbar-height;
+					grid-template-rows: 1fr @il-standard-page-small-mainbar-height; // for browsers not supporting env()
+					grid-template-rows: 1fr @il-standard-page-small-mainbar-height-w-bottom-safe-area;
 					width: 100%;
 					.il-mainbar {
 						.shadow-top();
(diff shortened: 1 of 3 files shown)


dsstrassner:
@catenglaender Ping me, after the Test Installation is ready. I will provide screenshots 😉

Amstutz (approved):
@catenglaender thx a lot. looking good for me. @dsstrassner , can you ping me here as soon as you have been able to test this an confirm, that this works as intended. If so, I would then merge this and pick it to trunk as well.

catenglaender:
The safe area inset for iOS can now be tested on the test instance previously created for the scroll bug: https://dev.cate-tms.de/ilias7mobile/

@dsstrassner Thank you vor volunteering to test this. It's very much appreciated 😄 On the newest iOS system you should also be able to activate a setting in Safari where the address bar is hovering at the bottom. This should also push the mainbar up, so that the address bar is not covering anything up.

dsstrassner:
Perfect @catenglaender
Dashboard:
IMG_FD7E48FB542D-1

Course: 
IMG_1BE40C1ABBBC-1

With changed Safari Setting: Tab Bar: 

IMG_47205C5E2BF3-1

But changing this setting is not the preferred way to use safari for many old school iOS Users.

catenglaender:
Thanks for the screenshot. 👍  It's great to finally see it in action!

Hm, the space above the iOS tab bar is maybe a bit too large. Unfortunately, some devices include a bit of padding with the safe area inset and some don't. But if the iOS safe area includes padding (and iOS is arguably the largest percentage of devices using thi)s, I should probably not put extra padding on top of it. 🤔

Amstutz:
Hi @catenglaender 

Thx a lot for the fix and special thx for the provided testplattform. I tested this myself on android and I can see no negative side effect. Further, I had a short chat with @dsstrassner . Everything seems to be as expected. @catenglaender if you like to improve the padding somewhat, best open a new PR. If I understand @dsstrassner, they look fine to him.

thx

jbraun-ilias:
maybe variable.less is not compiling for some of us...
the calc() produce an error.

`@il-standard-page-small-mainbar-height-w-bottom-safe-area: calc(~"@{il-standard-page-small-mainbar-height} + 1.2 * env(safe-area-inset-bottom)");`
on line 233 should be fine i think.

catenglaender:
Oh... thanks for your report!

Hmm, what's the version of your less compiler and is the math preference set to anything specific? Recent LESS builds with default settings should no longer attempt to calculate anything inside calc().

jbraun-ilias:
hmmm i have lessc 1.6.3  i think i'll have to look for an update ;-)
Thank you for your hint!

Amstutz:
@jbraun-ilias maybe best use the once shipped with ilias: 

`./node_modules/less/bin/lessc templates/default/delos.less templates/default/delos.css`

PR #4489, opened 2022-05-05
Setup: implement strict update path

klees:
Hi everyone,

just want to announce these changes to implement [strict update paths for the setup](https://docu.ilias.de/goto_docu_wiki_wpage_6911_1357.html).

The major changes are:
1. `ilSettings` will return the value for `common/ilias_version` from the database instead of ignoring the database and just returning a constant. The locations using `common/ilias_version` from `ilSettings` have been changed to directly use the constant to maintain their current behaviour.
2. The setup now blocks downgrades and updates that skip a major version, based on the version of the code from the aforementioned constant and the installed ILIAS version from `common/ilias_version`. The latter will be updated by the setup as of this PR.

The changes described in 1 could also be used to match `common/ilias_version` from the database with the version announced by the code to prevent the execution of the system when database and code don't match. Maybe this is something for `Services/Init` (@pascalseeland)

I kindly ask @alex40724 as current maintainer of `Services/Administration` to look into the change in `ilSettings`.

Best regards!

**EDIT**: The previous info that developers might manually update their db was removed. @chfsx had the aweseome idea to use the db_version as marker for the ILIAS version, so we can make sure that people only upgrade to ILIAS 8 from ILIAS 7, even if the version in the database indicates we are on ILIAS 3.4. Thanks!

### DIFF
--- a/Modules/DataCollection/classes/class.ilDataCollectionGlobalTemplate.php
+++ b/Modules/DataCollection/classes/class.ilDataCollectionGlobalTemplate.php
@@ -99,7 +99,7 @@ class ilDataCollectionGlobalTemplate implements ilGlobalTemplateInterface
         if (DEVMODE) {
             $php = ", PHP " . phpversion();
         }
-        $ftpl->setVariable("ILIAS_VERSION", $ilSetting->get("ilias_version") . $php);
+        $ftpl->setVariable("ILIAS_VERSION", ILIAS_VERSION . $php);
 
         $link_items = array();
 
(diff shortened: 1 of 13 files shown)


Amstutz:
Hi @klees 

Looked into it from a perf. perspective. One additional query is not much. However, a simple select on settings might give as some few milli secs extra since this table is still crowded (mostly with TA related stuff). 

Question:

- [x] Amount of Queries: This change will not result in more queries to the TB, since this will be part of the "common" module, being read anyways in each ILIAS request, right?

Thx for cleaning this up

alex40724 (approved):
Changes in Services/Administation are ok. Thanks.

klees:
Hi @Amstutz,

as far as I understand things, this won't result in more queries. For one, the locations where people ask for `common/ilias_version` now use a constant, so no change here. Second, `ilSetting` [reads modules (e.g. "common") at once and caches results](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Services/Administration/classes/class.ilSetting.php#L69). So yes, I cannot see more request to the DB being made.

Best regards!

chfsx (approved):
LGTM, thanks a lot @klees !

PR #4490, opened 2022-05-05
UI: Large Dropzones

chfsx:
@klees / @Amstutz we currently have defined "larger" dropzone for files, only the small one. We currently refactor some locations which uses a legacy forminput (see e.g. #4479) which could replaced with a large dropbox instead of a UI\Input.

A large Dropzone would look like this:
Bildschirmfoto 2022-05-05 um 08 34 05

### DIFF
--- a/src/UI/Component/Dropzone/File/Standard.php
+++ b/src/UI/Component/Dropzone/File/Standard.php
@@ -47,4 +47,8 @@ interface Standard extends File
      * Get the button to upload the files to the server.
      */
     public function getUploadButton() : ?Button;
+    
+    public function withLargeZone() : Standard;
+    
+    public function withSmallZone() : Standard;
 }
--- a/src/UI/Implementation/Component/Dropzone/File/Renderer.php
+++ b/src/UI/Implementation/Component/Dropzone/File/Renderer.php
@@ -79,10 +79,21 @@ class Renderer extends AbstractComponentRenderer
             $dropzone->getTitle(),
             [$dropzone->getForm()]
         );
-
-        $template = $this->getTemplate("tpl.dropzone.html", true, true);
+    
+        switch ($dropzone->getSize()) {
+            case Standard::SIZE_SMALL:
+            default:
+                $additional_class = 'ui-dropzone-small';
+                break;
+            case Standard::SIZE_LARGE:
+                $additional_class = 'ui-dropzone-large';
+        }
+        $template_file = "tpl.dropzone.html";
+        
+        $template = $this->getTemplate($template_file, true, true);
         $template->setVariable('MODAL', $default_renderer->render($modal));
         $template->setVariable('MESSAGE', $dropzone->getMessage());
+        $template->setVariable('DROPZONE_SIZE_CLASS', $additional_class);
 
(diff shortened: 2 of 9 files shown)


Amstutz:
Hi @chfsx 

Thx a lot for refactoring and locations using legacy inputs! Good stuff, we try to support this as much as possible. Thx for the examples and tests, including the rendering tests.

We have some questions:
- [x] Small vs Large:  What do you mean by "we currently have defined "larger" drop zone for files, only the small one". So far, we have not been aware that there are smaller and larger ones :-). Did we just miss that?
- [x] Use cases: Can you give some examples of use cases? When would I need this? When would I not need this? We are uncertain if we understand this atm.

- [ ] Large semantics: Large vs small is a matter of style. We try to avoid the need for clients of the UI Components to make design decision. We are fully aware, that we have not always been successful in doing so. However, cases where we failed (e.g. number of Cards per row in the Deck of Cards), tend to bite us eventually. Can we give this separation some other name that is somewhat more describing the use case, where you would use such a different type of zone? Since I do not know the cases, it is hard for me to give good examples. 
- [ ] Class via PHP: Do not set classes through PHP. I am certain, there are examples that slipped our attention before. However, use blocks inside the templates if there are different classes to be set. We try to be as transparent as possible for designers inside template classes and try to give them options to change classes. This is not possible if classes are set by variable.
 - [ ] Component Description: If possible, drop a sentence into the component's description, mentioning that this component can be set to large and small (or the improved name, if we find one). Maybe also add a rule, when to use which one. 

Please change:
 - [ ] Interface Documentation: Describe the two new Methods on the public interface by leaving some description of what they do and when to use them.
 
 Thx

chfsx:
Hi @Amstutz 
Thanks for the quick reply!

Until now, we didn't have any 'larger' Dropzones since there was no usage of them. As mentioned in #4479, we currently have to replace all usages of the legacy `ilFileStandardDropzoneInputGUI` which does no longer work after refactoring dropzone in ILIAS 8. 

In  #4479 we change the following place (as in ILIAS 7):
Bildschirmfoto 2022-05-05 um 09 30 00

In the PR we suggest to replace this with a `UI\Input\Form` with a File-input which looks like this: 
Bildschirmfoto 2022-05-05 um 09 31 58

But I think this screen would highly profit of a Dropzone without form which would look like this (without a larger variant):
Bildschirmfoto 2022-05-05 um 09 37 24 

Since this look somehow strange, a larger dropzone would be better at such screens:

Bildschirmfoto 2022-05-05 um 09 33 28

this is what this PR here suggest to add.

I will adapt your other points if there is a general "ok" that we would like to have such larger Standard-Dropzones.

Amstutz:
@chfsx and @amstutz discussed this. There seem to be the need for the two depending on the available space and how much weight the dropzones should carry on the screen. Similar to the bulky buttons vs buttons. We therefore discusssed that withBulky would be better that the "large vs small" wording. However, there is an other dependency to #4479 which we wait to be resolved before continuing our work here.

chfsx:
Since there isn't a need for a bulky dropzone, I close this PR (see #4479 )

PR #4492, opened 2022-05-05
Fix checking for url not empty

pascalseeland:
Fix check broken by php8 refactoring.


smeyer-ilias (approved):
Thanks @pascalseeland

PR #4494, opened 2022-05-05
PHPStan: Add docs and infrastructure for local configuration

mjansenDatabay:
This PR adds the infrastructure and documentation to create a local `PHPStan` configuration file to overwrite the shared configuration.

Furthermore a baseline file will be also supported to export the current issues of an analyzed component. This can be usefull to kind of suppress all errors, so new errors being reported could be spotted easier.


Amstutz:
Highly apprececiated, thx a lot. Merged by TB.

PR #4495, opened 2022-05-06
Bugfix0032606: .il_InfoScreenProperty should use @il-text-light-color

ezenzen:
Hi @all,

I fixed the color of .il_InfoScreenProperty to further streamline the ILIAS color scheme and to prevent accessibility issues.

See: https://mantis.ilias.de/view.php?id=32606

Greetings, 
Enrico


Amstutz:
thx a lot! picked to trunk as well.

PR #4496, opened 2022-05-06
Remove Personal Desktop Sidebar sorting from dashboard

iszmais:
https://mantis.ilias.de/view.php?id=32379

This PR removes leftovers of the PD sorting wich still affect the Dashboard.

As decided in https://docu.ilias.de/goto_docu_wiki_wpage_5593_1357.html all kinds of user specific sortings should be removed.

### DIFF
--- a/setup/sql/6_hotfixes.php
+++ b/setup/sql/6_hotfixes.php
@@ -1712,3 +1712,8 @@ if ( !$ilDB->tableColumnExists('cmix_settings', 'switch_to_review') ) {
     ));
 }
 ?>
+<#74>
+<?php
+$ilDB->manipulateF('DELETE FROM settings WHERE keyword = %s', ['text'], ['enable_block_moving']);
+$ilDB->manipulate('DELETE FROM il_block_setting WHERE ' . $ilDB->like('type', 'text', 'pd%'));
+?>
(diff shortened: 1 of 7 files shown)


setup/sql/6_hotfixes.php
```diff
@@ -1712,3 +1712,8 @@
     ));
 }
 ?>
+<#74>
+<?php
+$ilDB->manipulateF('DELETE FROM settings WHERE keyword = %s', ['text'], ['enable_block_moving']);
+$ilDB->manipulateF('DELETE FROM il_block_setting WHERE type LIKE %s', ['text'], ['pd%']);
+?>
```
lscharmer: Just a minor thing: Can you add a newline at the end of file so it is a proper line?

lscharmer (changes requested):

setup/sql/6_hotfixes.php
```diff
@@ -1712,3 +1712,8 @@
     ));
 }
 ?>
+<#74>
+<?php
+$ilDB->manipulateF('DELETE FROM settings WHERE keyword = %s', ['text'], ['enable_block_moving']);
+$ilDB->manipulateF('DELETE FROM il_block_setting WHERE type LIKE %s', ['text'], ['pd%']);
```
mjansenDatabay: @iszmais Please use `ilDBInterface::like` and be aware of using multile `%%`, which is required in some cases.

iszmais: Fixed the like. It should only check on the string start therefore the % is correct.

lscharmer (approved):

PR #4505, opened 2022-05-09
TA goto, fix redirect

nhaagen:
ran into a whoops from "last-visited". fixed it.

### DIFF
--- a/Modules/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php
+++ b/Modules/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php
@@ -1699,10 +1699,10 @@ class ilObjQuestionPoolGUI extends ilObjectGUI implements ilCtrlBaseClassInterfa
         if ($ilAccess->checkAccess("write", "", (int) $a_target)
             || $ilAccess->checkAccess('read', '', (int) $a_target)
         ) {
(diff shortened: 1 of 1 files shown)


Modules/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php
```diff
@@ -1694,13 +1694,15 @@ public static function _goto($a_target)
...
-            $_GET['cmd'] = 'questions';
-            $_GET['baseClass'] = 'ilRepositoryGUI';
-            $_GET["ref_id"] = $a_target;
-            include_once("ilias.php");
+        $ctrl = $DIC['ilCtrl'];
+
+        if ($ilAccess->checkAccess("write", "", (int) $a_target)
+            || $ilAccess->checkAccess('read', '', (int) $a_target)
+        ) {
+            $target_class = 'ilObjQuestionPoolGUI';
```
mjansenDatabay: @mbecker-databay / @nhaagen Just one thought: I personally would use `class-strings` (e.g.  `ilObjQuestionPoolGUI::class` here. This makes it easier for maintainers of these classes to change the dependent code if the class name changes (at least when refactoring the code with IDEs like `PHPStorm`).

PR #4511, opened 2022-05-10
UI/Substyles: fix overwriting of templates for substyles

nhaagen:
When using local/substyles, you can overwrite any html-tempalte of Modules or Services, but not from th UI-framework.
with this, you can.
eyes, @catenglaender


catenglaender:
Mantis Ticket: https://mantis.ilias.de/view.php?id=32624

PR #4512, opened 2022-05-11
TA Bugfix for 32629

dsstrassner:
Changed Button Label from 'Back' to 'Resume the Test'.

https://mantis.ilias.de/view.php?id=32629


swiniker:
Technically this is fine. Conceptually I don't see this being a big improvement, but this is not up to me to decide ;-),

kergomard (approved):

PR #4513, opened 2022-05-11
Delete Unicode-Folder from include/

kergomard:
As discussed during the PHP8-Meeting, this seems not to be needed.


kergomard:
@alex40724 : I had to remove the conditional include from `mediawiki/Title` otherwise the CI failed before running the tests. Agreed that we should still remove this?

kergomard:
As discussed in the PHP 8 on Mai 25 2022, I merge this.

PR #4514, opened 2022-05-11
Refinery: Remove test using float array key (PHP 8.1 / ILIAS 9 only)

mjansenDatabay:
In PHP 8.1 using a `float` array key results in an error (deprecation):

```bash
PHP Deprecated:  Implicit conversion from float 22.33 to int
```


klees (approved):

mjansenDatabay:
I will rebase the branch and merge if after the trunk is open for ILIAS 9.

PR #4516, opened 2022-05-11
Data/UUID: Upgrade composer dependency for PHP 8.1 (ILIAS 9 only)

mjansenDatabay:
This PR upgrades the `UUID` composer dependency used in ILIAS (`ramsey/uuid`). This does not work for PHP 7.4 (see: failed build) and MUST only be merged for ILIAS 9. The latest version of `ramsey/uuid` fixes some PHP 8.1 issues so I suggest to upgrade it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,7 +61,7 @@
 		"sabre/dav": "^4.1",
 		"symfony/console" : "^5.0",
 		"slim/slim": "^3.11",
-		"ramsey/uuid": "^3.8",
+		"ramsey/uuid": "^4.3",
 		"enshrined/svg-sanitize": "^0.15.0",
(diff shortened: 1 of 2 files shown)


klees:
Hi Jour-Fixe,

I have assigned the JF-label because we should discuss how we want to deal with dependencies that where introduced by people who are not active in the community anymore.

Best regards!

matthiaskunkel:
Jour Fixe, 16 MAY 2022 : We would like to handle external libraries like core components where a maintainer is taking care of the code. In this case, @chfsx (who uses it) will have a look at the library and notify us in case something is changing in this lib.

chfsx:
Merged in #4971

PR #4518, opened 2022-05-11
Bibliographic: Fix default value for `$limit` parameter (PHP 8.1 / ILIAS 9)

mjansenDatabay:
This PR fixes a PHP 8.1 issue with `null` used in `preg_split`. Of course it can be already merged for ILIAS 8.

See: https://www.php.net/manual/en/function.preg-split.php


chfsx:
Nice catch, thanks a lot @mjansenDatabay !

PR #4520, opened 2022-05-11
`ilStr`: Don't pass null as `$offset` parameter anymore (PHP 8.1 / ILIAS 9)

mjansenDatabay:
This PR fixes a PHP 8.1 issue with `null` being used as default value of `mb_stripos`. Of course it can be merged for ILIAS 8 as well. I did not want to adapt the consumer code and add type casts for unknown code, so I decided to add a corresponding check against `null`.

See: https://www.php.net/manual/en/function.mb-stripos.php


chfsx:
Thanks a lot!

PR #4521, opened 2022-05-11
A11Y Documents: Use correct return type declaration for `jsonSerialize` (PHP 8.1 / ILIAS 9)

mjansenDatabay:
This PR fixes a PHP 8.1 issue (deprecation) caused by a missing return value. Of course it can be already merged for ILIAS 8.


tfamula:
Thx @mjansenDatabay

PR #4526, opened 2022-05-12
UI Components, fix alt text of bulky buttons and images, see: #30308

Amstutz:
Fix of https://mantis.ilias.de/view.php?id=30308, redundant alt tags are an a11y issue. See report on mantis. 

This is intended to be pushed down to 7 as well if accepted.

Note that this could either be fixed on a client level by only setting a corresponding rule, or on directly while rendering. I decided to tackle this while rendering, since it will be hard, for all devs to a) remember setting always a tag except for b) when they have bulky buttons or links already offering a label.

### DIFF
--- a/Services/Init/classes/Dependencies/InitUIFramework.php
+++ b/Services/Init/classes/Dependencies/InitUIFramework.php
@@ -220,6 +220,14 @@ class InitUIFramework
                             $c["refinery"],
                             $c["ui.pathresolver"]
                         ),
+                        new ILIAS\UI\Implementation\Component\Symbol\Icon\IconRendererFactory(
+                            $c["ui.factory"],
+                            $c["ui.template_factory"],
+                            $c["lng"],
+                            $c["ui.javascript_binding"],
+                            $c["refinery"],
+                            $c["ui.pathresolver"]
+                        ),
                         new ILIAS\UI\Implementation\Component\Input\Field\FieldRendererFactory(
                             $c["ui.factory"],
                             $c["ui.template_factory"],
--- a/Services/User/classes/class.ilUserAvatarResolver.php
+++ b/Services/User/classes/class.ilUserAvatarResolver.php
@@ -120,7 +120,7 @@ class ilUserAvatarResolver
 
         if ($this->useUploadedFile()) {
             return $this->ui->symbol()->avatar()->picture($this->uploaded_file, $this->login)
-                            ->withAlternativeText($alternative_text);
+                            ->withLabel($alternative_text);
         }
     
         if ($this->letter_avatars_activated === false) {
@@ -130,7 +130,7 @@ class ilUserAvatarResolver
             );
         }
 
-        return $this->ui->symbol()->avatar()->letter($this->login)->withAlternativeText($alternative_text);
+        return $this->ui->symbol()->avatar()->letter($this->login)->withLabel($alternative_text);
     }
 
     public function getLegacyPictureURL() : string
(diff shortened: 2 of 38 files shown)


Amstutz:
I took the liberty of setting the jf tag myself just to make sure, that I will not forget this (contains public interface changes). @klees if you do not feel comfortable already discussing this during the next jf, feel free to remove it again.

klees (changes requested):
Hi @Amstutz,

thanks for your tenacious work on the accessibility of our system.

- [x] label-rule: I would strongly suggest to always remove the label on the `Icon` (or as I suggest later on: the `Symbol`) in the Bulky Button. I don't think that the "if the label is the same" is strong enough or even necessary. It is probably not strong enough because the `Symbols` will always have a more general notion (expressed in the label) than a button should have in an optimal world. Consider a Bulky Button in the Page Editor to insert a block. The icon on that button would have the label "block", while the button should read "insert new block". Hence, the label would not be removed and the thingy would read "block insert new block". So the "remove when label is same"-rule is too weak. Also, it is not necessary and we can always remove the label, since the label on the button should already tell what is going on and the `Symbol` really is decorational.
- [x] Symbol::withLabel 1: Consider using a context to modify the rendering instead of setting a value on the rendered thingy. In general, we would not allow anyone to set an empty label on a `Symbol`, because how could anyone create a `Symbol` without knowing a appropriate label for it. Also, I'm not sure that we really would want to have a way to change a label after creation. The label is strongly connected to the actual image, so how would I know later on what an appropriate value for the label would be? So I'm not very fond of the `withLabel` on the public interface and would prefer treating the problem on the rendering level (hence context). I would be okish with an `withLabel` on the internal interface if there are good reason for not using contexts in rendering. Context are exactly the mechanism for the problem at hand, after all: This thingy (the `Symbol`) should render differently in that context (the `Bulky Button`).

- [x] Symbol::withLabel 2: It seems as if we can safely assume, that every Symbol (Glyph, Icon and Avatar) should have a label. For the avatar it seems to be the "alternative text". So I think we should move the according methods to `Symbol` and also make the rules reference `Symbols`. The general problem would be the same for all `Symbol`s, right? The current situation regarding the label (and similar properties) is a little odd already, so lets use the occassion to clean it up.

Hope this helps, feel free to call on short notice, as always, to discuss stuff.

Best regards!

Amstutz:
Thx @klees 

I was thinking as well to go somewhat further with this change; however, I stopped due to the large footprint this immediately gets. However, your feedback is on the spot, and we need to change/align the symbol in this regard at some point. Further, this now resolves a second issue, we are dealing with: https://mantis.ilias.de/view.php?id=32169, _spans should not have aria-labels_. In Bulky Buttons, Glyphs are rendered as span and with an aria-label. It's not a huge issue, however an issue nevertheless. This also indicates me, that we really should tackle this now. 

However, since both bugs seem not to be super urgent issues, and the footprint is large (including the initialization of the UI Framework, see files changes), I highly recommend to only fix and tackle this for trunk.

Note, AFAIK the now failing 7.4 test should have nothing todo with this commit.

klees:
Hi @Amstutz,

thanks a lot for the additional effort. Currently I cannot see changes for the "label-rule"-item from my list. Is there some commit missing or do we need to discuss this?

Best regards!

Amstutz:
Hi @klees 

Thx for looking through this carefully! Indeed, I neglected to adapt the rule. This is fixed now. Note that there are exceptions to this rules specified in the rule-set of the Avatar. We did not touch the rules for the Avatar here. 

thx.

matthiaskunkel:
Jour Fixe, 16 MAY 2022: We highly appreciate this PR and accept it for trunk. PR will also solve another issue with glyphs.

Amstutz:
thx for the review!

PR #4528, opened 2022-05-13
0020916: missing function split() in /webservice/soap/lib/nusoap.php

katringross:
@smeyer-ilias I replaced deprecated split by preg_split according to fix in library https://github.com/f00b4r/nusoap/blob/e83219ee1add324124fea8e448b4cfddf782f8ff/src/nusoap.php#L3456


smeyer-ilias (approved):

smeyer-ilias:
Thanks @katringross for fixing this.

PR #4533, opened 2022-05-13
Docs: update coding style

klees:
Hi everyone,

I hereby propose to update our coding-style definition slightly:

* Update from PSR-2 to PSR-12. A lot of new PHP-stuff hasn't been adressed in PSR-2.
* Hence, remove definitions that are covered by PSR-12.
* Change return type declaration style from `method() : return_type` to `method(): return_type` to make our style match PSR-12.
* Unrelated: make `underscore_case` mandatory for variable and property names, there hasn't been any convention here until now.

If this is accepted, we'll need to adjust our tooling accordingly and most likely will have another commit to rectify the codebase. I'm confident that the TB will fund these changes. I would propose to only apply this to ILIAS 8 and future versions, since ILIAS 7 will already look vastly different then ILIAS 8 due to the PHP 8-project, hence no immediate benefit to retrofit this change.

Best regards!

### DIFF
--- a/docs/development/coding-style.md
+++ b/docs/development/coding-style.md
@@ -2,34 +2,20 @@
 
 This is the coding style for the [ILIAS project](https://github.com/ILIAS-eLearning/ILIAS).
 
-## PSR-2 and additions
+## PSR-12 and additions
 
-The ILIAS coding standard is aligned to the widely used and established [PSR-2 standard](https://www.php-fig.org/psr/psr-2/)
+The ILIAS coding standard is aligned to the widely used and established [PSR-12 standard](https://www.php-fig.org/psr/psr-12/)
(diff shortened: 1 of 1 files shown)


iszmais:
Thanks @klees for the proposal!

General approval from my side, im just a little confused about the spacing of return types. There is a rule that includes the space after the colon, but i couldnt find anything concerning the space before the colon. They seem to use this rule inside their examples but i couldnt find any rule to it.
Did i miss something here or is it maybe inherited from an older PSR standard?

Greetings,
@iszmais

matthiaskunkel:
Jour Fixe, 16 MAY 2022 : We highly appreciate this suggetion and accept the change of the guideline from PSR-2 to PSR-12, incl. the above mentioned handling of variable and property names. @klees will change the related tooling and notify of the change via dev list. In case of issues, please discuss this as dev issue in the next JF.

PR #4535, opened 2022-05-13
White Background for Delos

Amstutz:
This is a a draft PR for the "White Background" Project for ILIAS 8: https://mantis.ilias.de/view.php?id=32222

The various a11y tests and audits show did show a rather large set of contrast issues of the past. We fixed them case by case. @ezenzen along with @yvseiler and @Amstutz introduced a new color set with ILIAS 7 making it a lot easier to deal with those issues. Based on this work, we would like to go a step further and simplify the color scheme by introducing a white background for ILIAS with a reduced set of gray areas. This will help to avoid more a11y contrast issues in the future. 

Other than that, we see the following benefits:
- White is a powerful contrast with dark colors and thus strengthens accessibility. It can help to focus on certain design elements.
- White is considered a neutral color and thus increases the ease of customization.
- White feels simple, minimalist and uncluttered.

It is important to note that this change will result in adjustments elsewhere (e.g. forms, panels,...). In the weekly CSS-Squad Meetings on Tuesday afternoon 15:30, we reviewed the branch and generated a list of issues we need to work on before and/or right after merging: https://mantis.ilias.de/view.php?id=32222

This draft is available for testing thx to CaT here: https://dev.cate-tms.de/trunkwbg/ . Please contact timon.amstutz@unibe.ch if you wish to have access to this platform.

We like to present the stage of our work here, since the beta of ILIAS 8 is quickly approaching. @Amstutz recommends merging this work into core, even at this late stage and with still some open question. However, would like to get approval for this by the JF, and give fellow community member the opportunity to object to this approach. 

Note that if we merge this in this late stage, we expect to need to change some aspects of the skin during beta as well. We try to keep the footprint of those changes as small as possible.


matthiaskunkel:
Jour Fixe, 16 May 2022 : We highly appreciate this suggestion and accept the PR for trunk. Please report related issues to Mantis (CSS) so that @Amstutz can fix them during beta.

PR #4536, opened 2022-05-15
Fix a mysql error

padvincenzo:
This error happens when reloading the Ctrl structure, because `system` is a reserved word.

### DIFF
--- a/Services/Component/classes/class.ilCachedComponentData.php
+++ b/Services/Component/classes/class.ilCachedComponentData.php
@@ -138,7 +138,7 @@ class ilCachedComponentData
             $parent = $rec['parent'];
             $this->subobj_for_parent[$parent][] = $rec;
         }
-        $set = $ilDB->query('SELECT DISTINCT(id) AS sid, parent, il_object_def.* FROM il_object_def, il_object_subobj WHERE NOT (system = 1) AND NOT (sideblock = 1) AND subobj = id');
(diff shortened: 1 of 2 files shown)


mjansenDatabay:
Dear @padvincenzo ,

thanks a lot for you contribution. Unfortunately `release_5-4` is not maintained anymore, so we cannot accept further PRs for this release.

If this issue is still valid for `release_6` and/or `release_7` as well as `trunk`, we would highly appreciate a separate PR targeting to these branches.

Best regards,
Michael

chfsx:
@padvincenzo thanks for the PR, just because I'm curious: what version of MySQl/mariaDB are you using? Please note that ILIAS 5.4 is no longer supported and I won't merge the fix to this release. but if there is the same issue in newer versions of ILIAS, I'll pick the fix (or a slightly other variant) to newer versions.

padvincenzo:
Hi @mjansenDatabay and @chfsx, thanks for your replies.

The mysql version I'm currently using is 8.0.29, that is a bit too new for 5.4.

I saw that in `release_6`/`release_7`/`trunk` this problem is already solved in a different way (the file is `Services/Component/classes/Setup/class.ilObjDefReader.php`):

```php
grp," . $this->db->quoteIdentifier("system") . ",export
```

So there's no need to edit those branch. If this fix won't be merged you can close this PR.

chfsx:
@padvincenzo thanks for the feedback and for already checking newer branches.

PR #4539, opened 2022-05-16
Fix issues with access to uninitialized variables

kergomard:
Hi @smeyer-ilias 
Found a few issues with access to variables in your components.

Thanks,
@swiniker

### DIFF
--- a/Services/AccessControl/classes/class.ilAccess.php
+++ b/Services/AccessControl/classes/class.ilAccess.php
@@ -639,7 +639,7 @@ class ilAccess implements ilAccessHandler
                             ilObject::_lookupTitle($condition["trigger_obj_id"]) . " " .
                             $this->language->txt("condition_" . $condition["operator"]) . " " .
                             $condition["value"],
-                            $condition
+                            serialize($condition)
                         );
                     }
                     return false;
@@ -658,7 +658,7 @@ class ilAccess implements ilAccessHandler
                         ilObject::_lookupTitle($condition["trigger_obj_id"]) . " " .
                         $this->language->txt("condition_" . $condition["operator"]) . " " .
                         $condition["value"],
-                        $condition
+                        serialize($condition)
                     );
                 }
                 return false;
--- a/Services/Logging/classes/class.ilLog.php
+++ b/Services/Logging/classes/class.ilLog.php
@@ -59,7 +59,7 @@ class ilLog
         $this->MESSAGE = ilLogLevel::INFO;
 
         $this->default_log_level = $this->WARNING;
-        $this->current_log_level = $this->setLogLevel($a_log_level);
+        $this->current_log_level = $this->setLogLevel($a_log_level ?? $this->WARNING);
 
         $this->path = ($a_log_path) ?: ILIAS_ABSOLUTE_PATH;
         $this->filename = ($a_log_file) ?: "ilias.log";
--- a/Services/Logging/classes/class.ilLog.php
+++ b/Services/Logging/classes/class.ilLog.php
@@ -72,7 +72,7 @@ class ilLog
         $this->MESSAGE = ilLogLevel::INFO;
 
         $this->default_log_level = $this->WARNING;
-        $this->current_log_level = $this->setLogLevel($a_log_level ?? $this->WARNING);
+        $this->current_log_level = $this->setLogLevel($a_log_level ?? $this->default_log_level);
 
         $this->path = ($a_log_path) ?: ILIAS_ABSOLUTE_PATH;
         $this->filename = ($a_log_file) ?: "ilias.log";
--- a/Services/Logging/classes/class.ilLog.php
+++ b/Services/Logging/classes/class.ilLog.php
@@ -1,7 +1,20 @@
 <?php
 
-/* Copyright (c) 1998-2011 ILIAS open source, Extended GPL, see docs/LICENSE */
-
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
 
 /**
  * @deprecated
--- a/Services/User/classes/class.ilObjUser.php
+++ b/Services/User/classes/class.ilObjUser.php
@@ -590,17 +590,17 @@ class ilObjUser extends ilObject
         return $fullname;
     }
 
-    public static function _lookupEmail(int $a_user_id) : string
+    public static function _lookupEmail(int $a_user_id) : ?string
     {
         return self::_lookup($a_user_id, "email");
     }
     
-    public static function _lookupGender(int $a_user_id) : string
+    public static function _lookupGender(int $a_user_id) : ?string
     {
         return (string) self::_lookup($a_user_id, "gender");
     }
 
-    public static function _lookupClientIP(int $a_user_id) : string
+    public static function _lookupClientIP(int $a_user_id) : ?string
     {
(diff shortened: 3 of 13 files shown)


Services/User/classes/class.ilObjUserFolderGUI.php
```diff
@@ -811,10 +811,11 @@ public function confirmdeleteObject() : void
             );
         }
 
+        $obj_factory = new ilObjectFactory();
         // FOR ALL SELECTED OBJECTS
         foreach ($ids as $id) {
             // instatiate correct object class (usr)
-            $obj = &$this->ilias->obj_factory->getInstanceByObjId($id);
+            $obj = $obj_factory->getInstanceByObjId($id);
```
smeyer-ilias: Please can you replace it with ilObjectFactory::getInstanceByObjId()

Services/User/classes/class.ilObjUser.php
```diff
@@ -590,17 +590,17 @@ public static function _lookupFullname(int $a_user_id) : string
         return $fullname;
     }
 
-    public static function _lookupEmail(int $a_user_id) : string
+    public static function _lookupEmail(int $a_user_id) : ?string
```
smeyer-ilias, line 593: Changing the return type of this methods means that all calls of this method must also be checked/adapated.

Services/User/classes/class.ilObjUser.php
```diff
@@ -590,17 +590,17 @@ public static function _lookupFullname(int $a_user_id) : string
...
     {
         return self::_lookup($a_user_id, "email");
     }
     
-    public static function _lookupGender(int $a_user_id) : string
+    public static function _lookupGender(int $a_user_id) : ?string
```
smeyer-ilias, line 598: Changing the return type of this methods means that all calls of this method must also be checked/adapated.

Services/User/classes/class.ilObjUser.php
```diff
@@ -590,17 +590,17 @@ public static function _lookupFullname(int $a_user_id) : string
...
     {
         return (string) self::_lookup($a_user_id, "gender");
     }
 
-    public static function _lookupClientIP(int $a_user_id) : string
+    public static function _lookupClientIP(int $a_user_id) : ?string
```
smeyer-ilias, line 603: Changing the return type of this methods means that all calls of this method must also be checked/adapted.

Services/User/classes/class.ilObjUser.php
```diff
@@ -655,12 +655,12 @@ public static function _lookupFields(int $a_user_id) : array // Missing array ty
         return $user_rec;
     }
 
-    public static function _lookupLogin(int $a_user_id) : string
+    public static function _lookupLogin(int $a_user_id) : ?string
```
smeyer-ilias, line 658: Changing the return type of this methods means that all calls of this method must also be checked/adapted.

Services/User/classes/class.ilObjUser.php
```diff
@@ -655,12 +655,12 @@ public static function _lookupFields(int $a_user_id) : array // Missing array ty
...
     {
         return (string) self::_lookup($a_user_id, "login");
     }
 
-    public static function _lookupExternalAccount(int $a_user_id) : string
+    public static function _lookupExternalAccount(int $a_user_id) : ?string
```
smeyer-ilias, line 663: Changing the return type of this methods means that all calls of this method must also be checked/adapted.

Services/User/classes/class.ilObjUser.php
```diff
@@ -703,12 +703,12 @@ public static function _lookupId($a_user_str)
         return $ids;
     }
 
-    public static function _lookupLastLogin(int $a_user_id) : string
+    public static function _lookupLastLogin(int $a_user_id) : ?string
```
smeyer-ilias, line 706: Changing the return type of this methods means that all calls of this method must also be checked/adapted.

Services/User/classes/class.ilObjUser.php
```diff
@@ -703,12 +703,12 @@ public static function _lookupId($a_user_str)
...
     {
         return self::_lookup($a_user_id, "last_login");
     }
 
-    public static function _lookupFirstLogin(int $a_user_id) : string
+    public static function _lookupFirstLogin(int $a_user_id) : ?string
```
smeyer-ilias, line 711: Changing the return type of this methods means that all calls of this method must also be checked/adapted.

Services/Logging/classes/class.ilLog.php
```diff
@@ -59,7 +72,7 @@ public function __construct(
         $this->MESSAGE = ilLogLevel::INFO;
 
         $this->default_log_level = $this->WARNING;
-        $this->current_log_level = $this->setLogLevel($a_log_level);
+        $this->current_log_level = $this->setLogLevel($a_log_level ?? $this->WARNING);
```
smeyer-ilias: Please can you replace it with $this->setLogLevel($a_log_level ?? $this->default_log_level)

Services/AccessControl/classes/class.ilAccess.php
```diff
@@ -639,7 +639,7 @@ public function doConditionCheck(
                             ilObject::_lookupTitle($condition["trigger_obj_id"]) . " " .
                             $this->language->txt("condition_" . $condition["operator"]) . " " .
                             $condition["value"],
-                            $condition
+                            serialize($condition)
```
smeyer-ilias, line 642: Are you sure, that serialization is correct here? I cannot find a unserialize.

smeyer-ilias (changes requested):
Thanks for fixing these issues. 
I added @pascalseeland for the review of LDAP and ECS issues.

Services/AccessControl/classes/class.ilAccess.php
kergomard: No, I'm not, but you are throwing an array at a function that wants a string, so serialization was the poor man's choice, but you can decide whatever you want to pass here.

Services/User/classes/class.ilObjUser.php
kergomard: Yes, I know (and this goes for all your following comments). But:
* You are calling _lookup($a_user_id) from all these functions and there the return type is ?string. So it seems to me that the problem we have right now, ie. that a field can also be empty, is possible in all these cases. So we should either provide for this case by catching null and always returning a value or by allowing null. Which one it should be is not a decision I can make for you. I just chose one. We could alternatively add a null coalescing operator in all these functions. If you didn't change the basic structure (and I didn't check for this) these functions already might have returned null, thus all callers should already be aware of this possibility.

Services/Logging/classes/class.ilLog.php
kergomard: Done.

Services/User/classes/class.ilObjUserFolderGUI.php
kergomard: Done

kergomard:
You can find an alternative PR with Null-Coalescing in #4596

Services/User/classes/class.ilObjUser.php
smeyer-ilias: I do not agree. If you change the return type of a method, you have to check all callers of this method, since it is a breaking change. One example: 
lookupEmail() is used in ilBadgeHandler.php which uses the return value as first parameter in ilMail->enqueue() which expects a string. 

I would prefer a type cast to string, since distinguishing between null and empty strings is not necessary here, imo.

Can you please revert the return type changes and add type casts to the return statements? Otherwise i will fix this in the trunk.

smeyer-ilias (changes requested):

Services/User/classes/class.ilObjUser.php
kergomard: You have seen that there is another option for this PR below, right? There I use null coalescing. For your convenience, here is the link: https://github.com/ILIAS-eLearning/ILIAS/pull/4596

smeyer-ilias: No, sorry, haven't seen it. 
Perfect! So I can close this one?

kergomard: Yes, please do.

smeyer-ilias:
see https://github.com/ILIAS-eLearning/ILIAS/pull/4596

PR #4540, opened 2022-05-16
Fix one case where an access to an undefined variable can occure

kergomard:
Hi @mjansenDatabay 
Found one place where an access to an undefined array key can occur in the notification service.

Thanks,
@swiniker


mjansenDatabay:
Thanks a lot @swiniker 

FYI @iszmais

PR #4541, opened 2022-05-16
Pass over rendered HTML to use in an uihook plugin for replacements (DashboardGUI)

PurHur:
Hello,

i added the html to the $a_pars array so this UIHook call behaves like the other `UIHook->getHtml` calls in `ilTemplate`.

The problem is when you make an UIHookPlugin that wants to replace something in the rendered html the whole page will be blank, because there is no way you can not completely replace the dashboard html. 

The impact of this PR will be quite low. The UIHooks are always called, it justs adds a `new` parameter. As an uihook plugin developer i would assume this parameter is always filled with the rendered html in the first place, like its done here in the general ilTemplate hook: https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/Services/UICore/classes/class.ilTemplate.php#L180

Greetings
Purhur

### DIFF
--- a/Services/Dashboard/classes/class.ilDashboardGUI.php
+++ b/Services/Dashboard/classes/class.ilDashboardGUI.php
@@ -494,14 +494,15 @@ class ilDashboardGUI
                 
                 // user interface plugin slot + default rendering
                 include_once("./Services/UIComponent/classes/class.ilUIHookProcessor.php");
+                $html = $ilCtrl->getHTML($column_gui);
                 $uip = new ilUIHookProcessor(
                     "Services/Dashboard",
                     "left_column",
-                    array("personal_desktop_gui" => $this)
+                    array(
+                        "personal_desktop_gui" => $this,
+                        "html" => $html
+                    )
                 );
-                if (!$uip->replaced()) {
-                    $html = $ilCtrl->getHTML($column_gui);
-                }
                 $html = $uip->getHTML($html);
             }
         }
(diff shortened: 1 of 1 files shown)


PurHur:
@iszmais ping because i think you are the maintainer here.

PurHur:
The question here is if the getHtml method of UIHook plugins is really deprecated or not. Is there a link to this information in the wiki? I cant find it.

iszmais (approved):
Thx for the contribution @PurHur !
As far as i can see this isnt a bug since this UI Call shouldnt provide the HTML (there is a later one that does).
But since this whole stucture got a little messy over time i will merge this as an improvement so you can access the rendered HTML on both ways.
Just be aware that this can produce a conflict with other plugins which may cause the changes made in your plugin to not be rendered.

**I strongly recommend to use the function exchangeUIRendererAfterInitialization() inside your plugin object to prevent those kinds of conflicts!**

Greetings,
@iszmais

iszmais (changes requested):
1 Minor Change request:
Please squash your commits and set a commit message wich also hint onto changes inside the dashboard so its easier to differ these commit from the others in future

iszmais:
@amichelsdatabay

PurHur:
See https://github.com/ILIAS-eLearning/ILIAS/pull/4544#issuecomment-1130164999

Thanks for your documentation for the new system so plugin devs have a change to know where to migrate.

PR #4542, opened 2022-05-16
UIComponent/Explorer2: late init JS

nhaagen:
fixes "il not initialized"- JS error when using ilRepositorySelector2InputGUI


alex40724:
Thanks for the PR!

PR #4544, opened 2022-05-16
Pass over rendered HTML to use in an uihook plugin for replacements (ilGlobalTemplate)

PurHur:
Hello,

i added the html to the $a_pars array so this UIHook call behaves like the other UIHook->getHtml calls in ilTemplate.

The problem is when you make an UIHookPlugin that wants to replace something in the rendered html the whole partial will be blank, because there is no rendered html presend like in the other hook calls of getHtml. 

The impact of this PR will be quite low. The UIHooks are always called, it justs adds a new parameter. As an uihook plugin developer i would assume this parameter is always filled with the rendered html in the first place, like its done here in the general ilTemplate hook: https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/Services/UICore/classes/class.ilTemplate.php#L180

Note: You can close this PR if the getHtml is deprecated and should not be changed?

Greetings
Purhur

### DIFF
--- a/Services/UICore/classes/class.ilGlobalTemplate.php
+++ b/Services/UICore/classes/class.ilGlobalTemplate.php
@@ -1048,14 +1048,16 @@ class ilGlobalTemplate implements ilGlobalTemplateInterface
         $html = "";
         if (is_object($ilPluginAdmin)) {
             include_once("./Services/UIComponent/classes/class.ilUIHookProcessor.php");
+
+            $html = $ilLocator->getHTML();
             $uip = new ilUIHookProcessor(
                 "Services/Locator",
                 "main_locator",
-                array("locator_gui" => $ilLocator)
+                array(
+                    "locator_gui" => $ilLocator,
+                    "html" => $html
+                )
             );
-            if (!$uip->replaced()) {
-                $html = $ilLocator->getHTML();
-            }
             $html = $uip->getHTML($html);
         } else {
             $html = $ilLocator->getHTML();


thibsy:
Hi @PurHur 

I closed this PR because the mechanism you try to reinstantiate is deprecated and would allow an easy way to still use it. 
The segment you've referenced in `ilTemplate` was simply overlooked (I guess) and I'll remove this too in an upcoming commit. 

Thank you anyways, best regards!

thibsy:
Hi @PurHur 

I closed this PR because the mechanism you try to reinstantiate is deprecated and would allow an easy way to still use it. 
The segment you've referenced in `ilTemplate` was simply overlooked (I guess) and I'll remove this too in an upcoming commit. 

Thank you anyways, best regards!

PurHur:
Can you point me to the wiki where it was deprecated? If you remove the behaviour of ilTemplate i have to change a bunch of plugins. A lot... And i dont even know if there are renderer for all the things i want to change.

thibsy:
@PurHur I couldn't point you to a wiki entry, I myself just relied on the deprecation notice within `ilUIHookPluginGUI`. I guess I could hold back on the ilTemplate change for a while, but it'll be removed eventually.

PurHur:
I think this should be removed in Ilias8 so plugin devs dont fall into the error encountered here. Its not clear why it isnt working. Ingmar made a PR to properly change the HTML so people can migrate. 

See: https://github.com/ILIAS-eLearning/ILIAS/pull/4189
and https://github.com/ILIAS-eLearning/ILIAS/pull/4568

So if this method is deprecated since 2019 it should be removed if there is another concept for that.

PR #4552, opened 2022-05-16
Refinery/ReadingTimeTrafo: Handle unsupported entities when parsing X(HTML) documents

mjansenDatabay:
This PR addresses an issue with unsupported entities when parsing (X)HTML documents in the `EstimatedReadingTime` transformation of the `Refinery`.

See: https://ourcodeworld.com/articles/read/1375/how-to-solve-php-7-exception-warning-domdocument-loadhtml-tag-figure-nav-section-invalid

Mantis Issue: https://mantis.ilias.de/view.php?id=32682

### DIFF
--- a/src/Refinery/String/EstimatedReadingTime.php
+++ b/src/Refinery/String/EstimatedReadingTime.php
@@ -27,14 +27,20 @@ use DOMDocument;
 use DOMText;
 use DOMCdataSection;
 use DOMXPath;
+use LibXMLError;
 
 class EstimatedReadingTime implements Transformation
 {
+    private const LIBXML_CODE_HTML_UNKNOWN_TAG = 801;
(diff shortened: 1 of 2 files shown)


klees (commented):
Hi @mjansenDatabay,

LGTM in general, kudos for restoring the state of libxml after the operation =)

I would suggest to remove the check (preg_match) on the problematic entities and instead always handle errors in the supplied content with the approach used here. For one: the list of entities that are problematic will grow and we will be busy to answer to tickets add add new tags. Second, libxml is fairly strict anyway, AFAIK, so I would expect that we will see other problems when working on the content. Nobody uses strict XHTML anymore... Finally: The code will be much simpler.

That's it =)

Best regards!

PR #4555, opened 2022-05-17
Fixes #29125

mhilbert:
Handling of gender field:
sets 'n' to default (instead of 'f') and adds support of diverse gender.

### DIFF
--- a/Services/LDAP/classes/class.ilLDAPAttributeToUser.php
+++ b/Services/LDAP/classes/class.ilLDAPAttributeToUser.php
@@ -290,11 +290,6 @@ class ilLDAPAttributeToUser
                 switch ($field) {
                     case 'gender':
                         switch (strtolower($value)) {
-                            case 'n':
-                            case 'neutral':
-                                $this->writer->xmlElement('Gender', array(), 'n');
-                                break;
-
                             case 'm':
                             case 'male':
                                 $this->writer->xmlElement('Gender', array(), 'm');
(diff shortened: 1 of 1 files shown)


jochen-ac:
This solution also precisely maps our needs. It would be great if this could also be adopted for ILIAS 6.

Services/LDAP/classes/class.ilLDAPAttributeToUser.php
```diff
@@ -290,21 +290,23 @@ private function usersToXML()
...
                             case 'f':
                             case 'female':
-                            default:
                                 $this->writer->xmlElement('Gender', array(), 'f');
                                 break;
+
+                            case 'n':
+                            case 'neutral':
+                            case 'd':
+                            case 'diverse':
```
mjansenDatabay, line 306: IMO we should not list `d` an `diverse` explicitly, just because there are other locations where we have to introduce these cases as well. IMO we should just use the `default` to handle other values.

mhilbert:
I agree with Michael. Having `n`|`neutral` as `default` would be sufficient.

jochen-ac:
I also see it that way. As far as I know, the form of salutation would not differ in these cases.

mjansenDatabay:
Hi @pascalseeland
,

Best regards!

pascalseeland:
Superseed by be32359, which maps all unknown gender to n

PR #4556, opened 2022-05-17
Fix #32226: added hotfix for vanishing favourites main-menu item.

thibsy:
If the main menu wasn't configured between https://github.com/ILIAS-eLearning/ILIAS/commit/28962f2945712d231c94cb322f0e39e3ec9b6e13 and now, the hotfix updates the old identification to the new one which keeps all configurations intact. If adjustments were in between, we cannot tell whether the user has adjusted the newly generated favourites entry, therefore I just delete the entry for the old identification in this scenario.


chfsx (approved):
I removed some empty line changes, thx a lot!

PR #4560, opened 2022-05-17
[FIX] 0032241: Cookies for each video might lead to HTTP 431

chfsx:
@alex40724 this fixes the issue in https://mantis.ilias.de/view.php?id=32241 by using URL-tokens instead of Cookie-Tokens.


alex40724:
Thanks for the PR!

PR #4561, opened 2022-05-17
Add Heading for Actions Column in Pax-Table

kergomard:
One possible fix for: https://mantis.ilias.de/view.php?id=32697

Let me know if another naming would be better.


Amstutz:
Hi, 

Thx a lot @swiniker . The naming seems fine. Not that there are variables you can use for padding and margin. However, this legacy component does not make use of them atm. So IMO this is fine.

@smeyer-ilias : I merge this, hope this is fine. If I understood correctly, its fine if I directly push simple a11y fixes. Further, for this kind of fix in tables, we have a JF decision, here: https://mantis.ilias.de/view.php?id=32259 . Note that there will be certainly more of those.

Amstutz:
Thx, merged to trunk and picked to 7.

PR #4563, opened 2022-05-18
Fix 28308: Bibliographic link resolver

thibsy:
This PR solves an issue where link-targets were generated with the wrong attribute-names as mentioned in https://mantis.ilias.de/view.php?id=28308. The fix was quite simple because the mapping of the different attribute-names was already implemented but the comparison of the `$type` variable was wrong, since an `ilBiblTypeInterface` object is passed rather than a string.


chfsx:
Thanks a lot!

PR #4567, opened 2022-05-18
UI: allow null as false for checkbox input (#29645)

klees:
Hi everybody,

this fixes https://mantis.ilias.de/view.php?id=29645. Rather simple, small behaviour improvement of the checkbox for developers.

Best regards!


Amstutz:
Thx a lot, merged to 7 as well.

PR #4568, opened 2022-05-18
Added documentation for ui plugin exchange methods

iszmais:
This adds some documentation for the correct use of UI manipulations in UIHookPlugins.

### DIFF
--- a/docs/development/code-examples/ui-exchange/00_base_classes/ExamplePlugin.php
+++ b/docs/development/code-examples/ui-exchange/00_base_classes/ExamplePlugin.php
@@ -1,0 +1,28 @@
+<?php
+
+use ILIAS\DI\Container;
+
+class ExamplePlugin extends ilUserInterfaceHookPlugin
+{
+    public function exchangeUIRendererAfterInitialization(Container $dic): Closure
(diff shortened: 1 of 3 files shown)


iszmais:
@klees @Amstutz This is my first apporach to document the made changes of https://github.com/ILIAS-eLearning/ILIAS/pull/4189

Feel free to leave some feedabck if soemthing is wrong or if im missing something out.

Greetings,
@iszmais

Amstutz:
Hi @iszmais 

Thx a lot for your contribution. This will help various developers. Note that we believe the UI Hooks Plugin dev to be a pattern we do want to get a rid of eventually. The docu and examples are very detailed. This is helpful; however, it gives this pattern maybe a much larger weight than it should have. We therefore might want to shrink this or remove completely in upcoming releases.

Nevertheless, for the moment, this is certainly very helpful for many devs.

- [x] Other Releases: Can you provide the PR for 7 and trunk as well. In this case, I do not simply want to cherry pick, since we should really make sure that all of the content fits and works for both other releases as well.

Kudos and thx!

Kindly 
@Amstutz

PR #4569, opened 2022-05-18
Fix #28229: Notification aggregates aren't properly removed

thibsy:
Hi @Amstutz 

I've opened a PR that solves the first issue that occurred when reproducing https://mantis.ilias.de/view.php?id=28229 (where I uncovered your method-of-doom =)) that lead to an invalid object being passed to `UI.counter.getCounterObject()`. 

Now to recap the issue briefly: when aggregated notifications are displayed in the notification center, the notification.js should remove the "parent" slate too if it was the last child. Though I couldn't quite figure out how to achieve this and hoped you could point me in the right direction. As you can see in `removeNotificationItem()` there are already a few debug-statements, because I figured the handling of said problem should be done there (right?).

Best regards!

### DIFF
--- a/src/UI/templates/js/Item/notification.js
+++ b/src/UI/templates/js/Item/notification.js
@@ -425,13 +425,17 @@ il.UI.item = il.UI.item || {};
 			 * @param $close_button
 			 */
 			var removeNotificationItem = function () {
-				if(!hasSibblings()){
-					getParentSlateOfItem().hide();
-					if(isAggregate()) {
-                        getParentSlateOfItem().show().siblings().show();
-					}
-				}
-				$item.children().remove();
+				console.log("item: ", $item);
+				console.log("parents children: ", getParentSlateOfItem().children());
+				console.log("aggregates: ", getAggregatesOfItem());
+
+				// if(!hasSibblings()){
+				// 	getParentSlateOfItem().hide();
+				// 	if(isAggregate()) {
+                //         getParentSlateOfItem().show().siblings().show();
+				// 	}
+				// }
+				// $item.children().remove();
 			};
 
 			/**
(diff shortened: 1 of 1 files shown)


src/UI/templates/js/Item/notification.js
```diff
@@ -503,7 +507,7 @@ il.UI.item = il.UI.item || {};
 			var getNotificationsTriggererIfAny = function(){
 				var $meta_bar = getMetaBarOfItemIfIsInOne();
 				if($meta_bar.length){
-					var $notification_glyph = $meta_bar.find('.il-maincontrols-metabar > .btn-bulky .glyphicon-bell');
+					var $notification_glyph = $meta_bar.find('.btn-bulky .glyphicon-bell');
```
Amstutz, line 510: This change looks wrong an dangerous. This would detect any .glyphicon-bell as notification trigger. This might break if we would use this glyph in any place other than the meta bar. Better just add the missing li; 
.il-maincontrols-metabar > li > .btn-bulky .glyphicon-bell

Amstutz:
HI @thibsy 

Thx for this PR. Strange the part in the DOOM method I believed to be resolved with 7. However, see my inline comment, this should not fixed the way you are proposing. 

Concerning your question. If you believe something to be wrong inside the UI Component Notification Item, best pass me a ticket. However, atm I not sure this is the case (other than the missing li in the selector you mentioned). The you try just using replaceContentByAsyncItemContent, as documented here (sadly the demo has the same issue as you pointed out above)... See: https://test7.ilias.de/goto.php?target=stys_21_MainControlsMetaBarMetaBar_default_delos&client_id=test7

src/UI/templates/js/Item/notification.js
thibsy: That's true, though the selector `.il-maincontrols-metabar > li > .btn-bulky .glyphicon-bell` doesn't quite do the trick.

Amstutz:
We wil discuss this tomorrow Firday 10:30 via discord. If anybody else wants to join in, contact me.

thibsy:
As discussed with @Amstutz I've opened a ticket (https://mantis.ilias.de/view.php?id=32722) because further public functions are necessary in order to solve the previously mentioned issue im fixing.

PR #4580, opened 2022-05-20
Background Task: Update Nofication Item description on close

Amstutz:
Follow up of https://github.com/ILIAS-eLearning/ILIAS/pull/4569

See:  https://mantis.ilias.de/view.php?id=28229

According to the discussion with @thibsy . 

@thibsy ; thx a lot for the exchange. Always nice collaborating with you. 

Note: Best shortly re-test. I am not 100% sure, to have thought about all bg-task scenarios. This should also work with trunk, so you could just cherry pick.

### DIFF
--- a/Services/BackgroundTasks/classes/class.ilBTPopOverGUI.php
+++ b/Services/BackgroundTasks/classes/class.ilBTPopOverGUI.php
@@ -87,7 +87,7 @@ class ilBTPopOverGUI
             $actions = $this->getUserInteractionContent($observer, $redirect_uri);
             $primary_action = array_pop($actions);
             if ($primary_action instanceof Button) {
-                $title = $primary_action->withLabel($title);
+                $item = $item->withTitle($primary_action->withLabel($title));
             }
 
 
--- a/Services/BackgroundTasks/classes/class.ilBTPopOverGUI.php
+++ b/Services/BackgroundTasks/classes/class.ilBTPopOverGUI.php
@@ -76,20 +76,18 @@ class ilBTPopOverGUI
         $icon = $f->symbol()->icon()->standard("bgtk", $this->txt("bg_task"));
         $title = $observer->getTitle() . ($state === State::SCHEDULED ? " ({$this->txt('scheduled')})" : "");
 
-        $item = $f->item()->notification($title, $icon);
 
         $description_text = $this->txt('background_tasks');
-        $item = $item->withAdditionalOnLoadCode(function ($id) use ($description_text) {
-            return "il.BGTask.updateDescriptionOnClose('#$id','$description_text');";
-        });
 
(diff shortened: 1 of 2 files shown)


Services/BackgroundTasks/classes/class.ilBTPopOverGUI.php
```diff
@@ -76,13 +76,20 @@ public function getItemForObserver(Bucket $observer) : ILIAS\UI\Component\Item\N
...
+            return "il.BGTask.updateDescriptionOnClose('#$id','$description_text');";
+        });
+
         if ($state === State::USER_INTERACTION) {
             $actions = $this->getUserInteractionContent($observer, $redirect_uri);
             $primary_action = array_pop($actions);
             if ($primary_action instanceof Button) {
                 $title = $primary_action->withLabel($title);
             }
-            $item = $f->item()->notification($title, $icon);
```
thibsy, line 85: We can't remove this because $title might be adjusted in this scenario. It'd probably be best to not return the $item within this if-branch and add the closure that updates the description on close right before the last return statement.

thibsy (changes requested):

thibsy:
Thanks for looking into my issue as well @Amstutz, always a pleasure collaborating with you too :).

There's one little hiccup that needs to be fixed, but otherwise it's all good and should work with all background-task notifications. 

Kind regards!

Services/BackgroundTasks/classes/class.ilBTPopOverGUI.php
Amstutz: Good catch.

Services/BackgroundTasks/classes/class.ilBTPopOverGUI.php
```diff
@@ -87,7 +87,7 @@ public function getItemForObserver(Bucket $observer) : ILIAS\UI\Component\Item\N
             $actions = $this->getUserInteractionContent($observer, $redirect_uri);
             $primary_action = array_pop($actions);
             if ($primary_action instanceof Button) {
-                $title = $primary_action->withLabel($title);
+                $item = $item->withTitle($primary_action->withLabel($title));
```
thibsy: `withTitle()` is even better :), LGTM now!

thibsy (approved):

Amstutz:
@thibsy : thx for carefully checking this. Good catch. I shuffled things around somewhat. withTitle is not available on the public interface, so try II was no good. "Aller guten Dinge sind drei" :-)

thibsy:
@Amstutz yes that was a mean one, due to `withAdditionalOnLoadCode` PHPStorm suggests this method there, even though it doesn't exist on `Notification` :). 

Thx for the PR and the fix, I'll merge it now :)

PR #4581, opened 2022-05-20
Allow DateTimeImmutables as Value for DateTime Fields

klees:
Hoi zsämme,

currently the `DateTime Input Field` will only allow strings as input, which is irritating:

https://mantis.ilias.de/view.php?id=29267

This will allow to pass `DateTimeImmutable` as value as well. Also, this cleans up some stuff and adds a simplification for the `DateFormat`.

Best regards!

### DIFF
--- a/src/Data/DateFormat/DateFormat.php
+++ b/src/Data/DateFormat/DateFormat.php
@@ -1,6 +1,20 @@
 <?php declare(strict_types=1);
 
-/* Copyright (c) 2019 Nils Haagen <nils.haagen@concepts-and-training.de> Extended GPL, see docs/LICENSE */
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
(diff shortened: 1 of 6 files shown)


Amstutz (approved):
Thx a lot, the UI part seems fine to me. Feel free to merge if also good from a Data perspective.

klees:
@nhaagen Could you have a look into the stuff in `src/Data`?

src/Data/DateFormat/DateFormat.php
```diff
@@ -78,4 +92,14 @@ public function toString() : string
     {
         return implode('', $this->format);
     }
+
+    public function __toString() : string
+    {
+        return $this->toString();
+    }
+
+    public function format(\DateTimeImmutable $datetime) : string
```
nhaagen: This would be $format->format(), and is sort of the opposite direction of php native function.
Maybe we could "apply" it to a $datetime? applyTo?

nhaagen (approved):
Hi Richard, 
thanks a lot, this will ease using the input quite a bit. Except for a naming-suggestions - fine with me.

src/Data/DateFormat/DateFormat.php
klees: Love it =)

klees: Done...

PR #4582, opened 2022-05-20
Less compile bug

rfalkenstein:
This should solve the compiling error described in: https://mantis.ilias.de/view.php?id=32725


Amstutz:
Thx for the PR. Note that this is less version dependent. Make sure to use a 4.1.X version one, also see: https://github.com/ILIAS-eLearning/ILIAS/tree/release_7/templates#css-guideline

If you run less through ssh, you can also just use the one that is shipped with ILIAS (https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/node_modules/less/bin/lessc), by running from the ILIAS root:
./node_modules/less/bin/lessc path_to_less_you_with_to_compile path_to_css_output

PR #4583, opened 2022-05-20
T&A, JS inits for ClozeQuestions

nhaagen:
There were some JS-errors in the ClozeQuestion-GUIs, mainly due to wrong order inits.


mbecker-databay:
Thank you so much!

PR #4587, opened 2022-05-23
T&A, refining "suggested solutions" / GUI: "Content for Recapitulation"

nhaagen:
Dear fellow Devs and partners in crime,
originally, I wanted to convert the whole question-loading and storage into using a repository patten, but things are quite heavily entangled as you know. So as result of these endeavors, and to keep it manageble, this PR is about SuggestedSolutions.

First, SuggestedSolutions are typed classes now (as opposed to an array) and secondly, they are way more separated than before - as actually, they are not that specific at all and should be able to live without the test entirely.

However, there is some strategy involved, as I embarked with way bigger goals and had to shorten down a little.
A lot of the current pain in the T&A is due to the ruthless mixing of Logic an GUI, another one is making decisions very late, based on variables transported all the way through the GUIs.
With the assQuestionGUI and its specific derivates, it is even worse. There is $this->object at the GUI (which is a full-fletched question), and this->object is being knocked upon by the base-gui as well as the question-specifc guis and some other hidden functions. 
We will not disentagle all of it in one smooth stride, but maybe deviding things apart will subsequentlly identify and isolate the moving parts. So here is a first attempt in this direction. Again, I had to stop somewhere (in this case with resolving internal links and mediaobjects), but moving further from here on should be easier than before.

Next things for me would be the devision of "base-settings" - common properties all questions have - and specific settings, that are relevant for a certain type of question. There already is a mechanism like that with the 'getAdditionalTableName',
but then this is used to generalize funtions where it explicitly is meant ot be specific.
As hopefully shown in this PR, repo-pattern will lead to move responisibility and code into more isolated domains.

So, please kindly consider this PR, and I'll be very happy about comments and remarks.
Nils

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assFlashQuestion.php
+++ b/Modules/TestQuestionPool/classes/class.assFlashQuestion.php
@@ -165,7 +165,7 @@ class assFlashQuestion extends assQuestion implements ilObjQuestionScoringAdjust
             $this->setNrOfTries($data['nr_of_tries']);
             $this->setTitle((string) $data["title"]);
             $this->setComment((string) $data["description"]);
-            $this->setSuggestedSolution($data["solution_hint"]);
+            //$this->setSuggestedSolution($data["solution_hint"]);
             $this->setOriginalId($data["original_id"]);
             $this->setObjId($data["obj_fi"]);
             $this->setAuthor($data["author"]);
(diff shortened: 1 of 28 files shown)


klees:
Only merge when release_8-branch was created, I guess. We will present this either in the T&A Tech Squad or in an dedicated meeting, depending on available time and alternative topics.

dsstrassner:
There is one question, which I forget to ask at our TechSquad Meeting today. Why is the PHP class named "suggested solution" but in the GUI (Editing Questions) the section is called "Content for Recapitulation"?

dsstrassner:
Hi @nhaagen 

please see my last comment. Do you think, it is worthy to change the naming to match the GUI? 

Also, as discussed in today's TechSquad Meeting, we need to remove the dust a little, and then we are ready to merge this improvement for Trunk/ILIAS9.

PR #4589, opened 2022-05-23
Container: make the GUI base class respect that sub object does not support page editor

klees:
Hi @alex40724,

this fixes [Mantis 30959](https://mantis.ilias.de/view.php?id=30959). Unfortunatly I could not find another way to make the `ContainerGUI` respect the fact, that the Page Editor is not supported by the Study Programme without copying the method from the class verbatim. If you have a prefered way to implement this or if I have just missed some way to fix the issue, please let me know.

Best regards!


alex40724:
Thanks for the PR!

PR #4591, opened 2022-05-24
#32512: Added a sorting checkup and added the check to the workflow

catlauraherzog:
The actual script workes fine and like intended.

Regarding https://mantis.ilias.de/view.php?id=32512 I've added a check if the language file needs sorting. If so the commit will be rejected by the pre-commit-hook. The check is enabled by just adding a keyword to the execution of the script: `./CI/sort_langfile_entries.sh check`. If you don't pass the `check` argument the script will sort and modify the language files.


klees:
@mjansenDatabay Would you do the honour and look into this?

mjansenDatabay:
@klees I'll try to provide feedback this week.

-- UPDATE --

DONE :-)

CI/sort_langfile_entries.sh
```diff
@@ -21,9 +32,26 @@ do
         fi
     done
 
-    echo  "sorting entries in ${file}";
+    #echo  "sorting entries in ${file}";
```
mjansenDatabay: We don't need to keep this line, we have the GIT history :-).

mjansenDatabay (approved):
Thx @lauraherzog / @klees, I only had one remark.

I also tested the modified `CaptainHook` action. It worked like a charm.

CI/sort_langfile_entries.sh
catlauraherzog: Fixed.

matthiaskunkel:
Jour Fixe, 27 JUN 2022 : The current PR does not solve the reported problem (different sorting depending from OS). It will only mark a commit with an error message when the sorting is different than expected. We suggest to use PHP on the local computer for sorting with LOCALE EN to get an identical order of language entries independent from the used operation system.

mjansenDatabay:
Hi @alex40724 ,

today I had a chat with @lauraherzog . According to Laura, the actual issue that occured on `MacOS` should be fixed with this PR. Laura had to adjust some arguments used in the former version of `(head -n $header_length ${file} && tail -n +$((header_length + 1)) ${file} | sort) > ${file}.tmp`, so the current/fixed implementation in this PR should be fine. She doubts that there is a `Locale` based problem with the proposed script.

Can you please check whether this is still in issue with your infrastructure and provide a feedback here? If this is indeed still a problem, Laura will implement a PHP script as an alternative.

Best regards,
@mjansenDatabay

klees:
This uses php code for the sorting now. I hope this will be stable...

PR #4592, opened 2022-05-24
Setup: Fix infinite loop

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=32746

FYI @daniwe4


klees:
Thanks a bunch!

PR #4595, opened 2022-05-24
Setup: prevent config file in webroot (#30236)

klees:
Hi everyone,

this prevents people from doing stupid things (in this case: putting their setup config in the webroot).

https://mantis.ilias.de/view.php?id=30236

Thx @swiniker for reporting. Would you be so kind and review this?

Best regards!

### DIFF
--- a/src/Setup/CLI/HasConfigReader.php
+++ b/src/Setup/CLI/HasConfigReader.php
@@ -44,6 +44,10 @@ trait HasConfigReader
         }
 
         $config_file = $input->getArgument("config");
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much for this @klees !

This looks good to me.

One thought: People installing in a subfolder might still endanger their installation. We could check one level further up, unless this level is the filesystem-root. We then run into the issue, that the message is going to be tricky. Maybe something like:
"The config file can not reside inside the ILIAS-Directory or it's parent folder." This will give a little bit of added security, but I also understand if you do not want to do this.

Thanks again.

PR #4597, opened 2022-05-25
[FIX] 0032472: Plugin-Icons sollten über Skin anapssbar sein

chfsx:
Hi @Amstutz 

This improves the usage of Custom-Icons for Plugins in a skin (see https://mantis.ilias.de/view.php?id=32472 ). Prior to this, Plugin-icons had to be in a Subfolder of images which wasn't documented. With this commit, both options are possible, icons directly in images and in the old subfolders. I suggest to remove the subfolder variant with ILIAS 8.


Amstutz:
Thx a lot! Since this is highly System Style related, I feel responsible for this snippet of code. I hope that is fine. I had a short talk with @chfsx . Since we are only slightly extending the public interface and no breaking changes are introduced, we can imo safley merge this to 7. 

Thx a lot!

PR #4598, opened 2022-05-25
7/overhaul of test question editing tabs

mbecker-databay:


dsstrassner:
Thanks, @mbecker-databay!

This fixes https://mantis.ilias.de/view.php?id=30436

PR #4599, opened 2022-05-25
Fixed javascript creation and calling for cloze question

gvollbach:
The inclusion of Javascript seems to be broken in ILIAS 8, this PR fixes main part of this problem.


gvollbach:

mbecker-databay:
After speaking with gvollbach and evaluating other changes since this was opened, I close this PR. 
Still, thank you for providing it!

PR #4604, opened 2022-05-27
8/bugfix/matching question javascript

gvollbach:
Fixed javascript initialization for matching question preview and test view.


gvollbach:

PR #4605, opened 2022-05-27
Fixed char counter init for text question

gvollbach:
The inclusion of javascript seems to be broken in ILIAS 8, this PR fixes main part of this problem, for the preview and test view of the text question.


gvollbach:

PR #4606, opened 2022-05-27
8/bugfix/mc question charcounter

gvollbach:
Fixed char counter in mc preview and test output.


mbecker-databay:
Thank you so much!

PR #4608, opened 2022-05-29
fixed bug 32774: Item with Lead Avatar and Progress Meter Broken

alex40724:
See https://mantis.ilias.de/view.php?id=32774


Amstutz:
thx a lot! I will provide a unit test as well, to make sure, this is kept also in future revisions.

@klees : We seem to miss many revised copyright sections. we should fix that in one go.

PR #4612, opened 2022-05-30
Test/TestQuestionPool: Remove `strftime`

mjansenDatabay:
See: https://php.watch/versions/8.1/strftime-gmstrftime-deprecated

IMO there is no reason to use `local-dependant` functions for the found occurances, so I used `date` to replace the usages of `strftime` (except one HTML output case where I used `ilDatePresentation` / Impact: The date presentation will **not** look like this `Mon May 30 12:51:34 2022` anymore in the affected view).


mbecker-databay:

PR #4617, opened 2022-05-31
WAC check for assessment

lscharmer:
Add WAC check for the assessment file answers.

Mantis: https://mantis.ilias.de/view.php?id=29087


mjansenDatabay (approved):
We already did some internal review loops. I am fine with the result.

PR #4622, opened 2022-05-31
Dashboard: Take SAHS display options into account

lscharmer:
Mantis: https://mantis.ilias.de/view.php?id=31745

The Dashboard uses the `ilObjectListGUI::getAsListItem` method to create a list item. 
The `ilObjectListGUI::getAsListItem` method does not use the `ilObjectListGUI::modifySAHSlaunch` and does not set the `target` attribute to the anchor when it is needed.


iszmais (approved):
Thx @lscharmer for the solution!

Looks good and seems to work. :+1: 

Greetings @iszmais

PR #4624, opened 2022-06-01
Export: remove TargetRelese from exports

klees:
Hi everyone,

basically, the field `TargetRelase` in the export seems to be of no use, hence we might as well remove it.

Best regards!

### DIFF
--- a/Services/Export/classes/class.ilExport.php
+++ b/Services/Export/classes/class.ilExport.php
@@ -1,18 +1,21 @@
 <?php declare(strict_types=1);
 
-/******************************************************************************
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
  *
- * This file is part of ILIAS, a powerful learning management system.
- *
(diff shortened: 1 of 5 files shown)


alex40724:
The target release is passed to all exporter classes through getXmlExportHeadDependencies/getXmlExportTailDependencies
e.g.:
$sequence = $exp->getXmlExportHeadDependencies($a_entity, $a_target_release, $a_id);

The derived schema version is passed to all Dataset classes.

The idea was, to enable exports to previous versions of ILIAS, e.g. through a "Release" dropdown in the main export screen. However this has never been finally implemented (missing funding). But the idea has not been abandonded (yet). If we abandon this idea completely a removal of this parameter would require changes in almost all Export and Dataset classes.

alex40724:
@mjansenDatabay Not sure why this had a "bugfix" label, but there is no mantis ticket for this and this does not seem to be a bug at all. I removed the label.

dsstrassner:
32821

alex40724:
Removed from output, still passing it through the process now.

PR #4625, opened 2022-06-01
Fix reassignemt Cron for Study programms

iszmais:
https://mantis.ilias.de/view.php?id=32823
Fixes an syntax error and additionaly fix language loading.


klees:
Thx! I did not move this fix to >6 since the initial issue seems to be fixed, although differently, there.

PR #4626, opened 2022-06-01
Fix longmenu question

kergomard:
This has gotten bigger, then I would have liked, but as I fixed the first issue (gap removal in longmenu question, see  https://mantis.ilias.de/view.php?id=32440 for the mantis issue) new ones started popping up. What is in here:
* I renamed the js and template files to a consistent scheme. I did this also for the clozeQuestion as the structure is rather similar.
* I also fixed an issue whereby the js was not properly loaded in the editor of the clozeQuestion.
* I moved javascript out of all template files on the editing side, but didn't touch the player side. This I will do in a follow-up against ILIAS 8. While doing this I removed the debugPlayer, as it didn't seem to be worth the effort.
* I restructured the javascript in the longmenuQuestionGapBuilder as it was really entangled and confusing.
* I mostly removed jQuery.
* I made a minor change in TinyMCE to pass on the editor instance when calling the callbacks.
* ...and I actually also fixed the original issue.
I'm very much looking forward to the feedback and suggestions for improvement.


dsstrassner:
Hi @nhaagen, would you please do me, @kergomard and @mbecker-databay the favor to review the JS in this LM-Question PR? We have discussed this at the TechSquad and I think you have the skills to do this. Afterwards, @kergomard could merge this PR and will be mainly responsible for the LM Question Issues in ILIAS 8.

Amstutz:
@kergomard thx a lot for the proposal!

@dsstrassner, special thx for pushing this. According to @jweber-unibe it is our last major issue concerning the Long Menu Question. Would be great to have this fixed in the core soon.

Modules/TestQuestionPool/templates/default/longMenuQuestionGapBuilder.js
```diff
@@ -0,0 +1,457 @@
...
+	};
+
+	pub.setTextAreaValue = (text) =>
+	{
+		let cursor, inGap, inst;
+		if (isTinyActiveInTextArea())
+		{
+			if (navigator.userAgent.indexOf('Firefox') !== -1)
+			{
+				console.log('Here');
```
nhaagen:
```suggestion
```

nhaagen (approved):
Hi @kergomard ,
thumbs up for this one. Your adaption looks a lot cleaner and way more solid by means!
in fact, the only thing I saw was the leftover console.log, and maybe calling parseInt to increase a counter is a bit over, but then this was in the code before and will do no harm ;)
So, thanks a bunch, this will be way easier to work with.

kergomard:
I did some more fine-tuning and fixed a couple of errors I introduced. I will run a final quality control on this after my vacation in two weeks and then merge and cherry-pick. ...this way I'm around to fix the things I broke.

PR #4630, opened 2022-06-02
Fix issue with competences in Cloze Question

kergomard:
I had a look at https://mantis.ilias.de/view.php?id=32206 as part of the refinement efforts and decided to provide a PR. Here it is.


dsstrassner:
Alina will like this very much ;)

Thanks a lot @swiniker

kergomard:
I changed my solution, as after some more thought with a well rested brain, this seems more logical. I will also have a look at this next week with Alina.

kergomard:
I talked with Alina by the way and as far as I can see, this should be the right solution.

PR #4634, opened 2022-06-03
Update maintenance files

pascalseeland:
This PR updates all remaining maintenance.json files. 
@smeyer-ilias Please check if all modules/services that you wanted to hand over now have proper maintenance.json files.


smeyer-ilias (commented):
Thanks @pascalseeland 
OpenIdConnect is missing, correct?

PR #4637, opened 2022-06-04
Bugfix: no title/link for media used in snippets

padvincenzo:
In a mediapool, if a media object is used in a snippet page, the relative row in the usage tab is empty: no title, no link, no type...

Previous situation:

Previous situation

After fix:

After fix


alex40724:
Great, thanks for the PR!

alex40724:
Merged to 7, 8 and trunk

PR #4638, opened 2022-06-07
LSO: Fix lost characters during XML import

mjansenDatabay:
This PR fixes the handling of text nodes in the XML import of `lso` repository objects. The assumption that all characters of an XML text node (character data) are completely passed in a call of `handleEndTag` was wrong.

Mantis Issues:
- https://mantis.ilias.de/view.php?id=31944
- https://mantis.ilias.de/view.php?id=31945


klees:
Thanks @mjansenDatabay. It looked strange at first, but I think I understand now =)

klees:
Cherry-picked to release_6 and trunk.

PR #4639, opened 2022-06-07
Node/Dependencies: patch-packages

nhaagen:
There is a glitch in owl.carousel, which is added as library via  both node and bower.
The PR removes the bower-component and uses the node-module instead (AccordionGUI).
But then there still is the need of patching the library; there is a "patch-packages" mechanism, and I had it running in an isolated environment; however, I'm a little unsure, if this is really the way to go, since it brings a lot of dependencies on its own.
Opinions on this are very appreciated.


kergomard:
Thank you very much @nhaagen.
We merge this to 7 and cherry-pick it to trunk, but we will need to get rid of this again after the release of ILIAS 8.

kergomard:
@nhaagen : I just realized something when implementing this in trunk. I don't think the patch has been applied, right? Am I missing someting?
I thought I would ask you before diving deeply into the rabbit hole of patch-package ;-)

PR #4640, opened 2022-06-07
Object: move fallback to _SESSION to another place (#32831)

klees:
Hello everyone,

since `Services\Object` currently has no maintainer, neither implicit nor explicit, I kindly ask everyone to look into this change.

This is supposed to fix [#32831](https://mantis.ilias.de/view.php?id=32831). Instead of overwriting `$_POST`, the value in question `tit` is pulled from `$_SESSION` in the only consuming location that I could find.

Best regards!

### DIFF
--- a/Services/Object/classes/class.ilObjectCopyGUI.php
+++ b/Services/Object/classes/class.ilObjectCopyGUI.php
@@ -587,11 +587,8 @@ class ilObjectCopyGUI
         if ($this->post_wrapper->has('tit')) {
             $this->tpl->setOnScreenMessage('info', $this->lng->txt('wizard_search_list'));
             $_SESSION['source_query'] = $this->post_wrapper->retrieve("tit", $this->refinery->kindlyTo()->string());
-        } else {
-            // TODO: refactor this in next iteration
-            $_POST['tit'] = $_SESSION['source_query']; // TODO PHP8 Review: Remove/Replace SuperGlobals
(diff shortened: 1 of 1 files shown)


alex40724:
@smeyer-ilias Could you have a look at this, please?

Services/Object/classes/class.ilObjectCopyGUI.php
```diff
@@ -601,7 +598,11 @@ protected function searchSource() : void
             return;
         }
         
-        $query_parser = new ilQueryParser($this->form->getInput('tit'));
+        $tit = $this->form->getInput('tit');
+        if ($tit === "") {
+            $tit = $_SESSION['source_query']; // TODO PHP8 Review: Remove/Replace SuperGlobals
```
smeyer-ilias: Please use ilSession::get('source_query', '') here.

smeyer-ilias (changes requested):
Thanks for fixing this. 
Please remove the superglobal access.

Services/Object/classes/class.ilObjectCopyGUI.php
klees: Done!

klees:
Thx @smeyer-ilias, I introduced ilSession.

smeyer-ilias:
Thanks for fixing this.

PR #4644, opened 2022-06-09
Migration: fix method addNewType in ilDBUpdateNewObjectType

daniwe4:
Thanks Fabian Wolf for your hint.


fwolf-ilias:
I checked with the new database template file and everything works fine now. Thank you very much for looking through this.

PR #4646, opened 2022-06-09
Forms/Autocomplete: separation and late init of JS-parts

nhaagen:
https://mantis.ilias.de/view.php?id=31586


Amstutz:
> jstpl 

Nice one :-)

alex40724:
Thanks for the PR!

PR #4647, opened 2022-06-09
UI Components Icon: Remove Outlined Option, introduce new Milos Set

Amstutz:
This PR introduces the change from the previous "Milos" Iconset with Outlined and Standard Icons to the new "Milos" set with only outlined icons. This change has been approved on the 4th of April for ILIAS 8, see: https://docu.ilias.de/goto.php?target=wiki_wpage_7351_1357 .

In this PR we propose the necessary change to the UI Components public Interface, namely to get rid of the outlined Icons: See: https://github.com/ILIAS-eLearning/ILIAS/pull/4647/files#diff-4bb10a9e4dce4c1332aa117ecea6e3a4f282f1c8e46910ba66ed71e75c2ea6dc

Note that the Icons presented as the new Milos Icon here have yet "Draft" Status, meaning that they will changed in the upcoming days as soon as the compete set is done. The change in the code however should be final.

Further Note that one Unit test is failing. This is expected, since this test is checking, if all standard icons are available, which is not yet the case. This PR can and should only be merged, once the set is final an complete (all tests are passing).


Amstutz:
One question that we should discuss in the JF:

- In the CSS Squad, we had the request for publishing the illustrator binary files next to the used SVG files for Icons to ease the work necessary to redesign them. However, ai is a binary format (small, but binary still). Should we publish this along with the source?

matthiaskunkel:
Jour Fixe, 13 JUN 2022 :  Timon informed all participants about the change of the interface and the removal of the outlined icons. With ILIAS 8 there only will be standard icons. 
Concerning the question of providing the Illustrator filed of all Milos icons we prefer to have a separate repo in the ILIAS GitHub repo and to make them available there for all who want to modify the existing iconset.
Please merge new iconset and modified files to trunk.

tests/UI/Component/Symbol/Icon/IconTest.php
```diff
@@ -195,18 +173,8 @@ public function testAllStandardIconsExist() : void
             $path = self::ICON_PATH . "icon_" . $icon_abr . ".svg";
             $this->assertTrue(file_exists($path), "Missing Standard Icon: " . $path);
         }
-    }
+        exit;
```
iszmais, line 176: @Amstutz This seems to break the Tests in trunk. (Ends the Test run, therefore all tests after this (aprox 500) are inofficially skipped).

Im not sure if there is an solution for this by adaption of the pipeline but i assume the remove of this line should is the best option

_Thanks to @lscharmer for the initial hint._

iszmais (commented):
The Tests dont complete.

tests/UI/Component/Symbol/Icon/IconTest.php
klees: Good catch you both. Will look into this, @Amstutz is in the holidays.

klees: I have removed the thing: https://github.com/ILIAS-eLearning/ILIAS/commit/02c5234effc662cad1cd9b63a83858ae7944c2d1

PR #4649, opened 2022-06-10
Filter courses and groups on "right to leave" in unsubscribe multiple view

iszmais:
https://mantis.ilias.de/view.php?id=32130

Reduces the **unsibscribe multiple** view on items with the required permission `leave`

### DIFF
--- a/Services/Dashboard/ItemsBlock/classes/class.ilPDSelectedItemsBlockGUI.php
+++ b/Services/Dashboard/ItemsBlock/classes/class.ilPDSelectedItemsBlockGUI.php
@@ -513,10 +513,20 @@ class ilPDSelectedItemsBlockGUI extends ilBlockGUI implements ilDesktopItemHandl
         $top_tb->setLeadingImage(ilUtil::getImagePath('arrow_upright.svg'), $this->lng->txt('actions'));
 
         $button = ilSubmitButton::getInstance();
+        $grouped_items = $this->blockView->getItemGroups();
         if ($this->viewSettings->isSelectedItemsViewActive()) {
             $button->setCaption('remove');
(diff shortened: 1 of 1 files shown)


Services/Dashboard/ItemsBlock/classes/class.ilPDSelectedItemsBlockGUI.php
```diff
@@ -513,10 +513,20 @@ public function manageObject()
...
 
         $button = ilSubmitButton::getInstance();
+        $grouped_items = $this->blockView->getItemGroups();
         if ($this->viewSettings->isSelectedItemsViewActive()) {
             $button->setCaption('remove');
         } else {
             $button->setCaption('pd_unsubscribe_memberships');
+            foreach ($grouped_items as $group) {
+                $items = $group->getItems();
+                $group->setItems([]);
```
mjansenDatabay, line 523: I suggest to add a `filter` method and pass a corresponding callback instead of this kind of workaround.

iszmais: My thought too but i had difficulties to make a decision where to put it since the sturcture at this point is already unconvinient. Since all manager view on the dashboard use the same data source i would like to centralise it there since i dont want to encourage this behaviour. Otherwise a local filter method would imply the expandability of this blokcGUI wich is currently not given due to a very static approach wich can only differentiate between 2 states.

Therefore i would prefer to keep this bugfix as minimal inversive and approach a refactoring sometime in the future instead of partial approaches between different PRs.

lscharmer (approved):

PR #4652, opened 2022-06-13
Fix string in goto of forum

kergomard:
This fixes an error, when thread is a string (which it always is, as far as I can see).
You can see the behavior on this link: https://test8.ilias.de/goto.php?target=frm_68831_4042&client_id=test8


mjansenDatabay:
Thx @swiniker

PR #4654, opened 2022-06-14
Fix 32824 trunk

kergomard:
This is the sister pr for #4636


kergomard:
I keep the different commits to make a review easier. You might want to squash them when merging.

PR #4655, opened 2022-06-15
More fixes for edit question

kergomard:
This finalizes the work started in the following PR : https://github.com/ILIAS-eLearning/ILIAS/pull/4642

- It makes sure that the "Finish Editing"-Link in the Question Page always leads to the right page.
- It also makes sure that the "Back"-Links always points to the test and not suddenly to a none existent question pool

I opted for shortness of the implementation and thus added the same parameters to all calls. Not all of them are always necessary, but I would suggest to let this slide and fix it when we tackle the rather wonky command-flow.


mbecker-databay:

PR #4661, opened 2022-06-16
Fix clickthrough issues on PHP8

kergomard:
I clicked through the first level of the test and fixed the issues I found:
- Title and description missing in Settings.
- Dashboard could not be opened.
- Access to undefined array key.


mbecker-databay:
Awesome, thank you so much!

PR #4663, opened 2022-06-16
Fix undefined array pos_cens_date

kergomard:
Fixes an access to the potentially undefined array key pos_cens_date


mjansenDatabay:
Thx @swiniker

PR #4665, opened 2022-06-20
UI/Drilldown: 32724, add jsdom and fix tests

nhaagen:
https://mantis.ilias.de/view.php?id=32724

1. jsdom is required (19.0.0, until the bug in 20.0.0 is fixed)
2. Move mocks/global prerequisites into "beforeEach"

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "esm": "^3.2.25",
     "less": "^4.1.1",
     "mocha": "^8.0.1",
+    "jsdom": "19.0.0",
     "rollup": "^2.18.0",
     "rollup-plugin-terser": "^6.1.0"
   },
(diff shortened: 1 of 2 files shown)


klees (commented):
Hi @nhaagen,

looks as it this would work. But: When looking into this myself I thought we could improve this a little, to make code better and tests simpler with the same move. I thought we might:

* get rid of jquery and use browser functions directly to get rid of dependencies
* inject `window`, `document` and `il.Utilities.CookieStorage` to the appropriate modules of drilldown to make depencies explicit

Please have a look if these improvements. Are they feasible?

Best regards!

nhaagen:
Perfectly feasible, and certainly a good hint. however - some more than just 5 minutes. Id rather postpone, if OK with you?

Amstutz (approved):

PR #4671, opened 2022-06-20
Add ability to override filename of exported file

mBeym:
adds ability to override the filename of a table export by defining the **`getExportFileName`** function  
in a table class extending from **`ilTable2GUI`**

### DIFF
--- a/Services/Table/classes/class.ilTable2GUI.php
+++ b/Services/Table/classes/class.ilTable2GUI.php
@@ -2627,7 +2627,7 @@ class ilTable2GUI extends ilTableGUI
                 );
             }
 
-            $filename = "export";
+            $filename = $this->getExportFileName((int) $format);
 
             switch ($format) {
                 case self::EXPORT_EXCEL:
@@ -2827,5 +2827,10 @@ class ilTable2GUI extends ilTableGUI
             $this->limit_determined) {
             $this->rows_selector_off = true;
         }
+    }
+
+    protected function getExportFileName(int $format) : string
+    {
+        return 'export';
     }
 }


alex40724:
Thanks for the PR.

Is this part of any feature request? I currently do not extend the API of these legacy classes, unless there is a very urgent case or reason.

colin-kiegel:
@alex40724 My most recent encounter with this limitation of `ilTable2GUI` was a repository object plugin with some kind of tracking data presented in a table. Admins needed to be able to download the tracking data.

A generic filename "export.xls" is a No-Go for me from UX perspective - the filename should at least include some useful context information - that's why we introduced this patch.

With this PR a plugin like that wouldn't need an additional patch to provide meaningful filenames of table exports.

alex40724:
@colin-kiegel I think the plugin could simply extend the class. Anyway, this is legacy and under implicit maintainership only, open for anyone to take. As I wrote I would not like to further extend this personally.

mBeym:
@alex40724 It reduces the amount of duplicated code. Everytime a table needs to get exported. 
I would have to create an entire class. 
Do the exact same I would have done before. 
Copy the entire `exportData` method. 
Change the name.

Meaning. multiple lines are now duplicated (77 to be exact) and that for every Table I need to export with a logical sounding name.

Not even talking about the hassle when every export is called "export.xyz" and nobody has an idea WHAT it is until you look inside.

alex40724:
@mBeym Ok, I understand that this is more a convenience issue. I will close this PR. If someone is willing to maintain these classes this could be reopened. This might not seem to be very attractive, but if approached together with the KS data tables, there would definitely be some funding available.

PR #4672, opened 2022-06-20
Forum: Tweak 0032881: Reset Limited Thread View Message Box

fhelfer:
This fixes https://mantis.ilias.de/view.php?id=32881

### DIFF
--- a/Modules/Forum/classes/class.ilObjForumGUI.php
+++ b/Modules/Forum/classes/class.ilObjForumGUI.php
@@ -1317,11 +1317,8 @@ class ilObjForumGUI extends ilObjectGUI implements ilDesktopItemHandling, ilForu
             $thr_pk,
             $pos_pk
         );
-        if ($this->selectedSorting === ilForumProperties::VIEW_TREE && ($this->selected_post_storage->get($thr_pk) > 0)) {
-            $info = $this->getResetLimitedViewInfo();
-        }
-        $mainContent = $this->viewThreadObject();
-        $this->tpl->setContent(($info ?? '') . $mainContent);
+
+        $this->viewThreadObject();
     }
 
     /**
@@ -1748,7 +1745,7 @@ class ilObjForumGUI extends ilObjectGUI implements ilDesktopItemHandling, ilForu
         );
         $this->tpl->setOnScreenMessage('info', $this->lng->txt('forums_post_was_activated'), true);
 
-        $this->tpl->setContent($this->viewThreadObject());
+        $this->viewThreadObject();
     }
 
     private function deletePostingObject() : void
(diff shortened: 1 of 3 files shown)


Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -1315,8 +1314,9 @@ protected function selectPostObject() : void
             (int) $this->httpRequest->getQueryParams()['thr_pk'],
             (int) $this->httpRequest->getQueryParams()['pos_pk']
         );
-        $this->showResetLimitedViewInfo();
-        $this->viewThreadObject();
+        $info = $this->getResetLimitedViewInfo();
```
mjansenDatabay: IMO the following condition is missing here:

```php
if ($this->selectedSorting === ilForumProperties::VIEW_TREE && ($this->selected_post_storage->get($thr_pk) > 0)) {
}
```

mjansenDatabay (commented):
One (hopefully) last question: Why did you have to use `$this->tpl->setContent(...)` for all these occurances of `$this->viewThreadObject()`?

fhelfer:
> One (hopefully) last question: Why did you have to use `$this->tpl->setContent(...)` for all these occurances of `$this->viewThreadObject()`?

viewResetLimitedViewInfo() added the message box via $this->tpl->setOnScreenMessage()
I used the KS Message Box. $this->tpl->setOnScreenMessage() would have added another box around the box.
So i changed the method to return the html string and merged it with viewThreadObject() html
Threrefore i changed all occurences

mjansenDatabay:
Yes, but can't you just concatenate the two strings in `viewThreadObject` at the bottom of the function?

fhelfer:
Yes but i had to either pass it as a parameter
or call getResetLimitedViewInfo() inside
First option changes the the meaning of the function for only one occurence
Second option sounds good, but i didnt know when (in which case) i had to call it, to ignore it for all other calls

Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -115,8 +115,7 @@ public function __construct($data, int $id = 0, bool $call_by_reference = true,
                 $this->selected_post_storage->get($thr_pk) ?? 0,
                 $this->is_moderator
             );
-
-            $this->showResetLimitedViewInfo();
+            $this->ctrl->setParameter($this, 'thr_pk', $this->objCurrentTopic->getId());
```
mjansenDatabay, line 118: Do we really need this line? This seems to be duplicated in https://github.com/ILIAS-eLearning/ILIAS/pull/4672/files#diff-fd85924abce90fbeedabb3e3c4f121dfce9fedd4820943703ccf4a70ab4c31f9R5897

fhelfer: Deleting this line will result in error when returning from limited view
with message Could not find first posting by id: 0

PR #4673, opened 2022-06-20
Fix user id is zero

lscharmer:
This PR fixes that the user id is zero under certain circumstances. The user id will be the anonymous user in these cases.
Mantis Ticket: https://mantis.ilias.de/view.php?id=32063

### DIFF
--- a/Services/User/classes/class.ilObjUser.php
+++ b/Services/User/classes/class.ilObjUser.php
@@ -212,6 +212,7 @@ class ilObjUser extends ilObject
             $this->setId($a_user_id);
             $this->read();
         } else {
+            $this->setId(ANONYMOUS_USER_ID);
             // TODO: all code in else-structure doesn't belongs in class user !!!
             //load default data
             $this->prefs = array();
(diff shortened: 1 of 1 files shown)


chfsx:
sorry, wrong PR, will post this comment in #4169

Services/User/classes/class.ilObjUser.php
```diff
@@ -212,6 +212,7 @@ public function __construct($a_user_id = 0, $a_call_by_reference = false)
             $this->setId($a_user_id);
             $this->read();
         } else {
+            $this->setId(ANONYMOUS_USER_ID);
```
alex40724, line 215: Thanks for the PR.

Please note that @pascalseeland takes care of the authentication service, but not of the user service, which is still maintained by @smeyer-ilias.

This should not be fixed in the user class imo. The ID is usually set to 0 when new objects (here new users) are created. It is wrong to set the anonymous ID in these processes automatically. New user objects should not get the anonymous user ID. Whatever code calls the constructor here during initialisation should already pass the anonymous user ID.

alex40724 (changes requested):

alex40724:
Since @kergomard now took over this component, we left it to him to decide how to move forward here.

kergomard:
Thank you @lscharmer for this PR that has been around for so long.
I agree with @alex40724 , that this is most probably not the right solution and thus close this PR. I agree though, that the handling of users that are not authenticated is a mess even after the improvement and should be streamlined and improved.
Thanks again and best,
@kergomard

PR #4675, opened 2022-06-21
ContentPage: Tweak 00328235: Remove Action Menu from ContentPage Editor

fhelfer:
This fixes https://mantis.ilias.de/view.php?id=32235

### DIFF
--- a/Modules/ContentPage/classes/class.ilObjContentPageGUI.php
+++ b/Modules/ContentPage/classes/class.ilObjContentPageGUI.php
@@ -214,7 +214,7 @@ class ilObjContentPageGUI extends ilObject2GUI implements ilContentPageObjectCon
         $this->addToNavigationHistory();
 
         if (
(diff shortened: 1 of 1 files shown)


Modules/ContentPage/classes/class.ilObjContentPageGUI.php
```diff
@@ -213,7 +213,13 @@ public function executeCommand()
 
         $this->addToNavigationHistory();
 
-        if (strtolower($nextClass) !== 'ilobjstylesheetgui') {
+        if (
+            strtolower($nextClass) !== 'ilobjstylesheetgui' &&
```
mjansenDatabay: Please use the `::class` property when comparing this (lower case, of course)

mjansenDatabay (changes requested):
Thx @fhelfer , I had only one tiny remark.

mjansenDatabay (approved):

PR #4676, opened 2022-06-21
fixed resolution of some language variables in permissions

schmitz-ilias:
Fixed the resolution of language variables in the permissions, see [https://mantis.ilias.de/view.php?id=32909](url). This includes the name of custom roles from plugins, the name of all roles in the log, and the name of custom operations from plugins in the permissions table.


Services/AccessControl/classes/class.ilObjRoleGUI.php
```diff
@@ -325,7 +325,7 @@ protected function readRoleProperties(ilObjRole $role)
 
         $rbacreview = $DIC['rbacreview'];
         
-        $data['title'] = $role->getTitle();
+        $data['title'] = $role->_getTranslation($role->getTitle());
```
smeyer-ilias: Please change this to a static call, since the method is static: ilObjRole::_getTranslation(...)

Services/AccessControl/classes/class.ilObjRoleTemplate.php
```diff
@@ -46,6 +46,15 @@ public function __construct($a_id = 0, $a_call_by_reference = false)
     }
 
 
+    /**
+     * return translated title for autogenerated templates
+     * @return string
```
smeyer-ilias: Please remove the phpdoc-comment (\return string), since it does not provide additional information.

smeyer-ilias (changes requested):
Thanks @schmitz-ilias for fixing this.
Please have a look at the comments, there are only two minor issues.

smeyer-ilias (approved):
Thanks

iljalukin:
This PR causes a change in the display of the title in the "Edit Properties" dialog of the role.
Previously, the role title was displayed there in the form il_crs_admin_12345 and could be used in further workflows such as assigning the local role when creating registration codes.
Currently, the translated title "Course Administrator" is displayed there and you have no way to get to the role title. 
Is the change of displaying the title really intended and necessary at this point?

schmitz-ilias:
Hi @iljalukin, I talked to @smeyer-ilias about the issue you brought up. We propose to not revert to the previous behavior, but instead to add the role title in the form il_crs_admin_12345 in brackets to the end of the title in the header of the role GUI. 
I will implement this in the next few days and link the PR here.

PR #4682, opened 2022-06-22
Fix length for Mime Type Recognition

kergomard:
Sorry this is a fix for a previous error I committed: We need a few more bytes to be sure mime-type recognition works. 255 Bytes should be enough though. I couldn't find any resources on it, so the number is completely magical, it is the equivalent of the max-length of the full mime-type name. I will provide a pr back to ILIAS7 together with the MS-Office-Type fix. Let me know if you would prefer another number.

Thanks!


chfsx:
Thanks @kergomard !

PR #4685, opened 2022-06-23
Setup: Fix infinite loop (mantis #32746)

utesche:
... to fix reported error from reopening https://mantis.ilias.de/view.php?id=32746 on 23.6.2022.


klees:
Thx Uwe!

utesche:
Hi Richard,

perhaps,  here:
https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Services/AccessControl/classes/Setup/class.ilAccessRBACOperationClonedObjective.php#L56 

and here:
https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Services/AccessControl/classes/Setup/class.ilAccessRBACOperationClonedObjective.php#L87

are the next places the same kind of looping may pop up ...  ??

klees:
indeed...

klees:
@daniwe4 Could you have a look into the objectives you derived from the former migrations if they contain this pattern in any other places? Please fix the all!

PR #4688, opened 2022-06-23
introduces singularity for target notification types

iszmais:
https://mantis.ilias.de/view.php?id=32985

Fixes the possibility to have multiple notifications from a type, even if its inconvinient for that type, by introducing a register for notification type that are unique for their user.


mjansenDatabay (approved):
The notification related changes LGTM!

alex40724 (approved):
The changes in the awareness code are of course ok for me. I still don't understand the issue however. I was not able to reproduce any behaviour I would consider being a bug.

However imo it's the notification component that defines the behaviour here. As a consumer the awareness tool must only serve the interface correctly.

PR #4689, opened 2022-06-23
#33003: Fix by introducing new ConstantTransformation and changing ArrayBasedRequestWrapper::retrieve

klees:
Hi everyone,

this is supposed to fix [#33003](https://mantis.ilias.de/view.php?id=33003). Basically, the current behaviour of `ArrayBasedRequestWrapper::retrieve` to throw exceptions on unknown keys already bothered me some times before, but now I've had enough =)

Currently, we need to guard almost every usage of `retrieve` with a `has`. I propose to make `retrieve` pass `null` to the supplied transformation on unknown keys. Reasoning is as follows:

* If someone wants to use the old `has` -> `retrieve` pattern, they still can. No existing and working code should be affected by the change.
* We now can also use a pattern where the handling of non existing keys is internalized in the transformation, as demonstrated in the fixed file. From my POV this is a lot more readable, since the intent is declared (instead of the execution being controlled).
* A quick research (Caching PSR 6) suggests that this also might be expected by some people.

I briefly discussed with @mjansenDatabay, maybe he wants to add some aspect here...

I kindly ask @chfsx to look into the changes in src/HTTP. I also added a new transformation that always returns a constant value. @mjansenDatabay, would you please review?

Best regards!

### DIFF
--- a/Modules/IndividualAssessment/classes/class.ilIndividualAssessmentMembersGUI.php
+++ b/Modules/IndividualAssessment/classes/class.ilIndividualAssessmentMembersGUI.php
@@ -133,7 +133,8 @@ class ilIndividualAssessmentMembersGUI
 
     protected function addedUsers() : void
     {
-        if ($this->request_wrapper->retrieve('failure', $this->refinery->kindlyTo()->bool())) {
+        $r = $this->refinery;
+        if ($this->request_wrapper->retrieve('failure', $r->byTrying([$r->kindlyTo()->bool(), $r->always(false)]))) {
             $this->tpl->setOnScreenMessage("failure", $this->txt('iass_add_user_failure'));
         } else {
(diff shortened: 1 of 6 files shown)


mjansenDatabay (approved):
Thx for the new transformation @klees , I am sure there is already existing code where `always` can be used. We (@lscharmer and I) have thought about an `Always` transformation while we discussed the improvement of the `Refinery` as we analysed the consuming transformation code `\ILIAS\Setup\Agent::getArrayToConfigTransformation`.

And yes, @klees and I discussed the semantics of the `retrieve` and `has` methods, compared to similar mechanics used in PSR standards like caching, HTTP etc.. So I totally agree to the arguments of @klees .

chfsx (approved):
Thanks a lot, LGTM as well!

PR #4691, opened 2022-06-24
Implemented check if install works on daily schedule

catlauraherzog:
I have implemented an install check for ILIAS in the CI Pipeline. Once every day there will be an automatic checkup which tries to install the trunk and release_7 branch in github actions to see if it works.


kergomard:
Thank you very much, @lauraherzog !

PR #4692, opened 2022-06-24
Fix more Globals in test

kergomard:
This PR contains a few fixes for assignments to globals that blew up in my face when running more tests on test8. I simply removed two of them and changed the read out of a ref_id from "calling_test" as it might be a string. I also changed a few accesses to potentially undefined arrays and a wrong type.

I left the following assignment to $_GET/$_POST/$_REQUEST in: https://github.com/ILIAS-eLearning/ILIAS/blob/e150251c515f85cf0bf2464d20193a31e298f0d1/Modules/Test/classes/class.ilObjTestGUI.php#L955 . If this ever gets called on test8 test goes boom, but then at least we know, that we needed. If this doesn't happen during beta-testing I would strongly suggest to remove it.


mbecker-databay:
Thank you so much!

PR #4694, opened 2022-06-24
Container: read needs to respect target_id = null (#33052)

klees:
Hi @alex40724,

this will fix [#33052](https://mantis.ilias.de/view.php?id=33052). I looked into other locations in the original stack trace to fix the problem. My conclusion was: since `getTargetId` might return `null`, `read` will have to deal with that possibility anyway. This might also happen in other circumstances.

Best regards!


alex40724:
Thanks for the PR.

PR #4695, opened 2022-06-24
Proposal for `<meta>` tags in the Kitchensink

thibsy:
As discussed with @klees I have implemented (a proposal for ILIAS 9) to add support of `<meta>` tags to the Kitchensink.

The reason behind this is a [feature-request](https://docu.ilias.de/goto_docu_wiki_wpage_6931_1357.html) which will rely on meta-tags. Currently the metadata is managed via GlobalScreen, but only supports key => value pairs. These pairs will then be rendered like `<meta name="key" content="value" />` in the Page component. There is no flexibility in regard of the attribute-type (by attribute-type I mean e.g. the name= part of the tag). The feature will rely on whats called the open-graph-protocol (https://ogp.me) which uses `property=` instead of `name=`. Therefore we have to improve how metadata is currently handled so we can define what kind of attribute-type should be used. Possible options (except `property`) can be found at [mozilla.org](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta). The property-attribute is an exception, but must be used according to the OGP to ensure support of different platforms, therefore we'd have to support it as well.

Now why the Kitchensink? Well the kind of metadata we are talking about here is the one stored in an HTMLs `<head>` tag. This sort of data is required when creating a Page object (currently an associative array). With this proposal the Page would be able to require an array of meta-components and could delegate the rendering process. 
To be able to represent the data we must validate in our DTO (even if not in the Kitchensink) if the attribute-type is supported/known and whether it allows to have content, because there are two kinds of attribute-types: ones that allow content, and ones that don't. Now since the validation is solely based on the HTML-representation, it makes sense to implement it in the UI component. IMO it also makes sense to treat the implementation as a component, because like all other components it represents an HTML-tag and would be considered an entity (according to the Component interface).

Kind regards!

### DIFF
--- a/src/UI/Component/Meta/Complex.php
+++ b/src/UI/Component/Meta/Complex.php
@@ -1,0 +1,31 @@
+<?php declare(strict_types=1);
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ */
+
+namespace ILIAS\UI\Component\Meta;
+
+/**
+ * @author Thibeau Fuhrer <thibeau@sr.solutions>
+ */
+interface Complex extends Standard
+{
+    public const ATTRIBUTE_NAME = 'name';
+    public const ATTRIBUTE_PROPERTY = 'property';
+    public const ATTRIBUTE_HTTP_EQUIV = 'http-equiv';
+    public const ATTRIBUTE_ITEM_PROP = 'itemprop';
+
+    public function getContent() : string;
+}
--- a/src/UI/Component/Meta/Factory.php
+++ b/src/UI/Component/Meta/Factory.php
@@ -1,0 +1,94 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ */
+
+namespace ILIAS\UI\Component\Meta;
+
+/**
+ * @author Thibeau Fuhrer <thibeau@sr.solutions>
+ */
+interface Factory
+{
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *     this component is used to add metadata to an HTML document, that only consists
+     *     of key => value pairs.
+     *   composition: >
+     *     this component consists of an HTML <meta> tag.
+     *   effect: >
+     *     mostly provides metadata about the page that is used by search engines or the browser.
+     *   rivals:
+     *     Rival 1: Complex >
+     *          Complex should be used instead, if the specification allows a content to be provided.
+     *
+     * context:
+     *     - this component will only be used in the <head> element of a Page component.
+     *
+     * rules:
+     *   usage:
+     *     1: this component must only be used when building a Page.
+     *     2: this component must only be used these attributes-keys:
+     *          - charset
+     *          - media
+     *   composition:
+     *     1: <meta key=value />
+     *
+     * ---
+     * @param string $key
+     * @param string $value
+     * @return \ILIAS\UI\Component\Meta\Standard
+     */
+    public function standard(string $key, string $value) : Standard;
+
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *     this component is used to add metadata to an HTML document, that consists of key => value
+     *     pairs which also must be provided with a content.
+     *   composition: >
+     *     this component consists of an HTML <meta> tag.
+     *   effect: >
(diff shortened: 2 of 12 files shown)


klees (commented):
Hi @thibsy,

thanks a lot for looking into this. This most certainly is an area where we could make good use of improvements. Besides the OpenGraph protocol, I immediately see another use case: For our [ILIAS-based SaaS product cate](https://www.cate-tms.de) we are currently looking into deploying [Content Security Policies](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP). These will most probably need to somehow take the actual content of the page into account, hence be added in the page head. This groundwork will allow us to propose according changes for the ILIAS core.

Currently, this does not seem to be in a state where I can give a proper standard review with questions, suggestions and change requests. From my perspective, this is not close to any solution that I would want to accept for the UI framework. Iterative changes along the standard review template probably won't take it there. I will try to explain my thoughts about this to make the reasons for this assessment transparent. I will then propose some alternatives for ways forward.

Before going  into the code, I want to comment on one peculiar part of your description: You seem to identify a component in the UI framework with a certain tag in HTML. Although many UI components might be rendered as one tag with certain content, many UI components will also be rendered to different locations. Think, e.g., about the onload-code for certain components: It will be added to the bottom of the `<body>` hence the component won't be represented by a single HTML tag. The general idea is to provide our consumers (requirements engineers, solution engineers, designers and programmers alike) with a set of ready-made UI components [that prevent them from thinking about HTML](https://github.com/ILIAS-eLearning/ILIAS/tree/release_8/src/UI#use-kitchen-sink-concepts).

So from my perspective, a component `Metadata` is waaay to technical to be understood as a UI component. We can very well put the according classes in the UI framework, but I would not add them to the component hierarchy. The sparse kitchen sink documentation and rules, which are both super technical, and the lack of meaningful examples are a good hint that these aren't proper UI components. Also, the rendering is odd because there really aren't many choices regarding the rendering here.

Regarding the encoding of the metadata variants as data types, I would suggest to stick to the [classification given by MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#attr-name): Create a base class `HTMLMetadata` and derive four subclasses from it: `HTMLMetadata\DocumentLevel`, `HTMLMetadata\PragmaDirective`, `HTMLMetadata\UserDefined` and `HTMLMetadata\OpenGraph`. I doubt that `CharsetDeclaration` is required as a subclass, this seems to be fixed for our project (as it is now) or it might be some property on a `Page` component, but we most certainly won't be adding multiple char set declarations anywhere. These classes would at least give some meaningful names to the data types, and I could google something like "HTML Metadata Pragma Directive" to find more information about what this might be used for.

OpenGraph seems to be violating the available attributes of `<meta>` (according to MDN, at least), and also has even more restrictions on the keys and values that can be used. Hence, a distinct class. It might be helpful to specify the available (or required) properties for OpenGraph even more closely, e.g. by introducing `HTMLMetadata\OpenGraph\Title`. This would allow introducing additional typing, e.g. by only allowing a `Data\URI` as parameter for the construction of a `HTMLMetadata\OpenGraph\URL`. This will also play nicely with the [`LanguageTag` proposed by @lscharmer and @mjansenDatabay](https://github.com/ILIAS-eLearning/ILIAS/pull/4653).

I would suggest implementing `HTMLMetadata` and the derived classes as  (abstract) classes because to me, they seem to be simple value objects: There is a definite set of possible variations, there is no need to make some distinct object act as `HTMLMetadata`, we seemingly do not require different representations of the concept, we don't need to hide our implementation. Hence, I cannot see a reason to introduce `HTMLMetadata` via an interface.

For the base class, I would suggest adding only one method: `HTMLMetadata::toHTML(): string` that produces the required HTML representation. Although I will mostly argue that HTML should be located in templates only (and mean that), I don't see that this is the way to go here. Other than the HTML for the various UI components, there is no choice regarding the proper rendering for a given variant of `HTMLMetadata`. A `<meta>` tag only is proper OpenGraph metadata if it is rendered exactly like defined in the according specification. I cannot choose to render a pragma directive other than `<meta http-equiv="...>` because it simply is not a pragma directive if not rendered like this. So, the HTML rendering of the metadata is intimately tied to the variant we want to represent and should be attached to the according class.

Then, location in the code. This could very well be a part of the UI framework, but IMO would also fit into `Data` (as discussed in our call). A factory to produce `HTMLMetadata` objects would be nice  ergonomic and preferable to directly instantiating objects (afaic). For the UI framework, we currently do not have the concept of a factory for simple data types used in the context of the UI framework. So, we can either introduce that factory in the UI framework, add the objects to the UI framework and instantiate them with `new` or add them to `Data` and the factory that is already located there. I would go with `Data` and its factory. For one it's a bit lazy (no need to create an additional factory) but I also see benefit to be able to talk about HTML metadata without requiring the whole UI framework.

We could then make the `Standard Page` accept `HTMLMetadata` and simply output it. This would be the simple part then =)

So regarding the way(s) forward, I see two options:
* Either you iterate on this with the feedback I have provided. I think it wouldn't be necessary to implement the complete `HTMLMetadata` hierarchy, but only the `OpenGraph` part that you require. We could then add the other variants as required.
* I would offer to create an alternative proposal myself, implementing the pragma directive. We could then directly compare both variants, maybe together with other interested parties (@Amstutz, @chfsx? @mjansenDatabay? @kergomard? ...)

There also is another approach which might be viable and  involves components of the UI framework. I would not suggest it really, though, because I'm not sure how viable it really is, but I want to mention it anyway, so other people could champion it. We could add a component `OpenGraphDescription` (solid naming pending...), maybe in a group `Metadata`. I see a lot more possibilities for non-technical descriptions and useful examples there. We would not have a representation of HTML metadata then, but a get a tangible component instead. This would be nice, but I would expect the requirement to represent metadata to arise at some point again. So, not really sure about this.

Best regards!
@klees (in agreement with @Amstutz)

Amstutz:
To keep to list of open PRs on my watchlist somewhat short/manageable. I close this for the moment. @thibsy Feel free to reopen if there is further points to be discussed.

PR #4696, opened 2022-06-27
[FIX] #33024: public profile link targets

thibsy:
Hi @mjansenDatabay, I think I might have found the issue with https://mantis.ilias.de/view.php?id=33024.

The link-target generation in `ilBuddySystemRelationsTableGUI::fillRow()` didn't route through `ilContactGUI` because it is a base class itself. I think previously the base classes were considered last when finding a path, whereas they are now considered first. Therefore the target got recognized as a base class right away instead of finding a way though parent classes. You can force ilCtrl to route through `ilContactGUI` by providing part of the path to the target class `ilPublicUserProfileGUI`. (Im not a fan of this but) ilCtrl will then automatically find a valid path that is connected to the first provided class and prepends the `ilDashboardGUI` node in this scenario. 

I also found a bug when adopting ilCtrlContext's parameters that occurred when the command-class is also a base class but shouldn't be treated as such. Therefore I left it within this PR too.

Kind regards!

### DIFF
--- a/Services/Contact/BuddySystem/classes/tables/class.ilBuddySystemRelationsTableGUI.php
+++ b/Services/Contact/BuddySystem/classes/tables/class.ilBuddySystemRelationsTableGUI.php
@@ -13,8 +13,7 @@
  * us at:
  * https://www.ilias.de
  * https://github.com/ILIAS-eLearning
- *
- *********************************************************************/
+ */
 
 /**
  * Class ilBuddyList
--- a/Services/Contact/BuddySystem/classes/tables/class.ilBuddySystemRelationsTableGUI.php
+++ b/Services/Contact/BuddySystem/classes/tables/class.ilBuddySystemRelationsTableGUI.php
@@ -218,7 +218,7 @@ class ilBuddySystemRelationsTableGUI extends ilTable2GUI
         if ((!$this->user->isAnonymous() && $public_profile === 'y') || $public_profile === 'g') {
             $this->ctrl->setParameterByClass(ilPublicUserProfileGUI::class, 'user', $a_set['usr_id']);
             $profile_target = $this->ctrl->getLinkTargetByClass(
-                ilPublicUserProfileGUI::class,
+                [ilContactGUI::class, ilPublicUserProfileGUI::class],
                 'getHTML'
             );
             $a_set['profile_link'] = $profile_target;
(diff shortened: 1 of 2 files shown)


Services/Contact/BuddySystem/classes/tables/class.ilBuddySystemRelationsTableGUI.php
```diff
@@ -13,8 +13,7 @@
  * us at:
  * https://www.ilias.de
  * https://github.com/ILIAS-eLearning
- *
```
mjansenDatabay, line 16: IMO this change is wrong, according to `\ILIAS\CI\Rector\ChangeLicenseHeader`

thibsy: You're right, but then `CI/Copyright-Checker/copyright-checker.sh` doesn't work as expected, right? because the licence header count's as valid.

mjansenDatabay: Hmmm, the used header looks identical in `CI/Copyright-Checker/copyright-checker.sh` and `\ILIAS\CI\Rector\ChangeLicenseHeader`. If this header is considered invalid, it is a bug :-).

Services/Contact/BuddySystem/classes/tables/class.ilBuddySystemRelationsTableGUI.php
```diff
@@ -13,8 +13,7 @@
...
- *********************************************************************/
+ */
```
mjansenDatabay, line 16:
```suggestion
 *********************************************************************/
```

Services/Contact/BuddySystem/classes/tables/class.ilBuddySystemRelationsTableGUI.php
thibsy: I'll look into it

thibsy:
This issue has been resolved by #4702 where the priority of a baseclass has been decreased.

Services/Contact/BuddySystem/classes/tables/class.ilBuddySystemRelationsTableGUI.php
thibsy: FYI @mjansenDatabay the copyright-checker will only test if the content is similar, the `*`, `/` and space characters will be removed before the comparison. It looks like that was intended =).

PR #4700, opened 2022-06-28
Allow using AUTO_INCREMENT for integer table columns

mBeym:
- Allow the use of the MySQL/MariaDB AUTO_INCREMENT feature.
- The `nextId` method now returns the next AUTO_INCREMENT value if it's defined, otherwise for compatibility the previous sequence number
- deprecated `createSequence` method

### DIFF
--- a/Services/Database/classes/PDO/FieldDefinition/class.ilDBPdoFieldDefinition.php
+++ b/Services/Database/classes/PDO/FieldDefinition/class.ilDBPdoFieldDefinition.php
@@ -58,7 +58,7 @@ abstract class ilDBPdoFieldDefinition
      */
     public array $allowed_attributes = [
         "text" => ["length", "notnull", "default", "fixed"],
-        "integer" => ["length", "notnull", "default", "unsigned"],
+        "integer" => ["length", "notnull", "default", "unsigned", "autoincrement"],
         "float" => ["notnull", "default"],
         "date" => ["notnull", "default"],
         "time" => ["notnull", "default"],
--- a/Services/Database/classes/PDO/class.ilDBPdoMySQL.php
+++ b/Services/Database/classes/PDO/class.ilDBPdoMySQL.php
@@ -147,6 +147,16 @@ abstract class ilDBPdoMySQL extends ilDBPdo
 
     public function nextId(string $table_name) : int
     {
+        $autoIncrementQueryResult = $this->queryF(
+            "SELECT `auto_increment` FROM INFORMATION_SCHEMA.TABLES WHERE table_name = %s",
+            ["text"],
+            [$table_name]
+        );
+        $autoIncrementValue = $this->fetchAssoc($autoIncrementQueryResult)["auto_increment"];
+        if ($autoIncrementValue && is_numeric($autoIncrementValue)) {
+            return (int) $autoIncrementValue;
+        }
+
         $sequence_name = $this->quoteIdentifier($this->getSequenceName($table_name), true);
         $seqcol_name = $this->quoteIdentifier('sequence');
         $query = "INSERT INTO $sequence_name ($seqcol_name) VALUES (NULL)";
--- a/Services/Database/interfaces/interface.ilDBInterface.php
+++ b/Services/Database/interfaces/interface.ilDBInterface.php
@@ -55,6 +55,11 @@ interface ilDBInterface
 
     public function addPrimaryKey(string $table_name, array $primary_keys) : bool;
 
+    /**
+     * @param $table_name
+     * @param int $start
+     * @deprecated use 'autoincrement' in $fields instead
+     */
     public function createSequence(string $table_name, int $start = 1) : bool;
 
     public function getSequenceName(string $table_name) : string;


mBeym:
Next step if this is accepted will be to replace all usages of createSequence with autoincrement as well as removing the need to use nextId()

mjansenDatabay:
Hi @mBeym,

thanks for this PR.

What happens if there are concurrent requests where `nextId` is called for the same table within the exact same millisecond? I doubt that this approach is `Race Condition Safe`. This mechanism only selects an existing field vaue from a central database table without guaranteeing a uniqueness.
This will result in mutliple/concurrent requests retrieving the **same** value when executing the same operation. Even if all primary keys (single field) in ILIAS have been migrated to `AUTO_INCREMENT`, this problem is not solved. The current state of this PR  provokes that every `INSERT INTO` has to be wrapped in a `try/catch` and maybe a `do/while`-loop to guarantee a successful `INSERT INTO`-operation.

IMO the sequence tables (with `AUTO_INCREMENT`) **currently** used in ILIAS have an **advantage** over `AUTO_INCREMENT` primary keys (applied to the actual data table).

Currently an `INSERT` in ILIAS looks like this:

```php
$id = $db->nextId('table');

$db->manipulateF('INSERT INTO table ...', /*...*/, /*....*/); 
// ... or ...
$db->manipulate('INSERT INTO table ...'); 
// ... or ...
$db->insert('table', /*...*/);
```

The id is retrieved first and the uniqueness is guaranteed (more or less, but IMO a lot more reliably than in your approach). Especially if you are using (or plan to use) entities and repositories in your domain, you could refactor your code (like I'll outline in examples below) and possibly also use other `Identity Implementations` like (if relevant for the domain) `UUID`, without changing much code in ILIAS.

You can call this/the current approach: `Generating the identifier early` (there might be four different approaches where such an identity can be created)

Your future outlook/next steps will probably require code to look/be migrated like the following example. We will have to rely on the `LAST_INSERT_ID()` call if we like to use `AUTO_INCREMENT` integers on the actual database tables.

```php
interface MembershipRepository
{
    public function create(Membership $membership) : int;

    //...
}

class MembershipDatabaseRepository implements MembershipRepository
{
    private ilDbInterface $db;

    //...

    public function create(Membership $membership) : int
    {
        $this->db->manipulate('INSERT INTO table ...');
        
        $result = $this->query('SELECT LAST_INSERT_ID() AS pk');
        $value = $result->fetchObject()->pk;

        return (int) $value;
    }

    //...
}

class MembershipService
{
    private MembershipRepository $repo;

    //...

    public function createMembership(/*Command or Request or something similar*/ $request) : void
    {
        $membership = new Membership(/*...*/);

        $last_insert_id = $this->repo->create($membership);

        $membership->setId($last_insert_id);

        //... 
    }

    //...
}
```

Problems:
- The id (identity) of the `Membership` entity is only known to us when the repository has created the membership. That already reveals an implementation detail, namely that the repository uses `AUTO_INCREMENT` integers for the primary key.
- Beside this the bigger problem is, that it will be hard to change the implementation (think of a big project like ILIAS with different components and domains) or migrate to something I'll outline below.
- Furthermore the `Membership` entity could be supposed to be complete by creation, which is not possible with your approach.

An IMO better approach would be (**there might be other variants where the identity is not provided by repositories**):

```php
final class MembershipId // There could be also an interface
{
    private int $id;

    public function __construct(int $id)
    {
        $this->id = $id;
    }

    //...
}

// ... or ...

final class MembershipId // There could be also an interface
{
    private \ILIAS\Data\UUID\Uuid $id;

    public function __construct(\ILIAS\Data\UUID\Uuid $id)
    {
        $this->id = $id;
    }

   //...
}

interface MembershipRepository
{
    public function nextId() : MembershipId; //... Use one of implementations outlined above, we don't care

    public function create(Membership $membership) : void;

    //...
}

class MembershipService
{
    private MembershipRepository $repo;

    //...

    public function createMembership(/*Command or Request or something similar*/ $request) : void
    {
        $membership_id = $this->repo->nextId(); // Could be also provided by the application, the client, ....

        $membership = new Membership($membership_id, /*...*/);

        // Maybe the the complete entitry to some message queues etc.

        $this->repo->create($membership);

        //... 
    }

    //...
}
```
In this case, ...

- ... the `Membership`  entity is complete when created
- ... the repository does not reveal the type of identity
- ... and integer based (`AUTO_INCREMENT`) implementation could be replaced by a `\ILIAS\Data\UUID\Uuid` implementation
- ... existing ILIAS code does not need to be migrated that much to achieve such kind of decoupling

Best regards,
@mjansenDatabay

mBeym:
@mjansenDatabay How about.
1. Retrieve next id from AUTO_INCREMENT.
2. Updating the AUTO_INCREMENT value +1 using Alter Table.
3. Returning nextId.

*Optionally, add a second bool parameter too to nextId() that will make it not increment the AUTO_INCREMENT value and only returns the next one

This way If you don't need the id, no call to nextID is required anymore 
but if you do need it because another query requires the id of the previous query, it's still possible to get it

PR #4704, opened 2022-06-29
Forum: Bugfix 0033198: first posting missing when copying a forum

fhelfer:
This fixes https://mantis.ilias.de/view.php?id=33198

### DIFF
--- a/Modules/Forum/classes/class.ilForumTopic.php
+++ b/Modules/Forum/classes/class.ilForumTopic.php
@@ -254,13 +254,16 @@ class ilForumTopic
     */
     public function getFirstPostId()
     {
+        $this->db->setLimit(1);
         $res = $this->db->queryf(
             '
 			SELECT * FROM frm_posts_tree 
 			WHERE thr_fk = %s
-			AND parent_pos != %s',
-            array('integer', 'integer'),
-            array($this->id, '0')
+			AND parent_pos != %s
+			AND depth = %s
+			ORDER BY rgt DESC',
+            array('integer', 'integer', 'integer'),
+            array($this->id, '0', 2)
         );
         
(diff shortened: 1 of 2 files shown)


Modules/Forum/classes/class.ilObjForum.php
```diff
@@ -579,6 +580,27 @@ public function cloneObject($a_target_id, $a_copy_id = 0, $a_omit_tree = false)
                 1,
                 false
```
mjansenDatabay: @fhelfer Could it be sufficient to set this to `true` here?

fhelfer: setting this to true will still only copy the first post

mjansenDatabay: Hmmm, I checked this with the following result:

- If set to `true`, there is exactly one visible posting in the copied forum (+ the invisible root posting). This is the desired behaviour.

Maybe we should have a chat about this :-).

Modules/Forum/classes/class.ilForumTopic.php
```diff
@@ -258,9 +258,9 @@ public function getFirstPostId()
             '
 			SELECT * FROM frm_posts_tree 
 			WHERE thr_fk = %s
-			AND parent_pos = %s',
+			AND parent_pos != %s',
             array('integer', 'integer'),
-            array($this->id, '1')
+            array($this->id, '0')
```
mjansenDatabay: Hmmm, IMO we have to use the `depth` and `rgt` (`ORDER BY rgt DESC`) field in the condition and we should use a `$this->db->setLimit(1);`

Please consider that this might result in an empty SQL result, which has to be handled by the code below in this PR.

mjansenDatabay (approved):
Thx @fhelfer, LGTM now!

PR #4711, opened 2022-07-01
Forum: Bugfix 0033198: first posting missing when copying a forum

fhelfer:
This fixes https://mantis.ilias.de/view.php?id=33198

### DIFF
--- a/Modules/Forum/classes/class.ilForumPost.php
+++ b/Modules/Forum/classes/class.ilForumPost.php
@@ -113,7 +113,7 @@ class ilForumPost
                 ]
             );
 
-            if ($this->objThread->getPostRootId() === $this->id) {
+            if ($this->objThread->getFirstVisiblePostId() === $this->id) {
                 $this->objThread->setSubject($this->subject);
                 $this->objThread->update();
                 $this->objThread->reload();
--- a/Modules/Forum/classes/class.ilForumPost.php
+++ b/Modules/Forum/classes/class.ilForumPost.php
@@ -113,7 +113,7 @@ class ilForumPost
                 ]
             );
 
-            if ($this->objThread->getFirstPostId() === $this->id) {
+            if ($this->objThread->getPostRootId() === $this->id) {
                 $this->objThread->setSubject($this->subject);
                 $this->objThread->update();
                 $this->objThread->reload();
--- a/Modules/Forum/classes/class.ilForumTopic.php
+++ b/Modules/Forum/classes/class.ilForumTopic.php
@@ -214,6 +214,7 @@ class ilForumTopic
 
     public function getFirstPostId() : int
     {
+        $this->db->setLimit(1);
         $res = $this->db->queryF(
             'SELECT * FROM frm_posts_tree WHERE thr_fk = %s AND parent_pos != %s AND depth = %s ORDER BY rgt DESC',
             ['integer', 'integer', 'integer'],
@@ -285,6 +286,7 @@ class ilForumTopic
 
     public function getFirstPostNode(bool $isModerator = false, bool $preventImplicitRead = false) : ilForumPost
     {
+        $this->db->setLimit(1);
         $res = $this->db->queryF(
             '
 			SELECT *
--- a/Modules/Forum/classes/class.ilForumTopic.php
+++ b/Modules/Forum/classes/class.ilForumTopic.php
@@ -231,7 +231,7 @@ class ilForumTopic
     {
         $this->db->setLimit(1);
         $res = $this->db->queryF(
-            'SELECT * FROM frm_posts_tree WHERE thr_fk = %s AND parent_pos != %s AND depth = %s ORDER BY rgt DESC',
+            'SELECT pos_fk FROM frm_posts_tree WHERE thr_fk = %s AND parent_pos != %s AND depth = %s ORDER BY rgt DESC',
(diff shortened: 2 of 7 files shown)


Modules/Forum/classes/class.ilForumTopic.php
```diff
@@ -215,13 +215,13 @@ public function reload() : bool
     public function getFirstPostId() : int
     {
         $res = $this->db->queryF(
-            'SELECT * FROM frm_posts_tree WHERE thr_fk = %s AND parent_pos = %s',
-            ['integer', 'integer'],
-            [$this->id, 1]
+            'SELECT * FROM frm_posts_tree WHERE thr_fk = %s AND parent_pos != %s AND depth = %s ORDER BY rgt DESC',
```
mjansenDatabay: - Shouldn't we add an `$this->db->setLimit(1);` here as well?

Modules/Forum/classes/class.ilForumTopic.php
```diff
@@ -285,15 +286,18 @@ public function countActivePosts(bool $ignoreRoot = false) : int
...
     {
+        $this->db->setLimit(1);
         $res = $this->db->queryF(
             '
 			SELECT *
 			FROM frm_posts 
 			INNER JOIN frm_posts_tree ON pos_fk = pos_pk
-			WHERE parent_pos = %s
-			AND thr_fk = %s',
+			WHERE parent_pos != %s
```
mjansenDatabay, line 335: IMO This is wrong in this case, I checked the usages. We should rename this method to `getPostRootNode`. The `depths` MUST be 1 and the `parent_pos` MUST equal `0`.

Modules/Forum/classes/class.ilForumTopic.php
```diff
@@ -214,14 +214,15 @@ public function reload() : bool
 
     public function getFirstPostId() : int
```
mjansenDatabay: Maybe we should rename it to `getFirstVisiblePostId` (see comment below).

mjansenDatabay (changes requested):

Modules/Forum/classes/class.ilForumTopic.php
fhelfer: we could add a optional parameter for both methods ($exclude_root_node = false)

mjansenDatabay: Hmmm, I am personally not a fan of introducing parameters (we do it to often in legacy code).

> Don't use flag arguments. Split method into several independent methods that can be called from the client without the flag.

Modules/Forum/classes/class.ilForumTopic.php
```diff
@@ -212,16 +212,32 @@ public function reload() : bool
...
 
-    public function getFirstPostId() : int
+    public function getPostRootId() : int
     {
+        $this->db->setLimit(1);
         $res = $this->db->queryF(
-            'SELECT * FROM frm_posts_tree WHERE thr_fk = %s AND parent_pos = %s',
-            ['integer', 'integer'],
-            [$this->id, 1]
+            'SELECT * FROM frm_posts_tree WHERE thr_fk = %s AND parent_pos = %s AND depth = %s ORDER BY rgt DESC',
```
mjansenDatabay:
```suggestion
            'SELECT pos_fk FROM frm_posts_tree WHERE thr_fk = %s AND parent_pos = %s AND depth = %s ORDER BY rgt DESC',
```

Modules/Forum/classes/class.ilForumPost.php
```diff
@@ -113,7 +113,7 @@ public function update() : bool
                 ]
             );
 
-            if ($this->objThread->getFirstPostId() === $this->id) {
+            if ($this->objThread->getPostRootId() === $this->id) {
```
mjansenDatabay: The first visible posting id should be used here.

mjansenDatabay (changes requested):

Modules/Forum/classes/class.ilForumTopic.php
```diff
@@ -212,16 +212,32 @@ public function reload() : bool
...
+            return (int) $row->pos_fk ?: 0;
+        }
+        return 0;
+    }
+
+    public function getFirstVisiblePostId() : int
+    {
+        $this->db->setLimit(1);
+        $res = $this->db->queryF(
+            'SELECT * FROM frm_posts_tree WHERE thr_fk = %s AND parent_pos != %s AND depth = %s ORDER BY rgt DESC',
```
mjansenDatabay:
```suggestion
            'SELECT pos_fk FROM frm_posts_tree WHERE thr_fk = %s AND parent_pos != %s AND depth = %s ORDER BY rgt DESC',
```

Modules/Forum/classes/class.ilForumTopic.php
```diff
@@ -997,7 +1041,7 @@ public function updateThreadTitle() : void
             ['thr_pk' => ['integer', $this->getId()]]
         );
 
-        $first_node = $this->getFirstPostNode();
+        $first_node = $this->getPostRootNode();
```
mjansenDatabay:
```suggestion
        $first_node = $this->getFirstVisiblePostNode();
```

mjansenDatabay (approved):
Thx @fhelfer

PR #4713, opened 2022-07-04
ilCtrl: Clone structure to prevent temporary parameters are copied fo…

mjansenDatabay:
…r each clone

Mantis Issue: https://mantis.ilias.de/view.php?id=33091

### DIFF
--- a/Services/UICore/classes/class.ilCtrl.php
+++ b/Services/UICore/classes/class.ilCtrl.php
@@ -126,6 +126,11 @@ class ilCtrl implements ilCtrlInterface
         $this->component_factory = $component_factory;
     }
 
+    public function __clone()
+    {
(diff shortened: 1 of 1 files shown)


thibsy:
Thx for looking into this @mjansenDatabay! May I ask why there is a need to clone ilCtrl? (nevertheless, I understand this must be handled, just curious)

mjansenDatabay:
The need is, that `\ilMailFormCall::getTargetUrl` builds an URL to the mail system, which contains all query parameters relevant for the mail system, but also appends a backlink (created with `ilCtrl`) to the consuming GUI class (the referrer). To append the parameters I used `\ilCtrlInterface::setParameter`, but this call should not have any side-effects to the `ilCtrl` used when presenting the current view. Therefore I cloned ilCtrl.

PR #4718, opened 2022-07-04
Object/Setup: use ilDB from environment instead of global DIC

nhaagen:
err_setup


mjansenDatabay:
Today the TB decided to merge this PR. Thx @nhaagen !

PR #4721, opened 2022-07-05
Forum: Feature 0029875: Course/Group title in subject of forum mail notifications (trunk)

fhelfer:
This fixes: https://mantis.ilias.de/view.php?id=29875
(language have to be reloaded)

### DIFF
--- a/Modules/Forum/classes/Notification/class.ilForumMailEventNotificationSender.php
+++ b/Modules/Forum/classes/Notification/class.ilForumMailEventNotificationSender.php
@@ -429,6 +429,8 @@ class ilForumMailEventNotificationSender extends ilMailNotification
         $body .= "\n\n";
         $body .= $this->getLanguageText('thread') . ": " . $this->provider->getThreadTitle();
         $body .= "\n\n";
+        $body .= $this->getLanguageText($this->provider->getTopItemType()) . ": " . $this->provider->getTopItemTitle();
+        $body .= "\n\n";
         $body .= $this->getLanguageText($action) . ": \n------------------------------------------------------------\n";
 
         $body .= $this->getLanguageText('author') . ": " . $this->provider->getPostUserName($this->getLanguage());
--- a/Modules/Forum/classes/Notification/class.ilForumMailEventNotificationSender.php
+++ b/Modules/Forum/classes/Notification/class.ilForumMailEventNotificationSender.php
@@ -500,7 +500,7 @@ class ilForumMailEventNotificationSender extends ilMailNotification
     {
         $container_text = '';
         if ($this->provider->providesClosestContainer()) {
-            $container_text = " (" . $this->getLanguageText('obj_' . $this->provider->closestContainer()->getType()) .
+            $container_text = " (" . $this->getLanguageText('frm_noti_obj_' . $this->provider->closestContainer()->getType()) .
                 " \"" . $this->provider->closestContainer()->getTitle() . "\")";
         }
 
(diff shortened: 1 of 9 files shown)


Modules/Forum/classes/class.ilObjForumNotificationDataProvider.php
```diff
@@ -264,6 +278,19 @@ private function readForumData() : void
         $row = $this->notificationCache->fetch($cacheKey);
         $this->forum_id = (int) $row['top_pk'];
         $this->forum_title = (string) $row['top_name'];
+
+        $forum = ilObjectFactory::getInstanceByObjId($this->getObjId());
```
mjansenDatabay: We should extract this code to a separate base class/or a trait/or even better: some type injected as dependency.

Is it ensured that these (expensive) calls are not executed multiple times during a bulk processing of forum notifications? If not, a cache should be introduced (See: `ilForumNotificationCache`).

Modules/Forum/classes/Notification/class.ilForumMailEventNotificationSender.php
```diff
@@ -427,8 +427,12 @@ private function createMailBodyText(
         $body .= "\n\n";
         $body .= $this->getLanguageText('forum') . ": " . $this->provider->getForumTitle();
         $body .= "\n\n";
+        $body .= $this->getLanguageText($this->provider->getTopItemType()) . ": " . $this->provider->getTopItemTitle();
```
mjansenDatabay: We should not append the body if we could not find a relevant parent object. This is valid for all similar usages (budy/subject) and also effects the design of the language variables.

mjansenDatabay (changes requested):

Modules/Forum/classes/class.ilForumCronNotification.php
```diff
@@ -231,6 +233,15 @@ public function sendCronForumNotification(ilDBStatement $res, int $notification_
 
             $row['ref_id'] = $ref_id;
 
+            $top_item_ref_id = $this->tree->getParentId($ref_id);
```
smeyer-ilias: Have you considered to use $tree->checkForParentType(..., 'crs|grp') here or should only direct child elements be supported?

Modules/Forum/classes/class.ilObjForumNotificationDataProvider.php
fhelfer: Added top_item to notificationCache

Modules/Forum/classes/class.ilObjForumNotificationDataProvider.php
```diff
@@ -258,12 +272,26 @@ private function readForumData() : void
 
             $row = $this->db->fetchAssoc($result);
 
+            $top_item_crs_ref = $this->tree->checkForParentType($this->getRefId(), 'crs');
```
mjansenDatabay: We should not execute both operations here. We should check for a `crs` parent first. If a `crs` could **not** be found, we should check for a `grp`.

I suggest to extract this caching into a new class or a trait, nesting `if` statements should be avoided. We should work with early returns here.

Modules/Forum/classes/class.ilForumCronNotification.php
```diff
@@ -230,6 +232,16 @@ public function sendCronForumNotification(ilDBStatement $res, int $notification_
             }
 
             $row['ref_id'] = $ref_id;
+            $top_item_crs_ref = $this->tree->checkForParentType($ref_id, 'crs');
```
mjansenDatabay: We should not execute both operations here. We should check for a `crs` parent first. If a `crs` could **not** be found, we should check for a `grp`.

IMO we still need caching here. I suggest to extract this into a new method, nesting `if` statements should be avoided. We should work with early returns here.

mjansenDatabay:
@fhelfer Please check my latest changes in https://github.com/ILIAS-eLearning/ILIAS/pull/4721/commits/b1948473a38c4b64501c9f930e6a3c13410871b14 and test this thoroughly

fhelfer:
> @fhelfer Please check my latest changes in https://github.com/ILIAS-eLearning/ILIAS/pull/4721/commits/b1948473a38c4b64501c9f930e6a3c13410871b14 and test this thoroughly

works

mjansenDatabay (approved):

PR #4731, opened 2022-07-06
Fix Empty Headings in Results Table

kergomard:
This fixes: https://mantis.ilias.de/view.php?id=32258

An Empty Page-Heading coming from the ILIAS Page-Editor may force an Empty-Tag in certain circumstances. Maybe this is an overcautious fix as the whole if-statement might be unnecessary.

Should be cherry-picked to trunk if accepted.


Amstutz:
Thx a lot for fixing this a11y issue. I will pick it to trunk as well. kudos.

Amstutz:
picked to trunk

PR #4735, opened 2022-07-07
Forum: Refactor 0030928: Remove obsolete move Draft events

fhelfer:
This fixes: https://mantis.ilias.de/view.php?id=30928

### DIFF
--- a/Modules/Forum/classes/class.ilForum.php
+++ b/Modules/Forum/classes/class.ilForum.php
@@ -1792,16 +1792,6 @@ class ilForum
             $targetThreadForMerge->reopen();
         }
 
-        $this->event->raise(
-            'Modules/Forum',
-            'mergedThreads',
-            [
-                'obj_id' => $this->getForumId(),
-                'source_thread_id' => $sourceThreadForMerge->getId(),
-                'target_thread_id' => $targetThreadForMerge->getId()
-            ]
-        );
-
         $this->deletePost($sourceThreadRootArray, false);
     }
 }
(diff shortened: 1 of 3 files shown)


Modules/Forum/classes/class.ilForum.php
```diff
@@ -1792,16 +1792,6 @@ public function mergeThreads(int $source_id, int $target_id) : void
             $targetThreadForMerge->reopen();
         }
 
-        $this->event->raise(
```
mjansenDatabay: This `case` MUST NOT be deleted IMO. We need to refresh the learning progress if two threads are combined into one thread.

Modules/Forum/classes/class.ilForumAppEventListener.php
```diff
@@ -43,25 +43,6 @@ public static function handleEvent(string $a_component, string $a_event, array $
         switch ($a_component) {
             case 'Modules/Forum':
                 switch ($a_event) {
-                    case 'mergedThreads':
```
mjansenDatabay: This `case` MUST NOT be deleted IMO. We need to refresh the learning progress if two threads are combined into one thread.

Only the draft related code can maybe be removed (that's what the discussion in the Mantis issue is about).

mjansenDatabay (changes requested):
IMO the event in `\ilForum::moveThreads` has to be changed to `movedThreads`.

mjansenDatabay (approved):

PR #4742, opened 2022-07-08
Fixed bug #33289: ILIAS 7 XSS in erweiterter Sprachverwaltung

c-knof:


mjansenDatabay (commented):
LGTM in general (in regards to filtering the input), but I am wondering whether there might be other problematic characters we'll have to substitute.

Maybe we'll have to check the diff of the original file contents and the contents filtered by `ilUtil::stripSlashes()` to find other problematic chars. If there are none: Perfect :-)!

mjansenDatabay (approved):

katringross:
@mjansenDatabay I just checked the differences, but did not find other problematic chars.

PR #4746, opened 2022-07-08
UserSearch: 'support' searchable UDF WYSIWYG fields

nhaagen:
https://mantis.ilias.de/view.php?id=33027
https://mantis.ilias.de/view.php?id=32989


smeyer-ilias:
Thanks @nhaagen

PR #4750, opened 2022-07-11
UI item: move property key & value to same col

catenglaender:
See Mantis Issue https://mantis.ilias.de/view.php?id=32750

This is a suggestion to move key and values to the same column inside the UI item, to make the connection between them unmistakably clear.

Before
Bildschirmfoto 2022-05-24 um 15 07 40

After
2022-07-11 10_20_30-Window

The colon and spacing ": " is being added as a :after pseudo element.

This is a fully functional minimal implementation, but there are some points left to discuss if there is an interest to commit to this concept:
* Line height could be more clear as well: The lines of long properties breaking to the next line could be closer together, vertical spacing to the next key-value pair could be larger.
* There is something off with shy buttons (btn-link in Bootstrap). They don't seem to be on the exact same base-line as normal text. I suspect that this is an issue beyond the UI item.
* I think Bootstrap recommends filling columns until they wrap to the next row automatically as best practice instead of actually opening up a multiple row divs in succession. Not sure if it's worth the hassle to change it, because in this case it visually gives the desired result.

All the best,
Ferdinand

### DIFF
--- a/src/UI/templates/default/Item/item.less
+++ b/src/UI/templates/default/Item/item.less
@@ -45,6 +45,9 @@
 		font-size: @font-size-small;
 		color: @il-text-light-color;
 		overflow: hidden;
+		&:not(:empty):after {
+			content: ": "
+		}
 	}
 
 	.il-item-property-value {
--- a/src/UI/templates/default/Item/tpl.item_notification.html
+++ b/src/UI/templates/default/Item/tpl.item_notification.html
@@ -20,17 +20,11 @@
 				<hr class="il-item-divider">
 				<!-- BEGIN property_row -->
 				<div class="row il-item-properties">
-					<div class="col-md-12">
-						<div class="row">
-							<div class="col-sm-3 il-item-property-name">{PROP_NAME_A}</div>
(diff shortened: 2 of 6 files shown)


catenglaender:
Ah, I see the test fails because of a changed HTML structure. I haven't actually dealt with this before, so I would need a moment to look into that.

catenglaender:
Almost forgot about the notification item. Now, it also has key and value directly next to each other since it suffered from the same confusing spacing.

When taking care of the the unit tests I was a bit surprised that empty {PROP_NAME_B} & {PROP_VAL_B} still come with their row structure (which is totally logical and expected because that structure is in the HTML template). If we eventually render a whole magazine with let's say 30 items that only have the one property A, that would be 3 * 30 = 90 lines of DOM code with nothing in it for the empty property B structure. Since A and B are now exactly the same, I don't think this is a necessity anymore. But since other projects around the item and its properties are peaking up on the horizon, I think this is an optimization for later.

- [x]  Unit Test should pass now.

catenglaender:
@Amstutz Ready for merge?

Amstutz:
Thx a lot!

PR #4751, opened 2022-07-11
Fixes a few more access issues in ADT and Calendar

kergomard:
https://mantis.ilias.de/view.php?id=33078
https://mantis.ilias.de/view.php?id=33077
https://mantis.ilias.de/view.php?id=33110


smeyer-ilias (approved):
@kergomard Thanks for fixing this.

PR #4754, opened 2022-07-11
NPM: Update TinyMCE

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=33452

The `Diff` is quite big because someone changed the NPM files with an outdated version.


kergomard:
Thank you very much! We might think about adding a license header in the future (as this is our file), but this is not for now.

PR #4755, opened 2022-07-11
Page Layout: Sticky Breadcrumb Bar #32935

catenglaender:
After the location of the breadcrumbs has been changed, the need for a sticky breadcrumb bar was reported in this Mantis issue: https://mantis.ilias.de/view.php?id=32935

## Sticky after scroll

To create the desired effect I used the CSS attribute position: sticky, no JS. Browser support is good (96% of all web users): https://caniuse.com/css-sticky

The breadcrumbs remain in the document flow until scrolling begins, so no padding under the breadcrumbs is necessary.

## Cleaner page structure

While implementing the sticky breadcrumbs I noticed a few leftovers from previous page structures that seem to be no longer in use (and in the worst case block simple solutions):

There was a commented out HTML Code of a legacy footer in tpl.page_content.html which I removed, because it could potentially mess up div spacekeeping, grid and flexbox behavior. If this is some sort of templating language that shouldn't be removed, please let me know.

Breadcrumbs, mainspacekeeper and footer were nested in a nameless div that once played a big role in pushing the footer down and anchoring the mode info. However, it seems to no longer serve both of these purposes. Removing this div makes the implementation of the breadcrumb bar (and any other objects) above the mainspacekeeper easier.

The footer is now being pushed down with flexbox instead of a nested CSS grid, which is a common practice for areas inside a CSS grid that only require a one directional column.

This approach should handle additional "unexpected" divs way more gracefully than the CSS grid which would "count" the divs to assign heights.

## Testing

This implementation of the breadcrumb bar can be tested on this dev server provided by CaT: https://dev.cate-tms.de/febreadcrumbs/ilias.php?baseClass=ilrepositorygui&reloadpublic=1&cmd=&ref_id=1

Some magazine pages are visible without login, where the sticky breadcrumb effect can be seen in actions. Let me know if you require personal login credentials for testing purposes.

I am happy to answer any questions! 🙂

### DIFF
--- a/src/UI/templates/default/Layout/standardpage.less
+++ b/src/UI/templates/default/Layout/standardpage.less
@@ -235,16 +235,9 @@ main {
 	&:focus-visible {
 		outline: none;
 	}
-	display: grid;
-	grid-template-rows: 1fr auto;
-	grid-column: 2;
-	grid-row: 3;
+	display: flex;
+	flex-flow: column;
 	overflow: auto;
-	> div {
-		// contains breadcrumbs, mainspacekeeper of content
-		display: flex;
-		flex-flow: column;
-	}
 	#mainspacekeeper {
 	    flex-grow: 1; // stretches content from top to footer
     	width: 100%;
--- a/src/UI/templates/default/Layout/standardpage.less
+++ b/src/UI/templates/default/Layout/standardpage.less
@@ -113,6 +113,8 @@ header {
 
 //Breadcrumbs
 .breadcrumbs {
+	position: sticky;
+	top: 0;
 	align-items: center;
 	background-color: @il-standard-page-breadcrumbs-bg-color;
 	display: flex;
--- a/src/UI/templates/default/Layout/tpl.standardpage.html
+++ b/src/UI/templates/default/Layout/tpl.standardpage.html
@@ -65,15 +65,13 @@
 
 		<!-- html5 main-tag is not supported in IE / div is needed -->
 		<main class="il-layout-page-content">
-			<div>
-				<!-- BEGIN content_breadcrumbs -->
-				<div class="breadcrumbs">
-					{BREADCRUMBS}
-				</div>
-				<!-- END content_breadcrumbs -->
-
-				{CONTENT}
+			<!-- BEGIN content_breadcrumbs -->
+			<div class="breadcrumbs">
+				{BREADCRUMBS}
 			</div>
+			<!-- END content_breadcrumbs -->
+
+			{CONTENT}
 
 			<!-- BEGIN pagefooter -->
 			<footer role="contentinfo">
--- a/src/UI/templates/default/MainControls/mode_info.less
+++ b/src/UI/templates/default/MainControls/mode_info.less
@@ -22,15 +22,6 @@
 }
 
 
-//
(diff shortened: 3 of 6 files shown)


Amstutz:
Thx a lot for the excellent work and description. Special thx a nd kudos for cleaning up the CSS and DOM structure. 

There is one thing whicht needs to be adapted:
- [ ] Adapt Unit Tests: Unit Tests need adaptation; 4 Unit Tests currently fail. See tests/UI/Component/Layout/Page/StandardPageTest.php . Adapt the html to fit your proposed changes. Note that you can run the unit test to check if they work with running
`
php libs/composer/vendor/phpunit/phpunit/phpunit tests/ILIASSuite.php
`
to check the complete suite of ILIAS (takes some time) OR
`
php libs/composer/vendor/phpunit/phpunit/phpunit tests/UI/Component/Layout/Page/StandardPageTest.php
`
to only check to affected class.

thx a lot for all your work!
@amstutz

Amstutz (changes requested):
:point_up_2:

oliversamoila:
Thanks @catenglaender 

Under Chrome and Firefox, the adjustments work very well.
Under Safari, a different rendering of the padding-left value with 20px takes place than in the above mentioned browsers. 
» class: .breadcrumb_wrapper .breadcrumb

If this padding-left value is set to 80px, the breadcrumb is correctly positioned in Safari, but not in the other browsers.
80px is the width of the mainbar.

Sorry, but I don't know what needs to be corrected here.

Bildschirmfoto 2022-07-19 um 22 15 24
Bildschirmfoto 2022-07-19 um 22 14 58

catenglaender:
Very interesting issue with Safari not handling this very specific combination of `text-overflow: ellipsis` and `overflow: hidden` very well. Safari seems to let syllables flow over the padding, while Chrome and Firefox treat the padding as a strictly protected area that nothing can reach into. This issue exists since ILIAS 6, but it is less notable there because there is enough room to the left for a syllable or two to grow out of the container.

Currently, the only option I see is to use margin instead of padding as Safari seems to cut off the container beyond the margin like all the other browsers do. However, there will be moments (in Safari) when the syllables aren't broken so the container edge cuts into the word with the "..." not being visible. This is how that would look like:

Screen Shot 2022-07-25 at 3 21 08 PM

Maybe a gradient to white can further make the strict cut off more pleasant? Chrome and Firefox rendering output will visually remain unchanged and still display the "..." at the very left without cutting into it.

If we want to have a breadcrumb shortening that is rendered exactly the same in Safari, Chrome and Firefox, we might have to move away from the ltr ellipsis trick for something like a fade to white. I believe that the overflow into padding in Safari in this very specific case is a bug that could be reported to the webkit community.

I wonder if this issue deserves its own ticket and PR.

catenglaender:
@Amstutz Unit Tests are now passed. I would suggest we solve the Safari issue in its own Mantis ticket and separate PR because I think it needs a deliberate decision. What do you think?

Amstutz:
Thx a lot @catenglaender and thx @oliversamoila for testing. I would say we merge this at this state, so later PRs solving the safari issues can build up on this and also to avoid further conflictsl.

PR #4757, opened 2022-07-12
fixes 21972 by adding adding didactic templates to session settings

schmitz-ilias:
@smeyer-ilias This PR fixes [21972](https://mantis.ilias.de/view.php?id=21972) by adding the option to choose a session type to the settings tab of sessions.


smeyer-ilias (approved):
Thanks @schmitz-ilias 
I 'll will merge it to R6 and R7 
Regarding the trunk integration I will contact you,

PR #4758, opened 2022-07-13
  fixes 21972 by adding adding didactic templates to session settings (trunk)

schmitz-ilias:
@smeyer-ilias This PR integrates #4757 into the trunk.


smeyer-ilias (approved):
Thanks @schmitz-ilias

PR #4767, opened 2022-07-14
Makes some DataCollections with comments work

kergomard:
This is one possible fix for: https://mantis.ilias.de/view.php?id=32818


Amstutz:
thx a lot! Picked to trunk an 8 as well.

PR #4768, opened 2022-07-15
Fix few more issues in export and MetaData

kergomard:
Hi @smeyer-ilias 

Just a few more small things I stumbled over, when trying to fix an error in and export a DataCollection:
https://mantis.ilias.de/view.php?id=33398
https://mantis.ilias.de/view.php?id=33077 (see comment)
And an issue when exporting DataCollections.

Thanks and best,
@kergomard


smeyer-ilias (approved):
Thanks for the fixes.

PR #4772, opened 2022-07-15
search: ctrl comment

smeyer-ilias:
It is not necessary to merge this PR. 
I only created it to discuss changes in ilCtrl. 

The problem: 

- the main menu search creates a link to the "last search result" (./ilias.php?baseClass=ilSearchControllerGUI)
- ilSearchController forwards to the last search presentation class (e.g ilLuceneSearchGUI, ilSearchGUI,...)
- without the new (deprecated) call of setCmdClass() a ctrl error is triggered "ilCtrl cannot find a path for 'ilobjectcopygui' that reaches 'ilSearchControllerGUI'"

Prior releases of ilCtrl stored the "forward class" in ilCtrl->current_node, this seems to be not the case anymore.

I suspect that similar errors appear in other areas. E.g in the calendar service. 

How do we proceed here?

### DIFF
--- a/Services/Search/classes/class.ilSearchControllerGUI.php
+++ b/Services/Search/classes/class.ilSearchControllerGUI.php
@@ -128,7 +128,7 @@ class ilSearchControllerGUI implements ilCtrlBaseClassInterface
                 // no break
             default:
                 $search_gui = new ilSearchGUI();
-                $this->ctrl->setCmdClass(ilSearchGUI::class);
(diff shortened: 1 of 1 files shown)


thibsy:
Hi @smeyer-ilias and thanks for opening this discussion!

I assume your GUI is able to create valid link targets to the corresponding (searched) GUIs initially, right? If so, wouldn't it be possible to remember the last link-target instead of just the target-class?

Kind regards!

thibsy:
Hi @smeyer-ilias

I close this PR because IMO this issue should be addressed by a different approach. I won't be adding this automatic node-appendage-mechanism again because I really wan't to enforce some more static behaviour. Thanks for your understanding.

Feel free to contact me if you want to discuss alternative solutions.

Kind regards!

PR #4778, opened 2022-07-15
Bower: Remove moment.js & datetimepicker from bower (release_8)

fhelfer:
depends on #4773


mjansenDatabay:
@smeyer-ilias Any objections?

smeyer-ilias:
Thanks @fhelfer

PR #4779, opened 2022-07-15
Bower: Remove moment.js & datetimepicker from bower (release_7)

fhelfer:
depends on #4775


mjansenDatabay:
@smeyer-ilias Any objections?

smeyer-ilias:
Thanks, @fhelfer

PR #4786, opened 2022-07-15
[FIX] #32961: Input\File now uses default file-size and can render ex…

thibsy:
Hi @Amstutz and @klees!

I have fixed an Issue in the ui-file-input I recently refactored, also see https://mantis.ilias.de/view.php?id=32961.

One thing I'm not quite fond of is the change to the `FileUploadHelper` trait, whose property $max_file_size must be set in the constructor programatically. Because the trait is used in both, inputs and dropzones, I cannot implement the constructor in the trait. Should I move the property completely out of the trait or is the current attempt readable enough?

Thanks for your opinion, best regards!

### DIFF
--- a/Modules/IndividualAssessment/test/Settings/ilIndividualAssessmentInfoSettingsTest.php
+++ b/Modules/IndividualAssessment/test/Settings/ilIndividualAssessmentInfoSettingsTest.php
@@ -1,6 +1,19 @@
 <?php declare(strict_types=1);
 
-/* Copyright (c) 2021 - Daniel Weise <daniel.weise@concepts-and-training.de> - Extended GPL, see LICENSE */
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ */
 
 use PHPUnit\Framework\TestCase;
 use ILIAS\UI\Component\Input\Field\Section;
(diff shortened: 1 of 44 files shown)


thibsy:
P.S I'll look into the failing tests

Amstutz:
Hi @thibsy 

Thx for looking into this and proposing a fix. Special thx for improving and fixing the tests.

The fix currently has one major issue [here](https://github.com/ILIAS-eLearning/ILIAS/pull/4786/files#diff-6aa473ce6a784d1044324806de6d3c63658ba37ff16c65d1633033faeca64c21R89)

` return (int) \ilFileUtils::getUploadSizeLimitBytes();`

With this you introduce implicitly the dependency to ilFileUtils. IMO we should not do that. Passing via constructor seems a way to go, however seems problematic, since you use traits. Hmm, hmm, you could also pass a MaxFileSize Resolver by factory and pass it will initializing the UI Framework. This is how we usually work with such external deps, that we do not want to pass to the client for doing the work.

thibsy:
You're right @Amstutz. I introduced some sort of an `UploadLimitResolver`, but turns out introducing a new dependency to the input-factory is not easy =) (hence the 44 files now).

But it sort of solved the Issue I've been having with the `$max_file_size` attribute, which can now be nullable and return the default size of the resolver instead. I figured since the `UploadHandler` must be introduced by the implementations I can do the same for the resolver.

What do you think?

Amstutz:
Hi @thibsy 

Thx a lot! Looking mucht better.

> I introduced some sort of an UploadLimitResolver, but turns out introducing a new dependency to the input-factory is not easy

yes, you are right, especially in the Input Field case. I found 2 reasons (thx to your PR):

- UI Components external test classes use own pattern of injecting UI Components, instead of just using the centralized ready to be used UITestHelper, which can get you a Complete ready-to-be-used in tests ready Factory, Renderer and MainTemplate in one line. This then generates duplicate code.
- UI Components internally in have duplicate Code in the Inputs, this is something we should get rid of. 

However, both things are beyond this PR. IMO this can be merged. Can you perform a final check on the UI with the case described in mantis on PHP8 and 7.4. If both work, you can merge this directly to avoid conflicts. 

Thx a lot for you contribution and special thanks for improving the test base.

thx.

Amstutz (approved):

chfsx:
Hi @Amstutz ! thanks a lot, I tested the solution with PHP 7.4 and PHP 8, works like a charm. after CI has run, I'll merge this to trunk

PR #4791, opened 2022-07-18
Dashboard Administration: Save form without write right

lscharmer:
Mantis Bug: https://mantis.ilias.de/view.php?id=29592
When the user has visible and read rights but no write rights for the dashboard administration:
- all form fields will be disabled.
- an error message is shown when the user is trying to save the forms.


mjansenDatabay (changes requested):
I suggest to use `$DIC->rbac()->system()`.

mjansenDatabay (approved):

PR #4793, opened 2022-07-18
Fix maximum levenshtein distance

kunzeus:
Mantis: https://mantis.ilias.de/view.php?id=17861
Used the recommended code from the comment area.
I added the requested performance improvements:
Not running the Levenshtein distance when string sizes differ too much and leave the code execution when the maximum allowed distance is reached

However, I didn't want to start with integrating the Levenshtein distance as a library since I currently don't want to mess around  with the composer. 
Therefore, I integrated the Levenshtein distance in the class.assQuestion class since it's the same baseclass for all three classes it is used inside the test module.

Since this is my first pull request in ILIAS, I know I still have a lot to learn. Therefore, please let me know if additional things need to be taken into consideration or changes have to be done.


mbecker-databay:
Hi and thank you very much for your PR!

I have reviewed the code and I would ask you for one change prior to accepting it.

The current efforts in general are to try to free the base classes - assQuestion and assQuestionGUI - from any extra weight. 
I would like to have the Levenshtein implementation extracted to an own class. 
That would also keep the source code license clean: The MIT licensed part would live in its own file.

Cheers!

Max

mbecker-databay:
After sending my above remark, when discussing this with other T&A developers, I was told that the ideal place/way to handle this is by means of the Refinery / Transformation library.

While I personally do not find the naming/labelling intuitive (because we do not strictly "transform" here but "analyze"), it's the designated place for such functionality in ILIAS.

dsstrassner:
I would say, we talk about this PR in the next TechSquad Meeting tomorrow. Do you join us @kunzeus?

kunzeus:
i should be there, there should be no more all day events on a thursday for now.

kunzeus:
I added the code of the Levenshtein function into the refiner.
Please, let me know if I got something wrong or more adjustments are needed.

mbecker-databay:
As discussed in the TechSquad meeting, be so kind and provide tests for the transformation. @klees & @mjansenDatabay will be so kind and review that part in here.

src/Refinery/String/Levenshtein.php
```diff
@@ -0,0 +1,138 @@
...
+        return $matrix[$str1Length][$str2Length];
+    }
+
+    /**
+     * Helper function for levenshtein distance calculation, used to convert strings into arrays.
+     *
+     * @param string $str the string that is converted into an array
+     * @return array an array containing the characters of the string, each in a single cell
+     */
+    private function StringToArray(string $str): array
```
mjansenDatabay:
```suggestion
    private function stringToArray(string $str): array
```

Maybe `stringToChars` or `stringToCharCollection` would be a better name.

src/Refinery/String/Group.php
```diff
@@ -108,4 +109,39 @@ public function estimatedReadingTime($withImages = false) : EstimatedReadingTime
...
+
+    /**
+     * Creates an object of the Levenshtein class
+     * This class calculates the levenshtein distance with a default value of 1.0 per insert, delete, replacement.
+     *
+     * @param string $str string for distance calculation
+     * @param int $maximumDistance maximum allowed distance, limits the calculation of the Levenshtein distance. A maximum distance of 0 disables the function
+     * @return Levenshtein
+     */
+    public function levenshteinDefault(string $str, int $maximumDistance) : Levenshtein
```
mjansenDatabay:
```suggestion
    public function levenshteinDefault(string $str, int $maximumDistance) : Transformation
```

src/Refinery/String/Group.php
```diff
@@ -108,4 +109,39 @@ public function estimatedReadingTime($withImages = false) : EstimatedReadingTime
...
+     * @param float $costRep cost for replacement default 1.0
+     * @param float $costDel cost for deletion default 1.0
+     * @return Levenshtein
+     */
+    public function levenshteinCustom(string $str,
+                                      int $maximumDistance,
+                                      float $costIns,
+                                      float $costRep,
+                                      float $costDel
+    ) : Levenshtein {
```
mjansenDatabay:
```suggestion
    public function levenshteinCustom(
        string $str,
        int $maximumDistance,
        float $costIns,
        float $costRep,
        float $costDel
    ) : Transformation {
```

src/Refinery/String/Group.php
```diff
@@ -108,4 +109,39 @@ public function estimatedReadingTime($withImages = false) : EstimatedReadingTime
...
+    /**
+     * Creates an object of the Levenshtein class
+     * This class calculates the levenshtein distance with custom parameters for insert, delete, replacement.
+     *
+     * @param string $str string for distance calculation
+     * @param int $maximumDistance maximum allowed distance, limits the calculation of the Levenshtein distance. A maximum distance of 0 disables the function
+     * @param float $costIns cost for insertion default 1.0
+     * @param float $costRep cost for replacement default 1.0
+     * @param float $costDel cost for deletion default 1.0
+     * @return Levenshtein
```
mjansenDatabay:
```suggestion
     * @return Transformation
```

src/Refinery/String/Group.php
```diff
@@ -108,4 +109,39 @@ public function estimatedReadingTime($withImages = false) : EstimatedReadingTime
...
     }
+
+
+    /**
+     * Creates an object of the Levenshtein class
+     * This class calculates the levenshtein distance with a default value of 1.0 per insert, delete, replacement.
+     *
+     * @param string $str string for distance calculation
+     * @param int $maximumDistance maximum allowed distance, limits the calculation of the Levenshtein distance. A maximum distance of 0 disables the function
+     * @return Levenshtein
```
mjansenDatabay, line 117:
```suggestion
     * @return Transformation
```

src/Refinery/String/Levenshtein.php
```diff
@@ -0,0 +1,138 @@
...
+    {
+        $length = strlen($str);
```
mjansenDatabay: AFAIK the used string methods are not multibyte safe. `ilStr` could be used to handle multibyte strings in ILIAS. Appropriate unit tests for such strings would be great.

tests/Refinery/String/LevenshteinTest.php
```diff
@@ -0,0 +1,92 @@
...
+{
+    /**
+     * @var  Group
+     */
+    private $group;
+
+    public function setUp() : void
+    {
+        $dataFactory = new Factory();
+        $language = $this->getMockBuilder('\ilLanguage')
```
mjansenDatabay:
```suggestion
        $language = $this->getMockBuilder(ilLanguage::class)
```

mjansenDatabay (changes requested):
Hi @kunzeus,

thank you a lot for contributing to ILIAS and the `Refinery`. We highly apreciate your efforts.

In addition to the inline comments please implement the following changes:
- [x] Instead of having two methods in the factory, `levenshtein(...)` and `levenshteinCustom(...)`, please create a new sub-group `levenshtein(...)` with `standard(...)` and `custom(...)` as methods. Using `camelCase` in these factories could be an indicator for a "smell" where a new sub-group could be more appropriate.
- [x] Please use `snake_case` variable names, whish will be our variable naming convention in the future (see: #4533).
- [x] Please rename variables like `$str2Array` or `$str2Length`. Instead of `str1` and `str2` you could use words like `main`/`base` and `other`, `left` and `right` or similar pairs ...
- [x] There are two PSR-2 (our current code style) violations. Please add a blank line to the refinery related classes.
- [x] Please add our license header where missing (see: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/CI/Rector/ChangeLicenseHeader.php#L32).
- [x] Please use `declare(strict_types=1);` in the test class.

A separate PR for ILIAS >= 8 with `Typed Properties` would be highly appreciated as well.

Kindly,
@mjansenDatabay and @klees

src/Refinery/String/Transformation/LevenshteinTransformation.php
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * Helper function for levenshtein distance calculation, used to convert strings into character arrays.
+     *
+     * @param string $string_to_convert the string that is converted into an character array
+     * @return array an array containing the characters of the string, each in a single cell
+     */
+    private function stringToCharacterArray(string $string_to_convert): array
+    {
+        $length = ilstr::strLen($string_to_convert);
```
mjansenDatabay:
```suggestion
        $length = ilStr::strLen($string_to_convert);
```

src/Refinery/String/Group.php
```diff
@@ -104,8 +103,22 @@ public function caseOfLabel(string $language_key) : CaseOfLabel
...
+     */
+    public function levenshtein(): Levenshtein
+    {
+        return new Levenshtein();
+    }
+
+
+
+
```
mjansenDatabay: Can we please remove these empy lines :)?

src/Refinery/String/Levenshtein.php
```diff
@@ -0,0 +1,60 @@
...
+    public function custom(
+        string $str,
+        int $maximum_distance,
+        float $cost_insertion,
+        float $cost_replacement,
+        float $cost_deletion
+    ): Transformation {
+        return new LevenshteinTransformation($str, $maximum_distance, $cost_insertion, $cost_replacement, $cost_deletion);
+    }
+}
```
mjansenDatabay: We stillt need a line break here. Please use the `PHP-CS-FIXER`.

mjansenDatabay (changes requested):

src/Refinery/String/Group.php
```diff
@@ -104,8 +104,18 @@ public function caseOfLabel(string $language_key) : CaseOfLabel
...
      * @return EstimatedReadingTime
      */
-    public function estimatedReadingTime($withImages = false) : EstimatedReadingTime
+    public function estimatedReadingTime($withImages = false): EstimatedReadingTime
     {
         return new EstimatedReadingTime($withImages);
     }
+
+    /**
+     * Creates a between class.
```
mjansenDatabay: Please add a more detailed description here.

mjansenDatabay (changes requested):
Hi @kunzeus,

I am fine with these changes except for the description of the `levensthein()` (group) method.

After this has been changed, @maxbecker and @dsstrassner can merge this whenever they like.

Please keep in mind that (at least) the `Refinery` code has to be slightly changed for `release_8` and `trunk`.

Best regards,
@mjansenDatabay

PR #16, opened 2015-04-02
Remove duplicate Unicode lib (keeping the newer version + relocate)

colin-kiegel:
ok .. this looks a bit like a mess, but what I did is simple:

There were two unicode libraries present (almost identical)
- Modules/Wiki/mediawiki/normal (new version)
- include/Unicode

I moved the new version of mediawiki to include/Unicode and changed all references to point to the new location. Again to avoid classname collisions.

PR #15, opened 2015-04-02
removed *unused* JSON-Emulator (no references to it anymore)

colin-kiegel:
If still needed, there is a perfect duplicate at this path './Services/JSON/include/json.php'.

This also removes a classname collision like PR #14

PR #13, opened 2015-04-01
#15687: moving mail to... [HUN]

KissDaniGH:

PR #4799, opened 2022-07-18
bugfix: the Language to be set MUST not be NULL

utesche:
otherwise a TypeError Exception is raised:

TypeError thrown with message "Argument 1 passed to ilObjUser::setLanguage() must be of the type string, null given, called in /.../Services/User/classes/class.ilUserImportParser.php on line 467"

Stacktrace:
#16 TypeError in /.../Services/User/classes/class.ilObjUser.php:1479
#15 ilObjUser:setLanguage in /.../Services/User/classes/class.ilUserImportParser.php:467
#14 ilUserImportParser:importBeginTag in /.../Services/User/classes/class.ilUserImportParser.php:364
#13 ilUserImportParser:handlerBeginTag in [internal]:0

### DIFF
--- a/Services/User/classes/class.ilUserImportParser.php
+++ b/Services/User/classes/class.ilUserImportParser.php
@@ -464,7 +464,7 @@ class ilUserImportParser extends ilSaxParser
                     $ilias->ini->readVariable("layout", "style")
                 );
 
-                $this->userObj->setLanguage($a_attribs["Language"]);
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
@alex40724 I assigned this to you because you did the PHP 8 refactoring according to https://docu.ilias.de/goto_docu_dcl_10296_209.html

alex40724:
Thanks for the PR!

PR #4805, opened 2022-07-19
Fix writing to Globals

kergomard:
This fixes a write on $_REQUEST when creating a question in a new or existing pool, but forgetting to add a name or somehow breaking the ref_id in the dropdown.


mbecker-databay:
Thank you so much for the PR!

PR #4785, opened 2022-07-15
Fix/8/setup installation

chfsx:
There were several problem which made the installation of trunk fail.
@smeyer-ilias not sure if this is a feasible solution to have ilDBInterface injectable to ilTree

@mstuder or @d3r1w please have a look at the installation of employee-talks

PR #4807, opened 2022-07-19
EmployeeTalk: possible fix for db update steps

daniwe4:
Hi,
that's a little bit ugly but works in my tests. At the moment this is blocking a new installation, so i hope this could help.
Greetings Daniel


chfsx:
Another proposal in #4785

Amstutz:
Related https://mantis.ilias.de/view.php?id=33382

PR #4810, opened 2022-07-19
[FIX] 0025664: Double click on Appointment Title

chfsx:
@Amstutz this fixes the issue of https://mantis.ilias.de/view.php?id=25664
Seems to be more trivial than I thought...


Amstutz:
Wow, nice one. Did not remember this, but thx!

Amstutz:
picked to trunk as well.

PR #4811, opened 2022-07-19
Remove permanent link input field of footer

lscharmer:
This PR removes the input field in the footer with the permanent link to the current page. Instead one can click on the link to copy the permanent link to the clipboard.

Feature wiki: https://docu.ilias.de/goto.php?target=wiki_1357_Remove_the_input_field_of_Permanent_link_in_Footer

This request was originally proposed as a feature but in agreement (of Oliver Samoila) with the UI coordinators (Richard Klees and Timon Amstutz) this should be implemented as a PR.

#### Clipboard API
If the clipboard is not available the default behavior of the anchor is preserved.

#### Focus Highlight
Because the anchor is a normal anchor the highlighting of the link when one focuses the link e.g. with *tab* is the same as with `Legal Notice` and `Accessibility Control Concept` links.

### DIFF
--- a/lang/ilias_de.lang
+++ b/lang/ilias_de.lang
@@ -5156,6 +5156,7 @@ common#:#pdfg_visible#:#PDF-Erstellungs ist sichtbar.
 common#:#pdfg_write#:#Einstellungen der PDF-Erstellung bearbeiten
 common#:#pd_items_news#:#Einschließlich Neuigkeiten der ausgewählten Objekte
 common#:#perma_link#:#Link zu dieser Seite
+common#:#copy_perma_link#:#Link zu dieser Seite kopieren
 common#:#permission_denied#:#Kein Zugriffsrecht
 common#:#permission#:#Recht
 common#:#permission_settings#:#Rechteeinstellungen
--- a/lang/ilias_en.lang
+++ b/lang/ilias_en.lang
@@ -5136,6 +5136,7 @@ common#:#pdfg_visible#:#PDF Generation is visible
 common#:#pdfg_write#:#User can edit settings of PDF Generation
 common#:#pd_items_news#:#Include News of Personal Items
 common#:#perma_link#:#Permanent Link
+common#:#copy_perma_link#:#Copy Permanent Link
 common#:#permission_denied#:#Permission Denied
 common#:#permission#:#Permission
 common#:#permission_settings#:#Object Permission Settings
--- a/src/UI/Implementation/Component/MainControls/Renderer.php
+++ b/src/UI/Implementation/Component/MainControls/Renderer.php
@@ -32,6 +32,7 @@ use ILIAS\UI\Renderer as RendererInterface;
 use ILIAS\Data\URI;
 use ILIAS\UI\Implementation\Render\ResourceRegistry;
 use LogicException;
+use Closure;
 
 class Renderer extends AbstractComponentRenderer
 {
@@ -435,8 +436,11 @@ class Renderer extends AbstractComponentRenderer
         $perm_url = $component->getPermanentURL();
         if ($perm_url instanceof URI) {
             $url = $perm_url->__toString();
-            $tpl->setVariable('PERMA_LINK_LABEL', $this->txt('perma_link'));
-            $tpl->setVariable('PERMANENT_URL', $url);
+            $button = $this->getUIFactory()
+                           ->button()
+                           ->shy($this->txt('copy_perma_link'), $url)
+                           ->withAdditionalOnLoadCode($this->permanentJS($url));
+            $tpl->setVariable('PERMANENT', $default_renderer->render($button));
         }
         return $tpl->get();
     }
@@ -450,7 +454,6 @@ class Renderer extends AbstractComponentRenderer
         $registry->register('./src/UI/templates/js/MainControls/dist/mainbar.js');
         $registry->register('./src/UI/templates/js/MainControls/metabar.js');
         $registry->register('./src/GlobalScreen/Client/dist/GS.js');
-        $registry->register('./src/UI/templates/js/MainControls/footer.js');
         $registry->register('./src/UI/templates/js/MainControls/system_info.js');
     }
 
@@ -466,5 +469,21 @@ class Renderer extends AbstractComponentRenderer
             ModeInfo::class,
             Component\MainControls\SystemInfo::class
         );
+    }
+
+    private function permanentJS(string $url) : Closure
+    {
+        $url = json_encode($url);
+        return static function (string $id) use ($url) : string {
+            return "document.getElementById('$id').addEventListener('click', function(event){
+                    if (window.navigator.clipboard) {
+                        window.navigator.clipboard.writeText($url);
+                        event.stopImmediatePropagation();
+                        return false;
+                    } else {
+                        console.warn('Cannot copy link to clipboard. Please note that the clipboard is only available in secure contexts (HTTPS). See https://developer.mozilla.org/en-US/docs/Web/API/Clipboard for more information.');
+                    }
+                });";
+        };
     }
 }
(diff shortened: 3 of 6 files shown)


Amstutz (approved):
Hi @lscharmer 

Thx a lot for this. IMO this is good to be discussed with the JF.

Before merging however, one question:
- [x] Link Color: `color: white` looks strange for the footer, since the footer BG is white (since the white BG-Change a few weeks ago). Is this intended? You can probably just remove it, which would then use the default, which is set to: @il-footer-color. 

Thx.

lscharmer:
> Hi @lscharmer
> 
> Thx a lot for this. IMO this is good to be discussed with the JF.
> 
> Before merging however, one question:
> 
>     * [ ]  Link Color: `color: white` looks strange for the footer, since the footer BG is white (since the white BG-Change a few weeks ago). Is this intended? You can probably just remove it, which would then use the default, which is set to: @il-footer-color.
> 
> 
> Thx.

Hi @Amstutz
I did not know that the background color changed, I will remove this. Thank you for the info.

chfsx:
Jour Fixe, 25 JUL 2022 : Thank you very much for this PR, that we happily accept. We would kindly ask Lukas to allow a backup, if the clipboard is not available to Javascript (e.g. due to missing transport encryption or an access through a screen reader). Please also check, if there is some LESS/CSS for highlighting when tabbing, targeted at the current input field and remove any unused LESS/CSS.

klees (changes requested):
Hi @lscharmer,

thanks a lot for looking into this!

- [x] Button 1: Use a `Button` (probably `Shy`) instead of a link [here](https://github.com/ILIAS-eLearning/ILIAS/pull/4811/files#diff-9200c77ab023cc658566f3357f12c91a45ef728790fd9ecab3f3983e34db51dfR6). As far as I understand a11y requirements, this usage of the `<a>` is problematic. The code tells the user that the element links to a resource "Copy Permanent Link" with a certain URL. The URL, though, does not in fact point to any such resource but the current location instead. Moreover, the actual interaction is interrupted by JS and the user won't actually be directed to said URL when clicking the link. I would expect all this to be very confusing for user with assistive technology that relies on HTML semantics.
- [x] Button 2: Please use the UI framework to render said `Shy (?) Button` instead of adding custom HTML to the template of the footer.
- [x] JS-Binding: Please do not bind the JS with a specific CSS-selector. The structure of the html or the class names used to style the URL are not very well suited to bind JS. Both are likely to change, be it in future iterations on the skin of the ILIAS core or in custom skins. Also, you want to bind the behaviour to that specific element, not to every element specified by the selector. Please use `JavaScriptBindable::withAdditionalOnLoadCode` on the button you will be creating. You could still put the code into an extra `*.js` file and call it from there.

If there are any questions or doubts regarding the proposed changes, feel free to ask or voice them, of course.

Best regards!

klees (approved):

nicoroeser:
Many thanks to all of you for working on this!

I have just tested this change in an ILIAS 8 installation in several combinations:
* browser: Firefox with Tridactyl, Waterfox Classic with Pentadactyl,
* NoScript: set to allow JavaScript, set to disallow JavaScript,
* access via IP address, access via host name (changes Firefox interpretation of site).

In all tests, I used HTTP (not HTTPS) connections (to a local Docker container, but the latter should not matter).

When using Firefox with JavaScript, link copying works. First thing I noticed is that I do not get any feedback that the page link has been copied to the clipboard. Even the button does not indicate a press. Well, I can live with that. The link ends up in my clipboard, which is fine. :+1:

I’m a bit annoyed though that the look of the button makes me believe it is a hyperlink, but it behaves differently than a link: when I hover it, my browser does not show me the link target. When I use the way provided by my user agent to copy a link, I get an empty clipboard. (It is debatable whether Tridactyl and the like should be supported. But we cannot know each and every user agent, so should try to support many of them. And even the more common way – Right-click, Copy-link-URI – is not available.)

What bothers me more is that with JavaScript disabled I cannot find any (quick and accessible to non-expert users) way anymore to copy the link. If we declare JavaScript as a necessity, I’m fine with that, too.

When I use a host name instead of an IP address, I guess that Firefox does not interpret the site as local (not sure about this). When JavaScript is disabled, nothing happens when I click(*) the button. With JavaScript enabled, the page reloads when I click the button. Does not happen when accessed by IP address (→ local site, probably treated as secure, not analyzed in more depth).

(*) I tried with a pointing device.

To be honest, I’m not sure that a _button_ is the best way to solve this. (I had not raised objections earlier, because I had thought that such a button would fall back to a link if JavaScript is not enabled. Sorry about that.) After all, I like to get a URI to some document: the current object. Commonly you get such a URI from a hyperlink. And it does not matter to me if the page links to “itself”.

Having a button is fine for me, too, but
1. its look could be improved to signal to users that it is not a hyperlink, and
2. it should be usable in secure and in insecure contexts, with JavaScript available and not available (alternatively, let’s make JavaScript support mandatory).

I just feel that the current result is worse than before (again, sorry). Finally, I don’t want to start a bikeshed discussion here! – If you don’t agree, just go ahead and merge this. :smiley:

Amstutz:
Hi @nicoroeser 

Thx for looking into this.

> What bothers me more is that with JavaScript disabled I cannot find any (quick and accessible to non-expert users) way anymore to copy the link. If we declare JavaScript as a necessity, I’m fine with that, too.

A page needs to be able to be rendered without JavaScript enabled, as far as I know. However, at the same time, a page might Require JavaScript to fully operable as of WCAG 2.0. Note that 99% of the screen readers have JS enabled (see: https://webaim.org/techniques/javascript/). IMO, JS is required since quite some time for ILIAS to work properly (e.g. for Async Modals). 

> its look could be improved to signal to users that it is not a hyperlink, 

You certainly have a point here. However, to move things a long, I would propose to merge this as is. Then you could create a CSS ticket. @yvseiler and I would then look for an improved situation here.

To everybody involved. I am really not 100% sure if we are doing the best thing here. However, since we decided on the JF to go this road and the current situation is problematic as well, I am willing to give this a try. We might have to refine this solution further if we gathered some further feedback on this.

@lscharmer 

Change or briefly explain why not: 
- [ ] JS-Binding II: Thx for using the JS-Binding. There is one minor thing you could improve. You completely removed the JS File and now specify the complete code in your PHP rendering method. I am certain this is done in other places and might be OK for one-liners. In your case, it clutters the rendering with multiple lines. IMO it would be better, to keep the code in the JS file and use the scoping pattern we use in other places for giving this functionality than a public interface to call from your JS placed inside your rendering method. E.g. see [here](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/src/UI/templates/js/ViewControl/sortation.js) for an example. 

If you like to keep it this way, I am fine too. It's really an edge case. Just shorty tell us that you like to merge as is, and we do.

Thx
@Amstutz in agreement with @klees

lscharmer:
> Thx for using the JS-Binding. There is one minor thing you could improve. You completely removed the JS File and now specify the complete code in your PHP rendering method. I am certain this is done in other places and might be OK for one-liners. In your case, it clutters the rendering with multiple lines. IMO it would be better, to keep the code in the JS file and use the scoping pattern we use in other places for giving this functionality than a public interface to call from your JS placed inside your rendering method. E.g. see here for an example.

Hi @Amstutz,
I did not place the JS code into an own file because the code is so little. If it would be more code I would definitively put it into an own file.

Best regards

Amstutz:
I merged this to trunk (ILIAS 9) only. If I understand the PR correctly, this is intended for ILIAS 9 only, correct?

lscharmer:
> I merged this to trunk (ILIAS 9) only. If I understand the PR correctly, this is intended for ILIAS 9 only, correct?

This PR, yes I will create another one for ILIAS8. I just wanted to wait until there are no changes required anymore before opening the other.

matthiaskunkel:
Jour Fixe, 09 JAN 2023: We have two options to implement the feature to copy the permanent link of a page in ILIAS. As an alternative to the current suggestion made in PR4811 we could also just show the permanent link and let users copy it per mouse click. Decision is postponed to next JF.

PR #4813, opened 2022-07-19
Test/TestQPL: Fix dependency on `ilPluginAdmin`

mjansenDatabay:
This PR removes the dependency on `ilPluginAdmin` from the `Test` and `TestQuestionPool` components. According to the deprecation declaration of @klees the `ilComponentRepository` has to be used in the future. The triggered error resulted in failing ILIAS 8 tests, so we had to get rid of the dependency now.

Mantis Issue: https://mantis.ilias.de/view.php?id=33298

Because I touched many files (~100) of the T&A code base this should be tested thoroughly (e.g. by checking out this branch locally).


mjansenDatabay:
@alex40724 I also changed two files under your maintenance. Please check if the changes are okay for you.

alex40724 (approved):
Modules/LearningModule related changes are fine for me.

PR #4814, opened 2022-07-19
Remove setQuestionTabs from all Question Types

kergomard:
This removes the setQuestionTabs-Function from all Question-Types. I don't see any drawbacks of doing this as the whole infrastructure is already there, but it isn't used. I hit my head against this when trying to fix the misplaced statistics-tab.
If you accept this, I will provide a PR for trunk.


kergomard:
I changed my opinion on this and opted for a more comprehensive and in my view cleaner solution.

mbecker-databay:
I accept! :) Thank you so much!

PR #4816, opened 2022-07-19
Move Statistics Tab to Preview

kergomard:
This PR only has an effect if first #4814 is merged.

It moves the statistics-tab to the QuestionPreviewGUI. The functionality in itself is kept in the question though as it seems the right place until we clean up routing in this part of the T&A.

As is pointed out in https://mantis.ilias.de/view.php?id=33391 right now the use of the tab is very limited.

This PR, from my point of view has two parts:
1. The conceptual one: Do we want to move this Information to a place where it can be accessed after a test is started? Personally I would suggest to attach the JF-label to this and see if we can reach a conclusion.
2. The implementation: I'm glad for any feedback or better ideas and would spend a little more time to make sure that the statistics shown here are as specified. I would also provide a PR against trunk

### DIFF
--- a/Modules/Test/classes/tables/class.ilTestQuestionsTableGUI.php
+++ b/Modules/Test/classes/tables/class.ilTestQuestionsTableGUI.php
@@ -268,6 +268,12 @@ class ilTestQuestionsTableGUI extends ilTable2GUI
             $this->getEditLink($data, get_class($this->getParentObject()), $this->getParentCmd())
         );
         
+        $actions->addItem(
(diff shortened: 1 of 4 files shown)


mbecker-databay:
Mantis-Issue is set for JF discussion

matthiaskunkel:
Jour Fixe, 08 AUG 2022 : We highly appreciate this PR and accept it as bugfix for  https://mantis.ilias.de/view.php?id=33391. Please merge to 7, 8 and trunk.

PR #4822, opened 2022-07-20
Feature/configurable http path

l-4-l:
As a Sysadmin I want to create Ilias instance under `https://MYHOSTNAME/ilias/`.

I am using Traefik as a reverse-proxy, it means that:
- I route all incoming requests to `MYHOSTNAME/ilias/` directly to Ilias-instance and I can strip path-postfix (`/ilias`). (My ilias-instance runs under some IP or docker in my internal network)

Current situation:
- My ilias instance accepts the original request but most of redirects which it sends, it's doing to `https://MYHOSTNAME/`. Example: first request (user is not logged in) to `https://MYHOSTNAME/ilias/` is transformed to http://INTERNAL-ILIAS-IP/ and Ilias response is `HTTP 302 REDIRECT to https://MYHOSTNAME/login.php`

**Should be**
- Ilias response should be `HTTP 302 REDIRECT to https://MYHOSTNAME/ilias/login.php`.

**Tech description of the problem in the code**
- Ilias can correctly guess the hostname, but fails at guessing its own **path**. It always "thinks" that his path is `/`.

**How it can be solved without coding**
- Only if you have a reverse proxy which can rewrite redirect responses. Then proxy can rewrite (edit) response `HTTP 302 REDIRECT to https://MYHOSTNAME/login.php` to `HTTP 302 REDIRECT to https://MYHOSTNAME/ilias/login.php`. Currently Traefik cannot do it.

**How other apps are doing it and how I solved it here**
- I provide an additional env variable `ILIAS_HTTP_PATH` with "external" URL (including HTTP/HTTPS, domain name and path of the app in "external internet").
- if this env variable is not set, then behaviour is the same as before
- if it's set, then Ilias just picks it and not trying to ***guess*** ILIAS_HTTP_PATH.


l-4-l:
sorry, wrong orig branch, i will try again later

PR #4823, opened 2022-07-20
New env variable ILIAS_HTTP_PATH can override ilias path

l-4-l:
As a Sysadmin I want to create Ilias instance under `https://MYHOSTNAME/ilias/`.

I am using Traefik as a reverse-proxy, it means that:
- I route all incoming requests to `MYHOSTNAME/ilias/` directly to Ilias-instance and I can strip path-postfix (`/ilias`). (My ilias-instance runs as docker container in a docker network)

**Current situation**
- My ilias instance accepts the original request but most of response-redirects which it sends, it's doing to `https://MYHOSTNAME/`.
- Example:
1. In user-browser (user is not logged in), first request to `https://MYHOSTNAME/ilias/` is accepted by Traefik because it has path `/ilias`
2. Traefik transforms request to http://INTERNAL-ILIAS-IP/
3. Ilias on INTERNAL-ILIAS-IP processes it and sends a response: `HTTP 302 REDIRECT to https://MYHOSTNAME/login.php?...`

**Should be**
- Ilias response should be `HTTP 302 REDIRECT to https://MYHOSTNAME/ilias/login.php?...`.

**Tech description of the problem in the code**
- Ilias can correctly guess the hostname, but fails at guessing its own **path**. It always "thinks" that his path is `/`.

**How it can be solved without coding**
- Only if you have a reverse proxy which can rewrite redirect responses. Then proxy can rewrite (edit) response `HTTP 302 REDIRECT to https://MYHOSTNAME/login.php` to `HTTP 302 REDIRECT to https://MYHOSTNAME/ilias/login.php`. **Currently Traefik cannot do it. Many popular reverse-proxies cannot do it.**

**How other apps are doing it and how I solved it here**
- I provide an additional env variable `ILIAS_HTTP_PATH` with "external" URL (including HTTP/HTTPS, domain name and path of the app in "external internet").
- if this env variable is not set, then behaviour is the same as before
- if it's set, then Ilias just picks it and not trying to ***guess*** ILIAS_HTTP_PATH.

### DIFF
--- a/Services/Init/classes/class.ilInitialisation.php
+++ b/Services/Init/classes/class.ilInitialisation.php
@@ -376,6 +376,11 @@ class ilInitialisation
      */
     protected static function buildHTTPPath() : bool
     {
+        $ilias_http_path_from_env = getenv('ILIAS_HTTP_PATH');
+        if (isset($ilias_http_path_from_env)) {
+            return define('ILIAS_HTTP_PATH', $ilias_http_path_from_env);
+        }
+
         global $DIC;
 
         if ($DIC['https']->isDetected()) {


JohnnySepp:
Will this be merged in ilias 8 as well?

JohnnySepp:
scheduled for ILIAS JF at 31.10.2022

klees:
Hi @l-4-l,

this question might be super stupid, but why can't you just set the configuration variable [`http.path`](https://github.com/ILIAS-eLearning/ILIAS/tree/trunk/setup#about-the-config-file)? AFAIK this is doing exactly what you are looking to do.

Best regards!

landesportal-orca:
Hi @klees , 
Thank your for suggestion, the documentation in the mentioned file was not easy to find.

TL;DR: ***No, using `http_path` from ini-file is not the same as using ILIAS_HTTP_PATH. Using ILIAS_HTTP_PATH as in my code is better and gives global effect.***

Longread:

I made a quick code analysis and found that the variable `https_path` from file `ilias.ini.php` contains an ***internal*** Http Path which is used only in functionalities:
- "send email",
- "soap client" and 
- "cron manager".
("internal" here means that it's not considering that host URL can be behind a proxy. As a result all 3 functionalities can possibly fail. I haven't tested them. ILIAS_HTTP_PATH is an URL of the external proxy, it's usually guessed from HTTP-requests or something, my code makes it configurable instead guessing)

In my installation the `http_path` contains an HTTP URL for a docker container (`http://$CONTAINER_ID/`). I am sure that in a normal workflow this URL is not used (except from the mentioned functionalities).

The PHP variable ILIAS_HTTP_PATH is used anywhere else (except mentioned functionalities) and what's most important it's used in building ***external redirects***. 

I suppose that it is inconsistency in the code which is not really important, but changing `http_path` will not have any effect on a (for example) login process, unlike with ILIAS_HTTP_PATH. I think both of them finally should be controlled from a single variable. I see ***exactly one attempt*** to do it in `class.ilCronManager.php`:
```php
 if (!defined("ILIAS_HTTP_PATH")) {
   define("ILIAS_HTTP_PATH", ilUtil::_getHttpPath()); // _getHttpPath is a method that reads http_path from ini-file
 }
```

> this question might be super stupid, but why can't you just set the configuration variable [`http.path`](https://github.com/ILIAS-eLearning/ILIAS/tree/trunk/setup#about-the-config-file)? AFAIK this is doing exactly what you are looking to do.
> 
> Best regards!

klees:
Hi @pascalseeland

Best regards!

kergomard:
Hi @pascalseeland  

Best regards!
@kergomard

kergomard:
Hi @pascalseeland  
As I looked through the PRs that have been around for a while without progress, I found this one. Could you have a look at this and either see how to make it progress or close it?

Thank you very much!

dsstrassner:
hi @thojou, after the merge of #8553 - I thought you would also look into these old PRs.

PR #4833, opened 2022-07-21
fixed unit test due to new ilLink dependency

smeyer-ilias:
I introduced a new dependency in ilLink which broke the unit test for the main menu service. 
This fixes it.


chfsx:
thanks a lot, I'll update the license header in the file

PR #4840, opened 2022-07-21
Logging: Don't log to console for xhr request not sending `cmdMode` to prevent invalid JSON responses

mjansenDatabay:
This PR fixes an issue with popular community plugins (like `UserTakeover`) sending `xhr` requests which do not include a `cmdMode=async` query parameter.

If the console logging is enabled in the ILIAS logging administration, `JSON` HTTP responses are not valide anymore because `\Monolog\Handler\BrowserConsoleHandler` appends a `<script>` tag to the response body.

### DIFF
--- a/Services/Logging/classes/public/class.ilLoggerFactory.php
+++ b/Services/Logging/classes/public/class.ilLoggerFactory.php
@@ -110,20 +110,25 @@ class ilLoggerFactory
             }
         }
     }
-    
+
     /**
      * Check if console handler is available
-     * @return boolean
+     * @return bool
      */
-    protected function isConsoleAvailable()
+    protected function isConsoleAvailable() : bool
     {
-        include_once './Services/Context/classes/class.ilContext.php';
         if (ilContext::getType() != ilContext::CONTEXT_WEB) {
             return false;
         }
-        if (isset($_GET["cmdMode"]) && $_GET["cmdMode"] == "asynch") {
+
+        if ((isset($_GET['cmdMode']) && $_GET['cmdMode'] === 'asynch') || (
+            isset($GLOBALS['DIC']['http']) &&
+            strtolower($GLOBALS['DIC']->http()->request()->getServerParams()['HTTP_X_REQUESTED_WITH'] ?? '') === 'xmlhttprequest'
+        )) {
+            // In theory, we could analyze the HTTP_ACCEPT header and return true for text/html
             return false;
         }
+
         return true;
     }
     


Services/Logging/classes/public/class.ilLoggerFactory.php
```diff
@@ -110,20 +110,25 @@ public function initUser($a_login)
...
-    protected function isConsoleAvailable()
+    protected function isConsoleAvailable() : bool
     {
-        include_once './Services/Context/classes/class.ilContext.php';
         if (ilContext::getType() != ilContext::CONTEXT_WEB) {
             return false;
         }
-        if (isset($_GET["cmdMode"]) && $_GET["cmdMode"] == "asynch") {
+
+        if ((isset($_GET['cmdMode']) && $_GET['cmdMode'] === 'asynch') || (
```
smeyer-ilias, line 124: This changes the behaviour in case of cmdMode="async" and unset DIC['http']. 
Do you think this case is only of theoretical nature?

smeyer-ilias (commented):
Thanks @mjansenDatabay Please check the comment.

Services/Logging/classes/public/class.ilLoggerFactory.php
mjansenDatabay: @smeyer-ilias Changing the behaviour here is my intention :-).

Let me explain:
- The browser console logging handler of the `Monolog` library attached here is currently **not** aware of the HTTP response content type.
- If a HTTP client sends a requests and asks for a `JSON` response, the current behaviour will ALWAYS append a `script` tag to the response content. This leads to invalid `JSON` documents the client cannot parse anymore.
- So I state that browser console logging should be **disabled** for all HTTP requests with `cmdMode=async` **and** all requests with the `HTTP_X_REQUESTED_WITH` HTTP header. Why? Because we currently cannot properly differentiate whether a response is of type `JSON`, `HTML`, `XML` and so on. For `HTML` it might be okay to append the `script` tag with the `console.log` calls, but **not** for other types of response. So IMO the only way we could (currently) guess the content type of a response is to parse the `HTTP_ACCEPT` HTTP header. This is not part of this PR, but could be additionally be applied to allow the `console.log` for `HTML` responses.
- Of course we could `return true;`, if `$DIC['http]´ is **not** set (why ever this should happen in the context of a HTTP request.

smeyer-ilias (approved):
Thanks for fixing this.

PR #4656, opened 2022-06-15
Load full descriptions within object cache

iszmais:
https://mantis.ilias.de/view.php?id=32627

This enforces the loading off full decriptions (not just the cut offs o the object_data table) into the Object Data Cache

PR #4842, opened 2022-07-22
Fixed repository description limit 128 to be applied

iszmais:
Resolves https://mantis.ilias.de/view.php?id=32885 by removing deprecated parts

Also removes fetching of long descriptions because the loading of items already handles fetching of long descriptions since https://github.com/ILIAS-eLearning/ILIAS/pull/4656

### DIFF
--- a/Services/Container/classes/class.ilContainer.php
+++ b/Services/Container/classes/class.ilContainer.php
@@ -1102,37 +1102,11 @@ class ilContainer extends ilObject
     {
         global $DIC;
 
-        $ilSetting = $DIC->settings();
-        $ilObjDataCache = $DIC["ilObjDataCache"];
-        // using long descriptions?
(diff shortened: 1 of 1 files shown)


alex40724:
Sorry for the delay. Actually I don't know what to do with this, since this is a very minor issue and I do not trust this PR enough to just merge it. This is not your fault, the code structure is definitly a little messed up. It's just three calls, but I would like to have testing on this. E.g. the getSubItems variant calls getSubTreeQuery which does not seem to get the translations. Maybe they are added at another part of the code, but I hate to read to many control flow variants and to assume that this "just works". Part of this will be refactored with ILIAS 9.

alex40724:
Thanks for the PR. Fixed with minimal code change.

The code is a little bit obscure. I plan to refactor this.

PR #4845, opened 2022-07-22
UI Components/Learning Progress Icons, fix of: #33541

Amstutz:
Fix of https://mantis.ilias.de/view.php?id=33541 .

@klees I removed the htmlentities conversion for string values in the properties, since this makes it impossible to render Icons handled over from the ilObjectListGUIPreloader (they need to arrive as string, since this preloader also seems to feed legacy lists). At one point we need to change that, so Icons MUST be passed as Icons and not as string anymore. I prepared the rendering for that and already adapted the tests. But atm, this is not feasible. We should do that when replacing the remaining ListGUI classes.

While looking for a fix for https://mantis.ilias.de/view.php?id=33541, I came across some Issues with the Introduction of the Learning Progress Icons initially missed. @schmitz-ilias I am very sorry for that. I focused too much on the SVG items, so I missed parts of the ilLPStatusIcons. Since this is my bad, I fixed them in this commit. You work along with the tests is still nice btw ;-).

Most Importantly:
- No Custom HTML: Use UI Components. In this case, the Icon components. If you need to introduce your own HTML templates in Core ILIAS (apart from tables), you are probably doing something wrong. In case of questions, you may always contact me for help.
- Adapt UI Components: Here, some CSS change seems to have been needed for your Icons to be displayed properly. Maybe therefore you did not use the UI Components Icons. In such cases, look into options to improve the UI Components to fit your needs. If you find no good way or are uncertain, please contact us (@amstutz and/or @klees). In this case, some changes in the Icons CSS was necessary to make it work with your icons as well.
- Delos.less: Completely my bad, I missed that your changes where directly put into delos.less. If you really need custom less, put that into the services or modules respective less folder. In this case, no custom less is needed anymore.

Note that I injected the UI Factory and Renderer somewhat lazy by checking if it is set, and if not setting it through DIC in the constructor. Better would be to only inject if needed in the 2 specific functions. However, this would have needed the adaption of every instance those functions are used, which would have made the footprint of this change a bit too large.

Since this changes also the work of @schmitz-ilias it would be nice to get a thumb up from @schmitz-ilias before merging.

Thx.


schmitz-ilias:
Looks good to me, @Amstutz thanks for fixing, and thanks for the feedback :)

klees (approved):
Hi @Amstutz,

I'm (still) not happy how the `Item`s and their properties are sprawling, but that's nothing new =)
Still fine with this change for the time being, there are more general improvements visible on the horizon.

Please add the inline comment about `htmlentities` to the roadmap. After all, this is a possible vector for some injection (as many other locations) that we reintroduce.

Thanks!

Amstutz:
@klees thx for the feedback, done.

PR #4852, opened 2022-07-24
Fix https://mantis.ilias.de/view.php?id=33523 (ILIAS 8 & Trunk)

kergomard:
This is a PR to discuss at the next TechSquad-Meeting.

Mantis Issue for ILIAS 7: https://mantis.ilias.de/view.php?id=33523

PR #4853, opened 2022-07-24
Allow floats as point deduction in Hints

kergomard:
See https://github.com/ILIAS-eLearning/ILIAS/pull/4852


mbecker-databay:
Thank you so much for this one!

PR #4855, opened 2022-07-26
Fix for https://mantis.ilias.de/view.php?id=29786

kergomard:
I think we should get rid of this, but if you feel more comfortable, I can also provide a PR that checks before executing the while loop.


smeyer-ilias:
Thanks @kergomard for fixing this. 
I also think this can be removed.

pascalseeland:
Somehow github did not get the merge in 9797435b12ad86ad2232e28a68c32635c1832c16

PR #4863, opened 2022-07-26
#33552: setup now recognizes objectives that halt the program.

thibsy:
Hi all,

There currently is an issue with the setup, where an ILIAS-update (e.g. `php setup/cli.php update`) could abort when a certain objective is reached without any further information. In our case, this was due to a "bad" plugin which loaded some library in its constructor that contained an `exit` statement somewhere. 

We cannot prevent this from happening obviously, but we can offer some better information about why the setup aborted. This PR introduces a shutdown-function that recognizes if an objective halted the program and displays an according message in the terminal.

Best regards!

EDIT @klees:
The problematic behaviour is no described in [#33552](https://mantis.ilias.de/view.php?id=33552).

### DIFF
--- a/src/Setup/CLI/ObjectiveHelper.php
+++ b/src/Setup/CLI/ObjectiveHelper.php
@@ -1,5 +1,19 @@
 <?php
-/* Copyright (c) 2016 Richard Klees <richard.klees@concepts-and-training.de> Extended GPL, see docs/LICENSE */
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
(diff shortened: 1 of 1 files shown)


thibsy:
@klees do you think I should just throw a RuntimeException all the time instead of printing the message nicely if possible?

klees:
I would prefer a simple RuntimeException. This way, all errors will look the same.

klees:
And Symfony will take care =)

klees (approved):
Neat fix, thx a lot @thibsy!

klees:
Picked to trunk.

thibsy:
Thx!

PR #4864, opened 2022-07-26
Fix creation of questions from test in pool

kergomard:
The actual culprit is a wrong question_id when coming back from creating the question (as question has two ids, one for pool and one for test).


mjansenDatabay:
LGTM!

mbecker-databay:
Seems to make sense. Thank you so much for providing the PR!

PR #4867, opened 2022-07-27
introducing font weight variables

ezenzen:
Hi @all,

this PR fixt the issue 0033556. to use font weights consistently in the future, I created and introduced less-variables.

0033556: Font-weight:600 für Überschriften - https://mantis.ilias.de/view.php?id=33556

Many greetings,
Enrico

### DIFF
--- a/src/UI/templates/default/Breadcrumbs/breadcrumbs.less
+++ b/src/UI/templates/default/Breadcrumbs/breadcrumbs.less
@@ -50,7 +50,7 @@
 //** Breadcrumb margin
 @il-breadcrumb-margin-top: .8em;
 @il-breadcrumb-margin-right: 0;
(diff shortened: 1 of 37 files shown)


Amstutz (approved):
Hi @ezenzen 

Thx a lot for this proposal. IMO very good and highly appreciated. Cleans up a lot of things. 

One note, in future, please do not add changes in PRs that have nothing todo with the descripton of it. (E.g. the breadcrumb variable changes in this case). They are highly welcome, but should be done speratly. E.g. it could be, that we decline the changes here, in this case we would also lose the breadcrumb improvements.

Question, is there a reason for this being "Draft". IMO this is good to go, right?

Thx

ezenzen:
Hi @Amstutz,

yes, this is mergable. I changed the PR status. I just wanted to play it safe and fix any conflicts beforehand.

PR #4868, opened 2022-07-27
Fix for uninitialized "root" variable

iljalukin:
This entry appears in the error log while testing rewrite rules: using uninitialized "root" variable.
The $root variable declared in the previous location is apparently out of scope here and would probably be better declared in the server scope.


mjansenDatabay:
@tmuellerdatabay Could you please review this?

tmuellerdatabay (approved):
this fix from  @iljalukin is correct.

tmuellerdatabay:
@mjansenDatabay you can merge this.

PR #4869, opened 2022-07-27
centered login form

ezenzen:
we have discussed the following issue in the css squad. 

0033557 Loginform zentrieren - https://mantis.ilias.de/view.php?id=33557

I support the proposal. This customization only affects the default login form. If the login page is designed using the Page Editor, it will not be affected. The Page Editor content will remain. 

Since all authentication methods are displayed within the dic .ilStartupsection, the forms of all authentication methods are centered. 

However, I'm not sure if the div should be centered this way. If there is another/better way, discuss the CSS adjustments in the comments of this PR. 

Screenshot 2022-07-27 at 13-10-27 ILIAS Bei ILIAS anmelden

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -11127,7 +11127,7 @@ footer {
   grid-template-areas: "icon title close" "none description description" "none actions actions";
   grid-template-columns: 20px 1fr 20px;
   grid-gap: 9px;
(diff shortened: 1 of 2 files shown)


Amstutz (approved):
Thx a lot, see comments in the Ticket. IMO we can ship this as is. 

Note: this does not affect pages created by the Page Editor.

PR #4870, opened 2022-07-27
Possible Fix for Missing Info-Box After E-Mail-Confirmation

kergomard:
See: https://mantis.ilias.de/view.php?id=28383

Let me know if you would prefer another solution.

### DIFF
--- a/Services/Init/classes/class.ilStartUpGUI.php
+++ b/Services/Init/classes/class.ilStartUpGUI.php
@@ -282,17 +282,17 @@ class ilStartUpGUI
     protected function retrieveMessagesFromSession() : array
     {
         $messages = [];
-        if (isset($_SESSION[ilGlobalTemplate::MESSAGE_TYPE_FAILURE])) {
-            $messages[ilGlobalTemplate::MESSAGE_TYPE_FAILURE] = $_SESSION[ilGlobalTemplate::MESSAGE_TYPE_FAILURE];
+        if (ilSession::get(ilGlobalTemplate::MESSAGE_TYPE_FAILURE)) {
+            $messages[ilGlobalTemplate::MESSAGE_TYPE_FAILURE] = ilSession::get(ilGlobalTemplate::MESSAGE_TYPE_FAILURE);
         }
-        if (isset($_SESSION[ilGlobalTemplate::MESSAGE_TYPE_SUCCESS])) {
-            $messages[ilGlobalTemplate::MESSAGE_TYPE_SUCCESS] = $_SESSION[ilGlobalTemplate::MESSAGE_TYPE_SUCCESS];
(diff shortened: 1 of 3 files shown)


smeyer-ilias:
I changed the assignment, since it these changes affect Services/Initialisation

Services/Init/classes/class.ilStartUpGUI.php
```diff
@@ -271,6 +278,24 @@ public static function printToGlobalTemplate($tpl)
         $gtpl->setContent($tpl->get());
         $gtpl->printToStdout("DEFAULT", false, true);
     }
+    
+    protected function retrieveMessagesFromSession() : array
+    {
+        $messages = [];
+        if (isset($_SESSION[ilGlobalTemplate::MESSAGE_TYPE_FAILURE])) {
+            $messages[ilGlobalTemplate::MESSAGE_TYPE_FAILURE] = $_SESSION[ilGlobalTemplate::MESSAGE_TYPE_FAILURE];
```
smeyer-ilias: Please avoid the direct access to $_SESISON and use ilSession.

smeyer-ilias (changes requested):
Thanks for fixing this.

Services/Init/classes/class.ilStartUpGUI.php
kergomard: Done

mjansenDatabay:
IMO this is only an issue for `release_7`. This seems to be fixed in the meantime for ILIAS >= 8.x.

mjansenDatabay:
This solution differs from the (multiple) fixes in ILIAS >= 8.x (I am still not sure which appraoch I prefer :-)), but I'll accept this for `release_7`.

mjansenDatabay:
@kergomard Could you please resolve the conflicts and rebase/force-push your branch afterwards?

kergomard:
Thank you very much @mjansenDatabay for having a look at this. I rebased and force-pushed.

iszmais:
@kergomard Thanks for that improvment!
coincidentally i worked on the same problem yesterday i correlation of another Bug in ILIAS 8.

First comes first serves. I would revert my solution.
Could you provide this fix for release_8 and trunk, too?

Greetings,
@iszmais

kergomard:
@iszmais , I'm now unsure: Do we need parts of this for trunk or not? I can provide a separate PR, if needed, no problem.

iszmais:
@kergomard As far as i know the part wich provides the persistence of messages (so that they will not will be lost after deleting the session) is needed for 8 and trunk, therefore the first 3 additions of your PR.

I cant say anything for the other parts.

iszmais:
But as i would assume and , as noone seem to have a veto on that, that this should be fully Cerry picked in ILIAS 8 and TRUNK.

PR #4873, opened 2022-07-28
T&A: fix for 31161

okaufman:
Thsi will fix https://mantis.ilias.de/view.php?id=31161


Amstutz:
Hi @okaufman 

Thx a lot. We discussed this in the TA Squad today. Feedback:

- [ ] Private Const: Please move the public get Method to a private Const

Thx.

PR #4879, opened 2022-07-30
Make Previous Answers appear in Test with Manually Selected Pax

kergomard:
This fixes https://mantis.ilias.de/view.php?id=25184 . It came up as part of the ILIAS-Refinement effort.

Additionally the function contained an, as far as I can see, completely useless check for Javascript and was so overly complex that I had to simplify it, to make sure I understood. Please find the two parts in the separate commits.

As always: Should you accept this PR I will provide a PR against trunk.


mbecker-databay:
Looks good to me. Please add the PR for Release_8 or Trunk, so I can swoop this up in one go.

kergomard:
PR is here: #4990

PR #4881, opened 2022-08-01
Fix priority of mail settings between global and usersettings

iszmais:
https://mantis.ilias.de/view.php?id=31985

This will prevent the system from sending a mail with the user specific options if the configuration of these options is disabled for the user

### DIFF
--- a/Services/Mail/classes/class.ilMailOptions.php
+++ b/Services/Mail/classes/class.ilMailOptions.php
@@ -40,7 +40,7 @@ class ilMailOptions
     /** @var int */
     protected $emailAddressMode = self::FIRST_EMAIL;
     /** @var ilMailTransportSettings */
-    private $mailTransportSettings;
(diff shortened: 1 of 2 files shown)


mjansenDatabay (changes requested):
Hi @iszmais,

thx a lot for this PR.

I am not sure if we should make `read` public. `read` is one of the standard methods for this `ActiveRecord`-like class. and I doubt we would like expose this to the public, only for testing purposes.

Of course the general approach is debatable, too. I could discuss this with the JF.

Best regards,
@mjansenDatabay

iszmais:
@mjansenDatabay Fixed it.
Im not sure if i should move the anonym class inside an own file. I did not to prevent uneccessary requirements and loading.
Feel free to correct me on that

mjansenDatabay (approved):

PR #4882, opened 2022-08-01
Fix issue when copying DCL with MultiReference

kergomard:
See: https://mantis.ilias.de/view.php?id=28852

If you accept this I can provide a PR for ILIAS 8 and Trunk as well.

### DIFF
--- a/Modules/DataCollection/classes/Fields/Reference/class.ilDclReferenceRecordFieldModel.php
+++ b/Modules/DataCollection/classes/Fields/Reference/class.ilDclReferenceRecordFieldModel.php
@@ -168,15 +168,32 @@ class ilDclReferenceRecordFieldModel extends ilDclBaseRecordFieldModel
     {
         $field_clone = ilDclCache::getCloneOf($this->getField()->getId(), ilDclCache::TYPE_FIELD);
         $record_clone = ilDclCache::getCloneOf($this->getRecord()->getId(), ilDclCache::TYPE_RECORD);
-
+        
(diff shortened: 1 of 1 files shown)


Amstutz:
Thx a lot, fixed this as well for trunk and 8, but I like your version somewhat better, so I also manually picked this to 8 and trunk.

Modules/DataCollection/classes/Fields/Reference/class.ilDclReferenceRecordFieldModel.php
```diff
@@ -168,15 +168,32 @@ public function afterClone()
...
+            } elseif (!is_null($temp_value = $this->getCloneRecordId($clone_reference))) {
+                $value = $temp_value;
             }
+
+            $this->setValue($value, true); // reference fields store the id of the reference's record as their value
+            $this->doUpdate();
+        }
+    }
+    
+    protected function getCloneRecordId(string $clone_reference) : ?string
```
mjansenDatabay, line 192: @Amstutz / @kergomard This will result in `Return value must be of type ?string, none returned` if the condition evaluates to `false` (at least with 8.1).

Amstutz: thx bunch! ill fix asap.

Amstutz: fixed for 8 and trunk.

PR #4883, opened 2022-08-02
UI item shy button alignment 33324

catenglaender:
Mantis issue https://mantis.ilias.de/view.php?id=33324

# Issue

image

Shy buttons as value in properties is not aligned with the baseline of the key text.

This PR patches shy buttons inside of the UI item only to be aligned with the baseline of the surrounding text.

# Big picture issue still needs a decision

Despite looking like a text link on the inside, Bootstrap 3 (as well as the current Bootstrap version 5) very intentionally models the surrounding of the shy button (btn-link) like a real button - same padding, height and middle alignment so it seamlessly could be in a btn-group with "real" buttons and the text of all button types would be aligned.

We already broke Bootstrap's original intention by removing the padding in ILIAS button.less, which I suspect was done so the shy button could align better with normal text. However, it still doesn't align perfectly, so at the moment it's aligned with neither other buttons nor normal text.

The question is what route do we want to go by default? Should shy buttons look exactly like text links so they could be seamlessly part of a text line (like in the case of the item here)? Or should they have the spacing and padding of a normal button? This could be a system with exceptions, but I think it needs to completely follow one of the two options by default.

Maybe we should discuss this in the CSS Squad?


Amstutz:
Thx a lot for the fix @catenglaender !

Special thx for picking up on the bigger question behind it. As so often, I do not have the answer, but I hope we find it together in the squad today, I put it on the list to be discussed.

Thx!

PR #4884, opened 2022-08-02
UI Communication Item

fhelfer:
This PR proposes to introduce the UI-Component for implementing Communication Items
This item will initially be used for the new presentation of mails and forum postings

As discussed in https://docu.ilias.de/goto_docu_wiki_wpage_7421_1357.html mails should be represented in a view that is similar to standard item. 
Communication Items are introduced so that mails and forum postings can be presented in a view where contextual information is presented first.

### DIFF
--- a/src/UI/Component/Item/Communication.php
+++ b/src/UI/Component/Item/Communication.php
@@ -1,0 +1,45 @@
+<?php declare(strict_types=1);
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+namespace ILIAS\UI\Component\Item;
+
+use ILIAS\UI\Component\Symbol\Icon\Icon;
+use ILIAS\UI\Component\Symbol\Avatar\Avatar;
+
+/**
+ * Interface Notification
+ * @package ILIAS\UI\Component\Item
+ */
+interface Communication extends Item
+{
+    /**
+     * Set icon as lead
+     */
+    public function withLeadIcon(Icon $icon) : Communication;
+
+    /**
+     * Set avatar as lead
+     */
+    public function withLeadAvatar(Avatar $avatar) : Communication;
+
+    /**
+     * Get lead Icon or Avatar
+     * @return null|Icon|Avatar
+     */
+    public function getLead();
+}
--- a/src/UI/Component/Item/Factory.php
+++ b/src/UI/Component/Item/Factory.php
@@ -136,4 +136,42 @@ interface Factory
      * @return \ILIAS\UI\Component\Item\Notification
      */
     public function notification($title, Icon $lead) : Notification;
+
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *      The contextualisation of an item is prior to the content itself
+     *   composition: >
+     *      Communication items focus on a message to be communicated.
+     *      For a message, its context is of great relevance.
+     *      This context is given by information about the sender
+     *      and the receiver(s) of the message, as well as the time
+     *      at which the message was communicated.
+     *      This is followed by the main message content.
+     *      This can be supplemented by further information such as
+     *      attachments, references, ratings or status informations.
+     *   effect: >
+     *      Interactions can be present in the contextual information,
+     *      the content and metadata.
+     * context:
+     *     - Contexts can be single messages such as Mails, but also several related messages
+     *      such as in forums.
+     * rules:
+     *   usage:
+     *     1: Communincation Items SHOULD be used if it is a message.
+     *   ordering:
+     *     1: An lead icon or lead avatar can be used
+     *     2: Title
+     *     3: Context Informations
+     *     4: Message content
+     *     5: Metadata
+     *   accessibility:
+     *     1: All interactions offered by a notification item MUST be accessible
+     *      by only using the keyboard.
+     * ---
+     * @param $title
+     * @return \ILIAS\UI\Component\Item\Communication
+     */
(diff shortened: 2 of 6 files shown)


klees:
I'm on it, I think I'll need until the end of the week to write a reply.

klees (changes requested):
Hi @fhelfer,

Thanks for providing this PR. If we understand correctly, you are just the one who created the code, but the content was mostly provided by @oliversamoila. So, you might not be the right addressee for this feedback =)

We still want to start with something that might be for you: This looks very neat and tidy, go on like this. Thanks!

Regarding the proposal: We are not convinced that this component is a good fit for the `Item` category in the UI framework. Although visuals might be similar to the `Standard Item` for the component that you propose, we are looking to model our components according to semantic considerations, not visuals ones. In that regard, a `Communication Item` seems rather different from what the `Items` should embody. We will thus propose to create a new semantic category for the component you are looking to build. All of this requires some explanation, so let's go.

Let's begin by reiterating one important observation: The complete `Item` hierarchy turned out to be problematic almost since it was first created. This was discussed on multiple occasions, is [documented in the ROADMAP](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/UI/ROADMAP.md#balance-or-unify-cards-and-items) and recently spawned this [paper from @fenglaender](https://github.com/ILIAS-eLearning/ILIAS/pull/4916). This is an important reason why we have tried to keep new components out of the category for quite some time, since we would be getting even more stuff and disorganization if we accepted more "items". We must admit that we failed on several occasions, but since there is no time pressure or any other external factors for this proposal, we are stepping on the break here.

Even if we disregard the general problems and take the current description of `Item` at face value, this new "item" is not a good fit for the category. The description states that ["An item displays a unique entity within the system. It shows information about that entity in a structured way."](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/UI/Factory.php#L538). Let's unpack that a little. [Merriam-Webster states, that an entity is an "independent, separate, or self-contained existence", " the existence of a thing as contrasted with its attributes"](https://www.merriam-webster.com/dictionary/entity). As far as we understand, a "Communication Item" is intended to be used to display mails and forum postings (we don't really understand "purpose" from the KS entry, tbh).

Does the "mail" or a "forum posting" fit the description of entity, given above? We think: No. If the sender of a mail changes, it is not the same mail. If the content of the mail changes, it is not the same mail. If the date of a mail changes, it is not the same mail. There mostly seems to be no existence of a thing called "mail" that transcendents its concrete attributes. The same goes for forum posting.

From our perspective, a key question to be asked to find out if something is an "entity" in that sense is as such: If I have two things and I want to find out if they are "the same": Will I compare them by equality or by identity? If I compare them by equality, they are not entities. If I compare them by identity, they are. If my spectacles break, I have an insurance that grants me to replace them with the same spectacles. But these won't be the identical pair, they will just be equal to the pair I currently have. If I decide to replace the transparent glasses in my spectacles with shaded glasses, the spectacles will be the same, since they are identical. So, we need to find out how we discuss stuff regarding "sameness". This might be dependent on some context, though, but for mails and forums there hardly is any context where we would say that "sameness" is not exhaustively described by "equality".

This might feel rather philosophical, but we can see some more concrete problems when we try to match a "mail" to `Item` (all taken from the Feature Request that you have linked):

* The field that you have oddly called "Mail Title" is not the "title" (as in ["distinguished name"](https://www.merriam-webster.com/dictionary/title)) of that mail. It is the "subject", as in "the thing that the mail is about".
* The avatar that can be attached to the "item" is about one property (the sender of the mail) and not about the "item" (as it would be, e.g., if the `Item` was used for a user or a course).
* The text of the mail seem to be put in the "description" field. But it is not a "description" at all, it is THE most important part of the mail, the text.

Hopefully, this clarifies why the problem is not philosophical only at all and highlights why we want to focus on semantics of components in the ILIAS UI framework before we look into visuals: Only if we know semantics of the components, we can make good choices regarding the visuals. Let's take the mail from the feature request as an example again. The fields in the mockup are:

* the subject (oddly named "mail title")
* the text of the mail ("description")
* the avatar of the user who has sent the mail
* the name of the user who has sent the mail
* the date the mail was sent at
* "other information" 1
* "other information" 2

Think about how we would handle this information when space on the screen gets scarce. If we'd handle the mail as an `Item` we would hide the "description" rather early because it is just a support for the title. But we would have hidden the most important part of the mail, then. We would have no way to understand that we could hide the "avatar" rather early because the information is duplicated in the other field. Since all the fields "username", "date", "other info 1" and "other info 2" are just properties, we had no way to decide if one is more important than the other. This, however, is not a problem of this proposal in particular, but at the core of the problem we see with the general `Item` category.

To contain the mess of `Item` and add more semantics to your component, we would propose to simply create a new top-level category and, maybe, call it `Message` (what is a "communication item", anyway). The thing you are looking to create here could then be added as (e.g) `Simple (Message)`. We could then later add a `Listing` (for multiple messages one after another) or a `Thread` (where `Messages` show a back and forth in a tree like structure). We might well discuss if this concrete semantic structure is something we want or something that will lead to good results in the future. Furthermore, we would be happy to schedule a meeting to do weigh pros and cons (contact us!). But even if we implement this first (and maybe naive) idea, we will be better off then putting this into `Item`:

* The `Simple Message` could get descriptive and fitting field and property names (e.g. "Subject", "Text", "Author", ...).
* We will have a handle to style the `Message` according to requirements for it, if the need arises. There already seems to be hints for that requirement regarding the avatar, looking into the FW entry). We won't be tied to the `Item` then.
* We won't add another (non-)`Item` to the other hierarchy, so we'll be able to move both hierarchies independently and hopefully break down `Item` into more digestible parts at some point.
* You could look into a functional KS-description of the thing you want to create, since you could just drop the language that is just required to make the thingy fit the `Item` hierarchy.

As a final note, please have a look into [this proposal](https://github.com/ILIAS-eLearning/ILIAS/pull/4916), which addresses the problems mentioned here in a little more general way, and take part in the discussion. We can do better than what we have with the `Item`.

Note that even if we create a new semantic identity, PHP Logic might be reused by using traits if applicable and CSS and DOM logic by reusing classes/mixins/templates. The new component might be named different, however it still might behave and look similar to what we already have.

Best regards!
@Amstutz and @klees

klees:
@fhelfer @Amstutz @oliversamoila

After todays discussion: Can we close this PR?

Amstutz:
@fhelfer , you may Re-Open as needed.

PR #4877, opened 2022-07-29
Implemented PSR12 according to PR #4533

catlauraherzog:
Pull Request #4533 stated to update the Coding Standard to PSR12. I implemented it within the PHP CS Fixer. If anything needs to be changed, let me know.

PR #4886, opened 2022-08-02
Prepare ILIAS 9 development (docs, tools, composer deps, workflows)

mjansenDatabay:
This PR prepares the ILIAS code base for the ILIAS 9 development by ...

- ... bumping the version number (where relevant)
- ... updating the `Composer` dependencies based on the udpated supported PHP versions
- ... removing PHP 7.4 (where relevant)
- ... providing a new `PhpStorm` code style config and inspection profile (many changes have been made automatically by the latest `PhpStorm` versions, so there might be a need for further changes, other contributions are appreciated if this case). These files rely on a merge of the`PSR-12` related PRs #4877 #4533.


Amstutz:
We discussed this today in the TB. @mjansenDatabay is glad for further feedbacks concerning the profile. Note that we decided that it definitely will be **PHP 8.0 and 8.1 for ILIAS 9**.

PR #4887, opened 2022-08-02
Fix update errors in DB-Steps

kergomard:
I tried to update a (maybe rather broken, but basically running) version of ILIAS 7 to ILIAS 8 and stumbled over a few issues during DB-Update.


chfsx (approved):
thank you very much! interesting that i never encountered these errors with all the test upgrades. I'll cherry-pick to trunk as well

PR #4889, opened 2022-08-03
DEBUG: Remove global constant

mjansenDatabay:
This PR removes the global `DEBUG` constant and its usages. The effect and consequences of this DEBUG state are not documented (in contrast the `DEVMODE` is explained here: https://docu.ilias.de/goto_docu_pg_1082_42.html). It is unclear why (what's the purpose?) certain template file names are listed at the top of the rendered HTML page if the state evaluates to `true`.

The builds fail because of the wrong `License Header`, so no serious problems occur.

### DIFF
--- a/Modules/File/test/ilModulesFileTest.php
+++ b/Modules/File/test/ilModulesFileTest.php
@@ -15,7 +15,7 @@
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
- 
+
 use PHPUnit\Framework\TestCase;
 use ILIAS\Filesystem\Stream\Streams;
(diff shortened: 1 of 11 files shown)


klees (approved):

Amstutz:
TB: please fix the checks and feel free to merge. We discussed this in the TB today.

mjansenDatabay:
@Amstutz I fixed the licensed headers in the related files, except for the `setup/client.master.ini.php` file. There is an inline PHP starting/ending tag sequence, so in this mixed INI/PHP file the license header could not be easyly applied.

If there are no objections within the next hours, I'll merge this.

PR #4890, opened 2022-08-03
BF 0033595: section-input-header should use @il-font-weight-bold

ezenzen:
Hi @all,

as discussed in the css squad, headings in forms should use the variable @il-font-weight-bold. For legacy forms we have already adjusted this. This PR is used to adjust the headings of KS forms.

0033595: section-input-header should use @il-font-weight-bold - https://mantis.ilias.de/view.php?id=33595

Many Greetings, 
Enrico
Screenshot 2022-08-03 at 08-56-10 ILIAS Study Programme


Amstutz:
Thx a lot, looking good!

Amstutz:
picked to trunk as well.

PR #4892, opened 2022-08-03
Updated PHP-version check to match install.md

chfsx:
This fixed the PHP-Version check to match the supported version in install.md

### DIFF
--- a/src/Setup/Condition/PHPVersionCondition.php
+++ b/src/Setup/Condition/PHPVersionCondition.php
@@ -27,7 +27,7 @@ class PHPVersionCondition extends ExternalConditionObjective
         return parent::__construct(
             "PHP version >= $which",
             fn (Setup\Environment $env) : bool => version_compare(phpversion(), $which, ">="),
-            "ILIAS 6 requires PHP $which or later."
+            "ILIAS 8 requires PHP $which or later."
         );
     }
 }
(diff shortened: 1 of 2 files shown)


src/Setup/Condition/PHPVersionCondition.php
```diff
@@ -27,7 +27,7 @@ public function __construct(string $which)
         return parent::__construct(
             "PHP version >= $which",
             fn (Setup\Environment $env) : bool => version_compare(phpversion(), $which, ">="),
-            "ILIAS 6 requires PHP $which or later."
+            "ILIAS 8 requires PHP $which or later."
```
klees, line 30: I think we should be able to use some `const` here, right?

klees (approved):

src/Setup/Condition/PHPVersionCondition.php
chfsx: Definitely better for the future

mjansenDatabay: `ILIAS_VERSION_NUMERIC` could be used, but of course this is global scope :-). This could be either passed as a dependency or a new redundant constant could be introduced.

klees: This might be one of the few places where a global constant would actually be ok(ish).

klees:
I made the adjustment regarding the usage of the constant and picked this manually:

https://github.com/ILIAS-eLearning/ILIAS/commit/35e7461c70852090d8a305b514ba838bcea94916
https://github.com/ILIAS-eLearning/ILIAS/commit/a3a0308fe7d535dcd3e14a03481524e89c703dfc

klees:
Thanks a lot =)

PR #4901, opened 2022-08-05
Development Proposal: Text Handling

klees:
Dear colleagues,

after @nicoroeser [brought up the question about text handlling in ILIAS](https://docu.ilias.de/goto_docu_frm_5826_8136.html) (once again...) we had a [productive meeting with various interested developers](https://docu.ilias.de/goto_docu_sess_11733.html) on the initiative of @dsstrassner.

This proposal tries to answer the questions and requirements brought up in the meeting by proposing a tool set for text handling in ILIAS.

Best regards!

### DIFF
--- a/docs/development/text-handling.md
+++ b/docs/development/text-handling.md
@@ -1,0 +1,278 @@
+# How to Handle Text in ILIAS?
+
+As an application with content management functionality, ILIAS handles text in
+various locations: on handcrafted pages, in mails, in feedbacks, comments and
+forums and in various other locations.
+
+Currently, text in ILIAS is not abstracted in any way: It is passed as string.
+This causes various problems, especially when text is passed between components,
+that pop up from time to time. Does this text contain formatting in HTML? Can
+it be used with or without html escaping, or was HTML escaping already done by
+the other component? Or will it be performed later? Can I output my text in this
+or that context without introducing security problems, formatting problems or
+what not?
+
+This paper looks to propose an idea how this situation can be improved.
+
+
+## What do we Mean by "Text"?
+
+Certainly, we are not looking to target any `string` in the ILIAS codebase. There
+are various locations where `string`s are used for other purposes: as ids, as part
+of controlled vocabularies, as input from users that needs to be parsed. So what
+do we really target when we say "text" in this paper?
+
+**We understand text to be sequences of words of an unspecified length, possibly
+containing information about its structure (like paragraphs, headings or lists).
+The information content of these sequences of words won't be of interest for the
+application, but the structure is required to derive formatting for various output
+contexts.**
+
+Although this might leave some gray areas, this definition could be used to decide
+if a given piece of information is or isn't text, e.g.:
+
+* **login names of users**: *no text*, because just one word (no spaces...) and no
+further structure possible. Also, the contained information is of interest for the
+application since we need it to derive which user just logged in.
+* **content of a mail**: *text* because certainly structured words are possible and
+common. The actual information content does matter for the receiving person, but not
+for the application.
+* **titles of objects**: *might or might not be text*, because we support a little
+structuring (bold and italics) currently, but it is debatable if we really want to
+do so in the future.
+
+
+## Requirements for Text Handling
+
+To provide ILIAS developers with a tool set to solve the outlined problems with the
+current `string`-based text handling approach, a new approach needs to implement
+the following requirements:
+
+* For a given piece of text it should be known at any time which markup is used
+in the underlying string representation of the text.
+* For a given piece of text it should be known at any time which structure elements
+could be used in the text.
+* On programmatic interfaces it should be possible to specify which structure and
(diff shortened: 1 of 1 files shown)


klees:
Dear JF,

it would be nice if I could pitch this shortly on 08-08, so people can ask questions, propose changes, voice concerns, etc. If this still looks valid after 2 weeks, it would be nice to discuss and hopefully adopt this on 22-08.

Thanks!

matthiaskunkel:
Jour Fixe, 08 AUG 2022 : Richard explained his intention for this PR and invites all developers to discuss this proposal and to ask questions or post change requests. PR will be discussed again on the next JF at 22 AUG 2022.

docs/development/text-handling.md
```diff
@@ -0,0 +1,278 @@
...
+To make it possible to programmatically talk about structuring options for text,
+a central `class` (and later `enum`, once supported for PHP < 8.1 is cut) defines
+the options that we care about:
+
+```php
+class Structure
+{
+    const string BOLD = "bold";
+    const string ITALIC = "italic";
+    const string UNORDERED_LIST = "unordered_list";
```
mjansenDatabay, line 86: For me it is unclear how this list is maintained. If there are texts created with the TinyMCE (e.g. in the T&A or the forum) there might be dozens of possible allowed tags, depending on the dynamic configuration in the ILIAS administration.

Is this `Structure` a shared class? How can additional tags be suggested/added?

docs/development/text-handling.md
```diff
@@ -0,0 +1,278 @@
...
+
+    /**
+     * @throws \InvalidArgumentException if $text does not match format. 
+     */
+    public function toPlainText($text) : PlainText;
+
+    public function getMarkup() : Markup;
+
+    /**
+     * @return TextStructure[]
```
mjansenDatabay: Could give an example wthat this type could be and which kind of data it encapsulates?

docs/development/text-handling.md
klees: Yes, this would be a shared class. We would suggest additions via PR. Also, I would not want to target "everything that TinyMCE now does", but rather "a sensible list of text structures we could really support".

Also, I rather deliberatly tried to avoid the word "formatting" like hell (hope I suceeded) and talked about "structure" instead. AFAIK, many things that the Tiny does are about formatting, i.e. saying something like "make this 11px big", instead of giving structure, i.e. "this is a headline". I would target structure, not formatting. I see that "bold" and "italic" is questionable in that regard, though, but at least it is not about some defined font-weight, or something similar. We might find better names, though, and the line might be blurry in general.

The reason I would avoid talking about formatting is, that formatting mostly only makes sense in some context. Since we have a web app, this context can change easily, e.g. when switching from small to large display. So the formatting would be defined by some CSS (as it is supposed to be), while the text ist marked up with some structure information to do so.

docs/development/text-handling.md
klees: This is just wrong. Currently this would be consts from `Structure`. It could become an `enum` at some point. Will correct...

klees: :heavy_check_mark:

docs/development/text-handling.md
kergomard: I very much like the approach to avoid anything resembling "formatting". The distinction might sometimes be hard, though, as you already point out in the case of bold and italic. It could be a way out to seek inspiration in the semantic elements defined in the ILIAS-Editor, thus avoiding anything with the slightest smell of formatting.

kergomard:
Thank you very much @klees for this proposal!
I believe this to be a big step forward in the right direction.
- As stated in my comment above, I think it is a very good choice to avoid any formatting-related language and promises in here.
- I believe the choice of baseline representations to be sound. You state there that "[t]hese conversions can be lossfull". Currently I think this should explicitly not be true for the conversion to HTML. I can see cases where people could want that to add machine readable elements in a text, but I currently think that it would be useful to explicitly prohibit this case, in order to separate content for a user (ie. text) from content for a machine (ie. programming). All other conversion should explicitly be allowed to be incomplete as stated now.
- On a completely other level: As stated in another context I feel uncomfortable right now with the plan to implement this in a Markdown field for the kitchen sink. Personally I would prefer to first analyze the space of texts we need to cover in ILIAS and make sure that we can cover that with a maximum of two different user facing ways to input these texts into the system. I'm currently unsure if this is going to be possible with markdown and it makes me nervous to think that we would be permanently adding one more input method without getting rid of existing ones.

Thanks again and best!

klees:
@matthiaskunkel I won't be available on JF 22-08-22 for private reasons, please postpone this to the next JF.

matthiaskunkel:
Okay. Then we talk about it on the JF at September 19.

matthiaskunkel:
Jour Fixe, 19 SEP 2022 : We highly appreciate this suggested and ask Richard to complete the proposal for text handling.

PR #4903, opened 2022-08-05
Forum: Fix notification event settings

mjansenDatabay:
This PR ...

- ... fixes PHP 8 issues in the notifcation settings if a forum is a child object of a course or a group.
- ... uses the UI forms instead of legacy forms
- ... removes code that could never have worked


mjansenDatabay:
FYI @nmatuschek

PR #4905, opened 2022-08-05
Hides Action Row on RecordListTables where no Action is possible.

kergomard:
See: https://mantis.ilias.de/view.php?id=25669

If you accept this I can provide a PR for ILIAS 8 and Trunk as well.


Amstutz:
Thx a lot!

Amstutz:
picked to 8 and trunk as well.

PR #4908, opened 2022-08-05
Tree: Add missing tree impl. switch (sponsored by Mrs. Jansen)

mjansenDatabay:
This PR adds an `Achievable Setup Objective` to let administrators switch the `Tree Implementation` of the repository tree from a `Nested Set` implementation to a `Materialized Path` implementation and vice versa.

Debatable: Should we stick with the suggested objective class or should we segregate this into two separate objectives?

Using `ilTree` to migrate to `Nested Set` was not possible because of massive global scope usage, so I had to duplicate a few lines of code.

---

Furthermore this PR adds a `Metrics Collected Objective` to report the current tree implementation when executing the `status` command of the ILIAS setup.

---

- The builds fail because of the missing license header in `dbupdate_05.php`.
- If approved, this should be picked to `release_8` as well.

Mantis Issue: https://mantis.ilias.de/view.php?id=32277

### DIFF
--- a/Services/Tree/classes/Setup/class.ilTreeImplementationSwitch.php
+++ b/Services/Tree/classes/Setup/class.ilTreeImplementationSwitch.php
@@ -1,0 +1,147 @@
+<?php declare(strict_types=1);
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+use ILIAS\Setup\Environment;
+use ILIAS\Setup\UnachievableException;
+use ILIAS\Setup\NullConfig;
+
+class ilTreeImplementationSwitch extends ilSetupObjective
+{
+    public const NESTED_SET_TO_MATERIALIZED_PATH = 'nested_set_to_materialized_path';
+    public const MATERIALIZED_PATH_TO_NESTED_SET = 'materialized_path_to_nested_set';
+
+    private string $mode;
+
+    public function __construct(string $mode)
+    {
+        parent::__construct(new NullConfig());
+        $this->mode = $mode;
+    }
+
+    public function getHash() : string
+    {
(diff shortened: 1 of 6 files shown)


Services/Tree/classes/Setup/class.ilTreeImplementationSwitch.php
```diff
@@ -0,0 +1,141 @@
...
+        $childs = [];
+        while ($row = $res->fetchRow(ilDBConstants::FETCHMODE_OBJECT)) {
+            $childs[] = (int) $row->child;
+        }
+
+        foreach ($childs as $child) {
+            $i = $this->renumberNestedSet($db, $child, $i + 1);
+        }
+        $i++;
+
```
smeyer-ilias, line 120: @mjansenDatabay the rgt value update is missing here.

smeyer-ilias (changes requested):
Thanks @mjansenDatabay
Please add the rgt-value update fpr nested set trees.

Services/Tree/classes/Setup/class.ilTreeImplementationSwitch.php
mjansenDatabay: @smeyer-ilias DONE

chfsx:
As a technical board, we regularly check whether pull requests have been open for a long time.
Are there any updates on this? Please note that you can also close this request if you are not able or if you do not have the resources to look into it in detail.

At this point we would like to refer to the rules we all agreed on: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/contributing.md#rules-for-maintainers-assigned-to-prs.

With kind regards!

Services/Tree/classes/Setup/class.ilTreeImplementationSwitch.php
```diff
@@ -0,0 +1,147 @@
...
+        return $environment;
+    }
+    
+    private function renumberNestedSet(ilDBInterface $db, int $node_id, int $i) : int
+    {
+        $db->manipulateF(
+            'UPDATE tree SET lft = %s WHERE child = %s',
+            [ilDBConstants::T_INTEGER, ilDBConstants::T_INTEGER],
+            [$node_id, $i]
```
smeyer-ilias, line 105: I've switched these parameter. The result was quite surprising :)

smeyer-ilias: Thanks, to both of you
I 've cherry-picked your commits to R8 and will close this PR.

smeyer-ilias:
Cherry-picked to R8 and trunk

smeyer-ilias (commented):
See comments

PR #4909, opened 2022-08-07
Update Cloze Question to Allow Curly Braces

kergomard:
This is a fix for https://mantis.ilias.de/view.php?id=19117
I tried different solutions, among others the one reommended by the JF on 04 JUN 2018. After playing around with it, I don't believe it to be the right way to go as it would need a lot of changes and would make the treatment of strings even more convoluted than it is already. At the end I settled on this less than ideal but rather small workaround. This just works around the restrictions of our templating engine.


dsstrassner:
As JF only recommend a possible solution, I think we do not need a new JF decision, if we accept this PR. 

Thanks a bunch @kergomard

PR #4911, opened 2022-08-08
Test: Bugfix 0033693: Finishing Time cannot be changed after the start of the test (release_8)

fhelfer:
This fixes: https://mantis.ilias.de/view.php?id=33693


mjansenDatabay (commented):
My thoughts:

In general this PR looks good IMO, but I am unsure if we should go this way. This is not an atomic PR. In this PR, the legacy form elements `ilDateDurationInputGUI` and `ilDurationInputGUI` are modified. These elements are not part of the `T&A` components and belong to a separate maintenance, so they have to be reviewed by another maintainer (AFAIK @smeyer-ilias is responsible for the `Date/Time` related elements). The legacy form elements do not support `Unit Tests`, so it is hard to judge/verify whether these changes here will break existing behaviour.

My preferred solution:
- Apply a `Rollback` to the previous state of handling the start/end of the day. So keep the UI `as it was`.
- Add additional information to the `Byline` that the end date has to be **after** the start date.
- There should be just a new plausibility check as a constraint applied to start/end, not more, not less. If there is a violation, there **MUST** be a general (form related) and a specific (field related) error message (as usual).

PR #4912, opened 2022-08-08
Fix Profile Backlink in COPage

kergomard:
See: https://mantis.ilias.de/view.php?id=29719


alex40724:
@kergomard Thanks for the PR. Merged to all current branches...

PR #4918, opened 2022-08-09
PageEditor/MediaPool: Change access check

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=33756


oliversamoila:
Mantis-Issue: Consuming media from media pool requires the right "Edit settings" on the media pool
https://mantis.ilias.de/view.php?id=33756

alex40724:
No, it is correct to check for "write" access, see current business rules. I will link them in den bug report.

PR #4919, opened 2022-08-09
UI Key-Value Tree Node

schmitz-ilias:
This PR proposes the addition of a new type of tree node to the KS besides 'simple' and 'bylined', as proposed by @klees in [this jour fixe](https://docu.ilias.de/goto_docu_wiki_wpage_7593_1357.html). The new 'paired' node contains a value-string rendered behind the node label, forming a key-value pair. This value can then be used to distinguish between different nodes, or clarify their function, when the label alone is not sufficient.

The value is rendered to the right of the node label in a smaller font size, cursive, and in grey, to distinguish it as much as possible from the label (but maybe I went a bit overboard there...). See the image below for an example.

tree_with_paired_nodes

PS: The unit tests for this PR fail due to errors in the tests of a different component, as far as I can see the tests for the UI go through fine.


Amstutz (changes requested):
Hi @schmitz-ilias 

Thank you very much and congratulations on your first PR proposing a new UI component. More importantly, this is a really good showcase for the way UI Components are designed to work: We together identified a new `Tree Node` with slightly different semantics. You proposed a design fitting the new semantics. Overall, the look seems much improved comparing it to the `Bylined Tree Node` for the desired use case. Nice!

As it is most usual, some things to improve on...

- [ ] Label vs. Title: When looking into the new Node I wondered how it really compares to the other existing node types. I then noticed, that the `Simple Tree Node` and the `Bylined Tree Node` also are about a "label" on the node, which now feels odd since the label is something different for both cases: For the `Simple Node` it is a label for the whole node, while for the `Paired Node` here, the label seems to be a label for the value. I then looked up the meaning of "label": [a piece of paper, etc. that is attached to something and that provides information about it](https://www.oxfordlearnersdictionaries.com/definition/english/label_1?q=label). So, basically, the usage of "label" seems to be wrong here anywhere, if we take it figuratively, the label would be thing made up from the icon, the "label" (and the byline), not just the, well, "title" of the node. So I kindly ask you to fix a former mistake and: rename the according field to "title" for the `Simple Node` and `Bylined Node`, remove the inheritance for the `Paired Node` and rename the "label" to "key" for said node. Inheritance is odd here anyway: A `Paired Node` is not a `Simple Node`, but the current inheritance says just that.
- [x] Rival: Please add a ["rival" field](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/UI/docu/rules.md#interfaces-to-factories) and describe the difference to a "Bylined Node". From my perspective, the main difference is: For a bylined node, the "title" (see above) should be enough to identify the node, and the byline only adds some additional information. The `Paired Node`, on the other hand, can only be identified by the key together with the value.

- [x] Naming: Let's just name this node `KeyValue` instead of `Paired`. This seems to be a lot more hints for the use case.
- [x] Effect: The field is supposed to describe ["visual effects that relate to the item"](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/UI/docu/rules.md#interfaces-to-factories). Please remove the current "effect" description, it does not seem to fit.
- [x] Testcase: Please propose some testcases that we could add to [Testrail](https://testrail.ilias.de/index.php?/suites/view/390). We are very sorry that you are the first person that we ask to do so, but we only became aware again very recently that we ought to do so (see [README](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/UI/README.md): "Remember to adapt/add the Test Cases in [Testrail section UI Components](https://testrail.ilias.de/index.php?/suites/view/390) so that a tester with no technical expertise can confirm that all examples work as intended. They must be available and linked to the PR before the PR will be merged."). Frankly, we simply forgot about this for some time, but now we try to do better.

If there are any questions or concerns, feel free to ask here or contact us via our [Discord](https://discord.gg/bJFJwUMVHn) or any other channels.

Greetings!
@klees and @Amstutz

schmitz-ilias:
I implemented the changes you requested @klees, @Amstutz. I have added two testcases for the Key Value nodes to Testrail, [Key Value anzeigen](https://testrail.ilias.de/index.php?/cases/view/49987) and [Key Value validieren](https://testrail.ilias.de/index.php?/cases/view/49988). Unfortunately they do already show up in the current test run, so I marked them in the title as not to be tested for R8, and set their priority to 'Don't Test', but I'm not sure whether they nevertheless affect the statistics of the run. Is there any way to separate newly proposed testcases from rest?

I did not implement your suggestion to rename the node labels to 'title' or 'key' respectively, and to remove the inheritance of the Simple node from the Key Value node. Removing the inheritance would make it necessary to re-implement icons, links and the async loading functionality in the Key Value node, and doing so cleanly would require more restructuring of the implementation of all nodes than would be appropriate for this PR. I think the changes you suggested are worthwhile (maybe one could even remove the inheritance from Simple nodes to Bylined nodes, and move the icon, link and async loading functionality to bespoke traits?), but are in my opinion better positioned in a second, follow-up PR to this one.

I hope my changes are satisfactory, let me know if there is anything else I can improve on.

fab-kru:
> I have added two testcases for the Key Value nodes to Testrail, [Key Value anzeigen](https://testrail.ilias.de/index.php?/cases/view/49987) and [Key Value validieren](https://testrail.ilias.de/index.php?/cases/view/49988). Unfortunately they do already show up in the current test run, so I marked them in the title as not to be tested for R8, and set their priority to 'Don't Test', but I'm not sure whether they nevertheless affect the statistics of the run. Is there any way to separate newly proposed testcases from rest?

Hey @schmitz-ilias, thanks for the new test cases for ILIAS 9. I adapted the test run for 8 so that they won’t turn up for 8 anymore.

Cheers
Fabian (Test Management ILIAS open source e-Learning e.V.)

klees:
Thanks @schmitz-ilias, we are good to go for the JF.

Please be so kind and add the item regarding the restructured implementation to the ROADMAP.md in src/UI so we can keep track about the plan. Or will you providing the mentioned PR soon after this one?

schmitz-ilias:
I added a section describing the proposed restructuring to the roadmap. Thanks!

matthiaskunkel:
JourFixe, 22 AUG 2022 : We highly appreciate this suggestion and accept the PR for trunk. Please merge!

PR #4930, opened 2022-08-12
Composer/CaptainHook: Install the `Cptn` when using `Composer`

mjansenDatabay:
This commit adds the `captainhook/plugin-composer` to our development dependencies. With this we can ensure that every developer will install it and use our shared hooks. We tried to achieve this manually with https://github.com/ILIAS-eLearning/ILIAS/pull/3724 and https://github.com/ILIAS-eLearning/ILIAS/pull/3839, but relying on this plugin is the superior approach.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,8 @@
 		"vendor-dir": "./libs/composer/vendor",
 		"allow-plugins": {
 			"simplesamlphp/composer-module-installer": true,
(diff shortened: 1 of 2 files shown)


matthiaskunkel:
JourFixe, 22 AUG 2022: We highly appreciate this suggestion and accept the PR for ILIAS 9. Please merge into trunk.

PR #4916, opened 2022-08-08
UI Framework doc: UX Guide for properties and actions on repository objects

catenglaender:
We propose the guidelines in this PR as best practices for how to design and arrange properties and actions on repository objects to provide an intuitive, friction-less user experience.

We collected strategies that can assist developers and concept designers to create UIs that guide the user to the information and actions they are currently looking for while avoiding an overwhelming or cluttered presentation.

The main document describes relevant UX concepts like Information Architecture, User Intent and Mental Models and how they can be applied in the context of ILIAS repository objects and their elements.

A brief introduction to the topic of UX strategies and a link to the main document of this PR has been added to the Readme of the UI Framework.

We are interested in everyone's comments and are of course open to answer any questions about this PR. We would like to bring this proposal before the JF for approval, rejection or change requests.

This research and the proposal are part of a collaboration by CaT Concepts and Training GmbH and the University of Bern. Input and contributions by @yvseiler, @Amstutz, @nhaagen, @klees, @catkrahl shaped this proposal. Special thanks to the people who participated in the UI Item Workshop at CaT and @oliversamoila and Fabian Kruse for providing feedback as well as @jelieng for proof-reading.

PR #4932, opened 2022-08-12
Add kitchen sink element: LightboxCardDetailPage

lscharmer:
This PR adds a new element `LightboxCardDetailPage` to the UI Framework.
For the Feature Wiki: https://docu.ilias.de/goto_docu_wiki_wpage_5563_1357.html the currently available modal's do not satisfy the requirements for the proposed Feature. So this PR introduces a new lightbox modal which can be used to implement the feature in the future.

##### Why a new modal is needed:
The modal in the feature wiki **can** be implemented with existing modals like the `RoundTrip` or `Lightbox` + `LightboxTextPage`.
The issue with the `RoundTrip` is that there must be 2 buttons and that it is intended to be used for a sequence of steps. For the previously named feature a roundtrip with a `Close` and an `Add to profile` button and a sequence of `one` step does imo not respect the intention of the roundtrip modal.

For the `LightboxTextPage` the description and rules seem to loosely match the purpose of the modal that should be implemented in the feature but imo it misuses the `$text` parameter, which also accepts HTML and which would violate (among other things) the dependence of HTML as an output stream and hides the actual structure that one wants to implement.

### DIFF
--- a/src/UI/Component/Modal/Factory.php
+++ b/src/UI/Component/Modal/Factory.php
@@ -20,6 +20,7 @@ namespace ILIAS\UI\Component\Modal;
 
 use ILIAS\UI\Component;
 use ILIAS\UI\Component\Image\Image;
+use ILIAS\UI\Component\Item\Item;
 
 /**
  * Interface Factory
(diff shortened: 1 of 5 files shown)


Amstutz:
Hi @lscharmer 

We entirely agree that the context and use case described in the linked FR page are hard to implement properly through the existing state of the modals. We further agree that the roundtrip modal would not be the way to go in this case. In a short session on discord together with @nhaagen , @mjansenDatabay , @klees , @thibsy and @Amstutz we agreed that Round Trip should only be issued to get input from the user - not merely displaying information to him/her.

Your context is clearly different. The lightbox at this stage is somewhat limited. We also agree in this point. However, we do not believe that adding a new component Card Detail Page accepting an image and an item would improve the situation much eventually. Cards and Items are somewhat difficult components to begin with, marrying them in Lightbox modals semantically seems risky (also see the ongoing discussion here: https://github.com/ILIAS-eLearning/ILIAS/pull/4916/files ). 

However, we definitely believe that there would be great value, in having a Lightbox Modal that is ready to display some structured data along with maybe some actions to it would be of great use. However, best first contact the UI Clinic and be prepared to ask the following questions:

- What is the user intent, when opening the modal? 
- What experiences do the user bring from similar cases in ILIAS or other apps to match mental models they have already learned?

An example of how this can be done, can be found here: https://github.com/ILIAS-eLearning/ILIAS/pull/4916/files. You certainly do not need to go that far, but it could help to get a better understanding of the reason for the above question and the strategies to find answers to them. If done, then we might match an existing UI Component to match the user intent to pass to the modal or devise some new one.

Thx
@Amstutz and @klees

PR #4935, opened 2022-08-15
TA Roadmap: repo-strategy

nhaagen:
I was asked to expand on the idea of repo-pattern in T&A and how this might be a strategy as well. This is what I came up with so far; happily expecting comments ;)

(some of the mentioned thing I started out with in https://github.com/nhaagen/ILIAS/tree/_trunk/TA/repo , but that if far from ready/definite)

### DIFF
--- a/Modules/Test/ROADMAP.md
+++ b/Modules/Test/ROADMAP.md
@@ -31,3 +31,124 @@ Remarks on the individual items are marked with "@PHP8-CR"
 * \ilObjTestGUI::questionsObject / Undefined method. Left for review/analysis by TechSquad
 * \ilTestEvaluationGUI::passDetails / Undefined method. Left for review/analysis by TechSquad
 * \ilAssLacCompositeEvaluator::evaluateSubTree / Undefined method. Left for review/analysis by TechSquad
+
+
+## A [RepoPattern](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/repository-pattern.md) Approach to reduce complexity
+A lot of the current pain in the T&A is due to the ruthless mixing of Logic and GUI, making decisions very late based on 
+variables transported a long way and - in general - having properties set and evaluated in a quite obscured way by several components.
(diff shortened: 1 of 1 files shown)


dsstrassner:
Thanks @nhaagen. 
I would suggest, that we speak on this PR and the comments (at least I expect some ;) at the next TA TechSquad Meeting before merging it.

mbecker-databay:
I like this general writeup of the Repository Pattern, even if it's a bit "ranty". 
I think it would belong into a more central location in its full glory, other components may find it useful, too.

For T&A, my biggest concern is "standardizing" the implementation details, so that we have foundation and avoid looking at potentially n variants of the repository pattern in the code over time or - to avoid that - endless loops of review & makeovers.

Do you think there is a way to formulate such conventions for us?

nhaagen:
I'm sorry, if this is "ranty", I certainly meant no offense, it's more or less what I experienced when working on base settings and kept losing orientation. 
Concerning "standardizing the implementation details" - a lot of it is described in the repo-pattern-docs, and way better than I could do here. However, I do not think of a repo as "this is the blueprint, all others must act accordingly", but rather as the "single source of truth" and the _only_ place an object is built/stored in as to rule out side effects. Sometimes, collections may only exist in superior objects, but storage e.g. is across several tables. This would still be valid repo. Maybe a factory is needed, maybe stuff can be created directly, maybe I can use (local) DI; that quite depends on prerequisites or uniformity, and is specific to features, imo. Also, I expect quite a few iterations, but that's the whole point, right? Making it clearer, safer and more intuitive to rework the T&A code ;)

mbecker-databay:
My grin when I wrote that the text is a bit "ranty" was lost in typing it down. Do not get me wrong: It's a good read and building on the experiences made as a way to take the reader with you is a good approach. I meant this in a whole positive way.

I am not sure if my wish for a blueprint solution is realistic or going too far... that'd make a good point for the meeting in thursday.

PR #4936, opened 2022-08-15
BT 33814 and 33843

schmitz-ilias:
This PR  fixes [33814](https://mantis.ilias.de/view.php?id=33814) and [33843](https://mantis.ilias.de/view.php?id=33843). Would you be so kind to merge this in @smeyer-ilias 's absence into the trunk and R8, @alex40724?


alex40724:
@schmitz-ilias Done. Thanks!

PR #4937, opened 2022-08-16
TestQuestionPool: Remove unused JS library

mjansenDatabay:
I could not find any usage of these JS files.

If approved, this should be cherry-picked to `release_8`, `release_7` and maybe `release_6` as well.


mbecker-databay:
I take it, but only to 8 and trunk.

PR #4939, opened 2022-08-16
BT 33842

schmitz-ilias:
This PR fixes [33842](https://mantis.ilias.de/view.php?id=33842). Would you do the honors again @alex40724 ?


Services/Search/classes/Lucene/class.ilLuceneSearchObjectListGUIFactory.php
```diff
@@ -58,7 +53,7 @@ public static function factory(string $a_type) : ilObjectListGUI
         $full_class = "ilObj" . $class . "ListGUI";
 
         include_once($location . "/class." . $full_class . ".php");
-        $item_list_gui = new $full_class();
+        $item_list_gui = new $full_class(ilObjectListGUI::CONTEXT_SEARCH);
```
alex40724, line 56: I just did a quick check, if this parameter is correctly passed by the overwritten constructors. This is not always the case, I found: BookingPool, Forum, ItemGroup, LearningSequence, OrgUnit, RootFolder, SAHSLearningModule, Session, StudyProgramme, Survey, RemoteObjectBase, example:

```
class ilObjBookingPoolListGUI extends ilObjectListGUI
{
    protected \ILIAS\BookingManager\StandardGUIRequest $book_request;

    public function __construct()
    {
        global $DIC;

        $this->ctrl = $DIC->ctrl();
        $this->lng = $DIC->language();
        parent::__construct();
        $this->book_request = $DIC->bookingManager()
                                  ->internal()
                                  ->gui()
                                  ->standardRequest();
    }
...
}
```

Actually I am not sure, what CONTEXT_SEARCH controls exactly after quickly looking at the code.

But all these constructors might need to be fixed, or am I wrong? If yes, we might merge this one and create a separate PR.

alex40724 (changes requested):

Services/Search/classes/Lucene/class.ilLuceneSearchObjectListGUIFactory.php
schmitz-ilias: Setting the context to CONTEXT_SEARCH is needed here to make sure that uploading files to container objects that show up in the search results is disabled (the file dropzones are causing an error, and this functionality shouldn't be enabled in the search results anyways).

As far as I can see you are right, all overwritten constructors that don't already pass the context should be fixed to do so.

alex40724 (commented):
@smeyer-ilias Could you take care of this one, please.

alex40724 (commented):
.

alex40724 (approved):

smeyer-ilias:
Thanks to both of you. 
I will create a PR which adds the context-type to constructors

PR #4942, opened 2022-08-17
UI: render byline for Groups, e.g. when used as option in an OptionalGroup

nhaagen:
Hi,
when translating some form into UI Framework, I found the switchableGroup (the one with the radios...) not to render bylines; this is especially missing if a group w/o any further input is used to just express an option.
Please consider including bylines ;)
Best regards,
Nils


klees (approved):

Amstutz:
Thx a lot! Picked to trunk as well.

PR #4947, opened 2022-08-18
TOS: Refactor TosSettingsFromGUI to new KS UI Form

fhelfer:


mjansenDatabay (changes requested):
Hi @fhelfer ,

thx for this PR.

Please implement the following changes

- [ ] Instead of using hard coded names for the fields (e.g. `tos_status`) please use PHP constants like `private const F_TOS_STATUS = 'tos_status';`;
- [ ] Please add a check for the `POST` HTTP method in `saveSettings`.
- [ ] Depending on the state of the system you will get an error in `\ilObjTermsOfServiceGUI::saveSettings` if the `else` branch is reached and the key `tos_reevaluate_on_login` does not exist. Please make the code more robust for all combinations of checked or unchecked `checkbox` elements.
- [ ] Every state changing command (means: every successful form submission) should result in a HTTP redirect. Please implement this for the `elseif` branch.
- [ ] Please implement unit tests, if possible.

- [ ] Why do you always return `false` in line 187?

Please consider the following changes. You don't need to implement these (and not exactly as suggested), but please answer why you won't:

- [ ] Instead of calling the `Setter` methods in the `GUI` class, you could use a transformation to `bind` the form values to the object.

```php
$bindToModelTransformation = $this->refinery->custom()->transformation(function (array $values) : array {
    $this->object->bindFormInput($values); // Populates the object with the passed values
    return $values;
});

//...

$form = $form->withAdditionalTransformation($bindToModelTransformation);

// ...

$isFormValid = false;
if ('POST' === $this->dic->http()->request()->getMethod()) {
    $form = $form->withRequest($this->dic->http()->request());

    $formData = $form->getData();
    $isFormValid = !is_null($formData);
    if ($isFormValid) {
        $this->object->store();
    }
}

// ....

```
Best regards
@mjansenDatabay

fhelfer:
> Please implement the following changes
> 
> * [x]  Instead of using hard coded names for the fields (e.g. `tos_status`) please use PHP constants like `private const F_TOS_STATUS = 'tos_status';`;
> * [x]  Please add a check for the `POST` HTTP method in `saveSettings`.
> * [x]  Depending on the state of the system you will get an error in `\ilObjTermsOfServiceGUI::saveSettings` if the `else` branch is reached and the key `tos_reevaluate_on_login` does not exist. Please make the code more robust for all combinations of checked or unchecked `checkbox` elements.
> * [x]  Every state changing command (means: every successful form submission) should result in a HTTP redirect. Please implement this for the `elseif` branch.
> * [x]  Please implement unit tests, if possible.
> 
> Please answer the following questions:
> 
> * [x]  Why do you always return `false` in line 187?
> 
> Please consider the following changes. You don't need to implement these (and not exactly as suggested), but please answer why you won't:
> 
> * [x]  Instead of calling the `Setter` methods in the `GUI` class, you could use a transformation to `bind` the form values to the object.
> 
>

Services/TermsOfService/classes/class.ilObjTermsOfServiceGUI.php
```diff
@@ -189,17 +204,32 @@ protected function saveSettings(): void
...
-        }
+        $form = $form->withAdditionalTransformation($bindToModelTransformation);
+        $form = $form->withRequest($this->request);
+        if ('POST' === $this->request->getMethod()) {
+            $form = $form->withRequest($this->dic->http()->request());
 
-        $this->tpl->setContent($form->getHTML());
+            $formData = $form->getData();
+            if (!is_null($formData)) {
+                $hasDocuments = true;// ilTermsOfServiceDocument::where([])->count() > 0;
```
mjansenDatabay: This looks like forgotten debug code.

Services/TermsOfService/classes/class.ilObjTermsOfServiceGUI.php
```diff
@@ -222,6 +254,7 @@ protected function settings(): void
         $this->showMissingDocuments();
 
         $form = $this->getSettingsForm();
-        $this->tpl->setContent($form->getHTML());
+        $this->components[] = $form;
+        $this->tpl->setContent($this->dic->ui()->renderer()->render($this->components));
```
mjansenDatabay, line 271: We still need a new `MessageBox`  (type: info) component with the current state determined in `ilAdministrationSettingsFormHandler::FORM_TOS` and an additional link targeting to `\ilObjUserFolderGUI::generalSettingsObject`.

Please remove the `case` in `\ilObjUserFolderGUI::addToExternalSettingsForm` afterwads and delete the `ilAdministrationSettingsFormHandler::FORM_TOS` usage.

mjansenDatabay (changes requested):

Services/TermsOfService/classes/class.ilObjTermsOfServiceGUI.php
fhelfer: What message should be put inside the MessageBox? 
I am not quite sure what it is for

mjansenDatabay: TEST-7-Nutzungsvereinbarungen

mjansenDatabay:
@fhelfer We should also remove `ilAdministrationSettingsFormHandler::FORM_TOS`.

Services/TermsOfService/test/gui/ilObjTermsIfServiceGUITest.php
```diff
@@ -0,0 +1,200 @@
...
+ *********************************************************************/
+
+use ILIAS\UI\Factory;
+use ILIAS\UI\Renderer;
+use ILIAS\Refinery\Factory as Refinery;
+use PHPUnit\Framework\MockObject\MockObject;
+use ILIAS\HTTP\GlobalHttpState;
+use ILIAS\DI\Container;
+
+class ilObjTermsIfServiceGUITest extends ilTermsOfServiceBaseTest
```
mjansenDatabay:
```suggestion
class ilObjTermsOfServiceGUITest extends ilTermsOfServiceBaseTest
```

Services/TermsOfService/test/gui/ilObjTermsIfServiceGUITest.php
```diff
@@ -0,0 +1,200 @@
...
+        $this->tos = $this->getMockBuilder(ilObjTermsOfService::class)->disableOriginalConstructor()->getMock();
+        $this->tpl = $this->getMockBuilder(ilGlobalTemplateInterface::class)->getMock();
+        $this->ctrl = $this->getMockBuilder(ilCtrl::class)->disableOriginalConstructor()->getMock();
+        $this->lng = $this->getMockBuilder(ilLanguage::class)->disableOriginalConstructor()->getMock();
+        $this->rbacsystem = $this->getMockBuilder(ilRbacSystem::class)->disableOriginalConstructor()->getMock();
+        $this->error = $this->getMockBuilder(ilErrorHandling::class)->disableOriginalConstructor()->getMock();
+        $this->http = $this->getMockBuilder(GlobalHttpState::class)->disableOriginalConstructor()->getMock();
+        $this->refinery = $this->getMockBuilder(Refinery::class)->disableOriginalConstructor()->getMock();
+        $this->uiFactory = $this->getMockBuilder(Factory::class)->disableOriginalConstructor()->getMock();
+        $this->uiRenderer = $this->getMockBuilder(Renderer::class)->disableOriginalConstructor()->getMock();
```
mjansenDatabay: You only need to call `createMock` for interfaces like `GlobalHttpState::class`, `ilGlobalTemplateInterface::class` etc..

mjansenDatabay (changes requested):
Thx again @fhelfer.

Along with my inline comments I am questioning the new test class. Is it somehow possible to test the form behaviour without all these mocks?

If the answers is 'NO', I indeed suggest to remove this test. But maybe we could try to test this without the mocks, maybe be creating an anonymous class derived from `ilObjTermsOfServiceGUI`.

mjansenDatabay (approved):

PR #4948, opened 2022-08-18
Administration/Settings: Add missing database type switch for `settin…

mjansenDatabay:
…gs:value`

This PR adds an `Achievable Setup Objective` to let administrators switch the database field type of `settings:value` from `VARCHAR` to `CLOB` and vice versa.

Debatable: Should we stick with the suggested objective class or should we segregate this into two separate objectives?

---

Furthermore this PR adds a `Metrics Collected Objective` to report the current database field type when executing the `status` command of the ILIAS setup.

---

- If approved, this should be picked to `release_8` as well.

---

Mantis Issue: https://mantis.ilias.de/view.php?id=33860

### DIFF
--- a/Services/Administration/classes/Setup/class.ilSettingsMetricsCollectedObjective.php
+++ b/Services/Administration/classes/Setup/class.ilSettingsMetricsCollectedObjective.php
@@ -1,0 +1,51 @@
+<?php declare(strict_types=1);
+
+/**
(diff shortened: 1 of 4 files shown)


alex40724:
@mjansenDatabay Thanks for the PR. However we might not need it, if we get rid of the switch (see comments in mantis).

alex40724:
Fixed according to JF decision.

PR #4949, opened 2022-08-18
BT 33866

schmitz-ilias:
This PR fixes [33866](https://mantis.ilias.de/view.php?id=33866). Could you merge to trunk and R8 @alex40724 ?


smeyer-ilias:
Thanks @schmitz-ilias

PR #4950, opened 2022-08-18
BT 33876

schmitz-ilias:
This PR should fix [33876](https://mantis.ilias.de/view.php?id=33876).


pascalseeland:
As discussed in discord I will close this bug.

PR #4951, opened 2022-08-19
RoleTemplates: Bugfix 0033906 (ILIAS7): Updating internal templates no longer updates title 

fhelfer:
Editing the desription of internal role templates resulted in them updating their title and no longer being internal
So for il_crs_member for example the new title in database would be "Kursmitglied" and when creating a course in Ilias this role would no longer be created inside the course, which resulted in not being able to add course members

Mantis: https://mantis.ilias.de/view.php?id=33906


smeyer-ilias (approved):
Thanks for fixing this.

PR #4952, opened 2022-08-19
Fix for creating repository objects with long titles or descriptions

schmitz-ilias:
This PR fixes the creation of repository objects with titles and/or descriptions longer than object_data allows, see [33865](https://mantis.ilias.de/view.php?id=33865).

### DIFF
--- a/Services/Utilities/classes/class.ilStr.php
+++ b/Services/Utilities/classes/class.ilStr.php
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 
 /**
  * This file is part of ILIAS, a powerful learning management system
@@ -189,6 +189,9 @@ class ilStr
         bool $a_next_blank = false,
         bool $a_keep_extension = false
     ) : string {
+        if ($a_dots) {
+            $a_len--;
+        }
         if (ilStr::strLen($a_str) > $a_len) {
             if ($a_next_blank) {
                 $len = ilStr::strPos($a_str, " ", $a_len);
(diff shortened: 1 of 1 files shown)


klees:
Hi @schmitz-ilias and everyone,

we discussed this PR in the TB today. Currently we cannot see anyone assuming responsibility for `Service/Utilities` as maintainer or coordinator, only implicit maintainers. On the other hand, almost every ILIAS developer seems to dependent on this somehow. This is why we decided to take this to the JF to find someone to look into this PR.

In general we are looking to get rid of `Service/Utilities` completely. There is valuable code in there, but there also seems to be stuff that is not (much) used anymore. The code that is still in use should migrate to newer facilities, such as `src\Data`, the Refinery or the UI-Framework. We ask every developer to participate in this endeavour so we won't need to ask the question "who is responsible for `Services/Utilities` again some times soon.

Best regards!

alex40724 (changes requested):
Just my 2cents as an implicit maintainer. I think the length correction is ok. However I would not introduce strict type declaration during beta phase, since this might trigger additional errors in consumer code, which might have already been tested.

schmitz-ilias:
> Just my 2cents as an implicit maintainer. I think the length correction is ok. However I would not introduce strict type declaration during beta phase, since this might trigger additional errors in consumer code, which might have already been tested.

Done, thanks!

alex40724 (approved):
Thanks!

matthiaskunkel:
Jour Fixe, 19 SEP 2022 : We highly appreciate this PR and accept it for 8 and trunk.

PR #4955, opened 2022-08-19
System Styles Roadmap

Amstutz:
The System Styles, most importantly, the CSS contained in the System Styles are currently
lacking a coherent structure along with guidelines. They grew too large over time to be 
handled by one single developer and would therefore highly benefit from contributions among 
different developers and even service providers. They are of modular structure with different parts 
following a similar scheme. Further, they are critical importance for many other components. Therefore, we consider
them suited perfectly to be maintained by the Coordinator Model, see: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/maintenance-coordinator.md

This Roadmap is therefore a first step to meet the requirements listed above for such a maintenance.

### DIFF
--- a/templates/Roadmap.md
+++ b/templates/Roadmap.md
@@ -1,0 +1,15 @@
+# Roadmap
+The System Styles, most importantly, the CSS contained in the System Styles are currently
+lacking a coherent structure along with guidelines. They grew too large over time to be 
+handled by one single developer and would therefore highly benefit from contributions among 
+different developers and even service providers. They are of modular structure with different parts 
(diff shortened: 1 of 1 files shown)


matthiaskunkel:
Jour Fixe, 22 AUG 2022: Timon notified all JF participants about his activities to move the CSS maintainership into the coordinator model. A first step is a roadmap for the further development and maintenance of the system styles. An upcoming task will be the use of a frameworkless SASS for the UI framework (see roadmap).

PR #4957, opened 2022-08-22
Bugfix/kprim test add solution set check

mBeym:
May fix https://mantis.ilias.de/view.php?id=33635


mBeym:
Closed because according to reporter the issue no longer appears

PR #4958, opened 2022-08-22
Icon large in shared resources

Amstutz:
Icon too large, see https://mantis.ilias.de/view.php?id=33114

Please note:
- UI Components Icon has been used to fix the icon size issue
- The Alt text hast been set to "", to avoid wcag fail for duplicate naming of purely decorative icons. Note that the name will be given as well via visible text in the table.
- Outcommented code of another Icon usage has been removed in tpl and php code

thx for considering.


alex40724:
@Amstutz Thanks. Merged to trunk, too.

fsesterh:
This won't work with ILIAS 7 I guess? Personal and Shared resources with ILIAS 7 have the same trouble in case you don't use the default icons.

PR #4960, opened 2022-08-23
Trunk CI: adjust cs fixer to work with psr-12. See PR #4533.

daniwe4:


mjansenDatabay:
@daniwe4 This seems to be a (partly) duplicate of #4877

klees:
I kindly asked @daniwe4 to start fresh, this is the result =) I'll close the other one.

klees:
This does not include an automatic fix for the new rule regarding variable and property naming: "Names of properties and variables MUST be written in underscore-case."

I currently cannot find an according PHP-CS-Fixer rule to make this happen, so I'll take this 95% here...

klees:

PR #4962, opened 2022-08-24
[FIX] #33944: form-action of `ilTrObjectUsersPropsTableGUI`

thibsy:
Hi all

@mjansenDatabay asked me to look into https://mantis.ilias.de/view.php?id=33944 since it's an ilCtrl related issue which would be solved by this PR.

To briefly explain the behaviour:

`ilTrObjectUsersPropsTableGUI` tried to generate a link target to itself. It is a child-class of `ilLPListOfObjectsGUI` but not a direct child-class of `ilLearningProgressGUI`, which is the last command-class in the ilCtrl context at this point. Therefore `ilTrObjectUsersPropsTableGUI` could not have been appended to it because it isn't a direct child-class. As proposed by this PR, we could tell ilCtrl that there is a command-class in between by providing an array of classes where the first one is applicable to the current one. We could also make `ilTrObjectUsersPropsTableGUI` a direct child-class of `ilLearningProgressGUI ` by adding an according `@ilCtrl_isCalledBy`-statement to `ilTrObjectUsersPropsTableGUI` (if there are multiple cases where this issue might occur).

I'm not entirely sure, but it might be possible that ilCtrl in the past searched for indirect children to a certain depth. If that's the case I would consider this an ilCtrl issue and we can close this PR.

Kind regards!

### DIFF
--- a/Services/Tracking/classes/repository_statistics/class.ilTrObjectUsersPropsTableGUI.php
+++ b/Services/Tracking/classes/repository_statistics/class.ilTrObjectUsersPropsTableGUI.php
@@ -116,7 +116,7 @@ class ilTrObjectUsersPropsTableGUI extends ilLPTableBaseGUI
         $this->setExternalSegmentation(true);
         $this->setEnableHeader(true);
         $this->setFormAction(
-            $this->ctrl->getFormActionByClass(get_class($this))
+            $this->ctrl->getFormActionByClass([ilLPListOfObjectsGUI::class, get_class($this)])
         );
         $this->setRowTemplate(
             "tpl.object_users_props_row.html",


smeyer-ilias:
This is a similar issue as described here https://github.com/ILIAS-eLearning/ILIAS/pull/4772

The specific problem here is:
1) Different modules create LP tab links with cmdClass=ilLearningProgressGUI
2) ilLearningPrgressGUI forwards to other LP classes depending on LP settings. E.g. ilLPListOfProgressGUI using $ilCtrl->forwardCommand() 

In the past ilCtrl->forwardCommand() added the forwarded class to the ctrl-path.
https://github.com/ILIAS-eLearning/ILIAS/blob/c5c00f69ffd52853927f35f22d4f8b314ab20124/Services/UICore/classes/class.ilCtrl.php#L220
 Thus the ctrl path was: 
e.g ilRepositoryGUI -> ilObjGroupGUI -> ilLeanringPrgressGUI -> ilLPListOfProgressGUI

Now the ctrl path is:  ilRepositoryGUI -> ilObjGroupGUI -> ilLeanringPrgressGUI which produces the problems as described in the mantis ticket. 

There are many similar issues in the calendar and search service, ...

This could be fixed by manually calling $ilCtrl->setCmdClass() in combination with $ilCtrl->forwardCommand(), but set cmdClass is marked as depercated. eg. in ilLearningProgressGUI()

ilCtrl->setCmdClass(ilLPListOfProgressGUI.class)
ilCtrl->forwardCommand(new ilLPListOfProgressGUI())

thibsy:
@smeyer-ilias is this still an issue? because I couldn't reproduce the scenario from https://mantis.ilias.de/view.php?id=33944 anymore.

smeyer-ilias:
@thibsy I fixed it temporarily by using the deprecated setCmdClass().
So, it is still an issue.

mjansenDatabay:
@thibsy I agree with @smeyer-ilias , it is still an issue without his workaround.

thibsy:
Hi @smeyer-ilias 

I close this PR as well for the same reason as in #4772.

Kind regards!

PR #4963, opened 2022-08-25
Setup: Remove legacy database update scripts

mjansenDatabay:
This PR removes the 'legacy update steps' and the 'hotfixes' from the ILIAS code base. This includes all `dbupdate(_\d+)?\.php` scripts.

'Custom database update steps' are still kept. This mechanism could be still abandoned for ILIAS 10 (I guess), which will make `ilDBUpdate` obsolete.

The hierarchy of `ilDBUpdate` and `ilPluginDBUpdate` is still ugly. Maybe someone with a bit more sleep than me wants to tie up with this refactoring.


klees:
Thanks a lot @mjansenDatabay!

PR #4964, opened 2022-08-25
Fix: " Identifier 'ilDB' is not defined" when storing GlobalCache settings

utesche:
When performing a setup-install or setup-update (with nominated config.json):
*Store configuration of Services/GlobalCache...                             [FAILED]*
*[ERROR] Identifier "ilDB" is not defined.*
is raised  (in this case, the 'config.json' defines a memchache daemon to be used) .

I'm not skilled enough to judge whether this PR has other negative effects, but for the use by setup "update" and "install" it works apparently.
So take my suggestion as a hint to a problem at least.

wbr

### DIFF
--- a/Services/ActiveRecord/Connector/class.arConnectorDB.php
+++ b/Services/ActiveRecord/Connector/class.arConnectorDB.php
@@ -23,9 +23,7 @@ class arConnectorDB extends arConnector
 {
     protected function returnDB(): ilDBInterface
     {
-        global $DIC;
-
-        return $DIC['ilDB'];
+        return $GLOBALS['ilDB'];
     }
 
     public function checkConnection(ActiveRecord $ar): bool


klees:
Hi @utesche, @chfsx,

this seems to be something in ActiveRecord, so I have assigned this to Fabian.

Best regards!

chfsx:
Thanks @utesche ! For the moment, your suggestion is one way we can work around the problem. However, the unit tests fail with it. I will take your base and extend it and then merge it locally. 
For the future, I need to find a way to make these implicit dependencies explicit in the ActiveRecord.

chfsx:
@utesche could you send me please (the cache part of) your config.json to test against it? Thanks a lot!

chfsx:
ok I managed to test it with the demo-config from the readme. Thanks @utesche, I committed in https://github.com/ILIAS-eLearning/ILIAS/commit/0183cb3e10b7c49e57955c7fbf8c460f66f8eb39

utesche:
> @utesche could you send me please (the cache part of) your config.json to test against it? Thanks a lot!

Hello @chfsx,
I could - but it won't work for you anyway:  the addressed server isn't visible to the www.
(... and you manged it by yourself ...)

wbr,
Uwe

PR #4965, opened 2022-08-25
Refinery/HTTP: Fix PHP types for PHP 8.1

mjansenDatabay:
This PR fixes PHP 8.1 issues like ....

```bash
PHP Deprecated:  Return type of ILIAS\Refinery\KeyValueAccess::offsetGet($offset) should either be compatible with ArrayAccess::offsetGet(mixed $offset): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /home/runner/work/ILIAS/ILIAS/src/Refinery/KeyValueAccess.php on line 49
```
... for the `KeyValueAccess` (and derivate) type of the `Refinery` by using the types defined in the PHP core. This is one of the exceptions where we should accept `mixed` as a type.

### DIFF
--- a/src/HTTP/Wrapper/SuperGlobalDropInReplacement.php
+++ b/src/HTTP/Wrapper/SuperGlobalDropInReplacement.php
@@ -2,6 +2,22 @@
 
 declare(strict_types=1);
 
+/**
+ * This file is part of ILIAS, a powerful learning management system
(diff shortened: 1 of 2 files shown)


chfsx (approved):
Thanks a lot, LGTM!

klees (approved):

PR #4967, opened 2022-08-26
[FIX] declare(strict_types=1); is now on the first line again.

thibsy:
Hi @mjansenDatabay

I noticed that when generating PHP files with PHPStorm and the new coding style, the `declare(strict_types=1);` will be on the second line of the file due to the `NEW_LINE_AFTER_PHP_OPENING_TAG`. Up until now, the code-styles did not have this option enabled and the strict-types declaration was put on the first line with the PHP opening tag. If this was entirely intended though you can close this PR and I'll accept that the statement must be on the second line as of now :).

Kind regards!


mjansenDatabay:
Thx @thibsy

PR #4972, opened 2022-08-30
fix #34075: Failed test: Video-Player per Tabbing verlassen

alex40724:
Fixes https://mantis.ilias.de/view.php?id=34075


Amstutz:
Thx a lot, ill pick to trunk as well.

PR #4975, opened 2022-08-30
Data type LanguageTag with parsing

lscharmer:
This PR introduces a new data type LanguageTag.
The language tag is validated by the RFC as it is defined here: https://www.ietf.org/rfc/bcp/bcp47.txt.
This is the continuation of PR #4653.

This will be used in: https://github.com/ILIAS-eLearning/ILIAS/pull/4082

### DIFF
--- a/src/Data/Factory.php
+++ b/src/Data/Factory.php
@@ -2,7 +2,21 @@
 
 declare(strict_types=1);
 
-/* Copyright (c) 2017 Richard Klees <richard.klees@concepts-and-training.de> Extended GPL, see docs/LICENSE */
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
 
 namespace ILIAS\Data;
 
@@ -201,5 +215,10 @@ class Factory
         }
 
         return $this->open_graph_metadata_factory;
+    }
+
+    public function languageTag(string $language_tag): LanguageTag
+    {
+        return LanguageTag::fromString($language_tag);
     }
 }
--- a/src/Data/LanguageTag.php
+++ b/src/Data/LanguageTag.php
@@ -1,0 +1,170 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+namespace ILIAS\Data;
+
+use ILIAS\Data\LanguageTag\Standard;
+use ILIAS\Data\LanguageTag\Regular;
+use ILIAS\Data\LanguageTag\Irregular;
+use ILIAS\Data\LanguageTag\PrivateUse;
+use InvalidArgumentException;
+use ILIAS\Refinery\Custom\Transformation as Custom;
+use ILIAS\Refinery\Parser\ABNF\Brick;
+use ILIAS\Data\Result\Error;
+use Closure;
+
+/**
+ * This class represents a valid language tag that should be used instead of plain strings.
+ * The language tag is validated by the provided language tag definition.
+ *
+ * RFC 5646 compliant language tag definition (https://www.ietf.org/rfc/bcp/bcp47.txt).
+ */
+abstract class LanguageTag
+{
+    abstract public function value(): string;
+
+    public function __toString(): string
+    {
+        return $this->value();
+    }
+
+    public static function fromString(string $string): static
+    {
+        $brick = new Brick();
+        return $brick->apply(self::definition($brick), $string)->except(static fn (): Result => (
+            new Error('Given string is no valid language tag.')
+        ))->value();
+    }
+
+    /**
+     * This definition is directly translated from the ABNF definition on https://www.ietf.org/rfc/bcp/bcp47.txt.
+     */
+    private static function definition(Brick $brick): Closure
+    {
+        $extlang = $brick->sequence([
+            $brick->repeat(3, 3, $brick->alpha()),
+            $brick->repeat(0, 2, $brick->sequence(['-', $brick->repeat(3, 3, $brick->alpha())])),
+        ]);
+
+        $language = $brick->either([
+            $brick->sequence([
+                'lang' => $brick->repeat(2, 3, $brick->alpha()),
+                $brick->repeat(0, 1, $brick->sequence(['-', 'extlang' => $extlang])),
+            ]),
+            $brick->sequence(['lang' => $brick->repeat(4, 4, $brick->alpha())]),
+            $brick->sequence(['lang' => $brick->repeat(5, 8, $brick->alpha())]),
+        ]);
+
+        $script = $brick->repeat(4, 4, $brick->alpha());
+
+        $region = $brick->either([
+            $brick->repeat(2, 2, $brick->alpha()),
+            $brick->repeat(3, 3, $brick->digit()),
+        ]);
+
+        $alphanum = $brick->either([$brick->alpha(), $brick->digit()]);
+
+        $variant = $brick->either([
+            $brick->repeat(5, 8, $alphanum),
+            $brick->sequence([$brick->digit(), $brick->repeat(3, 3, $alphanum)])
+        ]);
+
+        $singleton = $brick->either([
+            $brick->digit(),
+            $brick->range(0x41, 0x57),
+            $brick->range(0x59, 0x5A),
+            $brick->range(0x61, 0x77),
+            $brick->range(0x79, 0x7A),
+        ]);
+
+        $extension = $brick->sequence([
+            $singleton,
+            $brick->repeat(1, null, $brick->sequence(['-', $brick->repeat(2, 8, $alphanum)])),
+        ]);
+
+        $privateuse = $brick->sequence([
+            'x',
+            $brick->repeat(1, null, $brick->sequence(['-', $brick->repeat(1, 8, $alphanum)]))
+        ]);
+
+        $new = static fn (string $class) => new Custom(static fn (string $arg) => new $class($arg));
+
+        $privateuse = $brick->transformation($new(PrivateUse::class), $privateuse);
+
+        $irregular = $brick->either([
+            'en-GB-oed',
+            'i-ami',
+            'i-bnn',
+            'i-default',
+            'i-enochian',
+            'i-hak',
+            'i-klingon',
+            'i-lux',
+            'i-mingo',
+            'i-navajo',
+            'i-pwn',
+            'i-tao',
+            'i-tay',
+            'i-tsu',
+            'sgn-BE-FR',
+            'sgn-BE-NL',
+            'sgn-CH-DE',
+        ]);
+
+        $regular = $brick->either([
+            'art-lojban',
+            'cel-gaulish',
+            'no-bok',
+            'no-nyn',
+            'zh-guoyu',
+            'zh-hakka',
+            'zh-min',
+            'zh-min-nan',
+            'zh-xiang',
+        ]);
+
+        $grandfathered = $brick->either([
+            $brick->transformation($new(Irregular::class), $irregular),
+            $brick->transformation($new(Regular::class), $regular)
+        ]);
+
+        $langtag = $brick->transformation(
+            new Custom(static fn (array $from): Standard => new Standard(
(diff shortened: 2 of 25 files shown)


lscharmer:
On the 27.09.2022 we (@nhaagen, @kergomard and I) discussed where the parser should be placed and how it should be accessed. We decided that the Parser itself should be a private API (there will be no way to access e.g. the `Brick` class from the `$DIC`). Publicly there will be a transformation for the refinery (e.g. `$DIC->refinery()->parseTo()->languageTag()`) and a creation method for the Data Factory (e.g. `$DIC->data()->languageTag('de')`) added for each data type that uses the parser.

The code for the refinery will be located under `src/Refinery/Parser`. And under `src/Refinery/Parser/Definitions` new Definitions for different data types can be added.

All Documentation for the parser will be moved from doc blocks into README files.

@klees @mjansenDatabay If you approve of this approach I will implement these changes.

Best regards
@lscharmer

mjansenDatabay:
@lscharmer / @klees 

For me the suggested changes/workshop results look/sound good. I like the new `parseTo` group of the `Refinery` and the suggested folder structure. From the consumer perspective it is good to hide the `Brick as we learned the `Parser` topic is hard and not many developers will (unfortunately) use it.

mjansenDatabay (approved):

Amstutz:
What is the status here? Note that there seems to be an other PR with a dep. to this one: https://github.com/ILIAS-eLearning/ILIAS/pull/4082

mjansenDatabay:
@Amstutz The status is: There was another workshop after the presentation of @lscharmer at the ILIAS development conference. In this workshop, @lscharmer , @nhaagen and @kergomard spoke about how to finally integrate this into the `Refinery` and `Data` components. 

As the coordinator of the `Refinery` I am fine with the workshop results and decisions.

The question is: Does @klees have any objections, especially for the `src/Data` part?

klees:
@mjansenDatabay I'm on it...

mjansenDatabay:
Any updates on this :)? A feedback is appreciated, so @lscharmer can continue to complete this until the end of the year.

klees (changes requested):
Hi @lscharmer,

thanks a lot for working on this. I had kind of a hard time to look through this and it took me longer than expected. Sorry for that... The hard times come for three reasons: I'm currently very tired, which has absolutely nothing to do with you. Building parsers is inherently complex. And you use some rather unusual constructs. We won't change the first, I'm afraid, but we can improve on the second and third item, which is what this review is mostly about.

Furthermore, some items are about improving the consumer interface of the new functionality.

- [x] `to` and `either`: If I'd use `Transform::to` on the result build by `either`: How would I find out if the left or right option was taken?


- [x] Homegrown Inheritance: I really had a hard time figure out the [construction around `Data\LanguageTag`](https://github.com/ILIAS-eLearning/ILIAS/pull/4975/files#diff-295829b75693494e324a987d92f150e215f671ab329e9a5eb498d8904bd9948cR29) (see change request about documention below). I might be wrong, but to me this looks as if the wrapper [`LanguageTag`](https://github.com/ILIAS-eLearning/ILIAS/pull/4975/files#diff-295829b75693494e324a987d92f150e215f671ab329e9a5eb498d8904bd9948cR29) is some kind of homegrown inheritance. As far as I understand, a language tag can be one these: an ordinary language tag according to the RFC, a private lang tag (not really cared about by the RFC) or an old, "grandfathered", language tag. This currently is implemented via two separate classes that carry the values (`LangTag` and `GrandFathered`) and a wrapper `LanguageTag` that may hold one of theses classes (each) and a string representing a private lang tag. To me, this is looked odd. Why is a `LanguageTag` potentially any of these sub classes? You check if it is none of them, but it may be each of them. Same thing goes for `Grandfathered`: It can be regular and irregular at the same time. It took me some time to figure it out and find the connection to the RFC. So, to me it looks as if this really is a case for an abstract base class `LanguageTag`, probably with a simple `__toString` (= `value`) as method. Then inherit this in the other classes. Probably rename `LangTag` to `LanguageTag\Standard` or something similar (RFC might have proper wording...). We can then hint on `LanguageTag` to mean any language tag. Or on `LanguageTag\Standard` to mean an ordinary lang tag. Introduce `LanguageTag\Private` to indicate a private-use-only lang tag. Either split `GrandFathered` into two classes `RegularGrandFathered` and `IrregularGrandFathered` or use a simple string value and a boolean (ir)regular-flag to get rid of the possibility to have a `GrandFathered` tag, that is irregular and regular at the same time.
- [x] Tests: Please consider adding some more tests for Brick. I'm not 100% sure all public methods are covered. I'm mostly sure, that for some of the building blocks tests for failing attempts to parse something are missing.
- [x] Transform: I would suggest to move the methods from `ABNF\Transform` to `Brick`. It seems as if these are just other building blocks for the parser and we will mostly require `Brick` and `Transform` besides each other anyway.
- [x] Short Closure: Please consider using short-form closures instead of the `function (..) use (..)` syntax if possible. There seem be mostly value building closures anyway, in deeply nested stacks sometimes, using most of the symbols from the outer scope. Hence `return` and `use` and `function` are mostly clutter. The code is hard to understand anyway and would benefit from this readability improvement a lot imo.
- [x] `as`: I think the creation of the parser would look a lot simpler if we'd baked that array building via `as` into `sequence`. Currently, the naming of array keys and the actual usage are far apart. E.g.: To understand, which array keys are available [here](https://github.com/ILIAS-eLearning/ILIAS/pull/4975/files#diff-f54c00a124a0aaa49047ac8df6aac234efae9e423b9db2b96116d37abbeb9eabR145), I would need to scroll up [there](https://github.com/ILIAS-eLearning/ILIAS/pull/4975/files#diff-f54c00a124a0aaa49047ac8df6aac234efae9e423b9db2b96116d37abbeb9eabR66). If I could write `$brick->sequence(["language" => $language, ...]) this would be a lot easier to understand. Please consider incorporating `as` into `sequence`.
- [x] `toTransformation`: Add a method `toTransformation` to `Brick` that basically does what `applyTo` does, but instead returns a `Transformation` from `string` to `Thing` according to the parser. This would allow to integrate the parser to the framework of the Refinery.
- [x] LangTagTransformation/LanguageTagTransformation/GrandFatheredTransformation: Consider removing these classes and create the transformations on the fly via `$refinery->custom()->transformation()` on the usage side. As far as I understand, these two classes are just required to build the final parser and do not need a name or factory since they won't be used elsewhere.
- [x] LanguageDefinition: I would suggest to move the construction of the parser to a static function on the base clase `LanguageTag` (which I hope you are going to create). AFAIK, the definition of the ABNF for the language tag is intimately tied to that class and the RFC. I would mostly reject static methods, but I think this might actually be a good use case for them. Especially because nobody will ever interact with that method directly but instead use the `Refinery` or `Data`.
- [x] Factory: I would suggest to expose the creation of a `LanguageTag` on the factory of `Data` similar to the one for `URI`. Consumers could then access them in the context of the `Refinery` as  `$refinery->to()->data("languageTag")`.

- [x] Unused `use`: Remove [these](https://github.com/ILIAS-eLearning/ILIAS/pull/4975/files#diff-3fa5e2f42a78eeff969635ade0d307b10485be928b81153d0382de5479fcb6a2R28) and [those](https://github.com/ILIAS-eLearning/ILIAS/pull/4975/files#diff-e49628890d75ceed00c7819a9e660393b722c6799c56c9b7675cf4dfcb3a818aR25).
- [x] Data/README: Please correct [the README.md](https://github.com/ILIAS-eLearning/ILIAS/pull/4975/files#diff-19ff14f8652abd51b667e794b4258fa8cc58e7845c95af91fbc8ac15f0a41f8aR569) in `src/Data`. The showcased way to create a lang tag won't work (but probably should). The [docu about the parser](https://github.com/ILIAS-eLearning/ILIAS/pull/4975/files#diff-19ff14f8652abd51b667e794b4258fa8cc58e7845c95af91fbc8ac15f0a41f8aR576) should probably go to a `README.md` in `src\Refinery\Parser\ABNF`.
- [x] Documentation: Please add some small documentation, probably refering to the [RFC](https://www.rfc-editor.org/rfc/rfc5646) to [`GrandFathered`](https://github.com/ILIAS-eLearning/ILIAS/pull/4975/files#diff-cac8cce5a68bbd55a4d01037b8c520b91fc51b3b64b3692d4fa2180d7e5445c3R23).
- [x] Transform::to: I think that building block is doing more than announced. It seems to apply the transformation and then wrap the result in an array.
- [x] Internal API: As far as I understand `Brick` is meant to be an internal tool for building transformations. Which I totally agree on. In `src\README.md` we require for code in the `src`-folder: "All libraries MUST expose a clear public interface to their consumers. Every class, interface or trait on the top level of the library is considered to be a part of the public interface, as well as all entities that are reachable via entities on the top level." So according to this, the parser stuff would currently be public. I think it would be ok to simply remove all the factories for the internal parser. If someone needs a `Brick`, it could simply be created (it has no other external dependencies...) or it could be required via dependency injection.

If anything is hard to understand, don't hesitate to contact me. We might even solve the most of this in a pair programming session, if you prefer to do so.

Best regards!

lscharmer:
Hi @klees,
I highly appreciate your feedback.
Your suggestions and change requests are great improvements that I will gladly implement.
The following 2 points are the only ones that need imo further discussion:

> - [ ] Factory: I would suggest to expose the creation of a LanguageTag on the factory of Data similar to the one for URI. Consumers could then access them in the context of the Refinery as $refinery->to()->data("languageTag").

I agree that the `LanguageTag` should be accessible from the `Data` Factory but I cannot make the `LanguageTag` available to the `Data` Factory (with constructor DI) because there is a chicken and egg problem:  The Refinery needs the Data Factory which needs the ABNF definition for the LanguageTag which in turn needs the Refinery.
The only way I can accomplish this is by accessing the global `DIC` or implement something like `ILIAS\Data\Factory::setRefinery(...)`, which are both not optimal.

> - [ ] Transform: I would suggest to move the methods from ABNF\Transform to Brick. It seems as if these are just other building blocks for the parser and we will mostly require Brick and Transform besides each other anyway.

Both `Brick` and `Transform` are solely there to form a structure. The classes are there to make it easier what the purpose of these methods is. Imo the purpose of these methods is different enough to have them split into multiple classes. I also think it is nice to call `$transform->to(...)` instead of `$brick->to(...)` or `$brick->transformTo(...)` which reflects this.

I would rather argue that even `Brick` could be split up into more classes (I placed appropriate comments as way to structure them _softly_: [here](https://github.com/ILIAS-eLearning/ILIAS/pull/4975/files#diff-c71420282ee476a9e2beb4fc48635b839e63eb9a0609126b7f4256ef8c9b69f0R83) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/4975/files#diff-c71420282ee476a9e2beb4fc48635b839e63eb9a0609126b7f4256ef8c9b69f0R125) which reminds me of those comments which should've been method names).

I would instead propose to use `Transform` in `Brick` to have something like: `$brick->transform()->to(...)`.

Best regards.

lscharmer:
Hi @klees,
I saw that there is one question that I need to answer:

> - [ ] to and either: If I'd use Transform::to on the result build by either: How would I find out if the left or right option was taken?

With `Transform::as(...)` you can give them a key which you can use to check for existence:
```php

$left = $transform->as('left', $brick->sequence(['foo']));
$right = $transform->as('right', $brick->sequence(['bar']));

$transformation = $refinery->custom()->transformation(static function (array $parts) {
    if (isset($parts['left'])) {
        // Do something with left.
   } elseif (isset($parts['right'])) {
        // Do something with right.
    }
});

$transform->to($transformation, $brick->either([$left, $right]));
```

Or, because either is basically an if else you can do something like:
```php

// No need to give them a key:
$left = $brick->sequence(['foo']);
$right = $brick->sequence(['bar']);

$leftTransformation = $transform->to($refinery->custom()->transformation(static function (string $left) {
    // Do something with left.
}), $left);

$rightTransformation = $transform->to($refinery->custom()->transformation(static function (string $right) {
    // Do something with right.
}), $right);

$transformation = $refinery->custom()->transformation(static function (array $parts) {
    // No need to check for left or right here.
});

$transform->to($transformation, $brick->either([$left, $right]));
```

Best regards.

lscharmer:
Hi @klees,

I implemented the requested changes. 

With the change of:

> -  [ ] as: I think the creation of the parser would look a lot simpler if we'd baked that array building via as into sequence. Currently, the naming of array keys and the actual usage are far apart. E.g.: To understand, which array keys are available [here](https://github.com/ILIAS-eLearning/ILIAS/pull/4975/files#diff-f54c00a124a0aaa49047ac8df6aac234efae9e423b9db2b96116d37abbeb9eabR145), I would need to scroll up [there](https://github.com/ILIAS-eLearning/ILIAS/pull/4975/files#diff-f54c00a124a0aaa49047ac8df6aac234efae9e423b9db2b96116d37abbeb9eabR66). If I could write $brick->sequence(["language" => $language, ...]) this would be a lot easier to understand. Please consider incorporating asintosequence`.

and the discussion of `$brick->transform()->to(...)` I decided to add a new class `Primitives` where the _old_ implementation lives and the `Brick::either` and `Brick::sequence` use `Primitives` and `Transform` to achieve the feature described above. Because `Transform` would with this change have only one public method for the API, I changed the brick so it serves as the sole interface for the API which uses internally `Primitives` and `Transform`. This keeps the public API clean and keeps the structure of the code.

I didn't implement the following change:

> - [ ] Transform::to: I think that building block is doing more than announced. It seems to apply the transformation and then wrap the result in an array.

Because the `Transform::to` method doesn't do more than announced, It just uses the same mechanism as `Transform::as` to transform its value. The method MUST wrap it's value into an array because `Intermediate::push` accepts an array.

Best regards.

klees:
Really awesome work, have merged: https://github.com/ILIAS-eLearning/ILIAS/commit/af60dbb2c2e2811b21611d1c7c67e620789dba9e

Thanks a lot!

mjansenDatabay:
Thx as well @klees for your ideas and review efforts.

klees:
Definitely worth it =)

PR #4976, opened 2022-08-30
T&A, #34015: getPreviewFileUploads shall return array

nhaagen:
https://mantis.ilias.de/view.php?id=34015


mbecker-databay:
Sorry, I missed seeing this one. From the description alone I made the same fix yesterday.

PR #4981, opened 2022-08-31
Fix/0034118/8/remove file german

chfsx:
Hi @Amstutz 
We had a problem with the dynamic inputs in file-dropzones with other languages that english (sorry, we just didn't thought about it enough when using the CSS-selectors): https://mantis.ilias.de/view.php?id=34118

with this PR, we
- fix the issue: https://github.com/ILIAS-eLearning/ILIAS/pull/4981/commits/54100bfdee4459a81837125c28a88f00ddc08df9
- removed dropzone.js from bower and moved to NPM: https://github.com/ILIAS-eLearning/ILIAS/pull/4981/commits/30d6c37ee4dde3fec6a9325154c441be98a34c0b


chfsx:
Tests are failing due to an issue with composer: 
```
Plugin simplesamlphp/composer-module-installer could not be initialized, class not found: SimpleSAML\Composer\ModuleInstallerPlugin
```
not related to this PR, I think.

thibsy (approved):
I only looked at the fix which LGTM, unit tests pass on my machine so it's probably the composer error like you said.

I've found another issue though, which is not caused or related to these changes, so I'll open another PR for it.

Amstutz:
Hi @chfsx 

Thx a lot. Uff, how could that have slipped our attention while reviewing ;-). Sry for that. 

I dislike the freshly added span layer with additional classes. Additional layers in the DOM tend to lead to issues down the road. Be it in terms of a11y or styling, or future refactorings.

Therefore, two questions / consider to change:
- [ ] CSS: Do you really need the classes attached for styling the glyphs? Could you not add the styles (such as margins) to the glyphicons themselves (and not the link/span around them)? Note that on the Glyphicons, you do have the respective classes to apply necessary styles for the upload.
- [ ] JS: In JS, you need the span and the respective classes to attach JS logic. I dislike to use CSS classes for such things. It is certainly not unprecedented in the UI Component, but again, those cases will and already have led to issues in the past. Note that Glyphs are JS bindable. So, you can attach an ID while rendering and pass this ID with onload code to the JS logic. This is much cleaner, less error-prone, and somewhat more efficient.

Thx a lot
Timon

thibsy:
@Amstutz I think we cannot use the bindable logic in the glyph scenario here, because they will be cloned by the dynamic-inputs-renderer on clientside, which would lead to duplicate IDs. I believe we can't properly work around this without introducing SSR to the dynamic-inputs-renderer (which IMO would be kind of an overkill), or do I miss an obvious approach?

Amstutz:
Hi @thibsy 

>  I believe we can't properly work around this without introducing SSR to the dynamic-inputs-renderer (which IMO would be kind of an overkill), or do I miss an obvious approach?

I am not so deep in the JS here, but right, if you clone the glyphs, then relying on the ID will be impossible. Duplicate IDs will be an automatic fail in any check of the DOM (along with the none-working JS). Maybe attach some unique string to the cloned ID or work with a data attribute? @klees any thoughts on this? I do not see an obvious solution either, but maybe we find a slightly better one.

thx

Amstutz:
Hi 

I had a short chat with @chfsx over discord for this. We agreed that he will look into possibilities to style the elements without the additional span and classes around it. Further he will look into using the existing classes on the glyphs for the js selector or a span with some data attribute. I further explained, why it is hard to change the ID of the cloned elements since cloning is done by the library.

Thx to everybody involved

chfsx:
@Amstutz I first tried to use the glyph-css-selectors such as `glyphicon-remove` but this doesn't feel right and a lot of JS code would depend on `.parent()`, and last but not least, CSS didn't work out well. I chase a `data-action` now (could be any other thing than `-action` but seems mostly right here, maybe a `data-contains-action` would be more precise here.

chfsx:
This will fix https://mantis.ilias.de/view.php?id=33899 as well

chfsx:
And with https://github.com/ILIAS-eLearning/ILIAS/pull/4981/commits/9523f08e4afbb896e2aa933e073efcca06b00221 the following  issue will be fixed as well: https://mantis.ilias.de/view.php?id=31645 (and with it, we can fix https://mantis.ilias.de/view.php?id=34353)

Amstutz:
Hi @thibsy , hi @chfsx 

Thx a lot for looking into this and going the extra mile by comparing the different approaches for binding JS to the DOM Elements. We believe the solution to be much better than before.

We will pick to trunk as well.

PR #4982, opened 2022-08-31
[FIX] file.js can remove files manualy again.

thibsy:
Hi @Amstutz

Here's the other issue I've mentioned in #4981. The problem was when uploading two (or more) files and then removing one using the close-glyph, the form submission aborted because the corresponding dropzone instance couldn't be found. It was merely a mixe-up of the query-selectors though and hence a minor fix (but caused me a headache anyways =)).

Kind regards!


thibsy:
sorry @Amstutz, @chfsx just pushed this commit to the other PR, therefore I closed this one.

PR #4983, opened 2022-09-01
CSS: Fix Big in LPG of Learning Modules, see #34163 and #34163

Amstutz:
Proposed fix for https://mantis.ilias.de/view.php?id=34163 and https://mantis.ilias.de/view.php?id=34161

Note that the alt text has been left empty, since we probably might consider the icons here as "decorative" since an title in plain text is given as well.

### DIFF
--- a/Services/Tracking/classes/repository_statistics/class.ilLPCollectionSettingsTableGUI.php
+++ b/Services/Tracking/classes/repository_statistics/class.ilLPCollectionSettingsTableGUI.php
@@ -14,6 +14,9 @@ class ilLPCollectionSettingsTableGUI extends ilTable2GUI
 
     protected ilObjectDefinition $obj_definition;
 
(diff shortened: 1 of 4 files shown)


Amstutz:
Note, the failing checks are only the copyright checks.

Amstutz:
@alex40724 why do you ignore my PR and "fix" this by using some ilIcon class, eventhough I asked you multiple times not to this and use the Standard Icon instead. Please revert your change above and consider this PR. thx.

alex40724:
@Amstutz  Personally I will not do refactorings by bug fixing. The migration to KS elements ends with beginning of beta to me, except in rare cases I will not introduce KS elements after this point.

The changes I did where predominantly to keep test cases from being usable, since Stefan was on vacation.

But Stefan is maintaining Services/Tracking anyway, so he will decide in this case.

PR #4995, opened 2022-09-05
TA: Repo for ScoreSettings, UI Form

nhaagen:
This adjusts the ResultSettings form of TestObject according to a suggestion from the ILIAS-NRW group.
I still need some clarification concerning the exact mapping of some former options to new fields, but this should be a rather non-tech issue.
Also, UI-forms are used as well as a repo-adaption to maybe devide the big number of settings into smaller units.
I'll still have to look into export/import and the template-funtionality.

Comments are most welcome already, though.

### DIFF
--- a/Modules/Test/classes/ScoreReporting/ScoreSettingsRepository.php
+++ b/Modules/Test/classes/ScoreReporting/ScoreSettingsRepository.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+interface ScoreSettingsRepository
+{
+    public function getFor(int $test_id): ilObjTestScoreSettings;
+    public function store(ilObjTestScoreSettings $settings): void;
+}
--- a/Modules/Test/classes/ScoreReporting/class.ilObjTestScoreSettings.php
+++ b/Modules/Test/classes/ScoreReporting/class.ilObjTestScoreSettings.php
@@ -1,0 +1,132 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+use ILIAS\UI\Component\Input\Field\Factory as FieldFactory;
+use ILIAS\UI\Component\Input\Field\Field;
+use ILIAS\Refinery\Factory as Refinery;
+
+use ILIAS\UI\Component\Input\Container\Form\Form;
+
+class ilObjTestScoreSettings
+{
+    protected int $test_id;
+    protected ilObjTestSettingsScoring $settings_scoring;
+    protected ilObjTestSettingsResultSummary $settings_result_summary;
+    protected ilObjTestSettingsResultDetails $settings_result_details;
+    protected ilObjTestSettingsGamification $settings_gamification;
+
+    public function __construct(
+        int $test_id,
+        ilObjTestSettingsScoring $settings_scoring,
+        ilObjTestSettingsResultSummary $settings_result_summary,
+        ilObjTestSettingsResultDetails $settings_result_details,
+        ilObjTestSettingsGamification $settings_gamification
+    ) {
+        $this->test_id = $test_id;
+
+        foreach ([
+            $settings_scoring,
+            $settings_result_summary,
+            $settings_result_details,
+            $settings_gamification
+        ] as $setting) {
+            $this->throwOnDifferentTestId($setting);
+        }
+
+        $settings_result_summary = $settings_result_summary
+            ->withShowPassDetails($settings_result_details->getShowPassDetails());
+
+        $this->settings_scoring = $settings_scoring;
+        $this->settings_result_summary = $settings_result_summary;
+        $this->settings_result_details = $settings_result_details;
+        $this->settings_gamification = $settings_gamification;
+    }
+
+    protected function throwOnDifferentTestId(TestSettings $setting)
+    {
+        if ($setting->getTestId() !== $this->getTestId()) {
+            throw new \LogicException('TestId mismatch in ' . get_class($setting));
+        }
+    }
+
+    public function getTestId(): int
+    {
+        return $this->test_id;
+    }
+    public function withTestId(int $test_id): self
+    {
+        $clone = clone $this;
+        $clone->test_id = $test_id;
+        $clone->settings_scoring = $clone->settings_scoring->withTestId($test_id);
+        $clone->settings_result_summary = $clone->settings_result_summary->withTestId($test_id);
+        $clone->settings_result_details = $clone->settings_result_details->withTestId($test_id);
+        $clone->settings_gamification = $clone->settings_gamification->withTestId($test_id);
+        return $clone;
+    }
+
+
+    public function getScoringSettings(): ilObjTestSettingsScoring
+    {
+        return $this->settings_scoring;
+    }
+    public function withScoringSettings(ilObjTestSettingsScoring $settings): self
+    {
+        $this->throwOnDifferentTestId($settings);
+        $clone = clone $this;
+        $clone->settings_scoring = $settings;
+        return $clone;
+    }
+
+    public function getResultSummarySettings(): ilObjTestSettingsResultSummary
+    {
+        return $this->settings_result_summary;
+    }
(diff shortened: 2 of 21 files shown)


dsstrassner:
https://mantis.ilias.de/view.php?id=33680

Modules/Test/classes/ScoreReporting/ScoreSettingsRepository.php
```diff
@@ -0,0 +1,23 @@
...
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+interface ScoreSettingsRepository
+{
+    public function select(int $test_id) : ilObjTestScoreSettings;
```
klees: Let's call this `getFor` (to avoid that idea, that this is about SQL).

Modules/Test/classes/ScoreReporting/ScoreSettingsRepository.php
```diff
@@ -0,0 +1,23 @@
...
+    public function store(ilObjTestScoreSettings $settings) : void;
+}
```
klees, line 25: I wonder if this is missing a `deleteFor`.

Modules/Test/classes/ScoreReporting/class.ilObjTestScoreSettings.php
```diff
@@ -0,0 +1,761 @@
...
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+use ILIAS\UI\Component\Input\Field\Factory as FieldFactory;
+use ILIAS\Refinery\Factory as Refinery;
+
+class ilObjTestScoreSettings
```
klees, line 27: This class very much feels like it really should be three classes, one for each concern. The prefixes tell...

We could than bundle these three classes into this one. Please either do this now or write it down in the ROADMAP and link to that entry in a comment above the class.

Modules/Test/classes/ScoreReporting/class.ilObjTestScoreSettingsDatabaseRepository.php
```diff
@@ -0,0 +1,155 @@
...
+            ->withPrintBestSolutionWithResult((bool) $row['print_bs_with_res'])
+            ->withResultsPresentation($row['results_presentation'])
+            ->withShowExamIdInTestResults((bool) $row['examid_in_test_res'])
+            
+            ->withExportSettings((int) $row['exportsettings'])
+
+//->withShowGradingDetailsEnabled()
+//->withShowSolutionSuggested
+//->withExportSettingsSingleChoiceShort
+//->withHidePagecontents
```
klees: Please throw away uncommented code...

Modules/Test/classes/class.ilObjTestGUI.php
```diff
@@ -3395,26 +3412,29 @@ public function applyTemplate($templateData, ilObjTest $object)
...
-            // result details properties
-            'solution_details' => 'setShowSolutionDetails',
-            'solution_feedback' => 'setShowSolutionFeedback',
-            'solution_suggested' => 'setShowSolutionSuggested',
-            'solution_printview' => 'setShowSolutionPrintview',
-            'highscore_enabled' => 'setHighscoreEnabled',
-            'solution_signature' => 'setShowSolutionSignature',
-            'examid_in_test_res' => 'setShowExamIdInTestResultsEnabled',
-            'exp_sc_short' => 'setExportSettingsSingleChoiceShort',
+    //TODO: score-settings...
```
klees: Undo indentation, this is a none-change (right?).

Modules/Test/classes/class.ilObjTestSettingsScoringResultsGUI.php
```diff
@@ -127,43 +179,60 @@ private function confirmedSaveFormCmd()
...
+        $settings = $result->value();
+        $this->storeScoreSettings($settings);
+        $this->tpl->setOnScreenMessage('success', $this->lng->txt("msg_obj_modified"), true);
+        $this->ctrl->redirect($this, self::CMD_SHOW_FORM);
+        return;
 
-        if (!$isConfirmedSave && $this->isScoreRecalculationRequired($form)) {
-            $this->showConfirmation($form);
-        }
+        /*
```
klees: Can we remove this uncommented code?

Modules/Test/classes/class.ilObjTestSettingsScoringResultsGUI.php
```diff
@@ -127,43 +179,60 @@ private function confirmedSaveFormCmd()
...
-
-        $errors = !$form->checkInput(); // ALWAYS CALL BEFORE setValuesByPost()
-        $form->setValuesByPost(); // NEVER CALL THIS BEFORE checkInput()
+        $form = $this->buildForm()
+            ->withRequest($this->request);
 
-        // return to form when any form validation errors exist
-
-        if ($errors) {
+        $result = $form->getInputGroup()->getContent();
```
klees: Please just use `getData` on the `$form` here. It will either return `null` when input is faulty or the actual result if that is not the case. `getInputGroup` is part of the internals only.

Modules/Test/classes/class.ilObjTestSettingsScoringResultsGUI.php
```diff
@@ -271,158 +357,173 @@ private function buildForm(): ilPropertyFormGUI
...
-        $fields = array(
-            'count_system', 'score_cutting', 'pass_scoring', 'pass_deletion_allowed'
-        );
 
-        if ($this->isSectionHeaderRequired($fields)) {
-            // scoring settings
-            $header = new ilFormSectionHeaderGUI();
-            $header->setTitle($this->lng->txt('test_scoring'));
-            $form->addItem($header);
+    /*
```
klees: Can we remove this?

Modules/Test/classes/class.ilObjTestSettingsScoringResultsGUI.php
```diff
@@ -271,158 +357,173 @@ private function buildForm(): ilPropertyFormGUI
...
-    /**
-     * @param ilPropertyFormGUI $form
-     */
-    private function saveScoringSettingsFormSection(ilPropertyFormGUI $form)
-    {
-        if ($this->areScoringSettingsWritable()) {
-            if ($this->formPropertyExists($form, 'count_system')) {
-                $this->testOBJ->setCountSystem($form->getItemByPostVar('count_system')->getValue());
-            }
+        private function saveScoringSettingsFormSection(ilPropertyFormGUI $form)
```
klees: Please remove =)

Modules/Test/classes/class.ilObjTestSettingsScoringResultsGUI.php
```diff
@@ -465,258 +566,310 @@ private function saveResultSummarySettings(ilPropertyFormGUI $form)
...
-        $highscore->addSubItem($highscore_wtime);
-
-        // show signature placeholder
-        $showSignaturePlaceholder = new ilCheckboxInputGUI($this->lng->txt('tst_show_solution_signature'), 'solution_signature');
-        $showSignaturePlaceholder->setInfo($this->lng->txt('tst_show_solution_signature_desc'));
-        $showSignaturePlaceholder->setChecked($this->testOBJ->getShowSolutionSignature());
-        if ($this->testOBJ->getAnonymity()) {
-            $showSignaturePlaceholder->setDisabled(true);
+    /*
+    <<<<<<< HEAD
```
klees: Remove this?

Modules/Test/classes/ScoreReporting/class.ilObjTestScoreSettings.php
```diff
@@ -0,0 +1,851 @@
...
+
+    public function getForm(
+        bool $anonymity_flag,
+        bool $disabled_flag,
+        string $action,
+        \ilLanguage $lng,
+        FormFactory $form_factory,
+        FieldFactory $f,
+        Refinery $refinery
+    ): Form {
```
klees: Please do not return a `Form` here, as this will break composability. Return the complete section, including the transformation that makes it output an `ilObjTestScoreSettings`.

The general idea is: All the objects that are filled via forms should only return `Fields`. These can be composed. The form should then be wrapped around the fields that we are interested in the presentation layer to handle the actual in and output. E.g.: We might be thinking about presenting this in some wizard some day. We will then be able to completely reuse this (if it returns a `Field`) and make the decision which form to use at very late point in time.

klees (changes requested):
Hi @nhaagen,

I'm really exited to see this. Looks awesome =)

I have some remarks and change requests, though, hope its ok I added them inline. If there's anything to discuss: You know how to contact me =)

Best regards!

Modules/Test/classes/ScoreReporting/ScoreSettingsRepository.php
nhaagen: In general, I'd agree. Here, we're working on a joint table, i.e. the settings-instance is build with a subset of properties; I figure, deletion/creation should be done somewhere else.

nhaagen:
From TA TechSquad today(22.09.22): the recalculation of results is still necessary, and this PRs lacks the possibility; it will have to be amended.

dsstrassner:
@nhaagen will finish the work on this PR after his return.

dsstrassner:
Update: There are still some issues which should be addressed ;)

dsstrassner:
When the issue of showing the icons in the "Review of Scored Answers" Subtab is gone, this PR is good to go from my perspective. If the PM decides to change the label from "Review of Scored Answers" to another one, this could be taken care of in the aftermath.

@klees has requested changes. I can't say, if where are finished on this or if there is anything left, at least a comment or so.

dsstrassner:
From my perspective we are good to go, to merge this with ILIAS 8 and Trunk. 

@mbecker-databay or @kergomard would someone of you two be so nice, to merge this accordingly?

nhaagen:
Thanks, @kergomard, for checking.
For now, I removed this merge: https://github.com/nhaagen/ILIAS/pull/43
I'll add it later as PR to be discussed seperatedly.

kergomard:
Thank you very much @nhaagen for all the work and for hugely improving these settings. As it was decided to port this back to ILIAS 8 we merge it at this late stage.

PR #4996, opened 2022-09-05
[FIX] phpstorm code-style config

thibsy:
Hi @mjansenDatabay

I've noticed another small issue with the new code-style config for PHPStorm. Up until now the declaration of return-types was like `function foo() : int` instead of `function bar(): int`, which would be fixed by this PR.

Since we would want to keep things consistent I also assumed this would fit in case of enums, therefore I adjusted this setting as well from `enum a: int` to `enum b : int`. If that's not the case though, you can just adopt the `SPACE_BEFORE_COLON_IN_RETURN_TYPE` option.

Kind regards!


mjansenDatabay:
Hi @thibsy ,

IMO this PR introduces a space we tried to get rid of with #4533

PSR-12 says:

> When you have a return type declaration present, there MUST be one space after the colon followed by the type declaration. The colon and declaration MUST be on the same line as the argument list closing parenthesis with no spaces between the two characters.

Best regards,
@mjansenDatabay

thibsy:
Oh I see, that slipped my attention, will close the PR again then.

PR #4998, opened 2022-09-07
Fix Language Detection For WebDAV

kergomard:
WebDAV is an HTTP-Protocol, thus it has usesHTTP set to true. Not all clients deem it necessary though to send a ACCEPTS_LANGUAGE Header (Windows Explorer doesn't), so that we need this additional check.


kergomard:
@katringross : It would be very kind if you could have a look at this soon. This blocks any further testing in WebDAV. Thank you very much!

katringross (approved):

katringross:
@kergomard sorry for taking so long! I overlooked this PR somehow :-(

PR #4999, opened 2022-09-07
Fix Issue When Saving New LongMenu Question

kergomard:
This Fixes an Issue when creating a new longmenu question. As original_id is null it will be set to an empty string and thus fail.


mbecker-databay:
Whoops. In a spree, I fixed this one today with 3dc90f03a7c6216e99853ee23608ddca37d560d6 ... in a different way.
Still, thank you so much for providing this one!

PR #5000, opened 2022-09-08
Prevent Invalidating Random Test

kergomard:
I don't see this directly as a bug and maybe it has been discussed already, but at least for us this caused troubles:
See: https://mantis.ilias.de/view.php?id=34334

This PR also adds a space where one is missing in the UI. I can move this elsewhere, if it bothers.


dsstrassner:
FYI @nhaagen ;-)

Thanks @kergomard! This is a bug. Editing should not be possible after participant data is present.

kergomard:
I merge and cherry-pick this to avoid further issues cropping up see: https://mantis.ilias.de/view.php?id=35076

PR #5001, opened 2022-09-09
TA OrderingQuestion: 33524, show form after error

nhaagen:
PR #4922, for v7.
https://mantis.ilias.de/view.php?id=33524


Amstutz:
@mbecker-databay , could this be merged, so we can re-test this. This would enable us to finally close: https://mantis.ilias.de/view.php?id=33524

PR #5005, opened 2022-09-12
Bugfix/0030392 ias active menu entry additional features

sasfalk:
#bugfix
Ziel: release_7
Mantis issue: 0030392
https://mantis.ilias.de/view.php?id=30392


klees:
Thanks a bunch! I have picked this to trunk and release_8.

PR #5006, opened 2022-09-12
Fix Removal of < in KPrim-Question

kergomard:
See: https://mantis.ilias.de/view.php?id=25428
So, this is not beautiful, but I don't find a better way to achieve this (if anyone sees one, I'm happy and will change it):
- I think the most correct way forward with the current infrastructure would be to change ilUtil::prepareFormOutput() to disallow double_encode, but I'm a little skeptical of side effects and ilUtil is on its way out anyways.
- The correct way on the middle run, is to completely replace the current way to build these forms in the test, but this is clearly out of scope here.
As always: If this is ok with you, I will provide the PR against 8.

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assKprimChoiceGUI.php
+++ b/Modules/TestQuestionPool/classes/class.assKprimChoiceGUI.php
@@ -310,7 +310,7 @@ class assKprimChoiceGUI extends assQuestionGUI implements ilGuiQuestionScoringAd
         }
         
         foreach ($answers as $key => $answer) {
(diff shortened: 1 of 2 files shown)


dsstrassner:
I think, we could talk about this PR at the TechSquad Meeting on 22/09/2022. On this Thursday, we talk about KS for questions.

mbecker-databay:
This is OK :) Thank you so much!

dsstrassner:
> I think, we could talk about this PR at the TechSquad Meeting on 22/09/2022. On this Thursday, we talk about KS for questions.

We talked about this PR at the Meeting on 15/09/22.

PR #5007, opened 2022-09-14
Bugfix 31495 FileInfo Error with no file

nim0elopez:
#bugfix
Target: release_7
Mantis issue: 31495
https://mantis.ilias.de/view.php?id=31495

The error was triggered every time the assessment form for a user in an individual assessment was loaded without a file having been added to it. The parent class calls the abstract function getInfoResult (which should return file information) and the current implementation in the ilIndividualAssessmentMemberGUI class could not handle an empty identifier.

This fix leans on the implementation as in the ilMMUploadHandlerGUI class which inherits from the same parent class.

An alternative solution would be to avoid calling the function in the first place by modifying the parent class AbstractCtrlAwareUploadHandler:100. I chose this safer approach to avoid unforeseen effects in other parts of the code, but I am of course open for discussion on that point :).

### DIFF
--- a/Modules/IndividualAssessment/classes/class.ilIndividualAssessmentMemberGUI.php
+++ b/Modules/IndividualAssessment/classes/class.ilIndividualAssessmentMemberGUI.php
@@ -212,7 +212,9 @@ class ilIndividualAssessmentMemberGUI extends AbstractCtrlAwareUploadHandler
 
         if ($grading->getFile() == '') {
             $storage = $this->getUserFileStorage();
-            $storage->deleteCurrentFile();
+            if (!$storage->isEmpty()) {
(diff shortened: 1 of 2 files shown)


klees (changes requested):
Hi @elopezmo ,

as discussed in our mentoring session: This does not seem to fix the root cause of the issue. Instead we would need to take care that we do not set a value on the form field for the file if we do not actually have a file. This would need to be done in `ilIndividualAssessmentUserGrading::toFormInput`.

Thanks for looking into this!

nim0elopez:
Hi @klees,
I modified the pull request as discussed :).

klees (approved):

klees:
Picked changes to release_8 and trunk.

PR #5008, opened 2022-09-14
Test: Fix Pax Overview in Test with Selected Pax

kergomard:
See: https://mantis.ilias.de/view.php?id=34412

As always: PR to ILIAS 8 as soon you verified I didn't overlook something.

### DIFF
--- a/Modules/Test/classes/class.ilTestParticipantList.php
+++ b/Modules/Test/classes/class.ilTestParticipantList.php
@@ -50,7 +50,9 @@ class ilTestParticipantList implements Iterator
     
     public function addParticipant(ilTestParticipant $participant) : void
     {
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
@maxbecker / @kergomard I am not sure if this introduces another issue if the test is used in the public area, where every user has the same user id, but different active_ids.

Maybe we should completely revert the latest changes in this iterator based class.

kergomard:
I close this PR after discussion with @mjansenDatabay , @dsstrassner , and @mbecker-databay . This will most probably not work for tests in public areas with anonymous participants. We will revert the commit causing this issue, reopen the corresponding performance-issue, and put `ilTestParticipantList` on our Roadmap for refactoring.

PR #5009, opened 2022-09-14
Bugfix/0030763 iass store title and description in metadata

sasfalk:
#bugfix
Mantis issue: 0030763
https://mantis.ilias.de/view.php?id=30763

Title and description were not stored in metadata. The missing method calls were added in the ilObjIndividualAssessment class.
The ilObjSurvey class served as a template for this fix. In this class the create method is called with „a_upload = false“ as a parameter. That didn't open up to me for the individual assessment class, so I left it out. But if the parameter is needed somewhere, I can add it to the function.


klees (approved):
Looks plausible. Thx!

klees:

PR #5011, opened 2022-09-14
Bugfix 31520 fix file upload error (ilias6)

nim0elopez:
#bugfix
Target: release_6
Mantis issue: 31520
https://mantis.ilias.de/view.php?id=31520

The changes in this PR are also a part of the changes included in the following commit (for ilias7):
https://github.com/ILIAS-eLearning/ILIAS/commit/e2a45f5a7eabb67856ff650ead4483c477f8fa98

it may be worth considering cherrypicking that commit for release_6 instead of using this PR - if the remaining changes in said commit all make sense for this version as well. But since the changes in that commit are plenty, it might be safer to merge this PR as the included changes specifically fix the issue in question.


klees (approved):

klees:
I have not move this up to newer releases because the file storage has changed in the meantime.

PR #5012, opened 2022-09-14
Notifications: Add JS toast API. Fix Bugs 34363, 34224

lscharmer:
Fixed Bug: 34224: Strict comparison of `string` against `int` in: `ilNotificationOSDRepository::ifOSDNotificationExistsById(...)`
Fixed Bug 34363: The Notifications are rendered with the initial page load.

A toast can now be created on the client side only with:
`OSDNotifier.toast(title, {icon, description, action})`


mjansenDatabay (approved):
Thx @lscharmer

PR #5013, opened 2022-09-14
TA: 34360, cast textsize to float on storing

nhaagen:
https://mantis.ilias.de/view.php?id=34360


kergomard:
Thanks at @nhaagen. I added this manually. Most casts seem to be separated by a space from the casted value, so I decided to stick to that, although I don't believe this to be regulated one way or the other. See: https://github.com/ILIAS-eLearning/ILIAS/commit/18fa6f8b16d49b831224d23906d7bcf537efa2b1

PR #5018, opened 2022-09-15
Bugfix/31867 upload file with spaces

nim0elopez:
#bugfix
target: release_7
mantis issue: 31867 and 32220 (these two seem to be duplicates)
https://mantis.ilias.de/view.php?id=31867
https://mantis.ilias.de/view.php?id=32220

During upload the file name is cleaned up, removing special characters and spaces. The wrong (unchanged) filename was being returned and therefore the wrong filename was being requested in subsequent calls.

Changes in `class.ilIndividualAssessmentMemberGUI.php` fix the issue.
Changes in `class.ilIndividualAssessmentFileStorage.php` are not all strictly necessary. The new function `getFileName` holds code that was repeated in the class before, and it makes the fix easy to read.
Before modifying the corresponding interface `interface.IndividualAssessmentFileStorage` I made sure it is not implemented anywhere else.


klees:
Thanks a bunch!

klees:

PR #5019, opened 2022-09-15
Orgu: 34375, check on create-permission for import tab

nhaagen:
https://mantis.ilias.de/view.php?id=34375


klees:
Picked to trunk.

PR #5024, opened 2022-09-15
Revert "Test: Remove iterable approach to find participants, use a ma…

mjansenDatabay:
…p instead"

Please re-open https://mantis.ilias.de/view.php?id=33596 and close https://mantis.ilias.de/view.php?id=34412 . And: Merge this to `trunk` as well.

### DIFF
--- a/Modules/Test/classes/class.ilTestParticipantList.php
+++ b/Modules/Test/classes/class.ilTestParticipantList.php
@@ -26,53 +26,70 @@
  */
 class ilTestParticipantList implements Iterator
 {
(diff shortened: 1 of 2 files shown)


kergomard (approved):
Dear @mjansenDatabay 
Thank you  very much!
This is more of a mental note for the future:

* As far as I can see this PR reverts the changes perfectly well, it also reverts some improvements (short array syntax, some typing, left over requires, strict comparisons). As we put this on the Roadmap I would suggest to keep it so and not invest the time right now. We will take care of this when refactoring.

Thanks again and best,
@kergomard

PR #5025, opened 2022-09-15
OrgU: 34397, fix type editing

nhaagen:
https://mantis.ilias.de/view.php?id=34397


klees:
Picked to trunk.

klees:
Picked to trunk.

PR #5026, opened 2022-09-15
Orgu (Tree): fix cancelMove and delete-Whoops

nhaagen:
ran into whoopses while moving/deleting OrgUs.

https://mantis.ilias.de/view.php?id=34495


klees:
Picked to trunk.

PR #5027, opened 2022-09-15
New UI-component "Launcher" proposal for JF decision

catkrahl:
The newly proposed top-level component "Launcher" should be used for launching a  object or a process. It can display general and status information, can contain inputs and consists of a primary button to execute an action.

It may substitute e.g. the info-tab when starting a test or a survey. 

Following the basic concept of UI-components in general, it can be used nested in other components or wrapped inside a modal depending on the intended purpose.

As a starting point we propose two derivations of the Launcher: an "Inline Launcher" and a  "Modal Launcher". But when it comes to other areas of application, more derivations are conceivable ...

Sceenshot(s) will follow and we would like to discuss this proposal at the next Jour Fix.

Hier ein erstes Mockup zum "Modal Launcher" am Beispiel "Test":

UI-Launcher-mockup

### DIFF
--- a/src/UI/Component/Launcher/Factory.php
+++ b/src/UI/Component/Launcher/Factory.php
@@ -31,21 +31,15 @@ interface Factory
      *     The Inline Launcher is meant to be used nested within Components (e.g. Cards or Panels).
      *     It provides description and status information at a glance.
      *   effect: >
-     *     Clicking the Primary Button starts an object or a process.
-     *     If the Component is configured with inputs, this will be communicated
-     *     to the user with an appropriate icon.
      *     The Form will be shown in an Interruptive Modal when clicking the Primary Button.
      *     On Submitting the modal, either a message is being displayed (in case of error)
      *     or the process/object is launched.
      *   rivals:
      *     Modal Launcher: >
-     *       The Modal Launcher will inform the user about a process after the basic
-     *       identification of the respective object was made or will further specify 
-     *       the initially chosen action.
-     * rules:
-     *   usage:
-     *     1: The Launcher SHOULD be nested in other Components like e.g. Panels or Cards.
-     *
+     *       Modal Launcher is visually only present via a Button and provides the required
+     *       information in a modal when the Button is clicked. Hence it is useful for
+     *       workflows where multiple launchable objects or workflows are presented and
+     *       the user choses among them.
      * ---
      * @return \ILIAS\UI\Component\Launcher\Inline
      */
@@ -55,25 +49,24 @@ interface Factory
      * ---
      * description:
      *   purpose: >
-     *     For further specification of a process or to avoid another view change,
-     *     the Modal Launcher can be used to show its contents wrapped within a Modal.
+     *     The Modal Launcher is presented via a Button, where the other information
+     *     are presented in a Modal that opens one the button is clicked. Hence it can
+     *     be used in cases, where multiple workflows and objects are presented and the
+     *     the user is expected to chose among them. The Modal Launcher allows to do
+     *     so without changing views.
      *   effect: >
-     *     Upon Signal, the Modal Launcher will open an Interruptive Modal and
-     *     displays a title, a desription, status information and a Primary Button.
-     *     If Inputs are configured with the Launcher, the resulting Form replaces
-     *     the contents of the Modal when the Primary Button is clicked.
-     *     Submitting the Form starts the object or process.
+     *     Upon click on the button, the Modal Launcher will open an Interruptive Modal
+     *     and display title, desription and status information and, if configured, the
+     *     inputs.
      *   rivals:
      *     Inline Launcher: >
      *       The Inline Launcher also identifies an object/process. The Modal Launcher
-     *       is a reaction to an interaction with another Component (e.g. item)
+     *       is only presented via a Button and hence requires other means to be identified
+     *       by the user correctly, such as Items or Cards.
      *     Interruptive Modal: >
      *       The Interruptive Modal is meant to display objects affected by a critical action. 
      *       The Modal Launcher presents relevant information to identify an object/process to launch
      *       or communicates that the intended launching action is permitted.
-     * context:
-     *   - List of launchable items
-     *
      * ---
      * @return \ILIAS\UI\Component\Launcher\Modal
      */
--- a/src/UI/Component/Launcher/Factory.php
+++ b/src/UI/Component/Launcher/Factory.php
@@ -1,0 +1,81 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+namespace ILIAS\UI\Component\Launcher;
+
+use ILIAS\Data\Link;
+
+interface Factory
+{
+    /**
+     * ---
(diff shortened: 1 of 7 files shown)


Amstutz:
Hi @catkrahl

Thx a lot for this important contribution. Note that we can only review this thoroughly, once the complete interface is available. However, since this is under some rather tight timeline, we will give some early feedback nevertheless.

I really do like the name "Launcher", good choice, kudos.

Maybe Change or gives some reason why not:
- [ ] Taxonomy: I am uncertain if "Launcher" really deserves a unique position at the very top of our Taxonomy. To me, it might be the first of several main content areas serving a very specific user intent (here, "Launch something"). It does not seem to belong to Layout, since Layout components carry certain structural decisions, they are also about the "where" of elements as opposed to the exclusive "what" in many other components. Launcher is clearly about "What". Maybe add a family "Views" would work (I was browsing through: https://github.com/ILIAS-eLearning/ILIAS/pull/4916/files, to get some inspiration). "View" seems to work, since it could be used to map some specific user intent to some specific View. E.g. "Launch Something" → "Launcher View". 
- [ ] A11y Rule: One of the reasons, such a "Launcher" would also be of high benefit in terms of a11y, is that we can arrange the items in a fashion, in which screen readers can properly work with. For the current launcher like info screen, we got the feedback that we need to place information necessary to read before launching the object **above** the launching button. Therefore, Maybe add something like: "All information required before launching the object MUST be placed before the Primary Button Launching the Object, so users working with screen readers will not miss it.

Please change:

- [ ] Typo: The Launcher is a component which contains **general** → general.
- [ ] Rules Wording: Instead of "it" use "The Launcher"
- [ ] Rules YAML Structure: Check out the YAML structure for the rules in other components and adapt:
```
     * rules:
     *   usage:
     *     1: ...
```

Thx a lot
@Amstutz

catkrahl:
Hi @Amstutz, 

thx a lot for your instant and helpful feedback! @nhaagen and me will construct the interfaces today and hopefully we'll finish them during the day. Regarding to your taxonomie-comments i think  your are right. During my first attempts to sort it under an existing component categorie, i had the same thoughts and problems. Even my early naming ideas had the namesuffix "view" ...

@A11y Rule: you are right, i will integrate your suggestion!

Thx a lot and best wishes
@catkrahl

alex40724:
Hi @catkrahl,

thanks for the proposal. This might be used for starting e.g. surveys, if I understand the concept correctly.

How would the consumer process the input data, e.g. the checkbox input in the example above?

Thx

klees:
Hi @alex40724,

yes, exactly, this could be used for the survey.

You could consume the data by constructing a form with the very fields that you have passed to the Launcher and provide them with the data in POST. The data would be POSTed against the `$target` asynchronously and then redirect to the target afterwards on HTTP-status 200. Does this sound plausible? If so, we will elaborate =)

Thanks for the feedback!

src/UI/Component/Launcher/Factory.php
```diff
@@ -0,0 +1,74 @@
...
+     * @return \ILIAS\UI\Component\Launcher\Inline
+     */
+    public function inline(Link $target): Inline;
+
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *     The Modal Launcher is presented via a Button, where the other information
+     *     are presented in a Modal that opens one the button is clicked. Hence it can
```
nhaagen, line 53:
```suggestion
     *     are presented in a Modal that opens once the button is clicked. Hence it can
```

src/UI/Component/Launcher/Factory.php
```diff
@@ -0,0 +1,74 @@
...
+     *     used in cases, where multiple workflows and objects are presented and the
+     *     the user is expected to chose among them. The Modal Launcher allows to do
+     *     so without changing views.
+     *   effect: >
+     *     Upon click on the button, the Modal Launcher will open an Interruptive Modal
+     *     and display title, desription and status information and, if configured, the
```
nhaagen, line 59:
```suggestion
     *     and display title, description and status information and, if configured, the
```

src/UI/Factory.php
```diff
@@ -896,4 +896,59 @@ public function symbol(): Component\Symbol\Factory;
...
+     *     inputs (e.g.: for access code input) and, finally, a Primary Button that
+     *     executes the launch action.
+     *   effect: >
+     *     Clicking the Primary Button starts the object, process or workflow. If the
+     *     Component is configured with inputs, they will be provided to make the user
+     *     fill them before the object can be launched. If the provided data is sufficient
+     *     the user is redirected to the target. Otherwise, a message is being displayed.
+     *     If the user cannot launch the object at all (precondition, unavailability etc.),
+     *     the primary button is disabled with unavailable action.
+     *     The label of the Primary Button may change, e.g. in relation to the statusu
```
nhaagen, line 924:
```suggestion
     *     The label of the Primary Button may change, e.g. in relation to the status
```

nhaagen (approved):

src/UI/Factory.php
```diff
@@ -900,44 +900,53 @@ public function toast(): C\Toast\Factory;
...
+     *     about the user's progress.
+     *   composition: >
+     *     The Launcher consists of receptive elements, i.e. text, status information
+     *     about progress, such as Icons or Progress Meter. It also contains optional
+     *     inputs (e.g.: for access code input) and, finally, a Primary Button that
+     *     executes the launch action.
+     *   effect: >
+     *     Clicking the Primary Button starts the object, process or workflow. If the
+     *     Component is configured with inputs, they will be provided to make the user
+     *     fill them before the object can be launched. If the provided data is sufficient
```
catkrahl, line 920: fill them in before the object can be launched.

matthiaskunkel:
Jour Fixe, 17 OCT 2022 : We highly appreciate the suggested Inline Launcher and accept it for trunk. Alexandra Tödt has some concerns about the Modal Launcher that should be written down here before a final decision about implementing also this modal version could be made. The button "Launcher starten" is not needed (at least for the Inline Launcher).

alex40724:
Thanks again for the proposal. I also see great value in the inline variant.

But I still struggle with the use cases for the modal variant, too. This might still be a misunderstanding on my side.

During the JF the mentioned main use case was the learning sequence, where users would like to start the object most of the time. Thus, the modal launcher should be featured.

I can't see any difference to the survey or test scenario (where the inline version on a start page has been proposed). To me all three object types should use the same process. At least for learners I assume that they want to start a test or a survey in most of the cases, too. Similar to the learning sequence. For other roles (tutors) this might not be the case, since they might want to configure the objects. But again this is similar for all three object types.

We are working very hard on consistency when it comes to processes, since this greatly supports the learnability of the system.

klees:
Hi @alex40724,

I still owe you an answer here. I'm sorry it took me so long, this always fell down behind.

As you see in the [follow up PR](https://github.com/ILIAS-eLearning/ILIAS/pull/5558), the inline variant carries some visual weight. IMO it works well, if there are one or two of these Launchers on a page, since the Launcher carries some visual weight. In the JF I tried to bring up a case, where there would be multiple launchable objects or processes on one page. My example was a list of "all uncompleted learning sequences". With the Modal Launcher these could be presented in a way that gives more overview, e.g. in a table, and the "Launch" would only be represented by a Button. The info would then be shown in a modal. Consistency and learnability would then be implemented by the "Launch"-glyph and the similar look of the information in the Inline Launcher and the Modal Launcher. So the difference here is not which object/process is launched but from where the object/process is launched: Is there a single (or maybe two or three) launchable processes on the page (e.g. the Info tab)? Or are there multiple launchable objects/processes on the page?

In a discussion with @atoedt we in fact found an example where a that "multiple launchable process"-case already exists: As far as I understood, in a 360 feedback I might be asked to give feedback for multiple other users. With the Inline Launcher I would have one of these components per other user, which might become overwhelming at some amount of other users. With a Modal Launcher, we could instead display a list of users, with each having one Launch Button only and the detail information in a Modal.

I think it's fine to start with the Inline Launcher, but perhaps we just keep that idea of a Modal Launcher in mind if we find other instances of that "multiple launchable process"-pattern.

Best wishes!

klees:
I close this one, as this is superseded by #5558.

PR #5029, opened 2022-09-16
OrgU: 34470, typcast params

nhaagen:
- addError with string-param
- ids as int

https://mantis.ilias.de/view.php?id=34470


klees:
Picked to trunk.

PR #5030, opened 2022-09-16
[FIX] PhpCsFixer will leave `declare(strict_types=1)` on the first line again.

thibsy:
Hi @mjansenDatabay 

I've noticed that my PhpCsFixer would put the strict-types declaration on the third line of the file again (similar as to the file generation with PHPStorm as mentioned in #4967). This PR introduces this rule to the code format config for the PhpCsFixer as well.

Kind regards!

### DIFF
--- a/CI/PHP-CS-Fixer/code-format.php_cs
+++ b/CI/PHP-CS-Fixer/code-format.php_cs
@@ -23,6 +23,7 @@ return (new PhpCsFixer\Config())
         '@PSR12' => true,
         'strict_param' => false,
         'concat_space' => ['spacing' => 'one'],
-        'function_typehint_space' => true
+        'function_typehint_space' => true,
+        'blank_line_after_opening_tag' => false,
 	])
 	->setFinder($finder);


mjansenDatabay:
Hi @thibsy,

I did some checks and it worked fine, but after some further readings I indeed found some additional rules in [`PSR-12`](https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements).

> The header of a PHP file may consist of a number of different blocks. If present, each of the blocks below MUST be separated by a single blank line, and MUST NOT contain a blank line. Each block MUST be in the order listed below, although blocks that are not relevant may be omitted.
> 
> - Opening <?php tag.
> - File-level docblock.
> - One or more declare statements.
> - The namespace declaration of the file.
> - ....

So in my understanding, the `declare` statement(s) MUST NOT be in the first line. If we all agree on this, we should also revert #4967 .

Maybe we need to move the license header above the `declare` statements to be fully compliant.

Best regards,
@mjansenDatabay

thibsy:
Hi @mjansenDatabay and thanks for checking this!

IMO we should stay PSR compliant and therefore revert #4967 and move the license header before the declare-statements. If @klees agrees as well you can close this PR.

Kind regards!

klees:
I agree. But I don't feel I'm the gatekeeper here =)

mjansenDatabay:
I reverted #4967 . The license header issue should be adressed in a `rector` IMO.

PR #5031, opened 2022-09-16
TA: 34475 cast points to float

nhaagen:
https://mantis.ilias.de/view.php?id=34475


dsstrassner:
I sincerely appreciate your help @nhaagen

nhaagen:
You're most welcome, @dsstrassner

PR #5032, opened 2022-09-16
[FIX] 0033654: Failed test: Übung exportieren - Invalid path supplied

chfsx:
This fixes specific issues with legacy filesystem changes, but export/import still fails due to other bugs (e.g. in metadata).
https://mantis.ilias.de/view.php?id=33654


alex40724:
Thanks!

PR #5033, opened 2022-09-16
src/Data: Add `Enum` trait

mjansenDatabay:
This PR suggests a `trait` to emulate an (backed) `Enum` type for PHP < 8.1.

@klees suggested to add this `trait` to `src/Data` after @thibsy, @klees and @mjansenDatabay discussed this approach in [`Discord`](https://discord.gg/H9v2v2Ar2T).

Of course it only makes sence if this `trait` is really used for `ILIAS 9`. I am confident that usages can be migrated to a native `Enum` with `rector` later on.

To discuss:
- The `value` property is `public`, so the state could be mutated. Unfortunately the `readonly` modifier, which would solve this issue, will be available with PHP 8.2, where we do not need this `trait` anymore. So to maintain the encapsulated state we could use a `value()` method instead.

### DIFF
--- a/src/Data/Enum.php
+++ b/src/Data/Enum.php
@@ -1,0 +1,113 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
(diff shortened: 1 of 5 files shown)


klees (commented):
Hi @mjansenDatabay,

thx a bunch! I would opt for a `value()`-method, so we do not throw away the guarantees that this will give.

Would you present this on todays JF so people become aware and use this?

Best regards!

mjansenDatabay:
@klees 

> thx a bunch! I would opt for a value()-method, so we do not throw away the guarantees that this will give.

Done

> Would you present this on todays JF so people become aware and use this?

Yes

klees (approved):

matthiaskunkel:
Jour Fixe, 19 SEP 2022 : Facing the changed timeline for ILIAS 9 we have to discuss again the required software for ILIAS 9 and the supported PHP versions. Probably, we do not support PHP 8.0 with ILIAS 9 and therefore do not need this emulation of enum. Decision will be made on the next TB meeting.

mjansenDatabay:
IMO this can be closed

PR #5037, opened 2022-09-19
Data: Proposal for improved HTML metadata

thibsy:
Hi @klees 

I'm sorry for the long delay, but this PR is an unfinalized follow-up of #4695. Let me explain my thoughts on this:

 - [ ] `Tag` I've decided to name this abstract representation of HTML metadata a "tag", since the namespace already tells us about what kind of data is stored inside it.
 - [ ] `Tag::toHtml` I didn't implement this in PHPs `__toString` method, because I thought there could be a need of other methods like `toJson` or even `toXml` in the future, and in this scenario `toHtml` would be more semantic.
 - [ ] `UserDefined` I only introduced this HTML metadata object because there is currently no need for further implementations except key => value pairs that are mapped to the meta tags name and content attribute.
 - [ ] `TagCollection` and `Undefined` were introduced, because I was inspired by your presentation on monoids at the DevConf =). The OpenGraph ([ogp.me](https://ogp.me)) protocol dictates a certain collection of HTML meta tags that differ depending on the OpenGraph [type](https://ogp.me#types). Therefore I thought, this looks monoid-ish because I add multiple tags to (again) a tag and I've tried to structure it similar to the FormInputs or Objectives.
 - [ ] I think there are far too many OpenGraph properties that would need to be implemented, so I've decided to be lazy and implement more generic properties like currently `Text` or `Link`. The properties are then grouped together properly in the `OpenGraph\Factory`.

To start-off I have a few questions:

- [ ] Do you think the OpenGraph metadata could benefit from such a monoid-ish structure? 
- [ ] Do you think we should implement more specific properties, like e.g. `OpenGraph\Title` or `OpenGraph\Image`?
- [ ] Do you think the namespace and name of `Tag` are fitting?
- [ ] Do you think the nesting of the `Html\Factory` and `OpenGraph\Factory` should already be separated in the main factory?

Best regards!

### DIFF
--- a/src/Data/Factory.php
+++ b/src/Data/Factory.php
@@ -8,6 +8,7 @@ namespace ILIAS\Data;
 
 use ILIAS\Data\Clock\ClockFactory;
 use ILIAS\Data\Clock\ClockFactoryImpl;
+use ILIAS\Data\Meta;
 
 /**
  * Builds data types.
@@ -24,6 +25,7 @@ class Factory
      */
     private ?Color\Factory $colorfactory = null;
     private ?Dimension\Factory $dimensionfactory = null;
+    private ?Meta\Html\Factory $html_metadata_factory = null;
 
     /**
      * Get an ok result.
@@ -180,5 +182,10 @@ class Factory
     public function dataset(array $dimensions): Chart\Dataset
     {
         return new Chart\Dataset($dimensions);
+    }
+
+    public function htmlMetadata(): Meta\Html\Factory
+    {
+        return $this->html_metadata_factory;
     }
 }
--- a/src/Data/Factory.php
+++ b/src/Data/Factory.php
@@ -26,6 +26,7 @@ class Factory
     private ?Color\Factory $colorfactory = null;
     private ?Dimension\Factory $dimensionfactory = null;
     private ?Meta\Html\Factory $html_metadata_factory = null;
+    private ?Meta\Html\OpenGraph\Factory $open_graph_metadata_factory = null;
 
     /**
      * Get an ok result.
@@ -186,6 +187,19 @@ class Factory
 
     public function htmlMetadata(): Meta\Html\Factory
     {
+        if (null === $this->html_metadata_factory) {
+            $this->html_metadata_factory = new Meta\Html\Factory();
+        }
+
         return $this->html_metadata_factory;
+    }
+
+    public function openGraphMetadata(): Meta\Html\OpenGraph\Factory
+    {
+        if (null === $this->open_graph_metadata_factory) {
+            $this->open_graph_metadata_factory = new Meta\Html\OpenGraph\Factory();
+        }
+
+        return $this->open_graph_metadata_factory;
     }
 }
--- a/src/Data/Meta/Html/Factory.php
+++ b/src/Data/Meta/Html/Factory.php
@@ -19,30 +19,26 @@ declare(strict_types=1);
 
 namespace ILIAS\Data\Meta\Html;
 
-use ILIAS\Data\Meta\Html\OpenGraph\Factory as OGFactory;
-
 /**
  * @author Thibeau Fuhrer <thibeau@sr.solutions>
  */
 class Factory
 {
-    public function __construct(
-        protected OGFactory $og_factory,
-    ) {
-    }
-
     public function userDefined(string $key, string $value): Tag
     {
         return new UserDefined($key, $value);
     }
 
-    public function undefined(): Tag
+    /**
+     * @param Tag[] $tags
+     */
+    public function collection(array $tags): TagCollection
     {
-        return new Undefined();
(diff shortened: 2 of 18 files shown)


klees (changes requested):
Hi @thibsy,

thanks a lot for working on this. Really nice to see that my monoid sermon was so fruitful =) This is indeed a very good example where a monoid is helpful.

Before I go into the stuff that I saw when reviewing, I try to answer you questions:

> `Tag`: I've decided to name this abstract representation of HTML metadata a "tag", since the namespace already tells us about what kind of data is stored inside it.

Sounds plausible =)

> `Tag::toHtml?  I didn't implement this in PHPs __toString method, because I thought there could be a need of other methods like toJson or even toXml in the future, and in this scenario toHtml would be more semantic.

Sounds plausible as well!

> `UserDefined` I only introduced this HTML metadata object because there is currently no need for further implementations except key => value pairs that are mapped to the meta tags name and content attribute.

This is fine for me. Maybe we should add some documentation on the factory along the lines of "Use something more specific if it exists or consider adding something more specific if you want to use it more often.".

> `TagCollection` and `Undefined` were introduced, because I was inspired by your presentation on monoids at the DevConf =). The OpenGraph (ogp.me) protocol dictates a certain collection of HTML meta tags that differ depending on the OpenGraph type. Therefore I thought, this looks monoid-ish because I add multiple tags to (again) a tag and I've tried to structure it similar to the FormInputs or Objectives.

Yeah, this definitly can be modelled with a monoid structure, which you prove in this very PR. Nice observation =)

> I think there are far too many OpenGraph properties that would need to be implemented, so I've decided to be lazy and implement more generic properties like currently Text or Link. The properties are then grouped together properly in the OpenGraph\Factory.

This sounds fine for me, as long as we do not look into inspecting these properties in our software. If I understand correctly, you would not expose these classes on the factory, but instead add more semantic constructors like `website` if needed, right? This means that we could always elaborate the internal representation, if necessary. So yes, go on like this.

To start-off I have a few questions:

> Do you think the OpenGraph metadata could benefit from such a monoid-ish structure?

Yeah, I definitly think so. [This here](https://github.com/ILIAS-eLearning/ILIAS/pull/5037/files#diff-23975a6e75073826f01b8423fcb17b2401d19825bd17c98ee301b8b5e685cc8cR43) already is a nice readability benefit.

> Do you think we should implement more specific properties, like e.g. OpenGraph\Title or OpenGraph\Image?

No, as long as we do plan to inspect these properties internally, semantics on the factory seems to be sufficient (see above).

> Do you think the namespace and name of Tag are fitting?

Yes, LGTM.

> Do you think the nesting of the Html\Factory and OpenGraph\Factory should already be separated in the main factory?

Yes, I thought so too.

So, in general, we are almost there, I think.

- [x] Linebreaks: Add them [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5037/files#diff-0fa4f1f4713ab86c3ee7066ebebacaca1efef398ac8a75118b0da8d315724562R44) for some reading improvements.
- [x] Nested Collections: You might want to collapse nested collections [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5037/files#diff-0fa4f1f4713ab86c3ee7066ebebacaca1efef398ac8a75118b0da8d315724562R34). This should make formatting for output (see previous point) simpler. Alternatively you could add something like `Tag::getSingleTags() : Iterator`, which might be even more handy. Both these points (Linebreaks, Nested Collections) are super advanced high gloss polish improvements, though.
- [x] Null: Please use `NullTag` or maybe `NoTag` instead of `Undefined` (first one prefered). This would clearly hint at the Null Object Pattern.

- [x] Remove [this use statement](https://github.com/ILIAS-eLearning/ILIAS/pull/5037/files#diff-23975a6e75073826f01b8423fcb17b2401d19825bd17c98ee301b8b5e685cc8cR26).
- [x] Tests: Please add automated tests.

Ping me on discord if you want to discuss something =)

Best regards!

thibsy:
Hi @klees

I recognized that there are [structured properties](https://ogp.me/#structured) like `Audio`, `Image` and `Video`, which may be added to any object-type as many times as required. Therefore I Introduced `OpenGraph\Resource` which is a `TagCollection` as well. It really came in handy to implement this whole thing as a monoid here =). Though, I don't feel quite sure about using the spreading operator like I did in this abstract class.

I also figured that ILIAS currently doesn't need any other object-type than website, therefore the factory was only extended with these structured properties. I've added documentation about not using these resources in any other way except this factory, because they are not valid metadata without other metadata added by `website()`, but I don't know if thats enough.

Kind regards!

klees:
I have merged this as https://github.com/ILIAS-eLearning/ILIAS/commit/74277fc7cb0a3438d12834bdaf686180fe4a29d5

klees:
Thanks a bunch, nice work =)

PR #5041, opened 2022-09-20
Added security.txt

fsesterh:
As requested here: https://docu.ilias.de/goto_docu_frm_7625_8323.html


security.txt
```diff
@@ -0,0 +1,5 @@
+Contact: email security at lists.ilias.de
```
corro: I don't think this meets the specification. Shouldn't it start with `mailto:` in case of an email address?
```
Contact: mailto:security@lists.ilias.de
```
Source: https://www.rfc-editor.org/rfc/rfc9116#name-contact

PR #5047, opened 2022-09-21
Form/DateDuration: Fix resetting optional date duration filter

mjansenDatabay:
https://mantis.ilias.de/view.php?id=34570

If approved, this should be cherry-picked to `trunk` as well.


alex40724:
Done.

PR #5048, opened 2022-09-21
OrgU: 34562, edit/create authorities for positions

nhaagen:
https://mantis.ilias.de/view.php?id=34562
needs https://github.com/ILIAS-eLearning/ILIAS/pull/5044


klees:
Picked to trunk.

PR #5049, opened 2022-09-21
[FIX] 0034523: Failed test: Datei hochladen

chfsx:
@Amstutz / @klees as mentioned in https://mantis.ilias.de/view.php?id=34523

But I misunderstood the Close-Signal, therefore I implemented an additional Closed-Signal (which naming is not good, i know, better Ideas appreciated...).

You therefore can listen to a triggered Signal on closing a Modal, the Dropzone\File\Wrapper uses this to clean the filelist (as mentioned in the initial bug ticket).

### DIFF
--- a/Services/Init/classes/Dependencies/InitUIFramework.php
+++ b/Services/Init/classes/Dependencies/InitUIFramework.php
@@ -13,7 +13,8 @@
  * us at:
  * https://www.ilias.de
  * https://github.com/ILIAS-eLearning
- */
+ *
+ *********************************************************************/
 
 /**
  * Responsible for loading the UI Framework into the dependency injection container of ILIAS
@@ -208,6 +209,7 @@ class InitUIFramework
         };
         $c["ui.factory.dropzone.file"] = function ($c) {
             return new ILIAS\UI\Implementation\Component\Dropzone\File\Factory(
+                $c["ui.signal_generator"],
                 $c["ui.upload_limit_resolver"],
                 $c["ui.factory.input"],
                 $c["lng"]
--- a/src/UI/Component/Closable.php
+++ b/src/UI/Component/Closable.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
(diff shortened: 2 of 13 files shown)


chfsx:
We could name it `onClosedSignal` instead of `ClosedSignal`

klees (changes requested):
Hi @chfsx,

the current nomenclature for the client side actions is actually really bad and I get confused all the time, but:

* A `Signal` is something I can connect to, that can cause what its name says. The `CloseSignal` is supposed to close the modal.
* The things that I connect to `Signal`s are called `Triggerer`s. If they trigger, the connected `Signal`s actually do the things they are supposed to 

AFAIK you are looking to build a Triggerer. Please have a look into `src/UI/Implementation/Component/Dropdown/Dropdown.php` for an example.

@Amstutz We need to devise a plan to change that wording some time...

Best regards!

chfsx:
ah of course, I knew I missed something. I will propose an Interface `Closable` (or `Dismissable`?) like other (`Hoverable` etc.) and remake the implementation of the whole PR...

chfsx:
@klees / @Amstutz I updated the PR accordingly:
- New Interface `Closable` which is implemented by Modals (we could rename it to `Dismissable` if needed
- Wrapper-Dropzones are now Triggarable and has a Clear-Signal
- Closing a Modal now triggers the Clear-Signal

chfsx:
Sorry, just wait a minute, there is some code wrong due to amending last commit, I'll update and force-push soon with running tests

chfsx:
There you go

klees (commented):
Hi @chfsx,

thanks for looking into this. The code looks fine, just two little things came to mind.

- [x] Trigger: I'm most sure that I'm missing something, but: Where is the closed-triggerer of the modal actually invoked? Or is this done by Bootstrap himself?

- [x] Debug: [These](https://github.com/ILIAS-eLearning/ILIAS/pull/5049/files#diff-2d9abcacf08b7e7e7a931e40d78ee734af912ab8af6659fa4b7a570dd349f632R494) are for debugging purpose, right? Can we remove them?

In general I find it really interesting, how boilerplatish most of the code looks. Although current naming is bad, this at least speaks to the fact that the abstractions behind it seem to be fine.

Best regards!

Amstutz:
> Trigger: I'm most sure that I'm missing something, but: Where is the closed-triggerer of the modal actually invoked? Or is this done by Bootstrap himself?

As far as I understand, by the hidden.bs.modal event thrown by bootstrap:
image

Which is again cause by closing the modal in il.UI.Modal.closeModal
image

Which again is registered in the renderer:
image

hidden.bs.modal then generates the new signal according to the proposal of @chfsx 
image

chfsx:
Thanks @Amstutz for the explanations, nothing to add :-) and I removed the two console.logs in my last commit as well.

klees (approved):

klees:
Picked to trunk.

klees:
Thanks a lot!

PR #5050, opened 2022-09-22
Survey: Don't use typed properties in ILIAS 7.x

mjansenDatabay:
https://mantis.ilias.de/view.php?id=34581


fsesterh:
Bugfix PR for https://mantis.ilias.de/view.php?id=34581

PR #5053, opened 2022-09-22
Test: Fix Copying Test in ILIAS 8

kergomard:
See: https://mantis.ilias.de/view.php?id=34508

There is a little bit of boy-scouting in here, but I think the PR is simple enough, to still keep it readable. If not: Let me know.


kergomard:
I will close this and fix directly.

PR #5054, opened 2022-09-23
Fix Saving Of Point After Enter

kergomard:
This fixes: https://mantis.ilias.de/view.php?id=33681
I additionally removed the dependencies from jquery from the function.


dsstrassner:
Wird sich Jens freuen :-)

jweber-unibe:
Hat er schon! ;-D

PR #5057, opened 2022-09-23
[FIX] removed duplicate showMessageForm() call.

thibsy:
Hi @pascalseeland

@mjansenDatabay drew my attention to the (IMO redundant `showMessageForm` call) within the callback I introduced while implementing the `CallbackDuration`, so I felt kind of responsible for leaving this code-snippet in there. First we thought the method-call triggers a `RuntimeException` due to the new shutdown-function of the HTTP-service, but then we realized that `ilGlobalTemplate::printToStdout` doesn't contain an exit-statement (anymore?).

Anyways, the `showMessageForm` will also be called after the callback finished.

Kind regards!

### DIFF
--- a/Services/Init/classes/class.ilPasswordAssistanceGUI.php
+++ b/Services/Init/classes/class.ilPasswordAssistanceGUI.php
@@ -210,11 +210,15 @@ class ilPasswordAssistanceGUI
 
             $usrId = \ilObjUser::getUserIdByLogin($username);
             if (!is_numeric($usrId) || !($usrId > 0)) {
-                \ilLoggerFactory::getLogger('usr')->info(sprintf(
-                    'Could not process password assistance form (reason: no user found) %s / %s',
-                    $username,
-                    $email
-                ));
+                \ilLoggerFactory::getLogger('usr')->info(
+                    sprintf(
+                        'Could not process password assistance form (reason: no user found) %s / %s',
+                        $username,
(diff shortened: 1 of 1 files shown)


thibsy:
P.S. if I'm wrong about it's redundancy you can close this PR, theres no real issue with the code AFAIK.

mjansenDatabay:
IMO the call is **not** redundant.

I see two options to fix this:

Option 1) Pass a `&$rendered` (initially `false`) flag to the callback and set it to `true` if something within the callback renders/prints HTML. Outside the callback, check if `!$rendered` and call `$this->showMessageForm(sprintf($this->lng->txt('pwassist_mail_sent'), $form->getInput('email')));` like it is done now.
Option 2) Move the `$this->showMessageForm(sprintf($this->lng->txt('pwassist_mail_sent'), $form->getInput('email')));` to the callback

thibsy:
@mjansenDatabay sorry for the late response, but I just checked and the call *is* redundant and leads to the page being printed to stdout twice (only visible when inspecting the page-source) if the user-id wasn't found. I therefore changed the label to bugfix as well because this could lead to signals not working since all IDs are used twice. I also think the intention of the `showMessageForm` I removed was to stop execution at this point, since the user-id is invalid.

mjansenDatabay:
LGTM @thibsy

PR #5060, opened 2022-09-24
Test: Fix Error in Feedback of Long Menu Question

kergomard:
See: https://mantis.ilias.de/view.php?id=34536


kergomard:
I merge this a cherry-pick to the other branches. I think this is trivial enough. This gets something off your plate.

PR #5061, opened 2022-09-25
Implement "Abandon Generate PDF by User in ILIAS Test"

kergomard:
I believe this should be it for the removal of PDF printing. But I'm completely unable to test this right now as the export and import is still thoroughly broken.

Thus I set this to draft and assign only you @dsstrassner .


kergomard:
As far as I can see, this is ready for review. The PDF-Export in the archive is very uuuglyy. But this it was also before. MathJax-rendering only works with server-side rendering. But this is again the same as before.

kergomard:
I merge this now, after approval by @dsstrassner .

PR #5062, opened 2022-09-26
Test: Fix Removal of "Pseudo-Tags" in Matching questions

kergomard:
See: https://mantis.ilias.de/view.php?id=25312

This is once more a sign of the bad design of the question forms. We will really need to tackle this.


mbecker-databay:

PR #5065, opened 2022-09-26
Text: Fix Removal of "Pseudo-Tags" in Single Choice Question

kergomard:
This is basically the same solution as applied in #5062 ported to the SingeChoice- and the MultipleChoiceQuestion.

See: https://mantis.ilias.de/view.php?id=32047

As always: Just let me know if you would like PRs to release_8, as soon as you have checked this.


kergomard:
And one more comment: There is a part in here, where we always array_map the same way, so we could centralize this. I would suggest not to do it though as we don't actually want to encourage the use of it. So I would suggest to duplicate for the moment.

mbecker-databay:

PR #5066, opened 2022-09-26
Test: Fix Removal of "Pseudo-Tags" in Textsubset

kergomard:
See: https://mantis.ilias.de/view.php?id=28982

Ok, one more of those, I  hope it is the last one. This is now the ugliest by far. But there seem to be no two Question-Types running through this exactly the same way.


mbecker-databay:

PR #5073, opened 2022-09-27
TA: 33524, fill form after error

nhaagen:
https://github.com/ILIAS-eLearning/ILIAS/pull/5072
https://mantis.ilias.de/view.php?id=33524
addendum; fill form after error


dsstrassner:
Thanks a bunch!

PR #5075, opened 2022-09-27
WIP: refactor deprecated update steps

iszmais:
This resolves deprecations withing the Skill update step, but isn't finished due to conceptional issues:
- Some deprecations should be resolved by objectives, wich (as far as i know) can not provide the same functional structure as update steps in total.
- e.g. Some of these refactored Objectives require the result/finished processing of another objective while they are collected inside the same group.
- The targeted objectives require a DB.

**Unfortunatly i dont have time to investigate this further. But im up for questions and further explanations.**

### DIFF
--- a/Services/AccessControl/classes/class.ilRbacReview.php
+++ b/Services/AccessControl/classes/class.ilRbacReview.php
@@ -1332,10 +1332,12 @@ class ilRbacReview
         self::$assigned_users_cache = array();
     }
 
-    public static function _getCustomRBACOperationId(string $operation): ?int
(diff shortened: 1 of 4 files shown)


klees:
Hi @iszmais,

welcome back =) I don't see how I would be involved here and assigned the responsible maintainers instead. If there are specific things to be asked or discussed I'm of course available.

Best regards!

tfamula:
Hi @iszmais 
Thank you very much for your PR, it helped me a lot and served as a template! I had a meeting with @klees and the result of our discussion can be found in the new PR #5123.

PR #5076, opened 2022-09-28
[FIX] PHP Deprecated:  Implicit conversion from float YX to int loses…

chfsx:
… precision

This is a followup to #5028. PHP8.1 needs an explicit conversion as proposed here. Without, a `PHP Deprecated:  Implicit conversion from float YX to int loses precision` is thrown.


klees:
Thx!

PR #5077, opened 2022-09-28
[FIX] Increased min-version of PHP-CS-Fixer library

thibsy:
Hi @klees

This PR changes the min-version of the php-cs-fixer library to one that supports php8.1, since this is the new min-version for ILIAS 9.

Kind regards!


klees:
@mjansenDatabay Could you have a look here? If I remember correctly you did some things regarding tooling and ILIAS 9 and I wonder why this wasn't covered.

Feel free to assign back, though.

Thanks!

mjansenDatabay:
Thx @thibsy

PR #5080, opened 2022-09-29
Bugfix 30706 Error commenting on a learning sequence

sasfalk:
#bugfix
Mantis issue: 0030706
Target: release_7
https://mantis.ilias.de/view.php?id=30706

Bug was reported for Ilias 6.9, but still occurs in Ilias 7, so I fixed it there.


klees:

PR #5085, opened 2022-09-29
Changes Language Variables in Test

kergomard:
See: https://mantis.ilias.de/view.php?id=34445

@mbecker-databay : It would be kind, if you didn't leave this too long, as it has changes to lang-files in it and they are a pain in the butt to fix once we run into conflicts.

I can take care of backporting and cherry-picking once you give me the green light.


mbecker-databay:
But of course!

PR #5086, opened 2022-09-29
TA: 34727, fix manual scoring of file-upload

nhaagen:
https://mantis.ilias.de/view.php?id=34727

### DIFF
--- a/Modules/Test/classes/class.ilTestScoringByQuestionsGUI.php
+++ b/Modules/Test/classes/class.ilTestScoringByQuestionsGUI.php
@@ -306,7 +306,11 @@ class ilTestScoringByQuestionsGUI extends ilTestScoringGUI
         }
 
         if ($ajax && is_array($correction_feedback)) {
-            $correction_feedback['finalized_by'] = ilObjUser::_lookupFullname($correction_feedback['finalized_by_usr_id']);
+            $finalized_by_usr_id = $correction_feedback['finalized_by_usr_id'];
+            if (! $finalized_by_usr_id) {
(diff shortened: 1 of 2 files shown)


kergomard:
If I'm getting it right you are solving three bugs here. Could you please just quickly tell me under what circumstance [finalized_by_user_id](https://github.com/ILIAS-eLearning/ILIAS/pull/5086/files#diff-39682dc33dedfc701bc6d7348101149b707dcdba10eb573336ec8b4438611694R309) is something resolving to false?
Thanks!

dsstrassner:
Also https://mantis.ilias.de/view.php?id=33664

nhaagen:
Hi @kergomard ,
I disklike the whole construct, however, this was "fixing while passing by".
$finalized_by_user_id is read from tst_manual_fb, and there it is allowed to be null.
It certainly SHOULD not be null, but then there seem to be cases where it is not enforced, or set to a clearly wrong value.

https://github.com/ILIAS-eLearning/ILIAS/blob/3a77acc9a83c1bf1cc491fd8b5d0bbf15f29525b/Modules/Test/classes/class.ilObjTest.php#L9508

show fields from tst_manual_fb;
| Field                | Type       | Null | Key | Default | Extra |
..
| finalized_by_usr_id  | bigint(20) | YES  |     | NULL    |       |

kergomard:
Thank you very much @nhaagen ! Sorry for having missed that! I will merge and cherry-pick in a minute.

nhaagen:
nono, good to have someone check on things like that ;)

PR #5087, opened 2022-09-29
Legacy Progress Bar: fix JS loading (#34595)

klees:
Hi @alex40724,

this fixes https://mantis.ilias.de/view.php?id=34595 in the most straight forward way. This is not neat or anything, but I would suspect this legacy component to vanish some time not too far away anyway, so I don't think we should invest more here.

Hope this is fine for you.

Best regards!

### DIFF
--- a/Services/UIComponent/ProgressBar/templates/default/tpl.il_progress.html
+++ b/Services/UIComponent/ProgressBar/templates/default/tpl.il_progress.html
@@ -14,7 +14,7 @@
 	<img src="{ICON_OK}" />
 </span>
 <script>
-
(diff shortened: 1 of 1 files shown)


alex40724:
I doubt that these are stable fixes for the issue. I had to refactor similar code already, e.g. here https://github.com/ILIAS-eLearning/ILIAS/commit/4e8ea7cd70c1ef9d1a8fd4f3679c1629c06ec4ec

It might work "sometimes" but it seems to produce race conditions or other strange effects depending on whatever circumstances.

PR #5091, opened 2022-09-30
Backport of https://github.com/ILIAS-eLearning/ILIAS/pull/5085

kergomard:
This is the backport of #5085 to release_7


kergomard:
Ok, I'm sorry. I actually already committed these changes together with another fix. As they are basically the same as in the other branches I will leave them in and not revert. Sorry again!

PR #5098, opened 2022-10-04
Fix for 34788

schmitz-ilias:
This should fix [34788](https://mantis.ilias.de/view.php?id=34788) (and also fixes the icons being oversized). As far as I can see, this also needs to be merged into R8 and trunk.


Amstutz:
Nice, thx a lot @schmitz-ilias ! Special kudos for also fixing the oversize issue.

klees:
Hi @dsstrassner,

Staff is maintained by @mstuder.

Best regards!

tfamula:
Thanks @schmitz-ilias! I cherry-picked this also for release 8 and trunk, but without the type cast (because of the return type of the method)

PR #5101, opened 2022-10-05
fix 31362 part 2

okaufman:
This will fix : https://mantis.ilias.de/view.php?id=31362#c86573 
By removing "id=tst_output" the view will be the same as if "Exam View/ Prüfungsansicht" was not activated. 
It removes "display:flex" in the CSS.

### DIFF
--- a/Modules/TestQuestionPool/templates/default/tpl.il_tst_question_hints_kiosk_page.html
+++ b/Modules/TestQuestionPool/templates/default/tpl.il_tst_question_hints_kiosk_page.html
@@ -1,4 +1,4 @@
 {KIOSK_HEAD}
-<div id="tst_output">
+<div>
 	{KIOSK_CONTENT}
 </div>


dsstrassner:
Hi @Amstutz since this is CSS related I assign you also 😉

Amstutz:
Hi @dsstrassner 

I removed myself. It's rather a question, if this div is needed for anything (besides css). If not we can remove it. However, this is more a JS or plugin interface question than css.

Amstutz:
I changed the state to draft here. I do not think this will work, since there seem to be JS deps towards tst_output, right @okaufman ?

okaufman:
tst-output appears only in ilCharSelector.js.
Since the CharSelector is not used in ilAssQuestionHintRequestGUI I don't think it would be a problem.

Amstutz:
@okaufman , thx for the feedback. What is ilAssQuestionHintRequestGUI doing? I see that it is called in ilTestPlayerAbstractGUI which again calls showQuestionEditable in some scenarios, which then again calls on showQuestionEditable which uses the char selector depending on the settings. Did you check that none of those scanrios apply to the tpl here?

okaufman:
ilAssQuestionHintRequestGUI shows up when a user takes a test with hints and the user clicks on "Lösungshinweis anfordern" in a specific question. It shows the different Hints for a question. The user can't edit the hints or write any new text and therefore has no use for the CharSelector. 
When editing the hint a different class is used(ilassquestionhintgui) that never uses this template.

The template is never used when the edit-Modus is enabled. showQuestionEditable is only called when edit mode is on. So these two never never influence eachother.

Amstutz:
oke, for the research.

PR #5104, opened 2022-10-07
Fix Question Edit Links not Adding to Locator

kergomard:
This is nothing big, but as I'm changing the command structure a second pair of eyes is appreciated.
See: https://mantis.ilias.de/view.php?id=34715

As always: You can leave the cherry-picking to me.

### DIFF
--- a/Modules/Test/classes/tables/class.ilTestQuestionsTableGUI.php
+++ b/Modules/Test/classes/tables/class.ilTestQuestionsTableGUI.php
@@ -15,6 +15,7 @@ require_once 'Services/UIComponent/Glyph/classes/class.ilGlyphGUI.php';
 
 class ilTestQuestionsTableGUI extends ilTable2GUI
 {
+    private const CLASS_PATH_FOR_EDIT_LINKS = ['ilrepositorygui', 'ilobjquestionpoolgui'];
 
     /**
      * @var bool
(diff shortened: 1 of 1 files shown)


Modules/Test/classes/tables/class.ilTestQuestionsTableGUI.php
```diff
@@ -15,6 +15,7 @@
 
 class ilTestQuestionsTableGUI extends ilTable2GUI
 {
+    private const CLASS_PATH_FOR_EDIT_LINKS = ['ilrepositorygui', 'ilobjquestionpoolgui'];
```
mjansenDatabay: I prefer to use the `::class` constants, but this is just personal taste. Maybe this could be also problematic because a `array_map('strtolower', self::CLASS_PATH_FOR_EDIT_LINKS)` has to be applied then.

mjansenDatabay: I am not 100% sure if this route is correct. Whether this route could be resolved depends on the value of `$target_class`. IMO this should be thoroughly tested. If you already tested the 3 occurances of this table, I am fine with this :-).

kergomard: Jup and jup:
1. I generally would agree with you that ```::class``` constant is more elegant. I will change that, as
2. I completely missed the corrections class! Thank you very much for the close look. Update follows...

kergomard:
Ok, I believe this should now cover all uses. I separated the creation of the PreviewLink from the creation of the EditLinks. This also tailors the URL-Paramters more to the situation.
Sorry for the oversight!

PR #5106, opened 2022-10-08
Test: Fix Formula Question Presentation

kergomard:
See: https://mantis.ilias.de/view.php?id=34227

We should actually have a look at this whole branching catastrophe for ILIAS 9, but this branches that cannot (```if (false &&```) or should not (see the current leakage of the result in the question view) be taken.

I wouldn't leave this too long, if you don't want another gazillion duplicate Mantis-Issues ;-)

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assFormulaQuestion.php
+++ b/Modules/TestQuestionPool/classes/class.assFormulaQuestion.php
@@ -299,67 +299,26 @@ class assFormulaQuestion extends assQuestion implements iQuestionCondition
                 ) {
                     $is_frac = true;
                 }
-                if (is_array($userdata) && isset($userdata[$result])) {
-                    if (is_array($userdata[$result])) {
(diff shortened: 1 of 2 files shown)


Amstutz:
A good, easy and almost risk-free option would be making short named functions with max up to 10 lines of those blocks, uncle bob style. This would then improve readability and ease of further refactoring greatly.

kergomard:
Yup, I'm all in favor! I would suggest though, to still merge this PR ASAP and then quickly coordinate the next step(s). I've put it on our list for discussion.

mbecker-databay:
I take it that this fix is for ILIAS 8 only and then based on our discussion on thursday we'll see another one for the trunk?

kergomard:
I would suggest to still cherry-pick to trunk and then take care of it. This way we keep trunk as usable as possible. Additionally this makes the code more readable anyway so any further changes will e easier.

PR #5108, opened 2022-10-10
Fix orderning of Question Export

kergomard:
See: https://mantis.ilias.de/view.php?id=14179

Ok, we had two different behaviors in the question export: In Excel questions where newly sorted before showing them in CSV they weren't.
This aligns the behavior. I refrained from entering deeply into the structure of these functions and kept my changes shallow. But the class needs some serious refactoring!

Behavior after the change:
* Questions are sorted by their sequence if the test is not based on a RandomQuestionSet. If it is based on a RandomQuestionSet questions are sorted by their qid.
* There is only one header row, if the test is not based on a RandomQuastioonSet. In the latter case one header row per result set is shown.
* I removed a small issue that the excel export would show the header row of the first result set on a new line, thus creating a split in the header rows of the first set.

As always @mbecker-databay : I will cherry-pick to 8 and 9 if this is accepted solution.

### DIFF
--- a/Modules/Test/classes/class.ilTestExport.php
+++ b/Modules/Test/classes/class.ilTestExport.php
@@ -401,7 +401,7 @@ abstract class ilTestExport
             $col = 0;
             
             // each participant gets an own row for question column headers
-            if ($this->test_obj->isRandomTest()) {
+            if ($this->test_obj->isRandomTest() && $firstwritten) {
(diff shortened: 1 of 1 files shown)


kergomard:
Ok, I will simply merge and cherry-pick. This gets us one more thin off our plates.

PR #5109, opened 2022-10-10
T&A: Feature: New option for excel export of ILIAS test statistics

fhelfer:
Feature Wiki: https://docu.ilias.de/goto_docu_wiki_wpage_7405_1357.html


dsstrassner:
As the client, did Alina check it out and give her OK?

mbecker-databay:
I requested an approval instance.

dsstrassner:
Hi @fhelfer, 
since the deadline for ILIAS 9 is moved to 2023, I see the chance for refactoring this complex methods to improve the T&A code wholesome. @klees will review this PR and suggest some improvements due to his engagement in the TA TechSquad.

@alinaseibt and I will test the whole implementation afterwards at the approval instance and if everything works like a charm, write also the test cases for TestRail and add the related information to the FR Wiki page.

mbecker-databay:
I am good with this one! Please advise once this one can be merged.

Modules/Test/classes/class.ilCSVTestExport.php
```diff
@@ -0,0 +1,96 @@
...
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+use JetBrains\PhpStorm\NoReturn;
```
mjansenDatabay: Should re really use `JetBrains` features here, even if the are optional and don't break any code if a developer does not use `PhpStorm`?

Modules/Test/classes/class.ilTestExportAbstract.php
```diff
@@ -0,0 +1,247 @@
...
+        global $DIC;
+        $this->lng = $DIC->language();
+        $this->test_obj = $test_obj;
+        $this->filterby = $filterby;
+        $this->filtertext = $filtertext;
+        $this->passedonly = $passedonly;
+    }
+
+    abstract public function deliver(string $title): void;
+    abstract public function getContent();
```
mjansenDatabay: Could we add a return type here?

Modules/Test/classes/class.ilExcelTestExport.php
```diff
@@ -0,0 +1,415 @@
...
+        }
+        return $this;
+    }
+
+    public function getContent(): ilAssExcelFormatHelper
+    {
+        return $this->worksheet;
+    }
+
+    public function withUserPages(): self
```
mjansenDatabay: This method is quite big. We should try to extract some methods, or event better, classes.

mbecker-databay:
@dsstrassner If you can give us a poke about this, we can merge it while it still smells good.

Modules/Test/classes/class.ilCSVTestExport.php
```diff
@@ -0,0 +1,96 @@
...
+        $separator = ";";
+        foreach ($rows as $evalrow) {
+            $csvrow = &$this->test_obj->processCSVRow($evalrow, true, $separator);
+            $csv .= implode($separator, $csvrow) . "\n";
+        }
+        $this->content = $csv;
+        return $this;
+    }
+
+    #[NoReturn]
```
mjansenDatabay: See comment above.

Modules/Test/classes/class.ilTestExportAbstract.php
```diff
@@ -0,0 +1,247 @@
...
+    protected bool $passedonly;
+
+    public function __construct(
+        ilObjTest $test_obj,
+        string $filterby = '',
+        string $filtertext = '',
+        bool $passedonly = false
+    ) {
+        global $DIC;
+        $this->lng = $DIC->language();
```
mjansenDatabay: If all other dependencies are passed in the constructor, we should consider passing the language as well.

Modules/Test/classes/class.ilTestExportAbstract.php
```diff
@@ -0,0 +1,247 @@
...
+
+/**
+ * @author Fabian Helfer <fhelfer@databay.de>
+ */
+abstract class ilTestExportAbstract
+{
+    protected ilObjTest $test_obj;
+    protected ilLanguage $lng;
+    protected string $filterby;
+    protected string $filtertext;
```
mjansenDatabay: IMO these two parameters are semantically weak. What is a `filterby`, what is a `filtertext`? What are valid criteria to filter the export by? I suggest do create explicit classes and replace this primitive string parameters.

Modules/Test/classes/class.ilTestExportAbstract.php
fhelfer: `abstract public function getContent(): ilAssExcelFormatHelper|string;`

dsstrassner:
Update: @fhelfer will provide a second instance. I will check the implementation there again and comment afterward.

dsstrassner:
I tested the feature along the [Feature Request ](https://docu.ilias.de/goto_docu_wiki_wpage_7405_1357.html) and everything works now as expected on the approval instance. I am ok with merging it @mbecker-databay ;)

Edit: 12:56h: Alina also tested it and only gets my test runs and I get now an error: "Sorry, an error occured. A logfile has been created which can be identified via the code "cjln6_6516"." Maybe we have to wait a little longer ...

fhelfer:
> Edit: 12:56h: Alina also tested it and only gets my test runs and I get now an error: "Sorry, an error occured. A logfile has been created which can be identified via the code "cjln6_6516"." Maybe we have to wait a little longer ...

Fixed
It was an Ilias Bug that had been fixed for 8. So rebasing did the trick

dsstrassner:
Now everything works fine. Thanks a lot @fhelfer

PR #5110, opened 2022-10-10
0029975 failed test online status

kunzeus:
Hello

This PR should fix the bug presented in: https://mantis.ilias.de/view.php?id=29975
It changes the default import behavior of ilTestObjects. Instead of defaulting to online it now uses the online/offline-state set by the ilContainerImporter.
It also fixes what, I believe, is a bug in the ilContainerImporter, at least in my test instance. The bug set other object lm, crs, grp etc. to online after the import, ignoring the online/offline-state of the imported XML file. With the proposed fix, the online/offline-state of the imported XML file is used for container imports.

I'm still learning ILIAS, so please let me know: if anything slipped my attention, or what can be improved further.

### DIFF
--- a/Services/Container/classes/class.ilContainerImporter.php
+++ b/Services/Container/classes/class.ilContainerImporter.php
@@ -134,7 +134,8 @@ class ilContainerImporter extends ilXmlImporter
                 if ($this->isRootNode($obj->getRefId(), $mapping)) {
                     $obj->setOfflineStatus(true);
                 } else {
-                    $obj->setOfflineStatus(false);
+                    // take the offline state of the XML file, and map it back: online to online and offline to offline
+                    $obj->setOfflineStatus(!((string)$offline === "0"));
                 }
                 $obj->update();
             }
--- a/Services/Container/classes/class.ilContainerImporter.php
+++ b/Services/Container/classes/class.ilContainerImporter.php
@@ -134,8 +134,8 @@ class ilContainerImporter extends ilXmlImporter
                 if ($this->isRootNode($obj->getRefId(), $mapping)) {
                     $obj->setOfflineStatus(true);
                 } else {
-                    // take the offline state of the XML file, and map it back: online to online and offline to offline
-                    $obj->setOfflineStatus(!((string)$offline === "0"));
+                    // use the offline status of the imported XML file and set the offline status for the new container objects accordingly
+                    $obj->setOfflineStatus(!((string) $offline === "0"));
                 }
                 $obj->update();
             }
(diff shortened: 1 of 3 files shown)


dsstrassner:
Hi @alex40724 

I assign you also, since @kunzeus want to change [Services/Container/classes/class.ilContainerImporter.php](https://github.com/ILIAS-eLearning/ILIAS/pull/5110/files#diff-51abd091353fb1633711a84c3e56aeba593385716da3fe4c8b1ffa21ad809423R138)

alex40724:
@smeyer-ilias Could you have a look at the changes of your code in the ContainerImporter, please.

kergomard:
@smeyer-ilias : Could you please have a look at the part that concenes you?

dsstrassner:
Dear JourFixe, 

how should we proceed?

Services/Container/classes/class.ilContainerImporter.php
```diff
@@ -134,7 +134,8 @@ protected function handleOfflineStatus(string $xml, ilImportMapping $mapping)
                 if ($this->isRootNode($obj->getRefId(), $mapping)) {
                     $obj->setOfflineStatus(true);
                 } else {
-                    $obj->setOfflineStatus(false);
+                    // take the offline state of the XML file, and map it back: online to online and offline to offline
+                    $obj->setOfflineStatus(!((string)$offline === "0"));
```
mjansenDatabay: This violates our coding standard, a space is missing after the `(string)` cast.

matthiaskunkel:
Jour Fixe, 09 JAN 2023: PM will contact @smeyer-ilias and ask for feedback.

dsstrassner:
Dear JF, I like to inform you all, that as there is no reaction occurred, so @mbecker-databay  or @kergomard will merge this PR accordingly in the next days.

smeyer-ilias:
Thanks  @kunzeus for fixing this. Your changes streamline the online / offline handling for imported objects using the centralized offline handling from ilObject. Objects which do not use that offline handline might still set a wrong status. 

Since this is tagged with the Jour-Fixe label, i want to ask whether it is really necessary to always import question pools with status "online".

smeyer-ilias (approved):

matthiaskunkel:
Jour Fixe, 23 JAN 2023: We highly appreciate this PR and accept it for ILIAS 6 and newer versions. We do not see any reason to set question pools always online. We prefer the behaviour described by the new import where only the parent object (e.g. course) is set offline and all embedded object keep their status (online or offline).

kunzeus:
Hello,

@smeyer-ilias  Currently, the Test Question Pool Object is not supporting offline Handling. Thereby, no offline state is written into the XML Export. But as far as I can say, changing this requires at least a DB update step (il_object_def: id:qpl, offline_handling:0->1) + other modifications to fit into the supporting offline handling category. I don't know what is required to support offline handling, and I would shelf it since it's not part of the fix this PR is about.

So it's either online or offline for all Question Pool imports in the meantime. However, I would address the issue in the next T&A TechSquad meeting.

PR #5111, opened 2022-10-11
KS colorpicker

pbechtold:
@Amstutz, @klees, @mjansenDatabay - Dear TB-Members, please have a look at the public colorpicker interface, if you accept we would like to discuss this interface at the JF on 17th Oct.

Colorpicker Input (closed):
screenshot_color-picker-compnent1

Colorpicker-Chrome (hex):
screenshot_color-picker-compnent_open-hex
Colorpicker-Chrome (rgb):
screenshot_color-picker-compnent_open-rgb
Colorpicker-Firefox:
screenshot_color-picker-compnent_open-firefox

### DIFF
--- a/src/UI/Component/Input/Field/ColorPicker.php
+++ b/src/UI/Component/Input/Field/ColorPicker.php
@@ -1,6 +1,7 @@
 <?php
 
 declare(strict_types=1);
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
--- a/src/UI/Component/Input/Field/ColorPicker.php
+++ b/src/UI/Component/Input/Field/ColorPicker.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+namespace ILIAS\UI\Component\Input\Field;
+
+/**
+ * This describes color picker inputs
+ */
+interface ColorPicker extends FormInput
+{
+}
--- a/src/UI/Component/Input/Field/ColorPicker.php
+++ b/src/UI/Component/Input/Field/ColorPicker.php
@@ -20,7 +20,7 @@ declare(strict_types=1);
 namespace ILIAS\UI\Component\Input\Field;
 
 /**
- * This describes color picker inputs
+ * @author Patrick Bechtold <patrick.bechtold@kroepelin-projekte.de>
  */
 interface ColorPicker extends FormInput
 {
--- a/src/UI/Component/Input/Field/ColorPicker.php
+++ b/src/UI/Component/Input/Field/ColorPicker.php
@@ -1,7 +1,6 @@
 <?php
 
 declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
--- a/src/UI/Component/Input/Field/Factory.php
+++ b/src/UI/Component/Input/Field/Factory.php
@@ -685,6 +685,8 @@ interface Factory
      *      Color Picker will render an input-tag with type="color".
      *   effect: >
      *     As soon as the Color Picker is clicked, a pop-up window opens, which contains the individual options of the color selection.
+     * context:
+     *   - The Color Picker input is used in UI-forms.
      * rules:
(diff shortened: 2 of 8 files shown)


tfamula:
Hi [pbechtold](https://github.com/pbechtold)!
Congratulations on your first PR for the kitchen sink! Just my two cents:

- [x] Accessibility: Do you know if the color picker is accessible? Can it also be used with e.g. a keyboard besides clicking with a mouse? I would like to see some accessibility rules in the public interface.

Amstutz:
Hi @pbechtold 

Thx a lot for this proposal and the screenshots showcasing it. Note to everybody:

Color Picker will render an input-tag with type="color".

This means, the proposal suggest to ONLY future render the Color HTML 5 Field (no Text field next to it connected to the input by JS or similar). We like this proposal since it seems to most standard and error-prone why to use this as possible. We also discussed this with our a11y at Webkey, and got the following feedback: 

"It is a very good color picker. It is keyboard accessible, and the elements are labelled and read out through a screen reader well. I think that this would be a good option to use as a color picker. Just be aware, that it has better compatibility and UX in Chrome than in Firefox."

Tests showed, that in Firefox exist problems with the keyboard navigation. However, since only standards are proposed to be used here, we believe this to be an issue of firefox, not the implementation.

However, since a11y is somewhat of an issue here, we propose to add the following a11y rule:

- [x] Accessibility: As with all Inputs, the Color Picker Input MUST be operable by only using inputs. If HTML5 Standards are used, it is the responsibility of the Browser to provide this functionality.

To document this here in the Component itself.

Thx a lot
@Amstutz and @klees

Amstutz:
Hi @tfamula 

>  Accessibility: Do you know if the color picker is accessible? Can it also be used with e.g. a keyboard besides clicking with a mouse? I would like to see some accessibility rules in the public interface.

We discussed this during a meeting prior to the PR. My feedback above probably partly answers the question. Note that there is the following rue for all Field Inputs:

>      *     1: >
>      *         All fields visible in a view MUST be accessible by keyboard by using the
>      *         ‘Tab’-Key.

However, the proposal suggests to only use HTML5 defauts, which leaves it to the browser to handle the control over the element. Chrome does a good job. Firefox not so much (also see screenshots above and feedback from webkey).

matthiaskunkel:
Hi @pbechtold , Thanks for the PR. Getting a new colour picker would be great. But it should also support the colour schemes used in the ILIAS page editor. Especially when using the page editor you want to present the supported colours easily to customers - and do not want them to use their own colours.

pbechtold:
> However, the proposal suggests to only use HTML5 defauts, which leaves it to the browser to handle the control over the element. Chrome does a good job. Firefox not so much (also see screenshots above and feedback from webkey).

Dear @tfamula, 
thx a lot. Just as Timon has nicely written, the advantage is that here a standard html input element is used and the accessibility is therefore also supported by the browser.

pbechtold:
> a11y

Dear @Amstutz ,

thank you very much for the useful advice. I still added the rule to the interface (PR), I think this is quite a useful hint for the documenation, thx!

pbechtold:
> Hi @pbechtold , Thanks for the PR. Getting a new colour picker would be great. But it should also support the colour schemes used in the ILIAS page editor. Especially when using the page editor you want to present the supported colours easily to customers - and do not want them to use their own colours.

Dear @matthiaskunkel , 
I hope I have understood your hint correctly. When integrating the color picker, u always have the possibility to specify the values, thus also the implementation of the default values.

alex40724:
> ...is that here a standard html input element is used and the accessibility is therefore also supported by the browser.

Unfortunately browsers do not always a good job when it comes to HTML5 input elements and accessibility. I looked into this for the Date/Time inputs some time ago, where we are still using a "custom widget" instead of the browsers native inputs. Even with its own accessibility issues, the custom solution was the better choice at that time imo. See e.g. https://www.hassellinclusion.com/blog/input-type-date-ready-for-use/

I did not do any research on the color inputs though, but I would not generally assume that the native inputs are the best choice when it comes to accessibility.

matthiaskunkel:
Jour Fixe, 17 OCT 2022: we highly appreciate this suggestion and accept the PR for trunk. In a second step we can try to tackle the problem with selecting pre-defined colours from the colour scheme of a content style.

pbechtold:
Dear UI coordinators,
this commit includes the implementation as well as the related tests, the integration within the slate does not present any difficulties either (see screenshot).
Screenshot 2022-10-19 161614

klees (changes requested):
Hi @pbechtold,

thanks a lot for working on this. Looks straightforward and mostly GTM.

I only have on change request:

* Unification: Please use [this form](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/UI/Implementation/Component/Input/Field/Factory.php#L47) [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5111/files#diff-03d0345eb7047ff616377ce4e735de2fb2a85ea54f2ec3b8675c4c7e8cd056f0R43). Its always easier if same things look the same.

Best regards!

Amstutz:
Since the requested change is only minor and this should be usable for upcoming ILIAS9 devs, I merged this and added the requested change myself just after merging: https://github.com/ILIAS-eLearning/ILIAS/commit/3fa544a15f39eac79baad5c85fac0a24e1c45df0

PR #5114, opened 2022-10-12
UI: Proposal to replace esm with babel-register

thibsy:
Hi folks,

I stumbled upon an issue when using mocha.js for clientside unit tests, where the `esm` module-bundler does not support the latest language features anymore, e.g. null-coalescing or optional-chaining):

```bash
trunk/src/UI/templates/js/Input/Field/src/Markdown/markdown.factory.js:40
        return this.instances[input_id] ?? null;
                                         ^

SyntaxError: Invalid or unexpected token
    at Module._extensions..js (node:internal/modules/cjs/loader:1173:10)
```

This is due to `esm` no longer being maintained (last official release in 2019). Therefore, I suggest with this PR to replace this package with [`@babel/register`](https://www.npmjs.com/package/@babel/register).
I'm honest, I didn't quite dig into this stuff with babel, and introduced another package called [`@babel/preset-env`](https://www.npmjs.com/package/@babel/preset-env). This allows to import the latest configuration for babel that supports ES2015+.

This is just a proposal, and I know introducing new dependencies isn't nice, but IMO we have to replace `esm` anyways in order to comply with the latest Javascript standards. So how do we proceed here?

Kind regards!


klees:
Oh, how I hate having that node_modules code in the repo.

matthiaskunkel:
Jour Fixe, 17 OCT 2022: We highly appreciate this suggestin and accept it for trunk.

Amstutz:
Hi @thibsy 

Thx a lot for this PR. Its certainly the right direction. Currently @klees and @Amstutz are assigned, however, this is somewhat bigger than the scope of the UI Components. Therefore we will need to discuss who will take responsibility for this. The Technical Board label is therefore correct. We definitely need additional JS expertise to look over PRs like this.

There might be hope in the future ;-), we keep you posted.

Thx
@Amstutz

Amstutz:
As previously noted, this is bigger than UI Components. However, we are very very happy, that we have additional expertise in terms of JS among the UI Coordinators. We can and will therefore take care of this. We shortly discussed this internally among the UI Coordinators. The reasoning Thibeau is giving in the description sound solid. The risk of merging that to 7, 8 and trunk is minimal, since only client side tests are affected. @thibsy since this has aged somewhat. Please have another look over the implementation and make sure, everything works as intended. If all is fine, feel free to merge this. Thx a lot!

PR #5116, opened 2022-10-13
Bugfix 31323 Next Button was deactivated in visited Content Page

sasfalk:
#bugfix
Mantis issue: 0031323
Target: release_7
https://mantis.ilias.de/view.php?id=31323

The next button of a Learning Sequence was incorrectly deactivated when the first page needs to be visited for the user to continue. This is because the next button has already been created when the content is loaded and the page is set to visted.
Thats why I swapped this steps and reload the items before creating the next-Button, so the visited-Status is set correctly.


klees:
Thanks a bunch!

klees:

PR #5121, opened 2022-10-13
fix js error on initial toast load

iszmais:
https://mantis.ilias.de/view.php?id=34590

This PR mostly reverts changes made in https://github.com/ILIAS-eLearning/ILIAS/pull/5012 and add all toasts that are preexisting to the toast container withing the pagebuilder to prevent an additional request.

### DIFF
--- a/src/GlobalScreen/Scope/Layout/Builder/StandardPageBuilder.php
+++ b/src/GlobalScreen/Scope/Layout/Builder/StandardPageBuilder.php
@@ -60,6 +60,7 @@ class StandardPageBuilder implements PageBuilder
         $title = $parts->getTitle();
         $short_title = $parts->getShortTitle();
         $view_title = $parts->getViewTitle();
+        $toast_container = $parts->getToastContainer();
 
         $standard = $this->ui->factory()->layout()->page()->standard(
             [$parts->getContent()],
@@ -69,7 +70,7 @@ class StandardPageBuilder implements PageBuilder
             $header_image,
             $responsive_header_image,
             $favicon_path,
-            $this->ui->factory()->toast()->container(),
+            $toast_container,
             $footer,
             $title,
             $short_title,
--- a/Modules/Chatroom/GlobalScreen/ChatInvitationNotificationProvider.php
+++ b/Modules/Chatroom/GlobalScreen/ChatInvitationNotificationProvider.php
@@ -43,7 +43,7 @@ class ChatInvitationNotificationProvider extends AbstractNotificationProvider
         $latest_time = 0;
         $osd_notification_handler = new ilNotificationOSDHandler(new ilNotificationOSDRepository($this->dic->database()));
         $invitations = [];
-        foreach ($osd_notification_handler->getNotificationsForUser(
+        foreach ($osd_notification_handler->getOSDNotificationsForUser(
             $this->dic->user()->getId(),
             true,
             time() - $leftIntervalTimestamp,
--- a/Services/GlobalScreen/classes/Setup/class.ilGlobalScreenBuildProviderMapObjective.php
+++ b/Services/GlobalScreen/classes/Setup/class.ilGlobalScreenBuildProviderMapObjective.php
@@ -1,5 +1,21 @@
 <?php
 
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
 declare(strict_types=1);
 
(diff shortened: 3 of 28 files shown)


iszmais:
@lscharmer @Amstutz @klees Pls leave your opinion on this.
I wouldnt see the need to take that to a JF until the is no major discussion point.
Feel free to open such.

Greeting,
@iszmais

klees:
Hi @alex40724,

you are listed as implicit maintainer for Services/Notification, hence I assigned you. Please reassign if there is a person you deem more suitable.

Thanks!

klees (approved):
I'm not much into the whole feature and problem, but this looks plausible indeed.

mjansenDatabay:
It is `Services/Notifications`, so I deassigned @alex40724 and assigned @mjansenDatabay

klees:
Oh, that's an intricate distinction =)

mjansenDatabay:
The changes LGTM in general.

@chfsx There is one change in `src/GlobalScreen` (https://github.com/ILIAS-eLearning/ILIAS/pull/5121/files#diff-5769142c345fc3a6bc2301605fe73f7876384f4504e108611a562c99fff70f2b). Do you have any objections?

src/GlobalScreen/Scope/Layout/Builder/StandardPageBuilder.php
```diff
@@ -60,6 +65,16 @@ public function build(PagePartProvider $parts): Page
         $title = $parts->getTitle();
         $short_title = $parts->getShortTitle();
         $view_title = $parts->getViewTitle();
+        $toast_container = $this->ui->factory()->toast()->container();
+
+        foreach (
```
lscharmer: This is just a minor thing but I would suggest to extract something into a variable for clarity.
Something like this:
```php
$toasts = (new ToastsOfNotifications($this->ui->factory(), new ilSetting('notifications'))
    ->transformNotificationsForToasts(
        (new ilNotificationOSDHandler())->getNotificationsForUser($this->user->getId())
    );
foreach ($toasts as $toast) {
    $toast_container = $toast_container->withAdditionalToast($toast);
}
```

lscharmer (changes requested):

lscharmer (changes requested):
Please squash the commits.

lscharmer (approved):
Thank you :)

src/GlobalScreen/Scope/Layout/Builder/StandardPageBuilder.php
```diff
@@ -60,6 +65,14 @@ public function build(PagePartProvider $parts): Page
         $title = $parts->getTitle();
         $short_title = $parts->getShortTitle();
         $view_title = $parts->getViewTitle();
+        $toast_container = $this->ui->factory()->toast()->container();
+
+        $transformer = new ToastsOfNotifications($this->ui->factory(), new ilSetting('notifications'));
```
chfsx: IMO notifications should be collected the same was as other types of notifications are collected (e.g. see SystemInfos, which has an own Collector and is called here src/GlobalScreen/Scope/Layout/Provider/PagePart/StandardPagePartProvider.php:225). we already have the whole `MainNotificationCollector` thing to collect notifications, would be great we could use this pattern for other types of notifications as well.

chfsx (changes requested):

iszmais:
@chfsx Is there still something missing your requirements?
If yes feel free to leave a comment or otherwise i would appreciate your approval.

Greetings,
@iszmais

chfsx:
Hi @iszmais 

Yes actually my comment from last time is still relevant, but I that you have pushed something to this. We should look at this together. The `StandardPageBuilder` should not get too much dependencies, but especially no new ones. You have addressed this by allowing the `PagePartProvider` to supply the TostContainer. But also in `StandardPagePartProvider` we should keep the dependencies small. For certain things this is not possible (e.g. breadcrumb with ilLocator). In the case of the toasts it would be a good idea to collect the single toasts by ToastProviders and a ToastCollector. Or you can use (maybe better) the existing `NotificationCollector`, like it is done for the `AdministrativeNotifications` (see for example `\ILIAS\GlobalScreen\Scope\Layout\Provider\PagePart\StandardPagePartProvider::getSystemInfos`). Otherwise the toasts will have some kind of special status and it will be difficult to keep track of them in the future.

would you like to have a chat on discord next week?

iszmais:
Hello @chfsx ,

I abstracted the toast provider into an own scope within the global screen. Imo it should not be combined or integrated inside the already exsiting notification scope since the intention of creating the toast was do separete the over the years tangled chat and notfication therefore i think we should start that tangeling again.

Maybe you already can take a look into this i will also fix the tests to it.
I would prefer to present the changes of the existing providers (ChatInvitation, Badge, WhoIsOnline, ...) in seperate PRs wich are depending on this just to make it easier for the target maintainers.

Greetings,
@iszmais

iszmais:
Correlation to https://github.com/iszmais/ILIAS/pull/7

chfsx:
Hi @iszmais 

Thank you for your comments and the discussion in https://github.com/iszmais/ILIAS/pull/7. I pushed you two more commits there that we should still include (one is not directly related to your change, but I noticed it because it didn't have any data on the ToastProviders in the artefact yet). 
Your explanation of the distinction between toasts and notifications convinced me, so we can merge it like that!

chfsx (approved):
Thanks a lot @iszmais !

PR #5124, opened 2022-10-13
Test: Fix PHP 8.1 issues with `ilAssQuestionHintList`

mjansenDatabay:


Modules/TestQuestionPool/classes/class.ilAssQuestionHintList.php
```diff
@@ -19,102 +19,53 @@
...
-     * iterator interface method
-     *
-     * @access	public
-     * @return	mixed
-     */
-    public function next()
+    public function next(): void
     {
-        return next($this->questionHints);
+        next($this->questionHints);
```
mbecker-databay, line 45: Is this correct? I'd think next is supposed to return a value of mixed

Modules/TestQuestionPool/classes/class.ilAssQuestionHintList.php
```diff
@@ -19,102 +19,53 @@
...
-    /**
-     * iterator interface method
-     *
-     * @access	public
-     * @return	mixed
-     */
-    public function key()
+    public function key(): int
     {
         return key($this->questionHints);
```
mbecker-databay, line 50: Is this correct? Per docs, key returns int|string|null

Modules/TestQuestionPool/classes/class.ilAssQuestionHintList.php
mjansenDatabay: According to the docs, next returns nothing

Modules/TestQuestionPool/classes/class.ilAssQuestionHintList.php
mjansenDatabay: You can add a precise type in derivatives

PR #5126, opened 2022-10-14
Setup: Make `ObjectiveIterator` iterator compatible with PHP 8.1

mjansenDatabay:


klees:
Thanks @mjansenDatabay!

PR #5127, opened 2022-10-14
Legacy-Form: Fix Error with Nested RadioGroups

kergomard:
This commit here
https://github.com/ILIAS-eLearning/ILIAS/commit/9aa888407b50c891fad6c0c53ea6c2c0ebada7a2#diff-442c3a79da205f00191d1f37414279a6c89e4543dccc90e3dcd7981e54b0f03bR132
has side-effects. `getParentForm()` can return null and does so in the case of nested ilRadioGroups. In these cases the code has been loaded anyways and is not needed.

Mantis report here: https://mantis.ilias.de/view.php?id=34974


alex40724:
Yes, thanks, this is fixed in the meantime.

PR #5128, opened 2022-10-14
[PROPOSAL] remove node_modules from ILIAS repository.

thibsy:
Hi folks,

In reaction upon @klees [comment](https://github.com/ILIAS-eLearning/ILIAS/pull/5114#issuecomment-1278610244) in #5114 I thought why not just remove the node_modules from the repository?

Reason alone could be my computer nearly crashing due to the huge diff of this PR alone. Same goes for the PR I just referenced above, which makes it really hard to figure out what the PR *actually* adds to the codebase.

It's also kind of inconsistent to include node_modules in the repository whereas the composer dependencies will have to be installed manually.

I believe I don't have to tell you about why this is a good thing :). If you strongly disagree of this though, please let me know.

Kind regards!


thibsy:
P.S. I adjusted the documentation of the repository in `install.md`, if this PR should be merged the official documentation on the website needs to be updated as well.

matthiaskunkel:
Jour Fixe, 17 OCT 2022: We see several advantages of this PR but need to discuss with the BB if we can go this way and introduce another dependency (npm) that are a problem for some installations. One option would be to offer ready builts additionally.

PurHur:
Please have a really good look on the security side of this issue. NPM is not the yellow from the egg when it comes to security. 

Adding
`--ignore-scripts`
to the docs should be done and only skipped if its really necessary. If a package needs this you should consider to replace it. Since the code of an npm package doesnt have to match e.g. a given Github Repo its nearly impossible to review the source correctly before deploying an update.

See:
https://cheatsheetseries.owasp.org/cheatsheets/NPM_Security_Cheat_Sheet.html

thibsy:
Thx @PurHur for drawing this to our attention. I updated the installation docs of the repository accordingly.

kergomard:
Thanks again @thibsy for taking the initiative for this! The Technical Board discussed your proposal at our meeting and we agree with the direction this takes. As this has the potential to complicate the installation and maintenance of ILIAS, we will need a little bit more time though, to come up with a good plan on how to provide a updates and installations in the future and how to communicate it well.

kergomard:
Hi @thibsy
Could you please be so kind and resolve the conflicts once more? I would then merge this right away.
Thank you very much!
@kergomard

thibsy:
Hi @kergomard 

This should be good to go now. Thx a lot!

Kind regards
@thibsy

kergomard:
Thank you very much @thibsy !

PR #5132, opened 2022-10-16
Legacy-Table: Fix missing Header on Checkboxes

kergomard:
This fixes the table checkbox headers for tables as they pose a problem for accessibility.

See: https://mantis.ilias.de/view.php?id=32871

If there is a top select all checkbox at the top they didn't work at all up to now, as they was a small typo. Additionally if there was no checkbox there was a `<th>` instead of a `<td>`.


Amstutz:
@kergomard thx a lot for the fix! Since this resolves an a11y issue in a legacy ui component with only implicit maintainance, Ill take the liberty and responsibility to merge this. @akill, I hope this is fine with you. Let me know if not. If any issues arise caused by this, you might forward it to me. 

Thx a lot.

PR #5136, opened 2022-10-18
ilCtrl: introducing new best-practises

thibsy:
This is yet a work-in-progress, which could serve as a baseline for the developer-workshop im planing for #5100.

### DIFF
--- a/Services/UICore/best-practises.md
+++ b/Services/UICore/best-practises.md
@@ -1,0 +1,62 @@
+# Introducing new ilCtrl best-practises
+
+This manual explains how ilCtrl should be used in the future to be able to move into a direction where static routes can
+become a possibility. It covers how ilCtrl should be used when implementing new features or components, and how current
+code could be migrated to meet ilCtrl's new standards.
+
+This guide will most definitely not be able to cover every scenario because there are some very unique use-cases that
+cannot be guided by some general best-practises (though it might still be helpful).
+
+## DON'TS (what you shouldn't do)
+
+### `ilCtrl::setCmdClass`
+
+**You MUST STOP using this method, it's been deprecated and will be removed with ILIAS 10.** This method-call can be avoided
+in most cases by simply generating absolute ilCtrl-links with `ilCtrl::getLinkTargetByClass` and supplying it with the
+complete array of classes from base-class to target-class.
+
+### `ilCtrl::setCmd`
+
+**You MUST STOP using this method, it's been deprecated and will be removed with ILIAS 10.** Normally, `setCmd` can simply
+be avoided by passing along the right command from the beginning when link-targets are generated with `ilCtrl::getLinkTargetByClass`. 
+
+### Manipulating `ilCtrl::getCmd`
+
+Apart from manipulating the output of this method with `ilCtrl::setCmd`, you also MUST NOT change the return-value of this
+method call in any way. Concatenating substrings to the command after it's retrieved prevents to properly assume which method
+will be called in the target class.
+
+### Using `ilCtrl::getLinkTarget` and `ilCtrl::getFormAction`
+
+You SHOULD NOT use these methods anymore, because the more fully-qualified link-targets are built right now the easier becomes
+the migration to static routes sometime in the future. Instead of using e.g. `getLinkTarget($this)` you SHOULD use 
+`ilCtrl::getLinkTargetByClass([BaseClass, MaybeAnotherGUI, self::class])`.
+Not being able to dictate ilCtrl the fully-qualified path indicates a design-flaw that should be restructured anyways.
+
+### Concatenating link-targets without using ilCtrl 
+
+You MUST NOT build link-targets on your own, by concatenating strings to look like an URL returned by ilCtrl. This could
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @thibsy for this proposal!
Even if we lost it out of our sights for a while, we think this is an important next step to soon be able to modernize our routing. We thus set it on the agenda for the next Jour Fixe and very much support the introduction fo these rules.

Thanks again and best,
@kergomard in the name of the Technical Board

alex40724:
Thanks for the PR. One question: getHTML main purpose is to allow to combine UI parts from different components. These components use ilCtrl, but are not part of the current route, e.g. on the dashboard or in the repository. Can this be achieved by forwardCommand? Currently I think it cannot, an example would be appreciated. Originally forwardCommand should only be used to forward within the current route. Otherwise we loose the possibility to combine blocks (or panels, if blocks/columns are abandoned) which make use of ilCtrl themselves.

alex40724:
Second, the suggestion to always use "full paths" contradicts general architecture principles like "separation of concers" imo. It adds additional coupling between components, e.g. subcomponents now need knowledge about their contexts, which was not the case in the past. If classes in the contexts are refactored, this might break paths in subcomponents (or plugins).

matthiaskunkel:
JourFixe, 12 JUN 2023: Thibeau explained the current PR and the suggested best-practices and recommends to longr use the deprecated methods in ilCtrl. Alexander asks for alternative solutions to reuse / combine parts of UI which won't no longer be possible when getHTML will be deprecated. We accept the PR to push the general code revision of ILIAS but asks to look for possible solutions for the aforementioned problems. A developer workshop to bring this topic forward would be highly appreciated.

thibsy:
Thx @alex40724 for bringing up your concerns. I will try to answer them briefly:

1. `ilCtrl::getHTML()`: This method has been used to combine parts of the UI from different services or modules in the past, but I discourage new implementations to rely on this mechanism, since IMO this should not be of any concern for the router. This cannot be achieved with `ilCtrl::forwardCommand()`. There is no alternative to this method yet, a new pattern and/or service has to be implemented first.
2. Fully-qualified link targets: I advise developers to use fully-qualified link-targets whenever possible because this will bring forward modules or services which heavily rely on the dynamic aspect of `ilCtrl`. I am almost certain, that modularity can still be maintained, even though routes will become static. This only means that controllers must be implemented in a different manner and developers must start thinking in terms of "services" instead of "GUI layers".

I will definitely follow-up on these topics, but for the time being I will merge these new best practises to discourage developers from using `ilCtrl` in a way it should not be used anymore. This focuses primarily on new implementations.

Kind regards,
@thibsy

alex40724:
@thibsy Thanks for your explanation. But sorry that I am not convinced. I brought up a very concrete architectural issue:

More fully qualified targets -> More knowledge necessary of other components internals  -> higher coupling

This is the current effect. It's not good.

The benefits of the best practice change is not really explained. The most concrete term (besides vague things like "being modern" or "not think in GUI layers") is currently "static routes". This needs a definition, since it is not a self-explaining concept and originally from the networking field. How will the static routes look like, what are the benefits and why do we need fully qualified targets in the code to foster them? How should controllers be implemented in a different manner? All this remains unclear. There is no ROADMAP.md in the UICore component that states the goal and the steps necessary to achieve it.

But the negative effect of the best practices (higher coupling) is imminent.

mjansenDatabay:
So with my maitainer 🎩 on, assessing the factual level, I would like to also emphasize that I share the thoughts of @alex40724 regarding the `fully qualified targets` (see: https://github.com/ILIAS-eLearning/ILIAS/pull/5136/files#diff-5f929fe0dffd9ffd6940583ae8fd6c347452fde5e8a3481c6099dcb65a65f3f9R29-R34).
In my opinion we just cannot remove `get(LinkTarget|FormAction)($this)` without providing an alternative mechanism. Re-using views (GUI classes) is a fundamental part of the "GUI Layer" in ILIAS. Almost every component providing user interfaces is affected by this approach.

Examples:
- Permission handling of all repository objects (in the repository, and the administration)
- Metadata
- Public profile
- Member gallery
- Info screen
- Export views
- Learning progress views
- Certificate editor
- Hints for test questions
- Feedback for test questions
- Taxonomies
- Page editor
- News
- LTI (provider)
- you name it ...

Without doubt, I agree that establishing this resuability in the `Routing` component(s) (`ilCtrl`) of an application is most probably a design flaw (when comparing this with modern application architecture). Furthermore I would say, creating an alternative MUST NOT be the sole task of @thibsy . Extracting this "GUI Layer Reusability" out of `ilCtrl` and providing an alternative approach will be a huge project IMO, which requires 🧠 power of many smart ILIAS developers. IMO this cannot be a side project while we all (providers of such resusable views, as well as consumers) are buisy with other "big projects" (like `Removing LUI` etc.).
If providers of this reusable views cannot rely on an elaborated and central approach, our code base will actually get even worse, because every developer will create own fancy solutions to provide/consume such reusable views.

I am absolutely fine with the other best-practices, above all the deprecation of the mutators.

PR #5137, opened 2022-10-18
TA: 34877, immutable assAnswerMatchingTerm

nhaagen:
https://mantis.ilias.de/view.php?id=34877

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assMatchingQuestion.php
+++ b/Modules/TestQuestionPool/classes/class.assMatchingQuestion.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -223,9 +225,9 @@ class assMatchingQuestion extends assQuestion implements ilObjQuestionScoringAdj
                 array(
                                     $next_id,
                                     $this->getId(),
-                                    $pair->points,
-                                    $termids[$pair->term->getIdentifier()],
-                                    $definitionids[$pair->definition->getIdentifier()]
+                                    $pair->getPoints(),
+                                    $termids[$pair->getTerm()->getIdentifier()],
+                                    $definitionids[$pair->getDefinition()->getIdentifier()]
                                 )
             );
         }
@@ -284,21 +286,25 @@ class assMatchingQuestion extends assQuestion implements ilObjQuestionScoringAdj
         if ($result->numRows() == 1) {
             $data = $ilDB->fetchAssoc($result);
             $this->setId($question_id);
-            $this->setObjId($data["obj_fi"]);
+            $this->setObjId((int)$data["obj_fi"]);
             $this->setTitle((string) $data["title"]);
             $this->setComment((string) $data["description"]);
             $this->setOriginalId($data["original_id"]);
-            $this->setNrOfTries($data['nr_of_tries']);
+            $this->setNrOfTries((int)$data['nr_of_tries']);
             $this->setAuthor($data["author"]);
-            $this->setPoints($data["points"]);
-            $this->setOwner($data["owner"]);
+            $this->setPoints((float)$data["points"]);
+            $this->setOwner((int)$data["owner"]);
             include_once("./Services/RTE/classes/class.ilRTE.php");
             $this->setQuestion(ilRTE::_replaceMediaObjectImageSrc((string) $data["question_text"], 1));
-            $this->setThumbGeometry($data["thumb_geometry"]);
+            $this->setThumbGeometry((int)$data["thumb_geometry"]);
             $this->setShuffle($data["shuffle"] != '0');
             $this->setShuffleMode((int)$data['shuffle']);
             $this->setMatchingMode($data['matching_mode'] === null ? self::MATCHING_MODE_1_ON_1 : $data['matching_mode']);
-            $this->setEstimatedWorkingTime(substr($data["working_time"], 0, 2), substr($data["working_time"], 3, 2), substr($data["working_time"], 6, 2));
+            $this->setEstimatedWorkingTime(
+                (int)substr($data["working_time"], 0, 2),
+                (int)substr($data["working_time"], 3, 2),
+                (int)substr($data["working_time"], 6, 2)
+            );
 
             try {
                 $this->setLifecycle(ilAssQuestionLifecycle::getInstance($data['lifecycle']));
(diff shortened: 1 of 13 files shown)


dsstrassner:
Thanks @nhaagen ;-)

Hi @mbecker-databay and @kergomard,  as @mbecker-databay is today AFK, would you do me the favor to review the pull request and @kergomard merge this afterwards, as this is a blocker for the test-run?

kergomard (changes requested):
Hi @nhaagen 

Thank you very much for this. I think this request makes our code more predictable and readable!

I would kindly ask you though to implement the following changes/suggestions:
- [ ] Right now, this only works because the classes creating the Terms and Definitions are not strictly typed. I would suggest to make sure we are passing properly typed parameters, when creating the Terms and Definitions or do you see good reasons not to do so?
- [ ] Please also check once more for other usages of the refactured classes and its child. There is eg. a usage in the corresponding Feedback-Class [here](https://github.com/nhaagen/ILIAS/blob/_8/TA/34877/Modules/TestQuestionPool/classes/feedback/class.ilAssMatchingQuestionFeedback.php#L52) and there might be others, I'm too tired right now to find.

Thanks again and best,
@kergomard

nhaagen:
please note that I changed the random-id generation here: https://github.com/ILIAS-eLearning/ILIAS/pull/5137/files#diff-46b8b111ec1f161c22ea2de2f883e139afe1c23ad2c467b356d23420dcfcc30aR41
with mt_random, I always got the same result.

kergomard:
@nhaagen : This is ready for review, right?
Concerning the rng: Fine with me, we should then also get rid of the seeding. It always was rather useless (as far as I can see), but is completely useless now. I would do so after merging, so you don't need to bother with it.

nhaagen:
Hi @kergomard ,
I removed the seed from the randomizer and from my point of view this good to go.
Thanks for checking.

kergomard (approved):

Modules/TestQuestionPool/classes/class.assMatchingQuestion.php
```diff
@@ -1129,7 +1130,9 @@ public function saveWorkingData($active_id, $pass = null, $authorized = true): b
     {
         global $DIC;
         $ilDB = $DIC['ilDB'];
-
+        $request = $DIC->http()->request;
+        var_dump($request);
```
kergomard: It think this shouldn't be here, right?

Modules/TestQuestionPool/classes/class.assMatchingQuestion.php
```diff
@@ -1069,19 +1065,24 @@ public function setImageFile($image_tempfilename, $image_filename, $previous_fil
...
 
-        foreach ($this->getDefinitions() as $definition) {
-            if (isset($postData[$definition->identifier])) {
-                foreach ($this->getTerms() as $term) {
-                    if (isset($postData[$definition->identifier][$term->identifier])) {
-                        if (!is_array($postData[$definition->identifier])) {
-                            $postData[$definition->identifier] = array();
+        $request = $this->dic->testQuestionPool()->internal()->request();
+        $post = $request->getParsedBody();
+        //var_dump($post);
```
kergomard: This shouldn't be here, right?

Modules/TestQuestionPool/classes/class.assMatchingQuestion.php
```diff
@@ -350,10 +355,20 @@ public function loadFromDb($question_id): void
...
+                    $termids[$data['term_fi']],
+                    $definitionids[$data['definition_fi']],
+                    (float)$data['points']
+                );
+                /*                    new assAnswerMatchingPair(
+                                        $termids[$data['term_fi']],
+                                        $definitionids[$data['definition_fi']],
+                                        $data['points']
+                                    ));
+                */
```
kergomard: Could you please remove the comment or is there a reason whey we should absolutely keept it?

Modules/TestQuestionPool/classes/class.assMatchingQuestion.php
```diff
@@ -284,21 +286,25 @@ public function loadFromDb($question_id): void
         if ($result->numRows() == 1) {
             $data = $ilDB->fetchAssoc($result);
             $this->setId($question_id);
-            $this->setObjId($data["obj_fi"]);
+            $this->setObjId((int)$data["obj_fi"]);
             $this->setTitle((string) $data["title"]);
             $this->setComment((string) $data["description"]);
             $this->setOriginalId($data["original_id"]);
```
kergomard: This now fails, because of the strict_types

Modules/TestQuestionPool/classes/class.assMatchingQuestion.php
```diff
@@ -318,11 +324,10 @@ public function loadFromDb($question_id): void
             array('integer'),
             array($question_id)
         );
-        include_once "./Modules/TestQuestionPool/classes/class.assAnswerMatchingTerm.php";
-        $this->terms = array();
+        $this->terms = [];
         if ($result->numRows() > 0) {
             while ($data = $ilDB->fetchAssoc($result)) {
-                $term = new assAnswerMatchingTerm($data['term'], $data['picture'], $data['ident']);
+                $term = $this->createMatchingTerm($data['term'], $data['picture'], (int)$data['ident']);
```
kergomard:
```suggestion
                $term = $this->createMatchingTerm($data['term'] ?? '', $data['picture'] ?? '', (int) $data['ident']);
```
This can now fail, if $data['term'] === null or $data['picture'] === null

Modules/TestQuestionPool/classes/class.assMatchingQuestion.php
```diff
@@ -334,11 +339,11 @@ public function loadFromDb($question_id): void
             array('integer'),
             array($question_id)
         );
-        include_once "./Modules/TestQuestionPool/classes/class.assAnswerMatchingDefinition.php";
+
         $this->definitions = array();
         if ($result->numRows() > 0) {
             while ($data = $ilDB->fetchAssoc($result)) {
-                $definition = new assAnswerMatchingDefinition($data['definition'], $data['picture'], $data['ident']);
+                $definition = $this->createMatchingDefinition($data['definition'], $data['picture'], (int)$data['ident']);
```
kergomard:
```suggestion
                $definition = $this->createMatchingDefinition($data['definition'] ?? '', $data['picture'] ?? '', (int) $data['ident']);
```
This can now fail, if $data['definition'] === null or $data['picture'] === null

Modules/TestQuestionPool/classes/feedback/class.ilAssMatchingQuestionFeedback.php
```diff
@@ -49,8 +49,8 @@ public function getAnswerOptionsByAnswerIndex(): array
     protected function buildAnswerOptionLabel(int $index, $answer): string
     {
         $caption = $ordinal = $index + 1;
-        $caption .= '. <br />"' . $answer->term->text . '" =&gt; ';
-        $caption .= '"' . $answer->definition->text . '"';
+        $caption .= '. <br />"' . $answer->term->getText() . '" =&gt; ';
```
kergomard: This should be like below and the same goes for the next line, right?
```suggestion
        $caption .= '. <br />"' . $answer->getTerm()->getText() . '" =&gt; ';
```

kergomard (changes requested):
Thanks @nhaagen , I think we are getting there. I've added comments inline, as I think in this case it is easier, but let me know if this doesn't work for you.

kergomard:
Ok, I think this good to go. I will Squash and merge.

PR #5140, opened 2022-10-19
TA: 34793, save ordering/nesting

nhaagen:
https://mantis.ilias.de/view.php?id=34793


nhaagen:
Vorschau/Prüfen still throws...

kergomard:
Thank you very much!

PR #5141, opened 2022-10-19
Bugfix T&A 0031518 negativ point deduction in hints leads to incorrect scoring

kunzeus:
Mantis: https://mantis.ilias.de/view.php?id=31518

Two commits, one should fix newly created hints, the other should fix negative signs already stored in the database.
I decided to rather change the setters in the different data structure classes, then all places where the hint deduction points are accessed. This has the advantage, I hope, that in the future the error is not recreated by accident. 
Also, the user is able to enter decimals for the point deduction. Therefore, I changed some of the comments to float instead of Integer. Since I think it's better to expect the more complex value.

Again, happy to do additional changes to the code or change my strategy if you see things differently. Just let me know so I can improve.


kergomard:

PR #5144, opened 2022-10-21
0035158: Keep online status

smeyer-ilias:
The description in mantis: https://mantis.ilias.de/view.php?id=35158


kergomard:
Thanks @smeyer-ilias .

PR #5145, opened 2022-10-21
[add] new line for hints with no point deduction

kunzeus:
Mantis: https://mantis.ilias.de/view.php?id=29621

Alternative line for Hints without a point deduction:

Screenshot_20221021_154423


dsstrassner:
Thanks @kunzeus!

Hi @matthiaskunkel, please check the new message.

matthiaskunkel:
I suggest to write "No points will be deducted for this hint." But thanks a lot for fixing this issue!!!

kergomard:
Thank you very much @kunzeus for this fix!
It looks good to me, just one question / request for change. Please just explain quickly, if you don't want to follow up on it:
- [ ] It seems to me that we tend to speak of "point deduction" generally. Why did you decide to name the language variable "...no_cost"? This might seem like a nitpick, but It seems to me that this will make it harder to find and follow up on when something changes. I would suggest to rename it "...no_deduction" if there is no technical or semantic reason not to do so.

Thanks again and best,
@kergomard

kunzeus:
The new commit fixes the issue of the naming scheme, the text was already correct, I uploaded the wrong picture.

kergomard:

PR #5146, opened 2022-10-24
35228: Category import fails

alex40724:
See https://mantis.ilias.de/view.php?id=35228, this is also already in release 8


mjansenDatabay:
Thx @alex40724

PR #5149, opened 2022-10-24
Portfolio: Fix counting non-countable variable

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=33381

If approved, this has to be cherry-picked to `releae_8` and `trunk`.


alex40724:
Thanks for the PR. Fixed slightly differnt.

PR #5151, opened 2022-10-24
Blog: Fix using absent "List GUI" in header actions (Mantis 35238)

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=35238

This seems to be no problem in ILIAS >= 8.x.


alex40724:
Thanks for the PR!

PR #4943, opened 2022-08-17
UI: introduction of a markdown input

thibsy:
Hi @Amstutz and @klees

As discussed in our UI-Clinic meeting today (see https://docu.ilias.de/goto_docu_dcl_8186_166_7568.html), I hereby open the PR to kick-off and announce the introduction of a new UI input for markdown (structured text).

It might be possible that the description of the public interface is still incomplete or not quite detailed enough, let me know if and where that's the case. A more detailed explanation of our plan can be found at: https://docu.ilias.de/goto_docu_wiki_wpage_7404_1357.html#ilPageTocA216.

@chfsx and I discussed that an input like this would already benefit from better text-handling in ILIAS as proposed in #4901. During the implementation of this input we might as well tackle the implementation of the markdown-part of this project or at least maybe await it.

I noticed that @klees mentioned the distinction between formatting and structuring in the PR linked above. The KS documentation now currently uses the word "formatting" or "formatted text" - should I adjust this?

Kind regards!

PR #5152, opened 2022-10-24
[FEATURE] UI: Markdown Input

thibsy:
Hi @Amstutz and @klees 

I'm almost finished with the implementation of the UI `Field\Markdown` input, as approved by the JF in #4943. There are some last polishings waiting to be done, but I'd like your opinion on the general structure before moving forward.  

In order to implement the `Field\Markdown` input I refactored almost the entire clientside part of the `Field\Textarea`, because the markdown input is basically a decorated textarea input that can use many of its functionalities. To sum-up the refactoring:

- [x] The markdown- and textarea-input now share a template in `Field\Renderer`, which will be used in the rendering-process of the markdown input to be further decorated.
- [x] The template has been restructured to be wrapped in a container (`<div>`), which makes it more simple to access elements of the input within the javascript.
- [x] The remainder is no longer added on-load. There's no need to pass along the language-variables anymore because I wrapped the actual (integer) remainder in a `<span>` element.
- [x] The max- and min-limit is now passed along to the official textarea-attributes, which prevents the user from adding more characters than possible automatically (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#attr-maxlength).
- [x] The javascript has been reimplemented as ES6 module.
- [x] I added some proper javascript unit-tests.

There is one open thing though:

- [ ] The Kitchensink previously added an additional info-text beneath the textarea (which was also not properly styled) that informed the user about the provided max- and min-limits. I've left a comment in `Field\Renderer::getPreparedTextareaTemplate()`.

This refactoring allowed me to a) share the rendering process of the actual textarea with the newly introduced markdown input and b) to extend my markdown ES6 module from the reimplemented textarea module.

Some open things for the markdown input are:

- [ ] `markdown.less` I noticed that it's really unpleasant to write markdown-syntax with a font-family like e.g. Arial. Therefore I changed the font-family for this particular input to Monospace. Sadly, I couldn't figure out how to do so (css noob here) without using `!important`.
- [x] `Field\Renderer::renderMarkdownField()` I think we need some additional glyphs for the markdown-actions Heading, Link, Bold and Italic. How do we proceed here? I currently just used the first letter of each action-name.
- [x] `Field\Renderer::renderMarkdownField()` the `edit` and `view` language variables are yet missing.
- [x] The PHP unit-tests are not yet complete.

Here's a brief preview of the current result:

Screenshot 2022-10-24 at 15 23 09


Amstutz:
@thibsy, thx a lot for this proposal. A lot of good stuff in here. Many thanks for improving the existing js code base. Thx for using rollup here and for adding tests.

We certainly need some time to diggest all of this, so could take a few days. 

> markdown.less I noticed that it's really unpleasant to write markdown-syntax with a font-family like e.g. Arial. Therefore I changed the font-family for this particular input to Monospace. Sadly, I couldn't figure out how to do so (css noob here) without using !important.

important is fine for certain exceptions. I might look if I find an alternative but for the PR it is oke.

>  Field\Renderer::renderMarkdownField() I think we need some additional glyphs for the markdown-actions Heading, Link, Bold and Italic. How do we proceed here? I currently just used the first letter of each action-name.

If possible, add them to the list of glyphs in the UI Component. Since the changes are minor we can fast-track those to not block further work here. Just add the changes in this PR here.

>   The Kitchensink previously added an additional info-text beneath the textarea (which was also not properly styled) that informed the user about the provided max- and min-limits. I've left a comment in Field\Renderer::getPreparedTextareaTemplate().

What would be your recommendation on dealing with this? Why?

thx.

thibsy:
Hi @Amstutz and thanks for the brief feedback!

I've forgot to mention, that I introduced an implementation of `Field\MarkdownRenderer` in `ilUIMarkdownPreviewGUI` which I planed on using similarly as the `Field\UploadHandler` implementations. It's currently a base-class which it shouldn't be, but the implementation is unfinished anyways because we still need to decide on a library to use for the parsing of markdown-syntax. 

I also introduced some PHPUnit tests for the markdown-input now and four new Glyphs like you said I could: `Header`, `Italic`, `Bold` and `Link`. I named them after their glyphicon classname so they don't get mixed up.

> What would be your recommendation on dealing with this? Why?

IMO we should leave it to the consumers to decide whether they want to include the limits info in their byline or not, 
otherwise these methods implicitly change the byline which I think is odd. One could argue that the same goes for the remainder that will automatically be displayed beneath the textarea, if a max-limit has been provided, and one would be right =). Maybe we could introduce some new methods like `withRemainder()` and `withLimitInformation()` which accept booleans that decide if the according infos should be rendered or not. I'm not a big fan of `withLimitInformation()` though, I think it'd be O.K. to leave this to the consumer entirely.

P.S. please note that the javascript tests will currently fail due to the module-bundler problem we have (#5114).

Kind regards!

Amstutz:
Hi @thibsy 

We did not forget this. However, this is rather huge, we still need a few days for the final review here. Let us know if you need the feedback urgently, e.g. if some other contribution is stuck behind this PR.

thibsy:
Hi @Amstutz, no problem. 

The PR is actually part of the feature-request https://docu.ilias.de/goto_docu_wiki_wpage_7404_1357.html, which is scheduled for ILIAS 9. Therefore, I think it's not yet very urgent =).

klees (changes requested):
Hi @thibsy,

thanks a lot for providing this input. This will help tremendously in the future.

Before we get to the meat, there are three things that we want to point out, for you, but for the general public as well.

Currently, this new `Markdown Input Field` will contain a simple string, as the `Textarea Input Field` does. We [did some thinking regarding text-representation](https://github.com/ILIAS-eLearning/ILIAS/pull/4901) not to long ago, which spawned this [proposal](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/text-handling.md). The `Markdown Input Field` is an obvious subject to this proposal, so we should expect some changes to the value of the input soon.

Second, we should talk about how we want to build our JS very soon. I.e., which patterns and principles we want to use. To the best of our current knowledge, the "action-model-projection"-loop from the  `Main Bar` ("functional reactive programming") is the best approach for building maintainable and testable frontends. Also, we could harness the possibilities of OOP more then we currently do (mostly). The code in this PR is somewhere in the middle between these principles and the old school "react-to-events"-style: Some classes, some separation from the creation of actions and the application of actions. We should get a shared understaning where we want to head here and document that understanding.

Third, this PR again shows that we need some connection between the ilCtrl-machinery and the UI framework some time soon. This connection is currently in `Services/UI`, which feels rather ad-hoc and lacks a maintainer. Nice to have you, as ilCtrl-maintainer, on board of the UI framework as well.

- [ ] Lang: Are you sure that we can change [this var] in the lang-file? I would suspect that we break some other usage. Same for header, bullet, ... and english variants.
- [ ] Attributes: Why do you pass the `minlength` and `maxlength` attributes via Javascript and not via the static rendering? We could even read that attributes from the element in the JS-code. Using HTML to pass attributes seems to be a lot cleaner than passing them via JS since we will get complete and readable HTML source then.
- [ ] Preview 1: What happens (visually...) if I click preview but the rendered text isn't returned from the backend fast enough?
- [ ] Preview 2: Could you ponder a little why we use the backend for rendering the preview instead of using some client side implementation? 

- [ ] Interface: I'm most sure that [Markdown::getMarkdownRenderer on the public interface](https://github.com/ILIAS-eLearning/ILIAS/pull/5152/files#diff-0ebacdb9da9acf01c6ec1eec1ead06e0f856e06414b65bac1bf549887a32ea38R25) doesn't need to be public.
- [ ] Un-Div: We would suggest to get rid of [this additional div](7https://github.com/ILIAS-eLearning/ILIAS/pull/5152/files#diff-0b444e7eb7d44eff993bab9549ceedeb5928e93f602a24995b3ce2bed4f1ef01R13). We try to use as little divs as possible, this one does not seem to be really necessary.
- [ ] Remainder: We would suggest to use some other class then `help-block` [for the info about the remainder](https://github.com/ILIAS-eLearning/ILIAS/pull/5152/files#diff-2926fea19e1429455978113bbbde49a60e40da725c947df79da13b08a95b34f3R4). This is the same class that is used for the byline, and the remainder seems to be something different.

- [ ] Copyright: Make the copyright-checks pass.
- [ ] LESS: Please turn the less-code into [sass code](https://github.com/ILIAS-eLearning/ILIAS/pull/5102). Feel free to ask @Amstutz for help, also regarding your font-question.

If there is anything we can help you with, feel free to contact us on short notice.

Best regards!
@Amstutz and @klees

thibsy:
Hi @klees 

It took a while, but I finally managed to go over this again. Thanks for the review btw!

- [x] Lang: I adjusted the glyphs used in this PR to use another naming convention (`*_action`), to indicate that these glyphs will do something if clicked. The language variables for the bullet-list and numbered-list were missing altogether, so I applied it to them as well.
- [x] Attributes: Done. This lead to a new issue with the JavaScript unit tests though, because apparently `JSDom` uses the wrong default value for `HTMLTextAreaElement.maxLenght` (see https://github.com/jsdom/jsdom/issues/2927). But, I managed to work around it for now.
- [ ] Preview 1: Nothing yet. The HTML will simply appear after a while.  I agree that it would make sense to indicate that some sort of loading is happening, but for a first version this will IMO suffice. If not: we could for starters implement an additional method like `getLoadingMessage()` on `MarkdownRenderer`.
- [x] Preview 2: We definitely want to have a backend library for transforming markdown-text into HTML, because when e.g. storing input of this field in the database to display it later on, we would like to already transform this text before sending our response. If we would use an additional library for dynamically  rendering HTML on clientside (like this input does), we'd have no guarantee that the two libraries produce the same HTML. As this PR also shows, the frontend can easily access the backend library, whereas the backend could not rely on the frontend library.

- [x] Interface: Done. You're right.
- [x] Un-Div: Done. I checked again and `ui-input-markdown-controls` _seems_ unnecessary, maybe someone with more better CSS-skills would like to have a look here =).
- [x] Remainder: I introduced a new class `ui-input-textarea-remainder`, which for now has similar properties as the `help-block`.

- [ ] Copyright: Tbh, I don't know why the copyright-check doesn't work in this case. Anyways, it's merely the examples which violate it.
- [x] LESS: Done. Hope this is correct.

Kind regards!
@thibsy

catenglaender:
Nice! That's a really useful little editor. :slightly_smiling_face: :clap: 

@klees asked me to have a look at the ITCSS/SCSS/CSS aspect of this PR and so far I noticed the following things:

* [x] Naming convention of CSS class names should follow BEMIT (see here: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/templates/Guidelines_SCSS-Coding.md#bemit-basics). So `ui-input-markdown-preview` would be `c-input-markdown__preview`. These rules are quite new and it will probably look a bit odd with all the legacy names next to it, but this is how CSS classes should be named in trunk now.
* [x] I think setting the `.ui-input-markdown-controls` to `display: inline-block` and then counteracting that with width and height settings is not a good practice, unless I overlook the reason why this has been done. The div's default `display: block` behaviour would already give you a full-width container and probably respect the btns height. So instead of defining width, height and display, I suspect just setting a margin-bottom with a variable from the layout layer should suffice.
* [x] I'm not the biggest fan of setting the children inside `.ui-input-markdown-controls` to float. I would suggest `display: flex;
    justify-content: space-between;   flex-wrap: wrap;` on it to pull the two children to opposite ends. When they stack on mobile they are both back to being aligned to the left without the need for any additional code.

Here is something we might want to address in the CSS Squad, but it can probably stay how it is in this PR
* Do we keep utility class ".hidden" or do we recommend using mixins into component specific classes like `c-input-markdown__preview--hidden`?

thibsy:
Thx @catenglaender for looking into the styling aspects of the PR :).

I'll work them in after receiving @klees second review, so I can implement them all at once.

Kind regards,
@thibsy

catenglaender:
Some more things I noticed, hope that's okay :smile: 

* [x] `.il-viewcontrol-mode` selects all viewcontrol mode buttons regardless where in ILIAS. It should probably be `.c-input-markdown .il-viewcontrol-mode`... if it still needs to be targeted at all.
* [x] The buttons in `ui-input-markdown-actions` do not have a bottom (or top) margin. This is problematic on smaller screens, when they stack and have no vertical spacing between them. However, the viewcontrol mode has the exact same problem. @Amstutz @klees Do we need a UI component button group that takes care of things like this? It's very common and not at all trivial design problem (if every single button has a margin-bottom, does the container need one too?).
* [x] The !important attribute could probably be avoided by using higher specificity taking `.ui-input-markdown textarea` to sth like `.ui-input-markdown .ui-input-textarea textarea`. Weirdly, the text is monospace for me (in Chromium) even without !important.

Alright, enough nitpicking the day :roll_eyes: Feel free to write in Discord if you have any questions.

chfsx:
Hi @catenglaender thanks for your point, I addressed two of them in https://github.com/ILIAS-eLearning/ILIAS/pull/5152/commits/ca047a674063d75f58929e1a051b90e6d4078acb, https://github.com/ILIAS-eLearning/ILIAS/pull/5152/commits/2ea6b508005817b0345c84203c6504bec4fe1772 and https://github.com/ILIAS-eLearning/ILIAS/pull/5152/commits/54b525deeb340dc226c3797acbeb8492d99820c5

the third one could be addressed in general and not for the markdown input only

chfsx:
@klees from our point of view, this should now be ready and we are looking forward to implementing it in https://docu.ilias.de/goto_docu_wiki_wpage_7404_1357.html after the merge.

klees (approved):

Amstutz:
Hi @thibsy

Thx a lot for this large contribution. We highly appreciate it and believe to be of excellent quality. We are looking forward to it being used in upcoming features and refactorings.

The PR is merged. However, for future generation of unit tests, one thing:

- Unit Tests: Provide tests of possible in a version without testing the fully rendered version each time. This tends to be error-prone and rather expensive in maintenance. It's good to test things as well for correct rendering, but note, that for those instances you can/should only test the rendering of the part of interest in e.g. using str_contains or similar. One complete rendering test is good to have, but for the rest, only test things of interest. 

Thx a lot.

klees:
:champagne: Love it!

PR #5153, opened 2022-10-25
Bugfix 32903 add Permanent Link in Media Pool

sasfalk:
#bugfix
Mantis issue: 0032903
Target: release_7
https://mantis.ilias.de/view.php?id=32903


alex40724:
@sasfalk Thanks for the PR!

PR #5154, opened 2022-10-25
Bugfix 32904 add Permanent Link in Exercise

sasfalk:
#bugfix
Mantis issue: 0032904
Target: release_7
https://mantis.ilias.de/view.php?id=32904


alex40724:
Thanks for the PR. Fixed slightly different.

PR #5155, opened 2022-10-25
Bugfix 0031264 data collection: the visibility-setting of the fields isn't copied

sasfalk:
#bugfix
Mantis issue: 0031264 / 0033859
Target: release_7
https://mantis.ilias.de/view.php?id=31264
https://mantis.ilias.de/view.php?id=33859


Amstutz:
Thx a lot. Sry, I missed this PR, since I only starting working on the DC some days ago. Therefore I fixed this already for trunk and 8 in the beta testing. So I will not pick this to trunk and 8. However, this ist not yet fixed for 7, so this is highly appreciated!

PR #5156, opened 2022-10-25
fixed 35042: Failed test: View shared items

alex40724:
See https://mantis.ilias.de/view.php?id=35042


chfsx:
Thanks a lot, cherry picker to trunk as well

PR #5159, opened 2022-10-26
VirusScanner: Allow _getInstace to Return Null

kergomard:
I think this needs to be as trivial as this, ...for the time being.


kergomard:
This has been fixed separately.

PR #5160, opened 2022-10-26
[FIX] 0034710: Failed test: Abbruch des Hochladens

chfsx:
This PR adds a progress indicator to the fileupload (for each file), as decided in https://mantis.ilias.de/view.php?id=34710

please note that there is another change with this PR:

dropped files are currently rendered with a dashed border (which indicates a dropzone, which is wrong). I introduced some less variables for this border style and hide it completely with this PR.

the fileupload now looks like this (with the progress indicator):

https://user-images.githubusercontent.com/6661332/197991104-ac9f9cb5-3a16-4d14-8506-ec28f15cba8b.mov

### DIFF
--- a/Services/Init/classes/Dependencies/InitResourceStorage.php
+++ b/Services/Init/classes/Dependencies/InitResourceStorage.php
@@ -18,13 +18,12 @@
 
 use ILIAS\DI\Container;
 use ILIAS\FileUpload\Location;
-use ILIAS\ResourceStorage\Consumer\InlineSrcBuilder;
 use ILIAS\ResourceStorage\Consumer\SrcBuilder;
+use ILIAS\ResourceStorage\Consumer\StreamAccess\StreamAccess;
 use ILIAS\ResourceStorage\Information\Repository\InformationDBRepository;
 use ILIAS\ResourceStorage\Lock\LockHandler;
 use ILIAS\ResourceStorage\Lock\LockHandlerilDB;
 use ILIAS\ResourceStorage\Policy\FileNamePolicy;
-use ILIAS\ResourceStorage\Policy\FileNamePolicyStack;
 use ILIAS\ResourceStorage\Policy\NoneFileNamePolicy;
 use ILIAS\ResourceStorage\Preloader\DBRepositoryPreloader;
 use ILIAS\ResourceStorage\Repositories;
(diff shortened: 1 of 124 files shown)


chfsx:
@Amstutz sorry for the force pushed commits, now the PR is ready for review

Amstutz:
Hi @chfsx

Thx a lot for extending this. Since this is a minor change with a JF decision already done, we can "Fast-Track" this, meaning merge and announce the public interface change later.

- [x] jQuery in JS: We try to minimize jQuery in JS as much as possible. Do you have a strong reason to use it here?
- [x] hasChunkedUpload vs supportsChunkedUploads: We do not understand, why we need both. Are they really both needed? Also, as far as we can see, only supports is used in the implementation, right? If so, please document them and also add a line concerning the difference between them. If not, remove one.

Please change:

- [x] Documentation: Please add some comment to document supportsChunkedUploads. What is it, why would I use it? Maybe even add an example-code snippet (up to you). Keep in mind, that if you add an example, you also need to change the corresponding testcase in ILIAS 8. I would personally perform the test then.

Thx a lot.

chfsx:
Hi @Amstutz 

Thanks for the review and reply!

> jQuery in JS

I understand the concern very well, in this case I used more jQuery functions because big parts of file.js are already built on it and it's better to replace this in one complete step and not change single parts to plain JS.

> hasChunkedUpload vs supportsChunkedUploads:

Yes, `hasChunkedUpload` is redudant and only `supportsChunkedUploads` is relevant. I have adjusted this accordingly and added a comment.

Perhaps in future there will be a `hasChunkedUpload` and `withChunkedUpload` at the UI\Input\File, so that one can decide for oneself whether a ChunkedUpload should be possible at the place used.

Currently, there's no place in ILIAS with ChunkedUpload. But there will come a lot of them: https://docu.ilias.de/goto.php?target=wiki_1357_Customizable_Maximum_File_Upload_Size

An example and a implemented UploadHandler with Chunk-support will come with this.

chfsx:
Rebased and force-pushed to not loose mergeability...

Amstutz:
Merged, please check if some test case in testrail needs updating an inform me to perform a (re)test of it.

Amstutz:
PR Merged, however, I will need to pick it manually and merge conflicts due to the change to ITCSS in trunk. This will take yet a few days. If urgent, you may create a seperate PR for trunk using the ITCSS structure for your CSS changes.

PR #5162, opened 2022-10-26
[FEATURE] UI/GS: Applied OpenGraph Metadata

thibsy:
Hi all,

I've implemented the feature-request (https://docu.ilias.de/goto_docu_wiki_wpage_6931_1357.html) with this PR. It applies the new and improved metadata from the data service to the global-screen and kitchensink. @chfsx I wasn't sure where exactly to handle duplicate-entries in case of `Html\UserDefined` metadata, but figured the `MetaContent` would be a good place. I think this can be avoided in the future if every available meta-tag (or at least some more) have been implemented, but for now I think this suffices. Duplicate entries weren't a problem until now, because the values were stored in an associative array anyways and the latter value just won. Now we have to handle this differently. IMO the kitchensink-part should be rather "dumb" and doesn't need to distinguish the two kinds of HTML-metadata. @Amstutz maybe you wan't to weigh in on this too?

@alex40724 I introduced a first "Exposer" for the OpenGraph protocol ([ogp.me](https://ogp.me)) in the repository service, because I thought it'll most likely be used there. The exposer currently handles all OpenGraph-metadata, but will deliver some more specific metadata according to the currently requested object.

@klees there were some small adjustments to the new `Html\UserDefined` and a further `Html\OpenGraph\TagCollection` class in the data service was needed, because I had to distinguish common HTML-metadata from OpenGraph-metadata. I also noticed that the type-hinting is rather special when handling OpenGraph-metadata, because to properly separate you have to check for both, `Html\OpenGraph\TagCollection` and `Html\OpenGraph\Tag`. It's also not entirely clear which one of these definitions you should use as a type-hint, but I guess that also depends on the case. What do you think?

Kind regards!


thibsy:
P.S. There's currently an issue with the OpenGraph-metadata that's exposed, because when scraping the "og"-tags the different social-media will send a (I think normal, but I haven't found specification on this) HTTP-request to the URL that's been shared. If ILIAS is initialized with the web-context the initialization will automatically redirect to the login-page. Funnily that's also the case when "public-area" has been activated and the shared object is visible for anonymous. That's a problem because the metadata-scraping will not be able to follow this request. Does anybody know how we could/should handle this?

Services/Repository/GlobalScreen/classes/RepositoryOpenGraphExposer.php
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+
+    protected function getDefaultImage(): OGImage
+    {
+        $image_path_resolver = new \ilImagePathResolver();
+
+        return $this->data->openGraphMetadata()->image(
+            $this->data->uri(
+                ILIAS_HTTP_PATH . ltrim(
```
chfsx, line 126: maybe ltrim `./` here and add a `/` again to be sure

chfsx, line 126: maybe `ltrim(..., "./")` here and add a `. "/".` to be sure here

src/GlobalScreen/Scope/Layout/MetaContent/MetaContent.php
```diff
@@ -116,7 +143,15 @@ public function getCss(): CssCollection
         return $this->css;
     }
 
-    public function getMetaData(): MetaDataCollection
+    public function getOpenGraphMetaData(): ?OpenGraph\TagCollection
```
chfsx, line 146: Do I understand it correct: OpenGraph will always be a TagCollection to avoid adding "normal" `MetaData`s here?

chfsx (approved):
Thanks a lot, the GS part looks good to me! 

The issue with the Initialisation is a bug IMO which has been here for a while, could you open a mantis issue for this? From my point of view it must be possible to open a ILIAS public repo without the redirect before.

src/GlobalScreen/Scope/Layout/MetaContent/MetaContent.php
thibsy: Correct, we need to distinguish OpenGraph metadata from common metadata at this level because the OpenGraph metadata would be invalid if it was added more than once. The `OpenGraph\TagCollection` was introduced to prevent `Html\TagCollection` from being added there.

klees (approved):
Hi @thibsy,

thanks a lot for improving this. I approve the changes but would kindly ask to remove [this loop](https://github.com/ILIAS-eLearning/ILIAS/pull/5162/files#diff-e091371a71733da679f4e602903e1f03191bfa56e8f5c1215e90de6f877dc325R113) by using monoid powers =)

Best regards!

thibsy:
@klees you're right =) 

I introduced the `Data\Factory` as a new dependency to the `AbstractComponentRenderer`, because I didn't wanted to create a "raw" `Html\TagCollection` instance in the `Page\Renderer`. This would've created an implicit dependance anyways. Therefore, I had to adjust the initialization of the renderer in several places.

I also noticed an "error" in the `InitUIFramework` where the closure for the `Field\Factory` created a new instance of the `Refinery\Factory` instead of using the existing one from the current container. The fix lead to a failed test case `ExamplesTest` because the `Field\Factory` has gotten a real instance of the `Refinery\Factory` instead of the mocked instance until now. I replaced the mocked instance by a real one, if that's too lazy let me know.

Since this is not just a minor tweak I re-requested the review from you.

Kind regards!

klees (approved):

klees:
Thx y'all!

PR #5163, opened 2022-10-26
SVG Files (.svg) for Milos Icon Set (version 26.10.2022)

yvseiler:
These are revised SVG Milos icons that will be replaced or added to the images folder (https://mantis.ilias.de/view.php?id=35320).


Amstutz:
Nice work, thx a lot!

PR #5165, opened 2022-10-27
Test: Fix calculation of starting time (Mantis 35353)

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=35353

This has to be cherry-picked to `trunk` as well.


dsstrassner:
hi @kergomard, 

 would you do me the favor? Thanks!

kergomard:
Thanks @mjansenDatabay ! I will close this though as I believe that this is not the right solution. I think the behavior before was, just by pure luck, the right one. I will fix this by removing the parameter from the method call completely.
If we pass the correct parameter the starting time will always be calculated from the last pass and this is not what we want, as it might also be limited by the duration since the first pass.
If anybody disagrees, please let me know.

I will fix in 7, 8 and trunk.

PR #5166, opened 2022-10-28
[PROPOSAL] CI: integrate node.js and automated js testing.

thibsy:
Hi @mjansenDatabay 

I think you're responsible for the automated-testing stuff, right?

I'm not too deep into github actions, but I tried adopting some configuration from https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs that might work and get automated javascript tests going =).

Since I'm quite a noob at CI-related stuff I also didn't know how I could test this easily, maybe you could help me there?
I also wasn't quite sure how this config-file should be structured, but I think they way I split the different steps is not entirely wrong.

Kind regards!

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -12,6 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         php: [8.1, 8.2]
+        nodejs: [ 19.x ]
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
@@ -26,11 +27,24 @@ jobs:
           tools: composer:v2
           coverage: none
 
-      - name: Install dependencies
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.nodejs }}
+
+      - name: Install Composer packages
         run: composer install --no-interaction --no-progress
+
+      - name: Install Node.js packages
+        run: npm install --ignore-scripts
 
       - name: PHP Unit Test
         run: CI/PHPUnit/run_tests.sh
+        env:
+          GHRUN: "yes"
+
+      - name: JS Unit Test
+        run: npm test
         env:
           GHRUN: "yes"
 


mjansenDatabay:
Hi @thibsy ,

> I think you're responsible for the automated-testing stuff, right?

No officially :-).

If I am right we will introduce a new dependency (at least for the GH actions) by using `actions/setup-node@v3`, so IMO we'll have to put this on a JourFixe agenda and present this to the community.

Best regards,
@mjansenDatabay

matthiaskunkel:
Jour Fixe, 19 DEC 2022: Thibeau notified us about the necessary use of node.js in our pipeline to support JS unit tests. A dependency already exists. Automated testing of JF should also support node.js v. 18 (as it might be the most used version at the time being). Thibeau asks for support by another developer who is more familiar with CI to have a look at the PR and comment it in case of possible improvements.

klees:
@thibsy The Technical Board regularly reviews PRs to identify road blocks in our processes. Is there anything we can do here to help you finish this?

Kind regards!

thibsy:
Hi @klees

Thanks for the ping. I rebased this onto the latest trunk and adopted the requested change by @mjansenDatabay and added Node.js v18.x to the workflow matrix.

However, this lead as shown in this PR to doubling the amount of checks which are run for every PR, because according to the workflow strategy, every PHP version is checked with every Node.js version. Now IMO this is obsolete, because PHP and Node.js are in no correlation. I have already asked on Discord, how to move forward with this. As I suggested we could split up the workflow into two, PHP- and JavaScript specific workflows.

Now, having automated JavaScript unit-tests is definitely something important, and I don't want to delay this any longer. Therefore my question: can we live with doubling the amount of checks for now? If not, can I simply remove the v18.x from the matrix to speed up this PR and maybe suggest splitting up the workflows in another PR, so we can have a JavaScript unit-test check _now_?

Kind regards,
@thibsy

thibsy:
Hi all,

As briefly discussed with @klees it makes sense to work in iterations here, so we can have automated JavaScript unit-tests _now_. It also makes sense to split up the workflows as I suggested, but this will happen in another PR.

For the time being, I have removed the Node.js version 18.x from the matrix so we can keep the amount of checks the same. Since we are also using the latest version of ECMAScript for ESLint, I don't think this will be an issue to only work with version 19.x.

I allowed myself to merge this PR, since it has already been accepted by JF some time ago.

Kind regards,
@thibsy

PR #5167, opened 2022-10-28
TA: 15673, save question after removing variables/results from questiontext

nhaagen:
https://mantis.ilias.de/view.php?id=15673


kergomard:
This looks good, thank you very much!

PR #5168, opened 2022-10-28
TA, 34659: rename some vars for better understanding

nhaagen:
https://mantis.ilias.de/view.php?id=34659


kergomard:
Thank you very much @nhaagen . This definitely makes things more understandable!

PR #5169, opened 2022-10-28
T&A 35100 review of scored answers causes error 

kunzeus:
This fixes the issue of: https://mantis.ilias.de/view.php?id=35100 
Also, some code cleanup

And again let me know about things I can improve.

### DIFF
--- a/Modules/Test/classes/class.ilTestEvaluationGUI.php
+++ b/Modules/Test/classes/class.ilTestEvaluationGUI.php
@@ -1121,7 +1121,7 @@ class ilTestEvaluationGUI extends ilTestServiceGUI
         $passOverViewTableGUI->setResultPresentationEnabled(true);
         $passOverViewTableGUI->setPassDetailsCommand('outParticipantsPassDetails');
         $passOverViewTableGUI->init();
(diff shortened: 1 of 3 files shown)


kergomard:
Thank you very much for this fix @kunzeus !

I would only have three little change requests, other than that this looks good to me:

- [ ] Thank you for correcting the doc-block [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5169/files#diff-2f0e8d5eafb6a326db0c3b8b1d6ef13629055f8d4ecddedfbc36bd5223180d6fR184). Could you please quickly specify the expected parameter types, too? This way we don't miss the opportunity to specify what we actually want.
- [ ] Could you please change the array-declaration [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5169/files#diff-2f0e8d5eafb6a326db0c3b8b1d6ef13629055f8d4ecddedfbc36bd5223180d6fR208) to short syntax.
- [ ] Could you please type the $isScoredPass of ilTestPassOverviewTableGUI::getRequiredActions() with a nullable Boolean. This would give us a better handle on it. I think you are right to set it to null, though, even if this all has a smell to it. ...or am I missing something?

Then we are ready and I will merge,
@kergomard

kergomard:
Thanks a bunch @kunzeus !

PR #5170, opened 2022-10-31
Info: IRSS Flavours

chfsx:
Flavours are derivatives of resources that can be created, managed and used alongside the original resource. A simple example are different resolutions of images: If a person uploads e.g. a profile picture with 20MB size, the IRSS can generate several flavours from it, e.g. a 500x500px flavour in an adapted quality, which can be used for the representation in a member gallery, without the full resolution with 20MB having to be loaded and displayed in the browser.
Similar approaches exist in ILIAS in many places, but always the components themselves must ensure that the desired "derivatives" are available, stored and can be retrieved. With the Flavours the IRSS offers this centrally.

Please see the [documentation](https://github.com/srsolutionsag/ILIAS/blob/6eb081047a5fc180b9998c64307a3881851bd874/src/ResourceStorage/README.md#resource-flavours) and ask me for infos/additions if needed.

Flavours are used as part of a refactoring on ilPreview. ilPreview is a service that was originally intended to provide preview images for various object types. Currently, however, it only works for file objects and is also very outdated.

ilPreview requires installed binaries and uses exec internally, which is to be abolished. 
In a separate [PR](https://github.com/ILIAS-eLearning/ILIAS/pull/5172), the default for PHP imagick instead of direct binaries is proposed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,4 @@ phpstan-baseline.neon
 
 # File Delivery
 override.php
+Services/ResourceStorage/artifacts/flavour_data.php
--- a/Services/Init/classes/Dependencies/InitResourceStorage.php
+++ b/Services/Init/classes/Dependencies/InitResourceStorage.php
@@ -18,6 +18,7 @@
 
 use ILIAS\DI\Container;
 use ILIAS\FileUpload\Location;
+use ILIAS\ResourceStorage\Artifacts;
 use ILIAS\ResourceStorage\Consumer\SrcBuilder;
 use ILIAS\ResourceStorage\Consumer\StreamAccess\StreamAccess;
 use ILIAS\ResourceStorage\Information\Repository\InformationDBRepository;
@@ -28,6 +29,7 @@ use ILIAS\ResourceStorage\Policy\NoneFileNamePolicy;
 use ILIAS\ResourceStorage\Preloader\DBRepositoryPreloader;
 use ILIAS\ResourceStorage\Repositories;
 use ILIAS\ResourceStorage\Resource\Repository\CollectionDBRepository;
+use ILIAS\ResourceStorage\Resource\Repository\FlavourDBRepository;
 use ILIAS\ResourceStorage\Resource\Repository\ResourceDBRepository;
 use ILIAS\ResourceStorage\Resource\ResourceBuilder;
 use ILIAS\ResourceStorage\Revision\Repository\RevisionDBRepository;
@@ -50,6 +52,7 @@ class InitResourceStorage
     public const D_LOCK_HANDLER = self::D_SERVICE . '.lock_handler';
     public const D_FILENAME_POLICY = self::D_SERVICE . '.filename_policy';
     public const D_STREAM_ACCESS = self::D_SERVICE . '.stream_access';
+    public const D_ARTIFACTS = self::D_SERVICE . '.artifacts';
 
     /**
      * @internal Do not use this in your code. This is only for the DIC to load the Resource Storage
@@ -80,7 +83,8 @@ class InitResourceStorage
                 new ResourceDBRepository($c->database()),
                 new CollectionDBRepository($c->database()),
                 new InformationDBRepository($c->database()),
-                new StakeholderDBRepository($c->database())
+                new StakeholderDBRepository($c->database()),
+                new FlavourDBRepository($c->database()),
             );
         };
 
@@ -118,6 +122,17 @@ class InitResourceStorage
             return new NoneFileNamePolicy();
         };
 
+        // Artifacts
+        $c[self::D_ARTIFACTS] = static function (Container $c): Artifacts {
+            $flavour_data = is_readable(ilResourceStorageFlavourArtifact::PATH) ?
+                include_once ilResourceStorageFlavourArtifact::PATH
+                : [];
+            return new Artifacts(
+                $flavour_data['machines'] ?? [],
+                $flavour_data['definitions'] ?? []
+            );
+        };
+
         // Stream Access for Consumers and internal Usage
         $c[self::D_STREAM_ACCESS] = static function (Container $c) use ($base_dir): StreamAccess {
             return new StreamAccess(
@@ -133,6 +148,7 @@ class InitResourceStorage
             return new \ILIAS\ResourceStorage\Services(
                 $c[self::D_STORAGE_HANDLERS],
                 $c[self::D_REPOSITORIES],
+                $c[self::D_ARTIFACTS],
                 $c[self::D_LOCK_HANDLER],
                 $c[self::D_FILENAME_POLICY],
                 $c[self::D_STREAM_ACCESS],
--- a/Services/Init/classes/Dependencies/InitResourceStorage.php
+++ b/Services/Init/classes/Dependencies/InitResourceStorage.php
@@ -111,7 +111,7 @@ class InitResourceStorage
 
         // Source Builder for Consumers
         $c[self::D_SOURCE_BUILDER] = static function (Container $c): ?SrcBuilder {
-            return null;
+            return new ilWACSrcBuilder();
         };
 
         // Filename Policy for Consumers
--- a/Services/ResourceStorage/classes/FlavourDBRepository.php
+++ b/Services/ResourceStorage/classes/FlavourDBRepository.php
@@ -1,0 +1,124 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\ResourceStorage\Resource\Repository;
+
+use ILIAS\ResourceStorage\Flavour\Definition\FlavourDefinition;
+use ILIAS\ResourceStorage\Flavour\Flavour;
+use ILIAS\ResourceStorage\Flavour\FlavourIdentification;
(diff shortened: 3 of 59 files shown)


matthiaskunkel:
I really appreciate this suggestion. It could help us to solve the problem that users often upload much too big pictures for the tile view - which becomes a performance problem in categories with a lot of tiles to be presented.

klees (commented):
Hi @chfsx,

this sounds like a super useful improvement of the fileservice. Thanks!

I haven't looked through this in detail (I hope I have time to do so...), but I just noticed one minor point:

* Could we reserve [`withXY` for immutable objects](https://github.com/ILIAS-eLearning/ILIAS/pull/5170/files#diff-874b178995a4dcae6e329b4a0aed19fc673ae81f6df6f0312d5b66c6cc0b2f7cR83)? This would allow our developers to form some expectation towards object behaviour by the naming of methods.

Thanks for considering =)

matthiaskunkel:
Jour Fixe, 31 OCT 2022 : We highly appreciate this suggestion and accept the PR for trunk. Fabian will wait another 2 weeks with the merge to give some time to other developers for feedback and change requests.

klees:
Hi @chfsx,

so now I had some time to look into this a little. Following things came to mind:

* Could we just rename the mechanism to `Derivatives`? The first sentence reads: "Flavours are derivatives of resources", so maybe it's just more self explanatory to use the explanation as a name. With the name `Flavour` I always wonder: Is this about the flavour in general ("sweet sour") or about the thing having the flavour ("Chicken with sweet sour sauce"). I think the name `FlavourDefinition` shows this problem in this context as well. The "flavour definition" would be "cropped to 100px X 100px" and the "flavour" would be "this specific image cropped to 100px X 100px", right? Sounds a a bit odd to me, so I would suggest looking for other names.
* I'm not super deep into the actual usage of the IRSS right now, but as far as I understand the state of affairs, I would try to interact with [`ILIAS\ResourceStorage\Consume\Consumers\`](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/ResourceStorage/Consumer/Consumers.php#L33) to get hold of my resources in a form suited for further processing. If I had a Flavour of my resource, how would I allow users to e.g. download it? Can I use a Flavour with that `Consumers` class?
* Why is a Flavour categorically different from other resources? In many regards I would not care if I had the "original" resource or a derivative thereof. E.g. when including an image into some HTML, it does not matter if it's the original image or some cropped version (right?). If original resources and flavours share the same type of Ids, many interfaces (such as `Consumers`) could be reused and complexity could be hidden in many places.
* What would I do with [`Flavour::clearNonStorageStreams`](https://github.com/ILIAS-eLearning/ILIAS/pull/5170/files#diff-874b178995a4dcae6e329b4a0aed19fc673ae81f6df6f0312d5b66c6cc0b2f7cR66) and [`Flavour::clearNonInMemoryStreams`](https://github.com/ILIAS-eLearning/ILIAS/pull/5170/files#diff-874b178995a4dcae6e329b4a0aed19fc673ae81f6df6f0312d5b66c6cc0b2f7cR74)?
* What happens if I request a Flavour for a resource that does not support the thing I request? E.g. crop an mp3 to 100x100px?
* Why do we need to abstract these `Engine`s? The machines currently use calls to functions of php extensions internally anyway, and the engines are not in fact used by the machines (or I'm missing something...).
* Why can there be machines without any engine, i.e. `withEngine` might never be called. If we need `Engine`s, this looks like a typical case for dependency injection: I cannot have a machine if I do not have an engine that fits...
* Why is the definition separated from the machine that is used to execute it? This could also be a case for dependency injection and a factory for the flavour definitions.

There might be more at some point, but I hope this feedback already helps a bit.

I really appreciate the direction this is taking. I also think we should look some more into how this will be used by others. The IRSS will be used by many, for the core and for plugins. It has the potential to enhance ease of use and security for many use cases at the same time. Every bit of effort we put into its programmatic interface will pay of a lot more.

Thanks for working on this!

chfsx:
Hi @klees

Thanks for your short review and your tips and questions, gladly I go directly to it:

> Could we reserve `withXY` for immutable objects? This would allow our developers to form some expectation towards object behaviour by the naming of methods.

Yes definitely, makes sense if we try to keep this constant, even if for me linguistically `withXY` sometimes seems more appropriate. I'll change this in one of the next commits

> Could we just rename the mechanism to `Derivatives`?

Short answer: I'm pretty sure i'll leave it at `Flavours`, unless there's a sudden 100% vote against from everyone here. Since you like to have arguments, here are my arguments for `Flavours` :-)

1. in software development we often use analogies or metaphors (correct would be "Bilder") for naming things. Of course these have to be chosen carefully in order to hit the point. `Flavours` as well as derivatives are such, one draws from the world of food or tastes, the other from linguistics, economics or chemistry. for me, derivatives have too broad an area of application across many topics and is not unique enough for this, even if it is an appropriate term.
1. The term `Flavour` nevertheless characterizes the matter for me: we have a basic product, from which different `Flavours` are offered. Let's take ice cream: We have a basic product on which we make different `Flavours`. But all of them are ice cream varieties: Sometimes because we add strawberries, sometimes chocolate pieces. Sometimes in a large cup, sometimes in a small one (not a good example, I know). Sometimes with color, sometimes not. I go to the ice cream store and want an ice cream. I get a `Flavour` of it.
1. Think a bit into the future: `Flavours` don't always have to be reductions of resources, don't always have to be vaporized, but can also be simply converted: Especially in the area of video, this will become very important, e.g. that consistency all uploaded videos are converted into mp4.
1. `Flavour` is my personal favorite: Software development has to be fun, I prefer to think of `Flavours` (and maybe even get hungry) than to think of derivatives, where stocks and obligations or chemical reactions come to my mind. Boring.
1. Other projects use the term partly for the same thing (see [Opencast](https://docs.opencast.org/r/9.x/admin/#workflowoperationhandlers/encode-woh/)) or in a similar way, but also to distinguish themselves from "distribution" and "version", at [Ubuntu](https://ubuntu.com/desktop/flavours).

One could say that some of the arguments apply equally to "derivatives", which is quite correct. However, this also shows me that both terms are equally appropriate or inappropriate, which leads to my argument 4: My personal favorite :-)

We will talk about these things again and again in the future, so I think it is better to have a relatively unique term for it and not a (very technical) one that we might also use for other things. A term that would also fit, for example, would be "artifact", but as is well known, this term has already been used.

> I would try to interact with `ILIAS\ResourceStorage\Consume\Consumers\`

Exactly, you have already found the right place for it, a method for downloading the `Flavour` is not there yet, but can be added very quickly if needed. In the current version it is only about flavours that are to be displayed in the browser (images). For this you can get pre-signed URLs for delivery via the consumer. Downloading a `Flavour` would be a simple extension. You have to check whether, for example, `Flavours` with more than one stream (such as thumbnails of a pdf) are downloaded as a zip or whether you can also simply get individual streams of them, we will probably want both.

> Why is a `Flavour` categorically different from other resources?

At the very beginning of the implementation, we also considered implementing `Flavours` as `Resources`, but quickly came away from it: `Flavours` belong directly to the `Resource` (better: directly to a`Revision` of a `Resource`) and both together can be understood as a package. If `Flavours` were their own `Resources`, it quickly becomes complicated: `Resources` should be reusable, i.e. different `Stakeholders` should be able to use them. Let's assume a `Resource` has a `Flavour`, but this `Flavour` is then again a `Resource` from the point of view of another `Stakeholder`. The other `Stakeholder` makes a new `Revision` of the `Flavour` (because it is a `Resource` for him), so the `Flavour` would be wrong for the first `Resource`. Also from the point of view of the `StorageHandler` it makes more sense to store the `Flavours` directly with the `Resource`, this makes it easier to move `Resources` between different `StorageHandlers` (what we have already done in ILIAS 7 for example). `StrageHandlers` also have cleanup processes, e.g. to delete empty directories in the filesystem, which is much easier and more reliable for the `StorageHandler`, if all files of a `Resource` are at the same place. 

> What would I do with `Flavour::clearNonStorageStreams` and `Flavour::clearNonInMemoryStreams`?

Good point, I need to internalize these, they don't belong in the public interface of a `Flavour`.

> What happens if I request a Flavour for a resource that does not support the thing I request?

You get the `Flavour` with zero streams for it.

> Why do we need to abstract these `Engines`?

The engine is used by the machine to check if e.g. the file type is supported or if the engine is present at all.

> Why can there be machines without any engine, i.e. withEngine might never be called.

`Machines` can certainly do something even if the engine is not present, for example a machine could return a "broken image" as a stream if the `Engine` is not present and nevertheless a `Flavour` is requestes that uses this `Machine`.

> Why is the definition separated from the machine that is used to execute it?

You can write arbitrary custom `Machines` and `Definitions`, the IRSS just comes with a default set of `Machines` and `Definitions` for cases we probably use most often. A `Machine` can in principle also support different `Definitions` and not just one. 

> I also think we should look some more into how this will be used by others.

That is the most important point for me. The public interface for `Flavours` has to be small and simple, as a consumer I don't have to and shouldn't have to worry about the internals. Also as a consumer I shouldn't have to worry about which engines are there (Except, of course, that they work the way they're supposed to and I'm enlightened enough by documentation as to how that is.), which settings are set etc. I request a `Flavour` and get one, I can use the existing streams and use.

Amstutz:
Hi there,

I just read through the documentation and browsed the discussion here. Thx a lot @chfsx , looks very promising.

My 5 cents for the flavor discussion, however, no strong opinion here:

- Flavor: Works best for me if it is intended, that at some point, also additional information could be added. In the Ice cream example, you have the base ingredients, in every ice cream, such as the cream (here this would the Resource Identification) and then add additional stuff to it, like strawberry or vanilla (here this could maybe be some mask for the image or similar). 
- Derivate: Works best for me, if the result is some kind of reduction of the initial data

The distinction seems not to be clear, since in base cases, you have data, apply a function with additional data and up with some transformed kind of data. However, the "flavor" has a more "extending" character to me, "derivate" more something of a "reducing" character. I can perfectly well live with both here....

chfsx:
Hi @Amstutz 

Thanks for your 5 cents! It's definitely the first thing you say: Different resolutions (even bigger ones!) of images, grayscale version, masked images, different video formats, preview images from videos, inverted SVGs, you name it.

with each piece of our discussion here, I can then also complete/add to the README to clarify the use cases. Zhank you!

chfsx:
Hi!
If there are any further feedbacks I'll merge the PR next week, thanks a lot for your inputs so far!

PR #5173, opened 2022-10-31
TA: 31815, apply rounding lately and calculate w/o precision

nhaagen:
https://mantis.ilias.de/view.php?id=31815

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assFormulaQuestionResult.php
+++ b/Modules/TestQuestionPool/classes/class.assFormulaQuestionResult.php
@@ -260,33 +260,25 @@ class assFormulaQuestionResult
         $math = new EvalMath();
         $math->suppress_errors = true;
         $result = $math->evaluate($formula); // baseunit-result!!
-
-        $resultWithRespectedUnit = ilMath::_mul($result, 1, $this->getPrecision());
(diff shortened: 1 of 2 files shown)


kergomard:
Thank you very much @nhaagen . I think I was able to wrap my head around it. For me this looks less error prone, yes.

One question: Why do we need the division by one on [line 314](https://github.com/ILIAS-eLearning/ILIAS/pull/5173/files#diff-b8b4411675745afd110e35777522dc8a3ba491c2581e8bcc3eb8974f70bd3ca4R314)?

nhaagen:
I guess you're right - we do not need this.

nhaagen:
I want to revise my last comment - apparently, there is is some normalization going on via the MathAdapter; I did not dig deep enough to fully comprehend consequences, but it feels saver to keep this.
e.g., look here: https://github.com/ILIAS-eLearning/ILIAS/blob/be28324af208016f19d63d6a9cc810e0128614dc/Services/Math/classes/class.ilMathBaseAdapter.php#L62

kergomard:
You are right. I don't like it (I think we agree on that), but you are definitely right. I merge and cherry-pick. Thanks!

PR #5174, opened 2022-10-31
TA: 26113, create only one file per user-result in archive

nhaagen:
https://mantis.ilias.de/view.php?id=26113


kergomard:
I've talked with @dsstrassner today: I would kindly ask you to go ahead and remove the "_v<int>" part completely. I will then merge. Thank you very much!

kergomard:
Thank you very much @nhaagen ! Will merge and cherry-pick.

PR #5175, opened 2022-10-31
TA: 30528, add title modifiers to kioskview

nhaagen:
https://mantis.ilias.de/view.php?id=30528


kergomard:

PR #5177, opened 2022-11-02
Release 8 iconremove: Deletes unneeded icon

yvseiler:
Fix for Mantis #35352 Deletes an icon that came in due to a misunderstanding during the last icon revision https://mantis.ilias.de/view.php?id=35352


Amstutz:
Thx a lot for housekeeping work here!

PR #5178, opened 2022-11-02
Prevent non-numeric inputs within the client

iszmais:
https://mantis.ilias.de/view.php?id=35267

This would be better solved with a whole refactoring of the ilAssClozeTestCombinationVariantsInputGUI since it has major issues with the verification of Inputs in general and is isolated from common input checks.

Due to time issues i would still recommend this PR as a temporary solution.

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assClozeGapCombination.php
+++ b/Modules/TestQuestionPool/classes/class.assClozeGapCombination.php
@@ -116,7 +116,7 @@ class assClozeGapCombination
                             $gap_combinations['select'][$i][$k],
                             $j,
                             $gap_values[$i][$j][$k],
-                            $gap_combinations['points'][$i][$j],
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @iszmais for this PR.
You are completely right, that this is actually only going to be solved by a real refactoring. I would still kindly ask you to implement another solution, aligned with what we do in other places : Just force cast the input to float. This will truncate it at the ','. And we are done with it ;-).

iszmais:
@kergomard Thanks for the response!

This approach may not solve the whole problem since the invalid input isnt limited to number conventions. You can also input string ATM wich results in the same issue.
I could of course also force them into floats but the result may be unsatisfying.
But i would be ok with that as an temporary solution to prevent hard error.

Pls give me your feedback on that.

Greetings,
@iszmais

kergomard:
You are right and I know it makes the heart bleed to solve problems this way, but still: Please do simply cast. We use this solution in different places and we should actually solve the whole points thing in a structured way.

kergomard:
Thanks @iszmais .

PR #5172, opened 2022-10-31
Replace some dependencies with PHP extensions

chfsx:
ILIAS uses `exec` internally for many operations, e.g. for zipping and unzipping archives or for generating different resolutions of images (`convert`). 

This PR generally suggests to replace these with appropriate PHP extensions:

- zip, unzip > php-zip: #5181
- imagick > php-imagick: #6015

corresponding implementations must follow if this is desired. i strongly recommend that we get rid of as many to all exec calls as possible. besides obvious reasons, we can also better test and reuse such code, for example.

PR #5181, opened 2022-11-04
Proposal/9/zip unzip replacement

chfsx:
This is a part of #5172 and replaces the current implementations of zip/unzip in ilFileUtils


alex40724:
@chfsx Looks great. Could you add some example consumer code in a readme?

chfsx:
@alex40724 I am working on a general readme on how to work with files. I will also talk about ZIP, but I will tell you that you should not use it anymore. To create a ZIP, you don't need a directory in the file system that you can zip. You can do that all in memory and you don't need expensive copying processes and possible relicts in the file system. The documentation will therefore rather be: "Don't use it" together with a documentation how to make ZIPs in the future.

I have changed the PR a little bit, now there is also a UnitTest for Legacy-ZIP included

chfsx:
@Amstutz I has to skin the following test in your component: https://github.com/ILIAS-eLearning/ILIAS/pull/5181/files#diff-3779baed56463a02084554a9828aa824093d505ecae2fe702eb2d3e7d45075adR122
With the changes Zipping/Unzipping is only possible in directories known to the `LegacyPathHelper`, there are Directories which can have User-Inputs. Your test tries zu Zip inside the ILIAS Code base which is now no longer possible for security reasons.

Amstutz:
Hi @chfsx 

De-activating this test is fine for me, should be improved anyway. However, can you anticipate if this has side effects on the importing (genarating a zip ou of files of the customizing dir)/exporting (unpacking a skin from and moving files to the customizing dir) of skins?

thx for the work! highly appreciated.

chfsx:
@Amstutz Customizing is one of the "allowed" directories, therefore this should work. i will test it an monday

matthiaskunkel:
Jour Fixe, 06 MAR 2023 : We highly appreciate this suggestion - especially the how-to - and accept the PR for trunk.

PR #5182, opened 2022-11-07
Proposal: DateFormat from User

schmitz-ilias:
Currently, using the KS datetime input field is a bit annoying: its format has to be given as a Data/DateFormat, but the format from the user's preferences is returned as an ilCalenderSettings constant. So, to have the input use the user-preferred format for dates, one has to write an adapter. This adapter is pretty simple, but especially with an eye toward the big KS initiative it would have to be replicated in quite a lot of Services and Modules. In my opinion it would be better to have it someplace central, to avoid this needless repetition.

To this end I added 'americanShort' as a preset DateFormat to the DateFormat factory, so that all three possible user preferences are available as presets. Further, I added the aforementioned adapter as a method 'fromUser', taking ilObjUser as an argument.

I'm aware that adding this kind of dependency on ilObjUser (and ilCalenderSettings) to src/Data is a big no-no. However, I'm not sure where else it should go, and want to open that question up to discussion.

So, in summary: making KS datetime inputs work with the format from the user preferences is annoying, and a publicly available adapter would help. Where should we put it?

### DIFF
--- a/src/Data/DateFormat/Factory.php
+++ b/src/Data/DateFormat/Factory.php
@@ -2,7 +2,21 @@
 
 declare(strict_types=1);
 
-/* Copyright (c) 2019 Nils Haagen <nils.haagen@concepts-and-training.de> Extended GPL, see docs/LICENSE */
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
 
 namespace ILIAS\Data\DateFormat;
 
--- a/src/Data/DateFormat/Factory.php
+++ b/src/Data/DateFormat/Factory.php
@@ -44,4 +44,24 @@ class Factory
         return $this->builder->weekday()->comma()->space()
                              ->day()->dot()->month()->dot()->year()->get();
     }
+
+    public function americanShort(): DateFormat
+    {
+        return $this->builder->month()->slash()->day()->slash()->year()->get();
+    }
+
+    public function fromUser(\ilObjUser $user): DateFormat
+    {
+        switch ($user->getDateFormat()) {
+            case \ilCalendarSettings::DATE_FORMAT_DMY:
+                return $this->germanShort();
+
+            case \ilCalendarSettings::DATE_FORMAT_MDY:
+                return $this->americanShort();
+
+            case \ilCalendarSettings::DATE_FORMAT_YMD:
+            default:
+                return $this->standard();
+        }
+    }
 }
--- a/src/Data/DateFormat/Factory.php
+++ b/src/Data/DateFormat/Factory.php
@@ -63,19 +63,4 @@ class Factory
     {
         return $this->builder->month()->slash()->day()->slash()->year()->get();
     }
-
-    public function fromUser(\ilObjUser $user): DateFormat
-    {
-        switch ($user->getDateFormat()) {
-            case \ilCalendarSettings::DATE_FORMAT_DMY:
-                return $this->germanShort();
-
-            case \ilCalendarSettings::DATE_FORMAT_MDY:
-                return $this->americanShort();
-
-            case \ilCalendarSettings::DATE_FORMAT_YMD:
-            default:
-                return $this->standard();
-        }
-    }
 }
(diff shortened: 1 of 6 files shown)


Amstutz:
Hi @schmitz-ilias 

Thx for the initiative here! Always happy to see neat little improvements and discussions about making things easier and better.

I am not coordinator of the data lib, so take everything with a grain of salt. Just my five cents.

As you have guessed, introducing the full public interface of ilObjUser as dep. Towards Data is no good. Havoc will start as soon as you will be required to Unit Tests your code (which will probably be very soon ;-) ). However, inverting this and Introducing Data to user, should be fine (best only the abstraction of it, for ilObjUser to be testable at some point as well). 

So, I would most likely just move the function to ilObjUser, as long as we store the answer to this question there and let user return DateFormat.

Looking forward to read different opinions.

klees (commented):
Hi @schmitz-ilias,

thanks a lot for looking into this. Always a good idea to improve ergonomics of our code, even if it only is a tiny bit =)

I would definitly accept the `americanShort`. No questions or remarks there =)

Regarding `fromUser` we all seem to agree on the general problem. I would ask you to either move that method to `ilObjUser`, potentially by simply introducing something like `getDataDateFormat` that does the necessary translation.

On the other hand, it looks as if the original `ilObjUser::getDateFormat` is only ever called in `ilCalendarUtil::getUserDateFormat`, which basically seems to do the thing that we are also doing in `Data\DateFormat`. So I think it would be even better to return a `Data\DateFormat` from `ilObjUser::getDateFormat` directly and get rid of most of the code in `ilCalendarUtil::getUserDateFormat`.

Feel free to contact me via Discord on short notice if we should look into this together some more.

Thanks for the initiative!

schmitz-ilias:
Hi @Amstutz, hi @klees, thanks for the feedback! I followed your advice and moved `fromUser` to `ilObjUser::getDataDateFormat`.

Sadly, changing the existing `ilObjUser::getDateFormat` to directly return a `DateFormat` doesn't really work well in my opinion. Apart from `ilCalendarUtil::getUserDateFormat` it's also used in `ilCalendarUserSettings::read`. From there the format in its incarnation as an `ilCalendarSettings` constant gets passed around to some GUI classes, and is eventually returned to `ilObjUser` to be saved there as a user preference. So short of either translating back the format in `ilObjUser` before saving it (which seems like a pointless exercise), or changing altogether how the format is saved in the database (which seems like a disproportionate headache), adding a second method seems like the way to go, at least for the moment.

kergomard:
Thank you very much for your proposal and your work on it, @schmitz-ilias !

I think now this PR landed partly in my responsibility concerning the User Service.

I think generally this is the right place to have this, but I do not like the introduction of an additional method for it. As far as I can see the general structure here is not what we would want because of  `ilObjUser::getPref()` and `ilObjUser::setPref()` being public and just working on an array. So this will have to change anyway in the near future. It also makes it a little harder to find all places where a date might be set or gotten.

I would still like to change `ilObjUser::getDateFormat()`. From where I stand, the right way of doing this, would be to change the internal representation of 'date_format' and then handle all the corresponding places and require a `DateFormat` on the Setters, too. I can take a closer look at this in March, but if you want, I'm very grateful if you speed things up and adapt this PR accordingly. If you want to do the first step, I'm also ok with just changing `ilObjUser::getDateFormat()` and I then take care of the rest later.

@kergomard

src/Data/DateFormat/Factory.php
```diff
@@ -2,7 +2,21 @@
...
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
```
klees, line 19: Thanks =)

klees (approved):
Changes in `src/Data` LGTM. Thanks!

schmitz-ilias:
Hi @kergomard!

I implemented your feedback, only one `ilObjUser::getDateFormat` remains. This necessitated some changes in `Services\Calendar`, where I had to immediately translate the `Data\DateFormat` back to an integer constant. As far as I can see, properly using the `DateFormat` throughout would entail changes in a couple of settings GUIs, and more importantly to how preferences are saved in `ilObjUser` - that seems like a bigger job than what I had in mind here =) 

I hope you and @smeyer-ilias are satisfied with what I'm proposing here as a temporary solution, otherwise we can also postpone this little improvement until you are making bigger changes to `User`.

kergomard:
Thanks a lot @schmitz-ilias .
One more little change, sorry, and then I'm fine: You are right that the DateFactory should be injected, but I don't think this function is the right place. Please just instantiate it in the constructor of `ilObjUser` and assign it to a class property. I will then move it on from there in a next step.
Best,
@kergomard

schmitz-ilias:
@kergomard done, thanks for the feedback!

Thanks @smeyer-ilias for merging, but I noticed and adapted another occurence of `ilObjUser::getDateFormat` in `Modules/DataCollection` in the last commit, I hope the responsible maintainer doesn't mind that that slipped in there. Sorry for that last minute addition!

PR #5189, opened 2022-11-08
Bugfix 33747 Entries of a copied DS are not displayed

sasfalk:
#bugfix
Mantis issue: 33747 and 32228
Target: release_7
https://mantis.ilias.de/view.php?id=33747
https://mantis.ilias.de/view.php?id=32228

Both tickets refer to the same bug.


jweber-unibe:
Dear @sasfalk 
_I am not a developer (and this is my very first Github-comment ever!), so please forgive if I am mistaken here ..._ but isn't this PR completely identical to #5155 (apart from the position of the added line)?!
In this case I believe it would be better to close one of the PRs in order to reduce redundancy and prevent irritation.
Best regards, Jens

Amstutz:
Hi @sasfalk 

I also gave this a long and hard look and also believe this to be a duplicate of https://github.com/ILIAS-eLearning/ILIAS/pull/5155 . In case we missed something, feel free to reopen.

Thx a lot for your kind contribution!

PR #5190, opened 2022-11-08
Bugfix 31872 show only plugin slots for available plugins

sasfalk:
#bugfix
Mantis issue: 0031872 and 0031873
Target: release_7
https://mantis.ilias.de/view.php?id=31872
https://mantis.ilias.de/view.php?id=31873

### DIFF
--- a/Services/Component/classes/class.ilCachedComponentData.php
+++ b/Services/Component/classes/class.ilCachedComponentData.php
@@ -84,6 +84,10 @@ class ilCachedComponentData
     /**
      * @var array
      */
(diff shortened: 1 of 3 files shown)


klees:
Merged as https://github.com/ILIAS-eLearning/ILIAS/commit/f35a9336eed64b39dbc5687aa94e50b3fa174c3c with a minor change for when there are no plugins at all.

klees:
@sasfalk: Would you be so kind and provide another version for release_8?

sasfalk:
Yes, of course. I'll take a look at it as soon as I can :)

PR #5083, opened 2022-09-29
Bugfix 32045 Test in imported learning sequence description error

sasfalk:
#bugfix
Mantis issue: 0032045
Target: release_7
https://mantis.ilias.de/view.php?id=32045

The error only occurred if the test in the imported learning sequence did not contain a description. The ilLegacyKioskModeView class couldn't handle an empty variable, so I added an if statement.

PR #5191, opened 2022-11-08
Bugfix 32045 Test in imported learning sequence description error

sasfalk:
#bugfix
Mantis issue: 0032045
Target: release_7
https://mantis.ilias.de/view.php?id=32045
Follow up PR for: #5083 

The error only occurred if the test in the imported learning sequence did not contain a description. The ilLegacyKioskModeView class couldn't handle an empty variable, so I added an if statement.

### DIFF
--- a/Modules/LearningSequence/classes/Player/class.ilLegacyKioskModeView.php
+++ b/Modules/LearningSequence/classes/Player/class.ilLegacyKioskModeView.php
@@ -133,8 +133,13 @@ class ilLegacyKioskModeView implements ILIAS\KioskMode\View
 
         $info = $factory->item()->standard($this->object->getTitle())
             ->withLeadIcon($icon)
-            ->withDescription($this->object->getDescription())
             ->withProperties($props);
+
+        $description = $this->object->getDescription();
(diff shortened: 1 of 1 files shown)


klees (commented):
Hi @sasfalk,

I'm sure this will alleviate the symptoms described in the ticket, but I'm quite convinced that this is not a fix for the underlying issue. I'm sorry that I did not point that out right away in the previous PR, I simply did not remember... =|

See e.g. [#32695](https://mantis.ilias.de/view.php?id=32695)

From my perspective, the LSO access a generic `ilObject`, where the interface of that class promises a `string` when [`getDescription`](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/Services/Object/classes/class.ilObject.php#L431) is called. This does not seem to be the case under the circumstances you describe.

So I think we should rather fix this in the import of the Test or in `Services\Object`, because the general problem will pop up elsewhere again.

@kergomard @dsstrassner: Thoughts?

Best regards!

klees:
Thanks for looking into this, I will close this for the time being.

PR #5197, opened 2022-11-10
TA: 35347, fix time extension gui's index errors

nhaagen:
https://mantis.ilias.de/view.php?id=35347


kergomard:
Thank you very much @nhaagen , especially for centralizing the usage of the $DIC.

Two little requests/questions, though:

- [ ] Why not just use the parts of the $DIC we actually need? I think it would only be language and ctrl. This would avoid usages of the $dic, where a class property is present (see e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5197/files#diff-875fad923759d3cb6c0feb72922835c6ac52686cc904a2afdaa38861a3160b5dR235)). Or am I getting this wrong?
- [ ] Please declare the class property/ies before using it/them.  

Thanks and best,
@kergomard

nhaagen:
@kergomard, yes, you're right and I was lazy ;)

kergomard:
Thank you @nhaagen !

PR #5198, opened 2022-11-10
TA: 34717, some array-shifting fixes in ordering images

nhaagen:
https://mantis.ilias.de/view.php?id=34717


kergomard:

- Are you sure about this change [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5198/files#diff-d625889a0323c69d5dada763c520ef596a78022fb2be4f66549062072d8487f9R157)? The value is initially null and is not set in the constructor. 
- Could you, in any case, also remove the lying DocBlock, please.

Best,
@kergomard

kergomard:
I will merge this. This does not make the ordering-question usable in ILIAS 8, but it is a step in the right direction. I will cherry-pick to ILIAS 7 and trunk.

PR #5199, opened 2022-11-10
Bugfix 0035599 (trunk): Multiple Choice Question negative points

fhelfer:
This fixes error message mentioned in https://mantis.ilias.de/view.php?id=35599


iszmais (changes requested):
LGTM, just some minor Change Request:

Since the origin of the problem is cause by the negative `$max` due to the test result i would suggest to extend the already existing prove `(!$max)` to the (also type correct) version `($max 
 <= 0.0)` and remove the correction within the precentage.

iszmais (approved):

kergomard:
Sorry, I automatically merged and cherry-picked your solution for ILIAS8 without thinking. I close this one.

PR #5202, opened 2022-11-10
TA: 35522, check on existing solutions for orderingQuestion horizontal

nhaagen:
https://mantis.ilias.de/view.php?id=35522
https://mantis.ilias.de/view.php?id=35604


kergomard:
Thanks @nhaagen !

PR #5203, opened 2022-11-10
[FEATURE] UI: `Form` compatibility with `Modal\Roundtrip`

thibsy:
Hi @Amstutz & @klees

We have once discussed the current state of using `Form`s within a `Roundtrip` modal, together with @mjansenDatabay and @nhaagen (and maybe others I can't recall?). This PR is a proposal to solve the issue by introducing a new type of form, called `NoSubmit`, which is currently only for internal use. I also refactored the `File\Dropzone` which also used a `Roundtrip` modal during its rendering process.

I named it `NoSubmit` form because it'll be rendered without any submit buttons. Instead, a submit `Signal` can be retrieved to submit the form on client side. I would have really liked to use the [form](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-form) attribute of buttons to submit `NoSubmit` forms, but since the IDs of components are currently only initialized during the rendering process it wasn't quite possible to link a button to such a form. To ease the rendering process of `Form`s I also split up the implementation into another additional class named `HasAction`, which indicates forms that support a POST-URL.

I refactored the `Roundtrip` modal to implement the `Form\Standard` interface, because the modal needs a POST-URL as well. I structured the implementation so the `Roundtrip` serves as a decorator and only delegates the `Form` method calls to the actual implementation, which is initialized in its constructor. New, while rendering the `Roundtrip` a submit  button is automatically added with the according on-click `Signal` to submit the form. The caption can be changed with `Roundtrip::withSubmitCaption`, similar as to the cancel caption. I obviously needed to change the constructor signature of `Roundtrip`s as well, but the change **is not breaking** any existing usages. I added an optional `Input` array and post-url to have a similar signature as to `Form`s. I also allowed the `Component` array, which serves as the modal content, to be null because there might be use-cases in the future where one will only want to display a `Form` in it.

What I didn't realize was, that the `File\Dropzone` implementation already used a `Roundtrip` modal internally, so I had to refactor this component as well with this PR. I was quite bold with this one because I know it doesn't have many usages. I changed `File\Dropzone`s to implement the `Roundtrip` interface, because IMO this component basically _is_ one. What makes this component unique, is merely the droppable area, which transfers dropped files on client side to the `Roundtrip`s `Field\File` input. Therefore, I thought why not let consumers be flexible with `File\Dropzones`? Up till now consumers weren't able to adjust the `Roundtrip` modal that's been used internally in any way. I also thought it redundant to require the same properties as the `Field\File` input, only so the `File\Dropzone` can implement the `FileUpload` interface as well. By doing this, it also lead to kind of an indefinite state, where the `Form` implementation of the dropzone had to be created dynamically on each `getForm` call, because the properties of the file-input could be changed after the component was initialized. This can be avoided now because the `File/Dropzone`s already accept a "finished" `Field\File` input. This component as well, serves as a decorator that only delegates the `Roundtrip` method calls to the actual implementation.

By transforming both `Roundtrip` and `Dropzone` components to decorators, the rendering process has become very nice IMO, because one can just retrieve the actual implementation by an according getter (which isn't in the public interface) and delegate the rendering to the passed along default renderer.

Kind regards!


chfsx:
@thibsy see also discussion in https://mantis.ilias.de/view.php?id=35971 concerning labels of the Inputs

thibsy:
Hi all,

I have rebased onto release_8 once again and implemented the new "required-inputs" feature for the `NoSubmit` form as well (see #5255) and added some unit tests.

@chfsx I also took care of the label and header-dropzones as described in https://mantis.ilias.de/view.php?id=35971.

Ignore the failing checks, since it is the example bug once again :/ ... I'll look into this right now.

Kind regards,
@thibsy

chfsx:
Related reports which can be closed after merging:
- https://mantis.ilias.de/view.php?id=35971
- https://mantis.ilias.de/view.php?id=33878

Testcases which should be runned again:
- https://testrail.ilias.de/index.php?/tests/view/54337
- https://testrail.ilias.de/index.php?/tests/view/54340

thibsy:
Hi all,

I did not realize that the ILIAS 8 release has been scheduled for tomorrow. I therefore moved the label-fix to a separate PR which will be ILIAS 9 only (for now). If need be, we can still cherry-pick the fixes to 8 as well.

Regardless, this PR needs to be in the ILIAS 8 release.

Kind regards,
@thibsy

chfsx:
@klees any change we could merge this before the release of 8 tomorrow? let me know if I can help in some way

klees:
@chfsx Yes, it's on my radar =)

klees (changes requested):
Hi @thibsy,

thanks a lot for looking into this. Definite improvement. Two general remarks:

There might be a second iteration regarding this, as we (CaT Concepts and Training GmbH) are looking into generalizing Roundtrip even more to get something like an "Input Wizard". Not sure where we'll end up there, though 🙂

Second thing, really important: Please do not include automated changes that are not warranted by our common coding-style, such as these [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5203/files#diff-7da0060b76695771e40e879022750775741454de17057dfe585acc16a4887174L26). I understand that you prefer some specific style for including names via use (and other style preferences). Please either try to find and codify a common style (for UI framework or the community) or just do not include these changes. I mean, if I'd like another style, would I simply change the code back in my next PR? 🙂 No need to roll back changes now, but please do not do this again...

- [ ] HasAction1: This probably should be a trait, because basically it seems to exist for code-sharing. "Extends" also implies an "is a"-relation, which is a semantic we don't seem to need. The constructor would become a protectec setPostURL then. Please also see HasAction2 below.
- [ ] NoSubmit: I would propose to just be real with names. Can we go with FormWithNoSubmitButton? Because we can in fact submit, right? It is strictly internal, right?

- [ ] HasAction2: Please rename this to HasPostURL.

Best wishes!

thibsy:
Thx @klees for your Feedback!

I implemented the changes and merged this right-away because of the ILIAS 8 release. If there are still some remarks, let me know, I can commit them directly to release_8 then.

Kind regards,
@thibsy

thibsy:
Picked for Trunk as well:
- https://github.com/ILIAS-eLearning/ILIAS/commit/92e38cb3140796f64d8680eab062871f9e6e69d3
- https://github.com/ILIAS-eLearning/ILIAS/commit/e51efb7fa8376f269b1e93e234aaa7c6e9688e6c

PR #5205, opened 2022-11-12
Reserved names

utesche:
The columns `system` and `export` are reserved names in MySQL/MariaDB and must be escaped when used explicitly in a SQL statement.
See https://docu.ilias.de/goto_docu_frm_1875_8392.html .


kergomard:
Thank you very much @utesche !

PR #5206, opened 2022-11-14
Add quality assurance review to md

kergomard:
This proposal introduces an ongoing assessment of code maturity in ILIAS to continuously improve the sustainability of our product.
It includes:
- Regular Reviews of all parts of the code.
- Criteria for these reviews that can partly be assessed automatically, but also allow room for qualitative feedback.

Please have a look at this proposal and give us feedback until the Jour Fixe on November 28th 2022.

### DIFF
--- a/docs/development/review.md
+++ b/docs/development/review.md
@@ -1,20 +1,79 @@
 # Review
 
-A review is given by a reviewer on code contributions (see: [contributing.md](contributing.md)) of 
+A review is given by a reviewer on code contributions (see: [contributing.md](contributing.md)) of
 a reviewee. There are no guidelines on how review has to be performed in ILIAS, however here we propose a pattern
 inspired by an article of Dan Munckton in [How to be a kinder more effective code reviewer](https://cultivatehq.com/posts/how-to-be-a-kinder-more-effective-code-reviewer/).
 
+ILIAS knows two different types of reviews:
+- Reviews that are given by developers to each other as part of the process to introduce improvements through Pull Requests
+- Reviews that are conducted regularly as part of the Quality Assurance Process as a way to assess Code Maturity.
+
+## General Pattern
 In our experience many conflicts arise due to feedback written in a fashion which makes it hard to understand,
 what exactly is asked of a reviewee. The pattern adapted from suggestions by Munckton is aimed
 to make reviews easier to read and to leave less room for misunderstandings and communicational deadlocks.
 
- 
-
-## General Pattern
-We try to lay a strong focus on making feedback **actionable**, meaning the reviewee should have 
+We try to lay a strong focus on making feedback **actionable**, meaning the reviewee should have
 a clear picture on how to react. Inspired by Munckton (see below for more details) we propose
-to use the following structure to reply to PRs:
+to use the [template provided below](#template-for-reviewing-prs) to reply to PRs.
 
+## Reasoning
+Munckton introduces the term **actionable** for feedbacks. An **actionable** feedback is one that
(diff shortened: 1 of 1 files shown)


matthiaskunkel:
14 NOV 2022 : @kergomard notified us about efforts of the TB to introduce quality measurements for the ILIAS code. Stephan pleases all developers to read the suggestions and to give feedback until the next Jour Fixe.

docs/development/review.md
```diff
@@ -75,51 +133,69 @@ kindly,
...
+	- [ ] The type `object` is not used
+	- [ ] `Mixed` is only used instead of a generic
+	- [ ] No named parameters are used
+- [ ] General Code Quality:
+    - [ ] There are no accesses to unknown classes
+    - [ ] There are no accesses to unknown methods
+    - [ ] There are no usages of undeclared class properties
+    - [ ] There is no dead/unreachable code
+    - [ ] There are no accesses to methods that are potentially missing (e.g. when Union Types or Magic Functions are used)
+    - [ ] All parameters and return values are either typed or their expected type is declared in a DocBlock, but not both
```
tfamula: I stumbled over this criterion. It says typing and DocBlock are not allowed at the same time. I often have an `array` typed as a parameter or return value and additionally make a more specified declaration in the DocBlock, like `@var int[]` or `@return int[]`. This should still be possible IMO.

kergomard: Thank you very much @tfamula  for the feedback and you are right. Even though we should mostly get rid of the arrays,  this is a unique case and I agree that it should be added, so I would suggest:
```suggestion
    - [ ] All parameters and return values are either typed or their expected type is declared in a DocBlock. No parameter or return value is typed and has a DocBlock aswell, unless the DocBlock specifies the types contained in an array or an iterator.
```

klees: Hmm, valid point. Should we loosen this a bit? I think something similar would apply to `Iterator<$TYPE>`.

kergomard: Personally, I would stick to the allow-listing approach. I added the iterator to the list above. If we see it getting to long, I will start the part of my brain responsible for synthesis. Ok?

docs/development/review.md
```diff
@@ -75,51 +133,69 @@ kindly,
...
-
-Do not leave any reactions in inline comments. They tend to crowd the discussion without giving
-the possibility of being resolved (seens not actionable). Move them to the summary.
+Please commit the following file as part of the review to the root of the component.
+To measure the statistics you can use:
+- Logical Lines of Code: [phploc](https://github.com/sebastianbergmann/phploc): `php phploc.phar <Path to Component>`
+- Nr of Tests: phpunit: `./libs/composer/vendor/phpunit/phpunit/phpunit --list-tests <Path to Component Test Suite File>`
+- Nr of LoCs changed in last 365 days: `git diff --shortstat HEAD $(git log --since=365.days --oneline Services/WebDAV/ | tail -1 | cut -f 1 -d " ")^ <Path to Component>`
+- Nr of Commits in last 365 days: `git log --since=365.days --oneline <Path to Component> | wc -l`
+- To check the general code quality: `./CI/PHPStan/run_checks.sh <Path to Component`.
```
klees:
```suggestion
- To check the general code quality: `./CI/PHPStan/run_checks.sh <Path to Component>`.
```

kergomard: Yup, that is it. Thanks!

matthiaskunkel:
Jour Fixe, 28 NOV 2022: Thanks for considering Thomas' suggestion and to adapt the PR. We highly appreciate the PR and accept it for trunk.

PR #5207, opened 2022-11-14
UI Key-Value Interruptive Item

schmitz-ilias:
Currently, the options to communicate to the user in an interrruptive modal what they are about to delete are fairly limited: one can only add 'interruptive items' to the modal. Consisting of title, icon, and description, they are tailor-made for e.g. repository objects, but less useful in other scenarios. For me, this came up in the context of the new metadata editor, but another prominent example in ILIAS would be deleting users from a table; if not impossible it is at the very least semantically strange to identify users by title and description.

In order to improve the situation, I propose adding a key-value type of interruptive item, in addition to the current standard one. It consists of a single line of text starting with a key, and appended to it in grey and cursive a corresponding value (similar to the key-value tree node, or the characteristic value listing). In the example with the table of users, one could then e.g. list the users' login as the keys, and their names or e-mails (or a combination) as the values. Due to its somewhat generic format, I expect this type of interruptive item to also be useful in many other scenarios.

KeyValueInterruptiveItems

Concerning the implementation, I promoted InterruptiveItem to a proper UI component, with its own factory and renderer. What is currently the only interruptive item is now accessible as interruptiveItem->standard. This should make it easier to add other types of interruptive items, if needed, but required small changes where interruptive items are already implemented (Services/MainMenu, Modules/Bibliographic, and Services/AdministrativeNotification).

Looking forward to your feedback!


thibsy:
Hi @schmitz-ilias 

Thank you for the contribution to the UI framework! We did not forget about this PR, but due to the fixing of ILIAS 8 issues we still need a little more time to finally look into this. Thanks for your understanding.

Kind Regards!

Amstutz:
Hi @schmitz-ilias

Sry for letting you wait for some time here. As @thibsy wrote, are working hard to get the Release 8 through the door. However, here is the first part of the feedback.  Since this has a rather large footprint on the public interface, we will need to discuss the interface change during the JF. Therefore, we only review the interface change here.

The extent of the list of possible interruptive items seems like a logical next step when extending the interruptive modal. So, in general, your approach seems fine. You could have resolved this by just using the title as kind of key and description as kind of value. However, we agree that this might prevent us from properly styling and layouting them in the future.

However, it also raises some questions:

- [ ] Mixing of Items: In your proposals, the interruptive Items could be mixed, correct? Should we allow that, those this make sense? Would this not set boundaries for the layout in the future, that are unnecessary (e.g. it could prevent is from generating easy to use lists by screen readers). Maybe we should better prevent mixing two different types of items inside one interruptive modal. If we prevent this, add a composition rule on the parent Item level.
- [ ] Keys: Can you provide examples of when we would need seperate keys in one list?


Please add or explain why not:

- [ ] Purpose: Since we need a clear picture for layouting this properly, of what type of data, we would need a clear picture of when this will be used. Can you sharpen the purpose description further?
- [ ] Usage: Why is Value only requiring by “SHOULD” affordance. It seems like part of the essence of what a keyValue item is. See also Composition description. Probably change to MUST.


Please change

- [ ] Accessibility: Key Value Pairs should be rendered as descriptive listings. Please add this rule. 
- [ ] Usage of ID: The “MUST” affordance of the ID is already given in the parent interruptive item. Omit it in the two children standard and key value.
- [ ] Firebase Key Usage: Why do you add the use Firebase\JWT\Key; namespace? Probably remove it.


thx a lot!

schmitz-ilias:
Hi @Amstutz, thanks for the feedback!

> Keys: Can you provide examples of when we would need seperate keys in one list?

For some cases it should be clear from context what the value is, so the key can be more specific. If you were to delete a bunch of weblinks from a list, it's obvious that an url as the value is the url of the weblink, so instead of identifying the links by 'Weblink: [URL]' one would use '[name of the weblink]: [URL]'. There might also be examples where you'd want to delete different kinds of objects, maybe submissions in an exercise. Those can be files or wikis or other stuff, right?

>  Mixing of Items: In your proposals, the interruptive Items could be mixed, correct? Should we allow that, those this make sense? Would this not set boundaries for the layout in the future, that are unnecessary (e.g. it could prevent is from generating easy to use lists by screen readers). Maybe we should better prevent mixing two different types of items inside one interruptive modal. If we prevent this, add a composition rule on the parent Item level.

Yeah, that "feature" is in there mostly because it's a side-effect of implementing different kinds of interruptive items while trying to change the interface as little as possible. As an attempt at a compromise I added a composition rule discouraging mixing item types, and played around with grouping different item types, should they get mixed anyways. This is what I came up with:
Screenshot from 2022-12-08 19-21-55
I agree that anyone ever using this is a stretch, but it would at least prevent any "dependencies" between layout and styling of the different types of items. Maybe it might be useful for showing an object and its content when deleting it, e.g. one could show title and description of a weblink object as a standard item, and then the individual weblinks below it as key-value items? 

I implemented your other suggestions, let me know if my changes are what you had in mind.

Amstutz:
Thx, IMO this is JF-Ready. I would rather not mix the two types and check this. However, If you have a strong opinion, we can also chat about this on the JF with fellow devs.

Thx for the contribution.

schmitz-ilias:
Thanks again for the feedback @Amstutz! I'm not that attached to the grouping idea, I'd also be happy with introducing a separate container object for each type of interruptive item to enforce a do-not-mix rule in the API. I guess we can present both options at the JF, and see how people feel about them.

I will not be at the next JF, so I'd suggest postponing the discussion of this PR to January.

Happy Holidays!

matthiaskunkel:
Jour Fixe, 09 JAN 2023:  We highly appreciate this suggestion and accept the PR and KS change for ILIAS 9 / trunk. We prefer Tim's solution for mixing items.

Amstutz:
Hi @schmitz-ilias 

Let us know, if this is ready for a final review.

Kindly
UI Coordinators

schmitz-ilias:
> Hi @schmitz-ilias
> 
> Let us know, if this is ready for a final review.
> 
> Kindly UI Coordinators

Hi all, I did some tidying up, should be ready now. The unsuccessful checks are the copyright checker complaining about KS examples, but I'm pretty sure these are not supposed to have the cp notice?
I did not add any cases to testrail yet, I'll save that as a last step until you are satisfied with the implementation.

Amstutz:
Hi @schmitz-ilias 

Yes, no need to worry about the checks for the examples. Actually I thought we got rid of them, but I might need to look into that again. Adding testrails after merge is fine, thx.

Amstutz:
Hi @schmitz-ilias 

Thx a lot for the implementation. Sry that the review took longer than we aim at. 

Special thx for the unit tests, examples and usage of the ITCSS structure!

Please change:
- [x] Stefan Wanzenried: Is listed as author of classes, please remove. Best remove the class docstrings if they don't provide additional value/documentation.
- [x] <br> Tag: Since you factored out some DOM part of the Item template into its own template (thx a lot for that!), please also move the <br> tag to the tpl (still residing in the PHP code here: https://github.com/ILIAS-eLearning/ILIAS/pull/5207/files#diff-138d50db4fa0634ee087a15fe78c97a960d884fda0b7153406d7a5494a05243eR58
- [x] CSS Class Naming conventions: Note that we have new Class Naming convention: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/templates/Guidelines_SCSS-Coding.md#naming-conventions and https://csswizardry.com/2015/08/bemit-taking-the-bem-naming-convention-a-step-further/ . Please use them for all new classes added. You may keep already existing classes as is.
- [x] Unit Tests: Since we allow the mixing of different types of items, we should also test the proper rendering (ordering of the items in the rendering) of those items. Add one or more unit tests, checking that mixed lists of items are rendered in the expected order.

Thx a lot
@amstutz, @thibsy , @klees

schmitz-ilias:
Hi all,
I made the requested changes. I hope I got the naming scheme right, some of the class names got a bit wordy but as far as I can tell that's intended.
Thanks for the feedback!

Amstutz:
Hi @schmitz-ilias 

Thx a lot, I looked into the class naming. I am still learning Bemit as well ;-). As far as I understand, your naming looks good to me. thx.

I resolved the conflicts and merged.

PR #5208, opened 2022-11-14
T&A: fixes 35034, save the unit of the answer also in the array $solutionSubmit

okaufman:
Mantis-Ticket: https://mantis.ilias.de/view.php?id=35034

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assFormulaQuestion.php
+++ b/Modules/TestQuestionPool/classes/class.assFormulaQuestion.php
@@ -1343,20 +1343,22 @@ class assFormulaQuestion extends assQuestion implements iQuestionCondition
 
         foreach ($this->getResults() as $index => $a) {
             $key = "result_$index";
-            if (
-                $post->has($key)
-                ||
-               $post->has($key . "_unit")
-            ) {
-                $value =$post->retrieve(
+            if ($post->has($key)) {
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @okaufman ! I think this fix is actually right, but as far as I can see, it doesn't fix the issue, right? What am I not seeing?

okaufman:
The function that I changed getSolutionSubmit() is called in savePreviewData in assFormulaQuestion.php. There the unit will be saved in the previewSession with `setParticipantSolution($userSolution)`. 
It is saved as a key-value-array with the key "$r1_unit". Then this array is accessed again in assFormulaQuestionGUI.php on line 907 with           
  `$participantsSolution = $this->getPreviewSession()->getParticipantsSolution();`
There the information about the unit gets passed to getBestSolution() in assFormulaQuestion.php where the error message showed up.

On my local machine this fixed the problem.  If you create a new question, the error does not show up. If you already have a FormulaQuestion where the error shows up as soon as you click preview, you have to somehow "reset the preview" and the error no longer shows up.

kergomard:
Yes, you are right! I went through the thing and was under the impression that $r_unit could also potentially not be set at all (when needed). But I also believe this case can not happen. So: Thank  you @okaufman, I will merge and cherry-pick.

PR #5211, opened 2022-11-15
New API section in documentation of Skill Service

tfamula:
I just want to inform that there is a new section called API in the `README.md` of the Skill Service. It gives an overview of all current Skill Services which are used in other components by the public interface. In the future, the documentation will be extended by more Services respectively more methods in existing Services.

### DIFF
--- a/Services/Skill/README.md
+++ b/Services/Skill/README.md
@@ -1,10 +1,84 @@
 # Skill aka Competence Service
 
-This README is work in progress.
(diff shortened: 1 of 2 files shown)


matthiaskunkel:
Jour Fixe, 28 NOV 2022: Thanks to @tfamula for extending the documentation of this service. Merge to trunk is highly appreciated.

PR #5212, opened 2022-11-15
T&A Fix 35321

okaufman:
Mantis: https://mantis.ilias.de/view.php?id=35321 
The only problem is, that after checking if the user did choose default-settings the rest of the function still execeuted.
Thats why I added "return".
I also noticed that there was the same problem, when a test already has results from participants. The settings of the test should not be changed, when results are already present. That's why I added a second "return".


kergomard:

PR #5216, opened 2022-11-16
SCSS broken Glyphicons

catenglaender:
Some Glyphicons went missing after the less to scss refactoring.
* I restored Glyphicons by fixing a path and also
* adjusted the brightness of icons in the mainbar (they were very, very light blue).
Color before
image
Color after
image


Amstutz:
thx, I still did not find the time to migrate all the 8 fixes. This one, was one of the list. thx for the fix!

PR #5218, opened 2022-11-17
removed the query_cache_size hint from install docu

acgruber:
See [MySQL](https://dev.mysql.com/doc/refman/5.7/en/query-cache-configuration.html): The query cache is deprecated as of MySQL 5.7.20, and is removed in MySQL 8.0.


chfsx:
Hi @acgruber and thanks you very much for your patience! Nice catch, I wasn't aware of this deprecation. Since ILIAS 9 will support MySQL 8.0 and no longer 5.7 it makes perfectly sense to remove that line

PR #5219, opened 2022-11-17
T&A #0035319: ECTS-FX Einstellungen speichern

fhelfer:
https://mantis.ilias.de/view.php?id=35319

removed string manipulation because ilObjTest::$ects_fx is already float
Only works because ilObjTest has no declare(strict_types=1)


kergomard:
How about we also fix the `ilObjTest::getECTSFX()` and `ilObjTest::setECTSFX()` as we are at it? I would suggest to remove the strlen check in the getter and just use a null coalescing operator. And then I would suggest to do the str_replacement in the setter and explicitly cast to float afterwards. What do you think? Me thinks this would make things more predictable.
Best,
@kergomard

kergomard:
Thanks @iszmais !

PR #5221, opened 2022-11-17
T&A #0035314: Delete Mark Step not removed from db

fhelfer:
https://mantis.ilias.de/view.php?id=35314


kergomard:
Thanks @fhelfer !

PR #5222, opened 2022-11-18
Bugfix/0026255 text field of data collection with limit

sasfalk:
#bugfix
Mantis issue: 0026255
Target: release_7
https://mantis.ilias.de/view.php?id=26255

When adding a new text field, the explanation now says that the field is limited to 4000 characters. Also, when writing in a text field, it shows how many remaining characters can be entered.
Increasing or removing the limitation of the field is complicated because the content of the text field is stored in a varchar(4000). You could replace this with a blob or clob, but this would require restructuring the database.


Amstutz:
Might only be a fix, until we really takle that issue, but a lot better than status quo. Thx a lot @sasfalk !

PR #5224, opened 2022-11-18
Bugfix: LearningSequence copy incl. LSItems 28896

nim0elopez:
#bugfix
Mantis issue: 0028896
Target: release_7
https://mantis.ilias.de/view.php?id=28896

This is the cleanest solution I could come up with.
The main Issue is that when a Learning Sequence is copied, LSItems for every child item of the LS (that is also selected for copy) one LSItem needs to be cloned too, but these LSItems are not handled by the individual child items, they are handled by the parent LS item. The child items themselves don't even know they are associated with an LSItem.

In this approach, the cloneObject Method was modified on the ilObject base class (which all other objects in the tree inherit from) in that at the end of the cloning process the parent is called with a new (per default empty) function "handleChildCloning".
Everywhere else this makes no difference as the function is empty in ilObject. But in the ilObjLearningSequence the function is overridden and the respective LSItem cloning is handled there.
In principle any other Structure that should potentially face a similar issue can now use the same approach here. Override the "handleChildClonig" and implement the required changes there.

One thing I am still not happy about is that, when creating the clone for the LSItem, I use getLSItems(), and iterate over all of them until I find the correct entry. I was unable to find a way to reconstruct the full LSItem directly without getting the entire List, mainly because without the full array I am unable to find out which position the LSItem should have. This value is required for the creation of the LSItem.

### DIFF
--- a/Services/Object/classes/class.ilObject.php
+++ b/Services/Object/classes/class.ilObject.php
@@ -1900,6 +1900,17 @@ class ilObject
             'cloned_from_object' => $this,
         ));
 
+        $tree = $DIC->repositoryTree();
+
+        $parent_nd = $tree->getParentNodeData($this->getRefId());
+
+        include_once './Services/Object/classes/class.ilObjectFactory.php';
+        $factory = new ilObjectFactory();
+        $parentObj = $factory->getInstanceByRefId($parent_nd['ref_id'], false);
+        if ($parentObj) {
+            $parentObj->handleChildCloning($this->getRefId(), $new_obj->getRefId());
+        }
+
         return $new_obj;
     }
     
@@ -2319,5 +2330,15 @@ class ilObject
     public function getPossibleSubObjects($a_filter = true)
     {
         return $this->objDefinition->getSubObjects($this->type, $a_filter);
+    }
+
+    /**
+     * @param $a_target_id
+     * @param $a_copy_id
+     * @return void
+     */
+    public function handleChildCloning($a_target_id, $a_copy_id)
+    {
+        // Kann ggf. überschrieben werden.
     }
 } // END class.ilObject
--- a/Services/Object/classes/class.ilObject.php
+++ b/Services/Object/classes/class.ilObject.php
@@ -2337,7 +2337,7 @@ class ilObject
      * @param $a_copy_id
      * @return void
      */
-    public function handleChildCloning($a_target_id, $a_copy_id)
+    public function handleChildCloning($a_child_ref_id, $a_clone_ref_id)
     {
         // Kann ggf. überschrieben werden.
     }
--- a/Services/Object/classes/class.ilObject.php
+++ b/Services/Object/classes/class.ilObject.php
@@ -1900,14 +1900,12 @@ class ilObject
             'cloned_from_object' => $this,
         ));
 
-        $tree = $DIC->repositoryTree();
 
-        $parent_nd = $tree->getParentNodeData($this->getRefId());
+        $parent_nd = $this->tree->getParentNodeData($this->getRefId());
 
-        include_once './Services/Object/classes/class.ilObjectFactory.php';
         $factory = new ilObjectFactory();
         $parentObj = $factory->getInstanceByRefId($parent_nd['ref_id'], false);
-        if ($parentObj) {
+        if ($parentObj instanceof ilContainer) {
             $parentObj->handleChildCloning($this->getRefId(), $new_obj->getRefId());
         }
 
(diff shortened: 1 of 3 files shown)


Services/Object/classes/class.ilObject.php
```diff
@@ -2320,4 +2331,14 @@ public function getPossibleSubObjects($a_filter = true)
     {
         return $this->objDefinition->getSubObjects($this->type, $a_filter);
     }
+
+    /**
+     * @param $a_target_id
+     * @param $a_copy_id
+     * @return void
+     */
+    public function handleChildCloning($a_target_id, $a_copy_id)
```
klees:
```suggestion
    public function handleChildCloning($a_parent_ref_id, $a_child_ref_id)
```

I think this would clarify what the parameters mean here...

klees (approved):
Hi @elopezmo,

thanks a lot for looking into this contrieved issue. The change is fine for me, but I will need approval from @kergomard for the changes in `Services/Object`.

Best regards!

klees:
@kergomard Any opinion here? If you approve, I will merge an cherry-pick to other releases...

Services/Object/classes/class.ilObject.php
nim0elopez: Good point, I changed the parameter names for clarification, thx!

kergomard:
Sorry, for missing this for so long! It got somehow lost in my inbox.
Thank you very much for the fix @elopezmo .

I've one conceptual question:
- [ ]  Why does this belong in `ilObject`? This for me sounds very explicitly like something that should be in `ilContainer`, as Containers handle their children, Objects (currently in the majority of cases) are not even able to have children.

If we decide that it has to be `ilObject` (I believe we would need very good reasons to do this, but there might be):
- [ ] Please move the usage of $DIC [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5224/files#diff-ff9d6240be485dd464c16fdaab28206db2450556800ef6c171c04631675836b9R1903) to the top of the function to keep them all together and make it more visible.
- [ ] Please remove the include [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5224/files#diff-ff9d6240be485dd464c16fdaab28206db2450556800ef6c171c04631675836b9R1907).

nim0elopez:
Hi @kergomard, thanks a lot for your feedback!
I think you are right and the IlContainer is the right place for this functionality. I will change this accordingly and also take into account your other suggestions.

nim0elopez:
@kergomard @klees just in case, the PR is ready to be reviewed again with the discussed changes :)

kergomard:
Thank you very much @elopezmo !
I'm still not happy though, sorry. From where I stand we should get rid of all changes in ilObject.  I think, we should override cloneObject in ilContainer and call the parent class from there, before then running the whole operations on the children. Am I missing something?

Thank again and best,
@kergomard

nim0elopez:
@kergomard :) That's all right, maybe I am the one missing something. This is how I understand it:
The Learning Sequence is a Container, containing different objects. These child objects themselves are not (necessarily) containers. When the child objects are cloned, the parent container needs to be notified. That is why I am still calling the parent on the object cloning, but defining the handler function (that the LS overrides and other Containers can, too) on the container.

kergomard (approved):
Thanks @elopezmo ! I apparently was asleep at the wheel. You are right and I was missing that part. I still think this somehow smells a little bit, but I don't think the smell comes from your fix ;-).

I assign @alex40724 for the small change in ilContainer.

Best,
@kergomard

nim0elopez:
thank you, @kergomard!

alex40724 (changes requested):
@elopezmo : Thanks for PR.

One question: A few lines before your changes in ilObject the "cloneObject" event is being raised. Couldn't you simply listen for this event in the Learning Sequence component and proceed accordingly?

See also https://docu.ilias.de/goto_docu_pg_43241_42.html

EDIT: Please try using the event handling. This should avoid any changes in Services/Container and Services/Object.

kergomard:
Thank you very much @alex40724 for this hint! I didn't have this on my radar, sorry. @elopezmo : Are you going to adapt this PR correspondingly? If not: I would like to close this to keep my backlog short and manageable.

nim0elopez:
Thank you! I do want to try the latest suggestion out, and adapt this PR accordingly. Will do so within next week :pray:

kergomard:
@elopezmo : I remove myself from this. If the changes in ilObject are really necessary, I kindly ask you to reassign me.

klees:
Merged @nhaagen's iteration. Thanks y'all!

PR #5225, opened 2022-11-18
T&A #0035463 (trunk): Same name in Test export

fhelfer:
https://mantis.ilias.de/view.php?id=35463

swap name for login in export sheet
also some other bugfixes otw

### DIFF
--- a/Modules/Test/classes/class.ilTestExport.php
+++ b/Modules/Test/classes/class.ilTestExport.php
@@ -473,7 +473,7 @@ abstract class ilTestExport
             $col = 0;
 
             // each participant gets an own row for question column headers
-            if ($this->test_obj->isRandomTest() && $firstwritten) {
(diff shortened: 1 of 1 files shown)


dsstrassner:
Thanks for your PR @fhelfer. @kergomard and I talked about your approach. We also take the feedback from Jens Weber in Mantis in credit.

So, I decided to choose the approach to add a duplicate check and take the first 28 characters of the surname & first name, truncate and to add a double-digit counter with a "-" (e.g. Leutheuser-Schnarenberger, M-01 / Leutheuser-Schnarenberger, M-02). If the surname, firstname are shorter than 28 characters, than the only the counter will applied. (e.g. Müller, Peter-01 / Müller, Peter-02)

fhelfer:
I changed it back to the old counting and made the duplicate check take the first 26 letters. With that excel should not run into problems.

kergomard:

PR #5228, opened 2022-11-18
TA: 31983 addendum - cast thumbsize to string for ilShellUtils

nhaagen:
https://mantis.ilias.de/view.php?id=31983#bugnotes


kergomard:

PR #5244, opened 2022-11-22
Util: generate webpath to data/client/

nhaagen:
incorporate relative path and filter surplus slashes.

I'm unsure about the place in ilUtil an the somewhat crowbar approach, but: this works.
If anybody has better solutions for this - you're most welcome.

https://mantis.ilias.de/view.php?id=35707

### DIFF
--- a/Modules/Test/classes/class.ilObjTest.php
+++ b/Modules/Test/classes/class.ilObjTest.php
@@ -5417,7 +5417,7 @@ class ilObjTest extends ilObject implements ilMarkSchemaAware, ilEctsGradesEnabl
     public function getImagePathWeb()
     {
         $relative_path = "/assessment/" . $this->getId() . "/images/";
-        return ilUtil::getDataWebPath($relative_path);
+        return self::getDataWebPath($relative_path);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


kergomard:
Thank you @nhaagen for this PR and sorry, I kept you waiting so long! It somehow fell outside of my field of vision!
Am I getting it right, that the problem you try to solve is the existence or lack of a trailing slashes for path information given be the administrator?
If yes: Which fields may contain them? They are coming from the ini-Files, right?

I see two approaches to this:
- Either we move a more targeted and simplified version of this to the test itself (I would most probably keep it completely localized as two of these functions will be gone anyway with ILIAS8).
- Or we move the clean-up to the actual place these paths(-segments) are brought into the system.

The approach with `ilUtil` / `ilFileUtils` seems to me to have al the disadvantage and non of the advantages of a centralized approach.

What do you think? Don't hesitate to contact me directly, if this is easier.

Thank you very much and best,
@kergomard

dsstrassner:
Hi @kergomard & @nhaagen!
What is the status of this PR?

nhaagen:
I'll move the getDataWebPath from Utils to ilObjTest and mark it as the anoyance it is.

kergomard:
I will merge and cherry-pick to 8 and trunk.
Best,
@kergomard

kergomard:
Ok, my bad, sorry! There seems to be another solution applied in ILIAS8 and above. I will thus not cherry-pick but simply keep this change for ILIAS 7.

PR #5245, opened 2022-11-22
Delos SCSS fixing tag priority colors

catenglaender:
Tag priority colors were broken after SCSS refactoring.This PR includes
* a new tool to generate a map of gradually desaturated colors
* a new tool to choose the higher contrasting color for each color of a map
* the tags (legacy and ui component) now use these tools to generate the tag background color and for choosing the highest contrasting text color on top of it
* a new set of radius variables have been introduced for border radius that should exist even in zero border radius skins (like the appearance of tags)
* some button variables from the same file have been redistributed to lower layers
* remaining variables that are shared between legacy tag and ui component tag have been placed in a clearly named file.


Amstutz:
Thx a lot for fixing and removing the legacy variables!

PR #5246, opened 2022-11-23
SCSS moved bootstrap button code to ui component, deactivate bootstrap button file

catenglaender:
Buttons used to be created with Bootstrap 3 defaults and then get overriden by our styling in the ui component button causing redundant CSS code.

With this PR...
* code for all standalone button types used by ILIAS from bootstrap's _button.scss moved into the ui component button, the original Bootstrap 3 button file is no longer imported.
* unused default values are left out so instead of overriding them, buttons are generated with the ILIAS styling right from the beginning
* buttons not used by ILIAS are left out (btn-warning, btn-info, btn-success, btn-danger, btn-block)
* results in 200 lines of CSS code less

This is still far from buttons being independent from Bootstrap 3 as the old mixins and variables are still being used, but it's a start.


Amstutz:
Thx a lot! Every CSS line in the delivered delos.css we can remove counts! Kudos.

PR #5247, opened 2022-11-23
A11y/UX, change dropdown and empty heading to link in listing panel

Amstutz:
This is the long time ago promised follow up of: https://github.com/ILIAS-eLearning/ILIAS/pull/2943 ( see JF decision of the second of May):

@yvseiler and @Amstutz finally found time to look for a solution. Since we would like to fix especially the a11y issue for ILIAS 8 as well, we decided to opt. For a solution that is as minimally invasive as possible

The solution in the PR resolves to main issues:
- A11y: Empty Heading issue in Listing Panel, see: https://mantis.ilias.de/view.php?id=33884, and https://mantis.ilias.de/view.php?id=32356 . Note that this is a WCAG AA fail and would prevent passing an audit.
- UX: Currently we hide a single option in a dropdown, even though there cannot be more one action currently there in any setting and there is more than enough space to take the link out of the dropdown.

Currently the favourites list looks like this:
image

With this change we would change it to this look:
image

There is no change to the look if the list is empty.

### DIFF
--- a/Services/Repository/Favourites/classes/class.ilFavouritesListGUI.php
+++ b/Services/Repository/Favourites/classes/class.ilFavouritesListGUI.php
@@ -65,15 +65,15 @@ class ilFavouritesListGUI
             $ctrl->setParameterByClass("ilPDSelectedItemsBlockGUI", "view", "0");
             $ctrl->setParameterByClass("ilPDSelectedItemsBlockGUI", "col_side", "center");
             $ctrl->setParameterByClass("ilPDSelectedItemsBlockGUI", "block_type", "pditems");
-            $panel = $f->panel()->secondary()->listing("", $item_groups);
-            $panel = $panel->withActions($f->dropdown()->standard([$f->link()->standard(
-                $this->lng->txt("rep_configure"),
-                $ctrl->getLinkTargetByClass(
-                    ["ilDashboardGUI", "ilColumnGUI", "ilPDSelectedItemsBlockGUI"],
-                    "manage"
+
+            $config_item = $f->item()->standard(
+                $f->link()->standard(
+                    $this->lng->txt("rep_configure"),
+                    $this->ctrl->getLinkTargetByClass(["ilDashboardGUI", "ilColumnGUI", "ilPDSelectedItemsBlockGUI"], "manage")
                 )
-            )
-            ]));
+            );
+            array_unshift($item_groups, $f->item()->group($this->lng->txt(""), [$config_item]));
+            $panel = $f->panel()->secondary()->listing("", $item_groups);
             return $this->ui->renderer()->render([$panel]);
         }
 


oliversamoila:
Hello @Amstutz , 
a little comment on this: With ILIAS 8 you can also switch to alphabetical sorting of favorites. There are no subheadings (as with sorting "by location" or "by type"). See [here](https://docu.ilias.de/goto_docu_wiki_wpage_6358_1357.html).
I assume that the direct sequential order of "Configure", "Object 1" ... "Object n" will be difficult to distinguish from each other. But we can have a look at it when the time comes. Unless you already have an approach how to separate this.
Many thanks to you for the improvements !!

Amstutz:
Thx for the hint. Uff. we have far, to many features and settings. 

Here is how this looks like with this setting active:
image

alex40724 (approved):
Change is fine for me. The original JF decision says "Timon will create a new PR and discuss it with the product manager." So if the product manager is fine with this, no additional JF discussion may be needed.

alex40724:
@Amstutz If you got the ok from the PM, I would merge this. Of course discussing this in the JF is fine for me, too. Anyway, I would like to merge this in the end.

Thanks for the PR.

oliversamoila:
Thanks for the explicit image @Amstutz 
In the discussions about the White Background we always talked about so-called "hairlines" for visual separation. This is also the case between the bulky links. Such a dividing line should at least be included.

Amstutz:
@alex40724 , thx for the feedback! I requested a feedback from @matthiaskunkel and gave him a short note to look over it.

@oliversamoila thx for the feedback concerning the design of the alphabetical ordering. I will look into this with @yvseiler.

alex40724:
Maybe the CSS hairline fix can be done separately, since this focuses currently on what KS elements are used by the Repository/Favourites service.

matthiaskunkel:
Thanks to @Amstutz for this PR. I am fine with it. So please merge it. No need to put it again on the JF agenda. 
But I would like recommend one single layout change for the presentation of the shy button "Konfigurieren": Let's place it right-aligned to avoid the misunderstanding that it is something similar to the following items. And it would appear at the same place where the dropdown was before. I guess this would be much more user-friendly. Thanks.

Amstutz:
@matthiaskunkel , thx for the feedback. Right aligned placement is not an option here, since we introduced this option as exactly what it looks like. Just another item in this list, without icon. So, there is no handle to grab this non-special item here and place it differently. The hairline is possible, and is just missing since, we placed this item in a separate group without title. It should be easy to fix on a CSS level.

We propose this in this fashion make as little change due to the stage late in beta as possible.

yvseiler:
Thanks for your design suggestions. I have now weighed up different possibilities (alignments, with hairline or with background colour):

I recommend the use of a hairline at the bottom (same as in list; border-bottom: 1px solid #ddd), which corresponds to @oliversamoila 's suggestion. This is because it follows the alignment of the other entries, stands out from the objects further down and at the same time makes it clear that I somehow belong to this list. 

I understand your concerns @matthiaskunkel . But for better differentiation from the favourites, I recommend working on the UX here in the context of the language (label) and not visually. In my opinion, the change from "Configure" to "Edit" already helps a lot so that this entry is better read as an action than as an object in the favourites list.

alinaseibt:
is it a possibility to lower the font-size of the shy button a bit, to differ it from the lists objetcs? The font-size seems to be the same as the objects titles and as Matthias said, "Konfigurieren" somehow seems to be just another favourites' object, but without an icon. I don't know if adding a hairline is enough differenciation.

Amstutz:
@alinaseibt , no. see comment above. in this proposal it is just a regular item in the list, just without item. no proper way to style it by css differently than the other items.

PR #5249, opened 2022-11-23
SCSS modern flex grid

catenglaender:
This PR brings a modern flexbox grid system (and potentially also its CSS grid alternative) into our codebase. Large parts come from Bootstrap 5 and are modified to hook into our variables. I had to customize it much less than I originally anticipated, which opens the question if we would rather want to add this as a dependency. It's a very solid modern grid system and very compatible to the one we currently have.
* class names are the same as the current Bootstrap 3 grid we are using in the Page Editor and in many html templates
* this uses flexbox instead of floats (well, actually our Bootstrap was already patched to also use flexbox, but that was a bit rough)
* there are a few new options for columns:
    * rows can define how many columns they have nested in them side by side instead of defining sizes on the columns individually
    * since this is flexbox, new grow/shrink/auto rules and utility classes can be used to dynamically fill up rows.
* because of these new options, this change adds css lines
* the css grid is currently not being generated. Since we never have boxes with varying paddings, the pre-build css grid might not bring any advantages over the flexbox grid.

## To Dos

- [ ] Do we want to add this (and in extension meaning a very selected, small part of Bootstrap 5) as a dependency, so updates from upstream are easier to diff?
- [x] The new grid reveals that the row class was misused to generate negative margins e.g. the "row" inside the "mainspacekeeper" with "fixed_content" inside which is not actually meant to be a column leaving it with -15px to both side (also "mainspacekeeper" is probably no real "container-fluid" anymore either and might be useful in fixing this)


catenglaender:
This changes how the make-col mixin works to the way it has evolved to in Bootstrap 5:

```SCSS
// no longer suported
.class {
    @include .make-col-sm-7;
}


// new way of doing this
.class {
    @include media-breakpoint-up(sm) {
        @include make-col(7);
    }
}
```

catenglaender:
The grid layout padding system uses (and has always used - even before this PR) a negative margin on the row that gets reset by adding the same amount as padding to the columns. This -margin +padding system should not be overridden on neither the row nor the column unless you override both equally. Unfortunately, we didn't stick to this rule everywhere and these places now come to the surface with the new grid being rightfully less forgiving as rows now push this system on their content.

The following things cause visual glitches now, which is the correct response for places where the row col system is used not quite as it was intended:
* .col-* classes not inside a .row
    * this is not allowed with this grid system anymore
    * It has been misused to create column layout with paddings in the past, but the padding should come from a surrounding container/panel instead
* .row classes but inside there is no .col-* class
    * this is allowed but now correctly creates no visual margin asmargin/padding compensation is applied automatically to any direct child (> *)

My recommendation: If we want padding for rows, we should add a container/panel with padding around the row. If we want padding inside a column, we should add a container/panel with padding inside the col-*. Or don't use this grid system and style semantic class names instead.

I fixed the two most obvious broken places:
* login panel
* legacy repository object lists

But I suspect there are more. Most of them can probably be fixed by either
* deleting the "row" class in the html template if there are no real columns inside of it
* adding a row around parentless col-* divs and a container with padding where necessary.

For now, I think this PR is ready to be merged, but I expect more discoveries where rows and cols were used incorrectly.

Amstutz:
Thx a lot for the proposal.

> If we want padding for rows, we should add a container/panel with padding around the row. If we want padding inside a column, we should add a container/panel with padding inside the col-*. Or don't use this grid system and style semantic class names instead.

I completely agree with that.

Note to everybody concerned, if we run into issue due to the change in the behavior concerning the padding. Please assign a ticket to me.

PR #5252, opened 2022-11-24
Test: Fix Updating of Test-Results Cache

kergomard:
See: https://mantis.ilias.de/view.php?id=35452

I would really love a second pair of eyes on this. From where I stand the original code is simply wrong on so many levels, but `assQuestion::_updateTestResultCache` should be a very central method. I'm at a loss. Thank you very much!


nhaagen:
I agree that the code is somewhat messy, but your changes look very sound.

kergomard:
Ok. I merge this. Cherry-pick it to trunk and see what needs to be backported to 7. Thanks @nhaagen for the second opinion.

PR #5256, opened 2022-11-24
Bugfix 10440 missing lang vars crs_visibility and crs_visibility_unvisible

sasfalk:
#bugfix
Mantis issue: 0010440
Target: release_7
https://mantis.ilias.de/view.php?id=10440

I added the two variables crs_visibilty and crs_visibility_univisible to the lang files. These are only called when creating remote courses, so I could not test if the variables are still needed at all in Ilias7 or higher.


matthiaskunkel:
Thanks @sasfalk for this PR. I merge it to release_7 and cherry-pick in 8 and trunk.

PR #5257, opened 2022-11-24
Bugfix 31142 root_entries[i] is undefined

sasfalk:
#bugfix
Mantis issue: 0031142
Target: release_7
https://mantis.ilias.de/view.php?id=31142

### DIFF
--- a/src/UI/templates/js/MainControls/dist/mainbar.js
+++ b/src/UI/templates/js/MainControls/dist/mainbar.js
@@ -539,7 +539,7 @@ var model = function() {
                 last_entry_id = entry_ids[entry_ids.length - 1],
                 more = state.entries[last_entry_id];
 
-            if(state.any_tools_visible()) {
+            if(state.any_tools_visible() && max_buttons > 0) {
                 max_buttons--;
             }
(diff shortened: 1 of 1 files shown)


klees:
Hi @sasfalk,

thanks a lot for looking into this. I don't think that this is a correct fix for two reasons:

* You can't simply change the code in the dist-folder, as the is the location where the code is compiled to. That means that this code will vanish when the source code in the src-folder will be compiled the next time. You may want to have a look into the according [docu](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/docs/development/js/js-bundling.md) and the [config](https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/src/UI/templates/js/MainControls/rollup.config.js). Feel free ask if there's something I shall elaborate.
* When looking into the actual source for the mainbar, you might notice that your fix now spans two modules, the model and the renderer. I think [this fix](https://github.com/ILIAS-eLearning/ILIAS/commit/1bfd7875f99395286e91f0ec6581306e048a428e) is a lot clearer, as it is only in one location. Also, this in fact is a rendering problem...

This is why I close that PR. Thanks for making me look =)

Best regards!

PR #5243, opened 2022-11-22
OrgU: Repo Pattern for Positions and Authorities

tbongers-cat:
As proposed by @klees 
See commit message for more details.

PR #5258, opened 2022-11-24
OrgU: Bugfixes for orgunit-specific permissions (release_8)

tbongers-cat:
This fixes [Mantis #35746](https://mantis.ilias.de/view.php?id=35746) for release_8.

This fix is also integrated in PR #5243 for trunk.


tbongers-cat:
Added a fix for [Mantis #34298](https://mantis.ilias.de/view.php?id=34928)

nhaagen (approved):

PR #5264, opened 2022-11-28
TA: 34411, preserve nesting when editing question in 'regular' form

nhaagen:
https://mantis.ilias.de/view.php?id=34411


kergomard:
Thank you very much @nhaagen for the fix!
This line [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5264/files#diff-6471e9b49d762f0f28cf576c8a0911241debb196aebfe477f3f93a73a940b5a8R185) gives me the hibbie jibbies, though. Could we maybe solve this a little more elegantly ;-).

nhaagen:
:) ouh, sorry.
This is still not refinery, but better?

kergomard:
Yes, thank you!

PR #5265, opened 2022-11-28
TA: 34411, prevent whoops on re-submitting after failed checkInput

nhaagen:
https://mantis.ilias.de/view.php?id=34411
this is not a real "fix", but rather hides the whoops when re-submitting a from that previously failed on checkInput.
The values are still not preserverd ;(
Please reject/close, if this feels awkward.


kergomard:
Thank you very much @nhaagen . I had a look at this a while ago and came to this point, too. This "fix" makes every bone in my body hurt, but if we don't find a better solution in a reasonable amount of time, I would still apply it.
Thus: I will keep this PR open for the time being, as a memento and in the hope that we do not need to merge it, but if necessary, I think we should.

PR #5268, opened 2022-11-28
[FIX] CI: copyright-checker now ignores UI examples.

thibsy:
Hi all

UI code examples located at `src/UI/examples` are always violating the copyright-checker because they do not contain the copyright license header at the top of each file. The reason for this is, that these snippets are displayed in the UI documentation **as is**. The documentation would get rather crowded if all examples would contain this license header, therefore it was agreed to leave them out.

I adjusted the copyright-checker to ignore files containing the substring `src/UI/examples` as done for the `./libs` directory.
This change would need to be picked for release_7 & release_8 as well.

@mjansenDatabay I know you're not officially responsible for this, feel free to reassign.

Kind regards!

### DIFF
--- a/CI/Copyright-Checker/copyright-checker.sh
+++ b/CI/Copyright-Checker/copyright-checker.sh
@@ -56,6 +56,10 @@ do
     continue
   fi
 
+  if [[ $FELONE == "./src/UI/examples" ]]; then
+    continue
+  fi
+
   # check for php extension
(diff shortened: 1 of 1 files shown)


Amstutz:
Note, the UI Coordinators all agree, that we want to exclude the examples folder. @thibsy thx a lot for implementing this. This was gaining dust on my personal todo list...

mjansenDatabay:
This LGTM, thx a lot @thibsy .

PR #5270, opened 2022-11-28
[FIX] #35763: fixed purification on NULL when no feedback was found.

thibsy:
Hi all

We found a bug (see https://mantis.ilias.de/view.php?id=35763) where `null` was being purified. This should fix the issue by checking the correct array-key with `array_key_exists`. I also initialized `$feedbackContent` properly.

Kind regards


kergomard:
Thank you very much @thibsy . I was apparently getting tired, when fixing this!

PR #5271, opened 2022-11-28
[FIX] Style: ilSystemStyleSkinContainerTest::testImportSkin

thibsy:
Hi @Amstutz 

We came across this failed test when **only unzip** has been installed on the system. Apparently, unzip requires zip to be installed as well, otherwise the operation fails. Therefore this unit test failed as well, because it ist only skipped if unzip isn't available.

Kind regards!


Amstutz:
Thx!

Amstutz:
picked to 8 and trunk as well.

PR #5276, opened 2022-11-30
SCSS panel more independent from bootstrap

catenglaender:
This PR makes the style code for panels more independent from Bootstrap:
* relevant code has been migrated from the bootstrap files into the legacy and ui component panel files.
* redundancies have been removed
* style code for panel-primary and panel-default is no longer compiled as they were styled to look exactly like a normal panel.
* reduces compiled CSS by 200 lines
* didn't use the new naming convention yet as the panel is tightly connected to the legacy panel and some js.

We might want to consider stopping the legacy panel from generating panel-primary and panel-default renderings and just use the panel CSS class only instead. However, I didn't dare to make that change as there are many places where these legacy panels are used. For now, they are still generated with the now empty CSS classes.


Amstutz:
Thx a lot!

PR #5278, opened 2022-12-02
[fix] for missing language entry

kunzeus:
Replace hard coded text with ILIAS  language entry

### DIFF
--- a/Modules/Test/classes/class.ilTestPlayerAbstractGUI.php
+++ b/Modules/Test/classes/class.ilTestPlayerAbstractGUI.php
@@ -1197,7 +1197,7 @@ abstract class ilTestPlayerAbstractGUI extends ilTestServiceGUI
 
         $this->tpl->setVariable(
             'LOCKSTATE_INFOBOX',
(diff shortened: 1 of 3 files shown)


kunzeus:
Also in ILIAS 7

dsstrassner:
Thanks a bunch @kunzeus. 

Fixes https://mantis.ilias.de/view.php?id=35548

kergomard:
I would have a little change request, though:
- Would you please be so kind as to change the name of the language-variable to 'tst_player_answer_saved_and_locked'. From where I stand the important information is the 'and locked' part and it should not be used if an answer is just saved.

mbecker-databay:
I had a faint memory the length of the identifier was important to look at due to the db field very short, but I see in current versions that it's set as varchar(200), so I second @kergomard s wish.

kunzeus:
Sorry for the wait, I was on sick leave.

kergomard:
Thank you very much @kunzeus ! I hope you are feeling well again. Will merge and cherry-pick to 8 and 7.

PR #5279, opened 2022-12-02
IRSS: Migration of Profile-Pictures

chfsx:
This PR is the first part of the migration of Profile-Picture into the IRSS.

It includes in the individual commits:
- The actual migration of the files: https://github.com/ILIAS-eLearning/ILIAS/pull/5279/commits/81f1b323d6760b0f4c08b325b8447a8c5fe9d0a7
- The usage of the migrated file (with fallback to the actual file, if not migrated yet): https://github.com/ILIAS-eLearning/ILIAS/pull/5279/commits/b4a1df1ff69b2b091f8976671aafd26d04854f77

The migration was performed as described in https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Services/ResourceStorage/MIGRATIONS.md. The following things are relevant for this:

- The usr_data table knows a new column rid: https://github.com/ILIAS-eLearning/ILIAS/pull/5279/files#diff-48da449156e9caffbfa0e3becf2e599003ae441222b4b5dd6d38805b1ff9d02cR41
- The migration is in https://github.com/ILIAS-eLearning/ILIAS/pull/5279/files#diff-48da449156e9caffbfa0e3becf2e599003ae441222b4b5dd6d38805b1ff9d02cR41https://github.com/ILIAS-eLearning/ILIAS/pull/5279/files#diff-48da449156e9caffbfa0e3becf2e599003ae441222b4b5dd6d38805b1ff9d02cR41
- The following new stakeholder is used for the use case:https://github.com/ILIAS-eLearning/ILIAS/pull/5279/files#diff-944cb99f76cc718f6197efc014e2f34fa0d95c7864fd0e35d683175ea1afff8eR26

You could already review these first steps before the next step follows:

- Many uses of the profile picture need to be slightly adjusted
- With the new flavours different resolutions of the profile pictures are created and used
- The upload options of profile pictures will be adjusted
- Import/export will be adapted

The work is already prepared, but to keep the PR lean and so that we can still discuss a few points in the implementation, these points follow only afterwards.

There has been some confusion as to who should review this PR now (@smeyer-ilias  or @kergomard ). However, since @kergomard is now responsible, I am assigning him for now. 

As always, let me know if we want to look at the PR (and also the next steps) together via discord for example.


chfsx:
@kergomard I rebased on trunk and force-pushed to stay up to date here. followup will be in https://github.com/srsolutionsag/ILIAS/pull/15

kergomard (commented):
Hi @chfsx 

Thank you very much for this improvement!

I have a couple of questions just to understand it better:
- We could also get rid of [this](https://github.com/ILIAS-eLearning/ILIAS/pull/5279/files#diff-3ff6df716bc6e703ff95d536542b3090fbda388e6ca05867d6d6c7078f4e453dR40), right?
- If I'm understanding this correctly, the migration does not yet remove the pictures, but copies them, right?

I assume the next steps concerning these two questions would be part of your plan laid out above.

Thanks again and best,
Stephan

chfsx:
Hi @kergomard I updated the PR after our meeting some days ago. Thanks a lot for looking into this

kergomard (changes requested):
Hi @chfsx 
Thank you very much for this PR! I like this a lot and have only a few small suggestions and things that are not working yet.

- [x] The [`ilUserProfilePictureDefinition::isGeyscale()`](https://github.com/ILIAS-eLearning/ILIAS/pull/5279/files#diff-6407817af2227387df2a905b53bae78899f5ab1ae4b2b46cc979bf07efdaad83R84) and [`ilUserProfilePictureMachine::makeGreyscale()`](https://github.com/ILIAS-eLearning/ILIAS/pull/5279/files#diff-027de37d0a18d57c37336ba0817b7df2d1f6e854c820539747b5ea7b73c5eda0R115) functions are only there for "completeness", right? I would suggest to remove them for now, if there is no very good reason against it, as they suggest that there actually is an option to generate greyscales, but there isn't. I think this would then also obsolete [this](https://github.com/ILIAS-eLearning/ILIAS/pull/5279/files#diff-6407817af2227387df2a905b53bae78899f5ab1ae4b2b46cc979bf07efdaad83R64).


- [x] Migration failed because `SYSTEM_USER_ID` is undefined in [`ilUserProfilePictureStakeholder::__construct()](https://github.com/ILIAS-eLearning/ILIAS/pull/5279/files#diff-ca34421f89cedcf161f0a82d0f0fe2e2207508520f84111428d74dd4e94b34a7R33), at least when I tested it, but I think this must always be the case, as $DIC->user()->getId() will be undefined and ILIAS not fully instantiated.
- [x] As far as I can see, the functions [`ilUserAvatarResolver::getLegacyPictureURL()`](https://github.com/ILIAS-eLearning/ILIAS/pull/5279/files#diff-9bff8443aa5dc07bf3e2e07aeffdca07634c51984e10bed06acf74d25826912aR187) and [`ilUserAvatarResolver::getAvatar()`](https://github.com/ILIAS-eLearning/ILIAS/pull/5279/files#diff-9bff8443aa5dc07bf3e2e07aeffdca07634c51984e10bed06acf74d25826912aR151) lack the checks for `ilUserAvatarResolver::force_image`.

Thanks again and best!
@kergomard

chfsx:
Hi @kergomard 
Thanks a lot for reviewing this!

- As desired, I removed the unused greyscale option
- Migrations is now more robust concerning dependecies
- force_image is already in https://github.com/ILIAS-eLearning/ILIAS/pull/5279/files#diff-9bff8443aa5dc07bf3e2e07aeffdca07634c51984e10bed06acf74d25826912aR107, but it's hart to read: if force_image is true, hasProfilePicture returns false if there is no irss nor legacy picture (otherwise it's impossible to return an picture)
There was also an issue when exporting users, this should be fixed now

PR #5280, opened 2022-12-03
Various DC PHP 8 Fixes

Amstutz:
Since I had to fix an a11y involving Icons in the DC, I needed to fix several things first in order to even check if my Icon fix works in the GUI. I labelled this a "Bugfix" without linking a specific one, since this targets a whole range of issues and should de-block a rather larger set of test cases.

With the here proposed fixes a was able to (PHP8):
-> Open an existin DC from Repo
-> Create a new DC
-> Create new Tables
-> Add Fields of to a DC (Text, Boolean, Reference and Formula has been tested, others are still with errors)
-> Change Fields in a DC (not all Types, see a above)
-> Use Formula Fields with Place Holder / Fix JS Loading
-> Add Entries (not all Types, see a above)
-> Change Entries  (not all Types, see a above)
-> Add Views to a DC 
-> Change Views in a DC
-> Add Detailed views to a Field
-> Look at detailled view in a DC

Note that overall state of the DC still remains very unstable. The migration performed by @mstuder, seems to have never been tested on the frontend. Many parts of the performed migration are broken or only partly done in order to make automated checks (style guide, MUST criteria of the projects etc) and the code review pass. 

I hope this helps somewhat. I would propose to merge this and then improve from this point on. Just picking individual improvements would also be an option.

### DIFF
--- a/Modules/DataCollection/classes/Fields/Base/class.ilDclBaseFieldModel.php
+++ b/Modules/DataCollection/classes/Fields/Base/class.ilDclBaseFieldModel.php
@@ -27,10 +27,7 @@
  */
 class ilDclBaseFieldModel
 {
-    /**
-     * @var int|string $id int for custom fields string for standard fields
-     */
-    protected $id = 0;
+    protected string $id = "";
     protected int $table_id = 0;
     protected string $title = "";
     protected string $description = "";
@@ -115,9 +112,8 @@ class ilDclBaseFieldModel
 
     /**
      * Get field id
-     * @return int|string
      */
-    public function getId()
+    public function getId(): string
     {
         return $this->id;
     }
@@ -570,7 +566,7 @@ class ilDclBaseFieldModel
     /**
      * @param mixed $value
      */
-    protected function normalizeValue($value): string
+    protected function normalizeValue($value): ?string
     {
         if (is_string($value)) {
             $value = trim(preg_replace("/\\s+/uism", " ", $value));
--- a/Modules/DataCollection/classes/Fields/Base/class.ilDclBaseRecordFieldModel.php
+++ b/Modules/DataCollection/classes/Fields/Base/class.ilDclBaseRecordFieldModel.php
@@ -28,7 +28,7 @@
  */
 class ilDclBaseRecordFieldModel
 {
-    protected int $id = 0;
+    protected ?int $id = null;
     protected ilDclBaseFieldModel $field;
     protected ilDclBaseRecordModel $record;
     protected ?ilDclBaseRecordRepresentation $record_representation = null;
@@ -361,7 +361,7 @@ class ilDclBaseRecordFieldModel
         return $this->field;
     }
 
-    public function getId(): int
+    public function getId(): ?int
     {
         return $this->id;
     }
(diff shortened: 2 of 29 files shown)


Modules/DataCollection/classes/Fields/Base/class.ilDclBaseFieldModel.php
```diff
@@ -27,10 +27,7 @@
  */
 class ilDclBaseFieldModel
 {
-    /**
-     * @var int|string $id int for custom fields string for standard fields
-     */
-    protected $id = 0;
+    protected string $id = "";
```
pascalseeland, line 30: If this is typed in any way, I would suggest using int. The constructor as well as the doRead rely on it being an int and the column in the db is int(11)

Modules/DataCollection/classes/Fields/Base/class.ilDclBaseFieldModel.php
```diff
@@ -570,7 +566,7 @@ public function checkValidity($value, ?int $record_id = null): bool
     /**
      * @param mixed $value
      */
-    protected function normalizeValue($value): string
+    protected function normalizeValue($value): ?string
```
pascalseeland, line 569: I would prefer to drop the return typing all together. Looking at that release7 code,  the intention was any value that is not a string gets passed through this method without being altered. In release8 anything that is not an string does not get checked but gets returned as a string. That also means that if it gets called with anything like an array an exception will be thrown

Modules/DataCollection/classes/Fields/Base/class.ilDclBaseRecordFieldModel.php
```diff
@@ -28,7 +28,7 @@
  */
 class ilDclBaseRecordFieldModel
 {
-    protected int $id = 0;
```
pascalseeland, line 31: I'm not so sure that this will break somewhere down the line. So I would like to avoid it. Can you point out at which places this is needed?

Modules/DataCollection/classes/TableView/class.ilDclTableView.php
```diff
@@ -472,7 +488,7 @@ public static function getCountForTableId(int $table_id): int
      * @param bool $create_default_settings
      * @return ilDclTableView|ActiveRecord
      */
-    public static function createOrGetStandardView(int $table_id, bool $create_default_settings = true): ActiveRecord
+    public static function createOrGetSatndardView(int $table_id, bool $create_default_settings = true): ActiveRecord
```
pascalseeland, line 491: This does not look like you wanted to break spelling

Modules/DataCollection/classes/class.ilObjDataCollectionGUI.php
```diff
@@ -223,7 +224,6 @@ public function executeCommand(): void
                 $this->prepareOutput();
                 $this->tabs->activateTab("id_records");
 
-                $tableview_id = 1;
```
pascalseeland, line 226: I did rework this by only setting a tableview_id if there actually is one. There is no need to define one in ildclrecordlistgui and to my understanding the issue is actually created by setting a default value like 1.

pascalseeland (changes requested):
Thanks for all the work that went into this pr. I did add some comments, but some are that I feel that having nulls might cause other issues. I did already fix some read methods that did fail on test8 where the database contains null values that do not exist on newly created dcls

pascalseeland:
Merged. Somehow I still manage to mess it up so github does not realizes it

PR #5281, opened 2022-12-05
Feature Forum (ILIAS 9): Rearrange Actions in Forum  

fhelfer:
Feature Wiki: https://docu.ilias.de/goto_docu_wiki_wpage_7462_1357.html


Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -5626,27 +5613,21 @@ static function (string $id) use ($formID): string {
...
-                    $sb_item->setUrl($url);
-
-                    $action_button->addMenuItem(new ilButtonToSplitButtonMenuItemAdapter($sb_item));
+                    $items[] = $this->uiFactory->button()->shy($this->lng->txt($lng_id), $url);
                 }
             }
-
-            $tpl->setVariable('COMMANDS', $action_button->render());
+            $action_menu = $this->uiFactory->dropdown()->standard($items);
+            $tpl->setVariable('COMMANDS', $this->uiRenderer->render([$action_button, $action_menu]));
```
fhelfer: @mjansenDatabay Do you know why the menu and button do not get rendered on the same line?

mjansenDatabay: Let's have a look tomorrow

fhelfer: new css class introduced

mjansenDatabay:
@Amstutz : @fhelfer had to change the forum specific scss file. Is this okay for you? AFAIK there are new guidelines which state, that changes to the delos css (which will be changed implicitly) have to be approved by the coordinator.

Modules/Forum/classes/class.ilObjForumAdministrationGUI.php
```diff
@@ -139,6 +140,7 @@ protected function populateForm(ilPropertyFormGUI $form): void
         $frma_set = new ilSetting('frma');
 
         $form->setValuesByArray([
+            'forum_enable_print' => (string) $this->settings->get('forum_enable_print', '0'),
```
mjansenDatabay: IMO we should cast the result to `(bool)` because it is a checkbox element, right?

Modules/Forum/classes/class.ilObjForumAdministrationGUI.php
```diff
@@ -159,6 +161,10 @@ protected function getSettingsForm(): ilPropertyFormGUI
         $form->setFormAction($this->ctrl->getFormAction($this, 'saveSettings'));
         $form->setTitle($this->lng->txt('settings'));
 
+        $print_cb = new ilCheckboxInputGUI($this->lng->txt('frm_enable_print_option'), 'forum_enable_print');
+        $print_cb->setInfo($this->lng->txt('frm_enable_print_option_desc'));
```
mjansenDatabay:
```suggestion
        $print_cb = new ilCheckboxInputGUI($this->lng->txt('frm_enable_print_option'), 'forum_enable_print');
         $print_cb->setValue('1');
        $print_cb->setInfo($this->lng->txt('frm_enable_print_option_desc'));
```

Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -2967,7 +2967,9 @@ public function viewThreadObject(): void
             ->current();
 
         $additionalDataExists = $toolContext->getAdditionalData()->exists(ForumGlobalScreenToolsProvider::SHOW_FORUM_THREADS_TOOL);
-        if (!$additionalDataExists && $this->selectedSorting === ilForumProperties::VIEW_TREE) {
+        $subtree_nodes = $this->objCurrentTopic->getPostTree($firstNodeInThread);
```
mjansenDatabay, line 3008: 1. If we add more checks before we display the tree in the tool slate, we should move this below the access control check (`ensureThreadBelongsToForum` IMO.
2. If you move the retrieval of the nodes, you also have to move `$this->objCurrentTopic->setOrderField($orderField);`. This means you have to move even more, because `$orderField` (should be renamed to `$order_field`) to be move as well (and so on) :-).

Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -3114,15 +3113,23 @@ public function viewThreadObject(): void
...
+                        $this->ctrl->getLinkTarget($this, 'createTopLevelPost', 'frm_page_bottom')
+                    );
+                }
                 $this->ctrl->clearParameters($this);
                 array_unshift($bottom_toolbar_split_button_items, $reply_button);
             }
 
             // no posts
-            if ($firstNodeInThread->getId() === 0 && !$numberOfPostings = count($subtree_nodes)) {
+            if (!$numberOfPostings && $firstNodeInThread->getId() === 0) {
```
```suggestion
            if ($numberOfPostings === 0 && $firstNodeInThread->getId() === 0) {
```

Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -3312,9 +3302,17 @@ static function (ilForumPost $post) use ($draft): bool {
             $threadContentTemplate->setVariable('TOOLBAR_BOTTOM', $bottom_toolbar->getHTML());
         }
 
-        $this->renderViewModeControl($this->selectedSorting);
-        if ($this->selectedSorting !== ilForumProperties::VIEW_TREE) {
-            $this->renderSortationControl($this->selectedSorting);
+        if ($bottom_toolbar_split_button_items !== []) {
+            foreach ($bottom_toolbar_split_button_items as $key => $item) {
```
mjansenDatabay: Instead of `$key` we should use `$component_index`.

Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -3312,9 +3302,17 @@ static function (ilForumPost $post) use ($draft): bool {
             $threadContentTemplate->setVariable('TOOLBAR_BOTTOM', $bottom_toolbar->getHTML());
         }
 
-        $this->renderViewModeControl($this->selectedSorting);
-        if ($this->selectedSorting !== ilForumProperties::VIEW_TREE) {
-            $this->renderSortationControl($this->selectedSorting);
+        if ($bottom_toolbar_split_button_items !== []) {
```
mjansenDatabay: Maybe we should not use the name `$bottom_toolbar_split_button_items` anymore, if we render the elements to the bottom toolbar *and* the regular toolbar here.

mjansenDatabay (changes requested):

Amstutz:
Hi @mjansenDatabay 

Thx a lot for the ping. Sry, this slipped my radar completely. Best Assign me if you need something from me to proceed, otherways this might fall under the table. This happened, sry for that.

The changes are fine for me. 
thx.

Amstutz (approved):

PR #5285, opened 2022-12-06
T&A #0035463 (trunk): Same name in Test export 

fhelfer:
https://mantis.ilias.de/view.php?id=35463#c88384

### DIFF
--- a/Modules/Test/classes/class.ilTestExport.php
+++ b/Modules/Test/classes/class.ilTestExport.php
@@ -805,11 +805,12 @@ abstract class ilTestExport
                     ? $userdata->getName()
                     : "ID $active_id";
                 $username = substr($username, 0, 26);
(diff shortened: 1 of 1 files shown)


fhelfer:
adds to #5225 , because issue was still not resolved

kergomard:

I have one little request again, sorry:

- [ ] Could you please reduce the amount of `strtolower()` needed? I would think we need exactly one.

fhelfer:
> Thank you very much @fhelfer !
> 
> I have one little request again, sorry:
> 
> * [ ]  Could you please reduce the amount of `strtolower()` needed? I would think we need exactly one.

@kergomard  less strtolower, new variable to keep original username for sheet naming

kergomard:
Thanks again @fhelfer !

PR #5286, opened 2022-12-06
Add kitchen sink element: LightboxCardPage

lscharmer:
This PR adds a new element `LightboxCardPage` to the UI Framework.
This PR is based on the discussions involving PR: #4932.

### DIFF
--- a/src/UI/Component/Card/Card.php
+++ b/src/UI/Component/Card/Card.php
@@ -70,6 +70,16 @@ interface Card extends Component, JavaScriptBindable, Clickable
     public function getSections(): array;
 
     /**
+     * @param \ILIAS\UI\Component\Component[] $sections
+     */
+    public function withHiddenSections(array $sections): Card;
+
+    /**
+     * @return \ILIAS\UI\Component\Component[] $sections
+     */
+    public function getHiddenSections(): array;
+
+    /**
      * Set the image of the card
      */
     public function withImage(Image $image): Card;
--- a/src/UI/Component/Modal/Factory.php
+++ b/src/UI/Component/Modal/Factory.php
@@ -23,6 +23,7 @@ namespace ILIAS\UI\Component\Modal;
 use ILIAS\UI\Component;
 use ILIAS\UI\Component\Image\Image;
 use ILIAS\UI\Component\Modal\InterruptiveItem;
+use ILIAS\UI\Component\Card\Card;
 
 /**
  * Interface Factory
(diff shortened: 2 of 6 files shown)


thibsy:
Hi @lscharmer

Sorry for letting you wait for some time now. Normally we try to get feedback to you within about 10 days. However, the TB decided to give the release 8 priority until it is released. Therefore we spend most of our time releasing it and try to get the Testrail percentage up.

However, we did not forget about this and we'll look into it.

Kind regards,
UI Coordinators

thibsy:
Hi @lscharmer

- [x] `Card` properties: could you explain what would happen to other properties of this component when added to a `LightboxCardPage`? E.g. what would happen if the `Card` has a title-action.

- [x] Further sections: the concept of adding further sections to the `Card` seems a bit odd, because these sections will only ever be used _if_ the component is added to a `LightboxCardPage`. Therefore we suggest to move `Card::withFurtherSections()` and `Card::getFurtherSections()` to the `LightboxCardPage`, where they will certainly be used.

- [x] `LightboxCardPage::getCard()`: please remove this method, `LightboxPage::getComponent()` should be used to return the `Card` component instead.

Kind regards,

@Amstutz, @klees and @thibsy

matthiaskunkel:
Jour Fixe, 23 JAN 2023: We highly appreciate this PR and accept it for trunk.

lscharmer:
Hi @Amstutz, @klees, @thibsy

> - [ ] Card properties: could you explain what would happen to other properties of this component when added to a LightboxCardPage? E.g. what would happen if the Card has a title-action.

  The Card is not rendered by the LightboxCardPage. The Card is merely used for it's information. So to show a Card the Card must be rendered somewhere else and therefore the title-action and all other actions won't change their behaviour.

> - [ ] Further sections: the concept of adding further sections to the Card seems a bit odd, because these sections will only ever be used if the component is added to a LightboxCardPage. Therefore we suggest to move Card::withFurtherSections() and Card::getFurtherSections() to the LightboxCardPage, where they will certainly be used.

Yes, we had a discussion about this some time ago and AFAIK this (the one in this PR) approach was the preferred one, as the information is not specific to a Modal / Lightbox but to a Card instead. If someday we don't want to show a detailed view of a Card in a Modal but also in another way (e.g as a fancy carousel) the new component can simply receive a Card an has all necessary information.

But AFAIK the Card is already too generic and is used in too many places, so I can move the `*FurtherSection` method to the `LightboxCardPage` if you prefer it.

> - [ ] LightboxCardPage::getCard(): please remove this method, LightboxPage::getComponent() should be used to return the Card component instead.

Done

Best regards,
@lscharmer

thibsy:
Hi @lscharmer

Sorry for the late response, we typically try to get things done in about two weeks and this flew under our radar a bit.

The reason behind `Card::withFurtherSections()` seems clear to us now, but we would prefer a more precise phrasing. Maybe something like `Card::withHiddenSections()`? Otherwise, one could assume `withSections()` and `withFurtherSections()` add to the same list, which might cause some confusion.

Since this PR was approved by the JF already, you can start the implementation now.

Kind regards,
@Amstutz, @klees and @thibsy

lscharmer:
Hi @Amstutz, @kless, @thibsy 
`*FurtherSections` is renamed to `*HiddenSections`.

Best regards
@lscharmer

Amstutz:
Thx. The interface looks good now. I close this to keep the list of open PRs short (so we not lose track of our todo list). Feel free to reopen once the final implementation is ready. looking forward to it.

PR #5290, opened 2022-12-07
Added info on ECS Interface

bogenschlag:
“Missing” is simply false. It used to be (just) me (even though it proved to be too herculean a task for me alone).

I suggest to name the SIG CampusConnect which _is_ jointly working on new test cases and test infrastructure (under the lead of Johannes Heim).


matthiaskunkel:
It is fine for me to list a group like a SIG as tester and responsible for test cases. But I would prefer to have an e-mail assigned instead of the permalink for the SIG. The e-mail contact should allow to contact a person for questions or comments.

bogenschlag:
You mean like `[SIG CampusConnect und ECS(A)](mailto:…)`? How about a list address on the Society’s list server then, so that all four members of the SIG administration would receive the mails?

matthiaskunkel:
@bogenschlag: forget my comment. Everything is fine. I merge your PR.

PR #5292, opened 2022-12-08
release_8 remove newline at the end of lng file which causes an error in lang sort script

daniwe4:


matthiaskunkel:
Thanks to @daniwe4 for fixing the language file.

PR #5300, opened 2022-12-09
Bugfix 35265: small change to cmd handling in LS

schmitz-ilias:
Hi @klees, this PR fixes [35265](https://mantis.ilias.de/view.php?id=35265).


klees:
Thanks a bunch!

klees:
Hi @schmitz-ilias,

I only just noticed that I also fixed this some days ago, but did not include into trunk until now: https://github.com/ILIAS-eLearning/ILIAS/pull/5310/commits/cdfc7a12f5743c7775802b5694e8ec9d1c374a03

To me the fixes look mostly similar, but my commit removes some additional cruft. I would appreciate an opinion from you, but I guess we could close this.

Still, thanks a lot for looking into this.

schmitz-ilias:
Yeah, that looks like pretty much the same fix ^^

PR #5301, opened 2022-12-09
Footer goto link: Use overlay instead of toasts

lscharmer:
As the current implementation is based upon the JS toast API which is causing (the JS API not the footer link) issues: https://mantis.ilias.de/view.php?id=34590
And the replacement is an ongoing discussion: #5121

Independent of the result of these discussions: the JS Toast API cannot be used in the future.
So this PR changes the goto link to use a JS overlay for a short period of time, to indicate that the link was copied to the clipboard.

Previous PR: #4811


Amstutz:
Hi @lscharmer 

Sry, for letting you wait for some time. Normally we try to get feedback to you within about 10 days. However, the TB decided to give the release 8 highest prio untill it is realeased. Therefore we spend most of our time into realasing it and try to get the testrail percentage up.

Please do not attach JF PRs the JF lable. You may request it, but we will do it. What exactly are the questions you wish to be answered by the JF?

thx.

oliversamoila:
Hi @Amstutz 
Many thanks for the feedback.
Sorry, the rule for assigning the label was not clear to me. (But maybe it is not fully comprehensible when I look in L103 under https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/contributing.md .)

In consultation with @mkunkel we have added a DevIssue to PR #4811. See: https://docu.ilias.de/goto.php?target=wiki_1357_JourFixe-2023-01-09#ilPageTocA2497 
Because this PR is directly related to it, the label seemed sensible to us in order to create transparency.

Many thanks to you 
Best regards

matthiaskunkel:
JourFixe, 09 JAN 2023: We have two options to implement the feature to copy the permanent link of a page in ILIAS. As an alternative to the current suggestion made in PR4811 we could also just show "Permanent Link" (with underlying URL) and let users copy it per mouse click. Decision about this PR is postponed to next JF.

klees:
According to the decision on [JF 2023-02-06](https://docu.ilias.de/goto.php?target=wiki_1357_JourFixe-2023-02-06) I close this.

Thanks to @lscharmer anyway for working on this. The effort might seem to be spend in vain, but IMO it's not: It helped us discuss and figure out which solution we want and how it fits in our general workload.

PR #5307, opened 2022-12-13
T&A Bugfix #0016771: Error when trying to add hint to question from within a test with Ilias Editor

fhelfer:
Fixes https://mantis.ilias.de/view.php?id=16771

### DIFF
--- a/Modules/TestQuestionPool/classes/class.ilAssQuestionHintGUI.php
+++ b/Modules/TestQuestionPool/classes/class.ilAssQuestionHintGUI.php
@@ -132,6 +132,17 @@ class ilAssQuestionHintGUI extends ilAssQuestionHintAbstractGUI
             $questionHint->setPoints($form->getInput('hint_points'));
 
             $questionHint->save();
+
+            if ($this->questionOBJ->isAdditionalContentEditingModePageObject() && !ilAssHintPage::_exists(
+                'qht',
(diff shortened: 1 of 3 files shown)


kergomard:

I have two questions:
- Could you please quickly explain why [this](https://github.com/ILIAS-eLearning/ILIAS/pull/5307/files#diff-466198434feda7c404355b7d9a2c0402357b25d5624bb72bf157eab7c2f09bc5R99) if-statement is needed? For me it looks like we are checking if a page exists before we create one anyway, within the ensure-function.
- Do you know if we have this same issue in newer versions of ILIAS?

fhelfer:
> Thank you very much for the PR @fhelfer !
> 
> I have two questions:
> 
> * Could you please quickly explain why [this](https://github.com/ILIAS-eLearning/ILIAS/pull/5307/files#diff-466198434feda7c404355b7d9a2c0402357b25d5624bb72bf157eab7c2f09bc5R99) if-statement is needed? For me it looks like we are checking if a page exists before we create one anyway, within the ensure-function.
> * Do you know if we have this same issue in newer versions of ILIAS?

* you are right, this line was edited while this pr was open
* seems not to be an issue in 8&9

kergomard:
Thanks again @fhelfer ! I will merge into ILIAS 7 and not perform any further action.

PR #5309, opened 2022-12-13
Bugfix 35452 et al.: filter out invalid users from an objects participants in Tracking

schmitz-ilias:
This PR should hopefully fix [35452](https://mantis.ilias.de/view.php?id=35452), [35762](https://mantis.ilias.de/view.php?id=35762), and [35473](https://mantis.ilias.de/view.php?id=35473).

I'm not able to reproduce these issues locally as they seem to require a very specific setup, but from what I can tell they are caused by users with an entry in obj_data, but no entry in usr_data showing up as an object's participants in tracking queries. The issues are reported to occur with multiple types of objects (at least tests and exercises), so the best fix I could come up with is this band-aid in `ilTrQuery`, where these faulty users are filtered out before being passed on.

### DIFF
--- a/Services/Tracking/classes/class.ilTrQuery.php
+++ b/Services/Tracking/classes/class.ilTrQuery.php
@@ -1012,6 +1012,9 @@ class ilTrQuery
 
         // begin-patch ouf
         if ($members_read) {
+            // BT 35452: failsafe against invalid users without an entry in usr_data
+            $members = self::filterOutUsersWithoutData($members);
+
             return $GLOBALS['DIC']->access(
             )->filterUserIdsByRbacOrPositionOfCurrentUser(
                 'read_learning_progress',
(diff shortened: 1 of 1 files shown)


Services/Tracking/classes/class.ilTrQuery.php
```diff
@@ -1087,6 +1093,27 @@ public static function getParticipantsForObject(int $a_ref_id): ?array
...
     }
 
+    /**
+     * @param int[] $user_ids
+     * @return int[]
+     */
+    protected static function filterOutUsersWithoutData(array $user_ids): array
+    {
+        $res = [];
+        foreach ($user_ids as $user_id) {
```
smeyer-ilias, line 1107: To avoid this iteration - under normal circumstances - you could call userExists() with all user_ids and iterate only in case the result is different to the number of users.

smeyer-ilias (changes requested):
Thanks @schmitz-ilias, please check the comment.

Services/Tracking/classes/class.ilTrQuery.php
schmitz-ilias: done, thanks!

smeyer-ilias (approved):

PR #5312, opened 2022-12-14
Bugfix Metadata & ADT (4)

schmitz-ilias:
This PR fixes [35888](https://mantis.ilias.de/view.php?id=35888), [35515](https://mantis.ilias.de/view.php?id=35515), [35914](https://mantis.ilias.de/view.php?id=35914), [33725](https://mantis.ilias.de/view.php?id=0033725), [35494](https://mantis.ilias.de/view.php?id=35494).

### DIFF
--- a/Modules/MediaPool/classes/class.ilObjMediaPool.php
+++ b/Modules/MediaPool/classes/class.ilObjMediaPool.php
@@ -21,7 +21,7 @@
  *
  * @author Alexander Killing <killing@leifos.de>
  */
-class ilObjMediaPool extends ilObject
+class ilObjMediaPool extends ilObject implements ilAdvancedMetaDataSubItems
 {
     protected ?int $default_width = null;
     protected ?int $default_height = null;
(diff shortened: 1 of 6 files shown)


alex40724 (commented):
Thanks for the PR! Looks good to me. @smeyer-ilias If ok for you, you may merge this. One Issue:

Do we need "implements ilAdvancedMetaDataSubItems" for ilObjPortfolio and ilObjPortfolioTemplate as well?

schmitz-ilias:
@alex40724, good catch, that functionality was indeed missing. I added it to `ilObjPortfolio`, but not to `ilObjPortfolioTemplate`: as  far as I can tell pages in portfolio templates can have assigned custom metadata sets, but those sets can never actually be filled with values (that only happens after a portfolio was created from the template). Since the subitem names are only relevant when already filled values need to be changed, the templates should never have to give out the names of their pages.

Since it's only a small change I added it to this PR, I hope that's alright.

alex40724 (approved):

smeyer-ilias (approved):
Thanks @schmitz-ilias 
btw. I've added the missing type and return types to ilAdvancedMetaDataSubItems

PR #5314, opened 2022-12-14
Add Chat toast provider for GS

iszmais:
Due to changes made inside the load of toast an provider interface was established to porvide the initial toasts without another request: https://github.com/ILIAS-eLearning/ILIAS/pull/5121

Therefore every provider of toast is obliged to provide a new toast Provider if his service want to add toast to the page.
ATM this only applies to initial toast but will probably be extended to all toast in near future.

This PR is an Proposal for toast inside your Service. Feel free to use and manipulate this template to your fitting and provide a PR for Trunk.
If you dont want to provide Toast with your Service feel free to close this PR.

### DIFF
--- a/Modules/Chatroom/GlobalScreen/ChatInvitationToastProvider.php
+++ b/Modules/Chatroom/GlobalScreen/ChatInvitationToastProvider.php
@@ -1,0 +1,67 @@
+<?php
+
+/**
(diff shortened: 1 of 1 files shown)


iszmais:
@mjansenDatabay This should only be merged when the dependencies are fullfilled.

ATM this only purpose is to present an example how the toast would be handeld if said changes are merged

PR #5315, opened 2022-12-14
New Caching Service

chfsx:
This PR shows a new implementation of GlobalCache in /src. After merging this, a replacement of the current implementation will follow (DropInReplacement). Additionally, the Memcached Implementation will follow asap

Consumers should use the new implementation in the future

### DIFF
--- a/Services/GlobalCache/classes/class.ilGlobalCache.php
+++ b/Services/GlobalCache/classes/class.ilGlobalCache.php
@@ -17,9 +17,7 @@
  *********************************************************************/
 
 /**
- * Class ilGlobalCache
- * @author  Fabian Schmid <fs@studer-raimann.ch>
- * @version 1.0.0
+ * @deprecated use ILIAS\Cache\Container\Request instead via $DIC->globalCache()
  */
 class ilGlobalCache
 {
--- a/src/Cache/Adaptor/APCu.php
+++ b/src/Cache/Adaptor/APCu.php
@@ -1,0 +1,62 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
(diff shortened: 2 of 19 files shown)


klees:
Hi @chfsx,

I really love that you are improving on the caching interface, I always needed to look up how things work when I was using the old interface ~twice a year.

IMO we should use this opportunity to move the read interface of the service to [the pattern that we use for HTTP requests and other inputs](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/input-processing.md#a-pattern-to-make-developers-remember).

From the perspective of our application, the data in the cache needs to be considered as "input". We simply cannot trust the data that is stored in the cache. Currently, an attacker could place prepared data in the cache and we would happily import it to our processes by using `unserialize`. This is a well known attack vector for every application that supports according functionality, see e.g. this [article by owasp](https://owasp.org/www-community/vulnerabilities/PHP_Object_Injection). A possible scenario would be that an attacker gains access to the cache via another, badly secured, PHP application, injects some hand-crafted object that e.g. force an error that shows some secret information.

Maybe this is a good occasion to discuss the general topic in the community again, e.g. on a workshop. I would be happy to organize something like this with you.

Best regards!

chfsx:
Hi @klees 

Thank you very much, that is certainly a valid point that we need to discuss. from my point of view, we could make good use of the refinery in the caching to address these things. however, the OWASP references are generally always about the fact that what is loaded from the cache, for example, also comes from user inputs in some way. from my point of view, that is not the case in our case, but of course it cannot be prevented. I think we will take this as an opportunity to talk about this again in general.

matthiaskunkel:
Jour Fixe, 19 DEC 2022: Fabian notified us that the interface for the new caching service is almost ready and that some further development (incl. drop-in replacement) is needed until this PR can be merged in January.

chfsx:
Addition: I will only merge this once the discussion on security has been concluded.

chfsx:
@klees as discussed I now only accept primitive types and there is no longer a serialize/unserialize involved. may you have a look at this again?

chfsx:
The most relevant part is here: https://github.com/ILIAS-eLearning/ILIAS/pull/5315/files#diff-5f3274e76a6e025cfd00a369434a9195c49d2b2659a97e0dc84332969a3bddbbR62

There is a union type return declaration which we do not want in ILIAS, but I think this is a location were we can rationalize this here

chfsx:
if there are no further objections, i would merge the pr next week.

Amstutz:
thumbs up for adding tests.

chfsx:
merging postponed since there came some more feedback in discord

PR #5316, opened 2022-12-14
Replace Event Cache with Artifacts

chfsx:
Events in ILIAS were previously cached in the GlobalCache - if activated. Since events are directly dependent on the code base, they can be better stored as artefacts, which significantly increases performance.


matthiaskunkel:
Jour Fixe, 19 DEC 2022: Fabian has implemented changed for artefact based caching in the event handling. The event handling itself hasn't been changed. Stephan will review the PR.

kergomard (approved):
Thank you very much @chfsx for this improvement!
This looks good to me. I have the feeling that we should start thinking about introducing a cache directory, but this is clearly not part of this improvement.
Best,
@kergomard

PR #5321, opened 2022-12-16
TA: 35945 call to undefined function

nhaagen:
https://mantis.ilias.de/view.php?id=35945


kergomard:
LGTM. Thank you very much @nhaagen. Will merge and cherry-pick to trunk.

PR #5325, opened 2022-12-19
T&A Bugfix: Fix Broken editQuestion GUI

fhelfer:


fhelfer:
@kergomard your commit https://github.com/ILIAS-eLearning/ILIAS/commit/74dd7869f6f0825a897119fc588e2c3d759a1ddc broke editQuestion GUI
Hope my bugfix doesnt break your intended changes from the commit :)

kergomard:

Quick question: Is there a mantis issue? If not: Can you please tell me, when exactly this fails? The fix looks actually trivial and ok, but as this parts are not really straight forward, I would like to give it a little bit more thought.

fhelfer:
https://mantis.ilias.de/view.php?id=35968
Fails when, creating a new SingleChoice Question in Test.
Edit Question Tab shows an empty page

kergomard:
As far as I can see, this only concerns trunk.

PR #5327, opened 2022-12-19
Fix Doubleencoding in List-View

kergomard:
See: https://mantis.ilias.de/view.php?id=35975
There are surely better solutions, but maybe this is just enough until I had some more time to get deeper into the ilObject-Service?


chfsx:
Thanks you very much! I'll cherry pick this down to other relevant branches as well

PR #5330, opened 2022-12-20
A11y, fix of focus search issue, see #35523

Amstutz:
Fix of https://mantis.ilias.de/view.php?id=35523 . 

Concerns: 
"WCAG Issue: When selecting one of the radio buttons in the "Search" fieldset, focus is redirected to the input field without warning to the user, and without the name of the radio button being read to the user."

Note, this fix comes at a convenience price for people with vision. For them it will probably feel like a decrease of usability, since the cursor will not jump automatically into the search field anymore after making a choice.


smeyer-ilias:
Thanks for fixing this.

PR #5334, opened 2022-12-20
Trunk: Use Plugins language handler to update language

mBeym:
Current version forces the use/creation of an "ilPluginLanguage" handler   
but doesn't even use it to actually update the languages. Instead it creates a new one that would ignore any plugins custom changes.


klees:
Thanks a lot!

PR #5341, opened 2022-12-22
CharSelector: Ensure correct types

mjansenDatabay:
If approved, this has to be cherry-picked to `release_8`.


fneumann:
also cherry-picked to release_8. Thx!

PR #5342, opened 2022-12-22
User: Fix wrong argument type for system notification

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=34248

If approved, this has to be cherry-picked to `release_8`.


kergomard:

PR #5344, opened 2022-12-22
T&A 0016123 fix deletion of mc answer also deletes corresponding feedback

kunzeus:
Hello,

This is a "fix" for https://mantis.ilias.de/view.php?id=16123

By adding a hidden field, a comparison is made between the answer IDs that are already exists in the database and the post answer IDs from the hidden field. This allows for the correct identification of feedback that need to be deleted and reordered. This fixes the error with the number 0016123.

However, the functionality of the JavaScript "+" button to add a new answer text duplicates all values of the table row where it is placed. This creates a new side effect. Allows retaining feedback that should actually be deleted.

This PR is open for evaluation to decide if the fix is sufficient as a temporary improvement until a full kitchen sink rework is done.  

Best wishes and a relaxing vacation
Ulf

@kergomard  wants to test the fix in the new year

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assAnswerBinaryStateImage.php
+++ b/Modules/TestQuestionPool/classes/class.assAnswerBinaryStateImage.php
@@ -44,7 +44,7 @@ class ASS_AnswerBinaryStateImage extends ASS_AnswerBinaryState
      */
     public function __construct($answertext = "", $points = 0.0, $order = 0, $state = 0, $a_image = "", $id = -1)
     {
-        parent::__construct($answertext, $points, $order, $id);
+        parent::__construct($answertext, $points, $order, $state, $id);
         $this->image = $a_image;
     }
 
--- a/Modules/TestQuestionPool/classes/class.assMultipleChoice.php
+++ b/Modules/TestQuestionPool/classes/class.assMultipleChoice.php
@@ -746,7 +746,7 @@ class assMultipleChoice extends assQuestion implements ilObjQuestionScoringAdjus
         $feedback = $ilDB->fetchAll($result);
 
         // Check if feedback exists
-        if (sizeof($feedback) >= 1){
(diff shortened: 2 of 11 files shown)


kergomard:
Thank you very much @kunzeus für this improvement!

As this improves the situation, I would suggest to merge it. Do you agree @dsstrassner ?

@kunzeus : Could you please remove the $tmp-Variables [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5344/files#diff-d992b9232822011560c2de0377d4c8b986578814bd87ef318f6b6290976b6beeR716) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5344/files#diff-d992b9232822011560c2de0377d4c8b986578814bd87ef318f6b6290976b6beeR726).

Thank you very much and best!
@kergomard

dsstrassner:
👍

Yeah. I think this would be great.

kergomard:
Thank you very much for the changes @kunzeus !
...and I'm terribly sorry, but I overlooked something: This changes make the question go boom, when the Page Editor is used for feedback. Might I kindly ask you have a look at it once more?

If you are at it, one very small thing I completely forgot:
- Please change the `array()` to `[]`. I would have done that myself (as I've forgotten), but as we need to touch it once more ...

Thank you very much and best,
@kergomard

dsstrassner:
Hi @kergomard, 

is this PR now ready to merge?

kunzeus:
No there is still an issue with the page editor objects. I'll look into it again next week.

dsstrassner:
Any Updates @kunzeus?

kunzeus:
Yes, I fixed the page editor error. 
However, I think it's wiser to fix the whole thing properly, on the Database level. Than further extending a workaround, which is only a partial solution.

kergomard:

You are right concerning your assessment of the situation. I will still merge this and cherry-pick to ILIAS 8 and trunk.
We then need to find a more intelligent solution.

Best,
@kergomard

PR #5346, opened 2022-12-31
Port must be an integer value instead of a string

legionth:
Moin everybody,

found a small detail in the setup docs that the `port` for `database` MUST be an `integer`.
Updated the docs accordingly.

Happy new year!


mjansenDatabay:
@klees Maybe something to assign to Ann-Christin in the future.

klees:
Thanks!

klees:
Picked to release_8.

PR #5347, opened 2023-01-01
Convert to string, because str_pad requires the first argument to be …

legionth:
…string. Found during initial call to index.php after initial setup.

Found in PHP Version: `8.1`

According to `setup/sql/ilias3.sql:9157` this is considered to be an `integer` which makes absolutely sense. This is just a quickfix for me, to keep the installation running. Feel free to give feedback, what should be done here! 

Cheers!

### DIFF
--- a/Services/Object/classes/class.ilObjectDefinition.php
+++ b/Services/Object/classes/class.ilObjectDefinition.php
@@ -139,7 +139,7 @@ class ilObjectDefinition
                 "rbac" => $rec["rbac"],
                 "group" => $rec["grp"],
                 "system" => $rec["system"],
-                "default_pos" => "9999" . str_pad($rec["default_pos"], 4, "0", STR_PAD_LEFT), // "unassigned" group
+                "default_pos" => "9999" . str_pad((string) $rec["default_pos"], 4, "0", STR_PAD_LEFT), // "unassigned" group
                 "sideblock" => $rec["sideblock"],
                 'export' => $rec['export'],
                 'repository' => $rec['repository'],


legionth:
```
Time: 00:23.540, Memory: 231.00 MB

There was 1 failure:

1) CronJobScheduleTest::testSchedule with data set "Weekly Schedule / Did run this Week" (ilCronJob@anonymous/home/runner/work/ILIAS/ILIAS/Services/Cron/test/CronJobScheduleTest.php:40$194 Object (...), false, DateTimeImmutable Object (...), 5, null, false)
Last run: 2022-12-26T00:00:00+00:00
Failed asserting that true is identical to false.

/home/runner/work/ILIAS/ILIAS/Services/Cron/test/CronJobScheduleTest.php:291
```

Didi I break the CI or is it a "new-years"-Bug 😄  I'll check later, when I got a little more time.

mjansenDatabay:
> Didi I break the CI or is it a "new-years"-Bug smile I'll check later, when I got a little more time.

Happy new year @legionth :-). I don't think you broke something, this might just be a date arithmetics issue. I will try to reproduce this, random seed for the unit tests was 1672570305 .

-- UPDATE --

Looks like an old issue with the determination if two dates are within the same week.

```php
case self::SCHEDULE_TYPE_WEEKLY:
    $last = $last_run->format('Y-W');
    $ref = $now->format('Y-W');

    return ($last !== $ref);
```

For the point in time when the CI executed the unit tests this resulted in:

```php
$ref =  '2022-52';
$now = '2023-52';
```

The desired behaviour is: `\ilCronJob::checkSchedule` should return `false`, because both dates are withing the same week. But instead `true` was returned, because the strings are not the same.


https://en.wikipedia.org/wiki/ISO_8601#Week_dates:

> As a consequence, if 1 January is on a Monday, Tuesday, Wednesday or Thursday, it is in week 01. If 1 January is on a Friday, Saturday or Sunday, it is in week 52 or 53 of the previous year (there is no week 00). 28 December is always in the last week of its year.

kergomard:
Thank you for the PR @legionth . I merge it like this. The ugliness of this lies elsewhere and needs more thinking.

PR #5348, opened 2023-01-02
Fix for Mantis 0036029

Akumatic:
Fixes https://mantis.ilias.de/view.php?id=36029


Amstutz:
hi @Akumatic 

Thx you very much and major kudos for the Fix. Highly appreciated. I will also pick that to trunk.

Kindly,
@Amstutz

PR #5349, opened 2023-01-02
Cron: Fix schedule check for weekly schedules

mjansenDatabay:
See:
- https://mantis.ilias.de/view.php?id=36118
- https://github.com/ILIAS-eLearning/ILIAS/pull/5347

### DIFF
--- a/Services/Cron/classes/class.ilCronJob.php
+++ b/Services/Cron/classes/class.ilCronJob.php
@@ -33,6 +33,44 @@ abstract class ilCronJob
     protected ?int $schedule_value = null;
     protected ?Closure $date_time_provider = null;
 
+    private function checkWeeklySchedule(DateTimeImmutable $last_run, DateTimeImmutable $now): bool
+    {
+        $last_year = (int) $last_run->format('Y');
+        $now_year = (int) $now->format('Y');
+
+        if ($last_year > $now_year) {
+            // Should never happen, don't execute otherwise
+            return false;
+        }
+
+        $last_week = $last_run->format('W');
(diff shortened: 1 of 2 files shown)


Services/Cron/classes/class.ilCronJob.php
```diff
@@ -55,9 +55,40 @@ private function checkSchedule(?DateTimeImmutable $last_run, ?int $schedule_type
...
+                    return $last_month !== $now_month;
+                }
+
+                if ($now_year - $last_year > 1) {
+                    // Always execute if the difference of years is greate than 1
+                    return true;
+                }
+
+                // Execute for week number 52 in 2022 (last run) and week number 52 in December of 2022 (now), but not for week number 52 in January of 2022 (now)
+                return $last_month === $now_month;
```
mjansenDatabay, line 91: This could be optimized, but I'd like to make all cases explicit.

Services/Cron/classes/class.ilCronJob.php
```diff
@@ -33,6 +33,44 @@ abstract class ilCronJob
...
+            return $last_month !== $now_month;
+        }
+
+        if ($now_year - $last_year > 1) {
+            // Always execute if the difference of years is greater than 1
+            return true;
+        }
+
+        // Execute for week number 52 in 2022 (last run) and week number 52 in December of 2022 (now), but not for week number 52 in January of 2022 (now)
+        return $last_month === $now_month;
```
mjansenDatabay, line 71: This could be optimized, but I'd like to make all cases explicit.

PR #5350, opened 2023-01-02
Link/User Registration: Fix permanent link building

mjansenDatabay:
This PR fixes an issue causing a TestRail test case to fail: https://testrail.ilias.de/index.php?/tests/view/56247

See:
- https://mantis.ilias.de/view.php?id=34853
- As a workaround for the regression caused by Mantis https://mantis.ilias.de/view.php?id=33088 / https://github.com/ILIAS-eLearning/ILIAS/commit/13b97c69 (@smeyer-ilias)

If approved, this MUST be cherry-picked to `release_8`.

### DIFF
--- a/Services/Link/classes/class.ilLink.php
+++ b/Services/Link/classes/class.ilLink.php
@@ -47,9 +47,13 @@ class ilLink
             }
         }
 
-        // workaround for administration links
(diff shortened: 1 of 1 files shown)


smeyer-ilias:
Thanks for the fix.

PR #5352, opened 2023-01-02
[WIP] Initial setup for docker

legionth:
Just wanted to propose to add some [Docker Compose](https://docs.docker.com/compose/) files to ILIAS.
I used it for my developer setup, and maybe some of you are interested in this.

This files are very basic and need a lot more love, before they are able to merge. There are components still missing like `Java RPC server ` and a proper `SMTP` configuration, but these things a "figureout-able".

Let me know what you think and if should share the progress of using this :)

Cheers!

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,53 @@
+version: '3'
+
+services:
+  nginx:
+    image: nginx:alpine
+    restart: unless-stopped
+    volumes:
+      - ./:/var/www/ilias:delegated
+      - ./docker/nginx/default.conf:/etc/nginx/conf.d/default.conf:delegated
+    command: "/bin/sh -c 'while :; do sleep 6h & wait $${!}; nginx -s reload; done & nginx -g \"daemon off;\"'"
+    ports:
+      - "80:80"
+      - "443:443"
+    depends_on:
+      - php
+    networks:
+      - app-network
+  php:
+    build: ./docker/php/
+    restart: unless-stopped
+    environment:
+      APP_ENV: dev
+      DATABASE_URL: mysql://ilias_user:my_password@mysql:3306/ilias?serverVersion=5.7
+    volumes:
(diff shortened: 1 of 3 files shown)


legionth:
And yes I use nginx, because I don't like apache. Sue me.

😄

mstuder:
hi @legionth 

I think docker is a very important topic, especially for ILIAS. I personally use docker also in big productive enviroments.

A colleague of mine has invested a lot of work in the last two years to make ilias docker capable.
Have a look to:
https://github.com/search?q=user%3Afluxfw+flux-ilias&type=
And:
https://hub.docker.com/search?q=fluxfw

Without having to decide anything here - but I am happy to share my knowledge. 

Personally I would do it without docker-compose.yml in the ILIAS repository. The docker compose file is a build guide for the containers and not for the image. If necessary you could think about a Dockerfile in the ILIAS repo. But even this could be maintained in a second project / repository. 

Regarding modularization and  Docker-(micro)-Services I would recommend to develop and maintain things separately  - if possible. This way they are not subject to the same release cycles. It could still be an official or official supported repository of ilias open source - but with the advantage of modularization, which in my view will be absolutely necessary in the future. If this is not done when Docker - which stands for modularization - is introduced, then when?

Best regards
Martin

klees:
Hi @legionth,

are you back in ILIAS development? You might be interested in [doil - Create ILIAS Development Environments with Docker and ILIAS](https://github.com/conceptsandtraining/doil) which delivers a lot of the elaboration that this is missing =)

Best regards!

legionth:
Thank you for your responses!

I see everybody has their solutions here and there. I just wondering why these are not part of the repository or the at least the `README`. But this shouldn't be discussed in this PR.

Thank you for clarifying @mstuder and @klees !

 P.S.:

> are you back in ILIAS development? 

Not really, just getting my feet wet ;)

PR #5354, opened 2023-01-03
Administration: Fix default view for `General Settings`

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=34278


alex40724:
Thanks!

PR #5356, opened 2023-01-03
Logging: Fix wrong PHP types in job configuration

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=35366


smeyer-ilias:
Thanks for fixing this

PR #5357, opened 2023-01-03
Tracking: Fix access on undefined key in object statistics job

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=35368


smeyer-ilias:
Thanks for fixing this

PR #5359, opened 2023-01-03
SystemCheck: Fix PHP types in cron task

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=35371


smeyer-ilias:
Thanks for fixing this.

PR #5367, opened 2023-01-04
User: Fix accessing undefined array key

mjansenDatabay:
See:
- https://mantis.ilias.de/view.php?id=35226
- http://testrail.ilias.de/index.php?/tests/view/54119


kergomard:
Thank you very much @mjansenDatabay ! This is simply ugly, ugly, ugly, ...but that it was already before. I merge and cherry-pick.

PR #5376, opened 2023-01-05
Search: Fix search filter handling

mjansenDatabay:
See:
- https://mantis.ilias.de/view.php?id=35961
- https://mantis.ilias.de/view.php?id=35277
- http://testrail.ilias.de/index.php?/tests/view/57515
- https://testrail.ilias.de/index.php?/tests/view/53337

The HTTP POST/session handling could still be optimized, but this PR fixes the filters which do not work at all currently.


smeyer-ilias:
Thanks for fixing and the other PRs

PR #5389, opened 2023-01-06
Init: Always use `-1` for `error_reporting` in `DEVMODE` (ILIAS >= 9)

mjansenDatabay:
PHP 8.1 is required for ILIAS 9.x, so we can remove the conditions and set `error_reporting` to `-1` if `DEVMODE` is active without any condition.

This is only relevant for ILIAS 9.x (current `trunk`).


Amstutz:
thx a lot!

PR #5392, opened 2023-01-06
User: Fix wrong sorting of `context` when listing role assignments

mjansenDatabay:
See:
- https://mantis.ilias.de/view.php?id=34957
- http://testrail.ilias.de/index.php?/tests/view/59398

### DIFF
--- a/Services/User/classes/class.ilRoleAssignmentTableGUI.php
+++ b/Services/User/classes/class.ilRoleAssignmentTableGUI.php
@@ -23,6 +23,8 @@ class ilRoleAssignmentTableGUI extends ilTable2GUI
 {
     protected ilPathGUI $path_gui;
     protected array $filter; // Missing array type.
+    protected \ILIAS\UI\Factory $factory;
+    protected \ILIAS\UI\Renderer $renderer;
 
     protected ilObjectDefinition $objectDefinition;
 
@@ -36,6 +38,8 @@ class ilRoleAssignmentTableGUI extends ilTable2GUI
         $lng = $DIC['lng'];
         $rbacsystem = $DIC['rbacsystem'];
         $this->objectDefinition = $DIC['objDefinition'];
+        $this->factory = $DIC->ui()->factory();
+        $this->renderer = $DIC->ui()->renderer();
 
 
         $lng->loadLanguageModule('rbac');
@@ -96,10 +100,6 @@ class ilRoleAssignmentTableGUI extends ilTable2GUI
 
     protected function fillRow(array $a_set): void // Missing array type.
     {
-        global $DIC;
-
-        $lng = $DIC['lng'];
-
         if (isset($a_set['checkbox']['id'])) {
             $this->tpl->setVariable('VAL_ID', $a_set['checkbox']['id']);
             if ($a_set['checkbox']['disabled']) {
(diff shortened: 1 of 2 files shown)


Services/User/classes/class.ilRoleAssignmentTableGUI.php
```diff
@@ -193,26 +214,14 @@ public function parse(int $usr_id): void
             ];
             $this->ctrl->setParameterByClass("ilobjrolegui", "ref_id", $rolf[0]);
             $this->ctrl->setParameterByClass("ilobjrolegui", "obj_id", $role["obj_id"]);
```
kergomard: I think we cannot set these two parameters here and have to move them to `ilRoleAssignmentTableGUI::fillRow()`, too.

Services/User/classes/class.ilRoleAssignmentTableGUI.php
```diff
@@ -109,9 +109,30 @@ protected function fillRow(array $a_set): void // Missing array type.
...
-        $this->tpl->setVariable("CONTEXT", $a_set["context"]);
+        $this->tpl->setVariable(
+            'ROLE',
+            $this->renderer->render(
+                $this->factory->link()->standard(
+                    ilObjRole::_getTranslation($a_set['title']),
+                    $this->ctrl->getLinkTargetByClass(ilObjRoleGUI::class, 'perm')
+                )
+            )
+        );
```
kergomard, line 124: I think we are missing the two parameters (`$obj_id` of the role and `$ref_id` of the role folder - see below) here when setting the link.

kergomard (changes requested):
Thank you very much @mjansenDatabay for this fix!

I think we still have a little issue with the linking of the role (see comments inline). I think we will need to carry them over through the array or do you see a better solution?

Best,
@kergomard

Services/User/classes/class.ilRoleAssignmentTableGUI.php
mjansenDatabay: Good catch @kergomard , I'll add the `obj_id` to the array.

Services/User/classes/class.ilRoleAssignmentTableGUI.php
mjansenDatabay: IMO we already have the `ref_id`, because the role folder is the parent.

Services/User/classes/class.ilRoleAssignmentTableGUI.php
```diff
@@ -109,9 +109,34 @@ protected function fillRow(array $a_set): void // Missing array type.
                 $this->tpl->setVariable('VAL_CHECKED', 'checked="checked"');
             }
         }
-        $this->tpl->setVariable("ROLE", $a_set["role"]);
-        $this->tpl->setVariable("DESCRIPTION", $a_set["description"]);
-        $this->tpl->setVariable("CONTEXT", $a_set["context"]);
+
+        $this->ctrl->setParameterByClass("ilobjrolegui", "ref_id", $a_set['ref_id']);
+        $this->ctrl->setParameterByClass("ilobjrolegui", "obj_id", $a_set["obj_id"]);
```
mjansenDatabay, line 114: @kergomard I moved the `ilCtrl` calls to `fillRow`.

kergomard:
Thank you very much @mjansenDatabay 

Yes, you are right, the role-folder is the parent. I just saw two different functions to fetch them and didn't go deeper.

PR #5393, opened 2023-01-06
TA: 36161, remove info message

nhaagen:
https://mantis.ilias.de/view.php?id=36161


kergomard:
Ok, in the best of all worlds this is none-sense. I merge and cherry-pick to to trunk. Thank you very much @nhaagen .

PR #5408, opened 2023-01-09
PRG: CertificatePlaceholders by longest validity and latest completion

nhaagen:
https://mantis.ilias.de/view.php?id=36175


klees:
Merged to trunk.

PR #5409, opened 2023-01-09
Release 8: actually use plugins language handler

mBeym:
Previously merged request: https://github.com/ILIAS-eLearning/ILIAS/pull/5335

Addition: Fixed `getLanguageHandler` method not public in default `ilPlugin` implementation
Fixes: https://mantis.ilias.de/view.php?id=36169

### DIFF
--- a/Services/Component/classes/Settings/class.ilObjComponentSettingsGUI.php
+++ b/Services/Component/classes/Settings/class.ilObjComponentSettingsGUI.php
@@ -168,10 +168,7 @@ class ilObjComponentSettingsGUI extends ilObjectGUI
     protected function refreshLanguages(): void
     {
         try {
-            $plugin_name = $this->request_wrapper->retrieve(self::P_PLUGIN_NAME, $this->refinery->kindlyTo()->string());
-            $plugin = $this->component_repository->getPluginByName($plugin_name);
-            $language_handler = new ilPluginLanguage($plugin);
-            $language_handler->updateLanguages();
+            $this->getPlugin()->getLanguageHandler()->updateLanguages();
             $this->tpl->setOnScreenMessage("success", $this->lng->txt("cmps_refresh_lng"), true);
         } catch (Exception $e) {
             $this->tpl->setOnScreenMessage("failure", $e->getMessage(), true);
--- a/Services/Component/classes/class.ilPlugin.php
+++ b/Services/Component/classes/class.ilPlugin.php
@@ -332,7 +332,7 @@ abstract class ilPlugin
     // Language Handling
     // ------------------------------------------
 
-    protected function getLanguageHandler(): ilPluginLanguage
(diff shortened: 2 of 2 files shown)


klees:
Hi @mBeym,

I relabeled this as an improvement, because currently there is no ticket attached that is fixed by this PR.

So I wonder: why is it that you want to add this change? I reread your initial PR, and you seem to want to use a custom language handler for your plugin. But this is the very thing that I wanted to prevent with the current structure. Or to put it differently: I just don't want to have that extension point for plugins right now. What do you intend to use it for? 

Best regards!

mBeym:
My reason for this change was that a feature that plugin developers "could" use was replaced with a proper Class/Object implementation, which is a great change, adding a method to retrieve that object.

but then, instead of using that object, with a single method call (two if you count getting the plugin object) and create an entirely new language handler object
```php
$language_handler = new ilPluginLanguage($plugin);
            $language_handler->updateLanguages();
```

skipping the entire plugin this language handler is supposed to... `handle`. 

Furthermore... In my opinion this doesn't add problems, concerns or anything else that would matter.
Plugin developers not needing this... won't use it, and plugin developers needing this... can use it.


In my case, developing plugins, I use it to load languages from a json file instead of the line by line unreadable .lang file implementation.

To me it's just bad approach to forcefully require it and then not use it **and** lock it down

klees:
Hi @mBeym,

I won't make this change for two reasons (and hence close the PR):

The first reason is from the perspective of the maintainer of `Services/Component` (me...). Every freedom that you, as a plugin developer, have for your plugin is less freedom that I, as the maintainer of the infrastructure, have in cleaning up and refactoring the component. If for some reason I come to the conclusion that the current structure with `ilPlugin::getLanguageHandler` isn't the structure I want to maintain, I would need to think about your plugin. Realistically, the situation will most likely play out something like this: I will announce to make the change in the method on the JF, no one will object, I will make the change, your plugin will break, you or your bosses will yell at me. A slim interface is for the benefit of all of us.

Second reason is from the perspective of the system and it's user. We have an existing infrastructure for language, as bad as it might be. For user this e.g. means, that they can change lang vars in the administration or find the according files in known formats and locations. As a system we try to behave consistent so users can build expectations. So no, just because you as a plugin developer do not like the format of the lang-files and want to use json instead, ILIAS does not need to support that as an extension point in its infrastructure. In reality, this would play out something like this: You build a plugin for someone. Someone tries to change the lang-vars in the administration but can't. She files a mantis report for ILIAS, that I need to look into and spend time on.

In fact, there already was considerable time spend on this none issue: A breaking change in your previous PR, bugreports, etc. So please, if you do not like the format of the lang-file, work on that format so that everyone can use the improvements. And try to understand that there are other people in this community that also might have valid motives and reasons.

Have a nice weekend.

PR #5410, opened 2023-01-09
Trunk: actually use plugins language handler

mBeym:
Previously merged request: https://github.com/ILIAS-eLearning/ILIAS/pull/5334

Addition: Fixed `getLanguageHandler` method not public in default `ilPlugin` implementation
Fixes: https://mantis.ilias.de/view.php?id=36169


klees:
I close this until we have discussed https://github.com/ILIAS-eLearning/ILIAS/pull/5409.

PR #5411, opened 2023-01-09
fixed sql syntax error during object deletion

smeyer-ilias:
See https://mantis.ilias.de/view.php?id=36188


smeyer-ilias:
Fixed with 38512a337a0ad05b3be5c38aacc50578c269a572

PR #5414, opened 2023-01-10
Bugfix 36127: do not show copyright entry if selection is inactive

schmitz-ilias:
This PR fixes [36127](https://mantis.ilias.de/view.php?id=36127) and [28893](https://mantis.ilias.de/view.php?id=28893#c88741). This should also be merged into R8 and trunk, let me know if you'd like a separate PR for that.


smeyer-ilias:
Thanks @schmitz-ilias

PR #5416, opened 2023-01-10
Fix for Mantis 0036140

Akumatic:
Implemented suggested change for https://mantis.ilias.de/view.php?id=36140


Amstutz:
Thx a lot!

Amstutz:
pushed to trunk as well.

PR #5417, opened 2023-01-10
Raise error after session logout

iszmais:
https://mantis.ilias.de/view.php?id=36185

This also provides messages for login and startuppages, even if a deleted session leads to them.

### DIFF
--- a/Services/Dashboard/classes/class.ilDashboardGUI.php
+++ b/Services/Dashboard/classes/class.ilDashboardGUI.php
@@ -61,6 +61,11 @@ class ilDashboardGUI implements ilCtrlBaseClassInterface
         $this->lng = $DIC->language();
         $this->ctrl = $DIC->ctrl();
 
+        if ($this->user->getId() === ANONYMOUS_USER_ID) {
+            $DIC->ui()->mainTemplate()->setOnScreenMessage('failure', $this->lng->txt("msg_not_available_for_anon"), true);
+            $DIC->ctrl()->redirectToURL("login.php?cmd=force_login");
+        }
+
         $this->tpl = $tpl;
 
         $this->ctrl->setContextObject(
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
The solution is similar to the one proposed by @kergomard for `release_7`, which I merged yesterday: https://github.com/ILIAS-eLearning/ILIAS/pull/4870 . In `release_8` (and above), the specific issue @kergomard fixed is not relevant  (anymore). But if there are still issues with the `MessageBox` elements on the login page, we should maybe implement this session based approach. It is still kinda ugly that we'll have to do it this way.

Please keep in mind that you'll have to involve @pascalseeland for the changes in `ilStartupGUI` (and also me as 2nd coordinator, which you already did :-)).

iszmais:
@mjansenDatabay thx for the info!

i reverted my changes on the startupGUI @kergomard solutions should do the trick too.

Therefore this is still ready for merge, but the the desired result of the mantis ticket still awaits the cherry pick of #4870 into ILIAS 8 and trunk.

iszmais:
UPDATE: Still awaits that fix https://github.com/ILIAS-eLearning/ILIAS/pull/4870 by @kergomard

kergomard:
Sorry for the long wait: Cherry-pick for ILIAS8 is here: #5625

lscharmer (approved):

iszmais:
@mjansenDatabay This is ready to merge if it gets your approval

iszmais:
Im not sure why i not initialy did that but i changed the redirect to a static force login since thats seems the most reasonable process of that error.

Further i tested some other cases of expired/lost session with and without referers on their behaviour which all seem to work good now, except one case:
- When a session is deleted (e.g. when i remove cookies from my browser and reload) no Error message is shown due to the fact that the erro cant linked to any call without an proper identifier. Im not sure if this is solvable (if it is even a problem) but when thats the case this most likely has to be solved inside the startupGUI imo. 

@mjansenDatabay @lscharmer Please leave me a feedback if you disagree on any of that.

mjansenDatabay:
The solution should be fine @iszmais

PR #5418, opened 2023-01-11
Auth/SAML: Store used auth_mode explicitly and move SAML logout to ev…

mjansenDatabay:
…ent listener

This commit removes the SAML specific code from `ilStartupGUI` to an event listener in `Services/SAML`.
Furthermore, it renames the session key `used_external_auth_mode` used for some external authentication modes and stores the used mode instead of a boolean flag.


klees:
Hi @pascalseeland

Best regards!

pascalseeland:
@mjansenDatabay this can be directly merged after rebase has happend

PR #5420, opened 2023-01-11
make more "defined" checks in ilInitialisation

chfsx:
This is a PR concerning https://mantis.ilias.de/view.php?id=35607 to make migrations work better


klees:
Merged to trunk.

PR #5425, opened 2023-01-11
Bugfix Conditions: fix copying of number of required materials

schmitz-ilias:
This PR fixes the copying of 'Number of Required Materials' of preconditions. It will be kept as is if still applicable, otherwise a fallback is calculated. Currently the fallback is always applied.


smeyer-ilias (approved):
Thanks!

PR #5426, opened 2023-01-11
Bugfix 36254: fix lp status icons in Staff

schmitz-ilias:
This PR fixes [36254](https://mantis.ilias.de/view.php?id=36254).


schmitz-ilias:
I just noticed that this is the exact same fix as #5098, which never got merged.... I'll leave both open until one of them gets lucky.

PR #5427, opened 2023-01-11
Implemented UFTNormalization as Transformation

chfsx:
Hi @klees 
This is a followup from a mantis-ticket, there are some places we should be able to normalize strings to one of the Unicode Normalization Forms (see https://unicode.org/reports/tr15/)

The usage currently (or as it was with ilStr::normalizeUtf8String) are bad and we will get rid of them anyway (by using the new Inputs).

But since this isn't a big deal to write this Transformation, here it is...

https://mantis.ilias.de/view.php?id=34241

### DIFF
--- a/src/Refinery/String/Group.php
+++ b/src/Refinery/String/Group.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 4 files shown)


klees:
Added @mjansenDatabay because we coordinate the Refinery together.

klees:
I have simplified this a little and merged this as https://github.com/ILIAS-eLearning/ILIAS/commit/85bac61bbc85ac9bea38fdeac50f4e861343e71e / https://github.com/ILIAS-eLearning/ILIAS/commit/3856386cf874077ab3ffe1e4805fed084cd862da

PR #5428, opened 2023-01-12
Bugfix 33273: type fix in MD

schmitz-ilias:
This PR fixes [33273](https://mantis.ilias.de/view.php?id=33273).


smeyer-ilias:
Thanks Tim

PR #5429, opened 2023-01-12
T&A: fix for 35577

okaufman:
fixes https://mantis.ilias.de/view.php?id=35577


kergomard:

PR #5430, opened 2023-01-12
PRG: 36262, fix list of ProgrammesMonitoringCategory

nhaagen:
https://mantis.ilias.de/view.php?id=36262


klees:
Picked to trunk.

PR #5437, opened 2023-01-13
PRG: 36276, accesshandling for PRGs as precondition

nhaagen:
https://mantis.ilias.de/view.php?id=36276


klees:
Cherry-picked to trunk.

PR #5439, opened 2023-01-13
Fix typo in "Remove server information"

iljalukin:
Fix misspelled command (should be Header instead of Headers)


klees:
Thanks a bunch =)

klees:

PR #5442, opened 2023-01-13
PRG: 36284, resetProgressesToSettings w/o auto-aquiring passed courses

nhaagen:
https://mantis.ilias.de/view.php?id=36284


klees:
Picked to trunk.

PR #5449, opened 2023-01-16
[FIX] make test more robust in macOS

chfsx:
This test fails after you opened the directory using Finder in macOS (which creates a .DS_Store file in it)


klees:
Picked to release_8.

klees:
Thanks a bunch. I'm not a huge fan of this, but this is more a complaint directed to MacOS than anything else =)

PR #5451, opened 2023-01-16
UI: introduce help topics and according facilities

klees:
Hi everyone,

with this PR we want to propose two things:

* A general way to bind help texts to UI components.
* A tangible usage of this way to make `Link`s and `Button`s have tooltips.

This PR proposes a new interface `HasHelpTopics` and attaches it to the base interfaces of `Link` and `Button`. Consumers can then use this interface to attach one or many help topics to any link or button. The UI framework will then use the `HelpTextRetriever` in the rendering process to look for appropriate text snippets and, if available, will render them alongside the original component as a tooltip. The `HelpTextRetriever` will need to be implemented to use the learning module, most likely by `Services\Help`, to recover the current state of the art of the Online Help Editorial Team.

This structure has the following properties:

* Any direct connection to a specific implementation of the help can be removed from consumers of the UI framework. So can handling of help texts.
* The `HelpTextRetriever` offers a straight forward way to implement multi language tooltips, custom help texts via plugins or even new help scenarios like a guided tour.
* Consumers of the UI Framework don't need to concern themselves with the question how certain help texts will be displayed. This decision will be made by the UI Framework, new insights from UI and UX can be incorporated, if need be, without affecting consumers code.
* Help topics will become available for computational evaluation. This could, e.g., be used to show help texts according to the visible features on a certain page, such as tabs or buttons.
* It is a more powerful superset of the current "screen-id" and "tooltip-id" model of the help. Current help systematic hence can be represented, but new possibilities (see above...) arise.

Thanks to FH Dortmund for funding this work and Kendra Grotz for discussing the options with me. Thanks to the UI coordinators for feedback and futher discussions.

Best regards!

Richard

---
Any questions? I might have anticipated some:

### Why don't we just add a component `Popover\Tooltip`.

The well known tooltip functionality, where hovering a certain element reveals a tiny popover containing some hopefully helpful text, has certain drawbacks: It doesn't work well with touch devices, user cannot know where to expect tooltips, they vanish when hovering out of the UI elements. [There are indeed other models for displaying said texts](https://inclusive-components.design/tooltips-toggletips/) without these drawbacks. Not introducing an explicit component allows us to adjust the way these texts are displayed and used in the future.

The second, more general point is as such: There are doubts, that a tooltip is a component at all. It barely has other properties than the actual string to be displayed. It is intimately tied to the component this string is about. The text for the tooltip looks rather similar to something like a byline for form inputs. Hence, tooltips seem to be a property of other UI components but not components in their own right.

### This is bad, because I won't be able to use this for my legacy component, right?

In general yes. But: When implementing this we will try to build this in a way that legacy UI components can display according tooltips as easy as possible. Since legacy UI should vanish at some point, this will be a temporary arrangement anyway.

### Why not simply attach texts directly to components? Why use these topics?

As outlined above, the new layer of indirection that we propose here, will open up some space for interesting new features. Also, this seems to make it quite simple for consumers of the UI-Framework and quite flexible on the other side.

### Why should we use an extra type for the help topics. Wouldn't plain strings suffice?

In general plain strings will be sufficient for all computations. Arrays of bytes ftw. But jokes aside: We deal with lots of plain strings in ILIAS, a hint on `string` mostly does not tell us, what we intend to do with this string. Also, plain strings allow us to do stuff that we shouldn't necessarily do: search and replace, concatenate, capitalize, ... So we attempt to fight primitive obsession here and tag our strings with a class. This will tell the world that they are indeed help topics, not just any string, and also forbid things that no one should do with help topics.

### How can I know, as a user of the UI framework, which help topics are available?

You can't. And also you shouldn't. This is why we called the `HelpTextRetriever` a `Retriever` and not, e.g. a provider or a resolver: To clarify that it is indeed possible that there are no help texts for a list of certain topics. Its the opposite: Consumers of the UI Framework should faithfully add appropriate topics to their components. These can be simple. `"logout"` might already do for the logout button in the Meta Bar. For some occasions, two topics might be appropriate: `"search", "user"`. Also, some components might introduce an even more controlled vocabulary for the topics: `ilMembership::HELP_TOPIC_ADD, "tutor"`. In some circumstances, we even might add help topics automatically, such as `"toolbar"` for buttons in the toolbar. At the beginning we can absolutely reuse the existing tooltip- and screen-ids as topics, but we can develop more elaborate schemes in the future based on the model proposed here.

To gather the topics that are actually used in ILIAS, we could simply write a `HelpTextRetriever` that just records the topics during, e.g., testing. This is what I meant by "Help topics will become available for computational evaluation.".

### Why did you name it topic and not just id, at is it now?

From my perspective, an id is meant to, well, identify a certain entity. Regarding the help, it is questionable what would be identified at all. "screen-ids" pretend to identify a certain screen. But the very concept of a "screen" didn't become very tangible in ILIAS until now and might simply not be helpful at all. For the tooltips, the ids seem to identify certain snippets of text in some circumstances, but certain components, such as main menu entries, in other. The very concept of identifiable components on the screen seems to brittle at best. For the main menu, e.g., the concept took huge hit when the configurable Main Bar was introduced. Is the "Repository"-entry still the same, when it located in a submenu? Also we may very well ask the question, what constitutes identity for components, just as we can ask this for screens. Is the "Member"-tab the identical tab at a course and group? Is it the identical tab at two different courses? Is it identical for different users that, due to their permissions, can see or do different things on that tab? These questions might or might not have good or helpful answers, but for this proposal here we simply avoid them by using "topics" instead of "id".

The basic model hence is: Consumers of the UI Frameworks can define topics that the help for a certain component should talk about. The `HelpTextRetriever` is responsible to produce a list of possible concrete texts for a certain purpose and topics. (Some of) these texts than can be displayed by the component or someone else, e.g. as a tooltip or in a sidebar. No need to talk about ids.

### Can this also be used to resolve screen-ids?

Yes. A screen-id would just be a super specific topic in the new wording. We would need to implement `HasHelpTopics` on the `Page` component, though.

### What would be other purposes than "tooltip"?

I think that we would at least want to have a purpose that targets the kind of texts that the actual online help displays. These might even be two or more different purposes. I would imagine that "how to" and "background info" could be good purposes to categorize the texts from the online help. If we think of new features, I think that "guided tour" would be another good purpose.


Amstutz:
Hi @rklees

We like the concept behind the proposal. Thx for the very detailed explanation.

Questions:

- [x] Documentation: Your explanations are very helpful to quickly grasp the idea behind the concept. Normally, we do not do that, but this goes somewhat beyond the info we provide through YAML. Maybe add it as readme into our docs folder of the UI Components? Parts of it, are already in comments, which is also fine.
- [x] "Purpose": could you describe a little what a purpose is? Perhaps some documentation would be nice as well, so other devs can quickly grasp the concept of the class and how they could possibly extend it.
- [x] Final classes: Why did you close the classes Purpose and Topic to inheritance?

Thx a lot! 

@Amstutz and @thibsy

klees:
at "Documentation": I would propose to add a document in src/UI/docu afterwards. Maybe something comes up during discussion and we need to expand. Will happily do so...

at "Purpose": I added according docu to the class. Hope this helps.

at "Final Classes": For `Purpose` I did so because it represents a controlled vocabulary (see other item). I closed `Topic` for inheritance because it is just a simple marker for strings. In general I deem inheritance to be problematic in many cases. So other way round: Why should someone extend `Purpose` or `Topic`?

matthiaskunkel:
Jour Fixe, 23 JAN 2023: We highly appreciate this PR and accept it for trunk.

klees:
I think this is solid now. @Amstutz/@thibsy, could you please review this? There also is a small part in here for @alex40724. Maybe @chfsx wants to have a look too? I know that this is much, so if anyone wants to have a walkthrough before giving a review, feel free to contact me.

Services/Help/classes/class.ilHelpUITextRetriever.php
```diff
@@ -0,0 +1,54 @@
...
+ * alternative mechanisms via plugins.
+ */
+class ilHelpUITextRetriever implements ILIAS\UI\HelpTextRetriever
+{
+    public function getHelpText(Help\Purpose $purpose, Help\Topic ...$topics): array
+    {
+        if ($purpose->isTooltip()) {
+            return array_filter(
+                array_map(
+                    fn ($topic) => ilUtil::getTooltipPresentationText($topic->get()),
```
alex40724: This would be ilHelp::getTooltipPresentationText(...)

alex40724 (approved):
Thank you for taking care of integration of help texts into the UI framework with this proposal. If I understand everything correctly, the use of the existing help content should be possible this way, if we find a way to integrate this into the legacy tabs. I see several ways to do this with interim solutions, e.g. using KS links in the legacy TABs rendering, since it will take some time before we have a tab concept in the UI framework in general.

A major difference is, that the indirection is moving from "one" topic ID to an array of topics and that we now have multiple retrievers possibly being chained up. This is not an issue for the current Help service and the existing content, since it just may continue to use its existing topic IDs.

I am not sure how retrievers would handle the topic array, if components come with multiple values. I could think of something like a taxonomy path here, e.g. course->settings->general.

If this is opened up for plugins, some namespace mechanism might be relevant to avoid conflicts on this level. Otherwise it will be just a "magic string" approach, even if wrapped as topics.

Having multiple chained retrievers also offers great flexibility, e.g. for plugins. I just wonder how this looks on the rendering side. How will multiple tooltip texts be rendered, are there any mockups?

So, again, if I do not oversee something fundamental, the solution should work well with the current Help content.

Services/Help/classes/class.ilHelpUITextRetriever.php
klees: Thanks, indeed!

Amstutz:
Hi @klees

Thx for the final implementation on this. Special thx for the JS and PHP Tests as well as the examples.

Since this PR contains some JavaScript code as well, we want to leave some notes for you but the general public as well (note, no direct action of yours needed for those two points):

- ILIAS copyright license: All PHP-files have to contain the official ILIAS copyright license as a PSR-12 document-level comment. We believe this should be true for JavaScript files as well and should be considered in the copyright-checker as well.
- JavaScript filenames: There are currently several approaches for naming our JavaScript files mixed together. We should think about how we want to name these files and streamline this as soon as possible. Ideas on a naming-concept are very welcome.


- [x] Order: See also point of @[alex40724](https://github.com/alex40724) above,  "I am not sure how retrievers would handle the topic array if components come with multiple values. I could think of something like a taxonomy path here, e.g. course->settings->general“. Is there a concept for that?
- [x] Devmode Echo: Why are you initializing the Retriever in Devmode with the Echoing version only? Could that lead to issues for Help Devs or other people mostly working in Devmode?
- [x] Examples Tags: Thx for the Examples. In the Tag examples, is there a point behind the $possible_relevances . IMO, this does not add any additional value to explaining how Tooltips work with tags and could probably be omitted.
- [x] Classes in JS: You introduce working with classes in JS. Is this a proposed pattern we should use in future more often?
- [x] replaceContent Marker: I needed to read the code twice until I completely got the function of the marker to replace content. Please drop a few more lines in the comments about what it does and why we need it. 
- [x] comparing elements: you are currently comparing two `HTMLElement` with the `===` operator in `Tooltip.onPointerDown()`. Are you sure this will lead to the expected behaviour in a robust manner? or should we maybe use [`isEqualNode`](https://developer.mozilla.org/en-US/docs/Web/API/Node/isEqualNode)?


Maybe change or explain why not:

- [x] Scoping of JS: The compiled JS puts e.g. the function replaceContent in the global scope. We should avoid this. In the counter example, this is done by using iife in rollup (putting everything into a self executing function). Can we do the same here? 
- [x] `TooltipRenderer::maybeGetTooltipEmbedding()`: This method currently only returns an array of `Closure` or `null`. Couldn't we change the return-type to `?array` with an according PHPDoc comment `Closure[]|null`? or is there a specific reason for `mixed`?

Please change: 

- [ ] Docu: Since this is in its final stages now, please add a document src/UI/docu as discussed above (see my and your comment in January).
- [x] ilHelp: Please change according to the proposal of @alex40724 above
- [x] Spacing: There seems to be a space too much behind (!) here: https://github.com/ILIAS-eLearning/ILIAS/pull/5451/files#diff-ce368bd026fc188f2c74b22b252dce4ee8b848851c5e5dcca2f02523973566cbR38
- [ ] Testrail: Do not forget to add/adopt Testrail cases for Tooltips changes in the examples.
- [x] JS structure: thx a lot for applying rollup for the core directory, but please use a `dist/` folder to output the `ui.js` file to. There are already several components following this structure and we want to keep things consistent.
- [x] JSDoc comments: please add some JSDoc comments (see e.g. [here](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html) for more info). This way modern IDEs can provide more concrete autocompletion and find probable bugs. Also be as concrete as possible, e.g. `MouseEvent` over `Event` or `HTMLDivElement` over `HTMLElement` and so on.
- [x] `Tooltip` instance variables: please declare each variable set in the constructor and provide some JSDoc comments as well. If instance variables are optional, initialize a default value instead of checking for `undefined`.
- [x] Undefined variable: [`foo`](https://github.com/ILIAS-eLearning/ILIAS/pull/5451/files#diff-80b7e5733fbe45fa24f2c66116cca1b084c2e925e36d632c8ef389ce65b3db88R15) is an undefined variable, right? should this be may be `tooltip_id` instead?
- [x] `Tooltip.isInMain()`: this method is possibly called several times right? We think it would be more efficient if `this.main` was initialized in the constructor instead. This would also respect the single-responsibility-principle as this this method currently has a side-effect consumers would not assume by just reading the method name.
- [x] type-safe comparisons: there are several places in your javascript code where type-unsafe comparisons are used (`==`), please use a strict comparison (`===`) whenever possible.

Thx very much,
@amstutz and @thibsy

klees:
Hi @alex40724,

thanks for the feedback and questions.

* about integration in legacy components: I tried to [bundle the rendering](https://github.com/ILIAS-eLearning/ILIAS/pull/5451/files#diff-4016e98c789a191357852561c6b7d539581dc42feb6fac78c7e0e558cf31f483) in a way that should be usable by legacy components. Usage is [shown here](https://github.com/ILIAS-eLearning/ILIAS/pull/5451/files#diff-a076d60b9a7852496094f6098709d398cc37829b13caa8aba18f3a811126cb41R252).
* about existing content: Your absolutely correct, it should be easy to integrate existing content in the new solution (and  [ilHelpUITextRetriever](Services/Help/classes/class.ilHelpUITextRetriever.php) seems to demonstrate that). Intentionally so... =)
* about multiple topics: From my perspective we should just start to attach relevant topics to buttons and links. These could be many per component (and of course include the tooltip-ids that we currently have). On the "New Object" button we could, e.g. attach "current_tooltip_id", "create", "object", maybe even more. For the tooltip-id we have an obvious strategy to resolve it. But a retriever could also return a specific tooltip for the combination of "object" and "create". It could also return some generic text about objects. Taxonomy could work as well. So gist is: I'm not really sure how we will end up using this either, but I'm mostly sure that we should try to attach topics with meaningful names instead of "just" ids. Plugins or core can then start to look for good strategies to provide helpful texts for single topics or combinations thereof. I would not see the topics as something that identifies a certain text (as the current ids are) but instead as a way for developers to mark up their components, even if we might not (yet) have texts for the concrete instance.
* rendering: Multiple tooltips will currently just be renderer beneath each other in the same box, each wrapped in a `<p>`. Looks really boring =) So a user probably won't even notice that these are "multiple" tooltips.

I hope this makes some sense...

I think that we will continue the discussion about how we find good topics, if we want to have conventions, etc. sometimes soon. AFAIK Kendra will reach out to stakeholders to put heads together and make a plan.

Best regards!

Amstutz:
Thx for the answers. IMO this is looking good. We might need to extend the docs somewhat further in future. I resolved the merge conflicst. The failing unit tests seemed not to be related to the PR. Thx!

PR #5454, opened 2023-01-17
Bugfix 34349: generate course member list from user gallery

schmitz-ilias:
This PR fixes [34349](https://mantis.ilias.de/view.php?id=34349). Some slight refactoring of `ilAttendanceList` was necessary, but nothing with too large a footprint.


smeyer-ilias (approved):
Thanks

PR #5456, opened 2023-01-17
Bugfix 36306: pass size in correct format to ilMapGUI

schmitz-ilias:
This PR fixes [36306](https://mantis.ilias.de/view.php?id=0036306).


kergomard:
Setting the values here this way for me seems wrong (why is ilObject concerned with setting any sizes in ADT?), but if this is what we do in ADT, then the changes in ilObject are fine with me.

kergomard:
@smeyer-ilias : Could you please have a look at this. It would be nice to get this off the list.

PR #5457, opened 2023-01-17
TA, 36269: write points for matchingPairs during question corrections

nhaagen:
https://mantis.ilias.de/view.php?id=36269

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assMatchingQuestion.php
+++ b/Modules/TestQuestionPool/classes/class.assMatchingQuestion.php
@@ -692,6 +692,17 @@ class assMatchingQuestion extends assQuestion implements ilObjQuestionScoringAdj
     }
 
     /**
+    * @param assAnswerMatchingPair[]
+    */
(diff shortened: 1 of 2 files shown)


nicoroeser:
@nhaagen , out of curiosity: why do you create a clone and then replace the object with it? Couldn’t the pairs just be loaded into the _current_ object?

(I have not tested nor deeply thought about this, only skimmed your commit. Please forgive me if I am missing something.)

nhaagen:
@nicoroeser , I did not see a method in assMatchingQuestion to actually _set_ a well configured pair.
There are methods to (newly) create them, based on information I would have to retrieve anyways. 
So, instead of taking the pairs apart and have them rebuild, I thought cloning and changing the relevant aspect the easier way.
Maybe, I'm also so used to do this due to repo-pattern stuff, that this is my standard approach ;)

mbecker-databay:
> 

withFunctionalCreep(); 

...but the dude abides... or are there objections?

kergomard:
No objections from my side.

PR #5458, opened 2023-01-18
Fixed Initialization of Anonymous User

chfsx:
This PR proposes a fix for https://mantis.ilias.de/view.php?id=36312

@pascalseeland, @kergomard and I discussed this issue last week and this is a proposal to avoid it.

We are not sure why there should be the state of initialization where a user with ID 0 exists. This PR ensures that the user session always has at least `ANONYMOUS_USER_ID`.

In addition, it is prevented that a request is redirected to the login page, when a request should actually reach a publicly accessible page: https://github.com/ILIAS-eLearning/ILIAS/compare/trunk...srsolutionsag:ILIAS:fix/0036312/9/user-initialization?expand=1#diff-5c2233ae6af0264344f1d3415e104429899c63359ffb40cd2503cbd7fd44d894R1694

By initializing the user with the `ANONYMOUS_USER_ID` a problem had to be fixed that a skin with the ID `''` is searched for: https://github.com/ILIAS-eLearning/ILIAS/compare/trunk...srsolutionsag:ILIAS:fix/0036312/9/user-initialization?expand=1#diff-1d6ee119cb221ac919799f18da915ebd8c7e44a3e427d851eac7bd2be1513d2cR111 . @Amstutz could you have a look at this as well?

### DIFF
--- a/Services/Authentication/classes/class.ilAuthSession.php
+++ b/Services/Authentication/classes/class.ilAuthSession.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 /**
  * @author Stefan Meyer <smeyer.ilias@gmx.de>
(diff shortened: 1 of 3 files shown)


kergomard (commented):
Thank you very much @chfsx ! I'm fine with the changes. As said in our meeting: My gut says that there will be consequences. I will tackle them in as far as RBAC or ilUser are concerned, if we decide to change this.
From where I stand, we can add the jourfixe-tag as discussed to see if somebody sees good reasons not to do this (I won't be around at the next jf though).

Best,
@kergomard

chfsx:
Thank you for your feedback! Yes, I think we have all been programming around the fact that there is a user with 0 and 13 for all these years, so there might be problems with changing it back. I also hope that someone still remembers why the whole thing is like this :-) I added the label

Amstutz (approved):
thx

matthiaskunkel:
Jour Fixe, 23 JAN 2023: We highly appreciate this PR and accept it for trunk. No more user 0...

chfsx:
I fixed the test failing in PHP8.1 here https://github.com/ILIAS-eLearning/ILIAS/commit/f5304be7d56171b56761925029f1ce30a59ee481

PR #5460, opened 2023-01-18
OrgU: Repo Pattern Refactoring (part 2)

tbongers-cat:
Dear colleagues,

this is a follow up of #5243. This PR switches User Assignments, Permissions, Operations and Operation Contexts from Active Record to Repository Pattern. It also includes a small adaption for _Modules/StudyProgramme_ (as discussed with @nhaagen) and a necessary change to _Services/AccessControl_ for handling the position permissions.

The goal of our refactoring of the organisational units (OU) according to the "Repository Design Pattern" was an optimization of the code by making it more structured and creating well-defined interfaces and methods. Another important goal was making the code testable and providing initial unit tests for each object, ensuring that our code base is future-proof and works well as an essential part of ILIAS.

The previously used "Active Object Pattern" has some flaws concerning its handling of large datasets and its code sometimes is a bit hard to read and analyze. The main reason for this is the ability of active objects to persist themselves to the database, which therefore happens in several parts of the code. The same is true for all queries to the database, which are scattered throughout the code and may be called more often than necessary. This makes code and query optimizations quite complicated, so we decided to use the Repository Design Pattern consistently in all parts of the OU, as this pattern consolidates all database operations in the object repositories.

Our focus during the refactoring process was designing a better code structure for all objects of the organisation units (e.g. user assignments and positions). To achieve this, in the first step we created single repositories for each object which serve as a "single point of truth" for anything concerning the object, i.e. the creation and deletion of objects as well as storing them to and reading them from the database. Besides that, each repository offers all necessary methods for querying and manipulating objects that are needed by other modules. The repository interfaces offer a well-documented overview of the available methods for each object type.

The new repositories provide us with more options for code optimizations, especially for instances with high user numbers and/or a big data volume. In addition, new methods can be easily provided for any upcoming requirements. The code becomes more scalable and adaptable, making future development much easier. Our next steps will be further optimizations and consolidations of the repositories into larger "meta" repositories, as we have already identified some internal dependencies between repositories. We would like to further optimize this to provide more and better documented interfaces and methods for all use cases.

On the user interface side, nothing has been changed for the organisational units (except for some small bugfixes), as most changes are happening "under the hood". We also tried to keep all existing interfaces and methods intact, so all calls from outside the OU are still working as intended. In a few cases, we have provided small adaptions to code that directly interacts with the OU, especially for permission handling. Please note that some of the available methods have been marked as deprecated and we kindly ask other developers to adapt their code in the future to use the newly provided methods of our repositories.

Thanks for the attention!

### DIFF
--- a/Modules/OrgUnit/classes/Positions/Operation/OrgUnitOperationContextRepository.php
+++ b/Modules/OrgUnit/classes/Positions/Operation/OrgUnitOperationContextRepository.php
@@ -1,0 +1,40 @@
+<?php
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ ********************************************************************
+ */
+declare(strict_types=1);
+
+interface OrgUnitOperationContextRepository
+{
+    /**
+     * @throws ilException
+     */
+    public function registerNewContext(string $context, ?string $parent_context): void;
+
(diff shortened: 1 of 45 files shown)


tbongers-cat:
@klees 
As discussed I will be doing some adaptations to this before it is "ready to go". Will keep you updated.

PR #5461, opened 2023-01-18
Bugfix 36292: set correct mode for Preconditions table

schmitz-ilias:
This PR fixes [36292](https://mantis.ilias.de/view.php?id=36292). I'm not sure why the row template for the Preconditions table is in `Services/AccessControl`, but it is only used in `Services/Conditions`.


smeyer-ilias:
Thanks for the hint: i 've moved the template to the correct location

PR #5462, opened 2023-01-18
[FIX] 0036074: Activation of the option "Save confirmation" => add upload entry

chfsx:
This is a fix for https://mantis.ilias.de/view.php?id=36074

in summary, the problem was a mixup of GET/POST, various array-access problems and just in general that the idea of "faking" uploads in ilFileInputGUI was not a good idea.


Amstutz:
Thx, a lot, pushed to trunk as well!

PR #5471, opened 2023-01-19
Fix reference modules

ingoj:
I discovered that course-, group-, category- and studyprogramme-references behave slightly different. Course-, group- and categoryreference set type to course (crs), group (grp) and category (cat) in init, studyprogamme-references set their type to studyprogramme-reference (prgr). 
In had no side-effects in my tests, but I did not test every variant. I am about to prepare a feature-request for tiles for these objects where this fix is helpful. But anyway I assume these 4 object-types should have a consistent behaviour.

Any Feedback is welcome :-)


smeyer-ilias:
Thanks @ingoj for this hint and fix. 
I'm also not sure about the side effects and therefor close this PR for R7 and R8. 

Nevertheless I fixed it trunk:
https://github.com/ILIAS-eLearning/ILIAS/commit/bbef70a6b25efe5fcc4bdc0b1f8ae619ecd57df2

PR #5474, opened 2023-01-24
PRG: 36391, hide settings subtab if orgu-positions are globally disabled

nhaagen:
https://mantis.ilias.de/view.php?id=36391


klees:
Thx! Picked to trunk...

PR #5475, opened 2023-01-24
MathJax URL updated to Cloudflare's recommendation

utesche:
The default URL to Mathjax is updated to Cloudflare's recommendation, 
   old:   'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.9/MathJax.js?config=TeX-AMS-MML_HTMLorMML,Safe'
   new:  'https://cdn.jsdelivr.net/npm/mathjax@2.7.9/MathJax.js?config=TeX-AMS-MML_HTMLorMML,Safe'
see: 
https://www.mathjax.org/MathJax-v2-7-9-available/

This PR do **NOT** adapt *./setup/sql/ilias3.sql*  and  *.setup/sql/ilDBTemplate.php*  accordingly, because imho these files are created directly from ILIAS' reference database, right?.   The reference-db's update with *./setup/sql/7_hotfixes.php* must be done beside this PR to keep the setup files consistent.

I'm not familiar with the usual workflow to bring up new DB-update-steps - so take my updated *./setup/sql/7_hotfixes.php* as a proposal.

### DIFF
--- a/lang/ilias_ar.lang
+++ b/lang/ilias_ar.lang
@@ -10864,7 +10864,7 @@ mathjax#:#mathjax_limiter#:#Inline Delimiters
 mathjax#:#mathjax_limiter_info#:#Select the inline delimiters as they are configured in your MathJax installation.
 mathjax#:#mathjax_mathjax#:#MathJax
 mathjax#:#mathjax_path_to_mathjax#:#URL to MathJax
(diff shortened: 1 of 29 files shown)


fneumann:
Merged the PR with changes:
- Added the ',Save' in all languages
- removed the update step for lng_data because that is automatically done in the setup
The update of the DB dump and template is not needed. These files are created whe a new version is releases. 
Thanks to utesche!

PR #5477, opened 2023-01-25
PRG: 36435, set user-points to settings-points for successful progres…

nhaagen:
https://mantis.ilias.de/view.php?id=36435


klees:
Picked to trunk.

PR #5135, opened 2022-10-18
T&A: fix for 22409

okaufman:
Fixes https://mantis.ilias.de/view.php?id=22409
I added a check to see if there are any obligatory questions present in the test before adding the text "You filled out all compulsory questions." in the final message.

I made a new function to make it easier to read.

PR #5480, opened 2023-01-26
T&A Bugfix #0036461: Empty Feedback Error

fhelfer:
https://mantis.ilias.de/view.php?id=36461

### DIFF
--- a/Modules/Test/classes/class.ilTestGradingMessageBuilder.php
+++ b/Modules/Test/classes/class.ilTestGradingMessageBuilder.php
@@ -34,7 +34,10 @@ class ilTestGradingMessageBuilder
      * @var integer
      */
     private $activeId;
-
+    /**
+     * @var array
+     */
+    private $messageText = [];
     /**
      * @param ilLanguage $lng
      * @param ilObjTest $testOBJ


nicoroeser:
Among other effects, your change removes displaying the state of required questions in the information about a test try for a participant, see `ilTestEvaluationGUI->outParticipantsResultsOverview()`.

fhelfer:
@nicoroeser reverted
but in that case, because of the fact that it is removed in ilias8+, isnt there the same issue?

nicoroeser:
I had thought that as well, but haven’t had the time yet to check ILIAS 8 and later. This is why I just noted the fact what happens in my earlier comment, but without stating whether this is good or bad. We may have to discuss the effect and the wanted behaviour and might have to change things in ILIAS 8+, or just for ILIAS 7. I am not sure right now, sorry.

dsstrassner:
Hi @nicoroeser 

should we talk about this tomorrow?

nicoroeser:
@dsstrassner , OK, fine for me. I do not have a strong opinion (yet?) on the (now removed) part of the change about the state of the questions (see [above](https://github.com/ILIAS-eLearning/ILIAS/pull/5480#issuecomment-1406346298)). As I said, I had just noticed the changed behaviour, but without analyzing more in detail. That said, I’m of course happy to discuss things with – would be good to have opinions from others who are affected and care much (and are better prepared than me :smiley: ).

nicoroeser:
T&A TechSquad 2023-03-23:

@nicoroeser is going to look at this pull request in more detail, including the ILIAS 7 vs. ILIAS 8+ thing. We intend to handle this pull request afterwards with some more input.

nicoroeser:
After having looked at [Mantis Issue #36461] and this pull request in more detail, it turned out that this initially (in 3e20d5d36acbd4597adbc1ca251a8a152bde01a8) was a slightly reduced version of 64efb7237ccdcdc4adbdb9c4272abeb3b3115010 (which is on the `release_8` branch). And now it is an even smaller part of it. I had not expected this at first.

Well, this pull request is about fixing [Mantis Issue #36461]. I have tested the code without and with the commit in this pull request applied. **The current form of the pull request is sufficient to fix the reported issue. :heavy_check_mark:**

[Mantis Issue #36461]: https://mantis.ilias.de/view.php?id=36461 "Mantis Issue #36461: Fehlermeldung Test abschließen leere Rückmeldung"

@fhelfer , thanks for the issue report and the pull request! :thumbsup:

(With my developer hat on, not in my role as TB or T&A TechSquad member:) For future commits, I’d recommend to include in the commit message where the changes were taken from (→commit hash if from repo). A few words of explanation do not harm. As a matter of courtesy we (ILIAS developers) should keep a trail of author information or/and point to the original commit(s) when commits are copied, cherry-picked, etc.

----

Looking a bit further (into 64efb7237ccdcdc4adbdb9c4272abeb3b3115010 again), I see that the original commit probably improves things in the same area, for ILIAS 8. I have the vague feeling that we have already discussed these things somewhere, but I cannot remember, sorry.

Unfortunately I cannot tell from the commit message what _exactly_ the problem was and what exactly the change was supposed to do. It has an effect on several parts of the code. I was not able to find further information on the change, neither in Mantis nor in the repo nor in the feature wiki nor in the T&A TechSquad minutes. As all that is separate from this pull request:
* @fhelfer (or anyone else who reads this at some time in the future): if you think the change should be ported to ILIAS 7, please open another issue report/pull request, reference this one, and explain why this is needed.
* @kergomard , you made that original commit – so you probably know best. If you think the topic is worth further discussion, we can put it on the agenda for one of our next T&A TechSquad meetings; and if it isn’t, we just stand by. :smile:

dsstrassner:
TechSquad, 13.04.23: @kergomard will look into it and will try to answer the questions ;-)

kergomard:
Thanks @nicoroeser for looking into this.

Ok, I will merge this, as it solves the issue for 7. This has a long history and went through many hands. The commit you found, was most probably a by product. I will try to be more explicit in my commit messages. The main issue and PR we actually discussed was: https://mantis.ilias.de/view.php?id=22409, the fix by Olivia: https://github.com/ILIAS-eLearning/ILIAS/pull/5135.

The other parts of my commit for ILIAS8 are just cleanups.

Best,
@kergomard

nicoroeser:
Many thanks for feedback, explanations and merging, @kergomard !

PR #5483, opened 2023-01-27
R7: Fix assFormulaQuestion undefined method usage

mBeym:
Fixes usage of undefined method "request->raw()"

also add access to http request in assQuestionGUI class (parent class)


dsstrassner:
Hi @mBeym, are there any Mantis tickets open with this?

mBeym:
@dsstrassner Not directly. Only noticed this while working on https://mantis.ilias.de/view.php?id=27520.

The method comes from newer ILIAS versions (Ilias 8 & trunk) but does not exist in ILIAS 7 and prevents saving (or rather throws an error when trying to save a question.

mBeym:
* I meant https://mantis.ilias.de/view.php?id=23502

PR #5488, opened 2023-01-27
Introduction of mustache for the mailing

shecken:
This is the implementation of the feature [Use Mustache in Mail Template Configuration](https://docu.ilias.de/goto_docu_wiki_wpage_7318_1357.html)

### DIFF
--- a/Services/Mail/classes/class.ilMail.php
+++ b/Services/Mail/classes/class.ilMail.php
@@ -65,7 +65,8 @@ class ilMail
         private ?int $mail_admin_node_ref_id = null,
         private ?int $mail_obj_ref_id = null,
         private ?ilObjUser $actor = null,
-        private ?ilMailTemplatePlaceholderResolver $placeholder_resolver = null
+        private ?ilMailTemplatePlaceholderResolver $placeholder_resolver = null,
+        private ?ilMailTemplatePlaceholderToEmptyResolver $placeholder_to_empty_resolver = null
     ) {
         global $DIC;
         $this->logger = $logger ?? ilLoggerFactory::getLogger('mail');
@@ -93,6 +94,7 @@ class ilMail
         $this->table_mail_saved = 'mail_saved';
         $this->setSaveInSentbox(false);
         $this->placeholder_resolver = $placeholder_resolver ?? $DIC->mail()->placeholderResolver();
+        $this->placeholder_to_empty_resolver = $placeholder_to_empty_resolver ?? $DIC->mail()->placeholderToEmptyResolver();
     }
 
     public function autoresponder(): AutoresponderService
@@ -566,6 +568,11 @@ class ilMail
         return $message;
     }
 
+    private function replacePlaceholdersEmpty(string $message): string
+    {
+        return $this->placeholder_to_empty_resolver->resolve($message);
+    }
+
     private function distributeMail(MailDeliveryData $mail_data): bool
     {
         $this->auto_responder_service->emptyAutoresponderData();
@@ -590,7 +597,8 @@ class ilMail
         ));
 
         if ($mail_data->isUsePlaceholder()) {
-            $this->sendMailWithReplacedPlaceholder($mail_data, $to_usr_ids, $cc_bcc_recipients);
+            $this->sendMailWithReplacedPlaceholder($mail_data, $to_usr_ids);
+            $this->sendMailWithReplacedEmptyPlaceholder($mail_data, $cc_bcc_recipients);
         } else {
             $this->sendMailWithoutReplacedPlaceholder($mail_data, $to_usr_ids, $cc_bcc_recipients);
         }
@@ -603,8 +611,7 @@ class ilMail
 
     private function sendMailWithReplacedPlaceholder(
         MailDeliveryData $mail_data,
-        array $to_usr_ids,
-        array $cc_bcc_recipients
+        array $to_usr_ids
     ): void {
         foreach ($to_usr_ids as $user_id) {
             $recipient = new Recipient(
@@ -612,15 +619,24 @@ class ilMail
                 $this->getUserInstanceById($user_id),
                 $this->getMailOptionsByUserId($user_id)
             );
-            $recipients = $cc_bcc_recipients;
-            array_unshift($recipients, $recipient);
 
             $this->sendChanneledMails(
                 $mail_data,
-                $recipients,
+                [$recipient],
                 $this->replacePlaceholders($mail_data->getMessage(), $user_id),
             );
         }
+    }
+
+    private function sendMailWithReplacedEmptyPlaceholder(
+        MailDeliveryData $mail_data,
+        array $recipients,
+    ): void {
+        $this->sendChanneledMails(
+            $mail_data,
+            $recipients,
+            $this->replacePlaceholdersEmpty($mail_data->getMessage()),
+        );
     }
 
     private function sendMailWithoutReplacedPlaceholder(
--- a/Services/Mail/classes/class.ilMail.php
+++ b/Services/Mail/classes/class.ilMail.php
@@ -30,68 +30,57 @@ class ilMail
     public const ILIAS_HOST = 'ilias';
     public const PROP_CONTEXT_SUBJECT_PREFIX = 'subject_prefix';
 
-    protected ilLanguage $lng;
-    protected ilDBInterface $db;
-    protected ilFileDataMail $mfile;
-    protected ilMailOptions $mail_options;
-    protected ilMailbox $mailbox;
     public int $user_id;
-    protected string $table_mail;
-    protected string $table_mail_saved;
+    private string $table_mail;
+    private string $table_mail_saved;
     /** @var array<string, mixed>|null */
-    protected ?array $mail_data = [];
-    protected bool $save_in_sentbox;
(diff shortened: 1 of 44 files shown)


mjansenDatabay:
Thanks @shecken and @klees . I will have a look at this PR when I am back in February (mid of).

Services/Mail/classes/class.ilMailTemplatePlaceholderResolver.php
```diff
@@ -24,62 +24,36 @@
...
+        $this->mustache_engine = $mustache_engine;
     }
 
+    /**
+     * @param ilMailTemplateContext $context
+     * @param string $message
+     * @param ilObjUser|null $user
+     * @param array $contextParameters
+     * @param $replaceEmptyPlaceholders boolean
+     * @return string
```
mjansenDatabay: Most of the parameter descriptions are obsolete because there is no added value

- Please remove them except for `$contextParameters`
- Tryp to specify the shape of `$contextParameters`

Services/Mail/classes/class.ilMailTemplatePlaceholderResolver.php
```diff
@@ -24,62 +24,36 @@
...
-        }
-
-        return $message;
+        return $this->mustache_engine->render(
+            $message,
+            new ilMailTemplateContextAdapter(
+                [$context],
+                $contextParameters,
+                $user,
+                $replaceEmptyPlaceholders
```
mjansenDatabay: IMO the argument here does not match the defined parameter. The meanings of `$replaceEmptyPlaceholders` (passed) and `$html_markup` (expected) are different.

`$html_markup` was introduced in the initial mail template implementation but is actually not used/supported, yet.

mjansenDatabay (changes requested):
Hi @shecken and @klees,

thank you a lot for this contribution, I really appreciate this.

First of all I like the minimally invasive approach you chose by introducing the `Adapter`.

- [x] Why did you rename the placeholders? You replaced all abbreviations like `crs`, `ass` et cetera. Furthermore you renamed `firstname_last_name_superior`.
- [x] Why did you replace `[` and `]` with `&lbrace` and `&rbrace;` in **some** (but not all) language files? In general we as developers only need to change the **de** and **en** language files AFAIK.
- [x] Isn't it a problem when `[` and `]` are replaced and used to determine the amount of steps in the `getRemainingAmountOfSteps` methods of both migrations, without checking the content between these delimiters? `[` and `]` could be theoretically used in the subject/body without referencing a supported placeholder. IMO this step has to be a bit more elaborated by replacing/searching the supported strings only (I know that this will be a big list).

- [x] Please adjust the PHP code style, it seems to be not PSR-12 compliant here and there (e.g. the strict type declaration must be in a seperate line, see: https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements).
- [x] Instead of a plain SQL `LIKE` we should use `ilDBInteface::like`.

Kind regards,
@mjansenDatabay

klees:
Hi @mjansenDatabay,

I can answer two of these:

The internal keys for the placeholders have changed to align them with the ids that are displayed to the users. The previous implementation looped over all placeholders, took the `[display ids]`  and just tried to replace them. So basically, the keys only were used for internal identification. With mustache, the engine asks for placeholders via the `display id`, so the internal identifiers need to match up with the display ids.

The `[` have been replaced by `&lbrace;` because otherwise `ilTemplate` would (in some circumstances) just replace the `{...}` with empty strings...

The rest of the stuff sound plausible and we will look into it.

Best wishes!

kergomard:
Hi @mjansenDatabay 

When looking though the older PRs I found this one. Are you still working on this or should it maybe be closed?

Best regards!

kergomard:
Please ignore my previous comment (I removed it). Got turned around by our nice new link ;-)

shecken:
Hi @mjansenDatabay,

would you please look at that again.

mjansenDatabay:
@shecken (FYI @klees): As discussed on `Discord`: Please add the missing changes regarding the content for the CC/BCC recipients for the case that placeholder replacement (= `Mustache`) should be used.

Currently (in ILIAS 8.x and current `trunk`) empty placeholders are **not** replaced and all CC and BCC recipeints receive the **same** content in the **same** physical email.

`\ilMail::distributeMail`:
 
```php
// ...
            $this->sendChanneledMails(
                $to,
                $cc,
                $bcc,
                $toUsrIds,
                $subject,
                $message,
                $attachments,
                $sentMailId,
                true
            );

            $otherUsrIds = $this->getUserIds([$cc, $bcc]);
            $this->logger->debug(sprintf(
                "Parsed CC/BCC user ids from given recipients for serial letter notification: %s",
                implode(', ', $otherUsrIds)
            ));

            $this->sendChanneledMails(
                $to,
                $cc,
                $bcc,
                $otherUsrIds,
                $subject,
                $this->replacePlaceholders($message, 0, false), // <----------------------------------------------
                $attachments,
                $sentMailId
            );
// ...
```
If `Mustache` does **not** support  placeholders which are **not** replaced, we'll have to make a change here.

Services/Mail/classes/class.ilMail.php
```diff
@@ -549,164 +541,148 @@ private function sendInternalMail(
...
+                $user_id,
+                $this->getUserInstanceById($user_id),
+                $this->getMailOptionsByUserId($user_id)
+            ),
+            $other_usr_ids
+        );
+        $this->logger->debug(sprintf(
+            "Parsed CC/BCC user ids from given recipients for serial letter notification: %s",
+            implode(', ', $other_usr_ids)
+        ));
```
mjansenDatabay, line 597: According to the JF discussion (see: https://docu.ilias.de/goto_docu_wiki_wpage_7969_1357.html) CC/BCC recipients **MUST NOT** receive individual copies (with interpolated placeholders) of the mails.

Means: We should keep the old behaviour and use **one** message body for **all** CC/BCC recipients. invidual placeholders should be replace with the name of the placeholder:

```
{{FIRSTNAME}} => FIRSTNAME

mjansenDatabay (changes requested):

mjansenDatabay:
Thx @shecken and @klees . I think we are ready to merge this :-). I'll do this next week.

mjansenDatabay (approved):

mjansenDatabay:
@smeyer-ilias / @alex40724: This PR changes some mail related files in your components (implementations of `ilMailTemplateContext`).
Furthermore a tiny change in `ilObjSurvey` was necessary to change the placeholder interpolation to be compatible with the new mechanisms of the mail component.

@kergomard In `Services/User` a new database update step class is introduced to migrate existing `New Account Mails`. 

I would like to merge this PR by the end of this week, so a short review is highy appreciated.

kergomard (commented):
Hi @shecken 
Thank you very much for this change and sorry for having missed my task here!
Changes look mostly good to me, one question though:
- Why are we only changing four placeholders in [`ilUpdateNewAccountMailTemplatesForMustache::step()`](https://github.com/ILIAS-eLearning/ILIAS/pull/5488/files#diff-4f0ac523721b2ed9ce4957650e4be2bb5439252a513f6c3d8866d4963f3c29f9R58)? First I thought that the other changes would be made elsewhere, but I do not see where. Sorry, if I'm missing something obvious here! I didn't read through the whole thing, but did just a little bit of searching.

Thanks again and best,
@kergomard

shecken:
Hello @kergomard,

i've added a replacement step for every other existing placeholders. This is done last, because of the syntax of optional blocks in mustache. (https://github.com/ILIAS-eLearning/ILIAS/pull/5488/commits/6986eaadc6afb38ff5b28919f6204b7eb99452b5)

kergomard (commented):
Thank you very much @shecken !

Two little questions/remarks:
- [ ] Shouldn't we maybe force case sensitivity for [this check](https://github.com/ILIAS-eLearning/ILIAS/pull/5488/files#diff-4f0ac523721b2ed9ce4957650e4be2bb5439252a513f6c3d8866d4963f3c29f9R116)? If I'm understanding [`REGEXP_REPLACE()`](https://mariadb.com/kb/en/regexp_replace/) right, this would only be true, if the collation is case sensitive.
- [ ] Does the `$1` also work on MariaDB? Again, if I'm getting the docu right, it would need to be `\\1`, but I'm probably missing something.
Thanks again and best,
@kergomard 

Thank you very much and best,
@kergomard

klees:
Hi @kergomard,

> Does the $1 also work on MariaDB? 

Yes, seems to be fine. We have tested the changes in [doil](https://github.com/conceptsandtraining/doil) which indeed uses MariaDB (10.5.19-MariaDB from Debian in my case, to be precise...) internally.

> Shouldn't we maybe force case sensitivity...

Maybe I do not understand something, but placeholders are in capitals always, right? So why should we force case sensitiv?

Kind regards!

kergomard (approved):
Hi @klees 

Thanks for checking the `$1` !

My comment was the other way around: Exactly because placeholders are always uppercase we should maybe make sure that the search would NOT replace lowercase occurrences of the strings. As I understand it, the current implementation would also replace lowercase placeholders. But this for me is not a condition as I'm not very concerned about it.

Thank you very much and best,
@kergomard

PR #5489, opened 2023-01-27
Reduce minimum size of bar charts

tfamula:
This PR reduces the minimum size of bar charts. The practical work with the bar charts has shown, that it was chosen too big in the first implementation. Especially if only a small number of bars is used in the chart, the bars appear oversized and take much space of the screen.

Current implementation:
bars_big

Proposal:
bars_small

### DIFF
--- a/src/UI/examples/Chart/Bar/Horizontal/custom.php
+++ b/src/UI/examples/Chart/Bar/Horizontal/custom.php
@@ -76,7 +76,7 @@ function custom()
     $dataset = $dataset->withPoint(
         "Item 6",
         [
-            "Target" => [-0.01, 0.01],
+            "Target" => [0, 0.01],
             "Dataset 1" => 0,
             "Dataset 2" => 3,
             "Dataset 3" => 0.2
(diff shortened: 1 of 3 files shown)


src/UI/examples/Chart/Bar/Horizontal/custom.php
```diff
@@ -76,7 +76,7 @@ function custom()
     $dataset = $dataset->withPoint(
         "Item 6",
         [
-            "Target" => [-0.01, 0.01],
+            "Target" => [0, 0.01],
```
tfamula, line 79: This PR also fixes a little bug in the example.

tfamula:
The tests fail because of:

> Copyright not as expected in src/UI/examples/Chart/Bar/Horizontal/custom.php

But AFAIK we do not use the Copyright in the example files. How can I avoid the fail?

klees:
@tfamula We indeed tried to disabled checks for the examples, as far as I remember. Seems we didn't succeed... :grimacing: 

I guess its best to just ignore the failing copyright check for now.

Amstutz:
Sry, for the fat finger in the first try ;-)

Thx a lot @tfamula for the contribution. Ill push to trunk as well.

Amstutz:
Note, we excluded the exceptions only for trunk. IMO this is fine. Most of the upcoming UI Components PRs involving examples will be trunk and then maybe pushed down.

PR #5490, opened 2023-01-30
PRG: 35479 fire events after db is updated, introduce internal handler

nhaagen:
https://mantis.ilias.de/view.php?id=35479
https://mantis.ilias.de/view.php?id=35969
https://mantis.ilias.de/view.php?id=35962


klees:
Picked to trunk.

PR #5494, opened 2023-01-30
Trunk: Fix missing array key exists check

mBeym:
Mantis Tickets: 
- https://mantis.ilias.de/view.php?id=36492
- https://mantis.ilias.de/view.php?id=36313
- https://mantis.ilias.de/view.php?id=36459


kergomard:
Thank you very much @mBeym !
One question: How sure are we, that the key $activeId really, really always exists? Shouldn't we test for both levels?

mBeym:
isset checks both on it's own.

kergomard:
Thank you @mBeym ! Learned something!

PR #5495, opened 2023-01-30
T&A Bugfix #0036331: SC-Question: Trying so save question in "Multi Line Answers" mode leads to fatal

fhelfer:
https://mantis.ilias.de/view.php?id=36331


dsstrassner:
This also blocks Testcases :(

PR #5498, opened 2023-01-30
PHPStan-Rule Legacy UI

chfsx:
As just discussed in the Legacy UI project, here is the PR with the PHStan rules for the first components (to be removed with ILIAS 9).

The current report can be downloaded here: https://github.com/srsolutionsag/ILIAS/actions/runs/4044385194 (with the tree rules presented in the meeting, but Button-usages is in it).
After merging, a new Report will be generate on schedule every day (with the rules for ILIAS 9).

### DIFF
--- a/.github/workflows/legacy-ui.yml
+++ b/.github/workflows/legacy-ui.yml
@@ -1,14 +1,8 @@
 name: legacy-ui-checks
 on:
-  workflow_dispatch:
-    inputs:
-      branch:
-        description: 'Branch to run checks on'
-        required: true
-        default: 'trunk'
   schedule:
-     # Runs every day bright and early
-    - cron: '0 3 * * 0'
+     # Run every 12 hours
+    - cron: '0 3,15 * * *'
 permissions:
   contents: read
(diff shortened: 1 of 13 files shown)


chfsx:
We still have a small "problem": scheduled github actions are only executed if the corresponding workflow file is in the configured default branch (currently release_7). this means we will have to merge this PR down to release_7, we can adjust the checkout of the source code so that 'trunk' is always examined. are there any opinions or counter-opinions on this?

https://stackoverflow.com/questions/63436541/github-action-workflow-schedule-not-working-on-non-default-branch

matthiaskunkel:
Jour Fixe, 06 FEB 2023: Fabian notified us about new reports about the usage of LegacyUI in the ILIAS repo 'trunk'. These reports should help developer to get rid of these components. Reports are generated each morning at 3 a.m. and available here: https://github.com/srsolutionsag/ILIAS/actions/runs/4044385194 . An additional notification will be sent to the admin list this week. Please merge the PR to trunk.

PR #5499, opened 2023-01-30
Bugfix 36373: shorten table id

schmitz-ilias:
This PR fixes [36373](https://mantis.ilias.de/view.php?id=36373).


schmitz-ilias:
Fixed by someone else at some point in the last three weeks.

PR #5500, opened 2023-01-31
CI: Let Captain Hook Just Fix the Coding Style

klees:
Hi folks,

could we please allow Captain Hook to just fix the coding style? Currently it only complains about non-compliance to the coding style and then I go on and fix the style by running that same command (minus a `--dry-run`) again. If we'd just allow Captain Hook to fix the style I (and presumably others as well...) will have one step less in my workflow and the style will end up fixed anyway.

Best wishes!

### DIFF
--- a/captainhook.json
+++ b/captainhook.json
@@ -36,7 +36,7 @@
         "conditions": []
       },
       {
-        "action": "libs/composer/vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix --config=./CI/PHP-CS-Fixer/code-format.php_cs --using-cache=no --dry-run -vvv {$STAGED_FILES|of-type:php}",
+        "action": "libs/composer/vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix --config=./CI/PHP-CS-Fixer/code-format.php_cs --using-cache=no -vvv {$STAGED_FILES|of-type:php} && git add {$STAGED_FILES|of-type:php}",
         "options": [],
         "conditions": [
           {


chfsx:
I am generally not a friend of actions that change the code or at least the git stage. i can understand the use case very well, but changing the stage should be left to the human and not done by an automatism. if i understand it correctly, "manual" work is still necessary afterwards: the action changes the code style, but these changes are then not yet part of the git stage but would be in the git worktre, right?

but if this is of general interest, i can live with it :-)

klees:
@chfsx: Ah right, changes would need to be added to the stage as well. Changed accordingly...

chfsx:
And that's exactly my point, you no longer commit what you wanted to commit :-) but as I said, if it makes the workflow easier for many, I won't oppose it.

klees:
I really don't get how this is any different from your PHPStorm workflow. PHPStorm also manipulates your files automatically. You then no longer save to disk what you wanted to save to disk...

klees:
Would you feel more comfortable with something like [just](https://github.com/casey/just)? Or does someone maybe know a good tool to watch filesystem changes and apply actions when things change?

chfsx:
In PHPStorm, it is up to you to decide whether you want to set up a so-called FileWatcher that directly applies the CodeStyle while editing. Otherwise it is always a manual process (key combination) that triggers this. In my case, it is always a conscious application of the style.

I don't know "just", what would be the idea of integrating this? would it then simply be a faster way to use commands like `CI/Special-Char-Checker/special-char-checker.sh`, `CI/PHP-CS-Fixer/run_check.sh` or `libs/composer/vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix --config=. /CI/PHP-CS-Fixer/code-format.php_cs --using-cache=no -vvv`? something like this would perhaps really serve us all, as there are more and more scripts that you want to run regularly and so you could address them via a somewhat clearer infrastructure.

mjansenDatabay:
The suggested change is already (partly) part of https://github.com/ILIAS-eLearning/ILIAS/pull/3839

klees:
@chfsx Yes, `just` is supposed to make common workflows accessible with simple commands like `just fix-style`. Some projects use `make` for a similar purpose and `just` just ditches the build part of make.

Amstutz:
no opinion with just. sounds neat...

however, I side with @chfsx concerning automatically apply fixes to commits. as @chfsx I like to control what I commit.

klees:
Ok, sounds like there is at least some "only check"-faction for automatisms =)

Will have a look into `just`.

klees:
I withdraw this for the time being =)

thojou:
Hey @klees,

While working on my own PR related to this topic, @mjansenDatabay pointed out that there are already two other PRs addressing the same issue. I’d like to request that we reopen this PR to integrate your changes into the current stable and trunk branches.

I understand that there might be different preferences regarding this, but I believe the current implementation is incomplete. At present, the process only halts further commits if coding style violations are detected. When this happens, like in your workflow, I manually copy the printed command, remove the --dry-run flag, run the command, and then retry the commit. The commit succeeds, but when I push the changes to GitHub, the pipeline fails because the PHP-CS-Fixer rewrites weren’t included before the commit.

This issue occurs because, after PHP-CS-Fixer modifies the files, they exist in two stages, but we only see one view of the "current code."

In my opinion, this partial automation adds unnecessary overhead. I also believe the benefits of fully automating this process outweigh concerns about losing control over the code. As you mentioned, enforcing coding styles isn't typically something developers handle manually. Additionally, since the CI pipeline checks for coding style compliance, committing code with invalid styles should be prevented. (I believe this was the original intent of integrating the pre-commit hook.) Otherwise, users would need to either add an extra CS-Fix commit or rewrite the repository history to integrate those changes into the previous commit.

klees:
Hi @thojou,

you're very welcome to move on here =)

Thanks!

PR #5501, opened 2023-01-31
Forum Bugfix #0036483: Latest articles links not working

fhelfer:
https://mantis.ilias.de/view.php?id=36483

@mjansenDatabay There are other solutions maybe we can discuss this sometime

### DIFF
--- a/Services/Object/classes/class.ilObjectGUI.php
+++ b/Services/Object/classes/class.ilObjectGUI.php
@@ -1446,7 +1446,7 @@ class ilObjectGUI
         $obj_id = ilObject::_lookupObjId($this->object->getRefId());
         $obj_type = ilObject::_lookupType($obj_id);
 
-        if ($this->ctrl->getNextClass() != "ilcolumngui") {
+        if ($this->ctrl->getNextClass($this) != "ilcolumngui") {
             // normal command processing
             return $this->getContent();
         } else {


mjansenDatabay:
Thx @fhelfer . The question is: Why does it work in ILIAS 7? Maybe changes should be applied to the news component, not to the forum. If it works in ILIAS 7 without this special news handling in the forum, I suspect the root of all evil is located in the news component.

fhelfer:
@mjansen It seems to be because of ilCtrl getNextClass Changes,
https://github.com/ILIAS-eLearning/ILIAS/blob/cbce4a6a80e9c43359ebd35f702a15dd3ecb3ae7/Services/Object/classes/class.ilObjectGUI.php#L1449
this has to be `$this->ctrl->getNextClass($this)` in order to work as in ilias7

mjansenDatabay:
If approved, this has to be picked to `release_8` as well.

mjansenDatabay:
@kergomard Since this change affects `ilObjectGUI`, I added you as the assigne here. The issue seems to be caused by `ilCtrl` changes, most probably caused by the `ilCtrl` revision by @thibsy . Maybe Thibeau has an idea why the behaviour changed so that `$this` has to be passed in ILIAS >= 8.x.

kergomard:
@thibsy : Yes, I would be very glad to have a second opinion on this, as I'm unsure if this can have side-effects. I will merge and cherry-pick thereafter.

thibsy:
Hi @mjansenDatabay and @kergomard

Sorry it took a while, but I finally managed to look into this.

I looked into the behaviour of `ilCtrl::getHTML()` and recognized the issue to be an `ilCtrl` bug.
Hence I close this PR, but thanks for looking into it anyways.

Kind regards!

PR #5502, opened 2023-01-31
Mail, 36487: encode and space ILIAS_URL's client id.

nhaagen:
https://mantis.ilias.de/view.php?id=36487

I'm a little unsure about this one; 
the placeholder [ILIAS_URL] is resolved to ilUtil::_getHttpPath() . '/login.php?client_id=' . CLIENT_ID;
A HTML-Mail will then be refined by String/MakeClickable;
the Text to transform is
`'http://ilias.de/login.php?client_id=test8: .....' `
and will be resolved to
`<a href="http://ilias.de/login.php?client_id=test8:">.....`
Also, when there is a space in the client id: 
`'http://ilias.de/login.php?client_id=test 8: .....' `will lead to `<a href="http://ilias.de/login.php?client_id=test"> 8: .....`

This PR encodes the client_id and adds a space after the placeholder, so that the transformation will produce a valid link.

### DIFF
--- a/Services/Mail/classes/class.ilMailTemplateContext.php
+++ b/Services/Mail/classes/class.ilMailTemplateContext.php
@@ -162,7 +162,8 @@ abstract class ilMailTemplateContext
             case 'ilias_url' === $placeholder_id:
                 $resolved = $this->envHelper->getHttpPath()
                     . '/login.php?client_id='
-                    . $this->envHelper->getClientId();
+                    . rawurlencode($this->envHelper->getClientId())
+                    . ' ';
                 break;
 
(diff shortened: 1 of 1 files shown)


klees:
@nhaagen, @mjansenDatabay: I guess we could drop the client_id, shouldn't matter anyway...

mjansenDatabay:
@klees / @nhaagen Yes, Richard is right IMO. We could remove the client_id query parameter here.

nhaagen:
Removed the client_id; for that matter, hard pointing to login.php is not necessary either, right?

PR #5505, opened 2023-01-31
PRG: 35891, do not recalc LP-Nodes, use constant for completion_by

nhaagen:
https://mantis.ilias.de/view.php?id=35891


klees:
Picked to trunk.

PR #5507, opened 2023-01-31
[fix] jump to question list

kunzeus:
Hello,

https://mantis.ilias.de/view.php?id=36508

This fixes the problem that the: "Back to Question List" buttons do not jump to the beginning of the question list, but to the end. I asked if jumping to the top of the page as described in Mantis is really desirable, or if jumping to the top of the question list would be the better behavior. It was decided to jump to the question list.

As always let me know where I can improve, I still have a lot to learn.


kergomard:
I had a closer look at the issue and don't believe this to be the right solution from a usability point of view. I think the error is very subtle: The anchor should actually not link to the top of the table, as this would make you loose your position if your test has a lot of questions. It should rather link to the top of the corresponding table row (see behavior in ILIAS 7) in ILIAS 8 it right now jumps to the bottom of the row. I hope I explain this more or less well otherwise don't hesitate to contact me. Would you be willing and able to give this another look?

Thanks again and best,
@kergomard

dsstrassner:
Thanks for looking into this. I second @kergomard.

kergomard:
Thank you very much @kunzeus for looking for a good solution!
Now: The inline-style really hurts, when looking at it. I would suggest to add

`#tst_pass_details_overview tr {
    scroll-margin-top: 30px;
}`

To the delos.less of the test and then to compile it. What do you think?

Thanks again and best,
@kergomard

kunzeus:
I agree it hurts, I'm looking into less and how it is used in ILIAS.

Amstutz:
no delos.less in trunk anymore. it is all ITCSS now. For ILIAS 8, less is still valid. See: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/templates/Guidelines_SCSS-Coding.md

dsstrassner:
I always forget that for trunk and ITCSS we need to add @Amstutz as Assignee. He needs to approve all changes there.

mbecker-databay:
@Amstutz I am good with it, please merge once you approve.

Amstutz:
Thx a lot. Merged to trunk. I guess this will need a pick to 8 as well. Note that we are working with less in 8. Should I pick this, if somebody else picks, feel free to directly push to 8.

kergomard:
Done.

PR #5512, opened 2023-01-31
EmployeeTalk: fix icons

schmitz-ilias:
This PR fixes a bunch of icons in EmployeeTalk, see [35564](https://mantis.ilias.de/view.php?id=35564). For some reason one is hiding in `Modules\OrgUnit`. I also added the icon for the EmployeeTalk Templates Administration to the standard KS icons. I'm not quite sure what the criteria are there, but that one seems to fit.

I also noticed that there is no icon yet for the Employee Talk objects themselves. I added a preliminary one to this PR (just the template icon inverted, as for Portfolios). If there is a good reason to not add a new icon I could also reuse existing ones, but I'd  prefer having a separate icon to distinguish between templates and the real thing.
(I hope you are the right person to tag for icons @matthiaskunkel, if not feel free to reassign.)

### DIFF
--- a/Modules/EmployeeTalk/classes/Provider/MyStaffListEntryProvider.php
+++ b/Modules/EmployeeTalk/classes/Provider/MyStaffListEntryProvider.php
@@ -58,7 +58,10 @@ final class MyStaffListEntryProvider extends AbstractStaticMainMenuProvider
 
         $title = $this->dic->language()->txt("mm_org_etal");
         $action = "ilias.php?baseClass=ilAdministrationGUI&ref_id=" . ilObjOrgUnit::getRootOrgRefId() . "&cmd=jump";
-        $icon = $this->dic->ui()->factory()->symbol()->icon()->standard('etal', $title);
+        $icon = $this->dic->ui()->factory()->symbol()->icon()->custom(
+            \ilUtil::getImagePath('icon_etal.svg'),
+            $title
+        );
 
         $items[] = $this->mainmenu->link($this->employeeTalkTemplateIdentifier)
                                   ->withAlwaysAvailable(false)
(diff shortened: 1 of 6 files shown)


matthiaskunkel:
Hi @schmitz-ilias,

I recently committed the two icons icon_tala.svg (Talk Templates Administration) and icon_talt.svg (Talk Template) to the images folder. Both have the same look. icon_tala.svg can also be used in the slate where currently the placeholder icon is still used. I wasn't aware that there is also a need for an inverted version. But if icon_etal.svg is used for a distinct employee talk that was derived from a template, the design (and iconography) is correct. It is the same behaviour like portfolio templates and portfolio.

klees (approved):
Hi @schmitz-ilias, hi @tfamula,

LGTM. Besides this change here, we should have a conversation about how to handle the interface between Org-Units and MyStaff sometimes soon. I expect that this wasn't the last location where code from MyStaff/EmployeeTalk resides in `Modules\OrgUnits`. Also I expect there to be huge dependencies between the components, maybe even down to the database layer. We are already [looking to improve the database layer of the OrgUnits](https://github.com/ILIAS-eLearning/ILIAS/pull/5460) and I wonder how well prepared you are to move MyStaff/EmployeeTalk along. Maybe you even have ideas or requirements for concrete changes in the OrgUnit code. I don't expect you to know any of this just yet, but I thinks it's important that we are considerate regarding the dependencies we most certainly have and start to straighten things up.

Best wishes!

schmitz-ilias:
@klees: A well defined interface between OrgUnits and MyStaff/EmployeeTalk is certainly in everyone's best interest. We'll keep this aspect in mind while bugfixing in the next few weeks, and I'm sure coming out of that process we will have a better idea about how we want to handle this on our end :)

@matthiaskunkel: While a separate icon for the talks is maybe not strictly necessary right now (as you said, the admin icon could be used in the slate and for the header of the talks list and it would probably not lead to too much confusion), I prefer adding the new icon for clarity and future-proofing. (Especially since one basically gets it for free as a derivative from the icons you added ^^)

@amstutz: Are you responsible for commits to `templates/default/images` as the css maintainer? Not to be overly careful, but I don't want to go over anyone's head here ^^

Amstutz:
Yes, I am. thx for the ping. LGTM. For the future, if @matthiaskunkel or @yvseiler are fine with icon changes, they are fine for me as well. Just one little thing, however, this is in your responsibility:

image

Maybe consider using the Constant from the Icon interface instead of the string here. If for some reason, someone changes the Icon in the future, this will quickly be spoted by e.g. inspections, if the constant does not exist anymore.

Amstutz (approved):

schmitz-ilias:
@Amstutz thanks for the tip, I'll look into that after merging

PR #5513, opened 2023-02-01
Fix check if week wraps year to only trigger at new year

pascalseeland:
The current check triggers every time a week spans two month, but should only trigger if its actually a year change with $last_month > $now_month


chfsx:
Thank you for the PR, as @mjansenDatabay is currently absent, I will take over the merging on my own responsibility. @mjansenDatabay  may hit me when he is back, if that was not good :-) 

So new (rebased) pushes should not fail from now on. I also cherry-pick in [trunk](https://github.com/ILIAS-eLearning/ILIAS/commit/847913d5315babb170e388e10716646fe613763d).

PR #5514, opened 2023-02-01
PRG: some gui-functionality fixed (36440, 36534, 36537, 36573)

nhaagen:
https://mantis.ilias.de/view.php?id=36440
https://mantis.ilias.de/view.php?id=36534
https://mantis.ilias.de/view.php?id=36537
https://mantis.ilias.de/view.php?id=36573


klees:
Picked to trunk.

PR #5515, opened 2023-02-01
[FIX] 0036179: Failed test: Prüfungsblatt (optional/obligatorisch) - …

chfsx:
…Byte vs Megabyte

@klees may you have a look at this? Not the solution we want for the future, but maybe for the moment. background is https://mantis.ilias.de/view.php?id=36179 which you already know. thanks in advance!

most crucial part is here: https://github.com/ILIAS-eLearning/ILIAS/compare/release_8...srsolutionsag:ILIAS:fix/0036179/8/size-formatting?expand=1#diff-f3892e7d92b0895c6b153d2f1754b755fc888a4b5199b1d4f8784a7344b81d5fR849

PR #5516, opened 2023-02-01
Implemented HumanReadable DataSize in __toString 

chfsx:
@klees best will be we use this new `HumanReadable` Transformation for #5515 to avoid using ilUtil again...


chfsx:
@klees as discussed I implemented this directly in __toString()

klees:
Picked to trunk with another commit message =)

PR #5517, opened 2023-02-01
add ADN toast provider for GS

iszmais:
Due to changes made inside the load of toast an provider interface was established to porvide the initial toasts without another request: https://github.com/ILIAS-eLearning/ILIAS/pull/5121

Therefore every provider of toast is obliged to provide a new toast Provider if his service want to add toast to the page.
ATM this only applies to initial toast but will probably be extended to all toast in near future.

This PR is an Proposal for toast inside your Service. Feel free to use and manipulate this template to your fitting and provide a PR for Trunk.
If you dont want to provide Toast with your Service feel free to cloase this PR.

### DIFF
--- a/Services/AdministrativeNotification/classes/GlobalScreen/ADNToastProvider.php
+++ b/Services/AdministrativeNotification/classes/GlobalScreen/ADNToastProvider.php
@@ -1,0 +1,70 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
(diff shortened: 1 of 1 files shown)


alex40724:
@chfsx This service is missing a maintenance json, but it is your code, right?

chfsx:
HI @iszmais and thanks a lot for the PR. one question: would it require a feature request and decision from the JF to adopt this or can maintainers decide and include this on their own?

iszmais:
@chfsx I guess for new Toasts this would require a decision from the JF but im not fully familiar this this progresses either.
@matthiaskunkel Could you provide some info to this?

matthiaskunkel:
Is this PR a change of a public interface? In this case, it needs to be presented at the Jour Fixe to make every developer aware of this change. A FR seems not be necessary IMHO as it does not looks like a proper feature.

mjansenDatabay:
IMO the situation is this: Administrative notifications do **not** provide any toasts until now (ILIAS 9 development). If they should provide any toasts, this is a change request and thus a feature wiki page should be created.

iszmais:
Due to changes in https://github.com/ILIAS-eLearning/ILIAS/pull/5725 this propasal in no longer valid.

If you are interested in Adding Toast to the ADN feel free to take a look ad the updated [README](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/GlobalScreen/Scope/Toast/README.md) or contact me or @chfsx for further details!

PR #5529, opened 2023-02-06
[FIX] 0036605: Updating ILIAS without MathJax config not possible

chfsx:
This fixes an issue described in https://mantis.ilias.de/view.php?id=36605


fneumann:
Thanks to @chfsx!

PR #5533, opened 2023-02-06
chatserver-README: navigation path to admin-GUI corrected

utesche:
.. a small typo as addressed by title above.

Edit:
Additional the first two screenshots should be replaced with recent ones - since ILIAS 7 there is no GUI configuration of the service itself: the sub-tab entry "Chatserver Settings" was removed and in consequence the sub-tab at all.


utesche:
@mjansenDatabay ,
I found some more aspects of the README to be discussed, currently I'm collecting them and will PM you...

This PR is closed now  
(changes to this README are processed by changing https://docu.ilias.de/goto_docu_pg_32889_367.html anyway, right?)

PR #5534, opened 2023-02-06
PRG: refine events

nhaagen:
This introduces delayed events (collect all events, trigger later) and handles internal events as such.


klees:
Picked to trunk.

PR #5540, opened 2023-02-07
Trunk: Add showing error message when test was already finished

mBeym:
Mantis Ticket: https://mantis.ilias.de/view.php?id=36512

- Clicking on the `Finish Test`, `Next question` or `Previous question` button will now show an error message if the user has no unfinished passes currently (meaning it was already finished from somewhere/someone else.


dsstrassner:
Hi @mBeym

thanks a bunch! Could this fix easily be cherry-picked to ILIAS 7

PR #5541, opened 2023-02-07
Setup-README: typo in internal link

utesche:
... to section of command 'achieve'.


dsstrassner:
Hi @klees 
in the near feature we will assign something like this to Ann-Christin, but for now, I think this is for you? Or do I miss something?

klees:
Thx @utesche!

PR #5543, opened 2023-02-08
UI/Symbol/Icon/Standard added status icons

catenglaender:
# Issue

the Standard Icons from Kitchen Sink are missing some icons to depict a larger variety of status information.

# Change
We propose adding the following Milos icons from the default image folder:
image
* icon_alert.svg
* icon_time.svg
* icon_checked.svg
* icon_not_ok.svg

# Larger Context

For the UI Launcher an icon is supposed to provide more information about the status of the connected process. We believe that the icons can be of value in a variety of use cases, for example:
* User is alerted to preconditions or prompted to take an action elsewhere > icon alert
* User is informed about an approaching or passed deadline > icon time
* User is informed that they already completed the connected process > icon checked
* User has failed the connected process > icon not ok

# Details & Conflicts

## alert css class

The name of the alert icon causes the renderer to use a css class named "alert", but this class is already styled by Bootstrap 3. As a quick and simple fix I counterstyled with the more specific selector ".icon.alert" to override the unwanted styling. When we make the alert system independent from Bootstrap, this issue could be addressed properly and the counterstyling could be removed.

## img tag styling

The checked icon was also already styled by a selector targeting the image name using a very broad `img[src$="icon_checked.svg"]`. I believe that this checkbox is no longer in use in the header/metabar, so I removed this styling. Should this cause visual glitches then the styling of that checked icon should be addressed more specific to this context with no side effect to the icon image in general.


klees (commented):
Hi @catenglaender,

to me it feels as if these should be `Glyph`s rather than `Icon`s. Icons "indicate the functionality or nature of the element they illustrate", while glyphs "map a generally known concept or symbol to a specific concept in ILIAS". The thingies here rather seem to be the latter than the former, right?

Best wishes!

Amstutz:
Hi @catenglaender 

@klees and @Amstutz had a longer discussion concerning the Glyph vs Icon topic. According to the current discussion, those new elements belong to neither. However, we agree that those seem rather to fit into what we understand as glyph. 

Please change or describe why you would rather not/prefer another solution:
- [ ] Changing the Glyph Description: Removing the parts on interactivity on Glyph. This seems to restrict the nature of the Glyph too far.
- [ ] Roadmap: If you change the description of Glyph, please also add an entry in the roadmap, explaining, that we need to check our usages of custom icons for elements, that should be migrated towards glyphs.
- [ ] Implement as Glyph: Implement the instances here as Glyph

Thx a lot for your feedback.
@klees and @Amstutz

Amstutz:
The changes in the PR are substantial and result in a more or less completely revised PR, I will close this for the moment to keep the list of open PRs short. Feel free to re-open if we can do something further here.

PR #5544, opened 2023-02-08
[FIX] UI: required file inputs will work now.

thibsy:
Hi all,

I noticed that `ILIAS\UI\Implementation\Component\Input\Field\File::getConstraintForRequirement()` returns a `ILIAS\Refinery\String\Group` instead of a `Constraint`. This PR introduces a proper constraint, similar as done in trunk (see [here](https://github.com/ILIAS-eLearning/ILIAS/blob/f75bea22c088dcdd24c2daa7eca811b04b54be8c/src/UI/Implementation/Component/Input/Field/File.php#L132)).

Kind regards,
@thibsy


klees (approved):
LGTM, thx!

PR #5546, opened 2023-02-08
[FIX] UI: file input will now call file-removal endpoint.

thibsy:
Hi all,

I noticed that the file-removal endpoint of `ILIAS\UI\Component\Input\Field\UploadHandler` when using file inputs is never called, even though files are successfully uploaded.

This is due to [this](https://github.com/ILIAS-eLearning/ILIAS/blob/d0dff3d405a9ea95a847df9106a0a24f1dc4b4ca/src/UI/templates/js/Input/Field/file.js#L132) if-statement, which checked an attribute that's never set. This PR fixes the issue by setting this attribute in the success hook after a file has been successfully uploaded.

Kind regards,
@thibsy

### DIFF
--- a/src/UI/templates/js/Input/Field/file.js
+++ b/src/UI/templates/js/Input/Field/file.js
@@ -84,6 +84,7 @@ il.UI.Input = il.UI.Input || {};
 				clone.attr('data-file-id', new_file_id);
 
 				files.file_id = new_file_id;
+				files.is_existing = true;
 
 				$(container).append(clone);
 			};


klees:
Hi @thibsy,

as briefly discussed via chat, I will close this PR, as I consider this a dangerous change.

As far as I understand the state of the implementation, the file-removal end-point will be called asynchronously, when the [x] next to the file in the form is clicked. This IMO breaks a deeply rooted assumption, i.e.: forms are only processed once they are submitted. If implemented, the removal endpoint will already be called before the form is submitted, and the file will be gone although the user never submitted the form.

I think we shouldn't introduce this change in a productive release, as we put users at risk to lose data.

Thanks for looking into this, anyway.

Best wishes!

PR #5548, opened 2023-02-09
[FIX] 0036668: .DS_Store Files break Objective

chfsx:
Fix for https://mantis.ilias.de/view.php?id=36668

### DIFF
--- a/Services/Component/classes/Setup/class.ilComponentBuildPluginInfoObjective.php
+++ b/Services/Component/classes/Setup/class.ilComponentBuildPluginInfoObjective.php
@@ -13,8 +13,7 @@
  * https://www.ilias.de
  * https://github.com/ILIAS-eLearning
  *
- ********************************************************************
- */
+ *********************************************************************/
 
 use ILIAS\Setup;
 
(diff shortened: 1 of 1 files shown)


klees (changes requested):
Hi @chfsx,

thanks a lot for looking into this!

The method you have modified serves as a slim wrapper around the according PHP methods to ease unit testing, so I'm not happy with the fact that business rules seem to start to creep in there.

Thus, I would suggest the following:
- [ ] is_dir: Add an `is_dir` wrapper, similar to the other PHP method wrappers.
- [ ] build: Add the business rule ("non-folders might exist but can be ignored") in the initial `build` method using that wrapper, instead of adding it to `scanDir`.
- [ ] dot-files: Consider one additional business rule: "hidden files or directories (starting with ".") can be ignored too" and implement it in `build` as well.

- [ ] unit test: Please write tests for the changes.

If this is just a drive-by fix I'm also happy if you just ignore my input. We'll look into this instead =)

Best wishes!

klees:
Hi @chfsx, we looked into your src/Filesystem suggestion, see @tbongers-cat feedback in #5633.

PR #5549, opened 2023-02-09
EmployeeTalk: replace ActiveRecord in update steps

schmitz-ilias:
This PR fixes [34063](https://mantis.ilias.de/view.php?id=34063) by getting rid of `ActiveRecord` methods from the `EmployeeTalkDBUpdateSteps`. Please have a quick look @klees as this involves `OrgUnit` tables.

### DIFF
--- a/Modules/EmployeeTalk/classes/Setup/class.ilEmployeeTalkDBUpdateSteps.php
+++ b/Modules/EmployeeTalk/classes/Setup/class.ilEmployeeTalkDBUpdateSteps.php
@@ -67,20 +67,22 @@ final class ilEmployeeTalkDBUpdateSteps implements \ilDatabaseUpdateSteps
         $this->useTransaction(function (\ilDBInterface $db) {
             $etalTableName = 'etal_data';
 
-            $db->createTable($etalTableName, [
-                'object_id' => ['type' => 'integer', 'length' => 8, 'notnull' => true],
-                'series_id' => ['type' => 'text', 'length' => 36, 'notnull' => true, 'fixed' => true],
-                'start_date' => ['type' => 'integer', 'length' => 8, 'notnull' => true],
-                'end_date' => ['type' => 'integer', 'length' => 8, 'notnull' => true],
-                'all_day' => ['type' => 'integer', 'length' => 1, 'notnull' => true],
-                'employee' => ['type' => 'integer', 'length' => 8, 'notnull' => true],
-                'location' => ['type' => 'text', 'length' => 200, 'notnull' => false, 'fixed' => false],
-                'completed' => ['type' => 'integer', 'length' => 1, 'notnull' => true]
-            ]);
+            if (!$db->tableExists($etalTableName)) {
+                $db->createTable($etalTableName, [
+                    'object_id' => ['type' => 'integer', 'length' => 8, 'notnull' => true],
+                    'series_id' => ['type' => 'text', 'length' => 36, 'notnull' => true, 'fixed' => true],
+                    'start_date' => ['type' => 'integer', 'length' => 8, 'notnull' => true],
+                    'end_date' => ['type' => 'integer', 'length' => 8, 'notnull' => true],
+                    'all_day' => ['type' => 'integer', 'length' => 1, 'notnull' => true],
+                    'employee' => ['type' => 'integer', 'length' => 8, 'notnull' => true],
+                    'location' => ['type' => 'text', 'length' => 200, 'notnull' => false, 'fixed' => false],
+                    'completed' => ['type' => 'integer', 'length' => 1, 'notnull' => true]
+                ]);
 
-            $db->addPrimaryKey($etalTableName, ['object_id']);
-            $db->addIndex($etalTableName, ['series_id'], 'ser');
-            $db->addIndex($etalTableName, ['employee'], 'emp');
+                $db->addPrimaryKey($etalTableName, ['object_id']);
+                $db->addIndex($etalTableName, ['series_id'], 'ser');
+                $db->addIndex($etalTableName, ['employee'], 'emp');
+            }
         });
     }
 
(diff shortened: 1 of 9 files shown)


klees:
Hi @schmitz-ilias,

would you be so kind and move these methods into single objectives along the lines of [the similar classes in AccessControl](https://github.com/ILIAS-eLearning/ILIAS/tree/release_8/Services/AccessControl/classes/Setup/)?

Best wishes!

schmitz-ilias:
> Hi @schmitz-ilias,
> 
> would you be so kind and move these methods into single objectives along the lines of [the similar classes in AccessControl](https://github.com/ILIAS-eLearning/ILIAS/tree/release_8/Services/AccessControl/classes/Setup/)?
> 
> Best wishes!

Done, I moved the creation of contexts and operations respectively to a new Objective for `OrgUnits`.

schmitz-ilias:
Hi @klees, did you have a chance to look at this yet? I'd like to get this into R8 and trunk sooner rather than later, to avoid further complications like those in #5570.

klees (changes requested):
Hi @schmitz-ilias,

I'm sorry that I lost track of this. Thanks for the reminder...

- [ ] **No Parent Context**: Why do we want to silenty discard objectives where the parent context does not exist? Isn't this a sign for an ill-configured objective?
- [ ] **No Context**: Same question: A missing context seems to indicate some problematic configuration. Why do we abort?

- [ ] **"New"**: Remove the 'New' in the name of the objectives. Regarding the objectives, we do not really care if the Op is "new" or not, we simply want to assert that is indeed registered.
- [ ] **Unit Tests**: Please write tests that exercise the objectives. We cannot really know if the objectives do the correct thing, but we can at least check that the code runs, parameters are passed along and cases are differentiated correctly.

 Thanks for improving our infrastructure! Feel free to contact me for further support, as always.

Best wishes!

schmitz-ilias:
I implemented your requested changes. You're right about the (parent) context, an ill configured objective should not fail silently. In these cases the objectives now throw an exception. 
Are exceptions the preferred way to go here? It is definitely more noticeable than before, but it does stop the whole setup dead in its tracks, as far as I can tell.

Thanks for the feedback, let me know if there is anything else.

klees (changes requested):
Hi @schmitz-ilias,

this looks almost good, but I still need to send you back once more. The [achieve method should be idempotent](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/Setup/Objective.php#L64), so IMO [this](https://github.com/ILIAS-eLearning/ILIAS/pull/5549/files#diff-48edc833d3076c5294584775cb62d6f024b7777d33f9d7bee83568c224adfe7aR82) and [that](https://github.com/ILIAS-eLearning/ILIAS/pull/5549/files#diff-47bdd1044a06fda9a65f55a7beb9cc9cad0df6818c76bcde3e3a3feef2ec62c9R67) should just `return`, as the environment looks as requested.

Thanks for looking into this again!

tfamula:
> The [achieve method should be idempotent](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/Setup/Objective.php#L64), so IMO [this](https://github.com/ILIAS-eLearning/ILIAS/pull/5549/files#diff-48edc833d3076c5294584775cb62d6f024b7777d33f9d7bee83568c224adfe7aR82) and [that](https://github.com/ILIAS-eLearning/ILIAS/pull/5549/files#diff-47bdd1044a06fda9a65f55a7beb9cc9cad0df6818c76bcde3e3a3feef2ec62c9R67) should just `return`, as the environment looks as requested.

Hi @klees and @schmitz-ilias,
shouldn't the `isApplicable()` method make sure, that the `achieve()` method is not executed more times than needed (by returning `false`)? I just wonder why the checks, which are already made in `isApplicable()` are also needed in `achieve()`. Other Objectives also don't do this.

schmitz-ilias:
> Hi @schmitz-ilias,
> 
> this looks almost good, but I still need to send you back once more. The [achieve method should be idempotent](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/Setup/Objective.php#L64), so IMO [this](https://github.com/ILIAS-eLearning/ILIAS/pull/5549/files#diff-48edc833d3076c5294584775cb62d6f024b7777d33f9d7bee83568c224adfe7aR82) and [that](https://github.com/ILIAS-eLearning/ILIAS/pull/5549/files#diff-47bdd1044a06fda9a65f55a7beb9cc9cad0df6818c76bcde3e3a3feef2ec62c9R67) should just `return`, as the environment looks as requested.
> 
> Thanks for looking into this again!

Done, thanks!

schmitz-ilias:
Hi @klees, did you have a chance to look at this yet?

klees (approved):
Hi @schmitz-ilias,

LGTM, thanks a lot.

I have squashed the OrgUnit commits and merged this to trunk and release_8.

Best regards!

klees:
Hi @tfamula,

in general you are correct. If some objective not `isApplicable`, then `achieve` should never be called. `isApplicable` and the idempotency of `achieve` serve a different goal, hence both exist:

* If an objective not `isApplicable` it won't need to be considered at all. E.g., not show up in the list of objectives that have been run, not being inspected for preconditions, etc. We simple don't need to bother with it anymore. This is good for the UX: Imagine if every objective that was ever achieved would always show up again in the list of objectives that is shown on update. Also, this improves the run time of `update`: We simply discard what we don't need to consider.
* Idempotency of `achieve` is required to be able to safely run some objective again to make things right. Imagine some situation where, for some reason, we simply don't know if some desired state of the system was achieved. If we would not require idempotency, we cannot decide if the objective needs to be run again: If the system is in the desired state, running the objective again would make things worse. We could, of course, inspect the code of the objective. But having a common abstraction ("idempotency") will scale a lot better.

I  hope this makes sense somehow. Feel free to continue asking if I there are open questions.

Best wishes!

PR #5550, opened 2023-02-09
T&A Bugfix #0036135: QP Sync Test not buildable skip

fhelfer:
https://mantis.ilias.de/view.php?id=36135


fhelfer:
Has to be cherry-picked to 7&8 aswell
Removed a disabled_form skip from displaying error messages

PR #5552, opened 2023-02-10
PRG: 36635, prevent error for mising default-language

nhaagen:
https://mantis.ilias.de/view.php?id=36635


klees:
Picked to trunk.

PR #5554, opened 2023-02-10
PRG: reset flag for sent mails

nhaagen:
... if validity or deadline changes to a point in the future; the user should receive mails again in this case.


klees:
Picked to trunk.

PR #5555, opened 2023-02-10
PRG: 36636, cert-gui w/o additional features

nhaagen:
https://mantis.ilias.de/view.php?id=36636
https://mantis.ilias.de/view.php?id=36634


klees:
Picked to trunk.

PR #5556, opened 2023-02-10
Bugfix for Mantis 25698

sz-tik:
Fixes Bug https://mantis.ilias.de/view.php?id=25698


Amstutz:
@sz-tik Thx a lot for the PR! This would solve the linked issue in the DC. Thx.

@alex40724 : Since the fix is in the Table2GUI, can you greenlight this? If so, I would pick this to 8 and trunk as well. Thx!

Amstutz:
merged it to fix this dc issue. Hope this is fine. did it since it's targeting implecetely mantained legacy UI. @alex40724 , please ping me, if this was not oke for you.

PR #5557, opened 2023-02-13
User: Fix `time_limit_unlimited` handling

mjansenDatabay:
PR: https://mantis.ilias.de/view.php?id=36710

It could be worth to introduce (in a first step of improvements) class constants for the different options, instead of hard coded '0' and '1' values.

If approved, this should be picked to `release_8` as well.


kergomard:
Thank you very much @mjansenDatabay . Merged and cherry-picked.
Yes, I'm looking forward to working on improving this Service.

PR #5558, opened 2023-02-13
UI: Launcher Component

nhaagen:
The Launcher Component is used to launch an object, process or workflow. 
If necessary, the component can be enriched with status information about the user's
progress and optional inputs (e.g. access code field) if launching is restricted.

Screenshot from 2023-02-13 13-15-46

### DIFF
--- a/src/UI/Factory.php
+++ b/src/UI/Factory.php
@@ -907,6 +907,70 @@ interface Factory
      * ---
      * description:
      *   purpose: >
+     *     The Launcher starts an object, a process or a workflow.
+     *     It conveys the smallest set of information needed to decide upon launching.
+     *     The Launcher communicates clearly whether or not a user can or cannot
+     *     launch at any given time.
+     *   composition: >
+     *     For a clear guidance of the users' intent the Launcher can present
+     *     a title and a descriptive text; it will always contain a Button for launching.
+     *     If necessary, the component can be enriched with status information
+     *     about the users' progress such as Icons or Progress Meter as well as
+     *     optional inputs (e.g. access code field) if launching is restricted.
+     *   effect: >
+     *     Clicking the Button starts the object, process or workflow. If the
+     *     Component is configured with inputs, they will be provided to make the user
+     *     fill them before the object can be launched. If the provided data is sufficient
+     *     the user is redirected to the target. Otherwise, a message is being displayed.
+     *     If the user cannot launch the object at all (precondition, unavailability etc.),
+     *     the Button is disabled with unavailable action.
+     *     The label of the Button may change, e.g. in relation to the status
+     *     of the progress.
+     *   rivals:
+     *     Item: >
+     *       Other than an item, the Launcher's focus is on an action rather than
+     *       the representation of an entity.
+     *     Link: >
+     *       Link's primary function is navigation; operating a Link must not
+     *       change the systems's status, while a launcher may well sign up a user
+     *       to a LearningSequence, e.g.
+     * rules:
+     *   usage:
+     *     1: If the user cannot launch the process, the Button MUST be disabled.
+     *     2: >
+     *       The Launcher SHOULD NOT be used to collect larger sets of information (e.g. a
+     *       full user registration) - that would be a process in itself.
+     *     3: >
+     *       The launcher SHOULD support the users' intent to make a quick choice if it is
+     *       the desired object/process/workflow to launch or not. Just relevant information
+     *       SHOULD guide this decision.
+     *     4: >
+     *       There MUST be but one Launcher per process or activity stream.
+     *       However, there MAY be multiple Launchers residing on a view or page
+     *       if there are multiple processes or activity streams on that view or page.
+     *   interaction:
+     *     1: The Launcher MUST start the object/procress when the Button is clicked.
+     *     2: >
+     *      The Launcher MUST provide ample inputs if the object is configured
+     *      with restricted access or the process needs further user decisions.
+     *   accessibility:
+     *     1: >
+     *       All interactions offered by a Launcher MUST be accessible by only using
+     *       the keyboard.
+     *     2: >
+     *       All information required before launching the object MUST be placed
+     *       before the Button launching the object/process, so users working
+     *       with screen readers will not miss it.
+     *
+     * ---
+     * @return \ILIAS\UI\Component\Launcher\Factory
+     */
+    public function launcher(): C\Launcher\Factory;
+
+    /**
+     * ---
+     * description:
+     *   purpose: >
      *     Help Topics can be attached to certain components. They make it possible
      *     that suitable help texts can be displayed alongside the component.
      *
@@ -914,4 +978,5 @@ interface Factory
      * @return \ILIAS\UI\Help\Topic[]
      */
     public function helpTopics(string ...$topic): array;
+
 }
--- a/src/UI/Factory.php
+++ b/src/UI/Factory.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 namespace ILIAS\UI;
 
@@ -978,5 +978,4 @@ interface Factory
      * @return \ILIAS\UI\Help\Topic[]
      */
     public function helpTopics(string ...$topic): array;
-
 }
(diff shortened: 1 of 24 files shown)


Amstutz:
Hi @nhaagen

Thx for revising the initial proposal for the launcher! The puplic interface looks ready for the next discussion on the JF. If accepted, the implementation is already looking good as well. However, we would take a final look after acceptance by the JF. 

Special thx for the examples, screenshots, usage of BEMIT Convention for CSS class names and Unit Tests.

One Question on the public interface:

- [ ] Launcher::ALLOWED_STATUS_COMPONENTS, Why is it on the public interface? We understand that the type is part of the public interface, but this is already communicated via comment. This could also be seen as implementation detail.

Thx
@Amstutz and @thibsy

klees:
Hi @nhaagen,

@thibsy, @Amstutz and me discussed this a little further, and we think that [this](https://github.com/ILIAS-eLearning/ILIAS/pull/5558/files#diff-32781d3085d2a15f827b5999a1d6df42db22e021ecddc8c96b0e2e330cd0fed9R32) can now be moved to [union types](https://php.watch/versions/8.0/union-types) as we only support PHP >= 8.0 starting with current trunk.

Yay!

Could you check that?

Best wishes!

src/UI/Factory.php
```diff
@@ -896,4 +896,67 @@ public function symbol(): Component\Symbol\Factory;
...
+     * ---
+     * description:
+     *   purpose: >
+     *     The Launcher Component is used to launch an object, process or workflow. From a
+     *     users' perspective the main goal of the Launcher component is to present just
+     *     relevant information for a quick decision to launch a certain object, process or
+     *     workflow and if all preconditions for launching are fullfilled. For a clear
+     *     guidance of the users' intent it  can present a title and a descriptive text and
+     *     offers a mandatory Button for launching.
+     *     If necessary, the component can be enriched with status information about the users'
```
atoedt: These sentences describe the composition not the purpose.

src/UI/Factory.php
```diff
@@ -896,4 +896,67 @@ public function symbol(): Component\Symbol\Factory;
...
+     *     1: If the user cannot launch the process, the Button MUST be disabled.
+     *     2: >
+     *       The Launcher SHOULD NOT be used to collect larger sets of information (e.g. a
+     *       full user registration) - that would be a process in itself.
+     *     3: >
+     *       The launcher SHOULD support the users' intent to make a quick choice if it is
+     *       the desired object/process/workflow to launch or not. Just relevant information
+     *       SHOULD guide this decision.
+     *   interaction:
+     *     1: The Launcher MUST start the object/progress when the Button is clicked.
```
atoedt: process or progress

src/UI/Factory.php
klees: @nhaagen Please move that sentence accordingly =)

src/UI/Factory.php
klees: @nhaagen: Please correct that typo.

src/UI/Factory.php
```diff
@@ -896,4 +896,67 @@ public function symbol(): Component\Symbol\Factory;
...
+
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *     The Launcher Component is used to launch an object, process or workflow. From a
+     *     users' perspective the main goal of the Launcher component is to present just
+     *     relevant information for a quick decision to launch a certain object, process or
+     *     workflow and if all preconditions for launching are fullfilled. For a clear
+     *     guidance of the users' intent it  can present a title and a descriptive text and
```
atoedt: The Launcher starts an object, a process or a workflow. It conveys the smallest set of information needed to decide upon launching. The Launcher communicates clearly whether or not a user can or cannot launch at any given time.

src/UI/Factory.php
```diff
@@ -896,4 +896,67 @@ public function symbol(): Component\Symbol\Factory;
...
+     * rules:
+     *   usage:
+     *     1: If the user cannot launch the process, the Button MUST be disabled.
+     *     2: >
+     *       The Launcher SHOULD NOT be used to collect larger sets of information (e.g. a
+     *       full user registration) - that would be a process in itself.
+     *     3: >
+     *       The launcher SHOULD support the users' intent to make a quick choice if it is
+     *       the desired object/process/workflow to launch or not. Just relevant information
+     *       SHOULD guide this decision.
```
atoedt, line 946: There is exactly on Launcher per process or activity stream. However there can be multiple Launchers residing on a view or page if these are multiple processes or activity streams on that view or page.

matthiaskunkel:
Jour Fixe, 20 MAR 2023: We prefer to use a rocket as the glyph for the bulky button. PM will contact our icon designer to suggest a possible glyph.

Amstutz:
Is the implementation ready for a final review? If so, please remove any conflicts and ping us, thx.

nhaagen:
Hi @Amstutz ,
please have a another look ;)
Thank you

Amstutz:
Hi @nhaagen

thx a lot for providing this Pull Request. Nice to see that this was completely handable without additional CSS. Thx for the examples and tests. 

Please Change or indicate, why not:
- [x] examples: please use the HTTP service instead of accessing $_SERVER when building the request url. You can use $DIC->http()->request()->getUri()
- [x] Inline::withRequest(): please put this method on the public interface as well, otherwise ::withInputs() will not work. Ideally, you would also already delegate this call to the Roundtrip. 


Please Change:
- [x] Renderer::renderInline() I: please type-hint the inline-launcher instead of just Component
- [x] Renderer::renderInline() II: please build the Roundtrip with using the third argument "$inputs", instead of using the form as content. This leads to submit-buttons being displayed inside the modals content instead of the footer.
- [x] interfaces: please always describe methods on the public interface if they are not just basic setters or getters. E.g. Launcher::withButtonLabel, what's the boolean for? Or Launcher::withInputs(), how should this method be used, what does the Closure look like?
- [x] call it like it is: Launcher::withStatusMessage -> ::withStatusMessageBox
- [x] Launcher::withInputs() does not hint to what this setter does. maybe rename it to e.g. ::withInputAndEvaluation()
- [x] Renderer::render() does not throw a logic exception
- [x] Unit Tests: Please provide at least one Uni Tests checking if the rendering works. This is important, to make sure, e.g. a11y parts of the rendering survive refactorings or similar. 
- [ ] Testrail Cases: Please add them

Thx
@amstutz and @thibsy

Amstutz:
Hi @nhaagen, 

This is a review only of the public interface, not the implementation.

- [ ] Property Label: Why would I need it, if not visible?
- [ ] Primary Identifier: how can this be a Symbol or Image?
- [ ] Secondary Identifier: how can this be a ShyButton, ShyLink or String, is this not supposed to be a Symbol or Image? Is it really always required?
- [ ] Data Factory Pattern: You use a similar pattern as done in the data table for filling in Data into Entity Listings. I understand why we go this way with the table, since the type of the content is not fully set. However, here I am wondering if this is not overly complicated, here. We could just pass in Arrays of Entities. To me, this is sufficiently typed. Why are you going all the way to the factory? Are you planning on making profit of the lazy evaluation here at some point?

- [ ] Document all methods of the public interface. Note that this is a good place to also extend somewhat on the wording introduced in the methods name/and or the composition section. E.g. what are, why would I set featured properties with the Method „withFeaturedProperties“?
- [ ] Property Composition: Maybe add, that the emphasis is on the value, not the label.
- [ ] Typing 1: Add a type for $value in Property::withProperty. We won't just accept anything as $value, right?
- [ ] Typing 2: There are many missing typehints on the main Entity-interface. We understand that these might be missing components in the UI framework. But please still add typehints: Do we accept strings? Maybe LegacyComponents to signal that this could contain HTML? There would be bonus points to sum up which additional components we would want to have, so we can dig into this after this PR is ready.


Please change:
- [ ] Wording 1: Drop the term "Representation" from "Entity Representation" in your prose, as it never appear in the PHP-classes. "Entity" would be good enough, no?
- [ ] Wording 2: Overhaul wording of Ferdinands doc according to the wording here. Only do this after we agreed on the wording in the JF.
- [ ] Property Listing: CharacteristicValue: Change in Characteristic Value in the rivals section.
- [ ] Property Listing as Rival: Add Properties to the rivals of Descriptive Listing
- [ ] Entity Composition: All possible properties should be named and summarized here. If too extensive, you might also refer to the background which you nicely listed below for further information and OR the public interface. However, it should be clear where to get the full set of all the properties that can be used, along with an explanation of what they are.
- [ ] Please remove the semantics Group png. We should not have binary files in git. The example should do the trick.
- [ ] EntityListing\EntityFactory and withData: What does withData do in the EntityListing, what is the factory for? I understood it only by looking at the example. Please provide a better documentation here. Note that you may refer to the example if needed. The working of the two should be clear by only looking at the public interface (and maybe an example if referenced).

thx a lot
@Amstutz and @klees

thibsy:
Hi @nhaagen,

Thx a lot for implementing the requested changes, they LGTM. I saw you did not yet check the box regarding Testrail Cases, please don't forget to add them. Since this has nothing to do with the component itself I will merge this PR now.

Thanks again for your work!

Kind regards,
@thibsy

PR #5561, opened 2023-02-13
Setup: Implement an Import/Export Mechanism

daniwe4:


klees:
Hi @daniwe4,

please implement this with https://github.com/ifsnop/mysqldump-php instead of mysqldump.

Thanks!

chfsx:
Thank you for this suggestion and also that it has already been switched to the library so that additional exec calls do not have to be introduced.

With the help of this implementation, we can finally integrate large parts of our build process into ILIAS, as until now we had to work with a bash script to generate the database template for new ILIAS versions: https://github.com/ILIAS-eLearning/ilias-build-release/blob/master/ilias-build-release#L60

In order to be able to replace this part of the process, the PR still needs the following extensions:

1. when generating a dump, we need to be able to export certain tables empty, for example. For example, the table il_plugin should not have any content, even if the instance used to create the template has installed some. For this we need a general hook to export such tables empty. some tables we don't want to export completely empty, but for example only individual entries. An example would be the user_data, where we only want to export the entries of root and anonymous, ignoring all others.
2. some data has to be modified when building the template, for example the md5-hash of "homer" has to be exported for root in the usr_data together with the enc_type md5, so that people can log in on a newly installed ILIAS.

@fwolf-ilias  could give more examples if necessary. For the two cases, however, we could certainly use the following hooks of the library:

- https://github.com/ifsnop/mysqldump-php#changing-values-when-exporting
- https://github.com/ifsnop/mysqldump-php#table-specific-export-conditions

However, both work with anonymous functions, which I would like to wrap in an interface or in small DTOs so that the use is clearer and the connection to the library is smaller.

Do you want to work out a proposal for this or do we want to get together to look at this in more detail? especially how consumers could then use the API and we can cover the core-build case.

chfsx (changes requested):
see comment above

daniwe4:
Hello @chfsx , I have tried to implement your above requirements and have recorded this in the setup/README.md. I hope it corresponds to what you imagined.
Greetings Daniel

Services/Database/classes/Setup/MysqlExecDumper.php
```diff
@@ -0,0 +1,59 @@
...
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+class MysqlExecDumper implements MysqlDumper
```
chfsx: The class should be renamed since this doesn't user `exec` anymore, right?

chfsx (changes requested):
@daniwe4 thanks for the updated PR! there is just onbe small change requested now, since the class doesn't use exec anymore it should be renamed to something else, maybe `PHPMysqlDumper`

Services/Database/classes/Setup/MysqlExecDumper.php
daniwe4: Done!

matthiaskunkel:
I set the Jour Fixe tag to this PR to notify us about the added library 'mysqldump-php' that was already committed to trunk in February 2023.

klees:
Description of the change [here](https://docu.ilias.de/goto_docu_wiki_wpage_8047_1357.html).

setup/README.md
```diff
@@ -12,7 +12,8 @@ main commands to manage ILIAS installations:
 
 `install` and `update` also supply switches and options for a granular control of the inclusion of plugins:
 
-* `--skip <plugin name>` will exclude the named plugin from the command
+* `--skip 
+* There are also named objectives for **import** and **export**. <plugin name>` will exclude the named plugin from the command
```
nicoroeser, line 16: Line broken at wrong offset, I think. `--skip` and ` <plugin name>` probably should be kept together.

Services/FileSystem/classes/Setup/class.ilFileSystemDirectoryCopiedRecursivelyObjective.php
```diff
@@ -0,0 +1,181 @@
...
+    protected string $target_folder;
+    protected bool $data_dir;
+    protected bool $bare;
+
+    /**
+     * Copies a directory from ILIAS root or from the outer ILIAS data directory
+     * depending on the flag $data_dir.
+     * With $source_folder you can select the directory to copy from selected
+     * root directory. Empty string for $source_folder means copy the whole root directory.
+     * $target_folder should always be the path wehre to copy into.
```
nicoroeser, line 35: Typo: “wehre”.

matthiaskunkel:
Jour Fixe, 30 OCT 2023: Fabian notified us about the library that has already been merged to trunk. We highly appreciate this suggestion and accept the PR for Trunk.

PR #5562, opened 2023-02-13
grand anonym session only when public section is enabled

iszmais:
https://mantis.ilias.de/view.php?id=36712

This only creates a anonymous sesssion when the public section is enabled since all other section wich an unlgogged user should have acces to are veryfied somewhere else  ( see ilInitialisation::blockedAuthentication())


pascalseeland:
Closing PR as issue cannot be reproduced. See bug report in mantis

PR #5563, opened 2023-02-14
LSO: show toast on LP-completion

nhaagen:
The user is notified by a toast on LP-completion of the LSO.
 
https://docu.ilias.de/goto_docu_wiki_wpage_7723_1357.html


nhaagen:
This is not based on the notification service, I'll amend and close this one.

PR #5564, opened 2023-02-14
LSO: add column 'LPSettings' to content-table

nhaagen:
https://docu.ilias.de/goto_docu_wiki_wpage_7722_1357.html
https://docu.ilias.de/goto.php?target=wiki_1357_JourFixe-2023-03-20#ilPageTocA2516

https://mantis.ilias.de/view.php?id=35704
https://mantis.ilias.de/view.php?id=36859
https://mantis.ilias.de/view.php?id=36015

### DIFF
--- a/Modules/LearningSequence/classes/Content/class.ilObjLearningSequenceContentGUI.php
+++ b/Modules/LearningSequence/classes/Content/class.ilObjLearningSequenceContentGUI.php
@@ -42,6 +42,8 @@ class ilObjLearningSequenceContentGUI
     protected LSItemOnlineStatus $ls_item_online_status;
     protected ArrayBasedRequestWrapper $post_wrapper;
     protected ILIAS\Refinery\Factory $refinery;
+    protected ILIAS\UI\Factory $ui_factory;
+    protected ILIAS\UI\Renderer $ui_renderer;
 
     public function __construct(
         ilObjLearningSequenceGUI $parent_gui,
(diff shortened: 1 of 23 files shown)


kergomard (changes requested):

To be honest: I'm completely unhappy with this class and it got a little worse with your changes, ...but this is not your fault! And right now I won't have the resources to get this to the point where you can make your changes in a more solid way.

So, I have a question/request:

- Do we really need to have [this function](https://github.com/ILIAS-eLearning/ILIAS/pull/5564/files#diff-6539c073d041b3cdae18a3da02ee6f716826ddef66ccd84d0ea96396157e2043R3411) `public`? I don't think we are using it outside the class itself, right? I would thus, if in any shape way or form possible, like to set it to `private`. If there is something preventing us from doing that: Could you please declare it `final` and `deprecate` it right away? We really need to find a better way.

Thanks again and best,
@kergomard

klees:
Thx @kergomard. I agree wholeheartedly.

kergomard (approved):

klees:
I have worked with the commits a little, so now this implements:

* [Learning Sequence: Improve Handling of Learning Progress Settings of Sub-Objects](https://docu.ilias.de/goto_docu_wiki_wpage_7722_1357.html)
* [Learning Sequence: Improve Usability of Objects without Kiosk-Mode (slightly)](https://docu.ilias.de/goto_docu_wiki_wpage_7707_1357.html)

PR #5565, opened 2023-02-14
Chatroom: providing some more defaults for client settings

utesche:
If installing a new client including (!) the settings for the chatserver, a new entry in table [chatroom_admconfig] with an empty (!) value for parameter 'client_settings' gets written.  That's ok so far.

When opening the admin GUI in ILIAS for the first time to configure the chat client, these missing value leads to several "Undefined array key" exceptions while initializing the GUI. This PR provides the necessary checks and defaults to handle this now.


utesche:
I renamed my implicit created branch - thus closing the PR. Whoops!
Don't know who to "rebase a PR", so I open a new PR...

PR #5569, opened 2023-02-15
[FIX] 0034063: EmployeeTalks: Errros when executing ilEmployeeTalkDBUpdateSteps::step_4

chfsx:
see https://mantis.ilias.de/view.php?id=34063

### DIFF
--- a/Modules/OrgUnit/classes/Positions/Operation/class.ilOrgUnitOperationContextDBRepository.php
+++ b/Modules/OrgUnit/classes/Positions/Operation/class.ilOrgUnitOperationContextDBRepository.php
@@ -13,8 +13,8 @@
  * https://www.ilias.de
  * https://github.com/ILIAS-eLearning
  *
- ********************************************************************
- */
+ *********************************************************************/
+
 declare(strict_types=1);
 
 class ilOrgUnitOperationContextDBRepository implements OrgUnitOperationContextRepository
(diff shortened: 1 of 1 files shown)


tfamula:
Hi @chfsx,
@schmitz-ilias provided a PR (#5549) last week regarding the update steps. I'm not sure if this already solves the bug or if we still need your PR. I will ask Tim when he is back at work.

chfsx:
From my point of view this is still valid. Problem ist that the parameter `$context` (see https://github.com/ILIAS-eLearning/ILIAS/pull/5569/files#diff-4f5259f290a229f4a37bef00c3cc89caa3ac381988a5035f3120855a1da2bef7L31)

is overridden by `$context = $this->find($context);` in Line 33, which leads to a fatal in `->withContext($context->getContext())` in Line 48 of ilOrgUnitOperationContextDBRepository.

chfsx:
Tested with a pulled trunk today, updating from e.g. ILIAS 7 or 8 fails

utesche:
I just created a similar PR, see:
https://github.com/ILIAS-eLearning/ILIAS/pull/5570
for trunk not recognising someone else was already working on this problem today...

Surprisingly I didn't got this problem for release_8 but only for trunk...!?
As I can't even select 'trunk' or 'v9.0_alpha' or so in mantis for a new issue, I decided to trace it down and to post a PR.

chfsx:
@utesche your version is much sliker and nicer and does exactly the same thing, I'd prefer your version! yes it occurs only in trunk, not release_8

chfsx:
Close this in favour of #5570

PR #5570, opened 2023-02-15
EmployeeTalk: failing DBUpdateStep 4

utesche:
During installation of a new client step 4 of class.ilEmployeeTalkDBUpdateSteps: 
https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Modules/EmployeeTalk/classes/Setup/class.ilEmployeeTalkDBUpdateSteps.php#L105 
fails on trying to register 'etal' as a new context.

The reason is an attempted access on $context as an object of class ilOrgUnitOperationContext here: 
https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Modules/OrgUnit/classes/Positions/Operation/class.ilOrgUnitOperationContextDBRepository.php#L48
$context is always null if processing reaches this point of code due to be overwrittten at the beginning of the function.

The PR fixes this failure for me, but check the proposed change CAREFULLY before merging!

### DIFF
--- a/Modules/OrgUnit/classes/Positions/Operation/class.ilOrgUnitOperationContextDBRepository.php
+++ b/Modules/OrgUnit/classes/Positions/Operation/class.ilOrgUnitOperationContextDBRepository.php
@@ -30,9 +30,9 @@ class ilOrgUnitOperationContextDBRepository implements OrgUnitOperationContextRe
 
     public function get(string $context, ?string $parent_context): ilOrgUnitOperationContext
     {
-        $context = $this->find($context);
-        if ($context) {
-            return $context;
+        $found_context = $this->find($context);
+        if ($found_context) {
+            return $found_context;
         }
 
         $parent_id = 0;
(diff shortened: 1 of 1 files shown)


utesche:
update:  see also  https://github.com/ILIAS-eLearning/ILIAS/pull/5569

schmitz-ilias:
Hi @utesche, thanks for the PR and the detective work (also to @chfsx as a close second ^^)!
I have provided a PR (#5549) which replaces the faulty `DBUpdateSteps` with objectives in `OrgUnit`. These objectives don't use any other `OrgUnit` classes, so #5549 should fix this new error along with the old one. The PR is not merged yet, I'll let you know once it is.

This PR still looks like it might be a worthwhile change even if it is not immediately needed as a fix, so I'll assign it to @klees as maintainer of `OrgUnit`.

klees:
We have a duplicate in the meantime: #5595

klees:
Thanks @utesche!

PR #5579, opened 2023-02-16
LSO: show toast on LP-completion

nhaagen:
The user is notified by a toast on LP-completion of the LSO.

https://docu.ilias.de/goto_docu_wiki_wpage_7723_1357.html

### DIFF
--- a/Modules/LearningSequence/classes/GlobalScreen/class.ilLSCompletionNotificationProvider.php
+++ b/Modules/LearningSequence/classes/GlobalScreen/class.ilLSCompletionNotificationProvider.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
--- a/Modules/LearningSequence/classes/GlobalScreen/class.ilLSCompletionNotificationProvider.php
+++ b/Modules/LearningSequence/classes/GlobalScreen/class.ilLSCompletionNotificationProvider.php
@@ -1,0 +1,87 @@
+<?php
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+use ILIAS\GlobalScreen\Scope\Notification\Provider\AbstractNotificationProvider;
+use ILIAS\Notifications\Repository\ilNotificationOSDRepository;
+use ILIAS\UI\Component\Symbol\Icon\Standard;
+use ILIAS\Notifications\ilNotificationOSDHandler;
+
+class ilLSCompletionNotificationProvider extends AbstractNotificationProvider
+{
+    public const NOTIFICATION_TYPE = 'lso_completion';
+    public const NOTIFICATION_TIME_PREFERENCE_KEY = 'lso_completion_note_ts';
+
+    public function getNotifications(): array
+    {
+        $current_user = $this->dic['ilUser'];
+
+        if ($current_user->getId() === 0  || $current_user->isAnonymous()) {
+            return [];
+        }
+
+        $osd_notification_handler = new ilNotificationOSDHandler(new ilNotificationOSDRepository($this->dic->database()));
+        $left_interval_timestamp = $current_user->getPref(self::NOTIFICATION_TIME_PREFERENCE_KEY);
+        $notifications = $osd_notification_handler->getOSDNotificationsForUser(
+            $current_user->getId(),
+            false,
+            time() - $left_interval_timestamp,
+            self::NOTIFICATION_TYPE
+        );
+
+        if (count($notifications) === 0) {
+            return [];
+        }
+
+        $icon = $this->dic['ui.factory']->symbol()->icon()->standard(Standard::LSO, 'lso_completion');
+        $item_factory = $this->dic['ui.factory']->item();
+        $lng = $this->dic['lng'];
+        $lng->loadLanguageModule('lso');
+
+
+        $group =  $this->globalScreen()->notifications()->factory()->standardGroup(
+            $this->if->identifier('lso_bucket_group')
+        )
+        ->withTitle($lng->txt('obj_lso'));
+
+        foreach ($notifications as $notification) {
+            $lso_title = $notification->getObject()->title;
+            $title = sprintf($lng->txt('notification_lso_completed_title'), $lso_title);
+            $notification_item = $item_factory->notification($title, $icon);
+
+            $group->addNotification(
+                $this->globalScreen()->notifications()->factory()->standard(
+                    $this->if->identifier('lso_bucket')
+                )
+                ->withNotificationItem($notification_item)
+                ->withNewAmount(count($notifications))
+                ->withClosedCallable(
+                    function () use ($current_user, $osd_notification_handler): void {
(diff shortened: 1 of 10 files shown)


Modules/LearningSequence/classes/GlobalScreen/class.ilLSCompletionNotificationProvider.php
```diff
@@ -0,0 +1,84 @@
...
+use ILIAS\Notifications\ilNotificationOSDHandler;
+
+class ilLSCompletionNotificationProvider extends AbstractNotificationProvider
+{
+    public const NOTIFICATION_TYPE = 'lso_completion';
+    public const NOTIFICATION_TIME_PREFERENCE_KEY = 'lso_completion_note_ts';
+
+    public function getNotifications(): array
+    {
+        $current_user = $this->dic['ilUser'];
```
iszmais, line 33: I would strongly recommend to use the access function instead of acces dic properties by array keys 
(just like `$this->dic->user()`)

Pls apply that for all array access calls on the DIC

Modules/LearningSequence/classes/GlobalScreen/class.ilLSToastProvider.php
```diff
@@ -0,0 +1,67 @@
...
+declare(strict_types=1);
+
+use ILIAS\GlobalScreen\Identification\IdentificationInterface;
+use ILIAS\GlobalScreen\Scope\Toast\Provider\AbstractToastProvider;
+use ILIAS\Notifications\ilNotificationOSDHandler;
+use ILIAS\Notifications\Repository\ilNotificationOSDRepository;
+
+class ilLSToastProvider extends AbstractToastProvider
+{
+    private const NOTIFICATION_TYPE = ilLSCompletionNotificationProvider::NOTIFICATION_TYPE;
```
iszmais, line 28: Even if current Toast providers still have a dependency on Notification the goal is to seperate those 2 completly in future, therefore i would not recommend to build such strong dependency on notification here.

Neverless, since that is still a thing inside the core its more a suggestion here and no mandatory change request.

Modules/LearningSequence/classes/GlobalScreen/class.ilLSToastProvider.php
```diff
@@ -0,0 +1,67 @@
...
+
+        if (count($notifications) === 0) {
+            return $toasts;
+        }
+
+        $factory = $this->globalScreen()->toasts()->factory();
+        $lng = $this->dic['lng'];
+        $lng->loadLanguageModule('lso');
+        $ui_factory = $this->dic['ui.factory'];
+        foreach ($notifications as $note) {
```
iszmais: This loop doesnt seem to depends on its iteration items. I would recommend 1 of 2:
- Add a specification to the toasts to present more information about the current osd.
- Just add a single toast instead of a loop.

iszmais (changes requested):
LGTM in generall and seems to work. Just have some minor style changes and a conceptional change request, cause im not sure of the toast. You may leave that one if this is intentional

Modules/LearningSequence/classes/GlobalScreen/class.ilLSToastProvider.php
nhaagen: Thank you, good point.

Modules/LearningSequence/classes/GlobalScreen/class.ilLSToastProvider.php
nhaagen: OK, that's valid; however, the main goal here was the toast and it certainly depends on this exact notification-type.
I'd rather keept it for now and split it up when there are more notes or different toasts.

Modules/LearningSequence/classes/GlobalScreen/class.ilLSCompletionNotificationProvider.php
nhaagen: Why is this? I usually switch it the other way around, since those functions are mere wrappers and somewhat artificial?

nhaagen:
@iszmais , thank you very much for looking through this, it's my first steps with notifications and toasts ;)

Modules/LearningSequence/classes/GlobalScreen/class.ilLSCompletionNotificationProvider.php
iszmais: Basiacly the use of wrapper is heavly recommended. Not only this provides a general API on wich the crowd of devleopers can base their inner dependencies, it also help in programming since hard wrapper allow autocompletions. The array access is a wrapper itsself with the exception that this is a dynamicly provided wrapper withing the construct of the PHP array. Therefore nothing is safed if you acces it by the array key, we only losing control of the wrapper on this point.

Majorly it for the same reason we dont use `$_GET` anymore but the `$DIC->http()->request()->getQueryParams()` since it gives us controll over the object structure and improves maintainability and extensibility of code in general (Not only ILIAS) due to developer handlet access wich can be regulated and swapped out.

There is also a small paragaprh inside the DI Docu about this
https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/DI/README.md

iszmais:
@nhaagen One addition i was missing myself in the example file:

The toast you provide are valid, but missing the presentation times and delay set by the ILIAS adminsitrator.
Atm a default is choosen for the presetation (5000 ms) and for the delay (500ms).
When you want your Toasts to work "just like the others" you need to set those by calling `withVanishTime()` and `withDelayTime()`. 

You can set them by yourself if you want your toast to behave special or you can just use the new implemented function within the abstract toast provider `getDefaultToast()` wich set them to the adminstration setting.

You can see the now updated example in **GlobalScreen/scope/Toast/REAMDE.md** for this too.

iszmais:
@klees Im kinda confused about the recent progress here (due to the force pushes).
Is there any major change made to the toast section since my last review i should be aware of?

Beside that i would appreciate if one of the responsible could apply the fixes on my last change request (Using dic wrapper instead of array key access).
Thats nothing crucial, but still offer i big benefit to the readability imo.

Greetings,
@iszmais

klees:
Hi @iszmais,

I looked into @nhaagen|s code and found that he indeed uses `getDefaultToast` as you recommended.

Regarding your request to use methods on the wrapper: I simply do not agree with your assertions.

The usage of the methods is neither heavily recommended (by whom?) nor is intended to be "a general API on which the crowd of developers can base their inner dependencies". It only exists for the reason stated [here](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/DI/README.md): "The standard pimple interface is enhanced with explicit methods to get the most common ILIAS services to help IDEs detect the types of services retreived from the container."

Moreover: by using the methods we loose a super nice property of the Pimple container, which is that we can replace it with a simple array. Which is incredibly handy for unit tests.

What in fact is heavily recommended (at least by me and some others =) ) is dependency injection, which we sadly cannot use here due to the restrictions presented by `AbstractNotificationProvider`.

So at [CaT](https://www.cate-tms.de) we use proper DI where possible and avoid the methods on the ILIAS container, mostly.

If we want to have a general rule like "always use methods on the wrapper of the container", we need to talk about this together, weigh pros and cons and agree on something. Until then we all need to trust individual maintainers to make sensible decisions for the code they maintain. Which is always a great cause to argue with each other, of course =)

Best regards!

iszmais:
Hi @klees ,

Thank you for your response.
As a like-minded i really appreciate the concern for the detail :smile: 

Tbh my personal experience (beside the ILIAS docu) on that is mostly based on educational content to PHP. I never saw examples doing otherwise but, after some research, i couldnt really find any qualified opinion on either of both ways in a convention, standard or by a VIP. I will let you know if that changes. Are you aware of any statement to that form a source external to ilias or its maintaining Insititutions?

I think both ways have their preferences wich are hardly subjectivly measured on their importance and heavly depends on the usecases of the individuals.

However i agree on the point that that should be unified in some way to preserve the codes constistency.
My first though was to take that to the jour fixe but im not sure if it isnt to "technical" for that. What do you think?

Greetings,
@iszmais

iszmais:
I moved that "issue" into the mantis -> https://mantis.ilias.de/view.php?id=37461

PR #5580, opened 2023-02-16
PHP 8.1 - some fixes

nhaagen:
with PHP 8.1, I ran into some trouble logging in.
These are first amendments.


klees:
Thx a bunch!

PR #5586, opened 2023-02-17
User: Don't display link to mail system without access

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=35876


kergomard:
Thank you very much @mjansenDatabay ! I think this makes the behavior more consistent and I merge it.

PR #5591, opened 2023-02-20
EmployeeTalks/Staff: Determine users recursively

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=36777

### DIFF
--- a/Modules/EmployeeTalk/classes/TalkSeries/class.ilObjEmployeeTalkSeriesGUI.php
+++ b/Modules/EmployeeTalk/classes/TalkSeries/class.ilObjEmployeeTalkSeriesGUI.php
@@ -104,7 +104,8 @@ final class ilObjEmployeeTalkSeriesGUI extends ilContainerGUI
                     return $access->filterUserIdsForUsersPositionsAndPermission(
                         $userIds,
                         $this->user->getId(),
-                        ''
+                        '',
+                        true
                     );
                 });
                 $this->container->ctrl()->forwardCommand($repo);
--- a/Modules/OrgUnit/classes/Positions/UserAssignment/class.ilOrgUnitUserAssignmentQueries.php
+++ b/Modules/OrgUnit/classes/Positions/UserAssignment/class.ilOrgUnitUserAssignmentQueries.php
@@ -87,12 +87,14 @@ class ilOrgUnitUserAssignmentQueries
 
     /**
      * @param $orgunit_ref_id
-     * @return ilOrgUnitUserAssignment[]
+     * @return list<int>
      */
     public function getUserIdsOfOrgUnit(int $orgunit_ref_id): array
     {
-        return ilOrgUnitUserAssignment::where(['orgu_id' => $orgunit_ref_id])
-                                      ->getArray(null, 'user_id');
+        return array_map(
+            'intval',
+            ilOrgUnitUserAssignment::where(['orgu_id' => $orgunit_ref_id])->getArray(null, 'user_id')
+        );
     }
 
     /**
--- a/Modules/OrgUnit/classes/class.ilOrgUnitPositionAccess.php
+++ b/Modules/OrgUnit/classes/class.ilOrgUnitPositionAccess.php
@@ -27,6 +27,7 @@ class ilOrgUnitPositionAccess implements ilOrgUnitPositionAccessHandler, ilOrgUn
     private \ilOrgUnitGlobalSettings $set;
     private ilAccess $access;
     private ilObjUser $user;
+    private ilTree $repository_tree;
 
     public function __construct(ilAccess $access)
     {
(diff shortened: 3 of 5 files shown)


schmitz-ilias:
Changes in `EmployeeTalk` look good to me, thanks for fixing!

klees (commented):
Hi everyone,

I'm very unhappy how things are moving forward regarding EmployeeTalk/MyStaff/OrgUnits. The testing of EmployeeTalks and MyStaff (fortunately) show various problems and shortcomings of the current implementation and in turn bring up requirements for the Org-Units.

In general, I'm looking to remove code and interface from the Org-Units instead of increasing it. This does the opposite. This change here, specifically, already smells like the next "performance is bad"-ticket landing on my desk through the issue-tracker. Also, institutions asking for this change here, have been reluctant to provide funding for the long-term maintenance of the Org-Units in the past.

So this change here is in a very bad position: It increases the surface and code of my component and likely introduces new issues, while it doesn't save ressources for the long term maintenance of the Org-Units.

Any idea how to deal with this issue? Keep in mind that this isn't a specific problem of this PR but instead a general problem that will likely come up again.

Best regards!

schmitz-ilias:
Hi @klees,

since we are currently (hopefully) in the very last stage of bugfixing for R8, I think it would be most productive to split your concerns in two parts (though they are of course closely related).

First, the changes made in this specific PR: naively looking in from the outside, it seems to me that what @mjansenDatabay did is roughly in line with how similar recursive(-ish) tasks are handled in other `OrgUnit` classes (see e.g. `ilOrgUnitRecursiveUserAssignmentTableGUI::loadData`). Is there anything in specific you'd like to be handled differently, in order to bring this fix more in line with your streamlining efforts? Is there a better way to do the same thing with the existing `OrgUnit` interface? I'd really like to avoid getting stuck on discussions about mid to long term strategy when fixing single tickets.
If there is really no way at all to move forward with this fix, I could look into using some of the methods from `ilMyStaffAccess` to filter users, but that would mean relying even more than we already are on direct access of `OrgUnit` tables from `EmployeeTalk`/`MyStaff`. This seems even more counterproductive to having a clear interface between the components than what is proposed here.

Second, what to do with similar issues moving forward: as discussed yesterday, it is a bit tricky to discuss this in full generality, as we do not yet have a complete overview over what exactly the demands of `EmployeeTalk`/`MyStaff` on `OrgUnit` are. There is really no way around the fact that our components are consumers of yours, so there is going to be a minimum viable surface area for `OrgUnit`. (Note that `OrgUnit` is a bit of a special case here, for other dependencies there is an argument to be had about reversing this relationship, e.g. for `Certificates`, but that is a different discussion.) That being said, from what I have seen so far, what `EmployeeTalk` and `MyStaff` need does not seem to be particularly outlandish or out of scope, and we'd be glad to work out the details together with you in the future.

klees:
Hi @schmitz-ilias,

I know that this situation is neither caused by your nor me, so this might be a "shooting the messenger"-kind of situation. Still, I think you did not quite get my point.

About your first part: I never talked about "streamlining efforts", so it won't be helpful to compare this to other Org-Unit code. In fact, existing code might already show the same problems that I don't want to reintroduce here. Specifically: A lot of roundtrips to the database, introduced by recursion and potentially bad for performance. This is something I want to get rid of, while this would be introducing new cases of the problem.

Then: I don't really get what one tries to do here. The code from `EmployeeTalk` filters some user ids by looking into a permission with the id `""`. What's that even supposed to mean? Why is that same permission never used in the newly introduced code? Does that lazy formulation in [the ticket](https://mantis.ilias.de/view.php?id=36777) even make sense in the context of [our model for permissions via org units](https://docu.ilias.de/goto.php?target=wiki_1357_Introducing_Positions_for_Granting_Permissions_over_Users_via_Organisational_Units)? Whatever that "thing" is that this code tries to accomplish, I'm quite confident that one might achieve it with existing interfaces of the `OrgUnits`. The proposed code actually only uses available interfaces, so no, you would not need to introduce direct access to tables from `OrgUnits`, you could just use the very same code proposed here.

Finally, that discussion is not about a mid term strategy. Its about resource usage. Instead of working on ILIAS 9, in this very moment I'm called upon looking into the fallout of some questionable design and implementation decisions taken by someone else (not you, obviously). The price of accepting this code would be even more usage of my resources in the future.

About your second part: I'm totally happy to talk about requirements from `EmployeeTalk` and `MyStaff` towards `OrgUnits` and move the service forward accordingly. These requirements will look a lot like requirements that other core components and plugins will have. So no problem there.

This is not my issue with this PR as this is not what's happening here. Instead someone tries to dump code in `OrgUnits`, that does a dubious thing and has long term consequences for the service, without providing resources to take care of this, while using questionable arguments of urgency. In fact, I have specifically rejected responsibility for `MyStaff`/`EmployeeTalk`, I'm neither funded nor mandated to look into problems in these services. This does not mean that I won't work with you on the interfaces to these services or even help you with my insights and knowledge, but this certainly does mean that I won't be available immediately and won't be accepting code or changes that are bad for the service that I took responsibility for.

I hope that this is comprehensible.

Best wishes!

schmitz-ilias:
I understand your concerns @klees, thanks for your time.

As it stands I don't really see us getting anywhere with this PR, so I'm closing it. I hope that's okay with everybody.
Thanks for the effort @mjansenDatabay, I'll look into adapting your fix to something internal to `EmployeeTalk`/`MyStaff`.

PR #5592, opened 2023-02-20
SCSS Container, Row, Form Padding Fix (caused horizontal page scrollbar)

catenglaender:
# Issue

## Visible Glitches compared to release_8
Both of these issues are caused by the same change in column layout structure:
* permanent horizontal page scrollbar
* some forms lost their padding and text was glued to the border of their containers

# Underlying Issue
The glitches were caused because the new grid system is less forgiving for deprecated div structures
* the new grid system strictly requires a .col to be inside a .row - this required changes to many legacy form html templates
* rows should be inside a .container (or a similar class) if padding is desired. The padding of a col should not be modified (unless by using the grid system's gap classes)
* a modern, Bootstrap 5 inspired way for constructing containers has been implemented (without the system that creates a space to the page border, as our standardpage is independent from the bootstrap page container)

# Result
in some details trunk should look more like release_8 now
* permanent horizontal scrollbar is gone
* many forms have padding like in release_8
* repository legacy object uses a better structure


Amstutz (approved):
Thx a lot. SCSS changes look good to me. Special thx for the detailed description!

Note, this is clearly targeting an issue in trunk (see title). Do we have a mantis report for this? If not, please ask the user running into the issue to create one. If we do, please link the report.

I will not need a review for every maintainer for every tpl file. Howevever, IMO it would be good if @alex40724 could take a short look into the tpl changes in Form and COPage to avoid layout issues with the Page Editor in Trunk.

Thx.

alex40724:
@Amstutz The changes in COPage are very limited, as maintainer they are fine for me. But I think they are on the same level as the changes of other maintainers components.

On the other hand I am not able to estimate if there are any issues in forms in general with this. The changes in Form simply need testing, as there are lots of forms, subforms, mobile versions of both and so on. I am still "implicit" maintainer of Form, but I would reduce this to the "functional" part only and ask you Timon, to coordinate the css/layout part at least from trunk / 9 on.

Amstutz:
Thx @alex40724 for the feedback!

> I am still "implicit" maintainer of Form, but I would reduce this to the "functional" part only and ask you Timon, to coordinate the css/layout part at least from trunk / 9 on.

That works for me, thx.

catenglaender:
This fixed Mantis Issues 36862 and 36863

PR #5597, opened 2023-02-21
OrgU: Fix for Mantis #36436

tbongers-cat:
Fix for the second part of Mantis #36436, should be viable for trunk as well.


klees:
Picked to trunk.

PR #5599, opened 2023-02-22
Portfolio: fix copying of AdvMD when creating from template (36565)

schmitz-ilias:
This PR fixes a couple of issues with advanced metadata sets not surviving the creation of a Portfolio from a template, including [36565](https://mantis.ilias.de/view.php?id=0036565).


alex40724:
@schmitz-ilias Thanks Tim!

PR #5601, opened 2023-02-22
Refactored notifications and adapted async fetchter onto new collector

iszmais:
This is a following PR to https://github.com/ILIAS-eLearning/ILIAS/pull/5121
wich changes the async toast load to the new implemented GS Collector to centralize the accew point.

Further this does some minor refactoring and renaming to avoid confusion with the "Non-OSD" Notifications and the Notification (without s) Module.


Services/Notifications/templates/default/notifications.js
```diff
@@ -1,17 +1,7 @@
 var OSDNotifier, OSDNotifications = settings => {
-    const createThrottle = timeout => {
-        let clear = () => {};
-        return callback => {
-            clear();
-            clear = clearTimeout.bind(window, setTimeout(callback, timeout));
-        };
-    };
+	let lastRequest = 0;
```
lscharmer: Please move this line into the `poll`  function (After line 54) as it is no where else needed.

Please change this file to use either only spaces or only tabs. Currently they are both in use.

lscharmer (changes requested):

klees:
Hi @iszmais,

may I kindly ask why you'd assigned me here and requested a review? I might miss something, but currently I see no reason why I would need to be involved here. I'm always happy to give reviews if I have enough time and energy, though.

Best wishes!

mjansenDatabay:
@iszmais We only change the DE and EN language files. Please revert all other language file changes.

iszmais:
@klees Since we had some major disccusions during the conception of the toast implementation i just wanted to make you aware of these changes wich slightly modifies the way toast handled now.

You dont have to feel obliged to verifiy this implementation, its more FYI and maybe to set a veto if you see something totaly wrong iyo. :+1:

iszmais:
> @iszmais We only change the DE and EN language files. Please revert all other language file changes.

@mjansenDatabay Does this also count for key changes?
I though that would only affect the translations itsself.

mjansenDatabay:
@iszmais I would say: Never, exept if the system crashes otherwise.

Services/Notifications/classes/class.ilNotificationGUI.php
```diff
@@ -107,18 +107,7 @@ private function getAvailableChannels(array $types = []): array
 
     public function getOSDNotificationsObject(): void
     {
-        $settings = new ilSetting('notifications');
-        ilSession::enableWebAccessWithoutSession(true);
```
mjansenDatabay, line 111: Please don't remove `ilSession::enableWebAccessWithoutSession(true);` here and elsewhere.

I know this is ugly but the only way to prevent the user session rewritten to database and extended in async. calls.

iszmais:
@mjansenDatabay @lscharmer Thx for oyur effort! I did some additional changes to the admin settings nad the abstract Toast provider.
I would appreciate if you could review those too.

Services/Notifications/templates/default/notifications.js
```diff
@@ -1,105 +1,86 @@
...
-			function() {
-				sound.play().then(() => {
-					console.log("Played sound successfully!");
-				}).catch((e) => {
-					console.log("Could not play sound, autoplay policy changes: https://developers.google.com/web/updates/2017/09/autoplay-policy-changes");
-					console.log(e);
-				});
-			}
-		);
+        const attach = new Promise((resolve) => {
```
lscharmer: Please remove the unnecessary promise creation or explain why it is needed.

lscharmer (changes requested):

Services/Notifications/templates/default/notifications.js
iszmais: I Agree on that but i guess more out of a conceptional way.

Initaly this promise was to ensure the sound are loaded before the sound is played. This doesnt affect "normal" setups but is recognizable with a low bandwidth.

Otherwise i assume its no wise, if a system already has a regognizable delay within that loading, to play that sound afterwards since its intention to draw the users attention to a specific time already failed then. Guess its better to miss that one out and hope for the next. 

@mjansenDatabay I would appreciate your opinion on that.

This maybe could be solve by preloading the sound file but i wouldnt recommend to preload medial content just for potential use due to high data use in general (even if thats not the case here).

However, i removed the promise.

lscharmer (approved):

mjansenDatabay (approved):

PR #5604, opened 2023-02-22
T&A Fixes #36289 FileUpload Question

jcopado:
Whoops error: Undefined array key "completion_by_submission"

I was not able to reproduce the bug, but this fix ensure the Undefined array key "completion_by_submission doesn't occur anymore


jcopado:
https://mantis.ilias.de/view.php?id=36289

PR #5607, opened 2023-02-23
fixed types of toast title

iszmais:
During my refactoring of the toast i noticed some typo inside the toast ui interface.
This doesnt affect or was carried further in any other section of the toast therefore its just an documentation issue.

PR #5608, opened 2023-02-23
fixed types of toast title for TRUNK

iszmais:
Sister PR to https://github.com/ILIAS-eLearning/ILIAS/pull/5607


klees:
This should be moved to release_8 as well, right?

PR #5611, opened 2023-02-23
Refinery: Prevent parser tests from failing when Xdebug is enabled (>=PHP8.1)

lscharmer:
The parser tests are adjusted so they can pass successfully with the default value of Xdebug's `max_nestining_level` (which is `256`).

The parser tests (Brick and Primitives) are now using smaller values for the tests so they won't exceed `256`. Previously loops up to 300 where used, which resulted in a call stack between 2000 and 3000 :)

As I don't want to remove the language tag tests they are now marked as _skipped_ when Xdebug is enabled and `max_nesting_level` is below `780`. Additionally they are skipped in the data provider, not in the test itself, so that only one message is shown to the user instead of 12 (If this is not preferred I will adjust this PR).

Interestingly all these tests run fine with PHP8.**0** + Xdebug.

_Tested with PHP8.0, 8.1 and 8.2._

### DIFF
--- a/tests/Data/LanguageTagTest.php
+++ b/tests/Data/LanguageTagTest.php
@@ -32,8 +32,8 @@ class LanguageTagTest extends TestCase
         $this->assertInstanceOf(LanguageTag::class, $tag);
     }
 
-        /**
-     * @dataProvider parseProvider
+    /**
+     * @dataProvider saveToRun
      */
     public function testParse(string $input, bool $isOk): void
     {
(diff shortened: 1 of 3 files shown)


mjansenDatabay:
LGTM (of course, because I noticed this Xdebug issue in PHP 8.2 and now it works based on this branch :-)).

This issue is a confirmation of what we discussed in the workshops last year. Based on such limits this must not be used on large strings or even shorter strings depending on the complexity of the ABNF defnitions.

mjansenDatabay (approved):

klees:
Thanks @lscharmer!

PR #5612, opened 2023-02-24
MainMenu Administration Custom Sub Item Form: `role_based_visibility`

lscharmer:
This PR fixes the following 2 bugs for the role_based_visibility in the sub item form for:

Visibility roles are always empty when editing a sub menu item: Role ID's where compared with the role titles in `array_intersect`.

With >=PHP8.0: Undefined array key `role_based_visibility` when trying to save a sub menu item with an unchecked visibility roles.


chfsx (approved):

chfsx:
And once more, thanks a lot!

PR #5616, opened 2023-02-24
Fix undefined array key in `ilObjectAccess::canBeDelivered`

lscharmer:
This PR adds a check in In the `ilObjectAccess::canBeDelivered` if the regular expression matched the given path, to prevent accessing an undefined array key (`$results[1]`).


kergomard:
Thank you very much @lscharmer!

PR #5255, opened 2022-11-24
Bugfix 25041 required comment in input UI Component

sasfalk:
#bugfix
Mantis issue: 0025041
Target: release_7
https://mantis.ilias.de/view.php?id=25041

Required fields in a form, are marked with an asterisk, but there was no explanation under the form what the asterisk means.
Now the system checks if any of the fields in the form are marked as required. If so, an explanation is displayed. If the individual fields are divided into sections or groups, the individual groups are called first and then the fields within the groups are checked.

PR #5618, opened 2023-02-24
UI, Form: 25041, unit tests

nhaagen:
unit tests for https://github.com/ILIAS-eLearning/ILIAS/pull/5255

### DIFF
--- a/src/UI/Implementation/Component/Input/Container/Form/Form.php
+++ b/src/UI/Implementation/Component/Input/Container/Form/Form.php
@@ -41,7 +41,6 @@ abstract class Form implements C\Input\Container\Form\Form, CI\Input\NameSource
      */
     protected $error = null;
 
-
     /**
      * @param array $inputs
      */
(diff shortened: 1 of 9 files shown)


klees:
@Amstutz or @thibsy, would you please have another look here? Thanks!

thibsy:
Hi @nhaagen

Thx a lot for adding the required testing to this PR. Also thx for your help in resolving the underlying issue.

While reviewing we stumbled upon the following two things:

Please change:
- [ ] Naming: Change `Form::getRequired()` to `Form::hasRequiredInputs()`. This works better on the form itself and is more concise in our opinion.

Question:
- [ ] Example: If possible, could you also provide an example of a form with at least one required field on the form example level. This PR is for ILIAS 7, but an example is not adding any risk to the productive release. Required is always a critical point for a11y conformance, therefore, it would be very useful to have such an example for easy testing (e.g. via testrail)?

Kind regards,
@thibsy and @amstutz

nhaagen:
Screenshot from 2023-03-01 13-38-20

thibsy:
Hi @nhaagen

Thanks for implementing the change and providing an example!

There are still a few tiny things, please change:

- [ ] Example: The UI examples should not contain the ILIAS license header, otherwise the documentation which is displayed in ILIAS will get rather crowded if there are several examples for one component.
- [ ] Comment: Please provide a brief comment above the example-function to explain what the example illustrates.
- [ ] Naming: Since you are adjusting the file anyways, we would also prefer it if the example had a more concise naming as well, like e.g. `form_with_required_input()` instead of just `required()`.

Afterwards this is good to merge.

Kind regards,
@thibsy and @Amstutz

Amstutz:
Thx a  lot!

PR #5621, opened 2023-02-24
GlobalScreen/MetaBar: Prevent accessing uninitialized property

lscharmer:
This PR initializes the `ILIAS\GlobalScreen\Scope\MetaBar\Factory\LinkItem::$symbol` with `null` to prevent an error when calling `LinkItem::hasSymbol` when no Symbol is set.


chfsx:
Thanks a lot!

PR #5623, opened 2023-02-27
[FIX] 0036800: Permissions in question pool is wrong

chfsx:
This fixes a wrong permissions check in the question pool, see Issue in https://mantis.ilias.de/view.php?id=36800


dsstrassner:
Thanks a bunch @chfsx

chfsx:
Please wait with merging, I'll update the PR with the solution for another issues which is mentioned in the issue as well: The Print-View (Tab) should be visible as well with read-access.

chfsx:
I updated the PR to show the missing tabs as well

PR #5628, opened 2023-02-28
Removing LUI: Allow UI Buttons in ilToolbarGUI

tfamula:
This is IMO a needed change to allow using UI Buttons in `ilToolbarGUI`.

### DIFF
--- a/Modules/Test/classes/toolbars/class.ilTestInfoScreenToolbarGUI.php
+++ b/Modules/Test/classes/toolbars/class.ilTestInfoScreenToolbarGUI.php
@@ -180,7 +180,7 @@ class ilTestInfoScreenToolbarGUI extends ilToolbarGUI
         }
     }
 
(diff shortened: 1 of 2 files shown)


alex40724:
@tfamula Thanks for the PR. There is an existing addComponent() function that should allow to add KS buttons imo. Please check.

mjansenDatabay:
Alex is right, the existing method is already widely used. The bigger problem is, that there is IMO no (non-hacky) way to replace instances of `ilSubmitButton` (means: a HTML button which is natively able to submit the toolbar form for state manipulating POST requests) in legacy toolbars.

mbecker-databay:
I am good with this PR, thank you!

tfamula:
Thanks for the hint, @alex40724! `addComponent()` already does the job. My bad!

tfamula:
@mjansenDatabay Maybe you should put the mentioned problem on the agenda for the next "Removing LUI"-meeting!

PR #5629, opened 2023-03-01
Fix #36866: getObjectsForOwner() needs to be static

Rillke:
As `getObjectsForOwner()` is in a factory, I assume that it should be a static method. That's also how it's been invoked.

There are only static calls to this method in the entire code base:

```
ajcay@dev:/ilias$ grep -r "getObjectsForOwner"
Services/Object/classes/class.ilObjectFactory.php:    public function getObjectsForOwner($object_type, $owner_id)
webservice/soap/classes/class.ilSoapGroupAdministration.php:            $owned_objects = ilObjectFactory::getObjectsForOwner("grp", $user_id);
webservice/soap/classes/class.ilSoapCourseAdministration.php:            $owned_objects = ilObjectFactory::getObjectsForOwner("crs", $user_id);
```

Fixes https://mantis.ilias.de/view.php?id=36866

### DIFF
--- a/Services/Object/classes/class.ilObjectFactory.php
+++ b/Services/Object/classes/class.ilObjectFactory.php
@@ -39,7 +39,7 @@ class ilObjectFactory
      * @param unknown_type $owner_id
      * @return unknown
      */
-    public function getObjectsForOwner($object_type, $owner_id)
+    public static function getObjectsForOwner($object_type, $owner_id)
     {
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @Rillke 
I don't think you are right with your argument, but concerning the current situation: You are right. I merge and cherry-pick.

PR #5631, opened 2023-03-01
Database: Show `InnoDb` migration errors

mjansenDatabay:
While executing the migrations on a local ILIAS installation I received an error while migration table `personal_pc_clipboard`. The actual problem was not shown, so I suggest to print the exception message with this PR.

The actual issue was:

> An undefined Database Exception occured. SQLSTATE[42000]: Syntax error or access      
         violation: 1067 Invalid default value for 'insert_time' 

If approved, this should be cherry-picked to `release_8` as  well.

### DIFF
--- a/Services/Database/classes/Setup/class.ilMysqlMyIsamToInnoDbMigration.php
+++ b/Services/Database/classes/Setup/class.ilMysqlMyIsamToInnoDbMigration.php
@@ -79,11 +79,16 @@ class ilMysqlMyIsamToInnoDbMigration implements Migration
     {
         $rows = $this->getNonInnoDBTables();
         $table_name = array_pop($rows);
-        if (is_string($table_name) && strlen($table_name) > 0) {
-            $migration = $this->database->migrateTableToEngine($table_name);
-        }
-        if (isset($migration) && $migration === false) {
-            throw new ilException("The migration of the following tables did throw errors, please resolve the problem before you continue: \n" . $table_name);
+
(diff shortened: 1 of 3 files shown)


Services/Database/classes/Setup/class.ilMysqlMyIsamToInnoDbMigration.php
```diff
@@ -79,11 +79,16 @@ public function step(Environment $environment): void
...
-            $migration = $this->database->migrateTableToEngine($table_name);
-        }
-        if (isset($migration) && $migration === false) {
-            throw new ilException("The migration of the following tables did throw errors, please resolve the problem before you continue: \n" . $table_name);
+
+        if (is_string($table_name) && $table_name !== '') {
+            try {
+                $this->database->migrateTableToEngine($table_name);
+            } catch (\ilDatabaseException $e) {
+                throw new ilException(
```
chfsx, line 87: @mjansenDatabay Should we use a `UnachievableException` exception here? I can change this while merging it

mjansenDatabay: Good point @chfsx , a `UnachievableException` would be indeed more appropriate.

chfsx:
@mjansenDatabay I cherry-picked the change and used the UnachievableException instead of the ilException there: https://github.com/ILIAS-eLearning/ILIAS/commit/abff76e9715f7e2752a312408860ebf9402fb0e0

PR #5633, opened 2023-03-02
Services/Component: Ignore dot files and non-directories in plugin scan

tbongers-cat:
This is a follow-up for #5548 and implements the change as suggested by @klees in the comments.

I also tried solving this via `src/Filesystem/Finder`, as that would ignore dot files by default and make scanning for only directories a bit easier. But as this objective is achieved when building artifacts (e.g. via `composer du`), the global $DIC is not initialised and therefore the Filesystem is not available at `$DIC->filesystem()`.

I considered doing the initialisation of the Filesystem manually in the Objective, but after going through its setup in `ilInitialisation::initIlias()`, I found it depends on a Factory which then depends on a FilenameSanitizer and that depends on reading the ILIAS settings ... so I gave up :)

If anybody has an idea on how to make accessing `src/Filesystem` inside the Objective easier, I would like to hear your input on this.


klees:
Picked to trunk. Thanks for investigating.

chfsx:
@tbongers-cat thanks for investigating, I'll have a look at the initialisation of the FileSystem-Service to make that easier from the outside

PR #5636, opened 2023-03-02
T&A #32195: Specific Feedback for Each Answer provided: Not necessa…

fneumann:
…ry "Feedback" Headline visible

Proposal:
Instead of hiding the headline (which would result in no visible action after clicking 'check'), a message is shown: 'There is no feedback available for your answer.' The message is always shown if the content under the headline would be empty.

Mantis: https://mantis.ilias.de/view.php?id=32195

### DIFF
--- a/lang/ilias_de.lang
+++ b/lang/ilias_de.lang
@@ -1439,7 +1439,7 @@ assessment#:#tst_failed#:#Durchgefallen
 assessment#:#tst_failed_imp_qst_skl_assign#:#Die Zuweisungen der Fragen zu folgenden Kompetenzen konnte durch den Import nicht hergestellt werden. Die entsprechenden Kompetenzen konnten im hiesigen System nicht identifiziert werden.
 assessment#:#tst_failed_imp_skl_thresholds#:#Der Import der Schwellenwerte für folgende Kompetenzen wurde übersprungen, weil die zugehörige Kompetenz im lokalen System eine abweichende Liste von Ausprägungen aufweist.
 assessment#:#tst_feedback#:#Rückmeldung
-assessment#:#tst_feedback_is_given_inline#:#Die Rückmeldung wird bei Ihrer Antwort angezeigt.
+assessment#:#tst_feedback_is_given_inline#:#Die Rückmeldung wird direkt unter Ihrer Antwort angezeigt.
 assessment#:#tst_feedback_not_available_for_answer#:#Für Ihre Antwort ist keine Rückmeldung verfügbar.
 assessment#:#tst_filter_lifecycle_enabled#:#Filter „Lebenszyklus“ verwenden
 assessment#:#tst_filter_question_type#:#Fragentyp
--- a/lang/ilias_de.lang
+++ b/lang/ilias_de.lang
@@ -1440,6 +1440,7 @@ assessment#:#tst_failed_imp_qst_skl_assign#:#Die Zuweisungen der Fragen zu folge
 assessment#:#tst_failed_imp_skl_thresholds#:#Der Import der Schwellenwerte für folgende Kompetenzen wurde übersprungen, weil die zugehörige Kompetenz im lokalen System eine abweichende Liste von Ausprägungen aufweist.
 assessment#:#tst_feedback#:#Rückmeldung
 assessment#:#tst_feedback_not_available_for_answer#:#Für Ihre Antwort ist keine Rückmeldung verfügbar.
+assessment#:#tst_feedback_is_given_inline#:#Die Rückmeldung wird bei Ihrer Antwort angezeigt.
 assessment#:#tst_filter_lifecycle_enabled#:#Filter „Lebenszyklus“ verwenden
 assessment#:#tst_filter_question_type#:#Fragentyp
 assessment#:#tst_filter_question_type_enabled#:#Fragentyp Filter verwenden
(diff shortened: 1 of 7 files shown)


dsstrassner:
Hi @yvseiler, 

would you be so kind, to review this PR, if this solves your issues?

Hi @matthiaskunkel, 

are the language changes ok for you?

matthiaskunkel:
Looks fine for me. Thanks!

yvseiler (approved):
Hi @fneumann and @dsstrassner 

Thank you very much for taking care of the issue and working out a solution. From a UI perspective, the comment below the title already helps more than without. Thanks for the suggestion. 

Please consider:
The statement "'There is no feedback available for your answer.'" could have the effect of ignoring the instant feedback (currently not strongly highlighted). Perhaps a separate improvement would be to make the instant feedback more visually striking to avoid this.

best regards,
@yvseiler

dsstrassner:
Hi @mbecker-databay and @kergomard  please wait with the merge.

kergomard:
Hi @fneumann 
Thank you very much for the fix and the PR!
On a conceptual level:
- If I'm getting it right, I would second @yvseiler concern: The feedback can actually be subtly wrong, when there is inline feedback. Shouldn't we take care of this situation?

On a more technical level one quick suggestion from my side:
- Couldn't we make [this](https://github.com/ILIAS-eLearning/ILIAS/pull/5636/files#diff-d269222b3b890956821abf200299d69ae0fc5f52e18be490208e84802331fffdL2324) more readable with early returns and without the nested conditions?
Please change:
- There seem to be Coding Style violation [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5636/files#diff-d269222b3b890956821abf200299d69ae0fc5f52e18be490208e84802331fffdR2347) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5636/files#diff-60422374998db83bbbd85d19a82fc0479c3cc1f3ed210d55b8001f59a0d3805bR351) additionally the two violations are subtly different.

Again, thank you very much and best,
Stephan

dsstrassner:
Hi @fneumann, 

thanks for your approach.
I have checked this now locally and I agree with @yvseiler and @kergomard, that this is the wrong approach: 

2023-03-07-p5vlt_Screenshot-

If the setting "Specific Feedback for Each Answer provided" is choosen and the there is answer specific feedback in the question, we should not show the info below the answer options. In cases, where really no feedback is given, the info is okay: 

2023-03-07-o90j0_Screenshot-

I change the title to WIP:

fneumann:
Hi @dsstrassner, @yvseiler, @kergomard 
Thanks for your comments and reviews. I would not say it'ts the wrong approach, because I think it is important to give a visual reply if "Check" is clicked, even if there is no feedbach available. 

But we surely have aproblem with three question types where feedback is given inline (SC, MC, KPrim). I reworked the PR to handle this case. The feedback header is still shown to keep things consistent, but for these three types the message now says 'The feedback is given in your answer', if no further feedbach is shown below the headline.

grafik

@kergomard 
I found it hard to work with early returns because there are different types of feedback that may be shown below the feedback header, depending on the settings of the question and the test. To the best chance to respect them all was to work with a flag that is set if any of these produces a feedback output and to to decide at the end if a message has to be shown.

kergomard:
Thank you very much @fneumann ! Now I think this is ok.
@yvseiler : Do you agree with the solution?
@matthiaskunkel : Language-Changes ok with you? Is Chris also on Github (couldn't find him)? Would he quickly check the english versions, before we introduce them?

@fneumann : Remove the WIP-flag as soon as you feel a final review would be due.

Thanks again and best,
@kergomard

yvseiler:
Thanks a lot @fneumann for the new suggestion, from my point of view it fits very well this way and even helps to recognize the inline feedbacks better.

lang/ilias_en.lang
```diff
@@ -1440,6 +1440,8 @@ assessment#:#tst_failed#:#Failed
 assessment#:#tst_failed_imp_qst_skl_assign#:#The question's assignments to the following competences could not be created. The according competences could not be identified within the local system.
 assessment#:#tst_failed_imp_skl_thresholds#:#The Import of thresholds for the following competences were skipped, because the according competences within the local system are configured with a different list of levels.
 assessment#:#tst_feedback#:#Feedback
+assessment#:#tst_feedback_is_given_inline#:#The feedback is given in your answer.
```
matthiaskunkel, line 1443: Better: "The feedback will be displayed along with your answer." Suggested by Chris Potter ;-)

lang/ilias_de.lang
```diff
@@ -1439,6 +1439,8 @@ assessment#:#tst_failed#:#Durchgefallen
 assessment#:#tst_failed_imp_qst_skl_assign#:#Die Zuweisungen der Fragen zu folgenden Kompetenzen konnte durch den Import nicht hergestellt werden. Die entsprechenden Kompetenzen konnten im hiesigen System nicht identifiziert werden.
 assessment#:#tst_failed_imp_skl_thresholds#:#Der Import der Schwellenwerte für folgende Kompetenzen wurde übersprungen, weil die zugehörige Kompetenz im lokalen System eine abweichende Liste von Ausprägungen aufweist.
 assessment#:#tst_feedback#:#Rückmeldung
+assessment#:#tst_feedback_is_given_inline#:#Die Rückmeldung wird bei Ihrer Antwort angezeigt.
```
matthiaskunkel, line 1442: How about "Die Rückmeldung wird direkt unter Ihrer Antwort angezeigt."? Makes it even clearer. But current version is fine for me, too.

matthiaskunkel (changes requested):
Language files reviewed. At least the English version should be changed.

lang/ilias_en.lang
fneumann: Changed the text as suggested.

lang/ilias_de.lang
fneumann: Changed the text as suggested

kergomard:
Thank you very much @fneumann !
This looks good to me now and I merge and cherry-pick to trunk ilias8 and trunk.

PR #5637, opened 2023-03-02
#29983 Adjust choice of best pass in test results export

fneumann:
See https://mantis.ilias.de/view.php?id=29983#c82580 
Now the first of equal passes wins


dsstrassner:
If I remember correctly, this PR traces the GUI behavior into the export? Like we discussed in TechSquad a few weeks ago?

fneumann:
Yes, it now chooses the same best pass that is also chosen in the GUI, if passes have equal score.

PR #5638, opened 2023-03-02
Fix permission check for MyStaff as possible starting point

tfamula:
Please also cherry-pick to release_8 and release_7 or ping me if I should do it.

### DIFF
--- a/Services/User/classes/class.ilUserUtil.php
+++ b/Services/User/classes/class.ilUserUtil.php
@@ -244,7 +244,7 @@ class ilUserUtil
             $all[self::START_PD_SUBSCRIPTION] = 'my_courses_groups';
         }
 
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you @tfamula for the PR.
Could you please explain to me how this improves the situation? For me this rather looks like a step back:
- The method name seams hard to understand to me.
- If I'm getting this right, we now don't check for activation of MyStaff anymore (I don't even know if this is a thing), but it feels like something that should be done.

Don't hesitate to contact me on discord.

tfamula:
As discussed with @kergomard on discord (and also with @schmitz-ilias), the current behaviour is correct. When selecting Staff as starting point, there should still be a check for the general right to Staff. In doubt, there will be an empty Staff List if an user has no employees (yet).

PR #5639, opened 2023-03-02
T&A Bugfix #36544, #36771, #36798 Undefined array key IlObjQuestionPoolGUI

jcopado:
Ensures Post entry is set, else return string


kergomard:
Thank you very much @jcopado ! Merged and cherry-picked to trunk.

PR #5640, opened 2023-03-02
T&A Bugfix #34296 TypeError Multiple Choice

jcopado:
Hardening the code, not able to reproduce


kergomard:
Thank you very much @jcopado for this fix. I don't think this is the right one though and will quickly apply the solution from the Single Choice Question.
Thanks again and best,
@kergomard

PR #5641, opened 2023-03-03
Added imagick and zip to preconditions

chfsx:
With https://github.com/ILIAS-eLearning/ILIAS/pull/5172 we decided to have php-imagick and php-zip as dependecies to install/update ILIAS 9. This PR adds those extensions as dependency to the setup.

### DIFF
--- a/Services/FileServices/classes/Setup/class.ilFileServicesSetupAgent.php
+++ b/Services/FileServices/classes/Setup/class.ilFileServicesSetupAgent.php
@@ -1,0 +1,47 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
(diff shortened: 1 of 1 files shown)


klees (changes requested):
Hi @chfsx,

could we move these to the components that actually require the extensions? Currently, these other extensions are just listed in the central setup file because I didn't bother to backtrace which component actually uses what. In general its better to have the dependencies declared where they are required IMHO. If components get removed or dependencies are not required any more for some other reasons, they simply can be removed locally. Also, you would  not need to pass me this code for review... =)

Thanks!

chfsx:
@klees yes, even better :-) I thought that all PHP packages would have to go in there, because all the previous ones are there, but I like to take them directly into the components! I will merge directly after the conversion

PR #5643, opened 2023-03-03
[FIX] DataSize can lead to segmentation fault in PHP 8.1.16

chfsx:
I will create another PT for ILIAS 8 as well (without the `map` syntax).
I'm pretty sure the segemntation fault comes from the huge array-indices (such as `1208925819614629174706176`) in `$suffix_map`, therefor I changed this to a `map`.


chfsx:
Sister-PR in https://github.com/ILIAS-eLearning/ILIAS/pull/5644

PR #5647, opened 2023-03-03
possible fix for bug #36884

c-knof:


webservice/soap/lib/nusoap.php
```diff
@@ -2638,7 +2638,7 @@ public function send($data, $timeout=0, $response_timeout=30, $cookies=null)
     * @access   public
     * @deprecated
     */
-    public function sendHTTPS($data, $timeout=0, $response_timeout=30, $cookies)
+    public function sendHTTPS($data, $timeout=0, $response_timeout=30, $cookies=NULL)
```
chfsx: I'm surprised that no typing was introduced here at all, that should have been done as part of the PHP 8 project. @smeyer-ilias did we forget something?

chfsx (changes requested):
CI/Copyright-Checker/copyright-checker.sh is currently failing.

@jeph864  may you want to have a look at this PR? From my point of view, this file hasn't been refactored in the PHP8 project, introducing types for parameters and return values was mandatory. I think this file was forgotten. do you want to adjust it? or maybe @smeyer-ilias  has information on why it could not be refactored.

webservice/soap/lib/nusoap.php
chfsx: @c-knof `NULL` must be `null` according to the code style, you can fix that with the cs-fixer: ```libs/composer/vendor/bin/php-cs-fixer fix --config=CI/PHP-CS-Fixer/code-format.php_cs --using-cache=no  webservice/soap/lib/nusoap.php ```

chfsx:
Thanks @c-knof for the PR, this message has been annoying me all the time :-) With the nusoap thing, I'm not sure why it's still needed, if you search the feature wiki for it, you'll find a few places that talk about removing it. but maybe I'm wrong :-)

webservice/soap/lib/nusoap.php
c-knof: @chfsx code style is fixed. Can I just insert the copyright section into this file? I was very unsure about it, so I didn't do it

chfsx:
That's what I was wondering, because if I see it correctly, the file is "foreign code". Therefore, it would probably be wrong to mark this code as "ours". But @smeyer-ilias could certainly say more about that.

In the meantime I was able to clarify further, as far as I understand it @smeyer-ilias  is still responsible for the previous releases, @jeph864  then for trunk/ILIAS 9.

smeyer-ilias:
Thanks @c-knof  for fixing this. 
@chfsx yes, this is "foreign code" and therefore no changes regarding the introduction of parameter and return types have been done.
nusoap.php is currently used only for the WSDL generation. "Normal" API calls do not use this dependency. There were ideas to remove this dependency and deliver a pregenerated WSDL with the ILIAS source code. But this is not easy to implement since the introduction of the SOAP-Plugin slot.

smeyer-ilias (approved):

smeyer-ilias:
Since it is foreign code with an own license header, I suggest to keep this license information. Ok, @chfsx ?

chfsx:
Thank you for your review @smeyer-ilias, yes I think in this case it is also ok to leave the licence header as it is.

PR #5648, opened 2023-03-03
Services/Badge: Apply PHP Rector 8.1 transformations

lscharmer:
This PR adds PHP8.1 features like `final const` construct properties and `readonly` for `Services/Badge`.


mjansenDatabay:
Thx a lot @lscharmer

PR #5649, opened 2023-03-03
Migrate DCL files from File-Objects to direct Access

chfsx:
This PR detaches the user uploads in the datacollection from the ilObjFile and uses the IRSS directly. Due to the migration to ILIAS 7, these files are already in the storage service, but were previously accessed via ilObjFile.

The PR again provides a migration. As long as this has not been executed, the files in DataCollections cannot be used (the old field type FileUpload has been modified so that it only displays information about the missing migration). A new field type "File" was added, which saves new files directly in the storage service without diversions via ilObjFile. 

If you have any questions, please contact me at any time. We can also discuss this via Discord, for example.

https://docu.ilias.de/goto_docu_wiki_wpage_6727_1357.html

### DIFF
--- a/Modules/DataCollection/classes/Content/class.ilDclRecordListGUI.php
+++ b/Modules/DataCollection/classes/Content/class.ilDclRecordListGUI.php
@@ -13,8 +13,7 @@
  * https://www.ilias.de
  * https://github.com/ILIAS-eLearning
  *
- ********************************************************************
- */
+ *********************************************************************/
 
 /**
  * @author  Martin Studer <ms@studer-raimann.ch>
@@ -37,6 +36,7 @@ class ilDclRecordListGUI
     public const CMD_CANCEL_DELETE = 'cancelDelete';
     public const CMD_DELETE_RECORDS = 'deleteRecords';
     public const CMD_SHOW_IMPORT_EXCEL = 'showImportExcel';
+
     /**
      * Stores current mode active
      */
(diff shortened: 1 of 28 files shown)


Amstutz:
Hi @chfsx 

Thx a lot for this large contriution. Highly appreciated. I somewhat dislike, that we still have to support the state before the migration is complete from a dev perspective. We would be able to get rid of more code otherwise. However, this is understandable and makes from a sysadmin point of view. Anyways not much you could change here.

Thx a lot for the migration.

There is one minor thing to change:
- [x] Inline HTML: Please do not use it, use the UI Component Link instead here: https://github.com/ILIAS-eLearning/ILIAS/pull/5649/files#diff-4307692b82f29e565d50ccb52772c20c4a3dc42fa8527540b3d597404627fd33R59

Feel free to directly merge this if the above is completed, so this will not run into merge conflicts with bugfixing soon again taking place.

Thx a lot
@Amstutz

chfsx:
Hi @Amstutz and thanks for the feedback, I changes to download link to a UI\Link\Standard.

I rebased on trunk and since there were conflicts in ilDataCollectionDBUpdateSteps9 (which was called ilDataCollectionDBUpdateSteps8 before) you should have a look again. maybe we should split up this class to have steps from 8 separated from 9? since this could lead to some steps performed more than once

chfsx:
@Amstutz seems to be OK concerning the ilDataCollectionDBUpdateSteps9 from my point of view

Amstutz:
thx for changing the to the UI Components Link. Also thx for resolving the conflicts around the DB Update Step. Looked through the class. Doing it twice seems better for most steps, since we are not sure, if we did not have any cases, writting a null into the DB after executing the first occurance (probably only valid, for test8 plattform, running since early beta, but it does not hurt much imo).

PR #5650, opened 2023-03-03
[FIX] 0036499: Failed test: Abbruch des Hochladens

chfsx:
This is a relatively small fix, unfortunately you can't cancel the upload when uploading large files. So that you don't have the impression that you can cancel the upload with a click on the X, it is hidden by the PR.

see https://mantis.ilias.de/view.php?id=36499


Amstutz:
Thx a lot!

PR #5656, opened 2023-03-06
UI: Data Table

nhaagen:
Screenshot from 2023-03-06 11-43-10

The Data Table is meant to list widely uniform records and work on them (as a whole, not modifying singular aspects/fields of a record).

There are, from the consumers point of view as well as for the user, some benefits I'd like to point out explicetly:

First, Data Table defines the nature of columns, so that formatting (both the appearance/CSS as well as the data-part)
is more easily to accomplish.

Second, it comes with actions that apply to single records, multiple records or both - however, the action
itself will not differ.
Screenshot from 2023-03-06 11-44-02 Screenshot from 2023-03-06 11-44-10 Screenshot from 2023-03-06 11-43-28

Third, the data-retrieval is an integral part of the table, and we tried to make it as easy as possible to 
map repository-data to a row by providing necessary parameter as order, range, etc by default.
Screenshot from 2023-03-06 11-45-29

For Users, some highlighting and colum-resizing was implemented. Also, the a11y was improved by things like
column-count and -index, aria labels indicating roles/groups and sortation.
When working with screenreaders, keybord navigation with arrow-keys comes in handy.

Of course, the Table should work (and come with) ViewControls: Pagination, Sortation, Columnselection and Filters.
Those are still missing and are worth PRs of themselves - this one is big enough, anyways.
Also, the set of available column-types is certainly not complete.

Find Description/Rules here: 
[for the Table](https://github.com/nhaagen/ILIAS/blob/_9/UI/DataTable/src/UI/Component/Table/Factory.php#L113-L190)
[for Columns](https://github.com/nhaagen/ILIAS/blob/_9/UI/DataTable/src/UI/Component/Table/Factory.php#L193-L228)
[(individual Columns)](https://github.com/nhaagen/ILIAS/blob/_9/UI/DataTable/src/UI/Component/Table/Column/Factory.php)
[and for Actions](https://github.com/nhaagen/ILIAS/blob/_9/UI/DataTable/src/UI/Component/Table/Factory.php#L230-L261)

I added some examples, and when dealing with this PR, it might help to checkout the branch and look at them ;)

This PR is a successor of (https://github.com/ILIAS-eLearning/ILIAS/pull/2724)

Happily awaiting comments and hints,
Nils and Ferdinand


src/UI/Implementation/Component/Table/Data.php
```diff
@@ -0,0 +1,299 @@
...
+        $this->setEnumeratedColumns($columns);
+        $this->initializeVisibleColumns();
+        $this->initializeViewDefaults();
+    }
+
+    protected function setEnumeratedColumns(array $columns): void
+    {
+        $counter = 0;
+        foreach ($columns as $id => $column) {
+            assert(is_a($column, T\Column\Column::class), \InvalidArgumentException($id . ' is not a column.'));
```
chfsx: This must be `new \InvalidArgumentException` instead of just `\InvalidArgumentException`

nhaagen: Thank you!

catenglaender:
I would like to highlight a couple (I think) exciting and useful details. :slightly_smiling_face: 

# Sticky elements

The header row and selection column stick to the edge when they reach the outer border of the scrolling page.

data-table_sticky

# Header Row cut off

After about 3 lines, long header texts are cut off, so they cannot cause odd one-syllable wide or infinitely tall header cells.

Data-Table_long-header-cell-text

# CSS Resize

While we implemented reasonable default sizes for the different column types, columns can also be resized. The current simple but functional implementation is pure CSS and doesn't use JS.

data-table_resize2

# Keyboard Navigation

Navigating with the keyboard arrows through the table looks like this:

data-table_keyboard

src/UI/Implementation/Component/Table/Data.php
```diff
@@ -0,0 +1,304 @@
...
+    {
+        $this->range = $this->data_factory->range(0, $this->number_of_rows);
+
+        $sortable_visible_cols = array_filter(
+            $this->getVisibleColumns(),
+            fn ($c) => $c->isSortable()
+        );
+
+        $order_by = current(array_keys($sortable_visible_cols));
+        $this->order = $this->data_factory->order($order_by, \ILIAS\Data\Order::ASC);
```
chfsx: This currently fails in when
a) you build a table without any sortable column (`Argument #1 ($subject) must be of type string, bool given`)
b) your array of columns doesn't have strings as array indices (`Argument #1 ($subject) must be of type string, int given`)
I think both should work :-)

src/UI/Implementation/Component/Table/Data.php
```diff
@@ -0,0 +1,304 @@
...
+use ILIAS\Data\Factory as DataFactory;
+use ILIAS\Data\Order;
+use ILIAS\Data\Range;
+
+class Data extends Table implements T\Data, JSBindable
+{
+    use JavaScriptBindable;
+
+    protected string $title;
+    protected DataRetrieval $data_retrieval;
```
chfsx: This must be `T\DataRetrieval` instead of `DataRetrieval`

chfsx (changes requested):
Hi @nhaagen I have made my first attempts with the new table and already love it :-) I noticed the first small things that could be fixed, then @thibsy and I would do another full review. thank you very much!

would it also be possible to rebase the branch on the current trunk? because the status of the branch is quite old and it has some errors (which have nothing to do with the table) which are annoying when using it.

thibsy (changes requested):
Hi @nhaagen and @catenglaender

First of all, thank you for implementing such a tremendously helpful component with lots of fun features!

This review will only cover the interface-changes, the technical review will follow later on.

Before diving into the review, we wanted to point out that this PR introduces a new concept of providing `Data\URI` objects instead of strings, which makes this a somewhat special component and something to think about. `ilCtrl` currently doesn't provide valid strings for `URI`s, so when using this component consumers must have in mind that targets will need to be prepended by the `ILIAS_HTTP_PATH` beforehand. There will certainly be a bridge to `ilCtrl` in the future, which might bring along another object of this sorts.

We also wanted to point out, that in the past consumers could easily delegate the sortation, pagination and filtering process to the `ilTable2GUI`, which cannot be done with the new data-tables. This is not something bad though, because the `ilTable2GUI`s implementation of these features is very inefficient. We believe, that there could be some abstractions in the future, for e.g. `PaginatedDataRetrieval`, since such sources will definitely share some common functionalities (not necessarily provided by UI).

- [ ] `Column\TimeSpan`: From our understanding, this component should always be used in a pair to represent a from-to timespan across exactly two columns. Is there a concept of how this will be checked when building the table? Would such a column maybe benefit from something like a `MultiColumn` component, which strictly expects multiple columns of a certain type, according to the concrete implementation (two `Date` columns for this instance)? In a table-context, is there a way of telling which `TimeSpan` columns belong together, since there could possibly be several?
- [ ] Sortation of entries: there are several legacy tables which contain input-fields where an integer value is used for the sortation of the data. This is something that could be implemented by some sort of drag-and-drop mechanism. Have you given this some thought? We could imagine that there will be a need of some mechanism for saving data which has been altered on client-side. This also comes to mind when e.g. thinking of `EditableColumn`s.
- [ ] Remembering state: up till now legacy tables stored the sortation and pagination in the session, so users would have the same view when returning to a table. Is there a plan for this scenario with the new data table as well?
- [ ] Data export: legacy tables could have been exported as a CSV or Excel file. Would it be possible in the future to have a similar mechanism for exporting the currently displayed data?
- [ ] `Filter`s: have you planed on having some sort of bridge between the data table and filter-service? We believe this is something that should definitely be considered, to avoid having multiple custom implementations by different consumers, since filtering will be something widely used.

- [ ] `Row` and `RowFactory`: Since there is a `PresentationRow` as well, we recommend to name these two interfaces `DataRow` and `DataRowFactory`, so there will be no confusion.
- [ ] PHPDoc comments: The `@param` and `@return` annotation for `array`s are invalid in `Row` and `Factory`. They should be `array<string, Column>` or `Action` instead. We also suggest to import these two interfaces, so modern IDEs can e.g. provide proper autocomplete.
- [ ] Documentation: Please provide some info about the usage of `DataRetrieval`s with comments. We know that the parameters are (only?) used for pagination, sortation, filtering, etc. but consumers and also developers would greatly benefit from some thoughts behind this. You could also add a class-level comment about how this interface would ideally be implemented by repositories directly.

Kind regards,
@thibsy and @Amstutz

nhaagen:
Dear @chfsx, @thibsy and @Amstutz,
thank you very much for looking into this.
I'll try to answer the comments as best as I can, however, please bear with me 
that some are of a very personally opionated and sowewhat speculative nature, since 
there are quite a few different application and thus needs for the Table Component; 
I expect this to grow, and not everything will be there in a first version.

I prefer the concept of Data/URI very much over a string, and we should use it wherever appropriate.
There are still shortcomings and pitfalls with it, yes, especially in construction.
With both ilCtrl and Data evolving (and other Services as well), I'm quite 
confident that those will be overcome soon, though, but entirely agree on the need to communicate to 
devs and give examples.

Before going to the common features (Sortation, Pagination, Filter,... ) let me touch on TimeSpan:
I certainly think that the concept of CombinedColumns is worthwile considering (and I'll think along); 
for now, there are both DateColumn and TimeSpanColumn, the latter taking a pair of dates.
For telling a correlation of dates, I think this is sufficient - either splitting into two cols or 
having both dates in one.

Which leads to sortation immediately: Overall, the idea is to delegate sortation 
(and pagination/filtering as well) to the repository, usually the database. 
This is reflected in DataRetrieval::getRows, where range, order, aspects and filter are mandatory params.

The current implementation with a click on the header is more or less a gimmick - i.e.  
the implementation, not the feature - and will change in the near future.
I'm currently working on viewcontrols (as in an input-container), and sortation is part of this.
With the PresentationTable, a consumer would have to manually align viewcontrol-paramters and pass them together,
since a change in one of them should not cancel the others. Bundling them together will
improve handling quite a bit.
Also, those controls will be built by the table itself, since all info is already there, and 
they should be completely transparent and of no concern to the consumer.
My plan is to indeed have them as dropdowns, and expanding just a little on the sortability of columns will 
give the necessary options: maybe we should enhance Column::withIsSortable to accept more than a boolean,
as to express sortability in only one direction or, in case of timespan, with sortation by a (sub-)aspect?

All view-parameters will be passed in the query; storing them (in a session) I do not really consider part of the table,
but if wanted, this should not be to hard to do ;) However, this hints to the need of initializing a view,
which can be done with the mutators on Data as withOrder, withSelectedOptionalColumns, etc.

The mentioned commonalities - a sortation, pagination et al. - in data retrieval is close to 
the db-layer, I suspect, and I currently have not given so many thoughts about this.
Do you have something like a blueprint/abstract repository in mind?

Finally, on the export: That's a good point, and I guess, a CSV is not much more 
than a somewhat altered rendering, triggered by a parameter (::asCSV() ?).

As for the changes - thank you very much for pointing out the issues, I'll fix; while at it,
I'll think up an example emphasizing on the real-world repository approach.

Best regards,
Nils

matthiaskunkel:
Jour Fixe, 20 MAR 2023: We highly appreciate the suggested UI component and accept the new Data Table for ILIAS 9. Once the UI component has been merged, it can be used for implementation - esp. within the Removal LUI project.

thibsy (changes requested):
Hi @nhaagen

Thank you for the feedback! We know that some features will be missing when introducing the new data-table, but like you said, we just wanted to point out these facts for you and the general public as well.

Please note that we will use the letters `C` and `I` as an abbreviation for the `ILIAS\UI\Component\Table` and `ILIAS\UI\Implementation\Component\Table` namespaces throughout this review when referring to objects.

Some of the changes will be interface-changes again because we overlooked some parts in the previous review.

PHP:

- [x] `C\TimeSpan`: if we don't provide some sort of `MultiColumn` or another concept of checking the validity of this column-type initially, we feel like removing the column altogether (incl. implementation parts). Right now timespan columns can and must be used like `C\Date` columns. We would of course highly appreciate if you take initiative here, but we don't see this as a necessity for the introduction of this table.
- [x] `I\Data::__construct()`: we don't really love how the object is created by calling "init"-methods which have side-effects on the instance and already use part of the instance variables. This way the order of method calls is crucial. We would prefer it if you would use methods with return-types and parameters to declare instance-variables directly. So e.g. `$this->initializeVisibleColumns()` becomes `$this->selected_optional_column_ids = $this->filterVisibleColumnIds($columns)`. Same goes for `setEnumeratedColumns()` and so on.
- [ ] `I\Data::getFilteredActions()`: using this method like `getFilteredActions(Single::class)` would lead us to believe we retrieve all actions which are instances of `Single`. We suggest to invert its behaviour, so it will only return actions of the provided type. This way `I\Data::getMultiActions()` and `::getSingleActions()` would also get more readable.
- [x] `I\Data::initializeVisibleColumns()`: you could use `ComponentHelper::checkArgInstanceOf()` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5656/files#diff-87a73ca4faa58bfa4f1071586b935e00c203f39f848da752b2b917ce60adae2bR82) which would be more consistent because you don't need to write error-messages yourself.
- [x] `I\Renderer::getActionRegistration()`: this method will inject values (the target-type) which are tightly coupled to the JavaScript logic. Any change of these values on either side will cause an error at some point. Do you have any idea how we could manage these kind of relations in the future, so we can prevent this (as good as we can)? We think it might help moving these values to constants which provide a brief description about them being coupled to the corresponding javascript file.
- [x] `I\Renderer::renderTableHeader()` I: you currently build the `Data\URI` using the `$_SERVER` variable. We think that it might be better if you'd use the `ServerRequestInterface` which can already be provided by the `C\Data` table. This would mean, that the request is required though and needs to be passed in the constructor, which wouldn't be so bad though. This way you could call `C\Data::getRequest()->getUri()` instead. Same goes for all the examples.
- [x] `I\Data::getVisibleColumns()`: you could and should probably use strict comparison in the `in_array()` call by passing `true` as third argument. Or are we missing something?
- [x] `I\DataRetrieval`: why does this class exist? please remove it if there is no good reason.

JavaScript:

- [x] Classes: thanks a lot for implementing your JavaScript logic as ES6 modules! Since a lot of components are still working with a revealing module pattern this point is highly optional, but we would still very much like to see it: [classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes). If you decide to refactor your logic accordingly, you could have a look at e.g. #5152 or the [newly introduced tooltips](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/UI/templates/js/Core/src/core.Tooltip.js). As of this year, there is also support for private fields and methods, which makes encapsulation way more easy (and makes revealing modules obsolete). Feel free to contact me on Discord if you have any questions about this.

PHP:

- [x] `C\Table::withTitle()`: this method is obsolete, as all tables require a title. Please remove this method and the according implementation-changes in `I\Table` and `I\Data` and test-cases.
- [ ] `I\Data::getRowFactory()`: as you already mentioned this is an anti-pattern and as we know they most likely stay there forever. The only thing standing in the way of creating an instance of the factory in the constructor already (injection would probably need too much internal logic outside the component), is `C\Data::withActions()`, which alters the factory's state when called. Such side-effects tend to cause confusion anyways, so we suggest to remove this method and expect the actions in table constructor.
- [x] `I\Data::initializeViewDefaults()`: this method looks rather fragile because of the mixed return types which could be returned by `array_keys()` and `current()`. It also feels weird using `current()` here, maybe use `array_pop()` instead? regardless, please implement this more safely by checking return-values (see also @chsfx comment).
- [x] `I\Renderer::renderStandardRow()`: please rename this to `renderStandardDataRow()` or something similar, by looking at the method name one can assume it renders _any_ table row, but it is very `C\Data` specific.
- [x] `I\Renderer` tsort_f & tsort_d: these strings occur throughout the renderer. Please move them to a constant, it might make sense to put them on the `C\Column\Column` interface.
- [x] `I\Column::format()`: please use the `ComponentHelper` over `assert()` calls in derived classes which override this method.
- [x] `I\Column\Number::withUnit()`: you should use `ComponentHelper::checkArgIsElement()` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5656/files#diff-eb63720c91de93fbb76f20176a0b32d6e107a643f5bba1efc9af12f30bb7bb7fR47) instead. You could also move the array-initialization to an internal constant, this way it doesn't get re-initialized on each call.
- [x] `I\Data::$title`: this instance variable is already declared in its parent, please remove it.
- [x] `C\DataRowFactory`: there is an unused import `Exception`, please remove it.
- [x] `I\Renderer::registerActionsJS()`: the `$js_onload` array is obsolete, you can just add the JavaScript binders directly to the component every time you retrieve a `Closure`. This way the second `foreach` loop can be removed as well.
- [x] `I\Renderer::renderTableHeader()` II: please remove the PHPDoc comment, it's not correct.
- [x] `I\Renderer::renderTableHeader()` III: please use the `Data\Order` constants of the class directly instead of accessing them dynamically. This makes code more readable.
- [x] `I\Renderer::renderTableHeader()` IV: on L320 and L322 you retrieve glyphs with an empty string as action. The action is optional though, so please remove the strings.
- [x] `I\Renderer::render()` I: please throw a `LogicException` at the end this method, if the provided `Component` cannot be rendered by this renderer.
- [x] `I\Renderer::render()` II: please remove the commented code.
- [x] `I\Factory`: please use dependency-injection for the action- and column-factory in `::column()` and `::action()` instead of creating a new instance on each call.
- [x] `I\Presentation`: you changed the the constructor signature of `I\Table` and now the presentation table is missing a parent-constructor-call. You can also remove the redundant `$title` instance variable.
- [ ] `I\Renderer::applyViewControls()` I: if we don't make the server-request required, you should use a null comparison on L240 and parenthesis when checking if there is a request available.
- [ ] `I\Renderer::applyViewControls()` II: you are parsing the request-parameters manually in there, but you should use `ServerRequestInterface::getParsedBody()` instead. Ideally we would even use the `ArrayBasedRequestWrapper`s here.
- [x] `I\Renderer::buildMultiActionsDropdown()` and `I\Renderer::getSingleActionsForRow()` : declaring `$dropdown` is unnecessary, please return the dropdown directly.
- [x] Anonymous functions: there are several occasions where anonymous functions are used without type-hints. Many of them could also be static, which improves scoping and performance in some cases. Please check your PHP files again and use type-declarations for arguments and return-values whenever possible. Cases so far:
	- [x] `I\Renderer::171`: missing type-hints, can be static
	- [x] `I\Renderer::175`: missing type-hints, can be static
	- [x] `I\Renderer::357`: missing type-hints, can be static
	- [x] `I\Renderer::364`: missing type-hints, can be static
	- [x] `I\Renderer::399`: missing type-hints, can be static
	- [x] `I\Renderer::410`: missing type-hints, can be static
	- [x] `I\Renderer::436`: missing type-hints, can be static
	- [x] `I\Renderer::504`: missing type-hints, can be static
	- [x] `I\Data::95`: missing type-hints, can be static
	- [x] `I\Data::106`: missing type-hints, can be static
	- [x] `I\Data::241`: can be static
	- [x] `I\Data::281`: missing type-hints
	- [x] `I\DataRow::83`: missing type-hints
- [ ] PHPDoc comments: there are still several cases across multiple files where the PHPDoc annotation for `array`s are invalid (see e.g. [here](https://phpstan.org/writing-php-code/phpdoc-types#general-arrays)). Please also note that classes might need to be imported when in a different namespace. Cases so far:
	- [x] `I\Data::$columns`, `I\Data::$actions`, `I\Data::getVisibleColumns()`
	- [x] `C\Data::withActions()`
	- [x] `C\Factory::data()`
	- [x] `I\DataRowFactory::__construct()` (also note, that the `@var` annotation is obsolete)
- [x] `I\Action\Action::getTarget()`: please remove the method-comment.
- [x] `I\Data::$actions`: please add the `array` type-declaration.
- [x] `I\Data::$columns`: please add the `array` type-declaration.
- [ ] `I\DataRow::$record`: please add the `array` type-declaration and rename it to `$records`. Also remove the PHPDoc as it doesn't add anything.
- [x] `I\DataRow::$disabled_actions`: please add the according `@var array<string, bool>` PHPDoc annotation.
- [x] `I\DataRow::__construct()`: the PHPDoc doesn't add anything here, please remove it.

JavaScript:

- [x] `table.data.js` L61 and L92: there is a semicolon instead of a comma, which will break the rest of the script.
- [ ] Variables: please use `let` or `const` whenever possible instead of `var`. This is due to `var` NOT being block scoped (see [here](https://developer.mozilla.org/en-US/docs/Glossary/Scope)). Please replace all `var` declarations in your JavaScript files.
- [x] Multi-Declarations: please declare each variable with a dedicated `let` or `const` keyword instead of using comma-separation.
- [ ] Code-Style: please have a look at the JavaScript [code-style](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js/js-coding-style.md) we are using. There are several things which don't comply with it yet, you could use the linter as described in the linked document to check where actions need to be taken.
- [x] JSDoc: there's currently no official documentation on our side about this, but we strongly recommend using [JSDoc](https://jsdoc.app/tags-type.html) whenever possible. Therefore we would really appreciate it if you provide at least `@returns`, `@param` and `@throws` annotations in your JavaScript files.

Kind regards,
@thibsy

klees:
Hi all!

Just want to give my two cents regarding various items of the discussion:

* `C\TimeSpan`: Could you explain a little more how and why you see this being related to a possible `MultiColumn` feature and why we should not be able to check validity of data here? From my perspective the afair is rather simple: A `C\TimeSpan` column accepts to dates (the data type, not the fruit) in an array. These are displayed in one column ("12.01.2023 - 14.01.2023"), where the individual dates are formatted according to date format applied to the column definition. So this is supposed to be one column, not two. This could be leveraged two simplify how the value is display: "12. - 14.01.2023". Of course, we are lacking a data-type to bundle express a timespan. But the same is true for the `email` column.
* JS-classes: Please do it =)
* `I\Data::getRowFactory()`: I like factories and proper DI and stuff just like the next person. But having the requirement that actions can only be added at the constructor would throw the baby out with the bathwater. We would not be able to say something like: "I want that table but with these actions". This would be a huge blow for composability and reusability. Imaging: I build one table for people who can only "view". Than use that table `withActions` for the people that can also "manage". So, we could inject a `DataRowFactoryFactory` instead of calling `new` in the method and go full Enterprise Java (with capitals!). I'm not really against doing so, but also: do we really want to?

Best regards!

Amstutz:
Hi @nhaagen 

I see that most of the points of the list are already taken care of, for some, @klees gave some feedback. is this ready for another go from your PoV, or are you still working on it?

Thx. looking forward to this component.

thibsy (changes requested):
Hi @nhaagen

I assume you are still working in some of the requested changes, since you haven't responded to the comment of @Amstutz yet. However, we want to push this PR a little since projects like "LUI" are certainly waiting for this component. I have adopted some points of my previous review which I think must still be implemented or answered before we can merge this.

Huge thanks for refactoring the JavaScript logic of the table with ES6 classes! Since you made a lot of changes I came up with this follow-up review which will primarily focus on the previously mentioned points and the JavaScript logic. You may ignore the previous review from here on out.

PHP:

- [x] `I\Data::getRowFactory()`: In my previous review I explained how we should not accept anti-patterns as they tend to stay forever. However, @klees made an excellent point on blowing this thing out of the bathwater when expecting all attributes on construction already. Hence my proposal: what if we think in terms of [builders](https://refactoring.guru/design-patterns/builder) instead of factories? By implementing `I\Data::withActions()` II we could already configure the attributes of such a builder with corresponding method calls instead of injecting the values on construction. This way `I\Data::getRowFactory()` or then `I\Data::getRowBuilder()` could reuse this builder instance. public configuration methods would also be hidden behind the public interface. If you need more details about this idea you can contact me on Discord.
- [x] `I\Data::withActions()` II: Since `I\Data::getSingleActions()` and `::getSingleActions()` are possibly called several times (also by the according `has...()` methods), we should probably already distinguish these actions in the `withActions()` method call and store them in separate instance variables. The `get...()` and `has...()` methods all end up calling `getFilteredActions()`, making these calls somewhat expensive otherwise.
- [ ] `I\Data::getVisibleColumns()`: Same goes for this method actually, we could already store visible columns on construction so this method call gets less expensive too. It is at least called two times right now, possibly more in the future.

JavaScript:

- [x] ES6 classes: I really appreciate you using `class` syntax now, but it seems like you just wrapped the previous module in a class. This was not the intention, because the beauty of implementing classes is, that we can write code which is tied to an instance using `this`. Having one `Data` class for all the data-tables is very impractical and defeats the purpose. When building classes for UI components, think of how one class instance represents the according HTML structure. I agree that we need a service which keeps track of our table instances, but this should be rather minimalistic. We have not yet established an official pattern for the UI components, but when developing the markdown input (#5152), using a "factory" for managing instances felt right. Taking this component as an example, please try to refactor your `Data` class to deal with exactly one table represented in HTML. Doing so, allows us to gather all required DOM elements on construction **once** and get's rid of additional init-methods like for keyboard navigation.
- [x] `Keyboardnav`: Once again, thinking in terms of classes representing HTML structures, the keyboard navigation seems to be something which is tightly coupled to the HTML structure of a data-table. The `Data` class is not very bloated right now and I think we should get rid of this class and merge it with `Data`, as this class should represent the table HTML. Methods like `focusCell()` would also be way more fitting on a table class.
- [ ] `Params`: This looks a lot like a util class, which is something we should avoid. As much as I love seeing classes, I don't think this is the use-case for one. Instead, we should be thinking in terms of functions here. You could name the file e.g. `UrlManipulations.js` and let it export each function which is currently a public class method. This way we don't inject some collection of methods into the `Data` table and can import and call internal functions instead.
- [x] `Keyboardnav.onKey()`: To accept an instance of itself `_self` instead of using `this` seems illogical to me. It looks like we only need to do this because we create one instance of `Keyboardnav` for all tables and then use `Keyboardnav.init()` to bind the instance to every data-table.
- [x] `Data.#actions_constants`: `@type` annotation should be more precise than `Object`. You can use syntax like `@type {{type: {url: string,}, opt: {...}}}` to explain how the object might look like. Please try to be as specific as possible with the JSDoc, since this is our only reference for type-hints in JavaScript.

PHP:

- [x] `I\DataRetrieval`: Please remove this class, you only use it to extend from in the UI examples, but anonymous classes should use `implements` instead.
- [x] `I\Data::withActions()` I: We need to check the provided array for `C\Action` instances with help from our `ComponentHelper`.
- [x] `I\Column::format()`: please use the `ComponentHelper` over `assert()` calls in derived classes which override this method.
- [x] `I\Renderer::registerActionsJS()`: the `$js_onload` array is obsolete, you can just add the JavaScript binders directly to the component every time you retrieve a `Closure`. This way the second `foreach` loop can be removed as well.
- [x] `I\DataRowFactory::__construct()`: the `@param` annotation is still invalid, there should be no space between `array` and `<>`.
- [x] `I\Data::$actions`: please add the `array` PHP type-declaration.
- [x] `I\Data::$columns`: please add the `array` PHP type-declaration.
- [ ] `I\DataRow::$record`: please add the `array` PHP type-declaration and rename it to `$records` (plural) to indicate it being a list.
- [ ] Code-Style: Please run the PHP-CS-Fixer with the latest configuration over all changed PHP files. There are a lot of places where our code-style isn't met.

JavaScript:


- [x] `Keyboardnav.onKey()`: please be more concise with event-listener hooks. Try to explain with the name already what the listener will do. Hence rename it to e.g. `navigateCellsWithArrowKeys()` or similar.
- [x] Private: I think you did not fully understand the syntax of [private](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) functions and properties. The `#`, unfortunately, is part of the function or property name. Therefore you need to access them like `this.#function()` and `this.#property`. There are some cases where you are now creating and using public properties by accident:
	- [x] `Data`
	- [x] `Keyboardnav`
- [x] Naming conventions I: According to our code-style we should stick with these [naming conventions](https://github.com/airbnb/javascript#naming-conventions). We should especially enforce this for our file-names, which should be named after their default export. I know that we currently don't do this anywhere, but we need to start somewhere.
	- [x] `table.data.js` would become `Data.js`
	- [x] `params.js` would become `Params.js`
	- [x] `table.keyboardnav.js` would become `Keyboardnav.js`
- [x] `Data.doAction()` I: Unfortunately our signal-management still depends on jQuery, therefore you had no choice when using it. You did it like a pro though by injecting it into the `Data` class as a dependency and considering it in the `rollup.js` config. However, you ignored your good intentions by accessing it globally (`$` instead of `this.#jquery`) in this method.
- [x] `Data.initActionConstants()`: Please avoid using arrays as parameter lists. There are exactly 4 values right now, please accept them explicitly. In regards to point "ES6 classes", these values should probably also already be expected in the constructor, so you could drop this method.
- [x] `Data.registerAction()`: I think we should throw an exception if `type` is not either SIGNAL or URI.
- [x] `Data.collectSelectedRowIds()`: Please use a [`forEach()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) statement instead of a common `for` loop. This way you don't need to declare `col` and `i` beforehand and the code gets way more readable because you don't need to retrieve the `col` from the array.
- [x] `Params.amendParameterToUrl()`: Same for the `for` loop here.
- [x] `Data.selectAll()`: Same for the `for` loop here.
- [x] `Keyboardnav.#keys`, `Keyboardnav.#supported_keys`: In JavaScript there's no such thing like "class constants", or even real "classes" for that matter. Regardless, the values contained in these two properties are only read and should probably also not be manipulated at runtime. I therefore suggest to move these values into `const` declarations (outside of the class).
- [x] `Keyboardnav.init()`: Use `document.getElementById()` instead of `document.querySelector()`. We should also throw an exception if the element isn't found or at least use optional chaining (`?.`) when adding the event listener.
- [x] `Keyboardnav.focusCell()`: The `@param` annotation for `table` and `cell` should be `HTMLTableElement` and `HTMLTableCellElement`.
- [x] `Data.#jquery`: `@type` annotation should be `jQuery` instead of `Object`.
- [x] `Data.#kbnav`: `@type` annotation contains an invalid object `Keyboardnavigation`.
- [x] `Keyboardnav.onKey()` I: The `@param` annotation for `event` should be a `KeyboardEvent` instead.

_Note that for a first iteration you don't need to consider my suggestions, it could just help for future revisions of this component and stays here for now as sort of documentation._

Kind regards,
@thibsy

Amstutz:
hi @nhaagen 

I see that you pushed a lot of work last week. However, there are still many checkboxes open. Is this close for a final review?

thx for some feedback.

nhaagen:
Hi @thibsy ,
I finally reworked the JS, and with the latest ECMA version even the linting looks ok to me ;)
Params are still in, though; @tbongers-cat is currently working on a general URL-abstraction in both PHP and JS, and I expect to replace the entire class with that, probably along with the integration of ViewControls.
The RowFactory is a builder now, and Actions are being split up early. This, however, felt awkward with the columns; I think, it's more intuitve to keep them together.
Please note that there is now an additional function in DataRetrival asking for a total count of the records; I discussed options for Paginations qo quite a length with our designers, and we will ultimately want this.
With PHPStan, the remaining issues mainly concern filters and additional parameters - we're not with filters, yet, so I ignored that for the time beeing.
So, thanks again for your deep introspection, and I certainly hope I caught the lot of the issues.
Best regards,
Nils

thibsy (changes requested):
Hi @nhaagen

Thx a lot for implementing the requested changes along with some of my suggestions!

- [ ] `C\DataRetrieval::getTotalRowCount()`: you have introduced a public interface change, which does not quite belong with the current state of the data-table. Since this would need to be presented at JF again, are you sure we cannot add this method in a separate PR, covering the view-controls? This might fit the discussion then and it certainly helps to have more context.
- [x] `I\DataRow::$record`: could you explain why this instance variable is public? is there also a reason why you cannot use PHPs `array` type-declaration here?
- [x] `I\StandardRow`: could you explain why this class exists? a quick search after the `*selectable` methods did not find any results, these methods are also hidden behind the public interface so why does it exist?

- [x] ID-attribute: you could probably set the HTML ID-attribute directly on the table instead of the wrapper. It looks like all your JavaScript selectors only ever search elements within the table. This would avoid recurring `this.#table.getElementsByTagName()` calls when focusing cells.
- [x] JS null-coalesce: you often directly access the first array-element with index 0, which could lead to type-errors. You should use either a null-coalesce (`??`) or something like [`.shift()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift) if the original array can be altered.

- [x] Annotations: there are still some invalid PHP annotations in:
	- [x] `C\DataRow::getColumns()`
	- [x] `C\DataRow::getActions()`
- [x] `I\StandardRow::$selectable`: you must initialize a default value, otherwise calling  `isSelectable` leads to a fatal.
- [x] `DataTable.#table`: the `@type` annotation should be a `HTMLDivElement`, since the ID is the surrounding div.

Kind regards,
@Amstutz and @thibsy

nhaagen:
Hi @Amstutz and @thibsy, 
thanks for bearing with me;)

> C\DataRetrieval::getTotalRowCount(): you have introduced a public interface change, which does not quite belong with the current state of the data-table. Since this would need to be presented at JF again, are you sure we cannot add this method in a separate PR, covering the view-controls? This might fit the discussion then and it certainly helps to have more context.

I cannot really decide on that; my intentions were to get the table useable soon and keep devs out of trouble then; adding the rowcount later would mean to re-touch implementations, while having it now, the further developements on the table would - as far as I can anticipate - be internal.
ViewControls, eg, should be built by the table itself w/o the consumer configuring more than the columns.
Same goes for filters; however, I did not give filters too many thoughts, yet, and there will have to be some sort of flag on the columns, so it evens out my reconning.
I'd gladly rely on your recommendation here, so maybe give me a ping and I'll remove the method ;)


> ID-attribute: you could probably set the HTML ID-attribute directly on the table instead of the wrapper. It looks like all your JavaScript selectors only ever search elements within the table. This would avoid recurring this.#table.getElementsByTagName() calls when focusing cells.

You are perfectly right that currently the JS is about cells, only. However, If need arises (in the future) to access elements, that are still inside the wrapper, but not in the table, I'd very much prefer searching downwards over upwards. Maybe my suggestion in the latest commit is a good compromise?

Best,
Nils

thibsy (approved):
Hi @nhaagen 

Thx a lot for your quick feedback!

You made an excellent point on touching the `C\DataRetrieval` again soon after the initial release. The change still needs to be presented at the JF, but our rules state [here](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/UI/docu/rules.md#modification-of-existing-ui-components) that coordinators may allow exceptions if the change is non-breaking and would slow down further development, as it would be the case for the LUI project in this scenario. Therefore we will present the change at the next JF.

Your compromise about the ID-attribute makes sense to me, the change of splitting up the elements in a `#component` and `#table` property looks good. Since everything else now looks good as well, I'll merge this right away!

Thanks again for your effort and this huge contribution to the UI Components.

Kind regards,
@thibsy

PR #5657, opened 2023-03-06
Fix crs to course typo

padvincenzo:
Hi
 I've noticed a typo in the property name: in two methods this variable is referred to as `course_start_time_indication` instead of `crs_start_time_indication`.


padvincenzo:
This fix is also for release 7 8 and trunk.

smeyer-ilias:
Thanks @padvincenzo for fixing this

PR #5661, opened 2023-03-07
Lightbox Dark Scheme and Indicators on top

chfsx:
This PR proposes the following two amendments:

- The Page Indicators are now above the LightboxPage for better readability.
- A withDarkScheme() for lightboxes that only display images.

Both items were discussed yesterday in JF and refer to the following feature: https://docu.ilias.de/goto_docu_wiki_wpage_7725_1357.html


Amstutz:
Hi @chfsx 

Thx a lot for this PR. There is definitely a need to revamp the lightbox modal. @yvseiler gave thumbs up to the optical changes, so they are fine with me as well. Thx a lot for the good work.

Note, that we introduced the BEMIT convention for naming classes better in our inverted triangle ITCSS structure, see: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/templates/Guidelines_SCSS-Coding.md#naming-conventions . In this case, you can leave as is, since all surrounding classes do not follow that structure yet, but make sure to use them if proposing new components in the future. 

There is just one question:

- [x] Interface Change: I dislike the design decision of Black vs Whit BG on a public interface level. We try to keep away any design decisions from Devs. If I read the rule correctly, we could just make this decision for the devs, by implementing something like "If there is at least one LightBoxImagePage Inside, we take black, otherwise white". Correct?

Thx for the examples and tests.
Kindly,
@Amstutz

chfsx:
@Amstutz thanks for the review! I updated the PR accordingly, Lighboxex now get the dark scheme if there is at least one ImagePage in it: https://github.com/ILIAS-eLearning/ILIAS/pull/5661/files#diff-d482f49ead303f97c7777234df7222a7630f15a80524e359589572aa7f5e6d2fR51

I therefore removed the interface to choose the scheme from the outside

PR #5662, opened 2023-03-07
T&A #37034 Crash when no entry has been selected in preview and malformed error text entries are included

jcopado:
In a malformed question (double closing )) without initial (() the question is anaylzed without error,s, but error come when previewing the question. I modified the  getPointsForSelectedPositions($positions) to ensure the user get 0 points if no word has been selected. and getBestSelection($withPositivePointsOnly = true) to ensure the malformed closing braket does not crash the question.

An error message telling the user that at least one word must be selected would be nice to inform the user.

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assErrorText.php
+++ b/Modules/TestQuestionPool/classes/class.assErrorText.php
@@ -650,11 +650,13 @@ class assErrorText extends assQuestion implements ilObjQuestionScoringAdjustable
                     }
 
                     if ($correct_solution && !$in_passage) {
-                        $errorobject = $this->errordata[$errorcounter];
-                        if (is_object($errorobject)) {
-                            $item = strlen($errorobject->text_correct) ? $errorobject->text_correct : '&nbsp;';
+                        if (isset($this->errordata[$errorcounter])) {
+                            $errorobject = $this->errordata[$errorcounter];
+                            if (is_object($errorobject)) {
+                                $item = strlen($errorobject->text_correct) ? $errorobject->text_correct : '&nbsp;';
+                            }
+                            $errorcounter++;
                         }
-                        $errorcounter++;
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @jcopado !
I'm unsure about this fix. If I'm getting this right, this would lead to a the passage of the example "wrods aer wrogn" to be expanded to "wrods aer wrogn)) I think" in `ilAssErrorText::getPointsForSelectedPositions()` and `ilAssErrorText::getBestSelection()`. This could actually be, what we want, but I'm rather sure it isn't. Am I'm not seeing something? My feeling is that we should rather work on the condition statements to avoid this situation, e.g [this statement here](https://github.com/ILIAS-eLearning/ILIAS/pull/5662/files#diff-b2e15ed310911bf8bc33ea76afe7fd2a030a4299c252ef907d54cc7c3adc4c61L847) should only be entered when we are either $inProgress or we are on an opening bracket. We don't actually care, if we find an opening bracket, but we don't have opening brackets, right?
Thanks again and best,
@kergomard

jweber-unibe:
I am not quite sure, but the reference to _'malformed question without initial (('_ might be looking in the wrong direction – @jcopado probably being prompted this way by an error in my bug report (https://mantis.ilias.de/view.php?id=36896) where I mistakenly closed the example 'Error Text' with such 'unmatched closing double brackets' [–> _"These ((wrods aer wrogn)) I think **))**",_ instead of _"These ((wrods aer wrogn)) I think"._]
If this is the case, I am sorry for the misunderstanding – fact is, the issue occurs _for any perfectly normally formed_ 'Error Text' with multi-word marking! (In the Mantis description, I have corrected my mistake now.)

kergomard:
Hi @jcopado

Can I help you in any form with this? I hope my question was clear, otherwise, please contact me, so that we can discuss this quickly and get rid of one more error in the ILIAS Test & Assessment.

Thank you very much and best,
@kergomard

dsstrassner:
This PR is linked in the protocol of [TechSquad meeting 51](https://docu.ilias.de/goto_docu_xpdl_12645.html):
-  JC: Solves a Problem, but not the intended one.
- SK: Please open an Issue. Please answer the question in the PR and then we can also tackle the problem in the report.

Hi @jcopado,
please change the Issue number to the new issue and link the new issue in your first statement. Also answer please the question from @kergomard, so the has the chance to help you and merge or close this PR accordingly. 

Greetings
@dsstrassner

jcopado:
I have changed the ID now solves 37034.
37034 happens only on malformed multi error text entries, where the inicial separator "((" is missing.
I hope at the end of the next week I have time to check this one more time.

dsstrassner:
Hi @kergomard & @jcopado!
What is the status of this PR?

kergomard:
Thanks again @jcopado for this PR! I will close this and come back to the isse, if we decide to not overhaul the whole ErrorText-Question. I will at least change the parsing in any case.

Thanks again and best,
@kergomard

PR #5663, opened 2023-03-08
T&A Bugfix #0036907: Reset Pool Sync Broken

fhelfer:
https://mantis.ilias.de/view.php?id=36907


kergomard:

PR #5664, opened 2023-03-08
[PROPOSAL] Add npm install to composer install and update commands

PurHur:
In addition to https://github.com/ILIAS-eLearning/ILIAS/pull/5128 i propose to combine the npm install/update with the corrosponding composer commands.

This removes one step for admins and devs to setup/update an ilias installation. It also respects the --no-dev argument. 

PS:
If someone is has some better bash skills for the --no-dev/--omit=dev flag and is able to merge the command in one line this would improve the change a lot. But i dont know how to bash ;)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,11 +18,10 @@
 		],
 		"post-install-cmd": [
 			"@php libs/composer/rmdirs.php",
-			"[ $COMPOSER_DEV_MODE -eq 0 ] && npm install --omit=dev --ignore-scripts || npm install --ignore-scripts"
+			"[ $COMPOSER_DEV_MODE -eq 0 ] && npm ci --omit=dev --ignore-scripts || npm ci --ignore-scripts"
 		],
 		"post-update-cmd": [
-			"@php libs/composer/rmdirs.php",
-			"[ $COMPOSER_DEV_MODE -eq 0 ] && npm update --omit=dev --ignore-scripts || npm update --ignore-scripts"
+			"@php libs/composer/rmdirs.php"
 		],
 		"test-php-all": [
 			"./libs/composer/vendor/bin/phpunit -c ./Services/PHPUnit/config/PhpUnitConfig.xml --colors=always --disallow-todo-tests"
--- a/composer.json
+++ b/composer.json
@@ -17,10 +17,12 @@
 			"@php setup/cli.php build-artifacts --yes"
 		],
 		"post-install-cmd": [
-			"@php libs/composer/rmdirs.php"
+			"@php libs/composer/rmdirs.php",
+			"[ $COMPOSER_DEV_MODE -eq 0 ] && npm install --omit=dev --ignore-scripts || npm install --ignore-scripts"
 		],
 		"post-update-cmd": [
-			"@php libs/composer/rmdirs.php"
+			"@php libs/composer/rmdirs.php",
+			"[ $COMPOSER_DEV_MODE -eq 0 ] && npm update --omit=dev --ignore-scripts || npm update --ignore-scripts"
 		],
 		"test-php-all": [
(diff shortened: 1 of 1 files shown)


PurHur:
Alternatively you could use this composer package which does the same:

https://github.com/eloquent/composer-npm-bridge

thibsy:
Hi @PurHur

I think something like this should do the trick:
```bash
[ $COMPOSER_DEV_MODE -eq 0 ] && npm install --omit=dev --ignore-scripts || npm install --ignore-scripts
```
But are you sure `$COMPOSER_DEV_MODE` is `0` for dev-mode instead of `1`?

Kind regards
@thibsy

PurHur:
That is true. It was switched....

Thank you for your bash learning. I changed that line.

nicoroeser:
TB 2023-04-04: many thanks for this pull request. We have looked at it in today’s Technical Board meeting, and found we need some more time to discuss this further. We are planning to get back to you in ~2 weeks.

klees:
Hi @PurHur,

thanks again for the proposal, we definitely see your point regarding ergonomics when installing ILIAS or updating dependencies. The former is definitly useful for a lot of people (admins, devs). Updates, on the other hand, should mostly be made on single dependencies, only @fwolf-ilias should regularly update all dependencies.

Regarding the implementation we think it will be odd if composer triggers another dependency management tool. It might just be irritating for `install`, as this will mostly just refresh the locked list of dependencies. It might be just dangerous for `update`, as we will change the dependencies there and it might not be obvious that this is true for php- and js-dependencies.

Would you be willing to provide an alternative implementation that bundles both installs and updates in two scripts `CI\install_dependencies.sh` and `CI\update_dependencies.sh`?

Thanks a lot!

PurHur:
Hello,

your argument with the `update` command is right. I see how that doesnt fit into the ilias dev cycle, so this could be removed.

But in the case of `install` i dont think it will be odd to combine different tools into one command. This is done in a lot of software projects. Using composer as single point of tool is also done very often and i see more and more stuff migrating to the composer 'build pipeline'. Since the introducing of the `build-artifacts` script, composer is not only used as dependency management tool but as a build tool (which is good). So adding npm install is just another script in the build process of an ilias installation.
In my opinion even the `php setup/setup.php update` command should be an added script there, but this is another topic =)

Here is just one example of an 'modern' php applications composer scripts:
https://github.com/mautic/mautic/blob/5.x/composer.json#L121
(Interesting point FYI: They do the npm ci on install and update and no npm update at all)

Also i disagree to create a complete new way to install dependecies. This would just create more noise in the cycle when there are standard tools to use.

If the jourfix/board doenst want to include npm in the composer process feel free to close this PR. 

Greetings
Purhur

klees:
Hi @PurHur,

I did not see this in the context of "using composer as a tool to manage the build pipeline in general". I'm intrigued, since we have been searching for some easy and unified way to run various scripts before.

I will take this to the JF and discuss there, as this is in some regard a directional decision we should be making with enough support. Care to join and present your argument?

Kind regards!

matthiaskunkel:
Jour Fixe, 10 JUL 2023 : We discussed the suggestion by @PurHur and its implications on the build process. There is a soft vote for an extension of composer to JS libraries and no one objects. Discussion will be continued within this PR.

PurHur:
I removed the automatic update and changed the install to clean install because of this PR: https://github.com/ILIAS-eLearning/ILIAS/pull/6365

kergomard:
Hi @PurHur, hi all
This PR was now open for quite a while and we believe that we could gain from integrating the install/update processes. We also believe that this would need some more conceptual work and sadly we can't find the resources right now, to take care of this. So I close the PR for the time being. If somebody would have the time to work on this, we would be very thankful and are available for discussions and support.

Thanks to all involved and best,
@kergomard in the name of the Technical Board

PR #5665, opened 2023-03-08
Object: Refactor AutoRating

kergomard:
Couldn't we move the whole auto-rating stuff to `ilObject::putInTree()`? This to me looks far more logical. It changes the behavior slightly, as this would also set the notifications, when an object with auto-rating is linked.

See: https://mantis.ilias.de/view.php?id=34112

### DIFF
--- a/Modules/File/classes/Processors/class.ilObjFileAbstractProcessor.php
+++ b/Modules/File/classes/Processors/class.ilObjFileAbstractProcessor.php
@@ -83,6 +83,8 @@ abstract class ilObjFileAbstractProcessor implements ilObjFileProcessorInterface
         if ($create_reference) {
             $file_obj->createReference();
         }
+
+        $file_obj->processAutoRating();
         $this->gui_object->putObjectInTree($file_obj, $parent_id);
 
         return $file_obj;
--- a/Modules/LTIConsumer/classes/class.ilObjLTIConsumerGUI.php
+++ b/Modules/LTIConsumer/classes/class.ilObjLTIConsumerGUI.php
@@ -273,9 +273,6 @@ class ilObjLTIConsumerGUI extends ilObject2GUI
                 $newObj->applyDidacticTemplate($dtpl);
             }
 
-            // auto rating
-            $this->handleAutoRating($newObj);
-
             $this->afterSave($newObj);
 
             return;
@@ -504,8 +501,7 @@ class ilObjLTIConsumerGUI extends ilObject2GUI
             if ($dtpl) {
                 $newObj->applyDidacticTemplate($dtpl);
             }
-            // auto rating
-            $gui->handleAutoRating($newObj);
+
             $newObj->setProviderId($provider->getId());
             $newObj->setProvider($provider);
             // custom params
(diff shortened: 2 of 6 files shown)


Modules/LTIConsumer/classes/class.ilObjLTIConsumerGUI.php
```diff
@@ -504,8 +501,7 @@ public function saveContentSelection(ilLTIConsumeProvider $provider, string $tok
             if ($dtpl) {
                 $newObj->applyDidacticTemplate($dtpl);
             }
-            // auto rating
-            $gui->handleAutoRating($newObj);
```
kergomard, line 508: @Uwe-Kohnle : I'm unsure why you are doing this manually here. Would the putInTree()-Approach cover your cases?

alex40724 (changes requested):
First of all, I think we agree that this is not well de-coupled code. But this change would not work:

"It changes the behavior slightly, as this would also set the notifications, when an object with auto-rating is linked."

Which would be simply wrong behaviour = bug. This is intended to be set for new objects, not for existing ones being linked. This might lead to great confusion if existing objects are suddenly changed due to linking in the background.

alex40724:
The rating service could set this listening to an object event that is fired on creation, but including the ref id. This would remove the rating code from the object service.

The event 'Services/Object'/'create' does not include the ref id. 
The event 'Services/Object' / 'putObjectInTree' is not only creation related.

In my opinion the main problem is, that ilObjectGUI->saveObject contains the interesting business logic, which is wronly placed in a GUI class. The BL here is the process "create > putintree > apply didactic templates > handle auto rating".

These steps should be performed in a method in ilObject instead (or any other non GUI BL class). And I think this method then should be called, e.g. in createFileObj of ilObjFileAbstractProcessor, which currently does putintree (but not applying didactic templates or handle auto rating) itself.

If this method would exist in ilObject, it could raise an event at the end. The rating service could listen to this event, so the rating code would disappear from the Object service.

kergomard:
@alex40724 : I think the behavior can actually be changed more easily. This is not the most elegant of all solutions, but it is a step in the right direction, from where I stand. See: https://github.com/ILIAS-eLearning/ILIAS/pull/5665/commits/3cbd38aefb5e64c9cf16fb57d39e89d42e01f2d4

Modules/File/classes/Processors/class.ilObjFileAbstractProcessor.php
```diff
@@ -83,6 +83,8 @@ protected function createFileObj(
         if ($create_reference) {
             $file_obj->createReference();
         }
+
+        $file_obj->processAutoRating();
```
kergomard, line 87: @chfsx : FYI.

chfsx: @kergomard Thanks a lot!

kergomard:
@alex40724 and @Uwe-Kohnle : Could you maybe have a look at the changes in your part of the code until the end of the week? This way, we could merge this and get rid of one more issue in ILIAS 8. Thank you very much!

kergomard:
Hi @alex40724 and @Uwe-Kohnle 
This is a notice in case you see this differently: I will merge this on Wednesday 2023-03-22. I think the changes in your components are trivial enough and we should get rid of this error, before the release of ILIAS8.
Thank you very much and best,
@kergomard

kergomard:
Hi @alex40724 and @Uwe-Kohnle 
After contacting you through multiple channels I didn't get any feedback on this PR. This is the last call. I will merge this tomorrow Friday 31 of March in the morning.
Best,
@kergomard

PR #5666, opened 2023-03-08
ilUtil::makeClickable: Open links in new tabs

lscharmer:
This PR fixes mantis bug: https://mantis.ilias.de/view.php?id=34573.

#### This PR resurrects the removed `$detectGotoLinks` parameter of `ilUtil::makeClickable`
Because of this the method `ilUtil::replaceLinkProperties` is readded (with some bug fixes).
Additionally a unit test is added for the `ilUtil::makeClickable` method, which also tests the `ilUtil::replaceLinkProperties` method.

#### MakeClickable Transformation
The MakeClickable Transformation will now add by default the attributes `target="_blank" rel="noopener"`.
This behaviour was extracted from the removed `ilMWFakery::makeExternalLink`: https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/Services/Utilities/classes/class.ilMWParserAdapter.php#L87 (pereviously used by `ilCtrl::makeClickable`)

I refactored the `MakeClickable` class. Previously some regex calls where using `preg_match` and others `mb_ereg`. To make this consistent i extracted this into one generic regex call.
Currently I have chosen to use `preg_match` as my research did not find any benefit of using one over the other (`preg_match` works in this case with multibyte, as it is tested in the unit tests).

@chfsx if there is a reason why we should use `mb_ereg` instead of `preg_match` I will gladly adapt the PR to use `mb_ereg` instead.


mjansenDatabay:
@smeyer-ilias / @alex40724 / @mjansenDatabay

According to the TB discussion/decission (21st MAR 2023) we would like to inform all consumers of `ilUtil::makeClickable` ...

that is:

- `Modules/Group`: @smeyer-ilias 
- `Modules/Course`: @smeyer-ilias 
- `Modules/Session`: @smeyer-ilias 
- `Services/ADT`: @smeyer-ilias 
- `Services/Calendar` :@smeyer-ilias 
- `Services/Mail` :@mjansenDatabay
- `Services/News`: @alex40724 

... , that this utility method is deprecated and will be removed with ILIAS 10. There is already an alternative in the `Refinery`.

**But**: There are some special additions (re-introduced with this PR) which are and *never* will be part of the `Refinery Transformation`, namely the detection of `goto`-links and the replacement of the URL text node with the corresponding object title. If this functionality is relevant for your components and use cases/business logic, feel free to add this functionality locally to your components or suggest a new service which provides this outside of `ilUtil`.

PR #5667, opened 2023-03-08
ilUtil::makeClickable: Open links in new tabs (trunk)

lscharmer:
Sister PR of #5666


mjansenDatabay:
@smeyer-ilias / @alex40724 / @mjansenDatabay

According to the TB discussion/decission (21st MAR 2023) we would like to inform all consumers of `ilUtil::makeClickable` ...

that is:

- `Modules/Group`: @smeyer-ilias 
- `Modules/Course`: @smeyer-ilias 
- `Modules/Session`: @smeyer-ilias 
- `Services/ADT`: @smeyer-ilias 
- `Services/Calendar` :@smeyer-ilias 
- `Services/Mail` :@mjansenDatabay
- `Services/News`: @alex40724 

... , that this utility method is deprecated and will be removed with ILIAS 10. There is already an alternative in the `Refinery`.

**But**: There are some special additions (re-introduced with this PR) which are and *never* will be part of the `Refinery Transformation`, namely the detection of `goto`-links and the replacement of the URL text node with the corresponding object title. If this functionality is relevant for your components and use cases/business logic, feel free to add this functionality locally to your components or suggest a new service which provides this outside of `ilUtil`.

PR #5669, opened 2023-03-10
Notifications: Upgrade to PHP 8.1

mjansenDatabay:


Services/Notifications/classes/Model/ilNotificationLink.php
```diff
@@ -18,30 +16,24 @@
...
 {
     /**
-     * @var string|ilNotificationParameter
+     * @param string|ilNotificationParameter $title
      */
-    private $title;
-    private string $url;
-
-    public function __construct($title, string $url)
+    public function __construct(private $title, private string $url)
```
iszmais: I recommend using Union type inside the constructor paramter instead of docu

Services/Notifications/classes/class.ilObjNotificationAdminGUI.php
```diff
@@ -119,6 +118,7 @@ public function saveOSDSettings(): void
     }
 
     /**
+     * @param array<string, mixed>|null $values
      * @throws ilCtrlException
      */
     protected function getForm(array $values = null): Form
```
iszmais: I recommend Union Type here (array|null)

iszmais (changes requested):
LGTM Just 2 minor missing unions types

Services/Notifications/classes/Model/ilNotificationLink.php
mjansenDatabay: We still should avoid union types in productive code in ILIAS (except for rare cases where native PHP classes/interfaces required to explicitly set a union type).

iszmais: I agree that we should the usage, but if its used it should be typed as such.

Or did i understand you wrong here?

iszmais: An alternative might be to stricten the paramter itsself to make the usage of a unions time unneccessary

iszmais (approved):

PR #5670, opened 2023-03-12
Fix spelling of Git

fabiobittner:
Git, the version control system developed by Linus Torvalds, is named after the British slang word "git" ([according to Wikipedia](https://en.wikipedia.org/wiki/Git#Naming)). Git is not an acronym, hence it's not written in capital letters.


klees:
Judging based on your profile pic you most probably are the youngest ILIAS contributor ever =)

Congrats, thanks and nice to have you!

PR #5673, opened 2023-03-13
DB: Check for RowFormat and Engine during Setup

chfsx:
Thanks @utesche for your finding in https://mantis.ilias.de/view.php?id=35697. I suggest to implement a hard check during setup to have a DB with DYNAMIC default raw format. May you have a look at the state of this PR and perform your tests with it again? Thanks in advance!

### DIFF
--- a/Services/Database/classes/Setup/class.ilDatabaseCreatedObjective.php
+++ b/Services/Database/classes/Setup/class.ilDatabaseCreatedObjective.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 use ILIAS\Setup;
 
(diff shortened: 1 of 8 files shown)


PurHur:
Good check. 

I dont know if the default table check would solve that problem. We had the same issue here in our team many times and the default value was correctly set on DYNAMIC. But the tables itself had another row format because they were old updated installations. 

A proper check for this issue would have to iterate over every table ilias created (or every tables with SHOW TABLES)

chfsx:
@PurHur Yes, indeed, in addition to this check here, one would also have to check the existing tables, as these could of course be historically incorrect.  however, we have no possibility to read out the RawFormat via the ilDBInterface, so I see this as difficult at present. This means that we would first have to extend the interface.

utesche:
@chfsx , hello Fabian,
I checked out exact your commit 23a7b91835570503ab5e68da1b91fb73186ac85f for an already existing release_8-instance and got this failure while updating via setup:

...
_Can create files in '/srv/www/8/release8/data/release8'...                 [OK]
Can create files in '/srv/www/8/release8'...                               [OK]
The database object is initialized....                                     [OK]
No major version is skipped in update....                                  [OK]
No downgrade is performed....                                              [OK]
The ILIAS Version is written to the settings....                           [OK]
The database server has valid settings....                                 [in progress]

 **! [NOTE] Default Row Format is DYNAMIC.**

The database server has valid settings....                                 [FAILED]


 **[ERROR] The default database engine is not set to 'InnoDB'. Please set the default database engine to 'InnoDB to
         proceed'.**_



My settings in DBMS:
- default storage engine : InnoDB
- innodb default row format : dynamic

I will test to install a new instance now with this codebase...
cu.

PR #5679, opened 2023-03-13
Components: Add parameter to links and verify request

mjansenDatabay:
If approved, this should be chery-picked to `release_8` as well.


klees:
Picked to release_8. Thanks a lot! Also, this looks suprisingly simple... Kudos @thibsy

PR #5681, opened 2023-03-14
Removed abstracted DB-Template

chfsx:
Since the abstracted DB-Template is no longer used since ILIAS 7, we propose to remove this completely.


matthiaskunkel:
JourFixe, 20 MAR 2023: We highly appreciate this suggestion and accept the removal of the abstracted DB template with ILIAS 8.

PR #5684, opened 2023-03-14
T&A 0036799 Akkordion content printed

kunzeus:
Issue:
https://mantis.ilias.de/view.php?id=36799

Problem:
The content of a page accordion element in a question gets printed regardless if visible in the preview.

Solution:
Add a media printing rule to the css to prevent unopened accordion panels from printing their content.

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -13368,12 +13368,6 @@ img.ilSmallIcon {
   height: 0px;
   display: none;
 }
-@media print {
-  .ilAccHideContent {
-    height: auto !important;
-    width: auto !important;
-  }
-}
 div.ilc_va_icont_VAccordICont {
   overflow: visible !important;
 }
@@ -20032,5 +20026,9 @@ table.mceToolbar td {
   }
   a[href]:after {
     content: "";
+  }
+  .ilAccHideContent {
+    height: auto !important;
+    width: auto !important;
   }
 }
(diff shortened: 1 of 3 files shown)


kergomard:
Thank you very much @kunzeus . I'm simply unsure if this is the behavior we expect. I would personally think that on print all accordion trays should be opened, as there is no way to open them afterwards. @dsstrassner : Do you have an opinion on this?

kergomard:
@kunzeus : I had a quick feedback from @dsstrassner on another channel. We agree that all trays of the accordions should be open and the content should be shown. Thank you very much again!

kunzeus:
@kergomard the new fix should result in the accordion always being open and printed correctly, at least in my tests in ILIAS 7/8.

kergomard:

I add @Amstutz to the reviewers, mainly with the following question:
This change will affect all accordions in ILIAS. You could be right, that this could be the right behavior, but I'm unsure, if we really want that. We could make this far more specific and only apply it to the accordions in TestQuestions. What do you all think?

Personally, I would make it more specific, as these legacy Accordions are hopefully going to be a thing of the past in the not too far future.

Thanks again and best,
@kergomard

Amstutz:
> This change will affect all accordions in ILIAS. You could be right, that this could be the right behavior, but I'm unsure, if we really want that. We could make this far more specific and only apply it to the accordions in TestQuestions. What do you all think?

Thx, this one is mine. correct. Its a CSS question with some footprint on TA, but mostly global. I agree with the solution. However, can you move the code into: https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/templates/default/less/print.less

Thx.

Amstutz:
thx a lot!

PR #5689, opened 2023-03-14
[FEATURE] UI: introduced league/commonmark as dependency.

thibsy:
Hi folks,

Additionally to the proposed markdown input in #5152 we needed to decide on a makdown-parser for our backend. As already mentioned in the linked PR, we definitely need a serverside markdown-parser, since we certainly want to convert our markdown-text into valid HTML before sending the initial response already.

With this PR I want to introduce the `league/commonmark` composer package, which is a well known and actively maintained markdown-parser. You can find the source-code on [GitHub](https://github.com/thephpleague/commonmark) where it is openly licensed under the BSD-3-Clause.

I recommend to use this library over other ones because of the following reasons:

- The library is built very flexible, we can easily extend the default parser with existing extensions, or we can easily add our own extensions. This will become very helpful in the future for developing our own ILIAS-markdown flavour.
- The parser-result will be a `Document` which can be traversed using the advanced syntax tree (AST). This will already be familiar to ILIAS programmers since the introduction of Rector.
- It's well documented, see https://commonmark.thephpleague.com/2.3/

P.S. You can ignore the Kitchensink changes, since these will already be covered by the referenced PR as well. They are only part of this PR to illustrate how the library will be used.

Please let me know if you have any concerns about this library.

Kind regards,
@thibsy

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -68,7 +68,8 @@
 		"simplesamlphp/simplesamlphp": "^1.19.8",
 		"seld/jsonlint": "^1.8",
 		"ext-xml": "*",
-		"jasig/phpcas": "^1.4"
+		"jasig/phpcas": "^1.4",
+		"league/commonmark": "^2.3"
 	},
 	"require-dev": {
 		"mikey179/vfsstream": "^1.6",
@@ -506,6 +507,16 @@
 			"wrapped_by": null,
 			"last_update": "2022-08-12",
 			"last_update_by": "",
+			"approved-by": "JF",
+			"approved-date": "2021-11-21"
+		},
+		"league/commonmark": {
+			"source": "https://github.com/thephpleague/commonmark",
+			"used_version": "2.3.9",
+			"added_by": "Thibeau Fuhrer <thibeau@sr.solutions>",
+			"wrapped_by": "ILIAS\\UI\\Component\\Input\\Field\\MarkdownRenderer",
+			"last_update": "2023-03-14",
+			"last_update_by": "Thibeau Fuhrer <thibeau@sr.solutions>",
 			"approved-by": "JF",
 			"approved-date": "2021-11-21"
 		},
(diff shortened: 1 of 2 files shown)


matthiaskunkel:
Jour Fixe, 20 MAR 2023: Thibeau notified us about a new dependency. The new markdown parser will become a new dependency in ILIAS 9. According to https://www.gnu.org/licenses/license-list.html.en#ModifiedBSD it is compatible with the GPL 3.0 used by ILIAS.

thibsy:
Hi folks,

Since this PR was already approved by JF and I haven't heard of any strong disregards, I've decided to merge this.

I removed the UI component changes because they will be covered by the previously mentioned PR, so only the new composer package `league/commonmark` will be added by this PR. The "wrapped_by"-comment however, is already pointing to the `MarkdownRenderer` interface which will be introduced later on.

Kind regards
@thibsy

composer.json
```diff
@@ -509,6 +510,16 @@
...
 		},
+		"league/commonmark": {
+			"source": "https://github.com/thephpleague/commonmark",
+			"used_version": "2.3.9",
+			"added_by": "Thibeau Fuhrer <thibeau@sr.solutions>",
+			"wrapped_by": "ILIAS\\UI\\Component\\Input\\Field\\MarkdownRenderer",
+			"last_update": "2023-03-14",
+			"last_update_by": "Thibeau Fuhrer <thibeau@sr.solutions>",
+			"approved-by": "JF",
+			"approved-date": "2021-11-21"
```
tfamula, line 521: @thibsy I guess this date is a copy-paste typo ;)

thibsy: thx @tfamula, you have a sharp eye ;) I fixed this with https://github.com/ILIAS-eLearning/ILIAS/commit/6a9aaed4ce16830cc6fab484cf73a6695d7cd8bf

PR #5690, opened 2023-03-15
T&A: fixes 36939

okaufman:
This will fix https://mantis.ilias.de/view.php?id=36939


kergomard:
Thank you very much @okaufman : This looks right to me. Merged and cherry-picked to trunk

PR #5694, opened 2023-03-16
OrgU: Fix for Mantis #36974

tbongers-cat:
Fix for the admin panel not being shown (should only be omitted for asynchronous calls)


klees:
Picked to trunk.

klees:
Thanks =)

PR #5695, opened 2023-03-16
Fix missing language vars if local lang file is installed Mantis #36972

PurHur:
Hallo,

this fixes https://mantis.ilias.de/view.php?id=36972

Please have a deep look in the change. I have no idea what i did there but it works. 

Greetings
Purhur

### DIFF
--- a/Services/Language/classes/Setup/class.ilSetupLanguage.php
+++ b/Services/Language/classes/Setup/class.ilSetupLanguage.php
@@ -567,7 +567,7 @@ class ilSetupLanguage extends ilLanguage
                 }
             }
     
-            $query = "INSERT INTO lng_modules (module, lang_key, lang_array) VALUES ";
+            $query = "REPLACE INTO lng_modules (module, lang_key, lang_array) VALUES ";
             foreach ($lang_array as $module => $lang_arr) {
                 if ($scope === "local") {
                     $q = "SELECT * FROM lng_modules WHERE " .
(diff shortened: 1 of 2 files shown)


Services/Language/classes/Setup/class.ilSetupLanguage.php
```diff
@@ -567,7 +567,7 @@ protected function insertLanguage($lang_key, $scope = '')
                 }
             }
     
-            $query = "INSERT INTO lng_modules (module, lang_key, lang_array) VALUES ";
+            $query = "REPLACE INTO lng_modules (module, lang_key, lang_array) VALUES ";
```
chfsx, line 570: @PurHur Wouldn't this mean that entries that are no longer in the file are never deleted from the database?

in the original code, isn't the sequence of INSERT and subsequent DELETE simply wrong?

PurHur: I tested that. I deleted "common_actions" in the core files and after a setup.php update run its not there anymore. 
I guess there is another DELETE query that handels that. Problably this one: https://github.com/ILIAS-eLearning/ILIAS/blob/release_7/Services/Language/classes/ilObjLanguageDBAccess.php#L140

I dont know how the whole module works but it looks like it works. 

Edit: The delete query in ilObjLanguageDBAccess just caused the same bug when installing a local lang file in the table. With the same fix deleted lang vars are still deleted somewhere.

katringross:
@PurHur thanks for your help and your PR! Like @chfsx I am not 100% sure, that the deletion still works with your PR. That's why I merged https://github.com/ILIAS-eLearning/ILIAS/pull/5705 from @c-knof today.

PR #5696, opened 2023-03-16
Properly handle input from user defined fields

schmitz-ilias:
This PR makes it so that input from user defined fields is read out using `getInput` instead of directly from post. This is needed for user defined fields (e.g. of a type from a UDFDefinition plugin) to be able to clean up their input after submission without directly writing into post.

In Addition, this PR also fixes a small mismatch in the signature of `ilUDFDefinitionPlugin::getFormPropertyForDefinition`.

As far as I can see this also needs to be merged to trunk, let me know if you want me to handle that, or if you want a second PR.

### DIFF
--- a/Services/Registration/classes/class.ilAccountRegistrationGUI.php
+++ b/Services/Registration/classes/class.ilAccountRegistrationGUI.php
@@ -436,10 +436,11 @@ class ilAccountRegistrationGUI
         $user_defined_fields = ilUserDefinedFields::_getInstance();
         $defs = $user_defined_fields->getRegistrationDefinitions();
         $udf = [];
-        foreach ($_POST as $k => $v) {
-            if (strpos($k, "udf_") === 0) {
-                $f = substr($k, 4);
-                $udf[$f] = $v;
+        foreach ($defs as $definition) {
+            $f = "udf_" . $definition['field_id'];
+            $item = $this->form->getItemByPostVar($f);
+            if ($item && !$item->getDisabled()) {
+                $udf[$definition['field_id']] = $this->form->getInput($f);
             }
         }
         $this->userObj->setUserDefinedData($udf);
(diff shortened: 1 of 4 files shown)


kergomard:
Sorry @schmitz-ilias to let you wait so long! Thank you very much for the PR. I think this is ok for the time being (there is more work to be done there ;-)).

I've one little question/request though:
- Do we really need to make [this parameter](https://github.com/ILIAS-eLearning/ILIAS/pull/5696/files#diff-3571517a52e7d1c39a2cfc20f6ce5de481af357d4f1a07441225790ec5ea93b6R93) nullable? I would really prefer not to. You don't need to pass it and the default value is `null`, so you really have a way to not set it, but if we also make it nullable any change in the future will be so much harder.

Thanks again and best,
@kergomard

schmitz-ilias:
Hi @kergomard, thanks for the feedback, I made the parameter not nullable. @pascalseeland, this included making another tiny change in `ilAccountRegistrationGUI`.
Best, Tim

kergomard (approved):

kergomard:
Thank you very much @schmitz-ilias ! Who ever takes care of this: We will, most probably need to quickly inform the JF before we merge, as we are changing the interface for plugins on a supported system. @pascalseeland : Please check your changes in here and set the JF-lable, if you agree. I can take care of merging and cherry-picking.

schmitz-ilias:
Hi @pascalseeland, would it be possible for you to take a look at this before Monday? The changes in `ilAccountRegistrationGUI` are as far as I can tell pretty minor, and if we can't take this to the JF on Monday we'd have to wait another full month (since there is no JF on May 1st).

kergomard:
I remove myself from the list of assignees to keep my backlog short. Let me know if you expect anything else from me.

alex40724:
@mjansenDatabay Could you have a look at the changes related to Auth/Reg as a coordinator?

mjansenDatabay (approved):
LGTM!

matthiaskunkel:
JourFixe, 12 JUN 2023: Tim notified us about a change in the signature of ilUDFDefinitionPlugin::getFormPropertyForDefinition that should be considered when writing or refactoring a related plugin for ILIAS 8.

PR #5697, opened 2023-03-16
Add setting approval for toast provider

iszmais:
_There is no own Mantis ticket for this Bug but a comment:_
https://mantis.ilias.de/view.php?id=36967#c91762

ATM the Global settings on Notifications are ignored for Toast.
This PR fixes that bug by adding a approval wihc was lost due to the last refactoring of the toasts.

### DIFF
--- a/src/GlobalScreen/Scope/Layout/Provider/PagePart/StandardPagePartProvider.php
+++ b/src/GlobalScreen/Scope/Layout/Provider/PagePart/StandardPagePartProvider.php
@@ -31,6 +31,7 @@ use ILIAS\UI\Component\MainControls\MainBar;
 use ILIAS\UI\Component\MainControls\MetaBar;
 use ILIAS\UI\Component\MainControls\Slate\Combined;
 use ILIAS\UI\Component\Toast\Container as TContainer;
+use ilSetting;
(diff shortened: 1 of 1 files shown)


src/GlobalScreen/Scope/Layout/Provider/PagePart/StandardPagePartProvider.php
```diff
@@ -269,10 +270,13 @@ public function getViewTitle(): string
     /**
      * @inheritDoc
      */
-    public function getToastContainer(): TContainer
+    public function getToastContainer(): ?TContainer
     {
-        $toast_container = $this->ui->factory()->toast()->container();
+        if ((new ilSetting('notifications'))->get('enable_osd') !== '1') {
```
chfsx, line 275: @iszmais I should have stayed firm about the scope Ttoasts`, this shows that they definitely belong to the `Notifications` :-)

See my comment to #5698. Here it is a little less tragic with the ilXY classes, there are (unfortunately) already others. However, I think that we should provide the settings relevant for the GS externally when initialising the service in order to abstract this cleanly. 

I will make you a proposal for this shortly.

PR #5698, opened 2023-03-16
refactor NotificationProvider

iszmais:
https://mantis.ilias.de/view.php?id=36967

The Notification Service has no knowledge about how to handle the persistence of OSD Notification.
This causes that, rn, the only option to remove a OSD Notification is by decay or with a direct interaction of the Notifications presentation, but not by its caused event.
This poses a problem since most Notification have mutliple events which should cause the removal of the notification (e.g. Contact Request canceled, ignored and approved) which mostly can be accessed not only over the UI but also can be caused by link etc.
Therefore the Notification System should oblige the creating sources to handle their Notifications and only should validate their scope of action.

This refactores the OSD Notification System to outsource the persistence of OSD Notification to the providing sources.
- It ensures that the handling source only acts in its own scope.
- It provides a key to for authetification and handling of created Notifications and enabled the deletion and cleanup of them.

_The PRs for the target components will follow soon_

### DIFF
--- a/Services/Notifications/classes/Repository/ilNotificationOSDRepository.php
+++ b/Services/Notifications/classes/Repository/ilNotificationOSDRepository.php
@@ -35,9 +35,6 @@ use ilDBConstants;
  */
 class ilNotificationOSDRepository implements ilNotificationOSDRepositoryInterface
 {
-    private const UNIQUE_TYPES = [
-        'who_is_online'
-    ];
     private ilDBInterface $database;
 
     public function __construct(?ilDBInterface $database = null)
@@ -65,10 +62,6 @@ class ilNotificationOSDRepository implements ilNotificationOSDRepositoryInterfac
             $base->getType()
         );
 
-        if (in_array($notification->getType(), self::UNIQUE_TYPES)) {
-            $this->deleteOSDNotificationByUserAndType($user_id, $notification->getType());
-        }
-
         $affected = $this->database->insert(
             ilNotificationSetupHelper::$tbl_notification_osd_handler,
             [
@@ -79,6 +72,7 @@ class ilNotificationOSDRepository implements ilNotificationOSDRepositoryInterfac
                 'visible_for' => [ilDBConstants::T_INTEGER, $notification->getVisibleFor()],
                 'type' => [ilDBConstants::T_TEXT, $notification->getType()],
                 'time_added' => [ilDBConstants::T_INTEGER, $notification->getTimeAdded()],
+                'provider_key' => [ilDBConstants::T_TEXT, $base->getProviderKey()]
             ]
         );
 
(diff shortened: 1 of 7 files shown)


Services/Notifications/classes/Repository/ilNotificationOSDRepository.php
```diff
@@ -148,17 +142,20 @@ public function deleteOSDNotificationById(int $id): bool
...
+    public function deleteOSDNotificationByProviderKey(string $type, string $provider_key, int $user_id = 0): bool
     {
-        $query = 'DELETE FROM ' . ilNotificationSetupHelper::$tbl_notification_osd_handler . ' WHERE usr_id = %s AND type = %s';
-        $this->database->manipulateF(
-            $query,
-            [ilDBConstants::T_INTEGER, ilDBConstants::T_TEXT],
-            [$user_id, $type]
-        );
+        $query = 'DELETE FROM ' . ilNotificationSetupHelper::$tbl_notification_osd_handler . ' WHERE type = %s AND provider_key = %s';
+        $keys = [ilDBConstants::T_TEXT, ilDBConstants::T_TEXT];
```
mjansenDatabay, line 148: Should be named `$types`.

chfsx:
Hi @iszmais 

I am not a fan of dependencies within src/GlobalScreen ([AbstractNotificationProvider.php](https://github.com/ILIAS-eLearning/ILIAS/pull/5698/files#diff-75372892a555a49c3c474b0c3e210bba413828e23e3e435e8010f13996db5e00) ), I try not to use ilXY classes within the GlobalScreen if possible. What I also don't quite understand is why this change is in Scope Notifiactions and not in Toasts?

I see that I have already accepted such ilXY dependencies in other places, unfortunately I overlooked this, e.g.: https://github.com/ILIAS-eLearning/ILIAS/blob/90eb02c21c20578ba58ffdd530acbfdc2bb0c7ee/src/GlobalScreen/Scope/Toast/Provider/AbstractToastProvider.php#L56

From my point of view, we should take a slightly different approach here. let's compare the whole thing with the scope of the notifications, where on the one hand an `IdentificationInterface` is required for each notification, obtained from the respective `IdentificationFactory` factory. Closures are also required for opening and closing, see https://github.com/ILIAS-eLearning/ILIAS/blob/22dab400bf1839c4d146dedc590cb6f1ae7757c2/src/GlobalScreen/Scope/Notification/Factory/isStandardItem.php#L26.

This has the advantage that the GS can send the actions to a general endpoint during rendering: https://github.com/ILIAS-eLearning/ILIAS/blob/7b801dffd796518f7223afa0fc4327814e7024ac/src/GlobalScreen/Scope/Notification/Collector/Renderer/AdministrativeNotificationRenderer.php#L50

Such actions (the callables) are then very easy to implement and are executed when an action is clicked (e.g. for administrative notifications, the click on the X to dismiss the notification). https://github.com/ILIAS-eLearning/ILIAS/blob/22dab400bf1839c4d146dedc590cb6f1ae7757c2/Services/AdministrativeNotification/classes/GlobalScreen/ADNProvider.php#L82

For this reason it is important that the GlobalScreen does not simply work with UI components but wraps its `items` per scope. In the case of notifications, this has already been done only halfway, but in my view it has been omitted completely in the case of toasts.

I would be happy to help refactor the toasts in this direction, but I am hesitant to merge a quick solution with additional dependencies.

From my point of view, an elaborate solution in the GS would also simplify the code on the part of services/notifications.

PR #5700, opened 2023-03-16
Use refactored notification provider functions

iszmais:
**DEPENDS ON https://github.com/ILIAS-eLearning/ILIAS/pull/5698**

With the merge of the depending PR the usage of OSDNotifcation obliges the Service to handle the cleanup of its own notifications. Therefore the refactored Notification System provider a new feature trough the NotifcationProvider from wich this service heritages, to delete your own Notifcation with a safe named key.

You can set that key within your Notification config with the function `setProviderKey()`

If an event causes the Notification to be irelevant it can be deleted by that key with a new function in your NotificationProvider Inehitance `removeOSDNotificationsByProviderKey()` wich optional recieves a user ID to remove this OSD for on user. Otherwise its removed for all users.

Further you can cleanup all old notifications with the function `deleteStaleNotifications()`. I encourage you to do so if your service produces a lot of notifications.

### DIFF
--- a/Modules/Chatroom/GlobalScreen/ChatInvitationNotificationProvider.php
+++ b/Modules/Chatroom/GlobalScreen/ChatInvitationNotificationProvider.php
@@ -29,8 +29,8 @@ use ILIAS\Notifications\ilNotificationOSDHandler;
 
 class ChatInvitationNotificationProvider extends AbstractNotificationProvider
 {
-    public const MUTED_UNTIL_PREFERENCE_KEY = 'chatinv_nc_muted_until';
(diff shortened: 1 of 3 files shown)


iszmais:
@mjansenDatabay This PR already covers the obvious parts wich can be handler by the provider and already sets a provider key.
Since im not familiar with the component i would appreciate if you could add the removal/cleanup calls on the required cases.
(Or Feel free to present your own approach on that).

mjansenDatabay:
See commends in https://github.com/ILIAS-eLearning/ILIAS/pull/5702

PR #5702, opened 2023-03-16
Use refactored notification provider functions

iszmais:
**DEPENDS ON https://github.com/ILIAS-eLearning/ILIAS/pull/5698**

With the merge of the depending PR the usage of OSDNotifcation obliges the Service to handle the cleanup of its own notifications. Therefore the refactored Notification System provider a new feature trough the NotifcationProvider from wich this service heritages, to delete your own Notifcation with a safe named key.

You can set that key within your Notification config with the function `setProviderKey()`

If an event causes the Notification to be irelevant it can be deleted by that key with a new function in your NotificationProvider Inehitance `removeOSDNotificationsByProviderKey()` wich optional recieves a user ID to remove this OSD for on user. Otherwise its removed for all users.

Further you can cleanup all old notifications with the function `deleteStaleNotifications()`. I encourage you to do so if your service produces a lot of notifications.

### DIFF
--- a/Services/Badge/classes/class.ilBadgeHandler.php
+++ b/Services/Badge/classes/class.ilBadgeHandler.php
@@ -27,6 +27,8 @@ use ILIAS\Badge\GlobalScreen\BadgeNotificationProvider;
  */
 class ilBadgeHandler
 {
+    public const BADGES = 'badges';
+
     protected ilComponentRepository $component_repository;
     protected ilDBInterface $db;
     protected ilTree $tree;
@@ -480,6 +482,8 @@ class ilBadgeHandler
         array $a_user_map,
         int $a_parent_ref_id = null
     ): void {
+        $provider = new BadgeNotificationProvider();
+
         $badges = [];
 
         foreach ($a_user_map as $user_id => $badge_ids) {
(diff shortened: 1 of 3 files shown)


iszmais:
@mjansenDatabay   This PR already covers the obvious parts wich can be handler by the provider and already sets a provider key.
Since im not familiar with the component i would appreciate if you could add the removal/cleanup calls on the required cases.
(Or Feel free to present your own approach on that).

Services/Badge/classes/class.ilBadgeHandler.php
```diff
@@ -541,14 +545,14 @@ public function sendNotification(
                     $url = new ilNotificationLink(new ilNotificationParameter('badge_notification_badges_goto', [], 'badge'), $url);
                     $osd_params = ["badge_list" => implode(", ", $user_badges)];
 
-                    $notification = new ilNotificationConfig(BadgeNotificationProvider::NOTIFICATION_TYPE);
+                    $notification = $provider->getNotificationConfig();
```
mjansenDatabay, line 548: What is unclear to me is the coupling to the notification provider here. Conceptually we know, that some (**not** all) of the `Toast` elements collected and handled by the `GS` are based on the `OSD` notification channel of `Services/Notifications`. So depending on the concrete `Toast Provider`, there might (**not** must) be a dependency to the `OSD` notifications from the direction of the `Toast` provider. But IMO there should be no coupling to the `GS Notifications` here.

Services/Badge/classes/class.ilBadgeHandler.php
```diff
@@ -541,14 +545,14 @@ public function sendNotification(
...
                     $notification->setTitleVar("badge_notification_subject", [], "badge");
                     $notification->setShortDescriptionVar("badge_notification_osd", $osd_params, "badge");
                     $notification->setLongDescriptionVar("");
                     $notification->setLinks([$url]);
                     $notification->setIconPath(ilUtil::getImagePath('icon_bdga.svg'));
-                    $notification->setValidForSeconds(ilNotificationConfig::TTL_SHORT);
-                    $notification->setVisibleForSeconds(ilNotificationConfig::DEFAULT_TTS);
+                    $notification->setProviderKey(self::BADGES);
```
mjansenDatabay, line 554: IMO this does not help. The notification emitted through `Services/Notifications` here must be provided with a **unique** `Identification`. Instead, you pass a static `badges` as key. How should the `Services/Badges` in turn emit a signal to `Services/Notifications` later, that the notification sent in the past is no longer valid and should be removed?

If we just want to rely on the UI part, which means, we only care about if the receiver of a `Toast` interacts with the provided actions (defined in the concrete `Toast Provider`), we might end up with a similar mechanism like here: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Services/Contact/classes/Provider/ContactNotificationProvider.php#L136 , where an interaction results in the execution of callable, which in turn could result in removing the `OSD` entry.

mjansenDatabay (changes requested):

PR #5706, opened 2023-03-19
Fixes T&A #36791 - Formula Question: Switches the ranges in case of malformed user input

jcopado:
…let better the machine do the work.


jcopado:
Solved in ILIAS 7.19

kergomard:
Thank you very much @jcopado .
I think though, that we should not implicitly change the user input. We should check this in checkInput and return an error on validation.
Could you please adapt the PR accrodingly?

Thanks again and best,
@kergomard

jcopado:
will try in the following weeks

jcopado:
Hi Stefan, It seems to be that filters were preventing variables and results to be properly validated. This does not prevents the crash when trying to access the preview in a question with wrong ranges, as reported in #36791, so I would say this is a work in progress PR.

dsstrassner:
TechSquad, 13.04.23: @nhaagen will look into it and give feedback next week.

nhaagen:
The filters fix https://mantis.ilias.de/view.php?id=15673, removing results/variables will lead to error otherwise.
It is true, though, that the filters are too greedy, i.e. - simply wrong: https://github.com/ILIAS-eLearning/ILIAS/pull/5802

kergomard:
Thank you very much @jcopado for looking into this.
As far as I can see @nhaagen 's made it impossible to save variables with inverted ranges. I don't think we should take care of the questions that already have inverted ranges. So I close this.
Best,
@kergomard

PR #5711, opened 2023-03-21
UI: Add an optional name to Input/Field and Input/Form

tbongers-cat:
_(Proposal for a request by @klees)_

Currently, all **Input/Field** components of the UI-Framework get a generated name like `form_input_<count>`. For future uses, we would like a new option to set a dedicated name for an input which makes it addressable, but is still unique. In addition, the "path" to the input should be visible by adding the names of its parent inputs (e.g. groups or sections) to its name. 

With this PR, all input fields get a slightly shorter name like `input_<count>`. If they are contained inside another input field, the path is added to the name, e.g. `input_1/input_2`. To set individual names per input, an optional parameter `$dedicated_name` was added to the methods of `Input/Field/Factory`. If this parameter is set, the name of the input is changed from `input_<count>` to `<dedicated_name>_<count>`, resulting in a name like `address_2/street_5`.
_(--> also see Revisions below for changes)_

**Input/Container/Form** components currently don't have a set name attribute. With the additional parameter `$dedicated_name` in `Input/Container/Form/Factory`, a name can now be set per form. If no dedicated name is set, the name attribute will still be omitted. 
Since each form contains a group at top level, this group is named the same as the form and is included in the "path" of all contained inputs. If no name is set on the form, the group will just be called `form`.

Looking forward to your input on this!

### DIFF
--- a/src/UI/Component/Input/Container/Form/Factory.php
+++ b/src/UI/Component/Input/Container/Form/Factory.php
@@ -62,8 +62,7 @@ interface Factory
      * ---
      * @param    string $post_url
      * @param    array<mixed,\ILIAS\UI\Component\Input\Field\FormInput>    $inputs
-     * @param    string|null $dedicated_name
      * @return    \ILIAS\UI\Component\Input\Container\Form\Standard
      */
-    public function standard(string $post_url, array $inputs, ?string $dedicated_name): Standard;
+    public function standard(string $post_url, array $inputs): Standard;
 }
--- a/src/UI/Component/Input/Container/Form/Factory.php
+++ b/src/UI/Component/Input/Container/Form/Factory.php
@@ -62,7 +62,8 @@ interface Factory
      * ---
      * @param    string $post_url
      * @param    array<mixed,\ILIAS\UI\Component\Input\Field\FormInput>    $inputs
+     * @param    string|null $dedicated_name
      * @return    \ILIAS\UI\Component\Input\Container\Form\Standard
      */
-    public function standard(string $post_url, array $inputs): Standard;
+    public function standard(string $post_url, array $inputs, ?string $dedicated_name): Standard;
 }
--- a/src/UI/Component/Input/Container/Form/Form.php
+++ b/src/UI/Component/Input/Container/Form/Form.php
@@ -67,4 +67,19 @@ interface Form extends Component
      * @return null|string
      */
     public function getError(): ?string;
+
+    /**
+     * Sets an optional dedicated name for this form which adds a NAME attribute
+     * to the form's HTML (otherwise no NAME attribute is set).
+     *
+     * The dedicated name is inherited by all child inputs of the form
+     * by setting it as dedicated name for the top level group that is generated
+     * for every form.
+     *
+     * Please see the description of withDedicatedName() on Field/Input for more details.
+     *
+     * @param string $dedicated_name
+     * @return $this
+     */
+    public function withDedicatedName(string $dedicated_name): self;
 }
--- a/src/UI/Component/Input/Field/Factory.php
+++ b/src/UI/Component/Input/Field/Factory.php
@@ -56,10 +56,9 @@ interface Factory
      * ---
      * @param string      $label
      * @param string|null $byline
-     * @param string|null $dedicated_name
      * @return    \ILIAS\UI\Component\Input\Field\Text
      */
-    public function text(string $label, ?string $byline = null, ?string $dedicated_name = null): Text;
+    public function text(string $label, ?string $byline = null): Text;
 
     /**
      * ---
@@ -83,10 +82,9 @@ interface Factory
      * ---
      * @param string      $label
      * @param string|null $byline
-     * @param string|null $dedicated_name
      * @return    \ILIAS\UI\Component\Input\Field\Numeric
      */
-    public function numeric(string $label, ?string $byline = null, ?string $dedicated_name = null): Numeric;
+    public function numeric(string $label, ?string $byline = null): Numeric;
 
     /**
      * ---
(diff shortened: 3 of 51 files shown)


tbongers-cat:
**Revision 1**: `$dedicated_name` has been moved from the constructor to `withDedicatedName()` method
(More changes coming soon, as discussed with @klees)

tbongers-cat:
**Revision 2:** A counter will only be added to unnamed inputs (`input_<counter>`) and for inputs with a dedicated name that is used more than once. This way, names like `form/address/street` can be generated.

tbongers-cat:
~~**Revision 3**: Only inputs with dedicated names are added to the "path" of sub-inputs, so that for example `section/input_7/firstname` is reduced to `section/firstname`. Also removed the standard name `form` from the top level group if no dedicated name is set for the form.~~
*(has been reverted, see comments below)*

Amstutz:
Hi @tbongers-cat 

Thx a lot for this proposal! This seems to be something, some devs waited for, as you can see by the reactions by the PR ;-).  Note that this will be presented to the next JF, since it contains a change to the public interface.

Question:
- [x] Use Case: Can you give us a use case, where this would be needed? 
- [x] Path Doc: Can you describe in the Doc Comment of the public interface method, how the "path" mechanism works? This seems not obvious by only looking at the public interface aspect of the code.
- [x] Revision 3: Why this change? This smells a bit like unpredictable magic to use. These could be more transparent, maybe, if described in somewhat more detail.

Probably change, or indicate why not:
- [x] Form::withDedicatedName(): Is missing on the public interface, correct? If so, add.


Please Change:
- [x] Example: Please document this useful new feature for inputs by providing an example documenting it. This could also answer the question for a use case above.
- [x] Unrelated Changes: There are a huge number of changes not directly related to the improvement at hand. While we highly appreciated (some even resolve small issues, thx a lot!), please put them in a separate PR, for the following reasons:
 - The vast amount of unrelated changes makes the PR rather hard to review.
 - Those changes also make it somewhat hard for interested other devs trying to have a brief look at it for the necessary JF discussion
 - The good changes would be lost in case the proposal would not find its way to the trunk.

IMO, this is ready to be presented to the JF. We are looking forward to it.

@Amstutz and @thibsy

klees:
Hi @Amstutz,

I think it's on me to answer your question regarding use cases, as I ordered this change at @tbongers-cat.

I want to start with a quick recap on the reason why we currently do not have names on the Input Fields in the UI-Framework. There are two main reasons:

* Names are bad for composability. If we'd conceptualize names as the "name" attribute of HTMLs input tags, it would be hard to combine huge forms from small parts. What should happen if two inputs had the same name? How could we coordinate naming over the complete system? Both questions have no good answer.
* Names in HTML imply names in `$_POST`. This is another problem for duplicated names, but also implies a specific habit of processing forms: Looking into `$_POST`, mostly, and start processing there. This, again, is bad for composability, as that style of processing does not compose very well, but also for security.

Hence, currently HTMLs "name" attribute is filled with a more or less random value and forms can be built from smaller parts. This, IMHO, works remarkably well for forms that are processed via HTML and POST.

Some time after we introduced the Form Container for these ordinary forms, we added the Filter Container, which uses the Input Fields from the UI Framework, but in the context of a filter. Hence, values are not passed to the server via POST, but instead added as query parameters to a GET request. This lead to the first problems with the naming of the fields: Random names don't really cut it, as they do not lead to intelligible URLs, which is sad. So, **first use case for named fields are the filters**.

The **second use case, then, are of course the View Controls**. They also should not POST their values, but instead add it to the GET query and should create intelligle URLs.

Another problem seemed unrelated at first. For the setup, the ILIAS components need to provide a direction how some configuration in the config.json of the setup map to a config object of the component. Currently, this direction is provided as a `Transformation` from the `Refinery`. These `Transformation`s currently are not self-documenting, so something like the list of available configurations and their meaning, as we have in the [readme of the setup](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/setup/README.md#about-the-config-file) is maintained seperatly. Also, the `Transformations` only can give little hint on which field in the json might be missing or why some input is not acceptible and definitely cannot provide a guided creation of the json or something alike.

First I thought we would want to add according information to the `Transformation` of the `Refinery`, but then I realized that it is basically the Input Fields `byline` that I'm going to reimplement. I then also realized that, due to randomized names, our forms are also hard to impossible to be used by machines (via JSON, probably). So **third use case is: make forms be fillable by JSON**. This will immediately help the setup, but probably also opens up some path to a system that can be automated better.

So, with this PR, we try to introduce names that can be displayed to users, without damaging the general composability of the forms and while still discouraging access to `$_POST` strongly.

I hope this makes some sense.

Best wishes!

tbongers-cat:
Hi @Amstutz and @thibsy ,

thanks a lot for your feedback. In addition to the use cases detailed above by @klees, I have now made the following changes:
- Added the missing interface for `withDedicatedName()` to Form and extended the description for both new methods to make their purpose clearer
- Added examples to Form/Standard and Input/Field
- After discussing it with @klees, I have reverted the changes of Revision 3 for now, so that all inputs are once again included in the path
- Removed all unrelated changes (will add them to another PR shortly)

Please let me know if you have any further questions :)

tbongers-cat:
PS: Just reverted a silly change I made to dynamic inputs, sorry!

Amstutz:
Is this ready for another go for a review?

tbongers-cat:
> Is this ready for another go for a review?

I have just rebased and resolved the conflict, so this should be ready for review now. The failed tests are caused by the omitted copyright notice in the new example files.

Amstutz:
thx a lot @tbongers-cat for adding this improvement here!

PR #5716, opened 2023-03-22
fixed local lang installation for setup aswell

c-knof:


klees:
Hi @c-knof,

thanks for the PR. Please have a look into [our guidelines for contributors](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/contributing.md#rules-for-contributors) to help process this. Specifically: Is there an issue in mantis that is fixed by this? If so: Please reference.

Thanks!

PR #5720, opened 2023-03-23
Cron: Allow local auth. only (without password)

mjansenDatabay:
This PR changes the CLI cron script of ILIAS by ...

- ... only trying a local authentication (so no more LDAP auth. etc.) - it is more a `Lookup` and not an authentication anymore ...
- ... therefore don't requiring a raw password to be provided as an argument in the PHP CLI process anymore
- ... using `Symfony Console` (similar to the setup). IMO this is a good foundation for possible future ehancements in the `Cron` component of ILIAS (e.g. introduce a command to run a specific job only etc.).

Depending on the JF decision I would like to integrate this in `release_8` as well.

### DIFF
--- a/Services/Cron/classes/CLI/Commands/RunActiveJobsCommand.php
+++ b/Services/Cron/classes/CLI/Commands/RunActiveJobsCommand.php
@@ -1,0 +1,84 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\Cron\CLI\Commands;
+
+use Symfony\Component\Console\Command\Command;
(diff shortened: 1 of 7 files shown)


Services/Cron/classes/CLI/Commands/RunActiveJobsCommand.php
```diff
@@ -0,0 +1,84 @@
...
+
+    /** @var OutputInterface|StyleInterface */
+    private $style;
+
+    protected function configure(): void
+    {
+        $this->setDescription('Runs cron jobs depending on the respective schedule');
+
+        $this->addArgument('user', InputArgument::REQUIRED, 'The ILIAS user the script is executed with');
+        $this->addArgument('client_id', InputArgument::REQUIRED, 'The ILIAS client_id');
```
PurHur, line 45: Is it really required anymore? 
There is only one folder allowed in the inner data dir and this is always the correct client_id.

mjansenDatabay: You are right, but as long as the `client_id` is scattered in our code base, we should just keep it for now.

There is already awareness in the community there there must be a project `Completely Remove the ClientId from the ILIAS Code Base`.

But as always: We need to find somebody who pushes this.

fsesterh:
We discussed this PR in the Security Group and support the changes.

matthiaskunkel:
Jour Fixe, 03 APR 2023 : We highly appreciate this suggestion and accept the PR for release_8 and trunk. We recommend to define a rule that cronjob must not check permissions to allow a cron user with only a little amount of permissions. Michael will check if the cron user could get the Anonymous role for this purpose. Please add related documentation at least to the cron README in GitHub and the cron chapter within the LM "Installation and Maintenance".

PR #5722, opened 2023-03-23
Folder: hide content tab when preconditions are not met (37015)

schmitz-ilias:
This PR fixes [37015](https://mantis.ilias.de/view.php?id=37015).


alex40724:
Great, thanks!

PR #5723, opened 2023-03-23
[FIX] UI: OptionalGroup now is only required if explicitly set.

thibsy:
Hi folks,

I noticed that `OptionalGroup`s are required if any of it's sub-inputs are, like in normal `Group`s.

This behaviour is IMO not quite as expected, because the sub-inputs should only be required **after** the `OptionalGroup` was checked, or if set explicitly (why-so-ever one would do this).

This PR therefore adds:
- a fix for this issue by overriding `Group::isRequired()`
- a fix for yet another issue, where the label of the optional group was not pointing to it's checkbox input.
- an example for the expected behaviour of required `OptionalGroup`s.
- a unit-test for the expected behaviour of required `OptionalGroup`s.

If you feel different about this behaviour let me know.

Kind regards
@thibsy

### DIFF
--- a/src/UI/Implementation/Component/Input/Field/OptionalGroup.php
+++ b/src/UI/Implementation/Component/Input/Field/OptionalGroup.php
@@ -51,6 +51,11 @@ class OptionalGroup extends Group implements Field\OptionalGroup
         return Input::withRequired($is_required);
     }
 
+    public function isRequired(): bool
+    {
+        return $this->is_required;
+    }
+
     /**
      * @inheritdoc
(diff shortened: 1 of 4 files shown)


Amstutz:
Hi @thibsy 

Thx a lot for the fixes. Special thx for adding the example and the test!

Note that in future it would be great to have Issue tickets for fixes for issues (especially for productive versions). IMO this one here is well enough documented and risk free that I would let it slide by as "improvement". For the "Bugfix" label linked issue ticket is required.

Thx a lot for the fix.

PR #5725, opened 2023-03-24
Improvement/8/refactoring toasts

chfsx:
@iszmais and @mjansenDatabay, As discussed, I have now modified the scope `Toast` in the GlobalScreen as follows:
- The scope follows the same pattern as all other scopes, namely that `Items` are wrapped and not directly worked with the corresponding UI components.
- The example has been updated, you can see there how, for example, the NotificationService can now offer a corresponding provider or an abstract provider for all consuming components of the NotificationService. 
- The toasts require a new event when displaying the action. From my point of view, this would have to be solved "correctly" in the UI component, e.g. with `withOnClose` ([here](https://github.com/ILIAS-eLearning/ILIAS/blob/a760829c6fc509a7885f9bfda7d5785032c6076f/src/UI/Component/Closable.php#L30)) and `withOnLoad` ([here](https://github.com/ILIAS-eLearning/ILIAS/blob/0e5241f8f4aa14443033a73ac2a9f1a5f2cf8145/src/UI/Component/Onloadable.php#L30)). However, this is not something I can offer in the course of this refactoring.

I removed the old ToastProviders for the moment, since they must be implemented again as mentioned above.

@iszmais you see an example here: https://github.com/srsolutionsag/ILIAS/blob/improvement/8/refactoring-toasts/src/GlobalScreen/Scope/Toast/README.md


chfsx:
There are some syntax problem I will address in a second --> done

iszmais:
Thank you very much for your effort @chfsx , especially for the perfect timing :smile: 

This look good to me, I just have 2 minor aspects i would reconsider:
- I implemented the toast initialiy with a `onshow` action wich was explicitly denied after a few discussions with responsibles. I guess it makes sense to reopen this discussion due to the major change we now doing, but I would open that to a wider group first before adding this to the toast to respect the initial effort put in that decision. This also applies to `onVanish`.
- The toast already has some kind of `onClose` action which is used for all interactions on the toast ATM. But this can be reused when we change the link handling to the target locations. Therefore there is no need in the global screen for doing that IMO.

Otherwise im fine with that PR.

Greetings,
Ingmar Szmais

chfsx:
I updated the PR accordingly after our meeting:
- There is currently no new Event in toast.js, please make a separate PR for the events onShown, onVanished and onClosed. The discription of the UI\Toast should provide a howto, which events should trigger what kind of action. I can adopt these rules then in the GlobalScreen part 
- The GS already could handle those events, see https://github.com/ILIAS-eLearning/ILIAS/blob/636b6a256385ac8e81a626445697cda7865d3349/src/GlobalScreen/Scope/Toast/Factory/isStandardItem.php

PR #5727, opened 2023-03-27
UI presentation table - fix overlap of box and header into action btn

catenglaender:
# Issues

Mantis: https://mantis.ilias.de/view.php?id=36531

The action button or dropdown is overlapped by other elements of the presentation table...

...when there is no sub-headline:

image

...when the headline is very long:

image

This happens because the action div is removed from the document flow with `position: absolute` so other elements cannot respect its space.

While I am guilty of having worked like this in the past to stick dropdowns and buttons to top right corners of containers with this method, I currently believe that placing these elements within the flow of a flex- or grid-layout is the more stable solution as nothing can ever overlap into the action's place. I would like to propose a refactoring of the presentation table as a flex-layout for Ilias 9 for consideration (maybe the CSS squad is a good place to discuss this?).

For now, this PR contains a less invasive temporary fix.

# Fix

This fix contains...
* a min-height for the header to make sure the box always starts below the action
* a limited width if the action is not empty, so header text cannot run into the action

image

Note that for the `:not(:empty)`selector to work, the action div has to be written as one line with no spaces in the html template.

### DIFF
--- a/src/UI/templates/default/Table/table.less
+++ b/src/UI/templates/default/Table/table.less
@@ -69,9 +69,15 @@
 
     .il-table-presentation-actions {
         margin: 5px 0;
+        // if there is an action, row header text should not run over action button
+        &:not(:empty) + .il-table-presentation-row-header {
+            max-width: calc(100% - 120px);
+        }
     }
 
     .il-table-presentation-row-header {
+        // if there is no sub headline, a min-height is needed so further fields box doesn't overlap with action button - Mantis 36531
+        min-height: @font-size-h4 * 2;
         h4 {
             font-size: @font-size-h4;
             font-weight: normal;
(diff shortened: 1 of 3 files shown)


Amstutz:
Hi @catenglaender 

Thx a lot for looking into this. 

I completely agree with your point concerning flex for the future for such caeses. I would warmly welcome a respective PR for 9 if somebody would likes to tackle this. 

For the fix for 7 and 8 (which I would also push to sacss/trunk ftm), I have one thing:

- [ ] REM vs Pixel: Imo this value hiere should be in REM, right? otherwise we run into issue, if somebody increases the font size a lot (e.g. a11y reasons): `max-width: calc(100% - 120px);`

thx a lot.

catenglaender:
@Amstutz: Done :heavy_check_mark:  the area reserved for the button is now calculated with rem instead of pixels. Thanks for the feedback :slightly_smiling_face:

Amstutz:
thx a lot!

thibsy (commented):
test

PR #5729, opened 2023-03-27
Cron: Replace `int` schedule type with `Enum`

mjansenDatabay:
Dear colleagues,

with this PR I would like to switch the PHP type representing the `Schedule Type` of a cron job from `int` to a backed enum. The usages of the public constants defined in `ilCronJob` have been marked as `deprecated` with ILIAS 8.

I changed all (core) `ilCronJob` implementations. If you have any objections, please provide feedback within the next two weeks.
Important: I did not change the license header of your files if this is already wrong.

Best regards,
@mjansenDatabay

### DIFF
--- a/Modules/BookingManager/classes/class.ilBookCronNotification.php
+++ b/Modules/BookingManager/classes/class.ilBookCronNotification.php
@@ -16,6 +16,8 @@
  *
  *********************************************************************/
 
+use ILIAS\Cron\Schedule\CronJobScheduleType;
+
 /**
  * Cron for booking manager notification
  * @author Alexander Killing <killing@leifos.de>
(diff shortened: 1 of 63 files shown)


alex40724 (approved):

mbecker-databay (approved):

smeyer-ilias (approved):

chfsx (approved):
@mjansenDatabay  Thank you for this change, in general I very much welcome the change to an enum. But do I understand correctly that this is a breaking change, e.g. for cron plugins? This means that a cron plugin cannot be offered for ILIAS 8 and 9 at the same time? For other reasons, we probably no longer have the possibility to offer plugins for several releases, so this would not be a blocker from my point of view, but perhaps it should be pointed out somewhere.

mjansenDatabay:
@chfsx Yes, this is a BC and a reason why I'll put this on the JF agenda (now :-)).

matthiaskunkel:
Jour Fixe 03 APR 2023:  Michael notified us about a breaking change in cronjobs that come with this PR. We highly appreciate his suggestion and accept the PR for trunk.

mjansenDatabay:
@klees / @kergomard / @Uwe-Kohnle : Do you have any objections?

kergomard (approved):
Thank you very much @mjansenDatabay ! Changes in ilUser are fine with me.

mjansenDatabay:
@Uwe-Kohnle / @klees I will merge this now

PR #5518, opened 2023-02-01
Add "Whos is online" toast provider for GS

iszmais:
Due to changes made inside the load of toast an provider interface was established to porvide the initial toasts without another request: https://github.com/ILIAS-eLearning/ILIAS/pull/5121

Therefore every provider of toast is obliged to provide a new toast Provider if his service want to add toast to the page.
ATM this only applies to initial toast but will probably be extended to all toast in near future.

This PR is an Proposal for toast inside your Service. Feel free to use and manipulate this template to your fitting and provide a PR for Trunk.
If you dont want to provide Toast with your Service feel free to cloase this PR.

PR #5730, opened 2023-03-27
Add Who is online Toast provider and removed notifcations

iszmais:
This is a further improvement to https://github.com/ILIAS-eLearning/ILIAS/pull/5518.
Since WIO Notifications are atm recreated and removed on every page reload they create kind of a mess within the notification Center handling, due to the fact they have a special mode onyl to handle their individual behavior.

Since Those behaviour doesnt match with the common ILIAS Notification and due to the improvement of the new ILIAS Toast feature this PR Proposes to remove the WIO feature from the Notification and make its presentation toast only.

That improves not only the Notification System but also refactores the ancient Fethcer of WIO Notification an put the into the more fitting, non-persisten Toast scope.

### DIFF
--- a/Services/Awareness/classes/Provider/AwarenessToastProvider.php
+++ b/Services/Awareness/classes/Provider/AwarenessToastProvider.php
@@ -1,0 +1,113 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\Notifications\Provider;
+
+use ilDateTime;
+use ILIAS\Awareness\User\Collector;
+use ILIAS\GlobalScreen\Scope\Toast\Provider\AbstractToastProvider;
+use ILIAS\UI\Implementation\Component\Symbol\Icon\Standard;
+use ILIAS\UI\Implementation\Component\Toast\Toast;
+use ilSetting;
+use ilUserUtil;
+
+/**
+ * @author Ingmar Szmais <iszmais@databay.de>
+ */
+class AwarenessToastProvider extends AbstractToastProvider
+{
+    private const PROVIDER_KEY = 'awareness';
+
+    public const MAX_ONLINE_USER_COUNT = 20;
+
+    /**
+     * @inheritDoc
+     */
+    public function getToasts(): array
+    {
+        $this->dic->language()->loadLanguageModule('awrn');
+        $settings = new ilSetting('awrn');
+
+        $toasts = [];
+        if (
+            $settings->get('use_osd', '0') !== '1' ||
+            0 === $this->dic->user()->getId() ||
+            $this->dic->user()->isAnonymous()
+        ) {
+            return $toasts;
+        }
+
+        $users = Collector::getOnlineUsers();
+        unset($users[$this->dic->user()->getId()], $users[ANONYMOUS_USER_ID]);
+        $users = array_slice($users, 0, self::MAX_ONLINE_USER_COUNT, true);
+
+        $new_user_ids = [];
+        foreach ($users as $id => $user) {
+            $time = (new ilDateTime($user['last_login'], IL_CAL_DATETIME, $this->dic->user()->getTimeZone()))->getUnixTime();
+            if ($time >= (time() - ($this->dic->http()->request()->getQueryParams()['max_age'] ?? 0))) {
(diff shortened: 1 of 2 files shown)


iszmais:
Hello @alex40724 ,

Im not aware of your scedule but i would highly appriate if this could go into the ILIAS 8 release to start the new Version with the improved user view and a clean Database.
Thanks in Advance!

Greetings,
@iszmais

alex40724:
Hi @iszmais ,
thanks for the PR. I have difficulties to understand the full impact. ILIAS 7 was old OSD only, right? With your changes from March 2022 this became notification center + toast? And now it should be toast only, right? I currently checked the ILIAS 8 test installation and neither got a toast or a notification, even if the users newly appeared in the who-is-online list. I think we should get JourFixe confirmation on the expected behaviour here, I set the label.

mjansenDatabay:
Hi @alex40724,

the reason why **NO** toasts were shown at all was a necessary (because of https://mantis.ilias.de/view.php?id=36967 and an issue described by Fabian Kruse in https://mantis.ilias.de/view.php?id=36967#c91762) change implemented with #5725 (by @chfsx). 

Afterwards, @iszmais made the toasts work again. The PR #5738 was merged **yesterday (late afternoon)**. With the latest state of `release_8` toasts should work again.

We noticed that the 'Who is online?' toasts could be provided without using `Services/Notifications` (they indeed misused `Services/Notifications` in the past).

Best regards
@mjansenDatabay

Services/Awareness/classes/Provider/AwarenessToastProvider.php
```diff
@@ -0,0 +1,110 @@
...
+            $links = [];
+            foreach ($new_users as $user) {
+                $uname = "[" . $user['login'] . "]";
+                if ($user['public_profile']) {
+                    $uname = $user['lastname'] . ", " . $user['firstname'] . " " . $uname;
+                }
+
+                $toast = $toast->withAdditionToastAction(
+                    $this->toast_factory->action(
+                        '',
```
chfsx: @iszmais this ToastAction should have a non empty 'id', maybe something like `'goToUser'`

iszmais:
Thank you for your quick reponse @alex40724 !
I really appreciate!

Aspointed out the current state of the test installation is due to another, unrelated problem wich @chfsx and myself fixed already for the release. While fixing that i stumbled upon that target issue here an worked out this proposal.

I totally agree than within the initial state of The FR of the Toast feature it would make completly sense to put the WIO Notifications inside the center. However within the implementation of the FR and the depending Jour Fixe meetings we set a goal to distinguish the Toast and Notifcation completly in Future wich comes no into reality now partialy over time.

We have now reached a point where we can create and handle Toast without the knowledge of notifications and the WIO Notifications are the mostly appropiate for this change.
Im pretty sure this change will be obligated at least for the first release patch, since it improves not only the code but has a convenient effect on the user interface too. I just wanted to anticapate the upcoming mantis ticket for it by putting it into the release, but if you feel unsure about it i totally respect your desicion to take that to the JF.

Greetings,
@iszmais

alex40724:
@iszmais  Thanks for the explanations. I was not aware of https://github.com/ILIAS-eLearning/ILIAS/pull/5725. Maybe this larger refactoring could have been also mentioned in a Jour Fixe, since we are "almost stable". Of course this is all fine if necessary to fix the issues. Since @mjansenDatabay  and @chfsx will review this and they are both TB members, it is ok to merge this when all reviews are "green" even without JF. So I set you to be responsible to merge.

I have one open issue: The old implementation seemed to limit the number of displyed users to "20" to not retrieve to much data if a higher number of users are online. I am not sure about the performance impact but the new implementation seems not to include this limitation. Please re-add it, if you consider this being appropriate, before merging.

Thanks for taking care of adopting the new concepts in the Who-Is-Online component.

alex40724 (approved):

Services/Awareness/classes/class.WidgetManager.php
```diff
@@ -63,60 +63,6 @@ public function __construct(
         $this->action_collector = \ilUserActionCollector::getInstance($a_user_id, new \ilAwarenessUserActionContext());
     }
 
-    /**
-     * Maximum for online user data
-     */
-    public function getMaxOnlineUserCnt(): int
-    {
-        return 20;
-    }
```
alex40724, line 72: This limitation seems to be missed in the new implementation, but may be helpful to prevent performance issues.

mjansenDatabay: Good catch Alex.

Services/Awareness/classes/Provider/AwarenessToastProvider.php
iszmais: I connected it to the scope awareness since this provide should handle all awareness toast atm.
If there are additional toast in future it may be usefull to split that but that for responsibles in future to decide.

Services/Awareness/classes/class.WidgetManager.php
iszmais: Due to the refactored Fetching this shouldnt cause any Performance issue no more, but i added anyways it for the enhanced user experience (nobody wants to see 200 new online users).

alex40724 (approved):

mjansenDatabay (approved):

PR #5731, opened 2023-03-27
centralized common property

iszmais:
Centralizes the Notifcation type, wich is used in every Notification provider, to ensure correct scoping when handling notifications.


mjansenDatabay:
Hi  @iszmais ,

we don't have to change any `src/GlobalScreen` code to solve the current issues for ILIAS 8.

Not every notification provider of the `GlobalScreen` has to provide a type, nor does a concrete provider for the notification center need to inherit a `MUTED_UNTIL_PREFERENCE_KEY` constant.
I know that **my** notification providers for the global screen use this concept and constant, but this is an implementation detail and should not be part of the abstraction.

Best regards,
@mjansenDatabay

PR #5732, opened 2023-03-27
UI docs ITCSS proposal wording, link to guide

catenglaender:
The ITCSS proposal document and the Guidelines for SCSS Coding were repeating the same information from different angles. With this PR each document was refined to have its own focus:

* **Using frameworkless SASS and ITCSS** is about why the big changes to the style code had to be made and what overarching strategy guides ongoing and future delos projects
* **Guidelines SCSS Coding** is about the detailed rules and best practices any contributions to the style code have to follow.

This should make more clear how these two documents build on each other instead of presenting the same or rivaling information. Each one now links to the other one and states its purpose and scope in comparison.


Amstutz:
Thx a lot for cleaning this up!

PR #5733, opened 2023-03-28
LSO: 36982, fix export (inits) & 35955, add setting for access_times

nhaagen:
https://mantis.ilias.de/view.php?id=36982
https://mantis.ilias.de/view.php?id=35955


klees:
Merged manually.

PR #5734, opened 2023-03-28
LSO: 36150, force 'add new item'

nhaagen:
https://mantis.ilias.de/view.php?id=36150


klees:
I opted to solve this in [ilObjLearningSequenceGUI::showPossibleSubObjects](https://github.com/ILIAS-eLearning/ILIAS/commit/aa0ae49f64a452d7c2e21a1cfb42131f2251e4f6). Seems to be the reason for using inheritance and overwriteable methods here, after all.

PR #5735, opened 2023-03-28
Container: fix AdvMD select filters (35621)

schmitz-ilias:
This PR fixes [35621](https://mantis.ilias.de/view.php?id=35621) and [36850](https://mantis.ilias.de/view.php?id=36850).

Please don't cherry-pick this to R8 and trunk, I'll provide a separate PR for that.


alex40724 (approved):
Thanks!

PR #5736, opened 2023-03-28
Container: fix AdvMD select filters for R8 (35621)

schmitz-ilias:
This PR is #5735 ported to R8 and trunk. It contains a bunch of additional php8 fixes in related AdvMD classes, and does not have the workaround necessary for R7.


alex40724 (approved):
Thanks!

PR #5737, opened 2023-03-29
Add kitchen sink element: LightboxCardPage Implementation

lscharmer:
This PR adds a new element LightboxCardPage to the UI Framework + Implementation.
This PR is based on: #5286.

### DIFF
--- a/src/UI/Component/Card/Card.php
+++ b/src/UI/Component/Card/Card.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 namespace ILIAS\UI\Component\Card;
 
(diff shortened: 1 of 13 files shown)


thibsy (changes requested):
Hi @lscharmer

Second of all, I apologize for the delay here!

- [x] `{ORIENTATION}`: please add the space (to separate if from the other class(es)) in the HTML template instead of adding it in the renderer. I'm certain this is more error-prone when added in the PHP code. I probably just discovered an error because of this for `CLASS_ACTIVE` as well, where we forgot to add a space in the PHP-string in `Renderer::renderLightbox()`. If you have the time, you could fix this for the other variable as well :).
- [x] `Renderer:renderPage()`: please rename `$i` to something more concise like `$page_number`. It wasn't clear what this variable is used for at first glance.
- [x] Example: please remove the copyright license header from the file.
- [x] Unit tests: please add some, at least one rendering test.

I also kindly ask you to rebase this onto trunk again, so the conflicts are resolved and this is ready for merge.

Kind regards,
@thibsy

thibsy (approved):
Hi @lscharmer,

Thx a lot for implementing the changes. Also thanks for improving the `LightboxTest` using a data-provider =). IMO this is now good to go so I'll merge it right away.

Kind regards,
@thibsy

PR #5739, opened 2023-03-29
Implement a notifcation indentification for future deletion handles

iszmais:
This PR introduces a identifation for notifications, wich enable the possibility to remove them on specific commands and trigger.

### DIFF
--- a/Services/Notifications/classes/Identification/NotificationIdentification.php
+++ b/Services/Notifications/classes/Identification/NotificationIdentification.php
@@ -1,0 +1,46 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\Notifications\Identification;
+
+/**
+ * @author Ingmar Szmais <iszmais@databay.de>
+ */
+class NotificationIdentification
+{
+    private string $type;
+    private string $key;
+
+    public function __construct(string $type, string $key)
+    {
+        $this->type = $type;
+        $this->key = $key;
+    }
+
+    public function getType(): string
+    {
+        return $this->type;
+    }
+
+    public function __toString(): string
+    {
+        return $this->key;
+    }
+}
--- a/Services/Notifications/classes/Setup/ilNotificationUpdateSteps.php
+++ b/Services/Notifications/classes/Setup/ilNotificationUpdateSteps.php
@@ -168,4 +168,16 @@ class ilNotificationUpdateSteps implements ilDatabaseUpdateSteps
         );
 
     }
+
+    public function step_10(): void
+    {
+        if (!$this->db->tableColumnExists('notification_osd', 'identification')) {
+            $this->db->addTableColumn('notification_osd', 'identification', [
+                'type' => 'text',
+                'length' => 255,
+                'notnull' => true,
+                'default' => ''
+            ]);
+        }
+    }
 }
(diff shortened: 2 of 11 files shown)


Services/Notifications/classes/Identification/NotificationIdentification.php
```diff
@@ -0,0 +1,46 @@
...
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\Notifications\Identification;
+
+/**
+ * @author Ingmar Szmais <iszmais@databay.de>
```
mjansenDatabay, line 24: You can remove the @author attribute, we have the Git history.

Services/Notifications/classes/Setup/ilNotificationUpdateSteps.php
```diff
@@ -168,4 +168,16 @@ public function step_9(): void
         );
 
     }
+
+    public function step_10(): void
+    {
+        if (!$this->db->tableColumnExists('notification_osd', 'identification')) {
+            $this->db->addTableColumn('notification_osd', 'identification', [
```
mjansenDatabay, line 175: What about existing entries? I see two options:

1) Truncate `notification_osd`
2) Using the internal ID as `identification`

Why? Because an empty ID can be an issue when passing it to the `GlobalScreen` or the `UI Toasts`, or even when handling the `callbacks` (maybe needs some extra handling for empty strings)`.

Services/Notifications/classes/Setup/ilNotificationUpdateSteps.php
```diff
@@ -168,4 +168,16 @@ public function step_9(): void
...
+                'type' => 'text',
+                'length' => 255,
+                'notnull' => true,
+                'default' => ''
```
mjansenDatabay, line 179: Therefore I suggest to omit the default.

mjansenDatabay (changes requested):

Services/Notifications/classes/Identification/NotificationIdentification.php
iszmais: As I Approve the reduction of pleonasms i would preserve the auther here since the information may already exists, but not in the project itsself but in its enviroment.

This may be reconsidered when the ILIAS Association sets git as a requirement for the usage of ILIAS but as far as i know its just a strong recommendation at the moment

Services/Notifications/classes/Setup/ilNotificationUpdateSteps.php
iszmais: I would favor to not set a default just for the purpose to have it, but i would not truncate the table without the adminstrators awareness, since it may contain old but upcoming notifications.
Atm the table will be truncated when the notifications are disabled. Else the empty identifications should cause any trouble. The global screen uses (and will further use) the real id of the notifications and for the UI Toast the setting of the default would be as inconvenient as the empty one.
And as far as i planned the callbacks should anyways be that strict to only handle only their own created notifications and not past ones wich would defy their limited scoping too. Therefore i would approve the disabling of the setting to be the only case in wich "lost" notifications can be deleted.

Therefore i dont see a need for a default here.

Services/Notifications/classes/Setup/ilNotificationUpdateSteps.php
iszmais: See above.
That default is the last stand to ensure the the data consistency but nothing more (its like the trash scope).
I would not want to encourage someone to really using it by giving it a semantic default.

Services/Notifications/classes/Identification/NotificationIdentification.php
mjansenDatabay: ?!? How do you contribute to  the ILIAS code base now, the last 8 years, and in the future **without** Git?. If we move to another VCS in the future, there will be a migration (like done in the past). But for me it's okay, removing this is not a hard requirement.

iszmais: Im not concerned about the contribution, this has definetly a requirement to git, but for the communication in general.
I agree that most profiteers of ILIAS get their product directly from github, but there are some exceptions (with more or less valid reasons) who recieving only the raw data, wich is still a valid distribution since git is only recommended for the installation.
And those profiteers may have not the additional information git providers but should still be able to get in contact with a maintainer or responsible over the code.

iszmais:
@mjansenDatabay Thank you for your review.
I noticed you merge the trunk version already, but not this one.
Causes aside, i would strongly encourage you to merge this one also to not differ the stage to much, especially since this is not a feature in any way but just a comfort cvhange for further development.

Otherwise we get in trouble of keeping track whats in what version when making other prs in the same context (like https://github.com/ILIAS-eLearning/ILIAS/pull/5827)

Greetings,
@iszmais

PR #5741, opened 2023-03-29
UI/Input/Field/Duration: allow same dates for start and end

nhaagen:
UI Input 'Duration' fails when start and end are equal; however, I cannot see a good reason for this (anymore).


Amstutz:
thx a lot!

PR #5742, opened 2023-03-30
UI: Typos and other corrections for README.md

tbongers-cat:
When I explored the concepts for new UI components via the `README.md`, I stumbled across a few typos and other errors, which are corrected with this PR.

In addition, the section for Less in "How to introduce a new component" (step 15 ff.) needs to be reworked for the switch to SCSS, but that is outside of my field of expertise.


Amstutz:
Thx a lot @tbongers-cat . The part about SCSS has been fixed in the meantime.

PR #5743, opened 2023-03-30
T&A: fix for #31918

okaufman:
This will fix https://mantis.ilias.de/view.php?id=31918
This will remove the function ilTestPlayerAbstractGUI::archiveParticipantSubmission since it is no longer used.


kergomard:

PR #5752, opened 2023-03-30
[FIX] UI: `Roundtrip` and `Interruptive` should no longer be translated

thibsy:
Hi all,

Please note that this PR contains a feature which has yet to be merged, see #5203. After the PR has been merged I will rebase this one again. Now to the issue at hand (second commit):

I noticed when using `Interruptive::withActionButtonLabel()`, `Interruptive::withCancelButtonLabel()` and `Roundtrip::withCancelButtonLabel()` a language variable has been expected. Since we'd like to keep things consistent I recommend to remove this behaviour and only translate **default values** (if no custom label was provided) while rendering. Otherwise, we mix up expecting language variables and translated strings which confuses the consumer. This was IMO inconsistent because when creating UI components we always expect translated strings instead of language variables, which would be the vast majority.

The change for cancel-labels is not exactly a breaking-change, as the same strings will still be displayed, they might just not be translated anymore. I checked the usages and there are not many places these labels are set, from which many even use the default string anyways.

What might be a breaking-change is that I removed the `Interruptive`s `name` attribute, where the language-variable of `Interruptive::withActionButtonLabel()` has been used as a command (so `$_POST['cmd']` contained this value). This wasn't great anyways, since any change of `ilCtrl` in regards to the request-parameters would break this. Instead, we should build proper links for the action in the first place, which as the maintainer of `ilCtrl` I strongly endorse anyways :). Furhtermore, it seemed odd that the action-label should serve as the command. I checked the usages of this method as well and the change might break up to 6 cases, which can easily be resolved though.

I will commit the according changes here as well once this PR has been approved.

Kind regards,
@thibsy

### DIFF
--- a/Modules/Blog/Exercise/class.ilBlogExerciseGUI.php
+++ b/Modules/Blog/Exercise/class.ilBlogExerciseGUI.php
@@ -188,7 +188,7 @@ class ilBlogExerciseGUI
 
         if ($tooltip) {
             $modal = $ui->factory()->modal()->roundtrip($lng->txt("exc_instruction"), $ui->factory()->legacy($tooltip))
-                ->withCancelButtonLabel("close");
+                ->withCancelButtonLabel($lng->txt("close"));
             $elements[] = $modal;
             $buttons[] = $ui->factory()->button()->standard($lng->txt("exc_instruction"), '#')
                 ->withOnClick($modal->getShowSignal());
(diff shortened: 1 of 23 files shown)


thibsy:
Update: I rebased, this is ready to be reviewed as well :).

Amstutz (approved):
Thx a lot for improving this. Special thx for the provided description in the PR and the example.

PR is hereby accepted. However, I let you click to merge button here, so you can sync this with the necessary fixes described above.

thx a lot.

thibsy:
Thx @Amstutz 

I applied the fixes or translated all labels (where necessary) and merged this right away.

Kind regards!

PR #5757, opened 2023-03-31
RootFolder: Change Function For Icon

kergomard:
Hi @alex40724 
While cleaning up the custom icons forms in `ilObject` I found this strange function that produces a broken out put and is only used in one place. Can we get rid of it please?
The changes in `ilObjectGUI` are most obviously fine with my, so if you agree you can just merge, ...or give me a thumbs up and I will take care of it.

Thank you and best,
@kergomard


alex40724 (approved):
Mmh, this seems to be some older code of @mjansenDatabay. I guess we missed that when migrating to the current implementation. So yes, the Rootfolder changes are fine for me, thanks.

kergomard:
Thanks @alex40724 for the fast reply!

PR #5758, opened 2023-03-31
User: Fix user language in `ilCronDeleteInactiveUserReminderMailNotif…

mjansenDatabay:
…ication`

See: https://mantis.ilias.de/view.php?id=37090

If approved this should be picked to `release_8` and `trunk`.


kergomard:
Thank you very much @mjansenDatabay ! Merged and cherry-picked to release_8 and trunk.

PR #5760, opened 2023-04-01
#37025 don't show instant response 

fneumann:
Don't show an instant response if a request for the instant response should fix the answer and the answer is not yet fixed.


kergomard:
Thank you @fneumann ! Will merge and cherry-pick to 8 and trunk, but without the part for the Dynamic Test, as it does not exist anymore.

PR #5761, opened 2023-04-01
fixed issues with test import

fneumann:
I tried to import an export of the test with participants results from https://test8.ilias.de/goto_test8_tst_78209.html to my test installation and stubled across two little issues.


kergomard:
Thank you very much @fneumann ! I will merge and cherry-pick to trunk.

PR #5762, opened 2023-04-03
Object: Implement KS-Forms for Common Properties

kergomard:
This PR implements the request of the Jour Fix [here](https://docu.ilias.de/goto_docu_wiki_wpage_7199_1357.html).
I put this in front of the JF, because of the [decision of the JF concerning "PR1211 • Object Service: Custom Settings"](https://docu.ilias.de/goto.php?target=wiki_1357_JourFixe-2018-10-08). This PR deprecates this service with ILIAS 11 and thus sets the options to use Object Properties  with Legacy-Forms to deprecated with ILIAS 11.

I believe the `CommonSettings` of an object are actually representations of `ObjectProperties` as form elements. Thus they belong to the `Object` and should also be accessed from there. The `ObjectService` should be a place for future factories for different aspects of the `Object` itself, but I'm not there yet.

This PR is still a draft and not ready to be merged. I would like to finalize and merge it by the end of this week.

Thanks and best,
@kergomard


matthiaskunkel:
Jour Fixe, 03 APR 2023: We highly appreciate Stephan's suggestion and accept the PR for trunk.

PR #5766, opened 2023-04-03
[FIX] 0036838 & 0036839 A11y improvements in File-Upload

chfsx:
This is a fix for
- https://mantis.ilias.de/view.php?id=36838
- https://mantis.ilias.de/view.php?id=36839

Currently, the glyphs for expanding and collapsing the additional fields as well as the glyph for removing a file cannot be operated with the keyboard, see the two issues.

To make these functions operable via the keyboard, the glyphs must first be accessible via tabs. Since this does not apply to all glyphs, I have introduced a withTabbable and isTabbable:

https://github.com/ILIAS-eLearning/ILIAS/compare/trunk...srsolutionsag:ILIAS:fix/0036838/9/focus-fileupload-elements?expand=1#diff-8f0e178a2a3228811d6a37900f164f4665486cbb2ca479247a94baaf0e14d11aR136

This is then used in the FileInput and the JS is adapted accordingly so that, in addition to a click, an enter-key also triggers the corresponding events:

https://github.com/ILIAS-eLearning/ILIAS/compare/trunk...srsolutionsag:ILIAS:fix/0036838/9/focus-fileupload-elements?expand=1#diff-2d9abcacf08b7e7e7a931e40d78ee734af912ab8af6659fa4b7a570dd349f632R176

If accepted, thus must be cherry-picked to release_8 as well.

### DIFF
--- a/src/UI/Component/Symbol/Glyph/Glyph.php
+++ b/src/UI/Component/Symbol/Glyph/Glyph.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 namespace ILIAS\UI\Component\Symbol\Glyph;
 
@@ -129,4 +129,10 @@ interface Glyph extends Symbol, Clickable
     * Get a Glyph like this with an action.
     */
     public function withAction(string $action): Glyph;
+
+    /**
+     * @return bool Whether the glyph is tabbable (e.g. for screen readers). This is true if a Glyph has
+     * an action or a onClock Signal.
+     */
+    public function isTabbable(): bool;
 }
--- a/src/UI/Implementation/Component/Input/Field/Renderer.php
+++ b/src/UI/Implementation/Component/Input/Field/Renderer.php
@@ -845,7 +845,7 @@ class Renderer extends AbstractComponentRenderer
     ): Template {
         $template->setCurrentBlock('block_file_preview');
         $template->setVariable('REMOVAL_GLYPH', $default_renderer->render(
-            $this->getUIFactory()->symbol()->glyph()->close()
+            $this->getUIFactory()->symbol()->glyph()->close()->withAction("#")
         ));
 
         if (null !== $file_info) {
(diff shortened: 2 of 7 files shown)


chfsx:
@Amstutz Another approach would be: Glyphs would need their own JS. The tabbable could be derived from it to trigger an action or an onClick signal. If so, the glyph itself would trigger a click event through a `keyup` event to trigger the behaviour of a click. This implementation would be somewhat larger, but perhaps more places would benefit from it.

Amstutz:
Hi @chfsx 

I was reading through your PR without reading your second comment and came more or less to the same conclusion. I dislike the "Tabbable" on the public interface, the Glyph itself should know whether it is tabbable or not. E.g. if an ID is generated successfully (JS attached) OR an Action is set, THEN it should be tabbable. However, as you say, there is still some manual work required, as you point out above.

What you propose would maybe be Gold standard. However, do you see an option to attach the tabbable functionality automatically without having to go all the way with refactoring the Glyphs?

chfsx:
Hi @Amstutz I adapted the PR accordingly, Glyphs are now Tabbable if they have an action or a click signal (see UniTest as well). But: This breaks a lot of other UI UnitTests since there are many consuments of Glyph during rendering.

There is now a JF for Glyphs which triggers the Click event on Keyup "Enter": https://github.com/ILIAS-eLearning/ILIAS/pull/5766/files#diff-8b7a0c1c55375012cf5db5ec1b03d8c558c5e88f00b1e52da7c6a0ddd2eee0eaR9

thibsy (changes requested):
Hi @fschmid

Thx a lot for improving the a11y in this context, especially for improving the situation for Glyphs in general. 

Your solution breaks quite a number of unit tests, as you correctly point out. @Amstutz will fix them right after merging so you do not have a huge additional task at hand here. You might ignore them. 

However, please change:
- [x] glyph.js: line 9, add a `?` for optional chaining when registering the event handler.
- [x] `registerEvents`: it might not be obvious why this function exists, please leave a brief comment (preferably incl. JSDoc annotations). Maybe also rename it to `triggerClickEventsOnEnter` or something more precise.

Maybe change, or quickly contact me
- [x] ES6: please use it.

Kind regards
@thibsy, @Amstutz

chfsx:
Hi @Amstutz and @thibsy 

I refactored this to address the following points:
- A JS is no longer needed, therefore no ES6 module is done
- A Glyph no return true on isTabbable, if there is either an Action or a onClick-Signal, no withTabbable available now

Please note there are many failing tests now, since a Glyph is used in many other Components as well. @Amstutz mentioned via Discord that he will look after those tests, thanks in advance. I adjusted the Tests for Input\File and for Glyph and added a new Test for the isTabbable decision.

chfsx:
@Amstutz Let me know if there is anything else I can do about this PR. Currently, I think only the failing tests are open.

Amstutz:
Hi @chfsx 

thx for the changes. I picked them manually to 8 and trunk and fixed the tests as promised. However the tests flagged a few more things, that I did not catch while reviewing. They are all fixed already, see: https://github.com/ILIAS-eLearning/ILIAS/commit/b0b15bf863c63f814a3dbc1068359f7f82fbb828

- Inactive Items: Should not be tabable (adapted isTabable)
- Unit Tests: Please do only provide rendering tests if really testing rendering. For testing functions such as isTabbable, please provide tests without rendering (removed some rendering tests, added some only testing the function).
- Public Interface: I removed isTabbable from the public interface, sinced it is only usesd internally for rendering.

Thx for your great contribution in fixing the above issues!

PR #5767, opened 2023-04-03
Setup Migrate-Command: malformed example message

utesche:
Hello @klees,

... the PR's title is it.
Of course, the PR concerns branches release_7 and trunk too.

Greetings,
Uwe.


klees:
Thx @utesche!

klees:
No need to move this to release_7, the string is escaped with `"` there and thus should be interpolated correctly.

PR #5770, opened 2023-04-04
UI: Cleanup for UI Input components

tbongers-cat:
These changes have been moved from PR #5711, as they came up during its development, but are not directly related:
- Added better type hinting and unified PHPDocs for `Form` and `Input` factories
- Removed unused parameters from test
- Corrected default value for group byline to null in factory (-> also removes empty DIV from render, see test)


Amstutz:
Hi @tbongers-cat

Thx for cleaning this up. Nice to see some small improvement in the output visible in the rendering test, by improving this (removed empty div).

If somebody might be concerned, as far as we understand, changing string $byline = null to ?string $bylilne = null is only a readibiliy improvement. However, we like it.

thx

PR #5771, opened 2023-04-04
User: Use argon2id hashes if supported (ILIAS 9)

mjansenDatabay:
See: https://docu.ilias.de/goto_docu_wiki_wpage_6330_1357.html

This PR changes the `User Password Manager` by using `argonid` as the desired password hashing algorithm for local user passwords.

We used configuration parameters based on the OWASP recommendation.

The default/fallback algorithm is still 'bcrypt', because it is available on every PHP compilation, which might not be the case for `argon2id` (most distributions do, and I added this dependency to the installation documentation of ILIAS, but who knows ...).

### DIFF
--- a/Services/User/classes/Setup/class.ilUserDB90.php
+++ b/Services/User/classes/Setup/class.ilUserDB90.php
@@ -53,4 +53,23 @@ class ilUserDB90 implements ilDatabaseUpdateSteps
             );
         }
     }
+
+    /**
+     * Modifies the 'passwd' field in table 'usr_data' to accept longer passwords
+     */
+    public function step_2(): void
+    {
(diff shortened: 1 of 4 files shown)


kergomard:

This looks good to be, I've just one question: What is the reason to keep the [second parameter in `ilUserPasswordEncoderFactory::getEncoderByName()`](https://github.com/ILIAS-eLearning/ILIAS/pull/5771/files#diff-d7e3819db05ef19906003db38edd58f4a9b51e7c84aa0cac60ae3a071b728ca4L131)? I don't think it is used anywhere now, thus I feel like it should go.

Thanks again and best,
Stephan

mjansenDatabay:
@kergomard You were right, I removed the optional parameter.

kergomard:

PR #5772, opened 2023-04-04
UI: Fix Representation of Minutes in JS

kergomard:
Maybe I'm also doing something wrong, but I was unable to fill a Duration-Input with Time correctly without this change.


kergomard:
Some more information. Sorry, for not adding this yesterday, I apparently was in my own world:
- The problem basically is, that I need to add a `DateFormat` with Time, otherwise when the value is added, the time is going to be stripped and `DateTime::withTime(true)` will present an entry ending in 00:00.
- On a sidenote: `DateTime::withTime(true)` has the effect of then adding a second Time-Display to the rendered Input. Thus rendering it, as far as I can see, useless.

Amstutz:

PR #5773, opened 2023-04-05
[PROPOSAL] UI: `Form` compatibility for `Interruptive` modals.

thibsy:
### Reusability

`Interruptive` modals have no client-side functionality which enables the reusability of them. Consider the following scenario: There is a table with multiple entries (or rows), which all have an action `Dropdown`. Several `Button`s of each `Dropdown` will trigger a "dangerous" action. 

The current implementation of this scenario will render an `Interruptive` modal for each of these dangerous actions for each entry. We can already slim down the implementation by rendering "dummy" `Interruptive` modals, which are replaced by their `withAsyncRenderUrl`, instead of actual ones, so they can share an endpoint which provides (sort of) the same `Interruptive` on each request. This still leads to much HTML needing to be rendered, since every modal could potentially be requested. The new `Data` table already does reuse one modal for all actions, but it still makes requests to replace the modal content.

This produces more network traffic, which would be fine, if it weren't unnecessary. The solution to this problem is not quite trivial though, but if the component provided client-side functionality to fill an `Interruptive`s affected items, this could get rid of asynchronous rendering altogether for this scenario.

### Submissions

An `Interruptive` modal contains several affected items (`InterruptiveItems`), which are identified by an ID. These items will be submitted via POST. Currently, there is no way of knowing how the affected items must be retrieved from the request-body, right now the key `interruptive_items` is hardcoded. Therefore, current implementations cannot define their own key for the request-body and use some magic string that could change at any time and break the endpoint.

The component can also not be provided with a `ServerRequestInterface`, like `Form`s, to obtain the submitted items with e.g. `getData()`. Thus, making implementations somewhat fragile.

## Solution

The first problem (reusability) points into the direction of dynamic inputs (`HasDynamicInputs`), which allows the client to render a template of an input on-demand. A template in this case would be an `InterruptiveItem`, whose ID would be used for submission. This also leads to the latter problem, which points to the behaviour of a `Form`. If `Interruptive` modals behaved like `Form`s, it would be way more robust because the consumer would not need to worry about retrieving the affected items from `$_POST`.

A possible solution would be:
- `InterruptiveItem` extends `Input\HasDynamicInputs`, so a template for client-side rendering can be used.
- `Interruptive` extends `Form\Standard`, so the modal can be treated in the same manner.
- `Interruptive` becomes a decorator for a `FormWithNoSubmitButton`, just like `Roundtrip`.
- `Interruptive` does only accept `InterruptiveItem`s as inputs, which need to be provided in the constructor.
- `modal.js` is extended by a public function to trigger the rendering process of a dynamic inputs (`InterruptiveItem`s)

There is currently an open discussion about displaying multiple `InterruptiveItems` inside the modal, which can be selected or not with an according `Checkbox`. The checkbox would decide whether the item should be considered on submission or not. If the `InterruptiveItem` already extends the `HasDynamicInput` and therefore `Input` interface, this looks also quite approachable:
- `Checkbox` is extended to accept another `Component` (or something more explicit) with e.g. `withAdditionalContent()`, which is displayed besides the checkbox-element.
- `SelectableItem` as new `InterruptiveItem` which returns a `Checkbox::withAdditionalContent()` as `getTemplateForDynamicInputs()`.

### DIFF
--- a/src/UI/Component/Modal/Interruptive.php
+++ b/src/UI/Component/Modal/Interruptive.php
@@ -21,13 +21,14 @@ declare(strict_types=1);
 namespace ILIAS\UI\Component\Modal;
 
 use ILIAS\UI\Component\Modal\InterruptiveItem\InterruptiveItem;
+use ILIAS\UI\Component\Input\Container\Form\Standard;
 
 /**
  * Interface Interruptive
  *
  * @package ILIAS\UI\Component\Modal
  */
-interface Interruptive extends Modal
+interface Interruptive extends Standard, Modal
 {
     /**
      * Get the message of this modal, displayed below the modals title
@@ -38,31 +39,6 @@ interface Interruptive extends Modal
      * Get the title of this modal
      */
     public function getTitle(): string;
-
-    /**
-     * Get a modal like this submitting the form to the given form action
-     */
-    public function withFormAction(string $form_action): Interruptive;
-
-    /**
-     * Get a modal like this listing the given items in the content section below the message.
-     * The IDs of the interruptive items are sent via POST to the form action of this modal.
-     *
-     * @param InterruptiveItem[] $items
-     */
-    public function withAffectedItems(array $items): Interruptive;
-
-    /**
-     * Get the label of the action button in the footer
-     */
-    public function getActionButtonLabel(): string;
-
-    /**
-     * Get a modal like this with the action button labeled
-     * according to the parameter.
-     * The label will be translated.
-     */
-    public function withActionButtonLabel(string $action_label): Interruptive;
 
     /**
      * Get the label of the cancel button in the footer
@@ -82,9 +58,4 @@ interface Interruptive extends Modal
      * @return InterruptiveItem[]
      */
     public function getAffectedItems(): array;
-
-    /**
-     * Get the form action where the action button is sending the IDs of the affected items
-     */
-    public function getFormAction(): string;
 }
--- a/src/UI/Component/Modal/InterruptiveItem/InterruptiveItem.php
+++ b/src/UI/Component/Modal/InterruptiveItem/InterruptiveItem.php
@@ -20,14 +20,14 @@ declare(strict_types=1);
 
 namespace ILIAS\UI\Component\Modal\InterruptiveItem;
 
-use ILIAS\UI\Component\Component;
+use ILIAS\UI\Component\Input\Field\HasDynamicInputs;
 
 /**
  * Interface InterruptiveItem
  *
  * Represents an item to be displayed inside an interruptive modal
  */
-interface InterruptiveItem extends Component
+interface InterruptiveItem extends HasDynamicInputs
 {
     /**
      * Return an ID of the item


thibsy:
Please not that I have created this PR merely to document the current situation, which I'd like to improve very much. The reusability-aspect (with `HasDynamicInputs`) can certainly be discussed, but I do insist on the `Form` part.

This PR is also somewhat related to #5203.

klees:
Hi @thibsy,

I have started to look into this, but as you noticed yourself: This doesn't seem to be an easy problem. Do I understand correctly that this is not urgent? I would then postpone this for about another month and solve some other hard problems before looking into this. There's only limited space in my RAM... Is this ok for you?

Thanks for your patience!

thibsy:
Hi @klees

No problem. I think it's not very urgent. However, I want to point out that as part of the LUI project there might be many new `Interruptive` modals in use soon, which would all need to be touched again since this improvement will most likely mean changes for consumers.

Kind regards,
@thibsy

klees:
@thibsy I think this is superseeded by the approach in #6309, so we could close this. Would you agree? If not, we should maybe discuss both approaches in person, maybe together with @Amstutz and @nhaagen (and whoever is interested too...). If yes, please close this. Thanks!

thibsy:
@klees I don't think this PR tackles the same issue, since this one specifically tackles `Interruptive` modals and the referenced PR does not really improve them, right?. Maybe we can reuse some functionality of dialogs though. I am happy to discuss this on Discord sometime.

klees:
Hi @thibsy,

I stumbled upon this PR while doing the regular checks for old PRs for the Technical Board. I apologize that it took me so long to move this forward. I will look into this again soon.

Kind regards!

klees (commented):
Hi @thibsy,

I think I kind of understand the problem and solution. The situation here is definitely far from ideal. Thanks for bringing this up.

In my words, in a nutshell: The `Modal` components, and especially the `Interruptive`, are just too dumb. They are merely concerned with the display logic and not with data processing logic. Even regarding the display logic they do not do enough, as they are not dynamic enough to deal with situations where many different sets of items could be deleted.

This tries to improve the situation by turning `Interruptive Modals` to `Standard Forms`, which would have two effects:
* We could reuse the dynamic inputs terminology and functionality of the forms.
* We could reuse the data processing logic of the forms.

What I currently don't understand:
* How does this relate to [the `Dialog` proposal](https://github.com/ILIAS-eLearning/ILIAS/pull/7045]? From my understanding, both approaches aim to solve a similar problem. They are quite similar wrt to the idea to somehow include `Form` in the mix for data processing. They are different wrt to their approach to the requirements regarding dynamic: This proposal would solve the problem client side, the `Dialog` proposal would solve this via a server roundtrip. They also differ regarding their scope: The `Dialog` proposal has a larger scope as it tries to address the problem for all kind of interactions that are `Dialog`s somehow, while this proposal immediately addresses the one `Interruptive Modal`.

My thoughts about this approach: I definitely see the problem and I also see how this would solve it. I also agree that we should tackle that problem. I'm a little hesitant regarding the way we would cross component boundaries here. This somehow feels if we would need to grow the complexity in the `Input` implementation, just like we needed to when incorporating the `View Controls` into `Input`. I think we won't be able to sneak around that complexity in the end, even with `Dialog` or any other solution that implements principled data processing, but this feels a little to ad-hoc to take that burden. This is rather a gut feeling then some well weighted opinion, though.

I have a stronger opinion regarding the upgrade of the `Interruptive Modal`: The `Modal` component in general feels semantically weak to me. From my POV it implements some mechanism (the JS, mostly), but it does little for our semantic understanding of the UI. The `Interruptive` part of component provides semantic meaning, which is fine. But why does this need to be a `Modal`? Why can't we, e.g., use a new page for that same interruptive workflow? Wouldn't that question, whether an interruptive workflow should use a modal or a new page, be something that we would let UX people decide, or even change it sometime in the future? From that perspective, I prefer something like the `Launcher`, that uses the `Modal` internally. Hence, I would prefer the `Modal` to stay relatively dump and instead move the decision if a modal should be used at all to the internals of some semantically richer component. It still might be ok to upgrade the `Interruptive Modal`, but someone would need to convince me.

To go on here: Shall we have a meeting to discuss this proposal and [the Dialog proposal](https://github.com/ILIAS-eLearning/ILIAS/pull/7045), so we can make some progress on the problem?

Kind regards!

dsstrassner:

PR #5774, opened 2023-04-05
Implement a notifcation indentification for future deletion handles for TRUNK

iszmais:
Sister PR to #5739

### DIFF
--- a/Modules/Chatroom/classes/class.ilChatroom.php
+++ b/Modules/Chatroom/classes/class.ilChatroom.php
@@ -18,6 +18,7 @@
 
 declare(strict_types=1);
 
+use ILIAS\Notifications\Identification\NotificationIdentification;
 use ILIAS\Notifications\Model\ilNotificationConfig;
 use ILIAS\Notifications\Model\ilNotificationLink;
 use ILIAS\Notifications\Model\ilNotificationParameter;
@@ -31,6 +32,7 @@ use ILIAS\Chatroom\GlobalScreen\ChatInvitationNotificationProvider;
  */
 class ilChatroom
 {
+    public const ROOM_INVITATION = 'invitation_to_room';
     private static string $settingsTable = 'chatroom_settings';
     private static string $historyTable = 'chatroom_history';
     private static string $userTable = 'chatroom_users';
@@ -691,7 +693,10 @@ class ilChatroom
             $notification->setIconPath('templates/default/images/icon_chtr.svg');
             $notification->setValidForSeconds(ilNotificationConfig::TTL_LONG);
             $notification->setVisibleForSeconds(ilNotificationConfig::DEFAULT_TTS);
-
+            $notification->setIdentification(new NotificationIdentification(
+                ChatInvitationNotificationProvider::NOTIFICATION_TYPE,
+                self::ROOM_INVITATION . '_' . $gui->getObject()->getRefId(),
+            ));
             $notification->setHandlerParam('mail.sender', (string) $sender_id);
 
             $notification->notifyByUsers([$recipient_id]);
--- a/Services/Notifications/README.md
+++ b/Services/Notifications/README.md
@@ -21,6 +21,45 @@ interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
 
 If enabled, users will be informed by pop-up notifications about certain events.
 
+If disabled, the users will no longer be informed and all existing OSD Notifications will be deleted
+
 ![](./docu/images/settings_enable_osd_en.png)
 
 ***Note**: Less time allows more timely notifications, but increases server load.*
+
+## Usage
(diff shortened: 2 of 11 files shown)


Modules/Chatroom/classes/class.ilChatroom.php
```diff
@@ -691,7 +693,10 @@ public function sendInvitationNotification(
             $notification->setIconPath('templates/default/images/icon_chtr.svg');
             $notification->setValidForSeconds(ilNotificationConfig::TTL_LONG);
             $notification->setVisibleForSeconds(ilNotificationConfig::DEFAULT_TTS);
-
+            $notification->setIdentification(new NotificationIdentification(
+                ChatInvitationNotificationProvider::NOTIFICATION_TYPE,
+                self::ROOM_INVITATION . '_' . $gui->getObject()->getRefId(),
```
iszmais, line 698: @mjansenDatabay I just removed the subscope here, as far as i remember right its removed for ilias 9, right?

Otherwise its the same PR (except the constructors) as for ILIAS 8

mjansenDatabay: Exactly

mjansenDatabay (changes requested):
@iszmais Please add some lines to the README and explain how notifications can be created with the identification.

iszmais:
@mjansenDatabay i would appreciate if you could do another full review since i changed a few tweaks and did some updates for PHP 8.1 also.

Services/Notifications/README.md
```diff
@@ -21,6 +21,45 @@ interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
...
+- `$config->setIdentification($identification)` A unique `NotificationIdentification` wich can be used to identify this specific notification for future deletions.
+- `$config->setHandlerParam($handler_name, $params)` Paramters that will be given to the handler for additional informations (not used for OSD).
+
+**A notification wich is no longer valid will not be seen in any UI, but still persits on the database. Therefore all users
+of the notification service are obliged to remove their notifications on the convient triggers.**
+
+You can remove a notification by using the following functions withing the `ilNotificationOSDHandler`:
+- `removeOSDNotificationByIdentification($type, $identification)` This removes all notifications withing your `$type` wich are
+compatible with the `NotificationIdentification` given by `$identification`. Optional you may add a user ID as parameter to remove
+this notification only for one user.
```
mjansenDatabay:
```suggestion
## Usage

### OSD Notifications

Services can provide notifications to the system by calling;

```php
$DIC->notifications()->system()->toUsers($config, $users, $async);
```

with a `ilNotificationsConfig` object, a list of user ids and a flag marking async calls.

The config needs a type to be created. This type defines your scope of notifications in which you are able to create and delete
notifications. You may use an existing type if communicated with the maintainer of its scopes service. If given the type used
should be identical with the type of your `NotificationProvider`.

Furthermore you can set following properties on your notification configuration:

- `$config->setTitleVar($title)` The title of your Notifications. The passed argument may be an ILIAS language variable.
  - The notification system tries to translate the title if a respective translation could be determined. Otherwise the passed string will be processed untouched and finally passed to the output channels. If you want to translate it you have the option
  to add an additional parameter for replaceables and another for a language module to load e.g. `$config->setTitleVar($title, [$username], 'usr')`
- `$config->setShortDescriptionVar($description)` The description for pace limited presentations (e.g. popups). It translates the same as title.
- `$config->setLongDescriptionVar($description)` The full description (not used for OSD). It translates the same as title.
- `$config->setLinks($links)` A list of `ilNotificationLinks` which will presented as translated links below the description.
- `$config->setIconPath($path)` The path to the icon for your notifciation type. It will be shown in the indent of the notification or on top of groups of your notification type.
- `$config->setValidForSeconds($time)` The time will be visible for the user in target presentations.
- `$config->setIdentification($identification)` A unique `NotificationIdentification` which can be used to identify this specific notification case for future deletions.
- `$config->setHandlerParam($handler_name, $params)` Parameters that will be given to the handler for additional informations (not used for OSD).

**Important: A notification, which is no longer valid, will not be seen in any UI, but still persists in the database. Therefore all consumers
of the notification service are obliged to remove their notifications on the appropriate system events (e.g. a notification emitted because of a contact request can be removed if the request is approved) .**

You can remove a notification by using the following functions within the `ilNotificationOSDHandler`:
- `removeOSDNotificationByIdentification($type, $identification)` This removes all notifications within your `$type` which are
compatible with the `NotificationIdentification` given by `$identification`. Optional you may add a user id as parameter to remove
this notification only for one user.
```

iszmais: I assume you are a victim of your on injection :smile: 

I applied the changes manualy, feel free to take a look

mjansenDatabay (approved):

PR #5775, opened 2023-04-05
[FEATURE] LUI: Refactoring of MainMenu, Bibliographic, AdministrativeNotification

thibsy:
This PR replaces:
- Legacy Advanced Selection List
- Legacy Button
- Legacy Confirmation Screen
  - Exception for: File-Table (Multi-Commands)

### DIFF
--- a/Modules/Bibliographic/classes/FieldFilter/class.ilBiblFieldFilterFactory.php
+++ b/Modules/Bibliographic/classes/FieldFilter/class.ilBiblFieldFilterFactory.php
@@ -29,7 +29,7 @@ class ilBiblFieldFilterFactory implements ilBiblFieldFilterFactoryInterface
     public function findById(int $id): \ilBiblFieldFilter
     {
         /** @noinspection PhpIncompatibleReturnTypeInspection */
-        return ilBiblFieldFilter::findOrFail($id);
+        return ilBiblFieldFilter::where(['id' => $id])->first();
     }
 
 
--- a/Services/AuthShibboleth/classes/class.ilAuthShibbolethSettingsGUI.php
+++ b/Services/AuthShibboleth/classes/class.ilAuthShibbolethSettingsGUI.php
@@ -151,7 +151,7 @@ class ilAuthShibbolethSettingsGUI
 
         $rule_ids = $this->wrapper->post()->retrieve(
             'rule_ids',
-            $this->refinery->to()->listOf($this->refinery->to()->int())
+            $this->refinery->to()->listOf($this->refinery->kindlyTo()->int())
         );
         foreach ($rule_ids as $rule_id) {
             $rule = new ilShibbolethRoleAssignmentRule($rule_id);
--- a/Modules/Bibliographic/classes/Admin/Library/class.ilBiblLibraryGUI.php
+++ b/Modules/Bibliographic/classes/Admin/Library/class.ilBiblLibraryGUI.php
@@ -71,12 +71,12 @@ class ilBiblLibraryGUI
     public function index(): bool
     {
         if ($this->checkPermissionBoolAndReturn('write')) {
-            $b = ilLinkButton::getInstance();
-            $b->setCaption(self::CMD_ADD);
-            $b->setUrl($this->ctrl()->getLinkTarget($this, self::CMD_ADD));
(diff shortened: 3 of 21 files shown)


src/UI/templates/js/Core/src/core.js
```diff
@@ -2,7 +2,7 @@ import replaceContent from './core.replaceContent.js'
 import Tooltip from './core.Tooltip.js'
 
 if (typeof il === 'undefined') {
-    il = {}
```
chfsx: @thibsy / @lzehnder please revert those changes since these are not part of this refactoring

src/UI/templates/js/Core/src/core.Tooltip.js
```diff
@@ -12,7 +12,7 @@ class Tooltip {
      * The container of the tooltip and the trigger element.
      * @type {Element}
      */
-    #container:
```
chfsx: @thibsy / @lzehnder please revert those changes since these are not part of this refactoring

src/UI/templates/js/Core/dist/ui.js
chfsx, line 1: Same her, this doesn't seem to be a part of this refactoring...

Services/AuthShibboleth/classes/class.ilAuthShibbolethSettingsGUI.php
```diff
@@ -151,7 +151,7 @@ protected function confirmDeleteRules(): bool
 
         $rule_ids = $this->wrapper->post()->retrieve(
             'rule_ids',
-            $this->refinery->to()->listOf($this->refinery->to()->int())
+            $this->refinery->to()->listOf($this->refinery->kindlyTo()->int())
```
chfsx, line 154: What's the reason to change this to kindlyTo()?

Modules/Bibliographic/classes/FieldFilter/class.ilBiblFieldFilterFactory.php
```diff
@@ -29,7 +29,7 @@ class ilBiblFieldFilterFactory implements ilBiblFieldFilterFactoryInterface
     public function findById(int $id): \ilBiblFieldFilter
     {
         /** @noinspection PhpIncompatibleReturnTypeInspection */
-        return ilBiblFieldFilter::findOrFail($id);
+        return ilBiblFieldFilter::where(['id' => $id])->first();
```
chfsx, line 32: Is this change relevant for the refactoring? What happens if there is no entry with id `$id`? will lead to a fatal as well since then `null` instead of `\ilBiblFieldFilter` will be returned.

chfsx (changes requested):
There are several changes which seem not to be part of this refactoring (in JS files), please have a look at the inline comments.

Services/AuthShibboleth/classes/class.ilAuthShibbolethSettingsGUI.php
lzehnder: The reason for changing this to kindlyTo was that I got an error (due to the value being a string) when testing around during refactoring. I thought it would make sense to fix this right away, even when this bug wasn't caused by the refactoring, as it occurred in the component being refactored.

chfsx: @lzehnder thanks for clarification 👍

thibsy:
@chfsx the JavaScript changes are not directly related to the LUI refactoring, but they were necessary due to a bug, which lead to interruptive modals not being rendered asynchronously. This was due to 'il' not being declared properly and a missing semicolon in the 'Tooltip' class. Without this, the confirmation when deleting file versions does not work.

chfsx:
@thibsy could you factor those changes out to a separate PR with a related Mantis-Ticket?

thibsy:
UI issue has already been fixed, see #5785

chfsx (approved):
Thanks a lot! I'll merge after I got the customers OK

Amstutz (approved):
I did not review the code, but looked at the changes on your testing env. Thx a lot for providing this. Changes look as expected. Thx a lot.

Can be merged.

PR #5776, opened 2023-04-05
fixed settings and maintain gui in language handling

c-knof:


dsstrassner:
Is there a mantis issue, which can be linked in this PR?

c-knof:
No, there is actually no mantis issue. There were two issues due to php8 refactoring overseen and they are now fixed.

PR #5782, opened 2023-04-06
T&A Feature #0033515: Move Results in Test Export one column

fhelfer:
https://mantis.ilias.de/view.php?id=33515


dsstrassner:
Hi @mbecker-databay or @kergomard: 
JourFixe said, we need this for ILIAS 7, ILIAS 8 and Trunk ;)

kergomard:
Thank you very much @fhelfer for this PR. I had a hard time really understanding the expected behavior from the description in Mantis, but looking from the result it becomes understandable. I will merge and port to ILIAS 7 and 8.

PR #5783, opened 2023-04-06
[FEATURE] CI: improved copyright-checker

thibsy:
Hey all,

As you all know our CI workflow will check changed files (according to git) for a valid ILIAS copyright license header when committing to the repository or creating a PR.

You may also know that we **don't** want said copyright-header in the examples of UI components. Therefore, many PRs that adjust or introduce UI examples will fail the workflows because of this.

With this PR I inverted the check for files located in `src/UI/examples/`, so the workflow will now fail **if there is** a copyright-header.

When improving this script I also noticed a few flaws:
1. `echo` has been used over `printf`, which lead to problems on macOS
2. the logic did not enforce PSR-12 file-level doc-blocks (see [here](https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements))
3. whitespace (`[:space:]`), `/` and `*` characters were stripped before comparison

Especially point 3 is crucial, because this means developers could have made any change they liked as long as it only affected these characters. Hence, this would have been a valid copyright-license until now:
```php
/**
 *ThisfileispartofILIAS,apowerfullearningmanagementsystem
 *publishedbyILIASopensourcee-Learninge.V.
 ********************************************************
 *ILIASislicensedwiththeGPL-3.0,
 *seehttps://www.gnu.org/licenses/gpl-3.0.en.html
 *Youshouldhavereceivedacopyofsaidlicensealongwiththe
 *sourcecode,too.
 */////////////////////////////////////////////////////*
 /*IfthisisnotthecaseoryoujustwanttotryILIAS,you'llfind
 *usat:
 *https://www.ilias.de
 *https://github.com/ILIAS-eLearning*/
```
This version of the copyright-check will compare all lines of the comment without altering it, so any adjustment will fail the workflow. The check will stop at line 13 of the comment (the repository URL) intentionally, because there are currently several versions of how the copyright-header will end. Some comments fill the line with `*` characters, others just close it directly with `*/`. This does not only serve as a "workaround" of sorts, because this also leaves the option for _actual_ file-level doc-blocks, which could be written beneath the copyright-header now. Starting a new comment would violate the PSR-12 standard. For this reason, PHP-files are also searched from line 3, which means the PSR-12 file-level doc-block will be enforced from now on. The check can also be applied to any other file now, whereas the starting line will be 1.

This also brings me to a new feature: the check can now be applied to one or more files or directories (which are traversed recursively) if supplied as arguments. So e.g. this will perform the copy-right check on all UI examples:
```bash
sh CI/Copyright-Checker/copyright-checker.sh "src/UI/examples"
```
Without arguments, this script defaults to either:
a) local: all files of the current directory (`$(pwd)`), or
b) GitHub: all changed files of the current PR or commit.

---

Another part of this PR is the consideration of JavaScript files. I find it inconsistent to only enforce the ILIAS copyright license header for PHP files, when JavaScript files are written by ILIAS developers as well. This is why I also made a small adjustment to the retrieval of files in the copyright-checker, so `*.js` files are now considered.

As previously mentioned JavaScript files will be checked from line 1, which means the file-level doc-block will be expected at the beginning of the file. According to our JavaScript code-style ([Airbnb](https://github.com/airbnb/javascript)) there's no rule for file-level comments and there is no "PSR" standard like there is for PHP. So I've decided to implement this behaviour.

Kind regards
@thibsy

### DIFF
--- a/CI/Import/Functions.sh
+++ b/CI/Import/Functions.sh
@@ -11,7 +11,7 @@ function get_changed_files() {
     CHANGED_FILES=$(git diff-tree --name-only --diff-filter=ACMRT --no-commit-id -r ${GH_SHA} | grep -e '.php$' -e '.js$')
   else
     URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/files"
-    CHANGED_FILES=$(curl -s -X GET -G $URL | jq -r '.[] | .filename' | grep -e '.php$' -e '.js$')
+    CHANGED_FILES=$(curl -s -X GET -G $URL | jq -r '.[] | .filename' | grep -e '\.php$' -e '\.js$')
   fi
   echo ${CHANGED_FILES}
 }
--- a/CI/Import/Functions.sh
+++ b/CI/Import/Functions.sh
@@ -8,10 +8,10 @@ function get_changed_files() {
 
   if [[ -z ${PR_NUMBER} ]]
   then
-    CHANGED_FILES=$(git diff-tree --name-only --diff-filter=ACMRT --no-commit-id -r ${GH_SHA} | grep '.php')
+    CHANGED_FILES=$(git diff-tree --name-only --diff-filter=ACMRT --no-commit-id -r ${GH_SHA} | grep -e '.php$' -e '.js$')
   else
     URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/files"
-    CHANGED_FILES=$(curl -s -X GET -G $URL | jq -r '.[] | .filename' | grep '.php')
+    CHANGED_FILES=$(curl -s -X GET -G $URL | jq -r '.[] | .filename' | grep -e '.php$' -e '.js$')
   fi
   echo ${CHANGED_FILES}
 }
--- a/CI/Copyright-Checker/copyright-checker.sh
+++ b/CI/Copyright-Checker/copyright-checker.sh
@@ -37,7 +37,7 @@ function is_copyright_valid() {
   local file="${1}"
 
   if ! [ -f "${file}" ]; then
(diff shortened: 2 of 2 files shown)


CI/Import/Functions.sh
```diff
@@ -8,10 +8,10 @@ function get_changed_files() {
 
   if [[ -z ${PR_NUMBER} ]]
   then
-    CHANGED_FILES=$(git diff-tree --name-only --diff-filter=ACMRT --no-commit-id -r ${GH_SHA} | grep '.php')
+    CHANGED_FILES=$(git diff-tree --name-only --diff-filter=ACMRT --no-commit-id -r ${GH_SHA} | grep -e '.php$' -e '.js$')
   else
     URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/files"
-    CHANGED_FILES=$(curl -s -X GET -G $URL | jq -r '.[] | .filename' | grep '.php')
+    CHANGED_FILES=$(curl -s -X GET -G $URL | jq -r '.[] | .filename' | grep -e '.php$' -e '.js$')
```
nicoroeser: The regular expressions in both `grep` calls need escaping of the period character. I suggest to use `grep -e '\.php$' -e '\.js$'` instead.

(This used to be wrong before, but I’d suggest to fix it while we’re at it.)

matthiaskunkel:
Jour Fixe, 17 APR 2023: We highly appreciate this suggestion and accept the PR for trunk and release 8 - but for PHP files only. Please think about how to handle merged JS files (when minimised) with different licence headers (removing licence information is not allowed due to licence).

thibsy:
Hi all,

I _temporarily_ ignore `*.js` files again until I have came up with a solution/concept for checking minified and bundled scripts. Since the PHP-part of this PR has been accepted, I will merge this into trunk and pick for release_8 right away.

Kind regards,
@thibsy

PR #5784, opened 2023-04-06
UI: Reinstate the CSS of Key-Value Tree Nodes

schmitz-ilias:
I noticed that in the trunk the CSS of key-value tree nodes is missing, I guess it got lost when moving everything to sass. This PR primarily reintroduces it again, but I took the opportunity to also make a few other improvements:

- I made a small change to the CSS of key-value nodes that makes it so that the value shrinks before wrapping below the node's label. This is to avoid the tree appearing inconsistent, with the value sometimes to the right and sometimes below the label, depending on length.
- I made some changes to the tree examples allowing me to test their behavior for long values/bylines.
- I changed the CSS class names for the tree according to BEMIT.
- I deleted a leftover `tree.less` in `src/UI/templates/Tree`.

Cheers, Tim


Amstutz:
Hi @schmitz-ilias 

What a nice easter egg :-) This is highly appreciated. Thx you very much. Special thx for improving the examples and migrating to BEMIT along the way!

thx.

PR #5785, opened 2023-04-06
[FIX] UI: fixed invalid character in core.Tooltip.js on line 15.

thibsy:
This fixed some invalid JavaScript code, causing https://mantis.ilias.de/view.php?id=37156 which prevented `Interruptive` modals from being replaced by their `withAsyncRenderUrl()`.

Cheers!


Amstutz:
thx a lot!

PR #5787, opened 2023-04-06
[FEATURE] UI: allow additional inputs for file-dropzones.

thibsy:
Hey all,

In light of the this [feature-request](https://docu.ilias.de/goto_docu_wiki_wpage_5318_1357.html) we need to be able to adjust the inputs of the `File\Dropzone` which appears in container-objects or -headers in the repository.

With the recent refactoring (#5203) this almost became possible, but since we don't want the copyright-selection to appear as a metadata-input for each file entry, we need to be able to pass `$additional_inputs` to the `Form` used in `File\Dropzone`s (or `Modal\Roundtrip` in the end).

We discussed on how this should be handled, because in our opinion, a `File\Dropzone` should expect a `Field\File` input explicitly, otherwise it wouldn't be a `File\Dropzone` anymore. Therefore we came up with the `$additional_inputs` as an optional array. This brings along one issue though: How do we merge the `Field\File` input into this optional array? A possible solution (which we also implemented) is mapping the `Field\File` input to the file-identifier of its `UploadHandler`. We believe with proper documentation (as written in the factory), this should be fine. However, if you have other ideas they're welcome. A nice side-effect of doing this is that now the `Field\File` input is also mapped to a key instead of just some index when retrieving data from `File\Dropzone::getData()`.

Examples are already there, unit-tests will follow.

Kind regards,
@thibsy and @lzehnder

### DIFF
--- a/src/UI/Component/Dropzone/File/Factory.php
+++ b/src/UI/Component/Dropzone/File/Factory.php
@@ -20,6 +20,7 @@ namespace ILIAS\UI\Component\Dropzone\File;
 
 use ILIAS\UI\Component\Input\Field\File as FileInput;
 use ILIAS\UI\Component\Component;
+use ILIAS\UI\Component\Input\Field\Input;
 
 /**
  * Describes a factory for file dropzones.
@@ -59,17 +60,19 @@ interface Factory
      *        Standard dropzones MUST offer the possibility to select files
      *        manually from the computer.
      * ---
-     * @param string $title
-     * @param string $message
-     * @param string $post_url
+     * @param string    $title
+     * @param string    $message
+     * @param string    $post_url
      * @param FileInput $file_input
+     * @param Input     $additional_input
      * @return \ILIAS\UI\Component\Dropzone\File\Standard
      */
     public function standard(
         string $title,
         string $message,
         string $post_url,
-        FileInput $file_input
+        FileInput $file_input,
+        ?Input $additional_input = null,
     ): Standard;
 
     /**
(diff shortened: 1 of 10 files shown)


klees:
Hi @thibsy,

not a review so far, but an answer to your question: Use the power of the monoid. The `Fields` are a monoid. So instead of using `array $additional_inputs` just use `?Field $additional_input`. People can always merge various monoid elements into one via the monoids "add", which is `Field\Group` for the `Field`-monoid. So we can just ask for one field. You could then just pass both `Field`s (the form and the additional one...) to the form.

Kind regards!

thibsy:
thx @klees for pointing this out, monoids FTW.

When adjusting the unit-tests due to the constructor signature change, I also noticed that the rendering tests did never really check for their rendering stubs - always an empty modal has been expected.

I fixed the main rendering test and tried implementing the other rendering tests in another manner according to the feedback in #5152. If this is the wrong direction please let me know. Otherwise, this would be good for review now.

Kind regards,
@thibsy

chfsx:
@klees I resolved the conflicts and updated the tests, please let me know if anything is missing for the merge

klees:
Hi everyone,

I added the JF label, because we can certainly discuss the interface change.

Best wishes!

klees (changes requested):
Hi @thibsy#0350,

LGTM, thanks for adding this. 

Only one fairly miner suggestion. You do not need to follow those, but but please indicate shortly why you prefer to do otherwise:
- [x] Ternary vs. if else: Please use if/else instead of ternary operator [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5787/files#diff-8062c00b69c4b472783ef0caf3e1ef0ca91774051835002f70ac2d03529bbb61R68). There are moments for the ternary op, but this made me stumble, so plain old if/else will be much clearer in situtations like this one.

Intersting way to check for the existence of the additional field, btw. Much appreciated, will suggest elsewhere.

Best wishes!

matthiaskunkel:
JourFixe, 12 JUN 2023: We highly appreciate this suggestion and accept the PR for trunk.

klees:
@thibsy Sry, still some things to do: Unit tests are failing. Please also remember to add testcases to testrail. Thx!

thibsy:
@klees sorry, must have rebased something wrong there. Tests should be fine now, I'll add some test cases on Testrail later today.

klees (approved):
LGTM, thx!

PR #5788, opened 2023-04-06
Add bitwise or, and to formula question

val-pu:
This PR adds bitwise AND and OR operators to formula questions. 

& and | are added as two new binary operators

_Why this addition?_
To simplify the and operation in formulas from
`(-1)^($v1+$v2)/-2+0.5` to `$v1 & $v2`

### DIFF
--- a/Services/Math/classes/class.EvalMath.php
+++ b/Services/Math/classes/class.EvalMath.php
@@ -338,10 +338,9 @@ class EvalMath
         }
     
         $stack = new EvalMathStack;
-        
         foreach ($tokens as $token) { // nice and easy
             // if the token is a binary operator, pop two values off the stack, do the operation, and push the result back on
-            if (in_array($token, array('+', '-', '*', '/', '^'))) {
+            if (in_array($token, array('+', '-', '*', '/', '^','&','|'))) {
                 if (is_null($op2 = $stack->pop())) {
                     return $this->trigger("internal error");
                 }
@@ -363,6 +362,10 @@ class EvalMath
                                             $stack->push(ilMath::_div($op1, $op2)); break;
                                         case '^':
                                             $stack->push(ilMath::_pow($op1, $op2)); break;
+                                        case '&':
+                                            $stack->push(ilMath::_and($op1, $op2)); break;
+                                        case '|':
+                                            $stack->push(ilMath::_or($op1, $op2)); break;
                 }
                 // if the token is a unary operator, pop one value off the stack, do the operation, and push it back on
             } elseif ($token == "_") {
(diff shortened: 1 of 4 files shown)


nicoroeser:
Thank you, @val-pu , for your pull request to enhance ILIAS! I have assigned it to our developers who take care of the Math code for further processing.

Just a quick note: I saw two things in the proposed change:
1. the commit speaks about _logical_ operations, but the added code uses _[bitwise](https://www.php.net/manual/en/language.operators.bitwise.php)_ operators (not [logical operators](https://www.php.net/manual/en/language.operators.logical.php));
2. parameters in the new functions are declared as `int|float` – bitwise or logical operations on floating-point numbers sound a little strange/dangerous to me.

I’m not in charge of the code in question, just leaving my findings. Guido and/or Michael are going to decide on this and comment on this pull request.

Thanks again!

mjansenDatabay:
Hi @val-pu ,

thanks a lot for this PR.

- [ ] In addition to the points @nicoroeser mentioned above, I am not sure why we need `ilMath::_not` here. As far as I can see there is no usage of this method at all. Please remove it.
- [ ] Furthermore please add some **unit tests**.

---------------

@dsstrassner : Maybe we should think about `ilMath` in general. The component is only implicitly maintained and the only (known) consumer is the `TestQuestionPool`.

Best regards,
@mjansenDatabay

dsstrassner:
Hi all, 
I will speak with @mbecker-databay and @kergomard after my vacation.

But right now, I'm not certain that we don't need a Feature Request if we accept this PR, since this would change the formula question.

mjansenDatabay:
> But right now, I'm not certain that we don't need a Feature Request if we accept this PR, since this would change the formula question. 

I agree

mbecker-databay:
Hey! Thank you for the PR!
I am coming from the practical side here. After asking my secret math advisor... can you give us examples for the questions that would require this feature in the formula question? 
(Not lost if you write it here, can be copied to a FR, too.)

Thanks!

val-pu:
Hey! Examples where this feature is extremely useful include: 

- Applying Bitmasks (e. g. extracting LINE, WORD, BYTE bits in caching)
- Calculating the parity of a Hamming-code (An xor function would be useful here)
- Generally in digital logic to calculate the output for a given input
Thank you for your comment!

dsstrassner:
Hello @val-pu, 
I close this PR, since I am the maintainer of the ILIAS Test & Assessment. These changes in the Formula Question need a Feature Request according to the [ILIAS Feature process](https://docu.ilias.de/goto_docu_wiki_wpage_4276_5307.html#ilPageTocA129). 

Since there is no feature request and no JourFixe decision, I must decline this.

Please note, that feature requests are only for new ILIAS versions possible and that there is Feature Freeze for next ILIAS versions.

But thank you for your contribution.

PR #5790, opened 2023-04-10
Fix broken mobile login view

Dennis1993:
How discribed in this bug report here a little fix for it: https://mantis.ilias.de/view.php?id=37097

Before:
image

After:
image

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -18647,6 +18647,13 @@ a#ilHelpClose {
 @media only screen and (max-width: 768px) {
   .ilStartupSection {
     padding-top: 15px;
+    width: auto;
+  }
+  .ilStartupSection .control-label {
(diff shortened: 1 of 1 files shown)


nicoroeser:
Thank you, @Dennis1993 , for your pull request to fix the mentioned issue and make ILIAS better! I have assigned it to our developers who take care of the affected part for further processing. You’ll hear from them soon.

Amstutz:
hi @Dennis1993 

Thx a lot for providing a PR for the issue. However, not that we do not directly write code in to CSS, we use SCSS which is then compiled into CSS. You find the documentation here: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/templates/Guidelines_SCSS-Coding.md

Let me know if I may be of further help.

PR #5792, opened 2023-04-12
Test: fix statistics tab selecting wrong pass (37163)

schmitz-ilias:
This PR fixes [37163](https://mantis.ilias.de/view.php?id=37163). Please have a quick look @fneumann, but I think the behavior should match what you intended in [this commit](https://github.com/ILIAS-eLearning/ILIAS/commit/2e8bcf0d94a22484ecdecd47f65b28c1449e5664).


fneumann:
Hi Tim, I checked the code and understand what went wrong. So your fix looks fine for me. Thanks!

PR #5793, opened 2023-04-12
T&A Solves #36836

jcopado:
With this changes, questions with no errors in the error text are saved as incomplete, and user gets the information message as an error (Please enter the error text. To mark a word as an error please insert a hash sign (#) in front of the word. To mark a group of words as an error please encapsulate the words group between double brackets (i.e.: here is ((an example)) of it). If you press the "Analyze Error Text" button, all marked words will be extracted for further processing.)

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assErrorText.php
+++ b/Modules/TestQuestionPool/classes/class.assErrorText.php
@@ -52,16 +52,17 @@ class assErrorText extends assQuestion implements ilObjQuestionScoringAdjustable
     }
 
     /**
-    * Returns true, if a single choice question is complete for use
+    * Returns true, if an error text question is complete for use
     *
-    * @return boolean True, if the single choice question is complete for use, otherwise false
+    * @return boolean True, if the error text question is complete for use, otherwise false
     */
     public function isComplete()
     {
         if (strlen($this->title)
             && ($this->author)
             && ($this->question)
-            && ($this->getMaximumPoints() > 0)) {
+            && ($this->getMaximumPoints() > 0)
+            && (!empty($this->getErrorData()) || !empty($this->getErrorsFromText($_POST['errortext'])))) {
             return true;
         } else {
             return false;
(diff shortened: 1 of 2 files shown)


dsstrassner:
TechSquad, 13.04.2023: @kergomard will give feedback.

dsstrassner:
Ping @kergomard

kergomard:
Thank you @jcopado for the PR and @dsstrassner for the ping!
This solution does actually solve the issue, you are right. There are few things that make me think, this is not the right solution for the problem yet:

* As far as I can see, the first part of [this statement here](https://github.com/ILIAS-eLearning/ILIAS/pull/5793/files#diff-b2e15ed310911bf8bc33ea76afe7fd2a030a4299c252ef907d54cc7c3adc4c61R65) doesn't do anything at all, as it will never be applied: If there is no `$errordata` `assErrorText::getMaximumPoints()` will return 0 anyways and thus obsolete the next check.
* The second part of the same statement, I'm rather unsure about in the context of the current implementation: `assErrorText::isComplete`, I think, is only called when the question is saved. It is called twice, as far as I can see: Once, when we save the QuestionGenericData and then again, when we call `parent::saveToDb()` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5793/files#diff-b2e15ed310911bf8bc33ea76afe7fd2a030a4299c252ef907d54cc7c3adc4c61R81). I don't think this change adds anything.
* Additionally I don't think we should start using  `assQuestionGUI::writePostData()` in `assQuestionGUI::editQuestion()` (however misnamed this function is). I think this is already done earlier in `assQuestionGUI::save()`.

I will close this PR. I found a solution, by adding the check to the `ilErrorTextWizardInputGUI::checkInput()`.  See [here](https://github.com/ILIAS-eLearning/ILIAS/commit/a465a6ca18a9adba423628434dbcd91bbfe36f3e).

Don't hesitate to give me some feedback, if you think my solution is wrong or if you see side effects.

Thanks again and best,
@kergomard

PR #5794, opened 2023-04-12
fix rector examples for newest code style

iszmais:
A bit picky but at least the header should be fixed imo.

The changes within the functions are just suggestions, if you may have a better approach to that take that.

### DIFF
--- a/CI/Rector/ilUtils/Example.php
+++ b/CI/Rector/ilUtils/Example.php
@@ -1,4 +1,4 @@
-<?php declare(strict_types=1);
+<?php
 
 /**
  * This file is part of ILIAS, a powerful learning management system
@@ -15,7 +15,9 @@
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
- 
+
+declare(strict_types=1);
+
 namespace ILIAS\CI\Rector\ReplaceWithDIC;
 
 class Example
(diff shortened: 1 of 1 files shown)


chfsx:
@iszmais thanks for the hint, i changed the code style and several other thing for the whole directory CI/Rector now and close this PR. The changes in `foo()`and `bar()` are not correct since we want here a ilUtil::sendXY since the rector `ReplaceUtilSendMessageRector` is meant to replace those with `setOnScreenMessage`, the Example Class is to try this rector.

iszmais:
@chfsx Thanks for the response!

LGTM, but (i might be missing something) arent the `sendXY` function removed for ILIAS 8. Why would the example should call an invalid function?

Greetings,
@iszmais

chfsx:
@iszmais the example is to show hoe the rector rewrites your code, you can try this with ```./libs/composer/vendor/bin/rector process --clear-cache --no-diffs --config ./CI/Rector/ilUtils/ilutil_rector.php CI/Rector/ilUtils/Example.php``` 
The rector can be used in plugins as well

iszmais:
@chfsx thank for the explanation!

I assumed that this isnt productive code and i might be extra picky here but even as an example i would still expect this to be executable in some way. Imo that falls under extended documentations. :sweat_smile: 

Greetings,
@iszmais

PR #5796, opened 2023-04-13
Logging: Respect HTTP header and disable console logging for JSON req…

mjansenDatabay:
…uests

This PR adds additional criteria (see also: #4840) to the browser console logging handler determination. If a HTTP client expects a JSON response by sending a corresponding `Accept` HTTP request header, the `ilLoggerFactory` MUST NOT attach the browser console handler. This currently results in a `<script>...</script>` string being appended to the JSON response string.

https://mantis.ilias.de/view.php?id=37167

If approved, this MUST be cherry-picked to `release_8` and `trunk` as well.

### DIFF
--- a/Services/Logging/classes/public/class.ilLoggerFactory.php
+++ b/Services/Logging/classes/public/class.ilLoggerFactory.php
@@ -121,11 +121,20 @@ class ilLoggerFactory
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


smeyer-ilias:
Thanks for fixing this.

PR #5798, opened 2023-04-13
TinyMCE/Logging: Use appropriate HTTP header when requesting JSON

mjansenDatabay:
https://mantis.ilias.de/view.php?id=37167

This PR adds the HTTP `Accept` request header when uploading an image in the `TinyMCE` `ilimgupload` plugin. The client always expects the response to be a JSON document, so it should state this.


kergomard:
Thank you very much @mjansenDatabay ! I merge this and the other two related PRs.
Best,
@kergomard

PR #5799, opened 2023-04-13
ECS: Allow to Add ECS to Kitchensink-Forms

kergomard:
Hi @pascalseeland 
This PR adds an Option to add ECS-Settings as a KS-Input-Section to Settings-Forms. I did this, to enable the move of the general settings of the Test to KS-Forms.

The changes in `ECSSettingsGUI` are just a fix for an issue with types of POST-Values that came to be a problem because of the use of strict types.
The naming in `ilECSObjectSettings` is a little funky for my taste, I tried to find a workable naming, but let me know, if you would like it otherwise.

I tested the functionality with some spoofed data and everything works as far as I can see, but sadly, I don't have an ECS-Server at my disposal.

Thank you very much and best,
@kergomard

### DIFF
--- a/Services/WebServices/ECS/classes/class.ilECSObjectSettings.php
+++ b/Services/WebServices/ECS/classes/class.ilECSObjectSettings.php
@@ -127,8 +127,9 @@ abstract class ilECSObjectSettings
     {
         $export_manager = ilECSExportManager::getInstance();
         $exportable_participants = (new ilECSParticipantSettingsRepository())->getExportableParticipants($a_type);
+        $obj_id = $this->content_obj->getId();
 
-        if (!$this->ecsIsActivatableForObject($export_manager, $exportable_participants)) {
+        if (!$this->ecsCanBeActivatedForObject($export_manager, $exportable_participants)) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


kergomard:
@pascalseeland : Any chance of a feedback on this?

Thank you very much and best!

kergomard:
Hi @pascalseeland 

I would be very grateful if you could have a look at this.

Thanks and best,
@kergomard

Services/WebServices/ECS/classes/class.ilECSObjectSettings.php
```diff
@@ -119,31 +125,11 @@ protected function isActive(): bool
      */
     public function addSettingsToForm(ilPropertyFormGUI $a_form, $a_type): bool
     {
-        $this->logger->debug('Show ecs settings.');
-        if (!$this->isActive()) {
-            $this->logger->debug('Object type is not active. => no settings.');
-            return false;
-        }
+        $export_manager = ilECSExportManager::getInstance();
+        $exportable_participants = (new ilECSParticipantSettingsRepository())->getExportableParticipants($a_type);
```
pascalseeland, line 129: This renames the used variables from exportManager to export_manager, which causes issues in the rest of this method. Same issue exists with exportableParticipants.

Services/WebServices/ECS/classes/class.ilECSObjectSettings.php
```diff
@@ -119,31 +125,11 @@ protected function isActive(): bool
...
 
-        $obj_id = $this->content_obj->getId();
```
pascalseeland, line 128: This variable is used in the rest of this method. If it is removed here, the rest of the usages need to be changed as well.

pascalseeland (changes requested):
Please fix the missing renaming of variables in lines 145, 160 and 180 as it breaks the code.

kergomard:
Hi @pascalseeland 
Thank you very much for your review! you are right and I fixed the issues. I also renamed a function I introduced from `ecsIsActivatableForObject` to `ecsCanBeActivatedforObject`.

Thanks and best,
Stephan

PR #5800, opened 2023-04-13
Services/Logging: Add PSR-3 README. Add `$context` parameter for `ilLogger`

lscharmer:
This PR exposes the `$context` parameter of the monolog logger for the `ilLogger` and placeholder interpolation for a logged message (PSR-3).
A short description on how and why is provided within a `README.md`.

Furthermore this PR replaces all occurrences of an anonymous class to create a dummy `ilLogger` with an actual `NullLogger` class.

See: https://peakd.com/hive-168588/@crell/using-psr-3-placeholders-properly

### DIFF
--- a/Modules/Test/test/tables/ilTestVerificationTableGUITest.php
+++ b/Modules/Test/test/tables/ilTestVerificationTableGUITest.php
@@ -18,6 +18,8 @@ declare(strict_types=1);
  *
  *********************************************************************/
 
+use ILIAS\Services\Logging\NullLogger;
+
 /**
  * Class ilTestVerificationTableGUITest
  * @author Marvin Beym <mbeym@databay.de>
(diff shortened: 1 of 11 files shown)


mjansenDatabay:
I added the `JourFixe` tag, not to push this, but to create awareness how to properly log :-).

klees:
Hi @lscharmer,

thanks for taking care about this. I have one suggestion for a change and one remark:

* `NullLogger`: Please rename `DummyLogger` to `NullLogger`. The class implements the [Null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern) (nice1) and should be called accordingly. We already use that naming pattern in many other locations in ILIAS (`grep -r Null libs/composer/vendor/composer/autoload_classmap.php`) and it will help people if we stay consistent here.
* This `$context` thingy from PSR-3 is not about "sensitive information", as you mention in your comment. It's a security measure that allows logger implementations to escape user input according to a logging backend. [This article](https://peakd.com/hive-168588/@crell/using-psr-3-placeholders-properly) explains the issue nicely. So basically, `$context` should be used whenever user input is passed to the `Logger`.

Greetings!

lscharmer:
Hi @klees
the PR is updated accordingly.

Best regards
@lscharmer

matthiaskunkel:
Jour Fixe, 17 APR 2023: Initiated by a security report, Michael gave a short presentation of the PR and notified us about proper logging according to PSR-3. In case, the PR is accepted by the responsible maintainer, all developers are pleased to have a look into the README and to consider the related recommendations.

smeyer-ilias:
Thanks for fixing this

PR #5803, opened 2023-04-13
UI, JS: import and inject global 'il' namespace

nhaagen:
While fiddling around with il.UI.core, we found that the global "il"-namespace was not always initialized or that the changes to submodules were not applied/available.
This ensures the existence of global il and injects it into the output-function.


klees:
Hi @thibsy and @Amstutz,

IMO this touches the general question where we want to put JS-stuff from ILIAS and how we want to do this. Is this the way?

Best wishes!

Amstutz:
Hi @nhaagen 

Thanks for bringing this to our attention. Like @klees already mentioned, this PR definitely touches the question how we will be handling our JavaScript bundles in the future. Since we understand, that this issue breaks some behaviour we already merge this.

This `rollup.js` configuration would need to be updated for every other bundle as well, to be consistent, but we'd like to publish some further rules or guidelines about this topic in the near future, where we will address this.

Thanks a lot and kind regards,
@thibsy & @Amstutz

PR #5805, opened 2023-04-15
Fix for missing background-color

Dennis1993:
Fix for https://mantis.ilias.de/view.php?id=37096

I added a minimal lighter color for the submenu for better overview.

Before:
image

After:
image


Amstutz:
Hi @Dennis1993 

Thx a lot for this PR. As you can see in the related ticket disccusion, we are working on some revamp here.

Further note that you need to provide SCSS code which is then compiled to CSS to propose CSS changes, see: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/templates/Guidelines_SCSS-Coding.md

PR #5807, opened 2023-04-17
[FIX] UI: deleted files will no longer be considered as values (#37161).

thibsy:
Hi all,

I opened this PR to solve an issue described in https://mantis.ilias.de/view.php?id=37161. In short, the problem is that `File` inputs provided with invalid file-ids (by `::withValue()`) will show "empty" entries on client-side.

To solve this issue I implemented a check for `UploadHandler::getInfoResult()` to skip file-ids which don't exist. We cannot implement this in `Fiel::isClientSideValueOk()`, because this would throw an exception when `false`. Doing so would implicitly add the requirement for `UploadHandler`s to update whatever storage-mechanism is used in the consumers "form-creation-logic", because `File::withValue()` calls would already need to know if files have been deleted before the form is even processed, since the form should be built in the same manner before `Form::withRequest()` is invoked.

With this PR, I also took the liberty of removing the unnecessary implementations of demo `UploadHandlers`, because I noticed they are no longer needed.

@chfsx I also fixed the MainMenu `UploadHandler`, so it does not return empty results if a file does not exist.

Kind regards,
@thibsy


chfsx (approved):
Thanks a lot, seem to be fine!

klees (changes requested):
Hi @thibsy,

I'm quite convinced that this fixes the acute issue, but IMO this does not fix the root cause of the issue. Also, I would expect that the fix will cause more problems down the line and that the root cause also causes other problems.

Let me explain:

In [your nice analysis in the ticket](https://mantis.ilias.de/view.php?id=37161#c92578) you list the steps that lead to the problem. But I think you sorted them wrong:

* 1) file-input receives existing file-id (by ::withValue())
* 2) user removes the existing file-id to upload a new one
* 4) file-input deletes the old file before final submission
* 5) file-input uploads the new file before final submission (not yet: $_POST receives new (existing) file-id))
* 3) user submits
* 5) (already happened: file-input uploads the new file before final submission) $_POST receives new (existing) file-id)
* 6) server builds the same form again, incl. ::withValue() call which is now a file-id that does not exist anymore.
* 7) file-input receives invalid file-id, due to missing validation (UploadHandler::getInfoResult()) the id will be added as dynamic input
* 8) ::withInput() is called because this is a submission
* 9) ::withValue() will be called again with the new POST data, which also contains a file-id, the existing one now.

The problem is this (and I think [@chfsx says something similar](https://mantis.ilias.de/view.php?id=37161#c92569)): We already delete the file asynchronously even before the user even confirmed that she really wants to have this file instead of that file by submitting the form. This also is problematic in other cases: When user uses `[Cancel]` or browsers `[Back]`. There might be more...

This is not a problem when we are adding files. Because then we only have an unused file lying around if the user decides to `[Cancel]`. But deleted files are just gone. Which is exactly what this PR would be hiding. This would not hide the `[Cancel]` or `[Back]`-problem, though.

So I guess the root cause is that we even want to [`getFileRemovalURL`](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/UI/Component/Input/Field/UploadHandler.php#L54) from our users. We cannot really use that URL in a sensible way IMO, because we will only know that a user does not want a file anymore once the POST hits the server. Also note, that even then we, as a UI framework, cannot know that we really can delete the file. Might be that we need to still keep it, because it is already used somewhere else. Hence: IRSS.

So, IMO, the fix from the perspective of the UI framework would be: remove `getFileRemovalURL` from the interface and do not delete files asynchronously. Which mirrors the situation we have in our internal usages (at CaT) in an interesting way: We never implement a handler for removal, to avoid these problematic situations altogether.

At this point I wonder if we even need to expose that `UploadHandler`-machinery to our users. As far as I understand affairs now, we could very well have one `UploadHandler` for the complete UI and only communicate via lists of IRSS-ressources with our users. For legacy storage we could instead use lists of paths to files. Maybe we should discuss this sometime.

Kind regards!

klees:
I close this because #5915 provides an improved solution.

PR #5808, opened 2023-04-17
PermaLink: Remove Permanent Link Service

kergomard:
At the Technical Board we had a look at the PermanentLink-Service as it seemed to us, that it was not needed anymore. I took a closer look and it actually really seems not to have a purpose anymore. It was used in a few places to show the permanent link on the Info-Page, but this is unnecessary, as we always show the permanent link in the footer.

Thus this PR proposes to remove it.


Modules/File/classes/class.ilObjFileGUI.php
```diff
@@ -375,7 +375,8 @@ protected function uploadFiles(): void
         if (self::UPLOAD_ORIGIN_DROPZONE === $origin) {
             $dropzone = new ilObjFileUploadDropzone($this->parent_id);
             $dropzone = $dropzone->getDropzone()->withRequest($this->request);
-            $files = $dropzone->getData()[self::PARAM_FILES] ?? null;;
+            $files = $dropzone->getData()[self::PARAM_FILES] ?? null;
+            ;
```
nicoroeser: :wink: Nitpick: extra line, caused by prior typo? Don’t think it marks an important spot in the code; let’s drop the extra semicolon (the new line 379), shall we?

kergomard: Is gone now!

alex40724 (approved):
Great! Thanks.

kergomard:
@smeyer-ilias , @pascalseeland , @chfsx , and @Amstutz could you please have look at this? I would like to merge it by the end of this week.
Thank you very much and best,
@kergomard

Amstutz (approved):
thx a lot!

chfsx (approved):
Thanks a lot, no objections from my side and thanks for the boyscouting (`;;`)

kergomard:
Ok, as announced, I do merge this now.

PR #5810, opened 2023-04-17
UI: Entity Representation

nhaagen:
Dear all,
this is an implentation for the excellent concept "UI Repository Item - Sorting and Highlighting Properties"
https://github.com/catenglaender/ILIAS/blob/ui/proposal-item-properties/src/UI/docu/UI-Repository-Item_proposal.md

Screenshot from 2023-04-17 14-17-30

Screenshot from 2023-04-17 14-20-19

It is pretty straight forward, but with some things probably worth discussing (please include @klees as well, this is a take at the issue not internally reviewed, yet):

@catenglaender certainly had the main parts in this, so I'll talk of we/us.

Since this is a pretty much direct rival for "Item", we chose "Entity" to parallel developement; also, we thought the name to be more specific in terms of ILIAS objects, but still leaving room for , e.g., user and the like.

First of course, there is the plain [Entitiy Representation](https://github.com/nhaagen/ILIAS/blob/13a153f41c8f5bcc5e4b7912820cc48f87dc38d6/src/UI/Factory.php#L918-L959), in which certain aspects will have to be grouped into Semantic Groups.
In order to properly display labled/unlabled props in an spacekeeping way, but still have CSS-hooks to their parts, we would like introduce a "[Property Listing](https://github.com/nhaagen/ILIAS/blob/13a153f41c8f5bcc5e4b7912820cc48f87dc38d6/src/UI/Component/Listing/Factory.php#L122-L144)" along; it blends nicely into entities.
And last, an Entity Representaion is not be be used standalone -  so we went for an [EntityListing](https://github.com/nhaagen/ILIAS/blob/13a153f41c8f5bcc5e4b7912820cc48f87dc38d6/src/UI/Component/Listing/Factory.php#L109-L120) (as opposed to a table, since the entries may be heterogeneous).

Looking forward to your comments, best regards,
Nils


yvseiler:
Love your work on this, thanks a lot! I like the terms you introduce like "Entity Representation", "Entity Listing" and "Property Listing".

src/UI/Component/Listing/Factory.php
```diff
@@ -106,4 +105,41 @@ public function workflow(): Workflow\Factory;
...
+     * ----
+     * @return \ILIAS\UI\Component\Listing\Entity\Factory
+     */
+    public function entity(): Entity\Factory;
+
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *     Property Listings will list characteristic, labled values in a space
```
klees:
```suggestion
     *     Property Listings will list characteristic, labeled values in a space
```

src/UI/Factory.php
```diff
@@ -914,4 +914,47 @@ public function toast(): C\Toast\Factory;
      * @return \ILIAS\UI\Help\Topic[]
      */
     public function helpTopics(string ...$topic): array;
+
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *     An Entitity Representation displays information about entities within
```
klees:
```suggestion
     *     An Entity Representation displays information about entities within
```

Amstutz:
Hi @nhaagen, 

This is a review only of the public interface, not the implementation.

- [ ] Property Label: Why would I need it, if not visible?
- [ ] Primary Identifier: how can this be a Symbol or Image?
- [ ] Secondary Identifier: how can this be a ShyButton, ShyLink or String, is this not supposed to be a Symbol or Image? Is it really always required?
- [ ] Data Factory Pattern: You use a similar pattern as done in the data table for filling in Data into Entity Listings. I understand why we go this way with the table, since the type of the content is not fully set. However, here I am wondering if this is not overly complicated, here. We could just pass in Arrays of Entities. To me, this is sufficiently typed. Why are you going all the way to the factory? Are you planning on making profit of the lazy evaluation here at some point?

- [x] Document all methods of the public interface. Note that this is a good place to also extend somewhat on the wording introduced in the methods name/and or the composition section. E.g. what are, why would I set featured properties with the Method „withFeaturedProperties“?
- [x] Property Composition: Maybe add, that the emphasis is on the value, not the label.
- [x] Typing 1: Add a type for $value in Property::withProperty. We won't just accept anything as $value, right?
- [x] Typing 2: There are many missing typehints on the main Entity-interface. We understand that these might be missing components in the UI framework. But please still add typehints: Do we accept strings? Maybe LegacyComponents to signal that this could contain HTML? There would be bonus points to sum up which additional components we would want to have, so we can dig into this after this PR is ready.


Please change:
- [x] Wording 1: Drop the term "Representation" from "Entity Representation" in your prose, as it never appear in the PHP-classes. "Entity" would be good enough, no?
- [ ] Wording 2: Overhaul wording of Ferdinands doc according to the wording here. Only do this after we agreed on the wording in the JF.
- [x] Property Listing: CharacteristicValue: Change in Characteristic Value in the rivals section.
- [x] Property Listing as Rival: Add Properties to the rivals of Descriptive Listing
- [x] Entity Composition: All possible properties should be named and summarized here. If too extensive, you might also refer to the background which you nicely listed below for further information and OR the public interface. However, it should be clear where to get the full set of all the properties that can be used, along with an explanation of what they are.
- [x] Please remove the semantics Group png. We should not have binary files in git. The example should do the trick.
- [x] EntityListing\EntityFactory and withData: What does withData do in the EntityListing, what is the factory for? I understood it only by looking at the example. Please provide a better documentation here. Note that you may refer to the example if needed. The working of the two should be clear by only looking at the public interface (and maybe an example if referenced).

thx a lot 
@Amstutz and @klees

nhaagen:
Thanks for the review ;)
I restricted parameters quite a bit, and added some doku. Probably, this is not complete, yet, and should be expanded
according to usecases I'm currently missing. I'd like to discuss with Ferdinand, too, but he's abroad ;(

>  Property Label: Why would I need it, if not visible?
- We don't really need it; so I'd switch label/value around and make it optional.
However, when reading the code, it is very helpful to still have it there and explicitely turn it of (provided, consumers see it the same way and not simply use an empty string). I first thought of aria-support, but then there should not really be a difference for screen-readers: when a value is self-explaining, it still is with readers.
I don't have strong feelings about this, rather a slight preference ;)

> Primary Identifier: how can this be a Symbol or Image?
- I can imagine contexts in which an image is sufficient to identify and distinct entitites; this is for a listing of images, of course, but also, if e.g. courses have good pics/symbols to identify them. Also, when e.g. wrapping an entity in a Tile (in the future), this might be a valid primary id.
 
> Secondary Identifier: how can this be a ShyButton, ShyLink or String, is this not supposed to be a Symbol or Image? Is it really always required?
- We thought of a scenario, where e.g. courses are identified by some logical id, like in university contexts, where subject, program,semester,course and the like are encoded in a string. For the educated reader in the right context, this is a very much better id than a pic. You'd still refer to the course by it's title, though, which then could be the primary id.

> Data Factory Pattern: You use a similar pattern as done in the data table for filling in Data into Entity Listings. I understand why we go this way with the table, since the type of the content is not fully set. However, here I am wondering if this is not overly complicated, here. We could just pass in Arrays of Entities. To me, this is sufficiently typed. Why are you going all the way to the factory? Are you planning on making profit of the lazy evaluation here at some point?
- The factory pattern stresses the uniformity of entities; of course, the pattern does not guarantee it, but it makes it harder to just bundle any entitites. An Entity Listing should always be looked at in regard of user intent, and although different reperesentations of the same entity are entirely feasible, they should not occur in the same context/listing.

best regards,
Nils

src/UI/Component/Entity/Entity.php
```diff
@@ -0,0 +1,130 @@
...
+ */
+interface Entity extends Component
+{
+    //Priority Areas
+
+    /**
+     * Set (and maybe explain) Blocking Availability Conditions when there is
+     * some access restriction for the user on this object.
+     *
+     * @param string|PropertyListing|StandardLink|Legacy|array<PropertyListing|StandardLink|Legacy> $blocking_conditions
```
klees: To simplify this typehint:

* lets drop `PropertyListing|StandardLink|Legacy` because these also could be passed via an array with one element
* lets drop `string` because we don't know much about what would be contained anyway and could just use `Legacy` instead

We could then hint like:

```suggestion
     PropertyListing|StandardLink|Legacy ...$blocking_conditions
```

and drop `@param` in the doc-block completely.

src/UI/Component/Entity/Entity.php
```diff
@@ -0,0 +1,130 @@
...
+        string|array|PropertyListing|StandardLink|Legacy $blocking_conditions
+    ): self;
+
+    /**
+     * Some Properties may be of higher relevance than others, be it as very
+     * significant properties of the entity itself or of greater importance
+     * within the context; you may "feature" them, i.e. put them in a prominent
+     * place.
+     *
+     * @param string|PropertyListing|StandardLink|Legacy|array<string|PropertyListing|StandardLink|Legacy> $featured_props
```
klees: I think this could be improved in the same way like `withBlockingAvailabilityConditions`.

src/UI/Component/Entity/Entity.php
```diff
@@ -0,0 +1,130 @@
...
+     */
+    public function withFeaturedProperties(
+        string|array|PropertyListing|StandardLink|Legacy $featured_props
+    ): self;
+
+    /**
+     * Main Details should provide a quick differentiation or choice on the entity.
+     * "Description" would be one of the most prominent examples.
+     *
+     * @param string|PropertyListing|Legacy|array<PropertyListing|Legacy> $main_details
```
klees: I think this could be improved in the same way like `withBlockingAvailabilityConditions`.

src/UI/Component/Entity/Entity.php
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Users may directly react to the entity, e.g. comment or tag it.
+     * When there are multiple possible reactions, split them e.g. into more
+     * common ones (here) and less often used ones.
+     * Another way of distinguishing Reactions might be the availability/significance
+     * for everybody in contrast to the current user (e.g. rating vs. my favorite)
+     *
+     * @param array<Glyph|Tag> $prio_reactions
+     */
+    public function withPrioritizedReactions(array $prio_reactions): self;
```
klees: Could we restrict this to some (small) amount of properties (maybe 3, check with @catenglaender) like so:

```suggestion
    public function withPrioritizedReactions(Glyph|Tag $prio_reaction1, Glyph|Tag $prio_reaction2 = null, Glyph|Tag $prio_reaction3 = null): self;
```

src/UI/Component/Entity/Entity.php
```diff
@@ -0,0 +1,130 @@
...
+
+
+    //Further Areas
+
+    /**
+     * Reactions that are less prominent than Prioritized Reactions go here.
+     * @param array<Glyph|Tag> $reactions
+     */
+    public function withReactions(array $reactions): self;
```
klees:
```suggestion
    public function withReactions(Glyph|Tag ...$reactions): self;
```

PHP will be able to check the typehints then and we could get rid of `@param`.

src/UI/Component/Entity/Entity.php
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Properties that could potentially limit a users access to the object
+     * belong to this group. If they _are_ actually blocking access, then
+     * you should place them into Blocking Availability Conditions.
+     *
+     * @param string|PropertyListing|StandardLink|Legacy|array<PropertyListing|StandardLink|Legacy> $availability
```
klees: I think this could be improved in the same way like `withBlockingAvailabilityConditions`.

src/UI/Component/Entity/Entity.php
```diff
@@ -0,0 +1,130 @@
...
+     */
+    public function withAvailability(
+        string|array|PropertyListing|StandardLink|Legacy $availability
+    ): self;
+
+    /**
+     * Details provide further information about the entity - worth knowing and helpful,
+     * but not as significantly important than Main Details or Featured Properties.
+     *
+     * @param string|PropertyListing|Legacy|array<PropertyListing|Legacy> $details
```
klees: I think this could be improved in the same way like `withBlockingAvailabilityConditions`.

src/UI/Component/Entity/Entity.php
```diff
@@ -0,0 +1,130 @@
...
+        string|array|PropertyListing|Legacy $details
+    ): self;
+
+    /**
+     * Actions are the things you can actually _do_ with the entity,
+     * e.g. in context of repository items: view, copy, delete, etc.
+     *
+     * @param Shy[] $actions
+     */
+    public function withActions(array $actions): self;
```
klees:
```suggestion
    public function withActions(Shy ...$actions): self;
```

will make PHP check our typehint (and allow us to drop `@param` in the doc-block).

src/UI/Component/Entity/Entity.php
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Personal Status properties indicate the status of a relation between
+     * the current user and the object. A most prominent example would be
+     * the learning Progress of a Course.
+     *
+     * @param string|PropertyListing|Legacy|array<PropertyListing|Legacy> $personal_status
```
klees: I think this could be improved in the same way like `withBlockingAvailabilityConditions`.

src/UI/Factory.php
```diff
@@ -914,4 +914,60 @@ public function toast(): C\Toast\Factory;
...
+
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *     An Entity displays information about entities within the system,
+     *     such as Repository Objects, Users, OrgUnits, etc, when the purpose is
+     *     to represent the entity itself.
+     *     Properties and relations of the entity are arranged in semantic groups
+     *     to structure and prioritize information.
```
klees: Add some sentences about "entity":

Entitys are things were "sameness" is determined by "identity" instead of "equality". A user is an "entity", because it is the same user, even if some property (e.g. the mail address) changed in the meantime. The mail adress is not an entity, because if, e.g., the street changed, it is not the same address anymore. The same thing (or what appears to be the same thing) might be an entity in one situation but not in others: A user (record) is not the same anymore once the address changed.

src/UI/Factory.php
```diff
@@ -914,4 +914,60 @@ public function toast(): C\Toast\Factory;
...
+     *   composition:
+     *     Entities will have a primary and secondary Identifier, which may be
```
klees:
```suggestion
     *     Entities will have a primary and secondary identifier, which may be
```

src/UI/Implementation/Component/ComponentHelper.php
```diff
@@ -173,6 +173,9 @@ protected function checkArgListElements(string $which, array &$values, $classes)
                 if ($cls === "string" && is_string($value)) {
                     return true;
                 }
+                if ($cls === "array" && is_array($value)) {
+                    return true;
+                }
```
klees: I don't think we want this here and I also think we don't need this. The method was originally introduced to deal with PHPs shortcomings in type checking. It should be replaceable by typehints in method/function declaration by now and is only here to accommodate older code, AFAIK. If I'm correct, please create another PR that adds a `@deprecated` to the method and puts the removal on the roadmap. Thanks!

src/UI/Implementation/Component/Entity/Entity.php
```diff
@@ -0,0 +1,279 @@
...
+use ILIAS\UI\Component\Link\Standard as StandardLink;
+
+use ILIAS\UI\Implementation\Component\ComponentHelper;
+
+abstract class Entity implements I\Entity
+{
+    use ComponentHelper;
+
+    /**
+     * @var null|string|PropertyListing|StandardLink|Legacy|array<StandardLink|Legacy>
```
klees: Is this hint really correct? Why can we have an `array<PropertyListing>` in `Entity::withBlockingAvailabilityConditions`, but not here? Also: I think we could reduce this to just a typehint to an array in a similar way then we could reduce typehints on `Entity::withBlockingAvailabilityConditions`. The same seems to be true for other properties here.

src/UI/Component/Listing/Entity/Factory.php
```diff
@@ -0,0 +1,40 @@
...
+     * description:
+     *   purpose: >
+     *     The Entity Listing yields uniform Entities according to a consumer
+     *     defined concept and lists them one after the other.
+     *
+     * ---
+     * @return \ILIAS\UI\Component\Listing\Entity\Standard
+     */
+    public function standard(
+        \ILIAS\UI\Implementation\Component\Listing\Entity\EntityFactory $entity_factory
```
klees: This should not be in the `Implementation` part of the UI-framework. We SHOULD never refer to `Implementation` in public interfaces.

src/UI/Implementation/Component/Listing/Entity/EntityFactory.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\UI\Implementation\Component\Listing\Entity;
+
+use ILIAS\UI\Factory as UIFactory;
+
+abstract class EntityFactory
```
klees: About `EntityFactory`/`Data Retrieval`: I like the pattern and share your argument regarding the usage of the pattern for the `Entity Listing`. But something is very off with the naming here, I had hard time understanding it. We have two similar yet sadly different approaches for the tables. This would add a third, AFAIK. Or I simply do not get it (and Timon and Thibeau apparently also didn't...), which would be bad as well =)

For the `Data Table`, we require our consumers to [define a `DataRetrieval`](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/UI/examples/Table/Data/base.php#L46). We pass a `RowFactory` with a `map`-method to that class, which the consumer can use to create rows for the table. So `DataRetrieval` is controlled by the consumer, `RowFactory` is controlled by the UI framework, `DataRetrieval` is expected to generate a number of rows according to passed specs.

For the `Presentation Table`, on the other hand, we require our user to [define an anonymous mapping closure](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/UI/examples/Table/Presentation/base.php#L24). That mapping closure is expected to fill one row that is passed from the outside with data from some record. The records, in turn, are passed to the table via `withData`.

This here, as far as I understand, would require users to define an `EntityFactory` that is required to generate, probably, `Entity`s from a list of records that was passed from the user via `withData`.

From my perspective, the `Data Retrieval`-pattern from the `Data Table` is better than `withData` from the `Presentation Table`, since it allows the UI framework to control the retrieval of data and also allows for better integration of `View Controls`. So regarding how we pass data to the component, I think we want to end up there, even for the `Presentation Table` at some point.

For the creation of single items in the collection (`RowFactory`, anonymous closure) the affairs are a little different for all three cases imo. For the `Data Table` we implicitly define a mapping via the definitions of the columns, because we want to enforce highly unified rows and control the colums as well. This is a little different for the `Presentation Table` and the `Entity Listing`, because individual items have a higher degree of freedom regarding content and look. We still want to nudge our consumers to some uniformity (as Nils points out), hence one mapping (closure) for a set of data.

At first I though it would be a good idea to completely unify the `Data Retrieval` for various kinds of collections. With @catenglaender|s [observation regarding different use cases of collections](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/src/UI/docu/ux-guide-repository-objects-properties-and-actions.md#user-intent-should-guide-data-presentation) in mind, I came to the conclusion that this is not something that we really want or need. Using a different type of collection (`Data Table`, `Presentation Table`, `Entity Listing`) is not merely a question of showing the same data in a different manner, but instead of accomodating a certain task with the required data and actions. Hence, having different `Data Retrieval`s for different collections (=use cases) is something we actually want. What we do want, though, is similar interfaces so the pattern becomes recognizable.

So IMO, the way to go here would be as such:

* Define an interface `RecordToEntity` with a method `map(UIFactory $ui_factory, mixed $record) : Entity`.
* Require that interface upon creation of an `EntityListing` (as we currently do with this class).
* Define an interface `Mapping` with a method `map(mixed $record) : Entity`.
* Define an interface `DataRetrieval` with a method `getEntities(Mapping $mapping, ?Range $range, ?array $additional_parameters)  : Generator<Entity>`. I would include `Range` to accommodate dynamic queries of data at some point. I would include `$additional_parameters` so we could add `View Control`s and `Filter`s to the `Entity Listing` in a same manner as we do for the table.
* Write a entry for the roadmap that asks for a similar change in the `Presentation Table`.

src/UI/Component/Listing/Entity/EntityListing.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public function withData(mixed $data): self;
+
+    /**
+     * This will call the Listing's EntityFactory::get-method, providing
+     * $ui_factory and $data (see above) as parameters.
+     *
+     * @return \Generator<Entity>
+     */
+    public function getEntities(\ILIAS\UI\Factory $ui_factory): \Generator;
```
klees: I don't think that this belongs to the public interface.

src/UI/Component/Listing/Factory.php
```diff
@@ -106,4 +110,43 @@ public function workflow(): Workflow\Factory;
      * @return \ILIAS\UI\Component\Listing\CharacteristicValue\Factory
      */
     public function characteristicValue(): CharacteristicValue\Factory;
+
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *     The Entity Listing yields Entities according to a consumer defined concept
+     *     and lists them one after the other.
```
klees, line 120: Let's add little more purpose here: reference @catenglaender|s paper and explain how this is different to `Data Table` and `Presentation Table`; name them as rivals.

src/UI/Component/Listing/Property.php
```diff
@@ -0,0 +1,46 @@
...
+        Legacy::class,
+        StandardLink::class
+    ];
+
+    /**
+     * @param string|Symbol|Legacy|StandardLink|array<Symbol|Legacy|StandardLink> $value
+     */
+    public function withProperty(
+        string $label,
+        string|array|Symbol|Legacy|StandardLink $value,
```
klees: Why do we want to support arrays of these thingies here? Isn't a single (e.g.) Symbol enough?

src/UI/Component/Listing/Property.php
```diff
@@ -0,0 +1,46 @@
...
+use ILIAS\UI\Component\Legacy\Legacy;
+use ILIAS\UI\Component\Link\Standard as StandardLink;
+
+/**
+ * Interface Property
+ * @package ILIAS\UI\Component\Listing
+ */
+interface Property extends Listing
+{
+    public const ALLOWED_VALUE_TYPES = [
```
klees: Why do we need this on the public interface?

klees (changes requested):
Hi @nhaagen,

I only looked at the public interfaces so far, as this is what we need to be approved by the JF next. Also, the implementations should be mostly uninteresting anyway, I guess. Nothing fancy there.

I general this looks good, but I also think this shows some open issues (that we won't be able to solve here...):
* If the [proposal for text handling](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/text-handling.md) was implemented, it would have helped a lot.
* Some unified concept for the Properties (that gets rid of that clunky union types) seems to be missing, something we could also use for the Tables.
* Some unification regarding `DataRetrieval` and mapping to particular entities are tbd.

I added my comments inline this time, it simply was to much for me atm to handle the review with our standard format. I hope that is ok. All comments are to be considered as strong SHOULDs. You know how to contact me if we need to clarify anything =)

Best regards!

nhaagen:
Hi @klees,

thanks for the detailed hints; I'm using splat operators now in favor of array|Thing widely,
and it certainly suites the interface.
Also, there is some more docu now and the data-retrieval/mapping is split up.

Best regards

klees (approved):
IMO this is good to go for the JF to discuss the interface. I would kindly ask @thibsy and @Amstutz to have a look into this before moving it onto the JF agenda. Thanks!

Amstutz:
The public interface seems ready tbd at the upcoming jf. The implementation will need another detailed review.

matthiaskunkel:
Jour Fixe, 10 JUL 2023: We highly appreciate this suggestion and accept the PR for trunk. We are looking forward to the first implementation of this UI element.

klees:
@nhaagen Please rebase!

Amstutz:
Hi @nhaagen

Thx for the implementation of this important component. Special thx for adding the type hints for the arrays in the doc comments.

- [x] Code Style I: You chose to use `switch(true)` instead of an `else if`clause. There is nothing wrong with it, but it adds an indentation, why do you prefer this way?
- [x] Code Style II: You use the function `is_a` instead of the language construct `instanceof`. Why? In doubt, we would prefer language constructs over functions.

Change or indicate why rather not:
- [x] Listing: Consider changing the `entitylsiting` dom to an actual, unsorted list. This would allow for much quicker navigation for people with screen readers.
- [x] Code Style III: There seems to be some inconsistent spacing at various places (e.g. `type1|type2|...` vs `type1 | type2 | ...` Did you run the cs fixer? If not, do so. If already done, best manually look through the spacing in the code again.

thx a lot!
@amstutz and @thibsy

nhaagen:
Hi,
very good point using a list.
The fixer does not support the spaces around union types, yet, but I hope I got them all.
Best regards,
Nils

PR #5811, opened 2023-04-17
UI: Input View Controls

nhaagen:
In https://github.com/ILIAS-eLearning/ILIAS/pull/5656, I promised to have a look into [Viewcontrols](https://github.com/nhaagen/ILIAS/blob/d7cb1150423f85c01481efc5ce8761d0661ae39f/src/UI/Component/Input/Container/Factory.php#L178-L207) for the Data Table, since we will at least need [Pagination](https://github.com/nhaagen/ILIAS/blob/d7cb1150423f85c01481efc5ce8761d0661ae39f/src/UI/Component/Input/ViewControl/Factory.php#L71-L99), [Column selection](https://github.com/nhaagen/ILIAS/blob/d7cb1150423f85c01481efc5ce8761d0661ae39f/src/UI/Component/Input/ViewControl/Factory.php#L27-L48) and [Order](https://github.com/nhaagen/ILIAS/blob/d7cb1150423f85c01481efc5ce8761d0661ae39f/src/UI/Component/Input/ViewControl/Factory.php#L50-L69).

As to keep it unobstrusive, I went for glyphs opening dropdowns; also, @yvseiler and @catenglaender were heavily working on the looks of viewcontrols, and this is incorporated here.

Screenshot from 2023-06-14 14-32-24
Screenshot from 2023-06-14 14-31-11
Screenshot from 2023-06-14 14-30-57

The idea of Input View Controls is to retrieve Range, Order and some kind of aspect selection from a bundle of Input Controls;
this will help to move Data Table forward.  If accepted, I'll bundle this up with #5656.

Screenshot from 2023-06-14 14-34-49

Best regards,
Nils


yvseiler:
Thanks a lot, it looks great. 

- For the glyphs for "Numbers of Entries", "Sortation" and "Display Columns" I could take them into our "Task Icon Force"-group to find a suitable visual symbol for the "non labeled" controls. 
- What do you think about whether it would make sense in the meantime to include in the description of the ViewControls that they may be represented with Label or with Glyph? This would also fit in with the FR about the [ViewControl Mode in the Dashboard](https://docu.ilias.de/goto_docu_wiki_wpage_7503_1357.html) (not part of this commit here). 

Best regards
Yvonne

klees:
I have reviewed this thoroughly (much change...) and will discuss certain aspects with @nhaagen tomorrow. We will then provide an updated version of the code.

klees:
I thought I pressed "Cancel Review", now my notes have been on the internet...

klees (approved):
Hi @Amstutz, hi @thibsy,

I will now pass this on. As this became so massive, I want to give you some assistance to digest this.

* Relevant changes in the public interface are [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5811/files#diff-1971b624539f066ac2e566130c38f2504fea5b1ba97536f1cfed85418432e7d9R53). They should not be suprising, as we simply add two missing View Control Inputs here. If you'd be so kind to greenlight them for the JF, we could move on regarding the public interface.
* There are implementations for the View Control Input (e.g. [Rendering](https://github.com/ILIAS-eLearning/ILIAS/pull/5811/files#diff-9957152d868b16ebd0847046247129f1a1ada960ee8ff8dd5ac90e194740ec4cR29) and auxiliary stuff in the expected locations. Much of this doesn't seem to be too interesting.

The interesting part is the rest of the changes: a huge refactoring, basically. General goal was to tighten the integration of View Controls into the framework of `Inputs`, align principles and reuse code. Most notably:

* Pull common code for `Input` and `Container` up one level to reuse it for the `ViewControl`s and their `Container`. This introduced a massive `Input` -> `Field` renaming, but the naming here was strange to begin with.
* Make the `Container` for the `ViewControl`s behave and feel similar to the other one for `Field`s. But use query params and GET instead of POST to retrieve the data. This is something we will want to repeat for the `Filter Container`, I would assume.

I'm sorry that this became so huge, I was also sorry for myself while reading this. I didn't anticipate the amount of refactoring that became necessary to make this nice. If there is anything we can help to understand or digest this, contact me, as always.

Best wishes!

klees:
On little addendum. This is part of the general quest to make the complete Data Table happen. Next steps in this quest, from my perspective, will be:

* integrate this with the [URL Builder](https://github.com/ILIAS-eLearning/ILIAS/pull/5938) and the current data table implementation
* solve [multi action modal problem](https://github.com/ILIAS-eLearning/ILIAS/pull/5985)
* clean up javascript
* profit

Best wishes!

klees:
Does anyone understand why the normal checks don't run?

Amstutz:
There was a conflict in the css map. I resolved this, now they are running. However, note that we will have many more conflicts in the classmap, since this is a one-liner changing with every css change. We might thinking about excluding that in git.

catenglaender:
This is probably a discussion for elsewhere, but: Could the server or a Github action generate the CSS map? Or actually do the entire SCSS compile of the CSS and the map (although the CSS diff on commit is a great opportunity to see if something went wrong and has unexpectedly large consequences). This would solve the problem of contributers using outdated compilers.

klees:
@catenglaender Yes, this would be really nice =)

thibsy (changes requested):
Hi @nhaagen

I will use the letters `C` and `I` as an abbreviation for the `ILIAS\UI\Component` and `ILIAS\UI\Implementation\Component` namespaces throughout this review when referring to classes.

For now you only need to concern about the interface changes (`C\*`), since this has yet to pass JF. However, I already did the technical review as well because I needed to dive into this topic to evaluate the public interface. (I tried sorting the points accordingly)

- [x] `C\Input\ViewControl\Factory`: the interface currently requires consumers to provide a label. As far as I can see, these labels will only ever be used as aria-labels, right? Now, if there is a use-case where consumers must have the possibility to use custom aria-labels, we should structure this similar to how `C\Modal\Modal`s implement this and use optional (nullable) labels, where the renderer will set the default translation. IMO, we can get rid of this logic altogether, because the consumer should not be concerned about aria-labels in general. This would mean, moving the constants to the renderer and removing the properties and methods from the inputs. Or do I overlook a specific use-case here?
- [ ] `C\ViewControl` namespace: could you elaborate what happens to the `Pagination` and `Sortation` in the other view-control namespace? Are the new components to replace these inputs? Do they co-exist, or do we need to refactor current usages? Also the documentation on the `C\ViewControl\Factory` about these components are somewhat more extensive. Can we maybe adopt some of this (with according changes)?
- [ ] `I\Input\Container\ViewControl\Renderer`: this renderer currently uses a mechanism which is not quite clear to me, using `getComponentInternalNames()` to sort of transform `C\Input\ViewControl\ViewControl` inputs into `C\Input\Field\Field` inputs. IMO we can achieve the same result here by reusing the `C\Input\Field\Hidden` component. You could either create a new input by passing along infos in the renderer, or even implement view-control inputs as a decorator for hidden inputs. Doing so would allow the renderer to only bother about container stuff and delegate along the rendering chain to the the view-control inputs, which would then render said hidden input. I think this would also scratch the need for `I\Input\Container\ViewControl\ViewControl::getRequest()`. Or am I missing something?

Please consider the following suggestions. You do not need to follow them, they are mostly here to document general improvements, not required for a merge.

- [x] `I\Input\ViewControl\ViewControl::getOnChangeSignal()`: do I understand correctly, that a view-control-input must always receive the containers submit-signal in order to work? If so we could throw a `LogicException` here.
- [ ] `I\Input\ViewControl\Renderer::renderPagination()`: you are currently cloning the internal `Signal` to provide it with different options for every page. This could be avoided, by implementing some minimal JavaScript, which initializes the pagination by passing the signal _once_. You could then render data-attributes for the page-buttons, which trigger said signal with the according data gathered from the HTML element(s) on click.

- [ ] `C\Input\Input` hierarchy: there is currently a specific hierarchy of input-interfaces* (see at the bottom), of which many are located in `Field` namespace. Since we are already at it and this is somewhat confusing, I recommend to move these interfaces into a more fitting namespace as well. Please also note the naming-change, as they don't need to be overly specific anymore:
	- [ ] `I\Input\Field\InputInternal` -> `I\Input\InputInternal`
	- [x] `C\Input\Field\FormInput` -> `C\Input\Container\Form\Input`
	- [ ] `I\Input\Field\FormInputInternal` -> `I\Input\Container\Form\InputInternal`
	- [x] `C\Input\Field\FilterInput` -> `C\Input\Container\Filter\Input`
	- [x] In addition to moving these interfaces, I think we should also break the link between the public and internal interfaces. This was pretty hard to follow. The implementation should rather implement the internal interfaces explicitly.
- [x] `C\Input\ViewControl\Sortation`: the way a consumer needs to pass `$options` to this view-control feels weird, since they need to concatenate their field-id, some magic separator (`:`) and a direction of `Order`. IMO this relies heavily on the consumer how this logical string is represented and thus making future refactoring hard. I would prefer some dedicated object, wrapping the field-id to an `Order` and maybe even the representable string.
- [ ] `C\Input\Field\Factory`: please import the moved `C\Input\Input` interface with an according `use` statement, so annotations will provide proper auto-completion.
- [ ] `C\Input\Field\HasDynamicInputs`: the same here.
- [x] `C\Input\ViewControl\Pagination`: please remove the commented method and provide a brief comment above each method to document the behaviour and intention.
- [x] `C\Input\ViewControl\Factory::fieldSelection()`: please add the missing `$button_label` to the PHPDoc as well for documentation reasons, if we don't remove it (see question one).
- [x] `C\Input\ViewControl\Pagination::withNumberOfVisibleEntries()` please rename the parameter to e.g. `$amount` or `$number_of_entries`.
- [x] `C\Input\ViewControl\Pagination::withTotalCount()`: why is the parameter optional? If we want to reset the count I'd prefer to pass in null explicitly. The functionality to reset the total might even be dropped, because the consumer would simply not call the method if he didn't want a total.
- [ ] `I\Input\Input`: this class does not implement the interfaces `FormInput` and `InputInternal` anymore. I think this is correct for `FormInput`, since this implementation or abstraction should now be located in `Field`. `InputInternal` however, is missing. Please implement the missing interface and remove the methods from `FormInput`.
- [x] `I\Input\ViewControl\Pagination::withLimitOptions()`: please use the component-helper to check the array for integer values, so you can add an according PHPDoc annotation and don't need to map `intval()` in the renderer.
- [x] `I/Input/Container/ViewControl/ViewControl::__construct()`: please inject an instance of the data-factory instead of creating a new instance. Please also use property promotion whenever possible.
- [x] `I\Input\Container::withRequest()`: declaring this method abstract is obsolete, since it is already on the interface. Please remove it.
- [x] `I\Input\Container::isSanePostRequest()`: this is something we got rid of in the meantime, please remove this as well.
- [x] `I\Input\ViewControl\Renderer`: there are several PHPDoc annotations where `\ILIAS\Data\Range` was not properly hinted (missing first `\`).
- [x] ViewControl examples I: please wrap the `print_r()` statement in `<pre>` tags and put it before (above) the actual component to keep our examples consistent.
- [x] ViewControl examples II: please retrieve your dependencies by the corresponding methods, which are type-hinted, instead of array-keys.
- [x] `VC*` prefix: please don't use abbreviations when naming things, unless it is a known acronym, which VC is not. Please rename your template-files and unit-tests accordingly.

Kind regards,
@thibsy

---
```
C\Input\Input
├─ C\Input\ViewControl\ViewControl
└─ C\Input\Field\Field
	├─ I\Input\Field\InputInternal
	└─ C\Input\Field\FormInput
		├─ I\Input\Field\FormInputInternal
		└─ C\Input\Field\FilterInput
```
*Hierarchy of input-interfaces

klees:
Hi @thibsy,

thanks a bunch, lots of extremely valuable stuff in there. Some small disagreement from me:

* `C\Input\ViewControl\Factory::fieldSelection()`: IMHO we should rather remove the docstring for `$label`, as we should only ever add docstrings if they provide additional information over the code that we write anyway. Hence, for a thoroughly typed parameter with a good name, we mostly don't need a docstring.

Best wishes!

nhaagen:
Hi @thibsy (and all),

again, thank you for detailed comments. 
Let me answer the questions first: 

> C\Input\ViewControl\Factory: the interface currently requires consumers to provide a label.

This was due to the input-nature of ViewControls (there is a label on Input),but I now used the pattern from Modals.

> C\ViewControl namespace: could you elaborate what happens to the Pagination and Sortation in the other view-control namespace?

I think that Input View Controls are superior, especially when using more than one Control; for the latter ones, parameters have to be manually aligned and synchronized.
However, the older Viewcontrols are still being used, and there is at least the ModeControl that is missing here.
Nontheless, I'd rather answer a 'yes' - probably, the old ones should go. I have no current plans of removing them, though.

> I\Input\Container\ViewControl\Renderer: this renderer uses getComponentInternalNames() ...

This is to retrieve the parameters controlled by the Container and tell the remaining ones apart; when passing a form via GET, the query-part of the forms's action is being replaced;
I add those params as hidden fields (not Inputs, in orer to preserve their names) and pass them along by submitting the form.


Changes
> C\Input\Input hierarchy

The Input-Interfaces were just hardcore. 
As you said, I moved the the required Inputs to their respective Container and renamed the internal interfaces and abstract classes to clarify and reflect hierarchy. 
I left the internal interfaces in the Implementation, though, as they have more depencies in implementation like InputData, Result, ... and I didn't want that in the Interface section. 

> C\Input\Field\Factory: please import the moved C\Input\Input interface

I couldn't find Input\Input being used; am I missing something?

> I\Input\Input: this class does not implement the interfaces FormInput and InputInternal anymore

It should not. The ViewControls are Inputs, but not real FormInputs.


I guess, there is the one or other round to go, but for now, I'd like to bounce this back.
So, thanks again, and best regards, 
Nils

thibsy (changes requested):
Hi @nhaagen

I am sorry for the delay here, I could only work on this in intervals and I didn't want to rush it. 

The refactoring of the input-hierarchy was indeed very hardcode, which is why I decided to implement some of the changes I would have requested myself, to ease to process for the both of us =). I think @klees or @Amstutz should have another look at this, since I can be considered a co-contributer now. However, I still have some changes which do not concern the input-hierarchy, which I'd like to cover, before this goes to JF.

Before I dive into these points, I will briefly explain the changes I have made. Note I will use the same abbreviations for the namespaces as before:

- I got rid of the `C\Input\Field\Field` interface because it merely extended from `C\Input\Container\Form\FormInput` and was therefore obsolete. In combination I also merged the abstract classes into one, which implements `C\Input\Container\Form\FormInput` directly.
- I moved `$label` and `$byline` attributes down to the `C\Input\Container\Form\FormInput` implementation, along with the `...Required()` and `...Disabled()` methods. They were not required by `C\Input\Input` but were implemented on this level. This lead to a benefit for `C\Input\Container\ViewControl\ViewControlInput`s, which can now extend the base input way more easily.
- As you may know, the inputs are structured in a monoid-ish manner, where creating a `C\Input\Field\Group` can be considered a monoid operation. The introduction of `C\Input\Container\ViewControl\ViewControlInput`s made it possible, to create a group which consists of multiple container-specific inputs, which violates the rule where the operation on two elements of the same type result in another element of the same type. Therefore, I applied some strictness to this structure by making `C\Input\Field\Group` only accept `C\Input\Container\Form\FormInput`s and introducing a new view-control `C\Input\ViewControl\Group` that only accepts `C\Input\Container\ViewControl\ViewControlInput`s. If filter-inputs are refactored to fit this structure, another group would be needed. The strictness is achieved in the constructor of these groups, where nested inputs are checked for according instances. 
- This lead to an issue in the `I\Input\Container\Container` abstraction, because the construction can no longer use one group for all kinds of inputs. I solved this by injecting the according input factories in the corresponding containers. I could not use the `I\Input\Factory`, because this would have led to a dependency cycle. In the same time, I centralized the `C\Input\Container\Container::withRequest()`, which uses a new method to let derived container extract the corresponding `I\Input\InputData`.
- While refactoring the groups, I also transformed the `C\Input\ViewControl\Sortation` and `C\...\Pagination` view-controls into proper decorator for `C\Input\Field\Group`s. I was kind of forced to do this, because this no longer worked because of the increased strictness. To indicate this, I introduced an internal interface `I\Input\ViewControl\HasGroupInput`, which is used in the `I\Input\Container\ViewControl\Renderer` to determine the component internal names now. However, this should definitely be improved in the future, by making `C\Input\Field\Hidden` implement `C\Input\Container\ViewControl\ViewControlInput` as well and transform these two view-controls into actual `C\Input\ViewControl\Group`s.

These are the most important changes. I might have made some small improvements in other places along the way, which are not notable.

---

Now to the review-part of this feedback:

- [ ] `C\ViewControl` namespace: your feedback sounds plausible to me, but I think we should already deprecate the replaced ones for e.g. ILIAS 10 then. Could you add an according deprecation on the factory? We can discuss the exact timeline at the JF.
- [x] `I\Input\Container\ViewControl\Renderer`: this seems to be another issue caused by having no real routing, otherwise we could have avoided this and used a normal form-action instead. I believe this mechanism is going to stay for a while then. Thanks for your explanation.
- [x] view-control labels: I think consumers of view-controls should not need to concern about a11y too much, so I would recommend to remove the following methods. If there is a good reason or a use-case I am missing, let me know, I would ask you to document the methods accordingly then. If the methods are only there for internal use, e.g. components rendering view-controls internally, we should remove the methods from the public interface, if other components really need to change these values.
	- [x] `C\Input\ViewControl\Pagination::withLabelOffset()`
	- [x] `C\Input\ViewControl\Pagination::withAriaLabelLimit()`
	- [x] `C\Input\ViewControl\FieldSelection::withAriaLabel()`
	- [x] `C\Input\ViewControl\Sortation::withAriaLabel()`

- [x] ViewControl examples I: please retrieve your dependencies by the corresponding methods, which are type-hinted, instead of array-keys. So e.g. `$DIC['ui.factory']` becomes `$DIC->ui()->factory()`.

Kind regards,
@thibsy

nhaagen:
Hi @thibsy,
thank you very much for the Input-Revision; this is certainly a tough topic, and I have a slight feeling that we will touch it again. However, your changes look fine to me, and I rebased and incorporated the parallel changes of other PRs. 
Since this here reaches into quite a lot of actual off-topics (forms, CSS, unittests, ...) on the one hand, but simultaneously introduces dependencies for others, I'd like to get it integrated as soon as possible. 

After all, I think that we have successfully made the point of ViewControls being Inputs,
and the general input/form-structure may easily be looked upon separetely, if need arises.

So, thanks again for your patience and commitment in this.

If @Amstutz does not object, I think this is ready to go. Maybe we (Thibeau, Timon, Nils) should gather for a joined look - it might make it a little easier to dig into the quite heavy changes?

Best regards,
Nils

thibsy (changes requested):
Hi @nhaagen

Thanks a lot for implementing the requested changes and for fixing the bug with groups, I have introduced :).

Before I approve of this PR, there is still one issue that occurred to me. Otherwise, this is ready for the the next JF.

- [x] `C\Input\Container\ViewControl\ViewControl::withRequest()`: this method call is currently mandatory for view-control containers to work, because without this call the `I\Input\Container\ViewControl\Renderer` will not generate hidden-inputs for query-parameters of the current request. This is a problem, because this leads to a request with **only** the view-control inputs as query-parameters, which would be an invalid request for ILIAS (because there is no static routing). IMO there are two ways to go:
	- [ ] good: removing `C\Input\Container\Container::withRequest()`, so containers can "decide" whether or not providing a request is optional, ideally move the method to a separate interface containers may extend. Then we could inject the request in `I\Input\Container\ViewControl\ViewControl::__construct()` directly, passing it down from the initialization.
	- [x] fast: throw a temporary `LogicException` when rendering an `C\Input\Container\ViewControl\ViewControl` without request and document this behaviour on the factory. However, this is merely treating the symptom, we should go for something similar like option A sooner or later.

Either option is fine for the moment, I leave it up to you how to handle this.

Kind regards,
@thibsy

nhaagen:
Hi @thibsy,
thank you very much for pointing that out. I went for the exception, but not because of time or difficulty, since the interface for an OptionalRequest is quite simple and straight forward, imo. However, having the request in the constructor of a Component seems to be a topic for a broader discussion.
I'm reluctant to introduce this dependency on Component-level, imo this would rather belong to the renderer.
Maybe extending the ComponentRenderer with a function to retrieve the request would be an option? 
Best regards,
Nils

thibsy (approved):
Hi @nhaagen 

Thanks for implementing the change so quickly and for sharing your concerns. I agree, this is probably something we want to discuss in more detail. It would also make this PR even bigger, which is not my intent, and the temporary solution is perfectly fine for this already large first iteration.

IMO this can be presented at the next JF, I therefore labeled this PR accordingly. 

Thanks again for your patience, and your effort!

Kind regards,
@thibsy

matthiaskunkel:
Jour Fixe, 21 AUG 2023 : We highly appreciate this PR and accept it for trunk. Please merge.

nhaagen:
Hi @thibsy - should we ship it? ;)

PR #5813, opened 2023-04-18
MD: move MD administration towards KS

schmitz-ilias:
This PR makes some first steps to move the Metadata Administration to the KS:

- I split up the Copyright tab, currently consisting of a small form with a table below it, into two subtabs for form and table respectively.
- I moved the Edit Copyright and Add New Copyright dialogues to asynchronoulsy rendered modals. This included moving the Add New Copyright button from legacy table to toolbar, and the action buttons in the table to a KS dropdown.
Note that I was only able to make asynchronous error handling of the forms in the modals work by using `Services/Repository/js/repository.js` (fyi @alex40724). I expect this to be a temporary solution, it should be possible to get rid of this dependency again in due time as improvements to the KS are made.
- I refactored the classes involved a bit, in order to prepare for further improvements.


smeyer-ilias:
Thanks, @schmitz-ilias

PR #5816, opened 2023-04-19
UI: Add optional custom constraint to Input::withRequired()

tbongers-cat:
The constraints for Inputs that are marked `withRequired(true)` are currently only visible inside the code of the individual implementations and differ depending on the type of field. To give developers better control on which constraints are checked, we have added a new parameter to `withRequired()` that allows us to REPLACE the default constraint with a custom constraint (instead of just being able to ADD more constraints with `withAdditionalTransformation()`).

The two added examples should make it more obvious how `withRequired()` is working with and without an additional constraint. Looking forward to your feedback!

_( This is based on a proposal by @klees )_

### DIFF
--- a/src/UI/Component/Input/Field/FormInput.php
+++ b/src/UI/Component/Input/Field/FormInput.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,12 +15,14 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+declare(strict_types=1);
 
 namespace ILIAS\UI\Component\Input\Field;
 
 use ILIAS\UI\Component\JavaScriptBindable;
 use ILIAS\UI\Component\OnUpdateable;
 use Closure;
+use ILIAS\Refinery\Constraint;
 
 /**
  * This describes inputs that can be used in forms.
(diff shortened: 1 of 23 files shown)


Amstutz:
Hi @tbongers-cat

Thx a lot for adding this. Special thx for adding additional documentation on the public interface. Special thx also for the additional examples and tests.

One Question:
- [x] Extend withRequired: You opted for extending withRequired with one parameter. Another option would have been to add something like withConstraintForRequirement. Can you elaborate somewhat on why you chose the first one?

One thing remaining to do:

- [ ] Update the Testrail Cases for the new examples.

Thx a lot

tbongers-cat:
Hi @Amstutz ,
thanks for your feedback. 

The functionality of `withRequired()` currently consists of two things: A marking for the required input in the output (red asterisk) and a (hidden) default constraint that is applied to it. These two functions can not be separated right now, as there is no way to just "flag" a field for being visually marked (while the constraint could theoretically be added as `withAdditionalTransformation()` if the default constraint did not exist).  

Because of this I opted for adding the custom constraint as a parameter, which changes the second part of withRequired()'s behaviour while keeping the other one. Making it an optional parameter also keeps the convenience of just using the already provided default constraint of an input. 
Using an additional `withConstraintForRequirement()` feels a bit wrong to me, as it isn't quite clear where the constraint came from in the first place. Besides that, `getConstraintForRequirement()` is currently protected, so adding `withConstraintForRequirement()` would make this internal property public, which imho is not needed.

Of course I'm open for you or others to convince me otherwise =)
I'll update the testcases accordingly once we decided on the way to go.

tbongers-cat:
From my side, this should be ready to go if the parameter for `withRequired()` is okay for everybody. In that case, I'll update the test cases as well.
~~The failed copyright checks are restricted to the examples I added, which should not have a copyright header for better readability afaik. If these need to be formatted differently, please let me know.~~ _(fixed)_

Amstutz:
Thx @tbongers-cat 

Looking good. We will need to shortly present the change at the upcoming JF due to the public interface change.

thx.

matthiaskunkel:
Jour Fixe, 15 MAY 2023: We highly appreciate this suggestion and accept the PR for trunk.

PR #5817, opened 2023-04-19
PDFGeneration: Remove `phantomjs`, and with it the setup agent 

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=31683

If approved, this has to be changed in `release_8` as well.


gvollbach:
@mjansenDatabay Thx. looks good.

PR #5818, opened 2023-04-19
VirusScanner: Add/Adjust config values to setup status

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=30368

If approved, this has to be cherry-picked/integrated in `release_8` and `trunk` as well.


gvollbach:
@mjansenDatabay Thx! Looks good to me.

PR #5819, opened 2023-04-19
Bugfix 0037173: Column layout in help slate

ezenzen:
Hi,

the online help learning module is to be revised and made visually attractive. The goal of the review is that the learning module is not only used in the help slate, but also used as a manual/handout. To do this, the pages are restructured using column layouts. f I create a column layout with 4 columns, they are displayed side by side on large screen. On mobile devices the columns are displayed below each other. This is also true for the Help Slate.

However, the size of the help slate ALWAYS corresponds to the mobile view, regardless of the screen size of the end device. For this reason, columns within the slate should always be displayed one below the other. 

See discussion: https://mantis.ilias.de/view.php?id=37173

Best regrards, 
Enrico

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,7 +40,6 @@
 /Customizing/clients
 virtual-data
 /libs/composer/vendor
-libs/bower/bower_components
 /node_modules/**/.cache
 
 # /dicto
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@
 /Customizing/clients
 virtual-data
 /libs/composer/vendor
+libs/bower/bower_components
 /node_modules/**/.cache
 
 # /dicto
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -15942,6 +15942,56 @@ div.il_SearchFragment {
 #ilHelpText .ilc_list_u_BulletedList .ilc_list_o_NumberedList {
   padding-left: 15px;
 }
+#ilHelpText .col-xs-1,
+#ilHelpText .col-sm-1,
+#ilHelpText .col-md-1,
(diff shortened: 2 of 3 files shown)


Amstutz:
This fix results from the workflow of using a learning module for generating content to be displayed in different contexts. In one, those columns seem to be an acceptable way of displaying the content, in the other (here slate), the columns are not wanted. 

We are not certain that CSS is the way to resolve this issue. In our opinion, this should be handled, by whatever is rendering this output here. The DOM seems just not fit the context here. Note that this is probably not the last issue we will face from this usage of a learning module for different purposes.

We should discuss on how to proceed in the next JF.

matthiaskunkel:
Jour Fixe, 15 MAY 2023: We see the need to use the Online Help learning module for the online help and for documentation purposes. But we prefer a clear solution for the problem of displaying an ILIAS learning module in the slate. The current suggestion is too much "hacky" for us. A concept workshop for a feasable solution would be highly appreciated.

Amstutz:
Closed, see JF decision.

Personal Comment: The current display in the Slate has not only layout implications, but a11y implications as well (it is hard to make screen readers be aware, that the trigger in the mainbar opening the slate, changes the slates content). Maybe we can again think about moving help content to external tabs and so also reduce the complexity of slates. If so, we would need to look into external control of tabs in order to context-sensitive display of the newly opened slate, see: https://stackoverflow.com/questions/4931308/how-to-control-one-window-from-another-in-html5

atoedt:
Currently the help slates does capture the focus upon clicking the help button. The screen reader user cannot miss the help slate.  This behavior does not need changing or improvement. The help was BITV tested successfully with the trigger.

PR #5820, opened 2023-04-19
T&A Bugfix #0037193: Cannot start a test with mark schema passed from 0%

fhelfer:
https://mantis.ilias.de/view.php?id=37193
If test results cannot be reset delete open runs from tst_active table


fhelfer:
Also cherry-pick to 8 & trunk

kergomard:
After looking at this, this looks like an important fix and makes a lot of code make a lot more sense ;-).
I will fix and cherry-pick upwards to ILIAS8 and trunk.
Best,
@kergomard

PR #5822, opened 2023-04-20
OrgU: Fix for Mantis #37176

tbongers-cat:
Do not include local DIC when WSDL is requested, because ILIAS is not initialized and the global $DIC is not available.
(see https://mantis.ilias.de/view.php?id=37176)


klees:
We should be using the `ServerRequest` from the DIC, but we won't have that either...

PR #5830, opened 2023-04-21
Bug fix mantis report 37208

chlulei:
Prevents error messages when unknown rbac_operations occur in the role operations array.


kergomard:
Thank you @chlulei for the PR

Could you explain to me, why we would want to do this. If this happens, the Didactic Template is invalid and we should rather give a feedback, right?

Best,
@kergomard

chlulei:
Hi @kergomard,
the bugfix seemed appropriate at the time, but in hindsight it is better to adjust the templates and give feedback.

Best regards
@chlulei

kergomard:
Thank you very much @chlulei ! I then close this PR, please feel free to provide another one providing better feedback to the users.
Best,
@kergomard

PR #5804, opened 2023-04-13
Dashboard (ILIAS 9): Feature: KS-ViewControls for Dashboard Sections

fhelfer:
Feature Wiki: https://docu.ilias.de/goto_docu_wiki_wpage_7503_1357.html
Depends on #4941

PR #5833, opened 2023-04-21
UI KS Feature: Standard Listing Panel gets ViewControls

fhelfer:
While implementing https://docu.ilias.de/goto_docu_wiki_wpage_7503_1357.html (PR https://github.com/ILIAS-eLearning/ILIAS/pull/5804) we discovered that Standard Listing Panels do not have the option to render ViewControls like Mode/Sortation and Pagination.
The Dashboard will however need these ViewControls in the Standard Listing Panel to move the options out of the Action Menu and render them to the Panels header section

### DIFF
--- a/src/UI/Component/Panel/Listing/Standard.php
+++ b/src/UI/Component/Panel/Listing/Standard.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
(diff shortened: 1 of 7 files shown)


Amstutz:
Thx a lot!

PR #5834, opened 2023-04-21
T&A 0037197: hint created when entering invalid input for points

kunzeus:
Ticket:
https://mantis.ilias.de/view.php?id=37197 hint created when entering invalid input for points

Problem:
When saving an invalid value in the point deduction field for a question hint, the control flow switched from the form with a questionHint object to an empty form, used to create new hints.

Solution:
Small fix in the control flow when a questionHint object is set.
Added a alert for non-numeric values. However here I'm a bit unhappy, because I belive checkInput should ususally handle this?

As always, let me know how to improve a solution or if I missed something.

### DIFF
--- a/Modules/TestQuestionPool/classes/class.ilAssQuestionHintGUI.php
+++ b/Modules/TestQuestionPool/classes/class.ilAssQuestionHintGUI.php
@@ -98,7 +98,7 @@ class ilAssQuestionHintGUI extends ilAssQuestionHintAbstractGUI
         $lng = $DIC['lng'];
         $ilCtrl = $DIC['ilCtrl'];
 
-        if ($form instanceof ilPropertyFormGUI) {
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much for this PR @kunzeus !

I actually like how small the changes are, ...but I still looked for another solution.
My reasoning:
- You are moving some checking of the form to the buildForm()-function. This is not the place where somebody familiar with ILIAS would look for it, so if for some reason we would start to change things in the form (I really hope we don't, but who knows), this would be a source for hard to trace errors.

As when I understood how the program flow runs in this part and what the consequences are, I already had the fix more or less ready, I just finalized it myself. You find it [here](https://github.com/ILIAS-eLearning/ILIAS/commit/377030158007e6170ce64719694d10b299e8debd). Please let me know, if you see something I missed!

So I will close this PR.
I'm sorry for being difficult once more!

Thanks again and best,
Stephan

PR #5838, opened 2023-04-24
fixed 32216: Object translation of titles are not considered when exp…

alex40724:
…orting an object.

https://mantis.ilias.de/view.php?id=32216

### DIFF
--- a/Modules/LearningModule/classes/class.ilLearningModuleExporter.php
+++ b/Modules/LearningModule/classes/class.ilLearningModuleExporter.php
@@ -89,6 +89,10 @@ class ilLearningModuleExporter extends ilXmlExporter
                     "component" => "Services/Object",
                     "entity" => "transl",
                     "ids" => $md_ids);
+                $deps[] = array(
+                    "component" => "Services/Object",
+                    "entity" => "transl_entry",
+                    "ids" => $a_ids);
             }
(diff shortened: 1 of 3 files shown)


kergomard:
Thank you for the PR @alex40724 

To speed things up: Could you please explain why we are doing this like this? I don't understand [your explanation in the mantis report ](https://mantis.ilias.de/view.php?id=32216#c92809) as it seems inconsistent with @thibsy 's statement, that descriptions are exported, but not titles.

- There are 'transl' and 'transl_entry' entries. Why?
- Why does this work without the change for descriptions, but not for titles? For all intents and purposes they should be identical from a translation point of view, right?
- Are MediaPools and LearningModules the only objects to allow translations of titles or why do we only need changes there and not in the other objects?

I can look into this myself, but then I will postpone as it will take me quite a while to untangle that ball of mud, I fear.

Best,
@kergomard

alex40724:
See my answer in the bug report. Best Alex

alex40724:
Changes in my components are obviously fine for me. You may merge this or fix the issue in another way.

kergomard:
Ok, I'll merge this and cherry-pick to ILIAS8 and trunk. This seems to be one big mess here of data saved in Metadata, Object, and Translations. This needs to change.

PR #5839, opened 2023-04-24
Metadata: Fix several metadata issues

mjansenDatabay:
See:
- https://mantis.ilias.de/view.php?id=37214
- https://mantis.ilias.de/view.php?id=37215

If approved, this MUST be cherry-picked to `trunk`.


smeyer-ilias:
This, I have overseen, thanks @mjansenDatabay

PR #5840, opened 2023-04-24
MarkdownFormattingToHTML

chfsx:
@klees this adds a new Transformation `MarkdownFormattingToHTML` which uses the library we decided to use to convert Markdown to HTML. This case is certainly an excellent example that can be converted within the framework of https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/text-handling.md. For the moment, however, it is important to offer this function centrally in order to be able to find all of its places.

### DIFF
--- a/tests/Refinery/String/MarkdownFormattingToHTMLTest.php
+++ b/tests/Refinery/String/MarkdownFormattingToHTMLTest.php
@@ -1,0 +1,69 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\src\Refinery\String;
+
+use ILIAS\Data\Factory;
+use ILIAS\Refinery\String\Group;
+use ILIAS\Tests\Refinery\TestCase;
+use ilLanguage;
+use InvalidArgumentException;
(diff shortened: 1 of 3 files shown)


tests/Refinery/String/MarkdownFormattingToHTMLTest.php
```diff
@@ -0,0 +1,70 @@
...
+                         ->disableOriginalConstructor()
+                         ->getMock();
+        $group = new Group(new Factory(), $language);
+
+        $this->markdown = $group->markdown()->toHTML();
+    }
+
+    public function stringProvider(): array
+    {
+        // Never ever try to change something on this array :-) e.g. a 'ä' isn't a 'ä' but a 'ä' ;-)
```
mjansenDatabay: Could you please provide more details here? I don't get it :-).

tests/Refinery/String/MarkdownFormattingToHTMLTest.php
```diff
@@ -0,0 +1,70 @@
...
+use ilLanguage;
+use InvalidArgumentException;
+use ILIAS\Refinery\String\Transformation\UTFNormalTransformation;
+use ILIAS\Refinery\Transformation;
+
+class MarkdownFormattingToHTMLTest extends TestCase
+{
+    private Transformation $markdown;
+
+    public function setUp(): void
```
mjansenDatabay:
```suggestion
    protected function setUp(): void
```

mjansenDatabay (changes requested):
Hi @chfsx ,

thank you a lot for contributing to the `Refinery`.

- [x] What do you mean by `a 'ä' isn't a 'ä' but a 'ä' ;-)`?

- [x] Fix the access modifier of the overwritten `setUp` modifier in the unit test.

You already addressed the other topic I would have raised, namely the correlation to Richard's document.

Best regards,
@mjansenDatabay

chfsx:
Hi @mjansenDatabay, thanks for the feedback!
> a 'ä' isn't a 'ä' but a 'ä' ;-)
this was a copy/paste mistake from https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/tests/Refinery/String/UTFNormalTest.php :-) I removed the line
and changed the access modifier of `setUp`

mjansenDatabay (approved):

PR #5842, opened 2023-04-24
add default selection prop type

iszmais:
https://mantis.ilias.de/view.php?id=33554

Fixes an issue for the export of dataCollections for ILAIS 6:

ILIAS 7 solution: https://github.com/ILIAS-eLearning/ILIAS/commit/612373906184950a9b6212f0f87803ab281af559

ILIAS 8 solution: https://github.com/ILIAS-eLearning/ILIAS/commit/b0485fe4903058f6396102a0ab3f76a9ed3f6de1


iszmais:
@mjansenDatabay Since the DataColletion has no maintainer oculd you take a quick look at that?
The change ittself is very minor, i just moved the block to match the ILIAS 7 and 8 structure.

Greetings,
@iszmais

Amstutz:
The Data Collection has no maintainer since it is handled by the Coordinator Model. @pascalseeland and @Amstutz or the coordinators.

The fix looks fine, thx for pushing it to 6. However, please encourage installations still using ILIAS 6 to update soon.

PR #5843, opened 2023-04-25
OrgU: Fix for Mantis #37149 (release_7)

tbongers-cat:
See https://mantis.ilias.de/view.php?id=37149

* Fixes "Add from clipboard" for "Staff" by adding a missing callback function
* Corrects the default value for the user_type/position_id to employee position


tbongers-cat:
I'll add a fix for release_8 and trunk in a second PR due to the refactoring of the org units.

tbongers-cat:
I'll add separate fixes for release_8 and trunk.

PR #5847, opened 2023-04-25
copyright-checker now also looks for line 4 and 5 to begin parsing

daniwe4:


klees:
Hi @daniwe4,

thanks for looking into this, but I will close this as the proposed change would lead to incomplience with PSR-12.

[PSR-12 defines, that a file level docblock appears before declare statements.](https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements) The copyright, after all, is such a block, and we decided that we want to use PSR-12 coding style.

This is inconvenient, as we need to switch `declare` and copyright for various files.

Best regards!

PR #5851, opened 2023-04-26
T&A: fix for 37076

okaufman:
This will fix https://mantis.ilias.de/view.php?id=37076 
I just added strtolwer when getting the cmdClass


kergomard:
Can you still reproduce this error? I can't. Neither on my local system nor on test8.
Best,
Stephan

okaufman:
I can reproduce the problem on this test: https://test8.ilias.de/goto_test8_tst_83463.html

kergomard:
Interesting: I still can't. Not with my main user and not with a user with reduced rights.
Ok, I merge this as I don't see how this can hurt.

Thanks again and best,
@kergomard

PR #5853, opened 2023-04-26
invalid return value for new memcache node

utesche:
Error thrown when installing a new ILIAS instance having a memcache node configured (respectively: updating with a new memcache node):

 [ERROR] ilMemcacheNodesRepository::create(): Return value must be of type ILIAS\Cache\Nodes\Node, none returned


chfsx:
Thanks a lot @utesche !

PR #5855, opened 2023-04-26
(8x) Migration: adapt notification about remaining steps

utesche:
The notification after a migration about the amount of remaining steps differentiates now between 0 or more remaining steps.
The amount of calls of getRemainingAmountOfSteps() is reduced by two.

The calculation of the remaining steps after a migration is fixed according to this commit in branch release_7: 
https://github.com/ILIAS-eLearning/ILIAS/commit/8342ff2dfae9b019b3377e8c71c213801c08d59b

### DIFF
--- a/src/Setup/Objective/MigrationObjective.php
+++ b/src/Setup/Objective/MigrationObjective.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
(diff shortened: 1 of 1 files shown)


utesche:
Sorry, but I can't figure out why the copyright-checker fails ...?

edit 2023-05-01,   reason was: the required Copyright-Comment-Block must begin at least at line 4...

PR #5857, opened 2023-04-27
LegacyUI/Forms: Make `max_length` optional again

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=37209#

If approved, this has to be cherry-picked to `trunk`.

### DIFF
--- a/Services/Form/classes/class.ilTextInputGUI.php
+++ b/Services/Form/classes/class.ilTextInputGUI.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -18,6 +16,8 @@ declare(strict_types=1);
  *
  *********************************************************************/
 
+declare(strict_types=1);
+
 /**
  * This class represents a text property in a property form.
  *
@@ -29,7 +29,7 @@ class ilTextInputGUI extends ilSubEnabledFormPropertyGUI implements ilTableFilte
      * @var string|array
      */
     protected $value = null;
-    protected int $maxlength = 200;
+    protected ?int $maxlength = 200;
     protected int $size = 40;
     protected string $validationRegexp = "";
     protected string $validationFailureMessage = '';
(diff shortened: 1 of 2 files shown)


alex40724:
@mjansenDatabay  Thanks for the PR. Please take care of the conflicts. Please note that I have already adopted the copyright header twice in this file accoring to JF decisions. Not sure what the check is currently complaining about exactly.

mjansenDatabay:
> @mjansenDatabay Thanks for the PR. Please take care of the conflicts. Please note that I have already adopted the copyright header twice in this file accoring to JF decisions. Not sure what the check is currently complaining about exactly.

@alex40724 The `declare` directive has to be below the license header. I fixed that.

alex40724:
@mjansenDatabay Really, is this serious? There are more than thousand(!) files in the Modules directory alone that start with the directive before the license. All maintainers do this (including yourself), Chatroom, StudyProgramme... How is this supposed to be migrated (again)?

The Service directory comes with another 2300 files having declare on top... and src with > 1000, too.

mjansenDatabay:
@alex40724 

> Really, is this serious? 

Hmm, I fear: Yes, see: https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements

I already had to migrate all my components (except the `Chatroom`, because of pending features to be integrated).

The `basic_rector.php` will be able to migrate this. It worked pretty good in my components.

alex40724:
@mjansenDatabay This is the third time I need to go through the license headers (in just one and a half years). Even with rector, this needs time, especially, since this is not only checked in the trunk. If you take this serious, you need to adopt this in all three major branches, right?

Btw. the rector script currently does not work for me: https://mantis.ilias.de/view.php?id=37240

mjansenDatabay:
> This is the third time I need to go through the license headers.

Yes, this was caused by an issue in the `ChangeLicenseHeader` rector, and: The `php-cs-fixer` does not care about this, even if the wrong order is a PSR-12 violation.

🤷

I personally only changed this in `trunk`. Of course I will change this if my commits break the CI in `release_8`. AFAIK `release_7` is not affected.

> Btw. the rector script currently does not work for me: https://mantis.ilias.de/view.php?id=37240

Yes, I guess Fabian will check this.

PR #5860, opened 2023-04-28
T&A Bugfix #0037232: Thumb set set to 0 instead of null

fhelfer:
https://mantis.ilias.de/view.php?id=37232

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assMultipleChoice.php
+++ b/Modules/TestQuestionPool/classes/class.assMultipleChoice.php
@@ -35,6 +35,8 @@ require_once './Modules/Test/classes/inc.AssessmentConstants.php';
  */
 class assMultipleChoice extends assQuestion implements ilObjQuestionScoringAdjustable, ilObjAnswerScoringAdjustable, iQuestionCondition, ilAssSpecificFeedbackOptionLabelProvider
 {
+    protected const DEFAULT_THUMB_SIZE = 150;
+    protected const MINIMUM_THUMB_SIZE = 20;
     /**
      * The given answers of the multiple choice question
      *
@@ -106,7 +108,7 @@ class assMultipleChoice extends assQuestion implements ilObjQuestionScoringAdjus
     ) {
         parent::__construct($title, $comment, $author, $owner, $question);
         $this->output_type = $output_type;
-        $this->thumb_size = 150;
+        $this->thumb_size = self::DEFAULT_THUMB_SIZE;
         $this->answers = array();
         $this->shuffle = 1;
         $this->selectionLimit = null;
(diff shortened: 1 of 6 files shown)


kergomard:
Might I suggest to go another way:
- Set the thumb_size Input to required.
- Define a constant in the `ass<Multiple|Single>Choice`-classes for the thumb_size, set it to protected and make public getter `ass<Multiple|Single>Choice->getDefaultThumbSize()` for it.
- Make sure `ass<Multiple|Single>Choice->getThumbSize()` never returns anything smaller than `ass<Multiple|Single>Choice->getDefaultThumbSize()`.

What do you think?
Thanks again and best!
@kergomard

fhelfer:
> 



> Thank you very much @fhelfer ! Might I suggest to go another way:
> 
> * Set the thumb_size Input to required.
> * Define a constant in the `ass<Multiple|Single>Choice`-classes for the thumb_size, set it to protected and make public getter `ass<Multiple|Single>Choice->getDefaultThumbSize()` for it.
> * Make sure `ass<Multiple|Single>Choice->getThumbSize()` never returns anything smaller than `ass<Multiple|Single>Choice->getDefaultThumbSize()`.
> 
> What do you think? Thanks again and best! @kergomard

It seems there is a default_thumb_size = 150 & a minimum thumb_size = 20
so maybe 2 new constants?

kergomard:
Thank you very much @fhelfer for checking this! 
Yes, please, two of those then. We need to make these values visible. Our magic values are a pain in the butt.
If I'm getting it right, we could then keep the default size private or protected and access it directly as we don't need to ever hand it out.
We should then default to the default size in `ass<Multiple|Single>Choice->getThumbSize()` (sounds logical, right ;-)). This would change the behavior slightly as  right now we default to the min_size, which is just wrong. I would then also suggest to check for the min_size in `ass<Multiple|Single>Choice->getThumbSize()` and raise if necessary.
Thanks again and best,
@kergomard

fhelfer:
@kergomard thumb size can now not be set to a value smaller than the minimum.
This will throw an error in the Import. I currently don't know if thats an issue maybe you can tell me more regarding this case.

dsstrassner:
hmm, if one export from ILIAS 7 (empty field) and import in the adapted ILIAS 8, this will end in an error?

kergomard:
Thanks again @fhelfer 
You went for the setter-approach. It is stricter, but I think it is also fine. Just see for some concerns below.
I've two last things, sorry:
- Please also change [this line here](https://github.com/ILIAS-eLearning/ILIAS/pull/5860/files#diff-d992b9232822011560c2de0377d4c8b986578814bd87ef318f6b6290976b6beeR782) and [this line here](https://github.com/ILIAS-eLearning/ILIAS/pull/5860/files#diff-773eb4bafdf336929e1f3df6e763d1906d76062457d147596f39ca3adfa72b9eL602) to use `$this->object->getMinThumbSize()` (thats why I think we need this function).
- The empty `if ($value)`-Condition like you user [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5860/files#diff-2a025bd1647e434e21eb1c2a533240fbe69296768d9bc5a548926841296f9093R220) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5860/files#diff-cb40336d6f956d4254e7ada13f1f636109583f52300f5c4489d135d1ee8866a6R216) has now bitten me in the butt multiple times, so I think we should be more explicit. What you do actually want is `if ($thumb_size !== null)`, right? I would actually propose to change it to `if ($thumb_size !== null && $thumb_size >= $this->object->getMinThumbSize()`. This way, if anybody gives us an import with a too small $thumb_size, your exception does not bite us (see @dsstrassner 's comment).

Best,
@kergomard

fhelfer:
Thank @kergomard,

that seems like the optimal solution, will update shortly :)

kergomard:
I had to change a little thing I had overlooked before, see: https://github.com/ILIAS-eLearning/ILIAS/commit/226cbf3d54bc4724fc7bde13a21cdb14d91f3e42

Merged and picked to 8.

PR #5862, opened 2023-05-01
[FIX] CI: applied PSR-12 standard for union-types and bitwise operator `|`

thibsy:
Hi all,

I recently discovered a small deviation between our code-style to the PSR-12 standard.

According to PSR-12 code-style we must follow these rules:
- [operators](https://www.php-fig.org/psr/psr-12/#62-binary-operators): all bitwise operators incl. `|` must be surrounded by single spaces.
- [catch structure](https://www.php-fig.org/psr/psr-12/#56-try-catch-finally): the `|` must be surrounded by single spaces when expecting multiple exceptions.

**However**, I could not find anything about using the `|` operator for union types. According to the two rules above, I made the assumption that the operator must also be surrounded by single spaces in this scenario. IMO it's important that we define how union types should be handled, so we don't start mixing up `a | b` vs `a|b`. Concretely, this
```php
function f(float|int $x): float|int
{
    $y = true|false;

    try {
    } catch (LogicException|InvalidArgumentException $z) {
    }
}
```
becomes
```php
function f(float | int $x): float | int
{
    $y = true | false;

    try {
    } catch (LogicException | InvalidArgumentException $z) {
    }
}
```

The PR changes the configuration for PHPStorm and the PHP-CS-Fixer, so the surrounding spaces are now always expected. I believe this might be announced or even discussed at the JF.

Kind regards,
@thibsy

### DIFF
--- a/CI/PHP-CS-Fixer/code-format.php_cs
+++ b/CI/PHP-CS-Fixer/code-format.php_cs
@@ -24,7 +24,7 @@ return (new PhpCsFixer\Config())
         'strict_param' => false,
         'concat_space' => ['spacing' => 'one'],
         'function_typehint_space' => true,
-        'binary_operator_spaces' => ['operators' => ['|' => 'single_space']],
-        'types_spaces' => ['space' => 'single'],
+        'binary_operator_spaces' => ['default' => 'single_space'],
+        // 'types_spaces' => ['space' => 'single'],
 	])
 	->setFinder($finder);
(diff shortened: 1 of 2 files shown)


mjansenDatabay (approved):

mjansenDatabay:
Thanks a lot @thibsy . I am also **unsure** regarding the `Union Types`. IMO we should put this on the agenda of the next JF and discuss this.

Most examples "out there" use `Union Types` and `Intersection Types` without surrounding spaces.

Maybe you (and other readers) already know this, but future coding styles will be discussed in the [`PER`](https://www.php-fig.org/per/coding-style/) in iterations. So there will be no PSR-X for coding style specifications in future PHP versions. But even in the PER I did not find any `Union Type` coding style definition.

matthiaskunkel:
Jour Fixe, 15 MAY 2023: We highly appreciate this PR and accept it for trunk with the following change: please remove all union-type related changes.

PR #5863, opened 2023-05-01
Switching height:100% to 100vh to better support full screen embedded PDFs

dmundra:
I believe this change will fix [#32535](https://mantis.ilias.de/view.php?id=32535). Related commit https://github.com/ILIAS-eLearning/ILIAS/commit/da83a245274dc54cd82aadb46dac1514eb5868bd. Let me know if I should make a new mantis ticket.

The steps to reproduce the problem/confirm it is fixed with the CSS changes in this PR:
1. Click on ‘Add New Item’ and then ‘Media Pool'.
2. Click on ‘Bulk Upload’ and then upload a PDF file.
3. Then click on ‘Save’.
4. Click the PDF and confirm the PDF opens to almost fullscreen instead of a squished narrow window in the modal popup.
5. Go back to the category and then click on ‘Add New Item’ again and then ‘Content Page’.
6. Click on ‘Content’ tab in the content page then ‘Customize the page’.
7. Click on the  sign and then ‘Insert Text’.
8. Add some text and then select it.
9. Then on the left sidebar click ‘Link’ and then ‘Internal Link’.
10. Select ‘Link Type' = ‘Media in Media Tool’.
11. Change Media Source to the media pool created in steps 1 to 3.
12. Select the PDF file and then click ‘Save and Return' and then ‘Finish Editing’.
13. Then click on the ‘Test PDF Link’.
14. Confirm it opens in a new tab.
15. Confirm the PDF opens to almost fullscreen instead of a squished narrow window.


Amstutz:
This is somewhat deeply nested CSS into a iframe. I have a hard time understanding what is going on here. @alex40724 can you look into this and if fine, merge this?

dmundra:
Sorry about that @Amstutz. Here are the missing details, `height:100%` for iframe doesn't do anything so it defaults to 150px height. The solution is to use `height:100vh` which represents the height of the viewport. Reference https://stackoverflow.com/questions/5867985/full-screen-iframe-with-a-height-of-100. For this change I only want it to apply to iframes so hence why the deeply nested CSS. I think it should not affect other content.

Amstutz:
Thx for the explanation. Since this touches page editor content, I will wait for the feedback of @alex40724 for further actions.

alex40724:
Thanks for the PR. We do not set the height at this position. Instead you may define a height in a Media content style accordingly now:
https://github.com/ILIAS-eLearning/ILIAS/tree/trunk/Services/MediaObjects#business-rules

PR #5864, opened 2023-05-02
Assistance: Add alternative approach for link param

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=37249

### DIFF
--- a/Services/Password/classes/class.ilPasswordUtils.php
+++ b/Services/Password/classes/class.ilPasswordUtils.php
@@ -15,20 +15,23 @@ class ilPasswordUtils
      */
     public static function getBytes(int $length) : string
     {
-        if (!defined('PHP_WINDOWS_VERSION_BUILD') && extension_loaded('openssl')) {
-            $secure = null;
-            $rand = openssl_random_pseudo_bytes($length, $secure);
-            if (false !== $rand && $secure === true) {
-                return $rand;
+        try {
+            return random_bytes($length);
+        } catch (Throwable $ex) {
+            if (!defined('PHP_WINDOWS_VERSION_BUILD') && extension_loaded('openssl')) {
+                $secure = null;
+                $rand = openssl_random_pseudo_bytes($length, $secure);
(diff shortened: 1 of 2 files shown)


corro:
LGTM, thank you!

Services/Password/classes/class.ilPasswordUtils.php
```diff
@@ -23,6 +23,10 @@ public static function getBytes(int $length) : string
             }
         }
 
+        if (function_exists('random_bytes')) {
```
pascalseeland: To my understanding from the php documentation (https://www.php.net/manual/en/random.installation.php) there should be no need to check for this function as it is always available since php7.

include/inc.pwassist_session_handler.php
```diff
@@ -64,10 +64,27 @@ function db_pwassist_session_close()
 * - Only a non-substantial number of bits can be predicted from
 *   previously generated id's.
 */
-function db_pwassist_create_id()
+function db_pwassist_create_id(): string
 {
-    // #26009 we use ilSession to duplicate the existing session
-    return \ilSession::_duplicate(session_id());
+    global $DIC;
```
pascalseeland, line 69: I would assume, that random_bytes, as according to its doc is suitable for cryptgraphic stuff would always be random enough to not need this extra round of checking if we already have identical key

pascalseeland (commented):
Generally looks good to me, but we might improve the overall code as mentioned in the comments

Services/Password/classes/class.ilPasswordUtils.php
mjansenDatabay: I took the improved code from `release_8`, this should be the superior approach.

include/inc.pwassist_session_handler.php
mjansenDatabay: You are most propably right, but I would like to make the code here as robust as possible. If it is not a show stopper for you, I would like to keep it.

PR #5882, opened 2023-05-05
Session: avoid accessing unused db table event_files (31608)

schmitz-ilias:
The database table `event_files` is an unused remnant of a previous refactoring. This PR makes a minimally invasive change to avoid accessing it (in lieu of introducing an additional index as suggested in [31608](https://mantis.ilias.de/view.php?id=31608)). I'll provide a separate PR for the trunk to remove the table and some other leftover stuff in `Session` altogether, so this PR should only be merged to R7 and R8.


smeyer-ilias:
Thanks @schmitz-ilias

PR #5883, opened 2023-05-05
Session: clean up removal of files (31608)

schmitz-ilias:
As mentioned in #5882, this PR removes the unused table `event_file` as well as related code.


smeyer-ilias:
Thanks @schmitz-ilias

PR #5889, opened 2023-05-08
T&A 0029121: Cloze question: Answer cannot be added in post-correction 

kunzeus:
Ticket:
https://mantis.ilias.de/view.php?id=29121

Symptom:
Answers for a question of a clozed question test, couldn't be added to the pool of correct solutions.

Problem:
The decision process relied on loose comparison of the existing and new solution.
Loose comparison type juggles the strings to a numeric, and therefore they are seen as equals.

Solution:
Switch to strict comparison. 
Example:
'01' == '1' -> True 
'01' === '1' -> False

Note:
TextSubset questions already uses strict comparison in this code path

tested the solution on:
ILIAS 7 
ILIAS 8 needed additional class call fixing

### DIFF
--- a/Modules/Test/classes/class.ilTestCorrectionsGUI.php
+++ b/Modules/Test/classes/class.ilTestCorrectionsGUI.php
@@ -314,6 +314,7 @@ class ilTestCorrectionsGUI
 
     protected function addAnswer()
     {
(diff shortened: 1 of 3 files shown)


dsstrassner:
Hi @mbecker-databay, could you please cherry-pick the solution to ILIAS 7 as well?

PR #5890, opened 2023-05-08
Dashboard: Remove ilAdvancedSelectionListGUI and ilSubmitButton

lscharmer:
Removing Legacay UI Components in the Dashboard.

### DIFF
--- a/Services/Dashboard/classes/class.ilDashboardGUI.php
+++ b/Services/Dashboard/classes/class.ilDashboardGUI.php
@@ -3,15 +3,18 @@
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
+ *
  * ILIAS is licensed with the GPL-3.0,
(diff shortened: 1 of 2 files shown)


Services/Dashboard/classes/class.ilDashboardGUI.php
```diff
@@ -3,15 +3,18 @@
...
  * You should have received a copy of said license along with the
  * source code, too.
+ *
  * If this is not the case or you just want to try ILIAS, you'll find
  * us at:
  * https://www.ilias.de
  * https://github.com/ILIAS-eLearning
- */
+ *
+ *********************************************************************/
```
iszmais, line 17: Can we add a `declare(strict_types=1);` here?

iszmais (approved):
LGTM :+1: 

just one minor question but nothing relevant.

Though this changes do cause some visuall inconveniences, but nothing in our responsibility so far. We should keep an eye on that but for now I would just await the further progress within the LUI removals. Maybe such thing resolve themself with the finalization

PR #5892, opened 2023-05-08
Setup: Fix Github installation workflow (#37261)

tbongers-cat:
The Github workflow for testing the installation failed, because the referenced config.json was located inside the webroot. This is now fixed by coyping the file to a location outside of webroot before performing the setup.

**Please note**: The workflow will still fail until the included fix to the config file (`CI/install-check/ilias-ci-config.json`) is merged into trunk, as the code is checked out from there during the workflow run.

_(PS: In my test runs, another (probably unrelated) error happened during installation which leads to the setup failing. If this just happens during the workflow run, we'll have to tackle it separately.)_

### DIFF
--- a/.github/workflows/install-check-trunk.yml
+++ b/.github/workflows/install-check-trunk.yml
@@ -22,6 +22,10 @@ jobs:
         with:
           ref: trunk
 
+      - name: Copy config file to location outside webroot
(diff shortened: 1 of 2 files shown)


klees:
Picked to release_8.

PR #5897, opened 2023-05-09
[FIX] UI: fixed access before initialization error in Table\Presentation

thibsy:
Hi all,

I noticed that rendering a `Table\Presentation` component without calling `::withData()` on it throws an error because the `$records` property is accessed before initialization.

I've fixed this by initializing the value by default, so `::withData()` stays an optional method call, and supplied an according unit-test for this behaviour. This should be picked for release_8 as well if accepted.

Kind regards,
@thibsy

### DIFF
--- a/lang/ilias_de.lang
+++ b/lang/ilias_de.lang
@@ -16743,6 +16743,7 @@ ui#:#ui_link_url#:#URL
 ui#:#ui_md_input_edit#:#Bearbeiten
 ui#:#ui_md_input_view#:#Vorschau
 ui#:#ui_numeric_only#:#Bitte geben sie eine ganze Zahl ein.
+ui#:#ui_table_no_records#:#Keine Einträge
 ui#:#ui_tag_required#:#Bitte geben sie mindestens ein Tag ein.
 ui#:#ui_transcription#:#Abschrift
 user#:#all_roles_has_starting_point#:#Alle Rollen mit definierter Startseite
--- a/lang/ilias_en.lang
+++ b/lang/ilias_en.lang
@@ -16740,6 +16740,7 @@ ui#:#ui_link_url#:#URL
 ui#:#ui_md_input_edit#:#Edit
 ui#:#ui_md_input_view#:#View
 ui#:#ui_numeric_only#:#Please insert a whole number.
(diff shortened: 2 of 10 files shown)


klees (changes requested):
Hi @thibsy,

as briefly discussed via Discord: We agree that simply showing a headline without anything below it will be confusing for users. We instead want to adapt the solution of `ilTable2GUI` and show some "There is no data here..."-message instead of the rows.

Thanks for taking care of this!

thibsy:
Hi @klees 

I have implemented empty table rows for the `Table\Presentation` and `Table\Data` table _in addition_ to the initial fix, which makes `Table\Presentation::withData()` optional again.

This PR introduces:
- a new language variable `ui_table_no_records`, which currently translates to `No records` and is used by both tables as a default message/content to indicate there's no data available.
- a new unit test for both renderings, which uses `str_contains()` to check only for the important aspects. Note the `PresentationTest:: testBareTableRendering()` case was not valid anymore, because this rendering test will now result in rendering the empty entry. I have replace the test accordingly.
- a new example for both tables, which demonstrate the default behaviour if no data has been provided. I think we might even provide some `Table::withNoRecordsMessage()` method in the future to provide more flexibility, or even add this to `DataRetrieval` because there might be different reasons for having no data (like permissions).
- unfortunately, some code-style changes, because #5970 got merged in the meantime, fixing arrow-function whitespaces.

I am open for better suggestions for the `ui_table_no_records` translations and the design for the new empty row for `Table\Presentation` tables =).

Kind regards,
@thibsy

klees (approved):
LGTM. Do we need @catenglaender or is this already styled nicely?

thibsy:
Hi @klees 

Here are the screenshots of how the empty tables look like right now:

Screenshot 2023-09-25 at 16 44 39
Screenshot 2023-09-25 at 16 44 00

I think it looks okay-ish. Unless you or @catenglaender disagree?

Kind regards,
@thibsy

klees:
Looks fine for the moment IMO. Would merge now if conflicts are resolved.

thibsy:
@klees I've rebased onto trunk, this should be ready for merge then =).

PR #5898, opened 2023-05-09
TA fix Cloze and Longmenu Question Form Issues

Amstutz:
This form should be removed asap. Soooo many things could go wrong here. One small thing should be fixed with this PR, see: https://mantis.ilias.de/view.php?id=37130 and https://mantis.ilias.de/view.php?id=37081


Amstutz:
Note test failings seem unrelated afais

smeyer-ilias:
@Amstutz the test failings are fixed in R8

kergomard:
Thank you very much @Amstutz for this PR!
And yes: This forms are generally speaking a bad idea built on top of other bad ideas and implementations.
One thing though: I don't believe this solves: https://mantis.ilias.de/view.php?id=37081
I still merge and cherry-pick to trunk.

Amstutz:
I do believe it does. Maybe reload js cache.

PR #5901, opened 2023-05-09
fixed 31419: fix broken layout with image maps

catkrahl:


Amstutz:
Fix for: https://mantis.ilias.de/view.php?id=31419

PR #5902, opened 2023-05-10
T&A Bugfix #0037248: Test Results Link Broken

fhelfer:
https://mantis.ilias.de/view.php?id=37248
Also the case for 8 and up


kergomard:
Thank you very much @fhelfer ! Will merge and cherry-pick to 8 and trunk.

PR #5903, opened 2023-05-10
UI, Mainbar: 37282, slate(cookie) still enganged after redirect

nhaagen:
https://mantis.ilias.de/view.php?id=37282

When a tool is active, the cookie il_mb_slates is set to "engaged"; this state is not lost (and shouldn't be) when redirecting.
However, the new page maybe has no more tools.

The entirely correct way woul be to first close slates and then redirect; this fix just re-writes the state after first calculation, which will result in the described behavior occuring _once_ only.

### DIFF
--- a/src/UI/templates/js/MainControls/src/mainbar.main.js
+++ b/src/UI/templates/js/MainControls/src/mainbar.main.js
@@ -271,8 +271,8 @@ var mainbar = function() {
         }
         mb.model.actions.initMoreButton(mb.renderer.calcAmountOfButtons());
         let state = mb.model.getState();
-        mb.persistence.storePageState(state.any_entry_engaged || state.tools_engaged);
         mb.renderer.render(state);
+        mb.persistence.store(state);
     },
     init_mobile = function() {
         var mb = il.UI.maincontrols.mainbar;
--- a/src/UI/templates/js/MainControls/src/mainbar.main.js
+++ b/src/UI/templates/js/MainControls/src/mainbar.main.js
@@ -270,7 +270,9 @@ var mainbar = function() {
             }
         }
         mb.model.actions.initMoreButton(mb.renderer.calcAmountOfButtons());
-        mb.renderer.render(mb.model.getState());
+        let state = mb.model.getState();
+        mb.persistence.storePageState(state.any_entry_engaged || state.tools_engaged);
+        mb.renderer.render(state);
     },
     init_mobile = function() {
         var mb = il.UI.maincontrols.mainbar;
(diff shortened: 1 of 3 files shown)


thibsy (commented):
Hi @nhaagen 

Thx a lot for providing a fix for this issue!

- [ ] `persistence.storePageState()`: you have exposed an internal function of this object, could you explain why we cannot call `persistence.store()` which is already exposed? I see that `persistence.store()` ultimately also calls `storePageState()` in the same fashion as you now do in `mainbar.init_desktop()`. Do I overlook some side-effect here, which makes exposing `storePageState()` necessary?

Kind regards,
@thibsy

src/UI/templates/js/MainControls/src/mainbar.main.js
```diff
@@ -270,7 +270,9 @@ var mainbar = function() {
             }
         }
         mb.model.actions.initMoreButton(mb.renderer.calcAmountOfButtons());
-        mb.renderer.render(mb.model.getState());
+        let state = mb.model.getState();
+        mb.persistence.storePageState(state.any_entry_engaged || state.tools_engaged);
```
thibsy: As from my question above: can we use `mb.persistence.store(state)` here instead?

nhaagen: Hey @thibsy ,
yes. In fact, I had it like this before, but in store(state), all entries are being walked through to compress them. 
This is of no concern to determine the slate-state, and I wanted max speed, since the whole thing is about "twiching".
Furthermore, I see no real disadvantage in exposing the function - closing the slate or setting the cookie on click of the button would actually have prevented the remarked behavior, and - it's a subset of a previously exposed function, without  there being any relevant calculations beforehand (in storeStates).
In general, I agree: The whole persistence layer should be internalized, but this is rather on the roadmap-side.
Best regards,
Nils

thibsy: Thx @nhaagen for the explanation. I hesitated because as I understand the mainbar js is written in a [redux-ish](https://redux.js.org/introduction/core-concepts) manner, which tries to avoid storing state "arbitrarily" and only does so through actions. As far as I understand, you always want to store the entire state, not just parts of it. I agree that speed is of the essence here, but is it really `.store()` vs `.storePageState()` that matters? IMO we could accept this, but I'd ask for a brief comment in `storePageState` explaining the exposure.

nhaagen:
I reversed to storing the entire state.

thibsy (approved):
Thx a lot, LGTM.

thibsy:
merged into trunk, picked for
- release_8
- release_7

PR #5904, opened 2023-05-10
Test: Fix ErrorText Behaviour

kergomard:
This PR is just in case somebody wants to look through the whole shenanigans, while we test the surface. 
It is rather encompasing, but actually started as a fix for: https://mantis.ilias.de/view.php?id=28393

But I didn't believe it useful to just tackle this and as we are still early in the cylce, this is a
refactoring of the ErrorText-Question.

It:
- Moves the parsing of the test to the creation time.
- It introduces saved positions for correct answers.
- It removes all jQuery and improves JS.
- It makes error-passages work in Learning Modules.
- It splits up the huge parsing functions into small, named functions.

### DIFF
--- a/Modules/Scorm2004/scripts/questions/question_handling.js
+++ b/Modules/Scorm2004/scripts/questions/question_handling.js
@@ -762,65 +762,78 @@ ilias.questions.selectErrorText = function(a_id, node) {
 	jQuery(node).blur();
 };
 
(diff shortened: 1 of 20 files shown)


kergomard:
Ok, I think we have now tested this thoroughly enough. Thanks @dsstrassner for the effort!

PR #5908, opened 2023-05-11
Removed nested styles from less file of ui icons ...

fneumann:
... because the template now uses an img directly without nested elements.

This PR is just cleanup of obsolete code - it shoud not have any visible effect. 

Note: the abbrevation feature of custom icons seems not to work at all.


Amstutz:
thx a lot @fneumann ! I manualy added your changes together with the fix for the abbreviation issue ( https://mantis.ilias.de/view.php?id=37294) here: https://github.com/ILIAS-eLearning/ILIAS/commit/18b79becd2729bcdf2fe8e9e0bcad09639df8302

PR #5909, opened 2023-05-11
#36506: fixed locally with added style only for the pdf generation

fneumann:
Works for both Wkhtmltopdf versions with or without patched QT. The small icon size is fixed to20px as we don't have the less variables available here. The icons produced by the version with patched QT  have a smaller width, but I think we can live with that.


Amstutz (approved):
thx a lot @fneumann

kergomard:
Thank you very much @fneumann also from my side!

PR #5910, opened 2023-05-11
Fix #36081: Antworten mit Umlauten für "Begriffe benennen" 

fneumann:
Correct answers are provided with encoded HTML entities. The fix decodes these before they are compared with the user input.

### DIFF
--- a/Modules/Scorm2004/scripts/questions/question_handling.js
+++ b/Modules/Scorm2004/scripts/questions/question_handling.js
@@ -1224,6 +1224,11 @@ ilias.questions.showCorrectAnswers =function(a_id) {
 					{
 						correct_answer = correct_answer.toLowerCase();
 					}
+
+					// decode the HTML entity encoding of the correct answer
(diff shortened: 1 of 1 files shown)


kergomard (commented):
Thank you very much @fneumann for the PR!

So, technically this looks right and it also seems to work for the comparison part, ...but there is something else that is interesting: For me the value in the Input field is deleted after it is sent. You should be able to see the behavior in the first question [here](https://test7.ilias.de/goto.php?target=pg_192079_81248&client_id=test7).

Thanks again and best,
@kergomard

fneumann:
Ok, the fix was also neeed at a second place. Now that is hadled, too.

kergomard:
Thanks again @fneumann ! I will merge and cherry-pick to 8 and trunk.

fneumann:
Wait a minute - I just saw that the second fix was after a toLowerCase - it should be before. Will provide a follow-up PR.

kergomard:
Thanks! I missed that one too ;-). I'll merge and cherry-pick right away.

PR #5911, opened 2023-05-11
[FIX] 0037295: Error after login: ILIAS\ResourceStorage\Flavour\Flavours::get()

chfsx:
see https://mantis.ilias.de/view.php?id=37295


kergomard:
Thank you very much @chfsx for the PR and your quick explanation, why this is actually the right thing to do.

PR #5912, opened 2023-05-12
Fix7 36081 text subset in lm postfix

fneumann:
Moved the second fix before the casting to lowercase.


kergomard:
Thanks for checking once more and the PR @fneumann !

PR #5913, opened 2023-05-12
Added UsageOfAttributeFinder

chfsx:
With this PR, I propose to provide a new `UsageOfAttributeFinder` in addition to the existing `ImplementationOfInterfaceFinder` in the setup. Attributes could now be used in the code base, since we support the corresponding minimum php version. 

A first use case will be presented in the context of "[Persisting Help IDs](https://docu.ilias.de/goto.php?target=wiki_1357_Persisting_Help_IDs)" (note that at the time of the opening of the PR, the page is not yet complete, but will be updated ASAP).

Furthermore, `ilCtrl` will also want to make use of attributes, for example to convert the `ilCtrlCalls...` comments into a stable and testable format.

I would be very happy about feedback and the merge of the PR.

I have taken the liberty of transferring the parts of the code that are identical in the `ImplementationOfInterfaceFinder` into an abstract class. If you need any adjustments, feel free to contact me.


klees (changes requested):
Hi @chfsx,

nice to see this code being useful to others =)

Thanks a lot for extracting the commonalities of the two classes. I think we should go one step further as there are even more commonalities to be extracted. The only actual difference between the two classes is `$r->implementsInterface($attribute)` vs. `$r->getAttributes($attribute) !== []`. In general, it seems to be quite a coincidence that the two classes can be subsummed with a generic public method. This is only the case because the interface we are looking to find is identified by a `string`, same as the attribute. If we, for example, would try to find classes where some attribute has a specified value, that commonality would break down, as we would require two parameters to find matching classes then. I think this indeed already visible in renaming `$interface` to `$attribute`. So even if the first parameter has the same type in both cases, it certainly has a different meaning and this cannot be abstracted in a common method.

So, I would suggest the following course of action (which also eases the implementation of unit tests). You do not need to follow this, but but please indicate shortly why you prefer to do otherwise:
- [x] Move `AbstractOfFinder::getMatchingClassNames` to a `protected AbstractOfFinder::genericGetMatchingClassNames`. Turn the first parameter into a `callable $is_matching`. The callable is supposed to take a `\ReflectionClass` as parameter and return a boolean that indicates if the class matches. Adjust code accordingly, of course.
- [x] Implement `ImplementationOfInterfaceFinder::isClassMatching(string $interface, \ReflectionClass $r)` and use it to recreate `ImplementationOfInterfaceFinder::getMatchingClassNames` (with `$interface` as first parameter name =)) by filling the first parameter and passing the rest to `AbstractOfFinder::genericGetMatchingClassNames`.
- [x] Implement `UsageOfAttributeFinder::isClassMatching` and use it to recreate `UsageOfAttributeFinder::getMatchingClassNames` (with the new `$attribute` as first parameter name =)) in the same way.

- [x] Tests: Implement them either way. If you follow my suggestion, you need to test `ImplementationOfInterfaceFinder::isClassMatching`, `UsageOfAttributeFinder::isClassMatching` and `AbstractOfFinder::getMatchingClassNames` with a generic `$is_matching`. The further and the latter should mostly be some reorganisation of the existing tests for the `ImplementationOfInterfaceFinder`. If you do not follow my suggestion, please split the tests according to the the three classes we then have.

I think we won't be at the end of the line when it comes to finding classes on the code base, so I expect this generalisation to become useful soon. If you have any questions or stuff to discuss, feel of course free to contact me, as always.

Kind regards!

chfsx:
@klees thanks for the feedback, I hope i have understood everything correctly. so that the test class is more appropriate, I have renamed it ClassFinderTest.

klees (approved):

klees:
Thanks a bunch, looks neat.

PR #5918, opened 2023-05-15
T&A: Bugfix #31232: Question Pool Permalink not working for only visible roles

jcopado:
Solves https://mantis.ilias.de/view.php?id=31232

### DIFF
--- a/Modules/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php
+++ b/Modules/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php
@@ -1737,14 +1737,16 @@ class ilObjQuestionPoolGUI extends ilObjectGUI implements ilCtrlBaseClassInterfa
         $lng = $DIC['lng'];
         $ctrl = $DIC['ilCtrl'];
 
-        if ($ilAccess->checkAccess("write", "", (int) $a_target)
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @jcopado for this PR.

Technically this looks right. I would have some formalities, though - one nitpick (I'm sorry) and two favors to ask:
* Could you please add the spaces in [these two lines](https://github.com/ILIAS-eLearning/ILIAS/pull/5918/files#diff-79341fc60c03ba53809fb097ae04d5d5a3e597a46a47406f01dd0fed78e926c2R1740-R1741) again. I believe this actually is a PSR-12 violation (see examples in [Section 6.1 here](https://www.php-fig.org/psr/psr-12/#61-unary-operators)).
* Could you please replace [this `exit`](https://github.com/ILIAS-eLearning/ILIAS/pull/5918/files#diff-79341fc60c03ba53809fb097ae04d5d5a3e597a46a47406f01dd0fed78e926c2L1747) with a `return`. We could actually simply get rid of it all together, but I think it makes things clearer.
* Could you please remove the two `elseif` statements and replace them with `if`s and early returns.

Thank you very much!

Berst,
@kergomard

kergomard:
Muchas gracias @jcopado and @JuanAguilarG ! I will merge and cherry-pick to 7 and trunk.

PR #5923, opened 2023-05-16
T&A: Bugfix #36142: Type error when direct Importing an XML file when TinyMCE Image is included in original question

jcopado:
Ensures no type error is raised when:

Steps would be like:

0a) Create a pool or test
0b) Therein create a question with TinyMCE and include some media (i.e. an image) in the question and/or answer text
1) Export the test or pool
2) Extract the "_qti_" labeled XML-file from the top level of the export ZIP-file
3) On the question list of any given pool click the "Import" button
4a) Select the 'qti'-XML-file from step 2) and click the "Upload" button
4b) On the following 'Import' screen click the "Import" button

### DIFF
--- a/Modules/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php
+++ b/Modules/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php
@@ -782,7 +782,9 @@ class ilObjQuestionPoolGUI extends ilObjectGUI implements ilCtrlBaseClassInterfa
             $newObj->setPermissions($this->qplrequest->getRefId());
         }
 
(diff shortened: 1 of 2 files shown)


jcopado:
No yet finished
Work in Progress

jcopado:
Without changing ilObjQuestionPoolGUI, the new implementation of ilSession:get() (in ILIAS7 was done directly with $_SESSION) didn't work properly, as the is_file only get "//manifest.xml" now we ensure qpl_import_dir and qpl_import_subdir exists, otherwise, we go to the XML Import as desired.

Now it should be solved.

kergomard:
Thank you very much @jcopado . Will merge and cherry-pick to trunk.

PR #5924, opened 2023-05-17
T&A 0036944: Mark Schema: 'Official Form'

kunzeus:
Ticket:
https://mantis.ilias.de/view.php?id=36944

Symptom:
in the Settings -> Mark Schema, the table column Official Form didn't show any information.

Problem:
Refactoring mistake 

Checked the solution on:
ILIAS 7, since described as only ILIAS 7 bug, see ticket.

Note:
The second bug, described in the comments of the ticket, occurs when no value is saved for a Mark Schema table row.   Easy to produce when you save by default nothing to the database for the Official Form name. However, with this fix, an empty failed message is now rather a willing result than a system error.


kergomard:
Thank you very much @kunzeus ! I will merge. If I'm getting this right, no cherry-pick is needed as this is ILIAS7 only.

PR #5925, opened 2023-05-17
Object/Container: Move Adopt Content

kergomard:
Hi @alex40724 
This PR moves Adopt Content to the Toolbar [as it was requested by the Jour Fixe](https://docu.ilias.de/goto_docu_wiki_wpage_5268_1357.html). After some thinking, I think this is the most elegant way to do it right now, opening up opportunities for the future.

I think we could actually mostly backport this to ILIAS 8, we should then just change the Link for a Button (I can do that, if you like). I think this also makes the behavior across Objects more consistent.

Best,
@kergomard

### DIFF
--- a/Services/Object/classes/class.ilObjectGUI.php
+++ b/Services/Object/classes/class.ilObjectGUI.php
@@ -780,17 +780,15 @@ class ilObjectGUI
 
     protected function addAdoptContentLinkToToolbar(): void
     {
-        if ($this->object->getType() === 'crs' || $this->object->getType() === 'grp') {
-            $this->toolbar->addComponent(
-                $this->ui->factory()->link()->standard(
-                    $this->lng->txt('cntr_adopt_content'),
-                    $this->ctrl->getLinkTargetByClass(
-                        'ilObjectCopyGUI',
-                        'adoptContent'
-                    )
+        $this->toolbar->addComponent(
+            $this->ui->factory()->link()->standard(
+                $this->lng->txt('cntr_adopt_content'),
+                $this->ctrl->getLinkTargetByClass(
+                    'ilObjectCopyGUI',
+                    'adoptContent'
                 )
-            );
-        }
+            )
+        );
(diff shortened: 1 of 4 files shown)


Services/Object/classes/class.ilObjectGUI.php
```diff
@@ -783,6 +783,21 @@ protected function addDidacticTemplateOptions(array &$a_options): void
     {
     }
 
+    protected function addAdoptContentLinkToToolbar(): void
+    {
+        if ($this->object->getType() === 'crs' || $this->object->getType() === 'grp') {
```
alex40724: I think this line is a strong hint, that this does not belong to the object service at all. You usually do not want to add these type specific checks to general classes like ilObjectGUI since this violates SoC principles. Additionally I have no idea how a non-container would make use of adopt content at all, so this is a container feature imo.

alex40724 (changes requested):

Services/Object/classes/class.ilObjectGUI.php
kergomard: Thank you very much @alex40724 for the quick reply.
But: You did look at the code, right? This here is just a trace I forgot to remove, coming from the previous implementation in `ilObjContainerGUI`. I will remove the if-statement right away as it is not needed anymore.

Why I would like to have this in ilObjectGUI for the time being (this time is not going to be very long):
* 'Adopt Content' is part of the copy process and copying is a concern of the `Object` right now (and most probably rightly so).
* I would like to start refactoring `Object` and this makes it (see the previous point) just this little bit easier.

Thanks again and best,
@kergomard

kergomard:
Hi @alex40724
Could you please have a look at this?
Thank you very much and best,
@kergomard

alex40724 (approved):
The changes in the container component are ok for me. Even if I am not aware of any non-container which could make use of adoptContent.

PR #5928, opened 2023-05-22
T&A Bugfix #0026320: Test generates empty data before password input

fhelfer:
Quick fix for: https://mantis.ilias.de/view.php?id=26320
Contrary to the JF decision, i just prevented data generation before password input.
Generating data only on manual finish has huge impact on test execution

Cherry-pick for 7&8 as well

### DIFF
--- a/Modules/Test/classes/class.ilCronFinishUnfinishedTestPasses.php
+++ b/Modules/Test/classes/class.ilCronFinishUnfinishedTestPasses.php
@@ -208,6 +208,19 @@ class ilCronFinishUnfinishedTestPasses extends ilCronJob
     {
         $processLocker = $this->processLockerFactory->withContextId((int) $active_id)->getLocker();
 
+        $testSession = new ilTestSession();
+        $testSession->loadFromDb($active_id);
+
+        $test = new ilObjTest($obj_id, false);
+
+        assQuestion::_updateTestPassResults(
+            $active_id,
+            $testSession->getPass(),
+            $test->areObligationsEnabled(),
+            null,
+            $obj_id
+        );
+
         $pass_finisher = new ilTestPassFinishTasks($active_id, $obj_id);
         $pass_finisher->performFinishTasks($processLocker);
 
--- a/Modules/Test/classes/class.ilTestEvaluationGUI.php
+++ b/Modules/Test/classes/class.ilTestEvaluationGUI.php
@@ -1981,8 +1981,20 @@ class ilTestEvaluationGUI extends ilTestServiceGUI
         $participantData->load($this->object->getTestId());
 
         if (in_array($activeId, $participantData->getActiveIds())) {
+            $testSession = new ilTestSession();
+            $testSession->loadFromDb($activeId);
+
+            assQuestion::_updateTestPassResults(
+                $activeId,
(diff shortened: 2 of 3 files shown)


dsstrassner:
Hi @fhelfer, normally, these issues are in the pre-test-phase when the test-admins test their settings.

This generates an entry in the test-dashboard. I could end the test run there, but the test is still blocked. But no entry on the result page to delete all test-runs. See the screenshots. 

The JF decided, that clicking on "Test beenden" in the test-dashboard should generate an entry on the result page ...
2023-05-22-js3gy_Screenshot-26320-1

2023-05-22-js3gy_Screenshot-26320-2

2023-05-22-js3gy_Screenshot-26320-3

Maybe @kergomard or @mbecker-databay could state, if your solution is good enough to avoid the initial problem ;-)

kergomard:
Hi @fhelfer 

Thank you very much @fhelfer , especially for thinking ahead!

Now, you confused me properly, but I think you are just not doing, what you write in your description above ;-):
You are not "preventing data creation". Quite the opposite, actually, you are "ensuring creation of a full dataset", right?
I actually do think this could also be a possible solution, but it also scares me. The program flow here is murky at best: Are we really sure we covered all possible entry-points? An alternative could be to keep the redirect where it was, to split of the population of question and such into a separate function (probably more or less [this part here](https://github.com/ILIAS-eLearning/ILIAS/pull/5928/files#diff-1a37eeff2a6c8cd43bf495aeda15b439f63c5eb32781b0f6a171d678a9482ba8L186-L216)) and to call it inside `ilTestPlayerAbstractGUI::handlePasswordProtectionReidrect()` to ensure we create a full set of data.

I'm really just thinking aloud. I just remember how long I debugged this password-catastrophe before the release and how finicky it actually is.

What do you think?

Thanks again and best,
@kergomard

fhelfer:
Hi @kergomard,
> Now, you confused me properly, but I think you are just not doing, what you write in your description above ;-): 

Yes i force-pushed some other approaches in the mean time. Sry for the confusion
The screenshots @dsstrassner provided helped me understand the JF decision, i completely misunderstood :sweat_smile: 
So now Test Results will be written after finishing a test run manually by a admin, so they can be safely deleted afterwards. It prevents blocked tests

kergomard:
Thanks again @fhelfer !
Now, this I understand ;-) . I looked through the consequences of this in detail as this call this late in the game worries me. I think there is one issue: You set `$obligationsEnabled` to `false` as far as I can see, this would lead to all tests having `$obligations_answered` to `1`. I don't think we can want that.

On a more general note: Could we maybe make this more specific? I think we need this only in three specific places:
in `ilTestEvaluationGUI::confirmFinishTestPassForUser()`, in `ilTestEvaluationGUI::confirmFinishTestPassForAllUser()`, and in `ilCronFinishUnfinishedTestPasses::finishPassForUser()`. Could you maybe try to limit this call to exactly this places? I'm a little bit worried about funky race conditions and similar if we simply call it on all passes.

Thanks again and best,
@kergomard

kergomard:
Thank you very much @fhelfer ! I believe this works. We have some redundancy, but we should solve that problem separately.

PR #5929, opened 2023-05-22
T&A Bugfix #0031276: Cannot add copied questions

fhelfer:
https://mantis.ilias.de/view.php?id=31276

@mbecker-databay  The "add from other test" feature did list only copied courses when filtering for a specific test and only original courses with no test filter
Changed the behaviour so that always all tests are shown
If thats not the intended behaviour please notify me

### DIFF
--- a/Modules/Test/Service/class.InternalRequestService.php
+++ b/Modules/Test/Service/class.InternalRequestService.php
@@ -70,6 +70,11 @@ class InternalRequestService
         return $this->int('q_id');
     }
 
+    public function getQuestionIds(): array
+    {
+        return $this->intArray('q_id');
+    }
+
     public function getCallingTest(): int
     {
         return $this->int('calling_test');
--- a/Modules/Test/classes/class.ilObjTestGUI.php
+++ b/Modules/Test/classes/class.ilObjTestGUI.php
@@ -3365,7 +3365,7 @@ class ilObjTestGUI extends ilObjectGUI implements ilCtrlBaseClassInterface
 
                 $clone = assQuestion::instantiateQuestionGUI($new_id);
                 $clone->object->setObjId($this->object->getId());
-                $clone->object->saveToDb();
+                $clone->object->saveToDb($new_id);
 
                 $this->object->insertQuestion($this->test_question_set_config_factory->getQuestionSetConfig(), $new_id, true);
 
--- a/Modules/Test/classes/class.ilObjTestGUI.php
+++ b/Modules/Test/classes/class.ilObjTestGUI.php
@@ -3335,10 +3335,10 @@ class ilObjTestGUI extends ilObjectGUI implements ilCtrlBaseClassInterface
             $this->ctrl->redirect($this, "infoScreen");
         }
 
-        if ($this->testrequest->raw('q_id') && !is_array($this->testrequest->raw('q_id'))) {
-            $ids = array($this->testrequest->raw('q_id'));
-        } elseif ($this->testrequest->raw('q_id')) {
-            $ids = $this->testrequest->raw('q_id');
+        if ($this->testrequest->hasQuestionId() && !is_array($this->testrequest->raw('q_id'))) {
+            $ids = [$this->testrequest->getQuestionId()];
+        } elseif ($this->testrequest->getQuestionIds()) {
+            $ids = $this->testrequest->getQuestionIds();
         } else {
             $this->tpl->setOnScreenMessage('failure', $this->lng->txt('copy_no_questions_selected'), true);
             $this->ctrl->redirect($this, 'questions');
--- a/Modules/Test/classes/class.ilObjTestGUI.php
+++ b/Modules/Test/classes/class.ilObjTestGUI.php
@@ -3365,7 +3365,7 @@ class ilObjTestGUI extends ilObjectGUI implements ilCtrlBaseClassInterface
 
                 $clone = assQuestion::instantiateQuestionGUI($new_id);
                 $clone->object->setObjId($this->object->getId());
-                $clone->object->saveToDb($new_id);
+                $clone->object->saveToDb();
 
                 $this->object->insertQuestion($this->test_question_set_config_factory->getQuestionSetConfig(), $new_id, true);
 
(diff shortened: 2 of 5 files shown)


kergomard:
Hi @fhelfer 
Thank you very much for this PR and for tracking this issue!
If I'm getting this right, you are doing two things here:
* You are changing the behavior of the question list filter.
* You are changing the behavior of the question creation.

I believe we should actually only do the first one. I don't think the general implementation of the concept of an "Original" and its "Copy" is in any way sound, but I think we are going even further in the wrong direction with this.
The crux of this whole thing:
* Do we really know what we need the `original_id` for?

Let me see, if I can define an intended behavior:
* Up to now, I assume, we want the `original_id` to track if a question is a copy of another question. This is in my mind only relevant for questions that are created in a Question Pool and then copied to a Test. If a question is duplicated inside a Test, it seems to me, that it is obvious that the copy is actually completely independent: Why the heck would somebody want to have the "same" (in the sense of two instances of the same object) question twice in a test? The copy-process seems to me to be used as a "template" to quickly create a similar question.
* Thus the `original_id` should contain the `obj_id` of the "original" if the question is a copy from a question in a pool. If the question itself is the "original", it should be `null`.
* So, if we create a question ("Question") anew the `original_id` will be set to `null`. If we add (copy) this question from a Question Pool to a Test ("Question Copy"), the copy will have the `original_id` set to the `id` of "Question". If we then make a copy of "Question Copy" ("Copy of Question Copy"), from where I stand, the `original_id` should be `null` again. The `original_id` should also be `null` if we copy a question inside a pool.
* I would also stipulate that the moment we change something in "Question Copy" above, without syncing it back to "Question" we should actually set the `original_id` to `null`. As otherwise the information contained in `original_id` looses all its value.

ADDENDUM 2023-08-31: While looking at the proposal to improve logging, we noticed that there was an additional hidden case where questions in a test can be changed: the corrections. Once a test has been run, questions can still be edited in part to compensate for errors. We suggest to remove the `original_id` in this case, too, as the question is clearly not the same one as in the Question Pool anymore.

This is only partly reality, but let us see, if this is where we actually want to go.

I will bring this to the attention in the next TechSquad tomorrow.

Thanks again and best,
@kergomard

kergomard:
Hi @fhelfer 
As promised, I discussed this with the TechSquad and we came to the conclusion that the behavior I describe above is basically correct. There are a few details in import and export I need to have a closer look at, but basically: The `$original_id` has really the use described above, so we should take care to keep it as clean as possible.

I thus kindly ask you, to have another look at this issue.

Thank you very much for tackling these hard questions!
@kergomard

fhelfer:
Hi @kergomard, thx for the explanantion i wil remove the changes made to original_id
Regarding the mantis bug report, whats the intended behaviour in using the filters? Currently if you filter for a test only questions with original_id are shown whereas without filters only tests without a original_id are shown. Theres never both. I changed it so it leaves the original_id out of the sql statement.

kergomard:
Hi @fhelfer 
Thank you very much for the update! I will have a look at it asap.
Now for your question:
- I don't think the original_id should have any relevance. Would we want to be precise, I think we should copy the question keeping the `original_id` linking back to the Question Pool, but as the use of the `original_id` might actually be - let us say - interesting, we shouldn't touch any copying behavior right now. So: Just ignore the `original_id`.

@kergomard

Amstutz:
note, if possible it would be nice if the PR could be picked down to 7 as well, see ticket.

kergomard:
Hi @fhelfer 
Thanks for adapting the fix! ...and now we run into issues because of the preexisting behavior.
With your fix: Which, from where I stand is essentially right, we now see every question in tests created in ILIAS7 or earlier twice, because they all have a copy that was created, when the question was created. But wait, it gets better: These copies of the questions are also not deleted, when the question in the test is deleted. So, with your fix, we now also find questions that have been deleted.
The good thing is: The `ilTestQuestionTableGUI` does it right and only shows the questions that really need to be there.
The questions are fetched here (if I'm not mistaken): [`ilObjTest::&getTestQuestions()`](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Modules/Test/classes/class.ilObjTest.php#L7575). Now the `ilTestQuestionBrowserTableGUI` does something completely different and gets its data from [`ilAQuestionList::load()`](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Modules/TestQuestionPool/classes/class.ilAssQuestionList.php#L617). I think the issue here arises because we are using the (partially) same query as for Questions from a Pool. I think we will need to split the two queries up more thorougly.

@kergomard

dsstrassner:
Hi @fhelfer and @kergomard, do we achieve the solution of this issue by 2023/07/18? Or do we have to wait until @kergomard comes back in August?

fhelfer:
Hi @kergomard,

I'm confident that this solution shows the correct amount of questions in the Browser.
Deleted questions are excluded due to the new joins, and copied questions are now included.
However duplicates in other tests will also be displayed. This happens because of how the "original_id" works. So, we need to choose whether we're okay with having both duplicates and copies, or if we'd rather have none.
If we want to escape these scenarios, we need to change the copy process.

But maybe im wrong about this. This whole thing kinda confuses me a bit :)

Best,
@fhelfer

kergomard:
Hi @fhelfer !
I think we should go with this and it is completely fine to show duplicates! Assume you copy a question with the copy-function: The only reason I could see to do this, is to create a variation of the question. Why would you want to have two identical copies of a question in one test? So: We can safely assume that the two are actually no duplicates, just two similar questions.

I've one question though, before I merge:
- Why do you need the `LEFT JOIN` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5929/files#diff-ea6bc8d559654b8d8f0fdccbba4104dfeb134a148ed6ee62f157ee6bbfdd7fddR519). I'm completely unclear what it should accomplish: You might have either a table with the right half all null or one with the right half set to something, but we never access this fields, right?

Thanks again and best,
@kergomard

fhelfer:
> * Why do you need the `LEFT JOIN` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5929/files#diff-ea6bc8d559654b8d8f0fdccbba4104dfeb134a148ed6ee62f157ee6bbfdd7fddR519). I'm completely unclear what it should accomplish: You might have either a table with the right half all null or one with the right half set to something, but we never access this fields, right?
> 
I missed this when copying from ilObjTest::&getTestQuestions()
Removed it

kergomard:

This now looks good to me. Could you please rebase and force push to resolve the conflicts? I will then merge.

Best,
@kergomard

PR #5931, opened 2023-05-22
Dashboard refactor for ILIAS9 / PHP8.1

iszmais:
Major Changes:
- Fixed Licencse Header
- Remove/updated trivila/missleading docs
- and `final` and `readonly`
- update constructors
- applied single quitation
- applied short array syntax
- removed trivial local variables
- applied dynamic called classnames
- removed unused functions
- enhanced type safety
- removed `class` prefixes
- removed `a_` prefixes
- partially changed names from desktop to dashboard
- minor bugfixes
- applied PSR-12


Services/Dashboard/Access/DashboardAccess.php
```diff
@@ -3,28 +3,36 @@
...
- * Dashboard permission wrapper
- *
  * @author Alexander Killing <killing@leifos.de>
  */
 class DashboardAccess
 {
-    protected \ilRbacSystem $rbac_system;
-    protected \ilDBInterface $db;
-    protected static int $setting_ref_id = 0;
+    protected readonly ?ilRbacSystem $rbac_system;
```
lscharmer: Why are the properties nullable? They where made readonly and assigned in the constructor, so they will never be null.

Services/Dashboard/Achievements/classes/ilAchievements.php
```diff
@@ -95,10 +84,9 @@ public function isAnyActive(): bool
     }
 
     /**
-     * Get active services
      * @return int[]
      */
-    public function getActiveServices(): array
+    final public function getActiveServices(): array
     {
         return array_filter($this->services, function ($s) {
```
lscharmer: As you changed something nearby, would you adjust this too?
```suggestion
        return array_filter($this->services, $this->isActive(...));
```
Please see: https://www.php.net/manual/en/functions.first_class_callable_syntax.php

Services/Dashboard/Administration/classes/class.ilObjDashboardSettingsGUI.php
```diff
@@ -363,9 +352,12 @@ private function maybeDisable(array $fields): array
             return $fields;
         }
 
-        return array_map(static function (FormInput $field): FormInput {
-            return $field->withDisabled(true);
-        }, $fields);
+        return array_map(
+            static function (FormInput $field): FormInput {
```
lscharmer: Please use the short function syntax.

Services/Dashboard/GlobalScreen/classes/DashboardLayoutProvider.php
```diff
@@ -38,13 +41,13 @@ public function isInterestedInContexts(): ContextCollection
...
         $this->data_collection = $screen_context_stack->current()->getAdditionalData();
-        if (!$this->data_collection->is(\ilDashboardGUI::DISENGAGE_MAINBAR, true)) {
+        if (!$this->data_collection->is(ilDashboardGUI::DISENGAGE_MAINBAR, true)) {
             return null;
         }
 
         return $this->globalScreen()->layout()->factory()->mainbar()
             ->withModification(
-                function (MainBar $mainbar): ?MainBar {
+                static function (MainBar $mainbar): ?MainBar {
```
lscharmer: Please use the short function syntax.

Services/Dashboard/ItemsBlock/classes/class.ilPDSelectedItemsBlockViewGUI.php
```diff
@@ -189,37 +190,40 @@ protected function groupItemsByStartDate(): array
...
+            }
+            if ($left['start']->get(IL_CAL_UNIX) > $right['start']->get(IL_CAL_UNIX)) {
                 return -1;
             }
 
             return strcmp($left['title'], $right['title']);
         });
 
-        uasort($groups['not_dated'], function ($left, $right) {
+        uasort($groups['not_dated'], static function ($left, $right) {
```
lscharmer: Please use the short function syntax.

Services/Dashboard/ItemsBlock/classes/class.ilPDSelectedItemsBlockViewGUI.php
```diff
@@ -244,7 +248,7 @@ protected function groupItemsByStartDate(): array
             $ongoing,
             $ended,
             $not_dated
-        ], function (ilPDSelectedItemsBlockGroup $group) {
+        ], static function (ilPDSelectedItemsBlockGroup $group) {
```
lscharmer: Please use the short function syntax.

Services/Dashboard/ItemsBlock/classes/class.ilPDSelectedItemsBlockViewGUI.php
```diff
@@ -254,20 +258,18 @@ protected function groupItemsByStartDate(): array
      */
     protected function groupItemsByLocation(): array
     {
-        $grouped_items = array();
+        $grouped_items = [];
 
         $items = $this->provider->getItems();
 
-        $parent_ref_ids = array_values(array_unique(array_map(function ($item) {
+        $parent_ref_ids = array_values(array_unique(array_map(static function ($item) {
```
lscharmer: Please use `array_column` because it does what the anonymous function does.

Services/Dashboard/classes/class.ilDashboardGUI.php
```diff
@@ -56,126 +60,105 @@ public function __construct()
...
                 break;
-
-                // pd notes
-            case "ilpdnotesgui":
-                if ($ilSetting->get('disable_notes') && $ilSetting->get('disable_comments')) {
+            case strtolower(ilPDNotesGUI::class):
+                if ($this->settings->get('disable_notes') && $this->settings->get('disable_comments')) {
                     $this->tpl->setOnScreenMessage('failure', $this->lng->txt('permission_denied'), true);
-                    ilUtil::redirect('ilias.php?baseClass=ilDashboardGUI');
+                    ilUtil::redirect('ilias.php?baseClass=' . __CLASS__);
```
lscharmer: Please use `self::class` instead of `__CLASS__`.

Services/Dashboard/classes/class.ilDashboardGUI.php
```diff
@@ -365,109 +330,101 @@ public function getRightColumnHTML(): string
...
-            $this->ctrl->setParameter($this, "cal_view", $cal_view);
+        if (array_key_exists('cal_view', $query_params) && $query_params['cal_view']) {
+            $cal_view = $query_params['cal_view'];
+            $this->ctrl->setParameter($this, 'cal_view', $cal_view);
         }
 
-        if (!empty($query_params["cal_agenda_per"])) {
-            $cal_period = $query_params["cal_agenda_per"];
-            $this->ctrl->setParameter($this, "cal_agenda_per", $cal_period);
+        if (!empty($query_params['cal_agenda_per'])) {
```
lscharmer, line 403: Please change this method to either use empty(...) or `array_key_exists(...) && ...`. Because they are both the same.

Services/Dashboard/templates/default/tpl.dash_obj_row.html
```diff
@@ -1,7 +1,7 @@
 <tr>
 	<td>
-	<input type="checkbox" name="id[]" value="{ID}" />
-	<img src="{ICON}" class="ilIcon">
-	{TITLE}
+		<input type='checkbox' name='id[]' value='{ID}' />
+		<img src='{ICON}' class='ilIcon'>
+		{TITLE}
 	</td>
 </tr>
```
lscharmer: Please add the newline at the end of the file.

Services/Dashboard/templates/default/tpl.dashboard.html
```diff
@@ -1,3 +1,3 @@
-<div class="ilDashboardMainContent">
-{CONTENT}
+<div class='ilDashboardMainContent'>
+    {CONTENT}
 </div>
```
lscharmer: Please add the newline at the end of the file.

Services/Dashboard/templates/default/tpl.dashboard_list_block.html
```diff
@@ -1,25 +1,29 @@
...
 <!-- END row_type_1 -->
 <!-- BEGIN row_type_2 -->
-<div class="ilObjListRow">
+<div class='ilObjListRow'>
 <!-- END row_type_2 -->
-<!-- <td class="std" valign="top" width="1">--><!-- BEGIN block_row_image --><!-- <img border="0" src="{ROW_IMG}" alt="{ROW_ALT}" title="{ROW_ALT}"/> --><!-- END block_row_image --><!-- {ROW_NBSP}</td>-->
-<div class="ilCLI">{BLOCK_ROW_CONTENT}</div>
+<!-- <td class='std' valign='top' width='1'>-->
+	<!-- BEGIN block_row_image -->
+	<!-- <img border='0' src='{ROW_IMG}' alt='{ROW_ALT}' title='{ROW_ALT}'/> -->
```
lscharmer: Why is there a comment of an image? Is this block useless?

lscharmer (changes requested):

Services/Dashboard/classes/class.ilDashboardGUI.php
iszmais: As far as i know its not the same since empty also checks for "empty" values like `null`, `""` and so on.

Further im pretty sure empty should be faster since its conceptional check is less complex.

lscharmer: The first expression is: `array_key_exists($key, $query_params) && $query_params[$key]` so it is equivalent to `empty`). I don't care which to choose but it should be the same in both cases.

Services/Dashboard/templates/default/tpl.dashboard_list_block.html
iszmais: Im not sure either but i assume thats template code wich will be activated by javascript somehow.

Im not a fan of this but i wouldnt consider it part of this refactoring to clean that up.

lscharmer: If this is modified with JS there would be an ID or non generic class name. Additionally the HTML would be probably not commented out but made invisible with display none. To use a HTML comment as a JS template should not be done and is very unlikely that it is the case in this file.
Please remove.

iszmais: Further more: i dont even know if that template is used anyway and i couldnt find any occurence of it.

Therefore i removed the complete file.

Services/Dashboard/classes/class.ilDashboardGUI.php
iszmais: Wich first expression?

lscharmer: https://github.com/ILIAS-eLearning/ILIAS/blob/2c335dde69dd84383621f5cc6596e1095d4f673b/Services/Dashboard/classes/class.ilDashboardGUI.php#L398

Services/Dashboard/templates/default/tpl.dashboard_list_block.html
iszmais: > If this is modified with JS there would be an ID or non generic class name. Additionally the HTML would be probably not commented out but made invisible with display none. To use a HTML comment as a JS template should not be done and is very unlikely that it is the case in this file. Please remove.

I agree that that should be the case and i wish i havent seen otherwise :smile:

PR #5932, opened 2023-05-23
Fix sending header instead of requesting return headers in return body

pascalseeland:


mjansenDatabay:
@jeph864 Can you please cherry-pick this to `release_8` and `trunk`?

PR #5934, opened 2023-05-24
Init: Add help screen ids

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=30705

If approved, this could be picked to `release_8` and `release_7`.


pascalseeland:
Cherry-Picked to 7 and 8

PR #5935, opened 2023-05-24
Authentication: Add SQL index to improve session statistics performance

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=31615

If approved, this could be picked to `release_8`.


pascalseeland (changes requested):
As this is bugfix I would target it at release8 and change the class name accordingly.

mjansenDatabay:
I created a new PR for `release_8` here: #6131

PR #5936, opened 2023-05-24
Remove alternative type short for learning histories

iszmais:
While refactoring i stumble upon 3 different types for learning history `lhist` `lhts` and `lht`.
While `lhist` seems to be widely used the other to have a rarely occurence in some unused languages.
only the symbol seem to use the icon of `lhts` istead of `lhist` but both icons look exact the same.

Im not sure if this are artifacts or inconsistencies but i would recommend removing the lesser used type short to prevent an entangled dublicated maintainence in future.

Or pls correct my if i understand the usage wrong here


alex40724:
Thanks for the PR. These are indeed different things. "lhts" is the object type of the "learning history settings"(!) node. Object types are currently stored in the database. So just changing it in the object will lose its reference to the database entry, I am pretty sure that this comes with some bad side effects, even if I did not test it. "lhist" is the ID of the component, it is also used for the page content ID.

iszmais:
@alex40724 
Ah ok my bad then. Thank you very much for taking the time for a further explanation. :+1: 

Greetings,
@iszmais

PR #5937, opened 2023-05-25
Test: Fix form handling

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=37383

I **strongly** recommend to **thoroughly** test this before merging this.

### DIFF
--- a/Modules/Test/classes/class.ilObjTestSettingsGeneralGUI.php
+++ b/Modules/Test/classes/class.ilObjTestSettingsGeneralGUI.php
@@ -200,7 +200,7 @@ class ilObjTestSettingsGeneralGUI extends ilTestSettingsGUI
         return $this->saveFormCmd(true);
     }
 
-    private function saveFormCmd($isConfirmedSave = false)
+    private function saveFormCmd($isConfirmedSave = false) : void
     {
         $form = $this->buildForm();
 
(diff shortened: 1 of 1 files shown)


Modules/Test/classes/class.ilObjTestSettingsGeneralGUI.php
```diff
@@ -353,22 +349,23 @@ private function saveFormCmd($isConfirmedSave = false)
...
                 (
-                    $starting_time instanceof ilDateTime &&
-                (int) $ending_time->getUnixTime() < (int) $this->testOBJ->getStartingTime()
-                )
-                ||
-                (int) $ending_time->getUnixTime() < (int) $this->testOBJ->getStartingTime()
+                    is_string($starting_time) && $starting_time !== '' &&
+                    $ending_time < $starting_time
+                ) ||
+                (new ilDateTime($starting_time, IL_CAL_DATETIME))->get(IL_CAL_UNIX) < (int) $this->testOBJ->getStartingTime()
```
fneumann: must be $ending time in this comparison - otherwise it is not longer possible to change the starting time to a value prior to the saved one

fneumann (changes requested):

Modules/Test/classes/class.ilObjTestSettingsGeneralGUI.php
mjansenDatabay: Good catch @fneumann , fixed!

fneumann (approved):

kergomard:
Ok, I merge this. No need for this to be cherry-picked to newer versions. The problem should not occur there, as far as I can see.

PR #5938, opened 2023-05-25
UI: URL abstraction for PHP and JS

tbongers-cat:
This is a new implementation of a **URL abstraction** class for both PHP and Javascript. 

The class provides better control over which component adds or removes parameters from a URL by adding tokens for each newly added parameter. All parameters are also required to have a designated namespace to avoid overwriting parameters with the same name. In addition, the PHP class provides a render function to generate the same URL object in Javascript, so that it can be used to dynamically control parameters on the JS side as well.

Please see the interfaces for a more detailed description of the usage of this class. Let me know if you think this should be put in a different location inside the code or if you have any further questions.

@klees might want to add more insights on the potential usage of this, as it was his proposal :)


klees:
Before I write a review of the code, I want to expand a little on the motivation to include a facility like this. So here we go:

Traditionally we use `ilCtrl::setParameter(ByClass)` to include parameters in generated URLs. The general idea is that multiple components need to have a possibility to attach parameters to the URL, since multiple components might be active for a given page. The `ilCtrl`-approach uses the global `ilCtrl` instance to store these parameters namespaced by command classes.

The general problem occurs again in the UI framework: various components (now UI components, not necessarily system components) need to contribute parameters to a URL. For instance, consider the Data Table: We will have various parameters contributed by `View Control`s or `Filter`s that need to be passed in URLs, e.g. for actions. For some controls, it even might not be possible or feasible to generate these URL server-side, e.g. because of combinatorics for something like a column selector.

This proposal aims to provide a solution to this problem, that:
* allows for coordinated usage of the (shared) namespace of URL parameters
* fits the functional style of the UI framework (hence no mutable global)
* provides a unified functionality on client and server side

The general idea is that UI components get access to certain URLs by a token. Tokens are attached to names of parameters and can be generated in a central location (e.g. a GUI class). The tokens for the parameters are passed to the UI components, which in turn can use them to create concrete URLs.

klees (changes requested):
Hi @tbongers-cat,

thanks a lot for building this.

- [ ] length: For my curiosity: Is [this](https://github.com/ILIAS-eLearning/ILIAS/pull/5938/files#diff-5a32c45e7425c7d6d063495e837f1969a7ab5cbc22a3ac12ae88cccfe7077c6bR25) really a thing for browsers? I know, that some other HTTP components (e.g. proxies) might indeed cut URLs after some length, but didn't know that browser might do the same...

- [ ] Location: Move the `URLBuilder` classes and interfaces to the UI framework. Several reasons: IMHO it is quite odd that we would have JS and PHP in different locations. I would not introduce JS in `src/Data` atm, though.
- [ ] interface 1: Remove the interface. I appreciate the muscle memory regarding seperating interfaces and implementation, but I don't think that we have anything to hide or abstract here, so I don't see value for that interface. I might be wrong, though =) Remember moving the nice documentation to the class, though.
- [ ] interface 2: Same, but for `URLBuilderToken`.
- [ ] constants: Do we really need to expose `URL_MAX_LENGTH` and `SEPARATOR` in both, PHP and JS implementations? They seem to be of no concern for the consumer. If so, please internalize them (move them outside of the `class` scope in JS).

- [ ] URI 1: Please use `URI`s instead of `string`s for in and out. If someone does not have an `URI` he could easily get one and we really want a valid `URI` not just any `string`... Also, if we give some an `URI`, he can easily turn it into a string.
- [ ] `getURI`: Do not internalize the `URI` to the builder. Instead implement `getURI(URI $base): URI`, which should mostly work as is internally. `renderObject` will then require an `URI` as well. We will regularly have the case, that we construct UI components without knowing exactly on which page they are actually located. The new interface will cater to that problem. Remember to adjust documentation accordingly. This will make the client-side interface slightly different from the server side interface. But I think this is fine: On the client side, we know which URL we're on.	
- [ ] `Exception`: Please throw an [`DomainException`](https://www.php.net/manual/en/class.domainexception.php) when a token is not valid. I mostly use PHP standard exceptions in my code. They feel expressive enough for my needs and also shouldn't be caught most of the time anyway... Document exceptions with `@throws`.
- [ ] `random_bytes()` might throw an exception, please implement similar to e.g. `ilCtrlTokenRepository::generateToken()`.
- [ ] `sizeof()`: We should use `empty()` instead of `sizeof(...) === 0`. several occasions.
- [ ] `URLBuilder.#analyzeURL()`: Please avoid functions with side-effects like this. Instead, implement functions with single responsibility accepting the necessary parameters and assign object properties explicilty by using them. E.g. `this.#parameters = getQueryParameters(query);`. Note, functions not using `this` should be put outside of the `class`es scope.
- [ ] `URLBuilderToken.#createToken()`: same here.
- [ ] `URLBuilderToken.getName()`: This looks like something that can already be done on construction, please add an according property.
- [ ] `set` and `get`: don't use it. Instead, introduce proper functions like e.g. `getFragment()` and `setFragment(...)`.
- [ ] JS unit tests: Please separate tests for `URLBuilder` and `URLBuilderToken`.

Kind regards!

tbongers-cat:
Hi @klees ,
thanks for the feedback =)

Question about **URL length**: 
There is no definitive answer to this according to my research. Different browsers have different limits for the URL length, both in their internal usage as well as in the length of a URL that can be entered in their address bars. In most modern browsers, these values seem to be pretty high, but former browsers (e.g. Internet Explorer) had a limit as low as ~2000 characters. 
The more important factor in our use case may be the limit set by the webserver. Both Apache and NGinx limit the length of the request to 8 kb in their standard configurations (see [LimitRequestLine](https://httpd.apache.org/docs/current/mod/core.html#limitrequestline) / [large_client_header_buffers](https://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers)), which amounts to ~8000 characters (_without taking into account characters that need to be URL encoded_). While these limits can be set higher, they will most likely be unchanged in most web servers.
The value of 2048 that I chose as maximum may be set a little low and checking the URL length may only be valid for some edge cases, but imho it can't hurt to be on the safe side here. I will gladly set this to a higher standard value if needed.

Suggestions:
- [x] **Location**: Moved the classes to `src/UI` and the tests to `test/UI` (*changed as requested*)
- [x] **Interface**: Removed both interfaces and moved the comments to the class files
- [x] **Constants**: The constants on the JS side have been made into private properties, as there is indeed no need to access them externally.

Changes:
- [x] **URI**: The PHP class now uses `\ILIAS\Data\URI` as in- and output, the JS class is using `URL` instead of simple strings. Both types can be cast to string easily, as you mentioned.
- [x] `getURI()`: I don't quite agree with your suggestion here, because the root URI is not passed when calling `getURI()`, but on **construction** of `URLBuilder`, which also handles all pre-existing parameters and makes them "claimable" once. The function of `getURI()` is to (now) return a URI object of the current state of the `URLBuilder` (which may or may not have been changed by writing or deleting parameters before). From my understanding, a UI component either gets an existing URLBuilder object from another component or it has to instantiate a new object itself and pass the base URI at that time. Please correct me if I misunderstood your intention here.
- [x] **Exception**: I have changed the Exception as proposed and also revised another exception to a native PHP type. If you think any of the other exceptions could be replace with native exceptions, please let me know.
- [x] `random_bytes()`: Adapted the solution from `ilCtrl `for any unwanted exceptions
- [x] `sizeof()`: Replaced all occurences
- [x] `URLBuilder.#analyzeURL / URLBuilderToken.#createToken`: Revised both functions as proposed
- [x] `URLBuilderToken.getName()`: Functionality moved to constructor
- [x] `set / get`: Removed/replaced by properly named methods
- [x] **Unit tests**: Added separate unit tests for `URLBuilderToken` (both PHP and JS). On the JS test for `URLBuilder`, the `URLBuilderToken` is still required to test the important variant of constructing the URLBuilder with existing tokens.

tbongers-cat:
@klees I've added `withURI()` and renamed `getURI()` to `buildURI()` as discussed. The parameters from the base URI are now only checked during the build process of the new URI, while claimed parameters always take precedence. Please have a look =)

klees (approved):
LGTM!

thibsy (changes requested):
Hi @tbongers-cat

Thank you for implementing the changes as requested. There still are some minor things left to do before we can merge this.

- [ ] Immutability: I also noticed that being immutable with the `URLBuilder` puts us at a disadvantage in the parameter-acquiring process. In order to be immutable you need to "acquire" the parameter in a clone of the builder, which would be lost however, since you need to return a `URLBuilderToken` in this method. This leads to returning an array|object instead, which has multiple values for different purposes, which is IMO something we should avoid. Which brings me to the question, do we really need to be immutable? If we really do, the JavaScript implementation should also be immutable, it is not at the moment.

- [x] `URLBuilder.acquireParameter()`: In PHP you return an associative array, which is good because you can access values by name. In JavaScript you should do the same with an object (`{key: val}`) to achieve a similar result. Also, the return values of both methods should be documented more precisely, if possible with according annotations.
- [x] `URLBuilder.#urlMaxLength`: This value is currently a private property of the object, not a constant. For public constants you should use `static`, as you correctly did with `URLBuilderToken.tokenLength`, but private ones should be declared with `const` outside of the classes scope. With this approach the property will be initialized for every new instance.
- [x] `URLBuilder.#separator`: Same here.

Kind regards,
@thibsy

**EDIT**: added question about immutability.

tbongers-cat:
Thanks again for your feedback @thibsy 

I have reworked some of the JS code to make it align better with the recent changes on the PHP side. Also I have moved all **constants** outside the class scope as requested, as they don't need to be public. 

The return type of `acquireParameter(`) has been changed to an object on the JS side and documented better on PHP and JS to reflect the returned values.

Concerning the question of **immutability**: We have discussed this internally and don't think that we can give up this principle here, especially as the URLBuilder is an object that will possibly be passed between components later. Therefore having immutable objects is the way to go to avoid any interferences and the return of an array/object instead of a single value seems to be the better choice.

For the JS side, immutability is not implemented yet, as I'm honestly at a loss on how to do this properly. The proposed way of cloning according to eslint would be spreading to a new object, but that leaves me with two major problems: 
**a)** We only create a **shallow copy** which would not respect the contents of the Maps included as properties and 
**b)** I have found no way to access the clone's **private properties** afterwards (especially `#parameters`) to include the necessary changes.
If you have any advice on how to do this or any good examples, please let me know and I will gladly adopt them.

On the failed copyright check for `dist/ui.js`: Is there a way to configure this via `rollup.config.js`? Otherwise I don't see how this auto-generated file can have a proper copyright header.

Amstutz:
Hi @tbongers-cat 

Thx for answering the questions above. Your answers reasonate with me. IMO this is fine for merging. Your questions are best answered by @thibsy, who is back in a few weeks. We definitely have to look into the failing copyright check in our dist js. However, this should not prevent a merge here.

Thx a lot.

thibsy:
Hi @tbongers-cat

I am sorry for the delayed response here. I didn't wanted to rush this but @Amstutz beat me to the punch :).

- [x] failing copyright-check: for future reference you can have a look at the [bundling documentation](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js-bundling.md#rollup-configuration). You can provide a `banner` property with the copyright license header as a value in order to include it for generated files. I'll fix this later today, so no action required.
- [ ] JavaScript deep copies: I think we could achieve the desired result of immutability by using [document.structuredClone()](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone). I'll have to check this though.
- [x] Immutability: I can see the importance of immutability, however, I still find the acquiring process should be improved in the future, so we don't need to work with an ambiguous value. I don't have an idea yet, but I have put this on my agenda :). This machinery is still extremely useful and perfectly fine to work with at this stage.

Thanks again for this contribution!

Kind regards,
@thibsy

PR #5939, opened 2023-05-25
Calendar: Fix handling of incoming date in `ilDateTimeInputGUI`

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=37383

This PR fixes an issue with the `ilDateTimeInputGUI` in case the container form is **invalid**.

Let's suppose the following settings are given:
* Global timezone in the ILIAS calendar administration: An arbitrary timezone
* User timezone: Europe/Berlin

Let's further suppose the user enters/selects the following datetime string with the calendar picker in the user interface:
* 25.05.2023 09:00

In ILIAS 7, the problem is:
1. In `\ilDateTimeInputGUI::checkInput` the HTTP POST variable (initially: `25.05.2023 09:00`) is mutated by setting a normalized datetime string (with an applied UTC timezone) when `$_POST[$this->getPostVar()] = $this->getDate()->get($post_format);` is called.
2. As a result, the value in $_POST changes from `25.05.2023 09:00` (Europe/Berlin) to `25.05.2023 07:00` (UTC). `\ilDateTimeInputGUI::$date` is a `ilDateTime` instance with an internal `DateTime` set to `2023-05-25 09:00:00.000000` with `Europe/Berlin` as the interpreted timezone.
3. If the containing form is invalid an the consumer calls `$form->setValuesByPost();`
4. The `$incoming` datetime string (`25.05.2023 07:00`) is parsed again. `\ilDateTimeInputGUI::$date` is  now a  `ilDateTime` instance with an internal `DateTime` set to `2023-05-25 07:00:00.000000` with `Europe/Berlin` as the interpreted timezone. So the offset is now 2 hours.
5. This datetime string is now rendered in the user interface, which is wrong.

This seems to be **not** an issue for ILIAS >= 8.x.

### DIFF
--- a/Services/Form/classes/class.ilDateTimeInputGUI.php
+++ b/Services/Form/classes/class.ilDateTimeInputGUI.php
@@ -32,6 +32,8 @@ class ilDateTimeInputGUI extends ilSubEnabledFormPropertyGUI implements ilTableF
      * @var bool
      */
     protected $side_by_side = true;
+    /** @var ilDateTime|null */
+    protected $valid_incoming_datetime = null;
 
     /**
     * Constructor
@@ -166,9 +168,13 @@ class ilDateTimeInputGUI extends ilSubEnabledFormPropertyGUI implements ilTableF
     */
     public function setValueByArray($a_values)
     {
-        $incoming = $a_values[$this->getPostVar()];
-        $this->setDate(ilCalendarUtil::parseIncomingDate($incoming, $this->getDatePickerTimeFormat()));
-                
+        if ($this->valid_incoming_datetime !== null) {
+            $this->setDate($this->valid_incoming_datetime);
+        } else {
+            $incoming = $a_values[$this->getPostVar()];
+            $this->setDate(ilCalendarUtil::parseIncomingDate($incoming, $this->getDatePickerTimeFormat()));
+        }
+
         foreach ($this->getSubItems() as $item) {
             $item->setValueByArray($a_values);
         }
@@ -231,6 +237,7 @@ class ilDateTimeInputGUI extends ilSubEnabledFormPropertyGUI implements ilTableF
                 $post_format = $this->getShowTime()
                     ? IL_CAL_DATETIME
                     : IL_CAL_DATE;
+                $this->valid_incoming_datetime = $this->getDate();
                 $_POST[$this->getPostVar()] = $this->getDate()->get($post_format);
             } else {
                 $_POST[$this->getPostVar()] = null;


mjansenDatabay:
FYI @mbecker-databay / @kergomard / @dsstrassner

mjansenDatabay:
@smeyer-ilias Any objections or questions regarding this change?

kergomard:
@smeyer-ilias : Any chance that you could have a look at this? It would be nice, if we could close the Issue mentioned above.

Thank you very much and best,
@kergomard

dsstrassner:
Dear JourFixe, we need this PR to resolve the mentioned issue. Many ILIAS users need correctly working start and end dates, even across time zone boundaries. 

Can this PR please be merged shortly after the next JourFixe?

alex40724:
This has been wrongly assigned (Services/Form, not Calendar).

@mjansen: Thanks for the PR. If there is any instance calling checkInput and providing the input array not via setValuesByPost this might fail, right? Hard to tell if there will be any side effects, but for the usual process this looks good. Only that setValueByArray() will keep it's name but the behaviour will be "set value by array only if checkInput has not been called."

There are 94 places using ilDateTimeInputGUI. I hope that none of these "works around" this somehow. There is no indication that this behaviour has been introduced recently, right? This might be this way since 10+ years.

ILIAS 8 doesn't have this issues, since it does not mess with the $_POST variable. But trying to adress this by changing `$_POST[$this->getPostVar()] = $this->getDate()->get($post_format);` and keep the value in the user timezone may make it worse for consumers that access $_POST directly afterwards.

mjansenDatabay:
Hi @alex40724 ,

> If there is any instance calling checkInput and providing the input array not via setValuesByPost this might fail, right? 

Maybe: But if legacy forms are used like intended/described here https://docu.ilias.de/goto_docu_pg_213_42.html or adapted by copying/pasting code from the majority of ILIAS forms, there should not be any issue at all.

>  There is no indication that this behaviour has been introduced recently, right? This might be this way since 10+ years.

You are right.

Best regards,
@mjansenDatabay

matthiaskunkel:
Jour Fixe label has been set before correct assignment of maintainer and related component (Alexander, Services/Forms). There is currently no need for a discussion of this PR in the JF. I assume the PR will be committed soon after Alexander's return from vacation.

alex40724:
Mmh, I thought the change of the setValueByArray() function could have been discussed. I assume that everyone is ok with this change.

PR #5941, opened 2023-05-26
[FIX] 0037374: Remove superfluous directories

chfsx:
Hi @mjansenDatabay 
As discussed in Discord, here is the PR for the improvement that directories after the first .php file should not be considered for the constant ILIAS_HTTP_PATH.


mjansenDatabay:
Thx a lot @chfsx !

PR #5943, opened 2023-05-27
Database Update Steps to convert whole database to UTF8MB4

Akumatic:
### Motivation
With current versions of MySQL and MariaDB, the character set `utf8` is an alias for `utf8mb3`. It uses up to three bytes per character to encode text. At the moment, ILIAS only supports utf8mb3.

The following points are incentives to convert the database to a more modern character set:

- `utf8mb3` only supports Basic Multilingual Plane (BMP), `utf8mb4` also supports supplementary characters
- According to the MySQL 8.0 documentation, `utf8mb3` is deprecated and is expected to be removed in future MySQL versions

### Implemented changes
This pull request implements database update steps to convert the character set of the ILIAS database itself, all tables and text containing columns to `utf8mb4`.

With the current state, most tables are converted without problems. Two tables have to be altered manually and feedback of the maintainer is required (see below). This is also implemented here and will either be kept or edited after feedback is received. In addition, there is a db update class in the trunk branch that leads to an index prefix that is too long causing the steps to stop.

No exception handling is implemented, as the database update should actually fail if a conversion step fails.

#### Service: ResourceStorage (@chfsx)
⚠️ This causes the update or setup to fail and needs to be resolved before merging. 
- Affected Class: [ilResourceStorageDB90](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Services/ResourceStorage/classes/Setup/DB/class.ilResourceStorageDB90.php)
- For the primary key, the columns `rid`, `revision`, `definition_id` and `variant` are used.
  - in total, an integer of length `8` and text of length `64 + 64 + 768 = 896`
  - with 4 bytes per character, the text part alone allocates at least `3584` bytes
  - with DYNAMIC row format of InnoDB, only up to `3072` bytes are supported for index prefixes

#### Module: Course (@smeyer-ilias)
- Affected database table: `crs_settings`
- The maximum row size of MySQL is `65535` bytes. The affected table contains four columns of type `varchar` and length `4000`, which sums up to `64000` bytes with the usage of `utf8mb4`. With the remaining columns, the row exceeds the limit and an error is thrown
- Columns with type `varchar(4000)` in this context are:
  - syllabus
  - contact_consultation
  - important
  - target_group
- To fix this issue, the affected four columns could be converted to type `text` which is stored in overflow pages.
- The necessary update steps are implemented in the function `convertCharsetColumnsSpecialCases`. 

#### Module: Session (@smeyer-ilias)
- Affected database table: `event`
- Same problem as with table crs_settings of Course above.
- Columns with type `varchar(4000)` in this context are:
  - description
  - location
  - tutor_name
  - details

### DIFF
--- a/Services/Database/classes/Setup/class.ilDatabaseUTF8MB4Agent.php
+++ b/Services/Database/classes/Setup/class.ilDatabaseUTF8MB4Agent.php
@@ -1,0 +1,32 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+use ILIAS\Setup;
+
+class ilDatabaseUTF8MB4Agent extends Setup\Agent\NullAgent
+{
+    public function getUpdateObjective(\ILIAS\Setup\Config $config = null): \ILIAS\Setup\Objective
+    {
+        return new \ilDatabaseUpdateStepsExecutedObjective(new ilDatabaseUTF8MB4UpdateSteps());
+    }
+
+    public function getStatusObjective(\ILIAS\Setup\Metrics\Storage $storage): \ILIAS\Setup\Objective
+    {
+        return new \ilDatabaseUpdateStepsMetricsCollectedObjective($storage, new ilDatabaseUTF8MB4UpdateSteps());
+    }
+}
--- a/Services/Database/classes/Setup/class.ilDatabaseUTF8MB4UpdateSteps.php
+++ b/Services/Database/classes/Setup/class.ilDatabaseUTF8MB4UpdateSteps.php
@@ -1,0 +1,227 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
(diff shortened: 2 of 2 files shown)


chfsx:
Hi @Akumatic 

Thank you for this PR and the push in the direction of being able to run ILIAS with mb4!

I have set up a [PR](https://github.com/Akumatic/ILIAS/pull/1) for you that adjusts a few small things, otherwise the PR already looks very good from my point of view. But what is certainly still needed is intensive testing. For example, I could imagine that, depending on the database configuration or version, some things are listed UPPERCASE and sometimes lowercase; I have had these problems in the past. You would certainly have to check whether this is the case. In addition, the PR with a database that is as old and as large as possible would have to be checked, especially with regard to performance, because otherwise we would have to switch to a migration instead of a DB update step.

We also still have to solve the three open issues (ResourceStorage, Course and Session); unfortunately I don't have a suitable idea for ResourceStorage yet.

Furthermore, the process requires that the feature request [Database: UTF8MB4 as default Character Set](https://docu.ilias.de/goto.php?target=wiki_1357_Database%3A_UTF8MB4_as_default%C2%A0Character_Set) is first discussed and accepted in the Jour Fix. Here, it is also important how to deal with future errors that occur due to this change. 

A short test (after the changeover) shows me that a lot of new problems may arise here, so you could try the following:

- Install trunk
- changeover to MB4
- patch ilDBPdoMySQL::doesCollationSupportMB4Strings to enable mb4
- Create a category in the magazine, and include an emoji in the title.

then various errors occur:
- Who\ExceptionErrorException thrown with message "unserialize(): Error at offset 177 of 758 bytes" in /var/www/html/Services/Navigation/classes/class.ilNavigationHistory.php:143
- PDOException thrown with message "SQLSTATE[22007]: Invalid datetime format: 1366 Incorrect string value: '\xF0\x9F\x98\x8D' for column `ilias`.`object_data`.`title` at row 1" in /var/www/html/Services/Database/classes/PDO/class.ilDBPdo.php:455
- sometimes you just get logged out of ILIAS, couldn't find out why yet... but probably has to do with the saved session in the database
- the emoji is not displayed

Let me know if I can be of any help.

Best regards


Fabian

kergomard:
Hi @chfsx  

Any Updates on this? Maybe this is also a case for a close due to inactivity?
Best regards!
@kergomard

chfsx:
@Akumatic please reopen if this gets relevant again (see my comments above)

PR #5944, opened 2023-05-30
User/Session: Allow Longer Session Ids

kergomard:
This increases the size of table-columns used for storage of the `session_id` to 256 varchars. It also shortens the value to 8 characters where it is used to generate length limited strings.
See: https://mantis.ilias.de/view.php?id=26954

### DIFF
--- a/Modules/DataCollection/classes/Helpers/class.ilDclPropertyFormGUI.php
+++ b/Modules/DataCollection/classes/Helpers/class.ilDclPropertyFormGUI.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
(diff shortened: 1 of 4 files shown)


Modules/DataCollection/classes/Helpers/class.ilDclPropertyFormGUI.php
```diff
@@ -51,7 +52,7 @@ public static function getTempFilename(
         $tmp_file_name = implode(
             "~~",
             array(
-                session_id(),
+                mb_substr(session_id(), 0, 8),
```
kergomard, line 55: @pascalseeland and @Amstutz : FYI (also for ILIAS 7 and trunk)

mjansenDatabay:
Thx @kergomard

PR #5945, opened 2023-05-30
Test: Update processing time on time-restricted tests without page reload

tbongers-cat:
Refering to the first part of https://mantis.ilias.de/view.php?id=22536#c75295 :
When extra time is added to time-restricted tests while the test is running, the timer for the participant is not updated unless the page is reloaded. This can lead to triggering the auto-submit too early without adding the extra time. 

This PR adds a polling method every 30 seconds that checks for the correct processing time (including extra time) and updates the timer accordingly. I have decided against just checking this before finally auto-submitting the test (as proposed in Mantis), as the participant might get irritated if the extra time is only displayed at the last second. 
This way, the update happens no more than 30 seconds after the proctor has added extra time via the dashboard. If you feel that the additional load of polling every 30 seconds is too much, the timing can of course be changed.

The second commit should be cherry-pickable for ILIAS 7 and 8 as well, the first one just contains some small fixes for PHP 8.1 and is only valid for trunk. The second part of the Mantis ticket (change of password during a test) will be handled separately in a future PR.

### DIFF
--- a/Modules/Test/classes/class.ilObjTestAccess.php
+++ b/Modules/Test/classes/class.ilObjTestAccess.php
@@ -541,7 +541,7 @@ class ilObjTestAccess extends ilObjectAccess implements ilConditionHandling
                 );
                 if ($result->numRows()) {
                     $row = $ilDB->fetchAssoc($result);
-                    if (trim($row['clientip']) != "") {
+                    if ($row['clientip'] !== null && trim($row['clientip']) != "") {
                         $row['clientip'] = preg_replace("/[^0-9.?*,:]+/", "", $row['clientip']);
                         $row['clientip'] = str_replace(".", "\\.", $row['clientip']);
                         $row['clientip'] = str_replace(array("?","*",","), array("[0-9]","[0-9]*","|"), $row['clientip']);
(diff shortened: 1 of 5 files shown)


dsstrassner:
Just for safety: Trunk / ILIAS 8 / ILIAS 7 @kergomard / @mbecker-databay 😺 

Thanks a lot @tbongers-cat ;-)

kergomard:
Thank you very much @tbongers-cat !
I think this is a simple and viable solution for the time being. The whole JS of the test needs refactoring, but that is not your problem.
I've got one question and one little request:
- Shouldn't we maybe also trigger this check just before sending the results (just in case)? We could then reset the time, if the new time is more than 60s in the future (or something like this). I'm unsure if this is really necessary, but it might give us an additional safeguard. What do you think?
- I quickly coordinated with Denis: It would be kind, if you would change the polling interval to 60s and move this number to a place, where it is a little bit more visible (maybe, for the time being, to [the definition of the vars](https://github.com/ILIAS-eLearning/ILIAS/pull/5945/files#diff-70778e02f29b6e65802088d9c214a225c49e5fc7062ddf3de343ce9f1881b335R28)?

Thanks again and best,
@kergomard

tbongers-cat:
@kergomard Thanks for the feedback. 
I have changed the interval to 60s and put it into a variable as suggested. The `redirect()` function now checks for changes in the processing time once more before submitting the form.

kergomard:
Thanks again @tbongers-cat !
Let's run with this.

PR #5946, opened 2023-05-30
User/Session: Allow Longer Session Ids

kergomard:
See: https://mantis.ilias.de/view.php?id=26954
Sister PR to: https://github.com/ILIAS-eLearning/ILIAS/pull/5944


mjansenDatabay:
Thx @kergomard

PR #5948, opened 2023-05-30
delos skin scss - remove Bootstrap dependency

catenglaender:
As part of the ITCSS SCSS stylecode refactoring, this PR removes Bootstrap 3 as a dependency from the delos default system style.

# Issues

* Bootstrap added thousands of lines of CSS code to Delos and any skin/system style derived from it.
* Many Bootstrap components were barely or not at all in use
* Some components have been heavily overwritten which resulted in a complex constructs that were half ILIAS specific - half outdated Bootstrap 3 code. This was very difficult to maintain.

# Goals

* Reducing many lines of unnecessary CSS code
* completely avoiding (or at least reducing) instances where code for the same place/element is overwritten
* Bootstrap code is not a bad thing in general. Some systems (like the grids and buttons) are very well-thought out and based on years of community experience. We want to keep those as starting points of high quality style code that can be further customized and optimized for the exact scope that we need in ILIAS.

# Changes

* Many lines of Bootstrap code have been removed completely as they were not used once in ILIAS.
* Parts of Bootstrap have been merged into their corresponding place within the ITCSS structure while also deleting classes and mixins that were not in use.
* Some of these sections were changed to be more specific to ILIAS or melted with the modern UI components into something that barely resembles the Bootstrap 3 origin.

# Impact

* at first glance there should only be very small visual differences compared to release_8
* margins, font-sizes, colors and paddings have slightly shifted due to the previous implementation not using the correct variables. This has been corrected for many components. Specifically, the alerts, form row/columns and headline typography are connected to variables and classes that have existed for years, but were never strong enough to surface on the places they were addressing.
* it's expected that we will find visual glitches and broken views as this has only been tested on a bare bone instance. These need to be fixed on a case by case basis. We may have to bring back more parts of Bootstrap if we realize that their removal has caused major problems or use this as a chance to implement ILIAS specific solutions.


Amstutz:
@catenglaender 

Thx a lot for proposing this huge milestone. We just lost 10'000 lines of CSS code. Excellent. 

There are several things in the layout that unintentionally changed. Overall, this PR should not lead to any changes in the UI. However, also the migration to SCSS left some minor issues in the current version of trunk. Since those are all minor design flaws and none-blocking errors, I decided to merge this as is. We will, however, open up the CSS squad and work through those issues piece by piece.

I hope this is fine for everybody.

Thx

PR #5949, opened 2023-05-31
css word-break in workflow listing, fixes long titles in learning sequence toc - mantis 37291

catenglaender:
Mantis: https://mantis.ilias.de/view.php?id=37291

# Issue

Long titles in learning sequence table of content overflow indefinitely.

This is actually the behavior of the UI component workflow listing.

image

# Changes

Long strings inside the UI workflow listing are now forced to break (using `word-break: break-word;`) even if `hyphens: auto` wouldn't add a break.

image

# Related

The page content shows the same problem (see heading in the screenshots). This should be tackled separately and with careful consideration for the scope of the change as it could negatively impact tables.

### DIFF
--- a/src/UI/templates/default/Listing/workflow.less
+++ b/src/UI/templates/default/Listing/workflow.less
@@ -1,3 +1,6 @@
+.il-workflow {
+	word-break: break-word;
+}
 .il-workflow-header {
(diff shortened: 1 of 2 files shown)


Amstutz:
Thx a lot! I will pick to 8 and trunk as well.

PR #5950, opened 2023-05-31
Test: Avoid data loss when changing password mid-test

tbongers-cat:
Refering to the second part of https://mantis.ilias.de/view.php?id=22536#c75295 :
When a test password is changed during a running test, every participant has to re-enter the password again. If the page is not reloaded (e.g. for essay questions), this mechanism is not triggered and autosaving as well as auto-submitting the test will fail. 

As proposed, the password will now just be checked **once** per test session and the session will be flagged as "authenticated". This allows autosave and auto-submit to work as expected. New test sessions will still require the changed password.

~~This PR adds a new field to the test session table (_tst_active_) for persistence, so it might just be viable for trunk.~~ It also handles an edge case that should not happen regularly.

### DIFF
--- a/Modules/Test/classes/class.ilTestPlayerAbstractGUI.php
+++ b/Modules/Test/classes/class.ilTestPlayerAbstractGUI.php
@@ -1793,7 +1793,6 @@ abstract class ilTestPlayerAbstractGUI extends ilTestServiceGUI
 
         if (!$this->passwordChecker->isPasswordProtectionPageRedirectRequired()) {
             $this->testSession->setPasswordChecked(true);
-            $this->testSession->saveToDb();
             return;
         }
 
--- a/Modules/Test/classes/class.ilTestPlayerAbstractGUI.php
+++ b/Modules/Test/classes/class.ilTestPlayerAbstractGUI.php
@@ -1777,11 +1777,23 @@ abstract class ilTestPlayerAbstractGUI extends ilTestServiceGUI
 
     protected function handlePasswordProtectionRedirect()
     {
+        /**
+         * The test password is only checked once per session
+         * to avoid errors during autosave if the password is
+         * changed during a running test.
+         * See Mantis #22536 for more details.
+         */
+        if ($this->testSession->isPasswordChecked() === true) {
+            return;
+        }
+
         if ($this->ctrl->getNextClass() == 'iltestpasswordprotectiongui') {
             return;
         }
 
         if (!$this->passwordChecker->isPasswordProtectionPageRedirectRequired()) {
+            $this->testSession->setPasswordChecked(true);
+            $this->testSession->saveToDb();
             return;
         }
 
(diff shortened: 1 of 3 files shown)


kergomard:
Hi @tbongers-cat 
Thank you very much for this PR! I think we are getting there.

I have one conceptual doubt, though: Doesn't this maybe take the term 'TestSession' slightly too literal? When reading through this, I wondered if it wouldn't just be enough - I would even think: better - to just set this in the PHP-Session? If I'm understanding this right, your solution would mean, that the password would be asked once per 'TestPass', but I would think, that, if a user needs to login again, because of a computer issue, or maybe because the duration is three days and there is work to be done in steps, she would need to reenter the - potentially changed - password. Am I missing something?

@kergomard

tbongers-cat:
Hi @kergomard 
Thanks for the feedback. I agree that storing this with the test session might a bit too much, therefore I have changed it to just store the information in the PHP session. Please have a look =)

kergomard (commented):
Thank you very much @tbongers-cat  for the adaption of the PR.

I'm a little bit confused about two things:
- [Here](https://github.com/ILIAS-eLearning/ILIAS/pull/5950/files#diff-be69b29b326e3c9417b8ccbe481943ea7a905c934408bb71874bf31bc6445bc3R291)  and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5950/files#diff-be69b29b326e3c9417b8ccbe481943ea7a905c934408bb71874bf31bc6445bc3R258)  you are setting `password_checked` to `false` just after loading the test pass. The class is called `ilTestSession`, but from where I stand, this is badly named, as the class does a bunch of disparate things. I think its main purpose is to load Test-PASS-data and not actually test-SESSION-data. As far as I see your session variable is reset on every page load. Am I missing something?

Thanks and best,
@kergomard

tbongers-cat:
Hi @kergomard 
sorry for getting back to this with such a delay. You are right about those to lines, they were an oversight by me and I actually meant to check them before finishing the PR. I have removed both lines, as the password check stored in session should not be affected by loading data.

kergomard:
Thank you very much @tbongers-cat ! Now this looks good to me. Could you please quickly see what causes  the conflicts and then I will merge right away. Thanks!
Best,
@kergomard

tbongers-cat:
@kergomard 
I see no conflicts here > _**"This branch has no conflicts with the base branch"**_
What am I missing?

kergomard:
Hi @tbongers-cat
Ok, this is fun:
image

Could you simply try a rebase and push? Maybe it helps.
Thanks and best,
@kergomard

tbongers-cat:
@kergomard This push should fix it (hopefully).

kergomard:
Thank you very much @tbongers-cat ! This did the trick. Will merge and pick to ILIAS7 and ILIAS8

PR #5951, opened 2023-05-31
UI: Allow dedicated names for inputs in a filter

tbongers-cat:
This is a followup for #5711 and allows dedicated names for inputs inside a filter component, as this implements its own namesource and did not implement `getNewDedicatedName()` yet.

### DIFF
--- a/src/UI/Implementation/Component/Input/Container/Filter/Filter.php
+++ b/src/UI/Implementation/Component/Input/Container/Filter/Filter.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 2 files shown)


Amstutz:
Hi @tbongers-cat

Looking good, thx for adding. 

one thing todo:

- [ ]  Unit Test: Pleas add one.

Thx.

tbongers-cat:
@Amstutz Done :)

Amstutz:
Hi @tbongers-cat 

Thx for adding the test. However, the tests is testing far more than what your implementing by checking the rendering for the complete component. Please reduce this to only testing the changed behaviour. This should be done without rendering. In cases such as this, wherer there is also some impact on the rendering, best only test this specific part, e.g. by checking if a string contains a certain element or pattern.

Thx.

tbongers-cat:
Hi @Amstutz , thanks for the feedback. I have changed the test accordingly, please have a look.

Amstutz:
thx a lot!

PR #5952, opened 2023-05-31
Fixes the Updatestep 102

kergomard:
I think we should simply notice users, if they are using MyISAM that we actually cannot update the length of session_id.
See: https://mantis.ilias.de/view.php?id=37403


kergomard:
I merge this, as otherwise this gives an ugly error on running setup/setup. Just let me know  @mjansenDatabay , if you would prefer another solution, I can always change it later on.

mjansenDatabay:
Thx @kergomard

PR #5954, opened 2023-06-01
UI: Layout/Alignment

nhaagen:
Dear all,

a [FR regarding the Presentation Table](https://docu.ilias.de/goto_docu_wiki_wpage_7764_1357.html) asked for Components being aligned horizontally next to each other.

This, however, seems like a reoccurring thing and should not be regarded solely 
in the context of single Commponents, but rather become a Layout principle.

Here is a proposal for [Alignment Layouts](https://github.com/ILIAS-eLearning/ILIAS/pull/5954/files#diff-53d0288e9cbb72ed3a885d65f434e9f8deb7c85b1d8b9fd9cdf6a0459a4b6b79); an [Alignment](https://github.com/ILIAS-eLearning/ILIAS/pull/5954/files#diff-833c4f7402c2097b4048fb2181049fa637a20224bd33ef895058d2550ecb121d) will take Components implementing
the according interface and wrap them into divs controlling their breakpoints.

Screenshot from 2023-06-01 09-21-55

Best regards,
Nils & Ferdinand


klees (changes requested):
Hi @nhaagen and @catenglaender,

I had a look into the interface to these new components but not the implementation yet. Also, I left @thibsy and @Amstutz out of the loop for this first iteration.

- [ ] structure 1:  Could we drop `prefer` and `force` and use some unified `horizonal` with an integer `strength` instead, where `strength` would define how long blocks stay next to each other when screen size gets smaller? Could we implement that behaviour?

- [ ] `Block`: Add a little context [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5954/files#diff-48b9d2492ac6da33bdf96bf99b6694d05b66fb96cf1ee2a928a784cb2da640b7R27) about what a `Block` is supposed to be.
- [ ] structure 2: Make `Alignment` be a `Block` as well. Remove the possibility to insert arrays of arrays of `Blocks` to an `Alignment` and instead just accept arrays of `Blocks`. If I understand correctly, an `Alignment` is supposed to align the included `Block`s according to their names. That would be an `array` of `Block`s in the alignment. But how would it treat the second `array` that we currently have? How will they be aligned? Semantics are not clear here IMO, or I simply do not understand the `effect` description and the difference between `prefer` and `force`.

- [ ] wording: Please exchange `blocksets` for `blocks`. No [sets](https://en.wikipedia.org/wiki/Set_(mathematics)) here =)

Kind regards!

klees:
I have reviewed this and will discuss the wording for the kitchen sink with @nhaagen tomorrow. We will then provide an updated version. Maybe @Amstutz and @thibsy could save some time on thursday or friday to review this then? IMO should be good to be discussed on JF on monday.

Amstutz:
@klees and @Amstutz discussed this today in discord session. We agree that something like alignments could be helpful for very generic content like output of the page editor.

However, there are several drawbacks in opening the interface of the presentation table to something as generic as alignments. As an example: We can no longer make a guess on the nature of content of those blocks, and therefore no longer decide whether they should be shown as a listing or in some other form.

Nevertheless, we believe this to be a good example to be discussed with fellow devs in the upcoming JF. It's a good showcase on the pros and cons of semantic vs layout-based components, and we would like to hear the opinion of others on the matter. 

I flagged this therefore for the upcoming JF discussion on Monday. Looking forward to it.

matthiaskunkel:
Jour Fixe, 26 JUN 2023: We highly appreciate this PR and accept it for trunk. Please change /Layout/factory.php line 73 to: 'Alignments MUST not be used when there is another component that fits the requirements (see purpose).' Due to the semantic weakness of the component, we request to decide using this UI element in new components at the Jour Fixe. In case someone finds a better title for the UI component with a semantic title (even if it is weak), please let us know here.

Amstutz:
Hi @nhaagen

Thx a lot for the PR. Special thx for the very detailed Component descriptions, examples and tests. Highly appreciated. Nice to see how little DOM and CSS is used for achieving this.


- [x] Inline Styles with Examples: Is it possible to devise an example without inline styles. We do really not want them, so it might be a bad example of show-casing inline styling. I am aware, that other examples are doing the same, but I think it is something we should not promote further. 

Please change:

- [x] JF Decision: Please add a MUST rule for usage, that requires a JF approval for using all Alignment Components
- [x]  CSS Classes as Variables: Do not use CSS Classes as Variables in Templates. This makes the templates impossible to read. A better Option might be the usage of different blocks which are parsed depending on the nature of the variable.

Kindly
@amstutz

nhaagen:
Hi @Amstutz ,
all done.
Thanks for looking through this ;)

Amstutz:
thx a lot!

PR #5955, opened 2023-06-01
Wiki Bugfix #0037406: Copying a Wiki with someone having enabled notifications crashes because of ctrl_struture

fhelfer:
https://mantis.ilias.de/view.php?id=37406


fhelfer:
cherry-pick for all above branches also

alex40724:
Thanks for the PR, see the comment in the bug report. This is definitely a bu, but needs a different fix, since the cause is somewhere earlier in the process when applying the notifications settings from the original wiki to the new one.

PR #5958, opened 2023-06-01
LSO: Disable multi-download for LSO

tbongers-cat:
I had a look at https://mantis.ilias.de/view.php?id=36631 and tried to figure out how to disable the header action "**Download Mulitple Objects**" just for the Learning Sequence. After some research, it looks to me as if `ilContainerGUI::multi_download_enabled` is ignored by `ilObjectListGUI::insertMultiDownloadCommand` (which inserts the Action button) and so there is no way to disable this option on a per container basis (as far as I can tell).

This PR adds a check to `ilObjectListGUI::insertMultiDownloadCommand` to see if multi-downloads are enabled for an `ilContainerGUI`. The downside with this is that the default value of `ilContainerGUI::multi_download_enabled` is `false` right now and thus the "**Download Mulitple Objects**" action disappears unless it is explicitly set to `true`. Therefore I'm putting this up for discussion, in case I'm missing some other option here to disable multi-downloads for a single container.

Looking forward to your feedback.

### DIFF
--- a/Modules/LearningSequence/classes/class.ilObjLearningSequenceGUI.php
+++ b/Modules/LearningSequence/classes/class.ilObjLearningSequenceGUI.php
@@ -226,6 +226,7 @@ class ilObjLearningSequenceGUI extends ilContainerGUI implements ilCtrlBaseClass
         $this->lng->loadLanguageModule($this->obj_type);
 
         $this->data_factory = new Data\Factory();
+        $this->multi_download_enabled = false;
 
         $this->ref_id = $this->request_wrapper->retrieve("ref_id", $this->refinery->kindlyTo()->int());
         parent::__construct([], $this->ref_id, true, false);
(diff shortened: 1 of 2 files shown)


klees:
Hi @alex40724,

could you have a look here, please? Is there any thing we could do to help?

Best wishes!

alex40724:
Hi @tbongers-cat,

thanks for the PR.

ilContainerGUI::multi_download_enabled does not control the availability of the command. It is set to true, after the user activated the multi download. In general the ilObjectListGUI part needs to be checked by @kergomard . I assume, that a new property or something similar is necessary to control the general availability of the command.

Kind regards

alex40724:
Currently there is no code change in my components, thus I deassign me. The current change needs to be reviewed by @kergomard

kergomard:
See Issue for further discussion.

kergomard:
Hi @tbongers-cat and @klees !

If I'm getting this right this PR can be closed after the JF decision.
I thus close it, but please reopen if I'm mistaken.

Thanks and best,
@kergomard

PR #5959, opened 2023-06-02
Membership: Show default role as default when adding users via search

schmitz-ilias:
This PR fixes [26025](https://mantis.ilias.de/view.php?id=26025).


smeyer-ilias (approved):
Thanks @schmitz-ilias

PR #5960, opened 2023-06-02
UI Components: Improved Mode Info Example

Amstutz:
The previous Mode Info example looked completely broken (see also https://mantis.ilias.de/view.php?id=33480). Since Mode Info is designed to be used once per page and in a very specific place, it was not fixable. To improve the situation and also somewhat better advertising, to make use of the flexibility of the Mainbar when in a special mode. I propose to revise the example completely. The current proposal is using a full-screen mode as done in other examples having a similar nature (e.g. Page Layout or Footer).

Further note, this would make a11y testing a lot easier/possible, since it is now rendered in the correct place on the page.

The new example looks like this:
image

---
image


klees:
Thanks @Amstutz, much appreciated.

PR #5963, opened 2023-06-05
UI Data Table - Mobile View

catenglaender:
# Issue

Data Tables are best to work with on large screens in landscape orientation. For small screens in portrait mode tables become quite cumbersome to navigate.

# Change

Below the mobile screen breakpoint, the Data Table now switches to a list view optimized for reading table items on a smartphone. This view is optimized for average users wanting to read two or three specific entries on the go and not meant for administrators to carry out large managing and sorting operations, as we expect these to always require a lot of screen space to be carried out efficiently.

### Data Table on large desktop screens (appears unchanged)

Desktop_DataTable-from-Repo

### Mobile small screen view (new with this PR)

Mobile_DataTable-from-repo

### With selection and dropdown action

image

# Further Considerations

The JavaScript for the keyboard navigation and the DOM of the table is currently not reacting to the mobile layout. This makes sense because we do not want screen readers to ever treat the Table any different from the original desktop view as we would loose some of the functions for navigating tables that screen readers have built in. Screen readers should always process the Data Table as a real table, not as a list. However, this means in the small screen view, the right arrow key causes the focus to go down to the next property of the list (which used to be the cell to the right), and the down arrow key jumps to the same property on another list item. Maybe the JavaScript could be deactivated below the mobile breakpoint?


yvseiler:
Hi! Thank you for the proposal. I will make a review as soon as possible, but I need a little time to think it through.

yvseiler:
Thank you very much for the proposed solution for small screens. 

**Basically**
- The display on small screens in the form of cards is a possible alternative to a table that has to be scrolled horizontally (see https://www.smashingmagazine.com/2019/02/complex-web-tables/). 
- Whenever possible, horizontal scrolling should be the last chosen display option for small screens. 
- There is no really clear solution for small screens, but there are approaches that go in a similar direction as suggested here (e.g. https://uxdesign.cc/designing-a-complex-table-for-mobile-consumption-nom-7472f7b11fe6).
- Whether it is the "Card" component in our framework or another suitable component (Listing Panel, in future Entity) is something I cannot answer yet. For the sake of clarity, I will continue to speak of "cards".

The primary goal should remain that users can scan information as easily as possible. In order to keep the list of displayed cards clear, we should make sure that
1. only some data is directly visible and not all (more can be displayed on click/tab)
2. functions are offered to limit the number of displayed cards (filters).

**1. Visible Data**
I recommend to add the possibility of a detailed view to the cards, this could be e.g. in the form of an expand/collapse behaviour. On the closed card, only a few data are visible at first. What might that look like?

- For the selection of the first data displayed, it can be defined that the first three columns are used (assuming that the columns on the left were considered more relevant when converting the data table). If the card is "expanded", all data of all columns of this entry will be shown in the card one after the other (as shown in mockups). The semantic groups introduced with the entity could possibly be reused here.
- Alternative: What do you think about the suggestion that the activated/visible columns could be displayed on the card. I.e. if columns are deselected or added, the number of displayed data on the card changes. In this case, in addition to filters (see 2.), it should also be possible to adjust the "columns" on small screens. This way, the user can dynamically adjust which information he wants to display on a card. Here, too, an "expand" behaviour could display all data for a card. (this would be my favorite way to go, but I do not know if the dynamic can be handled)

**2. Filter**
I recommend displaying the filter on small screens above the cards. By filtering, the number of cards displayed can be reduced, making the list more manageable.

What do you think about the suggestions?

Best regards
Yvonne

catenglaender:
@yvseiler and I talked about the Data Table mobile view today.

We concluded that from our UI/design point of view this is ready for merge.

We would like to contribute further thoughts how to get the most out of filters and view controls for the Data Table mobile view once Data Table view controls https://github.com/ILIAS-eLearning/ILIAS/pull/5811 and filters are available in trunk.

klees:
@Amstutz This LGTM, but much of style code. Would you be so kind and take a look here? Thx!

Amstutz:
@catenglaender 

I am not a fan of this view. I personally would much rather to see the 2d matrix view in tables to be kept as is, even on small view ports with horizontal scrolling. However, I do trust the judgement of you and @yvseiler seiler here. 

Technically, this looks fine, however, please fix the failing unit test and resolve the css.map conflict. Feel free to merge afterwards.

catenglaender:
@Amstutz Thanks for your input! :slightly_smiling_face: I strongly believe that (for now) this view is the right path for mobile. An unmodified table view could give users the wrong impression that tables are supposed to work perfectly on mobile and with touch, which they unfortunately don't. So I would much rather give the user the strong hint that this is okay for reading and single edits, but that mobile is just not meant for extended administrative work.

However, I do think we could always work on a mobile table later. Supporting long press context menus and selections would be awesome. But I think before we can do that we might want to work on a a general touch concept for ILIAS.

How about setting a flag in SCSS and wrap the code for the mobile view in an if statement so skins can decide against this table mobile view? Switching it off would require compiling, but at least it could be switched off. What do you think?

yvseiler:
I completely support @catenglaender arguments. Instead of a switch in the code, one possibility could be to give the user the choice. For example, the entities could be displayed one below the other by default on a small viewport; with a ViewControl Mode, one could switch to the display as a (horizontally scrollable) table.
For the moment, until the view can be used with the functions of Filter, ViewControls, Selection, etc. on small screens, this might be a feasible way.

Amstutz:
I do not agree. 

>  So I would much rather give the user the strong hint that this is okay for reading and single edits, but that mobile is just not meant for extended administrative work.

If so, then no issue here, nobody would open such a screen with a mobile device then, right. However, if somebody for some reason would like to do administrativ work and opens such a screen with such a huge table, then this person would be confronted with this new view.

Further I do believe that we will run into a rather large set of new tasks and issues with this change. 

Nevertheless. My opionion is not that important here. If @yvseiler and @catenglaender agree, that this is the way to go, I will follow the suggestions.

> Technically, this looks fine, however, please fix the failing unit test and resolve the css.map conflict. Feel free to merge afterwards.

- [ ] This is left to do.

templates/default/070-components/UI-framework/Table/_ui-component_table.scss
```diff
@@ -334,5 +355,96 @@ td.c-table-data__cell--highlighted {
...
+        //
+
+        td.c-table-data__cell--email,
+        td.c-table-data__cell--number {
+            text-align: inherit;
+        }
+        
+        // Highlighted Columns
+        td.c-table-data__cell--highlighted {
+            font-size: $il-font-size-xxlarge;
```
klees, line 424: @catenglaender There was a dangling reference to `btstrp3.$font-size-h2` which I replaced like this here. Hope that's fine.

klees:
Hi @Amstutz,

conflicts and tests are solved. I guess this is good to go now.

Kind regards!

Amstutz:
thx!

PR #5969, opened 2023-06-06
Search: Context Sensitive Search - Suggestions from ILIAS (11578)

schmitz-ilias:
This PR fixes [11578](https://mantis.ilias.de/view.php?id=11578) by disabling autocomplete in the main menu search when 'At Current Position' is selected as 'Search Area'.


smeyer-ilias (approved):
Thanks @schmitz-ilias

PR #5970, opened 2023-06-07
[FIX] CI: fixed arrow-function spacing for PHP-CS-Fixer.

thibsy:
Hi all,

According to PER, arrow-functions or the `fn` keyword must not be followed by a space character. See https://www.php-fig.org/per/coding-style/#71-short-closures for more info.

With the current configuration for the PHP-CS-Fixer however, this space has been added up till now. I have also checked with the PHPStorm code-style, which already did this correctly.

Kind regards,
@thibsy


mjansenDatabay (approved):
Thx @thibsy

PR #5973, opened 2023-06-07
StandardPage Modification: strict StandardPage return types

iszmais:
**ALTERNATIVE FOR https://github.com/ILIAS-eLearning/ILIAS/pull/5974**

This is a proposal to fix the subsequent problem of https://mantis.ilias.de/view.php?id=37271

Even if this problem should not occur an is always caused by other issues, the basic worklflow of the modifciation process should be unified withing the standard page and the accessing providers.

This PR does that buy stricten the Standard page on its (mostly already implemented) types without the possibility of an nullable return.
The alternative approach, wich targets the accessing provider, can be found [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5974).


chfsx:
As mentioned in #5976, I close this PR: The UI\Page allows certain parts to be `null`, the GlobalScreen must take this into account and allow this as well.

PR #5974, opened 2023-06-07
StandardPage Modification: lose modification parameter types

iszmais:
**ALTERNATIVE FOR https://github.com/ILIAS-eLearning/ILIAS/pull/5973**

This is a proposal to fix the subsequent problem of https://mantis.ilias.de/view.php?id=37271

Even if this problem should not occur an is always caused by other issues, the basic worklflow of the modifciation process should be unified withing the standard page and the accessing providers.

This PR does that buy losen the modification closure parameters and handling their dual type withing the modification itsself.
The alternative approach, wich targets the return types of the standard page, can be found [here](https://github.com/ILIAS-eLearning/ILIAS/pull/5973).

### DIFF
--- a/Modules/LearningModule/Export/class.ilLMHtmlExportViewLayoutProvider.php
+++ b/Modules/LearningModule/Export/class.ilLMHtmlExportViewLayoutProvider.php
@@ -54,7 +54,7 @@ class ilLMHtmlExportViewLayoutProvider extends AbstractModificationProvider impl
                 ->layout()
                 ->factory()
                 ->metabar()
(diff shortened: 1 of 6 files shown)


alex40724:
Thanks for the PR. I removed me from the list of reviewers. @chfsx should decide on this, since this seems is mainly a global screen issue.

kergomard (approved):
Hi @iszmais 
Thank you very much for the PR. I looked at the changes in Test in the name of @dsstrassner and they are fine. I didn't look at the rest.

Best,
@kergomard

chfsx:
This PR has been superseded with #5976 which should contain the same changes as here as well.

PR #5975, opened 2023-06-07
Fix headings for panel (Mantis 37445)

tfamula:
Hi @Amstutz and @klees,
I found out that the HTML heading of the Secondary Panel is not correct, if it appears within a "normal" panel. Currently, it is h2, but it should be h4 IMO.
I am not sure if my PR is the right way to solve this problem. I tried to keep the changes minimal for this little distinction. But I fear that I maybe have overlooked something in other places, which will cause new problems. However, if I raise awareness for this, it's already a good start.

If the PR should be accepted, I can also write a new unit test for the different behaviour.

https://mantis.ilias.de/view.php?id=37445

### DIFF
--- a/src/UI/Component/Card/Card.php
+++ b/src/UI/Component/Card/Card.php
@@ -98,4 +98,14 @@ interface Card extends Component, JavaScriptBindable, Clickable
      * Returns whether the Card is highlighted
      */
     public function isHighlighted(): bool;
+
+    /**
+     * Sets the component in the context as further information within another component
+     */
+    public function withFurtherInformationContext(bool $context): Card;
+
+    /**
+     * Is the component standalone or serves as further information within another component?
+     */
+    public function getFurtherInformationContext(): bool;
 }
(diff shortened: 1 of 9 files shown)


Amstutz:
Hi @tfamula 

Thx for the PR. This issue was already discussed several times during the introduction of the current headline scheme. We then mentioned that we would need some kind of rendering mechanism which is able to chose the currect heading for each context. We are somewhat unsure wheter this is the proposed way to go. We will discuss this among the UI Coordinators @thibsy , @Amstutz and @klees at get back to you as soon as possible.

Thx for taking the initiative.
@Amstutz

klees:
Hi @tfamula,

as promised, we discussed the issue at hand. Please note that mostly this feedback is not only for you or only about this change. They rather reflect our general discontent with the current situation regarding headlines, as this again shows the problems we have predicted [here](https://github.com/ILIAS-eLearning/ILIAS/pull/2812#pullrequestreview-470424018).

In a nutshell: The proposed change is an ad-hoc fix that does not attack the root cause. It damages composability as the component now needs to know where it is located. For a solution that attacks the root cause, the UI framework should keep track about the required headline level. We do not want to add another piece of duct tape as there  is a known better solution.

If anyone wants to work on the proposed solution or has budget to do so, feel free to contact us for guidance. We believe to have several ideas on how to tackle this and would be happy to discuss them with somebody willing to take on the task.

Kind regards!
@thibsy, @Amstutz and @klees

PR #5972, opened 2023-06-07
[37418] Fix missing tools if mainMenu has no items

PurHur:
Hello,

if the main menu has no items there are no tools visible.

The collector for the tools is still empty because collect wasnt called yet. This PR fixes this and the tools are visible if there are no menu items. 

Greetings
Purhur

PR #5976, opened 2023-06-08
Fix/7/layout modification

chfsx:
Thank you @PurHur and @iszmais for your PRs:
- #5972
- #5973
- #5974

This PR here replaces them and implements the following changes:

- The missing collect of tools as in #5972 (but the collectOnce at the bottom removed instead).
- LayoutModifications are provided as Closures, there already existed "definitions" how these have to look like, but these were not checked correctly. I have therefore adapted checkClosure to check the definitions. Violations are now thrown as InvalidModification exceptions, these are also output if DEVMODE is activated, otherwise the LayoutModifications are regarded as invalif.
- I have adapted the existing modifications (as in PR #5974).

What we cannot adapt are the changes in #5973: The UI\Page allows certain parts to be `null`, the GlobalScreen must take this into account and allow this as well.

I will therefore close the other three PRs and adopt the changes from here to and including trunk.

This solves the following issues:
- https://mantis.ilias.de/view.php?id=37418
- https://mantis.ilias.de/view.php?id=37271

### DIFF
--- a/Modules/LearningModule/Export/class.ilLMHtmlExportViewLayoutProvider.php
+++ b/Modules/LearningModule/Export/class.ilLMHtmlExportViewLayoutProvider.php
@@ -1,5 +1,21 @@
 <?php
 
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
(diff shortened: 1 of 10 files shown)


chfsx:
@PurHur and @iszmais perhaps you could test this PR briefly to see if it also solves all the problems you have found.

iszmais:
LGTM :+1: 

However, imo this does not resolve https://mantis.ilias.de/view.php?id=37271 but just resolves an error which occurs within the issue.
It fixes hte issue that occurs when the menu is empty, but it still leaves open the question why the menu is empty at that point.

Since i cant reproduce the origin problem im not able to test this specific problem.

Greetings,
@iszmais

PR #5979, opened 2023-06-09
Chat / OSC: Abandon emoticons

lscharmer:
Apply feature: https://docu.ilias.de/goto_docu_wiki_wpage_7791_1357.html


Services/Database/classes/class.ilDBGenerator.php
```diff
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
```
mjansenDatabay: Please revert all changes of this file.

lscharmer: The changes are reverted.

mjansenDatabay (approved):

PR #5983, opened 2023-06-12
LegacyUI/AdvancedSelectionList: Fix array access

mjansenDatabay:
If approved, this should be picked to `trunk` as well.

https://mantis.ilias.de/view.php?id=37459


alex40724:
Thanks!

PR #5985, opened 2023-06-12
UI, Modal/DataTable: relay signal options to Modal

nhaagen:
I was working on the removal of legacy ConfirmationGUI and did not find a good way to relay items to the Interruptive Modal via JS. This seems to be a common situation, though, when there are several items being picked in the GUI and the modal should react to this selection.
With this change, the signal options are sent along as POST to the ajaxRenderUrl.

### DIFF
--- a/src/UI/examples/Modal/RoundTrip/show_form_in_modal.php
+++ b/src/UI/examples/Modal/RoundTrip/show_form_in_modal.php
@@ -11,6 +11,7 @@ function show_form_in_modal()
     $renderer = $DIC->ui()->renderer();
     $request = $DIC->http()->request();
     $factory = $DIC->ui()->factory();
+    $request_wrapper = $DIC->http()->wrapper()->query();
 
     // declare roundtrip with inputs and form action.
     $modal = $factory->modal()->roundtrip(
@@ -20,7 +21,7 @@ function show_form_in_modal()
             $factory->input()->field()->text('some text'),
             $factory->input()->field()->numeric('some numbere'),
         ],
-        '#'
+        $request->getUri()->__toString() . '&rtwithform=1'
     );
 
     // declare something that triggers the modal.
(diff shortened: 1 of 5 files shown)


chfsx:
This seems to solve at least a part of https://mantis.ilias.de/view.php?id=37494

thibsy:
Note that I have opened a PR for documentation reasons a while back, which addresses exactly this issue: #5773

chfsx:
I think we need both of them… can we push this a litte? ping @klees :-)

klees:
Hi everyone,

I have looked into this and I'm quite convinced that this is not it. This looks way to unprincipled for my taste:

* As @chfsx points out, json-encoded data in the URL is strange and leads to problems.
* The ajax-call and the way that the response is included feels ad-hoc.
* This is not bound back to any server side facility, hence communication like in the `with_action`-example feels ad-hoc as well.

No accusation implied, though, this is just unchartered territory. I need to think a little about how we could handle this in a more principled fashion.

Even more stuff to discuss with @nhaagen. I hope you have a little patience @chfsx.

Thanks!

nhaagen:
Dear all,
as in above, my main concern was to relay selected columns to the modal.
There already is a way to do it without a modal, and this will still be improved with https://github.com/ILIAS-eLearning/ILIAS/pull/5938 (then, also taking care of the json_encoding).

@klees and me were talking about endpoints, mainly, and there are certain advantages in focussing on that.

In general, the table does not know about the nature of actions, its only concern should be to open some modal layer. 
The contents of that layer do not have to be limited to Affected Items, but may contain elaborated messages, further 
inputs (e.g. "give the filename of all selected files to zip") and more.

The idea is to extend the Table\Data\Action with another parameter and ditch the signal for good. 

So, instead of 
```
compare' => $f->table()->action()->multi('compare', 'ids', buildDemoURL($request, 'table_action=compare')),
'delete' => $f->table()->action()->standard('delete', 'ids', $signal)
```
we would stick to the url and ask for asynch rendering by an additional flag

```
'compare' => $f->table()->action()->multi('compare', 'ids', buildDemoURL($request, 'table_action=compare')),
'delete' => $f->table()->action()->standard('delete', 'ids', buildDemoURL($request, 'table_action=compare'), true)
```

having an Action factory like 
```
standard(string $command, string $parameter, URL $endpoint, bool $interrupt): Action
```

which will send an asynch request to $endpoint with $parameter[]=selected1&$parameter[]=selected2&...
and display the response in an interrupting layer over the table.

The response might then e.g. incorporate a Confirmation Message Box, but is not limited to it.
This would also emlinate the need to factor the modal first, and further unify the construction of actions.

Best regards

PR #5986, opened 2023-06-12
UI/Metabar: 33346, do not close metabar slates when there is no 'next focus'

nhaagen:
https://mantis.ilias.de/view.php?id=33346

please check against https://mantis.ilias.de/view.php?id=31367 (caused the issue), but I think this does not really break it.


klees (approved):
LGTM

Amstutz:
thx a lot.

PR #5990, opened 2023-06-13
UI Bugfix #0037466: Standard Listing Panel ui bugfix

fhelfer:
https://mantis.ilias.de/view.php?id=37466


Amstutz:
thx for fixing.

PR #5994, opened 2023-06-13
T&A Bugfix #0037334: Invalid Question UI

fhelfer:
https://mantis.ilias.de/view.php?id=37334


dsstrassner:
Thanks @fhelfer! Could you please also change the ilias_en.lang according to the JF Decision? ---> Question is incomplete!

kergomard:

Will merge and pick to ILIAS8.

Best,
@kergomard

PR #5995, opened 2023-06-13
[FIX] 0037125: Sorting of content in container objects does not work

chfsx:
This PR disables the dropzones in the Sort and Manage view for all objects, since the change is in Services/Object, I assign you to @kergomard.

Let me know if you want me to solve anything differently, unfortunately I had not found any other solution in the ObjectListGUI at the moment to find out which "mode" one is in.

This fixes https://mantis.ilias.de/view.php?id=37125

### DIFF
--- a/Services/Object/classes/class.ilObjectListGUI.php
+++ b/Services/Object/classes/class.ilObjectListGUI.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much for the PR @chfsx .
Nah, this doesn't make things perceivable worse.
Best,
@kergomard

PR #5996, opened 2023-06-13
T&A 0032158 mark scheme deletion confusion

kunzeus:
Ticket:
https://mantis.ilias.de/view.php?id=32158

Symptom:
you could delete a mark scheme even if it's required 

Problem:
The delete function directly delete the mark scheme in the database, and fixed the missing minimum requirement for the mark scheme with the next reload of the page.

Solution:
perform a check if the delete is feasible

Checked solution on:
ILIAS 7, ILIAS 8 works completely different and doesn't have the issue

Also, probably unavailable until the end of June.


kergomard:

I will merge this and cherry-pick to all newer branches.

Best,
@kergomard

PR #5997, opened 2023-06-14
Membership: allow checking whether user may mail to members of course, group or session (22764)

schmitz-ilias:
This PR makes some changes that allow [22764](https://mantis.ilias.de/view.php?id=22764) to be fixed, see the discussion there.

@mjansenDatabay, this should hopefully work as a minimally invasive solution for R7 and 8, but we could take this as an opportunity to come up with a more clearly defined interface between the Mail Service and Courses/Groups for the trunk. Stefan and I were thinking about something in the vein of a `MailBridge` in the Membership Service. Maybe we could have a quick discussion about this at some point.

### DIFF
--- a/Services/Membership/classes/class.ilParticipants.php
+++ b/Services/Membership/classes/class.ilParticipants.php
@@ -201,8 +201,12 @@ abstract class ilParticipants
         return false;
     }
 
+    /**
+     * This method was introduced as a band-aid fix for #22764.
+     * Please do not use this anywhere else.
+     */
     public static function canSendMailToMembers(
-        int $obj_id,
+        int $ref_id,
         ?int $usr_id = null,
         ?int $mail_obj_ref_id = null
     ) : bool {
@@ -211,8 +215,6 @@ abstract class ilParticipants
         $access = $DIC->access();
         $rbacsystem = $DIC->rbac()->system();
 
-        $refs = ilObject::_getAllReferences($obj_id);
-        $ref_id = end($refs);
         if (is_null($usr_id)) {
             $usr_id = $DIC->user()->getId();
         }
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
Hi @schmitz-ilias . The actual issue addresses the `Contacts` component IMO, which does currently have too much knowledge about courses and groups and business logic of these object types.

I currently offered a technical concept to decouple the contact services from ...

dimension 1) ... concrete membership providers like courses and groups, so other participants enabled object types should be (theoretically) able to provide lists of objects and their parcipants/members/role assignees as well.
dimension 2) ... consumers like the mail system, and (I was not aware of this until the PHP 8 project) the portfolio/workspace.

smeyer-ilias:
Thanks, @mjansenDatabay for the info. 
Do I understand you correctly: the concept is offered but not published yet?

If yes, we will merge this to R7  until trunk.

mjansenDatabay:
@smeyer-ilias Yes, the state is: Interest in funding, no contract is settled, yet.

Feel free to merge this to all branches.

smeyer-ilias:
Thanks @schmitz-ilias

Services/Membership/classes/class.ilParticipants.php
```diff
@@ -201,6 +201,51 @@ public static function hasParticipantListAccess($a_obj_id, $a_usr_id = null)
...
+        ) {
+            return true;
+        }
+
+        $part = self::getInstance($ref_id);
+        if (!$part->isAssigned($usr_id)) {
+            return false;
+        }
+
+        $object = ilObjectFactory::getInstanceByObjId($obj_id);
```
mjansenDatabay, line 237: @schmitz-ilias I tried to consume this new method, and found an issue:

* Where does the `$obj_id` come from?
* And: The contact service already has the instance. Creating a new one would cause obsolete SQL queries for N courses/groups. Maybe the instance can also be passed instead of the $ref_id.

A dynamic `$ref_id_or_instance` parameter would be good here.

PR #5998, opened 2023-06-14
Added Language How-Tos

acgruber:
As announced at the DevConf and in the [Jour Fixe of 12.06.2023](https://docu.ilias.de/goto_docu_wiki_wpage_7950_1357.html), all DevGuide pages that are how-tos and no training materials are changed to .md-files, stored in GitHub and embedded in the current DevGuide LM using the related page component plugin. This way we can ensure improved accessibility and maintenance.

### DIFF
--- a/Services/Language/use-language-object.md
+++ b/Services/Language/use-language-object.md
@@ -1,0 +1,31 @@
+# $lng language object
+
+ILIAS offers multi-language support. All text strings are stored in **language files** in the subdirectory `lang`. The lines of these files have the format:
+ 
+`module#:#variable_name#:#text content`
+ 
+It is best practice that as module a short ID for your component is used, e.g. frm for forums. This should also be used as a prefix for the variable names in this language module, e.g.:
+ 
+`frm#:#frm_new_posting#:#New Posting`
+ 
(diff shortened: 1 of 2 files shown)


Services/Language/use-language-object.md
```diff
@@ -0,0 +1,31 @@
...
+```php
+$lng->loadLanguageModule("frm");
+$tpl->setVariable("TEXT", $lng->txt("frm_new_posting"));
+```
+
+## Maintaining Languages
+
+The different languages supported by ILIAS are maintained by volunteers. We are offering language maintenance installations for every version. They are clients of the regular testing installation.
+
+- [http://lang54.ilias.de](http://lang54.ilias.de) is the language maintenance installation for 5.4
```
dsstrassner, line 28: Are you sure you even want to submit those old URLs?

acgruber: Thank you for your note! As stated in the JF: In this first step, the focus is on transferring the status quo to GitHub. While reviewing the PR, the responsible maintainers are kindly asked to check whether an update is necessary. Any improvements are highly appreciated :)

matthiaskunkel:
Thanks to @acgruber for transforming the DevGuide pages into md files. I merge the current status and improve the content in a second step. @dsstrassner is right. It doesn't make sense to mention outdated installations.

PR #5999, opened 2023-06-14
UI inline launcher button design

catenglaender:
# Issue

UI Inline Launcher Button is not notable enough.

# Change

As it's relatively certain to assume that launching is the most relevant action in the view, the launch button has been styled as a large button primary.

image


Amstutz:
Hi @catenglaender 

Thx for the proposal. @yvseiler and me looked at it, and we believe the result to be fine. However, there are some technical aspects that need some discussing.

- [ ] Darken for Buttons: You introduce a new color in the middle of the launcher scss. `darken(s.$il-btn-primary-bg,10%);` It is used twice, IMO that is a hint, that this should move to a variable.
- [ ] Button Variant: We still have the button-variant from bootstrap in use. IMO we should use that consistently to easy a future migration. Why not use that, and/or if necessary refactor, extend it. Maybe even move from the Legacy Bootstrap folder. This would perhaps be a good opportunity. It would certainly resolve the darken issue above (this is done like this in the mixin).
- [ ] Border 2px: Why 2px? Are we doing that in other places? If this something we do in other places?

thx a lot,
@Amstutz

catenglaender:
Very good points. Most of them are left overs because I tried an outlined button design, which is not supported by the bootstrap button-variant mixin. I wanted to modify the button-variant eventually to also support a larger variety of designs, but not in this PR. Now, because it ended up being a filled button again, I should indeed make it closer to the other button code we are used to. Thank you for the review!

catenglaender:
This PR is now using the button-variant and button-size mixins, which should adress all points @Amstutz mentioned.
# Screenshot
image

oliversamoila:
Hi @catenglaender 
I am not sure if this concerns the scope of the PR, but I would like to ask.

1. Why is there a failure message box? This is an information message box in terms of its content. Nothing is failed.
2. Could we swap the "Do this first" and "And this is mandatory, too" for targets in the KS doc? Usually there are navigable targets in the listing. Actions in turn are available as buttons at the bottom of the message box.

catenglaender:
This is probably a question that people who were involved more with the concept could answer better than me. This is the way how I understand it:
1. I would argue that it's a failure to launch. It's the error message you would get if you could try to launch it - we are just saving the user one extra click (of which we have more than enough anyway) and anticipate this failure message.
2. As far as I understand preconditions are courses or other magazine objects most of the time, which is why I think this is a list rather than an action button.

yvseiler:
The failure message shows exactly what the failure message asks for (why interaction has failed, how can the problem be fixed). But I can understand why @oliversamoila expects an information message here: 

- The KS description of the Information Message says _"The system informs the user about obstacles to the completion of a workflow or about side effects of its actions on other users"_ ([source information message](https://test8.ilias.de/goto_test8_stys_21_MessageBoxMessageBoxInfo_default_delos.html)). 

If we understand the failure message as "error from the system", I also tend to use an information message here (_"The system failed to complete some actions and displays information about the failure."_ ) [source failure message](https://test8.ilias.de/goto_test8_stys_21_MessageBoxMessageBoxFailure_default_delos.html). Because: No error occurred when I clicked on the title or called the action "start test", the action "show me the page with the launcher" worked. But there is something in the completion of my workflow that I am informed about. 
It could also be that the warning message makes users more hesistant to click on a link in it. 

From this point of view, I would tend to use an information message here.

Amstutz:
@catenglaender  Thx a lot for the improvement of the button design

@yvseiler and @oliversamoila : thx a lot for the feedback concerning the example. Always appreciated. I merge the PR nevertheless, since the point is outside of what has been changed/accomplished in the PR. However, If you feel strongly that the example can/should be improved, feel free to create an issue in mantis. We will take care of it.

PR #6000, opened 2023-06-14
Course: activate dormant participantHasPassedCourse event

schmitz-ilias:
This PR activates the already existing event 'participantHasPassedCourse'. It was implemented 2016 and never activated, but it seems to work fine, and should be sufficient for fixing [26314](https://mantis.ilias.de/view.php?id=26314).
Note that the event triggers when a course participant's 'passed' status is granted, but not when 'passed' is taken away.


matthiaskunkel:
Jour Fixe, 16 OCT 2023: This PR has already been merged but never presented in the JF. @schmitz-ilias notified us about the PR and its impact.

PR #6001, opened 2023-06-14
T&A 0036734 Partial solution (ILIAS7)

jcopado:
Ensures special characters are properly decoded to be shown in Form's Question field when addBasicQuestionFormProperties is used to add the 6 default form entries for questions in ILIAS, It will continue failing for other fields.


kergomard:
Hi @jcopado 
Sorry this got on my backburner for so long!
I implemented another solution: https://github.com/ILIAS-eLearning/ILIAS/commit/988e657779e9a6097402bedd1ae5d9e0b1334240

...and hope to tackle the issues at their roots. I might have missed something though.

Thanks and best,
@kergomard

PR #6003, opened 2023-06-15
T&A Bugfix #0037462: non-numerical mark scheme percentage

fhelfer:
https://mantis.ilias.de/view.php?id=37462

Used lng variable "mark_schema_invalid" => "One of your grade categories needs to start at the ‘Minimum Score Required (in %)’ level of 0% Your grading system hasn’t been saved." will be displayed when percentage is non numerical 

Change of behaviour to 7 where percentage is set to 0 instead

### DIFF
--- a/Modules/Test/classes/class.ilMarkSchemaGUI.php
+++ b/Modules/Test/classes/class.ilMarkSchemaGUI.php
@@ -97,12 +97,18 @@ class ilMarkSchemaGUI
                 //replace , with . for float values
                 $value = str_replace(',', '.', $value);
 
-                $this->object->getMarkSchema()->addMarkStep(
-                    ilUtil::stripSlashes($postdata["mark_short_$matches[1]"]),
-                    ilUtil::stripSlashes($postdata["mark_official_$matches[1]"]),
-                    str_replace(',', '.', ilUtil::stripSlashes($postdata["mark_percentage_$matches[1]"])),
-                    ilUtil::stripSlashes($passed)
-                );
+                $percentage = str_replace(',', '.', ilUtil::stripSlashes($postdata["mark_percentage_$matches[1]"]));
+                if (!is_numeric($percentage)) {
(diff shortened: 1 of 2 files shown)


fhelfer:
https://github.com/ILIAS-eLearning/ILIAS/pull/6003/commits/ac91848d6d90809673bb4f2a3bf39daa7e283138 fills data from mark sheme table with post data if the form fails so changes to the form doesnt get deleted if user makes a mistake in an input
If that leads to problems or is not intended let me know and i will remove the commit :)

kergomard:

I've one question/suggestion:
- I think the retrieval from the wrapper in in `ilMarkSchemaTableGUI` has an unwanted side-effect: The transformations applied to the value as part of saving, will not be visible in the form on success. On error everything will work as expected. I think we should move the data-retrieval out of the `ilMarkSchemaTableGUI` again. I would suggest to hand in the data as an array on error and initialize it in `ilMarkSchemaTableGUI::initData()` in all other cases. Let me know, if I'm missing something.

- [ ] While retrieving the percentage [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6003/files#diff-5b5cbfd7c3c0f2d0f7d03e2b9e6ff576e35dc9f04e9bf15bcd67e8e266b9ed8bR89), you have a variable `$percentage` that might contain a `string` or a `float`. Additionally this would, as far as I can see, give an error, if you would enter "german-style" numbers (e.g. 1,1). I would suggest to write a custom `Transformation` for this. If you change the first part, this might not be necessary or the corresponding checks should move.

Thanks again and best,
@kergomard

fhelfer:
Hello @kergomard,

i took a different approach, because i did not want to drag the post data through multiple Methods. So i removed the Exception and made the method ilMarkSchemaGUI::saveMarkSchemaFormData() return a bool on succes/failure.
Marksteps will be added to the Markschema nevertheless, but not saved to database.

kergomard:
Thank you very much @fhelfer . This actually looks good to me too. Will merge and cherry-pick to 8 and trunk.

PR #6005, opened 2023-06-15
Lang show mail settings

fsesterh:
Clarified the admin setting and how things work after the fix for Mantis 37426.


mjansenDatabay (changes requested):
Thx @fsesterh . Please remove the `security.txt` from the changed files and squash your commits (currently: 9 commits).

PR #6008, opened 2023-06-15
Nullable modifications in learning sequence

PurHur:
This PR fixes https://mantis.ilias.de/view.php?id=37473

Note: There could be more modifications in the sourcecode with missing nullable first parametes / nullable return type

### DIFF
--- a/Modules/LearningSequence/classes/GlobalScreen/class.ilLSViewLayoutProvider.php
+++ b/Modules/LearningSequence/classes/GlobalScreen/class.ilLSViewLayoutProvider.php
@@ -118,7 +118,7 @@ class ilLSViewLayoutProvider extends AbstractModificationProvider implements Mod
 
         return $this->globalScreen()->layout()->factory()->breadcrumbs()
             ->withModification(
-                function (Breadcrumbs $current) : ?Breadcrumbs {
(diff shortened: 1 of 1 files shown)


Modules/LearningSequence/classes/GlobalScreen/class.ilLSViewLayoutProvider.php
```diff
@@ -136,7 +136,7 @@ public function getContentModification(CalledContexts $screen_context_stack) : ?
         // away the header here.
         return $this->globalScreen()->layout()->factory()->content()
             ->withModification(
-                function (Legacy $content) use ($html) : Legacy {
+                function (?Legacy $content) use ($html) : Legacy {
```
klees, line 139: @chfsx This should return `?Legacy` as well, right?

klees: No, just `Legacy` seems to be fine: https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/GlobalScreen/Scope/Layout/Factory/ContentModification.php#L65

klees:
Thanks a bunch, also for debugging.

PR #6009, opened 2023-06-15
Added Development How-Tos

acgruber:
As announced at the DevConf and in the [Jour Fixe of 12.06.2023](https://docu.ilias.de/goto_docu_wiki_wpage_7950_1357.html), all DevGuide pages that are how-tos and no training materials are changed to .md-files, stored in GitHub and embedded in the current DevGuide LM using the related page component plugin. This way we can ensure improved accessibility and maintenance. Any update or improvement of the content by the responsible maintainer is highly appreciated.


acgruber:
Hi @klees,
Thanks for assigning. I was hoping for a delevoper to check if the content needs to be updated. Would you be so kind? If no changes are needed for now, just let me know and I'll be happy to merge.
Thanks and cheers!

PR #6010, opened 2023-06-15
clarify show_mail_settings_info for 37426

fsesterh:
Clarified show_mail_settings_info due to changes in Mantis 37426. Previously the info text did not contain the important information, that the mail settings can only be changed if admins also grant permissions for users to edit the corresponding field in the user profile. The mail setting option is only used to show/hide the current settings in the personal mail settings.


mjansenDatabay:
I will accept this PR. Nevertheless, there will be a revised version of this contributed by Chris Potter, soon.

Thx @fsesterh !

PR #6013, opened 2023-06-16
[FIX] remove funky whitespaces, zero joiner and soft hyphens

chfsx:
This PR fixes several problems that occur in ILIAS when uploading files with SoftHyphens or other hidden characters. 

- https://mantis.ilias.de/view.php?id=32339
- https://mantis.ilias.de/view.php?id=35372
- https://mantis.ilias.de/view.php?id=34241
- https://mantis.ilias.de/view.php?id=36879

the problems were partially solved locally, with this PR they should now be solved in general. I will also prepare the same fix for ILIAS 7. @alex40724 this may be interesting for you, as tickets for this topic are also assigned to you in mantis.

### DIFF
--- a/src/FileUpload/Processor/FilenameSanitizerPreProcessor.php
+++ b/src/FileUpload/Processor/FilenameSanitizerPreProcessor.php
@@ -1,5 +1,21 @@
 <?php
 
+/**
+ * This file is part of ILIAS, a powerful learning management system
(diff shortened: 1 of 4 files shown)


matthiaskunkel:
Jour Fixe, 26 JUN 2023: We highly appreciate this suggestion and accept the PR for trunk, ILIAS 8 and ILIAS 7. Please merge.

PR #6017, opened 2023-06-19
Plugins/Repository: Fix PHP type when consuming permanent link target

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=37481

If approved, this must be cherry-picked to `trunk` as well.


alex40724:
Thanks for the PR!

PR #6021, opened 2023-06-19
T&A Bugfix #0037463:  Ordering Question (vertical): Answer Picutre Error

fhelfer:
https://mantis.ilias.de/view.php?id=37463


fhelfer:
Old image from previous save gets send in $_FILES
dont know where to fix this but the added is_array check makes sure to skip the faulty entry in $_FILES

fhelfer:
Only an issue in ILIAS 7 no cherry-picking required

kergomard:
Thank you very much @fhelfer!
Ok, as this is ILIAS7 only I merge this fix, even if it is the definition of "suppression of symptoms".
Best,
@kergomard

PR #6025, opened 2023-06-20
Database Bugfix #0037504: Return value of ilDBAnalyzer::getPrimaryKeyInformation() must be of the type array, bool returned

fhelfer:
https://mantis.ilias.de/view.php?id=37504

### DIFF
--- a/Services/Database/classes/class.ilDBAnalyzer.php
+++ b/Services/Database/classes/class.ilDBAnalyzer.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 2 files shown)


mjansenDatabay:
@fhelfer I guess this has to be picked to `release_8` as well?

fhelfer:
> @fhelfer I guess this has to be picked to `release_8` as well?

Indeed

Services/Database/classes/class.ilDBAnalyzer.php
```diff
@@ -295,7 +295,7 @@ public function hasSequence(string $a_table)
         if (is_array($seq) && in_array($a_table, $seq)) {
             // sequence field is (only) primary key field of table
             $pk = $this->getPrimaryKeyInformation($a_table);
-            if (is_array($pk["fields"]) && count($pk["fields"]) === 1) {
+            if (isset($pk['fields']) & is_array($pk["fields"]) && count($pk["fields"]) === 1) {
```
chfsx: here one `&` is missing

chfsx:
I merge this for trunk only.  please note that the ilDBAnalyzer is an internal class and can NOT be understood as a public API, so it should not be used in plugins or anything. the class can change or be dropped at any time.

PR #6026, opened 2023-06-20
UI/standalone view controls/sortation/26634

nhaagen:
Screenshot from 2023-06-19 14-39-18

https://mantis.ilias.de/view.php?id=26634

This decouples ViewControl Sortation from Dropdown, so an item my be shown as "selected".
Please refer to the ticket, especially @yvseiler 's last comment.


oliversamoila:
Good morning to you, @nhaagen and @klees. 
Can you briefly help me to understand the status correctly? There is no reviewer assigned at the moment, would that have to be done in the next step or is something else required first?
Thank you very much for this work!

klees:
Hi @oliversamoila,

this just didn't move up to the top of my priority list until now. @Amstutz, @thibsy and me agreed that I'll do an internal QA for the code of my colleagues before I pass this on for a UI coordinator review.

Best wishes!

src/UI/Component/ViewControl/Sortation.php
```diff
@@ -36,10 +36,15 @@ interface Sortation extends Component, JavaScriptBindable, Triggerer
     public function withLabel(string $label): Sortation;
 
     /**
-     * Get the label.
+     * Get the label (w/o the prefix).
      */
     public function getLabel(): string;
 
+    /**
+     * Set the prefix of the label
```
klees:
```suggestion
     * Set the prefix of the label.
     *
     * Defaults to: "Sorted By: " (or translations).
```

klees (changes requested):
Hi @nhaagen,

I think we need to solve a problem that now becomes visible,  and I think this is also the source of the initial problem described in the ticket.

So, please implement the following changes:
- [x] **wording(minor)**: Add a dot [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6026/files#diff-3bf57f0def01a582fbd32a3901c45159878ef93afab8c5a7f73ba356a04f964aR44). Also: Tell what this defaults to: "Defaults to: "Sorted By: " (or an according translation).".
- [x] Default Option: Add a default option to the [factory](https://github.com/ILIAS-eLearning/ILIAS/pull/6026/files#diff-51e6055af6ac844eff052c52119ecf8d26ca5251165090334be6bcd0ee5472c0L105). It should give one of the values from the `$options` that is used when the user did not decide otherwise. We could get rid of [Sortation::withLabel](https://github.com/ILIAS-eLearning/ILIAS/pull/6026/files#diff-3bf57f0def01a582fbd32a3901c45159878ef93afab8c5a7f73ba356a04f964aL36) and [Sortation::withSelected](https://github.com/ILIAS-eLearning/ILIAS/pull/6026/files#diff-3bf57f0def01a582fbd32a3901c45159878ef93afab8c5a7f73ba356a04f964aR88). Also, I guess this is the core of the issue: The sortation should always be one of the options. No need to set another "label". Please modify call sites (luckily there are only some of these currently...), tests, etc. accordingly.

The rest LGTM imo.

Best regards!

nhaagen:
Hi @klees, 
I left ::withSelected; the idea was to (pre-)factor the control and then set the value.
Although all the places I saw use the constructor over the mutator, I'd really like to keep the option.
There are checks  in place, too ;)

nhaagen:
rebased to trunk...

oliversamoila:
A really quiet ping.
We would like to integrate this at least in the dashboard for the release 9. Can you give us a realistic time frame?
Thank you very much.

nhaagen:
rebased to trunk

klees (approved):

klees:
@nhaagen Please rebase to release_9 again. We need it there...

klees:
@thibsy @Amstutz Would you be so kind and look into this?

thibsy (changes requested):
Hi @nhaagen

Thx a lot for your contribution to the UI components! We apologise for the delay here.

- [x] `C\ViewControl` (namespace): this PR raises the question again, what we should do with this namespace. This is more of an observation than a direct question, it does not need your response. However, you are most likely fixing an issue here which also exists in the new implementation (`C\Input\ViewControl\Sortation`), which is why we will not close the related mantis ticket. This issue will reappear when consumers start using the new approach.
- [ ] `sortation.js`: why are there two different versions of this file (internal and public), and why is it necessary at all? It does not seem to be required anywhere. Maybe something went wrong when rebasing the changes?

- [ ] `examples\ViewControls\Sortation\small`: why did you remove this example? I assume this example would still be applicable, to e.g. showcase how labels may be omitted for limited space, as it is described on the factory. Could you adjust the example instead?

- [x] `I\ViewControl\Factory`: please rename the `$default` parameter to match the one defined on the interface.
- [x] `SortationTest`: please use `array_key_first()` instead of `current(array_keys(...))` [on line 124](https://github.com/ILIAS-eLearning/ILIAS/pull/6026/files#diff-a4f6eb5dc1be6b6963856413a64d9aef475084801095478586273dc60a61ed21R124).

Kind regards,
@Amstutz and @thibsy

nhaagen:
Hi @thibsy ,
thanks for your comments.
Let's head over to #7136, though, and close this.

I fixed the Factory's signature and the Test (thanks).

>- [ ]  C\ViewControl (namespace): this PR raises the question again, what we should do with this namespace. This is more of an observation than a direct question, it does not need your response. However, you are most likely fixing an issue here which also exists in the new implementation (C\Input\ViewControl\Sortation), which is why we will not close the related mantis ticket. This issue will reappear when consumers start using the new approach.

This is true. The Components are becoming more and more different, and I'd really like to ditch these here.
However, the selected-tick is implemented in the Input/ViewControl/Sortation as well.

>- [ ] sortation.js: why are there two different versions of this file (internal and public), and why is it necessary at all? It does not seem to be required anywhere. Maybe something went wrong when rebasing the changes?

Sorry, that's a glitch. sortation.js is removed in #7136.

>- [ ] examples\ViewControls\Sortation\small: why did you remove this example? I assume this example would still be applicable, to e.g. showcase how labels may be omitted for limited space, as it is described on the factory. Could you adjust the example instead?

Good observation, thanks again, a little embarassing for me. Fact is, with the proposed implementation, this would simply not work anymore.
I therefore added ::withHiddenLabel and reinstated the small-example.

All changes are in #7136.

Best regards, 
Nils

PR #6029, opened 2023-06-20
[FIX] switched property type from DefaultRenderer to Renderer for plugin compatibility

chfsx:
Plugins can introduce a DecoratedRenderer with exchangeUIRendererAfterInitialization, but this does not work if the declaration points directly to DerfaultRenderer, as here. As far as I can see, this is the only place.


klees:
Thanks a lot!

klees:
Picked to trunk.

PR #5871, opened 2023-05-04
PasswordAssistance: Remove `session_*` function calls

mjansenDatabay:
This PR removes the `session_*` function calls from the password assistance. These function calls are not required anymore.

**!!!Important!!!**: If #6030 is merged, this PR here can be closed

PR #6030, opened 2023-06-21
Init/PasswordAssistance/UsernameAssistance: Use UI framework for password assistance forms

mjansenDatabay:
This PR introduces the UI framework forms to the `ilPasswordAssistanceGUI`.

Furthermore I did a little bit of "boy-scouting" and introduced the UI `messageBox` for some texts:

See:
- https://mantis.ilias.de/view.php?id=32727
- https://mantis.ilias.de/view.php?id=32729
- https://mantis.ilias.de/view.php?id=32730
- https://mantis.ilias.de/view.php?id=32742

And it fixes: https://mantis.ilias.de/view.php?id=32290

If approved and merged, #5871 can be closed.

### DIFF
--- a/Services/Init/classes/class.ilPasswordAssistanceGUI.php
+++ b/Services/Init/classes/class.ilPasswordAssistanceGUI.php
@@ -193,14 +193,18 @@ class ilPasswordAssistanceGUI
 
         $tpl->setVariable(
             'TXT_ENTER_USERNAME_AND_EMAIL',
-            str_replace(
-                "\\n",
-                '<br />',
-                sprintf(
-                    $this->lng->txt('pwassist_enter_username_and_email'),
-                    '<a href="mailto:' . ilLegacyFormElementsUtil::prepareFormOutput(
-                        $this->settings->get('admin_email')
-                    ) . '">' . ilLegacyFormElementsUtil::prepareFormOutput($this->settings->get('admin_email')) . '</a>'
+            $this->ui_renderer->render(
+                $this->ui_factory->messageBox()->info(
+                    str_replace(
+                        "\\n",
+                        '<br />',
+                        sprintf(
+                            $this->lng->txt('pwassist_enter_username_and_email'),
+                            '<a href="mailto:' . ilLegacyFormElementsUtil::prepareFormOutput(
+                                $this->settings->get('admin_email')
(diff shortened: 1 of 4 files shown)


Services/Init/classes/class.ilPasswordAssistanceGUI.php
```diff
@@ -115,55 +110,83 @@ public function executeCommand()
...
-        $form->setTitle($this->lng->txt('password_assistance'));
-        $form->setFormAction($this->ctrl->getFormAction($this, 'submitAssistanceForm'));
-        $form->setTarget('_parent');
+        return $this->refinery->custom()->constraint(
+            static function ($value): bool {
+                return is_string($value) && ilUtil::is_email($value);
+            },
+            $this->lng->txt('email_not_valid')
+        );
+    }
```
pascalseeland, line 144: I would suggest adding an email-transformation to the refinery as this looks like a very common kind of data to be validated

pascalseeland (commented):
Looks good to me. The only thing I would suggest is adding the email transformation if you think this would make sense to have in the refinery.

Services/Init/classes/class.ilPasswordAssistanceGUI.php
mjansenDatabay: Or even better: An `Email` data type for `.src/Data`? But maybe this should be a follow-up PR?!?

mjansenDatabay: @klees What do you think? `Transformation` oder `Data Type`? None of these suggestions?

If we add an alternative to `ilUtil::is_email`, we should re-use (or move code from) `ilMailRfc822AddressParserFactory`.

mjansenDatabay: @pascalseeland IMO we should merge this and add a TODO for the ROADMAP for the `Email`

oliversamoila:
Hello @pascalseeland 
please give us feedback on this in a timely manner.
Do you need anything else?

oliversamoila:
Hello @pascalseeland, please give us feedback on this PR.

oliversamoila:
Dear @pascalseeland 
we need feedback by Thursday in order to integrate the changes into Release 9 before Coding Completed. 
In the meantime, we have to assume that if there is nothing else to remark on, we can merge this status.
Best regards.

PR #6032, opened 2023-06-21
loosing check for sql query ends

iszmais:
This is a proposal for https://mantis.ilias.de/view.php?id=37515

Alternativily the addition of a log or error output would be even helpfull for that usecase.
Whatever is prefered.

### DIFF
--- a/Services/Database/classes/Setup/class.ilDatabasePopulatedObjective.php
+++ b/Services/Database/classes/Setup/class.ilDatabasePopulatedObjective.php
@@ -147,7 +147,7 @@ class ilDatabasePopulatedObjective extends \ilDatabaseObjective
                 $stack = '';
                 continue;
             }
-            if (!preg_match('/;$/', $line)) { // Break after ; character which indicates end of query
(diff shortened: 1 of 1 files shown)


klees (commented):
Hi everyone, I really do not have an opinion here. Sry.

iszmais:
HI @klees 

i assigned you because you got the auto-assignment for setup mantis tickets, therefore i assumed you are the maintainer here :sweat_smile: 

Pls correct me if im wrong there.

Greetings,
@iszmais

klees:
Hi @iszmais,

the auto assignment in mantis is fine. The categories in mantis that are not prefixed with "¥" are understood to be categories that users should be able to understand and use. Something happens during the setup. Still, many things that happen during setup are not caused by the setup component. Like this one here, where you actually do not touch code in `src/Setup` or `setup`. User won't be able to know or understand this, so I get the tickets (because stuff happens during setup) and just pass them along if applicable. For the code you can mostly look into the `maintenance.json` file and see who is responsible.

Best wishes!

chfsx:
Thanks @iszmais for the PR, I will cherry-pick this to trunk as well.

PR #6033, opened 2023-06-21
LSO: Use centralized online status and replace legacy form for settings

tbongers-cat:
This is the implementation of https://docu.ilias.de/goto_docu_wiki_wpage_7199_1357.html for the Learning Sequence. It also replaces the legacy form for Settings with a new UI form. An ILIAS update is required for the change in module.xml to activate offline handling.

The LSO's own implementation of ilLearningSequenceActivation is still kept here, as it also includes options for optional start/end dates of the online status. It will be revised with the implementation of https://docu.ilias.de/goto_docu_wiki_wpage_6059_1357.html.

With this PR, the on-/offline status of LSO objects is additionally updated via the object properties of ilObject, as described in https://github.com/ILIAS-eLearning/ILIAS/tree/trunk/Services/Object#Common-Settings. The Dashboard and List GUI have been updated to use the online status of ilObject instead of their own implementations.

With the implementation of the new UI form, the options for editing the legacy pages for intro/outro have been removed, as they were already scheduled for removal with ILIAS 9.

~~(Note: This includes a temporary fix for the issue described in Mantis #37503, which will be fixed later)~~

### DIFF
--- a/Modules/LearningSequence/classes/Settings/class.ilObjLearningSequenceSettingsGUI.php
+++ b/Modules/LearningSequence/classes/Settings/class.ilObjLearningSequenceSettingsGUI.php
@@ -25,7 +25,8 @@ class ilObjLearningSequenceSettingsGUI
     public const PROP_TITLE = 'title';
     public const PROP_DESC = 'desc';
     public const PROP_ONLINE = 'online';
-    public const PROP_AVAIL_PERIOD = 'online_period';
+    public const PROP_AVAIL_FROM = 'start';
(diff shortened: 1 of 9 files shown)


tbongers-cat:
Removed the fix for Mantis #37503, as it is not needed anymore.

tbongers-cat:
- Changed the behaviour on saving the object (as discussed with @klees)
- Changed the use of `ilObjectPropertyIsOnline()` as discussed with @kergomard in Mantis #37507

PR #6036, opened 2023-06-21
Filter courses and groups on last admin in unsubscribe multiple view

iszmais:
https://mantis.ilias.de/view.php?id=32130

Since the rbac 'leave' does not cover last admins anymore an additional fix is required

### DIFF
--- a/Services/Dashboard/ItemsBlock/classes/class.ilPDSelectedItemsBlockGUI.php
+++ b/Services/Dashboard/ItemsBlock/classes/class.ilPDSelectedItemsBlockGUI.php
@@ -1,5 +1,22 @@
 <?php
-/* Copyright (c) 1998-2011 ILIAS open source, Extended GPL, see docs/LICENSE */
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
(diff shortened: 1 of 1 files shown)


Services/Dashboard/ItemsBlock/classes/class.ilPDSelectedItemsBlockGUI.php
```diff
@@ -523,6 +540,12 @@ public function manageObject()
                 $group->setItems([]);
                 foreach ($items as $item) {
```
lscharmer, line 541: I would suggest using `array_filter` instead of a foreach, as you are just filtering an array.

lscharmer (changes requested):

Services/Dashboard/ItemsBlock/classes/class.ilPDSelectedItemsBlockGUI.php
iszmais: Imo this wouldnt be an improvement to the readability here and therefore i would prefer to keep it that way, as long as no convention or perfomance factor would oblige that change.

Also im quite hestiant to interfere with the surroundings of a bugfix within its PR.
Overall i approve general refactoring but within such PR i value the claritiy of intentions higher than the global progress on that front.

Also that often becomes kinda a bottomless pit if you check withing the bug for the surroundings and then the surroundings of the surroundings ... :smile:

lscharmer (approved):

Services/Dashboard/ItemsBlock/classes/class.ilPDSelectedItemsBlockGUI.php
lscharmer: Thanks for improving the if block.

PR #6048, opened 2023-06-22
Membership: Printing list of participants in courses not possible (30617)

schmitz-ilias:
This PR fixes [30617](https://mantis.ilias.de/view.php?id=30617) / [32662](https://mantis.ilias.de/view.php?id=32662). 

~~I'm hiding the inner header in the print output by fiddling with its css classes via javascript. This is maybe a bit inelegant, but I don't think it's possible otherwise, given how printing of participants is currently implemented.~~

### DIFF
--- a/Services/Membership/classes/class.ilAttendanceList.php
+++ b/Services/Membership/classes/class.ilAttendanceList.php
@@ -524,6 +524,12 @@ class ilAttendanceList
 
         $tpl = $DIC->ui()->mainTemplate();
         $tpl->setContent($this->getHTML());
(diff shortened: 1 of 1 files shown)


oliversamoila:
Hi @schmitz-ilias , 
does this lead back to the state that the print dialogue of the browser is opened?

schmitz-ilias:
> Hi @schmitz-ilias , does this lead back to the state that the print dialogue of the browser is opened?

Hi @oliversamoila! Yes, when clicking 'Print List' you are shown the same table as before, but in addition the dialogue pops up.

PR #6049, opened 2023-06-22
T&A Bugfix #0024740: Fatal Error Message during Test Run

fhelfer:
https://mantis.ilias.de/view.php?id=24740


kergomard (commented):
Thank you very much @fhelfer for the PR!

I would have two little requests:
- Could you please move the duplicate code to a function `ilObjTestGUI::redirectAfterMissingRead()` parallel to the other one [`ilObjTestGUI::redirectAfterMissingWrite()`](https://github.com/ILIAS-eLearning/ILIAS/pull/6049/files#diff-1cb659c07cfa02fb1a744e11d5b0cf83003aa4d4a493c7f0943005f089d4cfffL932).
- Could you please also check the other usages of `->raiseError` in this function and change the pattern for those, too? That would be very kind!

@kergomard

kergomard:
Thank you very much @fhelfer ! Will merge and cherry-pick to ILIAS 8 and trunk.

PR #6050, opened 2023-06-22
Init/ErrorHandling: Hide sensitive information

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=37522

If approved, this MUST be chery-picked to `release_8` and `trunk`. Please introduce `Typed Properties` for these branches.

In my opinion this should be the **last** fix of this kind, at least for the user interface part. Sensitive data in log files should still be obscured in future.

Why?

1. Adding parameter names considered `sensitive information` to a list always fixes a local problem.
2. Using an exclusion list in general is not recommended
3. We have no mechanism to enforce a particular name for sensitive parameters (like `passwd`, `token`, etc.).
4. Using exclusion lists will be even harder if the `UI Input` forms are used more and more.
5. `PrettyPageHandler` supports an exclusion list natively (but only for values on the first level of a super global array), `ilPlaintextHandler` and `ilLoggingErrorFileStorage` have nothing in common (base class, interface), so the exclusion list must be injected from a central location by calling methods being not a part of any contract.
6. (All the arguments of the) Official statement of the vendor: https://github.com/filp/whoops/issues/571
7. With PHP 8.2 sensitive data could be handled on language level, see: https://stitcher.io/blog/new-in-php-82#redact-parameters-in-back-traces-rfc / https://wiki.php.net/rfc/redact_parameters_in_back_traces , allthough I doubt that the super global arrays could be tackled with this.

### DIFF
--- a/Services/Exceptions/classes/class.ilPlainTextHandler.php
+++ b/Services/Exceptions/classes/class.ilPlainTextHandler.php
@@ -18,6 +18,19 @@ class ilPlainTextHandler extends Handler
 {
     const KEY_SPACE = 25;
 
+    /** @var list<string> */
+    private $exclusion_list = [];
+
+    /**
+     * @param list<string> $exclusion_list
+     */
+    public function withExclusionList(array $exclusion_list) : self
+    {
+        $clone = clone $this;
+        $clone->exclusion_list = $exclusion_list;
+        return $clone;
+    }
+
     /**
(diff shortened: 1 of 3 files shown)


oliversamoila:
Hello @pascalseeland, please give us feedback on this PR.

mjansenDatabay:
Hi @pascalseeland, any objections to merge/pick this?

PR #6061, opened 2023-06-26
Mt 32014 release 8

chlulei:
Addresses mantis ticket 32014: https://mantis.ilias.de/view.php?id=32014

### DIFF
--- a/Modules/Course/classes/class.ilObjCourseAdministrationGUI.php
+++ b/Modules/Course/classes/class.ilObjCourseAdministrationGUI.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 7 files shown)


Modules/Course/classes/class.ilObjCourseAdministrationGUI.php
```diff
@@ -39,4 +42,31 @@ protected function getAdministrationFormId(): int
     {
         return ilAdministrationSettingsFormHandler::FORM_COURSE;
     }
+
+    protected function addChildContentsTo(ilPropertyFormGUI $form): ilPropertyFormGUI
+    {
+        $checkBox = new ilCheckboxInputGUI($this->lng->txt('crs_my_courses_groups_enabled'), self::SETTING_COURSES_AND_GROUPS_ENABLED);
+        $checkBox->setInfo($this->lng->txt('crs_my_courses_groups_enabled_info'));
+        $checkBox->setChecked((bool) $this->settings->get(self::SETTING_COURSES_AND_GROUPS_ENABLED, 0));
```
smeyer-ilias: Please set the default to "1"

smeyer-ilias (changes requested):
Thanks @chlulei, please fix the default value

smeyer-ilias:
Thanks @chlulei!

iszmais:
@chlulei Within the procession of the ticket i stumbled upon 1 criteria:
```Default status for both new settings is enabled.```
And since a default should be enforced by the providing source and not the the demanding i would appreciate if you could also add an update step to set such default, just like @alex40724 did in the similar setting: https://github.com/ILIAS-eLearning/ILIAS/commit/28962f2945712d231c94cb322f0e39e3ec9b6e13

PR #6064, opened 2023-06-26
0037524: 404-Error bei Lautstärkeänderung in einer Frage - 404 error when changing volume in a testquestion, status "editing"

PurHur:
Probably fixes https://mantis.ilias.de/view.php?id=37524#c94252

### DIFF
--- a/Modules/Test/js/ilTestPlayerQuestionEditControl.js
+++ b/Modules/Test/js/ilTestPlayerQuestionEditControl.js
@@ -633,7 +633,9 @@ il.TestPlayerQuestionEditControl = new function() {
         }
 
         // let the backend decide where a focus element is using the html id focus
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @PurHur for the PR! I merged the PR here https://github.com/ILIAS-eLearning/ILIAS/pull/6089 instead.

Best,
@kergomard

PurHur:
Note that the #focus achor is still an error in javascript actions.

Rillke:
@PurHur 

> Note that the #focus achor is still an error in javascript actions.

Can you elaborate on that and eventually provide a PR?

PurHur:
The line i fixed appends #focus to links. But if the link starts with javascript: is should not be appended since #focus is not valid js (it would be an undefined token/variable):

image

This is the PR that fixes this. Since the other fix fixes the eventhandling this does currently have no known error but its still wrong.

PR #6066, opened 2023-06-26
Calendar: Creating an appointment without edit calendar permission confused (25350)

schmitz-ilias:
This PR fixes [25350](https://mantis.ilias.de/view.php?id=25350) for R8 and trunk, see #6065.


smeyer-ilias:
Thanks @schmitz-ilias

PR #6067, opened 2023-06-27
addresses mantis bugreport 37262

chlulei:
Mantis ticket: https://mantis.ilias.de/view.php?id=37262

### DIFF
--- a/Services/Search/classes/class.ilRepositorySearchGUI.php
+++ b/Services/Search/classes/class.ilRepositorySearchGUI.php
@@ -1198,14 +1198,14 @@ class ilRepositorySearchGUI
                 $this->refinery->kindlyTo()->string()
             );
         }
-        $is_in_admin = $base_class === ilAdministrationGUI::class;
+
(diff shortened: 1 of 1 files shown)


Services/Search/classes/class.ilRepositorySearchGUI.php
```diff
@@ -1204,6 +1204,8 @@ protected function showSearchUserTable(array $a_usr_ids, string $a_parent_cmd):
             ilSession::set('usr_search_link', $this->ctrl->getLinkTarget($this, 'show'));
         }
 
+        // String value of 'baseClass' is in lower case
```
smeyer-ilias: Please adapt the line 1201, which also did not take the case into account.

smeyer-ilias (changes requested):

Services/Search/classes/class.ilRepositorySearchGUI.php
smeyer-ilias: This is not valid anymore, thanks @chlulei

PR #6068, opened 2023-06-27
mode-info: improvements to responsiveness and html-template-structure

catkrahl:
**some improvements to mode-info:**
* flex-layout
* separation into mode-info-content- and mode-info-close-container in the html-template
* mode-info-box will resize depending on the overall browser-window-width
* rounded lower corners for a clear visual separation to legacy elements in the GUI - following the concept of the new mode-switch-design


templates/default/070-components/UI-framework/MainControls/_ui-component_mode_info.scss
```diff
@@ -31,31 +38,37 @@ $il-mode-info-shadow: 0px 2px 2px 0px rgba(128, 128, 128, 1);
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: $il-mode-info-height;
+  pointer-events: all;
+  position: relative;
+}
+
+.il-mode-info-content {
+  border-bottom-left-radius: calc($il-border-radius-large + 10px);
```
catenglaender: Maybe use $il-border-radius-secondary-large instead of the calculation? We could consider making this variable 10 px as it doesn't seem to be currently in use anywhere else.

templates/default/070-components/UI-framework/MainControls/_ui-component_mode_info.scss
```diff
@@ -31,31 +38,37 @@ $il-mode-info-shadow: 0px 2px 2px 0px rgba(128, 128, 128, 1);
...
-
-   pointer-events: all;
+  padding: 0 4% 0 5%;
 }
 
-.il-mode-info span {
-  color: $il-mode-info-text-color;
-  font-size: $il-font-size-large;
+.il-mode-info-close {
+  border-bottom-right-radius: calc($il-border-radius-large + 10px);
```
catenglaender: Maybe use $il-border-radius-secondary-large instead of the calculation? We could consider making this variable 10 px as it doesn't seem to be currently in use anywhere else.

templates/default/070-components/UI-framework/MainControls/_ui-component_mode_info.scss
```diff
@@ -31,31 +38,37 @@ $il-mode-info-shadow: 0px 2px 2px 0px rgba(128, 128, 128, 1);
...
-.il-mode-info span.il-mode-info-content {
-  margin: auto;
+.il-mode-info-content,
+.il-mode-info-close {
   background-color: $il-mode-info-color;
-  padding: $il-padding-base-vertical $il-padding-base-horizontal $il-padding-base-vertical $il-padding-base-horizontal;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: $il-mode-info-height;
```
catenglaender: Maybe use min-height so multi-line text can push the area down? Unfortunately, this wouldn't push down the part of the bar with the close button in it. How likely is multi-line text in the mode info (especially on mobile)?

src/UI/templates/default/MainControls/tpl.mode_info.html
```diff
@@ -1,5 +1,9 @@
 <div class="il-mode-info">
-	<span class="il-mode-info-content">
-		{MODE_TITLE}{CLOSE_GLYPH}
-	</span>
+	<div class="il-mode-info-content">
```
catenglaender: Maybe there should be another div to contain both the content and the close button that could have the background color and the rounded corners and be pushed open by multi-line text?

catenglaender (commented):
I really like the design. It's so nice that we have rounded corners that purposfully communicate a layering where appropriate. I attached a couple comments in the code.

catkrahl:
image

catkrahl:
image

src/UI/templates/default/MainControls/tpl.mode_info.html
catkrahl: do you think we can use the outer div "il-mode-info" for this?

catenglaender: Isn't that outer container doing the fullwidth and fullheight frame around the page? I don't think it can double as the bar. Unless... maybe you could draw the page frame with il-mode-info::before/after and free up il-mode-info to be the bar this way.

catenglaender:
@catkrahl Wonderful! I really like the new structure and the clear separation of what div has what purpose. :+1: 

Two more things I would like to consider:

- [ ] How does this PR look with system and mode infos both visible?
- [ ] Is the code in the following comment really necessary or can we attach the font settings to close btn and label directly?

templates/default/070-components/UI-framework/MainControls/_ui-component_mode_info.scss
```diff
@@ -1,61 +1,101 @@
...
+  &__mobile-padding {
+    @media only screen and (max-width: $il-grid-float-breakpoint-max) {
+        height: $il-standard-page-small-header-height;
+    }
   }
 
-  span.il-mode-info-content {
-	display: block;
-	width: 100%;
+  * {
```
catenglaender: Is it really necessary to be so general?

templates/default/070-components/UI-framework/MainControls/_ui-component_mode_info.scss
```diff
@@ -1,61 +1,101 @@
...
+    color: $il-mode-info-text-color;
+    font-size: $il-font-size-large;
   }
 }
 
+@media only screen and (max-width: $il-grid-float-breakpoint-max) {
+
+}
+
+/*
```
catenglaender: commented out code can be removed

templates/default/delos.css.map
catenglaender, line 1: No need to commit the css map

catenglaender (changes requested):

templates/default/delos.css.map
catkrahl: danke für den Hinweis ...

templates/default/070-components/UI-framework/MainControls/_ui-component_mode_info.scss
catkrahl: it's just a bit shorter

catkrahl:
the combined view with system and mode info active looked similar to the view without the suggested changes ... 
image

catkrahl:
this was the combined view before the suggested changes ...

image

catenglaender (approved):
Looks good. Thank you so much! :+1: 
From my end, the review is completed. :slightly_smiling_face:

catenglaender (approved):

PR #6069, opened 2023-06-27
mantis ticket 36625

chlulei:
mantis ticket: https://mantis.ilias.de/view.php?id=36625

### DIFF
--- a/Modules/Session/classes/class.ilSessionOverviewGUI.php
+++ b/Modules/Session/classes/class.ilSessionOverviewGUI.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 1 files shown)


Modules/Session/classes/class.ilSessionOverviewGUI.php
```diff
@@ -120,9 +120,15 @@ public function exportCSV(): void
         $sortedMembers = [];
         foreach($part as $user_id) {
             $name = ilObjUser::_lookupName($user_id);
-            $sortedMembers[$name['lastname']] = [ $user_id, $name["firstname"] ];
+            $sortedMembers[] = [
+                'userid' => $user_id,
+                'firstname' => $name["firstname"],
```
smeyer-ilias: Please add string typecasts, since this $name['firstname'] might be null

Modules/Session/classes/class.ilSessionOverviewGUI.php
```diff
@@ -120,9 +120,15 @@ public function exportCSV(): void
...
+                'lastname' => $name['lastname']
```
smeyer-ilias: Same here

smeyer-ilias (changes requested):

PR #6070, opened 2023-06-27
UI/Table/Presentation: expand/collapse all, use Alignments

nhaagen:
As part of improving T&A-Results (https://docu.ilias.de/goto.php?target=wiki_1357_Improve_Presentation_of_T%26A-Results),
this adds a toggle-all and some icon in the row-title.

There are already references to the Alignment: https://github.com/ILIAS-eLearning/ILIAS/pull/5954

Screenshot from 2023-06-27 12-16-44

Screenshot from 2023-06-27 14-42-47


dsstrassner:
Hi @nhaagen 
there is a special question Icon: 

https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/templates/default/images/icon_ques.svg

Could you please use this? (Oliver Samoila has seen your mockup 😉)

nhaagen:
The icon is consumer-code ;)
```
$mapping_closure = function ($row, $record, $ui_factory, $environment) {
    return $row
        ->withLeadingSymbol(
            $ui_factory->symbol()->icon()->custom('templates/default/images/icon_ques.svg', '')
        )
```

klees:
@Amstutz @thibsy For me this is good to go from interface and implementation perspective. Only missing part is some SCSS, @catenglaender will take care once he is back.

catenglaender:
Can we wait for https://github.com/ILIAS-eLearning/ILIAS/pull/5811 to be merged? It would change the appearance of View Controls. I think it would make sense to style the "expand all" button in a similar way as it is part of the view control bar.

Amstutz:
The public interface changes are JF ready. The implementation will be reviewed as soon as @catenglaender considers them to be done.

matthiaskunkel:
Jour Fixe, 10 JUL 2023 : We highly appreciate this suggestion and accept the PR for trunk.

catenglaender:
@yvseiler and I did some research and had a discussion about the expand/collapse glyph/button, with the following result:

We propose the (unfortunately not so pretty) solution to have one button labled "expand all" and one labled "collapse all" instead of a toggle for the following reasons:

* State Ambiguity: Just like with a "mute microphone" button in video conferencing software, users don't know for sure if the toggle shows the state that is currently active or the state that will be executed on click.
* People we asked had difficulties reading the glyph in all drafts. They guessed that the glyph would expand the view controls. Glyphs that don't look like the accordion toggle have to be learned and we think the we would prefer spending the user's mental energy teaching them more widely used glyphs like the symbol for sortation.
* Users might have already opened some entries by hand and now might want to make the deliberate choice to collapse the opened ones or expand the closed ones. The need to offer both options at the same time is therefor valid.
* We believe that the view controls of the presentation table are likely to grow to include sortation, column/attribute show/hide and others, so maybe there will be an expand/collapse viewcontrols option in the future that should not rival the collapse/expand all in any way.

This is why we suggest to go this path that is certainly not the most elegant looking, but one that will be the most clear to operate for the user.

klees:
@nhaagen: Make it so, please =)

Amstutz:
@klees, I passed this to you ftm. Feel free to pass back for a final look as soon as you this is done.

nhaagen:
Screenshot from 2023-07-11 14-15-18

catenglaender:
This is how it looks now:
image

klees:
@Amstutz This is good for final review IMO.

Amstutz:
Hi @nhaagen 

Thx a lot for expending this important component. The implementation looks fine. 

Since the example has already been there, this can be kept as is, however, please try to avoid inline styling wherever possible in the future. We should try to set an as good an example as possible, it is very plausible that those examples are the first contact with ILIAS for new devs.

Please add, if possible, or shortly explain not:
- [ ] JS Testing: Can you test JS logic here? If this is beyond the scope of this rather small change, it is fine. However, JS Tests would be highly appreciated in general.

Final note, I had to check twice if I did not miss any SCSS code. However, seems to have worked without any change. Nice.

thx
@Amstutz

klees:
@nhaagen Also: Please add test cases for the UI component in testrail once this is merged. Thx!

nhaagen:
Hi @Amstutz ,
there was no chance (for me) to get the former JS under tests, so I rebuilt it and used the FactoryPattern with rollup.
However, the tests are still utterly basic, since Mocha/JSDOM does not know about querySelector; but at least, the path is open to proper testing and I'll have another look into it later today.
Best regards, 
Nils

Amstutz:
thx you very much!

PR #6074, opened 2023-06-28
Org.Units: Don't add missing CSS file

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=37540

If approved, this should be picked to `trunk` as well.


klees:
Removed a superfluous `;` and merged to release_8 and trunk manually. Thanks a lot!

PR #6077, opened 2023-06-28
T&A Bugfix #0034769: Order Question Standard Action for enter key

fhelfer:
https://mantis.ilias.de/view.php?id=34769

Makes the Save Button the default submit button on enter
If thats hacky Javascript is not desired, we should consider making the default button the first in Layout


fhelfer:
This seems to be also issue for 8 & trunk

kergomard (commented):
Hi @fhelfer 
Thank you very much for the PR as always!

Ok, this one is very painfull ;-), ...but I don't really have a better idea. So, let us make really clear, that this is a real hack. Could you please:
- Move the javascript to a `heredoc` - String and load it with `tpl::addOnLoadCode()`.
- While you do this: Could you please also move the retrieval of the `$DIC` and the `tpl`-Variable to the beginning of the constructor.
- I would personally move from `button.click()` to `form.requestSubmit(button)`, but this might be a matter of personal taste, I'm not sure. You can change it or not.

Thanks again and best,
@kergomard

kergomard:
Thank you very much @fhelfer . I would have added it with `<<<TAG` - Syntax, but this works good enough.

PR #6081, opened 2023-06-28
T&A 0016123 Fix Multiple issues related to linking answers and feedback

kunzeus:
Ticket:
https://mantis.ilias.de/view.php?id=16123 Continuation, I'm sorry

Symptom:
Import has no feedback
SQL error
Error on invalid saves

Problem:
Import: The importer has a code path running through the code I have changed and I was unaware this circumstance.
SQL: Mapping error
Invalid saves: on an save with missing points ids have been lost

Solution:
Import: the code reorganizing the feedback is no longer used on imports.
SQL: Fix mapping
Invalid saves: added missing id handover

Checked solution on:
ILIAS 7

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assAnswerMultipleResponseImage.php
+++ b/Modules/TestQuestionPool/classes/class.assAnswerMultipleResponseImage.php
@@ -1,8 +1,8 @@
 <?php
 /* Copyright (c) 1998-2013 ILIAS open source, Extended GPL, see docs/LICENSE */
 
-include_once "./Modules/TestQuestionPool/classes/class.assAnswerMultipleResponse.php";
-include_once "./Modules/Test/classes/inc.AssessmentConstants.php";
+include_once './Modules/TestQuestionPool/classes/class.assAnswerMultipleResponse.php';
+include_once './Modules/Test/classes/inc.AssessmentConstants.php';
 
 /**
 * ASS_AnswerBinaryStateImage is a class for answers with a binary state
@@ -37,16 +37,17 @@ class ASS_AnswerMultipleResponseImage extends ASS_AnswerMultipleResponse
     * @param integer $order A nonnegative value representing a possible display or sort order
     * @param double $points_unchecked The points when the answer is not checked
     * @param string $a_image The image filename
+    * @param integer $id The database id of the answer
     * @access public
     */
     public function __construct(
-        $answertext = "",
+        $answertext = '',
         $points_checked = 0.0,
         $order = 0,
         $points_unchecked = 0,
-        $a_image = "",
+        $a_image = '',
         $id = -1
-  ) {
+    ) {
         parent::__construct($answertext, $points_checked, $order, $points_unchecked, $id);
         $this->image = $a_image;
     }
--- a/Modules/TestQuestionPool/classes/class.assMultipleChoice.php
+++ b/Modules/TestQuestionPool/classes/class.assMultipleChoice.php
@@ -756,6 +756,7 @@ class assMultipleChoice extends assQuestion implements ilObjQuestionScoringAdjus
             foreach ($this->answers as $answer){
                 // Only the first appearance of an id is used
                 if ($answer->getId() !== null && !in_array($answer->getId(), array_keys($post_answer_order_for_id))) {
+                    // -1 is happening while import and also if a new multi line answer is generated
                     if ($answer->getId() == -1) {
                         continue;
                     }
--- a/Modules/TestQuestionPool/classes/class.assMultipleChoice.php
+++ b/Modules/TestQuestionPool/classes/class.assMultipleChoice.php
@@ -749,7 +749,7 @@ class assMultipleChoice extends assQuestion implements ilObjQuestionScoringAdjus
         // Check if feedback exists
         if (sizeof($feedback) >= 1
             && $this->getAdditionalContentEditingMode() == 'default'
-            && !$ctrl->getCmd() == 'importVerifiedFile'){
+            && !($ctrl->getCmd() == 'importVerifiedFile')){
             // Get all existing answer data for question
             $result = $ilDB->queryF(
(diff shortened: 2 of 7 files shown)


kergomard (changes requested):
Thank you very much @kunzeus for this PR!
I don't know how to test this right now and it is not easy to grasp, thus allow me a few questions, sorry:
- Under which circumstances does [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/6081/files#diff-a2b169b5bf9f80a17c6424d0472445235a33f80d5790027560f083f608d9d35eR759) return `-1`?
- [This here](https://github.com/ILIAS-eLearning/ILIAS/pull/6081/files#diff-a2b169b5bf9f80a17c6424d0472445235a33f80d5790027560f083f608d9d35eR777) looks very fragile. If I'm getting it right, we need to be completely sure, that `$db_order_ids` is a numeric array without any missing keys, but above we are setting the keys explicitly. If you are sure, that this is always the case, then I'm fine with this, but I would still ask you to break this line up in sub statements, naming all variables very verbosely, ie. `$sorted_question_id_from_db = array_keys($db_order_ids)` (if I'm getting this right).

The same applies to the SingeChoice-Question.

Please also resolve the conflicts, they are due to me fixing a few of the issues in the meantime, sorry!

Thanks again and best,
@kergomard

kunzeus:
> * Under which circumstances does [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/6081/files#diff-a2b169b5bf9f80a17c6424d0472445235a33f80d5790027560f083f608d9d35eR759) return `-1`?

First creation of a new question, but there is also no feedback. 
In the case someone fixes the JavaScript, and we don't get full row duplicates of answer rows, the default for the field is set to '-1'. Furthermore, it's set to '-1' in some calls to the constructor. This might prevent some unknown fuzzy edge cases we haven't encountered yet.
But it's defnetly the case on the import of a test, so this fixes the import since all questions are new questions and therefore have a -1 by constructor.

> * [This here](https://github.com/ILIAS-eLearning/ILIAS/pull/6081/files#diff-a2b169b5bf9f80a17c6424d0472445235a33f80d5790027560f083f608d9d35eR777) looks very fragile. If I'm getting it right, we need to be completely sure, that $db_order_ids is a numeric array without any missing keys, but above we are setting the keys explicitly. If you are sure, that this is always the case, then I'm fine with this, but I would still ask you to break this line up in sub statements, naming all variables very verbosely, ie. $sorted_question_id_from_db = array_keys($db_order_ids) (if I'm getting this right).

I made a commit that changes the naming scheme. I'd like to say that: "I'm sure this will always work", but I don't feel like I can promise this. The answers themselves come from the database and are as complete as they are stored there.  If you believe it's good to generally check the data from the database, I will implement some exceptions. 
For the diff calculation. I'm sorry if that complicates things. Here is a picture of what I try to achieve: I hope it eases the understanding of my code. I'm now half a day into testing stuff and wild input behaviors. I'm hoping to see something break, but currently it seems to work. Id like to here if you manage to break it, send me a message on discord. I'll continue testing for a bit cause I still don't trust the code fully. Especially the reordering of the feedback.
mapping

Still, I don't like my solution, and I'm in favor of reverting the original commit. Maybe I didn't say it long, or loud enough, or because of the long time it took, it's been forgotten: it's only a **partial** solution to the matching problem. Which means: no fix for the PageEditor feedback, and there's still the problem with the JavaScript duplications of the answers. 
Also, I believe it's the wrong strategy: adding to the existing code and not making a systematic change. I.e. create a relationship between answer and feedback that does not depend on the order of answers and feedback, but is rather by reference (foreign key). I'm open for every suggestion leading to a more stable solution, but I think it's not done by the smallest possible fix.

kergomard:
Thank you very much kunzeus! ...and sorry I kept you waiting for so long!
I understood the partial part and I think we should not move backwards (especially as I believe we will have a hard time removing your original commit, I think). I will test this on the surface, but your naming changes helped a lot (at least for me). We surely agree that it isn't a beautiful solution, but there was nothing beautiful about the point you started from, so: Absolutely and completely not your fault!
Thanks again and best,
@kergomard

kergomard:
Hi @kunzeus 

So, now I checked it. I can still reproduce the error noted by @PurHur [here](https://mantis.ilias.de/view.php?id=16123#c95132). To provoke it:
- Create a MC-Question with Feedback created through the TinyMCE.
- Delete one of the Answer-Options from the middle and save. -> If you check your DB here, there will still be all entries there. Nothing will have been deleted.
- Now try to copy the test containing the question and it will fail (with
```An undefined Database Exception occured. SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'answer' cannot be null QUERY: INSERT INTO qpl_fb_specific (feedback_id, question_fi, answer, tstamp, feedback, question) VALUES (<id>, <q_id>, NULL, <tst>, <cont>, 0)```. The problem is, that as there actually is no answer, as there are too many feedbacks, [`$post_answer_order_for_id[$db_answer_id]`](https://github.com/ILIAS-eLearning/ILIAS/pull/6081/files#diff-a2b169b5bf9f80a17c6424d0472445235a33f80d5790027560f083f608d9d35eR795) will always be `null`.

Don't hesitate to contact me, if you have any questions! Please also do so, if you don't feel comfortable to work on here.

Thanks for all your work and best,
@kergomard

kergomard:
Thank you very much @kunzeus for the update. Would this be ready for another review?

...and could you please rebase it to resolve the conflicts.

Thanks again and best!
@kergomard

kergomard (changes requested):

I think we are down to one final iteration: 
- This solves the issue now, ...but it doesn't tackle the problem of existing broken questions: I would suggest to catch this, as now a few already broken questions might have accumulated.

Thank you very much and best!
@kergomard

kergomard:
Thanks again @kunzeus ! This now looks right to me. I merge and cherry-pick to ILIAS8 and trunk.

PR #6083, opened 2023-06-28
TA: patch for #37559

klees:
Hello everyone!

This does not fix the root cause, though. The id should never be a string in the first place, hence we should tackle the form that causes this at some point. Obvious problem would be, that a missing value in the field would do something with something with id=0, because (int)null === 0, sadly.

Best wishes!


kergomard:
Thank you very much @klees !
Ok. This is one more reason why these forms should go, go, go, go ....!

PR #6085, opened 2023-06-28
UI: allow defining relationships in links

schmitz-ilias:
This PR adds the option to add relationships between current and referenced page to the `rel` attribute of a link. Just the interface for now, as I'm not quite sure whether what I propose here is the best way forward.

As discussed in the last Jour Fixe, this extension is needed to be able to mark links as a `license`, as required for [Moving Editing of Available Copyrights to Kitchensink](https://docu.ilias.de/goto_docu_wiki_wpage_7839_1357.html). Since `license` will probably only ever be used by `Services\MetaData`, and other relationships might be useful elsewhere, I expanded the scope a bit: I included all 13 possible values for the `rel` attribute as defined [here](https://html.spec.whatwg.org/multipage/links.html#linkTypes) (although information on what values can and can't be used in the `rel` attribute are all over the place).

What do you think? Is this actually useful, or should I pare this down to just the `license`?


klees (changes requested):
Hi @schmitz-ilias,

thanks a lot for this proposal and the tight presentation. This seems to be fine in general, so we are wondering a little why you are hesitant =)

From the perspective of the UI framework we are not quite sure that all the proposed `Relationship`s really are something that consumer could or should choose to express. E.g.: `prev` and `next` would be something that might be better processed via some `View Control` to control back and forth movement in some context. We might need these `Relationship`s to use them internally nevertheless. So, general approach looks fine, but we would suggest to strip down the list of available `Relationship`s and leave a comment to this effect in the enum.

- [ ] Alternatives: You mention in you comment that you are not sure if this is the way forward. Why is that? Is it only the extent of the list of `Relationships`? Did you ponder general alternatives?

- [ ] `Relationships`: Remove `next`, `prev`, `tag`, `help`, `search` from the enum. Alternatively: add some indication that these are for internal use.

Kind regards!
@Amstutz, @thibsy, @klees

schmitz-ilias:
Hi all,

I was a bit hesitant because I wasn't sure how much control the KS needs to have over the semantics of its links. If one wants greater control, one could alternatively add more types of links, something like `Link\License`, `Link\Author` and so on (which would maybe have the drawback of being overly specific). 
Further, as far as I understand it, some values of `rel` have a very direct effect on how browsers handle the link (`noopener`, `noreferrer`). At least `noopener` is already set when a link has `withOpenInNewViewport`. Again, not sure how much control over this should be given to consumers of the KS.

I'll split up the `Relationship` enum into two, the additional one being for those reserved for internal use in the KS. That way, we can very easily change what is available externally and what should be kept inside the KS, keeping the door open for more specific semantics in the future.

Cheers, and thanks for the feedback,
Tim

Amstutz:
Hi @schmitz-ilias

The public interface looks good. You are welcome to start the implementation if this helps the project behind this request. Note that since this is a minor change, we might fast track this as soon as implemented. Meaning, we can merge it, and announce it to the JF afterward.

thx.
@Amstutz , @klees , @thibsy

schmitz-ilias:
I implemented the relationships. I had to change one small detail in the interface: `Link:getRelationshipsToReferencedResource` needs to return `Relationship` as well as `RelationshipForInternalUse`.
A way to actually add a `RelationshipForInternalUse` is not included in the implementation, as how that should be done depends on how exactly the KS wants to use these relationships. They are already respected when rendering.

matthiaskunkel:
Jour Fixe, 10 JUL 2023 :  We highly appreciate this suggestion and accept the PR for trunk.

klees:
@schmitz-ilias Is this ready for review of the implementation?

schmitz-ilias:
> @schmitz-ilias Is this ready for review of the implementation?

Yes, it's ready, see my last comment.

klees (changes requested):
Hi @schmitz-ilias,

* [ ] Remove rel: We currently do not have the possibility to remove certain relationships (or purge all). Is this intentional? Any idea if we would need something like that?

* [ ] Interface: Please add an interface IsRelationship (could not come up with a better name...) and mark Relationship and RelationshipForInternalUse with that interface. Then hint on that interface to make Link accept RelationshipForInternalUse as well.

* [ ] RelationshipForInternalUse: Please move this enum to our internal Implementation namespace (=> UI/Implemenation/Component/Link/RelationshipForInternalUse).
* [ ] Separate Test: IMO the second part of you automated test checks if one relationship actually appears only once, even if it was added a second time. Thanks for checking that. Please move that separate test into a seperate method.
* [ ] Checks: Make them pass =)

Feel free to contact me on short notice if there is anything to discuss.

Best regards!

schmitz-ilias:
Hi @klees, thanks for the feedback!

>     * [ ]  Remove rel: We currently do not have the possibility to remove certain relationships (or purge all). Is this intentional? Any idea if we would need something like that?

I don't think removing relationships is necessary. In a pinch, it should be easy enough to create a new link with the required relationships.

>     * [ ]  Interface: Please add an interface IsRelationship (could not come up with a better name...) and mark Relationship and RelationshipForInternalUse with that interface. Then hint on that interface to make Link accept RelationshipForInternalUse as well.

Good idea, thanks! I got stuck on not being able to use inheritance with enums, but using a marker interface is a nice workaround.

>     * [ ]  RelationshipForInternalUse: Please move this enum to our internal Implementation namespace (=> UI/Implemenation/Component/Link/RelationshipForInternalUse).
> 
>     * [ ]  Separate Test: IMO the second part of you automated test checks if one relationship actually appears only once, even if it was added a second time. Thanks for checking that. Please move that separate test into a seperate method.
> 
>     * [ ]  Checks: Make them pass =)

Done!

In addition, I moved around how `noopener` is added as a relationship for links that open in a new viewport. In the previous implementation, toggling `Link::withOpenInNewViewport` on and off again would still have added the relationship.

Let me know if there is anything else. Cheers!

klees (approved):
LGTM, thanks for the good work and patience.

PR #6086, opened 2023-06-29
Fix/9/plugin event handling

chfsx:
This is a followup to the PR https://github.com/ILIAS-eLearning/ILIAS/pull/5316 and closes the problem in https://mantis.ilias.de/view.php?id=37532.

With this PR
- the plugin.xml is read in again and the events are saved accordingly: https://github.com/ILIAS-eLearning/ILIAS/pull/6086/commits/2d07f47958dfca52fdb0a6c0fc3a4259f592d950 this is for @kergomard 
- the information is no longer stored in the DB (which is no longer possible because the tables have been removed) and the corresponding code is removed from ilPlugin, which was responsible for this: https://github.com/ILIAS-eLearning/ILIAS/pull/6086/commits/19e52d83e818f365acf5795d4dfb40ee91a5f2ac this is for @klees 

In a further step, we should ensure that the plug-ins here approximate the general behaviour of components, so that such a distinction [here](https://github.com/ILIAS-eLearning/ILIAS/blob/aa29a8b24b57804d95009a23727e014a56beacdd/Services/EventHandling/classes/class.ilAppEventHandler.php#L126) is no longer necessary. I think, however, that we should look at event handling in general to upgrade it to a real event system and get away from the hook approach.


klees (approved):
Thanks a lot =)

chfsx:
So that seems to cause problems, at least in the CI: 

https://github.com/srsolutionsag/ILIAS/actions/runs/5409625499/jobs/9829991811#step:4:380

The reason for this is probably that artifact-building now requires the component repository: https://github.com/ILIAS-eLearning/ILIAS/pull/6086/files#diff-32010d332874bdf011e96107a217be640397fa34ee08d6a357b612d5ec0273daR33

klees:
Will debug...

klees:
Can reproduce without local `/data`-dir (which should be what GitHub sees...)

klees:
Ok, problem definetly is that we would now require a complete `ilComponentRepository` to build artifacts, which in turn needs an initialized database to pull state-info about plugins. This cannot exist during artifact building, as these are supposed to be static artifacts. Does event handling for plugins properly look into activation of plugins? Might be that we handlers where added on activation/deactivation of plugins before. I will look into a solution to the repository problem in the meantime...

klees:
I added another objective for to get a component repository during the build phase: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Services/Component/classes/Setup/class.ilStaticComponentRepositoryExistsObjective.php

chfsx:
Thanks @klees , I updated the PR accordingly

kergomard (approved):
Thank you very much @chfsx !
This looks good to me. It would be nice to fix the Copyright headers, but that's a trivial point ;-) .
Best,
@kergomard

kergomard:
Thank you very much @chfsx !

PR #6098, opened 2023-07-05
UI dropdown equal btn and link padding 37506

catenglaender:
https://mantis.ilias.de/view.php?id=37506

# Issue

Buttons in dropdown have different paddings than links.

image

# Changes

* btn-links and a tags are now styled the exact same
* deleted some unnecessary overrides

image


Amstutz:
thx a lot!

PR #6101, opened 2023-07-05
ilStartUpGUI / TOS: Valid logout URL

lscharmer:
Fix Mantis bug: https://mantis.ilias.de/view.php?id=37602

The logout URL works only with a valid CSRF token. To centralize the creation of a valid logout URL this PR adds a new static method `logoutUrl` to the `ilStartUpGUI`.

Additionally this PR replaces direct usages of `logout.php` in `ilStartUpGUI` and `Services/TermsOfService` (fixing the withdrawl process) with a call to `ilStartUpGUI::logoutUrl()`, as the CSRF token is missing from these links and will therefore not work.

The creation of a valid logout URL was taken from `UserMetaBarProvider::getMetaBarItems()` which is also replaced with a call to `ilStartUpGUI::logoutUrl()`.


mjansenDatabay:
@kergomard There is a tiny change in the `User` component, maybe you could have a look.

kergomard (approved):
Thank you very much for the improvements @lscharmer and @mjansenDatabay . Changes in User are fine with me.

mjansenDatabay (approved):

mjansenDatabay:
@pascalseeland I merge this, because the changes are quite trivial.

PR #6103, opened 2023-07-05
`Services/User`: `ilSetting` wrong argument type (trunk)

lscharmer:
Fix Mantis bug: https://mantis.ilias.de/view.php?id=37602

Fix wrong argument type for `ilSetting::get` for the Terms of Service withdrawl process.

PR #6104, opened 2023-07-05
`Services/User`: `ilSetting` wrong argument type (ILIAS8)

lscharmer:
Sister PR to #6103.
As the `ilPersonalProfileGUI` does not contain `declare(strict_types=1);` in ILIAS8, this does not throw an error but it is still wrong.

### DIFF
--- a/Services/User/Profile/classes/class.ilPersonalProfileGUI.php
+++ b/Services/User/Profile/classes/class.ilPersonalProfileGUI.php
@@ -346,7 +346,7 @@ class ilPersonalProfileGUI
             ($this->user->getAuthMode() === 'default' && $defaultAuth == ilAuthUtils::AUTH_LDAP)
         ) {
             $withdrawalType = 2;
-        } elseif ($this->setting->get('tos_withdrawal_usr_deletion', false)) {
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
@kergomard Could you have a quick look here? Thx in advance.

kergomard:
Hi @lscharmer and @mjansenDatabay 
Am I wrong in assuming that '0' would be more correct? I would then suggest to change this to:
`$this->setting->get('tos_withdrawal_usr_deletion', '') === '0'`
This would make this even clearer, as far as I'm concerned, ...but maybe I'm overlooking something?

Thanks for this improvement and best,
@kergomard

lscharmer:
> Hi @lscharmer and @mjansenDatabay Am I wrong in assuming that '0' would be more correct? I would then suggest to change this to: `$this->setting->get('tos_withdrawal_usr_deletion', '') === '0'` This would make this even clearer, as far as I'm concerned, ...but maybe I'm overlooking something?
> 
> Thanks for this improvement and best, @kergomard

Hi @kergomard
the PR's are adjusted accordingly.

Best regards
@lscharmer

kergomard:
Thank you very much @lscharmer ! I merge the two PRs.

PR #6106, opened 2023-07-05
9/UI/item media object layout 37599

catenglaender:
Mantis: https://mantis.ilias.de/view.php?id=37599

# Issue

Since the Bootstrap media object has been removed, some side by side positioning of images and icons no longer works as expected.

image

# Change

I suggest adding the media-object from Bootstrap again, but clearly mark it as not to be used in the future. Unfortunately, it is currently used in a few legacy components as well as in the modern UI item.

image


Amstutz:
thx a lot!

PR #6107, opened 2023-07-05
UI: buttons are labeled, not named or captioned

klees:
Hi everybody,

this tries to rectify some wording issue. On the one hand, this is a rather concrete change. On the other hand this is exemplary. So I try to explain a little why we should do this AFAIC.

So, there are three words that are involved here:

* **[caption](https://dictionary.cambridge.org/dictionary/english/caption)**: "a short piece of text under a picture in a book or article that describes the picture or explains what the people in it are doing or saying"
* **[label](https://dictionary.cambridge.org/dictionary/english/label)**: "a piece of paper or other material that gives you information about the object it is attached to"
* **[name](https://dictionary.cambridge.org/dictionary/english/name)**: "the word or words that a person, thing, or place is known by"

The thing we want to set here is not a "caption", because it does not describe or explain what we see in another piece of media. It is no "name", because the button will still be known by its name "submit button". (Also, it is no [title](https://dictionary.cambridge.org/dictionary/english/title)). It is a label, because it gives information (*What will happen exactly?*) about the submit button. Also, I think, this is how the general programming and internet community would call this thingy =)

Lets try to get this right an be consistent in the future when it comes to "labels", "names", "caption" and "titles".

Thanks =)

### DIFF
--- a/Modules/File/classes/class.ilObjFileGUI.php
+++ b/Modules/File/classes/class.ilObjFileGUI.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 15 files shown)


Amstutz:
HI @klees 

thx for the housekeeping in this matter. Naming is important, therefore highly appreciated. We also believe to be a good opportunity that this gets a call-out and short discussion during the upcoming JF.

Thx
@Amstutz

matthiaskunkel:
Jour Fixe, 10 JUL 2023 : Thanks for clarifying the wording with the PR. We highly appreciate this suggestion and accept the PR for trunk.

PR #6111, opened 2023-07-06
Docu: Added Data Tables How-To

acgruber:
As announced at the DevConf and in the [Jour Fixe](https://docu.ilias.de/goto_docu_wiki_wpage_7950_1357.html), DevGuide pages classified as how-tos are migrated to .md-files in GitHub and embedded in the current DevGuide LM. This way we can ensure improved accessibility and maintenance. Any update of the content by the responsible maintainer is highly appreciated.

### DIFF
--- a/Services/Table/use-data-tables.md
+++ b/Services/Table/use-data-tables.md
@@ -1,0 +1,267 @@
+# Data Tables
+
+Data tables should be implemented using the class `ilTable2GUI` provided by the Table Service. You usually need to extend this class and overwrite certain of its methods.
+
+```php
+// Calling GUI class
+class ilFooGUI
+{
+	[...]
+ 
+	/**
+	* Show a data table
+	*/
+	function showSomeDataList()
+	{
+		global $tpl;
+ 
+		include_once("./.../classes/class.ilMyTableGUI.php");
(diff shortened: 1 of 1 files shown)


klees:
@ILIAS-eLearning/technical-board As we are currently in the process of removing the tables that this is about, I would not include this in our docu anymore. Opinions?

mjansenDatabay:
@klees I agree!

dsstrassner:
But maybe, people need this for local developments for ILIAS 8 since the new data table is only in ILIAS 9 and above?

kergomard:
@dsstrassner : This is trunk only anyway, so I agree with @klees and @mjansenDatabay that we should remove this documentation and close this PR.

dsstrassner:
But, will this .md not integrated as external source for the LM and removed from the learning module as original content? Or will it only be removed, if this PR is merged?

acgruber:
I wouldn't remove content if it hasn't found a new place, but yes, the goal actually is to not have any original content in the LM and switch to .md only. Would a note - 'until ILIAS 8' or sth - help?

acgruber:
Update: @matthiaskunkel and I have discussed this and we're going to leave the content where it is for now. So I'm closing the PR. Thanks for taking your time to review.

PR #6112, opened 2023-07-06
UI/DataTable: async actions with tokens from URLBuilder

nhaagen:
Dear all,
last comment on https://github.com/ILIAS-eLearning/ILIAS/pull/5985 was describing a unification of actions and an idea of passing parameters to a modal-layer from a selection of DataTable rows.

This is a proposed implementation.

Screenshot from 2023-07-05 17-06-27

There are also some minor changes to the URLBuilder to make using it a little more comfy.


chfsx:
I am actually quite sure that I can solve my current problems with the DataTable via this PR :-) 

But i have now tried for quite a long time to integrate this PR into my current developments. I have tried left, right, straight ahead and back, unfortunately without success. I think a README with or other examples would be very helpful here. I can see the examples, but i can't get any further with them. Do you want to look at an implementation together via Discord? I think that would be helpful for a final version....

And since I know that you, @nhaagen,  is on the line, I know that we will have an awesome solution in no time :-)

Amstutz:
@nhaagen , @klees and @Amstutz discussed the public interface change in a short meeting on discord. We believe this to be the right way forward. However, we also agree the current solution does not specify the possible content, that can be rendered back in the async call, closely enough. Best would be a defined Modal Content, that can be rendered into the modals structure. However, this is not yet available yet. We should definitely work on providing this soon. FTM, we would, however, restrict this to Message Boxes and Interruptive Modals to be opened directly by the data table. 

@nhaagen : Therefore, please:

- [x] Content restriction rule: Please add a respective rule asking for Message Boxes or Interruptive Modals to be rendered in the async call.
- [x]  Roadmap: Please add an entry to the Roadmap, that describes the need for a Modal Content component to be used for Interruptive and Maybe also Round Trip Modals

Thx a lot

matthiaskunkel:
Jour Fixe, 10 JUL 2023 : We highly appreciate this suggestion and accept the PR for trunk.

thibsy (changes requested):
Hi @nhaagen

Thank you for your contribution to the UI components!

Since the public interface has already been agreed upon, I will now only do a technical review. After this, we can merge this into trunk.

I will use the letter `I` and `C` as an abbreviation for the `ILIAS\UI\Implementation\Component\Table` and `ILIAS\UI\Component\Table` namespace when referring to PHP classes.

- [ ] `tpl.datatabe.html`: what is the reason for the newly added HTML to create yet another kind of modal? IMO it would be better to reuse the `Roundtrip` modal without action buttons (except close) here. The client-side API could be extended by a `replaceContent()` function to set new content, this would already move towards the separation of modal and content. Doing so also brings other nice side-effects, like injecting signals to the data-table if necessary and getting rid of the custom `click` event-listener in `DataTable.asyncAction()` to close the modal.
- [x] Backdrop: you explicitly disable the `backdrop` property when showing an asynchronously rendered modal. Is this really something we should do? IMO we should try to be as consistent as possible with our components. If this really is something we want, we should achieve this with CSS, not JavaScript.

- [x] JavaScript bundle: I know you merely adjusted the files here, therefore this is only a suggestion: I think it makes sense to work with one bundled file per bundle, the `dist` directory may contain different version of such files, but ultimately only one should need to be loaded. This reduces network traffic and simplifies the logic that comes along with it. I therefore suggest to merge the existing `*table.js` files into one `table.js` file that gets bundled.
- [ ] `I\Action\Action` constants: I haven't quite made up my mind about how we should handle coupling between PHP and JavaScript like this, therefore I only added this as a suggestion: I think injecting values into JavaScript code, where they are used as constants to access object properties, makes the code unnecessarily complicated. To make future refactorings less invasive and keep property-names like this as close to the JavaScript as possible, I suggest to move the constants into the `I\Renderer` and mirror them in the `datatable.class.js`, so both sides can document their coupling. Afterwards you can remove the `DataTable.#signalConstants` property and use actual constants.
- [ ] `DataTable.asyncAction()` I: this method can and should be written without using jQuery. However, I am currently working on a proper async-renderer, which handles entire calls like this, so it wouldn't make sense to refactor this now. For future reference though, please use the [`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) API for asynchronous requests.

- [x] `datatable.js`: please don't use ESLint configuration comments, as also documented in the new [JavaScript code-style](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js-coding-style.md#3-eslint-configuration-comments). Using comments like this hints that something may be wrong with the JavaScript code, which is the case in this file: you are actually using two undefined global variables `$` and `il`, which must be considered as externals to the bundle. Otherwise, Rollup will not inject them properly into the IIFE that is ultimately generated. You need to introduce the values in the Rollup configuration and import the variables with an according statement. You can have a look at the [bundling document](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js-bundling.md#using-globals) where this is also described. Please note that `il` will not be considered as undefined by ESLint currently, I'll have to look into this later :).
- [x] `URLBuilder`(s): I totally forgot about array destructuring, nice improvement! Please implement the same changes for the JavaScript `URLBuilder`, since we want both backend and client-side builders to be (almost) identical. Btw, you can use the shorthand `[$a, $b] = [...]` for destructuring instead of `list()` if the array is not associative. This also works in JavaScript: `const [a, b] = [...];`.
- [x] `DataTable.asyncAction()` II: the `responseContainer` and `responseContent` variables seem like something we should already retrieve in the constructor. Otherwise these elements might end up being searched in the DOM multiple times for no reason.
- [x] `DataTable.asyncAction()` III: you are using jQuery directly here, to show the asynchronously rendered modal. Please use the existing client-side API to show the provided modal instead: `il.UI.modal.showModal(modalId, {}, { id: modalId });` Ideally, the namespace would be injected.
- [x] Examples: as described in `C\Factory::action()`, async actions must always render either a message-box or an interruptive modal. However, there are some examples which do not comply with this rule and e.g. render a characteristic-value listing. Please check your examples and only render either of those options.

Kind regards,
@thibsy

nhaagen:
Hi @thibsy,
again, thanks a lot for your review.
I incorporated the changes and there are some lessons to be learned (in regards to other components) in there.
E.g., I introduced dirty-chai in the package-json, so the linter will accept our tests ;)

> [ ] tpl.datatable.html

I'm not too happy with the modals (as they are) in general; using the factory means having information beforehand
that I retrieve from the response, so the direction of building them is somehow the other way.
I expect a major improvement by separating content and modal-wrapper, and will be happy to adjust the table accordingly.
For now, repeating the template just seemed a lot easier. 
I got rid of the clumsy click-handlers, though, also restoring the backdrop.

> [ ] I\Action\Action constants

I do not remember from whom, but I was explicitly asked to do that ;)
Either way is fine for me, although I understand the wish to only change one place
and do not have some "magic protocol".

> [ ] DataTable.asyncAction(), jquery

I'll keep that in mind. However, the modals are constructed via jQuery - something that imo should change, too.

Best regards,
Nils

thibsy (changes requested):
Hi @nhaagen

Thx a lot for incorporating the changes so quickly!

- [x] `tpl.datatabe.html`: I don't think copying the HTML is easier in the long run. This just adds another location which needs refactoring when the time comes, which may as well be overlooked. Yes, you need information beforehand if you are using the factory, but since we are pre-rendering an empty modal this shouldn't matter. I also found an issue which I think can be solved by reusing the existing modals, where additional JavaScript of asynchronously rendered modals will not be run. But, if this leads to some heavy changes again, we can still implement this in the 2nd iteration. However, maybe a note in the new roadmap entry would help to not forget this location when revisiting the modals.
- [ ]  `dirty-chai`: could you elaborate on the concrete issue you had when writing the JS unit tests in combination with the new linter configuration, and how the new dependency solves this? Maybe we can find another way instead of adding yet another library.

- [x] `I\Action\Action` constants: I can see the reasoning behind this, because you have only one value which needs changing in case of refactoring, yet this makes the JavaScript code a bit messy. You can leave this for now, since there is no rule against this (yet) and this is currently only a personal preference.

- [x] `DataTable.asyncAction()`: the way you currently render the retrieved HTML (`HTMLElement.innerHTML = ...`) does not run provided JavaScript code. If you decide to use existing modals, I think this can be solved by using `Modal::withAsyncRenderingUrl()` in combination with `il.UI.core.replaceContent()` on client-side. Otherwise, you probably need to use something like [`jQuery.parseHTML()`](https://api.jquery.com/jquery.parsehtml).

Kind regards,
@thibsy

nhaagen:
Hi @thibsy ,
I injected the JS and its evaluation; more importantly, I added to the roadmap to revert this again ;)
There is a related issue with consecutive forms and error messages in Modals with the StudyProgramme, and I really feel the need to get at the Modals soon. 

>[ ]  dirty-chai:

With a test-syntax of `expect(URLBuilder).to.not.be.undefined ` the linter will fail:

pre-commit: 
 - \CaptainHook\App\Hook\PHP\Action\Linting                          : done
 - libs/composer/vendor/friendsofphp/php-cs-fixer/php-cs-fixer fi... : skipped
 - npx eslint --fix {$STAGED_FILES|of-type:js}                       : failed

failed to execute: npx eslint --fix tests/UI/Client/URLBuilder.test.js

/home/nhaagen/projects/ilias/trunk/volumes/ilias/tests/UI/Client/URLBuilder.test.js
  30:5  error  Expected an assignment or function call and instead saw an expression  no-unused-expressions


With a test-syntax of  `expect(URLBuilder).to.not.be.undefined() ` the linter passes, but this is not a valid chai-call:
  1) URLBuilder and URLBuilderToken are available
       URLBuilder:
     TypeError: expect(...).to.not.be.undefined is not a function
      at Context.<anonymous> (tests/UI/Client/URLBuilder.test.js:30:34)
      at processImmediate (internal/timers.js:461:21)

dirty-chai allows for the brackets; this seemed like a good solutions for the tests.

Best regards,
Nils

thibsy (changes requested):
Hi @nhaagen

Thx for your explanation. The changes look good to me now and I think this is almost ready for a merge. There is still one thing I don't feel quite comfortable with:

- [x] `dirty-chai`: using this library means to use `require`, which is [discouraged by Airbnb's JavaScript code-style](https://github.com/airbnb/javascript#modules--use-them). I think we might need to discuss this in more detail and this would make more sense in a separate PR, since this does not immediately affect the UI components. I therefore ask to revert these changes and maybe use `expect().to.be.an(undefined)` to make ESLint happy for now. If this does not work you may exceptionally use ESLint configuration comments to disable the linter there.

Thank you for bearing with me :)

Kind regards,
@thibsy

nhaagen:
Hey @thibsy,
dirty-chai is gone again.
Not for this, but others, we will have to rework some JS-tests, I guess.
Best,
Nils

thibsy (approved):

thibsy:
Hi @nhaagen 

Thx again for your effort!

I agree, we will probably need to refactor some of the JS unit tests along the way. Nonetheless, the PR is good to go now, therefore I squash-merged this into trunk.

Kind regards,
@thibsy

PR #6113, opened 2023-07-07
[FIX] UI: JavaScript global `il` namespace

thibsy:
Hi all,

Because of an issue with the copyright license headers in #5938 I have opened this PR to fix the Rollup.js configuration, which now includes the header for generated files as well. To be consistent I have also renamed the output file to `core.js`.

This brought my attention again to another issue, initially fixed @nhaagen in #5803, which used the `banner` attribute to initialize the global `il` namespace if necessary. I realized later on, when implementing the JavaScript code-style for ILIAS, that we have misused this property here. According to Rollup.js it is not intended for injecting further JavaScript code. I have therefore fixed this issue in another way, by initializing it in the HTML template of the `Page\Standard` directly. This way, bundles don't need to worry about initializing the namespace anymore and it felt like the safest place to add it.

Kind regards,
@thibsy


klees (approved):
LGTM, love it.

PR #6114, opened 2023-07-07
LUI, replaces ilLinkButtons of Services Calendar, Search, AdvancedMet…

chlulei:
LUI, replaces ilLinkButtons of Services Calendar, Search, AdvancedMetadata


smeyer-ilias (approved):
Thanks, @chlulei

PR #6116, opened 2023-07-07
28240 - Add possibility to always offer custom meta data set in study…

lukastocker:
… programmes.

[Mantis Ticket 28240](https://mantis.ilias.de/view.php?id=28240)

PR #6121, opened 2023-07-10
AMET: Add possibility to always offer custom meta data set in organ…

lukastocker:
…isation units.

[Mantis Ticket 28239](https://mantis.ilias.de/view.php?id=28239)

**Note:**
The values "12","33","34" for the contexts in "class.ilAdvancedMDPermissionHelper.php" were chosen because the values "11", "31" and "32" are already assigned in PR [6116](https://github.com/ILIAS-eLearning/ILIAS/pull/6116).


klees:
Hi @smeyer-ilias,

these are completely located in Services/AdvancedMetadata, thus I assigned you. Please ping me if there is anything we can do here.

Best wishes!

smeyer-ilias:
Thanks, I 've cherry-picked this fix to R8 and trunk and used different IDs since there were conflicts in R9.

PR #6122, opened 2023-07-10
PRG: remove 'Add' button if membership automation search result is em…

lukastocker:
…pty. (#29857)

https://mantis.ilias.de/view.php?id=29857

### DIFF
--- a/Modules/StudyProgramme/classes/class.ilObjStudyProgrammeAutoMembershipsGUI.php
+++ b/Modules/StudyProgramme/classes/class.ilObjStudyProgrammeAutoMembershipsGUI.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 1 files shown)


nhaagen (changes requested):
Hi @lukastocker ,
thanks a bunch for tackling this issue.
Your changes are working just fine and the save-button is gone, however, I feel that the anticipated behavior of the ticket is slightly different. From a user's perspective, I embarked with the intention of setting up a, e.g., group as auto-source.
Then I had a typo in my request, but still - I want to add this source. 

- [ ] Please return to the form in case of no entries, 
- [ ] add a message that the search was not successful
- [x] and still hide the save-button

Best regards,
Nils

lukastocker:
> Hi @lukastocker , thanks a bunch for tackling this issue. Your changes are working just fine and the save-button is gone, however, I feel that the anticipated behavior of the ticket is slightly different. From a user's perspective, I embarked with the intention of setting up a, e.g., group as auto-source. Then I had a typo in my request, but still - I want to add this source.
> 
>     * [ ]  Please return to the form in case of no entries,
> 
>     * [ ]  add a message that the search was not successful
> 
>     * [x]  and still hide the save-button
> 
> 
> Best regards, Nils

Important Note:
I had a conversation with Nils Hagen and this implementation differs too much from the needed result. The current modal (UI) can't do what we want and need. Nils Hagen rewrote the modal system - I am currently waiting for it to get merged. With this new modal system it's possible to adjust this PR accordingly.

klees:
@nhaagen @lukastocker I understand correctly: We need to drop this for the moment? I'll close, then... Please reopen if I missunderstood. Thanks!

PR #6124, opened 2023-07-10
TA fix for 37028

okaufman:
Thsi will fix https://mantis.ilias.de/view.php?id=37028


kergomard:
Thank you very much @okaufman . Will merge and adapt it for ILIAS8 and trunk.

PR #6125, opened 2023-07-10
PRG: filter user by permissions if 'mail to assigned user' is calle…

lukastocker:
…d. (#31413)

https://mantis.ilias.de/view.php?id=31413

### DIFF
--- a/Modules/StudyProgramme/classes/class.ilObjStudyProgrammeMembersGUI.php
+++ b/Modules/StudyProgramme/classes/class.ilObjStudyProgrammeMembersGUI.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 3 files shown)


nhaagen (changes requested):
Hi @lukastocker ,
At first, I stumbled about the specific dic and "$this->object", since $this->obect may be null.
Are we sure, there is no path this might be the case when calling ilStudyProgrammeMailMemberSearchGUI?
Otherwise, it's a good idea to use the specific dic and thus hook in to the permission helper.

However, doing so: 
Is there a reason (real question) to re-loop over the users in in getProcessData?
The array is beeing built for the first time, it feels easier to not add a user in the first place than to filter them out later.
You probably need getUserIdsSusceptibleTo rather than filterUserIds, or getUserIdsInPrgAccessibleForOperation to get a positive list.

I also tried running the code - sadly, I directly ran into a whoops. Please see the mantis-ticket for the report.

Best regard,
Nils

nhaagen (approved):

PR #6126, opened 2023-07-10
TA fix for 37031

okaufman:
This will fix https://mantis.ilias.de/view.php?id=37031
It is basically the same fix as in https://github.com/ILIAS-eLearning/ILIAS/pull/6077 

I had to adjust the Tests so that they will run through.


kergomard:
The fix you copied overlooked a sideeffect: Enter doesn't work in textareas anymore. I will still merge this and apply the fix i added here fc35b183bc .
Thanks again and best,
@kergomard

PR #6128, opened 2023-07-10
Refinery: Fix `mb_convert_encoding` in `EstimatedReadingTime`

mjansenDatabay:
This PR suggests a replacement for the (in PHP 8.2) deprecated `mb_convert_encoding`.

### DIFF
--- a/src/Refinery/String/EstimatedReadingTime.php
+++ b/src/Refinery/String/EstimatedReadingTime.php
@@ -53,7 +53,7 @@ class EstimatedReadingTime implements Transformation
     public function transform($from): int
     {
         if (!is_string($from)) {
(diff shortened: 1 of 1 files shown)


klees:
Explanation was provided by @mjansenDatabay:

This fixes an issue with `mb_convert_encoding` introduced by PHP 8.2. The function is used to fix problems that page-editor-content might cause when moved to dom-documents.

`HTML_ENTITIES` could be used as "pseudo-type" (see [here](https://www.php.net/manual/en/mbstring.supported-encodings.php)) for HTML inputs. This doesn't work anymore with PHP 8.2.

This uses `mb_encode_numericentity` instead, where numeric representations of entities are used. This makes it confusing, since encodings need to be passed as "maps" and people wonder "what's this?". See [here](https://3v4l.org/0UMGv) for an example.

PR #6129, opened 2023-07-10
PRG/LUI: remove submit button in toolbar

nhaagen:
as part of the LUI-removal: also remove the submit-button in toolbar


klees (approved):
I hate this, but I don't know any better. We will stumble upon this again when we tackle the toolbar, so this is ok for now.

PR #6130, opened 2023-07-10
9/UI/itscss clean up and bootstrap issue guide

catenglaender:
Some clean up and documentation for the ITCSS style code after the Bootstrap 3 removal:

* Documented the process for fixing an issue caused by the removal of Bootstrap 3
* updated SCSS Guidelines with lessons learned from coding first ILIAS 9 components
* deleted unused scss files
* moved files out of legacy folders that are actually okay to use in future style code
* renamed some tools (e.g. vendor-prefixes is now _tool_browser-prefixes)
* clarified some comments in the ITCSS index files
* delos.css.map has been deleted and now excluded via .gitignore as it continously brought merge conflicts and holds local server paths anyway - so it should only be created by the designer locally (or be generated by the ILIAS instance in the future).

This should make it a little easier to keep the overview and to contribute to the ITCSS SCSS refactoring and Bootstrap removal project.


Amstutz:
Thx a lot!

PR #6132, opened 2023-07-11
Docu: Added Time Zones How-To

acgruber:
As announced at the DevConf and in the [Jour Fixe](https://docu.ilias.de/goto_docu_wiki_wpage_7950_1357.html), DevGuide pages classified as how-tos are migrated to .md-files in GitHub and embedded in the current DevGuide LM. This way we can ensure improved accessibility and maintenance. Any update of the content by the responsible maintainer is highly appreciated.

_Note: This content probably requires a thorough update. Thanks in advance!_


schmitz-ilias:
Thanks @acgruber!

PR #6142, opened 2023-07-12
Search: Fix Double Encoding of Search String

kergomard:
See: https://mantis.ilias.de/view.php?id=37537


smeyer-ilias:
Thanks for fixing this. 
I close this report, since it inspired me to make a commit, which fixes a similar issue for UDF selects.

9dba2f9c688e541cd9a5dae2f16efda5b3918b58

PR #6145, opened 2023-07-12
Test: Fix Highlighting of Ordering Elements

kergomard:
See: https://mantis.ilias.de/view.php?id=36735

This is a port of a css-fix to trunk. I actually do not like this, because:
- We cannot change the naming, without change quite a few elements outside of the test, as far as I can see.
- We have a hard coded color, but it is in the context of another hard coded color (background of the element) defined in content.css of the COPageEditor.

Let me know, if you have good ideas for improvements. I'm a happy taker!

Thanks and best,
@kergomard


dsstrassner:
Hi @Amstutz, I have me also assigned because of the TechSquad statistic. This PR and the code are all yours ;-)

Amstutz:
Hi @kergomard 

I do not understand, why you used the hard-coded color here. Why not use the $il-main-color variable? This will be compiled into CSS as part of delos.css and so should not be an issue, right? Am I missing something?

For the magic 9px possibly $il-padding-base. This is not a hard requirement. With spacing / borders, we haven't found a very fixed system yet. 

lg
Timon

kergomard:
Thank you very much for the feedback @Amstutz . As discussed I moved both hard-coded values to variables. Let us see if somebody complains.
Best,
@kergomard

Amstutz:
I talked about this with @kergomard a few days ago. We are both not sure how to handle this with content style related code. Merged as is.

dsstrassner:
Surely, someone will complain, as the test / test questions are involved. ;)
Thanks @Amstutz & @kergomard :-)

PR #6031, opened 2023-06-21
Init/Login/Migration/Codes: Use UI framework for startup forms

mjansenDatabay:
This PR introduces the UI framework forms to the `ilStartupGUI`.

Furthermore I did a little bit of "boy-scouting".

See: https://mantis.ilias.de/view.php?id=33816

**!!!Important!!!:** The `Code` and `Account Migration` forms do not provide a `Cancel` button anymore, just because the UI forms don't support further buttons.

PR #6149, opened 2023-07-13
UI/Init: Add autocomplete="off" for password fields

tbongers-cat:
As a result of a penetration test for one of our customers, it was critized that the ILIAS password fields are not setting the attribute `autocomplete="off"`. This attribute is meant to tell the browser to not show any (locally stored) autocompletion options when entering data, to avoid the security risk e.g. for computers in a shared environment.

From our research, this attribute is currently **ignored** by a lot of modern browsers, especially when they offer some sort of internal password manager for their users. There is discussion if the use of a password manager rectifies doing this, but it seems to be a common usage today.

Nevertheless, we think it won't hurt to have this attribute in our password input fields, if not only to send a signal to any browser displaying the site. And as an added bonus, we might satisfy at least one point of future penetration tests =)

This PR enables the autocomplete attribute for the main login form and adds it to the UI password input. If you know any other places where this should be added, please let me know and I'll gladly add them to the PR.

### DIFF
--- a/Services/Init/classes/class.ilStartUpGUI.php
+++ b/Services/Init/classes/class.ilStartUpGUI.php
@@ -528,7 +528,7 @@ class ilStartUpGUI implements ilCtrlBaseClassInterface, ilCtrlSecurityInterface
         $pi->setRetype(false);
         $pi->setSkipSyntaxCheck(true);
         $pi->setSize(20);
-        $pi->setDisableHtmlAutoComplete(false);
+        $pi->setDisableHtmlAutoComplete(true);
         $pi->setRequired(true);
         $form->addItem($pi);
 
--- a/src/UI/Component/Input/Field/Factory.php
+++ b/src/UI/Component/Input/Field/Factory.php
@@ -320,6 +320,12 @@ interface Factory
      * rules:
      *   usage:
      *     1: Password Input MUST be used for passwords.
+     *   composition:
+     *     1: >
+     *        The input MUST always be rendered with the attribute autocomplete="off".
+     *        This advises browsers to NOT autofill the input field with cached passwords
+     *        and avoids potential exposure of confidential data, especially in
+     *        shared environments.
      *   interaction:
      *     1: >
      *         Password Input SHOULD NOT limit the number of characters.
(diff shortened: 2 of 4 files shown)


Amstutz:
Hi @tbongers-cat

thx a lot for your PR and solving this issue. The implementation is fine. 

Please Change:

- [ ] Composition Rule: Can you add a composition rule, that autocomplete MUST be set to off, along with the reason. Like this, we will not miss it in further refactorings.

thx a lot

Services/Init/classes/class.ilStartUpGUI.php
```diff
@@ -528,7 +528,7 @@ protected function initStandardLoginForm(): ilPropertyFormGUI
         $pi->setRetype(false);
         $pi->setSkipSyntaxCheck(true);
         $pi->setSize(20);
-        $pi->setDisableHtmlAutoComplete(false);
+        $pi->setDisableHtmlAutoComplete(true);
```
mjansenDatabay, line 531: I stumbled upon this while working on #6031. This was disabled here based on a JF decision some years ago. Maybe we'll have to discuss this again if we like to change the behaviour.

tbongers-cat:
Hi @Amstutz
I have added the composition rule as requested. Please have a look if the wording is ok.

Amstutz:
Hi @tbongers-cat 

Thx for adding the rule. Wording is find. However, you added the rule to the description section. However, rules belong into their own section, see: https://github.com/ILIAS-eLearning/ILIAS/tree/release_8/src/UI#how-to-implement-a-component .

Further did you see the comment of @mjansenDatabay ? We best also discuss/decide this shortly during the next JF.

thx

tbongers-cat:
@Amstutz 
I have moved the rule to the correct section now, sorry.

I'm fine with discussing this at the JF. From my code research during implementation, the option has been set to `false` for years, but I was not aware that this was based on a JF decision and would like to hear the reasons for it.

matthiaskunkel:
Jour Fixe, 07 AUG 2023: We highly appreciate the PR and accept it for ILIAS 9. Please merge into trunk.

Services/Init/classes/class.ilStartUpGUI.php
dsstrassner: > I stumbled upon this while working on #6031. This was disabled here based on a JF decision some years ago. Maybe we'll have to discuss this again if we like to change the behaviour.

To document the findings of @mjansenDatabay: [Decision on JourFixe-2014-04-14](https://docu.ilias.de/goto_docu_wiki_wpage_3106_1357.html)

Amstutz:
Thx a lot, I merge this, see also comment in: https://mantis.ilias.de/view.php?id=37361#c95548

PR #6150, opened 2023-07-13
ModeInfo/UI Demo: fix example for modeinfo

nhaagen:
fixes the example for ModeInfo; also, the button was _very_ unobstrusive.

### DIFF
--- a/src/UI/Implementation/Component/Layout/Page/Renderer.php
+++ b/src/UI/Implementation/Component/Layout/Page/Renderer.php
@@ -166,9 +166,6 @@ class Renderer extends AbstractComponentRenderer
         }
 
         if ($for_ui_demo) {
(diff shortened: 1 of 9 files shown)


Amstutz:
Thx for improving the example.

what exactly was broken? note that tests are failing atm.

nhaagen:
Links to the fullscreen were not working (script-url was changed), and the page was rendering without CSS/JS.
Checks failed due to the horrible $_GET-access. Now it's something with the footer?!

nhaagen:
Sorry about the fuss; I found some more issues, basically all examples rendering a full page. 
This now should fix links and responses by using refinery, requestwrapper and ilCtrl over script-names and server-globals.

Amstutz:
Looking much better know, thx a lot for improving this!

PR #6151, opened 2023-07-13
Allow Different Sources for Images Depending on Actual Size on Screen

kergomard:
This is my first try at expanding the UI-Framework, so please be patient with me:
This PR would introduce a way to add [`srcset`](https://html.spec.whatwg.org/multipage/images.html#srcset-attribute) and [`sizes`](https://html.spec.whatwg.org/multipage/images.html#sizes-attributes) attributes to images to allow us to offer different image-sizes to the browsers. This PR comes out of the work on moving Tile-Images to the IRSS and taking maximum advantage of the flavors offered there.

Screenshots are not very useful, as there is actually nothing to see ;-), but this PR contains, besides the interface changes (7d787187422f9e7235fe0cf7dfe5fa9ca82c8bac), also the implementation (fe0635194aacb8fad9d712a41972f86bb76743f0), and an example(8cda90d034283f9a9a9a8d1e23b80fc7e76e786a). I will add the tests if and as soon as we agree on the interface.

If you think a synchronous exchange would be more fruitful, don't hesitate to contact me!

Looking forward to your feedback, thank you  very much, and best,
@kergomard

Addendum: I just used an image that already was in the example folder, ...and multiplied it. I'm pretty sure we don't want the images here, but am a little fuzzy on where to put them.


thibsy:
Hi @kergomard 

Thank you for contributing to the UI Components!

We haven't forgotten this PR, but our plate was currently quite full. This is definitely something I would like to discuss with @klees before giving a definite review, but from my point of view I already have some concerns about this:

- [ ] By looking at the examples, it feels like this feature heavily relies on consumers. I had lookup to image-source-sets myself, just to understand the example. It not only gives developers very much control about this attribute, but it also requires developers to know about image source-sets. I think this is an important feature nonetheless, I merely think we should take away some of this control and required knowledge. IMO the only concern should be to decide which "media" (viewport in px) should receive which "source" (path to the asset). Therefore I would much rather have code like this: `$image->withSourceSetMapping(['desktop' => '/path/to/large.png', 'mobile' => '/path/to/small.png']);` Doing so would also allow some consistency, otherwise we could easily end up with very different source-sets, because every developer has a different preference :).

Kind regards,
@thibsy

kergomard:
HI @thibsy 

Thank you very much for your feedback!

Just as a short reaction: I don't think we can actually simplify the use of this much.
If I'm understanding your suggestion correctly, this is actually more powerful, than you think: The browser can choose the image-file it needs, from the size it will be shown in. The problem there clearly is: It would only know the size it needs, after having rendered the whole page, so it would need to wait for the DOM to render and only then it could load the images. What you do with the `sizes`-attribute is give it information about how much place the image is going to use in the viewport and this can be done by using media-queries. As the browser knows its viewport, it can then instantly decide which image it is going to ask for. I would e.g. like to use this for images in cards, where the image-size can actually vary quite a bit, depending on the size of the viewport, as an image might only use a sixth of the space on a big screen, but the whole page on a small one, so there might actually be the need for a bigger picture on a smaller screen.

If it would make things easier, we could also schedule a call, when @klees  is back.

Thanks again and best,
@kergomard

klees (changes requested):
Hi everyone,

we just had a meeting with @kergomard, @Amstutz and @thibsy. We came to the following conclusions:

We would need to improve the interface regarding the way that additional sources are included. This is next to trivial and not the real problem.
The real problem is, that the people creating Image components won't be able to define helpful sizes selectors. The person creating the component won't know in which context the component is used, it even might be used in many different locations (as we made the UI components much composable and reusable). More crucially, even the component receiving the Image (such as a Card, Standard Item, ...) won't be able to set sizes correctly, because they also depend on some context etc. In general, sizes of components will depend on style code, hence won't be known in general when PHP code is executed. So we cannot set sizes on the server and we couldn't manage to find a helpful heuristic.
Hence, we need to rely on the client somehow to select a proper alternative source, hence JS.

@kergomard than had to leave, but the rest discussed a little further. We came to the conclusion that a good solution would look as such:

The source given to the Image via constructor or withSource will always be the src when the Image is rendered.
A new method withAdditionalHighResSource(string $source, int $min_width_in_pixel): Image should be introduced to define additional resources. The semantic is: If the rendered image is bigger than $min_width_in_pixel, use that $source. If more than one source could be used, take the one with the biggest $min_width_in_pixel.
Do this once, after the page is rendered, on the client side.
Advanced consumers could then use a low-res image as "source" and set the normal image as "additional source" with a super low $min_width_in_pixel.

I hope this suits the desired use cases.

Kind regards!

kergomard:
The interface has been updated according to the results outlined in the previous comment:

The implementation will follow asap.

Thanks and best,
@kergomard

Amstutz:
Hi @kergomard 

Thx a lot for adapting the proposal according to the Feedback. From our point of view, this is ready to be discussed at the upcoming JF. Note that the Review here only considered the public interface, the implementation will be reviewed separately.

- [x]  Image Composition Rule: Do we want to add a rule, that is, suggesting the usage of this for certain scenarios? Something like: "In case of highly varying sizes of the images display, multiple sources MAY/SHOULD be used to reduce the load on the web servers and speed up of the initial display."

Please add, or explain why not:
- [x]  Image Description: Please add one or two lines concerning this new option in the Composition section of the Image description. Otherwise, it might be missed by non-programmers.

Thx a lot!

kergomard:
Hi @Amstutz 

I completely missed this part!

I added some more information on composition in the description and a `SHOULD` rule in the composition section. I tried to follow the language used in other places, but I'm more than happy to amend, if it is not as expected.

Thank you very much and best,
@kergomard

matthiaskunkel:
Jour Fixe, 18 SEP 2023 : We highly appreciate this suggestion and accept the PR for trunk.

kergomard:
Hi @thibsy , @klees , @Amstutz 
Hi all

I now did a first implementation of the async fetcher for the additional high resolution images. While doing so I found a few possible issues/inconsistencies:

1. The CSS for a responsive image is a `mixin` that is only inserted in certain contexts, so the example is completely without function as long as I don't put it inside e.g. a Card/DeckOfCards. Is this intended and should I just leave the example this way?
2. Thinking about it, this feature only makes sense on Responsive Images. Standard Images just set their size depending on the image inside. The image might be resized, but, as far as I can see, only in one direction (smaller). Should we change something concerning the interface under these circumstances or am I missing something?
3. And then: I tried to write JS tests for this and I failed. Maybe somebody has a good idea? The only perceivable change of the javascript is a change of the source of the image. All else is completely internal and we shouldn't open it up. I was unable to find a good way to get a size of the image on the screen and then to retrieve a file through the Promise, that then changes the src, once the file is loaded.

Thanks and best,
@kergomard

klees:
Hi @kergomard,

from my perspective the fixed width image will vanish sometime soon, as they don't play along very well with a design system that should be responsive in general (ping @yvseiler and @catenglaender for their opinion). So for me the example and your other work is fine. 

Regarding your tests, I think you need to figure out what you should test exactly, which weight your code is pulling. Then you need to isolate that weight accordingly. For your code, the pulled weight seems to be: determine the correct source based on image size; load the image and assign it to the src. So, I guess, for the test: Pass a simple custom made object as `image` to `loadHighResImage`, where you fill properties as required for a certain test. Mock `#loadBestSource`. Check inputs to `#loadBestSource` and `src` on your custom image object.

Contact me on short note for further help if you like.

Kind regards!

kergomard:
Hi @thibsy , @klees , @Amstutz
Hi all

I've added JS tests for the Image.

Thank you very much @klees and @thibsy for the support!

Best,
@kergomard

thibsy (commented):
Hi @kergomard

I apologise for the delay on my side!

When reviewing I have noticed several issues with the current implementation, which needed refactoring. Since this would have led to some back and forth between us, I have decided to speed up the process and refactored it myself.

So instead of a review, I will point out the underlying issues and how I solved them:

- [ ] Example: The example wasn't quite working, mostly because responsive images don't work. I fixed this by using a `Card` component because this container treats responsive images correctly. I also added a layout to demonstrate different sizes being loaded for different widths.
- [ ] `HighResImageLoader`: I was glad to see you using classes, but unfortunately this was not the ideal case to do so. The image loader gained no benefit from being a `class`, which is why I transformed this into an ES6 module with a function export.
- [ ] `loadHighResImage()` I: This function compared the [`nodeName`](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeName) attribute of an element, which is not a good practice. `instanceof` should be used instead. I solved this by introducing another module which will retrieve the `` element of an image component so this function has not to worry about it.
- [ ] `determineBestSource()`: This function only worked with an ordered `Map`, which I fixed by introducing another condition to compare previous values. This takes away some required knowledge for other developers and consumers.
- [ ] `loadHighResImage()` II: I noticed that images were loaded twice. This is because you created a new `Image` to check if the source is available, but when updating the existing images `src` attribute, the image will be loaded again. I fixed this by cloning the image element, to keep properties alive, and replace it if another source is successfully loaded.
- [ ] Injection of source-mappings: I think we should avoid using `json_encode()` when possible to inject values into our JavaScript logic. I improved this by converting the array into a JavaScript compliant `Map` string, similar as the `URLBuilder` does.
- [ ] Implicit trigger: It felt unintuitive to know _when_ the images are loaded. By looking at the injected JavaScript code, one needs to know that `withAdditionalOnLoadCode()` fires on `jQuery.ready()`. I explicitly added a statement which registers the loading mechanism to `jQuery.load()`, which also fires at a later point when all JavaScript logic should be initialised.
- [ ] JavaScript unit-tests: I actually refactored the JavaScript logic to a degree, where it isn't quite necessary to use JSDom anymore.

Let me know if you disagree with anything. Maybe @klees wants to have another look, but IMO this is good to merge now.

Kind regards,
@thibsy

kergomard:
Thank you very much @thibsy ! The changes are a clear improvement and they help me learn and understand where we are heading!

Best,
@kergomard

klees (approved):

PR #6152, opened 2023-07-14
[FIX] renamed proposed directory to artifacts

chfsx:
Can we rename the directory for artefacts from `cache` to `artifacts`? for the following reasons:

- we try to make the names in the repo close to what exactly happens or is located there. in this case there are artifacts there. 
- `cache` sounds like you can just empty the directory and ILIAS will rebuild it itself, which is not the case here without a build process.


kergomard:
Hi @chfsx 
I was one of does pushing for `cache` and I would still vote against renaming it to `artifacts`:
- I think it is a convention that a folder containing the ephemeral parts of the code should be called `cache`. I do understand your second point, but I don't believe a `cache` always needs to be rebuilt automatically. Rebuilding manually, as in our case, is a valid strategy and doesn't invalided its properties as a cache.
- Right now, I'm unable to ascertain that there might not be other ephemeral content, that is not public, ie. doesn't belong in the folder `public`. So, I would like to offer a place for these potential things and thus this folder could contain more than what we call `artifacts`. 
- I would be all in favor of creating a folder inside called `artifacts`, thus increasing the verbosity of the naming.

Just my five cents, but I will clearly not start a religious war over this ;-).

Thank you very much and best,
@kergomard

klees:
Hi everyone,

I might be missing for the discussion on the JF this afternoon (dentists appointment...), so I post my five cents here. To me it mostly does not matter that much how this folder is named. I initially went with `artifacts`, then changed the name after @kergomard`s input. I would be happy to move back to `artifacts` again.

In general, I don't think that future changes to this directory name will have much impact on anyone, so if we find that we'll have other files that might fit into a directory like this, we could change the name quite easily, imo.

Kind regards!

pascalseeland:
I would vote for naming it artifacts. Cache is something which you can clear without any consequences, which you clearly cannot do here.

matthiaskunkel:
Jour Fixe, 24 JUL 2023 : We prefer the name 'artifacts' instead of cache, esp. for the reason mentioned by @pascalseeland. PR is highly appreciated and accepted for trunk.

PR #6155, opened 2023-07-14
0037633: T&A Administration Crash when "Activate Manual Scoring for" all questions are deactivated

kunzeus:
Ticket:
https://mantis.ilias.de/view.php?id=37633

Sympthom:
When setting a lot of inputs empty there have been errors

Problem:
unset post variables

Solution:
Add checks for the post variables and null coalescing
 
Tested:
8

Code compatible with:
ILIAS 8, Trunk

PR #6156, opened 2023-07-14
T&A 0037632: T&A Administration Crash when all "Available Question Types" are deactivated

kunzeus:
Ticket:
https://mantis.ilias.de/view.php?id=37632

Symptom:
Deactivation of all question types leads to a null error.

Solution:
add null check

broken on
7, 8, Trunk

Tested:
7, 8 

Code compatible with:
ILIAS 7, 8, Trunk


kunzeus:
closed changes will be included in: https://github.com/ILIAS-eLearning/ILIAS/pull/6155

PR #6159, opened 2023-07-14
Course: Remove Setting of Online Status

kergomard:
Hi @alex40724 

Could we maybe get rid of this? If not: Could you please explain to me, why this is needed, so that I can figure out a way to do this?

Thank you very much,
@kergomard


alex40724:
Imo this can be removed.

kergomard:
Thank you very much for the feedback @alex40724 . Then I merge.

PR #6161, opened 2023-07-17
LSO: add possibility to add LSO to favourites. (#36632)

lukastocker:
Mantis Ticket [36632](https://mantis.ilias.de/view.php?id=36632)


klees:
Picked to trunk.

PR #6164, opened 2023-07-17
UI Components: Refactor/Extend Toast Interfaces

iszmais:
This PR involves a proposal to extend the toast UI Component by 2 functionalities:
- vanish Timing
- delay Timing

This where already implemented within the dynamic implemetation process of the toast feature request which included the origin toast UI component (https://github.com/ILIAS-eLearning/ILIAS/pull/3411), but where never added to the interface itsself and are therefore not used within the actual presentation through the GS (https://github.com/ILIAS-eLearning/ILIAS/commit/5570810c79015b87000e36a5310c5411d5466a7d).

To keep a clean workflow this PR should preceed the establishment of any usage of the future within the GS.

### DIFF
--- a/src/UI/Component/Toast/Container.php
+++ b/src/UI/Component/Toast/Container.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -18,14 +16,12 @@ declare(strict_types=1);
  *
  *********************************************************************/
 
+declare(strict_types=1);
+
 namespace ILIAS\UI\Component\Toast;
 
 use ILIAS\UI\Component\Component;
 
-/**
- * Interface Container
- * @package ILIAS\UI\Component\Toast
- */
 interface Container extends Component
 {
     /**
--- a/src/UI/Component/Toast/Factory.php
+++ b/src/UI/Component/Toast/Factory.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -18,13 +16,12 @@ declare(strict_types=1);
  *
  *********************************************************************/
 
+declare(strict_types=1);
+
 namespace ILIAS\UI\Component\Toast;
 
 use ILIAS\UI\Component\Symbol\Icon\Icon;
 
-/**
- * This is how a factory for Toast looks like.
- */
 interface Factory
 {
     /**
--- a/src/UI/Component/Toast/Toast.php
+++ b/src/UI/Component/Toast/Toast.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -18,6 +16,8 @@ declare(strict_types=1);
  *
  *********************************************************************/
 
+declare(strict_types=1);
+
 namespace ILIAS\UI\Component\Toast;
 
 use ILIAS\UI\Component\Button\Shy;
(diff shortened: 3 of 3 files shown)


iszmais:
@klees could you make a decision if this should walk past another Jour Fixe befor merging?
Im would assume not as it is just a minor change, but if you think otherwise i can provide a presentation for such.

klees:
Hi @iszmais,

general rule is, that all changes in public interfaces need to pass the JF (see [here](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/UI/docu/rules.md#modification-of-existing-ui-components)). We may fast track this as UI-coordinators if there are good reasons to do so.

Is this time critical?

Best wishes!

iszmais:
@klees No this is not time critical.
 I will prepare a presentation to the JF.

klees:
Hi @iszmais,

my personal two cents: I'm not convinced that these timings are something that really belongs on the public interface of singular `Toast`s. Having these methods would imply that each and every toast may behave differently. Which IMO is detrimental for user experience, as it does not allow to build an expectation towards these timings.

Luckily, we have the `Container` for the `Toast`s. I would suggest to add the timings there.

Kind regards!

iszmais:
Thank you for your response @klees ,
This was actually the intend of this PR. The origin FR defined some implication in that direction but had no specified ruling for that, what caused this to be left out on the initial implementation.
After some rereview with @oliversamoila and in investigation of the origin FR we came to the conclusion that this would be indeed usefull for specific toast when the information they are carring passed a certain limit.

Amstutz:
Hi @iszmais

Please elaborate why this would be needed to be attached to the Toasts and not the container as @klees suggested. 

thx
@amstutz

iszmais:
Hi @Amstutz 

as said i assume this would only be usefull for specific toast.

E.g. in the current core situation the toast with the highest information density is the contact request with around 20 words to read and 3 different actions beside the default toast actions. This makes it really hard to select a proper reaction to the toast within the default 5 seconds, at least for "cozy" users. Therefore an increase of its presentation time would be a benefical improvement to the user experience.

On the other hand there are toast like Who-is-online or badge-recieved with only around 5 words an 1 or 0 neccessary actions required from the user. Those can be easily handled within 5 seconds and an increase of there presentation time would only contribute to an information overload on the UI.

Further this is would be an advancement for future development and encourages the usage of this new component to balance the information presentation between the different level inside ilias.

Greetings,
@iszmais

Amstutz:
Hi @iszmais

thx for the helpful explanation.

I believe it to be very hard for devs to make such a calculation. I am aware that there are formulas to calculate reading time and similar, but would this not be a bit much to ask for devs. Should they perform user tests to get the right number for their use case? Would we provide guidance with rules or description of the component? What would this guidance be? What if the devs come to different results?

A user would get used to having 5 seconds to read, since usually notifications would be displayed this long. Suddenly, a toast with only 2 seconds appears. Would that be not very irritating?

I would propose to have restrictions on the amount of information to be displayed in a toast. Then provide an example on the example page, displaying exactly this maximum amount of information so that we have a place to point to if somebody wants to perform some kind of testing to set the best possible time for the container for this exact case.

Obviously, we would try to default the time with some qualified guess on the time needed to process this maximum amount of data.

I would expect much better results this way.

Thx

iszmais:
Hi @Amstutz ,

i would see that not as an requirement thats the devs have to fullfill but as an option that they can use when it is ensured that its usage is neccessary. If the feature isnt used, atm the default is applied wich i would keep that way.
The decission for a proper specific value would therefore be discussed individualy in targeted Feature Requests and therefore hopefully by non developing individuals with sufficent expertise.
Also i would assume this is mostly the usage for longer toast to display them longer. I agree this maybe should also be defined within the interface.

Greetings,
@iszmais

Amstutz:
Hi @iszmais 

We discussed among the coordinators. We believe that we do not want the option on the public interface for individual toasts due to the above-mentioned concerns.  Please let us know if we should discuss such an option for Toast Container.

Thx a lot.

PR #6166, opened 2023-07-17
Prevent toast from vanish when an interaction happens

iszmais:
This PR prevent the vanishment of a toast when the user hovers it.
This is limited to the global refreshintervall of toasts. When another Intervall start the toast still disappears.

### DIFF
--- a/src/UI/templates/default/Toast/tpl.toast.html
+++ b/src/UI/templates/default/Toast/tpl.toast.html
@@ -1,5 +1,5 @@
 <div class="il-toast-wrapper" <!-- BEGIN id -->id="{ID}"<!-- END id --> data-vanishurl="{VANISH_ASYNC}" data-vanish="{TOAST_VANISH}" data-delay="{TOAST_DELAY}">
-    <div class="il-toast">
+    <div class="il-toast" tabindex="0">
         <div class="icon">{ICON}</div>
         <div class="title">{TITLE}</div>
         {CLOSE}
--- a/src/UI/templates/js/Toast/toast.js
+++ b/src/UI/templates/js/Toast/toast.js
@@ -4,7 +4,7 @@ il.UI.toast = ((UI) => {
     let vanishTime = 5000;
     let delayTime = 500;
 
-    let setToastSettings = (element) => {
+    const setToastSettings = (element) => {
         if (element.hasAttribute('data-vanish')) {
             vanishTime = parseInt(element.dataset.vanish);
         }
@@ -13,11 +13,11 @@ il.UI.toast = ((UI) => {
         }
     }
 
-    let showToast = (element) => {
+    const showToast = (element) => {
         setTimeout(() => {appearToast(element);}, delayTime);
     }
 
-    let closeToast = (element, forced = false) => {
+    const closeToast = (element, forced = false) => {
         element.querySelector('.il-toast').addEventListener('transitionend', () => {
             if (forced && element.dataset.vanishurl !== '') {
                 let xhr = new XMLHttpRequest();
@@ -30,10 +30,22 @@ il.UI.toast = ((UI) => {
         element.querySelector('.il-toast').classList.remove('active');
     };
 
-    let appearToast = (element) => {
+    const appearToast = (element) => {
         element.querySelector('.il-toast').classList.add('active');
-        element.querySelector('.il-toast .close').addEventListener('click', () => {closeToast(element, true);});
-        setTimeout(() => {closeToast(element);}, vanishTime);
+        queueToast(element);
+        element.querySelector('.il-toast .close').addEventListener('click', () => closeToast(element, true));
+        element.querySelector('.il-toast').addEventListener('mouseenter', () => stopToast(element));
+        element.querySelector('.il-toast').addEventListener('focusin', () => stopToast(element));
+        element.querySelector('.il-toast').addEventListener('mouseleave', () => queueToast(element));
+        element.querySelector('.il-toast').addEventListener('focusout', () => queueToast(element));
+    }
+
+    const stopToast = (element) => {
+        clearTimeout(element.vanisher);
+    }
+
+    const queueToast = (element) => {
+        element.vanisher = setTimeout(() => closeToast(element), vanishTime);
     }
 
     return {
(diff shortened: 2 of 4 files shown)


thibsy (changes requested):
Hi @iszmais

- [x] Background: could you elaborate on the background of this improvement? I think this might be something we also want to put on the factory description of toasts.

- [x] `setTimeout()`: generally speaking, using `setTimeout()` is a bad practice. This PR introduces an odd behaviour, which recursively calls the `closeToast()` callback until the element is not hovered anymore. Using `setTimeout()` here also makes this inaccurate, because a fixed interval is used before the callback is invoked again. We would prefer an approach using events, which clearly enable or disable the `closeToast()` callback. This could e.g. be achieved by using the `mouseover` and `mouseleave` events.
- [x] Unit tests: this is a feature which could be easily overlooked in future refactorings. Could you please add a simple JS unit test, which covers this behaviour. If you need any help, you can have a look at [the JS testing document](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js-testing.md) or contact us on Discord.

Kind regards,
@Amstutz and @thibsy

iszmais:
@thibsy Thank you for your review!
To your feedback:
- The background of this is to increase the users experience with the toast component since the enforced removal of a toast (wich is given to its self destroying nature) may feel unsatisfying for the user. Therefore it should be ensured that an automatic action (like defaultly vanishing the toast) should never be exectuted when it contraditcs a manual action (like a user focusing its control in any way on the toast), at least as long its not required to be executed (like the clearing of toast within the start of a new refresh interval). This way its ensured that the toast implies its service as a beneficial (and not obligational) source of information for the user.
However, there was an initial approach to summarize the automatic toast effects within the factory description wich was dismissed in the intention to keep the description abstract enough to focus on the toast basic apperance without a special context. I Agree that this should be mentioned there but only in addition to the other chronologically behaviour of the toast. I would leave this open to reconsider maybe within a JF.
- Fix it. But i used the `mouseenter` Event instead of `mouseover` since this is mostly the same as `hover`
- Added 3 test cases

thibsy (changes requested):
Hi @iszmais

I apologise for the delay here, this has been off my radar for a while.

- [x] Background: thanks a lot for your explanation. I think we can consider this an implementation detail, for which it is sufficient to be covered with a corresponding unit test. There's no need for JF approval.

- [x] Events: thanks for implementing this with events, its already a lot cleaner =). However, I noticed that using the `mouseenter` and `mouseleave` events is not quite accessible. We should switch to `focusin` and `focusout`, which are events that can also be triggered if a toast is focused via tab key. Unfortunately this is not exactly possible right now, because toasts are only focusable by accident because of their actions and/or links. We should not try to gather elements in JavaScript by the different actions and links, since this would be too inconsistent, but  make the toast itself focusable instead. This can most likely be achieved by adding a `tabindex` attribute to the toast HTML. Since this feature is not accessible otherwise, I kindly ask you to implement this change as well.
- [x] Debug statements: there are still some `console.log()` statements in the JavaScript logic, please remove them.
- [x] Unit tests: the first `stopToast()` unit test currently fails, please fix it.

Kind regards,
@thibsy

iszmais:
@thibsy Thank you for your response!

I removed the console logs and fixed the tests (this was actually a more complex problem in the test enviroment initialisation, therefore no change made to the actual code).

I understand your concerns about the accessibility but im not sure if a tabindex is an improvement here.
Imo when encountering an conceptional issue where its not clear how the apperance of a toast should interfere with the tab controll. E.g. if the index is put under a to high prio this might disturb the users workflow when a toast "suddenly" appears. Otherwise when its to low the accessibility of the toast itself is not perceptibly increased at all.

I have no expertise on how other websites handle that problem. Maybe there is a common convention for a shortcut to switch focus to non-persistent elements? Or we could handle that through the aria-live regions?

However since this PR doesnt change the current accessibility (since thats already a problem within toast) i would ignore this open issue for the moment and maybe discuss this in a ticket or JF separatly.

Greetings,
@iszmais

thibsy (changes requested):
Hi @iszmais

Thanks a lot for the quick feedback.

- [ ] A11Y: I unterstand your concerns about the `tabindex`. You are probably right, if this attribute is not used correctly, this could disturb a users workflow. However, one could also argue, due to the self-destructive nature of toasts, the priority should always be high. Nonetheless, I kindly ask you to look into this topic, because this PR introduces behaviour which is not accessible. Even though this might be the product of toasts not being fully accessible right now. If you need any help, you can also refer to the UI/UX experts of [the UI-Clinic](https://docu.ilias.de/goto_docu_grp_12155.html).
- [ ] Timeout-queue: you previously used an array to store the timeout-ids internally. Why did you switch to exposing this id on the element? This introduces the possibility to change this value, which could lead to clearing timeouts of entirely different components. If there is no good reason, please use an internal [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) instead.

Kind regards,
@klees & @thibsy

iszmais:
@thibsy Thank you for the quick response, too!

- I dont mind implementing an tabindex i just dont think its very beneficial. But since this has no direct drawbacks its fine for me. Could you give me a quick response on what you want the tabindex to be. I assume due to your last input it should be `0`? I guess altenativly only `1` is an option but i wouldnt put it any higher due to the current tab handling this is already a low prio in Ilias. Pls correct me on that if needed otherwise i will implemend the assumption.
- I switched the handling due to something i encountered while fixing the failed test. This failed because the toast inside the test html had no ID due to the fact that it not get assigned one from the ILIAS Signal structure. If i want to proceed using the map i need an external identifier for the toast and therefore would be dependend on the workflow of the signal handling. Further this also means the vanish handling only works for toasts with an active signal. To cut the dependence between these unrelated workflows i had to cut the dependencie to an external Identifier. And since JS dont provide an identifier on its own the only possible way known to me is to work with the element itsself.

Greetings,
@iszmais

iszmais:
Also there seems to neither a technical connection nor a gobal accepted convetion for aria regions and tab-accessbility therefore we might consider introducing one for ILIAS wich provides the accesibility (like some kind of forced focus on key combination).

EDIT: The only approach on that i found was [here](https://zweiterblick.at/index.php?site=toast-message) and [here](https://www.barrierefreies-webdesign.de/knowhow/pop-up-fenster/sinn-und-unsinn.html#:~:text=Die%20Anforderungen%20der%20barrierefreien%20Gestaltung,%C3%BCber%20dieses%20Linkverhalten%20zu%20informieren.) after wich popups of any kind only required to be presentated correctly but not interacted.

PR #6168, opened 2023-07-17
Feature Contacts #0037643: Contact Toasts Shy Button

fhelfer:
https://mantis.ilias.de/view.php?id=37643


iszmais:
@fhelfer Thank you for your contribution!

I spoke to the author of the mantis ticket (@oliversamoila) and since this doesnt seem to be a bug to be fixed and is more of a conceptional kind i assume the problems resolvement is not withing the general toast implementation and if it is it has to surpass the jour fixe as a UI Component change.
Also i assume that this is most likely a bit "drastic" since i assume noone intended to remove the links from the toast entirely. :smile: 

Therefore i will close this PR and we will evaluate a proper solution for the general public within the ticket

PR #6171, opened 2023-07-18
fix_37646 allow to add maps to content pages

rfalkenstein:
If this is fine for you I can cherry pick this to release_8 and trunk, too. 
"Filelist" is already supported and I don't really know what effect "Tabs" is supposed to have (as you suggested in: https://mantis.ilias.de/view.php?id=37646)


mjansenDatabay:
Thx, @rob0403. Picks for `release_8` and `trunk` are highly appreciated.

PR #6172, opened 2023-07-18
T&A Feature #0018295: Modal needed finish Test

fhelfer:
https://mantis.ilias.de/view.php?id=18295


dsstrassner:
Hi @fhelfer, I have installed a doil instance with ILIAS 8 and your PR to check the result and I like it so far. But I think we have forgotten the "Dashboard - German: Bearbeitungsstand" page and the according settings. In your PR the Modal is triggered and afterwards the "Bearbeitungsstand - Dashboard" is shown (Setting: "Show List of Questions" & --> "Show at end of test (give participants overview before final ‘Finish the Test’)".

Screenshot Settings Page ILIAS Test

I am not sure, if we need, in this case, to only show the Modal, after clicking the "Finish Test" Button at the "Bearbeitungsstand Dashboard" Page:
SCR-20230718-mbov

And one small thing:
SCR-20230718-mhtu
The Primary Button has 2 "-" too much.

fhelfer:
Hi @dsstrassner, thy for testing this PR
> The Primary Button has 2 "-" too much.

This I cannot reproduce, maybe you have a custom variable set or missed a lng reload? In the lng file & and when setting the lng variable to the button label no "-" are used. 

> I am not sure, if we need, in this case, to only show the Modal, after clicking the "Finish Test" Button at the "Bearbeitungsstand Dashboard" Page:

For the Beareitungstand/Dashboard view i am again unsure if its at all possible that we can mix Legacy Elements with new KS Buttons inside ilTable2GUI 

Best, @fhelfer

klees:
Hi @fhelfer,

I hope I can help with the second question. Please forgive me for the high-level answer, I currently could not look into the specifics here. Feel free to contact @nhaagen for further advice, he will be available again next week.

> For the Beareitungstand/Dashboard view i am again unsure if its at all possible that we can mix Legacy Elements with new KS Buttons inside ilTable2GUI

Technically it should be possible to use the standard renderer of the UI framework on some UI components and put the results into a ilTable2GUI. So from this perspective you should be fine. It might be harder or next to impossible to wire up UI components with the ilTable2GUI appropriately, to e.g. trigger some actions. So it will depend on you concrete use case how good this works out for you. From my perspective: Take your UI-component of choice, render it and give it a try.

Hope this helps a little...

Kind regards!

dsstrassner:
From my perspective, I do not need these two buttons in the iltable2GUI. For ILIAS 9, I could live with 2 buttons above the table. 
With ILIAS 10, this table has to go anyway.
I do not know if this helps in this situation?¿?

fhelfer:
> From my perspective, I do not need these two buttons in the iltable2GUI. For ILIAS 9, I could live with 2 buttons above the table. With ILIAS 10, this table has to go anyway. I do not know if this helps in this situation?¿?

@dsstrassner  Buttons in Dashboard table are now placed above. Please feel free to test again if the changes are working correctly

dsstrassner:
@fhelfer thanks. Everything looks fine, except one thing:

- Activate 'Show List of Questions'
- Activate 'Show at end of test'
- Try to finish the test at the end with Finish test --> the answer of the last question is not saved, with the click of 'Finish Test' - when you click 'Dashboard' the last question is saved.

See my screencast

When I open the detailed results an error is shown. But I do not know if this related or from my minor knowledge of cherry-picking ;-)

https://github.com/ILIAS-eLearning/ILIAS/assets/66684170/b1cee6a0-dc5a-46d2-8524-7abc43f46ee1

fhelfer:
@dsstrassner Somehow the difference in KS-Buttons and them using data-action and a classic \<a\> tag doesnt save the answer. So if the Command is outQuestionSummary the finish button iwll stay the way it was before

dsstrassner:
Hi @fhelfer, please attend the next TechSquad meeting (03 AUG 2023), so we could discuss this there: [BBB @ docu.ilias.de](https://docu.ilias.de/goto_docu_xmvc_10976.html). I hope, this is okay with you @mjansenDatabay?

fhelfer:
Hi @kergomard, i could not manage to remove the default data-action click event so the only way i could manage to keep the old behaviour was is by using         `event.stopImmediatePropagation();` 

@dsstrassner please retest again if the behaviour is as desired. thx :)

kergomard (changes requested):
HI @fhelfer
Thank you very much for this PR!

- [ ] `ilTestNavigationToolbarGUI::additional_render_items` never contains anything else, but the one single modal, right? I assume you implemented it the way you did to allow for further development. I think I would prefer though, to change it to `ilTestNavigationToolbarGUI::finish_test_modal` and strictly type it. The same would be true for `ilListOfQuestionsTableGUI::additional_items` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6172/files#diff-be84d55865e90bf8b0f37ef1f6b236dfd32c45a1b811da55fd66137015d90c37R37). I would most probably also remove the getters and setters in this case as they are for internal use only. If you decide not to change it: Why are the variables named differently in the two classes? Is this really necessary? And the same would go for `ilListOfQuestionsTableGUI::commandButtons` (where the naming becomes completely confusing, as the getters and setters are named `get/setTopItem(s)`.  This would also hugely simplify `ilListOfQuestionsTableGUI::getHTML()`.
- [ ] To me it looks wrong, that we are adding an `onClick` and then we are also adding a click-listener [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6172/files#diff-23f59f53478dbf8f08f8432474d48e94a9e3d07b3b71e3666c1c9e22be67f2e3R360-R372). It just feels like a recipe for desaster. I think we should bundle the two and just add them both together in the `onClick`. This way we could really define the order they are executed in. What am I missing?

Please change the following:
- [ ] [This line](https://github.com/ILIAS-eLearning/ILIAS/pull/6172/files#diff-23f59f53478dbf8f08f8432474d48e94a9e3d07b3b71e3666c1c9e22be67f2e3R78) isn't exactly in the right place. Please move.
- [ ] `$invited_user["clientip"] !== null` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6172/files#diff-2f0e8d5eafb6a326db0c3b8b1d6ef13629055f8d4ecddedfbc36bd5223180d6fR674) is superfluous as it is contained in the `isset($invited_user['clientip'])`. I would replace it with `$invited_user['clientip'] !== ''`, but we may also be able to remove it.
- [ ] Please make sure all variables are named in snake-case (no [camelCases](https://github.com/ILIAS-eLearning/ILIAS/pull/6172/files#diff-be84d55865e90bf8b0f37ef1f6b236dfd32c45a1b811da55fd66137015d90c37R33)).
- [ ] Please rebase to resolve the conflicts.

I did not test the behavior, I defer to @dsstrassner for this. I will provide him with a version of trunk with your changes and some necessary adaptations to make it run once you have gone through me comments above.

Thanks again and best,
@kergomard

fhelfer:
> HI @fhelfer Thank you very much for this PR!
> 
> Please consider the following suggestions. You do not need to follow those, but please indicate shortly why you prefer to do otherwise:
> 
> * [x]  `ilTestNavigationToolbarGUI::additional_render_items` never contains anything else, but the one single modal, right? I assume you implemented it the way you did to allow for further development. I think I would prefer though, to change it to `ilTestNavigationToolbarGUI::finish_test_modal` and strictly type it. The same would be true for `ilListOfQuestionsTableGUI::additional_items` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6172/files#diff-be84d55865e90bf8b0f37ef1f6b236dfd32c45a1b811da55fd66137015d90c37R37). I would most probably also remove the getters and setters in this case as they are for internal use only. If you decide not to change it: Why are the variables named differently in the two classes? Is this really necessary? And the same would go for `ilListOfQuestionsTableGUI::commandButtons` (where the naming becomes completely confusing, as the getters and setters are named `get/setTopItem(s)`.  This would also hugely simplify `ilListOfQuestionsTableGUI::getHTML()`.
> * [x]  To me it looks wrong, that we are adding an `onClick` and then we are also adding a click-listener [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6172/files#diff-23f59f53478dbf8f08f8432474d48e94a9e3d07b3b71e3666c1c9e22be67f2e3R360-R372). It just feels like a recipe for desaster. I think we should bundle the two and just add them both together in the `onClick`. This way we could really define the order they are executed in. What am I missing?

Different Event Listeners if for when modal needs to be triggered or if Dashboard Navigation is required. Missed this one
> 
> Please change the following:
> 
> * [x]  [This line](https://github.com/ILIAS-eLearning/ILIAS/pull/6172/files#diff-23f59f53478dbf8f08f8432474d48e94a9e3d07b3b71e3666c1c9e22be67f2e3R78) isn't exactly in the right place. Please move.
> * [x]  `$invited_user["clientip"] !== null` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6172/files#diff-2f0e8d5eafb6a326db0c3b8b1d6ef13629055f8d4ecddedfbc36bd5223180d6fR674) is superfluous as it is contained in the `isset($invited_user['clientip'])`. I would replace it with `$invited_user['clientip'] !== ''`, but we may also be able to remove it.
> * [x]  Please make sure all variables are named in snake-case (no [camelCases](https://github.com/ILIAS-eLearning/ILIAS/pull/6172/files#diff-be84d55865e90bf8b0f37ef1f6b236dfd32c45a1b811da55fd66137015d90c37R33)).
> * [x]  Please rebase to resolve the conflicts.
> 
> I did not test the behavior, I defer to @dsstrassner for this. I will provide him with a version of trunk with your changes and some necessary adaptations to make it run once you have gone through me comments above.
> 
> Thanks again and best, @kergomard

Thx for the review. I managed to check all your requested changes

Best @fhelfer

kergomard:
Thank you for the quick fixes @fhelfer !
I found two more things when preparing the branch for @dsstrassner :
- [ ] The "Finish Test"-Button on the Question List (so when you show the question list before finalizing the test), is not working.
- [ ] There is an access to an uninitialized variable.

This two things would be fixed with: https://github.com/fhelfer/ILIAS/pull/5 (as far as I can see).
Best,
@kergomard

dsstrassner:
The Branch @kergomard has provided, shows me the PR within a functional ILIAS 9 instance. I could not find any flaw there any more. Thanks @fhelfer and @kergomard!

kergomard:
Thank you very much for the merge @fhelfer ! I merge now.

dsstrassner:
@fhelfer Yeah! Thanks!

PR #6173, opened 2023-07-18
Add ext modules in composer.json (ext-zip;ext-imagick)

PurHur:
These php extensions are missing and are a requiremend since the migration of the bare exec commands. 

There is no mantis issue yet, but will prevent a lot of mantis issues in the future 🤓

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,6 +41,9 @@
 		"ext-json": "*",
 		"ext-simplexml": "*",
 		"ext-libxml": "*",
+		"ext-xml": "*",
+		"ext-zip": "*",
+		"ext-imagick": "*",
 		"cweagans/composer-patches": "^1.7",
 		"filp/whoops": "^2.15.1",
 		"monolog/monolog": "^2.3.5",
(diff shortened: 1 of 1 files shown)


iszmais:
@chfsx 

While trying to install ILIAS 9 on some "slim" PHP foundations, I did some research and came to notice that the ImageMagick extension ist considered "Experimental" https://www.php.net/manual/en/extensions.state.php, which seems to raise minor concerns for one of our clients.
I assume this is nothing to worry about but therefore I just wanted to double-check if this extension really needs to be a dependency or if we might want to consider it to be optional.

Greetings,
@iszmais

chfsx:
@iszmais yes, this is definitely needed (generating previews and other image processing in ILIAS) and it's the very first time i've seen it listed as experimental. as far as i can see, this information is only available on this page, nowhere else. i'm not sure if the information is correct.

iszmais:
@chfsx Thank you for the qualified feedback!
This was indeed verified to be "deprecated" by the maintainers of the documentation: https://github.com/php/doc-en/issues/3643#issuecomment-2273103313

PurHur:
Note that i wrote this comment 2022 as you switched from bare exec to an outdated extension. 

https://github.com/ILIAS-eLearning/ILIAS/pull/5172#issuecomment-1302591170

iszmais:
@chfsx I just feel obliged to inform you that the PHP documentation was updated, but the Extension ImageMagick is still listed as experimental.

No Change Request. Just FYI.

Greetings,
@iszmais

PR #6176, opened 2023-07-20
Remove unused code with potentially high impact on performance (ILIAS 9)

tfamula:
The variable `$nodes` is never used in the code. Because the method `ilObjOrgUnitTree::getAllChildren` can potentially lead to performance issues, the unused should be better removed.


smeyer-ilias:
Thanks. I've already merged this

PR #6180, opened 2023-07-20
Fix LP determination depending on course setting

iszmais:
https://mantis.ilias.de/view.php?id=37659


smeyer-ilias:
Thanks @iszmais for this PR
I close this report according to the notes in the mantis report.

PR #6183, opened 2023-07-21
Fixed type hint of assignment type

iszmais:
_There is no Mantis ticket since this issue doesnt cause anything rn.
But i will do when a strict type is enforced on the file._

The Type of assignments is, in contrast to object types like `crs` or `grp`, no string but a number specified in constants and is required as such by other components.

### DIFF
--- a/Modules/Exercise/Assignment/class.ilExAssignment.php
+++ b/Modules/Exercise/Assignment/class.ilExAssignment.php
@@ -1067,7 +1067,7 @@ class ilExAssignment
         return self::lookup($a_id, "title");
     }
 
-    public static function lookupType(int $a_id): string
+    public static function lookupType(int $a_id): int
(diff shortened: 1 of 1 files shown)


Modules/Exercise/Assignment/class.ilExAssignment.php
```diff
@@ -1067,7 +1067,7 @@ public static function lookupTitle(int $a_id): string
         return self::lookup($a_id, "title");
     }
 
-    public static function lookupType(int $a_id): string
+    public static function lookupType(int $a_id): int
     {
         return self::lookup($a_id, "type");
```
alex40724, line 1072: Note that lookup returns a string here.

alex40724:
Thanks, I made this change in trunk only.

PR #6185, opened 2023-07-24
Fix 0037685: updateRunInformation for running cron jobs

rsheer:
Fix according to mantis issue: https://mantis.ilias.de/view.php?id=37685


mjansenDatabay:
Thx @rsheer

PR #6186, opened 2023-07-25
Re-Added ImageMagick as dependency since T&A uses it

chfsx:
Most of the usages of `convert` have been replaced by an abstraction with php-imagick, unfortunately there are still places where the `convert` command has to be used, so this dependency does not completely disappear with ILIAS 9.

The following pages have also been adapted: 
- https://docu.ilias.de/goto_docu_wiki_wpage_7458_1357.html
- https://docu.ilias.de/goto_docu_wiki_wpage_7974_1357.html


matthiaskunkel:
Jour Fixe, 07 AUG 2023 : Fabian notified the JF that ImageMagick is still used by T&A and therefore still exists as dependency. Thanks for updating the related documents.

PR #6189, opened 2023-07-25
Fix 0037690: Cast string to int for calendar consultation

rsheer:
Cast string to int for admin_id value from database


smeyer-ilias (approved):
Thanks for fixing this

PR #6190, opened 2023-07-26
T&A Bugfix #0037651: No question pool selected when deleting from Test

fhelfer:
https://mantis.ilias.de/view.php?id=37651

also cherry-pick for 7&8 as well


kergomard:
Thank you very much @fhelfer ! This looks good, but I would prefer another naming for the lang-var. I will merge and quickly change it myself (not worth the round-trip).

PR #6194, opened 2023-07-27
TA Bug 37663

dsstrassner:
Change in PR 4610 seems to be the cause for the error in https://mantis.ilias.de/view.php?id=37663

Please look carefully. Need to cherry-pick to trunk and ILIAS 7, if solution is practicable.

### DIFF
--- a/Modules/Test/classes/class.ilTestExportGUI.php
+++ b/Modules/Test/classes/class.ilTestExportGUI.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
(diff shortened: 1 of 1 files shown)


kergomard:
Yup, you are right with your change. We can clean-Up a little bit more. [this](https://github.com/ILIAS-eLearning/ILIAS/pull/6194/files#diff-ae90e98bc0091df04010780f14e47475f213a4d3b0c2f0487aba4fed5b9283a8L215-L217) is unnecessary, too. I will merge and then add this clean-up, too.

PR #6200, opened 2023-08-01
UI KS Feature: ViewControl Mode moved to the right side

fhelfer:
For the development of https://docu.ilias.de/goto_docu_wiki_wpage_7503_1357.html we decided that the Mode Selection should be placed with the other View Controls as well.
This will look like the following screenshot.
image


oliversamoila:
Just a small note in case there is any question about the required space (especially for smaller devices): The labels in ViewContol mode will certainly be even shorter. Glyphs are used in the final state, but they are currently in the making.

Amstutz:

PR #6201, opened 2023-08-02
Chatroom: Use centralized `online/offline` handling and `Object Prope…

mjansenDatabay:
…rties`


Modules/Chatroom/classes/class.ilChatroomFormFactory.php
```diff
@@ -82,96 +88,135 @@ private function addDefaultBehaviour(ilPropertyFormGUI $form): ilPropertyFormGUI
...
-
-        $cb = new ilCheckboxInputGUI($this->lng->txt('allow_custom_usernames'), 'allow_custom_usernames');
-        $txt = new ilTextInputGUI($this->lng->txt('autogen_usernames'), 'autogen_usernames');
-        $txt->setRequired(true);
-        $txt->setInfo($this->lng->txt('autogen_usernames_info'));
-        $cb->addSubItem($txt);
-        $form->addItem($cb);
-
-        return $form;
+        return $this->refinery->custom()->transformation(static function (array $values): array {
```
lscharmer: Can be simplified to:
```suggestion
        return $this->refinery->custom()->transformation(static fn (array $values): array => array_merge(...$values));
```

Modules/Chatroom/classes/class.ilChatroomXMLParser.php
```diff
@@ -101,7 +101,9 @@ public function handlerEndTag(XMLParser $a_xml_parser, string $a_name): void
                 break;
 
             case 'OnlineStatus':
-                $this->room->setSetting('online_status', (int) $this->cdata);
+                $this->chat->setOfflineStatus(
+                    ((int) $this->cdata) ? false : true
```
lscharmer:
```suggestion
                    ((int) $this->cdata) !== 0
```

Modules/Chatroom/classes/class.ilObjChatroomAccess.php
```diff
@@ -80,25 +78,19 @@ private static function checkRoomAccess(string $a_permission, int $a_ref_id, int
 
         switch ($a_permission) {
             case 'visible':
```
lscharmer: Duplicated code. The cases `visible` and `read` are the same, can be simplified with:

```suggestion
...
            case 'visible':
            case 'read':
                if (!$is_online) {
                    $DIC->access()->addInfoItem(
                        ilAccessInfo::IL_NO_OBJECT_ACCESS,
                        $DIC->language()->txt('offline')
                    );
                    return false;
                }
                break;
....
```

Modules/Chatroom/classes/gui/class.ilChatroomSettingsGUI.php
```diff
@@ -29,50 +29,57 @@ class ilChatroomSettingsGUI extends ilChatroomGUIHandler
...
-                    $value = $settingsForm->getInput($setting);
-                }
+        $room = ilChatroom::byObjectId($this->gui->getObject()->getId());
+        $mutated_settings = $room->getSettings();
+        if (!$room) {
+            $room = new ilChatroom();
+            $mutated_settings['object_id'] = $this->gui->getObject()->getId();
+        }
+
+        foreach ($mutated_settings as $setting => &$value) {
```
lscharmer, line 69: Can be simplified with:
```suggestion
        $mutated_settings = array_merge($mutated_settings, array_intersect($mutated_settings, $values));
```

lscharmer (changes requested):

lscharmer (approved):

PR #6203, opened 2023-08-02
T&A: fix for 37202

okaufman:
This will fix https://mantis.ilias.de/view.php?id=37202


kergomard:
Ok, after a quick discussion @dsstrassner ,  we merge this and I will cherry-pick. We will see if somebody complains.

PR #6204, opened 2023-08-07
Trunk: T&A - Fix int value used as parameter in ilRadioOption obj creation

mBeym:
Fixes problem causing a crash when trying to access the **Create question** view in a Test

- ilRadioOption only takes a string **value** parameter. `declare(strict_types=1)` added to this class in the past (31.03.2023) makes this fail then.

There are a lot of those wrong value parameters for ilRadioOption in ILIAS,  
most/all of which are in classes that don't yet have `declare(strict_types=1)` enabled.

If desired I can fix those too.

I stumbled over this error specifically while working on implementing:

https://docu.ilias.de/goto_docu_wiki_wpage_7423_1357.html

### DIFF
--- a/Modules/Test/classes/class.ilObjTestGUI.php
+++ b/Modules/Test/classes/class.ilObjTestGUI.php
@@ -1648,15 +1648,15 @@ class ilObjTestGUI extends ilObjectGUI implements ilCtrlBaseClassInterface
         // use pool
         $usage = new ilRadioGroupInputGUI($this->lng->txt("assessment_pool_selection"), "usage");
         $usage->setRequired(true);
-        $no_pool = new ilRadioOption($this->lng->txt("assessment_no_pool"), 1);
+        $no_pool = new ilRadioOption($this->lng->txt("assessment_no_pool"), "1");
         $usage->addOption($no_pool);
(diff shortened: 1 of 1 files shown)


dsstrassner:
hi @mBeym, could you please create a Mantis Issue, so I can add the bugfix Label? Or is this an improvement?

kergomard:
Hi @mBeym 

Thank you very much! This will not be the last of these, yes.
I would have a little request for this:
- Could you please use single quotation-marks instead of doubles. We should only use double ones when really needed (e.g. because there are variables in the string or because we need some formatting). If we implement this thoroughly our code will get easier to read and search.
- If you find the time, it would be a big improvement, if you could replace all occurrences, but I will merge this as soon as you implemented the single quotes.

I don't think we necessarily need a mantis issue for this. We can actually declare it an improvement (especially if you get around to replacing all of the wrong assignments). Nobody is going to complain if you create one, though.

Thanks again and best,
@kergomard

mBeym:
@kergomard I replaced the double quotes with single ones.

I'll create a mantis ticket later and work on fixing all of the occurrences I can find when I have some free time

kergomard:
Thanks again @mBeym ! You don't need to worry about the report. I merge this.

PR #6206, opened 2023-08-07
Trunk: remove limit number of concurrent participants

mBeym:
Implements [Remove Setting «Limit Number of Concurrent Participants»](https://docu.ilias.de/goto_docu_wiki_wpage_7423_1357.html) (Feature Wiki)


kergomard (changes requested):
Thank you very much @mBeym for removing this!

Just a couple of things:
- [x] I don't think [this line here](https://github.com/ILIAS-eLearning/ILIAS/pull/6206/files#diff-5d6ab1ffacd501d89a09d5369268119388245d55f0629ea07b307b6326a5f0f2L260) should go, right?
- [ ] Is there a reason you are not dropping `limit_users_enabled` from the db?

I think this is it.

Thanks again and best,
@kergomard

kergomard:
Sorry, @dsstrassner just pointed out to me, that the line is actually there and I had simply not seen it. My bad!

mBeym:
@kergomard good catch! looks like that table slipped through. Fixed!

kergomard:
Thank you @mBeym for checking and adding! I will merge and the limit concurrent pax feature is history.

PR #6207, opened 2023-08-08
Database: Fix Uninitialized Variable in ilDatabaseUpdateObjective

kergomard:
See: https://mantis.ilias.de/view.php?id=37737


chfsx:
Thanks a lot!

PR #6209, opened 2023-08-08
024085: Imagemap question does not show correct solution

fneumann:
see https://mantis.ilias.de/view.php?id=24085


kergomard:
Thank you very much @fneumann for this PR!

I think you are right and we don't need this here. If I'm getting this right, it is about showing the clickable areas and we are in the context of having a correct solution, so we shouldn't need that.

I removed the code with a direct commit https://github.com/ILIAS-eLearning/ILIAS/commit/1637f94216b1195cc9ccfff158d74710ee552ac1 as I strongly believe we should stop commenting code out as this quickly leads to a  huge mess, but it is also not worth another extra round.

Thanks again and best,
@kergomard

PR #6211, opened 2023-08-08
9/test/feature completely remove ctm

kergomard:
This PR is purely informational and I will be merging this myself:

- This removes most of the remnants of the CTM and adds a migration to remove the actual tests.
- The migration will not run due to: https://mantis.ilias.de/view.php?id=37738 , but I would still merge this.
- This also contains a small fix for the suggested-solutions-repository, as the column containing the type of a suggested solution can also be `null` or an empty string, and removes two `strcmp`s in Logging. They are both in separate commits. Do not Squash!


dsstrassner:
Please merge, after fixing the tests ;)

PR #6212, opened 2023-08-08
9/UI buttons, mainbar, slate style refactoring

catenglaender:
# Issue

* Button code was a leftover from Bootstrap
* For the disabled state colors are forced on top as an afterthought
* Buttons are tricky to style for custom system style skins if the former bootstrap mixin doesn't support what you are going for (e.g. custom hover or disabled colors)
* In extension mainbar, metabar, slate buttons and drilldown buttons are tricky to style too
* Glyphs and other elements are sometimes hard to position in the precise middle

# Change

* **new button mixin `make-button()`** to fully replace bootstraps `button-variant()` and `button-size()`
  * all you need at minimum is `$button-color`, everything else like hover colors are generated from it
  * but it also has parameters for all kinds of states and settings that would otherwise be generated from the button color (e.g. a very specific button color like `$engaged_hover_bg-color` can be adressed to custom override the default)
  * mixin can output code for basic button structure, button design, button size only by setting `$set-basics`, `$set-design`, `$set-size` to `true/false` respectively, so you can override something with just the parts you need to override
  * basics can also be set on import additionally to or instead of passing them to the mixin as parameters using `@use "../../../030-tools/_tool_buttons" as * with (...);`
  * a preset styling for menu style buttons for slate/drilldown can be set with `$set-full-width: true`
* bulky buttons now have a base styling that already resembles their look in drilldown and slate - this saves some lines of code
* square metabar and mainbar buttons are styled as an override of the bulky button without using the mixin so only the minimum is repeated in the resulting CSS
* Many variables have been moved from _layout_standardpage.scss into the respective components, so e.g. mainbar background colors are now in _ui-components_mainbar.scss. This exposed such component specific variables were mis-used in another component.
* The new button-mixin is not designed to handle unnecessary empty span elements (e.g. an empty dropdown label) and hardcoded whitespaces between label and caret/glyphs as the button manages the distance between caret/glyphs. These have therefor been removed in html and php files.

# What do we gain...

* -20 lines of CSS code... I had hoped for more, but I guess you can only trim down button code to a certain extend if you want to reach all states
* code is more organized, design changes and tweaks are easier to do and maintain in the future
* Even very specific settings variables can now actually travel into the mixin instead of being attached as an override or patch.

This is the default delos look:
Screenshot from 2023-08-08 12-54-26

And this is what you can get playing around with the SCSS variables for a couple of minutes for a custom system style skin as some service providers would build. For now it's not planned to make such heavy customization possible through the ILIAS Admin GUI, customizing all colors and changing the look drastically is only possible if your fork and compile your own SCSS - but if you do so, you can get very far by adjusting variables alone.

Screenshot from 2023-08-08 12-44-46

Screenshot from 2023-08-08 13-51-31

# Still sub-optimal

* I wasn't expecting to feed so many parameters to the mixin, but I have to keep them exposed to the settings layer. I encourage service providers with heavily styled skins to try the make-button mixin with only the bare minimum because the default fallbacks have been very carefully chosen. Or maybe we decide that we don't want to have some states on the settings layer anymore?
* no detailed refactoring for breadcrumbs, metabar, standardpage yet
* old naming convention as all standardpage components and many unittests are connected to this
* keyboard focus outlines are still functional, but not quite the same as they were in ILIAS 8. The way how focus outlines were set in ILIAS 8 was a bit patchy, using the il-focus mixin but overriding it when a distance to the outer border or a second ::after element outline wasn't wanted. I plan to add a focus-inner mixin in another PR that can be used in such cases to unify the inconsistent overrides.


Amstutz:
thx a lot @catenglaender . There are lots of changes. However, most have been discussed and serve the main purpose of having better control over buttons. To avoid further conflicts and to include the changes in early upcoming a11y tests, I merge this.

PR #6217, opened 2023-08-14
9/legacy-UI/form Textarea height 37674

catenglaender:
Mantis https://mantis.ilias.de/view.php?id=37674

# Issue

Textarea for messages is currently single line. Users don't know immediately that they can write a longer message. Used to by a multi line field in ILIAS 8.

image

# Change

Setting textarea heigth back to browser default

image


Amstutz:
thx a lot! special thx for the really good description with screenshot (as always)!

PR #6218, opened 2023-08-14
fix mantis bug #37702: Placeholders get lost along with HTML tags in langvars

c-knof:
https://mantis.ilias.de/view.php?id=37702

in ilUtil::secureString the allow_array and secure_tags differ, therefore all secure tags and additionally span and br are transmitted

### DIFF
--- a/Services/Language/README.md
+++ b/Services/Language/README.md
@@ -1,0 +1,6 @@
+# Language Service
+
+document in progress
+
+## HTML Tags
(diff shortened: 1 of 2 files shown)


mjansenDatabay:
LGTM! Maybe we should document this in the README.md of `Services/Language`?

katringross:
@mjansenDatabay  thanks for your support, we added a readme

PR #6221, opened 2023-08-15
T&A Feature #0037744: Question pool titles in tests

fhelfer:
https://mantis.ilias.de/view.php?id=37744

### DIFF
--- a/Modules/Test/classes/tables/class.ilTestQuestionBrowserTableGUI.php
+++ b/Modules/Test/classes/tables/class.ilTestQuestionBrowserTableGUI.php
@@ -21,6 +21,7 @@ declare(strict_types=1);
 use ILIAS\HTTP\GlobalHttpState;
 use ILIAS\Refinery\Factory as Refinery;
 use ILIAS\Test\InternalRequestService;
+use ILIAS\Modules\Test\traits\QuestionPoolLinkedTitleBuilder;
 
 /**
  * @author Helmut Schottmüller <ilias@aurealis.de>
@@ -28,6 +29,7 @@ use ILIAS\Test\InternalRequestService;
  */
 class ilTestQuestionBrowserTableGUI extends ilTable2GUI
 {
+    use QuestionPoolLinkedTitleBuilder;
     private const REPOSITORY_ROOT_NODE_ID = 1;
 
     public const CONTEXT_PARAMETER = 'question_browse_context';
@@ -47,7 +49,6 @@ class ilTestQuestionBrowserTableGUI extends ilTable2GUI
 
     /** @var array<string, mixed> */
     private array $filter = [];
-    private ILIAS\DI\UIServices $ui;
 
     public function __construct(
         private ilTabsGUI $tabs,
@@ -64,7 +65,6 @@ class ilTestQuestionBrowserTableGUI extends ilTable2GUI
     ) {
         global $DIC;
         $this->setId('qpl_brows_tabl_' . $this->test_obj->getId());
-        $this->ui = $DIC->ui();
 
         parent::__construct($this, self::CMD_BROWSE_QUESTIONS);
         $this->setFilterCommand(self::CMD_APPLY_FILTER);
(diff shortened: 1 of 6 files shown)


fhelfer:
cherry-pick down to 7 as well

kergomard (changes requested):
Thank you very much for this PR @fhelfer !

I will let this wait until after the merge of the Clean-Up PR as I would kindly ask you to change the approach to accessing UIElements. Maybe we will already have that, after the PR, if not: Please inject the ui_factory and the ui_renderer separately through the constructor (you will find a consistent pattern through my changes (I hope).

But then I've deeper doubts about this:
[Here](https://github.com/ILIAS-eLearning/ILIAS/pull/6221/files#diff-bc7692fae178852001326f5d9b350e6b4e274d924015ecbaa4b0900efa922a7fR444) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6221/files#diff-2d337a7313a471838d9aa68b847fdc1b8f29b5d6d225216df5eef24988dd1e60R227) you are just retrieving all references and then choosing a random one to display (actually the first one, but this is basically the same). Now there might be multiple references (as far as I can see) and the user might have access to only some of them, so I think we would need to check that:
- If the array contains only one element: Perfect! We use that.
- If it contains multiple elements we search for the first one the user has access to and we use that.
- If she doesn't have access to any of the references we just don't use a link.

I also believe we shouldn't link the title, if it is a link to another test. Personally I believe we should not show that Title in this case as it does not have any informational value.

I hope I make sense here otherwise don't hesitate to ask and we can always discuss, if I went completely off the rails!

Thanks again and best,
@kergomard

dsstrassner:
Maybe we wait until after my return to discuss linking other tests. It may be that I did not think enough about that.

dsstrassner:
As discussed in 2023-09-14 TechSquad Meeting, @fhelfer will check, if we could limit the presentation of questions pool links according to the read / view rights (onto the question pool) of the person who is looking into the particular test and his questions.

Also, he will limit the presentation to questions from question pools. We will not show the link to other tests.

The question of original or copied and changed question will be tackled by others in the future (cc @kergomard).

kergomard:
Hi @fhelfer 
Could you please rebase to resolve the conflicts and give me a go ahead as soon as it is ready for another review?
Thank you very much and best,
@kergomard

fhelfer:
Hi @kergomard,

PR has been rebased and is ready. Please review again 
Best, @fhelfer

kergomard (commented):
...especially for also fixing the linking in the `ilTestRandomQuestionSetSourcePoolDefinitionListTableGUI`!
I would kindly ask you to change the following things or to explain, why not:
- [ ] This  [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6221/files#diff-2d337a7313a471838d9aa68b847fdc1b8f29b5d6d225216df5eef24988dd1e60R234) looks very much like we only link a question when it appears in only one reference of a pool a user has access to, but this shouldn't be the case: If there are multiple references to the same pool the user has access to, we should just link to the first one. Now, this is random, it can also be the last, but we might implement this more cheaply if we take the first one - see next point. I would, as a rule, never use `count()` unless you really do need the amount. Otherwise `$v !== []` sounds like the better and more readable option in our case.
- [ ] This leads me to the next point: Could you please split this up into functions with early returns? I would probably create two functions. One `getFirstQuestionPoolReferenceWithCurrentUserAccess(array $all_qpl_ref_ids)` (if you find a better wording, I'm a happy taker, this one has the advantage of being precise, but it is loooonnnng ;-) ) and then I would create one `buildPossiblyLinkedQuestonPoolTitle(int $qpl_obj_id)`. Probably the first one would be called from within the second one, ...but there might be better solutions for the issue!
- [ ] From where I stand [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/6221/files#diff-2d337a7313a471838d9aa68b847fdc1b8f29b5d6d225216df5eef24988dd1e60R249) is a lie ;-) . This should be fixed by the previous point, but why not add a "No access" / "Kein Zugriff" in parenthesis behind the title (I would probably use the language variable `status_no_permission` from `common`), so the unlinked title in this case would look something like ('My Shiny Question Pool (No permission granted)'.
- [ ] And finally something very small: We have two variables `$qp_ref_ids` and `qp_references`. Their names don't tell us, what the difference between them is. I would ask you to give them more verbose names, so inside the function `getFirstQuestionPoolReferenceWithCurrentUserAccess(array $all_qpl_ref_ids)`, there will be a variable `$ref_ids_of_qpls_with_access` or most probably after the change above `$ref_id_of_qpl_with_access`, so afterwards we have an easy time understanding it.
- [ ] Similar changes would apply to the `ilTestRandomQuestionSetSourcePoolDefinitionListTableGUI` and `ilTestQuestionBrowserTableGUI`. Maybe you find a solution where you need to implement this only once? Maybe in `ilTestObj`?

@kergomard

dsstrassner:
I am thrilled how much afford you both invest in this issue. I like your dedication to improving the T&A every time, we change a little bit of code!

kergomard:
Hi @fhelfer 

This PR is in your field, right? You are not waiting for me? Let me know, if I should do something.

Thank you very much and best,
@kergomard

fhelfer:
Hi @kergomard,

thx for the reminder, i unforunately lost my stashed changes / forgot to push them.

As discussed with you via discord, i created a new trait for the duplicated code lines and addressed your comments. You might still find issues with this approach, pls just let me know if theres anyhting still in the wrong.

Best, @fhelfer.

kergomard:
Thank you very much @fhelfer!
There is always something ;-), but I will merge this and do a lilttle bit of refactoring myself. Just for your information:
- I will move the parts you are taking from the `$DIC` in the trait to function parameters.
- I will remove `trait` from the namespace.
- I will restructure this [if-monster here](https://github.com/ILIAS-eLearning/ILIAS/pull/6221/files#diff-c714f0e71c76a6c6630414660796e0e948da1dbc6ce2a07b3bacb3bde4619daaR37) a little bit.

...and then I will run some tests.

Thanks again and best,
@kergomard

kergomard:
So, this led to more work than expected ;-) .

My fixes for ILIAS 9 are:
https://github.com/ILIAS-eLearning/ILIAS/commit/441bff564e152e4e5f042e33d44ecf48f3a0c924
https://github.com/ILIAS-eLearning/ILIAS/commit/30618462c961327039a549cb64334261cb4ee0d4

Best,
@kergomard

PR #6222, opened 2023-08-15
UI/DataTable: View Controls for the Table

nhaagen:
Building upon #5811, this integrates the ViewControls into DataTable.


klees (changes requested):
Hi @nhaagen,

thanks for working on this so tenaciously. I'm not 100% sure I understand everything in here, so please bear with me.

Really nice to see all the stuff coming together, though...

- [ ] Some Code 1: Why do we need [this](https://github.com/ILIAS-eLearning/ILIAS/pull/6222/files#diff-6f7f25104712ede5c719cab13082a881862cd82c8b9c01a77228ebd6e666ff15R88)? Looks like something that `URLBuilder` should do?
- [ ] Some Code 2: What is [this](https://github.com/ILIAS-eLearning/ILIAS/pull/6222/files#diff-9957152d868b16ebd0847046247129f1a1ada960ee8ff8dd5ac90e194740ec4cR156) supposed to do? Looks brittle somehow.
- [ ] `Sortation`: This would exist twice, right? Once as `View Control` above the table, once via column headers? Can we keep the URLs with `<a>`s in the column headers somehow?
- ...

- [x] `getViewControls` 1: We probably will want to allow users to add custom `View Controls` to `Data Tables` sometime. Also, `Renderer::getViewControls` does not look as if really has to do much with rendering. I think the code should be moved `Table\Data` and we won't lose a thing.
- [x] `getViewControls` 2: The method looks like an excellent target for some unit tests and refactoring: It has clearly discernible parts (`getPaginationViewControl`, `getSortationViewControl` and `getFieldSelectionViewControl`) where the behaviour depends in a non-trivial way on various inputs. This seems to be at the heart of the issue. Please break apart and write tests.
- [x] `Range`: [This snippet here](https://github.com/ILIAS-eLearning/ILIAS/pull/6222/files#diff-a4f983edeb2e721f6d1a2d0eaf11f4beba3760482f48a9c422ce89e4f5c79ba5R337) seems to be something for the `Range` data type. Name might be something like `withMaximum`.
- [x] Empty View Controls: [This](https://github.com/ILIAS-eLearning/ILIAS/pull/6222/files#diff-a4f983edeb2e721f6d1a2d0eaf11f4beba3760482f48a9c422ce89e4f5c79ba5R336), [that](https://github.com/ILIAS-eLearning/ILIAS/pull/6222/files#diff-a4f983edeb2e721f6d1a2d0eaf11f4beba3760482f48a9c422ce89e4f5c79ba5R343) and [that](https://github.com/ILIAS-eLearning/ILIAS/pull/6222/files#diff-a4f983edeb2e721f6d1a2d0eaf11f4beba3760482f48a9c422ce89e4f5c79ba5R347) seem to indicate that the table should get the feature to remove range, order and field selection. The code looks a little odd, because basically, if no value is set by the view control, we use whatever value was set before. IMO this could lead to strange behaviour, such as no field is selected, but some field is actually selected in the displayed data. I think `Data::withOrder`, `::withRange` and `::withSelectedOptionalColumns` should allow to pass `null` to remove all selections. This could then be used in `Renderer::applyViewControls`.

Contact me if we need to discuss this...

Best regards!

klees:
Hi @thibsy, hi @Amstutz,

this LGTM now, besides one issue: We didn't manage to turn the headers of sortable columns into links until now. They work, but they are buttons, not links as they are supposed to be. I'll create an according a11y-ticket once this is merged to document the issue and will fix this during beta.

Kind regards!

Amstutz:
Hi @nhaagen

Thx a lot for this large contribution. Special thx for adding on and documenting the examples. Highly appreciated. Further, I really like how our different puzzle peaces are coming together in the data table. Kudos.

You added a minimal change on the public interface here https://github.com/ILIAS-eLearning/ILIAS/pull/6222/files#diff-38bb22893d0f8ee91d55b1524ad2bf0bbe35a1f8e94c1974777dd3255d091fbfR60. However, this is such a minor detail, that we can easily fast lane your changes.

Please Change (note, some may have been existing before your change, but need fixing nevertheless):

- [x] JS Code: You expended on the inline JS Code here: https://github.com/ILIAS-eLearning/ILIAS/pull/6222/files#diff-38bb22893d0f8ee91d55b1524ad2bf0bbe35a1f8e94c1974777dd3255d091fbfR60 . Please move that to an external JS File. Try to keep inline JS as minimal as possible. Fix this.
- [x] Base Example I: If you hop to page 2 in the pagination an then set the page count to 500, you still only see the one example from page 1.
- [x] Base Example II: If you set the range to unlimited, I currently get a fatal: ILIAS\Data\Range::getEnd(): Return value must be of type int, float returned. Fix this.
- [x] Base Example III: There are 2 optional success fields. One is initially hidden. However, both do not initially appear and both appear and both appear in the Field Selection to be shown. Probably give them different names or remove one.
- [x] Repo Example: Same as above 2optional success field. Further a progress Field. Is this intended? Maybe remove them.
- [x] Params: There are too many, here, https://github.com/ILIAS-eLearning/ILIAS/pull/6222/files#diff-87a73ca4faa58bfa4f1071586b935e00c203f39f848da752b2b917ce60adae2bR425
- [x] Constants: Should be on the very top https://github.com/ILIAS-eLearning/ILIAS/pull/6222/files#diff-87a73ca4faa58bfa4f1071586b935e00c203f39f848da752b2b917ce60adae2bR340

Perhaps Change or indicate why not:
- [x] Initially hidden/Optional: Optional fields that are NOT initially hidden, seem to be hidden initially nevertheless, and need to be activated through the field control. Is this intentional? Potentially fix.


Thx a lot
@Timon Amstutz

klees:
Hi @Amstutz,

thanks for your review!

* Regarding the change in the public interface: That was me =)
* Regarding the JS of the ViewControls container: We currently do not have any structured JS for the view controls I could attach that too. I fear that we will fall into a JS-review spiral here that I cannot afford until coding completed. So I added an entry to the Roadmap. Hope that is ok for now.
* Regarding initially hidden: I think this is the same problem as described in [#38070](https://mantis.ilias.de/view.php?id=38070). I guess the interface is not good atm, didn't catch that before. I'll propose an according change in a separate PR to make that breaking change more visible.

Kind regards!

klees:
@Amstutz This is updated!

klees:
@Amstutz I was wrong about the last point. I'll still provide an interface improvement separately, but the other issue will be fixed here. On it.

Amstutz:
thx a lot. Entries on the Roadmap are fine with me.

PR #6224, opened 2023-08-15
Remove legacy type safety check

PurHur:
Just a small cleanup commit that removes an obsolete check.


smeyer-ilias:
Thanks for fixing this

PR #6225, opened 2023-08-15
PRG: filter users by permissions if 'mail to assigned users' is called. (#31413)

nhaagen:
https://github.com/ILIAS-eLearning/ILIAS/pull/6125


nhaagen:
picked from 8

PR #6227, opened 2023-08-15
Object: fix import

schmitz-ilias:
This PR fixes [37774](https://mantis.ilias.de/view.php?id=37774). As far as I can see, it does not need to be picked to R8 and above.


kergomard:
Thank you very much @schmitz-ilias ! This looks right, thanks!

PR #6228, opened 2023-08-15
Course: expand export/import (24733)

schmitz-ilias:
This PR fixes [24733](https://mantis.ilias.de/view.php?id=24733) by solving a problem with importing the Course setting 'News', and implementing import/export for the settings 'Determination of Status 'Passed'' and 'Mail to Members' as well as for the Course Map.

The setting 'Add to Favourites' and the 'Files for Download' from Course Information were intentionally left out: the former because of the low relevance of the setting, and the latter because of the larger footprint of the necessary changes.


smeyer-ilias (approved):
Thanks, @schmitz-ilias

PR #6229, opened 2023-08-16
SOAP: Fix plugins not working in ILIAS 8

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=37776

If approved, this commit MUST be cherry-picked to `trunk` as well.


jeph864:
@mjansenDatabay 
Thanks for the PR. It looks like a huge chain of bugs

PR #6230, opened 2023-08-16
User: Self-registration: no pre-selection of roles

kergomard:
Hi @pascalseeland  and @mjansenDatabay 

I don't think we need a formal review of this, it is more FYI: The implementation of the feature
See: https://docu.ilias.de/goto.php?target=wiki_1357_Self-registration%3A_no_pre-selection_of_roles
requires [a little change in Registration](https://github.com/ILIAS-eLearning/ILIAS/pull/6230/files#diff-81f281a543d649ee19d5ecfe7621c9735693d65eda2074977314427d7dc7d376R337) otherwise it will always show an error message. I will add another commit to this with clean-ups in User today. This doesn't need to concern you.

Thank you very much and best,
@kergomard


mjansenDatabay (approved):
LGTM!

kergomard:
Sorry @mjansenDatabay and @pascalseeland : I didn't think far enough!

While cleaning-up `ilUserProfile` I found a few static-methods I strongly believe we should get rid of. So there are two more commits here with changes for you:
- https://github.com/ILIAS-eLearning/ILIAS/pull/6230/commits/153f11971141f39972a9ec5e22444233fce8752c handles `Auth`
- https://github.com/ILIAS-eLearning/ILIAS/pull/6230/commits/d7adc2f2c45bd1476b53189c989b170b3b8c11e5 `Registration`

Sorry again and best,
@kergomard

mjansenDatabay (approved):

PR #6231, opened 2023-08-16
9/legacy-UI/carousel correct import 37762 37484

catenglaender:
https://mantis.ilias.de/view.php?id=37762

# Issue

Carousel in modal appears unstyled.

image

# Changes

/legacy/component_carousel.scss wasn't properly imported by component index.

image

### DIFF
--- a/templates/default/070-components/_index.scss
+++ b/templates/default/070-components/_index.scss
@@ -61,6 +61,7 @@
 
 @use "./legacy/component_animated-collapse-fade";
 @use "./legacy/component_btn-group";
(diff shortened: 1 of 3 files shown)


catenglaender:
Idea for sass Unit tests: Check if all scss partials are ever imported at some point, so we notice if components get completely disconnected. I don't think this check can easily be done by sass itself. Maybe php can look through all import lines and compare them against the file list?

Amstutz:
thx for the fix. I agree that some measure could be helpful here. But how would php get the list of files needing to be imported?

PR #6232, opened 2023-08-16
Session: fix import of material assignment (25992)

schmitz-ilias:
This PR fixes [25992](https://mantis.ilias.de/view.php?id=25992).


smeyer-ilias:
Thanks for fixing this.

PR #6233, opened 2023-08-16
9/UI/dropdown repository add item menu font size and padding

catenglaender:
# Issue

Font size and padding of the "add item" repo dropdown was off due to the bootstrap grid refactoring.

image

# Changes

Tweaked font size and padding.

image


catenglaender:
I wonder if we should take some time to think about if we want to tweak this look (in a follow-up PR). Our dropdown menus rarely have padding, why should this one be different? I do like the spacious look, but it kinda doesn't fit to the rest of delos. Or should delos in ILIAS 9 become more spacious?

Amstutz:
Nice to see this fix by even losing one line :-). 

I agree with the assessment, that the "add new item" menu could need a closer look!

PR #6236, opened 2023-08-17
T&A Bugfix #0025306: Ordering Question (Vertical) : problems with code "<T>"

fhelfer:
https://mantis.ilias.de/view.php?id=25306

cherry-pick down to 7 as well

### DIFF
--- a/Modules/TestQuestionPool/classes/forms/class.ilAssFormSubmitEntitiesEscaper.php
+++ b/Modules/TestQuestionPool/classes/forms/class.ilAssFormSubmitEntitiesEscaper.php
@@ -1,0 +1,50 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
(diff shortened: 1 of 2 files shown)


fhelfer:
image

<T> will not be stripped by its tags and therefore be empty, but displayed in its htmlspecialchars() version

kergomard:
Thank you very much @fhelfer ! I will need to look at this close, so I will leave it until next week and concentrate on the logging right now

kergomard:
Hi @fhelfer

Sorry, this got stuck for so long with me! I lost it out of sight for a sec. I don't think this solution is satisfactory. We will need to find an option that:
* Doesn't alter the output (to this degree).
* Doesn't leave us upon for XSS attacks.
I think the problem that we have now, is double encoding. Could you please see if you can figure how to show our users an output without html-entities, without opening us up to all sort of attacks?

Thanks and best,
@kergomard

Amstutz:
@fhelfer : Any update on this?

kergomard:
Hi @fhelfer
Do you think you could have another look at this?

Thank you very much and best,
@kergomard

kergomard:
As discussed in the TechSquad today. I close this PR for the time being and assign the issue to me.
Thank you very much @fhelfer for all your efforts!

PR #6240, opened 2023-08-17
[trunk] Introducing ilDBInterface::primaryExistsByFields  to check if a specific primary key exists in database

fwolf-ilias:
Dear @chfsx

as discussed in [#37750](https://mantis.ilias.de/view.php?id=37750) an ilDB function to check if a specific primary key exists for a database table. It is later used to check if the tree table has the primary key child set or uses an index instead.

Regards
Fabian Wolf


chfsx:
Thanks a lot @fwolf-ilias for the three PRs!

PR #5964, opened 2023-06-06
Feature/9/irss exc instruction files

chfsx:
Dear @alex40724 

This pull request contains the adjustments in the exercise to save instruction files via a migration to the IRSS on the one hand, and the adjustments to upload new files directly to the IRSS on the other.

The implementation uses the ResourceCollections as a basis. In the following, I will go through a few commits that contain the relevant changes for you. Please note that the PR also contains the implementation of the ResourceCollection GUIs, which are treated separately (see [feature request](https://docu.ilias.de/goto_docu_wiki_wpage_7841_1357.html)). However, so that you can also try out the branch, this implementation is also already in it.

Commits:

- https://github.com/ILIAS-eLearning/ILIAS/pull/5964/commits/c6462b11a1ac374cf0cff312c9b705cafe2024a2 : Only small code style adjustment
-https://github.com/ILIAS-eLearning/ILIAS/pull/5964/commits/bf258879c322c8575be51e7c7fbf36519853986e : Adding a new stakeholder for the instruction files
- https://github.com/ILIAS-eLearning/ILIAS/pull/5964/commits/fe17905f109f20e3abae42e36503a542ad275ab0 : New database field 'if_rcid' (for Instruction Files Resource Collection ID) in the table exc_assignment
- https://github.com/ILIAS-eLearning/ILIAS/pull/5964/commits/c9746373d65c4fba8197edc85bd7c4664dd642e8 : The actual migration of the existing files
- https://github.com/ILIAS-eLearning/ILIAS/pull/5964/commits/3ab5bfe87fb1e5e48c22652c9fbbe12bccfba9a8 : Adjustments in the assignments so that new files get into the IRSS and existing files are displayed correctly
- https://github.com/ILIAS-eLearning/ILIAS/commit/cc0438b44a92a39f81553300899dd9c25f9b797e : The Resource Collection GUI mentioned above, which does not have to be reviewed by you.

Within the scope of this PR, no further refactorings were made, as the implementation is only to contain the migration and the application of the IRSS. of course, you can also change things afterwards, e.g. to simplify things even more.

There are at least some points to discuss from my side:
- https://github.com/ILIAS-eLearning/ILIAS/commit/3421ee3684cdce0b4c4177126d42b289d36a7232#r116693370
- https://github.com/ILIAS-eLearning/ILIAS/commit/3421ee3684cdce0b4c4177126d42b289d36a7232#r116693681
[...]

PR #6245, opened 2023-08-21
Bulky Dropzone

chfsx:
This proposal allows a drop zone to be marked as "Bulky". The following sentence serves as a description here:

`Makes a dropzone bulky, this to ensure better visibility when the dropzone is to be displayed standalone.`

This is only a proposal. in the following screenshots you can see how the difference shows up. this would be used e.g. in the new ilResourceCollectionGUI (which is used e.g. in the exercise, see PR #5964).

current:
Bildschirmfoto 2023-08-21 um 07 49 16

with bulky: 
Bildschirmfoto 2023-08-21 um 07 48 50

In principle, however, we also have the option of visually modifying the current standard dropzone or leaving the dropzone as it is. a bulky variant is possibly another variant we could make.

this PR does NOT refactor the dropzone in terms of naming variables and classes.


Amstutz:
Hi @chfsx 

Thx a lot for adding this. We will need to announce the change at the upcoming JF. I therefore added the label. However, the changes seem fine and since the public interface change is minimal, we will fast track this change. 

So please:
- [x] Conflicts: Resolve the conflicts.

If done, you may directly merge, to avoid future conflicts. 

thx a lot!

matthiaskunkel:
Jour Fixe, 16 OCT 2023: @chfsx has notified us about this PR that has already been merged to trunk. We highly appreciate this PR belatedly.

PR #6247, opened 2023-08-21
Removed author for Badges, added "Author wanted"

fab-kru:


matthiaskunkel:
Thanks for the update.

PR #6248, opened 2023-08-21
Update badges

fab-kru:
As requested by @matthiaskunkel! Cheers…


fab-kru:
Doh, this didn’t work as planned. Will try again later.

PR #6250, opened 2023-08-21
UI KS Fix&Feature: Panel ViewControls Placement + Examples

fhelfer:
Follow up for #6200


Amstutz:
Hi @fhelfer 

Thx for improving the View controls here and adding the examples. 

- [ ] ID for examples: Why did you add an ID [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6250/files#diff-d5b487706646519c33121a9828761e80f7e927fbd4a56f5924de9d7885505a98R113). Probably remove.

thx a lot.

@Amstutz

fhelfer:
> * [x]  ID for examples: Why did you add an ID [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6250/files#diff-d5b487706646519c33121a9828761e80f7e927fbd4a56f5924de9d7885505a98R113). Probably remove.

It allows for the links in view_control to navigate directly to the section. Used [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6250/files#diff-54c01e05006dd1177d6f9521fd6703840a8b40be9c364a78e0814bfcb444c2adR48)
Its a usability change i preferred, so i dont have to scroll through all the examples to see what my click did to the panel.
Unfortunately this only works with the mode switch. For ViewControls with method ->withTargetUrl() it does not work (But would be nice for ux to have smth like a direct scoll to the interacted vc)
If that complicates the PR approval, I will remove it.

Best,
@fhelfer

Amstutz:
Hi @fhelfer 

Thx a lot for the explanation. We believe that added functionality to be highly useful and would appreciate such an enhancement. However, we would need to look into the details of how we would generate those links consistently for the examples. 

Since this, those not tackle the original intent of the PR, we would recommend removing this proposal here, so this can be merged soon. 

So for this PR, best:

- [ ] ID for examples: Remove them here.

Thx a lot
@Amstutz

fhelfer:
Hi @Amstutz, the id has been removed
Best,
@fhelfer

Amstutz:
thx a lot!

PR #6251, opened 2023-08-21
[FIX] 0031534: Identical Page Titles

chfsx:
this PR solves the described A11y problem from https://mantis.ilias.de/view.php?id=31534. in my opinion, this is a workaround and will certainly be revised in the context of general refactoring, how a page is put together. for the moment, however, we can probably live with it...

Some examples:
- Settings: My funny Course Title: ILIAS
- Dashboard: ILIAS


Amstutz (approved):
Thx a lot for providing this fix. I agree with this being a workaround we should improve on later. However, this is an important a11y fix. kudos.

chfsx:
CP to trunk as well

PR #6252, opened 2023-08-21
Fix/0037784/8/save confirmation with fileupload

chfsx:
This is a solution for https://mantis.ilias.de/view.php?id=37784. this was the only place where only the first 8 characters of the session id was used for the temp-file.


Amstutz:
Thx a lot!

PR #6255, opened 2023-08-22
Soap response timeout in  setup config

jeph864:
Different components deal with response timeout differently and mostly  use  a default value defined in the soapclient class. For requests which needs  to run a longer time than the default value, this posits an issue. For example, the soap copyObject Method is expected to return reference ids in the response. However,  large objects which usually take longer, send a response to client before the request has finished processing. The IDs sent will be zero since the function returned before they were available. This PR adds the possibility to set this value globally in the setup config. Components are still allowed to set their own values if they don't want to use the default value.

### DIFF
--- a/Services/WebServices/SOAP/classes/class.ilSoapClient.php
+++ b/Services/WebServices/SOAP/classes/class.ilSoapClient.php
@@ -54,7 +54,7 @@ class ilSoapClient
         }
         $this->connect_timeout = $timeout;
         
-        $this->response_timeout = self::DEFAULT_RESPONSE_TIMEOUT;
+        $this->response_timeout = $ilSetting->get('soap_response_timeout', self::DEFAULT_RESPONSE_TIMEOUT);
     }
     
     /**
--- a/Services/WebServices/SOAP/classes/class.ilSoapClient.php
+++ b/Services/WebServices/SOAP/classes/class.ilSoapClient.php
@@ -54,7 +54,7 @@ class ilSoapClient
         }
         $this->connect_timeout = $timeout;
         
-        $this->response_timeout = $ilSetting->get('soap_response_timeout', self::DEFAULT_RESPONSE_TIMEOUT);
+        $this->response_timeout = (int) $ilSetting->get('soap_response_timeout', self::DEFAULT_RESPONSE_TIMEOUT);
     }
     
     /**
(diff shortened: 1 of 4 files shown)


Services/WebServices/classes/Setup/class.ilWebServicesSetupConfig.php
```diff
@@ -69,4 +76,12 @@ public function getRPCServerPort() : int
     {
         return $this->rpc_server_port;
     }
+
+    /**
+     * @return int
+     */
```
mjansenDatabay:
```suggestion
```

Services/WebServices/classes/Setup/class.ilWebServicesConfigStoredObjective.php
```diff
@@ -53,6 +53,7 @@ public function achieve(Setup\Environment $environment) : Setup\Environment
         );
         $settings->set("soap_wsdl_path", $this->config->getSOAPWsdlPath());
         $settings->set("soap_connect_timeout", (int) $this->config->getSOAPConnectTimeout());
+        $settings->set("soap_response_timeout", (int) $this->config->getSoapResponseTimeout());
```
mjansenDatabay:
```suggestion
        $settings->set("soap_response_timeout", (string) $this->config->getSoapResponseTimeout());
```

Services/WebServices/classes/Setup/class.ilWebServicesConfigStoredObjective.php
```diff
@@ -53,6 +53,7 @@ public function achieve(Setup\Environment $environment) : Setup\Environment
         );
         $settings->set("soap_wsdl_path", $this->config->getSOAPWsdlPath());
         $settings->set("soap_connect_timeout", (int) $this->config->getSOAPConnectTimeout());
```
mjansenDatabay:
```suggestion
        $settings->set("soap_connect_timeout", (string) $this->config->getSOAPConnectTimeout());
```

Services/WebServices/SOAP/classes/class.ilSoapClient.php
```diff
@@ -54,7 +54,7 @@ public function __construct($a_uri = '')
         }
         $this->connect_timeout = $timeout;
         
-        $this->response_timeout = self::DEFAULT_RESPONSE_TIMEOUT;
+        $this->response_timeout = $ilSetting->get('soap_response_timeout', self::DEFAULT_RESPONSE_TIMEOUT);
```
mjansenDatabay:
```suggestion
        $this->response_timeout = (int) $ilSetting->get('soap_response_timeout', self::DEFAULT_RESPONSE_TIMEOUT);
```

PR #6256, opened 2023-08-22
[FIX]  0025388: Group  and Course XML parser ignore empty spaces in ContainerSetting element

jeph864:
This PR fixes https://mantis.ilias.de/view.php?id=25388


smeyer-ilias:
Thanks @jeph864 for fixing this.

jeph864:
@smeyer-ilias ,
Apparently the same issue appears in courses as well. I will open another PR with the fix

PR #6258, opened 2023-08-22
9/legacy-UI header image repo page title 37668

catenglaender:
# Issue

On ILIAS 9 repo pages, the econd description / status line doesn't float next to the icon image as the headlines line-height ist larger than it was in ILIAS 8

image

# Change

Rather than decreasing the line-height to match ILIAS 8, I propose increasing the margin around the image to not have the text lines look too crowded.

image

image


Amstutz:
thx a lot!

PR #6259, opened 2023-08-22
9/UI/modal backdrop fade opacity 37669

catenglaender:
Solves Mantis: https://mantis.ilias.de/view.php?id=37669

# Issue

The dark background behind the modal was not transparent enough.

image

# Change

This was caused by the general animation tool being more specific than the specific modal animation. Added the .fade class to the end of animation class .in for the modal backdrop.

image


Amstutz:
thx a lot

PR #6262, opened 2023-08-23
9/UI/dropdown fix one line in table header, set inline, menu position 37670

catenglaender:
Fixes: https://mantis.ilias.de/view.php?id=37670

# Issue

## Initial issue
Dropdown menu was misaligned. Dropdown itself should be inline with other buttons.

image

## Newly discovered issue
Technically, the buttons changing rows and columns and showing/hiding filters are now considered view controls in ILIAS 9. Since the new input viewcontrols are now merged, the row button should be an input viewcontrol pagination. The other ones should be input viewcontrols as well.

# Changes

## Initial issue
Fixed. The buttons are now in one line. The menu opens correctly directly below the dropdown.

image

## Newly discovered issue
My knowledge of php is not good enough to change this. Is it worth fixing the legacy table2? Will they be replaced by UI framework data table soon?

### DIFF
--- a/templates/default/070-components/UI-framework/Dropdown/_ui-component_dropdown.scss
+++ b/templates/default/070-components/UI-framework/Dropdown/_ui-component_dropdown.scss
@@ -55,6 +55,7 @@ $cursor-disabled: not-allowed !default;
   .dropup,
   .dropdown {
 	position: relative;
+	display: inline-block; // so dropdown can be in one line with a button
   }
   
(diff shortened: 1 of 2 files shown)


Amstutz:
Hi @catenglaender, the whole table along with it's controls can be considered legacy components. Fixing the SCSS is fine (and must be done), but I would go no further than that other for some a11y fixes.

PR #6263, opened 2023-08-24
LegacyUI/SplitButton: Fix padding of buttons in split button

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=34577

We don't need this for ILIAS >= 9 because the `SplitButton` will be replaced with alternative KS elements.


Amstutz:
thx, does this need a pick down to 7?

mjansenDatabay:
@Amstutz Yes, should I provide a PR for `release_7` as well?

PR #6264, opened 2023-08-24
Fix 37803: Frequent errors in ILIAS log »Duplicate entry 'MathJax-dummy' for key 'PRIMARY«

fwolf-ilias:
Remove the usage of ilSettingsFactory::settingsFor outside of a setup environment because it tends to trigger database integrity constraint violations if called simultaneously on crowded installations

### DIFF
--- a/Services/MathJax/classes/class.ilMathJax.php
+++ b/Services/MathJax/classes/class.ilMathJax.php
@@ -124,7 +124,7 @@ class ilMathJax
         global $DIC;
 
         if (!isset(self::$_instance)) {
-            $repo = new ilMathJaxConfigSettingsRepository(new ilSettingsFactory($DIC->database()));
+            $repo = new ilMathJaxConfigSettingsRepository();
             self::$_instance = new self($repo->getConfig(), new ilMathJaxFactory());
         }
         return self::$_instance;
--- a/Services/MathJax/classes/class.ilMathJaxConfigSettingsRepository.php
+++ b/Services/MathJax/classes/class.ilMathJaxConfigSettingsRepository.php
@@ -25,9 +25,13 @@ class ilMathJaxConfigSettingsRepository implements ilMathJaxConfigRespository
     /**
      * Constructor
      */
-    public function __construct(ilSettingsFactory $factory)
+    public function __construct(?ilSettingsFactory $factory = null)
     {
-        $this->settings = $factory->settingsFor('MathJax');
+        if($factory === null) {
+            $this->settings = new ilSetting('MathJax', true);
+        } else {
+            $this->settings = $factory->settingsFor('MathJax');
+        }
     }
 
     /**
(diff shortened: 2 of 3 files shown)


fneumann (commented):
Thanks for your analysis and the quick fix. As you wrote the issue is caused by  ilSettingsFactory::settingsFor which always does a write - even when it is just called for reading the settings. In the comments of the class I don't see that it is intended to be only used in a setup context. So I would prefer the issue to be fixed there. 
@klees:  what do you think? The class ilSettingsFactory was introcuded by you.
If it is possible to use the SettingsFactory in normal requests, then I would like to keep this depenency as  required in the constructor of ilMathJaxConfigSettingsRepository. If not, then I would like to avoid it completely and use 'new ilSetting' in all cases instead. An implict
fallback looks pragmatic but somehow strange to me.

klees:
I agree with @fneumann. I think we should try to tackle the root cause of the problem instead of just avoiding it. So, the `set` is there to force a (re)set of `$value_type`. Hence, another solution would be to just use `_getValueType` from `ilSettings` directly somehow.

Maybe this `$value_type`-thingy can be removed altogether. I vaguely remember that we have removed the possibility to use `clob` for the settings some time ago. Anyone can point to a the according ticket or change? @mjansenDatabay maybe?

mjansenDatabay:
> Anyone can point to a the according ticket or change? @mjansenDatabay maybe?

I can, one minute ...

-- UPDATE --

* PR: https://github.com/ILIAS-eLearning/ILIAS/pull/4948
* Mantis Issue: https://mantis.ilias.de/view.php?id=33860
* Jour Fixe, 19 SEP 2022 : **We have no objections against changing to varchar(4000). Please adapt as suggested.**

klees:
Thx @mjansenDatabay!

Looks exactly like what we need. Seems as if this didn't make it into release_8 or trunk until now. Any idea why the change is missing? (@alex40724 maybe?)

fwolf-ilias:
@fneumann in case you want to switch completely to `new ilSetting()` keep in mind, that the database might not be defined in all cases. With my patched installation I've got this error when running `php setup/setup.php update`: 

```
Check configuration of Services/MathJax...                                 [FAILED]
                                                                                                                        
 [ERROR] Identifier "ilDB" is not defined.
```
This is also the reason i suggested keeping SettingsFactory when calling from the setup cli.

Nevertheless I agree with y'all. I also would prefer a fix within SettingsFactory. But as this is an urgent error from my side, I have decided on a pracmatic solution that would only affect one developer.

fneumann:
I now merged a similar PR which avoids the fallback condition in the constructor of ilMathJaxConfigSettingsRepository but gets the prepared ilSetting object differently for setup and requests. 

@kless, @mjansenDatabay 
Thanks for investigating the underlying problem with ilSettingsFactory. If you like to go on with that, please create a new mantis issue.

PR #6265, opened 2023-08-24
DC: check uniqueness for all types of fields - #0030758

sz-tik:
fixed Mantis Bug #0030758: https://mantis.ilias.de/view.php?id=30758

check uniqueness for all types of fields:
complements for formula field, rating field

### DIFF
--- a/Modules/DataCollection/classes/Content/class.ilDclRecordEditGUI.php
+++ b/Modules/DataCollection/classes/Content/class.ilDclRecordEditGUI.php
@@ -512,13 +512,56 @@ class ilDclRecordEditGUI
         }
 
         //Check if we can create this record.
-        foreach ($all_fields as $field) {
-            try {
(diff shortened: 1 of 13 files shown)


Amstutz:
Hi @sz-tik 

Thx a lot for this fix for the Data Collection. Highly appreciated. I struggle somewhat to completely comprehend all things at work here. Maybe it would be best here, if we have a short talk through discord, teams, skype, whatever you prefer, and you can walk me through your changes, and we can discuss several aspects directly? What do you think?

Thx 
@Amstutz

Amstutz:
@sz-tik and @amstutz had a session over Webex discussing this PR. Most of the changes are good and important. Thx a lot @sz-tik for tackling this challenging ticket. We found some aspects that can be improved and refined somewhat.

Nevertheless, the change is rather large and complex. We agree that we would prefer to publish it only for trunk. We therefore close this PR and @sz-tik will open a new version of it for trunk.

PR #6266, opened 2023-08-24
[FIX] 0037785: Search does not show the correct suggestions

chfsx:
Hi @smeyer-ilias 
the only problem in https://mantis.ilias.de/view.php?id=37785 is the wrong escaping of quotes here: https://github.com/ILIAS-eLearning/ILIAS/compare/release_8...srsolutionsag:fix/0037785/8/wrong-quote-handling-in-search#diff-a3ee71ff1123f159bf6b2a4d84c52fb31f4932623ba559f7b1193ac391cb7e27R159

### DIFF
--- a/Services/Search/classes/class.ilQueryParser.php
+++ b/Services/Search/classes/class.ilQueryParser.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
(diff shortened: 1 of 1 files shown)


smeyer-ilias (approved):
Thanks for fixing this.

mjansenDatabay:
@smeyer-ilias / @chfsx These changes result in issues when searching for local roles in the LDAP administration (role assignment rules) https://mantis.ilias.de/view.php?id=38006

PR #6267, opened 2023-08-24
DC refactor SelectInputGUI Switcher to Dropdown for LUI

Amstutz:
Replaced SelectInputGUI with KS Dropdown for FR: https://docu.ilias.de/goto_docu_wiki_wpage_7975_1357.html

Note that the label for the Dropdowns acting as Switchers has yet to be set.

Before: 
image

After:
image


yvseiler:
+1 for dropdown label "Tabelle wechseln"

PR #6268, opened 2023-08-25
Fix  0036953: SOAP method copyObjects returns no ref_ids for long requests

jeph864:
https://mantis.ilias.de/view.php?id=36953
This PR allows  the methods to use the response timeout defined in  settings instead of the hardcoded default one.


alex40724:
Thanks for the PR! Merged to 7, 8 and trunk.

PR #6269, opened 2023-08-25
[FIX] 0037812: Identical Page Title for Subtabs

chfsx:
This is a follow-up to #6251 and adds the subtab as well to the page-title, see https://mantis.ilias.de/view.php?id=37812


Amstutz:
I believe some of the code would better be placed closer to the tabs than here. However, the way you did it, we contained it all in one scope, which should ease the way for future refactorings. I see no better way. Thx a lot for the comments explaining this.

And many thx for resolving the issue.

PR #6270, opened 2023-08-25
Trunk: Fix ilRadioOption value parameter type used

mBeym:
Done according to: https://github.com/ILIAS-eLearning/ILIAS/pull/6204#issuecomment-1667432304


kergomard:
Ohhh, I didn't mean the whole ILIAS-Code-Base, sorry, just the Test. The changes for the test are ok with me. Thank you very much!

@alex40724 : Is it ok, if we merge this. I think most of the other changes are in your code and  @mBeym is right, that this is more correct then before.

kergomard:
Thank you very much @mBeym for this PR. As @alex40724 did not react, I changed the parts in the test with: https://github.com/ILIAS-eLearning/ILIAS/commit/78bbc9420dbc3a7504841a288f182541ed936bed and would now close this PR, to keep my backlog short.

Thank you very much and best,
@kergomard

PR #6271, opened 2023-08-25
UI, Docu: allow docu-entries for array-returns and outside of UI/Component

nhaagen:
https://mantis.ilias.de/view.php?id=37753

### DIFF
--- a/src/UI/Implementation/Crawler/EntriesYamlParser.php
+++ b/src/UI/Implementation/Crawler/EntriesYamlParser.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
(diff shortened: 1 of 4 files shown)


klees (approved):

klees:
@thibsy @Amstutz Please look into this.

I had doubts if we really want to have Non-UI-Components on the factory. I came to the conclusion that we indeed want this: As soon as have more data structures that aren't UI components (like the help topics) we need to create them somewhere. And I expect that there will be indeed more of these data types, as the content of the UI framework becomes more complete and we indeed need to talk about (some) concepts that do not have a visual representation.

Maybe we should discuss this some time...

thibsy (approved):

PR #6272, opened 2023-08-25
Test: Fix datetime handling with start/end

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=37757


kergomard:

Just one question: What does prevent us from just setting the timezone on the `DateTimeImmutable` and stop there? This to me looks cleaner, but I'm probably missing something.

Thanks again and best,
@kergomard

mjansenDatabay:
Hi @kergomard ,

> What does prevent us from just setting the timezone on the `DateTimeImmutable` and stop there? 

Nothing :-). I just implemented it like shown in the example. I'll change that.

Best regards,
@mjansenDatabay

kergomard:

PR #6273, opened 2023-08-25
UI, Input/SwitchableGroup: fix #37805, disable in current block

nhaagen:
https://mantis.ilias.de/view.php?id=37805


klees:
Thx!

PR #6274, opened 2023-08-25
UI: DataTable #37493, make action-ids optional

nhaagen:
https://mantis.ilias.de/view.php?id=37493

ids for the actions are no longer mandatory

### DIFF
--- a/src/UI/Implementation/Component/Table/Renderer.php
+++ b/src/UI/Implementation/Component/Table/Renderer.php
@@ -211,7 +211,7 @@ class Renderer extends AbstractComponentRenderer
 
         $actions = [];
         foreach ($component->getAllActions() as $action_id => $action) {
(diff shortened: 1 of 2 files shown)


klees:
Hi @nhaagen,

do we currently need that id for anything?

Best wishes!

nhaagen:
No, the id is really optional. all strings are within the action itself, and there is no reference to the array key.

klees:
Ok, so I guess we should drop them.

tfamula:
@nhaagen When there is no action-id for an action, then `DataRow::withDisabledAction` will not work for this action, right? Because you have to pass the action-id as parameter here. Or did I understand something wrong?

klees:
@tfamula This seems to be right. Currently we have only dropped the `$action_id` in the example and made it possible to use integer ids as well. So technically, your use case should still work. Don't know if we have improved regarding comprehensibility, though. Thanks for the hint!

PR #6275, opened 2023-08-28
Test: Fix availability datetime handling

mjansenDatabay:
This PR fixes the timezone handling of the test availability when populating the form with the values from the persistence layer.

See comments of https://mantis.ilias.de/view.php?id=37757


kergomard:

PR #6276, opened 2023-08-28
UI: refactor DateTime's picker to HTML5 for accessibility.

lukastocker:
Also remove datetime-picker javascript from DateTime rendering.

Adjust DateTime tests.

**NOTE:**
Currently it's not possible to remove the bootstrap-datetimepicker.js files because the Button > Months still uses it: There's no Month-Picker for HTML5 until now - only a <input type="month".. /> but it's just functioning like a text input - so it's not a picker. Therefore the Button "Month" is still not accessible.

The duration.js file isn't needed anymore, because duration is now using the HTML5 types too.

PR #6278, opened 2023-08-28
Test: Fix scoring datetime handling

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=37823

### DIFF
--- a/Modules/Test/test/ScoreSettingsTest.php
+++ b/Modules/Test/test/ScoreSettingsTest.php
@@ -256,7 +256,12 @@ EOT;
 
         $s = new ilObjTestSettingsResultSummary(666);
         $actual = $this->getDefaultRenderer()->render(
-            $s->toForm(...$ui)
+            $s->toForm(...array_merge($ui, [[
+                'user_time_zone' => 'Europe/Berlin',
+                'user_date_format' => $data_factory->dateFormat()->withTime24(
+                    $data_factory->dateFormat()->standard()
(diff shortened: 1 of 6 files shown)


Modules/Test/test/ScoreSettingsTest.php
```diff
@@ -294,7 +299,7 @@ public function testScoreSettingsSectionSummary(): void
                                     <label for="id_3" class="control-label col-sm-4 col-md-3 col-lg-2">tst_reporting_date<span class="asterisk">*</span></label>
                                     <div class="col-sm-8 col-md-9 col-lg-10">
                                         <div class="input-group date il-input-datetime" id="id_3">
-                                            <input type="text" name="" placeholder="YYYY-MM-DD HH:mm" class="form-control form-control-sm" />
+                                            <input type="text" name="" placeholder="YYYY-MM-DD H:mm" class="form-control form-control-sm" />
```
mjansenDatabay, line 302: I wonder why a `HH` is transformed to a `H` in the UI renderer, but this should not be the problem of te T&A.

mjansenDatabay: Interesting, could be related to https://github.com/ILIAS-eLearning/ILIAS/pull/6276

kergomard:
Thank you very much @mjansenDatabay ! Will merge and cherry-pick to ILIAS8.

PR #6279, opened 2023-08-28
UI, URLBuilder: distinguish and apply array-parameters

nhaagen:
https://mantis.ilias.de/view.php?id=37494
mutli-action parameters should be sent as array.

### DIFF
--- a/src/UI/URLBuilder.php
+++ b/src/UI/URLBuilder.php
@@ -223,6 +223,7 @@ class URLBuilder
     {
         $params = array_merge($this->uri->getParameters(), $this->params);
         $query = (! empty($params)) ? '?' . http_build_query($params) : '';
+        $query = preg_replace('/%5B[0-9]+%5D/simU', '%5B%5D', $query);
         return $query;
     }
 
--- a/src/UI/URLBuilder.php
+++ b/src/UI/URLBuilder.php
@@ -217,7 +217,9 @@ class URLBuilder
 
     /**
      * Create the query part of the URL from all parameters
-     * Claimed parameters overwrite base parameters in array_merge()
+     * Claimed parameters overwrite base parameters in array_merge(),
+     * numeric indizes of array-parameters are being removed to ensure
+     * continous numeration (p[1]=A&p[2]=B --> p[]=A&p[]=B).
      */
     private function buildQuery(): string
     {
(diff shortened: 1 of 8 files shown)


klees (commented):
Hi @nhaagen,

thanks a bunch. Nice that we don't seem to require changes in PHP.

Please check one thing for me: How will PHP behave if we pass an array via URL where indizes are missing. E.g. `title[0]` and `title[2]` are passed, but not `title[1]`. Will this still become a list with continuous indizes? If not, I think we should just remove the indizes in the URL.

- [ ] Encoding: Do we really need to urlencode the `[` and `]`? Why? Will PHP still parse that stuff?

Kind regards!

nhaagen:
foo.php?indexed[12]=A&indexed[7]=B&arr[]=C&arr[]=D
will provide
array(2) { [12]=> string(1) "A" [7]=> string(1) "B" }
and 
array(2) { [0]=> string(1) "C" [1]=> string(1) "D" }


However, the continuity should be ensured by ".writeParameter()" setting the indizes itself.

klees:
Yeah, but URLs should be readable and editable by humans. So when you provide me with `https://my.domain/foo.php?tag[]=bar&tag[]=baz` I want to be able to add and remove tags etc. This will be harder if I feel pressured to correct enumerations. We probably would want some according safe guards server side when we retrieve the values, though.

So, please change the rendering to `[]` instead of `[1]`. Thanks!

nhaagen:
will do, but this will cause quite some changes in the internal param storage of the builder.
Maybe it's enough to ditch the indizes on the output.

nhaagen:
updated the url-output instead of storage, works nicely.

klees:
Seems that the `[]` should be encoded as of [RFC 3986](https://stackoverflow.com/questions/40568/are-square-brackets-permitted-in-urls). Firefox seems to do its best to hide that masking for its users, though. I will roll back that change.

klees:
@thibsy This LGTM. Please have a quick look and, hopefully, hit the merge button. Thanks!

thibsy (changes requested):
Hi @nhaagen 

Thx a lot for looking into this, it mostly LGTM as well. I have one small request though:

- [x] `URLBuilder::buildQuery()`: could you extend the comment of this function, to explain why we are using a `preg_replace()` in there? I think we should document somewhere, that we are removing indices from the URL and why we are doing it.

I'll merge right after this!

Kind regards,
@thibsy

thibsy (approved):

thibsy:
@alex40724, FYI: this concerns the issue you have mentioned during the last LUI project meeting. With this change, values of all actions can be retrieved as arrays, sync and async.

PR #6281, opened 2023-08-28
fix sql quotation of identifiers

iszmais:
This is a partial fix for https://mantis.ilias.de/view.php?id=37825


chfsx:
Thanks a lot @iszmais, I committed in https://github.com/ILIAS-eLearning/ILIAS/commit/d0f2f72b7e2c8bb3389140671be9929779fc3925 with another small change and will port this to release_8 and trunk

PR #6285, opened 2023-08-30
Fix redirect loop when session expires

PurHur:
Hello friends of ilias,

i am here again to solve my favorite issue of all the time: https://mantis.ilias.de/view.php?id=31639

The problem is that there is a redirect loop also if the anonymous area is not enabled. An expired session forces the user to clear the cookies or to login again. Most other pages are not reachable if the redirect loop is triggered. 

I had to make some changes to the session handling and destroying expired session to make it work. Also the handling for the "Your session expired" message is changed.

This works for me but there could be some side effects. Someone with deeper initialisation/session knowledge should review this change. 

Greetings Purhur

### DIFF
--- a/Services/Authentication/classes/class.ilSession.php
+++ b/Services/Authentication/classes/class.ilSession.php
@@ -256,9 +256,6 @@ class ilSession
         try {
             // only delete session cookie if it is set in the current request
             if (isset($_COOKIE[session_name()]) && $_COOKIE[session_name()] === $a_session_id) {
-                //$params = session_get_cookie_params();
-               // setcookie(session_name(), '', 0, $params['path'], $params['domain'], $params['secure'], isset($params['httponly']));
-
                 \ilUtil::setCookie(session_name(), '', false,true);
             }
         } catch (\Throwable $e) {
--- a/Services/Authentication/classes/class.ilSession.php
+++ b/Services/Authentication/classes/class.ilSession.php
@@ -253,9 +253,17 @@ class ilSession
 
         $ilDB->manipulate($q);
 
-        $params = session_get_cookie_params();
-        setcookie(session_name(), '', 0, $params['path'], $params['domain'], $params['secure'], isset($params['httponly']));
-        
+        try {
+            // only delete session cookie if it is set in the current request
+            if (isset($_COOKIE[session_name()]) && $_COOKIE[session_name()] == $a_session_id) {
+                $params = session_get_cookie_params();
+                setcookie(session_name(), '', 0, $params['path'], $params['domain'], $params['secure'], isset($params['httponly']));
+            }
+        } catch (\Throwable $e) {
+            // ignore
+            // this is needed for "header already"  sent errors when the random cleanup of expired sessions is triggered
+        }
+
         return true;
     }
 
--- a/Services/Authentication/classes/class.ilSession.php
+++ b/Services/Authentication/classes/class.ilSession.php
@@ -256,7 +256,7 @@ class ilSession
         try {
             // only delete session cookie if it is set in the current request
             if (isset($_COOKIE[session_name()]) && $_COOKIE[session_name()] === $a_session_id) {
-                \ilUtil::setCookie(session_name(), '', false,true);
+                \ilUtil::setCookie(session_name(), '', false, true);
             }
         } catch (\Throwable $e) {
             // ignore
--- a/Services/Authentication/classes/class.ilSession.php
+++ b/Services/Authentication/classes/class.ilSession.php
@@ -255,9 +255,11 @@ class ilSession
 
         try {
             // only delete session cookie if it is set in the current request
-            if (isset($_COOKIE[session_name()]) && $_COOKIE[session_name()] == $a_session_id) {
-                $params = session_get_cookie_params();
-                setcookie(session_name(), '', 0, $params['path'], $params['domain'], $params['secure'], isset($params['httponly']));
+            if (isset($_COOKIE[session_name()]) && $_COOKIE[session_name()] === $a_session_id) {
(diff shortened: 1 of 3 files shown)


Services/Init/classes/class.ilStartUpGUI.php
```diff
@@ -265,7 +265,7 @@ protected function showLoginPage(ilPropertyFormGUI $form = null)
         $page_editor_html = $this->purgePlaceholders($page_editor_html);
 
         // check expired session and send message
-        if ($GLOBALS['DIC']['ilAuthSession']->isExpired()) {
+        if ($GLOBALS['DIC']['ilAuthSession']->isExpired() || $this->httpRequest->getQueryParams()['session_expired']) {
```
mjansenDatabay:
```suggestion
        if ($GLOBALS['DIC']['ilAuthSession']->isExpired() || ($this->httpRequest->getQueryParams()['session_expired'] ?? false)) {
```

Services/Init/classes/class.ilInitialisation.php
```diff
@@ -1407,6 +1406,8 @@ public static function resumeUserSession()
             !$GLOBALS['DIC']['ilAuthSession']->isAuthenticated() or
             $GLOBALS['DIC']['ilAuthSession']->isExpired()
         ) {
+            ilSession::_destroy($_COOKIE[session_name()],ilSession::SESSION_CLOSE_EXPIRE);
```
mjansenDatabay:
```suggestion
            ilSession::_destroy($_COOKIE[session_name()], ilSession::SESSION_CLOSE_EXPIRE);
```

Services/Authentication/classes/class.ilSession.php
```diff
@@ -252,7 +252,18 @@ public static function _destroy($a_session_id, $a_closing_context = null, $a_exp
         ilSessionIStorage::destroySession($a_session_id);
 
         $ilDB->manipulate($q);
-        
+
+        try {
+            // only delete session cookie if it is set in the current request
+            if (isset($_COOKIE[session_name()]) && $_COOKIE[session_name()] == $a_session_id) {
```
mjansenDatabay:
```suggestion
            if (isset($_COOKIE[session_name()]) && $_COOKIE[session_name()] === $a_session_id) {
```

Services/Authentication/classes/class.ilSession.php
```diff
@@ -252,7 +252,18 @@ public static function _destroy($a_session_id, $a_closing_context = null, $a_exp
...
+                $params = session_get_cookie_params();
+                setcookie(session_name(), '', 0, $params['path'], $params['domain'], $params['secure'], isset($params['httponly']));
```
mjansenDatabay: Is it possible to use `ilUtil::setCookie` instead?

Services/Init/classes/class.ilInitialisation.php
mjansenDatabay: Could we use `$DIC->http()->cookieJar()` instead (and for ILIAS >= 8.x `$DIC->http()->wrapper()->cookie()`, a PR for 8.x would be highly appreciated)?

Services/Authentication/classes/class.ilSession.php
mjansenDatabay: Could we use `$DIC->http()->cookieJar()` instead (and for ILIAS >= 8.x `$DIC->http()->wrapper()->cookie()`, a PR for 8.x would be highly appreciated)?

mjansenDatabay (changes requested):

Services/Init/classes/class.ilInitialisation.php
PurHur: I think i am too early for using that:

```
var_dump($DIC->http()->cookieJar()->get(session_name()),$_COOKIE[session_name()]);exit;
```
results in:

```
NULL string(26) "i85ho4batlfir49ijh17vkgo6t"
```

Services/Authentication/classes/class.ilSession.php
PurHur: See comment above

mjansenDatabay (approved):

Services/Authentication/classes/class.ilSession.php
```diff
@@ -252,7 +252,17 @@ public static function _destroy($a_session_id, $a_closing_context = null, $a_exp
         ilSessionIStorage::destroySession($a_session_id);
 
         $ilDB->manipulate($q);
-        
+
+        try {
+            // only delete session cookie if it is set in the current request
+            if (isset($_COOKIE[session_name()]) && $_COOKIE[session_name()] === $a_session_id) {
+                \ilUtil::setCookie(session_name(), '', false,true);
```
mjansenDatabay:
```suggestion
                \ilUtil::setCookie(session_name(), '', false, true);
```

Services/Init/classes/class.ilInitialisation.php
```diff
@@ -1407,6 +1406,8 @@ public static function resumeUserSession()
             !$GLOBALS['DIC']['ilAuthSession']->isAuthenticated() or
             $GLOBALS['DIC']['ilAuthSession']->isExpired()
         ) {
+            ilSession::_destroy($_COOKIE[session_name()], ilSession::SESSION_CLOSE_EXPIRE);
```
mjansenDatabay: @amichelsdatabay Do we need a `if (isset($_COOKIE[session_name()]))` check here?

PurHur: I added 
```
if ($GLOBALS['DIC']['ilAuthSession']->isExpired()) {
```

The cookie is always set if the session is expired because to expire the session it has to be looked up before in the database

PR #6287, opened 2023-08-30
Dashboard Feature: no dashboard sortation for one item

fhelfer:


iszmais:
@fhelfer Thank you very much for your contribution!

Unfortunatly, as far as i know, this doesnt disable the sorting when only 1 item is available, but when 1 group of items is available. E.g. when i have 2 Courses on the root node of the repository and sort by location these 2 are grouped inside date as 1 item since `$this->data` is an multidimensional Array.

I still want to be able to sort entries within such groups. The sorting shoul only be diabled when there ist only one item or less in any group.

iszmais (changes requested):
@fhelfer Thank you very much for your contribution!

Unfortunatly, as far as i know, this doesnt disable the sorting when only 1 item is available, but when 1 group of items is available. E.g. when i have 2 Courses on the root node of the repository and sort by location these 2 are grouped inside date as 1 item since `$this->data` is an multidimensional Array.

I still want to be able to sort entries within such groups. The sorting shoul only be diabled when there ist only one item or less in any group.

iszmais:
@fhelfer Thank you for the adjustment.
I rethought that and came to the conlusion that this may however not be a desired feature due to 2 concerning points.
- First its technical hard to specifiy since the data array is mostly very vague defined itself since it could be a n-dimensional array with a complete disjunkt content. There i wouldnt want to put additional time into defining it here as long as there is not strict rule for that.
- Second there might be a case where users want to predefine their sortings even if there are no items atm but to come. I would be contraintuitive to require at least 2 items to be able to change my standard sorting, especially if the user is not incrontroll of the shown items.

Therefore i would close this ticket.
We might open a FR for that in future wich may specify on a certain usecase and leaves addtional configruation options to tackle the second problem, but i dont see any prio for that right now.

Thank you for your effort and my apologize,
Ingmar Szmais

PR #6288, opened 2023-08-31
ks: main action for items

alex40724:
This adds a main action to KS items to provide a button for the most important action of an item.

This is needed to implement the feature https://docu.ilias.de/goto_docu_wiki_wpage_7444_1357.html

The tests are currently failing. I will fix them and add new ones as soon as we decide on the interface and the rendering.

I am sure there will be a more appropriate sass variable for defining the margin, currently it says
`margin-right: $il-padding-small-horizontal;` Any suggestions?

Bildschirmfoto 2023-08-31 um 12 04 12


src/UI/templates/default/Item/tpl.item_standard.html
```diff
@@ -36,7 +36,10 @@
 		<div class="media-body">
 			<!-- END lead_start_icon -->
 			<div class="il-item-title">{TITLE}</div>
+			<div class="il-item-actions" style="display: inline-block; float: right;">
```
catenglaender: This should work without the inline styles.

Alternatively, I would suggest using classes from the brand new layout tool "element-bar" (located in `templates/default/050-layout/_layout_element-bar.scss`), which is designed for exactly this kind of button bars. In this case you can remove all CSS but `float: right` from `.il-item-actions` in `_ui-component_item.scss`. Just assigning the classes should suffice to place them next to each other with the margin that other button bars have (or will have) as well:
```html
<div class="il-item-actions l-bar__container">
    <div class="l-bar__element">{MAIN_ACTION}</div>
    <div class="l-bar__element">{ACTIONS}</div>
</div>
```

templates/default/070-components/UI-framework/Item/_ui-component_item.scss
```diff
@@ -43,6 +43,14 @@ $il-item-gap-xs: 1px;
 		}
 	}
 
+	.il-item-actions {
+		display: inline-block;
```
catenglaender: As far as I know `float: right` forces `display: inline-block` to be rendered as `display: block` anyway, so I suspect that we can do without this line.

If we decide to use the element-bar layout tool for spacing the buttons, almost no CSS code is needed on `.il-item-actions`.

src/UI/templates/default/Item/tpl.item_standard.html
```diff
@@ -36,7 +36,10 @@
 		<div class="media-body">
 			<!-- END lead_start_icon -->
 			<div class="il-item-title">{TITLE}</div>
```
catenglaender, line 38: If the title is supposed to float around the actions, the title div has to come after the actions. This has been in the wrong order before this PR and is already broken in ILIAS 8.

templates/default/070-components/UI-framework/Item/_ui-component_item.scss
```diff
@@ -43,6 +43,14 @@ $il-item-gap-xs: 1px;
...
+		float: right;
+		.btn-default {
+			margin-right: $il-padding-small-horizontal;
```
catenglaender: This would also push the last button in the row to the left not aligning with the edge of the padding.

I would suggest either adding
```scss
&:last-child {
    margin-right: 0;
}

```
or using the element-bar as it handles the end of the line (and as a bonus also adds bottom margin for top rows should the bar ever stack over multiple lines... which is unlikely with just two buttons, but I like to cover these multi-line edge cases).

catenglaender (commented):
Awesome! @yvseiler and I were thinking about main action buttons on items and entities for a while now. That's a useful feature for prioritizing an action. Thank you very much.

Here you can find a few code comments from my html/css focused perspective. I am a bit nitpicky here considering that one comment is also talking about a flaw in the current implementation.

There is a brand new element-bar scss layout tool that could be quite useful for controlling the button spacing from a central pattern. For two buttons it's not really needed, but it might remove the need to add any scss code with this PR.

I am looking forward to having main action buttons like this in the item and other UI components, as I don't think the often most important action should be burried in a dropdown.

Amstutz:
Hi @alex40724

Thx a lot for this PR. As @catenglaender we agree that this is a very useful feature. 

Please or indicate why not:

- [x] Inline Styles: Please removed them, as suggested by @catenglaender in the comment.
- [x] Layout tool: Use the element-bar layout tool for spacing the buttons, as suggested by @catenglaender in the comment.
- [x] Margin right: Implement one of the suggested option by @catenglaender.

Please change

- [x] Tests: If accepted by JF, add the tests
- [x] Testrail: If accepted by JF, please add an according testrail case to update the Testcase for the Standard Item.

Thx
@amstutz, @thibsy, @klees

matthiaskunkel:
Jour Fixe, 18 SEP 2023 : We highly appreciate this suggestion and accept the PR for trunk. Please allow also Standard Links as action. Alexander will try to consider all comments made by Ferdinand and confirmed that he will add tests and a testcase as required by Timon.

klees:
@alex40724 Just want to make sure that we have no deadlock here. IMO the potato is in your hand, right?

alex40724:
@klees @thibsy All things mentioned by @Amstutz are implemented now. I added Standard Links as actions, too. The testrail case https://testrail.ilias.de//index.php?/cases/view/31882 has been updated accordingly. From my side this is ready for a final review.

thibsy (changes requested):
Hi @alex40724

Thx for implementing the requested changes.

Please consider the following suggestions. We do understand that suggestions are not common sense currently, but we still kindly ask you to help us keep our shared UI framework in a good shape. Hence, you do not need to follow those, but please indicate shortly why you prefer to do otherwise:

- [ ] `ItemTest` I: Please use a mock for the tests `test_with_main_action_button()` and `test_with_main_action_link()` instead of a real button and link instance.
- [ ] `ItemTest` II: We try to get rid of rendering tests where we compare the entire HTML to check if only a specific piece exists. Please refactor `test_main_action_button()` and `test_main_action_link()` similar to [`WrapperTest::testRenderWrapperWithAdditionalInputs()`](https://github.com/ILIAS-eLearning/ILIAS/blob/9f37d15a6736d66196bfc39e94e81c52958e1d1f/tests/UI/Component/Dropzone/File/WrapperTest.php#L69), which uses the canonical name of a mocked component to search for a hash with `str_contains()` to check if a specific HTML part will be rendered.

- [ ] `ItemTest` III: #6376 got merged in the meantime, which changed snake-case function names to camel-case. Please do the same for the newly added tests.
- [ ] Conflicts: please rebase onto the latest trunk to resolve the conflicts.

Kind regards,
@klees & @thibsy

alex40724:
Rebase went wrong.... Please check https://github.com/ILIAS-eLearning/ILIAS/pull/6460 instead.

PR #6289, opened 2023-08-31
Fix 37848: Remove return type

fwolf-ilias:
According to ilDclBaseRecordFieldModel::getValue which is the only source for this function argument the type is a mixed string|array but the object variable it could also be int|float|array|null (only as type hinting). 

Please check those hints and adjust them if necessary.


Amstutz:
thx a lot!

PR #6290, opened 2023-09-01
UI, Table: an OrderingTable to change position of records

nhaagen:
Screenshot from 2023-09-01 11-16-03

This introduces the Ordering Table. The Ordering Table is very close to the DataTable, but provides an input for each row to change the order of records (you may also drag rows around ;)). It is wrapped in a form and writes back the order of rows to its data binding.


klees (commented):
Hi @nhaagen,

this looks fine. Can you propose a description for the interface so we can move this on to @thibsy and @Amstutz?

Thanks!

klees:
@thibsy @Amstutz AFAIC we could take this to the JF. Would you be so kind and have a look into the interface?

tfamula:
Hi @nhaagen,
thank you for the PR, I'm really looking forward to this table :)

While checking my legacy tables, I stumbled over the following use case (and I guess I will not be the only one): Depending on permissions, a user must see the "Position"-column and the "Save Order"-button – or not. As far as I can see, it will not be possible to hide/disable the column and the button within the `OrderingTable`. So the only option would be to create two tables (`Ordering` and `Data`) in my code and show one of them in the GUI, depending on whether the user is allowed to change the order or not. Is this the way to go or are you planning to make the ordering in the `OrderingTable` optionally deactivatable?

Thanks
Thomas

klees:
Hi @tfamula,

thanks for bringing that use case. Seems to make sense, so I guess no one will object to an according addition. Doesn't seem to be very complicated as well. Could you propose that addition once the ordering table is included?

Kind regards!

thibsy (changes requested):
Hi @nhaagen

Thanks a lot for yet again another important contribution to the UI components =)!

Please note this review is only for the public interface yet, so we can already present this at the upcoming JF. I'll do a more detailed review to a later point.

I will use the letter `C` as an abbreviation for the `ILIAS\UI\Component\Table` namespace throughout the review.

- [ ] Submissions: So, if I understand this correctly, the `C\Ordering` table uses the same endpoint for rendering the table and processing submissions. This is due to the unchangeable form-action `#`. The storage process is also triggered implicitly, by rendering the table, which feels quite unintuitive. I think consumers should have better control over the processing of an `C\Ordering` table, because currently it is e.g. not possible to give any feedback to the client. Maybe consumers would need to perform some sanity-check before storing the table, which IMO should probably also not be implemented in the `C\OrderingBinding` as well. Using the same endpoint for this also results in resubmitting the form on every page-reload, after the form has been submitted once. And, consumers need to implement some `if ($request->getMethod() === 'POST') { ... }` check, in order to know if `withRequest()` should be called. IMO, there are two ways to move forward with this:
	- [ ] `I`: An `C\Ordering` table can be used similar to forms, by providing a post-url and having methods like `getData()`. I think it would even be possible to make `C\Ordering` extend `ILIAS\UI\Component\Input\Container\Container`. However, this may be too much for a first iteration, but we could also implement similar behaviour on the table itself. Anyways, getting rid of `C\OrderingBinding::withOrder()` and its implicit call.
	- [ ] `II`: An `C\Ordering` table can be used more like the `C\Data` table, which means, it can be provided with an endpoint that will be called asynchronously once the table is submitted. This way, we don't need to reload the page at all, since you have already implemented the mechanism to reorder the entries on the client. The endpoint can then e.g. return a message-box, which will be displayed in a modal, also allowing _some_ control over the processing. Now, personally, I think we should move towards option `I`, but for a first iteration this would also be fine.

- [ ] `static` return-type: I had to look up the [`static` return-type](https://php.watch/versions/8.0/static-return-type), because I was concerned this return-type could expose public methods of the implementing class. Luckily, this is not the case. Instead, it is a great improvement for fluent method calls, if they are scattered across different levels of the inheritance, but remain the same entity. Is there a reason you only changed `C\Data::withActions()` and `withRequest()` of this interface? Otherwise, I suggest to also change the other return-types from `self` to `static`. Same for:
	- [ ] `C\OrderingBinding::withOrder()`
	- [ ] `C\OrderingRow::withPosition()`
	- [x] `C\Row::withDisabledAction()`

- [ ] `withRequest()` and `withActions()`: these methods are currently implemented in both `C\Data` and `C\Ordering` tables, which is IMO redundant. How to get rid of this redundancy depends on whether `withRequest()` is coupled to `withActions()` or not. If thats true, we can introduce a `HasActions` interface, for both methods, otherwise we probably want a separate interface for `withRequest()`, like `HasRequest`.
- [x] `C\Row`: this interface suggests, it holds commonalities between all possible table-rows. However, this is not the case for `PresentationRow`. Now, since this interface is essentially a `C\DataRow`, please revert this change and make `C\OrderingRow` extend `C\DataRow` instead, so we can remove this interface.
- [x] `C\OrderingRow::withPosition()`: I think the argument name `$idx` is not concise enough, and maybe even misleading. Please rename it to something like `$position_index` and add a brief comment about how this method should be used.
- [ ] `C\OrderingBinding`: I think this interface follows the same concept as the `C\DataRetrieval`, right? If so, please also stick to the naming so the pattern is recognisable.
- [x] `C\Ordering`: there are three unused imports, please remove them.

If you have any questions regarding the submissions, feel free to contact me on Discord.

Kind regards,
@thibsy

oliversamoila:
Hello @nhaagen 
Thank you for this table variant.

* I have two more comments: [Save Order] should not be a primary action button in my opinion. The name of the table suggests that this is the central purpose, but there will also be use cases in which the ordering table is used, but the ordering is not primary, but simply functional or appropriate. Examples are the configuration of the main menu (since ILIAS 5.4) and the dashboard (since ILIAS 9).
* Could it be useful to explicitly point out in the description of the ordering table that the controls for selection (checkboxes) per line and actions per line are an COULD condition?

klees:
Hi @thibsy, hi @Amstutz,

thanks for your time to discuss your feedback. I'll try to document our results here, please feel free to correct me if required.

I will put **some general point regarding `withRequest`** here first, so this comment can be referenced and people can stop reading afterwards: It is intentional, that `withRequest` is neither injected via the constructor nor injected via the `Renderer`. The reason that we do not inject it via constructor is as such: There will be components, that produce components to be rendered by other components. `Membership`, e.g., could produce a member table to be rendered in the `Course`. To allow `Membership` to pass a request via constructor, it either would need to pull it out of thin air somehow (think: `$DIC`, static call) or the request would need to be passed down by the Course. The request would mostly become available in the controller of the page that is currently viewed and would need to be passed down, maybe through various layers, if we would require it in the constructor. Also, the table would then somehow be tainted with that contextual info from the request and we would not be able to use it in some other context. This is also the reason, why we do not insert the request in the renderer. Imagine that we use the table to create some export. The original request would, most likely, not be correct in that case. So the request is something the consumer should be able to control easily and also something where the consumer should be reminded that he indeed should handle that issue with care. Also, the request should, mostly stay in the top layers, close to the endpoint that a user called. The same argumentation works for the inputs and also, most likely, for other locations where we want to include the request.

Hence, regarding `HasRequest` and `HasActions`, creating interfaces is not the way to proceed, as we currently do not see that we will need to write code that can act on all components that have a request or actions.

Regarding the item *`Submissions`*: We discussed that there is another problem, besides the ones mentioned in the item. When the interactions is mediated via a form with POST where the result is rendered immediately, a reload of the page will lead to that irritating "form resubmission" message from the browser. This is a well known problem, which is solved by processing the POST and than redirecting, which does not happen for the ordering table currently. We did agree though, that the proposed solutions both are not what we want. The first solution would turn an implementation detail to something essential: While we currently use a form to submit the desired order to the server, this is not essential to the ordering table. Other implementations are well possible: async, some custom json protocol, .... So, an ordering table is not an Input Container, but something on its own. Async, on the other hand would require us to somehow actually reorder the table or refresh the table with some content for the server, which both is something we currently cannot do very well in a principled way. A good solution to this might be to have a second URL to receive the POST and a slightly updated version of `OrderingBinding` with something like `setOrder(): ?MessageBox\Failure` instead. We will look into this.

Regarding `OrderingBinding` we will stick to that name for the moment, as the class indeed expresses a two way binding between a storage and the table and hence is more then the one way retrieval for the table. We'll first look to the activities regarding the submissions will turn out.

That's it, I think @nhaagen can go on here now =)

Best wishes!

nhaagen:
Thank you all for discussing that, I rebased and added some changes.

>  [Save Order] should not be a primary action button in my opinion. The name of the table suggests that this is the central purpose, but there will also be use cases in which the ordering table is used, but the ordering is not primary, but simply functional or appropriate. Examples are the configuration of the main menu (since ILIAS 5.4) and the dashboard (since ILIAS 9).

It's a StandardButton, now.

> * Could it be useful to explicitly point out in the description of the ordering table that the controls for selection (checkboxes) per line and actions per line are an COULD condition?

I added a comment in the example; the actions have to be added explicitly, so I think this is rather obvious (and a MAY, if).
Actions have a full doc-entry on their own, too.


> it will not be possible to hide/disable the column and the button within the OrderingTable

There is now Ordering::withOrderingDisabled (see example)


> C\Row: this interface suggests, it holds commonalities between all possible table-rows. However, this is not the case for PresentationRow. Now, since this interface is essentially a C\DataRow, please revert this change and make C\OrderingRow extend C\DataRow instead, so we can remove this interface.

Changed back to OrderingRow extending DataRow; however, an OrderingRow now is a DataRow, too, so the Renderer has to distinguish by negative exclusion (or, worse, by the _order_ of if-clauses).
Ultimately, the PresentationRow should be a Row as well...

> static return-type: 
> C\OrderingBinding::withOrder()
> C\OrderingRow::withPosition()

Those are endemic to the ordering; I'd always prefer self over static. Same goes for the others; 
withRequest/withActions returns static because of the implementation in AbstractTable, only. 


> A good solution to this might be to have a second URL to receive the POST

I added Ordering::withTargetURL -
although not too refined, we might take it as a base for another iteration?

Best,
Nils

klees:
Hi @thibsy and @Amstutz,

I think we have addressed all items of your review. Would you be so kind and move this forward?

Kind regards!

klees:
Maybe we can take this to JF on monday...

thibsy (changes requested):
Hi @nhaagen

Thanks a lot for working on this, and sorry for the delay here.

I have already reviewed the interface and implementation at once, but until the interface passes JF you can ignore the requested changes regarding implementation details. The only important change for JF is the `OrderingBinding` one.

- [x] `OrderingBinding`: I don't think there is a benefit to have a bidirectional data-binding like this at the moment. This is mostly because we can't incorporate such a mechanism properly into our framework right now. I think we should get rid of `OrderingBinding::setOrder()` again and implement the data-processing similar to forms (post-URL in constructor, `withRequest()`, `getData()`). This way the processing will not be triggered implicitly and consumer have control over the feedback they wan't to give, along with eliminating the resubmission issue. You may add something about a proper bidirectional data-binding to the roadmap though - I think we can address this once the UI framework has its own endpoint.
- [x] `Column` functionality: the `OrderingTable` uses the same `Column`s as the `DataTable`. The problem is, that the column determines whether or not it is e.g. sortable or not, but the surrounding `Table` must be able to handle it. Currently all functionality (sortation, ordering-labels, initial and optional visibility) is silently ignored. Consumers should be able to expect similar results, even though they are used in different contexts. I hence ask you to implement all column-related functionality, except for the sortation, which you may add to this to the roadmap instead for this iteration. This is mostly because this functionality cannot simply be reused and must be implemented on the client separately.
- [x] JavaScript: you have implemented functionality which is specific to the ordering table inside the data table. Since the ordering table will most likely gain more functionality in the future, like the client-side column sortation, this would lay some groundwork for future developments. Please implement the ordering table as its own class in JavaScript, which may share some functionality with the data table. I know this would be much boilerplate code at the moment, so you can ignore this for the first iteration.
- [x] jQuery usage: you are [using jQuery to update input](https://github.com/ILIAS-eLearning/ILIAS/pull/6290/files#diff-fc1d41495a0f272721cef849138935636e7192b892597030ddf128a710896cd4R357) values. Please use vanilla JS if there is no good reason.
- [ ] `DataTable.#tmpDragRow`: I don't think we need this property. You can simply create a local variable inside `DataTable.#addDraglisteners()` and pass the row (state) as an argument to the event handlers. This way we get rid of any side effects.

- [ ] Row builders: since all concrete row builder implementations are basically identical, I believe we should only implement one (mighty) `RowBuilder`, implementing all the concrete row builder interfaces. Since there may be more row builders in the future, we should stick to a naming like `build<XY>Row()` on the interfaces though.
- [x] `DataTable.#isDraggedElementValidRow()`: please use an `instanceof` check instead of the first three comparisons in the return statement. Use `HTMLElement.ownerDocument.defaultView.HTMLTableRowElement` for testability.
- [x] Sortable rows: please store all sortable rows in a new property instead of creating an array from `DataTable.#table.rows` and removing the first element on each drag-over.
- [x] Examples: please use methods from the `$DIC` instead of array-access, or provide `@var` annotations for proper autocompletion.
- [x] `DataRowBuilder::buildRow()`: this method is never used, please remove it.
- [ ] `AbstractTable`: please rename this class to `Table`.

Kind regards,
@thibsy

nhaagen:
Hi @thibsy ,
I think I addressed the main suggestions and issues;

> [ ] DataTable.#tmpDragRow: I don't think we need this property. 

I did not find a vanilla way to pass a var through events

> [ ] Row builders: 

Having but one RowBuilder would enable the consumer to return mixed row types from getRows; I really would not like to lose the current strictness.

> [ ]  AbstractTable: please rename this class to Table.

Sadly, the PresentationTable is still in the way. I'd really like to move everthing from AbstractTable to Table, once the data binding and View Controls are aligned

Kind regards,
Nils

thibsy:
Thx a lot @nhaagen for implementing the changes! I will add the JF label so we can discuss the interface on Monday.

matthiaskunkel:
Jour Fixe, 13 MAY 2024: Richard presented us the current implementation of the UI Ordering Table. We highly appreciate this suggestion and accept the PR for trunk.

thibsy (approved):
Thx a lot for working on this @nhaagen! This LGTM so I will merge this now.

PR #6291, opened 2023-09-01
Course & Group: small performance improvements for members view

schmitz-ilias:
This PR improves the performance of the members view in Courses and Groups with many members by reducing the number of calls to `ilObjUser::_lookupName`, `ilObjUser::_lookupLogin` and `ilObjUser::_lookupActive`, see [31394](https://mantis.ilias.de/view.php?id=31394). I wouldn't expect any major leaps, but the changes here should at least mitigate problem.

Let me know if you prefer a second PR for R8 and trunk.


smeyer-ilias (approved):
Thanks @schmitz-ilias

PR #6293, opened 2023-09-04
Test: Don't use $this in static context

mjansenDatabay:
Test: https://mantis.ilias.de/view.php?id=37867


kergomard:
Thank you very much  @mjansenDatabay ! It is absolutely obvious that I shouldn't have done that ;-) !

PR #6294, opened 2023-09-04
0037852: bugfix - Typed property ilDclBaseRecordModel::$last_edit_by must not be accessed before initialization

msalterhslu:


Amstutz:
thx a lot.

PR #468, opened 2017-03-27
UI Rating Input

nhaagen:
Proposal to add a new input for rating things.
Uses, but not introduces input-facory (which is in PR #360).
The implementation is incomplete, mainly backing the proposal's examples.

PR #6295, opened 2023-09-04
UI/Input/Rating: a five-star rating input

nhaagen:
This introduces a Five-Star-Rating as UI FormInput.

Screenshot from 2023-09-04 12-07-11

https://mantis.ilias.de/view.php?id=33588


src/UI/Component/Input/Field/Factory.php
```diff
@@ -756,4 +756,39 @@ public function colorPicker(string $label, ?string $byline = null): ColorPicker;
...
+     *      The Rating Input is presented in form of the known five-star-concept,
+     *      where users can select a specific number of stars to indicate their
+     *      level of satisfaction or preference, where more stars generally signify
+     *      a higher rating.
+     *   composition: >
+     *      Internally, the input consist of radios; visually, the user is presented with
+     *      five stars to select. A preceding text may be used to describe the issue to be rated.
+     *   effect: >
+     *      When clicking a star it is highlighted as well as all prior/lower stars.
+     *      If options are labeled, the according label is shown next to th input.
```
klees:
```suggestion
     *      If options are labeled, the according label is shown next to the input.
```

klees (commented):
Hi @nhaagen,

also internal review for this one. Thanks for moving this over to the core...

- [ ] usage rule: What would be a use case to change the wording for the individual options? Can't we always move the problem over to the label of the input and the question and name the options from good to bad?
 
- [x] Enum: Please introduce an enum for the value contained in the input field.
- [x] Question-Aria: Please look if there is an appropriate way to bind the question to the input for screenreaders, probably with some aria-label.

- [x] HTML: Please adjust the HTML according to [this docu](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio). Looks as this might also provide a solution for my suggestion regarding the question.

Feel free to contact me for a discussion, as always.

Best regards!

klees (approved):
LGTM! @thibsy/@Amstutz, would you be so kind to carry on with this?

Amstutz:
Hi @nhaagen

Thx a lot for this PR. Special thx for providing the rendering tests (which helped me a lot to evaluate this from an a11y point of view) and for the detailed description of the component. Further thx for using standard HTML radio controls along with field sets, which make this control a lot more accessible than earlier controls.

Questions:
- [x] Currently we have a similar legacy control for ratings that almost could be replaced by this. However, the current control allows to also display an average of all votes in the past (a small blue dot). Would this be feasible to add here? This would allow replacing current implementations a lot quicker.
- [x] Reset Option: You implemented the reset option as further Radio Input. I see how this works nicely for required inputs if none should be set and having a double function for also resetting this. However, I wonder if “reset” is really what is going on here. Should we not keep this a neutral option and name it that way („No rating given“) or similar. If we do so, I wonder if we should even use some kind of “Reset-Glyph” here. How about just a neutral Radio Option (HTML default)?

Please Change
- [x] Naming I: `withQuestionText` should be renamed to `withAdditionalText()` and provided with a description on the public interface. Please also change the return-type to `static`.
- [x] Naming II: Rename "FiveStarRatingScale" to "Five-Star Rating Scale" in the component's description. 
- [x] Remove @inheritdoc comments
- [x] Comments: remove unnecessary comments like "this implements interface XY" or "this is what interface XY looks like“
- [x] Add Accessibility Rule: Something like "Screenreaders MUST have a way of accessing, a) the currently selected option, b) the amount of total options and c) the control to reset the selection.“ Note that this already works with the current implementation. However, I would like to see this stated as a rule, to make sure we keep this working in future refactorings.

Please consider the following changes or indicate, why rather not:
- [x] Naming III: `getEnumTrafo()` should be renamed to `getFiveStarRatingScaleTransformation()`.
- [x] Return Statement: in `getEnumTrafo`, [this return statement](https://github.com/ILIAS-eLearning/ILIAS/pull/6295/files#diff-067f067ba84fe036ba032cd54f75fcadcf4fb096f96132d4cae42fbb3d1a5256R55) should be outside the if statement; otherwise the transformation is invalid for integers and does not have a default return.
- [x] `FiveStarRatingScale`: We think the gaps between the severities are not clear enough. We suggest the following scale instead: `terrible`, `bad`, `average`, `good` and `excellent`.

Thx a lot
@Amstutz and @thibsy

nhaagen:
Screenshot from 2023-12-05 14-24-35

Amstutz:
Hi @nhaagen
 
Thx a lot for the huge amount of work you already put into this. A further review together with @yvseiler led to the following three point. Please adapt or indicate, why not:
 
- [x] Description: The code says "The selection can be reset with a small "x" next to the input." (components/ILIAS/UI/src/Component/Input/Field/Factory.php). This is outdated, correct? If so, the description should be gone/adapted.
- [x] Neutral Option: Don't we have cases, when we do not need/want this? Could this be optional? I can think of cases, when neutral, must be avoided (if I want a rating for EVERY input given).
- [x] !important: There are some in the SCSS? Can you check with @catenglaender if we might get rid of them?

Please change:

- [x] Magic numbers I: you have introduced some magic numbers [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6295/files#diff-a1ef4e5bb04ac3db9bb200ac63f85bcfd5bd5c43c4c9315588948d047ae0ed63R102), please remove them by using values from the `FiveStarRatingScale`.
- [x] Magic numbers II: there are some more magic numbers [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6295/files#diff-ef03ec89ebf29bde4d4f7277b852dd624870adbb45d3d1f08a36c630a4cab2edR1095), please remove `5` by `count(FiveStarRatingScale::cases())` and `100` with a constant like `PERCENTAGE_FACTOR`.

Kindly

@amstutz & @thibsy

nhaagen:
Hi @Amstutz and @thibsy ,
I removed the neutral option for "required" - I think, that's what is meant in most cases and it makes only little sense to force the user to give a "neutral" statement. Are you OK with that?

templates/default/070-components/UI-framework/Input/_ui-component_rating.scss
```diff
@@ -0,0 +1,95 @@
...
+}
+.il-input-rating .glyphicon-star:before {
+  content: "\2605";
+}
+.il-input-rating .glyphicon-star-empty:before {
+  content: "\2606";
+}
+
+.il-input-rating .il-input-rating-star {
+  float: right;
```
catenglaender: Instead of floats, maybe we could use the SCSS tool element-bar to place the stars next to each other. Advantage: They would be perfectly vertically centered.

templates/default/070-components/UI-framework/Input/_ui-component_rating.scss
```diff
@@ -0,0 +1,95 @@
...
+  position: relative;
+  display: block;
+  margin-top: -3px;
```
catenglaender: Ah, the old ILIAS glyph positioning fun. Again, I wonder if the SCSS element-bar tool would be a better solution that doesn't need correcting.

templates/default/070-components/UI-framework/Input/_ui-component_rating.scss
```diff
@@ -0,0 +1,95 @@
...
+  white-space:nowrap;
+}
+.il-input-rating .il-input-rating__options {
+    float: left;
+}
+
+.il-input-rating input.il-input-rating-scaleoption {
+  float: right;
+  padding: 0;
+  margin: 0 0 0 -20px;
```
catenglaender: Would a margin variable from the settings layer work here?

templates/default/070-components/UI-framework/Input/_ui-component_rating.scss
```diff
@@ -0,0 +1,95 @@
+@use "../../../010-settings/" as *;
+
+.il-input-rating .il-input-rating__text {
```
catenglaender: Maybe some nesting could improve the overall structure and readability and also give us a chance to really use the BEMIT naming convention:
```SCSS
.c-input-rating {
    &__text {
    }
    &__stars {
    }
    &__star {
        &--empty {
        }
    }
}
```

templates/default/070-components/UI-framework/Input/_ui-component_rating.scss
```diff
@@ -0,0 +1,95 @@
...
+  .il-input-rating__none {
+    display: none;
+  }
+}
+
+.il-input-rating-scaleoption:checked + label + span {
+  display: block;
+}
+
+/* targets hovering in IE10+ */
```
catenglaender: Do we still need the IE10 hacks? Edge should not need this.

catenglaender (commented):
Some thoughts on how to maybe get the SCSS a little bit more tidy. @klees shall I take care of this?

nhaagen:
Hi @catenglaender ,
please, feel free to tackle the CSS-parts! 
However, I'm unsure about the tool element-bar, since the rating revereses the order of options in order to use the ~ operator to select/hover previous stars and avoid using JS. Does that still work with the bar?

Amstutz:
Thx a lot. I had a short chat with @nhaagen . We both believe that the CSS might need some further adjustements. However, we merge this, since the rest is good. Note that we might need to factor out the glyphs part, once glyphs are available as symbol only:

image


thx @nhaagen for you work!

alex40724:
@Amstutz Is there any Jour Fixe decision on this, maybe in another PR? Could you please point me to this?

alex40724:
If so, you may remove the JF label.

Amstutz:
Hi @alex40724 

Thx a lot for checking up on this. I was sure, that this was only the implementation of a concept we decided upon at the JF some time ago. However, there seems only this old PR without JF decision: https://github.com/ILIAS-eLearning/ILIAS/pull/468.

In short, my mistake and sorry. Should not have happened and thx for adding the JF label. Nice to see that the proesses here are well established.

matthiaskunkel:
Jour Fixe, 04 MAR 2024: We highly appreciate this suggestion and accept the PR for trunk.

PR #6302, opened 2023-09-08
ILIAS 8: Increase max length of redirect_url for tst redirection to 256

mBeym:
Increases the length for the **redirection_url** column for tests.

This is used for the "Redirection" after a test is completed by a user.

because of the nature of ILIAS having a lot of query parameters (and very long ones), it's very easy to reach the current limit of 128 characters when a link like `ilias.php?cmdClass...&baseClass...&cmd...&Id...` is saved.

### DIFF
--- a/Modules/Test/Setup/class.ilTest8DBUpdateSteps.php
+++ b/Modules/Test/Setup/class.ilTest8DBUpdateSteps.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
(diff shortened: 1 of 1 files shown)


kergomard:
For a review see: https://github.com/ILIAS-eLearning/ILIAS/pull/6303

kergomard:
Thank you very much @mBeym !

I created[ an issue in mantis](https://mantis.ilias.de/view.php?id=37948) and will take care of this by simply making sure urls are not silently cut. I would close these PRs as to me it seems like more work to amend them, then to fix it separately.

Thanks for bringing this to my attention.

Best,
@kergomard

PR #6303, opened 2023-09-08
ILIAS Trunk: Increase max length of redirect_url for tst redirection to 256

mBeym:
Increases the length for the **redirection_url** column for tests.

This is used for the "Redirection" after a test is completed by a user.

because of the nature of ILIAS having a lot of query parameters (and very long ones), it's very easy to reach the current limit of 128 characters when a link like `ilias.php?cmdClass...&baseClass...&cmd...&Id...` is saved.

### DIFF
--- a/Modules/Test/classes/Setup/class.ilTest9DBUpdateSteps.php
+++ b/Modules/Test/classes/Setup/class.ilTest9DBUpdateSteps.php
@@ -162,4 +162,16 @@ class ilTest9DBUpdateSteps implements ilDatabaseUpdateSteps
             $this->db->dropTable('tst_seq_qst_postponed');
         }
     }
+
+    public function step_8(): void
+    {
+        if ($this->db->tableExists('tst_tests') && $this->db->tableColumnExists('tst_tests', 'redirection_url')) {
+            $this->db->modifyTableColumn('tst_tests', 'redirection_url', [
+                'type' => 'text',
+                'length' => 256,
+                'notnull' => false,
+                'default' => null
+            ]);
+        }
+    }
 }


kergomard:
Hi @mBeym

Thank you for this PR.

This short review is for all three versions:
* https://github.com/ILIAS-eLearning/ILIAS/pull/6302
* https://github.com/ILIAS-eLearning/ILIAS/pull/6301

Could you please provide an Issue-Report in Mantis?

Ok, I don't think your rational is actually sound (the URL you are showing above would be invalidated with every update and for ILIAS links you would always need to use the very much shorter permalinks), but you are actually right, that 128 characters are a completely random number and simply cutting them is a real issue.

- [ ] Could you please explain how 256 is any less random than the 128 Characters? A little bit of looking around informed me, that the limit for URL length for modern browsers is around 64KB of data. Couldn't we find a principled number for this?
- [ ] Could we drop this for ILIAS7 and only make this change for ILIAS8 and above?

- [ ] If we update this, we should actually fix the whole input. I checked for ILIAS7, there the input is simply cut when saving without any information to the user. Could you please make sure the user is informed, when she enters an uri that is too long at form validation time and make sure that the string is not silently cut.
- [ ] 
Thanks again and best,
@kergomard

mBeym:
@kergomard 

Agree about my example not being correct. I forgot about that and was made aware of that through discord as well.
For that reason. At least I don'T need this change anymore as a goto link is probably going to always fit.

1. The 256 was also just a random number but considering that this "issue" hasn't come up so far, I think having double of what's currently possible should be enough...
2. I needed this for ILIAS 7 but as I don't need it myself anymore... I don't have a problem with that.
3. I can do that if I should continue with this change (as again, I don't need it, nobody else has needed it so far so....)

so in my opinion the pull requests could also be closed...

kergomard:
Thank you very much @mBeym !

I created[ an issue in mantis](https://mantis.ilias.de/view.php?id=37948) and will take care of this by simply making sure urls are not silently cut. I would close these PRs as to me it seems like more work to amend them, then to fix it separately.

Thanks for bringing this to my attention.

Best,
@kergomard

PR #3394, opened 2021-05-17
Fix and Improve OpenID Connect integration

JuliusPC:
The current OIDC integration has three major flaws:

1. It uses the Implicit Flow instead of the Code Flow. The Implicit Flow should be considered deprecated (and will be in OAuth 2.1) and provides no relevant advantages over using the Code Flow, especially when the application is a confidential client (which ILIAS is, since the code is run on the server and not on the client). Since I couldn't find any comment why Implicit Flow is used, I think it was introduced by copying example code from the library's documentation.
2. The Logout is totally broken and likely never worked. There is no need to pass any Access Token to the Endsession Endpoint. Instead, you just need to provide the ID Token. This may be caused by [incorrect parameter naming in the used library](https://github.com/JuliusPC/OpenID-Connect-PHP/commit/b4eacdb384bc39fedf04c4dd16faedf3ba1668f3).
3. The used library is poorly maintained and has [multiple flaws](https://github.com/JuliusPC/OpenID-Connect-PHP/wiki/Progress-on-fixing-upstream-issues). Therefore, I forked the library and included it in this PR.

All points above are fixed by this PR.

The first and the last change allow to use PKCE if the OpenID Provider supports it.

PR #6305, opened 2023-09-11
0037842: Fix OIDC Logout and remove implicit workflow

pascalseeland:
This PR takes the code changes from https://github.com/ILIAS-eLearning/ILIAS/pull/3394 without the switch of the used library.

### DIFF
--- a/Services/OpenIdConnect/classes/class.ilAuthProviderOpenIdConnect.php
+++ b/Services/OpenIdConnect/classes/class.ilAuthProviderOpenIdConnect.php
@@ -12,6 +12,7 @@ use Jumbojett\OpenIDConnectClient;
  */
 class ilAuthProviderOpenIdConnect extends ilAuthProvider implements ilAuthProviderInterface
 {
+    const OIDC_AUTH_IDTOKEN = "oidc_auth_idtoken";
     /**
      * @var ilOpenIdConnectSettings|null
      */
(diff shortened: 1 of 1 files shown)


Services/OpenIdConnect/classes/class.ilAuthProviderOpenIdConnect.php
```diff
@@ -41,16 +42,21 @@ public function handleLogout()
...
 
-        $auth_token = ilSession::get('oidc_auth_token');
-        $this->getLogger()->debug('Using token: ' . $auth_token);
+        $id_token = ilSession::get(self::OIDC_AUTH_IDTOKEN);
+        $this->getLogger()->debug('Logging out with token: ' . $id_token);
 
-        if (strlen($auth_token)) {
-            ilSession::set('oidc_auth_token', '');
+
+        if (strlen($id_token)) {
```
mjansenDatabay: We are reading this from the session storage, so it might theoretically be `null`.

```suggestion
        if (is_string($id_token) && $id_token !== '') {
```

Services/OpenIdConnect/classes/class.ilAuthProviderOpenIdConnect.php
```diff
@@ -108,7 +105,7 @@ public function doAuthentication(\ilAuthStatus $status)
             $_GET['target'] = (string) $this->getCredentials()->getRedirectionTarget();
 
             if ($this->settings->getLogoutScope() == ilOpenIdConnectSettings::LOGOUT_SCOPE_GLOBAL) {
-                $token = $oidc->requestClientCredentialsToken();
+                $token = $oidc->requestUserInfo();
                 ilSession::set('oidc_auth_token', $token->access_token);
```
mjansenDatabay: We are still storing the "Access Token" here. In the changed code above, you renamed the variable to `$id_token`, 
 and "ID Token" != "Access  Token". We should use the correct wording.

You used a different implementation in https://github.com/ILIAS-eLearning/ILIAS/pull/6306/files#diff-e62021cd52eaff709b9d78dfcd64f416d3f31b24915a119cfb21763f0050203cR110

mjansenDatabay (changes requested):

matthiaskunkel:
Jour Fixe, 18 SEP 2023 : We highly appreciate this suggestion and accept the PR for all maintained versions. Please add a notice to the changed behaviour section of the release notes and send a related notification in the admin mailing list as well as in Discord.

Services/OpenIdConnect/classes/class.ilAuthProviderOpenIdConnect.php
pascalseeland: Sorry,  I messed up c&p from release_8 branch.

mjansenDatabay (approved):

PR #6307, opened 2023-09-11
UI/Link/Bulky: 37135, add lang attribute to label

nhaagen:
https://mantis.ilias.de/view.php?id=37135

### DIFF
--- a/components/ILIAS/Form/classes/class.ilSelectInputGUI.php
+++ b/components/ILIAS/Form/classes/class.ilSelectInputGUI.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 /**
  * This class represents a selection list property in a property form.
--- a/components/ILIAS/Form/classes/class.ilSelectInputGUI.php
+++ b/components/ILIAS/Form/classes/class.ilSelectInputGUI.php
@@ -27,6 +27,7 @@ class ilSelectInputGUI extends ilSubEnabledFormPropertyGUI implements ilTableFil
 {
     protected array $cust_attr = array();
     protected array $options = array();
+    protected ?\Closure $langresolve = null;
     /**
      * @var string|array
      */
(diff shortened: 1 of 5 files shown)


klees (changes requested):
Hi @nhaagen,

this most likely will do the right thing, but I'm not convinced that this is indeed the correct solution. 

For `ilSelectInputGUI`: I currently wonder why we should pass a closure that only ever is set to a one value. Also, where is `getOptionsLangAttribute` actually called? I would suggest to move the changes that are not related to the UI-framework into their own PR. Depending on @kergomard|s plans with `Services/User` we even might not want to do these changes now, as we will be switching the form to the UI framework someday soon. (any comments, @kergomard?)

Regarding the UI-framework part: I do not think that it is quite correct to use the language of the target resource as the language of the label. These are not the same. We seem to be on the verge to require a "text in language xy" here instead of just a string. Will think about this.

Unrelated, but curious still: I wonder why these thingies are even `links`. Is there really another `resource` they link to? Will think about this too...

Call to action will follow, maybe after  @kergomard|s input... =)

Kind regards!

klees:
Hi @nhaagen,

I looked into this today with @Amstutz a little. So, we introduced [`HasContentLanguage`](src/UI/Component/HasContentLanguage.php) already, and it is already implemented in `Link` and seems to work properly. So no changes in the UI components necessary from my POV.

I'll restructure commits accordingly...

Kind regards!

klees:
Hi @chfsx, @kergomard,

thanks to @Amstutz we figured this out. I assigned you both because ticket is a little broad... Please ping me if I can be of any help...

Kind regards!

kergomard (approved):
Changes in User LGTM! Thank you very much!

I very much look forward to switching these forms to UI-Inputs, but this just depends on anybody providing funding, as personally I set my priorities on `Services\Object` first and will put my energy to find the resources there first.

I agree that the `Closure` might be a little bit overengineered for a temporary solution, but this, I leave up to you ;-) .

Best,
@kergomard

chfsx:
Changes LGTM as well, thanks! failing tests don't seem to be related with this changes, right?

klees:
@chfsx: I don't see how we would have made installation of node.js fail...

thibsy:
Node.js seems to be working again, the checks now fail due to copyright notices.

nhaagen:
Just found this molding around - should be good to go, now ;)

kergomard:
Ok, I de assign me here. You do have my green light to merge the changes in `User`.

Thanks and best,
@kergomard

kergomard:
I merge this now, otherwise this will be rotting here forever. Will backpick to ILIAS 9. There is no easy way to pick it back to ILIAS 8 as far as I can see. If we would like to pick it back to ILIAS 8 we would need to find another solution.

Best,
@kergomard

PR #6309, opened 2023-09-11
UI/Modals: use dialog element for modal and introduce a Response Component

nhaagen:
- structure async response
- vanilla JS to replace modal parts
- promote Forms w/o buttons to factory


klees (commented):
Hi @nhaagen,

thanks a lot for working on this. Please understand that this is "only" my review, as I agreed with @Amstutz and @thibsy that we do CaT-QS for our contributions before I pass them on to them. You will get feedback from them once I pass this on.

- [ ] `Horizontal`: Why should a `Horizontal Divider` be [content of a `Modal`](src/UI/Component/Divider/Horizontal.php)?	
- [ ] `Link`: Same for `Link`. Why would we want to have a singular `Link` in a `Modal`?
- [x] `withAsyncUrl`: Do we really want to be able to exchange the URL for an already existing modal? Would imagine this to be a possible source of confusion, since we would need to update client side signals for that modal as well?
- [ ] Design: Could you argue a little why you have separated `ModalContent` from `ModalResponse`. I think this also partly answers my two first questions... This, to me, looks like a way to build generic modal content from lists of UI components. I see at least one other possible point in our design-space: That we put the required methods form `ModalResponse` on `ModalContent` (the getters, basically). This would mean that we can't have `Horizontal` and `Link` on their own but need more complete components to be displayed in the modal. From my perspective the latter point (no separate response/content) would be preferable, because it shifts attention on building UI components that really can be displayed in a modal as a whole. The current approach would remove control about what "really is" in the modal from the UI framework and our processes. I might be missing something, hence I'm interested in your rationale for the current proposal.

- [x] Wording: I think we should use that "Dialog" wording for the complete shebang instead of "Modal". "DialogResponse", ... After all, "Dialog" and "Modal" will be two different worlds, basically, right? 

I guess we want to have one iteration on the interface, the rest seems to be fine. Maybe we could look into this together on wednesday.

Best regards!

nhaagen:
Hi @klees,
thanks for looking through this. I'll start from the bottom, and reach a little into the history of this PR.

With the current modals, there are some shortcomings that imo make using them quite difficult; there is, first of all, the pre-definition of contents. 
That often leads to several modals on one page (quite horrible in table-actions...) or inflexibility in the contents, e.g. when confirming a selection.
Of course, there are asynchronous calls, but the current implementation will not work well with multiple steps or e.g. errors in forms, since the modal is re-rendered completely and context (ids, js) is lost. Control of Buttons is always hacky.

This is why the layer in the frontend and the reponse in the backend are separated. They form the "mechanics" of this Modal without telling anything about the actual contents.
Async is default (and currently the only way), but you get proper endpoints/params, and stay in context; I'd encourage you to have a live-look at the examples, forms and completely changing contents feel quite nice, imho.

Looking into the allowed contents of a Modal, there is the not to heavy ModalContent-interface; in general, I think this is a good, if not even the correct way to flag Components as possible parts of others. 
However, it is perfectly arguable whether a Divider or Link should implement it.

As said, I'm not (yet) too concerned with the actual contents, and just got a few components I've regulary seen to be used within Modals (and to showcase possibilites, of course).
The various usecases for a modal (confirmation, additional input, showing details) lead to an equal variety of needed "high order content components", and the UI-typical question between appreciated restriction and desired flexibility.
I totally agree that there should be some 'pre-fabricated' responses, like confirmation on selected items, and would suggest an according factory, like, modal()->response()->confirmation().
I'd still stick to content-interface for the lower components, since even there not everything should be allowed. Maybe it's just a personal whim, but, joining Response and Content also feels like mingling domains of content and presentation layer.

As for the naming, I guess that you are right and there will be coexisting worlds, so I'll switch to Dialog... overall.
The "withAsync" was the idea of re-using the modal, but it turns out that is hardly a case; I'll remove the method.

Looking forward to where this is going, 
cheers, 
Nils

klees (changes requested):
Hi @nhaagen!

Thanks for the elaborate answer and the changes.

Regarding `DialogContent` and `DialogReponse`. As often, I will argue for restricting the content of the `Dialog` more than it is currently the case. Currently I could build very odd constructs (multiple `Horizontal Dividers` following each other...) because of the list of `DialogContent` the `Response` will accept. This will also be bad in the long run. Imaging, e.g., we would allow `Form Input Container` as `DialogContent`. We would then allow multiple `Form`s in one `Dialog Modal` via interface, which would be extremely odd. We would need to disallow this by some text in the KS description then, which people will break. And then there are two save buttons in one modal?

So we should move on like this:

* `DialogContent` flags components, that can be a complete content of a `Dialog Modal`. Hence, `Dialog Response` need to accept one of these.
* Every `Dialog Content` should be able to describe itself enough to be displayed in the `Dialog Modal`. Hence I think we need this as interface:

```php
interface DialogContent
{
	public function getDialogTitle(): string;
	public function getDialogButtons(): array;
	
}
```

* For the `DialogResponse` we would need two variants, as far as I understand: One that simply closes the modal, one that contains one `DialogContent`. This, from my POV, would be something for the `Factory`. I'm not sure that we would need anything on a public interface of a `DialogResponse`, so I currently suspect that one interface for the response would be fine.

Then, I wonder why we need `DialogResponse::getCloseButton`. We always seem to append it to the list of `Buttons` somewhere in the `Renderer`. What am I missing?

Shall we discuss this a little via VC?

Kind regards!

nhaagen:
Sorry, accidentally deleted the branch

PR #6312, opened 2023-09-12
LearningModule: Unneeded Setting of Offline-Status

kergomard:
Hi @alex40724 

See: https://mantis.ilias.de/view.php?id=37820

This call is unnecessary. In cases it absolutely needs to be set to ONLINE, this should be done after creation of the object.

Best,
@kergomard

### DIFF
--- a/Modules/LearningModule/classes/class.ilObjContentObject.php
+++ b/Modules/LearningModule/classes/class.ilObjContentObject.php
@@ -103,7 +103,6 @@ class ilObjContentObject extends ilObject
     public function create(
         bool $a_no_meta_data = false
     ): int {
-        $this->setOfflineStatus(true);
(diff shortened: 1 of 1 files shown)


alex40724:
@kergomard: Sorry if I disagree here. This is an issue of the setOfflineStatus method. It must not lookup things, this is not what a setter is supposed to do. This might lead to all kind of other issues. Usually you should be able to call setters without having persistent data before.

kergomard:
Just to be clear: The setter has been fixed (https://github.com/ILIAS-eLearning/ILIAS/commit/97c9f0ecad460213318f7ab4275bc9b3b0654aa5). This setter has to go though and it will (it is just wrong on every level). As the line was removed otherwise, I close this PR.

PR #6313, opened 2023-09-12
DC, Active Record, fix of #37935

Amstutz:
See Ticket https://mantis.ilias.de/view.php?id=37935

PHP 8.2 issue with Active Record


chfsx:
thanks a lot!

PR #6314, opened 2023-09-12
`Refinery`: Transformation for escaping html/css/js

lscharmer:
This PR introduces 5 new transformations to escape: Javascript, HTML, HTML attributes, CSS and URL's.
The implementation is based upon this repository: https://github.com/laminas/laminas-escaper/

### DIFF
--- a/components/ILIAS/Refinery/src/Encode/Transformation/HTMLAttributeValue.php
+++ b/components/ILIAS/Refinery/src/Encode/Transformation/HTMLAttributeValue.php
@@ -1,0 +1,107 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\Refinery\Encode\Transformation;
+
+use ILIAS\Refinery\Transformation;
+use ILIAS\Refinery\DeriveInvokeFromTransform;
+use ILIAS\Refinery\DeriveApplyToFromTransform;
+use ValueError;
+
+/**
+ * Inspired by: Laminas escaper: https://github.com/laminas/laminas-escaper.
+ * This class expects a valid UTF-8 string. Conversion between encodings is not the responsibility of this class.
+ *
+ * This class encodes a given string so it can be used as a value inside a HTML attribute tag (e.g. <div class="ENCODED_VALUE"></div>).
+ * The given string is encoded so it can be used in any HTML Tag regardless if the value is enclosed in double quotes ("), single quotes (') or no quotes at all (e.g. <div class=ENCODED_VALUE></div>).
+ * Please see for more information of the syntax of HTML attributes: https://html.spec.whatwg.org/multipage/syntax.html#attributes-2
+ * Please see https://html.spec.whatwg.org/multipage/dom.html#phrasing-content for a list of all allowed characters and escape sequences.
+ */
+class HTMLAttributeValue implements Transformation
+{
+    use DeriveInvokeFromTransform;
+    use DeriveApplyToFromTransform;
+
+    public function transform($from)
+    {
+        return $this->encode($from);
+    }
+
+    private function encode(string $from): string
+    {
+        return preg_replace_callback(
+            '/[^a-z0-9,._-]/iSu',
+            fn(array $m): string => $this->replace($m[0]),
+            $from
+        ) ?? throw new ValueError('Invalid UTF-8 string given.');
+    }
+
+    private function replace(string $utf8_char): string
+    {
+        $codepoint = $this->utf8CharacterToCodepoint($utf8_char);
+
+        // All unicode control characters besides white space codepoints as well as noncharacters are not allowed in HTML attributes.
+        if ($this->isNonPrintableControl($utf8_char, $codepoint) || $this->isNonCharacter($codepoint)) {
+            return '&#xFFFD;'; // Unicode "replacement character", indicating a non printable character.
+        }
+
+        return match ($codepoint) {
+            34 => '&quot;',
+            38 => '&amp;',
+            60 => '&lt;',
+            62 => '&gt;',
+            default => sprintf($codepoint > 255 ? '&#x%04X;' : '&#x%02X;', $codepoint),
+        };
+    }
+
+    /**
+     * Decodes a given UTF-8 character (which may be multibyte) to a Unicode codepoint.
+     * E.g. The multibyte character "€", encoded in UTF-8 as the byte sequence 0xE2 0x82 0xAC,
+     * will return the integer 0x20AC, which is the corresponding unicode codepoint.
+     */
+    private function utf8CharacterToCodepoint(string $utf8_char): int
+    {
+        // UTF-32 encodes Unicode codepoints as itself. BE stands for Big Endian.
+        return hexdec(bin2hex(strlen($utf8_char) > 1 ? (mb_convert_encoding($utf8_char, 'UTF-32BE', 'UTF-8')) : $utf8_char));
+    }
+
+    /**
+     * The unicode range for control codepoints is from U+0000 NULL to U+001F (inclusive) and from U+007F to U+009F (inclusive).
+     * The range U+0000 NULL to U+001F and the codepoint U+007F are all one byte long and are covered by the PHP function ctype_cntrl.
+     *
+     * See https://infra.spec.whatwg.org/#control for more information.
+     */
+    private function isNonPrintableControl(string $utf8_char, int $codepoint): bool
+    {
+        return strlen($utf8_char) === 1 ?
+            ctype_cntrl($utf8_char) && !ctype_space($utf8_char) :
+            $codepoint <= 0x9F;
+    }
+
+    /**
+     * Unicode specifies a fixed list of 66 codepoints to be "noncharacters".
+     * Please see https://infra.spec.whatwg.org/#noncharacter for the complete list.
+     */
+    private function isNonCharacter(int $codepoint): bool
+    {
+        return 0xFDD0 <= $codepoint && $codepoint <= 0xFDEF ||
+               in_array($codepoint, [0xFFFE, 0xFFFF, 0x1FFFE, 0x1FFFF, 0x2FFFE, 0x2FFFF, 0x3FFFE, 0x3FFFF, 0x4FFFE, 0x4FFFF, 0x5FFFE, 0x5FFFF, 0x6FFFE, 0x6FFFF, 0x7FFFE, 0x7FFFF, 0x8FFFE, 0x8FFFF, 0x9FFFE, 0x9FFFF, 0xAFFFE, 0xAFFFF, 0xBFFFE, 0xBFFFF, 0xCFFFE, 0xCFFFF, 0xDFFFE, 0xDFFFF, 0xEFFFE, 0xEFFFF, 0xFFFFE, 0xFFFFF, 0x10FFFE,  0x10FFFF], true);
+    }
+}
--- a/components/ILIAS/Refinery/README.md
+++ b/components/ILIAS/Refinery/README.md
@@ -15,6 +15,11 @@ processed by the ILIAS project.
       - [in](#in)
         * [series](#series)
         * [parallel](#parallel)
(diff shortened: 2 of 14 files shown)


mjansenDatabay (approved):

src/Refinery/Escape/Transformation/HTMLAttribute.php
```diff
@@ -0,0 +1,72 @@
...
+            '/[^a-z0-9,._-]/iSu',
+            fn(array $m): string => $this->replace($m[0]),
+            $from
+        ) ?? throw new ValueError('Invalid UTF-8 string given.');
+    }
+
+    private function replace(string $chr): string
+    {
+        if ($this->isNonPrintableControl($chr)) {
+            return '&#xFFFD;';
```
mjansenDatabay: Could you add a comment which describes what `&#xFFFD;` is?

src/Refinery/Escape/Transformation/HTMLAttribute.php
```diff
@@ -0,0 +1,72 @@
...
+        return match ($ord) {
+            34 => '&quot;',
+            38 => '&amp;',
+            60 => '&lt;',
+            62 => '&gt;',
+            default => sprintf($ord > 255 ? '&#x%04X;' : '&#x%02X;', $ord),
+        };
+    }
+
+    public function isNonPrintableControl($chr): bool
```
mjansenDatabay: Why not `private`?

mjansenDatabay (approved):

src/Refinery/Escape/Transformation/HTMLAttribute.php
lscharmer: Could question :)

mjansenDatabay (approved):

klees (changes requested):
Hi @lscharmer,

please be so kind and try to provide minimal PRs. In this case, one PR per Transformation would be nice. This would improve various things:

* I could look into PRs incrementally or even ask other people to look into smaller PRs.
* We could merge PRs incrementally. I guess I could have merged some of the changes already without further ado, but now they are all tied together.
* Smaller PRs will be easier to review because I reviewers aren't repelled by the amount of code to be reviewed.

These, and maybe even more, are the reasons we specifically ask to make small PRs in our [rules for contributors](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/docs/development/contributing.md#rules-for-contributors). There are times when big PRs are warranted. This is not one of these times.

So, now to the content =)

- [x] CSS: Why do we need this transformation? Do we have a usecase for it?
- [x] HTMLAttributes: Why do we need this transformation? Do we have a usecase for it?
- [x] JavaScript: Why do we need this transformation? Do we have a usecase for it?
- [x] u: What does the `u` modifier in your regexps do? [This](https://github.com/ILIAS-eLearning/ILIAS/pull/6314/files#diff-92e0d450d8ec63ba77a6e87023294d0d2a6d9f061728da368f85befca262de78R39) seems to suggest that it somehow only matches UTF-8 strings, but I could not find according docu. Could you link some? Could you give me an example of a string that would make that expression fail?

- [x] Wording Group: I think what we do here is "Encoding" not "Escaping". "Escaping" means to have a certain character (say `\`) that makes the next character(s) behave differently than normal (make `\n` a linebreak). We "escape the n". Encoding on the other hand is about how data is represented. This is what we are doing here: Move data in UFT8-space to special character sequences that might contain escape characters. I would propose to rename the group to "Encode".
- [x] Wording HTML: Please rename the transformation to "HTMLSpecialCharsAsEntities" 
- [x] Wording CSS: Please rename the transformation to "NonAlphaNumToCSSEscapes" 
- [x] Docu CSS: Please document what [this transformation](https://github.com/ILIAS-eLearning/ILIAS/pull/6314/files#diff-a6b677866a51479f2ca08ef1f729a9de1a7fc21b25cc9018f5ef27df19d1a72fR32) is supposed to do. Maybe link some source such as [this](https://www.w3.org/International/questions/qa-escapes#cssescapes).
- [x] Docu HTML: Please document what [this transformation](https://github.com/ILIAS-eLearning/ILIAS/pull/6314/files#diff-92e0d450d8ec63ba77a6e87023294d0d2a6d9f061728da368f85befca262de78R32) is supposed to do. Maybe link some sources.
- [x] Docu URL: Same as for HTML.
- [x] Docu HTMLAttributes: Please document what [this transformation](https://github.com/ILIAS-eLearning/ILIAS/pull/6314/files#diff-b83583f7eb3a52d71a791c02009ccfe2a3a47bbeb67bd969b324d3105dc2997dR30) is supposed to do. Maybe link some sources. Judging on the examples in the tests, this transformation seem to do various things at once and, frankly, I cannot grasp how and where I would want to use it or what it is even supposed to do. I might ask to iterate on this further.
- [ ] Docu JavaScript: Same as for `HTMLAttributes` 
- [ ] Docu UTF8Chars: Please document what the methods in that trait are supposed to do.

In general, there are many things here that I currently do not understand here. Also, it took me quite some time to look through this. Besides smaller PRs a better exposition would have helped a lot: Explain stuff in the the summary of the PR, add docu to the classes and maybe methods you expose, give names that are suitable or even can be used as a starting point to search for stuff on the internet.

Kind regards!

klees:
Hi @lscharmer,

I will assign this to you, so this does not appear on my agenda anymore.

Please assign back if this is ready for the next iteration.

Kind regards!

lscharmer:
Hi @klees,

I updated this PR and addressed your points:

>     * [ ]  CSS: Why do we need this transformation? Do we have a usecase for it?
In agreement with @mjansen, we decided that we don't and it is therefore removed from this PR :)

>     * [ ]  HTMLAttributes: Why do we need this transformation? Do we have a usecase for it?
There are many places where PHP variables are passed to templates where they are used as HTML attributes.
E.g. https://github.com/ILIAS-eLearning/ILIAS/blob/0514bac85249b86002d05915fcab7427e15d3a63/components/ILIAS/UI/src/Implementation/Component/Link/Renderer.php#L74 would be one use case where the value should be escaped as a HTML attribute.

>     * [ ]  JavaScript: Why do we need this transformation? Do we have a usecase for it?
This is useful in places where JavaScriptBindable::withAdditionalOnLoadCode(...) is used to add JS from PHP.
E.g. Here: https://github.com/ILIAS-eLearning/ILIAS/blob/0514bac85249b86002d05915fcab7427e15d3a63/components/ILIAS/UI/src/Implementation/Component/Button/Renderer.php#L94

>     * [ ]  u: What does the `u` modifier in your regexps do? [This](https://github.com/ILIAS-eLearning/ILIAS/pull/6314/files#diff-92e0d450d8ec63ba77a6e87023294d0d2a6d9f061728da368f85befca262de78R39) seems to suggest that it somehow only matches UTF-8 strings, but I could not find according docu. Could you link some? Could you give me an example of a string that would make that expression fail?
The `u` modifier is very important. The behaviour of the modifier is documented here: https://www.php.net/manual/en/reference.pcre.pattern.modifiers.php
Additionally the first user contributed note on that page describes the behaviour with invalid UTF-8 strings with this modifier in detail.
I use the `u` modifier to ensure that the given string is a valid UTF-8 string. IMO this is the most fail prove way to check for a valid UTF-8 string, as it is handled by PHP itself, instead of implementing this oneself. If you know a better way to check for a valid UTF-8 string I will change this of course.
The following expression will fail: `preg_match('//u', "\x80")` but without the `u` modifier it will not.

>     * [ ]  Wording Group: I think what we do here is "Encoding" not "Escaping". "Escaping" means to have a certain character (say `\`) that makes the next character(s) behave differently than normal (make `\n` a linebreak). We "escape the n". Encoding on the other hand is about how data is represented. This is what we are doing here: Move data in UFT8-space to special character sequences that might contain escape characters. I would propose to rename the group to "Encode".
Done.

>     * [ ]  Wording HTML: Please rename the transformation to "HTMLSpecialCharsAsEntities"
Done. I hope the documentation and links I added are enough, as this is just a wrapper around a builtin PHP function.

>     * [ ]  Wording CSS: Please rename the transformation to "NonAlphaNumToCSSEscapes"
Is removed instead.

>     * [ ]  Docu CSS: Please document what [this transformation](https://github.com/ILIAS-eLearning/ILIAS/pull/6314/files#diff-a6b677866a51479f2ca08ef1f729a9de1a7fc21b25cc9018f5ef27df19d1a72fR32) is supposed to do. Maybe link some source such as [this](https://www.w3.org/International/questions/qa-escapes#cssescapes).
Is removed instead.

>     * [ ]  Docu HTML: Please document what [this transformation](https://github.com/ILIAS-eLearning/ILIAS/pull/6314/files#diff-92e0d450d8ec63ba77a6e87023294d0d2a6d9f061728da368f85befca262de78R32) is supposed to do. Maybe link some sources.
Done. I hope the documentation and links I added are enough, as this is just a wrapper around a builtin PHP function.

>     * [ ]  Docu URL: Same as for HTML.
Done. I hope the documentation and links I added are enough, as this is just a wrapper around a builtin PHP function.

>     * [ ]  Docu HTMLAttributes: Please document what [this transformation](https://github.com/ILIAS-eLearning/ILIAS/pull/6314/files#diff-b83583f7eb3a52d71a791c02009ccfe2a3a47bbeb67bd969b324d3105dc2997dR30) is supposed to do. Maybe link some sources. Judging on the examples in the tests, this transformation seem to do various things at once and, frankly, I cannot grasp how and where I would want to use it or what it is even supposed to do. I might ask to iterate on this further.
Done, this class is renamed to `HTMLAttributeValue`. This transformation only encodes a given string to be a valid HTML attribute value. This just includes all of the requirements of the HTML spec.
There are just a lot of requirements.
I add documentation to elaborate on all of the requirements of an HTML attribute value and added links to the corresponding HTML specs where necessary and I renamed all variables to be more fitting to the domain, so one can search for these words online.

>     * [ ]  Docu JavaScript: Same as for `HTMLAttributes`
After some research I noticed that this can be implemented much simpler. It uses `json_encode` with correct flags now.
The transformation is renamed to `JsonEncode`.

>     * [ ]  Docu UTF8Chars: Please document what the methods in that trait are supposed to do.
Done, this trait is also renamed to `ProvideUTF8CodepointRange`.



Best regards
@lscharmer

alex40724:
@lscharmer Thanks for this PR. This is definitely needed. I would strongly suggest to extend the readme file https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/Refinery/README.md with this PR, too, since this is what most devs will read. Outlining the main use cases would be very helpful.

lscharmer:
> @lscharmer Thanks for this PR. This is definitely needed. I would strongly suggest to extend the readme file https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/Refinery/README.md with this PR, too, since this is what most devs will read. Outlining the main use cases would be very helpful.

Hi @alex40724,
the `README.md` is adjusted accordingly.

I also renamed the `JsonEncode` transformation to `Json` as `encode` is already in the group / path name (`$refinery->enode()->json()`).

mjansenDatabay (approved):

mjansenDatabay:
I added this to the JF agenda for educational purposes :-).

alex40724:
@lscharmer Thanks!

matthiaskunkel:
Jour Fixe, 27 MAY 2024: @mjansenDatabay notified us about the transformations for escaping html/css/js and asks all developers to use this methods for ILIAS 10. PR will be merged to trunk.

components/ILIAS/Refinery/src/Encode/Transformation/HTMLAttributeValue.php
```diff
@@ -0,0 +1,107 @@
...
+     * The unicode range for control codepoints is from U+0000 NULL to U+001F (inclusive) and from U+007F to U+009F (inclusive).
+     * The range U+0000 NULL to U+001F and the codepoint U+007F are all one byte long and are covered by the PHP function ctype_cntrl.
+     *
+     * See https://infra.spec.whatwg.org/#control for more information.
+     */
+    private function isNonPrintableControl(string $utf8_char, int $codepoint): bool
+    {
+        return strlen($utf8_char) === 1 ?
+            ctype_cntrl($utf8_char) && !ctype_space($utf8_char) :
+            0x1F <= $codepoint && $codepoint <= 0x9F;
```
nicoroeser: Nitpick: the `0x1F <= $codepoint && ` part is a bit confusing after you have read the function comment. It does not matter, but it will always be true; we could remove it. Or replace `0x1F` with `0x7F` (which would make the check less strict, but still does not matter).

If I’m mistaken, please enlighten me; if not, consider dropping the part from the code. Thanks. :smiley:

lscharmer: You are correct, I will remove that part. Thanks for the correction :)

lscharmer: Is removed

nicoroeser: Thanks! :smiley:

PR #6147, opened 2023-07-13
TA: Use Presentation Table for results

nhaagen:
https://docu.ilias.de/goto_docu_wiki_wpage_7764_1357.html

PR #6317, opened 2023-09-13
Icons: switch ok/mostly_ok around

nhaagen:
In https://github.com/ILIAS-eLearning/ILIAS/pull/6147, we stumbled over the icons for ok/mostly_ok/wrong,
and I'd like to suggest a (for me, at least) more intuitive approach:

E.g., a question, that scores zero points, is clearly "false". 
With full pionts, it's clearly "correct".
Those should be the filled icons, imo.
Everything in between is "partially correct" at least; the checkmakrk indicates this, and the unfilled body suggests a 'not fully' state.

Screenshot from 2023-09-13 09-28-04
vs (new)
Screenshot from 2023-09-13 09-34-28

Best regards


yvseiler:
I have contacted the icon designer so that we can solve the issue with the three ok/mostlyok/notok icons. Unfortunately, the change alone is not enough, because if both ok / not ok have the "filled" variant, it is difficult to distinguish in a data collection when quickly looking over it. Therefore, I believe that only the change of the two is not useful for all cases.  I will try to discuss the problem with the icon designer as soon as possible so that we can find a solution for the three combinations that works well both in the test and in the data collection.

PR #6320, opened 2023-09-13
Draft: UI: Resource Selector Prototype

thibsy:
Hi @Amstutz,

As promised (although a little late), here is a working example of the resource selector prototype. I will briefly recap for you and the general public as well.

We have discussed several ideas of resource selectors (formerly known as tree picker) in two UI Clinic meetings, where we have decided to pursue a drilldown approach for browsing possible items. This leaves the option to implement a sort of multi-column drilldown for larger devices, like we know it from the MacOS finder. Now to test the user experience of this approach, as well as the accessibility, we have been tasked to implement a clickable prototype. 

This prototype therefore merely implements some minimal functionality for selecting items visually. I could reuse most of the drilldown components logic and only needed to tweak the CSS a little bit. The JavaScript code was almost entirely compatible except for minor changes when fetching the entry label.

I have implemented this as a working UI example, which you can find in the documentation under `UI Component > Input > Field`.

Kind regards,
@thibsy

### DIFF
--- a/src/UI/examples/Input/Field/resource_selector_prototype.php
+++ b/src/UI/examples/Input/Field/resource_selector_prototype.php
@@ -1,0 +1,179 @@
+<?php
+
+declare(strict_types=1);
+
+namespace ILIAS\UI\examples\Input\Field;
+
+use ILIAS\UI\Implementation\Render\ilTemplateWrapper;
+use ILIAS\Data\URI;
+use ILIAS\UI\Factory;
+use ILIAS\UI\Component\Menu\Drilldown;
+use ILIAS\UI\Implementation\Render\Template;
+use ILIAS\UI\Component\Symbol\Symbol;
+use ILIAS\UI\Renderer;
+
+function resource_selector_prototype(): string
+{
+    global $DIC;
+
+    $main_template = $DIC->ui()->mainTemplate();
+    $renderer = $DIC->ui()->renderer();
+    $factory = $DIC->ui()->factory();
+
+    $prototype = getInputTemplate($main_template, 'tpl.resource_selector_prototype.html');
+
+    $c_icon = $factory->symbol()->icon()->standard('', '')->withAbbreviation('C');
+    $g_icon = $factory->symbol()->icon()->standard('', '')->withAbbreviation('G');
+
+    $structure = [
+        new PseudoEntry('1 Category', $c_icon, [
+            new PseudoEntry('1.1 Sub Category', $c_icon, [
+                new PseudoEntry('1.1.1 Course', $c_icon),
+                new PseudoEntry('1.1.2 Group', $g_icon),
+            ]),
+            new PseudoEntry('1.2 Course', $c_icon),
+            new PseudoEntry('1.3 Group', $g_icon),
+        ]),
+        new PseudoEntry('2 Category', $c_icon, [
+            new PseudoEntry('2.1 Course', $c_icon),
+            new PseudoEntry('2.2 Group', $g_icon),
+        ]),
+        new PseudoEntry(
+            '3 Course with an extremely long title which should be handled by the component somehow in an unfrastrating manner, which IMO is not yet the case, right?' .
+            'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed vitae nisl eget nunc aliquam aliquet.' .
+            'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed vitae nisl eget nunc aliquam aliquet.' .
+            'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed vitae nisl eget nunc aliquam aliquet.',
+            $c_icon
+        ),
+        new PseudoEntry('4 Group', $g_icon),
+        new PseudoEntry('Test breadcrumb nesting', $c_icon, [
+            new PseudoEntry('Level 1', $c_icon, [
+                new PseudoEntry('Level 2', $c_icon, [
+                    new PseudoEntry('Level 3', $c_icon, [
+                        new PseudoEntry('Level 4', $c_icon, [
+                            new PseudoEntry('Level 5', $c_icon, [
+                                new PseudoEntry('Level 6', $c_icon, [
+                                    new PseudoEntry('Level 7', $c_icon, [
+                                        new PseudoEntry('Level 8', $c_icon, [
+                                            new PseudoEntry('Level 9', $c_icon, [
+                                                new PseudoEntry('Level 10', $c_icon, [
+                                                    new PseudoEntry('You have reached the target, yay!', $c_icon),
+                                                ]),
+                                            ]),
+                                        ]),
+                                    ]),
+                                ]),
+                            ]),
+                        ]),
+                    ]),
+                ]),
+            ]),
+        ]),
+        ...generateManyEntries($c_icon),
+    ];
+
+    $html = '';
+    foreach ($structure as $pseudo_entry) {
+        $html .= renderPseudoEntry($main_template, $renderer, $pseudo_entry);
+    }
+
+    $prototype->setVariable('ENTRIES', $html);
+
+    $prototype->setVariable('DROPDOWN', $renderer->render($factory->dropdown()->standard([
+        $factory->button()->shy('Entry 1', '#'),
+        $factory->button()->shy('Entry 1', '#'),
+        $factory->button()->shy('Entry 1', '#'),
+    ])));
+
+    return $prototype->get();
+}
+
+function getNewContainerEntry(
+    \ilGlobalTemplateInterface $main_template,
+    Renderer $renderer,
+    PseudoEntry $pseudo_entry,
+    int $level = 0
+): string {
+    $entry = getInputTemplate($main_template, 'tpl.resource_selector_container_item.html');
+    $entry->setVariable('ITEM_ICON', $renderer->render($pseudo_entry->icon));
+    $entry->setVariable('ITEM_NAME', $pseudo_entry->title);
+    $entry->setVariable('SELECT_ACTION', getSelectActionHtml($main_template));
+    $entry->setVariable('MENU_LEVEL', $level);
+
+    $sub_entries = $pseudo_entry->entries ?? [];
+    foreach ($sub_entries as $sub_entry) {
+        if (null === $sub_entry->entries) {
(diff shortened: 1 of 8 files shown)


thibsy:
Screenshot 2023-09-13 at 15 50 37

--- 

Screenshot 2023-09-13 at 15 51 11

yvseiler:
Hi @thibsy 

Thank you a lot for your work on this topic. A big step forward since our last workshop, thanks. @Amstutz and I looked at the Picker in action. Based on our first impression when operating the treepicker, we asked ourselves the following questions:

- [ ] What happens if only a single choice selection is possible? It would be expected that the add glyphs disappear or something else?
- [ ] How do I recognise objects that are not selectable, or what happens if one is not selectable? F.e. I have a file, a learning module and a folder in a course. I want to copy the file into the folder. How do I know that the learning module is not selectable? At the moment in ILIAS non-selectable objects are not even displayed in the tree.
- [ ] When a category is selected, should the sub-categories also be selected? Currently in the new treepicker, only the category itself is selected, not the objects within it (as ILIAS interacts now). I'm not sure, if there is a case, where the sub-categories should also be selected or not.
- [ ] What happens when I click on the "Choose" button? Nothing happens in the framework. We can imagine that the modal closes and the selected objects are listed in the form, or what did you have in mind here?
- [ ] Is the last place I was before the modal was closed saved? The example suggests that at the moment, so we are unsure here. Or do I always start at the highest level "repository"? I think we should only save the "last visited place" if we have a good option to go back fast to the highest level ("Repository")
- [ ] We had talked in a further workshop about the fact that it can be helpful if the objects already selected are listed/displayed in the case of a multiple selection. Is there still any progress on this or reasons/problems why we do without it?

- [ ] The "end-object" (f.e. 4 Group in screenshot above) has a hover, although it is not clickable in the sense of "go down one hierarchy level". Here I would expect that when clicked, the end object would be selected. If this is not possible, the end object should not have a hover effect (only the add glyph on the right). Or would it be possible to highlight the Title too when hovering the Add-Glyph?
- [ ] Use a coloured icon for all objects (not greyed out). Currently, in the proposal, the icon is not greyed out for the objects that I can go one hierarchy level lower. For those that I can select, but cannot go "deeper", the icons are greyed out. This "inactive" status irritates me and I rather interpret it as "cannot be selected" (which is not the case). Greyed-out icons should only be used if the object cannot be selected (i.e. the add icon is missing).

I have received a feedback that the white part (i.e. an end object) is interpreted as a part of the light grey part above it (e.g. 3 and 4 are parts of 2). Maybe it would make sense to design everything with the same background and only with the glyph on the right (>) it becomes obvious that you can go one level deeper... I'm not sure yet. I will ask two more users here how they interpret this.

It takes a moment to understand the screen when you first see it. But the controls and symbols itself are simple and easy to understand. It seems, that it encourages you to try it out and it can easily be reset (click on the tick to "no longer select"). I am looking forward to the next steps. 

Please ask if anything is unclear or you need thoughts or suggestions.

kindly,
@yvseiler

dsstrassner:
Hi @thibsy, 
Out of curiosity and the need to have a modern selector in the whole ILIAS world: What will be the next steps in this project, and when will they happen? Will this be implemented with ILIAS 10?

thibsy:
Hi @yvseiler, 

Thanks a lot for your detailed feedback, I'll try to answer your questions as best I can :).

> What happens if only a single choice selection is possible? It would be expected that the add glyphs disappear or something else?

Disabling the select-buttons seems to be a good choice, but I don't think this is sufficient visually. Hence, I also improved the colouring concept in a sense of disabled vs active items inside of the selector. Note that this is not fully implemented, but the idea is to either disable the selection alone or recursively, depending on the action and/or target.

> How do I recognise objects that are not selectable, or what happens if one is not selectable? F.e. I have a file, a learning module and a folder in a course. I want to copy the file into the folder. How do I know that the learning module is not selectable? At the moment in ILIAS non-selectable objects are not even displayed in the tree.

To ignore items which cannot be selected is definitely the easy choice. However, I think it could make sense to show such items anyways, depending on the context, because if users are already familiar with a certain tree, it might be confusing or not even recognisable. But I think we should leave this to the consumer of this component.

> When a category is selected, should the sub-categories also be selected? Currently in the new treepicker, only the category itself is selected, not the objects within it (as ILIAS interacts now). I'm not sure, if there is a case, where the sub-categories should also be selected or not.

This this depends on the action, but I certainly had in mind to provide an according mechanism for a multi-select, I just didn't implement this for the prototype. However, I'm not sure yet if this should be defined on the resource selector, or on an item-level, but this should not be our concern at this point.

> What happens when I click on the "Choose" button? Nothing happens in the framework. We can imagine that the modal closes and the selected objects are listed in the form, or what did you have in mind here?

I did not implement this mechanism for the prototype, but you can imagine that the data is transferred from the modal to some sort of list in the actual form, where items can also be removed by clicking an icon. Maybe we could also reuse the tag input for this, where items will be added as tags.

> Is the last place I was before the modal was closed saved? The example suggests that at the moment, so we are unsure here. Or do I always start at the highest level "repository"? I think we should only save the "last visited place" if we have a good option to go back fast to the highest level ("Repository")

From a user perspective, I think it is more friendly to keep the location in the tree persistent, as well as the selected items. But I agree that an action to return to the root of the tree should be implemented. I therefore added breadcrumbs to the prototype which can be used to navigate back in bigger chunks. This also helps with knowing the context in the tree. Note, the breadcrumbs will reset if you reload the page but the drilldown does not, because I did not yet couple their state to the drilldown.

> We had talked in a further workshop about the fact that it can be helpful if the objects already selected are listed/displayed in the case of a multiple selection. Is there still any progress on this or reasons/problems why we do without it?

My idea was to reuse a dropdown menu, because the space inside of the modal is very limited. The menu can be opened via button in the footer of the modal, which would be disabled initially unless items have been selected. Selecting an item enables the button to show a list of all selected items, each one featuring a removal-glyph to unselect them again. However, for the life of me, I could not figure out why the dropdown component does not work inside of the prototype. There seems to be an issue with the JavaScript, because the menu cannot be opened via its button. So for now, you have to imagine this...

> The "end-object" (f.e. 4 Group in screenshot above) has a hover, although it is not clickable in the sense of "go down one hierarchy level". Here I would expect that when clicked, the end object would be selected. If this is not possible, the end object should not have a hover effect (only the add glyph on the right). Or would it be possible to highlight the Title too when hovering the Add-Glyph?

That is true, I adjusted the prototype accordingly. 

> Use a coloured icon for all objects (not greyed olut). Currently, in the proposal, the icon is not greyed out for the objects that I can go one hierarchy level lower. For those that I can select, but cannot go "deeper", the icons are greyed out. This "inactive" status irritates me and I rather interpret it as "cannot be selected" (which is not the case). Greyed-out icons should only be used if the object cannot be selected (i.e. the add icon is missing).

I agree, I adjusted this too.

> I have received a feedback that the white part (i.e. an end object) is interpreted as a part of the light grey part above it (e.g. 3 and 4 are parts of 2). Maybe it would make sense to design everything with the same background and only with the glyph on the right (>) it becomes obvious that you can go one level deeper... I'm not sure yet. I will ask two more users here how they interpret this.

I adjusted the prototype to try this out. IMO, it's now more recognisable that all currently listed elements are content of the top-most title or back action.

Additionally, I also adjusted the following:

- the modal has now a fixed height, meaning that the actual content becomes scrollable whereas the header and footer stay fixed. IMO this is more user friendly because all important actions are always visible on the screen.
- as already mentioned, I added breadcrumbs to at the top of the modal content. However, I failed to make them sticky in a sense where they would not move on scroll. IMO this provides much more context for orientation in the tree already.

Kind regards,
@thibsy

thibsy:
@dsstrassner thx for checking in. The next steps would be to check if this prototype is a viable option and then define the public interface for such a component. Our goal will be to implement a first version of the resource selector for ILIAS 10.

kergomard:
Hi @thibsy , hi all
Sorry for weighing in on this so late and maybe you have already tackled this issue. I've a conceptual question, though: Where would we use this?
I believe that many places we are using repository pickers right now are not good usages.
My reasoning:
In ILIAS we have a separate access right for visibility, so it can happen that parts of a tree are accessible, but invisible to a user when following a tree based visualization. We cannot ignore the visibility of tree branches as this would lead to serious information disclosure issues. How would we deal with this?
My feeling is, that an autocomplete search field with a locator based search list and a fuzzy search algorithm would be a better solution for the problem we approach with tree-pickers today and additionally use far less screen real estate.

Sorry again, thanks a lot for all the prototype, and best,
@kergomard

thibsy:
Hi @kergomard,

Thanks for weighing in here.

The resource selector will primarily be used to replace usages of `ilExplorer`s and numeric inputs, where an ILIAS object is referenced. I think if we implement this with some flexibility, we could also use this for entirely different purposes, like e.g. creating IRSS collections from a set of uploaded files.

From an UI framework perspective the visibility issue is not our concern. We will need to use some sort of items-provider, since trees need to be rendered asynchronously after a certain depth, which shifts the issue outside of our scope. However, I don't think this should be problematic.

About a search-based input field: there are certainly use-cases where this could be more fitting - I prefer prompt-based things myself. However, I think there will always be use-cases where discoverability is the main purpose, in which cases the search-based approach would not be feasible. We certainly had in mind to provide filters like a fuzzy-search in the future, to "trim the tree".

I'd like to hear @yvseiler opinion in this :).

Hope this helps, kind regards,
@thibsy

kergomard:
Thank you very much @thibsy 

This is exactly my point: We shouldn't replace any `ilExplorer` used to select an object in a tree by a new tree based resource selector, unless we are completely and absolutely sure only system administrators will have access to it (and I think this is only very rarely the case), because it is a conceptually flawed approach. I understand that this is actually not a primary concern for the UI-Coordinators, but as I don't think there is a solution for the problem of visibility, I don't think this ui-element can be used for what it is designed for and it would be sad to spend too much time on it.

Best,
@kergomard

Amstutz:
I see your point. However, a tree like this is a strong and good mental model, if used uniformly as a platform. It features an extremely high external consistency among operating systems and a huge range of applications. 

I would definitely not abandon such a tree - like way of searching for things. However, we might think on offering an alternative way (like a search), for the edge cases, featuring read-only (none visible) nodes in the tree. Note sure, if we want to overload this here with the case however. Afaik we did not support it so far in pickers used. We propably should thow in a further iteration.

PR #6322, opened 2023-09-14
9/composer/update psr7 and getallheaders

kergomard:
This PR updates `psr7` and `getallheaders` to newer versions.
To not create errors I also removed the TCPDF-Patch already removed in ILIAS 7 + 8. It is in a separate commit.


chfsx:
Thanks a lot @kergomard!

PR #6323, opened 2023-09-15
[FIX] allow finder to find classes in src as well 

chfsx:
@klees this changes the rules which directories are excluded from a ImplementationOfInterfaceFinder or UsageOfAttributeFinder which makes it possible to provide e.g. Agents and Objectives in src as well.

src/Setup, src/GlobalScreen and src-directories in Plugins are still excluded.


klees (approved):
I'm not really sure if this captures all problematic `src`-folders, but we can try for sure. Amending this is cheap...

PR #6325, opened 2023-09-15
Fix for 37964: Database Exception with MySQL

tfamula:
https://mantis.ilias.de/view.php?id=37964


kergomard:
Thank you very much @tfamula for the report and the PR. For me this solution feels like hiding the issue. I will fix it by quoting like in ILIAS 8.

Thanks again and best,
@kergomard

PR #6326, opened 2023-09-15
`Refinery`: Add more transformations (to improve `Refinery` usage in `ILIAS\Setup\Agent::getArrayToConfigTransformation`)

lscharmer:
This PR introduces new transformations for the refinery, mostly designed to be usable for the ILIAS Setup:

## `$refinery->to()->memberOf()`
Enum validation: Checks if a value is in a specified set.

Example use case in the config would be: `chatroom.log_level` which must be one of `emerg`, `alert`, `crit`, `error`, `warning`, `notice`, `info`, `debug` or `silly`

## `$refinery->executable()`
Checks if the given string specifies an executable path.

I could not find a suitable place for this transformation so I placed it on the top level of the refinery.

An example use case would be for config paths like: `mediaobject.path_to_ffmpeg`, `style.path_to_scss`, `virusscanner.path_to_scan` etc.

## `$refinery->int()->isBetween()`
This is just a convenience transformation but IMO very nice to have.
There is no actual class for this constraint as it can be composed from other transformations and constraints.

Example use case would be in the config would be: `chatroom.deletion_interval.deletion_value` which have both min and max values.

### `DeriveTransformWithProblem`
This trait is implemented to prevent boilerplate when implementing new transformations, as some methods are always implemented the same.

### `ProblemBuilder`
No behavior is changed, just some simplifications (with one exception (pun intended): the `InvalidArgumentExcpetion` is now a `ArgumentCountError`)

-----------------

Just FYI: Another PR will follow which will introduce deeper changes to the refinery and the relation between Transformations and Constraints but that PR will be independent of the changes made in this PR.

### DIFF
--- a/components/ILIAS/Refinery/README.md
+++ b/components/ILIAS/Refinery/README.md
@@ -187,6 +187,11 @@ transformation to create structures like `list`, `dictonary`,
                  parameters on the objects.
 * `data()`     - Returns a data factory to create a certain data type
 
+###### Other
+
+* `inArray()` - Returns an object that validates that a value is a member of the given array.
+                 E.g.: `$t = $refinery->to()->memberOf(['red', 'green', 'blue']); $t->transform('blue'); /* => 'blue' */ $t->transform('yellow'); /* => Exception */`
+
 ##### in
 
 The `in` group is a group with a dict of `Transformations`
@@ -352,7 +357,7 @@ Just the `applyTo` method needs to be created in the new transformation class.
 
 ##### Error Handling
 
-Exceptions thrown inside the `applyTo` method will be
+Exceptions thrown inside the `applyTo` method will
 **not be** caught.
 On return of an [error result object (`Result\Error`)](/src/Data/README.md#result)
 the `transform` method will throw an exception.
@@ -361,6 +366,15 @@ the `transform` method will throw an exception.
   thrown.
 * If the content of the error object is an string this string will be added
   to a an `Exception` which will be thrown.
+
+#### DeriveTransformWithProblem
+
+This trait is used to simplify the creation of new constraints and reduce duplicated code.
+For a constraint only the methods `accepts($value): bool` and `getError()` must be implemented.
+
+#### Other Transformations
+
+* `$refinery->executable()` Returns an object that validates that a given path designates an executable OS path.
 
 ## Libraries
 
--- a/components/ILIAS/Refinery/src/DeriveTransformWithProblem.php
+++ b/components/ILIAS/Refinery/src/DeriveTransformWithProblem.php
@@ -1,0 +1,57 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\Refinery;
+
+use UnexpectedValueException;
+
+/**
+ * This trait is a convenience trait which uses `DeriveApplyToFromTransform`, `DeriveInvokeFromTransform` and `ProblemBuilder`
+ * and implements some methods that are always implemented the same.
+ * The main purpose of this trait is to reduce duplicated code and make it easier to implement new constraints.
+ */
+trait DeriveTransformWithProblem
+{
+    use DeriveApplyToFromTransform;
+    use DeriveInvokeFromTransform;
+    use ProblemBuilder;
+
+    public function transform($from)
+    {
+        $this->check($from);
+        return $from;
+    }
+
+    public function check($value)
+    {
+        if (!$this->accepts($value)) {
+            throw new UnexpectedValueException($this->getErrorMessage($value));
+        }
+    }
+
+    public function problemWith($value): ?string
+    {
+        if (!$this->accepts($value)) {
+            return $this->getErrorMessage($value);
+        }
+
+        return null;
+    }
+}
--- a/components/ILIAS/Refinery/src/Factory.php
+++ b/components/ILIAS/Refinery/src/Factory.php
@@ -44,7 +44,7 @@ class Factory
      */
     public function to(): To\Group
     {
-        return new To\Group($this->dataFactory);
+        return new To\Group($this->dataFactory, $this->language);
     }
 
     /**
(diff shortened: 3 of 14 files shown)


src/Refinery/To/DefaultToNull/Group.php
```diff
@@ -0,0 +1,38 @@
...
+
+declare(strict_types=1);
+
+namespace ILIAS\Refinery\To\DefaultToNull;
+
+use ILIAS\Refinery\To\DefaultToNull\Transformation\RecordOfTransformation;
+use ILIAS\Refinery\Transformation;
+use ilLanguage;
+
+class Group
```
mjansenDatabay: Could you try to describe (with PhpDocs) the purpose of this group? Maybe it is also worth to add some lines to the `README.md` of the `Refinery`.

src/Refinery/To/DefaultToNull/Group.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    public function __construct(
+        private readonly ilLanguage $language
+    ) {
+    }
+
+    public function recordOf(array $transformations): Transformation
```
mjansenDatabay: Could you try to describe (with PhpDocs) the purpose of this transformation (here or/and in the respective class)? Maybe it is also worth to add some lines to the `README.md` of the `Refinery`.

src/Refinery/Factory.php
```diff
@@ -175,4 +175,9 @@ public function always($value): Transformation
     {
         return new ConstantTransformation($value);
     }
+
+    public function executable(): Transformation
```
mjansenDatabay: Although the name is already descriptive: Could you try to describe (with PhpDocs) the purpose of this transformation (here or/and in the respective class)? Maybe it is also worth to add some lines to the `README.md` of the `Refinery`.

src/Refinery/To/Group.php
```diff
@@ -179,4 +180,14 @@ public function dateTime(): Transformation
     {
         return new DateTimeTransformation();
     }
+
+    public function memberOf(array $valid_members): Transformation
+    {
+        return new MemberOfTransformation($valid_members, $this->language);
```
mjansenDatabay: Could you try to describe (with PhpDocs) the purpose of this transformation (here or/and in the respective class)? Maybe it is also worth to add some lines to the `README.md` of the `Refinery`.

src/Refinery/To/Group.php
```diff
@@ -179,4 +180,14 @@ public function dateTime(): Transformation
...
+    }
+
+    public function defaultToNull(): DefaultToNull
+    {
+        return new DefaultToNull($this->language);
```
mjansenDatabay: Although the name is already descriptive: Could you try to describe (with PhpDocs) the purpose of this transformation (here or/and in the respective class)? Maybe it is also worth to add some lines to the `README.md` of the `Refinery`.

src/Refinery/DeriveTransformWithProblem.php
```diff
@@ -0,0 +1,52 @@
...
+    use DeriveInvokeFromTransform;
+    use ProblemBuilder;
+
+    public function transform($from)
+    {
+        $this->check($from);
+        return $from;
+    }
+
+    public function check($value)
```
mjansenDatabay: Could we use `: void` here as return type?

mjansenDatabay (changes requested):
Thx a lot for this PR @lscharmer . After a first glimpse I only have some minor remarks, most of them are "documentation".

src/Refinery/DeriveTransformWithProblem.php
lscharmer: Even though the method in `Constraint::check()` has none?

mjansenDatabay (approved):

klees (changes requested):
Hi @lschamer,

the [request for minimal PRs from the other PR](https://github.com/ILIAS-eLearning/ILIAS/pull/6314#pullrequestreview-1813234053) applies here as well. 

Thanks a lot for adding these transformations, they seem to be quite useful and currently missing.

- [ ] ProblemBuilder: You seem to be fixing some problem with the `ProblemBuilder`, but do not really understand what it is. Or is it just a refactoring? Please split that change into a single commit. If it is a problem: please add unit test, that fail without the changes.
- [ ] DefaultToNull: I also came accross this problem several times. I'm not convinced that "all fields might be null" is the thing that really needs to be expressed. Instead we would want to say: this and that field could be null (and it is fine). Also, we would not only want to say it for the strict `to` variant, but also for the less `kindlyTo` variant. Hence I propose the following implementation: On toplevel, we add a transformation `withDefault($transformation, $default)` that replaces a `null` value with `$default` and passes non-null values to the `$transformation`. We could derive `optional($transformation)` from there. We could then introcude special cases for `WithDefault` in the `recordOf`-transformations for the `to` and `kindlyTo` variant. We could the write `recordOf(["key" => $f->optional(...)]);` or something similar for `withDefault`. Maybe we should discuss this, but I would at least suggest to split this topic from the PR at hand.
- [ ] MemberOf: Although I would personally also prefer `memberOf`, I think we should name this transformation `inArray`. This is something most PHP developers will recognize immediately, while "member of" comes from a more mathy background imo.

- [ ] Executable: Please rename to `IsExecutable`. We use this `has`/`is` naming scheme for constraints consistently, mostly.

Best wishes!

klees:
Hi @lscharmer,

as with #6314, please assign back if back if this is ready for the next iteration.

Kind regards!

lscharmer:
Hi @klees,

this PR is updated an can be reviewed again. Below are my answers to your review:

> - [x] ProblemBuilder: You seem to be fixing some problem with the `ProblemBuilder`, but do not really understand what it is. Or is it just a refactoring? Please split that change into a single commit. If it is a problem: please add unit test, that fail without the changes.

Yes, it is just refactoring. It is now placed in it's own commit.

> - [x] DefaultToNull: I also came accross this problem several times. I'm not convinced that "all fields might be null" is the thing that really needs to be expressed. Instead we would want to say: this and that field could be null (and it is fine). Also, we would not only want to say it for the strict `to` variant, but also for the less `kindlyTo` variant. Hence I propose the following implementation: On toplevel, we add a transformation `withDefault($transformation, $default)` that replaces a `null` value with `$default` and passes non-null values to the `$transformation`. We could derive `optional($transformation)` from there. We could then introcude special cases for `WithDefault` in the `recordOf`-transformations for the `to` and `kindlyTo` variant. We could the write `recordOf(["key" => $f->optional(...)]);` or something similar for `withDefault`. Maybe we should discuss this, but I would at least suggest to split this topic from the PR at hand.

I will extract the `DefaultToNull` stuff into it's own PR (it is removed from this one). The Transformation was not ment to address the `default` / `optional` problem, but I see how this is relevant when one chooses a specific approach on how one wants to pursue to implement such a concept.

We can further discuss the details in that PR (coming soon).

> - [x] MemberOf: Although I would personally also prefer `memberOf`, I think we should name this transformation `inArray`. This is something most PHP developers will recognize immediately, while "member of" comes from a more mathy background imo.

Done

> - [x] Executable: Please rename to `IsExecutable`. We use this `has`/`is` naming scheme for constraints consistently, mostly.

Done



Best regards
@lscharmer

klees:
Hi @lscharmer,

thanks again and sorry that it took me so long to rereview this. I just merged this manually with two rather minor changes.

Kind regards!

mjansenDatabay:
Thx @klees for your efforts

PR #6327, opened 2023-09-16
[FIX] #37959: remove default timeout for file uploads.

thibsy:
Hi all,

We have noticed an issue while uploading large files on slow(er) systems, where a default timeout of 30 seconds has prevented files from being uploaded, see https://mantis.ilias.de/view.php?id=37959.

The issue occurs because of a default value in [Dropzone.js](https://docs.dropzone.dev/), the library we use for file uploads.
There are two ways of fixing this:

- a) set the default value of 30 seconds to null when initializing the dropzone instance for our file input.
- b) update the library to a version where the default value is null, see [latest documentation](https://docs.dropzone.dev/configuration/basics/configuration-options) for details.

Since option a) would have been a workaround of sorts and updates are always nice to have, I decided to go with the latter option. Since we have stopped using `bower` and because I have had some problems when updating the package (it did not pull the latest version for some reason), I have decided to install the package with `npm` instead and uninstalled the bower package accordingly. I hope thats okay.

I have also made some adjustments in `Field\Renderer` because the format of max-file-sizes changed from mb to byte. While at it, I also fixed an [issue in `file.js`](https://github.com/ILIAS-eLearning/ILIAS/commit/f6f45c5dfc70ecb5eb2dd7567b7066f5025a26cc), solved in release_8 and trunk, where the max-file-size attribute was misspelled.

Kind regards,
@thibsy


klees (commented):
Hi @thibsy,

the general solution to move the dropzone library to npm is fine.

There are some specifics I really wonder about... Maybe you can help:

* [x] Why does fsevents appear in node_modules now? I doesn't seem to be included somewhere or even be added to package-lock.json? Feels strange.
* [x] Something similar for owl.carousel: Why are there updates in the folder?

Then:

* [x] Why'd you remove the [feedback about the actual sizes](https://github.com/ILIAS-eLearning/ILIAS/pull/6327/files#diff-2d9abcacf08b7e7e7a931e40d78ee734af912ab8af6659fa4b7a570dd349f632L21)?

Rest looks fine...

Thanks!
@Amstutz and @klees

thibsy:
Hi @Amstutz and @klees,

Thanks for having a closer look into this. I'll answer your questions as best I can:

I could figure out, that the `fsevents` got installed because I ran the command to update the installed `dropzone` package without the `--omit=optional` flag and it turns out, that this library will automatically be installed on macOS. I therefore removed these changes from the PR. However, the changes to the `owl.carousel` package seems odd, since the package is not listed as transitive dependency of `dropzone`. I simply updated the package with npm though (`npm update dropzone`), so I assume there is some relation between the packages we are missing.

About the feedback on file-sizes: I removed the placeholders from the default option because a) it was never used, since the `Field\Renderer` injects a proper translation, and b) the dropzone.js did not seem to replace the placeholders (`{{}}`) automatically - and implementing the prettification of file-sizes myself felt odd. However, it turns out that our entire `addedfile` handler was unnecessary and somehow prevented the default handler from properly setting the error message. I fixed this now and reverted the change, because the placeholders would be filled now. While investigating, I also stumbled upon some issues with dropzone.js:

- The latest documentation states, [the `maxFilesize` attribute must be provided in bytes](https://docs.dropzone.dev/configuration/basics/configuration-options). However, I found out while looking into the code that this is false. I also found an according [GitHub issue](https://github.com/dropzone/dropzone/issues/2197). So my previous comment was wrong.
- There is also a `filesizeBase` attribute, which defaults to `1000`, that _should_ be used for file-size calculations. However, I found out this is also false. There are several cases where `1024` is hardcoded when performing such calculations. Important for us is:
	- Prettification of file-sizes use base `1024` if the file is too big, which means MiB will be displayed instead of Mb if we don't provide a custom string.
	- Comparisons between the [`Blob.size`](https://developer.mozilla.org/en-US/docs/Web/API/Blob/size) and `maxFilesize` attributes also use `1024`, which means we must provide MiB instead of Mb as `maxFilesize`.

Kind regards,
@thibsy

klees:
Hi everybody,

I had a long look into the changes in `owl.carousel` because, I guess, this is exactly how some supply chain attack would look like. An unwarranted update, that goes through unnoted. The changes in the unminified files seem to be good, though, and the minified file contains the exact same change.

So changes are strange, because no one seems to understand why they are included. But the content is ok.

Kind regards!

klees:
No need to port this, library in later version already up to date.

PR #6330, opened 2023-09-18
Utilities: Fix min pwd length > max pwd length

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=37767

The method is used in `Services/SOAPAuth`, `Services/Init`, `Services/Registration` and `Services/AuthShibboleth`. Since the `Services/Utitlies` folder is not explicitly maintained anymore, I'll assign this correspondingly.

If approved, it has to be picked to `release_8` and `trunk` as well.


pascalseeland:
Already fixed in 8 and trunk

PR #6332, opened 2023-09-19
#0037988 - php 8 compatibility ilCourseParticipants

sroenchen:
change result check for query and add check for usable date


smeyer-ilias (commented):
Thanks for the report and the pull request. 
I fixed this issue with the commit https://github.com/ILIAS-eLearning/ILIAS/commit/fbb19cb6dfb3f25bd7dd6f53d35dacca35749d22

"origin_ts" is a unix time integer. strtotime expects a string date as first parameter and always evaluates to false if unix timestamps are passed as parameter.

smeyer-ilias:
Fixed in R8

PR #6334, opened 2023-09-19
#0037994 fix error when copying large container objects

sroenchen:
https://mantis.ilias.de/view.php?id=37994

Writing the log entry can occasionally cause an error. To fix that, the log entry could be shoved into the if-clause. 
Switching from 'isset' to '!empty' may bring an additional safeguard against undefined array key exceptions.


alex40724:
Fixed slightly different. Thanks for the PR.

PR #6335, opened 2023-09-19
T&A Bugfix #0029216: Test Import Results are being skipped

fhelfer:
https://mantis.ilias.de/view.php?id=29216

Caution when cherry-picking to 8
will do a PR for 8 as well, because not everything can be safely picked

### DIFF
--- a/components/ILIAS/Test/classes/class.ilObjTestGUI.php
+++ b/components/ILIAS/Test/classes/class.ilObjTestGUI.php
@@ -1454,18 +1454,16 @@ class ilObjTestGUI extends ilObjectGUI implements ilCtrlBaseClassInterface, ilDe
             $questionPageParser->setQuestionMapping($qtiParser->getImportMapping());
             $questionPageParser->startParsing();
 
-            if (isset($_POST["ident"]) && is_array($_POST["ident"]) && count($_POST["ident"]) == $qtiParser->getNumImportedItems()) {
-                // import test results
-                if (@file_exists(ilSession::get("tst_import_results_file"))) {
-                    $results = new ilTestResultsImportParser(
-                        ilSession::get("tst_import_results_file"),
-                        $newObj,
-                        $this->db,
-                        $this->logging_services->root()
-                    );
-                    $results->setQuestionIdMapping($qtiParser->getQuestionIdMapping());
-                    $results->startParsing();
-                }
+
+            if (@file_exists(ilSession::get("tst_import_results_file"))) {
+                $results = new ilTestResultsImportParser(
+                    ilSession::get("tst_import_results_file"),
+                    $newObj,
+                    $this->db,
+                    $this->logging_services->root()
+                );
+                $results->setQuestionIdMapping($qtiParser->getQuestionIdMapping());
+                $results->startParsing();
             }
             $newObj->update();
         }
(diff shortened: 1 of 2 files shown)


kergomard (commented):
Thank you very much for this PR @fehlfer!
I've an additional question on this version of the PR:
- [ ] Your changes concerning the ZIP file [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6335/files#diff-03ef5196fa034eb9a4da3167daf79d8de700679402306e6de30154341c7640e3) seem to change the behavior. Before, if there already was a zip-file, the file would have been added to the existing zip, now it looks to me like, the zip-file would be replaced by one only containing the new files. What am I missing?

Thanks and best,
@kergomard

fhelfer:
> Thank you very much for this PR @fehlfer! I've an additional question on this version of the PR:
> 
> * [ ]  Your changes concerning the ZIP file [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6335/files#diff-03ef5196fa034eb9a4da3167daf79d8de700679402306e6de30154341c7640e3) seem to change the behavior. Before, if there already was a zip-file, the file would have been added to the existing zip, now it looks to me like, the zip-file would be replaced by one only containing the new files. What am I missing?
> 
> Thanks and best, @kergomard

Hi @kergomard,

It seemed to me that this would be an incorrect behaviour because you get an infinite loop of zip folders enclosed in each other.

Best, @fhelfer

kergomard:
Hi @fhelfer 
Sorry, I didn't formulate well:
If you have a zip-file containing the files 'bli.txt' and 'bla.txt', before your changes after passing your function you would have a zip-file containing 'bli.txt', 'bla.txt', and 'results.xls', after your changes the file would only contain 'results.xls'. Am I missing something?

Thank you very much and best,
@kergomard

kergomard:
Hi @fhelfer 
Do you think you could have another look at this?

Thank you very much and best,
@kergomard

mbecker-databay:
Hi @kergomard !
From my point of view, the exports zip file name is generated with the current timestamp. So there should not be any leftover files from other processes in existing zips.
I think the process as you describe it, not having bli and bla, appears correct to me.

fhelfer:
Hi @kergomard,

i reverted the changes made to zip, because the bug i noticed earlier in the PR seems to be fixed by now :)
So now the PR only contains fixes for the bug mentioned in mantis

Best, @fhelfer

dsstrassner:
So, @mbecker-databay and @kergomard, who would be the one, who presses the Rebase and merge button? ;-)

kergomard:
Thank you very much @fhelfer for this PR:
These two lines [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6335/files#diff-9928e816095be3a3b2c72e86e61f040b8d80dddd16e96c42b8fe6a18d458d455R149-R150) led me to believe that there was something fundamentally wrong with the import. The problem is very basic: We are not importing the questions, ...and we are doing so very consciously as we are setting `ilQTIPerser::setIgnoreItemsEnabled(true)`. I will get around to this, but not in the next days. If you want to give it another try, please go ahead. This is also still broken in ILIAS 8.

Thanks again and best,
@kergomard

dsstrassner:
We discussed this at today's TechSquad Meeting and the import of Tests with Results from Random Questions Tests are broken beyond repair within the framework of reasonable use of resources.
We let this open until the JF, where the Abandon of the Feature is discussed and decide afterward, what we will do with this issue.

PR #6337, opened 2023-09-19
Setup/Data: Fix minor version compare

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=38002

If approved, this has to be picked to `trunk`.

### DIFF
--- a/src/Data/Version.php
+++ b/src/Data/Version.php
@@ -1,5 +1,21 @@
 <?php
 
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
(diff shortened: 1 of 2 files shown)


klees (changes requested):
Hi @mjansenDatabay,

would you be so kind a write an automated test for the issue?

Thanks!

mjansenDatabay:
@klees 

1) I fixed the license header
2) I added a new case in the data provider (and provided names for the cases as well)
3) I fixed the checks by adding negated assertions with the corresponding inverted method call.
4) I used `version_compare`, I did not see any reason why the versions are compared manually

klees:
@mjansenDatabay I didn't use `version_compare` because it felt a little to obscure for me. But since all tests still pass I'm actually fine with it now =)

Thanks for the other improvements.

klees (approved):

klees:
Picked to trunk!

PR #6339, opened 2023-09-20
Exclude 1 and 0 from quotation

iszmais:
Temporary solution for https://mantis.ilias.de/view.php?id=38013

The best practice here where to remove the requirement from the WHERE clause, since its trivial in said occurences.
However since this might be a bit time critical to fix an exception for the redunant `WHERE 1=1` or `WHERE 0 = 0` calls might be sufficent for the moment.

### DIFF
--- a/Services/Database/classes/PDO/class.ilDBPdo.php
+++ b/Services/Database/classes/PDO/class.ilDBPdo.php
@@ -251,7 +251,7 @@ abstract class ilDBPdo implements ilDBInterface, ilDBPdoInterface
      */
     public function quoteIdentifier($identifier, $check_option = false)
     {
-        return '`' . $identifier . '`';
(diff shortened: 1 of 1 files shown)


chfsx:
I don't want to introduce such exceptions. from my point of view they are only needed when working with the `ilDBInterface` method `update`. but from my point of view only "real" identifiers should be passed with this method and no fake identifiers like 1=1.

if you write your query yourself anyway, you don't need to quote `1` as a fake-identifier.

1 and 0 are chosen arbitrarily (because they seem to be the most obvious), but i could work with 3=3 or 4=4 as well, in the end we have an infinite list of exceptions.

iszmais:
see ticket: https://mantis.ilias.de/view.php?id=38013

PR #6341, opened 2023-09-20

iszmais:
https://mantis.ilias.de/view.php?id=38015
https://mantis.ilias.de/view.php?id=38017

This resolves both of the bugs above and further allows custom Inputs to use the core based `Multi` flag even for non-single-layered values while simultaneously increase overall robustness.

### DIFF
--- a/Services/Form/classes/class.ilPropertyFormGUI.php
+++ b/Services/Form/classes/class.ilPropertyFormGUI.php
@@ -817,7 +817,7 @@ class ilPropertyFormGUI extends ilFormGUI
             $multi_values = $item->getMultiValues();
             if (is_array($multi_values) && sizeof($multi_values) > 1) {
                 $multi_value = new ilHiddenInputGUI("ilMultiValues~" . $item->getPostVar());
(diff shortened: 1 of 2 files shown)


alex40724:
@iszmais Thanks for the PR. This is definitely a better way to handle the values. But we/I need to test some scenarios before merging. Did you check different usages of multi inputs already?

iszmais:
@alex40724 Thx for your fast response.

I already texted that for SelectInputs and TextInputs (both, Multi `on` and `off`) and for an own heritance of an CombinationInput wich includes a Text and Select Field.

I didnt checked more Input types since these already where not supported be the current/origin Multi select JS and therefore i assumed these Inputs are also not allowed for the multi flag (dont implement `ilMultiValuesItem`).

But i dont know if this might already have a bug for the current state, this woudl need further testing.

Greetings,
@iszmais

iszmais:
Also evry usage in the current ILIAS Core of Multi flagged Inputs only uses Selects or Texts and, at least all i could find, worked as intended with these changes.

PR #6343, opened 2023-09-21
CMPS: adjust global for logger regarding ilComponentUpdatePlugin…

lukastocker:
…Objective setup. ([#37960](https://mantis.ilias.de/view.php?id=37960))

https://mantis.ilias.de/view.php?id=37960


klees:
Thanks!

PR #6345, opened 2023-09-21
T&A Bugfix 0038004: Aufruf von Fragenpools nicht möglich

fhelfer:
https://mantis.ilias.de/view.php?id=38004


kergomard:
Thank you very much for the PR @fhelfer . I will merge it and then apply a strict comparison to it, but I don't think it is worth the roundtrip.

PR #6352, opened 2023-09-22
T&A Bugfix #0037817: Ordering Question (vertical): Upload does not update form

fhelfer:
https://mantis.ilias.de/view.php?id=37817


kergomard:
Thank you very much @fhelfer ! This looks good to me. Will merge, ILIAS 7 only.

PR #6353, opened 2023-09-22
Introduces calendar appointment export limit (1mb)

chlulei:
Fix for https://mantis.ilias.de/view.php?id=31342


matthiaskunkel:
Jour Fixe, 02 OCT 2023: We highly appreciate this suggestion to fix the above mentioned issue. But we prefer to offer this restriction of the calender event export as an additional URL to the existing one. This will offer a sound solution for Google calendar user (who have the mentioned problem) but won't restrict the export for users of other calendars (they will still get all events even above 1MB). Backport to all maintained versions is accepted. Please add a short notice to the release report of all related releases (7.x, 8.x)

smeyer-ilias (approved):

PR #6354, opened 2023-09-22
UI/Launcher: Improved Launcher form input data mapping

matheuszych:
I want to use the `Launcher` to implement [Test: Use Launcher to Start the ILIAS Test
](https://docu.ilias.de/goto_docu_wiki_wpage_979_1357.html), but I noticed that the data mapping for the inputs of the `Launcher` form is not stable. As of now, it is unclear which value belongs to which input. The values are stored in a numeric array where the keys correspond to the order of the inputs inside the form. This makes it difficult to maintain.

Therefore, I propose this improvement, which allows the values to be mapped to the input names.

### DIFF
--- a/src/UI/examples/Launcher/Inline/with_fields.php
+++ b/src/UI/examples/Launcher/Inline/with_fields.php
@@ -32,8 +32,7 @@ function with_fields()
             $ui_factory->input()->field()->checkbox('Understood', 'ok')
     ]);
     $evaluation = function (Result $result, Launcher &$launcher) use ($ctrl, $ui_factory) {
-        $values = $result->value();
-        if ($result->isOK() && reset($values)) {
+        if ($result->isOK() && ($values = $result->value()) && reset($values)) {
             $ctrl->redirectToURL(
                 (string)$launcher->getTarget()->getURL()->withParameter('launcher_redirect', 'terms accepted (' . $launcher->getButtonLabel() . ')')
             );
--- a/src/UI/examples/Launcher/Inline/with_fields.php
+++ b/src/UI/examples/Launcher/Inline/with_fields.php
@@ -75,8 +75,7 @@ function with_fields()
             $ui_factory->input()->field()->password('pwd', 'Password')
     ]);
     $evaluation = function (Result $result, Launcher &$launcher) use ($ctrl, $ui_factory) {
-        $values = $result->value();
-        if ($result->isOK() && reset($values)?->toString() === 'ilias') {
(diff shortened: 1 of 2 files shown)


src/UI/examples/Launcher/Inline/with_fields.php
```diff
@@ -32,7 +32,8 @@ function with_fields()
             $ui_factory->input()->field()->checkbox('Understood', 'ok')
     ]);
     $evaluation = function (Result $result, Launcher &$launcher) use ($ctrl, $ui_factory) {
-        if ($result->isOK() && $result->value()[0]) {
+        $values = $result->value();
+        if ($result->isOK() && reset($values)) {
```
mjansenDatabay:
```suggestion
        if ($result->isOK() && ($values = $result->value()) && reset($values)) {
```

src/UI/examples/Launcher/Inline/with_fields.php
```diff
@@ -75,7 +76,8 @@ function with_fields()
             $ui_factory->input()->field()->password('pwd', 'Password')
     ]);
     $evaluation = function (Result $result, Launcher &$launcher) use ($ctrl, $ui_factory) {
-        if ($result->isOK() && $result->value()[0]->toString() === 'ilias') {
+        $values = $result->value();
+        if ($result->isOK() && reset($values)?->toString() === 'ilias') {
```
mjansenDatabay:
```suggestion
        if ($result->isOK() && ($values = $result->value()) && reset($values)?->toString() === 'ilias') {
```

klees:
Hi @matheuszych,

could you elaborate a little what you mean by "not stable"? Values should be ordered in the result just as the inputs in the form. This should not change if inputs don't change order. Also, you should be able to label the inputs via keys in the array you pass them and then retrieve values accordingly.

In general, we do not use HTML-names for internal processing of data. You even won't be guaranteed that the HTML your form generates will have the exact names that you assign. Background is that fixed names don't play well with composability. If I, e.g. build an combined input to gather some names (firstname, lastname, title, ...) from a user and then want to use two of these to, e.g., gather names of her parents, having a fixed "firstname" as name of an input will lead to surprising results. Hence, users of the Inputs from the UI Framework should not rely on HTML names and instead have other means to capture their inputs.

Kind regards!

matheuszych:
Hey @klees,
The wording 'not stable' was not quiet what I ment. I ment more like 'not practical'. 

I have looked into what you said and realized that there already is a way to achieve exacly what I need. This makes this PR redundant. I did not fully understand the way the Launcher should be used :)

Thank you for your help!

PR #6298, opened 2023-09-05
Trunk: Test: Use Launcher to Start the ILIAS Test

matheuszych:
Featurewiki-Link: https://docu.ilias.de/goto_docu_wiki_wpage_4925_1357.html

PR #6355, opened 2023-09-25
 UI/Launcher: Modal button label customization

matheuszych:
As of right now, you are unable to set the labels for the buttons inside the `Launcher` modal. The default labels **Cancel** & **Save** are not always suitable for every purpose. Therefore I propose this change to make the labels adjustable if needed. If no label is provided, the default value **Cancel** or **Save** will be applied accordingly.

### DIFF
--- a/src/UI/Component/Launcher/Launcher.php
+++ b/src/UI/Component/Launcher/Launcher.php
@@ -23,7 +23,6 @@ namespace ILIAS\UI\Component\Launcher;
 use ILIAS\UI\Component\Component;
 use ILIAS\UI\Component\Chart\ProgressMeter\ProgressMeter;
 use ILIAS\UI\Component\Symbol\Icon\Icon;
-use ILIAS\UI\Component\Input\Container\Form\Form;
 use ILIAS\UI\Component\Input\Field\Group;
 use ILIAS\UI\Component\MessageBox;
 use Psr\Http\Message\ServerRequestInterface;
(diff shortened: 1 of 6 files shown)


dsstrassner:
hi @thibsy, could you estimate a response to this PR? I my understanding, we need this for the feature in PR #6298.

thibsy:
Hi @Dennis1993, sorry this was not on my radar since @klees has been assigned, but I'll try to look into it this week. Please note that the interface part is missing, which also needs to pass JF. However, since this is a trivial change and the FR is probably for ILIAS 9, we can announce this change after the PR has been merged.

dsstrassner:
@thibsy thanks for your quick response. I asked you, because you  set the labels.

thibsy (approved):
Hi @matheuszych

I am aware this functionality is linked to a feature scheduled for ILIAS 9, which is why we cannot wait until next JF to announce this change. Luckily the changes are minimal and all LGTM, which allows us to announce it afterwards. I will therefore merge this if the checks pass and label it for next JF.

Kind regards,
@thibsy

dsstrassner:
Thanks a bunch @thibsy!

matthiaskunkel:
Will the modified labels become part of the language files? Or are they stored in another way?

klees:
Hi @matthiaskunkel,

this introduces the possibility to set according labels of the modals, it does not introduce actual values they are set to. These values need to be determined by the component that is using the launcher (in this case: the T&A) and I absolutely expect the component to use the language files. The default values ("save" and "cancel") that the UI framework uses are in fact read from the lang-files.

Hope this helps!

matthiaskunkel:
Thanks @klees for this clarification.

matthiaskunkel:

PR #6356, opened 2023-09-25
LSO: add error message if "Download Multiple Objects" is used. (#…

lukastocker:
…36631)

This implementation should be temporary, as the multi-download feature has to be handled differently / should not be a property of the container service. A decision about the handling will be made later.

Mantis Ticket: https://mantis.ilias.de/view.php?id=36631

### DIFF
--- a/Modules/LearningSequence/classes/class.ilObjLearningSequenceGUI.php
+++ b/Modules/LearningSequence/classes/class.ilObjLearningSequenceGUI.php
@@ -85,6 +85,7 @@ class ilObjLearningSequenceGUI extends ilContainerGUI implements ilCtrlBaseClass
     public const CMD_SHOW_TRASH = 'trash';
     public const CMD_UNDELETE = 'undelete';
     public const CMD_REDRAW_HEADER = 'redrawHeaderAction';
+    public const CMD_ENABLE_ADMINISTRATION_PANEL = "enableAdministrationPanel";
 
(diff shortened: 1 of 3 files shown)


klees (commented):
Hi @lukastocker,

- [ ] redirect: Why do we redirect [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6356/files#diff-207fe857d1f1d1339bbcc61fcfd235955e151314e30bd4b151380252ab8334b3R223). Couldn't we simply use `false` as third parameter for `setOnScreenMessage` and the just call `manageContent` directly?

The rest LGTM.

Kind regards!

lukastocker:
> Hi @lukastocker,
> 
> thanks a lot for looking into this.
> 
> Please answer the following questions:
> 
>     * [x]  redirect: Why do we redirect [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6356/files#diff-207fe857d1f1d1339bbcc61fcfd235955e151314e30bd4b151380252ab8334b3R223). Couldn't we simply use `false` as third parameter for `setOnScreenMessage` and the just call `manageContent` directly?
> 
> 
> The rest LGTM.
> 
> Kind regards!

Hi @klees ,

you're absolutely right about that. It seems I was thinking way too complicated. Just changed my implementation and successfully tested it on my local instance.

Thanks!

Best wishes,
Luka

klees:
Picked to trunk!

PR #6357, opened 2023-09-25
9/UI/buttons no press transform on active state

catenglaender:
Mantis: https://mantis.ilias.de/view.php?id=37838

# Issue

Buttons appear pressed down on active state

image

# Change

This is a flag of the button mixin that has now been switched to "none", so no transform is set for the active state.

image


Amstutz:
Thx a lot, resolved conflicts and merged.

PR #6359, opened 2023-09-25
Form: fix start and end labels of duration input (36124)

schmitz-ilias:
This PR fixes [36124](https://mantis.ilias.de/view.php?id=36124). The duration input only falls back to the default start- and end-labels if their respective properties are `null`, not if they're empty strings. Empty strings are used when the labels should be skipped.


alex40724:
@schmitz-ilias Thanks for the PR! Merged to trunk.

PR #6365, opened 2023-09-27
Docs: use npm clean-install instead of install

klees:
Hi everybody,

we found that `npm install` does not exactly do what we thought it would do: Instead of using the versions of packages that we we have pinned in our `package-lock.json` it recalculates the versions based on the `package.json`. Hence, we cannot guarantee, that people use the exact versions of packages that we have tested or even that the same versions are used on different systems.

Luckily, there is `npm clean-install`, that does the exact versions defined in the `package-lock.json`.

Take care when using npm, everybody.

Kind regards!

### DIFF
--- a/docs/configuration/install.md
+++ b/docs/configuration/install.md
@@ -389,7 +389,7 @@ This requires that the php dependency manager [composer](https://getcomposer.org
 is available in your $PATH.
 
 ```
-npm install --omit=dev --ignore-scripts
+npm clean-install --omit=dev --ignore-scripts
 ```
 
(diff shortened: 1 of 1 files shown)


thibsy:
There's also a nice [StackOverflow answer](https://stackoverflow.com/a/53325242), which explains the difference between `npm install` and `npm clean-install`, for anybody interested.

matthiaskunkel:
Jour Fixe, 02 OCT 2023 : We highly appreciate this suggestion and accept the PR for trunk. Please add a short notice about this change in the release notes of ILIAS 9.x

PR #6366, opened 2023-09-27
[FIX] CI: install npm packages using clean-install.

thibsy:
Hi all,

As a follow-up to the our findings in #6365, I have also adjusted the CI workflows, so packages are now installed deterministically. I think this can be merged right away.

Kind regards,
@thibsy

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -36,7 +36,7 @@ jobs:
         run: composer install --no-interaction --no-progress
 
       - name: Install Node.js packages
-        run: npm install --ignore-scripts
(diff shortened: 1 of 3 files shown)


thibsy:
There is currently something wrong with the GitHub actions. Setting up the node.js environment using `actions/setup-node@v3` takes extremely long. Checks in the referenced PR also took up to 20 minutes to setup the environment, one workflow even failed because of this.

Note, this is not caused by the changes in this PR.

thibsy:
Seems to be working now. Checks also pass with `npm clean-install`.

PR #6367, opened 2023-09-27
`Refinery`: Refinery overhaul Proposal

lscharmer:
This PR proposes an overhaul of the refinery.
This is a solutiuon to the issues mentioned in the _Refinery/ROADMAP.md_ under _Mid Term_.

### Motivation

Currently I see several issues with the Refnery which this PR tries to address:

#### Undefined properties
The `ProblemBuilder` expects 2 properties to be defined `$error` and `$lng` but this is not enforced so **many** Constraints simply **do not define** them.

#### ProblemBuilder
For the user of the Refinery it is unintuitive to be able to call `->withProblemBuilder(...)` on `$refinery->to()->string()` but not on `$refinery->kindlyTo()->string()`.

Common use cases (at least for me) is to use `$container->http()->wrapper()->query()->retrieve('foo', $refinery->kindlyTo()->int())`. 
To be able to define a problem builder which directly shows a translated message would be very useful, as it is already possible with constraints.

#### Transformation / Constraint

As already recognized in the Refinery/ROADMAP.md the distinction between the two is not clear.
Many features of the constraints are useful for transformations too and there are cases where one wants to treat a Transformation as a Constraint (e.g. to supply to another Constraint which only accepts other Constraints).

#### Duplicated code / Cumbersome to implement

Many Constraints extend from `Refinery\Custom\Constraint` whilst there are by no means a custom Constraint.
IMO this is a symptom of the fact that it is not easy to implement new Constraints or Transformations.
The usages of the different traits tries to prevent duplicated code / share a common concept. But even when implemented correctly they still require some code that is always the same and can be simplified IMO. Additionally and more importantly as they are optional and not enforced some Transformations don't use them and implement their own methods, which are first of all the same code as it is defined in the traits and secondly this leads to an possible inconsistent usage for the user of the Refinery, because one `->withProblemBuilder` call must not behave or provide the same arguments as another.

### New Implementation

The Refinery is split into 3 parts:
- Public API: the new `Transformation` class.
- Constraints: Still an interface but with one method.
- Transformables (formally `Transformation`): An interface with one method.

#### First glimpse

The implementation of _Transformations_ and Constraints is now reduced to as little as possible, for Transformables and Constraints one must only implement **one** method:

```php
class SampleUpperCase implements Transformable
{
    public function($from)
    {
        return strtoupper($from);
    }
}
````

```php
class SampleStringConstraint implements Constraint
{
    public function problemWith($value)
    {
        if (is_string($value)) {
            return null;
        }
        return 'Given value is not a string';
    }
}
```

**From here on: Transformation means the new class not the old interface unless explicitly stated**.

#### Defining a consistent API

The public API is represented by the `ILIAS\Refinery\Transformation` class which is now not an interface but a class.

This class implements all methods of the old Constraint and Transformation interfaces (not as interfaces).

This class is responsible for providing the error builder mechanism, automatic error translation, useful methods (like `ILIAS\Data\Result` does), bridging the 3 concepts and some "legacy" methods for backwards compatibility.

The purpose of the new `Transformable` interface is to provide the core of a Transformation (or business logic).
From a Transformable everything else that is provided by the `Transformation` can be deduced and isn't repeated.

If a value cannot be transformed an exception should be thrown (same as the old transform).

Transformables only require other Transformables not Constraints or the Transformation.

##### Constraint

The purpose of the Constraint is to report a useful error when a value doesn't satisfy the Constraint.

For this the Constraint requires the method `problemWith($value): null|Exception|string` which must return an error string or exception when the value should be rejected and `null` when it is accepted (this is taken from the old `Constraint::problemWith(...)` method).

Constraints only require other Constraints not Transformables or the Transformation.

##### Unifying Transformable and Constraints 

Currently Transformables can be converted to a Transformations but neither can a Transformable be used as Constraint nor a Constraint as Transformable or Transformation.

To accomplish this (beside the technical aspect) the rules the old Transformations and Constraints must be adjusted: 

###### Using Constraints as Transformable

This has not changed from the old rules because the old Constraints where already instances of the old Transformations. But the previous definition was only a _soft_ requirement: In the README Constraints are constraint as they MUST NOT modify the supplied value. But that didn't stop some Constraints to doing so nevertheless :D.
With this approach **it is not possible** to **return** a different value (technically an object could still be modified).

To convert a Constraint to a Transformable: 
```php
$transformable = new \ILIAS\Refinery\ConstraintTransformable(new FooConstraint());
```
The `ConstraintTransformable` class will throw the problem the constraint reported when transforming a value and will return the same value when there is no problem.

The ConstraintTransformable can now, like any other Transformable, be converted to a Transformation.

 ###### Using Transformable as Constraints

For this the Transformation is used. To be able to compose Transformables and Constraints the Transformation class implements both interfaces. With this a Transformable which is converted to a Transformation can be passed to any Constraint that expects another Constraint and vice versa.

The Transformation achieves this by calling ->transform on the given transformable an catch a thrown Exception as the problem to report. The return value of the Transformable is discarded.
This approach was chosen for two reasons:
- It is simple (The transformable doesn't need an extra method for this)
- Many of the old Transformations  rejected values before the actual computation but also in the middle of the computation. With this approach it is ensured that a given value will have no problem with the used Transformable.

As the old Transformations are already defined to not to have any side effects this will be save.

 ###### Using Transformations as Transformable / Constraints

The Transformation implements Transformable and Constraint.

 ###### Groups and building Transformations

To lift Transformables and Constraints to Transformations a new class exists: `BuildTransformation` with two methods: `fromTransformable` and `fromConstraint`. This class is supplied to all groups (instead of `ilLanguage`).
All groups will now look like this:

```php
class Group
{
    public function int(): Transformation
    {
        return $this->build_transformation->fromTransformable(new IntegerTransformation());
    }

    public function isLessThan(int $max): Transformation
    {
        return $this->build_transformation->fromConstraint(new LessThan($max));
    }
}
```

#### Changes required outside the Refinery

All consumer code which uses the `Refinery/Factory` to access the transformations **has to change nothing**. 99% of the changes necessary are contained inside `src/Refinery`.
I could only find 2 instances where something outside of the Refinery has to be changed (because they derive from `Constraint` (or more precisely Custom Constraint):
- `ilAccessibilityDocumentCriterionAssignmentConstraint`
- `ilTermsOfServiceDocumentCriterionAssignmentConstraint`

#### Issues for this approach

As the Transformation class cannot be made final (for compatibility with existing code and to be able to mock it in Unit Tests) people may add new Transformations / Constraints which derive from the Transformation class instead of the corresponding interface which will defeat the whole purpose of this approach.

Transformable: The Transformable interface may confuse people, especially as the Transformations are already a used concept. The Transformable interface + conversion is not as simple as extending from a super class named Transformation.

Similar to the previous: The construction of the classes in the groups is more complicated now (with the BuildTransformation class).

It is not possible to get an error message before supplying an invalid value to a Constraint. This was previously possible and was used in the `ILIAS\Refinery\Logical\Not` class and in the `ILIAS\Refinery\Logical\LogicalOr` class (which makes the class obsolete).

#### Open Todos

All classes in the refinery are already adjusted but not the Test Cases.

The README must be adjusted (Concept, ideas behind the three classes, examples, etc).

There might be a better name for the Transformable interface.

Renaming all old Transformation folder and suffixes to Transformable. With this change there might be more changes outside the Refinery (in src/Data some of the old transformations are instantiated directly).

#### Finally

This PR introduces deep changes for the refinery and as such will require a discussion to validate and adjust if this is an improvement to the Refinery and an implementation which accommodates with the vision that is intended for the Refinery.

Additionally if you see it fit I would be happy to discuss this in a meeting :)

Best regard
@lscharmer


lscharmer:
After a discussion with @klees and @mjansen this proposal wil not be implemented as is.

Instead there will be several new PR's which will address the problems discussed in this PR:

- Fixing undefined properties.
- Change the base class for constraints which extend from custom constraint in the Refinery itself.
- Changing transformations to constraints where necessary (e.g. in `$refinery->kindlyTo()->*`).
- Removing the requirement for constraints to return the same value.
- Removing unused public methods from `Constraint`.
- Fix error reporting for individual problematic constraints.
- Decoupling the Problem Builder concept from `Constraint` (withProblemBuilder, translation callback). Solving the Setup 0 Problem (last to be implementend).

Best regards

@lscharmer

PR #6369, opened 2023-09-29
core: 0037940: fix downloadFile-Links

unibwstabil:
See: https://mantis.ilias.de/view.php?id=37940


alex40724:
Thanks for the PR!

PR #6370, opened 2023-09-29
Registration: Valid Until Should be INT

kergomard:
I think the INT is right here and thus added a cast. See:

See: https://mantis.ilias.de/view.php?id=38057

Thank you very much and best,
@kergomard

### DIFF
--- a/Services/Registration/classes/class.ilAccountRegistrationGUI.php
+++ b/Services/Registration/classes/class.ilAccountRegistrationGUI.php
@@ -481,7 +481,7 @@ class ilAccountRegistrationGUI
 
                         case "relative":
                             $rel = unserialize($code_data["alimitdt"], ['allowed_classes' => false]);
-                            $access_limit = $rel["d"] * 86400 + $rel["m"] * 2592000 + $rel["y"] * 31536000 + time();
+                            $access_limit = (int) ($rel["d"] * 86400 + $rel["m"] * 2592000 + $rel["y"] * 31536000 + time());
                             break;
                     }
                 }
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
Hmmm, I'd prefer to only cast the expression in `Services/Registration/classes/class.ilAccountRegistrationGUI.php:484`:

```php
                        case "relative":
                            $rel = unserialize($code_data["alimitdt"], ['allowed_classes' => false]);
                            $access_limit = (int) ($rel["d"] * 86400 + $rel["m"] * 2592000 + $rel["y"] * 31536000 + time());
                            break;
```

IMO it is **not** possible that the other assignments to `$access_limit` can be of type `float`. But for me it would be also okay to merge and pick (to `trunk`) the suggestion of @kergomard .

@pascalseeland What do you think?

mjansenDatabay (approved):

kergomard:
Yep, that was my first solution. Adding [this line here](https://github.com/ILIAS-eLearning/ILIAS/pull/6370/files#diff-81f281a543d649ee19d5ecfe7621c9735693d65eda2074977314427d7dc7d376R503) that I was not sure about. So just to be sure. ...but then I thought that we were inside a `if ($access_limit)`-block and the change was just smaller, so I went with the one single cast. Both are fine with me and I can change the implementation to whatever you decide.

mjansenDatabay:
@kergomard  Since there is no feedback provided by @pascalseeland , I prefer the `(int)` cast suggested in https://github.com/ILIAS-eLearning/ILIAS/pull/6370#issuecomment-1741324111 . 

Afterwards I'll merge and pick your changes.

kergomard:
Thank you very much  @mjansenDatabay ! Done as requested.

PR #6372, opened 2023-10-02
Core: 0038102: Bei der Bearbeitung der Rechtevorgaben kommt es in manchen Fällen zu einem Fehler

faborg:
See: https://mantis.ilias.de/view.php?id=38102

### DIFF
--- a/Services/Tree/classes/class.ilNestedSetTree.php
+++ b/Services/Tree/classes/class.ilNestedSetTree.php
@@ -1,7 +1,22 @@
 <?php
 
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ ******************************************************************** */
+
 declare(strict_types=1);
-/* Copyright (c) 1998-2009 ILIAS open source, Extended GPL, see docs/LICENSE */
 
 /**
  * Base class for nested set path based trees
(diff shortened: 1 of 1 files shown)


klees:
@fwolf-ilias Feel free to call me on short notice if you want to discuss this.

fwolf-ilias:
@klees Thank you for your offer. Im struggling with the following questions:
- is the `parent` information mandatory for a nested set node (i think yes)
- what is the reason `ilNestedSetTree::getSubtreeInfo` wont delivering it
 (Maybe it seemed irrelevant because it should already be known in a "what is my subtree" context) 
 
Nevertheless i think im going to fix it within ilNestedSetTree::getSubtreeInfo by adding parent.
`ilMaterializedPathTree::getSubtreeInfo` for example is delivering its path, which includes, among other things, this information.

klees:
Hi @fwolf-ilias!

> is the parent information mandatory for a nested set node (i think yes)

I think technically it would not be required to have a `parent` field in the database to implement a [nested set model](https://en.wikipedia.org/wiki/Nested_set_model). But looking into [our implementation](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/Services/Tree/classes/class.ilNestedSetTree.php#L241) it seems save to assume that we need that field. Even if we wouldn't have that field in the database we could for sure determine a parent for every node. Only one exception: the root node of any tree would not have a parent. In the database this seems to be represented by a `0` in the parent field. It seems as this is a special case that is missing in `ilNestedSetTree::getRelation` and could be easily inserted in the beginning: If parent is 0, every node is by definition a child.

> what is the reason ilNestedSetTree::getSubtreeInfo wont delivering it

I also can't think of any reason, besides "not needed".

Only usage side seems to be `ilTree::getRbacSubtreeInfo`, which is only used in `./Services/AccessControl/classes/class.ilObjRole.php` and `./Services/AccessControl/classes/class.ilObjRole.php`. So should be easy to check if it is doing any harm there, I guess. I think we should introduce `parent` in ilMaterializedPathTree::getSubtreeInfo as well, though, so interfaces stays uniform.

Also, I don't think this is a correct fix. Semantics seem to be off: if two nodes have no parent, they are not siblings...

Your approach seems to be better...

Hope this helps!

fwolf-ilias:
Thank you @rklees for your opinion and @faborg for this PR.

PR #6373, opened 2023-10-02
9/UI/keyboard focus outline 37635

catenglaender:
Mantis [37635](https://mantis.ilias.de/view.php?id=37635)

# Issue

Keyboard focus is missing on many elements after Bootstrap removal.

image

# Change

There are now three mixins to choose from when setting the look of the keyboard focus.

## Default

The default that should be used most of the time
Uses ::after to create a contrasting outer border to the outside and the inside of the focus line.

```scss
@use "../correct-relative-path/030-tools/tool_focus-outline" as f;

.most-elements {
    @include f.il-focus();
}
```
image

## Fallback: Outline for Inputs

Fallback for any elements where the ::after contrast border causes issues (e.g. inputs, breadcrumbs).
Uses shadow instead of ::after to create a contrasting outer border.

```scss
@use "../correct-relative-path/030-tools/tool_focus-outline" as f;

input {
    @include f.il-focus-outline-only();
}
```

image

## Fallback: Border for elements touching corners or edges

Fallback for anything where the focus outline would be cut off by the edge of the screen or container.
Uses shadow instead of ::after to create a contrasting outer border.

```scss
@use "../correct-relative-path/030-tools/tool_focus-outline" as f;

.corner-button {
    @include f.il-focus-border-only();
}
```
image


Amstutz:
@catenglaender presented this PR today in the CSS Squad. Thx a lot! He provided a lot of very helpful explanations concerning the usage of the various outline mixins. The session has been recorded and will hopefully be published in the ILIAS Youtube Channel.

PR #6374, opened 2023-10-03
filter: fixed 38103: Deleting a Comment throws error

alex40724:
https://mantis.ilias.de/view.php?id=38103


klees:
Hi @alex40724 and @tfamula,

I think that you introduced that class during the overhaul of the for the UI framework. If I remember correctly this class incorporates some bindings that we could not cleanly incorporate to the UI framework at that time. I'm not really sure what's going on there tbh...

Kind regards!

PR #6375, opened 2023-10-03
33679 - Includes a new Icon for Questions being edited in a question …

jcopado:
…Pool

### DIFF
--- a/Modules/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php
+++ b/Modules/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php
@@ -1383,10 +1383,6 @@ class ilObjQuestionPoolGUI extends ilObjectGUI implements ilCtrlBaseClassInterfa
             $q_gui = assQuestionGUI::_getQuestionGUI('', $this->qplrequest->getQuestionId());
             if ($q_gui->object instanceof assQuestion) {
                 $q_gui->object->setObjId($this->object->getId());
-                $title = $q_gui->object->getTitle();
-                if (!$title) {
-                    $title = $this->lng->txt('new') . ': ' . $this->questioninfo->getQuestionType($q_gui->object->getId());
-                }
                 $this->tpl->setTitle($this->object->getTitle());
                 $this->tpl->setDescription($q_gui->object->getComment());
                 $this->tpl->setTitleIcon(ilObject2::_getIcon($this->object->getId(), 'big', $this->object->getType()));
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @jcopado for the PR!

Now, I'm sorry I didn't see this earlier, but as the maintainer of `ilObject` I take issue with the general approach: I don't like the fact that we are changing the Icon that right now, even if not explicitly specified, seems to be a property of the `object` e.g. you can set `CustomIcons` there, in the context of something (the question) that is not an `object`.

I had a talk with @dsstrassner and @oliversamoila concerning this and we would kindly ask you to:
- [ ] Adapt this PR to NOT change the icon, but to change the code to the effect that it always shows the `title` of the `object` even if we are in the context of a question.
- [ ] To provide us with a list of all views in the Test and the Question Pool that are affected by this change. We will then find a solution for them.

Thank you very much and best,
@kergomard

jcopado:
Hello everyone,

I would need more clarification rearding the change which is now required by the TB.

- [ ]  Adapt this PR to NOT change the icon, but to change the code to the effect that it always shows the title of the object even if we are in the context of a question.

Does this mean that the icon won't be changed?
Does this mean that the change that is now required from the TB is to make the title of the question to always appear?

The issue here to me is the word Object thing. By object you mean the question object, the question pool object, the test object or any object in ILIAS?

I'm sorry but I don't understand what it is now required from us.

kergomard:
Hi @jcopado 
Sorry, that I was not precise enough!
But first: This has absolutely nothing to do with the TB. I'm doing these reviews here in my role as one of the two programmers with the right to merge PRs into the Test & Assessment (the other one is @mbecker-databay ).

So, by `object` now I mean the test or the question pool (the thingy that derives from `ilObject`). You are completely right that in many wordings also a question could be an `object`. I was thinking too much as the maintainer of `ilObject`, sorry.

We will need to see what we do with the icon of the question, but yes, it means that we don't want to change the icon shown at the top of the page (see my explanation above).

At the top of the page it should NOT be the title of the question that is shown, but always the title of the Question Pool or of the Test respectively.

Now, that is why I kindly asked you to look into the second question: This will lead to views, where the question title is nowhere shown on the page which is a rather problematic situation. This is what I meant with:

> * [ ] To provide us with a list of all views in the Test and the Question Pool that are affected by this change. We will then find a solution for them.

I hope I was clearer now. Don't hesitate to contact me directly on Discord to discuss this.

Thanks again and best,
@kergomard

chfsx:
TB: I see that the related mantis-issue ticket https://mantis.ilias.de/view.php?id=33679 has been closed in the meantime, so I am closing this PR as well. If this is not correct, please reopen the PR rebased.

kergomard:
Thanks @chfsx for closing this and sorry @jcopado this good completely lost in my stack.

It was correctly to close this, together with @dsstrassner and @oliversamoila we decided to make things a little clearer, by adding the title of the Test/Questionpool to the title of the Question when showing the latter. I had kept this PR alive to remind me of this, but then ... . Please find the alternative solution [here](https://github.com/ILIAS-eLearning/ILIAS/commit/712837195db8b4976864ab34287ef9e48c19d25a).

Sorry again and best,
@kergomard

PR #6377, opened 2023-10-04
T&A: remove signature for submissions in tests

okaufman:
This removes the option to digitally sign a submission and the corresponding plugin-slot.
It implements this request: 
https://docu.ilias.de/goto_docu_wiki_wpage_7661_1357.html

I still have a question:
Inside the database still is a column 'sign_submission' on the table tst_tests. I am unsure if/how to remove it. There is also a reference to that column in ilObjTestMainSettingsDatabaseRepository.


kergomard:
Thank you very much for the PR @okaufman !

I looked through it and it looks good to me.

For your question:
- [ ] Please drop the table column by adding a corresponding step to the [database update file for ILIAS 9](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Modules/Test/classes/Setup/class.ilTest9DBUpdateSteps.php). You find examples there e.g. in step5. Please make sure to also implement the checks for the existence of the column.
- [ ] Please then remove [the corresponding line of the select query in `ilObjTestMainSettingsDatabaseRepository`](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Modules/Test/classes/MainSettings/ilObjTestMainSettingsDatabaseRepository.php#L95). This would be the reference you mention above, right?

Thanks again and best,
@kergomard

okaufman:
Thank you for your help @kergomard 
I made the changes you asked for.

kergomard:
Hi @okaufman 

I was just talking with Oliver Samoila when I saw this and he reminded me of something important, that I had overlooked, so I've to ask you for one more favor: Could you please remove the language variables, we don't need anymore from the language files?

I only found the following two:
* sign_submission
* sign_submission_info

Then we are good to go and I will merge immediately.

Thank you very much and best,
@kergomard

okaufman:
Okay, I removed the lang variables.
Thank you for merging.
Best wishes Olivia

kergomard:
And a final "Thank you"! I merge this now.

PR #6378, opened 2023-10-04
Fix: 0038109: Wenn es obligatorische Fragen in einem Test gibt, wird …

faborg:
…ein Fehler geworfen und der Test nicht geöffnet

### DIFF
--- a/Modules/Test/classes/tables/class.ilListOfQuestionsTableGUI.php
+++ b/Modules/Test/classes/tables/class.ilListOfQuestionsTableGUI.php
@@ -16,6 +16,9 @@
  *
  *********************************************************************/
 
+use ILIAS\UI\Factory;
+use ILIAS\UI\Renderer;
+
 /**
 *
 * @author Helmut Schottmüller <ilias@aurealis.de>
@@ -35,6 +38,8 @@ class ilListOfQuestionsTableGUI extends ilTable2GUI
     protected ?bool $obligationsNotAnswered = false;
 
     protected ?bool $finishTestButtonEnabled = false;
+    protected Factory $factory;
+    protected Renderer $renderer;
 
     public function __construct($a_parent_obj, $a_parent_cmd)
     {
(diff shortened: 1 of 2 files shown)


faborg:
See: https://mantis.ilias.de/view.php?id=38109

kergomard:
Thank you very much for the PR @faborg !

This looks right, but as you need to change it anyways, to fix the tests, I would have a few small additional wishes:

- [ ] So, one is: Could you please fix the unit tests.
- [ ] Could you please alias the imports to `UIFactory` and `UIRenderer`. The reason for the first alias, I think, is self evident, the reason for the second one, less so. I simply like to make it very visible, they belong together.
- [ ] Coul you please correspondingly rename the class properties to `ui_factory` and `ui_renderer`.
- [ ] And finally: Could you please integrate them with `$DIC['ui.factory']` and  `$DIC['ui.renderer']`. One day we might have a discussion about this, but for the time being, I would suggest to avoid the convenience functions. This would also allow us a simpler mocking, if need be.

Thanks again and best,
@kergomard

kergomard:
Thank you very much for the fast update @faborg !

Now, there is one last little issue (and I have not made that explicit above, sorry): You will need to change the calls [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6378/files#diff-be84d55865e90bf8b0f37ef1f6b236dfd32c45a1b811da55fd66137015d90c37R165) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6378/files#diff-be84d55865e90bf8b0f37ef1f6b236dfd32c45a1b811da55fd66137015d90c37R170) to conform to the new names of the properties.

I will then merge immediately and cherry-pick to trunk (this is ILIAS 8 and trunk only).

Thanks again and best,
@kergomard

kergomard:
For documentation purposes: This is only ILIAS 8. No cherry-pick to trunk needed.

faborg:
Thank you :+1:

PR #6380, opened 2023-10-04
Mantis 0038043: ilDatePresentation::formatDate may return null instea…

mbecker-databay:
…d of string


schmitz-ilias:
Merged and picked to 9 and trunk. Thanks for the fix!

PR #6382, opened 2023-10-04
[FEATURE] UI: add chunked-upload for file-inputs.

thibsy:
Hi folks,

I am currently working on a FR for ILIAS 9 (see https://docu.ilias.de/goto_docu_wiki_wpage_6387_1357.html), which introduces a customisable upload-limit that should be applied to all `Field\File` inputs implicitly.

The current implementation of `Field\File` inputs is either limited to a "local" upload-limit, explicitly set by `withMaxFileSize()`, or implicitly, to the PHP-limit (the smaller of both ini-options `post_max_size` and `upload_max_filesize`). Local limits exceeding the PHP-limit will fallback to the PHP-limit silently, making it impossible to upload larger files than this.

In order to upload files exceeding this limit, we need to implement a mechanism called chunked-upload, where files are sent to the server in smaller pieces where they are assembled back together again, so to avoid this limitation.

The new feature will take advantage of this mechanism, but for full support all `Field\UploadHandler` implementations need to be able to provide such a mechanism. However, because of timeline issues, I have decided to implement this feature following the same principle as `Field\File::withMaxFileSize()`, where invalid upload-limits silently fallback to the next-best value. I have added a roadmap-entry so we can tackle this in a later iteration. The new priority is: `local limit > custom global limit > PHP limit`, whereas the local and custom limit will only apply, if they are smaller than the PHP limit or the corresponding upload-handler supports chunked-uploads.

The custom global limit can be injected in the initialisation of the UI framework when creating the `Input\UploadLimitResolver`, leaving the task of determining this limit up to ILIAS.

Kind regards,
@thibsy

### DIFF
--- a/Services/Init/classes/Dependencies/InitUIFramework.php
+++ b/Services/Init/classes/Dependencies/InitUIFramework.php
@@ -267,7 +267,8 @@ class InitUIFramework
                             $c["refinery"],
                             $c["ui.pathresolver"],
                             $c["ui.data_factory"],
-                            $c["help.text_retriever"]
+                            $c["help.text_retriever"],
+                            $c["ui.upload_limit_resolver"]
                         ),
                         new ILIAS\UI\Implementation\Component\Symbol\Glyph\GlyphRendererFactory(
                             $c["ui.factory"],
@@ -277,7 +278,8 @@ class InitUIFramework
                             $c["refinery"],
                             $c["ui.pathresolver"],
                             $c["ui.data_factory"],
-                            $c["help.text_retriever"]
+                            $c["help.text_retriever"],
+                            $c["ui.upload_limit_resolver"]
                         ),
                         new ILIAS\UI\Implementation\Component\Symbol\Icon\IconRendererFactory(
                             $c["ui.factory"],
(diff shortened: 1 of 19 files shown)


thibsy:
Hi @klees 

I've implemented the according FR now and could make chunked-uploads work in one of our `UploadHandler`s. Injecting a custom global limit also works.

Please note the change where I use [~90% of the PHP-limit](https://github.com/ILIAS-eLearning/ILIAS/pull/6382/files#diff-f3892e7d92b0895c6b153d2f1754b755fc888a4b5199b1d4f8784a7344b81d5fR952) to inject as the chunk-size. I noticed that file-uploads failed when using the exact size-limit or slightly less and could make it work with this amount of bytes when testing.

Kind regards,
@thibsy

klees (approved):
Hi @thibsy,

thanks for implementing this. Thanks also for [small improvements](https://github.com/ILIAS-eLearning/ILIAS/pull/6382/files#diff-11283614eb1a80aab33a9d2c520f66aa468f09f15caa5ee7518123e2091df16fR50).

- [x] Magic Value: Please turn [that `0.9`](https://github.com/ILIAS-eLearning/ILIAS/pull/6382/files#diff-f3892e7d92b0895c6b153d2f1754b755fc888a4b5199b1d4f8784a7344b81d5fR952) into a `const` on the `class`, which also gives a nice opportunity to add the explanation from the PR to the code.

Rest LGTM, go ahead and press merge after the change =)

Kind regards!

matthiaskunkel:

PR #6383, opened 2023-10-04
fix several TypeErrors and makes the code more DRY

unibwstabil:
See: https://mantis.ilias.de/view.php?id=38114


kergomard:
Thank you very much for the PR @unibwstabil and a very special "Thank you" for removing code duplication!
This looks good and I will merge and cherry-pick to trunk.

PR #6384, opened 2023-10-05
Search: Fix User Search

kergomard:
See: https://mantis.ilias.de/view.php?id=38005

This is ILIAS 7 only, it already works for ILIAS 8 and above.


smeyer-ilias:
Thanks @kergomard

PR #6385, opened 2023-10-05
OrgU: Fix Access Rights to Import Sub-Tab

kergomard:
I think, this would be the right solution, but this is ILIAS 7 only for ILIAS 8 the access fails earlier, as `ilObjOrgUnitGUI::getTabs()` will never return anything, if you are not root. The problem being that the right to check [in this line](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/Modules/OrgUnit/classes/class.ilObjOrgUnitGUI.php#L477) should not be 'visible, read', but 'visible,read' (without space after the comma). But this fix is still not enough.

See: https://mantis.ilias.de/view.php?id=29896


klees:
Thanks a lot!

PR #6388, opened 2023-10-06
9/object/feature tile image irss

kergomard:
This PR moves the Tile Images to the IRSS.

@chfsx : You find the changes for the FileTypeTileImages in [this commit](https://github.com/ILIAS-eLearning/ILIAS/commit/e30adf534b7458550b3a0145a214b9aa63f44bfe).

The merge of this PR depends on: https://github.com/ILIAS-eLearning/ILIAS/pull/6151

Best,
@kergomard


chfsx:
@kergomard thanks a lot, LGTM!

alex40724:
@kergomard Thanks for this change. The ilObjectTypeSpecificProperties interface seems to be a general object related interface, is this correct? If so, will it come with documentation and a JF discussion? This might be something affecting performance in the repository.

kergomard:
Hi @alex40724 

Yes, it will come with documentation. Right now, this is just a first step driven by specific needs. Why do you believe there would be a JF decision concerning this? At a certain point, hopefully very early in the development of ILIAS 10, the whole interface will be ready and would then be presented to everyone and discussed in bigger circles. My hope right now is that this will give us a better handle on performance by reducing calls to the database and clarifying interfaces, but their is still quite some work to be done.

Best,
@kergomard

alex40724:
@kergomard I think a discussion is needed at some point, to ensure developers are aware about the performance implications of using interfaces like `ilObjectTypeSpecificPropertyModifications` and `ilObjectTypeSpecificPropertyProviders` since they are potentially called for each single item in the repository presentation (or other repository object lists, like search, dashboard, ...). These implementations must be lightweithed, otherwise performance will drop substantially. This current change is using the ilAccess preloading mechanism to some extend, which is good. If the rest of the implementation doesn't do any heavy DB requests or other things, it will be ok. But in general developers might not be aware of these things when implementing these interfaces.

kergomard:
`RBAC` is on my todo list. I don't believe the current approach to be the right one. I can just not do everything at the same time. So yes: We will have a few moving parts at the foundation of ILIAS, but I think this is very much needed. I became maintainer of `ilObject` and `RBAC`, with a program aimed at improving, not just managing them.

kergomard:
I merge this now. Thanks to everybody who helped!

PR #6390, opened 2023-10-06
fix: Undefined array key "frac_helper"

faborg:
See: https://mantis.ilias.de/view.php?id=38155


kergomard:
Thank you very much @faborg for your PR!

I would kindly ask you to change the solution for this issue:
- [ ] Please remove the `strlen` completely and replace it with an `isset($userdata[$result]["frac_helper"]) && $userdata[$result]["frac_helper"] !==''`. The reasoning behind it: This will simply break with the Update to ILIAS 9 as `strlen()` does not accept `null` anymore. But actually we don't care how long the string is, we just care that there is something.

Thanks again and best,
@kergomard

kergomard:
Thank you very much @faborg. I will merge and cherry-pick to trunk.

PR #6392, opened 2023-10-08
Taxonomy Settings: Remove Legacy Table, use Panels, Move Activation to Additional Feature Section

alex40724:
Feature Wiki
https://docu.ilias.de/goto_docu_wiki_wpage_7694_1357.html
and
https://docu.ilias.de/goto_docu_wiki_wpage_7696_1357.html

This moves the taxonomy listing from a legacy table to a listing panel. Activation is generally under the "Additional Features" section, see FW above.

@maxbecker
The second commit contains the quesstion pool changes: https://github.com/ILIAS-eLearning/ILIAS/pull/6392/commits/89ec33380e1db8e6462ae8892f4ff2ab4913dd44

This integrates the general activation settings according to https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Services/Taxonomy/README-technical.md#taxonomy-settings-in-repository-objects-ilias-9 under "Additional Features".

This also removes the direct instantiation of ilObjTaxonomyGUI and uses the service available through DIC. I did not touch the "Show/Taxonomy Filter as Navigation Tree" setting. Not sure if this is to be abandoned.

### DIFF
--- a/Modules/Category/classes/class.StandardGUIRequest.php
+++ b/Modules/Category/classes/class.StandardGUIRequest.php
@@ -87,4 +87,10 @@ class StandardGUIRequest
     {
         return $this->str("term");
     }
+
+    public function getTaxId(): int
+    {
+        return $this->int("cat_tax_id");
+    }
+
 }
--- a/Modules/Category/classes/class.ilObjCategoryGUI.php
+++ b/Modules/Category/classes/class.ilObjCategoryGUI.php
@@ -29,13 +29,14 @@ use ILIAS\Category\StandardGUIRequest;
  * @ilCtrl_Calls ilObjCategoryGUI: ilPermissionGUI, ilContainerPageGUI, ilObjUserGUI, ilObjUserFolderGUI
  * @ilCtrl_Calls ilObjCategoryGUI: ilInfoScreenGUI, ilObjStyleSheetGUI, ilCommonActionDispatcherGUI, ilObjectTranslationGUI, ilObjectContentStyleSettingsGUI
  * @ilCtrl_Calls ilObjCategoryGUI: ilColumnGUI, ilObjectCopyGUI, ilUserTableGUI, ilDidacticTemplateGUI, ilExportGUI
- * @ilCtrl_Calls ilObjCategoryGUI: ilObjTaxonomyGUI, ilObjectMetaDataGUI, ilContainerNewsSettingsGUI, ilContainerFilterAdminGUI
+ * @ilCtrl_Calls ilObjCategoryGUI: ilTaxonomySettingsGUI, ilObjectMetaDataGUI, ilContainerNewsSettingsGUI, ilContainerFilterAdminGUI
  * @ilCtrl_Calls ilObjCategoryGUI: ilRepositoryTrashGUI
  * @ingroup      ModulesCategory
  */
-class ilObjCategoryGUI extends ilContainerGUI
+class ilObjCategoryGUI extends ilContainerGUI implements \ILIAS\Taxonomy\Settings\ModifierGUIInterface
 {
     public const CONTAINER_SETTING_TAXBLOCK = "tax_sblock_";
+    protected \ILIAS\Taxonomy\Service $taxonomy;
 
     protected ilNavigationHistory $nav_history;
     protected ilHelpGUI $help;
@@ -81,12 +82,14 @@ class ilObjCategoryGUI extends ilContainerGUI
                 ilObjectServiceSettingsGUI::INFO_TAB_VISIBILITY,
                 '1'
             );
+
         }
         $this->cat_request = $DIC
             ->category()
             ->internal()
             ->gui()
             ->standardRequest();
+        $this->taxonomy = $DIC->taxonomy();
     }
 
     public function executeCommand(): void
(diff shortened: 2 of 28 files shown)


kergomard (changes requested):
Hi @alex40724 
Thank you for this PR. I only looked at the parts for the test and I would kindly ask you to change the following:
- [x] Please use short array syntax.
- [x] Please move all name spaces to a use-statement and don't address them inline.
- [x] Please move [this access of the `$DIC`](https://github.com/ILIAS-eLearning/ILIAS/pull/6392/files#diff-79341fc60c03ba53809fb097ae04d5d5a3e597a46a47406f01dd0fed78e926c2R88) together with all other accesses a few lines further up.
- [ ] Please remove the 'Show Taxonomies'-Setting (unless you have discussed something different with @mbecker-databay ). As I understand it, this is part of this feature request (see: https://docu.ilias.de/goto_docu_wiki_wpage_7696_1357.html). I've already removed the "Taxonomy Filter as Navigation Tree' a while ago.

Thanks and best,
@kergomard

alex40724:
I took care of the first three issues in the T&A code changes.

With this PR I provide the integration of the taxonomy settings screen into the question pool settings, but I will not take care of the legacy setting / the integration of the object service fully here. This would require a) a migration of the setting (if wanted), b) taking care of copy/import/export processes and c) replace the calls to the old legacy settings methods. I also noticed that the skill management setting is legacy too (not using ilObjectServiceSettingsGUI), so I think d) moving this to ilObjectServiceSettingsGUI would be appropriate, too. However this would include a lot of code changes with different possible alternatives depending on the maintainers preference, especially c). This is beyond what I can currently offer.

alex40724:
Btw. that last commit triggered breaking unit tests in unrelated T&A tables:
1) ilTestPassOverviewTableGUITest::testActiveId
Pimple\Exception\UnknownIdentifierException: Identifier "ui.factory" is not defined.

/home/runner/work/ILIAS/ILIAS/libs/composer/vendor/pimple/pimple/src/Pimple/Container.php:105
/home/runner/work/ILIAS/ILIAS/src/DI/UIServices.php:37
/home/runner/work/ILIAS/ILIAS/Modules/Test/classes/tables/class.ilTestPassOverviewTableGUI.php:53
/home/runner/work/ILIAS/ILIAS/Modules/Test/test/tables/ilTestPassOverviewTableGUITest.php:53

Running the tests locally does not trigger these errors, so this seems to depend on the unit test execution sequence.

kergomard:
Hi @alex40724 
There is still [one inline Namespace left](https://github.com/ILIAS-eLearning/ILIAS/pull/6392/files#diff-79341fc60c03ba53809fb097ae04d5d5a3e597a46a47406f01dd0fed78e926c2R51).
Max fixed the Tests, if I'm getting this right.

So, there is the last issue:
Why don't we just move the current setting for filtering by taxonomy to the "Additional Features" section and make the subtab "Taxonomy" depend on that?

Best,
@kergomard

alex40724:
> Hi @alex40724 There is still [one inline Namespace left](https://github.com/ILIAS-eLearning/ILIAS/pull/6392/files#diff-79341fc60c03ba53809fb097ae04d5d5a3e597a46a47406f01dd0fed78e926c2R51). Max fixed the Tests, if I'm getting this right.

The namespace issue is fixed.
 
> So, there is the last issue: Why don't we just move the current setting for filtering by taxonomy to the "Additional Features" section and make the subtab "Taxonomy" depend on that?

I removed the original setting. It is already under "Additional Features" due to the use of ilObjectServiceSettingsGUI. The subtab is already aware of the setting, as documented. It will not appear, if the taxonomy service is deactivated in the pool settings.

kergomard:
Hi @alex40724 
Thank you very much for moving the last namespace import!

Now, I don't think we can simply remove the Checkbox for the setting. Right now we save a field 'show_taxonomies' in the table `qpl_questionpool`. All preexisting settings would be lost. And then we also use `ilTestQuestionPool::getShowTaxonomies()` to define if we would like to show the filter on the list of questions.

As far as I can see, we need another solution.

Thanks and best,
@kergomard

alex40724:
This is basically what I wrote in my comment https://github.com/ILIAS-eLearning/ILIAS/pull/6392#issuecomment-1752732560 above. There are several ways how to deal with object settings in a component, use a wrapper/manager class, access them in set/getters directly, remove the setter/getter but don't user a wrapper class, ... I will not do this part, since I am unaware of the maintainers preferences with these things. An I also do not want to dig to deep into the T&A code currently.

kergomard:
Hi @alex40724 
I rephrase my question from [above](https://github.com/ILIAS-eLearning/ILIAS/pull/6392#issuecomment-1752908566): Do the taxonomies in the test question pool continue working as they do today without any of your changes under `Modules\TestQuestionPool`?

Best,
@kergomard

alex40724:
No, the old integration using the legacy table via ilObjTaxonomyGUI for the taxonomy settings directly will not work. But this change is handled with this PR. Missing is the full integration of ilObjectServiceSettingsGUI. Other parts of the taxonomy service did not change (they will in the future, since everything needs to be moved to the DIC published API). This is just about the setting (Pool vs Object, open) and the integration of the manage screens (Pools vs. Taxonomy, done).

kergomard:
Hi @alex40724 
I don't think the legacy table has anything to do with it. We cannot implement this in the `TestQuestionPool` with your current proposal, so I'm actively searching for good options to allow the improvement to go through with minimal fuzz for everybody.

If I understand this correctly, the whole problem is [this `if`-statement here](https://github.com/ILIAS-eLearning/ILIAS/pull/6392/files#diff-01b4767940262e6f90d648b13b2a9f4d5eab80e8c29a713fdeca4a7955167841R60). If you would remove the activation checks from the `addSubTab()`-function things would work (alternatively, but I personally don't like this option, you could also provide a `skip_checks` or `always` or `force` parameter). We could then simply check the need for the sub tab on our side and things would work. Am I missing something?

Thank you very much and best,
@kergomard

alex40724:
@kergomard  So you do not want to integrate ilObjectServiceSettingsGUI, but keep the legacy setting? I can re-add these old lines, remove the ilObjectServiceSettingsGUI part and add the tab in the pool code directly (as it was similar before).

kergomard:
If I'm understanding you right: Yes, this would be what we would want. I would then take care of moving the setting down to the section "Additional Features" and making the sub-tab dependent on our setting and we have a working solution that should be feasible for everyone without much pain.

kergomard (approved):
Thank you very much @alex40724 !

Now this looks and works right for the TestQuestionPool. Thank you very much!

On a side note: 
- The Taxonomies do not work on PHP8.2 as there is a missing property, declaration in `UsageManager`. `UsageManager::db_repo` is set without declaring it first. This happens in quite a few of the Components you have changed lately, e.g. the `MediaTypeManager`. I kindly ask you to test on PHP8.2.

Thanks and best,
@kergomard

mbecker-databay (approved):
I am fine with this!

alex40724:
Merged manually due to conflict.

PR #6394, opened 2023-10-09
Remove timing functions from toast

iszmais:
In repsonse to the rejection of the proposal of individual timing function per toast (https://github.com/ILIAS-eLearning/ILIAS/pull/6164) this removes all occurences prepared for that in consequence.

### DIFF
--- a/Services/Awareness/classes/Provider/AwarenessToastProvider.php
+++ b/Services/Awareness/classes/Provider/AwarenessToastProvider.php
@@ -84,9 +84,7 @@ class AwarenessToastProvider extends AbstractToastProvider
                     $this->if->identifier(self::PROVIDER_KEY . '_' . $this->dic->user()->getId()),
                     $this->dic->language()->txt('awareness_now_online')
                 )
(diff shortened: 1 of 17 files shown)


iszmais:
@klees im having some conceptional issues here targeting the UI factory.
Imo the decision of the coordinators implies the complete removal of said attributes from the toast (to prevent a misleading usage of a function that does nothing). However since the timing of toast ist still an established customizable value inside ILIAS this also means that that setting needs to be applied somewhere else and the only and most fitting central spot for that i found seems to be the renderer itself.

Unfortunatly the rendering process connected to the ui factory does not seem to allow dependencies to global scope stuff such as setting, or at least is not able to test those properly.

Is that intended?
If yes, whats is the desired workflow to append dynamic property to a component within the ui factory, when this property should not be indivdually adaptable?
If no, should such extention target the whole example rendering testing area or should be specified for target components and when yes where?

Greetings,
@klees

iszmais:
@klees i fixed the tests and extended all depending mocks.

Feel free to review.

PR #6397, opened 2023-10-09
9/UI/image responsive max width fix 37743

catenglaender:
Fixes
https://mantis.ilias.de/view.php?id=37743
and in extension
https://mantis.ilias.de/view.php?id=38123
https://mantis.ilias.de/view.php?id=38122

# Issue

Responsive images can overflow their container. This is very noticeable on the UI Component Items:
image

# Changes

The existing img-responsive mixin now also generates a utility css class that is already in use by the image UI component.

image

We might want to discuss...

- if we even need an img-responsive class. Shouldn't all images always be responsive? When do we ever want the image to not fit inside its container?
- how/where to generate CSS utility classes in one clearly defined location as CSS utility classe are only allowed as an exception to the rules stated in the SCSS Guidelines: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/templates/Guidelines_SCSS-Coding.md

### DIFF
--- a/templates/default/050-layout/_layout_responsive-img.scss
+++ b/templates/default/050-layout/_layout_responsive-img.scss
@@ -3,3 +3,7 @@
     max-width: 100%; // Part 1: Set a maximum relative to the parent
     height: auto; // Part 2: Scale the height according to the width, otherwise you get stretching
 }
+
+.img-responsive {
+    @include img-responsive();
+}
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -1,4 +1,7 @@
 @charset "UTF-8";
+/*
+* Dependencies
+*/
 /*!
  * Datetimepicker for Bootstrap 3
  * version : 4.17.37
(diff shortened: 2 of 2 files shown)


yvseiler:
Thanks a lot @catenglaender. Interesting questions... I'll give my take on question 1 "When do we ever want the image to not fit inside its container?":

As long as the container specifies the space, the image itself should always be responsive and use that space. For our standard ILIAS icons, we basically decided against offering them in a fixed size, because we felt that the place where they are embedded should decide how big they are displayed. However, we had the case in Cards and Tables that icons took as much space as possible, making the icon huge (f.e. https://mantis.ilias.de/view.php?id=32426). Here it makes sense that the icon gets a maximum size. If the container solves this appropriately, all the better (I think this was always the solution in these issues). I can imagine that there are places where we have to temporarily work with a fixed size image until a suitable solution is implemented via the UI framework or the container. But the goal should always be to have a container that offers the right size for that place (both for maximum and minimum size - an icon that is only 2px wide doesn't make sense then either).

Amstutz:
Thx a lot. 

Concerning your question; I have a hard time thinking of a non-responsive example of outside user-generated content. However, in user-generated content we seem to have all kinds of (strange) use cases. There are certainly also such, requiring fixed size images. 

Thx for your fix!

PR #6395, opened 2023-10-09
9/UI/symbol/icon width on mobile 38166

catenglaender:
https://mantis.ilias.de/view.php?id=38166

# Issue

Some icons in the UI component entity or item collapse to a 0 px width on small screens / mobile breakpoint. 

This is due to the svgs not having any witdh information saved with  them. Max-width: 100% of a responsive SVG apparently turns into 0 px  width.

# Change

Icons now counteract the max-width: 100% responsive safety measure meant for images. They now always have the size defined for small, medium and large icons.

image

PR #6400, opened 2023-10-11
9/UI/item drodown pull 38054

catenglaender:
Mantis: https://mantis.ilias.de/view.php?id=38054
This should be merge before testing and merging: https://github.com/ILIAS-eLearning/ILIAS/pull/6395

# Issue

Below the mobile breakpoint, the dropdowns on UI items in the dashboard jump right to the end of the inner item content. They need to be pushed all the way to the edge on the right.

This only effects KS UI items, not UI entity or legacy item despite them looking very similar.

image

# Change

Removed the media query that would shrink the size of the display: table-cell .media-body container.

image

# Bigger Picture

As it looks like the UI entity will replace the UI item in the long run, I decided against refactoring the UI item to be like the UI entity. Also I forgot that the Bootstrap 3 media object has already been brought back for the page content headline icon, so we might as well use it here until we progress with a page content revision.

### DIFF
--- a/templates/default/030-tools/_tool_clearfix.scss
+++ b/templates/default/030-tools/_tool_clearfix.scss
@@ -1,12 +1,9 @@
 @mixin clearfix() {
-    &:before,
-    &:after {
-      display: table; // 2
-      content: " "; // 1
-    }
-    &:after {
-      clear: both;
-    }
+  &::after {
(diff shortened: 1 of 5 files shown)


Amstutz:
Thx a lot!

PR #6402, opened 2023-10-12
AccessControl: Fix return value of ilRbacAdmin::assignUserLimited() (#37627)

tbongers-cat:
https://mantis.ilias.de/view.php?id=37627

Fixes the return value of `ilRbacAdmin::assignUserLimited()`, which is expected to be boolean in `ilCourseRegistrationGUI::add()`


kergomard:
Thank you very much for the PR @tbongers-cat ! This looks right to me. I will merge and cherry-pick to trunk.

PR #6406, opened 2023-10-16
[FIX] 0038223: Undeletable filter in Trash

chfsx:
This fixes the mantis issue: https://mantis.ilias.de/view.php?id=38223


fwolf-ilias:
Thank you very much @chfsx

PR #6408, opened 2023-10-16
SVY: adjust english language variable for deleting...

lukastocker:
... all participant data to avoid confusion about the functionality, as user could think that the old phrasing could lead to thinking that every participant gets deleted instead of their data. (35538)

Mantis Ticket: https://mantis.ilias.de/view.php?id=35538


alex40724:
Thanks for the PR!

PR #6409, opened 2023-10-16
Navigation: 37409 check if type of item is a known plugin id

shecken:
Suggestion to solve the problem of
https://mantis.ilias.de/view.php?id=37409

System should check if type of item is a known (installed) plugin.


alex40724:
Thanks!

PR #6410, opened 2023-10-16
Search: 38092 cast db result for user id to int during result entry creation

shecken:
Suggestion to solve
https://mantis.ilias.de/view.php?id=38092

User is a string in db result and must be a integer.


smeyer-ilias:
Thanks @shecken

PR #6411, opened 2023-10-16
IndividualAssessment: 36471 if mapping was succesfull cast new id to int

shecken:
https://mantis.ilias.de/view.php?id=36471


klees:
Picked to trunk. Thanks!

PR #6412, opened 2023-10-16
TA: 37233, use internal wrapper instead of POST

nhaagen:
https://mantis.ilias.de/view.php?id=37233


kergomard:

This looks right and far more elegant. ...although I'm unsure where we really use this. Plugin? Will merge and cherry-pick to trunk.

Thanks and best,
@kergomard

PR #6413, opened 2023-10-16
UI: Fix Race Condition When Loading Image

kergomard:
Hi @thibsy 
Please find the announced PR for the problem when loading images here.

I label this an improvement as there is no corresponding issue.

Best,
@kergomard


thibsy (approved):
Hi @kergomard 

Thx for testing this, I too have already stumbled upon an issue once, where the `jQuery.load()` function fires before `jQuery.ready()`, which is not the intended behaviour. 

I found an according [StackOverflow question](https://stackoverflow.com/questions/61910610/window-onload-fires-before-jquery-document-ready-in-firefox) back then which also explained the issue in my case. I think this may be a similar issue. 

However, I think we don't need a comprehensive fix and can just implicitly load the images on `jQuery.ready()` for now.

Kind regards,
@thibsy

PR #6416, opened 2023-10-17
TA/MediaObject: #31994, export image caption to CDATA-section

nhaagen:
https://mantis.ilias.de/view.php?id=31994


alex40724:
Hi @nhaagen. Thanks for the PR, but sorry I have to close it, since this seems to be a T&A only issue, not a general page/media object issue, see my comment in the bug report for details.

PR #6418, opened 2023-10-17
SVY: throw error message if previous button is ...

lukastocker:
...clicked and at least one answer was already given. (#25211)

To go back one page the user has to clear the input first.

Mantis Ticket: https://mantis.ilias.de/view.php?id=25211

### DIFF
--- a/Modules/Survey/templates/default/tpl.il_svy_svy_content.html
+++ b/Modules/Survey/templates/default/tpl.il_svy_svy_content.html
@@ -11,20 +11,24 @@
 <!-- BEGIN top_prev -->
 		<input type="submit" id="prevButton" class="btn btn-default" name="cmd[previous]" value="{BTN_PREV}" />
 <!-- END top_prev -->
+
 	</td>
 	<td class="submit">
-<!-- BEGIN suspend_survey -->
 		<div align="center">
+		<!-- BEGIN top_clear -->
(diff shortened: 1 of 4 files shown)


Modules/Survey/templates/default/tpl.il_svy_svy_content.html
```diff
@@ -11,20 +11,24 @@
...
 <!-- END top_prev -->
+
 	</td>
 	<td class="submit">
-<!-- BEGIN suspend_survey -->
 		<div align="center">
+		<!-- BEGIN top_clear -->
+			<input type="submit" id="prevButtonClear" class="btn btn-default" name="cmd[clearInput]" value="{BTN_CLEAR}" />
+		<!-- END top_clear -->
+		<!-- BEGIN suspend_survey -->
```
alex40724, line 21: We do not want to add this kind of HTML in templates. This should all UI framework stuff.

alex40724 (changes requested):
@lukastocker Thanks for the PR. You may have noticed that this involves some legacy HTML templates. We do not want to add more HTML to these templates, this needs some refactoring. I will put this on the roadmap, but currently this is a "needs funding" bug to me.

alex40724:
I added this to the roadmap, see https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/Modules/Survey/ROADMAP.md

Since this is a minor inconvenience and the bug report does not have any votes I postpone this to ILIAS 10.

PR #6422, opened 2023-10-17
SCORM: #38025, check for empty result when checking (Tracking)Permissions

nhaagen:
https://mantis.ilias.de/view.php?id=38025


Uwe-Kohnle:
I checked it. Thank you Nils

PR #6425, opened 2023-10-17
Delete `Survey` skill data when deleting competences

tfamula:
Related feature: [Improvement of deleting competences](https://docu.ilias.de/goto_docu_wiki_wpage_6588_1357.html)

With the following PR, all `Survey` data, which is related to competences, will be deleted as soon as competences are deleted in the administration. Otherwise, this superfluous data remains unnecessarily in the database, which probably would lead to errors, because old competence assignment would not be found anymore.

### DIFF
--- a/Modules/Survey/Skills/Service/class.SkillInternalManagerService.php
+++ b/Modules/Survey/Skills/Service/class.SkillInternalManagerService.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+/**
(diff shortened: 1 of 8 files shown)


alex40724 (approved):
Thanks! Looks good to me.

tfamula:
Merged manually with https://github.com/ILIAS-eLearning/ILIAS/commit/33f6a913c8c493db03b08b2651a96cde55d90ec8 due to conflicts

PR #6426, opened 2023-10-17
Delete `Test` skill data when deleting competences

tfamula:
Related feature: [Improvement of deleting competences](https://docu.ilias.de/goto_docu_wiki_wpage_6588_1357.html)

With the following PR, all `Test`/`TestQuestionPool` data, which is related to competences, will be deleted as soon as competences are deleted in the administration. Otherwise, this superfluous data remains unnecessarily in the database, which probably would lead to errors, because old competence assignment would not be found anymore.


kergomard (changes requested):
Thank you very much for this improvement @tfamula !

Ok, this review will mostly be about structure and I will keep it short. Please implement the following changes:
- [x] Everything called `Manager` or `Internal` or `Service` must go. These words have no real meaning and are a smell. The best combination is `InternalService`, it should actually be called `SecretService` as, as I understand its uses all over the place, it is a `Service` nobody else should know about. If this is the case, they should never be on the `$DIC`. As far as I can see, you can simply delete all of them and we loose absolutely nothing. Just instantiate the corresponding repositories in the EventListeneners and we are done.
-  [x] Please also remove the nested switch statements [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6426/files#diff-ff81f17390af5f3de80a68392abeeb904d4d34e95201f00a04078bc43a97a7b9R36) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6426/files#diff-ff81f17390af5f3de80a68392abeeb904d4d34e95201f00a04078bc43a97a7b9R36). Why not just use one single `if`?

Thank you very much and best,
@kergomard

tfamula:
Hello @kergomard!

Thanks for your feedback. I am wondering a little bit about your opinion to the structure, because I have found first approaches of a `Service` structure in the `Test` and `TestQuestionPool` modules. Also, an `InternalService` is used there, which can be accessed via `$DIC`.
However, I changed the code according to your wishes. Please have a look if I have understood you correctly.


>  Please also remove the nested switch statements [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6426/files#diff-ff81f17390af5f3de80a68392abeeb904d4d34e95201f00a04078bc43a97a7b9R36) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6426/files#diff-ff81f17390af5f3de80a68392abeeb904d4d34e95201f00a04078bc43a97a7b9R36). Why not just use one single `if`?

I just copied this style from other EventListerners. I guess it can be practical when more than one component and/or event are used in the future. But your suggestion is also fine for me.

kergomard:
Hi  @tfamula 

Yes, I wondered if you would find that, but then decided against a lot of explanations. You are right. We discussed it and will be removing it, there are just a few things we need to do first. So: It is on the way out.

Thank you very much and best,
@kergomard

tfamula:
~~By the way, I have no idea why the tests are failing...~~

I will have a look into the failing tests!

kergomard:
Thank you very much @tfamula !

I do know: Please move `declare(strict_types=1)` below the Copyright (and keep an empty line between the `<?php` and the Copyright) and tests will sail through ;-).

As you need to touch the files anyway, could you please do me a very small favor and move the full namespace to the use-statement [here]($attach_preferences) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6426/files#diff-ff81f17390af5f3de80a68392abeeb904d4d34e95201f00a04078bc43a97a7b9R34). I find that clearer and easier to read.

I will then merge this.

Best,
@kergomard

tfamula:
> I do know: Please move declare(strict_types=1) below the Copyright (and keep an empty line between the <?php and the Copyright) and tests will sail through ;-).

Very interesting! It's the first time I stumble over this. Yesterday, I pushed a bunch of classes into the trunk, which have the wrong order, and the tests passed without any problem.

kergomard:
Thank you very much @tfamula ! I merge now.

PR #6428, opened 2023-10-18
MediaCast: 31722 fix ordering issues at copying object

shecken:
https://mantis.ilias.de/view.php?id=31722

If an item will be deleted there is no reordering of existing items. This causes the mentioned problem in the ticket. There is an old ordering entry for a non existing item. At the copy process the mapping will fail.

I add a reordering after delete of an item and a check the old ordering item id is existing in mapping array.


klees:
Hi @alex40724,

could you look into this sometime soon? As always, a "no" would be fine, just looking for some feedback regarding the status here.

Kind regards!

alex40724:
Thanks!

PR #6430, opened 2023-10-18
9/UI/keyboard focus outline 37635 tweaks

catenglaender:
This is a small addition to https://github.com/ILIAS-eLearning/ILIAS/pull/6373
After the presentation in the CSS squad yesterday, I got some ideas how to improve this.

Clearing the :after element is now optional for the two fallback mixins. Meaning you can safely use il-focus-outline-only, il-focus-border-only for things that already have an after element as it will no longer interfer with tem by default.

However, if you do want to clear the after element for example to avoid the weird visual glitch in the breadcrumbs, you can set $clearAfter to true when including the mixin: `@include il-focus-outline-only($clearAfter: true);`

I also noticed that with a small tweak to the form component, radio and checkbox elements actually do not need to use the fallback solution - at least in Chrome and Firefox.

I will record a new quick video. ;)


Amstutz:
Nice thx a lot. Looking forward to the video.

PR #6431, opened 2023-10-18
Course: Handle autofill after changing course member settings (#29388)

tbongers-cat:
https://mantis.ilias.de/view.php?id=29388

Calls autofill handler after changing max number of members and/or autofill settings.


smeyer-ilias:
Thanks @tbongers-cat

PR #6433, opened 2023-10-18
TA: #38270, declare class properties

nhaagen:
https://mantis.ilias.de/view.php?id=38270


kergomard:

I will merge and then also remove the $lng parameter. As far as I can see it is completely nonsensical, as it will be over in the parent constructor.

PR #6434, opened 2023-10-18
Calendar: Change length of title field to match length of object title (#31368)

tbongers-cat:
https://mantis.ilias.de/view.php?id=31368

Adjusts length of title field in _cal_categories_ to match the length of the title in _object_data_


smeyer-ilias:
Thanks @tbongers-cat

PR #6435, opened 2023-10-18
User: #37887, increase possible length of firstname/lastname/email

nhaagen:
https://mantis.ilias.de/view.php?id=37887

I'm not sure if a length of 128 is excessive?
32 seems to be too small, though.

### DIFF
--- a/Services/User/Setup/ilUserTableUpdateSteps.php
+++ b/Services/User/Setup/ilUserTableUpdateSteps.php
@@ -1,0 +1,48 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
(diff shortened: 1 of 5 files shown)


kergomard:

I actually think this is an ilUser-Issue, so I assign this to me and I also assign the Issue to myself.

I think this is only half of the solution: Whatever arbitrary number we are going to select here somebody will try to enter more. We would need to make sure the field length is limited.

Now. I know that the first- and lastname can be set in `ilObjUserGUI`, but there we already have a defined max_length for the field of 32. So, we would need to change that, if we change the length, but it can't actually be the place the guy in the report fails. Another  place I know of is `ilUserProfile`, but there it seems to be the same thing and maxlength is also set to 32. Are you going to communicate with the reporter in the report or shall I?

@kergomard

nhaagen:
Hi @kergomard ,
thank you hinting the input field sizes!
I reckon, the user fails to update and thus cannot login; the error-message hints to that.
I'll ask the reporter, though.

kergomard:
Hi @nhaagen 

My guess would be that the user is somehow created at login time most probably by the authentication-method. I'm not sure I like shortening the fields on every update, as I feel this might introduce problems later on: We decide to allow even longer names, adapt the corresponding limits on the database and the input-fields and wonder why the names are still shortened (I'm referring to [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/6435/files#diff-e6c81eef5e6261f6eba82ec6de2d7aeecfae82d154fdef8c383b6aa36938a3e6R449-R451)). If in any way possible I would like to be more specific.

Thanks and best,
@kergomard

kergomard:
Hi @nhaagen 

Thanks again for the PR. I reflected on this a little more and think we should go with this solution. I will merge and cherry-pick to trunk.

Best,
@kergomard

PR #6441, opened 2023-10-19
Test: Remove Month From Waiting Time

kergomard:
This PR removes the month field from the setting for waiting between test passes.

This was approved by the JF on august 7th 2023. See: https://docu.ilias.de/goto_docu_wiki_wpage_7987_1357.html#ilPageTocA2527

Current settings that contain a month value will add 31 days per month to the  count.


dsstrassner:
Quick test with 5 Minutes Waiting Time works like a charm. Also updated an old local instance. 31 Days are added to the "days"-field, as promised. (Setting was 1 month before that).

From my point of view, you can merge @kergomard! Thanks for the PR!

kergomard:
Thank you very much for the quick feedback @dsstrassner !

PR #6443, opened 2023-10-19
[FIX] broken file-objects without RID did throw exception

chfsx:
@kergomard On test9 there are a few file-objects that have no RID (and are therefore actually broken anyway). nevertheless I would like to catch the case in the list guis, because otherwise the whole container can no longer be displayed. see small change in ilObjectListGUI


chfsx:
I found another solution for this.

PR #6444, opened 2023-10-20
DateTime and Duration Inputs for Filter

tfamula:
Hello everybody!

This PR adds the [recently revised](https://github.com/ILIAS-eLearning/ILIAS/pull/6276) `DateTime` and `Duration` Inputs for the Filter. 

Although the general implementation seems to be fine, there are some issues I could not solve at the moment:

- Layout in Popover content: The Duration Input is rendered as a "complex" Filter Input in a Popover (see screenshot). This does not look that bad, but it could still use some polish. Maybe some UI experts could help here with the HTML/CSS. (Side note: Please ignore the wrong placement of the Popover. This is a general problem of the Popovers within the Filter – and maybe some other UI components – in the current trunk and should be tackled separately.)
- Method getUpdateOnLoadCode: This method is primarily used for "complex" Filter Inputs to synchronize the changes within the Popover with the Filter Field, which triggers the Popover. Unfortunately, it turned out to be difficult to make this method work for the new Inputs, especially for the Duration Input, which technically consists of two DateTime Inputs, what makes it more difficult. I've run out of ideas and probably will need some help here. But it's important to mention here, that this will not obstruct the functionality of the Inputs within the Filter. The problem is limited to usability and could theoretically be tackled as a separate issue.

@tfamula 

Bildschirmfoto vom 2023-10-20 11-58-02

### DIFF
--- a/src/UI/Component/Input/Field/DateTime.php
+++ b/src/UI/Component/Input/Field/DateTime.php
@@ -20,7 +20,7 @@ declare(strict_types=1);
 
 namespace ILIAS\UI\Component\Input\Field;
 
-use ILIAS\UI\Component\Input\Container\Form\FormInput;
+use ILIAS\UI\Component\Input\Container\Filter\FilterInput;
 use ILIAS\Data\DateFormat\DateFormat;
 use ILIAS\UI\Component\Component;
 use DateTimeImmutable;
@@ -28,7 +28,7 @@ use DateTimeImmutable;
 /**
  * This describes the datetime-field.
  */
-interface DateTime extends FormInput
+interface DateTime extends FilterInput
 {
     /**
      * Get an input like this using the given format.
--- a/src/UI/Component/Input/Field/Duration.php
+++ b/src/UI/Component/Input/Field/Duration.php
@@ -20,14 +20,14 @@ declare(strict_types=1);
 
 namespace ILIAS\UI\Component\Input\Field;
 
-use ILIAS\UI\Component\Input\Container\Form\FormInput;
+use ILIAS\UI\Component\Input\Container\Filter\FilterInput;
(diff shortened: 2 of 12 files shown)


thibsy (changes requested):
Hi @tfamula

I apologise for the delay here, normally we try to get a first response out within about 10 days, but unfortunately we had to prioritise some other things for the release of ILIAS 9.

- [ ] `FilterContextRenderer`: Could you explain why we need `wrapInPopoverContext()` in addition to `wrapInFilterContext()`? The outcome seems to be similar-ish.

- [ ] `FilterInput`: This class is only used to communicate information between rendering steps. However, the composable DTOs should not carry such information, so I suggest to implement a context-renderer for the `DateTime` input and get rid of this class again. If you have questions about this feel free to contact me on Discord.
- [ ] Unit tests: Please add some unit tests, especially for the newly added `tpl.context_filter_popover.html` template (if its necessary, see question above), making sure your changes keep working in the future.
- [ ] SCSS: It's fine if something does not look completely polished. However, please improve the look of the popovers, so they do not look broken. The placement is a different issue, as you point out. See SCSS Guidelines here: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/templates/Guidelines_SCSS-Coding.md
- [ ] Conflicts: Please rebase these changes onto the latest trunk. This will most likely be a bit more complicated than usual due to the late changes of the component revision.

Kind regards,
@Amstutz, @klees and @thibsy

tfamula:
Hi @thibsy 
Thank you for your feedback!
I could not manage the conflicts on the existing branch. That's why I decided to create a new branch and thus a new PR. Please check my answer in https://github.com/ILIAS-eLearning/ILIAS/pull/6646.

PR #6445, opened 2023-10-20
Improvements for DateTime/Duration Input

tfamula:
These are some improvements / fixes / cleanup for the rendering of the recently revised `DateTime` and `Duration` Inputs.

### DIFF
--- a/src/UI/Implementation/Component/Input/Field/Renderer.php
+++ b/src/UI/Implementation/Component/Input/Field/Renderer.php
@@ -637,8 +637,6 @@ class Renderer extends AbstractComponentRenderer
         $tpl = $this->getTemplate("tpl.datetime.html", true, true);
         $this->applyName($component, $tpl);
 
-        $f = $this->getUIFactory();
-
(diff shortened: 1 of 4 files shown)


src/UI/Implementation/Component/Input/Field/Renderer.php
```diff
@@ -810,9 +801,6 @@ protected function renderHiddenField(F\Hidden $input): string
     public function registerResources(ResourceRegistry $registry): void
     {
         parent::registerResources($registry);
-        $registry->register('./node_modules/moment/min/moment-with-locales.min.js');
-        $registry->register('./node_modules/eonasdan-bootstrap-datetimepicker/build/js/bootstrap-datetimepicker.min.js');
```
tfamula, line 814: I am not 100% sure about these two lines. But since we use HTML5 now, we actually do not need Javascript anymore, right? When I tested the Inputs locally without these two lines, everything was still working as intended.

klees:
Thanks @tfamula!

PR #6446, opened 2023-10-21
UI: Fix withValue with DateTimeImmutable

kergomard:
Apparently the html5 datetime-local input expects the given value to be in the format 'Y-m-d H:i'.

See: https://mantis.ilias.de/view.php?id=38265

### DIFF
--- a/Modules/IndividualAssessment/classes/ilIndividualAssessmentUserGrading.php
+++ b/Modules/IndividualAssessment/classes/ilIndividualAssessmentUserGrading.php
@@ -194,7 +194,7 @@ class ilIndividualAssessmentUserGrading
 
         if (!is_null($this->getEventTime())) {
             $event_time = $event_time->withValue(
-                $this->getEventTime()->format($date_format->toString() . ' HH:mm')
+                $this->getEventTime()
             );
         }
 
--- a/src/UI/Implementation/Component/Input/Field/DateTime.php
+++ b/src/UI/Implementation/Component/Input/Field/DateTime.php
@@ -89,12 +89,9 @@ class DateTime extends FormInput implements C\Input\Field\DateTime
         // TODO: It would be a lot nicer if the value would be held as DateTimeImmutable
         // internally, but currently this is just to much. Added to the roadmap.
         if ($value instanceof \DateTimeImmutable) {
-            $value = $this->format->applyTo($value);
+            $value = $value->format(Renderer::HTML5_NATIVE_DATETIME_FORMAT);
         }
-
-        $clone = clone $this;
-        $clone->value = $value;
-        return $clone;
+        return parent::withValue($value);
     }
 
     public function withFormat(DateFormat $format): self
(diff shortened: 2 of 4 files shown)


kergomard:
@thibsy and I had a short exchange. We agreed that the first solution was actually not the right one as it changes the value on  the DTO, but it should actually only be changed in rendering.

Thus I changed the solution. Now the value is only adapted before rendering it.

First for reference: The allowed formats for the value-attribute of datetime-local can be found under https://developer.mozilla.org/en-US/docs/Web/HTML/Date_and_time_formats#local_date_and_time_strings .

Now, I think it would be worth investing a little bit of time to clarify, what we expect on this input:
* Most usages seem to set the value as `string`. But to do so, they mostly create a `string` from a variable they hold as `DateTimeImmutable` (and now, before rendering we change it to a `DateTimeImmutable` to get a string out again).
* Quite a big subset of the usages sets the value to the format preferred by the user, but not all.
* If time is needed it is mostly appended by concatenating a string and there are two different formats used: "H:i" or "HH:mm", but it could also be gotten out of `ILIAS\Data\DateFormat`.

I'm willing to work on this for ILIAS 10, ...but I think it would be worth discussing the right solution first.

Best,
@kergomard

klees:
Hi @kergomard,

from the POV of the inputs in the UI framework it is correct to set the value as string. In that framework, the "value" is the raw data that is displayed to users, while the "content" is the structured data derived from the transformations attached to the single inputs. The flow of data always is `values -> machinery -> structured data`. We do not make a difference if values originate from the server (e.g. as defaults or as previously set values) or the client (via POST).

The fundamental reason for this is, that, via the transformations, we only have a function `value -> data`. It is coincidental that for some data types, such as `DateTimeImmutable` there is a function `data -> value` (via `format`), but we cannot assume this to be the case for arbitrary data. Hence when setting data on inputs, we only accept "raw" values, strings, ints, arrays. One could argue if `DateTimeImmutable` could be included in that list, but the general behaviour ("You need to use 'raw' values when setting up inputs.") won't change.

There is an (internal) method [`Input::isClientSideValueOk`](https://github.com/ILIAS-eLearning/ILIAS/blob/6c4340641218e131cafbbab0f1a159bd241fbf1a/src/UI/Implementation/Component/Input/Field/DateTime.php#L180) that should be implemented to find out if a provided raw value can actually be understood. I.e., to determine if client or user are passing garbage to `withValue`. I think we should adjust it accordingly to throw an exception if the provided string cannot be interpreted as a `DateTimeImmutable`.

If we would want to change this general behaviour, we would require people to create transformations in pairs (`value -> data`, `data -> value`) to do the back and forth. That back and forth cannot be a property of transformations themselves. It would limit the available transformations to these that are information preserving, because if we use `data -> value` and then `value -> data'`, we want to would assume that `data === data'` to keep our minds sane. Thus, we could not up- or downcast in some sense, but only change representations. This would, e.g., mean, that we cannot transform data objects to json, as we loose the information about classes then. So, demanding bijectivity from transformations would make them a lot weaker then they currently are.

We should still strive to move into a world where we do not need to `withValue` in GUI code, as this requires knowledge of internal structure of data objects from the GUI classes. This is what we solve with the [`toFormInput`-pattern](https://github.com/ILIAS-eLearning/ILIAS/blob/6c4340641218e131cafbbab0f1a159bd241fbf1a/Modules/StudyProgramme/classes/class.ilStudyProgrammeTypeSettings.php#L45) at [CaT](https://www.concepts-and-training.de).

Hope this helps to build an understanding about the stuff that is going on in the inputs...

Kind regards!

klees:
Hi everyone,

we found that we could indeed remove `DateTime::withFormat` now, as we cannot control formatting on client side anymore. This seems to be a bigger task, though, so I will present the solution in a separate PR once this is merged.

@thibsy This is fine now, IMO.

Kind regards!

thibsy (approved):
Thx @klees & @kergomard, LGTM now.

PR #6447, opened 2023-10-22
Fix `fixPosition` - don't constantly expand dropdown on click any more

Mr-Pine:
[This default style](https://github.com/ILIAS-eLearning/ILIAS/blob/6c4340641218e131cafbbab0f1a159bd241fbf1a/templates/default/delos.css#L12815) in combination with [this](https://github.com/ILIAS-eLearning/ILIAS/blob/6c4340641218e131cafbbab0f1a159bd241fbf1a/templates/default/delos.css#L2338) places the top right “action” dropDown `15px` from the right side of the screen.

Fix position then checks that the right side is at least 20px from the edge of the screen. Since this is not the case it tries to move the dropdown to the left by decrementing the `left` style property by `5px`. But since the element still has the `right: 0` property this only increases the width of the dropdown, not the right position, leaving the “problem” unresolved.

This causes every click on the “action” button to increase the width of the dropdown

This PR fixes this behaviour by dropping the threshold in `fixPosition` to `15px`, the value that's used in the stylesheets.

### DIFF
--- a/Services/JavaScript/js/Basic.js
+++ b/Services/JavaScript/js/Basic.js
@@ -230,8 +230,8 @@ il.Util = {
 			return;
 		}
 
-		if (vp.right - 20 < r.right) {
-			il.Util.setX(el, r.x - (r.right - vp.right + 20));
+		if (vp.right - 15 < r.right) {
+			il.Util.setX(el, r.x - (r.right - vp.right + 15));
 		}
 
 		r = il.Util.getRegion(el);


klees:
Hi @Mr-Pine,

thanks for contributing!

FYI: I have labelled this as `improvement` because for the `bugfix` label we would [require a ticket from mantis to be linked](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/docs/development/contributing.md#rules-for-contributors).

Still sounds plausible, though =)

Kind regards!

klees:
@Amstutz I assigned you because this seems to be about css more then about that old `Basic.js`-class. Hope that is ok. Maybe @catenglaender could have a look into this?

Amstutz:
Hi @Mr-Pine 

Thx a lot! LGTM from my part. As soon as @catenglaender greenlights this, ill merge. @catenglaender , I assigned you ftm. Feel free to pass back to me as soon as reviewed.

Thx.

mjansenDatabay:
Hi @Mr-Pine , hi @catenglaender 

this PR has been lying around for quite a while.

@Mr-Pine: Please resolve the conflicts
@catenglaender: If the conflicts have been solved and you don't have any objections, please merge this PR and pick it to the other branches.

Thanks and best,
@mjansenDatabay

catenglaender:
I think for 8 this is okay.
For 9, I wonder if exposing the 15 px as a css variable --il-page-left-padding or something like this on the :root element. Javascript can fetch these with `getComputedStyle(document.body).getPropertyValue('--il-page-left-padding')`. This way we would know where the 15 px come from and why they are there. What do you think?

Mr-Pine:
Well, while trying to fix the conflict I noticed that the conflict was a fix for exactly this problem: https://github.com/ILIAS-eLearning/ILIAS/commit/e5ba4c3b0ac64ab768dc41adf49fd2c7ac0ff1be :D So I guess this pr can be closed

Mr-Pine:
But yes @catenglaender this does sound like a better idea than hardcoding values separately

PR #6448, opened 2023-10-23
UICore: Remove Lightbox Related Functions

kergomard:
Hi @thibsy 

While implementing the time limits in Object, I found these remnants of the legacy lightbox. We discussed in the UILegacy-Meeting that the lightboxes are gone and I thoroughly searched the code, I don't see any usages anywhere. So I propose to remove them.

@mjansenDatabay : I also removed it from the RTE-Template.
@pascalseeland and @Amstutz : And also from the DataCollecition.

Best,
@kergomard


mjansenDatabay (approved):
LGTM @kergomard

thibsy (approved):
Thx @kergomard, LGTM.

PR #6450, opened 2023-10-23
UI Components, Select Input: Display null if value set and required

Amstutz:
Some small change, which proposes to not display the null value if a value has been set and the selction is set to required. See: https://mantis.ilias.de/view.php?id=38310.

Note that the Unit Test for this is somewhat extensive. This is due to the fact, that we had no test for rendering required for Select Inputs. So this is included as well. Otherwise I would have only tested the abscance of the null value in the rendering.


thibsy (approved):
Thx a lot @Amstutz, LGTM. Appreciate the extensive unit test =).

thibsy:
Note, the failing unit tests are not related to this PR.

PR #6451, opened 2023-10-23
ILIAS 9 Services/LegalDocuments. Services/TermsOfService. Services/DataProtection

lscharmer:
This PR implements the following FR: https://docu.ilias.de/goto_docu_wiki_wpage_7451_1357.html.

Because the new DataProtection Service is very similar to the TermsOfService, a new Service, LegalDocuments, is added which provides the basis for implementing the two services and serves as an entry point for other services.

A lot of code which belongs to the terms of services where scattered across other services, this code is moved to the LegalDocuments service. 
The places where another service must call the ToS, is replaced with a call to the LegalDocuments. 
The LegalDocuments service delegates these calls to the ToS and the new DataProtection service and therefore hides the concrete implementation of the actual ToS (and others).
If there will be Services in the future which have similar semantics to the ToS or DataProtection they can be implemented with the LegalDocuments too, without the need of adjusting other services.

The following changes are made outside Services/TermsOfService (and Services/DataProtection, Services/LegalDocuments) and need approval from other maintainers, most changes are replacements from ToS to LegalDocument or code deletion because it is moved to the LegalDocuments service:

@kergomard:
## `Services/User`
In File `Services/User/classes/Profile/class.ilPersonalProfileGUI.php`:
- Withdrawal process
- Show Agreement

In file `Services/User/classes/class.ilObjUser.php`:
- Method `hasToAcceptTermsOfService`
- Method `hasToAcceptTermsOfServiceInSession`
- Method `getUsersAgreed`
- In `_getUsersOnline` the call to ToS is replaced with a call to LegalDocuments

In file `Services/User/classes/class.ilObjUserFolderGUI.php`. ToS as well as DataProtection is displayed as an option for user deletion (as described in the FR).

File `Services/User/classes/class.ilObjUserGUI.php`.
`Services/User/classes/class.ilUserAppEventListener.php`.

@alex40724:
## `Services/Administration`
File `Administration/GlobalScreen/classes/AdministrationMainBarProvider.php`.

## `Services/Awareness`
File `Services/Awareness/User/class.Collector.php`.

@pascalseeland / @mjansenDatabay:
## `Services/Authentication`
File `Services/Authentication/classes/Frontend/class.ilAuthFrontend.php`.

## `Services/Init`
File `Services/Init/classes/StartupSequence/StartUpSequenceDispatcher.php`.

File `Services/Init/classes/class.ilInitialisation.php`.

In File `Services/Init/classes/class.ilStartUpGUI.php`:
- Display for the Terms Of Service is moved to LegalDocuments
- Withdrawal process is moved to LegalDocument

## `Services/Registration`
File `Services/Registration/classes/class.ilAccountRegistrationGUI.php`

@jeph864:
## SOAP / Webservices 
File `webservice/soap/classes/class.ilSoapAdministration.php`.

@thibsy 
## `Services/UICore`
File `Services/UICore/classes/Screen/PageContentProvider.php` (Footer).

@mjansenDatabay:
## Services/Mail
File `Services/Mail/classes/class.ilMail.php` and corresponding UnitTests.
File `Services/Mail/classes/Recipient.php` and corresponding UnitTests.


Services/UICore/classes/Screen/PageContentProvider.php
```diff
@@ -229,8 +229,8 @@ public function getFooterModification(CalledContexts $screen_context_stack): ?Fo
 
             $footer = $f->mainControls()->footer($links, $text);
 
-            $tosWithdrawalGui = new \ilTermsOfServiceWithdrawalGUIHelper($this->dic->user());
-            $footer = $tosWithdrawalGui->modifyFooter($footer);
+            global $DIC;
+            $footer = $DIC['legalDocuments']->modifyFooter($footer);
```
thibsy: Hi @lscharmer,

I think we can get rid of this snippet if your new service implements its own `ModificationProvider`, or am I missing something? If so, I would appreciate it very much to get rid of this modification here =).

Kind regards,
@thibsy

lscharmer: Ah yes, thank you @thibsy, I will change this.

thibsy: Thx a lot!

lscharmer: Hmm, I looked into this but it seems that one is only able to replace the **whole** footer and I don't get the previous footer passed to `getFooterModifications`.

thibsy: @lscharmer you're right, it isn't possible to provide multiple modifications as I thought. The one with highest priority wins and replaces the lower priority modifications. But @chfsx notified me about an upcoming revision of the footer (see https://docu.ilias.de/goto_docu_wiki_wpage_7883_1357.html), which will introduce an easier mechanism to provide footer-parts. So for the time being this change LGTM, but please change `$DIC` to `$this->dic`, since the container is already available there.

lscharmer: `$DIC` is replaced with `$this->dic`. Ah ok, when the footer is updated I will remove the code in another PR.

Services/LegalDocuments/classes/Setup/Agent.php
```diff
@@ -0,0 +1,89 @@
...
+    public function getInstallObjective(Config $config = null): Objective
+    {
+        return new NullObjective();
+    }
+
+    public function getUpdateObjective(Config $config = null): Objective
+    {
+        return $this->list([
+            new ilDatabaseUpdateStepsExecutedObjective(new UpdateSteps()),
+            new ConsumerObjective(),
```
thibsy: A minor addition: you should return the `ConsumerObjective` in `getBuildArtifactObjective()` instead and leave it up to this service to decide at what point the artifact should be built. Otherwise this artifact will also be ignored when the `build-artifacts` command is run.

lscharmer: @thibsy: Done.

thibsy (approved):

kergomard (commented):
Hi @lscharmer 

Thank you very much for the improvements here!

This looks good as far as I can see, I only have two nit-picks:
* Could you please remove the explicit `$DIC` in the middle of the function [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6451/files#diff-e6c81eef5e6261f6eba82ec6de2d7aeecfae82d154fdef8c383b6aa36938a3e6R3701). The `$DIC` is already there and if I decide to remove it, I would like things to brake, so that I can react.
* Could we get rid of the two double casts [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6451/files#diff-ca68ff2332ed52e681485f1c58b7a71a16b67d73c3c09c85900720267d35ccdaR1943) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6451/files#diff-ca68ff2332ed52e681485f1c58b7a71a16b67d73c3c09c85900720267d35ccdaR1947)? We want to have a numeric string '0' or '1' at the end and the Input is null or '1', right? So we should be explicit and it would be something like `$this->form->getInput('dpro_withdrawal_usr_deletion') === '1' ? '1' : '0'.

If you are able to apply the two changes, you can just merge from my side. No need to ask me again.

Thank again and best,
@kergomard

lscharmer:
> Hi @lscharmer
> 
> Thank you very much for the improvements here!
> 
> This looks good as far as I can see, I only have two nit-picks:
> 
>     * Could you please remove the explicit `$DIC` in the middle of the function [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6451/files#diff-e6c81eef5e6261f6eba82ec6de2d7aeecfae82d154fdef8c383b6aa36938a3e6R3701). The `$DIC` is already there and if I decide to remove it, I would like things to brake, so that I can react.
> 
>     * Could we get rid of the two double casts [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6451/files#diff-ca68ff2332ed52e681485f1c58b7a71a16b67d73c3c09c85900720267d35ccdaR1943) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6451/files#diff-ca68ff2332ed52e681485f1c58b7a71a16b67d73c3c09c85900720267d35ccdaR1947)? We want to have a numeric string '0' or '1' at the end and the Input is null or '1', right? So we should be explicit and it would be something like `$this->form->getInput('dpro_withdrawal_usr_deletion') === '1' ? '1' : '0'.
> 
> 
> If you are able to apply the two changes, you can just merge from my side. No need to ask me again.
> 
> Thank again and best, @kergomard

Is both adjusted.

mjansenDatabay (approved):

mjansenDatabay:
If there are no objections of the remaining reviewers, I will merge this today, because the changes in the corresponding components (`Services/Administration`, `Services/Authentication` and `SOAP / Webservices`) are trivial.

PR #6454, opened 2023-10-24
9/UI/fix submit button on wrong side

kergomard:
Hi @Amstutz , @klees, and @thibsy 

This is a PR changing the submit button in the Roundtrip Modal from the right to the left of the button to cancel. I think this is in accordance with the Usage Rules.

See: https://mantis.ilias.de/view.php?id=38322

Thanks and best,
@kergomard


thibsy (approved):
Hi @kergomard, thx a lot! LGTM.

PR #6455, opened 2023-10-24
LDAP: Fix accessing undefined array key "member"

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=38049#c96875

If approved, this must be cherry-picked to `trunk` (and maybe `release_9`) as well.

### DIFF
--- a/Services/LDAP/classes/class.ilLDAPQuery.php
+++ b/Services/LDAP/classes/class.ilLDAPQuery.php
@@ -429,13 +429,13 @@ class ilLDAPQuery
 
         // All groups
         foreach ($group_data as $data) {
-            if (is_array($data[$attribute_name])) {
+            if (isset($data[$attribute_name]) && is_array($data[$attribute_name])) {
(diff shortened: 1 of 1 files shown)


pascalseeland:
@mjansenDatabay please rebase onto updated release8. That should get rid of the copyright issue

mjansenDatabay:
@pascalseeland Please have a look at https://mantis.ilias.de/view.php?id=38363#c97055 . It seems this fix is not necessary.

Dennis1993:
The problem seems to arise from something else. The real problem is that with PHP8 the LDAP limit is handled differently. Unfortunately I don't know how to fix this.

Theoretically, importing more than <LDAP limit> should not be possible with anyone using PHP8.

More information in the link from mjansenDatabay. I hope it can be fixed.

Thanks!

mjansenDatabay:
IMO we don't need this PR anymore if https://github.com/ILIAS-eLearning/ILIAS/pull/6909 gets merged.

PR #6458, opened 2023-10-24
Proposal for new Footer-Interface

chfsx:
This is the interface change as a suggestion for the rebuilding of the footer according to [Revamping Footer with Link-Groups](https://docu.ilias.de/goto_docu_wiki_wpage_7883_1357.html).

The new footer is divided into three areas: Sharing-Links, Utility-Link-Groups and Meta-Infos:

image

Rendered, the new footer looks like this: 

Desktop:
image

Mobile:
image

The interface loses all getter methods, as these do not have to be part of the interface. these are for internal use in the renderer and are only done on the implementation.

### DIFF
--- a/src/UI/Component/MainControls/Factory.php
+++ b/src/UI/Component/MainControls/Factory.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 namespace ILIAS\UI\Component\MainControls;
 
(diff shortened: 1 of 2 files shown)


klees (changes requested):
Hi @chfsx,

thanks for looking into this nice improvement that short before the deadline for ILIAS 9. I really like it and I'll try to do my best to make this happen in time.

I formatted and rectified your comments a little, hope this is ok. Changes are [here](https://github.com/klees/ILIAS/tree/proposal/9/footer-interface), would have pushed directly to your branch but wasn't allowed to. Two things I want to point out specifically: [@description](https://jsdoc.app/tags-description.html) is not needed when the doc-block starts with the description, so I removed it. And: Please remind yourself to [line wrap at ~80 chars](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/docs/development/docs-guidelines.md#content). 

- [ ] Sharing Links: I understand that we have one sharing link for each page. Could you provide some examples for situations where there are more of these links?
- [ ] Sanitized Plain Text: What do you mean with `sanitized plain text` in the description of `withMetaInfo`?

- [ ] Remove Modal: Frankly, I'm not convinced that we would want to trigger modals from the footer. I'm sure you have some use case (you would not suggest it otherwise, I guess) but I can imagine it to require some discussion. Also, I'm not convinced that the way to add these modals (via `withModalsToRender`) will provide a good developer experience. We should discuss this as well. So I would propose to remove the option for modals from the PR for now to make the general improvement happen for ILIAS 9 and iterate on it in 10.


- [ ] docu: Please include some informative sentence about the three sections in the KS-entry, so non-programmers can understand what these sections are about.

Best regards!

klees:
Hi @chfsx,

I will close this for the moment to get if of my agenda. Feel free to reopen, though, or ping me for further discussion.

Kind regards!

PR #6460, opened 2023-10-24
ks: main action for items

alex40724:
Follow up of https://github.com/ILIAS-eLearning/ILIAS/pull/6288

This adds a main action to KS items to provide a button for the most important action of an item.

This is needed to implement the feature https://docu.ilias.de/goto_docu_wiki_wpage_7444_1357.html

### DIFF
--- a/src/UI/Component/Item/Factory.php
+++ b/src/UI/Component/Item/Factory.php
@@ -34,8 +34,9 @@ interface Factory
      *       This is a standard item to be used in lists or similar contexts.
      *   composition: >
      *       A list item consists of a title and the following optional elements:
-     *       description, action drop down, audio player element, properties (name/value),
-     *       a text, image, icon or avatar lead, a progress meter chart and a color.
+     *       description, main action button or main action link, action drop down,
+     *       audio player element, properties (name/value), a text, image, icon or
+     *       avatar lead, a progress meter chart and a color.
      *       Property values MAY be interactive by using a Shy Buttons or a Link.
      * rules:
      *    accessibility:
--- a/src/UI/Component/Item/Factory.php
+++ b/src/UI/Component/Item/Factory.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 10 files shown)


alex40724:
@thibsy Tests are using mocks now instead. Link/Button HTML is replaced by Canonical.

thibsy (commented):
Hi @alex40724

Thx a lot for implementing our suggestions! There is one minor thing left to do: 

- [ ] `ItemTest`: the renderer which is supplied with the mocked instance using a hash-code as canonical name will render only this hash instead, which allows us to perform the assertion like `assertTrue(str_contains($html, $hash_code));` instead of comparing the HTML structure with `assertHTMLEquals()`. This will allow us to modify said structure more easily in the future without adjusting too many unit-tests. Could you replace this check?
- [x] Copyright: I also noticed the checks currently fail because of malformed copyright license headers, please adjust them as well.

Afterwards this is ready to be merged now.

Kind regards,
@thibsy

alex40724:
> Hi @alex40724
>     * [ ]  `ItemTest`: the renderer which is supplied with the mocked instance using a hash-code as canonical name will render only this hash instead, which allows us to perform the assertion like `assertTrue(str_contains($html, $hash_code));` instead of comparing the HTML structure with `assertHTMLEquals()`. This will allow us to modify said structure more easily in the future without adjusting too many unit-tests. Could you replace this check?

@thibsy  Thanks for the feedback, but I did that, as far as I understand the approach. Maybe you could be more specific.

The remaining assertHTMLEquals will test for the HTML of the item (which we want to test here, since this is an Item test) and it replaces the button/link HTML with the hash-code of the Button/Link. So it will check its position in the HTML, but it will not be influenced by HTML changes in buttons or links. So changes in Button or Link renderers will not affect this test, e.g.
```
<div class="il-item il-std-item ">
            <div class="il-item-title">Title</div>
            <div class="il-item-actions l-bar__container">
                <div class="l-bar__element">$expected_button_html
            </div>
            </div>
        </div>
```
All the HTML code is item specific, so it should be tested, `$expected_button_html` is the button hash (`$expected_button_html = md5(I\Component\Button\Standard::class);`). So the position is tested (and should) but not the button HTML. `str_contains` would only check for existence, but not for the correct position. Still changes in Button or Link renderers will not affect this test.

thibsy (approved):
Hi @alex40724

Thanks for your explanation. I can see your reasoning and will merge this PR now, since this was also merely a suggestion and using a hash-code is already a nice improvement. In the future we can ask ourselves, if a rendering test can be sufficiently covered by just a `str_contains()` check, to reduce redundancy if e.g. elements are always rendered in the same place. We must be careful not be too minimalistic though.

Kind regards
@thibsy

PR #6463, opened 2023-10-25
FR, Data Table for QuestionPool

nhaagen:
https://docu.ilias.de/goto_docu_wiki_wpage_7878_1357.html


kergomard (commented):
Thank you very much  @nhaagen !
I went quickly through this and this looks good. I did no functional testing.

Here are just a few small things:
* [This here](https://github.com/ILIAS-eLearning/ILIAS/pull/6463/files#diff-79341fc60c03ba53809fb097ae04d5d5a3e597a46a47406f01dd0fed78e926c2R97) does not work in PHP 8.2 as I don't see `$df` being declared anywhere. As you change this, could you please rename `$df` to `$data_factory` and then move the Namespace to a `use` statement. I would suggest to alias it to `DataFactory`.
* [This here](https://github.com/ILIAS-eLearning/ILIAS/pull/6463/files#diff-79341fc60c03ba53809fb097ae04d5d5a3e597a46a47406f01dd0fed78e926c2R98) should be unnecessary, right? Request is set in `ilObjectGUI`.
* Versions of the same apply to all class method assignments on the next few lines.

The rest actually looks good, as far as I can see. I assume @dsstrassner , you did or will do some testing, right?

Thanks again and best,
@kergomard

dsstrassner:
When I try to copy all questions in the pool to a new pool, I got: 

Error thrown with message "Class "GUI" not found"

Stacktrace:
#7 Error in /var/www/html/Modules/TestQuestionPool/classes/class.ilObjQuestionPool.php:268
#6 ilObjQuestionPool:createQuestion in /var/www/html/Modules/TestQuestionPool/classes/class.ilObjQuestionPool.php:290
#5 ilObjQuestionPool:copyQuestion in /var/www/html/Modules/TestQuestionPool/classes/class.ilObjQuestionPool.php:817
#4 ilObjQuestionPool:pasteFromClipboard in /var/www/html/Modules/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php:1284
#3 ilObjQuestionPoolGUI:pasteObject in /var/www/html/Modules/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php:534
#2 ilObjQuestionPoolGUI:executeCommand in /var/www/html/Services/UICore/classes/class.ilCtrl.php:118
#1 ilCtrl:forwardCommand in /var/www/html/Services/UICore/classes/class.ilCtrl.php:91
#0 ilCtrl:callBaseClass in /var/www/html/ilias.php:24


Not sure, if related to this PR.

---
Cells not closed:

image

---

Sorting does not work for Taxonomies and the Label is somehow outside the symmetry:

image

---

I miss the filter option for taxonomies:

old: 

image

new:

image

We need then a "No Taxonomy" - Option in this filter (see FR)

---
The Apply and Reset "Button" "protrudes" into the Filter Element above. Not sure, if we need an update of the Filter UI Component ;)

image

---

Could you please move  "Created and Updated" at the right end of the table?

---
The "Only Commented Questions"  Filter does not work.
---



- [ ] Copying
- [ ] All cells closed
- [x] Sorting Taxonomies in Table
- [x] Taxonomies missing in Filter
- [ ] Apply-Button
- [ ] Order
- [ ] Comment-Filter

nhaagen:
@kergomard you were, of course, right. Thank you. The class-properties are declared, now.

kergomard:
Hi @nhaagen ! Thank you very much!

Ok, I'm going to merge this: There is still one missing class property, but I'm going to quickly add it myself.

Thanks again and best,
@kergomard

dsstrassner:
@nhaagen sorry to bother you, but it was important for Uni Stuttgart to get this "No taxonomy"-Option in the Filter. Everything else is possible to repair after coding complete, but we need this option ...

nhaagen:
@dsstrassner you're not bothering at all, thanks for bearing. It's just putting all those pieces together and amending for missing ones. My bad, I'll add it, of course

dsstrassner:
@kergomard From my point of view, are the functional changes are ready to merge, and the small issues could be resolved after merging.
This would be: 

- [ ] Column setting is kept when using the filter
- [ ] Comment-Filter need to work

kergomard:
Thank you @nhaagen !

Ok, this looks good and I merge now.

There are two more functional issues, that need to be addressed, but seem to be a problem in the table:
- [ ] The table initially only shows the column with the question title and this can also not be changed persistently.
- [ ] The table only shows five rows and this can also not be changed persistently.

We will need to fix that in a separate step.

Best,
@kergomard

PR #6464, opened 2023-10-25
User: Fix `Mustache` migration for welcome mail

mjansenDatabay:
This PR applies some fixes for the `Mustache` syntax migration of welcome mail subjects/bodies.

* `()` were missing in `getWhere`, so the former code was logically wrong.
* `ilDbInterface::setLimit` is used instead of `LIMIT 1` in raw SQL
* The SQL `REGEXP_REPLACE` statements did not work at all. I replaced the code by PHP `preg_replace` calls.


shecken (approved):

shecken (approved):

kergomard:
I will merge this as it is and take care in a next commit that we avoid writing empty strings into cells containing null.

Best,
@kergomard

PR #6465, opened 2023-10-25
UI: datetime_required LangVar in Wrong Module

kergomard:
The language variable datatime_required is in the module ui, but if I'm understanding the structure right it should be in validation.
See: https://mantis.ilias.de/view.php?id=38341


kergomard:
Ok, as this is now assigned to you @matthiaskunkel . I just merge this.

PR #6466, opened 2023-10-25
#38346 - fix saml metadata.php

sroenchen:
Replace array_key_exists() with isset()


mjansenDatabay:
The problem is: This will not be sufficient to make SAML work in ILIAS 8. `simplesamlphp` also uses a lot of array functions, so there is currently no alternative to the INI approach I mentioned in https://mantis.ilias.de/view.php?id=38346 .

sroenchen:
ini-approach seems to be working. Closing the PR

PR #6475, opened 2023-10-25
PRG: add Page Editor Element for Required Actions

lukastocker:
Also add SCSS for the Action Note and Statusinformation element.

Adjust Statusinformation element template and fix errors, as it crashed while trying to create.

The Design might have to be adjusted at a later time.

This feature is mentioned as No. 1 in the [Feature Wiki](https://docu.ilias.de/ilias.php?baseClass=ilwikihandlergui&cmdNode=16m:r6&cmdClass=ilobjwikigui&cmd=viewPage&ref_id=1357&page=Study_Programme%3A_Page_Editor_Element_for_Required_Actions)


Services/COPage/xsl/page.xsl
```diff
@@ -4376,6 +4376,24 @@
...
+		</div>
+	</xsl:if>
+	<xsl:if test="$mode != 'edit'">
+		<div class="il-prg-actionnote-container">
+			<Launcher>
+				[[[PRG_ACTION_NOTE]]]
+			</Launcher>
+		</div>
+	</xsl:if>
+</xsl:template>
```
alex40724, line 4386: @lukastocker The text "Study Programme Action Note" will not be translated here. Since you are already using the `modifyPageContentPostXsl()` approach you could reduce the XSL part to:
```
<xsl:template match="PRGActionNote">
    [[[PRG_ACTION_NOTE]]]
</xsl:template>
```

And handle the different edit/presentation rendering completely in `modifyPageContentPostXsl()` and use $DIC->language().

alex40724 (changes requested):
Hi @lukastocker 

thanks for implementing this feature. There is one note I have regarding the changes in COPage:

- [ ] Please further reduce the code in page.xsl and fix the translation issue as suggested in the comment.

kindly,
@alex40724

klees:
@alex40724 Thanks a bunch for looking into this and for the valuable feedback as well!

lukastocker:
> Hi @lukastocker
> 
> thanks for implementing this feature. There is one note I have regarding the changes in COPage:
> 
> Please implement the following changes:
> 
>     * [ ]  Please further reduce the code in page.xsl and fix the translation issue as suggested in the comment.
> 
> 
> kindly, @alex40724

Hi @alex40724 ,

thank you for your feedback!
I implemented the changes for the Action Note and also Statusinformation Element as they are both very similiar and I already adjusted the tpl and scss for the Statusinformation element in the first commit too.

The Element will be translated now too.

Let me know what you think of the changes.

Kind regards,
Luka

klees:
@alex40724 The changes in PageEditor look LGTM (according to your feedback...) now. I would assume that this is ok now for you too if I don't hear otherwise.

alex40724 (approved):

PR #6459, opened 2023-10-24
UI, DataTable: add a LinkColumn

nhaagen:
Screenshot from 2023-10-24 17-30-20

PR #6482, opened 2023-10-26
UI DataTable: Add LinkList Column

tfamula:
Bildschirmfoto vom 2023-10-26 17-32-16

### DIFF
--- a/src/UI/Component/Table/Column/Factory.php
+++ b/src/UI/Component/Table/Column/Factory.php
@@ -114,10 +114,21 @@ interface Factory
      * ---
      * description:
      *   purpose: >
-     *      The Link Column features a Link.
+     *      The Link Column features a Standard Link.
      *
      * ---
      * @return \ILIAS\UI\Component\Table\Column\Link
      */
     public function link(string $title): Link;
+
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *      The Link Column features a Listing of Standard Links.
+     *
+     * ---
+     * @return \ILIAS\UI\Component\Table\Column\LinkListing
+     */
+    public function linkListing(string $title): LinkListing;
 }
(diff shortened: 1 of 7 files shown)


tfamula:
It would be great if we could fasttrack this before Coding Completed tomorrow, like https://github.com/ILIAS-eLearning/ILIAS/pull/6459.

src/UI/examples/Table/Column/Link/base.php
```diff
@@ -55,6 +55,7 @@ public function getTotalRowCount(
         }
     };
 
-    $table = $f->table()->data('Link Columns', $columns, $data_retrieval);
+    $table = $f->table()->data('Link Columns', $columns, $data_retrieval)
+        ->withRequest($DIC->http()->request());
```
tfamula, line 59: This PR also includes a little fix for the the example of the Link Column

thibsy (changes requested):
Hi @tfamula

It is possible to fast-track this as well, like we did with #6459. So if you can provide the requested changes today, I will gladly merge this afterwards.

- [x] `Column\LinkList::format()`: this method currently compares the given `$value` against the `Listing\Listing` interface. IMO this is too broad and would allow any kind of listing in the future (even though they might fail `Link\Standard` check like `Listing\Property` currently does). I believe we should stick with only `Listing\Ordered` and `Listing\Unordered` for the moment.

- [x] `LinkList`: please rename this interface to `LinkListing`, indicating its association with `Listing\*` components. Also rename the factory method accordingly.
- [x] `ColumnTest`: please also add a unit-test for invalid values passed into `LinkList::format()`. One where another component is passed, and, one where a listing with non-link values is passed.

Kind regards,
@Amstutz, @klees and @thibsy

tfamula:
Hi @thibsy 
Thanks for the quick feedback! All done!

thibsy (approved):
Thx @tfamula, this LGTM now. I added the JF label so we can announce this as well.

matthiaskunkel:

PR #6485, opened 2023-10-27
T&A Bugix #0038178: Random Question rules that cannot be implemented generate error when starting test.

fhelfer:
https://mantis.ilias.de/view.php?id=38178
trunk as well

### DIFF
--- a/Modules/Test/classes/class.ilTestRandomQuestionsQuantitiesDistribution.php
+++ b/Modules/Test/classes/class.ilTestRandomQuestionsQuantitiesDistribution.php
@@ -140,14 +140,9 @@ class ilTestRandomQuestionsQuantitiesDistribution
         $this->questRelatedSrcPoolDefRegister = array();
     }
 
-    /**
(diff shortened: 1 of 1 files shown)


dsstrassner:
@kergomard - maybe HJL wants this in his pretesting ;)

kergomard:
Shouldn't we just remove the  `is_numeric()` check completely? The only reason for it to not be numeric, should be that it is null, but then the check does not help us at all. What am I missing?

Best,
@kergomard

fhelfer:
> Shouldn't we just remove the `is_numeric()` check completely? The only reason for it to not be numeric, should be that it is null, but then the check does not help us at all. What am I missing?
> 

Thx @kergomard , you are right. We can completely remove the numeric check

Best, @fhelfer

kergomard:
Thank you very much @fhelfer ! I think this should work. I will merge and cherry-pick to ILIAS 9 and trunk

PR #6502, opened 2023-10-30
Services/Test: Add wac check for question file upload preview images (ILIAS 9)

lscharmer:
Sister PR to: #6500.

### DIFF
--- a/Modules/Test/classes/AccessFileUploadAnswer.php
+++ b/Modules/Test/classes/AccessFileUploadAnswer.php
@@ -23,12 +23,13 @@ namespace ILIAS\Modules\Test;
 use ILIAS\Data\Result;
 use ILIAS\Data\Result\Ok;
 use ILIAS\Data\Result\Error;
-use ILIAS\DI\Container;
 use ilObjTest;
 use ilObject;
 use ilSession;
 use ilTestSession;
 use ilTestAccess;
+use ilDBInterface;
+use ilObjUser;
 use Closure;
 
 class AccessFileUploadAnswer implements SimpleAccess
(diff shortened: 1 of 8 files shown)


kergomard:
Thank you very much @lscharmer !

I leave the review to @mbecker-databay , just one thing:
Please remove the $this->dic in `ilObjTestAccess` completely and move all the creation of Objects in `canBeDelivered()` to the constructor. 

For me also [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/6502/files#diff-280ab40c70a54a6d18164beb627ad5dc75ed79423cd19c2484db4e37b2043625L64-R83) looks way over complicated for what you want to achieve. I would kindly ask you to keep it as simple as it actually is: If you already created the Access-classes in the constructor, as far as I can see, this is one conditional statement with three conditions. What am I missing?

Thanks again and best,
@kergomard

lscharmer:
> Thank you very much @lscharmer !
> 
> I leave the review to @mbecker-databay , just one thing: Please remove the $this->dic in `ilObjTestAccess` completely and move all the creation of Objects in `canBeDelivered()` to the constructor.
> 
> For me also [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/6502/files#diff-280ab40c70a54a6d18164beb627ad5dc75ed79423cd19c2484db4e37b2043625L64-R83) looks way over complicated for what you want to achieve. I would kindly ask you to keep it as simple as it actually is: If you already created the Access-classes in the constructor, as far as I can see, this is one conditional statement with three conditions. What am I missing?
> 
> Thanks again and best, @kergomard

Hi @kergomard,
I will remove the `DIC`.
Concerning the second point: Yes, currently this could be seen as an if with 3 conditions but, as this PR shows, this will change. The point of the `findMatch` method is that it receives a list of simple access classes and checks them one after the other. As this PR shows, this list will be extended in the future, because the simple access classes (should) cover only one URL. So in my point of view this is not an `if` with three cases but a list with paths to go through.
But I will change the `array_reduce` to a `foreach` for clarity.

Best regards
@lscharmer

kergomard:
Thank you very much @lscharmer ! This is already fine by me.

PR #6504, opened 2023-10-30
Use Skill Usage Service in `TestQuestionPool` classes

tfamula:
Hi @kergomard,
after some refactoring, I would like to get rid of the static function `ilSkillUsage::setUsage` (and the class `ilSkillUsage` as a whole). That's why I now want to use the new `SkillUsageService` instead.

Because this change is quite trivial, I would also like to cherry-pick it to release 9.

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assQuestion.php
+++ b/Modules/TestQuestionPool/classes/class.assQuestion.php
@@ -20,6 +20,7 @@ use ILIAS\Refinery\Transformation;
 use ILIAS\TA\Questions\assQuestionSuggestedSolution;
 use ILIAS\TA\Questions\assQuestionSuggestedSolutionsDatabaseRepository;
 use ILIAS\DI\Container;
+use ILIAS\Skill\Service\SkillUsageService;
 
 require_once './Modules/Test/classes/inc.AssessmentConstants.php';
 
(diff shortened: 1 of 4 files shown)


kergomard:
Thank you very much @tfamula !

Could you please fix the tests.

Other than that: This looks good and I'm grateful for every static function that leaves us. I will merge and cherry-pick to ILIAS 9.

Best,
@kergomard

tfamula:
Hi @kergomard 
I really don't understand why the tests are failing. In my eyes, the errors are not caused by the changes of this PR (but they are triggered by this PR because I touched the classes now)

kergomard:
Hi @tfamula 

Yes they are: In the `SkillProfileManager` on line 50 you try to get `rbac_review` out of the `$DIC`, but the `$DIC` here does not contain any `rbac_review` and things go sideways from there.

Best,
@kergomard

tfamula:
Thanks for the hint, @kergomard! I was blind :) 
Seems like I fixed them.

kergomard:
Thanks @tfamula ! We all tend to be blind, when it comes to our own code. I merge and cherry-pick now.

tfamula:
@kergomard I have seen that you cherry-picked the second commit of this PR (the one with the unit test fix) to Release 9 (https://github.com/ILIAS-eLearning/ILIAS/commit/42b04bd851faeecccd65c90d2a5e457003392d78), but not the first commit (the one which replaces the static function). Could you please also cherry-pick the first commit? Thanks :)

kergomard:
Thank you very much for informing me! Should have squashed before merge, sorry. Done.

PR #6512, opened 2023-10-31
Services/LegalDocuments: Remove default value for longtext

lscharmer:
In MySQL it is not possible to set a default value "" for a field of type longtext, blob, etc. This PR removes this and adjusts the code to conform to this.


mjansenDatabay:
@lscharmer Damn... :-). We cannot remove this stab. ILIAS 9 is already released, so we'll have to add an empty `step_2` in `trunk` and `release_9`.

PR #6513, opened 2023-10-31
[FIX] Tile Image Upload in Legacy Forms

chfsx:
In legacy forms with more than one upload, the tile upload may not work (depending on the order in which the uploads get temp-names...). The PR fixes the prblem, but for this the HTTP service must be pulled into the classes.


kergomard:
Thank you very much  @chfsx ! This looks right and I will merge and cherry-pick to ILIAS 9.

chfsx:
Thanks! The PR was already for trunk, so cherry-picking for release_9 would be needed

PR #6515, opened 2023-11-02
[FIX] maybe return image from IRSS

chfsx:
This Commit fixes a rare problem (on test9) when there is actually a tile image in the IRSS, but something goes wrong or has gone wrong when creating thumbnails.


kergomard:
Thank you very much  @chfsx !

PR #6517, opened 2023-11-02
OrgUs: fix 37895, Orgunit-Specific Permissions

nhaagen:
https://mantis.ilias.de/view.php?id=37895


klees:
Will merge manually, thanks anyway!

PR #6519, opened 2023-11-02
[FIX] 0038444: Static-URL for Users not working

chfsx:
This fixes the issue https://mantis.ilias.de/view.php?id=38444

@lscharmer for your info since you introduced the code.


kergomard:
Thank you very much @chfsx ! I will keep this open and merge after the trunk is unblocked to avoid forgetting the cherry-pick.

kergomard:
Change my mind and found a good mode to keep track of this. Will merge and port later.

PR #6518, opened 2023-11-02
UI/ViewControls, DataTable: fix initial values of ViewControls

nhaagen:
ViewControls were rendered without initially set values. 
@klees is fundamentally behind this solution, I just finalized. Thank you!

PR #6520, opened 2023-11-02
UI ViewControls/DataTable: fix initial values of ViewControls

nhaagen:
this is #6518 for v9
https://mantis.ilias.de/view.php?id=38598


klees (approved):
I just saw that this duplicates #6518. I will close the other and kindly ask @Amstutz/@thibsy to go on with this version...

thibsy (changes requested):
Hi @nhaagen and @klees

Thanks a lot for looking into this.

- [x] Input names: This PR introduces some magic strings which are used as input names that also have to be used by consumers to provide values. This is most likely also the case for other components, but best move these strings into a constant on the public interface for now. We probably want to streamline this in the future.
- [x] `StackedInputData` I: This class uses a `try`-`catch` approach which could be avoided if we introduce a `InputData::has(string $key): bool` method. IMO this feels more intuitive and also aligns nicely with the same pattern as http `RequestWrapper`s.
- [x] `StackedInputData` II: Please add a brief comment to the constructor about how the stacks will be searched in the same order as provided, meaning earlier stacks will take precedence over the latter ones.

- [x] Probable bug: when testing the behaviour introduced by this PR in the base example of a data-table, the selected columns reset to the initial selection if none of the options are selected and applied. Im not entirely sure if this is a bug in the example or the data-table. Maybe null-values are not handled correctly. Could you look into this?
- [x] `ViewControl::extractCurrentValues()`: IMO this method is unnecessary and I don't think we want derived or other classes to use this method. We can create this instance directly in `extractRequestData()`. The unit test should cover `getComponentInternalValues()` and/or `extractRequestData()` instead.
- [x] `ViewControl::getComponentInternalValues()`: please rename the `$names` argument to e.g. `$input_values`, since the behaviour slightly changed. Maybe also provide a brief doc-comment about the name => value mapping of the returned value.
- [x] Conflicts: please rebase this onto the release_9 branch to resolve conflicts.

Kind regards,
@Amstutz and @thibsy

nhaagen:
Hi @Amstutz and @thibsy,
it took a while, but I think I have adressed the issues.
Would you mind to have another look?
thank you

thibsy (changes requested):
Hi @nhaagen

Thanks a lot for implementing our requested changes along with our suggestions!

There are still some minor things left to do, please implement the following changes:

- [x] `StackedInput::getOr()`: this method still uses the `try`-`catch` approach. Please implement this similar to `get()` but with a `return $default` instead of a logic exception.
- [x] `StackedInputDataTest` I: there are several test-cases which should use `getOr()` instead of just `get()`, as indicated by the additional arguments and test-names.
- [x] `StackedInputDataTest` II: you create a dynamic property [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6520/files#diff-e0a2288edf70fbdba34d61a5d470cc33314916e3fac4439c11915ea046dc043aR37), please declare it explicitly.
- [x] `Pagination` input names: there are still `offset` and `limit` usages as strings in several unit-tests, which can be replaced by the newly added constants.

Thanks for bearing with us!

Kind regards,
@Amstutz and @thibsy

thibsy (approved):

thibsy:
Thx @nhaagen for bearing with us. This LGTM now, so I'll merge and pick this to trunk.

thibsy:
Picked to trunk in 14fab546ae4702b7b33dc0c2254377fc298d43cb.

PR #6522, opened 2023-11-02
OrgUnits: Change Access To Translations

kergomard:
Hi @klees 

I would suggest to get rid of the translations-class in OrgUnits for ILIAS 9 already. This would also remove one of the two usages of the old Table. If you would prefer to only do this with ILIAS 10, I can provide a PR after the trunk is open again.

Best,
@kergomard


kergomard:
Ok, the copyright checker is complaining about a copyright in a class that was removed ;-).

klees:
Love to get rid of some code.

klees:
Thanks @kergomard, picked to trunk.

PR #6525, opened 2023-11-02
38454: Failed test: Enable/Disable Staff View

alex40724:
See https://mantis.ilias.de/view.php?id=38454


klees:
Change already included in trunk. Thx @alex40724?

PR #6526, opened 2023-11-02
UI DataTable: Fix total row count in examples

tfamula:
Since the ViewControls are implemented in the DataTable now, `getTotalRowCount` in the KS examples of the DataTable should not stay `null` anymore.

Please also cherry-pick to ILIAS 9.


klees:
Picked to release_9 and trunk manually. Thanks a bunch!

PR #6527, opened 2023-11-03
UI, ViewControl Pagination: #38495, fix 'off by one'

nhaagen:
https://mantis.ilias.de/view.php?id=38495


tfamula:
@klees Since the bug is quite significant IMO, I would be very happy if we could merge this until the community testing of ILIAS 9 starts next week.

klees:
Picked to trunk.

PR #6529, opened 2023-11-03
UI, DataTable: #38429, typecheck for Icons in Columns

nhaagen:
https://mantis.ilias.de/view.php?id=38429

### DIFF
--- a/Modules/TestQuestionPool/classes/QuestionTable.php
+++ b/Modules/TestQuestionPool/classes/QuestionTable.php
@@ -137,8 +137,8 @@ class QuestionTable extends ilAssQuestionList implements Table\DataRetrieval
         $f = $this->ui_factory->table()->column();
         $df = $this->data_factory->dateFormat();
         $date_format = $df->withTime24($this->data_factory->dateFormat()->germanShort());
-        $icon_yes = $this->ui_renderer->render($this->ui_factory->symbol()->icon()->custom(ilUtil::getImagePath('standard/icon_checked.svg'), 'yes'));
-        $icon_no = $this->ui_renderer->render($this->ui_factory->symbol()->icon()->custom(ilUtil::getImagePath('standard/icon_unchecked.svg'), 'no'));
+        $icon_yes = $this->ui_factory->symbol()->icon()->custom(ilUtil::getImagePath('standard/icon_checked.svg'), 'yes');
+        $icon_no = $this->ui_factory->symbol()->icon()->custom(ilUtil::getImagePath('standard/icon_unchecked.svg'), 'no');
 
         return  [
             'title' => $f->link($this->lng->txt('title')),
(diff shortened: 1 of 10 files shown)


tfamula:
Hi @nhaagen!

Thanks for the PR!
Do I see it correctly that in future only the `Icon` Component is allowed for the `StatusIcon` Column?
Please note, that this Column type already has some usages since ILIAS 9 (e.g. Skill Management, but also other components), which have to be adapted.
Moreover, I would be happy about unit tests which test the new `Boolean::format()` and `StatusIcon::format()` methods.

nhaagen:
Hi @tfamula ,
it's just been very recent that Colum::format may return Components. 
I really hope to not cause too much adaption trouble in enforcing the icons, since I think it's a valuable improvement over pre-rendering.
I added some tests, too.

klees (changes requested):
Hi @nhaagen,

LGTM, mostly. At first I wondered if we would really want to have flexible icons for true/false, but then I realized that a boolean could express various states, e.g. "happy", where we could use smiley faces or something. Then this occured to me:

* [x] Glyph: add `Glyph` as another option for the `Boolean` column.

That's it.

Thanks!

nhaagen:
Screenshot from 2024-01-11 09-23-03

klees:
@Amstutz, @thibsy Would you be so kind and move this on? Thanks!

oliversamoila:
Hello @klees , 
can you briefly explain to me why we need a glyph column as a representation of booleans?
The boolean indicates a value - like in the example yes/no , 0/1, cross, tick. So far so good. I also understand that there are symbols that represent this well.
But the glyph would serve to trigger an underlying action. With a value pair in the sense of a boolean, I would then expect the function of a toggle. 
Is that the case? If so, shouldn't it rather be a toggle labelled with two glyphs? And if not, what do we do with the Boolean glyph?

Thank you very much in advance.

klees:
Hi @oliversamoila,

no, we do not intent to introduce a toggle in disguise here.

The glyphs are currently designed to include an action, as you mention. We (i.e. the UI coordinators) found this to problematic on various occasions: Bulky Buttons with Glyphs could have had two actions. We had to jump extra hoops for a11y, although we already solved problems for buttons. Glyphs behave strangely e.g. in the Pagination. We are currently looking to remove the action from the Glyphs and shift it to the Buttons instead. #6922 shows the current code. Should come out a lot cleaner once we have done the move.

So basically, we want to the symbolism from the glyphs (booleans could stand for e.g. sad/happy) but not the actions.

@nhaagen: Could you clarify that in the description and then check programmatically if the provided actions do indeed carry no action?

Hope this makes sense...

Kind regards!

klees (approved):

tfamula:
Fix usages in Skill Management: https://github.com/ILIAS-eLearning/ILIAS/commit/38f1f568a6fd53a3b5d3d9c791719bb6a370c8da

PR #6534, opened 2023-11-03
remove timing functions from toast

iszmais:
(This PR was recreated since its origin fork was deleted: https://github.com/ILIAS-eLearning/ILIAS/pull/6394)

In repsonse to the rejection of the proposal of individual timing function per toast (https://github.com/ILIAS-eLearning/ILIAS/pull/6164) this removes all occurences prepared for that in consequence.

### DIFF
--- a/Services/Awareness/classes/Provider/AwarenessToastProvider.php
+++ b/Services/Awareness/classes/Provider/AwarenessToastProvider.php
@@ -84,9 +84,7 @@ class AwarenessToastProvider extends AbstractToastProvider
                     $this->if->identifier(self::PROVIDER_KEY . '_' . $this->dic->user()->getId()),
                     $this->dic->language()->txt('awareness_now_online')
                 )
-                ->withIcon($this->dic->ui()->factory()->symbol()->icon()->standard(Standard::USR, ''))
-                ->withVanishTime((int) $setting->get('osd_vanish', (string) Toast::DEFAULT_VANISH_TIME))
-                ->withDelayTime((int) $setting->get('osd_delay', (string) Toast::DEFAULT_DELAY_TIME));
+                ->withIcon($this->dic->ui()->factory()->symbol()->icon()->standard(Standard::USR, ''));
             $links = [];
             foreach ($new_users as $user) {
                 $uname = "[" . $user['login'] . "]";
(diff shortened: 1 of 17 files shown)


iszmais:
@klees  I had to recreate this PR due to the update/removal of my own fork.

The test are fixed the way we agrred on in our meeting.

Otherwise feel free to merge!

Greetings,
@iszmais

klees (approved):
LGTM!

@Amstutz/@thibsy Could you please have a quick look here as well? I supported @iszmais to find the general scheme for the solution and I think someone should double check...

Kind regards!

thibsy (changes requested):
Hi @iszmais

Note that we have merely reviewed the changes to the UI framework, you may request a review from the other affected components maintainers.

- [ ] Magic numbers: You are introducing [magic numbers here](https://github.com/ILIAS-eLearning/ILIAS/pull/6534/files#diff-63b53dda562adbd47190b76d2dd456d70b59eb88052cb3001c1043d7ccde6c01R591-R592). We suggest to move them into a constant or introduce an according getter for them.

- [ ] Ambiguity: [This return value](https://github.com/ILIAS-eLearning/ILIAS/pull/6534/files#diff-3bd0462fb4d216f715814f268c5ea0c8399d6abef3de459ea4918c56013f5dc4R63-R66) is ambiguous. We should either introduce a corresponding new data-type or split up the values into separate functions which always return an integer. If you decide to go with a new data-type, I suggest to implement it in the toast namespace.
- [ ] `ToastRendererFactory` I: Please use the `$contexts` argument to determine if the toast-renderer should be returned and do so explicitly, since you are using additional options for instantiation. As default, the parent method should be called.
- [ ] `ToastRendererFactory` II: Please rename the variables `$c`-something with their actual names and add type-hints to each argument.
- [ ] ILIAS versions: since this PR is implemented for ILIAS 8 it must be cherry-picked for ILIAS 9 and trunk as well. Due to the rather big changes between these versions, we kindly ask you to provide separate PRs for these versions as well.

Kind regards,
@Amstutz and @thibsy

iszmais:
Thank you for your response @thibsy & @Amstutz 

unfortunatly this seems this seems to be more time consuming than expected.
And since this dont create any issue i will discard this changes and close the PR.

But i encourage the coordinators to implement their decision based on this PR.

Greetings,
@iszmais

thibsy:
Hi @iszmais

I have added a roadmap-entry to the UI components for this task and referenced this PR as a proposal (see https://github.com/ILIAS-eLearning/ILIAS/commit/3231be1af86ee1773eb2e7dac9d95f2b63a264b7).

Kind regards,
@thibsy

PR #6537, opened 2023-11-03
Prevent toast from vanish when an interaction happens

iszmais:
(This PR was recreated since its origin fork was deleted: https://github.com/ILIAS-eLearning/ILIAS/pull/6166)

This PR prevent the vanishment of a toast when the user hovers it.
This is limited to the global refreshintervall of toasts. When another Intervall start the toast still disappears

### DIFF
--- a/src/UI/templates/default/Toast/tpl.toast.html
+++ b/src/UI/templates/default/Toast/tpl.toast.html
@@ -1,5 +1,5 @@
 <div class="il-toast-wrapper" <!-- BEGIN id -->id="{ID}"<!-- END id --> data-vanishurl="{VANISH_ASYNC}" data-vanish="{TOAST_VANISH}" data-delay="{TOAST_DELAY}">
-    <div class="il-toast">
+    <div class="il-toast" tabindex="0">
         <div class="icon">{ICON}</div>
         <div class="title">{TITLE}</div>
         {CLOSE}
--- a/src/UI/templates/js/Toast/toast.js
+++ b/src/UI/templates/js/Toast/toast.js
@@ -3,8 +3,9 @@ il.UI = il.UI || {};
 il.UI.toast = ((UI) => {
     let vanishTime = 5000;
     let delayTime = 500;
+    let queue = new WeakMap();
 
-    let setToastSettings = (element) => {
+    const setToastSettings = (element) => {
         if (element.hasAttribute('data-vanish')) {
             vanishTime = parseInt(element.dataset.vanish);
         }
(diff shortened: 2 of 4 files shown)


iszmais:
@thibsy I had to recreate this PR due to the update/removal of my own fork.

However, this should still be ready to merge imo since it tackled the targeted problems.
The accesebility of toast is an will be an issue in future already because of its volatile nature but is no concern of this PR i assume since it dont intefere in a bad way with that topic. I might even improve it at least a bit.

kergomard:
Hi @thibsy  

I found this PR while looking for PRs without acitivity as part of the preparation of the meeting of the TB. This got stuck somewhere on the way, right? May I to have a look at this and see how you want to proceed?

Thank you very much and best,

iszmais:
@kergomard  Thanks for remembering me about that.

Yes, this is still a useful feature of IMO since it prevents the automatic toast vanishing from "stealing" the toast from an interacting user, which is a very unpleasant user experience.

In https://github.com/ILIAS-eLearning/ILIAS/pull/6166 we had a few discussions about its accessibility. To summarize my point here: Yes, this is not fully accessible, but because volatile content **cannot** be fully accessible due to the incompatibility of these two concepts. Therefore, this is less an issue of the toast here than an issue in general, and most likely an unsolvable one.

Under That circumstance I would appreciate if you could take another look into this.

Greetings,
Ingmar Szmais

thibsy (changes requested):
Hi @iszmais,

We had another look into this and think, since the suggested feature now considers keyboard **and** mouse interactions, the solution is accessible enough for a merge. We can always improve the accessibility in further iterations. 

However, before we can merge this, please implement the following change(s):

- [x] As mentioned in [a previous review](https://github.com/ILIAS-eLearning/ILIAS/pull/6166#pullrequestreview-1644943401), we think that exposing the timeout identifier is not a good practice. Please initialise each toast using some exposed `createToast()` function inside the component renderer, which will take the ID as an argument. This way we should be able to store the timeout identifiers inside a `Map`.

Feel free to contact us on Discord if you have any questions about this.

Kind regards,
@thibsy

iszmais:
Hi @thibsy ,

I implemented the change, but I used a `weakMap` (recommended by @lscharmer) since this seems more practical for volatile objects in any way.

Pls have another look.

Greetings,
@iszmais

thibsy (approved):
This LGTM now - using a `WeakMap` makes sense. I'll merge later today and pick for all supported versions. Thx a lot!

PR #6538, opened 2023-11-03

iszmais:
Resolves https://mantis.ilias.de/view.php?id=38015
Resolves https://mantis.ilias.de/view.php?id=38017

### Initial Problem

Values of a multi input fields are transmitted to the Form/JS but string concatination and are therefore susceptible for a lot of context related issues of the form value.
This causes all input field of that kind to misfunction on specific input value and makes this feature unusable for some input types (event if it is enabled for those atm).

### Solution

This PR encodes the values to a context unrelated format (base64) before transmitting them to the JS where they are decoded and proper handled. Through that the inputs are preserved similar in every case, not matter of their semantically value.
Further this increases the support of that feature to all input types as long as their value is a multi scalar type and therefore makes it usable not only for all current types where its enabled, but for upcoming and custom types of any kind, too.

### Conflicts

- Performance: The transformation causes a slight increase in workload and space. However, these changes are negligible in relation to the overall performance. A form where this increase would be noticeable would already struggle with displaying such large information via a user interface.
- Current persisted values: The internal persistence of input values is not affected by those changes therefore all current occurrences will stay the same. Also this feature has only 1 current usage wich isn't scalar (ilOrgUnitPositionFormGUI =>  ilOrgUnitGenericMultiInputGUI)

If there is any struggle or concern with the proposed changes please leave me a feedback.

Greetings,
@iszmais

### DIFF
--- a/Services/Form/classes/class.ilPropertyFormGUI.php
+++ b/Services/Form/classes/class.ilPropertyFormGUI.php
@@ -817,7 +817,7 @@ class ilPropertyFormGUI extends ilFormGUI
             $multi_values = $item->getMultiValues();
             if (is_array($multi_values) && sizeof($multi_values) > 1) {
                 $multi_value = new ilHiddenInputGUI("ilMultiValues~" . $item->getPostVar());
-                $multi_value->setValue(implode("~", $multi_values));
+                $multi_value->setValue(base64_encode(json_encode($multi_values)));
                 $this->addItem($multi_value);
             }
             $cfg["multi_values"] = $multi_values;
--- a/Services/Form/js/ServiceFormMulti.js
+++ b/Services/Form/js/ServiceFormMulti.js
@@ -48,7 +48,7 @@ var ilMultiFormValues = {
 	 */
 	addEvent: function(e) {
 		var id = $(e.delegateTarget).attr('id').split('~');
-		ilMultiFormValues.add(id[1], id[2], '');
+		ilMultiFormValues.add(id[1], id[2], []);
 	},
 
 	/**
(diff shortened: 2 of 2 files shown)


iszmais:
@alex40724  I had to recreate this PR due to the update/removal of my own fork.

In my opinion this is still a major improvement to the form. Do you need any further assistence on testing this beside my previous protocol?

> I already texted that for SelectInputs and TextInputs (both, Multi on and off) and for an own heritance of an >CombinationInput wich includes a Text and Select Field.
>
>I didnt checked more Input types since these already where not supported be the current/origin Multi select JS and >therefore i assumed these Inputs are also not allowed for the multi flag (dont implement ilMultiValuesItem).
>
>But i dont know if this might already have a bug for the current state, this woudl need further testing.
>
>Also evry usage in the current ILIAS Core of Multi flagged Inputs only uses Selects or Texts and, at least all i could find, >worked as intended with these changes.

Otherwise feel free to merge!

Greetings,
@iszmais

iszmais:
@alex40724 I fixed the timely related conflicts here. Do you need any more input from me here (beside my response to your questions of the [origin PR](https://github.com/ILIAS-eLearning/ILIAS/pull/6341))?

Otherwise i would appreciate this inside the core since its risk-free and widely beneficial for form usage.

Greetings,
@iszmais

mjansenDatabay:
Hi @iszmais,

thx for your pull request. Unfortunately, there is currently no one who feels responsible for this code. We will discuss this kind of PR during the next JF as part of a new agenda point "Looking for Shepherd".

Best regards,
Michael

iszmais:
Reopened for ILIAS 9 here: https://github.com/ILIAS-eLearning/ILIAS/pull/8881

PR #6546, opened 2023-11-06
Style: make style.zip pass validation (38442)

schmitz-ilias:
This PR fixes [38442](https://mantis.ilias.de/view.php?id=38442) by making libs/ilias/Style/basic_style/style.zip comply with the new import validation. Two changes were required:

- The current style.zip unzips as style/style/. I removed the unnecessary nested style folder.
- I removed the `TargetStyle` attribute of the `Manifest` element in the manifest.xml. It doesn't seem to show up in any other import file, so it is not in the associated schema file. If it is required for Styles and can't be taken out, we could of course change the schema.


alex40724:
Thanks.

PR #6550, opened 2023-11-07
(ILIAS7) Course Bugfix #0028404: Index for loc_settings

fhelfer:
https://mantis.ilias.de/view.php?id=28404


schmitz-ilias:
Thanks for the improvement @fhelfer (and sorry for the wait)

PR #6552, opened 2023-11-07
(ILIAS9) Block Bugfix #0038031: Open page 2 of To Do block: Dashboard throws "ilBlockGUI::setOffset(): Offset out of range."

fhelfer:
https://mantis.ilias.de/view.php?id=38031


alex40724:
Ok, I didn't see this on. You usually do not need to provide multiple PRs for the same issue. Change is in 9 branch already. (Will be merged to trunk as soon as I have a working trunk)

PR #6556, opened 2023-11-07
Dashboard Bugfix #0038612: Clicking View Controls in Block Membership Block redirects to Dashboard

fhelfer:
https://mantis.ilias.de/view.php?id=38612


Services/Dashboard/Block/classes/class.ilDashboardBlockGUI.php
```diff
@@ -186,7 +188,12 @@ protected function initAndShow(): void
             $this->http->sendResponse();
             $this->http->close();
         }
-        $this->returnToContext();
+        if ($this->parent === ilDashboardGUI::class) {
+            $this->returnToContext();
+            return '';
```
iszmais: Pls Remove that return. In case of a interruption in `returnToContext()` we dont want to see an empty page but preferably a falls back on the other view.

Services/Membership/classes/class.ilMembershipOverviewGUI.php
```diff
@@ -64,12 +64,8 @@ public function executeCommand(): void
...
     {
-        $main_tpl = $this->main_tpl;
-        $lng = $this->lng;
-
-        $main_tpl->setTitle($lng->txt("my_courses_groups"));
-
+        $this->lng->loadLanguageModule('pd');
         $block = new ilMembershipBlockGUI();
-        $main_tpl->setContent($block->getHTML());
+        $this->main_tpl->setContent($block->getHTML());
```
iszmais: Pls Remove all changes inside the Membership service.
These dont seem to be neccessary for the resolvement of the Issue, except the loading of the language module but wich should also be part of the preceeding process inside the dashboard (The memberships do not depend on the dashboard and should therefore dont need to fullfill its duties).

iszmais (changes requested):
@fhelfer Thank you again for your continous contribution!

This looks good to me, i have just 2 minor concerns.

Greetings,
@iszmais

iszmais (approved):
Thx for the quick fix!

PR #6558, opened 2023-11-07
Exercise: Fix Removal of Background Task

kergomard:
This is a fix for: https://mantis.ilias.de/view.php?id=38620

Best,
@kergomard


alex40724:
@kergomard Thanks for the PR. Could you explain, what happens here? How is this related to the removal of the task?

kergomard:
Hi @alex40724 

This does simply not work, because the `input` you try to set on `->setValue()` is an `array`. As this throws an exception, execution stops there and the UserInteraction never finishes, thus it will also never be fully performed and it will remain in the list. I'm actually unsure if the return value does really matter. I tried returning `$download_name` or `$zip_name`, both do the trick.

Best,
@kergomard

alex40724:
Thanks!

PR #6559, opened 2023-11-07
LSO: #34712, return LP_MODE_COLLECTION in LP defaults

nhaagen:
https://mantis.ilias.de/view.php?id=34712

### DIFF
--- a/Modules/LearningSequence/classes/LearnerProgress/class.ilLSLP.php
+++ b/Modules/LearningSequence/classes/LearnerProgress/class.ilLSLP.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -18,6 +16,8 @@ declare(strict_types=1);
  *
  *********************************************************************/
 
+declare(strict_types=1);
+
 /**
  * lp connector
  */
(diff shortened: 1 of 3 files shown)


Modules/LearningSequence/classes/LearnerProgress/class.ilLSLP.php
```diff
@@ -18,14 +16,19 @@
...
  */
 class ilLSLP extends ilObjectLP
 {
     public static function getDefaultModes(bool $a_lp_active): array
     {
-        return [ilLPObjSettings::LP_MODE_DEACTIVATED];
+        return [
+            ilLPObjSettings::LP_MODE_DEACTIVATED,
+            ilLPObjSettings::LP_MODE_COLLECTION
+        ];
```
schmitz-ilias, line 36: Hi @nhaagen, this should be something like
```
if ($a_lp_active) {
    return [
        ilLPObjSettings::LP_MODE_DEACTIVATED,
        ilLPObjSettings::LP_MODE_COLLECTION
    ];
}
return [ilLPObjSettings::LP_MODE_DEACTIVATED];
```
Otherwise it would be possible to choose 'collection' as the default mode for LSO when the learning progress is globally deactivated. (Semantically this all makes very little sense, sorry about that...)

klees:
Picked to release_7 and release_8.

klees:
Picked to trunk.

PR #6561, opened 2023-11-08
UI: adjust SCSS for Apply and Reset Input Filter Buttons. (#38425)

lukastocker:
[Mantis Ticket 38425](https://mantis.ilias.de/view.php?id=38425)

The position of those buttons was adjusted as the background padding for bottom removed. It looked good in the documentation but not for e.g. in Users and Roles > Didactic Templates - therefore an adjustment was needed.


dsstrassner:
Hi @Amstutz, a quick review would be very nice, because this drives me crazy in the question pool on test9 😁

PR #6562, opened 2023-11-09
T&A: Bugfix #0038398: Bugfix Lock Answers Vertical Ordering Question

fhelfer:
https://mantis.ilias.de/view.php?id=38398


kergomard:
Thank you very much for this PR @fhelfer !

...but there is a `!` missing here. I fixed it with: https://github.com/ILIAS-eLearning/ILIAS/commit/b63559afe57e12d56b4f0d27a4597cf5d9412cab as this was not worth the trouble of an additional roundtrip.

Thanks  again and best,
@kergomard

PR #6563, opened 2023-11-09
LSO, Export: fix 38647

nhaagen:
https://mantis.ilias.de/view.php?id=38647


klees:
Cherry-picked to trunk.

PR #6564, opened 2023-11-09
T&A Bugfix #0038404: Last scored access column can not be sorted

fhelfer:
https://mantis.ilias.de/view.php?id=38404


kergomard:

I fixed this the other way around, as I feel "scored_pass_finished_timestamp" is by far clearer than "pass_finished".

See: https://github.com/ILIAS-eLearning/ILIAS/commit/069fd074eaa1cb9f36a254d5dc016ab34ddcc84f

So, I close this.

Thank you again and best,
@kergomard

PR #6566, opened 2023-11-10
UI: adjust 'dropdown' class to fix displaying issues with the dropdow…

lukastocker:
…n buttons in the Page Editor Column Layout Element.(#38150)

[Mantis Ticket 38150](https://mantis.ilias.de/view.php?id=38150)

### DIFF
--- a/templates/default/070-components/legacy/Services/_component_copage.scss
+++ b/templates/default/070-components/legacy/Services/_component_copage.scss
@@ -798,6 +798,7 @@ p#copg-auto-save {
 	> .dropdown {
 		overflow: hidden;
 		position: static;
+		display: block;
 		> ul {
 			position: absolute;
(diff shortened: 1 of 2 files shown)


Amstutz:
@catenglaender : can you check for possible side effects here. If done and good to merge, assign back to me.

catenglaender:
I don't think we should solve it like this. Many legacy components don't use btn-groups and might rely on dropdowns being inline-block. I will discuss this with @lukastocker

lukastocker:
> I don't think we should solve it like this. Many legacy components don't use btn-groups and might rely on dropdowns being inline-block. I will discuss this with @lukastocker

After discussing this PR with @catenglaender I pushed the changes right now. Therefore I will assign this PR back to @Amstutz .

PR #6570, opened 2023-11-10
UI: adjust poll progressbar SCSS. (#38157)

lukastocker:
[Mantis Ticket 38157](https://mantis.ilias.de/view.php?id=38157)

### DIFF
--- a/templates/default/070-components/legacy/Services/UIComponent/_component_progressbar.scss
+++ b/templates/default/070-components/legacy/Services/UIComponent/_component_progressbar.scss
@@ -7,10 +7,13 @@ $progress-bg: #bbb;
 .progress {
 	height: 15px;
 	min-width: 100px;
+	background-color: lighten($il-neutral-color, 55%);
 }
 
 .progress-bar {
 	font-size: $il-font-size-small;
 	line-height: 15px;
-	min-width: 30px;
+	min-width: 50px;
+	text-align: center;
+	color: $il-main-darker-bg;
 }


klees:
@catenglaender Again, fixed width. Please advice...

lukastocker:
> @catenglaender Again, fixed width. Please advice...

This implementation was discussed with @catenglaender today. 
As there are more issues which need a decision when and if we want to fix those issues too and if we still want to merge this PR, this PR and the Mantis Ticket will be presented in the CSS-Squad Meeting.

Amstutz:
We discussed the fix today in the squad. We decided to merge the solution as is, to come close to the state as is in ILIAS 8. 

However, @catenglaender will look into the code and check if e.g. variables could be used for the color.

templates/default/070-components/legacy/Services/UIComponent/_component_progressbar.scss
```diff
@@ -7,10 +7,13 @@ $progress-bg: #bbb;
 .progress {
 	height: 15px;
 	min-width: 100px;
+	background-color: rgb(68, 74, 76);
+	box-shadow: rgba(6, 6, 6, 0.1) 0px 1px 2px inset;
```
lukastocker: Hi @catenglaender ,

haben wir hierfür bereits Variablen die dem entsprechen bzw. sollten wir ggf. neue hinzufügen?
mixin box-shadow ist ja auf jeden Fall vorhanden.

catenglaender: Just to record it here as well, in a conversation we decided to use $il-neutral-color for the gray background of the bar and no box shadow.

lukastocker: I just pushed the changes to this PR.

catenglaender:
Unless we want to refactor the existing code more (which might introduce more bugs), this is good to go from my point of view.

Just a note what could be done:
* removing the min-width for the outer container and rely on the implied display: block for the full width.
* check if we really need to rely on the line height for the height of the bar

But maybe that's something for another day as it could break existing implementation.

catenglaender (approved):

Amstutz:
I somewhat dislike the `lighten` in the code, since it basically creates a new color at this point. also, the hard-coded 50px seems somewhat odd.

> this is good to go from my point of view.

I can follow that reasoning and merge.

catenglaender:
We will can the left over messy parts with a refactoring. whenever someone wants the bars to be able to show 0% correctly.

PR #6571, opened 2023-11-10
UI: adjust Matrix Question (Question Pool Survey) input field width. …

lukastocker:
…(#38154)

https://mantis.ilias.de/view.php?id=38154


Amstutz:
@catenglaender can you look into this and check if a solution without !important is possible?

lukastocker:
> @catenglaender can you look into this and check if a solution without !important is possible?

Hi @Amstutz ,

I discussed this PR with @catenglaender  yesterday and pushed the new solution. Therefore I am assigning this back to you.

Kind regards
Luka

PR #6574, opened 2023-11-10
Object: fix array access (38657)

schmitz-ilias:
This PR is a potential fix for [38657](https://mantis.ilias.de/view.php?id=38657), the array access issue makes the copying of some folders fail.


kergomard:
Thank you very much for the PR @schmitz-ilias !

I fixed this differently in: https://github.com/ILIAS-eLearning/ILIAS/commit/66433f7a88f2bb2af4525df6f01a80be4cf7eda0

Best,
@kergomard

PR #6575, opened 2023-11-13
User Bugfix #0038724: ilObjUserFolderGUI PHP8 strict type bugs

fhelfer:
https://mantis.ilias.de/view.php?id=38724

### DIFF
--- a/Services/User/classes/class.ilObjUserFolderGUI.php
+++ b/Services/User/classes/class.ilObjUserFolderGUI.php
@@ -666,14 +666,14 @@ class ilObjUserFolderGUI extends ilObjectGUI
 
         foreach ($user_ids as $user_id) {
             $ufield = new ilHiddenInputGUI('id[]');
-            $ufield->setValue($user_id);
+            $ufield->setValue((string) $user_id);
             $form->addItem($ufield);
         }
 
         // return to search?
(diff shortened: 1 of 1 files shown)


kergomard:

This looks mostly right, ...apart from this [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6575/files#diff-ca68ff2332ed52e681485f1c58b7a71a16b67d73c3c09c85900720267d35ccdaL1553-R1564). If you remove the `return`s we would need breaks, as far as I can see or we need to reorganize the block.

I would prefer to either keep the current version or to reorganize the block.

Thanks again and best,
@kergomard

fhelfer:
> 
> This looks mostly right, ...apart from this [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6575/files#diff-ca68ff2332ed52e681485f1c58b7a71a16b67d73c3c09c85900720267d35ccdaL1553-R1564). If you remove the `return`s we would need breaks, as far as I can see or we need to reorganize the block.
> 
> I would prefer to either keep the current version or to reorganize the block.
> 

Sharp eyes, @kergomard. I removed the line from the PR, this will most likely not run into an error but will keep existing as just a red mark in my ide for now.

Best, @fhelfer

PR #6577, opened 2023-11-13
(trunk) User Bugfix #0038724: ilObjUserFolderGUI PHP8 strict type bugs

fhelfer:
https://mantis.ilias.de/view.php?id=38724
#6575, because cherry-picks are problematic atm :)


kergomard:
Thank you very much @fhelfer for the port!

PR #6578, opened 2023-11-13
LSO: #38714, remove trailing zeros from order-input

nhaagen:
https://mantis.ilias.de/view.php?id=38714


klees:
Cherry-picked to trunk.

PR #6581, opened 2023-11-13
LSO: 38713, enable online/offline settings for all LSO content

nhaagen:
https://mantis.ilias.de/view.php?id=38713


klees:
Cherry-picked to trunk.

PR #6585, opened 2023-11-13
UI: include missing class style for example and code sections in the …

lukastocker:
…UI documentation. (#38133)

https://mantis.ilias.de/view.php?id=38133

Running the PHPUnit tests does not finish because of [this Commit](https://github.com/ILIAS-eLearning/ILIAS/commit/244ad9cae6158142ca4c58db0c5e7dca24375935).

### DIFF
--- a/templates/default/030-tools/_tool_highlighted-box.scss
+++ b/templates/default/030-tools/_tool_highlighted-box.scss
@@ -1,0 +1,30 @@
+@use "sass:map";
+@use "../010-settings/" as *;
+@use "../050-layout/basics/" as *;
+@use "../030-tools/tool_browser-prefixes" as *;
+
+// Generates highlighted boxes
+$background-color:$il-main-darker-bg !default;
+$text-color: $il-text-color-invert !default;
+$border-color:$il-main-border-color !default;
+$border: $il-main-border !default;
+$shadow-box: inset 0 1px 1px rgba(0, 0, 0, .05) !default;
+$min-height: auto !default;
+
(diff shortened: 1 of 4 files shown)


Amstutz:
Hi @lukastocker 

Thx for fixing this. The fix work, however, I am not sure, that it is correct, to have the same styling for .php and .well. Are those not different things? IMO not all code that should be in code should also be in well and maybe also the other way around.

- [ ] Split .php and .well in two different files and add as few values to each.

@catenglaender , do you agree? If so, where could the .php code be located? What should it be named?

thx

lukastocker:
> Hi @lukastocker
> 
> Thx for fixing this. The fix work, however, I am not sure, that it is correct, to have the same styling for .php and .well. Are those not different things? IMO not all code that should be in code should also be in well and maybe also the other way around.
> 
> Please change or indicate why not:
> 
>     * [ ]  Split .php and .well in two different files and add as few values to each.
> 
> 
> @catenglaender , do you agree? If so, where could the .php code be located? What should it be named?
> 
> thx

Hi @Amstutz ,
@catenglaender and myself discussed this implementation yesterday. I just pushed the changes like we discussed. 
I will assign this PR back to you.

We also discussed, that it would be best, if the alerts would act with the highlighted boxes or at least like them. This would be a new implementation/refactoring, therefore I will create a new Mantis ticket with a reference to this PR/Issue.

Kind regards,
Luka

Amstutz:
Thx a lot. nice, the code looks a lot better/more reusable now.

PR #6589, opened 2023-11-13
LSO: 38713, wrong return type

nhaagen:
https://mantis.ilias.de/view.php?id=38713
https://mantis.ilias.de/view.php?id=38714 (2nd part)


klees:
Picked to trunk.

PR #6593, opened 2023-11-14
LSO: Export and XSD

nhaagen:
This is an updated version of Import/Export for the LSO, including its ContentPages (Abstract/Extro).
It also adds the XSD.


klees:
Picked to trunk: https://github.com/ILIAS-eLearning/ILIAS/commit/061cb96f9d4cfe2d6ecf04ffcf792e262402a703

PR #6596, opened 2023-11-14
AdvMD/Wiki: fix updating page lists on reordering select options

schmitz-ilias:
This PR fixes an issue where Page Lists filtering for specific values of a (multi-)select Custom Metadata field in a Wiki get scrambled when the options of the select are reordered.

The issue also seems to occur in R7 (but I haven't checked this fix there).


alex40724:
Thanks. This fixes the reordering issue on my installation. I will check 7, too.

alex40724:
Merged down to release 7, too.

alex40724:
I added a mantis report for this for documentation purposes. https://mantis.ilias.de/view.php?id=38788

PR #6597, opened 2023-11-14
User: Fix passing data to mail component

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=38772

If approved, this has to be cherry-picked to `release_9` and `trunk` as well.

### DIFF
--- a/Services/User/classes/class.ilObjUserFolderGUI.php
+++ b/Services/User/classes/class.ilObjUserFolderGUI.php
@@ -4030,15 +4030,15 @@ class ilObjUserFolderGUI extends ilObjectGUI
 
         $umail->savePostData(
             $mail_data['user_id'],
-            $mail_data['attachments'],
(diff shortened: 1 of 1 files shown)


kergomard:

One question though:
- This for me looks like a band-aid, shouldn't we actually fix this in `ilMail::fetchMailData()`? We would need to potentially pepper every use of `ilMail` with these safeguard otherwise.

Thank you very much and best,
Stephan

mjansenDatabay:
You are right, but I think that `fetchMailData` already provides a valid record (of course as we know, arrays are not a good solution) or `null`.

IMO the problem occurs because of this line: https://github.com/ILIAS-eLearning/ILIAS/pull/6597/files#diff-ca68ff2332ed52e681485f1c58b7a71a16b67d73c3c09c85900720267d35ccdaR4025

PR #6598, opened 2023-11-14
9/UI/DataTable full width placeholder 38364 38072

catenglaender:
https://mantis.ilias.de/view.php?id=38072
https://mantis.ilias.de/view.php?id=38364

# Issue

Empty space next to a table looks odd. Other page elements do not line up at the far right edge.

image

image

# Change

As suggested by @catkrahl a background extension below the table now spans over the entire width of the page and "reserves" the space to the right of the table. This should make especially the right edge of the page feel more balanced when a table is small.

Also added a margin between viewcontrols and table.

image


Amstutz:
looks good, thx a lot! Kudos for adapting the Uni Tests.

PR #6599, opened 2023-11-15
a11y fix 37761 for ilias7

okaufman:
This will fix a11y issue 37761 for ilias7
https://mantis.ilias.de/view.php?id=37761


Amstutz:
thx a lot!

PR #6602, opened 2023-11-15
Favourites Bugfix #0038792: Configure Button in Favourites shows empty page

fhelfer:
https://mantis.ilias.de/view.php?id=38792

### DIFF
--- a/Services/Repository/Favourites/classes/class.ilFavouritesListGUI.php
+++ b/Services/Repository/Favourites/classes/class.ilFavouritesListGUI.php
@@ -25,7 +25,6 @@ declare(strict_types=1);
  */
 class ilFavouritesListGUI
 {
-    private ilGlobalTemplateInterface $main_tpl;
     protected \ILIAS\DI\UIServices $ui;
     protected ilCtrl $ctrl;
     protected ilLanguage $lng;
@@ -73,7 +72,7 @@ class ilFavouritesListGUI
                 $this->lng->txt('rep_configure'),
                 $this->ui->factory()->legacy('PH')
             )->withAdditionalOnLoadCode(function ($id) {
-                return "document.getElementById('$id').setAttribute('data-depth-level', null)";
+                return "document.body.appendChild(document.getElementById('$id'));";
             });
             $roundtrip_modal = $roundtrip_modal->withAsyncRenderUrl(
                 $this->ctrl->getLinkTargetByClass(
--- a/Services/Repository/Favourites/classes/class.ilFavouritesListGUI.php
+++ b/Services/Repository/Favourites/classes/class.ilFavouritesListGUI.php
@@ -25,6 +25,7 @@ declare(strict_types=1);
  */
 class ilFavouritesListGUI
 {
+    private ilGlobalTemplateInterface $main_tpl;
     protected \ILIAS\DI\UIServices $ui;
     protected ilCtrl $ctrl;
     protected ilLanguage $lng;
@@ -63,20 +64,34 @@ class ilFavouritesListGUI
             }
         }
         if (count($item_groups) > 0) {
+            $tpl = new ilTemplate("tpl.favourites.html", true, true, "Services/Repository/Favourites");
             $ctrl->setParameterByClass(ilSelectedItemsBlockGUI::class, "view", "0");
             $ctrl->setParameterByClass(ilSelectedItemsBlockGUI::class, "col_side", "center");
             $ctrl->setParameterByClass(ilSelectedItemsBlockGUI::class, "block_type", "pditems");
-
             // see PR discussion at https://github.com/ILIAS-eLearning/ILIAS/pull/5247/files
+            $roundtrip_modal = $this->ui->factory()->modal()->roundtrip(
+                $this->lng->txt('rep_configure'),
+                $this->ui->factory()->legacy('PH')
+            )->withAdditionalOnLoadCode(function ($id) {
+                return "document.getElementById('$id').setAttribute('data-depth-level', null)";
+            });
+            $roundtrip_modal = $roundtrip_modal->withAsyncRenderUrl(
+                $this->ctrl->getLinkTargetByClass(
+                    [ilDashboardGUI::class, ilColumnGUI::class, ilSelectedItemsBlockGUI::class],
+                    'removeFromDeskRoundtrip'
+                ) . '&page=manage&replaceSignal=' . $roundtrip_modal->getReplaceSignal()->getId()
+            );
+
             $config_item = $f->item()->standard(
-                $f->link()->standard(
+                $f->button()->shy(
                     $this->lng->txt("rep_configure"),
-                    $this->ctrl->getLinkTargetByClass(["ilDashboardGUI", "ilColumnGUI", "ilSelectedItemsBlockGUI"], "manage")
+                    $roundtrip_modal->getShowSignal()
                 )
             );
             array_unshift($item_groups, $f->item()->group($this->lng->txt(""), [$config_item]));
             $panel = $f->panel()->secondary()->listing("", $item_groups);
-            return $this->ui->renderer()->render([$panel]);
+            $tpl->setVariable("FAV_CONTENT", $this->ui->renderer()->render([$panel, $roundtrip_modal]));
+            return $tpl->get();
         }
 
         return $favoritesManager->getNoItemFoundContent();
--- a/Services/Repository/Favourites/classes/class.ilFavouritesListGUI.php
+++ b/Services/Repository/Favourites/classes/class.ilFavouritesListGUI.php
@@ -63,34 +63,21 @@ class ilFavouritesListGUI
             }
         }
         if (count($item_groups) > 0) {
-            $tpl = new ilTemplate("tpl.favourites.html", true, true, "Services/Repository/Favourites");
             $ctrl->setParameterByClass(ilSelectedItemsBlockGUI::class, "view", "0");
             $ctrl->setParameterByClass(ilSelectedItemsBlockGUI::class, "col_side", "center");
             $ctrl->setParameterByClass(ilSelectedItemsBlockGUI::class, "block_type", "pditems");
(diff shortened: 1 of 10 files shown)


alex40724:
@fhelfer Thanks for the PR. I am losing a bit overview regarding the dashboard changes. Could you point me to the feature request that specifies the modal behaviour, please? Thanks.

EDIT: Ok, found it. It's a dev issue here https://docu.ilias.de/goto_docu_wiki_wpage_7504_1357.html

Services/Repository/Favourites/classes/class.ilFavouritesListGUI.php
```diff
@@ -63,20 +63,34 @@ public function render(): string
...
+                $this->ui->factory()->legacy('PH')
+            )->withAdditionalOnLoadCode(function ($id) {
+                return "document.body.appendChild(document.getElementById('$id'));";
+            });
+            $roundtrip_modal = $roundtrip_modal->withAsyncRenderUrl(
+                $this->ctrl->getLinkTargetByClass(
+                    [ilDashboardGUI::class, ilColumnGUI::class, ilSelectedItemsBlockGUI::class],
+                    'removeFromDeskRoundtrip'
+                ) . '&page=manage&replaceSignal=' . $roundtrip_modal->getReplaceSignal()->getId()
+            );
```
alex40724: The implementation inside the modal is part of the Dashboard service (ilSelectedItemsBlockGUI). Please move these lines to the Dashboard service, too. E.g. the Dashboard could provide the complete modal instance.

Services/Repository/Favourites/templates/default/tpl.favourites.html
```diff
@@ -0,0 +1,2 @@
+<script src="Services/Dashboard/Block/js/ReplaceModalContent.js"></script>
+<div>{FAV_CONTENT}</div>
```
alex40724: The only purpose of this template seems to be to integrate a js from the Dashboard service, please move this template to the Dashboard service, too.

Services/Repository/RecommendedContent/classes/class.ilDashboardRecommendedContentGUI.php
```diff
@@ -118,14 +118,12 @@ public function confirmedRemoveObject(): void
...
-            $this->ctrl->redirect($this, 'manage');
-        }
-
-        foreach ($refIds as $ref_id) {
-            $rec_manager->declineObjectRecommendation($this->user->getId(), (int) $ref_id);
+        if ($refIds !== []) {
+            foreach ($refIds as $ref_id) {
+                $rec_manager->declineObjectRecommendation($this->user->getId(), (int) $ref_id);
+            }
+            $this->main_tpl->setOnScreenMessage('success', $this->lng->txt('pd_remove_multi_confirm'), true);
```
alex40724: Is the PR mixing favourite and recommended content issues on purpose or is this a mistake?

alex40724 (changes requested):
Thanks for the PR. I think we are all struggeling with the borders of these services.

Please check my comments, may be @iszmais can assist here. I do not want to make it too complicated, but also not mix up things even more.

If the suggestions are feasible, please implement them. If not, please let me know.

Best regards
Alex

fhelfer:
> fhelfer Thanks for the PR. I am losing a bit overview regarding the dashboard changes. Could you point me to the feature request that specifies the modal behaviour, please? Thanks.
> 
> EDIT: Ok, found it. It's a dev issue here https://docu.ilias.de/goto_docu_wiki_wpage_7504_1357.html

Exactly thats the Feature request. Unfortunately a few buttons still point to dashbaord despite the user being in any other contexts. 
A better approach to handling favorite items will be targeted in later FW articles, targeting also dashboard blocks in general. Also sortations in the modal will be my next PR :)

> Thanks for the PR. I think we are all struggeling with the borders of these services.
>Please check my comments, may be iszmais can assist here. I do not want to make it too complicated, but also not mix up things even more.
>If the suggestions are feasible, please implement them. If not, please let me know.

Unfortunately i cannot implement your changes until after my vacation. But will look after this asap after.

Best, @fhelfer

Services/Repository/Favourites/classes/class.ilFavouritesListGUI.php
fhelfer: Hmm, this is very specific code to FavouritesList. The AsyncRenderURL and onload are different to what DashboardBlocks need.
I dont think this is a pretty solution. Not much code would be moved.

Services/Repository/Favourites/templates/default/tpl.favourites.html
fhelfer: This is also specifically needed for favourites in slate. DashboardBlock Modals do not want to send js contents on every modal render.

Services/Repository/RecommendedContent/classes/class.ilDashboardRecommendedContentGUI.php
fhelfer: removed

Services/Repository/Favourites/templates/default/tpl.favourites.html
alex40724: The repository component should not be responsible to load dashboard javascript code. This violates SoC principles. This results in high coupling and bad maintainability. As I wrote I understand that there is currently no clear interface, but these lines may produce severe issues in the future. A Dashboard maintainer should be able to change or even delete this JS file without the need to check other components.

Services/Repository/Favourites/classes/class.ilFavouritesListGUI.php
alex40724: It is not about the amount of code. It is about separation of concerns and coupling.

The Dashboard service provides the management view, since ilPDSelectedItemsBlockGUI is a part of it. This is already Favourites specific code. On the other hand the knowledge of things like the ctrl path ilDashboardGUI::class, ilColumnGUI::class, ilSelectedItemsBlockGUI::class are Dashboard internals. Providing only the modal instance would reduce the coupling and keep these Dashboard internals out of the Repository service.

iszmais:
Hi @alex40724 ,

i see this more as a conceptional issue (wich already made issues within the past) than a technical one. I try to summarize my general concern:
ATM the have a not fully abstracted concept of favorites. The functionality, location and naming of the favorites imply that they are exclusively part of the dashboard functionality (e.g. the command for add Favorites is "addToPersonalDesktop" aka Dashboard), but the handling within the menu implies otherwise that the favorites are an abstract component. This state proposes 2 solutions which can be considered best practice IMO:
- Case 1 "Favorites are an independent functionality": This would require the abstraction of said modal into an independent space in ILIAS. However this would also imply a full refactoring of the functionality since, even separated into a different component, the dependency to the dashboard may be reduced but cannot be fully eradicated.
- Case 2 "Favorites are a dashboard functionality": In this case the menu entry should not interfere with the configuration process of the dashboard and therefore the "Configure" link should be removed. It could be replaced with a link to the dashboard on wich the user than can configure his favorites (Similar to the current handling of mail, tags, .. inside the mainmenu and metabar). The menu would only provide a readonly view of said component just like for the repository tree or last visited.

I would highly favor this last approach since it unifies the menu entry with the convention of other menu entries and reduces another dual configuration and therefore the complexity of configruation within ILIAS in general.

Pls let me know if you have any concerns about that conceptional change.
If not i would prepare a proposal to take this to the Jour Fixe.

Greetings,
@iszmais

alex40724:
Hi @iszmais,

thanks for the answer.

My understanding was, that the Dashboard component aims to provide an interface that other components (like favourties) should serve in the future. This it at least stated in the current roadmap of the Dashboard: "In detail a component should provide all the necessary information on its presentation and required settings to the `Dashboard` by implementing a `Provider Interface`."

I am aware that we don't have an implementation for this now, but it would be great to have at least an outline of this future interface. This would make it easier to come to conclusions for these kind of discussions now imo.

You are writing "ATM the have a not fully abstracted concept of favorites.".

If the favorites component serves an interface of the dashboard, we would not need that imo.

We can decide on this in the Jour Fixe. Currently the favorites core classes are located in the Repository service. It would be ok for me to move these somewhere else.

Best

matthiaskunkel:
Jour Fixe, 15 APR 2024: We had a short discussion about the future of the Favourites service and accept the suggestion by @fhelfer and @alex40724 to move the related classes to the Dashboard component and make the function Favourites a part of the Dashboard. This change is accepted for 9 and trunk.

lang/ilias_de.lang
```diff
@@ -12996,6 +12996,7 @@ pd#:#pd_import_personal_data#:#Persönliche Daten importieren
 pd#:#pd_my_memberships_intro#:#Welcome to Your Memberships###12 12 2023 new variable
 pd#:#pd_my_memberships_intro2#:#This is your personal space for all courses and groups you participate in. Currently, there aren't any memberships assigned to you. To become a member of a course or group, there are two ways:<br /><ol><li>You can join courses or groups provided that joining them has been allowed by their administrators.</li><li>An administratrator of a course or group can assign a membership to you.</li></ol><br />In any case, the respective objects will appear automatically in this place.###12 12 2023 new variable
 pd#:#pd_my_memberships_sort_default#:#Standard-Sortierung
+pd#:#pd_no_items_to_manage#:#Keine Objekte zum entfernen verfügbar.
```
nicoroeser: Nitpick: „zum entfernen“ → „zum Entfernen“ (capitalisation).

iszmais:
@fhelfer Can you move the Favourites-classes to the Dashboard?
I would recommend the same structure as in Repository (simple subdirectory inside classes).

Thanks alot!
@iszmais

fhelfer:
@iszmais Code has been moved and configure Modal is working atm.
Please give the PR another review and also port changes to trunk after merge

Thanks, @fhelfer

iszmais:
@fhelfer Works fine thanks a lot for your collaboration!

Greetings,
@iszmais

PR #6607, opened 2023-11-16
Component: Add hasActivatedPlugin() to ilComponentRepository

tbongers-cat:
Adds a small QoL method to ilComponentRepository to make checking for active plugins a bit easier (as discussed with @klees).

### DIFF
--- a/Services/Component/classes/class.ilArtifactComponentRepository.php
+++ b/Services/Component/classes/class.ilArtifactComponentRepository.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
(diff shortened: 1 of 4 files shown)


klees (commented):
Hi @tbongers-cat,

LGTM in general. Basically, there is a 2x2 matrix of states that the new method covers. 3 of them are interesting, but you only test one. Please add the missing two testcases.

Kind regards!

tbongers-cat:
@klees Fixed, please have a look

klees:
Picked to trunk and release_9.

PR #6610, opened 2023-11-16
UI: Fix Value Can be Empty String

kergomard:
I'm not completely sure when this happens, but when rendering a form containing a DateTime Field `->withRequest()` the value for the field can be an empty string. When applying this value, the `DateTimeImmutable` will contain the current date and things go sideways.

A report has been created for the test here:
See: https://mantis.ilias.de/view.php?id=38838

### DIFF
--- a/src/UI/Implementation/Component/Input/Field/DateTime.php
+++ b/src/UI/Implementation/Component/Input/Field/DateTime.php
@@ -86,6 +86,10 @@ class DateTime extends FormInput implements C\Input\Field\DateTime
      */
     public function withValue($value): self
     {
+        // This is necessary, otherwise DateTimeImmutable will replace
+        // the empty string with the current date during rendering.
+        // The empty string is the default value posted by the input, if
+        // it is left empty.
         if ($value === '') {
             $value = null;
         }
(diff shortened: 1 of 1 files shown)


klees (changes requested):
Hi @kergomard,

thanks for looking into this. I don't think that the renderer is the appropriate location to fix this, because this would mean, that now users would see another value (none...) then we had processed on server side.

IMO, [DateTime::isClientSideValueOk](https://github.com/ILIAS-eLearning/ILIAS/blob/release_9/src/UI/Implementation/Component/Input/Field/DateTime.php#L174) would be the correct location. We should reject an empty string outright, because there is no obvious way to interpret it as DateTime, even if PHP thinks so =) If you want to go an extra mile, I would be very happy if you could come up with a better test for "good" strings then we currently have.

Kind regards!

kergomard:
Thank you for the feedback @klees 

You are right, wrong end of the changes. I've moved the conversion of the empty string from the rendering to `DateTime::withValue()`. We cannot reject the value in `DateTime::isClientSideValueOk()` as the empty string is actually the value a DateTime-Input will post, if it is left empty.

Thank you very much and best,
@kergomard

klees:
Picked to trunk.

PR #6611, opened 2023-11-17
Add missing int cast. Fixes #38829

pascalseeland:


kergomard:
Thank you very much for the PR @pascalseeland ! Will merge and pick to trunk.

PR #6615, opened 2023-11-17
TA: 38731, remove inc.AssessmentConstants.php and distribute constants

nhaagen:
https://mantis.ilias.de/view.php?id=38731

I removed the inc.AssessmentConstants.php and its according includes and distributed the remaining constants in a quite mechanical way. 
What's left are the QuestionIdentifiers - and that's another topic. I've seen questions returning their classname, the identifier, and whatnot. We should discuss in TechSquad.


dsstrassner:
Thx @nhaagen, we discussed this today at the TechSquad and liked your work. @kergomard will contact you, is anything further to discuss or to change. Also, you may approach him ;)

kergomard:
Thank you very much @nhaagen for this PR!

This again a big step in cleaning up our code!

Just one little suggestion, you don't have to follow it, but please quickly explain your reasoning:
- Couldn't we move the QuestionIdentifiers to the `QuestionInfoService`? I think we should get rid of it completely and its use is rather localized. I would then take care of removing it completely together with the QuestionInfoService itself in January when I should have some time for clean-up.

And then: Could you please rebase this quickly, I will then merge immediately.

Best,
@kergomard

kergomard:
Thank you very much @nhaagen for the rebase.

@nhaagen and @kergomard had a quick chat and decided to leave the constant in the separate class to make it easier to search for it.

Best,
@kergomard

PR #6618, opened 2023-11-17
OpenId: Fix post_logout_redirect_uri Relative

kergomard:
The post_logout_redirect_uri should not be relative, as far as I understand the protocol, otherwise the redirect can, one, not work and, two, not be checked.

See: https://mantis.ilias.de/view.php?id=38873


mjansenDatabay:
My fault, I will take care of this.

PR #6622, opened 2023-11-20
Add additional ignores to implementation finder

iszmais:
As far as i know the directories:
- docs
- lang 

should be safe to be added to the finder.

Additional, the directories (but im not 100% sure about this):
- data
- templates
- node_modules

too

`test` can be removed since it doesnt exist.

This may need some customization for the new trunk.

Further i might would consider using a white-list instead of a blacklist for this process since it seems to me the we know the specific scopes that should be searched but defining them by excluding all other, wich may cause further maintainance effort in future. But thats just an idea of mine.

### DIFF
--- a/src/Setup/ImplementationOfInterfaceFinder.php
+++ b/src/Setup/ImplementationOfInterfaceFinder.php
@@ -33,9 +33,13 @@ class ImplementationOfInterfaceFinder
      * @var string[]
      */
     protected array $ignore = [
+        '.*/data/',
+        '.*/docs/',
+        '.*/lang/',
+        '.*/templates/',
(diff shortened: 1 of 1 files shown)


klees (commented):
Hi @iszmais,

I wonder what the benefit of this would really be. The initial set of inspected files is derived from the classmap, so adding most of the folders should make no difference or we should exclude them for the classmap already. Did you observe something else?

Kind regards!

iszmais:
@klees My bad... i didnt see that this was already prefiltered by the classmap :smile:

klees:
@iszmais No problem, always good to have many eyes looking at things =)

PR #6624, opened 2023-11-20
TA: 38878, ProviderContext to main

nhaagen:
https://mantis.ilias.de/view.php?id=38878


kergomard:
Thank you very much @nhaagen ! Will merge and cherry-pick to trunk.

PR #6625, opened 2023-11-20
Fixed Mantis Bug #37783:  language update objective fails because of missing db precondition

c-knof:


katringross:
@klees it seems, that many preconditions fail, after we added this one line of code!? Is this another problem?

klees:
Hi @katringross,

could you be a bit more specific regarding the other preconditions that fail? Maybe ping me via discord for a quick debugging session, I'm available today starting at ~16.

Kind regards!

PR #6626, opened 2023-11-20
Add language svensk version

M4X1L:
New Swedish Language Translation for Ilias


matthiaskunkel:
Thanks a lot to @M4X1L for adding Swedish language support to ILIAS. That's a great contribution to ILIAS. Language file is added to Release_8 but ordering needs to be fixed and file also merged to 9 and Trunk.

PR #6628, opened 2023-11-20
[FEATURE] ilCtrl: remove deprecated methods `setCmd()` and `setCmdClass()`.

thibsy:
Hi everyone,

This PR is the follow up for #5100 and removes the deprecated methods `ilCtrl::setCmd()` and `ilCtrl::setCmdClass()`.

I have opened this PR for visibility. These changes will most likely lead to some unexpected behaviour in each affected controller as well as other ones which route through one or more of them. I have commented the usages of these methods and added an according `@todo` annotation, which most likely needs inspection by a maintainer.

There have been some heavy usages in the `COPage`, `TestQuestionPool` and `Test` services and modules as well as index files, such as `login.php`.

While going through the usages of these methods I also recognised a pattern across controllers which need to show an info-page:

```php
$ilCtrl->setCmd('showSummary');
$ilCtrl->setCmdClass(ilInfoScreenGUI::class);
```

This points into the direction of transforming this controller into a service rather than using it as an endpoint. However, this is just a realisation, this PR is not meant to discuss this particular issue.

Kind regards,
@thibsy

### DIFF
--- a/components/ILIAS/AccessControl/classes/class.ilPermissionGUI.php
+++ b/components/ILIAS/AccessControl/classes/class.ilPermissionGUI.php
@@ -150,7 +150,8 @@ class ilPermissionGUI extends ilPermission2GUI
     protected function confirmTemplateSwitch(): void
     {
         $this->ctrl->setReturn($this, 'perm');
-        $this->ctrl->setCmdClass('ildidactictemplategui');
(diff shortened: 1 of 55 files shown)


Amstutz (approved):
Thx for proposing. I will find a workaround for my components.

PR #6629, opened 2023-11-21
CSS: Fix Overlapping Focus and Text

kergomard:
This fixes an issue whereby a focus border on an input field overlaps the warning-message for erroneous entries in the field. I think, if we do anything about it, we should keep this as simple as this. The error occurs from ILIAS 7 onward.

It was reported for the test here
See: https://mantis.ilias.de/view.php?id=38867
but it is a general issue.

Thank you very much and best,
@kergomard


kergomard:
Sorry, ...I think another solution might be better...

kergomard:
I close this as I don't think this is the right solution.

PR #6630, opened 2023-11-21
#38953 fix awareness popups showing with disabled awareness tool

sroenchen:
The awareness toast-provider ignores if awareness is disabled, as long as the popup is activated.  This commit should fix that.


alex40724:
Thanks for the PR! Applied manually.

PR #6632, opened 2023-11-21
[FIX] 0038947: Update from Ilias 5.4 -> 7.26 fails because of Quotation of WHERE `1` = 1

chfsx:
This fixes the issue https://mantis.ilias.de/view.php?id=38947 and should be cherry-picked to release_8 at least

### DIFF
--- a/setup/sql/dbupdate_04.php
+++ b/setup/sql/dbupdate_04.php
@@ -24790,11 +24790,7 @@ ilDBUpdate3136::addStyleClass(
 ?>
 <#5430>
 <?php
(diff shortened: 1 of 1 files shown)


alex40724:
I think I used this construct since the update method was not able to act on empty arrays at that time if I remember correctly.

alex40724:
Thanks for the PR!

alex40724:
@chfsx Does this need cherry picking? Afaik installations should be migrated to 7 before updating to 8. In 8 this is not executed anymore.

PR #6633, opened 2023-11-21
TA: 38586, wrong offset in pagination

nhaagen:
https://mantis.ilias.de/view.php?id=38949


kergomard:
Thank you very much  @nhaagen ! Will merge and pick to trunk.

PR #6634, opened 2023-11-21
UI: adjust card title position for buttons which look and act like li…

lukastocker:
…nks. (#38682)

https://mantis.ilias.de/view.php?id=38682


lukastocker:
Hi @Amstutz ,

this implementation was discussed with @catenglaender yesterday.

Kind regards
Luka

Amstutz:
thx a lot!

PR #6641, opened 2023-11-21
Add support for IPv6 in ILIAS\Data\URI (ILIAS 8)

lscharmer:
We recently discovered in our logs more and more IPv6 addresses (explicit or through other means), which caused the initialization to fail. This PR adds the support for this.


klees (changes requested):
Hi @lscharmer,

thanks a lot for looking into this. This was indeed missing as the comments say =)

One general remark: If you open PRs for me, it is enough to create the PR for one branch, preferably the latest named branch (so `release_9` in this very moment). In general I will then port the changes myself or in rare cases ask you to do so and open additional PRs.

About this change here: Please give some love to the docu of the class to keep the current standard, e.g.: link the RFC; explain what you did or did not implement; refer to according references in the RFC.

Thanks a bunch!

lscharmer:
Hi @klees,
> Hi @lscharmer,
> 
> thanks a lot for looking into this. This was indeed missing as the comments say =)
> 
> One general remark: If you open PRs for me, it is enough to create the PR for one branch, preferably the latest names branch (so release_9 in this very moment). In general I will then port the changes myself or in rare cases ask you to do so and open additional PRs.

Ah ok good to know, I will do this in future PR's :)
> 
> About this change here: Please give some love to the docu of the class to keep the current standard, e.g.: link the RFC; explain what you did or did not implement; refer to according references in the RFC.

Will be added 👍.
> 
> Thanks a bunch!

lscharmer:
@klees I updated this PR.
As I forgot about the case for embedded IPv4 addresses in an IPv6 address the regex is now much more complicated.
I **would be happy** to move the IPv6 regex to into several methods instead of the current constants. Because it is currently cumbersome to extract meaningful regex parts and reuse them in a very similar regex.
This would also have the benefit of highlighting the differences between the regexes instead of hiding them (because the differences are method arguments).

As I don't know the reason why all regexes are written as constants I don't want to change this on a whim.

The sibling PRs will be updated when this PR is reviewed.

Best regards
@lscharmer

klees:
Hi @lscharmer,

the reason for the regexps as constants basically is maintainability. Instead of having on huge regexp chunk, where we would need to find out what is what later on, the idea was that we would be able to map individual parts of the regexp to the definitions from the RFC. This might have come to an end now...

If I understand you correctly, you are basically proposing to build a top down parser based on regexps, right? Or do you want to compose regexps via methods? Is the current state the one we could merge? I see your point that this looks quite complicated, but this still might very well be the most simple form, as the structure of the problem itself is quite complicated. So, if you really think that you will gain something in presenting the regexp differently, go on, but please do not discard the current state if it doesn't turn out to well... If you want to build a top down parser I would object. In that case we should use you parser combinators imo. We might then want to do a little benchmarking regarding the speed of the combinator... Maybe contact me on short notice via Discord to discuss the approaches.

I'm sorry that it currently takes some time regarding your PRs. You have touched some issues and topics that are somewhere in the substrate of our system, so I think they deserve time and attention, which I currently don't have much. I assume that the time around new year will be a lot more quiet, so I semi-promise that I will look into your various improvements then.

Kind regards!

lscharmer:
> Hi @lscharmer,
> 
> the reason for the regexps as constants basically is maintainability. Instead of having on huge regexp chunk, where we would need to find out what is what later on, the idea was that we would be able to map individual parts of the regexp to the definitions from the RFC. This might have come to an end now...
> 
> If I understand you correctly, you are basically proposing to build a top down parser based on regexps, right? Or do you want to compose regexps via methods? Is the current state the one we could merge? I see your point that this looks quite complicated, but this still might very well be the most simple form, as the structure of the problem itself is quite complicated. So, if you really think that you will gain something in presenting the regexp differently, go on, but please do not discard the current state if it doesn't turn out to well... If you want to build a top down parser I would object. In that case we should use you parser combinators imo. We might then want to do a little benchmarking regarding the speed of the combinator... Maybe contact me on short notice via Discord to discuss the approaches.
> 
> I'm sorry that it currently takes some time regarding your PRs. You have touched some issues and topics that are somewhere in the substrate of our system, so I think they deserve time and attention, which I currently don't have much. I assume that the time around new year will be a lot more quiet, so I semi-promise that I will look into your various improvements then.
> 
> Kind regards!

Hi @klees,
I'm proposing to compose regexps via methods.
This is the working versions with constants.
I will simply create another PR which uses methods (for things like _repeating N X times_) and you can choose which one you want to adopt.

Best regards
@lscharmer

klees (approved):
Thanks @lscharmer, very nice!

klees:
I have picked this to release_9 and trunk. I did not bother to pick this to release_7 since support will cease soon.

PR #6645, opened 2023-11-21
UI: adjust 'select all' checkbox and label on a manage page.(#38183)

lukastocker:
https://mantis.ilias.de/view.php?id=38183


Amstutz:
thx a lot!

PR #6647, opened 2023-11-22
Update Guidelines_SCSS-Coding, fix typo

yvseiler:


yvseiler:
on second thought we do not need a : here :)

PR #6648, opened 2023-11-22
IRSS: Fix Write Access to Storage on Migration

kergomard:
Hi @chfsx 

This is one possible fix for:
https://mantis.ilias.de/view.php?id=38993

I think this should actually be enough, but it also happens really early and thus stops listing of migrations with users who do not have write access to storage. I do actually believe that this is the right behavior though. The check is also slightly more specific than the one in the Migration of the Files-Module as I think we should be sure to have write rights on the fsv2-folder.

Best,
@kergomard


chfsx:
Thanks a lot @kergomard , I cherry-picked to trunk as well

PR #6649, opened 2023-11-22
Node: Test and change the nodejs versions (ILIAS 9)

lscharmer:
This PR adjusts the required nodejs version in `docs/configuration/install.md` and changes the github ci builds to use the node LTS version **20**.

PR #6650, opened 2023-11-22
Node: Test and change the nodejs versions (ILIAS 10)

lscharmer:
Sister PR to #6649.


klees:
Hi @mjansenDatabay,

I hope it is ok that I have assigned you, you always care about software versions so diligently.

Thanks!

PR #6653, opened 2023-11-23
Composer: Add `simplesamlphp/simplesamlphp` as dependency

mjansenDatabay:
This PR adds `simplesamlphp/simplesamlphp` as composer dependency.

Usage:
* `components/ILIAS/Saml`

Wrapped By:
* `\ilSimpleSAMLphpWrapper` (implements `\ilSamlAuth`)
* `\ilSimpleSAMLphplIdpDiscovery` (implements `\ilSamlIdpDiscovery`)

Reasoning:
* `SimpleSAMLphp` is an award-winning application written in native PHP that deals with secure authentication. The main focus of `SimpleSAMLphp` is providing support for SAML 2.0 as both "Service Provider" and "Identity Provider".
* SAML is a very complex standard (in my personal opinion, see https://datatracker.ietf.org/doc/html/rfc7522 and linked documents) and we should not try to/do not want to reinvent the wheel in ILIAS. So we should rely on this library crafted by these authentication/security standard specialists.

Maintenance:
* `SimpleSAMLphp` is actively maintained by multiple contributors (see: https://simplesamlphp.org/contrib/). There is recent activity (even this week, see: https://github.com/simplesamlphp/simplesamlphp/commits/master).
* Security issues are always fixed in a timely manner followed by new releases. A proper security process is implemented: https://simplesamlphp.org/security/ .

Outlook:
* We are currently working on upgrading the library to the new major release (2.x) to receive all future (security) bugfixes and to be compliant with future PHP versions.

Links:
* Packagist: https://packagist.org/packages/simplesamlphp/simplesamlphp
* GitHub: https://github.com/simplesamlphp/simplesamlphp
* Documentation: https://simplesamlphp.org/docs/

### DIFF
--- a/composer_new.json
+++ b/composer_new.json
@@ -44,6 +44,7 @@
 		"ext-xml": "*",
 		"ext-zip": "*",
 		"ext-imagick": "*",
+		"simplesamlphp/simplesamlphp": "^1.19.8",
 	},
 	"require-dev": {
 	},


matthiaskunkel:

chfsx:
see #6725 for the major release update of this library

lukastocker:
Hello @mjansenDatabay ,

I am currently creating an PR via cherry-pick with all dependencies approved in the JF in december 2023. Sadly I can't include your composer simplesamlphp PR because the included package doesn't work with the approved symfony/console v.7.0 package. As far as I check correctly the linked update in https://github.com/ILIAS-eLearning/ILIAS/pull/6725 won't work with symfony/console v7.0 too. We do need a decision how to proceed.

Kind regards
Luka

mjansenDatabay:
Hi @lukastocker ,

thanks for the analysis. You are right, version `7.x` of `symfony/console` is not supported by the latest/current `simplesamlphp` package, and will be most probably not supported by the next major release as well (see link below).

The question is: Why do we need to upgrade `symfony/console` to version `7.x`? I guess to support PHP 8.3 ?!?
Did somebody check if v. `6.4` would possibly be sufficient for PHP 8.2/8.3?

If v. `6.4` ist **not** sufficient/possible, I suggest to continue as follows:

1. You ignore this PR (should be closed anyway as far as I understood our process). You can continue your work on the `composer.json` file.
2. PR #6725 can also be ignored for now.
3. If there is no version that supports version `7.x` of `symfony/console` later this year (I doubt that, see `dev-master`: https://packagist.org/packages/simplesamlphp/simplesamlphp#dev-master), there are several options:
  * We abandon `Services/SAML` for ILIAS 10.x
  * We find funding to evaluate other libraries and apply the relevant changes to #6725 
  * We fork the latest version into the ILIAS GitHub organization and remove the unnecessary dependencies.
  * We find another way to include this library, because the `symfony/console` is NOT USED AT ALL in the `SAML` context of ILIAS, so it is NOT a "Runtime Problem", but a "Build Problem".

Best regards,
Michael

lukastocker:
Hi @mjansenDatabay ,

I tried to install different versions of symfony/console and simplesamlphp with following results:

- simplesamlphp v1.19.8 needs symfony ^4.4 || ^5.0  - v6 is not supported
- simplesamlphp v2.0.0 needs symfony v5 max - v6 is not supported
- simplesamlphp v2.1.0 works with symfony/console v6 

How shall we proceed?

Symfony/console v6.4 should work with PHP8.2 and up as you can see here: https://github.com/symfony/console/blob/7.0/composer.json

Kind regards 

Luka

mjansenDatabay:
Hi @lukastocker ,

thanks for taking care. Let's use `^2.2` (see: https://packagist.org/packages/simplesamlphp/simplesamlphp#v2.2.2), which also supports `symfony/console 6.4`. Minor release updates without breaking changes should be fine for the SAML component in ILIAS. @lscharmer  and I will take care if there are any SAML-related problems.

Best regards,
Michael

klees:

Kind regards!

PR #6659, opened 2023-11-23
Certificate: Use `file://` schema to make background images work with…

mjansenDatabay:
… new `ilServer`

See: https://mantis.ilias.de/view.php?id=38729


smeyer-ilias:
Thanks, found the same here [1] (uri-specification) https://www.w3.org/TR/xsl11/#datatype
Could this be merged to 7 and 8 to keep the option to use the java server for older releases?

PR #6663, opened 2023-11-24
UI: remove padding in infoscreen value class to align texts in rows. …

lukastocker:
…(#39020)

https://mantis.ilias.de/view.php?id=39020


lukastocker:
Hi @Amstutz ,

the removed padding is not needed in that scenario. I am not sure if there's a scenario where the class is used too and a padding is needed. In that case we probably need to add a parent class to the specific scenarios. If you find some, please let me know.

Kind regards
Luka

PR #6664, opened 2023-11-24
#39062 prevent 403 on loginpage

sroenchen:
Added a check for a valid user login, to prevent a failing access check on the login page, due to the HelpGUI

[#39062](https://mantis.ilias.de/view.php?id=39062)


sroenchen:
PR closed due to misdiagnosing the issue. 
The actual problem being helpGUI not being shown while not being pseudo-loggedin as anonymous (as described in https://mantis.ilias.de/view.php?id=39062#c99873 ) is in not remedied by the submitted commit.

PR #6665, opened 2023-11-24
Implementation of new Integration Mechanism for Components

klees:
Dear colleagues,

this implements the [integration mechanism as agreed upon for the Component Revision](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/components-and-directories.md#types-of-dependencies-and-integration-strategies).

This is here for the general visibility and for the curious. I will continue to work on this. Next step will be:

* Implement an actual bootstrap using the facilities from this PR.
* Use the mechanism to reimplement the Setup.

I will leave this PR open for some days and merge it when I'm satisfied with the current state. Please feel free to look into this, ask questions and question my design choices. This will be the base on which other components will stand and we should absolutely make sure that inner workings and general functionality are known and understood by many people.

Kind regards!

Addendum 1: This fails 8.1-tests because I currently use an 8.2-only construct. I expect ILIAS 10 to use PHP >= 8.2, hence I already removed 8.1 in this PR. I will remove the according commit before merge.


klees:
Hi Jour-Fixe,

there are three things in here that would need you acknowledgment:

* Initially there has been the possibility to explicitly define a Null Object for services. Due to advice from @pascalseeland I removed that possibility and made dependency resolution more explicit then I originally planned.
* In my initial paper, components would have had a definition in a file that is named according to the component itself. E.g. `AccessControl.php` for `ILIAS\AccessControl`. I changed this to a generic filename `Component.php` for every component. The class is still named according to the component.
* In aggreement with @kergomard  as maintainer, I renamed the component `ILIAS\Object` to `ILIAS\ILIASObject`. `Object` is a reserved word in PHP and hence not well suited for identifiers.

Kind regards!

matthiaskunkel:
Jour Fixe, 27 NOV 2023: Richard notified us about three topics: the removal of the Null Object, the changed naming convention and the relabeling of the Object service to ILIASObject. We fully agree with the removal of the Null Object and the relabeling of the Object service. But there is no clear tendency in favour of the naming convention. Therefore, Richard will make a short poll to get a decision about how classes and files are labeled.

klees:
Hi everyone,

the poll is here: https://nuudel.digitalcourage.de/mHtCjp9HquMCsUoU

Kind regards!

klees:
Hello everyone!

The poll had a super small minority (11 to 10 votes) for the $COMPONENT.php naming (over `Component.php`). We still changed accordingly.

In the current state the setup is partly moved to the new mechanism.

I would kindly ask @katringross to have a look into the changes in language. I merged the "components" `ILIAS/Language` and `ILIAS/Language_` and introduced an interface `ILIAS\Language\Language` to introduce the core service of the component.

I would also ask @mjansenDatabay to have a look into this, as I started to wire up `ILIAS/Refinery` and `ILIAS/Data` according to the component revision.

If these two agree I would merge the current state and the go on. Next to tasks would be:

* Completely move the setup to the mechanism of the component revision which basically means removing the `ImplementationOfInterfaceFinder`.
* Creating infra to build different bootstraps, which would basically replace `ILIAS/Context`.

Kind regards!

mjansenDatabay (approved):
The refinery changes LGTM!

katringross (approved):
Looks fine, thanks @klees!

klees:
Thx both!

PR #6667, opened 2023-11-24
Release 9: Chatroom, Forum, ContentPage xml import validation

mBeym:
added Validation of the uploaded xml file for the objects:
- Chatroom
- Forum
- ContentPage

=> https://docu.ilias.de/goto_docu_wiki_wpage_6315_1357.html


Modules/Forum/classes/class.ilForumXMLParser.php
```diff
@@ -183,10 +183,6 @@ public function handlerEndTag(XMLParser $a_xml_parser, string $a_name): void
                 $propertyValue['Id'] = $this->cdata;
                 break;
 
-            case 'StyleId':
```
mjansenDatabay, line 186: Are you sure this can be removed? Have you used a content style for the forum page editor content?

xml/SchemaValidation/ilias_copa_9.xsd
```diff
@@ -0,0 +1,31 @@
...
+                <element ref="t:description" minOccurs="1" maxOccurs="1"/>
+                <element ref="t:info-tab" minOccurs="1" maxOccurs="1"/>
+            </sequence>
+        </complexType>
+    </element>
+
+    <element name="id" type="t:Data" />
+    <element name="title" type="t:Data" />
+    <element name="description" type="t:Data" />
+    <element name="info-tab" type="t:Data" />
```
mjansenDatabay, line 30: You removed the `style-id` Are you sure this is not exported anymore?

xml/SchemaValidation/ilias_frm_9.xsd
```diff
@@ -0,0 +1,107 @@
...
+<xs:schema
+		xmlns:xs="http://www.w3.org/2001/XMLSchema"
+		elementFormDefault="qualified">
+
+	<xs:element name="Forum">
+		<xs:complexType>
+			<xs:sequence>
+				<xs:element name="Id" type="xs:integer" />
+				<xs:element name="ObjId" type="xs:integer" />
+				<xs:element name="Title" type="xs:string" />
```
mjansenDatabay, line 11: You removed the `StyleId` Are you sure this is not exported anymore?

mjansenDatabay (commented):
Thx  @mBeym

xml/SchemaValidation/ilias_copa_9.xsd
mBeym: it was at least never included in the exports I created.

I could also not find it anywhere else in the code except some other Modules/Services that still had it in the switch case.

mjansenDatabay: If it is no more part of the XML exporter, it is not used anymore.

But as far as I can see, the `Services/Style` component is still used a sa `Tail Dependecy` (see: `\ilContentPageExporter::getXmlExportTailDependencies`), so it might be included in the export and parsed during the import.

mBeym: but don't those end up in their own folder in the exported archive?

From **manifest.xml**:
```<?xml version="1.0" encoding="utf-8"?><!--Generated by ILIAS XmlWriter--><Manifest MainEntity="copa" Title="bgggg" InstallationId="0" InstallationUrl="http://ilias8"><ExportFile Component="Modules/ContentPage" Path="Modules/ContentPage/set_1/export.xml"/><ExportFile Component="Services/MediaObjects" Path="Services/MediaObjects/set_1/export.xml"/><ExportFile Component="Services/MetaData" Path="Services/MetaData/set_1/export.xml"/><ExportFile Component="Modules/File" Path="Modules/File/set_1/export.xml"/><ExportFile Component="Services/MetaData" Path="Services/MetaData/set_2/export.xml"/><ExportFile Component="Services/Object" Path="Services/Object/set_1/export.xml"/><ExportFile Component="Services/COPage" Path="Services/COPage/set_1/export.xml"/><ExportFile Component="Services/Style" Path="Services/Style/set_1/export.xml"/><ExportFile Component="Services/Object" Path="Services/Object/set_2/export.xml"/></Manifest>```

Comment (mBeym) [reply xml/SchemaValidation/ilias_copa_9.xsd:30]: I can get a different component (LearningModule) to include the StyleId in the main export.xml

but neither Chatroom, Forum nor ContentPage export that value for me.
And the LearningModule classes had some other mentions of "StyleId" in the code.

Comment (mjansenDatabay) [reply xml/SchemaValidation/ilias_copa_9.xsd:30]: Sounds good :-).

**Review by mjansenDatabay — CHANGES_REQUESTED** [review]
@mBeym 

There is a new documentation available, see: https://github.com/ILIAS-eLearning/ILIAS/blob/release_9/Services/Export/README-technical.md

I had a chat with @smeyer-ilias about the export/import and got an idea how this is intended to work.

- [x] Please add the following schema files to support ILIAS 8 and ILIAS 7 (they have to be verified first of course, if they are correct) exports as well

* ilias_copa_5_4.xsd (should be used by ILIAS 7 and 8 as well)
* ilias_chtr_5_3.xsd (should be used by ILIAS 7 and 8 as well)
* ilias_frm_5_1.xsd (should be used in ILIAS 7 an 8 as well)

- [x] Please delete all other `frm`, `copa` and `chtr` schema files (AFAIK there are only`frm` files to be deleted)

This means: In ILIAS 9, we will support imports from ILIAS 7, 8, and 9. I would like to drop the support for imports from ILIAS <= 7.

- [x] Please remove the entries from `\il[TYPE]Exporter::getValidSchemaVersions` we do **not** want to support anymore

For the imports, the `Schema Version` can be fetched in the `\il[TYPE]Importer` classes by calling `$this->getSchemaVersion()`, so the XML parsing can vary depending on given/imported schema (e.g. to handle mandatory attributes/elements differently depending on the import file).

File: Modules/Chatroom/classes/class.ilChatroomExporter.php [hunk 4]
```diff
@@ -66,7 +66,14 @@ public function getValidSchemaVersions(string $a_entity): array
...
                 'min' => '5.3.0',
-                'max' => ''
+                'max' => '8.999'
+            ],
+            '9.0' => [
+                'namespace' => 'https://www.ilias.de/Modules/Chatroom/chtr/9',
+                'xsd_file' => 'ilias_chtr_9.xsd',
+                'uses_dataset' => false,
+                'min' => '9.0',
+                'max' => '9.999'
```
mjansenDatabay: Please set the `max` value to `''`.

mjansenDatabay (changes requested):

Modules/Chatroom/classes/class.ilChatroomExporter.php
mBeym: done

mjansenDatabay (approved):

PR #6668, opened 2023-11-24
Composer: Add `phpstan/phpstan` as dependency

mjansenDatabay:
This PR adds `phpstan/phpstan` as composer dependency.

Usage:
* `PHPStan` finds issues in the ILIAS code without writing tests. It detects architectural smells in our application based on rule definitions (see: scripts/PHPStan/Rules). It is used in the GitHub CI pipeline (see: .github/workflows/legacy-ui.yml). `PHPStan` makes use of the `AST` (abstract syntax tree) and is capable to analyze our ILIAS PHP code statically.

Wrapped By:
* Not applicable

Reasoning:
* `PHPStan` is one of the two or three industry standards (beside `psaml` and `phan`) to statically analyze PHP code. In ILIAS it is/can be used as a PHP alternative to `dicto`. `PHPStan` is also implicitly used by other libraries (like `rector`). Providing a toolset to enable rule based static analyses is a hard task (in time/costs for the maintenance), so we should not try to reinvent the wheel here by trying to do this analyses ourselves.

Maintenance:
* `PHPStan` is a very active project (see: https://github.com/phpstan/phpstan/graphs/commit-activity), code was committed even this week (see: https://github.com/phpstan/phpstan/commits/1.11.x). There is one main contributor (ondrejmirtes) supported by other open source enthusiasts (and financially by some well known organizations).
* The risk of relying on this library is relatively small. It is a development dependency and our production code does not rely on this library. But: The project mainly relies on one contributor and there might be funding issues in future (a commercial plan has been indroduced, see: https://phpstan.org/blog/introducing-phpstan-pro).

Outlook:
* The existing UI rules could be further extended by  https://github.com/mjansenDatabay/ILIAS/commit/5e4dfeaf4a471de51712681f078758ed9f44b26c / https://github.com/mjansenDatabay/ILIAS/commit/a0907099d64080d9c0959436304d2b5a536866cd

Links:
* Packagist: https://packagist.org/packages/phpstan/phpstan
* GitHub: https://github.com/phpstan/phpstan
* Documentation: https://phpstan.org/user-guide/getting-started

### DIFF
--- a/composer_new.json
+++ b/composer_new.json
@@ -46,6 +46,7 @@
 		"ext-imagick": "*",
 	},
 	"require-dev": {
+		"phpstan/phpstan": "^1.10",
 	},
 	"autoload": {
 		"psr-4" : {


matthiaskunkel:

klees:

Kind regards!

PR #6669, opened 2023-11-24
Fix #37778: Fix for SOAP Plugin Reinitialization Issue in ILIAS 8

jeph864:
This PR addresses a bug identified in ILIAS 8, specifically concerning the malfunctioning of SOAP plugins. The issue originated from a series of interconnected bugs:

- https://mantis.ilias.de/view.php?id=37778
- https://mantis.ilias.de/view.php?id=37781
- https://mantis.ilias.de/view.php?id=37780

 They  can all be  traced back to the  reinitialization performed by the SOAP administration. During the initial Initialisation, certain superglobals are replaced within the system. However, when reinitialization occurs, ILIAS attempts to replace these superglobals again, leading to a cascade of errors.  This PR solves this issue but only considers SOAP contexts.

### DIFF
--- a/Services/Init/classes/class.ilInitialisation.php
+++ b/Services/Init/classes/class.ilInitialisation.php
@@ -1528,6 +1528,9 @@ class ilInitialisation
         );
 
         if ($replace_super_globals) {
+            if(ilContext::getType() == ilContext::CONTEXT_SOAP_NO_AUTH){
+                return;
+            }
             $throwOnValueAssignment = defined('DEVMODE') && DEVMODE;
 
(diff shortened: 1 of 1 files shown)


jeph864:
@pascalseeland is there any update on this PR?

pascalseeland:
I would prefer to merge #7116 which, together with disabling the super global replacement should fix this issue instead of making the replacement code depend on the context. We should probably include some documentation in #7116 PR that if someone needs SOAP one has to disable it.  Or we could add the documentation to the UI where one enables SOAP.

mjansenDatabay:
I'll close this PR since the issue seems to be solved in the meantime.

PR #6671, opened 2023-11-27
DataCollection - Release 7: Fix new lines in written text causing mail to be malformed

mBeym:
Fixes https://mantis.ilias.de/view.php?id=36630


Amstutz:
Thx a lot for this PR!

Nice work in code cleanup with the statics! Highly appreciated. Also thx for adding the Unit Test!

Would it be possible in this case to provide a PR for 8 and 9 as well? Ill pick to trunk.

Kudos and thx.

Amstutz:
Note, tests are fine, something with language sort seems of in 7. Merged.

PR #6672, opened 2023-11-27
T&A Bugfix #0037248: Wrong cmdNode by using the action menu

fhelfer:
https://mantis.ilias.de/view.php?id=37248


fhelfer:
Fixes https://mantis.ilias.de/view.php?id=39124 as well

kergomard:
Thank you very much for the PR @fhelfer ! Let's go with this. I usually don't like these redirects all over the place, they just seem plain wrong, but here I think it is the right move. Merged and picked to ILIAS 8, ILIAS 9, and trunk.

PR #6674, opened 2023-11-27
InfoScreen: use LOM API (38859, 38868)

schmitz-ilias:
This PR fixes [38868](https://mantis.ilias.de/view.php?id=38868) and [38859](https://mantis.ilias.de/view.php?id=38859) by using the new LOM API instead of the old one.


alex40724:
Thanks for the PR!

PR #6676, opened 2023-11-27
LUI: fix check #38699

klees:
Hi @chfsx,

these should be the missing changes for the LUI scripts in trunk.

Kind regards!


chfsx:
I do not get any result if I run the script locally (all the results files are empty or do not have any violations in). I try to find out why :-)

PR #6677, opened 2023-11-27
fix 0038490: Failed test: MathJax im Zertifikat anzeigen

fneumann:
Added MathJax processing to PDF generation for real certificates, not only in preview


mjansenDatabay:
Thx @fneumann

PR #6679, opened 2023-11-28
Bugfix: 0039089: Forum migration "forum.ilForumDraftsFilesMigration"

playamongthestars:


mjansenDatabay (changes requested):
Thx @nmatuschek , I have a few questions.

1. You use `scandir`. This function works recursively and might be a performance issue. Is this really necessary?
2. If we do not find a valid directory, shouldn't we set the `rcid` to `-`?

Furthermore:
- [x] Please remove the nested `if` statement and use `&&` in a single `if` instead.

Thx @mjansenDatabay

playamongthestars:
No. If the directory doesn't exist, it's no difference between "NULL" and "-" for the RCID-Value. So we can avoid this unneccessary 
step to update the rcid for every post to "-" in the db.
It is also possible, that an directory exists but contains no files. So there is the need to split the if-statements. 
scandir returns false if the directory contains no files. And count() on false ends in a fatal error. 

Hard to believe that this single migration step could cause serious performance issues. Maybe the migration-step takes a little more time, but only if it is a real big forum and only if it contains a lot of files. 
I am not sure what's better for the performance. 
Use scandir and skip if the directory is empty and don't let the stakeholder iterate over an empty directory? 
or: You could also feed the stakeholder with an empty array, but the stakeholder would (and I think that's worse) iterate with a foreach over the directory and returns a NULL-Value in this case anyway.

If you have better ideas, please let me know. :-)

mjansenDatabay:
* If you use `&&`, the second operation is not executed if the first (`is_dir(...)`) evaluates to a boolean `false`.
* There is indeed a difference between `-` and (`NULL` or ``), please have a look at `\ilForumDraftsFilesMigration::step` and `\ilForumDraftsFilesMigration::getRemainingAmountOfSteps`. So we MUST use a `-` in my opinion.

PR #6681, opened 2023-11-28
LSO: add missing lang-vars for editor

nhaagen:
(and remove duplicate entry dateplaner#:#cal_ical_url_google#:#iCal-URL)

https://mantis.ilias.de/view.php?id=39131


klees:

PR #6687, opened 2023-11-29
UI: adjust presentation table layout. (#37754, #38126)

lukastocker:
https://mantis.ilias.de/view.php?id=38126
https://mantis.ilias.de/view.php?id=37754

Following issues were fixed =>

1. Correction of missing ul html in example
2. Setting cols and rows for presentation table layout in the template to get correct cols and widths
3. Deletion of margins and widths in children and also some paddings
4. Set order of elements depending on screen size
5. Moving 'on-screen-size' mixin to tool 'layout_breakpoint' 
6. Remove 'margin-bottom' from highlighted_box' tool as it shouldn't be set there
7. Adjust presentation table tests

Old layout desktop:
pres-table-old-desktop

New layout desktop:
pres-table-new-desktop

Old layout mobile:
pres-table-old-mobile

New layout mobile:
pres-table-new-mobile-2
pres-table-new-mobile

This implementation was discussed with @catenglaender . We decided to solve both mantis tickets in one PR to avoid partly duplicated implementations.

### DIFF
--- a/src/UI/examples/Table/Presentation/base.php
+++ b/src/UI/examples/Table/Presentation/base.php
@@ -76,7 +76,7 @@ function included_data()
             'bookings_available' => '3',
             'target_group' => 'Employees, Field Service',
             'goals' => 'Lorem Ipsum....',
-            'topics' => '<li>Tranportations</li><li>Europapolice</li>',
+            'topics' => '<ul><li>Tranportations</li><li>Europapolice</li></ul>',
             'date' => '30.09.2017 - 02.10.2017',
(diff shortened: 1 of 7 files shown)


catenglaender:
Finally! The slightly rough edges of the presentation table have been bothering me for years. This is much more balanced and stable. Also, I love the on-screen-size layout mixin.

Amstutz:
Hi @lukastocker 

Thx a lot. I like the mixin as well. special thx for the housekeeping and cleaning up things. Nice to see that we are further losing CSS LOC.

PR #6688, opened 2023-11-29
UI: adjust position and width of column(s) in bulk action of the book…

lukastocker:
…ing pool modal. (#38840)

https://mantis.ilias.de/view.php?id=38840
https://mantis.ilias.de/view.php?id=38921

Old layout:
bulk-style-modal-booking-pool

New layout:
booking-pool_bulk-modal-new

Notice: The labels should not be full width as suggested first as they are part of the legacy ui and therefore in a column directly beside the input field. The issue was a false display setting in the "old" layout. Now the label + input fields are positioned correctly. 

This implementation was discussed with @catenglaender today.


Amstutz:
Thx a lot!

PR #6689, opened 2023-11-30
T&A Bugfix #0038431: Subset Question Answer gets lost

fhelfer:
https://mantis.ilias.de/view.php?id=38431


dsstrassner:
@kergomard or @mbecker-databay, could someone of you two, please review and merge this before Christmas?

kergomard:
Thank you very much @fhelfer for this PR!

I close it though and fix it on the other end: Instead of counting the answers up when setting them and again when reading them, I think it would be more economical to keep a 0-based index.

My change is [here](https://github.com/ILIAS-eLearning/ILIAS/commit/1f805b7a12d70326351a7a33c9227b63afd871a1)

Best,
@kergomard

PR #6696, opened 2023-11-30
T&A Bugfix #0039154: Failed test: Editing Question not syncing with original

fhelfer:
https://mantis.ilias.de/view.php?id=39154

These lines got deleted with https://github.com/fhelfer/ILIAS/commit/d71562733c1e339109a9e024065844d11474a806 which results in the orginal question not updating. @kergomard maybe you can look after why you made these changes in the first place.

Best, @fhelfer

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assQuestion.php
+++ b/Modules/TestQuestionPool/classes/class.assQuestion.php
@@ -1608,6 +1608,9 @@ abstract class assQuestion
             return; // No original -> no sync
         }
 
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much for the PR @fhelfer ! You pose a very good question, I don't have an answer to, but I would think that this was an accident. I have already put the lines back in a little bit differently (see: https://github.com/ILIAS-eLearning/ILIAS/commit/2dc5c17). Sorry!

PR #6697, opened 2023-12-01
UI: add margin-top to viewcontrol mode. (#38581)

lukastocker:
https://mantis.ilias.de/view.php?id=38581

### DIFF
--- a/templates/default/070-components/legacy/Services/_component_copage.scss
+++ b/templates/default/070-components/legacy/Services/_component_copage.scss
@@ -774,6 +774,10 @@ div.ilCOPGMediaPrint {
 	padding: 0 $il-padding-xlarge-horizontal;
 }
 
+#copg-table-top-actions .il-viewcontrol-mode {
+	margin-top: $il-margin-xxlarge-vertical;
+}
+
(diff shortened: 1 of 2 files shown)


catenglaender:
Maybe the code might have less side effects at another place. Generally, components should not set a margin above or below them. It's better if the surrounding component/view would do that, so the margin can be different in different contexts.

How about this in /070-components/legacy/Services/_component_copage.scss:
```scss
#copg-top-actions .il-viewcontrol-mode {
    margin-top: $il-margin-xxlarge-vertical;
}
```

lukastocker:
> Maybe the code might have less side effects at another place. Generally, components should not set a margin above or below them. It's better if the surrounding component/view would do that, so the margin can be different in different contexts.
> 
> How about this in /070-components/legacy/Services/_component_copage.scss:
> 
> ```css-scss
> #copg-top-actions .il-viewcontrol-mode {
>     margin-top: $il-margin-xxlarge-vertical;
> }
> ```

Hi @catenglaender ,

thanks for the review. I absolutely agree with your suggestion and therefore adjusted the PR accordingly.

PR #6698, opened 2023-12-01
UI: adjust position of glyphicon in ViewControl's pagination. (#38964)

lukastocker:
https://mantis.ilias.de/view.php?id=38964


catenglaender:
Oh these glyphs keep acting up in view controls. Long term, I hope that we can get rid of most of the code overriding glyphs standard styling in components and have glyphs just work in every context with no adjustments.

Be aware that there are two view controls pagination. One in View Controls Pagination and one in Input View Controls Pagination. Sometimes you fix one, which makes the other one break.

PR #6699, opened 2023-12-01
Maps: 38720, rollup as iife

nhaagen:
https://mantis.ilias.de/view.php?id=38720


klees:
@nhaagen Please prepare a similar PR for trunk. Thanks!

PR #6701, opened 2023-12-01
UI/DataTable: 38503, add a session storage to the table

nhaagen:
https://mantis.ilias.de/view.php?id=38503

### DIFF
--- a/Services/Init/classes/Dependencies/InitUIFramework.php
+++ b/Services/Init/classes/Dependencies/InitUIFramework.php
@@ -145,7 +145,8 @@ class InitUIFramework
                 $c["ui.data_factory"],
                 $c["ui.factory.table.column"],
                 $c["ui.factory.table.action"],
-                $row_builder
+                $row_builder,
+                $c["ui.session"]
             );
         };
         $c["ui.factory.table.column"] = function ($c) {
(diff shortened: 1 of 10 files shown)


oliversamoila:
Hey @nhaagen ,
Thank you very much.
Can you explain in just a few sentences what comes with the PR?
Selected columns, selection of rows, selected sorting, customised column widths, last called page, or other? Does this also apply if I log in again days later?

nhaagen:
With this PR, DataTable stores all viewcontrol-parameters (Pagination, Ordering, FieldSelection) in the session.
Session is - as far as I know - destroyed on logout.
However, since the storage is injected, one could easily build a more permanent storage.

oliversamoila:
Thank you very much, @nhaagen.

klees (changes requested):
Hi @nhaagen,

this mostly LGTM and I will pass on to @Amstutz and @thibsy soon.

Please make some minor improvements before:

* [x] Docu: Please add some documentation about the behaviour in the KS entry of the table.
* [x] Refactoring: Please factor out that storage reading sequence in `Table\Data::applyViewControls`. Write a unit test for the logic.
* [x] id-Prefix: Please add some prefix to the id (in `Table\Data::getStorage`, probably) so we do not pollute the namespace of the session. Maybe we should simply use the FQN of the class... 
* [x] Unit Test: Please write an integration test for the complete functionality.
* [ ] Random Question: Can we in fact write arrays to the session?

Thanks!

klees:
Dear @Amstutz and @thibsy, please carry this on. Thanks!

thibsy (approved):
Thx @nhaagen for looking into the changes I have made. Since I am a co-contributor now I will pass this on to @Amstutz for a final review, afterwards this should be good to go.

oliversamoila:
Hello @nhaagen, 
I have another question: is this something that now works for all previous table migrations or must IDs be provided everywhere as with the question table in https://github.com/ILIAS-eLearning/ILIAS/pull/7272?
Thank you in advance.

klees:
Hi @oliversamoila,

the tables that want to use this need to have an id, so as long as they are not provided now they need to be added.

Kind regards!

PR #6702, opened 2023-12-01
UI/DataTable: 39206, generate translated ordering options by columntype

nhaagen:
https://mantis.ilias.de/view.php?id=39206


klees (changes requested):
Hi @nhaagen,

this mostly LGTM and I will pass on to @Amstutz and @thibsy for a formal review once we made small improvements:

* [x] Custom Labels: Currently, the name of the column is always prefixed, even if consumers use custom labels. Since they should now how the column is named, we should give them complete freedom in naming IMO. This would allow for, e.g., having a column named "Prize" but have the sortation named "Cheapest first".
* [x] Unit Test: Write some for `OrderOptionsBuilder`. 
* [x] Interface: We should detach the labels from the flag IMO. I.e.: `withOrderingLabels(        string $asc_label = null, string $desc_label = null)`. Two reasons: I wondered if labels would be discarded when isSortable is set to `false`. Separating the thingies would clarify that. Also, a little more symmetry to `getOrderingLabels` should help people to understand the usage of the thingy.

Thanks!

nhaagen:
> [ ] Custom Labels: Currently, the name of the column is always prefixed, even if consumers use custom labels. Since they should now how the column is named, we should give them complete freedom in naming IMO. This would allow for, e.g., having a column named "Prize" but have the sortation named "Cheapest first".

I cannot find the paper atm, but I vaguely remember that the column name should be stated; anyways, the ordering targets one col only, dismissing the name might lead to the idea of more complex sorting-methods.

klees:
@Amstutz/@thibsy: Please continue =) Thanks!

thibsy (changes requested):
Hi @nhaagen

Thanks a lot for your contribution to the UI components and sorry for the delay here!

- [ ] Unicode character: you are using [a unicode character](https://www.compart.com/en/unicode/U+2192) for some translations. Is this A11Y friendly? If not, we should probably stick with "to" or similar.

- [x] `OrderOptionsBuilder::getLabelsByColumn()` I: the `switch` statement uses magic strings. This is because `Column::getType()` does not return some constant value we can use. Since this has nothing to do with the actual issue and leads to additional refactoring, this is highly optional, but you could improve this by returning the fully qualified class name there.

- [x] `OrderOptionsBuilder::buildFor()` I: please move the separator for `$sort_options` keys into a constant.
- [x] `OrderOptionsBuilder::buildFor()` II: please use `Data\Order::ASC` or `::DESC` instead of strings when calling `Data\Factory::order()`.
- [x] `OrderOptionsBuilder::getLabelsByColumn()` II: please don't build language variables dynamically. This makes it quite hard to track down which variable is being used in what scenario. Please assign the entire value of the language variable directly and maybe move this value into a constant as well.
- [x] `OrderOptionsBuilder::getLabelsByColumn()` III: please improve this method by getting rid of the `$asc` and `$desc` variables and returning the proper values directly in the `switch` statement. This should simplify the `return` statement a lot.
- [x] Conflicts: please rebase this onto the latest release_9 branch to resolve the conflicts.

Kind regards,
@Amstutz and @thibsy

nhaagen:
> [ ] Unicode character: you are using [a unicode character](https://www.compart.com/en/unicode/U+2192) for some translations. Is this A11Y friendly? If not, we should probably stick with "to" or similar.

The char is read as "right arrow" - I figure, that is quite understandable, but then, of course, I'm not the one to tell.

thibsy (changes requested):
Hi @nhaagen

Thanks for implementing the changes so quickly! There are some minor things left to do before we can merge this.

- [x] `OrderOptionsBuilder::getLabelsByColumn()` I: IMO you have added a bandaid to the problem here, by dynamically determining the interface. Please revert this change or address the issue at its root, the implementation of `Column::getType()`.
- [x] Unicode character: thx for looking into this. @Amstutz and me think that it would be more intuitive if the ordering is e.g. read as "A to Z", instead of "A right arrow Z". Please replace the unicode character accordingly.

Kind regards,
@Amstutz and @thibsy

nhaagen:
[2fa12a8](https://github.com/ILIAS-eLearning/ILIAS/pull/6702/commits/2fa12a8c10e02454c84aaa8ec2ab0d97e77bf4bd) places a LabelBuilder on the columns; 
[bcfe53d](https://github.com/ILIAS-eLearning/ILIAS/pull/6702/commits/bcfe53d7e0db9d80d5184ea879a0ab13c99d906e) injects ilLanguage, so columns can create their labels themselves.
Frankly, I'd still prefer a builder on the table as in [d80beb1](https://github.com/ILIAS-eLearning/ILIAS/pull/6702/commits/d80beb11a919926ff834d5a214c1281651ed8213), admittably with alterations.

Please point me in to the preferred direction to go, I'll amend and rectify the tests, then.

thibsy:
Thx @nhaagen for providing multiple solutions!

@Amstutz and I think that bcfe53d7e0db9d80d5184ea879a0ab13c99d906e (`ilLanguage` approach) and d80beb11a919926ff834d5a214c1281651ed8213 (`OrderOptionsBuilder` approach) are possible solutions. We have a slight preference for the `ilLanguage` approach, mainly because this keeps the logic in one place and makes new column implementations easier.

Therefore I kindly ask you to pursue the currently committed solution and adjust the unit tests accordingly.

Thanks a lot and kind regards,
@thibsy

thibsy (changes requested):
Hi @nhaagen

Thx a lot for finalising the PR! This needs some last polishing, afterwards this is good to be merged.

- [x] `Boolean::getOrderingLabels()`: please use `Symbol` in the `instanceof` checks, since this is the common interface of `Icon` and `Glyph`, which provides the `getLabel()` method you are using.
- [x] Unit tests: you are using several anonymous classes which directly extend form `ilLanguage`. Please use mocks instead, otherwise we couple the unit-tests to the logic of the actual implementation. You can also use [`willReturn()`](https://docs.phpunit.de/en/11.0/test-doubles.html#willreturn) to configure `ilLanguage::txt()` return values, or intercept argument values using anonymous functions.
- [x] Language files: the workflow seems to fail due to unsorted language variables. Please run the [sort script](https://github.com/ILIAS-eLearning/ILIAS/blob/release_9/CI/sort_langfile_entries.sh) to fix this.

Thanks for bearing with us!

Kind regards,
@thibsy

thibsy (approved):
Thx @nhaagen for implementing the changes!

This looks good to go now. I will merge and pick this for trunk, and add the Jour Fixe label to announce the small interface change at the upcoming JF.

Kind regards,
@thibsy

klees:
@nhaagen Please add this to the "Data Table for ILIAS 8"-PR. Thanks!

matthiaskunkel:
Jour Fixe, 18 MAR 2024: We highly appreciate this PR and accept it for Release 9 and trunk.

PR #6704, opened 2023-12-04
T&A Bugfix: ilObjTestSettingsMainGUI missing parameter

fhelfer:


fhelfer:
@matheuszych FYI

kergomard:
Thank you very much @fhelfer ! There is more wrong than this, but I will fix it myself:
We are now passing `ilTabsGUI` in the constructor, but still then taking it from the `$DIC` ;-)

Thanks again and best,
@kergomard

PR #6705, opened 2023-12-04
(RemovingLUI) T&A: replace Toolbar with dropdown in QPL Printview

fhelfer:
FW: https://docu.ilias.de/goto_docu_wiki_wpage_7975_1357.html


oliversamoila:
Thanks @fhelfer 
Here is the specific request for removing the forms into toolbars for this case: https://docu.ilias.de/goto_docu_wiki_wpage_7975_1357.html

kergomard:
Thank you very much for the PR @fhelfer ! I backported this to release_9, too. I think this is ok.

PR #6711, opened 2023-12-05
UI: align input field's footer save button with header save button. (…

lukastocker:
…#38121)

https://mantis.ilias.de/view.php?id=38121


Amstutz:
thx a lot!

PR #6712, opened 2023-12-05
FORM: remove btn-sm class from property_form template. (#38189)

lukastocker:
https://mantis.ilias.de/view.php?id=38189


Amstutz:
thx a lot!

PR #6715, opened 2023-12-05
Init: 38903, check for parameter

nhaagen:
https://mantis.ilias.de/view.php?id=38903


mjansenDatabay:
Thx @nhaagen / @pascalseeland I'll merge this because this is trivial.

PR #6716, opened 2023-12-06
UI: change height of Subform to avoid overlapping items and add missi…

lukastocker:
…ng 'row' class to property form template. (#39111, #39281)

https://mantis.ilias.de/view.php?id=39281
https://mantis.ilias.de/view.php?id=39111


Amstutz:
thx a lot!

PR #6717, opened 2023-12-06
Init: 39263, re-init refinery with user to pass correct language setings

nhaagen:
https://mantis.ilias.de/view.php?id=39263


klees:
Hi @pascalseeland,

this is about `Services\Init`, so I pass this on to you. I'm not super happy with this solution, but I think this needs to do for now. I guess we will be able to do better with improved initialisation once component revision fully hits.

Kind regards!

PR #6718, opened 2023-12-06
Add restriction for plugin element insert for valid parent types

iszmais:
https://mantis.ilias.de/view.php?id=38234
https://mantis.ilias.de/view.php?id=39295


alex40724:
Please note that this check is about page parent types, not repository object types.

See https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/COPage/README-technical.md#using-the-page-editor-in-another-components
and
https://docu.ilias.de/goto_docu_pg_56942_42.html

PR #6721, opened 2023-12-06
Release_8: Fix new lines in written text causing mail to be malformed

mBeym:
Fixes https://mantis.ilias.de/view.php?id=36630 for ILIAS 8


Amstutz:
thx a lot!

PR #6724, opened 2023-12-06
UI/Filter: 38597, anchor position of il-popover-container

nhaagen:
https://mantis.ilias.de/view.php?id=38597

### DIFF
--- a/templates/default/070-components/UI-framework/Input/_ui-component_filter.scss
+++ b/templates/default/070-components/UI-framework/Input/_ui-component_filter.scss
@@ -222,3 +222,6 @@ $il-standard-filter-add-input-popover-list-button-width: 100%;
 		width: $il-standard-filter-add-input-popover-list-button-width;
 	}
 }
+.il-popover-container {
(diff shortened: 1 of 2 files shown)


tfamula:
Hi @nhaagen,
thanks for the PR. The `Guidelines_SCSS-Coding.md` says in chapter "Hacks and Tweaks": 

> !important MUST NOT be used outside of this layer and SHOULD not be used here either.

But I guess that you considered this as only solution for this specific problem. For me, this is fine. But maybe @Amstutz or @catenglaender should decide here.

nhaagen:
well, it's only "relative !important" ;)

Amstutz:
@nhaagen , what @tfamula says ;-). Did you check with @catenglaender for an alternative? I guess we need to overwrite some inline js styling, right? if so, could be hard to find something better.

Amstutz:
I believe to have found a fix without the relative important by fixing the popover js. Hopes it works, if not, we will fallback to this one here.

PR #6726, opened 2023-12-07
UI: adjust the second example for presentation table. (#38124)

lukastocker:
The second example uses alignments instead of further fields. The styling (CSS) is just an example and therefore styled with a light green color and an icon to point out the difference between example 1 and 2 visually.

The general styling should be done in the specific component who will use the presentation table with alignments.

https://mantis.ilias.de/view.php?id=38124

### DIFF
--- a/src/UI/examples/Table/Presentation/base1.php
+++ b/src/UI/examples/Table/Presentation/base1.php
@@ -5,19 +5,21 @@ declare(strict_types=1);
 namespace ILIAS\UI\examples\Table\Presentation;
 
 /**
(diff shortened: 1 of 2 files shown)


lukastocker:
Hi @Amstutz ,

diese Umsetzung wurde bereits mit @catenglaender besprochen.

Amstutz:
please also check if changes are needed in the testrails cases. all examples are covered there, if so, adapt. See: https://testrail.ilias.de/index.php?/runs/view/940

PR #6728, opened 2023-12-07
UI: adjust metabar layout in general and the avatar in smaller screen…

lukastocker:
…s. (#38331)

https://mantis.ilias.de/view.php?id=38331

The avatar-letter's 'abbreviation' class got adjusted as good as possible with the given options as a temporary fix. It was discussed that the avatar needs a refactoring and therefore a new ticket will be created.


lukastocker:
Hi @Amstutz ,

this implementation was already discussed with @catenglaender .

Amstutz:
thx a lot!

PR #6729, opened 2023-12-07
Exceptions: Fix `Plain Text` error handler

mjansenDatabay:
This commit fixes the `Plain Text` error handler. It is broken since
ILIAS overwrites the super global variables. When transforming those
back to arrays for debugging/output purposes, `NULL` bytes are introduced.
This breaks the HTTP response.
Furthermore, this commit fixes the `ilDelegationHandler` by also
returning the return value of `handle` of the current `HandlerInterface`
to the caller. The same applies for the `contentType` value provided by
the actual/wrapped handler.

If approved, this has to be picked to `trunk` and `release_8`.

See: https://mantis.ilias.de/view.php?id=39316

### DIFF
--- a/Services/Exceptions/classes/class.ilDelegatingHandler.php
+++ b/Services/Exceptions/classes/class.ilDelegatingHandler.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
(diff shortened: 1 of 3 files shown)


klees:
Thanks @mjansenDatabay!

klees:

PR #6730, opened 2023-12-07
TA: 39312, JS in modals did not prevent defaults and was looking for button-ids

nhaagen:
https://mantis.ilias.de/view.php?id=39312


kergomard:
Thank you very much @nhaagen for the PR! Merged and picked to trunk.

PR #6734, opened 2023-12-07
Refinery: 39300, pass language to ByTrying so that ProblemBuilder works

nhaagen:
https://mantis.ilias.de/view.php?id=39300


mjansenDatabay (commented):
Hi @nhaagen , thanks for the suggested changes. Could you please try to explain why it is not sufficient to load the `ui` module in the UI classes :-)?

nhaagen:
Hi @mjansen, 
ilLanguage is loaded super-early into refinery; it is then encapsuled and passed within down to the components.
But then, the langvar may well belong to validation, too, as it is already the case for 'datetime_required' in the DateTime-Input.
Thanks's for pointing it out

mjansenDatabay (approved):

klees:
Picked to trunk!

PR #6743, opened 2023-12-08
Calendar: Fix accessing undefined array key

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=39337


smeyer-ilias:
Thanks for the PR

PR #6748, opened 2023-12-08
NPM: Lib tinymce

alex40724:
Usage:

WYSIWYG text editing.

Wrapped By:

components/ILIAS/COPage (internally)

Reasoning:

Main use is the text element in the page editor. Tiny provides a lot of browser independent dom and editing handling we are relying on. I did not test version 6 yet. The page editor does not need the current fork additions.

Note: This is not about the legacy form integration and its configuration in the ILIAS administration, which is abandoned more or less (hopefully completely in ILIAS 10).

Maintenance:

Well maintained, commercial background, still widely used and integrated.

Links:

Packagist: https://www.npmjs.com/package/tinymce
GitHub: https://github.com/tinymce/tinymce

### DIFF
--- a/public/package_new.json
+++ b/public/package_new.json
@@ -8,6 +8,7 @@
     "test": "tests"
   },
   "dependencies": {
+    "tinymce": "^6.9.1"
   },
   "devDependencies": {
   },


matthiaskunkel:

mjansenDatabay:
Hi @alex40724 ,

after thinking about this, I doubt we can switch to the core `tinymce/tinymce` package without abandoning some "features" **first**. The TinyMCE (if globally enabled) is used in different core components:

1. CoPage
2. SurveyQuestionpool
3. Survey
4. TestQuestionPool
5. Forum
6. Exercise

There are ILIAS specific plugins, which can be used within the components 2 - 6:

1. Create/edit LaTeX code
2. Paste from Latex
4. Insert/edit image (implemented by Minvervis ~ two years ago after upgrading the TinyMCE to version 5.x)
5. Quotation (this is forum specific and will not be a problem,)

If we use the standard `tinymce/tinymce` package in the future, our (components 2 - 6) users will not have the option to use `MathJax` or upload images (this results in `Media Objects`) anymore.

So IMO we should talk about and agree upon the consequences (for the different components) first if we do this, I doubt that this is clear for (or has been understood in the JourFixe by) all stakeholders.

@dsstrassner / @kergomard / @mbecker-databay  : What do you think? 

To be clear: It is **not** your (@alex40724) task to take care of this alone. I just wanted to point out, that these TinyMCE plugins are appreciated by our users.
So if you all agree, we could/should have a chat about this (Discord, BBB, etc.). If you all say: No, we plan to replace the TinyMCE in our compoments in general for ILIAS 10 and we will abadon `MathJax` and image uploads at all, then it's my problem alone ;-).

Best regards,
Michael

mjansenDatabay:
Hello dear colleagues, any thoughts about this?

alex40724:
I did not introduce these plugins and thus would not support them in 2,3,6.

mjansenDatabay:
> I did not introduce these plugins and thus would not support them in 2,3,6.

Thx for your feedback @alex40724 .

One last question: Do you want to add the library without any wrapping component (to be concrete: the `RTE` component), or do you request the "Authority" for the `RTE` component when (re-)adding the library?

klees:

Kind regards!

PR #6754, opened 2023-12-08
NPM: Add `dropzone` as dependency

chfsx:
This PR adds `dropzone` as npm package.

Usage:
* `ILIAS\UI\Component\Input\Field\File`
* `file.js`

Wrapped By:
* `file.js`

Reasoning:
* This package is used by the UI file input to handle file uploads on client-side. It also supports chunked uploads which have been implemented in ILIAS 9. Besides that, the library also provides further functionality like drag-and-drop file selection and dynamic rendering of HTML previews.

Maintenance:
* The package is NOT actively maintained anymore (last release 2021), see https://github.com/dropzone/dropzone/releases, but still seems to be the quasi standard

Links:
* Packagist: https://www.npmjs.com/package/dropzone
* GitHub: https://github.com/dropzone/dropzone
* Documentation: https://docs.dropzone.dev

### DIFF
--- a/public/package_new.json
+++ b/public/package_new.json
@@ -8,6 +8,7 @@
     "test": "tests"
   },
   "dependencies": {
+    "dropzone": "^5.9.3",
   },
   "devDependencies": {
   },


matthiaskunkel:
JourFixe, 11 DEC 2023: Dependency is accepted for trunk. But we have to look for alternatives in case security problems arise.

klees:

Kind regards!

PR #6766, opened 2023-12-10
Composer: Add `jumbojett/openid-connect-php` as dependency

pascalseeland:
This PR adds jumbojett/openid-connect-php as composer dependency.

Usage:
* components/ILIAS/OpenIdConnect

Reasoning:
* ILIAS uses this library to enable login via oidc. It is the defact standard when using oidc with php

Maintenance:
* Not to many commits in the last years, but there are automatic tests for php up to 8.2.
* There are several people contributing to the software

Links:

### DIFF
--- a/composer_new.json
+++ b/composer_new.json
@@ -44,6 +44,7 @@
 		"ext-xml": "*",
 		"ext-zip": "*",
 		"ext-imagick": "*",
+		"jumbojett/openid-connect-php": "^0.9.10"
 	},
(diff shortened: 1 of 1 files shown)


matthiaskunkel:
JourFixe, 11 DEC 2023: Dependency rejected for security reasons. Library has a certain risk for us. We should try to substitute this dependency by another one with similar functions but better maintenance.

PR #6777, opened 2023-12-11
Language: missing lang vars for obj permissions

nhaagen:
https://mantis.ilias.de/view.php?id=39246


matthiaskunkel:
Thanks for the fixes and sorry for the delay of my review.

PR #6783, opened 2023-12-12
Maps: 39144, check for input

nhaagen:
https://mantis.ilias.de/view.php?id=39394


klees:
Hi @alex40724,

this is in the old forms where you are listed as implicit maintainer. Feel free to assign back if you do not feel responsible.

Kind regards!

PR #6784, opened 2023-12-12
PRG: 39395, do not resolve user-link for deleted user

nhaagen:
https://mantis.ilias.de/view.php?id=39395
https://mantis.ilias.de/view.php?id=39396


klees:
Picked to trunk.

PR #6785, opened 2023-12-12
PRG, 39299: double check on numeric for points

nhaagen:
https://mantis.ilias.de/view.php?id=39299


klees (changes requested):
Hi @nhaagen,

I think this would be a general problem of the `Numeric Input Field`. Why should consumers need to specify twice that the field is indeed numeric...

Please check if I'm right and move the fix to the UI framework if so.

Thanks!

nhaagen:
There is 
        
```
       $trafo_numericOrNull = $this->refinery->byTrying([
            $this->refinery->kindlyTo()->int(),
            $this->refinery->kindlyTo()->null()
        ])
        ->withProblemBuilder(fn($txt) => $txt("numeric_only"));
```
on the input.
We need that, since the field might be empty.

That, however, is the crucial point (with both upper and lower limits in place for the PRG points instance): the check should not be against "numeric", but "required".
I fixed it accordingly.

klees:
Picked to trunk.

PR #6786, opened 2023-12-12
PRG: 39377, there _really_ should not be a fixed id on tables ;(

nhaagen:
https://mantis.ilias.de/view.php?id=39377


klees:
Picked to trunk.

PR #6790, opened 2023-12-13
Project assigned cache directory for composer and npm

monogonom:
On RHEL and binary compatible distributions, such as AlmaLinux9, problems can occur when using ```composer``` and ```npm``` when accessing the globally configured directory paths of the cache directories.

With a project-based and RHEL-compliant configuration for the cache directory of the respective application, such conflicts can be resolved very easily.

## Cache directory

- the cache directory should be created outside the webroot
- the cache directory should be ignored by git, the directory name ".cache" is suitable for this purpose
- the cli user must have write access to the cache directory.  For the command ```sudo -u apache composer install``` the cache directory should be assigned to the user / group ```apache``` and the access rights should be set to ```0700`` or ``0770``.

## Composer

- here a simple entry in the composer.json in the config block is sufficient, for example directly at ```vendor-dir``` in a new line ``` "cache-dir" : "./.cache", ```

## npm

- a file with the name ```.npmrc``` must be created in the public directory with the content ``` cache=./../.cache ```
- here the access rights for the cli user and this group could be set with ```0660```, ```0640``` or ``0600```

## Links

- https://getcomposer.org/doc/06-config.md#cache-dir
- https://docs.npmjs.com/cli/v8/commands/npm-cache
- https://docs.npmjs.com/cli/v8/configuring-npm/npmrc


kergomard:
Thank you very much @monogonom and sorry this PR completely left my field of view for so long!

I looked at the specifications and your conclusions and I think this is not the right way forward. As far as I understand it, it should be the user who takes care of setting the right `.cache`-directory with the right permissions, we should actually not do that, as we should not draw any conclusions on write rights for the webserver user (outside of the strictly necessary for running ILIAS this is). I would thus see this as a responsibility of the user. Am I missing something?

Thanks and best,
@kergomard

kergomard:
I will close this to keep my backlog clean. Please reopen if you disagree with my assessment in the previous comment.

Thanks and best,
@kergomard

monogonom:
Hi kergomard,


During the installation of ILIAS release_9 in a RHEL9 environment, the
following problem with the npm-cache (cache = “/usr/share/httpd/.npm”)
occurred again. I executed the following on the command line

*npm clean-install --omit=dev --ignore-scripts*


The Apache user does not have write access to this directory. There is also
no home directory. This is fixed with the following command line:

*sudo -u apache echo “cache=”./../cache“” >> .npmrc*

Are there any workarounds? If not, you can include this in the
documentation (e.g. troubleshooting).


Kind regards


Christian




Am Fr., 27. Sept. 2024 um 16:42 Uhr schrieb kergomard <
***@***.***>:

monogonom:
*sudo -u apache **npm clean-install --omit=dev --ignore-scripts*

Am Fr., 1. Nov. 2024 um 21:04 Uhr schrieb Christian Stepper <
***@***.***>:

PR #6798, opened 2023-12-14
UI: Fix Substr on MB-String

kergomard:
In the Letter-Avatar `getAbbreviation()` function a `substr` is applied to a value that might contain mb-strings, thus potentially breaking the letter avatar.

See: https://mantis.ilias.de/view.php?id=39468

If accepted this should to be cherry-picked to all supported versions.


Amstutz:
thx a lot!

PR #6799, opened 2023-12-15
User: Replace usages of `strftime` with `date`

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=39496


kergomard:
Thank you very much @mjansenDatabay for the PR! Will merge and pick to trunk.

PR #6801, opened 2023-12-15
Administration: Add workaround for failing `Status` command

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=39355

### DIFF
--- a/Modules/SystemFolder/classes/class.ilObjSystemFolderGUI.php
+++ b/Modules/SystemFolder/classes/class.ilObjSystemFolderGUI.php
@@ -1099,8 +1099,12 @@ class ilObjSystemFolderGUI extends ilObjectGUI
         $r = $DIC->ui()->renderer();
         $refinery = $DIC->refinery();
 
+        $dic_copy = $DIC;
+
         $metric = $this->getServerStatusInfo($refinery);
         $report = $metric->toUIReport($f, $this->lng->txt("installation_status"));
+
+        $DIC = $dic_copy;
 
         $this->tpl->setContent($r->render($report));
     }


klees:
Hi @alex40724,

I removed myself, since I'm not maintainer here. I'm available for discussions and questions, ofc, as always.

Cheers!

alex40724:
@daniwe4 Not sure why, but I am not able to assign you as reviewer here. Could you have a look at this change please. Messing with the DIC does not seem to be a good way to deal with this.

daniwe4:
Hi,
The workaround prevents the execution of two objectives that are actually the problem. 
These are:

 - class.ilDatabaseMetricsCollectedObjective.php
 - class.ilGlobalCacheMetricsCollectedObjective.php

For the former the solution is easy, see here 
https://github.com/daniwe4/ILIAS/blob/6df49ee7757ecf079e9ef0bc448cf207f51b7010/Services/Database/classes/Setup/class.ilDatabaseMetricsCollectedObjective.php#L146,
 but the latter requires a little more time. 
So a hotfix could be a TryFinally block around the code so that the objective doesn't break but also, it doesn't run correctly. See here:
https://github.com/daniwe4/ILIAS/blob/6df49ee7757ecf079e9ef0bc448cf207f51b7010/Services/GlobalCache/classes/Setup/class.ilGlobalCacheMetricsCollectedObjective.php#L49C1-L49C1

But the ilGlobalCacheMetricsCollectedObjective has to be updated to the new Cache implementation.

klees:
I had a look into this, but I don't think this is the correct fix. Some status objectives seem to mess with the DIC, these should be fixed, not the call-side.

In general, DIC does not exist for code in the setup context and hence should not be used. I'll try to track down the offenders =)

klees:
@daniwe4 is indeed right, `ilDatabaseMetricsCollectedObjective` and `ilGlobalCacheMetricsCollectedObjective` cause the problem.

daniwe4:
@klees , yes, its right

klees (changes requested):
I would suggest to close this and instead open tickets/PRs for the two offending objectives.

alex40724:
Ok, I close.  The follow-ups seem to be
https://github.com/ILIAS-eLearning/ILIAS/pull/6870
and
https://github.com/ILIAS-eLearning/ILIAS/pull/6871

PR #6804, opened 2023-12-15
TA: 39168, add col 'comment', enable filter, fix js-init of notes

nhaagen:
https://mantis.ilias.de/view.php?id=39168


kergomard:
Thank you very much for the PR @nhaagen! There was more to it though: We had a huge confusion between Notes and Comments. Additional fixes in: https://github.com/ILIAS-eLearning/ILIAS/commit/1151ae2d851e67fe8bd20747fc65ed7f97e3d415
Everything picked to trunk.

Best,
@kergomard

PR #6806, opened 2023-12-15
ilias mantis ticket 39461

chlulei:
@alex40724

Link to ticket: https://mantis.ilias.de/view.php?id=39461

Additional changes to the 'Services/Style/set_1/export.xml' contained by style.zip, located in the directory 'libs/ilias/Style/basic_style/' are necessary. The export.xml uses the attribute 'TargetRelease'. The attribute is not allowed by the schema file. A possible could be to remove the attribute 'TargetRelease' from the export.xml.


smeyer-ilias (approved):
Thanks @chlulei

PR #6807, opened 2023-12-18
Revert "Init: 39263, re-init refinery with user to pass correct language setings"

nhaagen:
Reverts ILIAS-eLearning/ILIAS#6717

https://mantis.ilias.de/view.php?id=39526


fwolf-ilias:
@pascalseeland I think the underlying issue there is less bad than the current blocker and I am also in favor of a revert for now. Testing is not possible at the moment.

nhaagen:
Alternatively, re-init of refinery might move to init of language (instead of user):
https://github.com/ILIAS-eLearning/ILIAS/pull/6808

PR #6811, opened 2023-12-18
PRG: 39380, message for mails on non-top-assignments

nhaagen:
https://mantis.ilias.de/view.php?id=39380

Screenshot from 2023-12-18 12-13-03


klees:
Picked to trunk.

PR #6813, opened 2023-12-18
xml import validation namespace fix

chlulei:


smeyer-ilias (approved):
TThanks for fixing this.

PR #6815, opened 2023-12-19
Fix type error in LegacyGotoHandler

tfamula:


chfsx:
Thanks a lot, picked to trunk as well. the code was moved from the old goto.php, therefore this was already wrong before.

PR #6816, opened 2023-12-19
9/UI/DataTable sticky select all rows header column  39210

catenglaender:
Mantis https://mantis.ilias.de/view.php?id=39210

# Issue

* "Select all rows" header column didn't stick to the left border.
* Sometimes text is peaking through on the left edge of the sticky elements

data-table_non-sticky-row-selection

# Changes

* "Select all rows" header column now also sticks to left edge
* gradient to white as a gentle transition to the other header columns
* negative left positioning for sticky elements, so no text from below can glitch through

data-table_sticky-row-selection


Amstutz:
Thx a lot

PR #6817, opened 2023-12-19
UI: fix Tooltip/CoreJS to find visible main element

nhaagen:
https://mantis.ilias.de/view.php?id=39562

### DIFF
--- a/src/UI/templates/js/Core/dist/core.js
+++ b/src/UI/templates/js/Core/dist/core.js
@@ -62,197 +62,196 @@
      * This represents one tooltip on the page.
      */
     class Tooltip {
-        /**
-         * @type {HTMLElement}
-         */
-        #tooltip;
+      /**
+       * @type {HTMLElement}
+       */
+      #tooltip;
 
-        /**
-         * The tooltip element itself.
-         * @type {Element}
-         */
-        #element;
+      /**
+       * The tooltip element itself.
+       * @type {Element}
+       */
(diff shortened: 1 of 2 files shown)


Amstutz (changes requested):
Hi @nhaagen 

This is almost impossible to review, due to all the whitespace changes. Can you directly push the whitespace changes, if according to guidelines, and only provide the changes here? 

This would make the changes much less difficult to spot. Thx.

Lg

Amstutz:
Learned something new today:

image


Still, there are many unrelated changes. please ship them seperately.

nhaagen:
Hi @Amstutz ,
I split up the changes into two commits, this is the actual change: https://github.com/ILIAS-eLearning/ILIAS/pull/6817/commits/21b2da05b2097b16bf0d380798e0f21e546b8db1

Amstutz:
Thx lot, very helpful.

I understand the array cast. However, why did you replace: element.hasOwnProperty('hidden') ? What was wrong with that. The current solution looks a bit heavy handed...

Otherwise it is good to go.

thx.

nhaagen:
Again, the linter forbade ;(
There is a [no-prototype-builtins - rule](https://eslint.org/docs/latest/rules/no-prototype-builtins) in eslint, apparently because considering all objects _really being_ prototypes is unreliable; also, the method might be injected/overwritten, e.g. by a JSON object.

PR #6818, opened 2023-12-19
TA: 39486, only authorize _changed_ answer for 'check' ('überprüfen')

nhaagen:
https://mantis.ilias.de/view.php?id=39486

### DIFF
--- a/Modules/Test/classes/class.ilTestOutputGUI.php
+++ b/Modules/Test/classes/class.ilTestOutputGUI.php
@@ -729,7 +729,7 @@ abstract class ilTestOutputGUI extends ilTestPlayerAbstractGUI
         );
 
         if (!$this->isParticipantsAnswerFixed($questionId)) {
-            if ($this->saveQuestionSolution(true)) {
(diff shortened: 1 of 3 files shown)


kergomard:
Thank you very much @nhaagen for the PR!

I don't think, this is right, though. The problem is specific to the ordering questions (or at least certain question types, it doesn't happen with single choice questions. This here would actually lead to saving an intermediate answer that would get deleted in the next step.

On a sidenote: Same problem in ILIAS 8, I changed the Issue correspondingly.

Would you be able to look into this more deeply?

Thanks again and best,
@kergomard

nhaagen:
The solution-arrays were built no matter what - so the result got saved, and as authorized as well.
This checks for the query-parameter "test_answer_changed" before retrieving the arrays.

kergomard:
Thanks for looking into this once more @nhaagen. I think this solution is in the right place now. This is still wrong from where I stand, but it now looks like the usual wrongness of the T&A ;-). Merged and ported to ILIAS 8 and trunk.

PR #6819, opened 2023-12-19
9/UI/DataTable breadcrumbs cover up header fix

catenglaender:
https://mantis.ilias.de/view.php?id=39204

# Issue

Breadcrumbs cover up the sticky data table header.

data-table_breadcrumb-covers-header

# Changes

Sticky header gets pushed down by the height of the breadcrumbs, if the breadcrumbs div is present. Might break special edge cases where breadcrumbs are hidden using display: none rather than actually not rendering them.

Standardpage variable for breadcrumb height wasn't used anywhere, now it defines the min-height.

data-table_breadcrumb-fix


Amstutz:
Thx a lot.

PR #6820, opened 2023-12-19
TA: 39402, do not save answer for formula question on initial call

nhaagen:
https://mantis.ilias.de/view.php?id=39402

### DIFF
--- a/Modules/Test/classes/PassPresentedVariablesRepo.php
+++ b/Modules/Test/classes/PassPresentedVariablesRepo.php
@@ -1,0 +1,79 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+/**
+ * Stores random-generated parts of questions
+ * in order to present the user with a fixed question during a test attempt
+ */
+class PassPresentedVariablesRepo
+{
+    public const TABLE_NAME = 'tst_qst_var_presented';
+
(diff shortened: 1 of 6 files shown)


dsstrassner:
We will discuss this at the TechSquad Meeting 04/01/24 or 11/01/24.

nhaagen:
@dsstrassner , @kergomard and @nhaagen are of the opinion that
a) pinning random values as user answer is still wrong
b) that pariticipants should not be confronted with different values for the question during one test.

This produces a need to store randomized values per question/user/testpass in a persistent way.
getInitialVariableSolutionValues() should then check this storage for existing values or generate and store new ones.

We think that this storage should not be question-specific to the formula, but that there might be other question types
that will profit from it.

Attention should be payed to correction and results in regard of this functionality.

nhaagen:
Ultimately, the generated vars must still be part of the user's solution, if not at least for backward compatibility. 
Furthermore, the question does not really have a single point of storing/generating those answer-objects, and question/result presentation ist quite mixed while both depent on user anwers to substitute variable part in the question text - without submitting the latter ones.
I did not disentangle the whole, but please consider this a step in the right direction.

kergomard:
Thank you very much @nhaagen ! I do consider this a good step forward!
I didn't test it yet, because if one little issue first:
- [ ] [This here](https://github.com/ILIAS-eLearning/ILIAS/pull/6820/files#diff-498f6047f032462823f878b0200920c74b32188507aed6136fbd6d26ae59a855R280) looks wasteful. If I'm understanding this right, we are removing all variables and resetting them on every variable. So we are running through this quite a few times and most of them unneeded. What am I not understanding?

And then one little suggestion, you don't need to follow, just please quickly explain why you wouldn't:
- [ ] I would suggest to increase the field size for the value in the new table. For now the 64 chars are clearly enough, but if we think forward to other uses ... . The 64 also seem rather arbitrary.

Thanks again and best,
@kergomard

kergomard:

First: @nhaagen and I had a talk and we came to the conclusion that while the section marked is actually wasteful, we don't have a better solution right now, so we are going to stick with it, unless somebody has a good idea.

Then:
I tested this and it seems to mostly work. There is one place where it doesn't yet:
- [x] Could you please fix the Manual Scoring for questions a participant did not answer? There no variables are shown.

Thanks again and best,
@kergomard

kergomard:
Thank you very much @nhaagen ! Merged and picked to trunk.

PR #6821, opened 2023-12-19
9/UI/Panel Header ViewControls dynamic stacking 38937

catenglaender:
https://mantis.ilias.de/view.php?id=38937

# Issue

New view controls in panel are too prominently placed: centered and below the panel header

image

# Changes

View Controls are now in panel header
View Controls button size is now reduced using the exact values suggested by @yvseiler (Thank you! :slightly_smiling_face: )
View Controls drop into a bottom header row on smaller screen sizes
Sortation uses only a glyph to save space

ViewControl_Pabel_Listing

# Outlook

A PR to make glyphs possible in the Dashboard View Control Mode will follow in January.

### DIFF
--- a/Services/Block/classes/class.ilBlockGUI.php
+++ b/Services/Block/classes/class.ilBlockGUI.php
@@ -765,7 +765,7 @@ abstract class ilBlockGUI
             ->withTargetURL($href, $this->getNavParameter() . "page")
             ->withTotalEntries($this->max_count)
             ->withPageSize($this->getLimit())
-            ->withMaxPaginationButtons(5)
+            ->withMaxPaginationButtons(1)
             ->withCurrentPage($this->getOffset() / $this->getLimit());
     }
 
@@ -977,8 +977,6 @@ abstract class ilBlockGUI
             )->withTargetURL(
                 $this->sort_target,
                 'sorting'
-            )->withLabel(
-                $this->activeSortOption
             );
             $viewControls[] = $sortation;
         }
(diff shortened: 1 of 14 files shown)


Amstutz:
Hi @catenglaender 

I am amazed with how little CSS change this is possible. Really speaks the new structuring, nice work @catenglaender . 

Since DOM changes have been necessary, the rendering tests of the panels need fixing. Things left to do:

- [x] Unit Tests: Fix them.

 I see and completely approve the change to work without label for the sortation view control in the suggested case. However, to make this also keep working for screen readers, we need two things:
- [ ] Aria Label Rule: A new rule, stating that the label MAY be empty. However, if done so, an Aria Label MUST be set.
- [ ] Aria Label Option: Add the option for an Aria Label for Sortation View Control

Considering the high visual impact of this change, I would be willing to merge this without the removed label for the moment, and add the removal of the label and the necessary changes in a further iteration. 

Thx you a lot for the important contribution.

Kindly

klees:
Hi @Amstutz,

I fixed the unit tests.

Shouldn't we just use the former "Label" as aria label as well and treat the fact that the label is hidden as an implementation detail? I mean, would there be a case where the label and the aria-label should be different? I think we already had that discussion some time ago, but I think it's a mistake to treat label and aria-label as different things.

Kind regards!

Amstutz:
You are right. We can work with the hidden label. if we will come to a point, when we need both version, we could then add this accoringly. To make some progress here before the holidays, I will:

1. Merge this as is.
2. Re-add the label 
3. Make the label screenreader only: https://mantis.ilias.de/view.php?id=39638

Thx a lot for the fix and the feedback!

PR #6822, opened 2023-12-20
TA: 39410, remove 'finalized_by' if finalization is not set

nhaagen:
https://mantis.ilias.de/view.php?id=39410


kergomard:

PR #6823, opened 2023-12-20
UI/Password: 38662, fix css for revelation/masking

nhaagen:
https://mantis.ilias.de/view.php?id=38662

### DIFF
--- a/templates/default/070-components/UI-framework/Input/_ui-component_password.scss
+++ b/templates/default/070-components/UI-framework/Input/_ui-component_password.scss
@@ -12,7 +12,7 @@
     display: none;
   }
 
-  .revealed & {
+  &.revealed {
     .revelation-reveal {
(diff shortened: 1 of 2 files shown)


Amstutz:
Hi @nhaagen 

There seems to be unrelated CSS code and the css map. Remove the changes in the map from the commit. And maybe check with @catenglaender if you are compiling to CSS in the same manor. Note I am struggling myself to have to right version of SCSS compiler ready. Maybe we need to refine the specs here somewhat.

So please:
- [x] Map: Remove it from the commit
- [x] Unrelated changes: Remove them.

other than that, this is ready to be merged. Thx a lot.

nhaagen:
Hi @Amstutz , 
I cleaned this, thanks for looking through

Amstutz:
thx a lot!

PR #6825, opened 2023-12-20
Orgu: 36658, tolerate deleted users and mark inactive

nhaagen:
https://mantis.ilias.de/view.php?id=36658


klees (changes requested):
Hi @nhaagen,

this will probably work, but I think we should go a little bit deeper with the fix.

After all, this is an inconsistency in the database. We should not fail without any chance for help, I agree here. But: This fixes just one location where this could fail. There might be even more locations in the code, where we assume that users indeed exist if they are assigned. 

So I propose that we instead add a named objective that rectifies the database in that regard: look for users that do not exist anymore in the assignment table, delete the according entries.

I bet this will come in handy some time again...

Thanks!

nhaagen:
This will not tolerate deleted users anymore;
run php setup/setup.php achieve orgunit.removeDeletedUsersFromOrgUnits to rectify the DB.

klees (approved):

klees:
Picked to trunk.

klees:
Picked to trunk.

PR #6829, opened 2023-12-20
UI: 39430, do not set attribute 'name' if empty

nhaagen:
https://mantis.ilias.de/view.php?id=39430


Amstutz:
thx a lot for this important fix. One thing left to do:

-> resolve the conflicst, then we can merge this.

thx

nhaagen:
Very sorry, @Amstutz , I just killed all my changes when rebasing. Please have another glance at them, since I just re-did the ticket...

PR #6831, opened 2023-12-20
9/legacy-UI/toolbar fix inline form 39565

catenglaender:
https://mantis.ilias.de/view.php?id=39565

# Issue

Toolbar form in media pool > video item edit > subtitles is misaligned

image

# Changes

Toolbar is in conflict with a previous fix which made only direct children be subject of the toolbar's special styling for forms. https://github.com/ILIAS-eLearning/ILIAS/pull/6688
However, simply reverting was also out of the question. Too general of a styling could affect modals with forms popping up from "within" the toolbar. So this very specific, local styling should fix the problem without breaking modals triggered from the toolbar.

image


Amstutz:
Thx a lot for this @catenglaender. I researched myself, but could not resolve this. So highly appreciated. I think the solution is OK, since toolbar is a legacy component we are trying to fade out...

PR #6836, opened 2023-12-21
TA: 39189, enable dragging in post-correction

nhaagen:
https://mantis.ilias.de/view.php?id=39189


kergomard:
Thank you very much for the PR @nhaagen ! Merged an picked to trunk.

PR #6837, opened 2023-12-21
Orgu: 39383, fix confirmationGUI for deletion of position

nhaagen:
https://mantis.ilias.de/view.php?id=39383


klees:
Picked to trunk.

PR #6840, opened 2023-12-22
TA/Utils: 39276, mb_strpos does not accept null as offset

nhaagen:
https://mantis.ilias.de/view.php?id=39276


kergomard:
Thank you very much  @nhaagen ! FYI @chfsx . I merge this as I think it is only logical. Merged and picked to trunk.

PR #6841, opened 2023-12-22
PRG: 39633, catch exception for inactive cert

nhaagen:
https://mantis.ilias.de/view.php?id=39633


klees:
Picked to trunk.

klees:
@mjansenDatabay I could not find any method on the service to check if a certificate is indeed available for a given object. Is there any viable way to do so?

mjansenDatabay:
@klees You can check whether there is an existing template for a specific object. There is no "official" API, yet, but you could ask `\ilCertificateTemplateDatabaseRepository::fetchCurrentlyUsedCertificate` and check if the current state of this template `isCurrentlyActive`.

PR #6842, opened 2023-12-22
PRG: 39639, status of course is not relayed to PRG

nhaagen:
https://mantis.ilias.de/view.php?id=39639
(also see https://github.com/ILIAS-eLearning/ILIAS/pull/6727)


klees:

PR #6843, opened 2023-12-27
9/UI/COPage js needs content in p and surrounding section dropdown

catenglaender:
# Issue

Dropdowns in page editor for paragraph style and surrounding section appear broken.

image

This is because dropdown text is rendered into caret span which should be empty so it creates the triangle for the dropdown. Dropdown text should be in front of the caret span.

# Changes

It appears that the currently selected style name is placed into the dropdown via js. The js seems to have trouble finding the correct place to insert the text when the dropdown is rendered using `""` as a placeholder string. When using `" "` as a placeholder string the dropdown is the filled in the correct place as a sibling of the span, causing the empty span to correctly render a triangle.

image

### DIFF
--- a/Services/COPage/classes/class.ilPageObjectGUI.php
+++ b/Services/COPage/classes/class.ilPageObjectGUI.php
@@ -2033,7 +2033,7 @@ class ilPageObjectGUI
 
         if ($a_paragraph_styles) {
             $sel = new \ParagraphStyleSelector($ui_wrapper, $a_style_id);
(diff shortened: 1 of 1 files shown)


Amstutz:
thx a lot, assinged to @alex40724

Amstutz:
Ticket: https://mantis.ilias.de/view.php?id=38926

alex40724:
@catenglaender Thanks for the PR.

PR #6844, opened 2023-12-27
9/UI/print no general display block 39256

catenglaender:
Mantis: https://mantis.ilias.de/view.php?id=39256

# Issue

Most grids and complex layouts look terrible in print e.g. the kprim question:

image

This is mostly due to all divs under .il-layout-page-content being set to "display: block", which makes css grids and flex-box layouts no longer functional.

# Changes

As discussed during the CSS Squad we would prefer to not apply such a general one size fits all solution and rather only apply locally optimized print styling when the browser default print styling actually needs fixing.

Therefor we removed the far too general styling of .il-layout-page-content div.

image

If a flex-box layout doesn't work in print we might want to consider using flex-box options to fix the issue.

### DIFF
--- a/templates/default/070-components/UI-framework/Layout/_ui-component_standardpage.scss
+++ b/templates/default/070-components/UI-framework/Layout/_ui-component_standardpage.scss
@@ -509,8 +509,7 @@ footer {
 				display: none !important;
 			}
 		}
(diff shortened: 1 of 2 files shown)


dsstrassner:
I assign myself for statistic purposes only.

PR #6847, opened 2024-01-02
UI/PresentationTable: 39476, use CSS instead of deprecated align-attribute

nhaagen:
https://mantis.ilias.de/view.php?id=39476


klees (approved):
Hi @nhaagen,

this LGTM. 

For documentation purpose (I know, you know this =) ) we should neither use inline HTML nor inline styles.

I still think this is ok in the context of the UI examples, since our methods to create interesting looking and appropriately styled output are currently rather limited for the presentation table. I rather have an interesting example with inline html/css then a bland and useless example without...

Kind regards!

PR #6848, opened 2024-01-02
TA: 39656, calculate points as floats

nhaagen:
https://mantis.ilias.de/view.php?id=39656


kergomard:
Thank you very much for the PR @nhaagen . Merged and cherry-picked to trunk.

PR #6850, opened 2024-01-02
T&A Bugfix #0039699: Starting test without test sequence

fhelfer:
https://mantis.ilias.de/view.php?id=39699

### DIFF
--- a/Modules/Test/classes/class.ilTestPlayerAbstractGUI.php
+++ b/Modules/Test/classes/class.ilTestPlayerAbstractGUI.php
@@ -808,6 +808,11 @@ abstract class ilTestPlayerAbstractGUI extends ilTestServiceGUI
         $this->ctrl->redirectByClass(ilTestScreenGUI::class, 'testScreen');
     }
 
+    protected function backToInfoScreenCmd(): void
+    {
+        $this->ctrl->redirectByClass('ilObjTestGUI', 'redirectToInfoScreen');
+    }
+
     /*
     * Presents the final statement of a test
     */
--- a/Modules/Test/classes/class.ilTestSequence.php
+++ b/Modules/Test/classes/class.ilTestSequence.php
@@ -536,7 +536,7 @@ class ilTestSequence implements ilTestQuestionSequence, ilTestSequenceSummaryPro
             return reset($correctedsequence);
         }
 
-        return 0;
+        return 1;
     }
 
     public function getLastSequence(): int


kergomard:
Thank you very much for the PR @fhelfer :
I understand the changes in ilTestPlayerAbstractGUI, although I would kindly ask you to change the the string 'ilObjTestGUI' to `ilObjTestGUI::class`.
I don't think the change of the return value of `ilTestSequence::getFirstSequence()` is right though. Could you explain please, why we should do that?

Thank you very much and best,
@kergomard

dsstrassner:
Hi @fhelfer and @kergomard could you please discuss this here or on discord, as @kergomard could not join the TechSquad Meeting this week? Thanks in advance!

fhelfer:
I don't think the change of the return value of `ilTestSequence::getFirstSequence()` is right though. Could you explain please, why we should do that?
returning 0 will lead to line ilTestOutputGUI::showQuestionCmd():323 resulting to true and showing failure message. @kergomard  Do you have other fixes in mind which are maybe better?

mjansenDatabay:
Hi @dsstrassner / @kergomard / @fhelfer   ,

the Technical Board regularly monitors PRs with no recent activities. This PR has been lying around for quite a while. As far as I can see, there are some open questions before @fhelfer can proceed with this. Please provide feedback or close the PR, if the topic is not relevant anymore.

Best regards,
@mjansenDatabay

kergomard:
Just to leave a trace here: Feedback in Mantis-Issue as I cannot reproduce this (anymore).

fhelfer:
@kergomard aufgrund des Ausfalls von Mantis hier
https://test9.ilias.de/go/tst/92999
Starting the test will result in Error:
The started pass for test "Einstiegstest" does not contain any questions.
Despite having questions. 
This is because there is a generic error thrown because questionId is null in ilTestOutputGUI::showQuestionsCmd():327
You can export/import the course local to reproduce.

dsstrassner:
I tried the linked with 2 users and got no Error Message from ILIAS. I also could finish the test. I think we can close this issue.

kergomard:
Hi @fhelfer 

Sorry, I left this so long (as so often)! I will close this PR, my feeling is that the problem you experienced was a side-effect of another problem, that is now fixed (if I remember correctly there was a problem with the function that detected if a test in a learning objective driven course can  be started).

Additionally: Even if we encounter the issue again, we would need a completely different solution for it, so there is no reason to keep this PR open.

Thank you very much for you patience and your effort!
@kergomard

PR #6851, opened 2024-01-02
TA: 39695, stricter handling of image for multiple/single choice

nhaagen:
https://mantis.ilias.de/view.php?id=39695

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assAnswerBinaryStateImage.php
+++ b/Modules/TestQuestionPool/classes/class.assAnswerBinaryStateImage.php
@@ -58,11 +58,16 @@ class ASS_AnswerBinaryStateImage extends ASS_AnswerBinaryState
         return $this->image;
     }
 
-    public function setImage(string $image = null): void
+    public function setImage(?string $image = null): void
     {
         if($image === '') {
             throw new \Exception('imagename must not be empty');
         }
         $this->image = $image;
+    }
+
+    public function hasImage(): bool
+    {
+        return $this->image !== null;
     }
 }
--- a/Modules/TestQuestionPool/classes/class.assAnswerBinaryStateImage.php
+++ b/Modules/TestQuestionPool/classes/class.assAnswerBinaryStateImage.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -14,6 +15,8 @@
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 /**
  * Class for answers with a binary state indicator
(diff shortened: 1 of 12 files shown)


kergomard (changes requested):

This improves consistency quite a bit. ...we will see how the  Exception on passing an empty string fares. I really hope it holds up to reality ;-). I've a few things, I think we need to change, though:
- [x]  [`assAnswerBinaryStateImage::setImage()`](https://github.com/ILIAS-eLearning/ILIAS/pull/6851/files#diff-c4502538fc70a48b19e222dffb966a2c5ce2425333969217414fd1580902e98aR61) requires a string, but we are calling it in [`assAnswerBinaryStateImage::__construct()`](https://github.com/ILIAS-eLearning/ILIAS/pull/6851/files#diff-c4502538fc70a48b19e222dffb966a2c5ce2425333969217414fd1580902e98aR53) where it might be set to `null`. Please adapt `assAnswerBinaryStateImage::setImage()` accordingly.
- [x] Please remove the duplicate empty line [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6851/files#diff-1999900d8a8edb16d38055c49ca1a4fcf26d89b13172c0383d6997d18df3bdfcR21).
- [x] Please add a space after the `if`[here](https://github.com/ILIAS-eLearning/ILIAS/pull/6851/files#diff-1999900d8a8edb16d38055c49ca1a4fcf26d89b13172c0383d6997d18df3bdfcR71).
- [x] Could we get rid of the error suppression operators, e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6851/files#diff-2040f0385338d5d22a57475b396c8cea7ff23e1b9442ba1d06ab1d3c6c7ea54dR284)? We should at least try to not introduce new ones.
- [x] I would kindly ask you to get rid of the conditions of the type `if ($answer->getImage())`, e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6851/files#diff-773eb4bafdf336929e1f3df6e763d1906d76062457d147596f39ca3adfa72b9eR289) (but there are many more). I believe, it will make our life a lot easier, if we only use those on functions returning booleans. I think the only case we care about here is that `$answer->getImage()` should not return `null`? We already took care of the empty string case. Personally I prefere these comparisons to be normal comparisons and not `is_null()`. I think they  are far more readable.

And finally one question:
- [ ] The changes [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6851/files#diff-2040f0385338d5d22a57475b396c8cea7ff23e1b9442ba1d06ab1d3c6c7ea54dR283-R285) are unrelated, right? I think they are basically right (apart from the error suppression and the $value->getPicture()  ;-)), but this is a separate case, ...or am I missing something?

Thanks again and best,
@kergomard

nhaagen:
> [ ] The changes [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6851/files#diff-2040f0385338d5d22a57475b396c8cea7ff23e1b9442ba1d06ab1d3c6c7ea54dR283-R285) are unrelated, right? I think they are basically right (apart from the error suppression and the $value->getPicture() ;-)), but this is a separate case, ...or am I missing something?

Yes, this is not exactly from the bugreport; but so are the changes from the InputWizards. 
I went through questions with Images and found/fixed the same error pattern as described in 39695

kergomard:
Thank you very much for the PR @nhaagen ! I very much like the `hasImage()`-Function! Merged and picked to trunk.

PR #6852, opened 2024-01-02
OrgU: 39676, UI Form for AMD Selection

nhaagen:
https://mantis.ilias.de/view.php?id=39676


klees (changes requested):
Hi @nhaagen,

thanks for choosing this solution and making it right instead of simply fixing the width. Good housekeeping =)

Please iterate one steps further:

- [x] Unit Test: Parametrize `getAMDForm` to its inputs, which **I think** is a list of AMD records. Make it protected. Write a smoke test that calls the method with some records and asserts that a form is indeed created.

Thanks a bunch!

klees (approved):

klees:
Picked to trunk.

PR #6854, opened 2024-01-02
9/UI/repository add item dropdowns limit height add vertical scroll 38055

catenglaender:
https://mantis.ilias.de/view.php?id=38055

# Issues

The add item dropdown in the repository shows up awkwardly long on mobile and creates two vertical scrollbars, in the worst case even scrolling outside of the layout frame. In some cases there would also be horizontal scrollbars. This makes for a very poor user experience on narrow viewports.

image

# Changes

* dropdowns in general now have a max viewport height after which they stop expanding and the scrollbar appears on the dropdown menu itself rather than on the page content or layout grid
* a loosely connected issue of notification badges in the header causing horizontal scroll bars even when the width of the layout was capped at 100% width also has been fixed

image


Amstutz:
thx a lot! 50vh seems to work for the views I just tested.

PR #6856, opened 2024-01-03
TA: 39709, hints - we do not need a special case for KioskContent anymore

nhaagen:
https://mantis.ilias.de/view.php?id=39709


kergomard:

PR #6857, opened 2024-01-03
Rep: 39713, wrap exception in onScreenMessage

nhaagen:
https://mantis.ilias.de/view.php?id=39713


alex40724:
Thanks for the PR, I fixed it slightly different only catching ilRepositoryException.

klees:
For reference: https://github.com/ILIAS-eLearning/ILIAS/commit/69730b4475f0fd6a37d3ea6ff0a53aa50b5122c9

PR #6860, opened 2024-01-04
IASS: adjust language variable for 'iass_finalize_info'. (#39423)

lukastocker:
https://mantis.ilias.de/view.php?id=39423


klees:
Picked to release_8, release_9 and trunk.

PR #6861, opened 2024-01-04
UI: 38290, Paginations with one page only

nhaagen:
https://mantis.ilias.de/view.php?id=38290


klees (approved):
LGTM

Will @Amstutz and @thibsy take over, please?

Amstutz:
Hi @nhaagen 

Thx a lot for fixing this issue. Note that this will trigger a change in the testrail case for the view controls. However, since we will adapt them for 10 and the test are mostly done for 9, I think we will wait for the new examples to testrails workflow to adapt this.

thx

PR #6865, opened 2024-01-05
Fix for Survery detailexport (#37733)

mdezi:
excel detail export didnt function when trying to export surverys containing freetext answers - tried to access an array without checking wether the array key is set.

### DIFF
--- a/Modules/Survey/Evaluation/class.ilSurveyEvaluationGUI.php
+++ b/Modules/Survey/Evaluation/class.ilSurveyEvaluationGUI.php
@@ -594,7 +594,7 @@ class ilSurveyEvaluationGUI
             $a_excel->setCell($a_excel_row, 0, $this->lng->txt("freetext_answers"));
 
             // mc/sc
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @mdezi for the PR!

I labeled it as 'improvement' as there is no Mantis-Issue for it, but it looks rather like a 'bugfix'. Please see our guidelines for PRs here: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/contributing.md#rules-for-contributors

Thanks again and best,
@kergomard

alex40724:
@mdezi Thanks for the PR. Could you please provide a bug report according to the mentioned guidelines. Thanks.

mdezi:
I had commented my changes in this ticket ( https://mantis.ilias.de/view.php?id=37733 ) as it describes a similar case with the same issue. Shall i make a new ticket for it instead?

alex40724:
Thanks, this should be fixed in the meantime.

PR #6866, opened 2024-01-05
Dashboard Bugfix #0039733: Confirmation Message Remove Multiple on Dashboard missing

fhelfer:
https://mantis.ilias.de/view.php?id=39733


iszmais:
@fhelfer Thanks alot for your effort!
But as stated within the ticket i assume this is more an issues within the presentation of confirmations in modals and should therefore be fixed inside the confirmationGUI. Especially since the usage of modal is encouraged now.

iszmais:
@fhelfer Thanks alot for your effort!
But as stated within the ticket i assume this is more an issues within the presentation of confirmations in modals and should therefore be fixed inside the confirmationGUI. Especially since the usage of modal is encouraged now.

PR #6870, opened 2024-01-08
GlobalCache: Fix Metric Objective

klees:
Hi @chfsx,

I think this broke when some internal of GlobalCache have been refactored.

Kind regards!

--

See:

* https://github.com/ILIAS-eLearning/ILIAS/pull/6801
* https://mantis.ilias.de/view.php?id=39355


chfsx:
Thanks a lot and sorry for the delay! I merged locally after rebasing and picked to trunk

PR #6871, opened 2024-01-08
Database: Fix Metric Objective

klees:
Hi @chfsx,

I fixed the objective and also made it a little more failsafe by making sure that at least the DIC is restored to its previous state, even if something fails.

Kind regards!
---
See:

* https://github.com/ILIAS-eLearning/ILIAS/pull/6801
* https://mantis.ilias.de/view.php?id=39355


chfsx:
Thanks a lot and sorry for the delay! I merged locally after rebasing and picked to trunk

PR #6872, opened 2024-01-08
Refinery: Rename test classname

mjansenDatabay:
This commit renames class `MyValidationConstraintsConstraint` to `CustomTest` to prevent issues with duplicate execution of **other** test cases. On some systems (even with an uninstalled ILIAS) some test cases (although added by the respective suites only **once**) were executed up to three times.

Furthermore, this commit removes the other classes from the PHP test case file. Class `MyToStringClass` was not used a all and I could get rid of `MyValidationConstraintsConstraint` by using an anonymous class instead.
I want to point out that the actual problem with the unit tests was not related to the "1 file 3 classes" situation.

### DIFF
--- a/components/ILIAS/Refinery/tests/Custom/Constraint/CustomTest.php
+++ b/components/ILIAS/Refinery/tests/Custom/Constraint/CustomTest.php
@@ -22,29 +22,11 @@ use ILIAS\Refinery\Custom\Constraint as CustomConstraint;
 use PHPUnit\Framework\TestCase;
 use ILIAS\Data\Factory as DataFactory;
 
-class MyValidationConstraintsConstraint extends CustomConstraint
-{
-    public function _getLngClosure(): Closure
-    {
-        return $this->getLngClosure();
-    }
-}
-
-class MyToStringClass
-{
-    private string $str_repr = '';
-
-    public function __toString(): string
-    {
-        return $this->str_repr;
-    }
-}
-
-class ValidationConstraintsCustomTest extends TestCase
+class CustomTest extends TestCase
 {
     private string $txt_id = '';
     private ilLanguage $lng;
-    private MyValidationConstraintsConstraint $constraint;
+    private CustomConstraint $constraint;
 
     protected function setUp(): void
     {
@@ -56,7 +38,12 @@ class ValidationConstraintsCustomTest extends TestCase
             return $txt($this->txt_id, $value);
         };
         $this->lng = $this->createMock(ilLanguage::class);
-        $this->constraint = new MyValidationConstraintsConstraint($is_ok, $error, new DataFactory(), $this->lng);
+        $this->constraint = new class ($is_ok, $error, new DataFactory(), $this->lng) extends CustomConstraint {
+            public function _getLngClosure(): Closure
+            {
+                return $this->getLngClosure();
+            }
+        };
     }
 
     public function testWithProblemBuilder(): void
@@ -127,8 +114,6 @@ class ValidationConstraintsCustomTest extends TestCase
             ->method("txt")
             ->with("id")
             ->willReturn("%s-%s-%s-%s-");
-
-        $to_string = new MyToStringClass();
 
         $res = $lng_closure("id", [], new stdClass(), "foo", null);
 


klees:
Hi @sebastianbergmann,

this might be interesting for you.

We had a curious case where the amount of testcases that PHPUnit reported was different for different developers. Some had PHPUnit report ~10300 tests, while others reported ~12000 tests. We tried to minimize differences, checked for php versions, PHPUnit versions and execution parameters. We were able to pin the problem down to the usage of `TestSuite::addTestFiles`, `TestSuite::addTest` for us didn't seem affected. Turned out, that for the devs with more tests some of the tests ran twice and hence the reported amount of tests was too high. We also observed that the execution order  was different for the two groups (we use random execution order), even if we'd use the same random seed.

@mjansenDatabay further pinned this down to this change here. He was one of the people with ~12000 cases, with this change he also has ~10300 as others (including GitHub-CI).

If we can do anything to help you debug this, feel free to get back to me.

Thanks for PHPUnit!

Kind regards

sebastianbergmann:
Thank you for letting me know.

If you can provide a **minimal**, **self-contained**, reproducing test case that shows this problem: great, please open a ticket for PHPUnit.

Without such a minimal, self-contained, reproducing test case I will not be able to investigate this issue.

klees:
@sebastianbergmann Since we do not have such an example I did not open a ticket. Sry. Still wanted to let you know. Cheers!

sebastianbergmann:
@realFlowControl (or @epdenouden) Maybe you can have a look at this? Both of you worked on this area in the past.

mjansenDatabay:
The reason seems to be located in: `\PHPUnit\Framework\TestSuite::addTestFile`

```php
...
        $filename   = FileLoader::checkAndLoad($filename);
        $newClasses = array_slice(get_declared_classes(), $this->declaredClassesPointer);

        // The diff is empty in case a parent class (with test methods) is added
        // AFTER a child class that inherited from it. To account for that case,
        // accumulate all discovered classes, so the parent class may be found in
        // a later invocation.
        if (!empty($newClasses)) {
            // On the assumption that test classes are defined first in files,
            // process discovered classes in approximate LIFO order, so as to
            // avoid unnecessary reflection.
            $this->foundClasses           = array_merge($newClasses, $this->foundClasses);
            $this->declaredClassesPointer = count(get_declared_classes());
        }

        // The test class's name must match the filename, either in full, or as
        // a PEAR/PSR-0 prefixed short name ('NameSpace_ShortName'), or as a
        // PSR-1 local short name ('NameSpace\ShortName'). The comparison must be
        // anchored to prevent false-positive matches (e.g., 'OtherShortName').
        $shortName      = basename($filename, '.php');
        $shortNameRegEx = '/(?:^|_|\\\\)' . preg_quote($shortName, '/') . '$/';

        foreach ($this->foundClasses as $i => $className) {
            if (preg_match($shortNameRegEx, $className)) {
                try {
                    $class = new ReflectionClass($className);
                    // @codeCoverageIgnoreStart
                } catch (ReflectionException $e) {
                    throw new Exception(
                        $e->getMessage(),
                        $e->getCode(),
                        $e,
                    );
                }
                // @codeCoverageIgnoreEnd

                if ($class->getFileName() == $filename) {
                    $newClasses = [$className];
                    unset($this->foundClasses[$i]);

                    break;
                }
            }
        }
...
```

Precondition: Our unit tests are **not** part of the composer class map (seems to have no impact on the problem)

The `$newClasses` value is initially determined by utilizing a "Pointer" and the return value of `get_declared_classes()`. The declared classes list contains class names of test suites which have already been added **before**.
If a test suite is added first (which is the case for the tests being executed multiple times) and a test file (where the contained class name does not match the name of the PHP file) is added **afterwards**, the `$newClasses` array contains many class names (of already added test suites). The subsequent regular expression based comparison when iterating over `$this->foundClasses` does **not** match (and of course the inner filename comparison as well), so `$newClasses = [$className]` is **not** executed (so the list not overwritten by the matching class name) and thus this list contains too many (and incorrect) class names added (again) as tests.

@sebastianbergmann Maybe this explanation helps.

I think (beside the actual fix in our application with this PR or by a more robust mechanism in `PHPUnit`) there are other possible fixes/workarounds like adding test files first and test suites afterwards to circumvent this problem.

PR #6874, opened 2024-01-09
Calendar: only show consultation hours of current object in repo context (26706)

schmitz-ilias:
This PR fixes [26706](https://mantis.ilias.de/view.php?id=26706) by filtering out consultation hour appointments based on context: when in repository mode, calendar entries are not shown if they are from consultation hour calendar categories, and their target object is set and doesn't match the current calendar context.

Finding the categories each entry is assigned to is a bit fiddly, I couldn't make it work without one additional db query for each entry via `ilCalendarCategoryAssignments::_lookupCategories`. If there is a better way to do this, let me know.


smeyer-ilias (approved):
Thanks @schmitz-ilias,
everything fine!

PR #6878, opened 2024-01-11
T&A Bugfix #0039816: SubsetQuestion correct_answers integer required

fhelfer:
https://mantis.ilias.de/view.php?id=39816


kergomard:
Thank you very much for the PR @fhelfer ! I think we can go with this approach here. Merged and cherry-picked to trunk.

PR #6881, opened 2024-01-11
IASS: fix action link to iass settings. (#39835)

lukastocker:
https://mantis.ilias.de/view.php?id=39835

The implementation removes some code which was not needed and resulted to using the wrong ref_id and object (Course Object instead of Individual Assessment Object) and therefore could not find the iass access handler.


klees:
Picked to trunk.

PR #6882, opened 2024-01-11
TA: 39853, 39840, 39841 - check on null/bool instead of strlen for images

nhaagen:
https://mantis.ilias.de/view.php?id=39853
https://mantis.ilias.de/view.php?id=39840
https://mantis.ilias.de/view.php?id=39841


kergomard:
Thank you very much @nhaagen for the PR, this looks good. Merged and picked to trunk.

One thing (and this is not for you @nhaagen, just if anyone reads this): I removed this `strlen()` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6882/files#diff-2879ed233c72ce34fa24b068d461689193b97a94e88b450247270f353ac22a71R108) too. Why do we want to know the length of this? We just want to be sure it is not an empty `string`. No counting required whatsoever ;-) .

PR #6885, opened 2024-01-12
UI: build Testrail Cases based on factory docs and examples

nhaagen:
Testcases for UI Components are pretty much equal for every component.
This Objective generates an XML-file to be imported into testrail, based on the actually implemented components.

run `php setup/setup.php achieve style.buildUIFrameworkExampleTestCases -y` to build the XML.

In here is a proposal to not only have factory docs for the components, but also encourage and ultimately force developers to add a description to examples. The format of those descriptions and the way we would display them in the UI Docu is of course open for debate, so I went for a very low-level implementation of that for now.
Also, I found no other way than a static list to add ids for existing test cases - suggestions are very welcome.

Here is an example-import of the generated cases: https://testrail.ilias.de//index.php?/suites/view/830&group_by=cases:section_id&group_order=asc&display_deleted_cases=0&group_id=13446


klees (changes requested):
Hi @nhaagen and @fab-kru,

@Amstutz, @thibsy and me discussed this today. We would like to implement this approach. We would make the following chances before we consider this ready:

1. cases in english
2. automated tests that expected output in every example exists
3. add expected output to every example
4. seperate yaml-section for examples with `---`, make them look like in kitchen sink in general
5. ask Fabian Kruse to report ids for new cases, so we can add them

@nhaagen: Please implement 1., 2. and 4.. @Amstutz will try to find someone that makes a PR against @nhaagens branch to implement 3. @klees will inform @fab-kru about 5. and general progress. @Amstutz will also look into the generic texts for the cases.

Thanks y'all!

klees:
Merged as: a63d1696c598b483a6129e9508916e8d8e1cc957

PR #6886, opened 2024-01-12
TA: 39863 error when switching to multiline

nhaagen:
https://mantis.ilias.de/view.php?id=39863


kergomard:

PR #6889, opened 2024-01-12
UI: Add Categorised Items Drilldowns

kergomard:
This PR proposes to add a new type of drilldown menu to be added that will in a first step be used for the modal "Add New Item" as specified in the FR: https://docu.ilias.de/goto_docu_wiki_wpage_6597_1357.html

This is currently only about the interface. The implementation is very rudimentary to make the tests pass

A few introductory notes:
- I decided to create a separate type of drilldown menu because I came to the conclusion that this is the clearest solution. By giving the first of nodes another nomenclature, I think it becomes semantically clearer, what this is about and how it can be used. Additionally this allows to reduce the accepted item types.
- I only implemented the itemsFilter for this drilldown type. I could imagine to move this a level up in the future, but am not sure it is really necessary. I would thus suggest, to keep it localized for the time being.

At the end the idea is for it to render like this:
00_AddNewItem_Modal_Search

Thank you very much and best,
·@kergomard

### DIFF
--- a/components/ILIAS/GlobalScreen/src/Scope/MainMenu/Collector/Renderer/TopParentItemDrilldownRenderer.php
+++ b/components/ILIAS/GlobalScreen/src/Scope/MainMenu/Collector/Renderer/TopParentItemDrilldownRenderer.php
@@ -44,7 +44,7 @@ class TopParentItemDrilldownRenderer extends BaseTypeRenderer
             $entries[] = $this->buildEntry($child);
         }
 
-        $dd = $this->ui_factory->menu()->drilldown($item->getTitle(), $entries);
+        $dd = $this->ui_factory->menu()->drilldown()->standard($item->getTitle(), $entries);
 
         $slate = $this->ui_factory->mainControls()->slate()->drilldown(
             $item->getTitle(),
--- a/components/ILIAS/UI/src/Component/MainControls/Slate/Factory.php
+++ b/components/ILIAS/UI/src/Component/MainControls/Slate/Factory.php
@@ -181,6 +181,6 @@ interface Factory
     public function drilldown(
         string $name,
         \ILIAS\UI\Component\Symbol\Symbol $symbol,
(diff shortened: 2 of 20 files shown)


klees (changes requested):
Hi @kergomard,

thanks a lot for moving this forward. We had a look into the interface, as requested =)

* [ ] Effect: What happens if we have more then two levels in the menu tree? It seems as if the user can be on two levels simultanously. So, first level will be displayed on the left side and the second level on the right side. If the user clicks on some item on the right side to open the third level: will the second level be shown on the left? Or will the left side keep showing the first level. How will the user know that he is now on the third level? We would appreciate a little more explanation how this would deal with more deeply nested menu structures.
* [ ] Title: "Categorised Items Drilldown Menus can have a title or a filter, but not both." -> Why is that? Can't we, e.g., just use the title for the modal where we display the menu? Or as an identifier to include it somewhere else in an upper structure? No actual need to always display it naively as a <hX> as far as we are concerned.

* [ ] Standard: We really wonder if there would ever be a situation where one would have enough space and would not want to use the enhanced version you present here. Also, this enhanced version would collaps to the Standard version if space is scarce. We would suggest to just improve the existing drilldown with the enhanced functionality you present here.
* [ ] Search: Are there really use cases, where one wouldn't like to have the search? We would suggest to simply include the search and not make it optional.

We hope this makes sense. If not, feel free to contact us, as always.

Best regards!
@klees, @Amstutz, @thibsy

klees:
Hi @kergomard,

I return this to you for the time being. Please be so kind and assign back if this is ready for the next iteration.

Kind regards!

kergomard:
Thank you very much for the fast reply @klees !

I had a very elaborate rational why my approach was actually better, but after thinking some more, I think I was wrong.
I thus created another PR [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7024) and close this PR. I would very much appreciate a feedback there. The new PR does not require an interface change.

Thanks again and best,
@kergomard

PR #6890, opened 2024-01-12
PRG: 39836, fix LP check for user not being assigned

nhaagen:
https://mantis.ilias.de/view.php?id=39836


klees:
Picked to trunk.

PR #6891, opened 2024-01-12
NPM: Add `moment`

mjansenDatabay:
This PR adds `moment` as NPM dependency.

Usage:
* Calendar: ./components/ILIAS/Calendar/classes/class.ilCalendarUtil.php
* Legacy Forms: Implicitly by `ilCalendarUtil` in the `ilDateTimeInputGUI` field type
* OnScreenChat: ./components/ILIAS/OnScreenChat/js/moment.js
  * It is mainly (but not only) used to display relative dates and especially times (locale aware) in the conversation windows. These date/time presentations are dynamically updated (without expensive server round-trips) every 60 seconds:

```js
			setInterval(() => {
				$('[data-livestamp]').each(() => {
					let $this = $(this);
					$this.html(dateTimeFormatter.fromNowToTime($this.data("livestamp")));
				});
				$('[data-message-time]').each(() => {
					let $this = $(this);
					$this.attr("title", dateTimeFormatter.format($this.data("message-time", "LT")));
				});
			}, 60000);
```

Wrapped By:
* OnScreenChat: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/OnScreenChat/js/moment.js

Reasoning:
* `moment` is a JavaScript library to parse, validate, manipulate and display dates and times in JavaScript.
It has been initially introduced by the `Calendar` component and is still used there. I guess it has been added as dependency because at that time JavaScript did not support any locale aware date/time formatting. As a project we could agree upon a migration to the native JavaScript `Intl.DateTimeFormat` and `Intl.RelativeTimeFormat` packages for client-side date.

Maintenance:
* `moment` is actively maintained.

Links:
* NPM: https://www.npmjs.com/package/moment
* GitHub: https://github.com/moment/moment
* Documentation: https://momentjs.com/docs/

### DIFF
--- a/public/package_new.json
+++ b/public/package_new.json
@@ -8,6 +8,7 @@
     "test": "tests"
   },
   "dependencies": {
+    "moment": "^2.29.4"
   },
   "devDependencies": {
   },


mjansenDatabay:
Closes because of wrong upstream

PR #6892, opened 2024-01-12
TA: 39744, exception for next-button after feedback

nhaagen:
https://mantis.ilias.de/view.php?id=39744

The JS for the buttons was not loaded/initialized.


kergomard:
Thank you very much @nhaagen for the PR! Merged and picked to trunk.

PR #6896, opened 2024-01-15
OrgU: 39826, do not assign null to icon during creation

nhaagen:
https://mantis.ilias.de/view.php?id=39826


klees:
Picked to trunk.

PR #6897, opened 2024-01-15
Refinery: 39859, do not call strlen with null

nhaagen:
https://mantis.ilias.de/view.php?id=39859


klees:
It's too early in the morning and I did not understand that this fix is in the Refinery. I reverted the commit in release_9.

IMO its fine that the Refinery throws here, as the consumer of the transformation should make sure that a string is passed.

klees:
@nhaagen Please look for a fix in the Individual Assessment.

PR #6898, opened 2024-01-15
#39893 define -ui before using it

fwolf-ilias:
ilHelpGUI::ui was not initialised and preventing help to be shown.


alex40724:
Fixed differently. Thanks for the PR!

PR #6904, opened 2024-01-16
UI/DataTable: add title to actions-column

nhaagen:
https://mantis.ilias.de/view.php?id=39207

add a title to the actions-col:

Screenshot from 2024-01-16 12-09-44

### DIFF
--- a/src/UI/Implementation/Component/Table/Renderer.php
+++ b/src/UI/Implementation/Component/Table/Renderer.php
@@ -331,8 +331,11 @@ class Renderer extends AbstractComponentRenderer
         }
 
         if ($component->hasSingleActions()) {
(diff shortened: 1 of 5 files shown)


klees (approved):

klees (changes requested):
Hi @nhaagen,

I needed to look up why we would want to have an `aria-colindex` and found [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-colindex). Maybe this is of use for someone else.

* [x] Please shorten the test by not testing complete HTML but instead just look for the interesting bits via `str_contains` (or something similar).

Thanks!

klees:
@Amstutz/@thibsy: Please continue!

PR #6907, opened 2024-01-16
OrgU: 39275, check for general activiation of object-type before filtering users

nhaagen:
https://mantis.ilias.de/view.php?id=39756


klees:
Picked to trunk.

PR #6908, opened 2024-01-16
Services/Init: Store correct goto url for redirection

lscharmer:
Fix Mantis Bug \#39411.
URL's like `/goto.php/cat/123` where truncated to `/123` when saving as the original request target.

@chfsx AFAIK your recent changes in the goto link behaviour might have broken this, as this code assumes that there are no slashes after the target php script.

### DIFF
--- a/Services/Init/classes/StartupSequence/StartUpSequenceDispatcher.php
+++ b/Services/Init/classes/StartupSequence/StartUpSequenceDispatcher.php
@@ -29,6 +29,7 @@ use ilTermsOfServiceWithdrawalStartUpStep;
 use ilUserProfileStartUpStep;
 use SplQueue;
 use ILIAS\LegalDocuments\StartUpStep;
+use ILIAS\StaticURL\Builder\StandardURIBuilder;
 
 /**
  * Class StartupSequenceDispatcher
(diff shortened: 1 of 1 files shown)


chfsx:
@lscharmer yes, this was certainly broken by the change of the goto links, but the `orig_request_target` is completely unknown to me so far and also seems to me to be a somewhat strange construct.

I have not tried it, but if you activate the "nice" urls in the administration, /goto.php/ is changed to /go/, presumably your change will then no longer work. Is this the only place where this is "processed"? I see 9 usages of `ilSession::get('orig_request_target')`, which then do something with this "URL" again. This would probably have to be solved somehow more centrally.

lscharmer:
> @lscharmer yes, this was certainly broken by the change of the goto links, but the `orig_request_target` is completely unknown to me so far and also seems to me to be a somewhat strange construct.
> 
> I have not tried it, but if you activate the "nice" urls in the administration, /goto.php/ is changed to /go/, presumably your change will then no longer work. Is this the only place where this is "processed"? I see 9 usages of `ilSession::get('orig_request_target')`, which then do something with this "URL" again. This would probably have to be solved somehow more centrally.

@chfsx Ah ok, I didn't know the readable / nice URL's have changed too :). I will test it with them too.
All other places, where `orig_request_target` is used, are just reading and redirecting to the value (without modification) and removing it afterwards, so this is the central / only place where the URL is actually written.

lscharmer:
@chfsx I updated (and tested) the code to handle `/go/` and `/goto.php/` correctly. With this all target scripts can have slashes after the script name, not just `goto.php`.

mjansenDatabay:
I agree that the URL handling was not an elaborated approach and still isn't with this PR, but as long as we do not have a semantically rich representation of our current URL with the option to be aware of the new shot syntax and corresponding methods to retrieve the correct relative path it is OK for me to accept this as a bugfix.

PR #6909, opened 2024-01-16
LDAP: Fix possible non-strings when calling `readUserData`

mjansenDatabay:
In some cases it might occur that different
PHP type issues might occur when ...:

* `\ilLDAPQuery::readUserData` is called from multiple locations
* The keys of `$this->users` in `\ilLDAPQuery` might be of type `int`, resulting in problems when processing the data in `ilLDAPAttributeToUser`


mjansenDatabay:
Hello @pascalseeland ,

any objections?

Best regards,
Michael

PR #6910, opened 2024-01-16
Object: handle reloading table in creation via copy better (33815)

schmitz-ilias:
This PR fixes [33815](https://mantis.ilias.de/view.php?id=33815) by only trying to grab the search term of the copy process from the form when it isn't in the session.
This is only a quick fix, and there is a lot more refactoring one could do here. Feel free to close this PR if you want to take this opportunity to do so.


kergomard:
Thank you very much for the PR @schmitz-ilias !
There is no need for any refactoring here as the functionality is gone with ILIAS 9 if there is no obscure use, I was unable to find. Clean-Up with ILIAS 10. So: Merged and no picks needed.

Best,
@kergomard

PR #6911, opened 2024-01-16
Tree: remove unused methods in ilTree class.

lukastocker:
This implementation removes following methods in ilTree class as they are not used anywhere:

- getLangCode
- getLeftValue
- setTreeId

Also set those methods in ilTree to protected which are only referenced in the ilTree class file.


fwolf-ilias:

PR #6912, opened 2024-01-16
WCAG: adjust outline color for 'focus-visible' for a better visibilit…

lukastocker:
…y. (#39931)

https://mantis.ilias.de/view.php?id=39931


Amstutz:
The reason behind this not working in firefox before, was that firefox seemed to use focus while chrome seem to use focus-visible.

PR #6915, opened 2024-01-16
Exercise: Fix mailing in assignment reminder

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=39763


alex40724:
@mjansenDatabay Thanks for the PR. Why not using $DIC->mail()->placeholderResolver()? This would remove a "new" and the engine instance from the consumer code, or am I wrong?

mjansenDatabay:
@alex40724 You are right, I updated this PR.

alex40724:
Great thanks for the PR! Merged to trunk.

PR #6917, opened 2024-01-17
T&A Bugfix #0039805: Create export result file empty

fhelfer:
https://mantis.ilias.de/view.php?id=39805

### DIFF
--- a/Modules/Test/classes/class.ilExcelTestExport.php
+++ b/Modules/Test/classes/class.ilExcelTestExport.php
@@ -23,7 +23,7 @@ declare(strict_types=1);
  */
 class ilExcelTestExport extends ilTestExportAbstract
 {
-    private bool $bestonly;
+    protected bool $scoredonly;
     protected ilAssExcelFormatHelper $worksheet;
 
     public function __construct(
@@ -31,10 +31,10 @@ class ilExcelTestExport extends ilTestExportAbstract
         string $filter_key_participants = ilTestEvaluationData::FILTER_BY_NONE,
         string $filtertext = '',
         bool $passedonly = false,
-        bool $bestonly = true,
+        bool $scoredonly = true,
         ilLanguage $lng = null
     ) {
-        $this->bestonly = $bestonly;
+        $this->scoredonly = $scoredonly;
         $this->worksheet = new ilAssExcelFormatHelper();
         parent::__construct($test_obj, $filter_key_participants, $filtertext, $passedonly, $lng);
     }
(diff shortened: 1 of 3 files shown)


kergomard:

Two things though:
- This seems to not address the issue presented in the comment by  @dsstrassner [here](https://mantis.ilias.de/view.php?id=39805#c99856), right?
- For the issue to be addressed, we would also need to set the `$bestonly`-parameter, right? ...and this leads me directly to the second issue: Am I right assuming that `$bestonly` is a misnomer? What we actually mean here is `$scoredonly`. This might be the best pass, but it might also simply be the last one. If I'm right: Could you please change the variable name.

@kergomard

dsstrassner:
Hi @fhelfer! As ILIAS 9 should be released in April, we need to prioritize this PR. Please check the comment from @kergomard

fhelfer:
Hi @kergomard,

Export now inludes only scored passes and also passes that are not passed
Thats why the result file was empty. There were no passed runs
Also PHP Type Erro Bugfix otw

pls review again
Best, @fhelfer

kergomard:
Thank you very much @fhelfer and sorry to have let you wait for so long!

I've one last issue, sorry! I'm having so many issues with wrong type-casts that I'm getting a little paranoid:
Are you sure these casts [here](https://github.com/ILIAS-eLearning/ILIAS/pull/6917/files#diff-2b2e8009fe3ee31b614f56f8227ba6bb44b5305b43b5f08e86705705dcf34bc6R2675) are really right? My feeling is that we are casting to avoid `null`-values. This would e.g. mean for the sequence value, that all values where sequence is `null` will then be 0. But 0 is the first question in a sequence and should thus already exist, right? And further the field is `NOT NULLABLE` in the db, so if it is null, the question should actually not exist, right? The same goes for `question_fi`. `original_id` is `NULLABLE` and right now I think in this case 0  is equal to `null`. Personally I think this is wrong, but the cast might still be correct for now. I also suspect that the method signature of `ilTestEvaluationUserData::addQuestion(int $original_id, int $question_id, float $max_points, int $sequence = null, int $pass = 0)` is not completely correct and the type of the `$sequence`-paramater should actually be `?int`  (I know, the error would be mine, but as we are now looking into this, I would like to straighten this out if possible). My educated guess would be that we should cast the `original_id` and `continue` on `question_fi`, `sequence`, or `points` being `null` which would then mean, the method signature would hold up for the moment.
I hope it becomes clear that I'm not 100% sure here what the right solution is, so if you think we should simply cast the whole shebang, it would be kind, if you could quickly explain why?

Thanks again and best,
@kergomard

fhelfer:
Hey @kergomard, I added a check with a continue.
Could you give it another review?

Best, @fhelfer

kergomard:
Thanks again @fhelfer !
This change at least guards against my fears, so lets run with it.
Merged and picked to trunk.

Best,
@kergomard

PR #6921, opened 2024-01-18
Test: Fix use statement with non-compound name 'ilDBInterface'

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=39956


kergomard:
Thank you very much @mjansenDatabay for the PR! Was apparently asleep at the wheel. This is ILIAS 8 only.

PR #6926, opened 2024-01-18
UI/DataTable: translate aria-label in header's sorting glyph

nhaagen:
https://mantis.ilias.de/view.php?id=39724
Please also see https://github.com/ILIAS-eLearning/ILIAS/pull/6702 / https://mantis.ilias.de/view.php?id=39206


klees:
Picked to trunk.

PR #6929, opened 2024-01-19
TREE: set method to protected in NestedSetTree if only used internally.

lukastocker:
Note: No methods had to be removed as all are used.


fwolf-ilias:

PR #6930, opened 2024-01-19
TREE: set method to protected in MaterializedPathTree if only used in…

lukastocker:
…ternally.

https://github.com/lukastocker/ILIAS/branches/yours


fwolf-ilias:

PR #6933, opened 2024-01-19
Fix typo in setOnScreenMessage doc

iszmais:
I assume we need to go forward, not backward :smile:


thibsy (approved):
You're right, we do! Thx a lot for providing this :).

PR #6934, opened 2024-01-19
TREE: move specific methods to NestedSetTree file and adjust calls in…

lukastocker:
… ilTree accordingly.

getLeftValue() was not moved as it is removed in PR [6911](https://github.com/ILIAS-eLearning/ILIAS/pull/6911)

### DIFF
--- a/components/ILIAS/Tree/classes/class.ilNestedSetTree.php
+++ b/components/ILIAS/Tree/classes/class.ilNestedSetTree.php
@@ -1,7 +1,22 @@
 <?php
 
+/**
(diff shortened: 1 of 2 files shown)


klees:
Hi @fwolf-ilias,

Luka found that the methods in question are actually only used for trees that exclusively us nested set, so the `\LogicException`s won't ever be triggered. This is an intermediate step on the journey to allow these locations to use the according implementation directly...

Kind regards!

fwolf-ilias:

PR #6940, opened 2024-01-19
UI: adjust the z-index for popover elements to avoid scrolling above …

lukastocker:
…the metabar. (#32769)

https://mantis.ilias.de/view.php?id=32769

This mantis ticket is solved partly. After consideration it was decided
to exchange the current webui popover library with another library in the
future, because it's not maintained anymore and hasn't been for a few
years. For this task an entry was written in the UI ROADMAP.


Amstutz:
thx a lot!

PR #6944, opened 2024-01-19
Tracking/LearningProgress: Fix accessing undefined array key

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=39540

@smeyer-ilias If approved, this should be picked to `trunk` and maybe to `release_8` as well.


smeyer-ilias (approved):
Thank @mjansenDatabay

PR #6949, opened 2024-01-21
Composer: Add firebase/php-jwt as dependency

eqsoft:
This PR adds firebase/php-jwt as composer dependency.

Usage:
* components/ILIAS/LTI

Reasoning:
* ILIAS uses this library for jwt handling (JSON Web Token: https://jwt.io/)

Maintenance:
* frequently commits (https://github.com/firebase/php-jwt/commits/main/).
* There are several people contributing to the software

Links:
* Packagist: https://packagist.org/packages/firebase/php-jwt
* GitHub: https://github.com/firebase/php-jwt
* Documentation: https://github.com/firebase/php-jwt/blob/main/README.md

### DIFF
--- a/composer_new.json
+++ b/composer_new.json
@@ -44,6 +44,7 @@
 		"ext-xml": "*",
 		"ext-zip": "*",
 		"ext-imagick": "*",
+		"firebase/php-jwt": "^6.10"
 	},
 	"require-dev": {
 	},


matthiaskunkel:
Jour Fixe, 22 JAN 2024: @Uwe-Kohnle will call a meeting to check if this lib is used by others, too. Or if it could be removed from trunk. Decision postponed after meeting.

jcopado:
I don't know if this meeting has been already called, but I would like to point that we use the JWT library, and we would like to keep using it.

klees:
Hi @Uwe-Kohnle,

since there doesn't seem to be any activity here, I close this for the time being. Feel free to reopen or recreate of course.

Kind regards!

PR #6950, opened 2024-01-21
Composer: Add celtic/lti as dependency

eqsoft:
This PR adds celtic/lti as composer dependency.

Usage:
* components/ILIAS/LTI

Reasoning:
* ILIAS uses this library for basic LTI feature

Maintenance:
* Not to many but frequently commits
* Not to many but experienced persons

Links:
* Packagist: https://packagist.org/packages/celtic/lti
* GitHub: https://github.com/celtic-project/LTI-PHP
* Documentation: https://github.com/celtic-project/LTI-PHP/blob/master/README.md

### DIFF
--- a/composer_new.json
+++ b/composer_new.json
@@ -44,6 +44,7 @@
 		"ext-xml": "*",
 		"ext-zip": "*",
 		"ext-imagick": "*",
+		"celtic/lti": "^5.0"
 	},
 	"require-dev": {
 	},


matthiaskunkel:
Jour Fixe, 22 JAN 2024: Decision about dependency is postponed. First, it should be tested if lib-lti1p3-core could replace 'celtic' and if LTI 1.1 should be dropped with ILIAS 10.

klees:

Kind regards!

PR #6953, opened 2024-01-22
Auth: Make `ilSessionTest` work

mjansenDatabay:
The tests in `ilSessionTest` are always ignored and reported as "skipped" when executing our `PHPUnit` test suite.

* With this PR I `mocked` the database by delegating calls to the `ilDBInterface` to an "in memory" database (`PDO sqlite`).
* Furthermore I removed a method which was/is not a test method at all: `tstBasicSessionBehaviour`

There is an alternative to this PR: Completely **delete** the test file.
Since the code unter test only contains wrapped database calls, it is IMO only possible to test the database calls.

### DIFF
--- a/components/ILIAS/Authentication/tests/ilSessionTest.php
+++ b/components/ILIAS/Authentication/tests/ilSessionTest.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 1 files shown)


pascalseeland:
The main reason why I did skip the old ilSession test was because it was flappy. The issue occured because of https://github.com/ILIAS-eLearning/ILIAS/blob/30e89b3bb44ea6bf4031d3d3563147fc5089ab2e/components/ILIAS/Authentication/classes/class.ilSession.php#L194C1-L199C1, which did run on arbitrary times.  With the cron check around it, I would propose the following:

- [x] Keep the old test with methode renamed back to testBasicSessionBehaviour
- [x] Add a mock to that test that will return ```true``` when querying of the job auth_destroy_expired_sessions is active
- [x] Add the current test as a second test

mjansenDatabay:
@pascalseeland Done

components/ILIAS/Authentication/tests/ilSessionTest.php
```diff
@@ -71,180 +63,279 @@ protected function setGlobalVariable(string $name, $value): void
...
             1,
             1
         );
 
-        $result = "";
-        ilSession::_writeData("123456", "Testdata");
-        if (ilSession::_exists("123456")) {
-            $result .= "exists-";
+        $cron_manager = $this->createMock(ilCronManager::class);
+        $cron_manager->method('isJobActive')->with($this->isType('string'))->willReturn(false);
```
pascalseeland: This should return true to my understanding, so the ilSessionTest assumes that the cleanup will be run by Cron and not cause the flacky behavior.

components/ILIAS/User/classes/Settings/class.ilPersonalSettingsGUI.php
```diff
@@ -784,7 +784,7 @@ protected function deleteOwnAccountStep3(): void
             $mmail->Send();
         }
 
-        $this->log->root()->log('Account deleted: ' . $ilUser->getLogin() . ' (' . $ilUser->getId() . ')');
+        $this->log->root()->log('Account deleted: ' . $this->user->getLogin() . ' (' . $this->user->getId() . ')');
```
pascalseeland: This commit does look like it does not belong to this PR

pascalseeland (commented):
Looks good besides the two remarks

components/ILIAS/User/classes/Settings/class.ilPersonalSettingsGUI.php
mjansenDatabay: You are right

mjansenDatabay:
@pascalseeland This is ready to merge. I you don't have any objections, go go go!

mjansenDatabay:
Merged, with all checklist items completed!

PR #6958, opened 2024-01-24
AdvancedMetadata: Don't initialise global page template in constructor

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=39472


smeyer-ilias:
Thanks @mjansenDatabay for fixing this.

PR #6959, opened 2024-01-24
Search: Transform query value before output

mjansenDatabay:
If approved, this must be cherry-picked to `release_7`, `release_8` and `trunk`.


smeyer-ilias (approved):
Thanks for fixing, @mjansenDatabay

PR #6961, opened 2024-01-24
9/UI fix focus outline for links across multiple lines 36738

catenglaender:

# Issue

release_9 Chrome:
image

release_9 Firefox:
image

# Changes

Links (and all inline elements) should not use the default il-focus mixin for keyboard focus as the two lines constructed by the ::after pseudo element will draw a different path than a real border would take. Therefor we will no be using il-focus-border-only for managing the focus of an inline element like a link.

Furthermore, I added a mixin for clearing that can clear all traces of a previously applied focus mixin when overriding with another focus mixing is necessary.

Ironically, il-focus which used to be the default mixin is now barely used by non-legacy components. Most modern components use the il-focus-border-only and ...-outline-only variants to avoid some specific bugs. 

release_9 Chrome:
image

release_9 Firefox:
image

# Outlook

Maybe we should try to retire il-focus in ILIAS 10 and decide to use il-focus-border-only or ...-outline-only as the default? Further testing will be necessary. I have not tested on Safari yet.

WIP porting il-focus-border-only down to release_7 and release_8 as the same (and some worse) visual issues happen there with focus lines on multi-line links.

### DIFF
--- a/templates/default/030-tools/_tool_buttons.scss
+++ b/templates/default/030-tools/_tool_buttons.scss
@@ -140,6 +140,9 @@ $button_active-transform: scale(.95) !default;
         font-size: inherit; // see set-size
         font-weight: $font-weight;
         text-decoration: $text-decoration;
+
+        // KEYBOARD FOCUS
+        @include focus.il-focus-outline-only();
     }
 
     @if $set-size {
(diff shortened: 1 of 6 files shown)


Amstutz:
only 8 needed.

Amstutz:
Thx a lot!

PR #6962, opened 2024-01-24
7/UI fix focus outline for links across multiple lines 36738

catenglaender:
ported down from release_9 version: https://github.com/ILIAS-eLearning/ILIAS/pull/6961

# Issue

## release_7 in Firefox
image
image

## release_7 in Chrome
image
image

# Changes

Ported down mixin il-focus-border-only and use it for links.
Overriding nav-tabs and sub-tabs <a> so they still use the unmodified mixin il-focus. The advantage of the classic il-focus is that it doesn't shift text.

## Firefox
image
image

## Chrome
image
image

# Outlook

Port to release_8 is WIP

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -12038,24 +12038,18 @@ a:hover {
   color: #3a4c65;
   text-decoration: underline;
 }
-a:focus-visible {
-  position: relative;
-  outline: 2px solid #FFFFFF;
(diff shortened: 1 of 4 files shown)


Amstutz:
We decided in the CSS Squad to not merge this down to 7, due to possible issues with existing skins.

PR #6963, opened 2024-01-24
8/UI fix focus outline for links across multiple lines 36738 

catenglaender:
ported down from release_9 version: https://github.com/ILIAS-eLearning/ILIAS/pull/6961

# Issue

## release_8 Firefox
image

## release_8 Chrome
image

# Changes
Ported down mixin il-focus-border-only and use it for links.
Overriding nav-tabs and sub-tabs so they still use the unmodified mixin il-focus. The advantage of the classic il-focus is that it doesn't shift text.

## Firefox
image
## Chrome
image

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -12288,24 +12288,18 @@ a:hover {
   color: #3a4c65;
   text-decoration: underline;
 }
-a:focus-visible {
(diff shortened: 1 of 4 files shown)


Amstutz:
Thx a lot!

PR #6967, opened 2024-01-25
#40120 - Fix undefined array key exception

sroenchen:
Fixes an error that can occur on Scorm upload

Mantis ticket: https://mantis.ilias.de/view.php?id=40120


Uwe-Kohnle:
href is required according to "The SCORM Content Aggregation Model" 2-143. Please check your manifest file.

PR #6970, opened 2024-01-26
Init: Modify global screen for forced startup steps

mjansenDatabay:
This PR adds a `ModificationProvider` for the startup steps a user has to fulfill after a successful login.
Unnecessary elements of the screen should be hidden because otherwise users gets confused.

See: https://mantis.ilias.de/view.php?id=40061


matthiaskunkel:
Jour Fixe, 19 FEB 2024: We highly appreciate this PR and accept it for 7, 8, 9 and trunk.

mjansenDatabay:

PR #6974, opened 2024-01-26
LegacyForms: Add `aria-describedby` for (not all) legacy form elements (`il*InputGUI`)

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=36764

Issues/Quesions:
* First of all I do not see a generic approach to add the support of the `aria-describedby` attribute, other then investing many resources (time/money) for a revision of the legacy form component. For many input fields the "byline" is rendered in the `ilPropertyFormGUI` and the corresponding HTML template (`tpl.property_form.html`).
The input element itself is rendered in the respective `il*InputGUI`. But there are (several) exceptions, where both elements are rendered in `il*InputGUI`. Furthmore some elements have multiple "bylines" (this is supported by the specification, multiple IDs can be passed as a space-separated list), and some elements do have multiple HTML input elements. Which of those elements should be described by the one (or multiple) "byline(s)"?
* Currently I prefixed the HTML `id` element with `desc_` (maybe better?: rename the prefix to `byline_`). This could be moved to the server-side (by trying to centralize this code as well) to prevent code-duplication in multiple/different HTML templates.

Fields:
* `ilTextInputGUI`: Straight forward implementation ...
* `ilCheckboxInputGUI`: Straight forward implementation ...
* `ilTextareaGUI`: This component has 0 - 2 "bylines", depending on whether the consumer called `setInfo` and/or the consumer enabled the "Character Counter". There were several issues with this "Character Counter" (no unique ID, then I went down the rabbit hole and also fixed an error in the forum `TinyMCE` usage of it). if prefixed the element with `counter_`.
* `ilNumberInputGUI`: This component has 0 - 2 "bylines", depending on whether the consumer called `setInfo` and/or the consumer wants this component to display the "constraints" (min/max). I prefixed the "constraints" block with `con_`.
* `ilRadioGroupInputGUI`: Although I added the support for the **options** of `ilRadioGroupInputGUI`, there is no support for the "byline" of the outer radio group itself. Reason: Which element does this "byline" describe? The first? None? Some kind of new introduced pseudo element?

### DIFF
--- a/Services/Form/classes/class.ilCheckboxInputGUI.php
+++ b/Services/Form/classes/class.ilCheckboxInputGUI.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 /**
  * This class represents a checkbox property in a property form.
@@ -148,6 +148,9 @@ class ilCheckboxInputGUI extends ilSubEnabledFormPropertyGUI implements ilToolba
         }
 
         $tpl->setVariable("ARIA_LABEL", ilLegacyFormElementsUtil::prepareFormOutput($this->getTitle()));
+        if ($this->getInfo() !== '') {
+            $tpl->setVariable('DESCRIBED_BY_FIELD_ID', $this->getFieldId());
+        }
 
         return $tpl->get();
     }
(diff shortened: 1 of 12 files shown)


Amstutz:
Hi @mjansenDatabay 

Thx you VERY much for these big improvements for A11y. This should make the situation much better for many use cases.

One thing, I am really not sure on the Char Counterpart of this. I had a very recent discussion with Amanda on exactly this fashion. It could be highly obtrusive to have the quickly changing char counter in the aria described field. I would first like to have feedback on possible unwanted side effects for people relying on readers before introducing this.

However, we should not keep the merging of the rest on hold for so long. So could you:

- [x] Char Counter: Can you move the counter out of the aria described, until we are sure, we are not introducing new barriers with this?

Again, thx a lot for your work.

mjansenDatabay:
@Amstutz Thanks for your feedback. I removed the "Character Counter" code from this PR.

Amstutz:
Thx a LOT!

Amstutz:
> If approved, this has to be picked to release_9 and trunk as well.

Done

PR #6978, opened 2024-01-29
UIComponent/Tabs: Transform URL before template rendering

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40225

If approved, this has to be picket (and adapted) for `release_8`, `release_9` and `trunk`.


kergomard:
Thank you very much for the PR @mjansenDatabay ! Merged and picked to all supported branches.

PR #6980, opened 2024-01-29
T&A 0017930 & 0033898

kunzeus:
https://mantis.ilias.de/view.php?id=33898
https://mantis.ilias.de/view.php?id=17930

With the PR, the right or wrong icons should now work and be displayed again. In addition, the desired change to the display text for the answers. Let me know if you have any change requests.


kergomard:
Thank you very much for the PR @kunzeus ! Merged and picked to ILIAS 8, ILIAS 9, and trunk.

PR #6981, opened 2024-01-29
bugfix #39505: TypeError in SelfRegistration with reg-code & local roles

utesche:
The PR fixes 3 TypeErrors for
issue:  https://mantis.ilias.de/view.php?id=39505
forum: https://docu.ilias.de/goto_docu_frm_1875_8811.html
.

### DIFF
--- a/Services/Registration/classes/class.ilAccountRegistrationGUI.php
+++ b/Services/Registration/classes/class.ilAccountRegistrationGUI.php
@@ -575,6 +575,7 @@ class ilAccountRegistrationGUI
         // local roles from code
         if ($this->code_was_used && is_array($code_local_roles)) {
             foreach (array_unique($code_local_roles) as $local_role_obj_id) {
+                $local_role_obj_id = (int) $local_role_obj_id;   // bugfix  https://mantis.ilias.de/view.php?id=39505
                 // is given role (still) valid?
                 if (ilObject::_lookupType($local_role_obj_id) === "role") {
                     $this->rbacadmin->assignUser($local_role_obj_id, $this->userObj->getId());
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
Thx for the PR @utesche . I suggest to cast the values in the variable assingment.

Instead of:

```php
        $code_local_roles = [];
```

AND

```php
                $code_data = ilRegistrationCode::getCodeData($code);
                if ($code_data["role_local"]) {
                    // need user id before we can assign role(s)
                    $code_local_roles = explode(";", $code_data["role_local"]);
                }
```

Should: All values of the `$code_local_roles` list should be of type `int`.

```php
        /** @var list<int>|null $code_local_roles */
        $code_local_roles = [];
```
AND

```php
                $code_data = ilRegistrationCode::getCodeData($code);
                if (isset($code_data['role_local']) && is_string($code_data['role_local'])) {
                    // need user id before we can assign role(s)
                    $code_local_roles = array_filter(array_map(
                        static fn (string $value): int => (int) $value,
                        explode(';', $code_data['role_local'])
                    ));
                }
```

Whenever `$code_local_roles` is used, the consuming code can expect all values to be of type `int`.

---

Of course, it is still not nice that consumers (like `class.ilAccountRegistrationGUI.php`) have to know internals (separator) of `$code_data['role_local']`. But that should not be part of this PR, this requires some deep refactorings.

Best regards,
Michael

mjansenDatabay:
@utesche Do have time/energy and want to include the review remarks into your PR?

utesche:
@mjansenDatabay,
as suggested, 
the strict casting to int of separate local_role_ids at usage time is changed now 
to a conversion of the whole array of ids to int when creating the array (by exploding the given input).

Services/Registration/classes/class.ilAccountRegistrationGUI.php
```diff
@@ -575,6 +580,7 @@ protected function createUser(int $a_role): string
         // local roles from code
         if ($this->code_was_used && is_array($code_local_roles)) {
             foreach (array_unique($code_local_roles) as $local_role_obj_id) {
+                $local_role_obj_id = $local_role_obj_id;
```
mjansenDatabay: IMO this line can now be deleted.

```suggestion
```

mjansenDatabay (approved):

mjansenDatabay:
@pascalseeland Please check my suggested change and merge the PR afterwards, if you don't have any further objections.

PR #6983, opened 2024-01-30
#40301 Forum Bugfix: Missing user_noti_id for cloning notification settings

playamongthestars:
https://mantis.ilias.de/view.php?id=40301
Added a missing notification setting, that caused a fatal error on cloning a forum object


mjansenDatabay (changes requested):
Thx @nmatuschek :

- [x] Please fix the unit tests
- [x] Please squash your commits (after fixing the test), so the merge results in one amotic commit.
- [x] Please reference a Mantis issue or create a new one, it the problem is not filed, yet.

mjansenDatabay (approved):

PR #6985, opened 2024-01-30
User: Fix request parameter handling for missing role selection

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40273

If approved, this has to be picked to `trunk` and maybe to `release_8` as well.

### DIFF
--- a/Services/User/classes/Cron/class.ilCronDeleteInactiveUserAccounts.php
+++ b/Services/User/classes/Cron/class.ilCronDeleteInactiveUserAccounts.php
@@ -471,14 +471,17 @@ class ilCronDeleteInactiveUserAccounts extends ilCronJob
         if ($delete_period > 0) {
             $roles = implode(',', $this->http->wrapper()->post()->retrieve(
                 'cron_inactive_user_delete_include_roles',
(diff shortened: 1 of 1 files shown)


Services/User/classes/Cron/class.ilCronDeleteInactiveUserAccounts.php
```diff
@@ -471,14 +471,17 @@ public function saveCustomSettings(ilPropertyFormGUI $a_form): bool
...
+                    $this->refinery->always([])
+                ])
             ));
 
             $this->settings->set('cron_inactive_user_delete_include_roles', $roles);
             $this->settings->set('cron_inactive_user_delete_period', (string) $delete_period);
         }
 
-        if ($this->reminderTimer > $reminder_period) {
+        if ($this->reminder_period > $reminder_period) {
```
mjansenDatabay, line 484: @kergomard This property was undeclared.

mjansenDatabay: This single fix MUST NOT be picked to `release_8`.

kergomard:
Thank you very much for the PR @mjansenDatabay ! Merged and picked to trunk. ILIAS 8 should be ok.

PR #6986, opened 2024-01-30
LegacyForms: Don't show empty `maxlength` attribute for text fields

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=39878

If approved, this should be picked to `trunk` and `release_8` as well.


Amstutz:
Thx a lot for the fix. Looks right. I will merge this. I hope it is fine if done so. Furthermore, I am not strictly responsible for the component, but look over the a11y issues, and there is only implicit maintenance active for this. Please let me know if I should not do so in the future.

Amstutz:
Picked to trunk as well.

PR #6987, opened 2024-01-30
User: Fix overwriting user attributes in XML import

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40290


kergomard:
Thank you very much @mjansenDatabay for the PR! This looks very logical ;-) . Merged and picked to ILIAS 9 and trunk.

PR #6988, opened 2024-01-30
User: Fix request parameter handling for missing role selection

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40288

If approved, this must be picked to `trunk` and `release_8` as well.


kergomard:
Thank you very much @mjansenDatabay for this PR! Merged and picked to ILIAS 8 and trunk.

PR #6989, opened 2024-01-31
Container: Fix PHP types in `\ilContainer::getSubItems`

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40238


alex40724:
Thanks for the PR, should be fixed in the meantime.

PR #6990, opened 2024-01-31
Cron: Fix timezone issues when determining due date

mjansenDatabay:
This PR fixes timezone issues when comparing the
"Last Run" datetime with the "Current" datetime.
Furthermore, this PR removes infrastructure dependencies to `time()` by using the `ClockInterface` type instead.

See: https://mantis.ilias.de/view.php?id=40302


corro:
LGTM

PR #6991, opened 2024-01-31
ItemGroup: Fix PHP types in `\ilItemGroupItems::getAssignableItems`

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40238

### DIFF
--- a/Modules/ItemGroup/classes/class.ilItemGroupItems.php
+++ b/Modules/ItemGroup/classes/class.ilItemGroupItems.php
@@ -152,6 +152,10 @@ class ilItemGroupItems
                 continue;
             }
 
+            if (!isset($node['title'])) {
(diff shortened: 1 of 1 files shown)


Modules/ItemGroup/classes/class.ilItemGroupItems.php
```diff
@@ -152,6 +152,10 @@ public function getAssignableItems(): array
                 continue;
             }
 
+            if (!isset($node['title'])) {
+                $node['title'] = '';
```
mjansenDatabay, line 156: I set the title here because the `$materials` list will be sorted by title, and I liked to prevent further issues with a `null` title.

alex40724:
Thanks for the PR, should be fixed in the meantime.

PR #6992, opened 2024-01-31
Remove array wrapper from data table token

iszmais:
https://mantis.ilias.de/view.php?id=40311

This seems to serve no purpose and as far as i can tell can be straight up removed since every consumer of the "Token Feature" in single actions has already a workaround implemented to handle array values and plain values.

### DIFF
--- a/src/UI/Implementation/Component/Table/Action/Action.php
+++ b/src/UI/Implementation/Component/Table/Action/Action.php
@@ -89,7 +89,7 @@ abstract class Action implements I\Action
         if ($target instanceof URI) {
             $target = $this->url_builder->withParameter(
                 $this->row_id_parameter,
-                [$row_id]
+                $row_id
             )
             ->buildURI();
         }
--- a/tests/UI/Component/Table/Action/ActionTest.php
+++ b/tests/UI/Component/Table/Action/ActionTest.php
@@ -90,7 +90,7 @@ class ActionTest extends ILIAS_UI_TestBase
     {
         $act = $this->link_action->withRowId('test-id');
         $this->assertEquals(
-            'ref_id=1&namespace_rowids[]=test-id',
+            'ref_id=1&namespace_rowids=test-id',
             urldecode($act->getTarget()->getQuery())
         );
     }
--- a/tests/UI/Component/Table/DataRendererTest.php
+++ b/tests/UI/Component/Table/DataRendererTest.php
@@ -464,8 +464,8 @@ EOT;
     <div class="dropdown">
         <button class="btn btn-default dropdown-toggle" type="button" data-toggle="dropdown" id="id_3" aria-label="actions" aria-haspopup="true" aria-expanded="false" aria-controls="id_3_menu"><span class="caret"></span></button>
         <ul id="id_3_menu" class="dropdown-menu">
-            <li><button class="btn btn-link" data-action="http://wwww.ilias.de?ref_id=1&namespace_param%5B%5D=row_id-1" id="id_1">label1</button></li>
-            <li><button class="btn btn-link" data-action="http://wwww.ilias.de?ref_id=1&namespace_param%5B%5D=row_id-1" id="id_2">label2</button></li>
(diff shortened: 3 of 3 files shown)


klees:
Hi @iszmais,

this actually serves a purpose, so we won't merge this 🙂

You kind of already mentioned the purpose yourself: This replaces several individual workarounds in consumer code with a single convention: "Row Ids in Data Tables will alway be transferred in an array."

We feel that this makes sense as a convention, even for single actions. Actions that are available as a single action will very likely be also available as multiaction. So consumers will be able to just implement one handler, without any workaround, for these kinds of actions. This even helps for actions that are really only available as single actions, as people won't need to think too much as well about how Row Ids will be passed in that case. So this the behaviour is just as intended.

This PR would actually roll back a previous change. @tfamula reported that it was indeed irritating to find out how the Row Ids would be transmitted, so we introduced the current behaviour and convention. Can't seem to find the ticket, though. Maybe @tfamula can...

Hope this makes sense.

Kind regards! 
@klees and @Amstutz

tfamula:
https://mantis.ilias.de/view.php?id=37494

iszmais:
Hi @klees ,

thanks a lot for the explanation. I might not share this approach but i totally get the intention behind it, therefore fine for me.

Just some minor question that approached me while dealing with this is the incompatibility with the "old" parameter approach for control structure calls is that those actively deny the usage of array parameters.
To summarize i have no problem with an parameter delivered as an array but that im obliged to handle both cases since "UI-action" parameters are always provided as arrays and "ilControl" parameter are never provided as such.

Is there a unified solution to fix this communication issue between both cases?

Greetings,
@iszmais

klees:
Hi @iszmais,

you're welcome.

Regarding your question: I don't really understand, what you are looking for, tbh. As far as I see things, you would need to know for every parameter from the query what you expect as input. Is it a string? A ref_id? Some other internal id? This just adds another type. How would you need to handle both cases? It might help if you point at some code...

Regarding the creation of URLs, there is `ILIAS\UI\URLBuilder` (and his sister for the client side). Idea is to have a facility to easily create URLs of the same form on the server and on the client side. I think that facility should play nicely with `ilCtrl`, where `ilCtrl` provides the base URL (with ilCtrl-related params) and the builder enhances it. Nonetheless there is some potential for simplification and unification here, and I think @thibsy and me will tap it at some point.

Hope this helps a little.

Kind regards!

iszmais:
@klees Thank you for the quick response.

I guess its the easiest if i make a quick fictional example:
I have a table where i want to add an action for every entry e.g. `edit`. To provide this action with an identifier i need to add such over the token and therefore my identifier would be wrapped in an array creating some url like `....&cmd=edit&my_identifier[]=123`.
So far so good, but now i also have a different access to this Editing e.g. a user has a link somewhere wich directly lead him to the edit of the only relevant entry for him. This link, since its not embedded i a table would most likely build by using the **ilCtrl** just like `ctrl->getLinkByTargetClass()` and therefore would append a parameter via `setParameterByClass()`. Since this does not allow array parameters im forced to use a disjunctive format for my link here wich results in something like `....&cmd=edit&my_identifier=123` and following have to handle **both** inputs.

To clarify my point is not about that i favor one of both approaches, but that im obliged to cover both. This seems a bit unpleasant to me beside our general approach to increase our type strictness.

Do i miss something there or is this just how it is?

Greetings,
@iszmais

thibsy:
Hello @iszmais 

Thanks for bringing this use-case to our attention. It seems your issue is with `ilCtrl` rather than the UI components, which does not yet utilise the `URLBuilder` and does not support array-parameters. 

I certainly have in mind to incorporate the new machinery into `ilCtrl` somewhen soonish, but until then, I think you will have to use a workaround. Maybe you could wrap your target URL in an `URLBuilder` and set parameters using the new machinery instead of `ilCtrl::setParameterByClass()` - which prevents global state manipulation at the same time.

Kind regards,
@thibsy

klees:
@iszmais: I see, this is indeed not covered very well currently. For the moment, I would go with the approach of @thibsy, but there for sure is some unnecessary boilerplate involved, either on the side where the URL is created, or the side where it is consumed.

PR #6999, opened 2024-02-01
COPage, RTE: Enable browser spellcheck in TinyMCE

lscharmer:
This PR enables the browser spellcheck feature for CO Page Elements using the TinyMCE and the RTE.

### DIFF
--- a/components/ILIAS/RTE/templates/default/tpl.tinymce_frm_post.js
+++ b/components/ILIAS/RTE/templates/default/tpl.tinymce_frm_post.js
@@ -149,6 +149,7 @@ tinymce.init({
     editor_deselector: "noRTEditor",
     branding: false,
     language: "{LANG}",
+    browser_spellcheck: true,
     block_formats: ilTinyMCETranslateFormats(),
     plugins: "{ADDITIONAL_PLUGINS}",
     menubar: false,
--- a/components/ILIAS/COPage/Editor/js/src/components/paragraph/ui/tiny-wrapper.js
+++ b/components/ILIAS/COPage/Editor/js/src/components/paragraph/ui/tiny-wrapper.js
@@ -208,6 +208,7 @@ export default class TinyWrapper {
         menubar: false,
         statusbar: false,
         language: 'en',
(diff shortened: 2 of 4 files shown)


oliversamoila:
Hello @alex40724, hello @matthiaskunkel, 
There is a frequently recurring request that the ILIAS editors (ILIAS Page Editor and TinyMCE) also use the spell check of the user's system. We would like to provide the solution for these two editors for ILIAS core as well.
We would be happy to briefly present this proposal at the upcoming JourFixe. 
@matthiaskunkel , you are added here because the Tiny is without an explicit maintenance.

Best regards
Oliver

matthiaskunkel:
I've added the JourFixe label so I won't forget to add the PR to the agenda for Monday.

components/ILIAS/RTE/templates/default/tpl.tinymce_frm_post.js
```diff
@@ -149,6 +149,7 @@ tinymce.init({
     editor_deselector: "noRTEditor",
     branding: false,
     language: "{LANG}",
+    browser_spellcheck: true,
```
mjansenDatabay, line 152: This LGTM!

mjansenDatabay (approved):

mjansenDatabay:
One question @oliversamoila / @lscharmer:

Did you check whether the "Spellcheck" is also enabled during a test run for "Essay/Text Questions" if the `TinyMCE` is used? I doubt that a spellcheck should be necessrarily active during an online exam.

alex40724:
Thanks for the PR. Since this is a feature request, please provide a feature wiki entry as always. We will need test cases for this, also a spec how language dependency is expected to work, browser support, how to disable this and so on.

alex40724:
Actually I think I will oppose this even as a feature request. After vaguely checking browser support and reading through issues like https://support.mozilla.org/bm/questions/1337289 the browser dependency seems to be quite high. I would not like to have to explain how to install dictionaries for various browsers and how to configure these in mantis tickets the future.

PR #7002, opened 2024-02-02
Container: Fix Adopt Content in Wrong Toolbar

kergomard:
See: https://mantis.ilias.de/view.php?id=36725


alex40724:
Thanks for the PR

PR #7003, opened 2024-02-05
[IMPROVEMENT] added event-system to IRSS

chfsx:
Issue [39642](https://mantis.ilias.de/view.php?id=39642) shows that the IRSS needs an event system in order to be informed as a component about certain operations. however, the event system must remain very general (e.g. as an event only "resource was added to collection") and can only provide trivial data. the components must decide for themselves on the basis of this data whether they want to react to it. the system follows the general observer pattern.

in a first step, only this first event is offered, but further events can be added later.


alex40724 (approved):
LGTM. Thanks.

PR #7004, opened 2024-02-05
PRG: 40070, render (default) icon in types-settings

nhaagen:
https://mantis.ilias.de/view.php?id=40070


klees:
Picked to trunk.

PR #7005, opened 2024-02-05
CSS: remove margin-top from radio button to make sure the label and r…

lukastocker:
…adio button are in line. (#27295)

https://mantis.ilias.de/view.php?id=27295


Amstutz:
thx.

PR #7010, opened 2024-02-05
UI: change modal titles from h4 and span tags to h1 tags... (#36754)

lukastocker:
... and adjust PHP Unit tests for modals.

https://mantis.ilias.de/view.php?id=36754


Amstutz:
thx a lot!

PR #7012, opened 2024-02-05
Refinery/MarkdownToHtml: Escape HTML and don't accept and unsafe links

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40204

### DIFF
--- a/src/Refinery/String/MarkdownFormattingToHTML.php
+++ b/src/Refinery/String/MarkdownFormattingToHTML.php
@@ -29,9 +29,18 @@ class MarkdownFormattingToHTML
 {
     private \League\CommonMark\CommonMarkConverter $converter;
 
-    public function __construct()
+    public function __construct(bool $escape = true)
     {
-        $this->converter = new \League\CommonMark\CommonMarkConverter();
+        $config = [
+            'allow_unsafe_links' => false,
+            'max_nesting_level' => 42 // https://commonmark.thephpleague.com/1.5/security/#nesting-level
+        ];
+
+        if ($escape === true) {
+            $config['html_input'] = 'escape';
+        }
+
+        $this->converter = new \League\CommonMark\CommonMarkConverter($config);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


corro:
Maybe consider setting `max_nesting_level` too in order to avoid excessively long rendering times and possible DoS.

tests/Refinery/String/MarkdownFormattingToHTMLTest.php
```diff
@@ -40,6 +39,7 @@ protected function setUp(): void
         $group = new Group(new Factory(), $language);
 
         $this->markdown = $group->markdown()->toHTML();
+        $this->markdown_wihtout_html = $group->markdown(true)->toHTML();
```
klees:
```suggestion
        $this->markdown_without_html = $group->markdown(true)->toHTML();
```

src/Refinery/String/MarkdownFormattingToHTML.php
```diff
@@ -29,9 +29,17 @@ class MarkdownFormattingToHTML
 {
     private \League\CommonMark\CommonMarkConverter $converter;
 
-    public function __construct()
+    public function __construct(bool $strip_html = false)
```
klees: Lets default this to `true` so people would need to allow HTML willingly.

klees (changes requested):
LGTM, please make the two changes. Feel free proceed afterwards. Thanks!

mjansenDatabay:
@klees I applied the requested changes.

Furthermore, I added the `max_nesting_level` configuration option (it is not configurable by our users), this was also a recommendation. I did not find a meaningful value for that (so I chose `42`, maybe you or @chfsx have any serious suggestion here, currently it is just a "Magic Number"), the documentation suggests a value between 10 and 50.

According to the suggestion of @chfsx I changed the HTML handling to `escape` instead of `strip`.

klees (approved):

PR #7013, opened 2024-02-05
[FIX] several path issues

chfsx:
I tried to get an HTML learning module to run in trunk, in particular to test the functions of the WebAccessChecker. I noticed that there is a lot of confusion about the paths within ILIAS. i tried to rebuild relative paths wherever necessary so that they are dependent on the current location from where the path is called (using `__DIR__`). this is necessary, for example, because the base path is different in the WAC context than when ilias.php is called.

I suspect that there are still similar problems in many places. when working with trunk - especially when creating content that creates files via outdated paths (aka not IRSS), many errors occur. for example, a directory /data is created from time to time, although /public/data should actually be used, etc. I fear that this will keep us all quite busy. not affected are all components that already use the IRSS. However, I do not see it as my responsibility to fix all these issues.

It is also clear that the many constants (such as `ILIAS_WEB_DIR`, `CLIENT_WEB_DIR`, ...) are problematic. these were introduced a long time ago without being specifically described, e.g. when which one must be used, etc. there are relative as well as absolute paths. we need to get rid of them all...


chfsx:
@Amstutz unfortunately I haven't found a way how this works for the SystemStyles in general, so all tests fail. in my opinion we could move the custom skins into a ContainerResource of the IRSS, that might have further advantages anyway. do we want to discuss this?

klees (approved):

chfsx:
Changes are in 
- Mail, LegalDocuments @mjansenDatabay 
- Exercise, HTMl Learning Module @alex40724 
- SystemStyles @Amstutz 
- Language @katringross 
- SOAP @jeph864 

Please have a look at the changes, if there is no feedback I would merge the PR on tuesday, 13.2.24

chfsx:
OK @Amstutz I found a workaround to have tests running again, see https://github.com/ILIAS-eLearning/ILIAS/pull/7013/commits/cad7a5add43ca53617ebefc7481c81ad60e99206

Amstutz:
> @Amstutz unfortunately I haven't found a way how this works for the SystemStyles in general, so all tests fail. in my opinion we could move the custom skins into a ContainerResource of the IRSS, that might have further advantages anyway. do we want to discuss this?

I believe we mentioned moving System Styles to IRSS several times already, I am definitely all for it. Let's finally do this. Ill ping you.

components/ILIAS/Mail/classes/class.ilMail.php
```diff
@@ -1367,7 +1367,7 @@ public static function _getInstallationSignature(): string
         $signature = $DIC->settings()->get('mail_system_sys_signature', '');
 
         $clientUrl = ilUtil::_getHttpPath();
-        $clientdirs = glob(ILIAS_WEB_DIR . '/*', GLOB_ONLYDIR);
+        $clientdirs = glob(CLIENT_WEB_DIR . '/*', GLOB_ONLYDIR);
```
mjansenDatabay: IMO this is not correct. `ILIAS_WEB_DIR` pointed to the value of the `path` setting in the `clients` section of the `ilias.ini.php` file and seemed to be **data** (where all client folders were located in).

`CLIENT_WEB_DIR` on the other hand is ...

```php
            define("CLIENT_WEB_DIR", ILIAS_ABSOLUTE_PATH . "/" . ILIAS_WEB_DIR . "/" . CLIENT_ID);
```

..., so it already contains a specific client id.

But: The code tries to determine the amount of ILIAS clients. With current ILIAS versions it should always be 1, so IMO the code is obsolete  in general.

katringross:
Ok for Language, Thx @chfsx

chfsx:
Thanks @mjansenDatabay, you are absolutely right! I removed the client determination in ilMail: https://github.com/ILIAS-eLearning/ILIAS/pull/7013/commits/84d499993c74ff9cd2b984702f9b48d9b1ecc382

chfsx:
@mjansenDatabay ah you did as well :-) I rebased the brach now

chfsx:
@klees please have again a look at https://github.com/ILIAS-eLearning/ILIAS/pull/7013/commits/c866bb2ebed28a0e4fc5bdc63249829ea6257bf3

Artifacts have a lot of different paths, relative ones, absolute ones, ... . this commit resolves the correct location for all of them

mjansenDatabay:
@chfsx 

> @mjansenDatabay ah you did as well :-) I rebased the brach now

Sorry, I forgot to mention that we changed this in `trunk` already :).

chfsx:
@klees  the longer I worked with Artifacts in trunk, the more I see we should not define the path we store artifacts by the consumer. Have a look at https://github.com/ILIAS-eLearning/ILIAS/pull/7013/commits/0716d431bba96e67d9700eaa6aab2cffcc14bd7e which stores all artifacts in /artifacts. Please note that there is a static method to access the path. This method is final AND static in it's meaning.

klees:
@chfsx Yes, I agree. I already have some ideas. In general, we could hide the complete storage and loading from the consumer, which would also allow to have a "DEVMODE" where artifacts are always created on the fly. Isn't very high on my agenda currently, because the current mechanism basically works, but it does not feel right to me as well.

If you want to invest some time here I would be very happy. If so, I would suggest to huddle up to discuss an approach.

klees (changes requested):
@chfsx I do not really understand what the code regarding path resolution is supposed to do. I think it would help tremendously to have some testcases for the new methods. These would also document what the code is supposed to do.

In general I'm fine with a change like this.

klees:
Also: I have looked into so many string based path manipulations, I think we could really use some solid "path" abstraction in data... Not that you have to write one, but this string based path manipulation does not feel right as well...

chfsx:
@klees I realise with my latest changes in https://github.com/ILIAS-eLearning/ILIAS/commit/0716d431bba96e67d9700eaa6aab2cffcc14bd7e those path methods are no longer needed, I removed them but made a local copy since I see the need for a Path abstraction in refinery as well. maybe we can need some snippets then

chfsx:
@klees this PR fixes https://mantis.ilias.de/view.php?id=40353 as well

klees (changes requested):
Hi @chfsx,

there's one thing I noticed regarding the new `BuildArtifactObjective::PATH` you have introduced. I appreciate the thought, but I think this change will leave the system in a state that is incoherent.

So, before the change, the `BuildArtifactObjective` was a way to define how an artifact is to be build. The path of the artifact (which was also the name, kind of) would have been given via `getArtifactPath`. In general, I could have created a `BuildArtifactObjective` that parametrises over the path and some build instruction to create various artifacts. The objective and the artifact were different entities.

After the change, the `BuildArtifactObjective` would also double as the artifact itself. Since `PATH` encodes the class in the path, I cannot have parametrized artifact objectives anymore. This will be the general trajectory how we deal with artifacts IMO, but I'm not sure if we would want to go there now, since this is definitely not just a fix but a real interface change.

If think we could have a slightly lighter version, where `PATH` only gives the base path for artifacts (and `getArtifactPath` will keep its name but have a changed description) which would show the same benefits.

I we want to go through with this now, we would at least need to do the following things imo:
* change the description of `getArtifactName`, maybe even remove it, as we already have a class name that is also a name
* get rid of the `Objective` in the  name (as this is now the artifact itself)
* add some according documentation to the class

I also thought about leaving this in the current incoherent state since we will tackle this anyway some time. But nothing is more permanent then a provisional arrangement, so I'm not really fond leaving this as is.

I hope this understandable.

Kind regards!

chfsx:
Richard and I have been talking about artifacts, we are going to do some major rebuilding for trunk on the whole infrastructure around artifacts, but it will take a while. for now we would merge this PR so that at least trunk has a few less problems regarding paths.

PR #7015, opened 2024-02-05
9/UI/forms alert help block spacing 38867

catenglaender:
https://mantis.ilias.de/view.php?id=38867

# Issue

When focusing on an input field, the focus outline covers up error messages for that field.

image

# Change

With this PR, we suggest keeping the default look of alert messages in forms for the following reasons:
* Because these are critical alerts, blocking the user to proceed, it should have the breathing room these messages deserve.
* The resulting breathing room solves the visual issue with the focus outline
* It's less CSS. Less overrides are likely to lead to less bugs.

image

## Conflicts resolved
There has been a conflict with a previous fix adding row to an existing div for date-and-time fields. Instead of adding the class to an existing div, the date-and-time html template now has its own row div with no other class applied to it.


Amstutz:
thx a lot

PR #7017, opened 2024-02-06
T&A 0039949 ideographic space and other Non-Printable-Characters

kunzeus:
https://mantis.ilias.de/view.php?id=39949

This is a, dear I say it, "temporary" fix that should be removed with PHP 8.4 where mb_trim() will be introduced.
The provided method trims all Unicode characters of the [category](https://www.php.net/manual/de/regexp.reference.unicode.php)  "other" and "seperator".

Note: Using the provided method on Multi-line input is not recommended. 

Tested on ILIAS: 8, 9 
Same code on ILIAS: 8, 9

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assQuestion.php
+++ b/Modules/TestQuestionPool/classes/class.assQuestion.php
@@ -61,6 +61,8 @@ abstract class assQuestion
     protected const DEFAULT_THUMB_SIZE = 150;
     protected const MINIMUM_THUMB_SIZE = 20;
 
+    public const TRIM_PATTERN = '/^[\p{C}\p{Z}]+|[\p{C}\p{Z}]+$/u';
+
     protected ILIAS\HTTP\Services $http;
     protected ILIAS\Refinery\Factory $refinery;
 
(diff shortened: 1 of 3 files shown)


dsstrassner:
Thanks, @kunzeus, for the prompt fix. I think we need to discuss the Mantis Issue at the TechSquad Meeting on Thursday.

kergomard:
I've one question and one change request:
- Could you please explain, why you are also removing the category other? everything in there, as far as I can see, is printable. Worstcase it is just tofu, but ok.
- This expression, as far as I can see, only replaces characters at the beginning OR at the end, ...we actually need both and should thus add the 'g'-flag.


Thank you very much and best,
Stephan

kunzeus:
Good morning

The category "other" contains two groups with [content](https://www.compart.com/en/unicode/category). The other two are currently empty: Control and Format.
[Control](https://www.compart.com/en/unicode/category/Cc) contains stuff like: Line Tabulations, End of line. Which are similar or equivalent to the trim() function filtering for \t \n and so on. It's probably not necessary to filter out everything, but it takes care of all control characters, which are otherwise invisible to the ILIAS users. 
[Format](https://www.compart.com/en/unicode/category/Cf) contains stuff like: invisible plus,  invisible separator, Zero width spaces, and other invisible stuff. I could imagine problems arising from this one. It might be better to only filter out control or to use a more specialized filter.

It worked in my testing on both sides with multiple spaces, ideographic space and other stuff. Likewise, it also works in a PHP online [sandbox](https://onlinephp.io/c/df579), but I'll change it if you prefer.

kergomard:
Thank you very much for the fast reply @kunzeus . You are completely right! Merged and picked to ILIAS 9 and trunk.

PR #7020, opened 2024-02-06
T&A 0040420 Redirect after test completion

kunzeus:
https://mantis.ilias.de/view.php?id=40420

The Code was updated to properly include the behavior of the checkbox. 

Tested and fixed for 8, also the test redirect itself works
ILIAS 9 is not affected, ILIAS 7 runs the same codebase, so should work there too.

### DIFF
--- a/Modules/Test/classes/class.ilObjTestSettingsGeneralGUI.php
+++ b/Modules/Test/classes/class.ilObjTestSettingsGeneralGUI.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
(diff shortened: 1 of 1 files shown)


dsstrassner:
Thanks a bunch @kunzeus!

dsstrassner:
Currently, there is no funding for the code review available. So, we need to wait with review, until some time is ordered at Databay or Kergomard.

mbecker-databay:
Funding secured. Patch looks good to me, thank you so much!

If you fix the copyright issue the build threw up, I'll merge it.

PR #7021, opened 2024-02-06
OrgU: 40104, wrong target for admin actions (delete)

nhaagen:
https://mantis.ilias.de/view.php?id=40104

you have to enable AdministrationPanel _before_ setting output; otherwise form target (ref_id) is wrong


klees:
Picked to trunk.

PR #7022, opened 2024-02-06
UI/maincontrols/metabar focus in engaged slate 32262

catenglaender:
https://mantis.ilias.de/view.php?id=32262

This is just a suggestion for a possible temporary fix. I outline a potentially better solution further down.

# Issue

When opening a metabar slate on the mobile small screen breakpoint, the keyboard focus selection doesn't jump into the slate. This makes getting to selectable input elements like the search input very confusing

Focus-metabar-search

# Change

After opening a slate when pressing a metabar button, keyboard focus jumps to the next focusable element in the engaged slate. This does not fix the focus tabbing order, but in most cases the focus will jump to the element the user wanted to select.

Focus-metabar-search_auto-focus

This also feels good on desktop. Clicking on the search button and immediately having the keyboard focus inside the search bar, does save a click.

Focus-metabar-search_auto-focus_desktop

# Up for debate

There are a couple of issues I should mention:
* selection order is not actually fixed, back tabbing throws you into a completely unexpected position just as before
* I am not very experienced in JS, so please treat this implementation as a working proof of concept. I am sure somebody with more experience will have some suggestions on how to improve this - speaking of which:
* There is a timeout function to wait for the slate to become engaged. Otherwise this function did search for the engaged slate too early and couldn't find it. I know event management like this only from UI animation and don't know what the best practices are in ILIAS community for queuing javascript functions.
* How do screen readers perceive such a jump in focus?
* And then arguably, the perfect solution is probably way different:

## A better solution

Right now, the metabar has duplicate content. There is a version of the entries to be shown on mobile and a separate version of the same entries for desktop. This is the core of this problem. The engaged slate on desktop is right underneath the engaged button. Consequently, on desktop the focus order is automatically correct. The mobile version is somewhere else in the DOM structure and can therefor not make the jump back there without overriding or forcing the default focus order.

In an ideal world, the DOM should always be in the correct order and all elements should only be there once. This way focus order would be correct just by browser default.

However, for ILIAS 9 it's likely too late for such a fundamental change to the standardpage layout. Since the metabar and slate design and scripts are quite extensive, such a streamlining would require extensive testing on various devices and with a variety of browsers.


catenglaender:
@nhaagen suggested to not use jquery :focusable to find the search input field, but input and buttons in general, so that at least this part is java script without using a dependency. I will update the code next week.

Amstutz:
@catenglaender , thx for the update. Since we now wait for a next action of yours, I assigned you ftm. Feel free to re-assign, as soon as ready for another look.

catenglaender:
Done, the new functions should be jquery independent now. :heavy_check_mark:

Amstutz:
thx a lot for fixing this a11y issue!

PR #7023, opened 2024-02-07
Roles: 40049, typo in reading 'assign_users' flag

nhaagen:
https://mantis.ilias.de/view.php?id=40049


kergomard:
Thank you very much for the PR @nhaagen ! This is actually already ILIAS 8 so I've merged and picked to ILIAS 8 and trunk.

PR #7030, opened 2024-02-07
Fix language set on compare call of not default languages

iszmais:
https://mantis.ilias.de/view.php?id=40441


iszmais:
Resolved per pick https://github.com/ILIAS-eLearning/ILIAS/commit/e56b2adaeba4ec869901f083a7a712b19d8d0022

PR #7031, opened 2024-02-07
UI: set text-decoration and color for btn mixing on focus. (#39678)

lukastocker:
https://mantis.ilias.de/view.php?id=39678


Amstutz:
thx a lot!

PR #7033, opened 2024-02-07
LSO: 40166, im/export title, description and container settings

nhaagen:
https://mantis.ilias.de/view.php?id=40166


klees:
Picked to trunk.

PR #7036, opened 2024-02-08
[FIX] #39318 UI: improve Modal\Interruptive label handling.

thibsy:
Hi all,

This PR should solve the following mantis issue: https://mantis.ilias.de/view.php?id=39318.
In addition this fixes labels being translated in the renderer, which has led to custom labels being translated twice so they were wrapped by `-` characters.

Please note this might be a breaking change, because I removed the `name` attribute from the submit button entirely, which used to set the `cmd` parameter to the language variable of the action button. If this is an issue for the productive release, we probably need to come up with an alternative, but I still suggest to remove the attribute for ILIAS 9 onwards.

Kind regards,
@thibsy

### DIFF
--- a/src/UI/Component/Modal/Interruptive.php
+++ b/src/UI/Component/Modal/Interruptive.php
@@ -53,7 +53,7 @@ interface Interruptive extends Modal
     /**
      * Get the label of the action button in the footer
      */
-    public function getActionButtonLabel(): string;
+    public function getActionButtonLabel(): ?string;
 
     /**
      * Get a modal like this with the action button labeled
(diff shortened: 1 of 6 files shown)


klees (changes requested):
Hi @thibsy 

thx a lot for your change. We approve it, and believe it to be a step in the right direction for the interruptive modal. Special thx for the Example adaptions and tests.

However, since this is a breaking change, we recommend not the push it down to 8 and 9 but instead only include it in trunk. There might be many cases already making use of the name.

Please:
* Change the target of the PR to 9 (or trunk?)
* Find an appropriote place to document the behaviour. Best on the public interface of the Interruptive, maybe on the getActionButtonLabel method?

Thx a lot
@amstutz and @klees

thibsy:
Hi all,

While rebasing I noticed that I already fixed this issue once with #5752 or more concretely 44329e8d67fb02b728fd3d5c218d9c410d99c958. 
This means the issue is already fixed for ILIAS 9 onwards and I therefore close this PR.

Kind regards,
@thibsy

PR #7037, opened 2024-02-08
PRGR: 40328, asynch loading of objectlist (actions) with same targets

nhaagen:
https://mantis.ilias.de/view.php?id=40328


klees:
Picked to trunk.

PR #7040, opened 2024-02-08
OrgU: 39824, error retrieving table parameters

nhaagen:
https://mantis.ilias.de/view.php?id=39824 (part2)


klees (changes requested):
Hi @nhaagen,

@mjansenDatabay and me are trying to establish this idiom:

https://github.com/ILIAS-eLearning/ILIAS/blob/release_9/Modules/LearningSequence/classes/Members/class.ilLearningSequenceMembershipGUI.php#L75

Please change accordingly.

Thanks!

klees:
Picked to trunk.

PR #7042, opened 2024-02-09
Bugfix #0040454: Backtarget 'All threads' redirects to repository

playamongthestars:
https://mantis.ilias.de/view.php?id=40454 and
https://mantis.ilias.de/view.php?id=40333
Fixes the backtarget to the threads overview of a forum


Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -1724,10 +1724,13 @@ public function prepareThreadScreen(ilObjForum $a_forum_obj): void
         $this->tpl->setTitleIcon(ilObject::_getIcon(0, "big", "frm"));
 
         $ref_id = $this->retrieveRefId();
-
+
```
```suggestion

```

PR #7045, opened 2024-02-12
UI/Prompt: use html dialog-tag for a modal prompt and introduce a Response Component 

nhaagen:
was: https://github.com/ILIAS-eLearning/ILIAS/pull/6309


klees:
Hi @thibsy, hi @Amstutz,

this is good for a review from my POV.

Let me outline the idea briefly:

* We introduce a `Dialog` component that provides the framework for an actual dialog.
* The content of the `Dialog` is capture in a `Dialog Response` that requires some `Dialog Content`.
* `Dialog Content` can be any other component that can implement the interface, such as `Message Box` or `Form`. We can attach that interface to the UI components where we see fit.
* The `Dialog Response` renders the content in a structure way, to the `Dialog` can destructure it on the client side and shove the parts (the title, the actual content, the buttons, js) to the according places.
* The dialog then captures further interaction in the content and again asyncly asks for another `Dialog Response`. Until it recieves a `Close` response and it's done.

Hope this makes some sense...

Kind regards!

Amstutz:
We (@Amstutz and @thibsy) discussed the PR with @klees.

We agree that:
- The name "Dialog" is easily confused with the Dialog element in HTML. We prefer therefore "Prompt". Other suggestions are also welcome.
- The new element is related to the modal. The differentiation between the two needs to be more clearly differed, especially the Rivals section, which is missing atm.
- We want to fade out the external usage of the modals in future, since the are semantically week. They describe how they look, not what they are supposed to do. 
- We would like to introduce a new category for components that are menat to be used internally.

So please:
- [ ] Modal Rival: Please refine the description of the Dialog and/or Modal, especially the Rivals section to clarify the differentiation between Modals and Dialogs. If applicable also add one or more usage rules for one or the other.
- [ ] Naming: Use the name Prompt.
- [ ] Roadmap: Add a section in which you describe the plan for a new internal category for the UI Components.

Thx a lot

Amstutz:
We (@Amstutz and @thibsy) discussed the PR with @klees.


We agree that:

- The name "Dialog" is easily confused with the Dialog element in HTML. We prefer therefore "Prompt". Other suggestions are also welcome.
- The new element is related to the modal. The differentiation between the two needs to be more clearly differed, especially the Rivals section, which is missing atm.
- We want to fade out the external usage of the modals in the future, since they are semantically week. They describe how they look, not what they are supposed to do.
- We would like to introduce a new category for components that are meant to be used internally.

So please:

- [x]  Modal Rival: Please refine the description of the Dialog and/or Modal, especially the Rivals section, to clarify the differentiation between Modals and Dialogs. If applicable, also add one or more usage rules for one or the other.
- [x]  Naming: Use the name prompt.
- [x]  Roadmap: Add a section in which you describe the plan for a new internal category for the UI Components.

Thx a lot

Amstutz:
Feel free to reassign if ready for e next round of review.

thibsy (changes requested):
Hi @nhaagen and @klees,

Thanks a lot for your contribution to the UI components, and special thanks for the elaborate roadmap entry - much appreciated!

Please note, this review will only focus on the public interface, so we can pass this through JF first.

- [x] `UI\Factory::prompt()` usage: with "Prompts MUST NOT be used standalone", do you mean that `Prompt\Prompt`s should always be used together with a `Prompt\PromptContent` or `Prompt\Response`, or do you mean it should always be used in a flow that consists of multiple steps? (Maybe answer this by extending the rule)
- [x] `Prompt\Factory::standard()`: can you explain why we do not accept an initial `Prompt\PromptContent` when creating a `Prompt\Standard`? Do you want to keep this component strictly asynchronous?

- [x] `Prompt\Response`: please move this interface to a new namespace and rename it to `Prompt\Instruction\Instruction`. The term response is already used in other contexts such as HTTP, which could be misleading for our case. Please also extract the `response()` method as `show()`, and the `redirect()` and `close()` methods from the `Prompt\Factory` and move them to a new `Prompt\Instruction\Factory`. This will allow us to document the concept of "HTML over the wire" a bit when populating the new factory under `Prompt\Factory::instruction()`. I think we can use this concept for other components in the future, and even shift the handling of instructions closer to the UI framework, once we have our own endpoint.
- [x] `Prompt\PromptContent`: please remove both methods from this interface and instead introduce a `$title` and optional `$buttons` argument on the (new) `Prompt\Instruction\Factory::show()` method. This decouples the logic from our other components and does not leave consumers wondering, how these properties are generated/derived. Nonetheless, a description may be necessary to explain, how `Form\Standard` submit buttons are added to the prompt.
- [x] Accessibility: there aren't any rules about the accessibility of `Prompt\Prompt`s yet. Please add one or more accessibility rule(s) to `Prompt\Factory::standard()` or `UI\Factory::prompt()`.
- [ ] Spacing: please make sure that the new factory descriptions adhere to the template in terms of spacing. I noticed you are not grouping the entries with newlines properly and some indentations are too short. This could be an issue when parsing the descriptions to yaml.
- [x] Unnecessary comments: please remove the class-level comments of the following interfaces, because they do not really add anything to it: `Prompt\Factory`, `Prompt\Prompt`, `Prompt\Response`.

Kind regards,
@Amstutz and @thibsy

nhaagen:
Dear @Amstutz and @thibsy,
thanks for your input and the streamlining of Instructions, especially in regard to the ProgressBar-PR.

> - [] Prompt\Factory::standard(): can you explain why we do not accept an initial Prompt\PromptContent when creating a Prompt\Standard? Do you want to keep this component strictly asynchronous?

and
> - [] Prompt\PromptContent: please remove both methods from this interface and instead introduce a $title and optional $buttons argument on the (new) Prompt\Instruction\Factory::show() method. This decouples the logic from our other components and does not leave consumers wondering, how these properties are generated/derived. Nonetheless, a description may be necessary to explain, how Form\Standard submit buttons are added to the prompt.

are sort of touching the same issue:
I want to really stress the different nature of the Prompt as opposed to its contents, i.e. the focus should not be on factoring a dialog element; it's merely a means of transportation.
In order to really make use of Prompts, consumers will have to set up an endpoint anyways.
I've seen places where modals were built/rendered per action, sometimes even for the same action on different objects - 
I really would like to avoid these patterns in advance.

Same goes for the title and buttons: I think, consumers _should_ maybe not wonder, but completely neglect the origin of those. They belong to the content, not the prompt-layer, and when creating new PromptContent, they should be defined according to the content itself, not the calling of it by consumers.
My hopes are to create reusable PromptContent-Components with strong semantics that may - or may not - be displayed in a modal.

Kind regards,
Nils

nhaagen:
renamed "Instruction" to "State",  "PromptContent"  to "isPromptContent"

components/ILIAS/UI/src/Factory.php
```diff
@@ -1047,4 +1046,57 @@ public function helpTopics(string ...$topic): array;
...
+     * description:
+     *   purpose: >
+     *     A Prompt requires a user to make some inputs to the system, like making
+     *     choices, acknowleding an important information or filling out a form. The
+     *     Prompt urges the user to make the input instead of simply giving the
+     *     opportunity to do so. Hence, Prompts are used to request certain user input
+     *     that is required to move on in an ongoing workflow.
+     *   composition: >
+     *     The Prompt provides a wrapper for Prompt Content; the content is transmitted
+     *     asynchrounously via a Prompt Response. The Response consists of a Title, the
```
thibsy:
```suggestion
     *     asynchrounously via a Prompt State. The State consists of a Title, the
```

components/ILIAS/UI/src/Component/Prompt/Factory.php
```diff
@@ -0,0 +1,80 @@
...
+     *      actions of Forms and targets of Links are "wrapped" to RPCs and thus
+     *      stay in context of the Prompt, i.e. you may take roundtrips to the
+     *      server and modify the Prompt's content without closing it.
+     * context:
+     *   - The Prompt requires a Prompt State.
+     *
+     * rules:
+     *   usage:
+     *     1: >
+     *      The server MUST answer with an Instruction Component
```
thibsy:
```suggestion
     *      The server MUST answer with an Prompt State Component
```

components/ILIAS/UI/src/Component/Prompt/State/Factory.php
```diff
@@ -0,0 +1,96 @@
...
+
+interface Factory
+{
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *      Build a Prompt State to show the Contents given here.
+     *   composition: >
+     *      The Prompt Instruction accepts Prompt Content to be handled by
```
thibsy:
```suggestion
     *      The Prompt State accepts Prompt Content to be handled by
```

components/ILIAS/UI/src/Component/Prompt/isPromptContent.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace ILIAS\UI\Component\Prompt;
+
+use ILIAS\UI\Component\Component;
+use ILIAS\UI\Component\Button\Button;
+
+/**
+ * Flags some Component to be a valid content of a Prompt
+ */
+interface isPromptContent extends Component
```
thibsy:
```suggestion
interface IsPromptContent extends Component
```
(Along with all usages)

components/ILIAS/UI/src/Implementation/Component/Prompt/isPromptContent.php
```diff
@@ -0,0 +1,39 @@
...
+namespace ILIAS\UI\Implementation\Component\Prompt;
+
+use ILIAS\UI\Component\Component;
+use ILIAS\UI\Component\Button\Button;
+use ILIAS\UI\Component\Prompt\isPromptContent as IPromptContent;
+
+/**
+ * Flags some Component to be a valid content of a Prompt
+ */
+interface isPromptContent extends IPromptContent
```
thibsy:
```suggestion
interface IsPromptContentInternal extends IPromptContent
```
(Consistent with e.g. `FormInputInternal`)

thibsy (approved):
Hi @nhaagen and @klees,

Thx for answering our questions and implementing the requested changes. Special thanks for streamlining the "HTML over the wire" concept with the progress bar PR!

I found some remnants of "response" and "instruction" terms and some minor misspelling, which I have added as suggestions to this PR. This should not prevent this from going to JF though, so I have added the label accordingly.

Kind regards,
@thibsy

matthiaskunkel:
Jour Fixe, 14 OCT 2024: We highly appreciate this suggestion and accept the PR for trunk. We ask all developers to have a look at this concept and to ask questions and post comments.

Amstutz:
Thx a lot @nhaagen, for the great work. This is a rather large and important improvement for our client side work.

PR #7047, opened 2024-02-12
9/UI/worflow darker colors for unavailable steps 39873

catenglaender:
Mantis: https://mantis.ilias.de/view.php?id=39873

# Issue

Contrast is really low for steps marked as unavailable in workflow listing UI component. Text color to background color contrast ratio fails accessibility tests.

image

#  Changes

Applied color values as suggested by @yvseiler in Mantis ticket.

image


Amstutz:
Thx for improving this component!

PR #7052, opened 2024-02-13
UI: change text color of placeholder when field input is required. (#…

lukastocker:
…36768)

https://mantis.ilias.de/view.php?id=36768

This change will provide a better contrast for all placeholder texts where a required field is implemented.

placeholder-required-field-color-with-alert-box

placeholder-required-field-color

Placeholder texts in input fields without required inputs aren't changed as the background is still light instead of reddish:
placeholder-without-required-field-color


Amstutz:
thx a lot!

PR #7056, opened 2024-02-13
Session: Fix Missing Container Function

kergomard:
See: https://mantis.ilias.de/view.php?id=38876

I think this is the best solution I can come up with: As far as I can see this is a complete unicorn and I didn't even realize this was a possibility.

Best,
@kergomard


smeyer-ilias:
Thank for fixing this.

PR #7057, opened 2024-02-14
PRG: 40495, mark assignment-src for automatic restart

nhaagen:
https://mantis.ilias.de/view.php?id=40495


klees:
First commit picked to trunk.

PR #7059, opened 2024-02-14
T&A: fix for mantis-Issue #35942

okaufman:
After failing a test the user will see a info-messagebox instead of a failure-messagebox. 
If they are using a screenreader "information message" (Informationsmeldung) will be read instead of "failure message"(Fehlermeldung)


kergomard:
Thank you very much for the PR @okaufman ! Sorry, it took me so long to get around to it. Merged, no need for picking.

PR #7063, opened 2024-02-15
UI/Input/DateTime: 39859, add 'not null' to required-constraint of DateTime

nhaagen:
https://mantis.ilias.de/view.php?id=39859
(https://github.com/ILIAS-eLearning/ILIAS/pull/6897)


klees:
Picked to trunk.

PR #7065, opened 2024-02-16
CSVUtil: Remove usage of `utf8_decode`

mjansenDatabay:
The PHP functions `utf8_encode` and `utf8_decode` are deprecated as of PHP 8.2. Since ILIAS 9 supports PHP 8.2 and using this functions will raise errors, we'll have to get rid of them.

The class `ilCSVUtil` is used by the `Test` and `User` components.


kergomard:
Thank you very much @mjansenDatabay for fixing this! Merged and picked to trunk.

PR #7066, opened 2024-02-16
CSVWriter: Remove usage of `utf8_decode`

mjansenDatabay:
The PHP functions `utf8_encode` and `utf8_decode` are deprecated as of PHP 8.2. Since ILIAS 9 supports PHP 8.2 and using this functions will raise errors, we'll have to get rid of them.

The class `ilCSVWriter` is used by several components which provide CSV exports, but `setDoUTF8Decoding` is **not** used at all. so I removed the usage.

- BookingManager
- Poll
- Scorm2004
- ScormAicc
- Session
- StudyProgramme
- Survey
- Authentication
- Membership
- MyStaff
- Table
- Tracking
- WebServices

### DIFF
--- a/Services/CSV/classes/class.ilCSVWriter.php
+++ b/Services/CSV/classes/class.ilCSVWriter.php
@@ -1,19 +1,20 @@
 <?php
 
-/******************************************************************************
+/**
+ * This file is part of ILIAS, a powerful learning management system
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @mjansenDatabay for finding this and for the PR! Merged and picked to trunk.

PR #7067, opened 2024-02-16
40554 Bugfix: Cannot assign string to property ilUserImportParser::$user_id

playamongthestars:
Fixes this Bug: https://mantis.ilias.de/view.php?id=40554


kergomard:
Thank you very much for the PR @nmatuschek ! Merged and picked to trunk.

PR #7069, opened 2024-02-19
Setup: fix command of skipping installation of plugin. (#40470)

lukastocker:
https://mantis.ilias.de/view.php?id=40470

`strtolower` was removed as it prevented a distinct comparison with the plugin name and therefore didn't skip installing and activating the plugin.


klees:
Picked to trunk.

PR #7070, opened 2024-02-19
Setup: show message if user tries to skip a major upgrade. (#40142)

lukastocker:
https://mantis.ilias.de/view.php?id=40142


klees:
Hi @lukastocker,

I don't think that this is the correct fix. We need to fix this in the current release, not a previous one.

Kind regards!

PR #7071, opened 2024-02-19
Provide correct function names in ilTree

iszmais:
I share the point that a general refactoring is difficult here because of the heavy usage, but the class should at least provide the correct approach to prevent the situation to become worse.

### DIFF
--- a/Services/Tree/classes/class.ilTree.php
+++ b/Services/Tree/classes/class.ilTree.php
@@ -28,15 +28,15 @@ class ilTree
 
     const POS_LAST_NODE = -2;
     const POS_FIRST_NODE = -1;
-    
-    
+
+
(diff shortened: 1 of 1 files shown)


iszmais:
@fwolf-ilias  this also fixed the partial trailing whitespaces, but the main concern is about the change of `...Childs()` functions to `...Children()` functions.

klees:
Hi  @iszmais,

@fwolf-ilias now has the maintainership for ILIAS/Tree.

Kind regards!

kergomard:
Hi @fwolf-ilias 
As I looked through the PRs that have been around for a while without progress, I found this one. Could you have a look at this and either see how to make it progress or close it?

Thank you very much!

PR #7073, opened 2024-02-20
bugfix 40577: SCORM player ignores favicon.ico from custom style

utesche:
This PR is related to issue:  https://mantis.ilias.de/view.php?id=40577

(1)
The SCORM player now respects a given 'favicon.ico' from the active style.

(2)
Additionally a missing icon was detected:
./templates/default/images/scorm/browsed.svg
and is provided hereby as a proposal.

It is linked to from here:
https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/Modules/Scorm2004/classes/class.ilSCORM13PlayerGUI.php#L527


Uwe-Kohnle:
Thank you. browsed.svg would be a proposal and probably changed.

PR #7074, opened 2024-02-20
StudyProgramme: add new config option search recursive 

nhaagen:
https://docu.ilias.de/goto_docu_wiki_wpage_6957_1357.html


utesche:
Hi,
the new column [prg_auto_membership].search_recursive acts as a true/false-flag, right?
If so, I recommend to declare it as 'NOT NULL' to avoid the null value handling and storing by the DBMS.

klees:
@nhaagen As @utesche says.

klees:
Merged as https://github.com/ILIAS-eLearning/ILIAS/commit/2eb8f8770300e3895a758c227013a52d1a882313

PR #7077, opened 2024-02-20
Setup: adjust text for numbers of found migrations. (#40438)

lukastocker:
https://mantis.ilias.de/view.php?id=40438


klees:
Picked to trunk and release_9.

PR #7084, opened 2024-02-21
9 didactic template, fixes ref to class StatusType

chlulei:
Namespace of class StatusType was changed. This fixes the reference to StatusType in ilDidacticTemplateSettingsGUI.


smeyer-ilias (approved):
Thanks @chlulei

PR #7089, opened 2024-02-21
[FIX] 0040540: Unzip behaviour changed in ILIAS 9

chfsx:
Hello @alex40724 , hello everyone else (see below)

thanks for the feedback! i have taken this as an opportunity to replace all existing calls of `ilFileUtils::unzip` with `\ILIAS\Filesystem\Util\Archive\Archives` or `\ILIAS\Filesystem\Util\Archive\LegacyArchives`, because this was only a wrapper anyway. it also allows consumers to make finer settings in the future as to how unzip should behave if it needs adjustments. i have restored the default that NO top-directory is created but the structure is kept. the behavior with ZIP is also relevant, which i have also restored. i have tried to test as many of the locations as possible, so far i have not noticed any errors.

@alex40724  here a hint: in `ilFileUtils` (formerly in `ilUtil`) there was also a method `processZipFile`. this was used exclusively in `ilExSubmission`. i have therefore moved the method there, but i think you can simply replace it later.

here are the affected components:
- Modules/CmiXapi: @Uwe-Kohnle 
- Modules/Exercise: @alex40724 
- Modules/ScormAicc: @Uwe-Kohnle 
- Modules/Survey: @alex40724 
- Services/Certificate: @mjansenDatabay 
- Services/Export: @smeyer-ilias 
- Services/FileServices: @chfsx 
- Services/FileSystem: @chfsx 
- Services/MediaObjects: @alex40724 
- Services/Style/Content: @alex40724 
- Services/Style/System: @Amstutz 
- Services/User: @kergomard 

I would merge this PR next thursday if there is no objection.

Info: Failing tests are due to License-Header, but the header seems correct in the file mentioned
----

Former text of PR:
@alex40724 this should fix the issues in exercise concerning unzip, https://mantis.ilias.de/view.php?id=40540

### DIFF
--- a/Modules/CmiXapi/classes/class.ilCmiXapiImporter.php
+++ b/Modules/CmiXapi/classes/class.ilCmiXapiImporter.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -18,6 +16,8 @@ declare(strict_types=1);
  *
  *********************************************************************/
 
+declare(strict_types=1);
+
 /**
  * Class ilCmiXapiImporter
  *
@@ -30,6 +30,7 @@ declare(strict_types=1);
 class ilCmiXapiImporter extends ilXmlImporter
 {
     private array $_moduleProperties = [];
+    private \ILIAS\Filesystem\Util\Archive\LegacyArchives $archives;
 
     public array $manifest = [];
 
@@ -64,6 +65,7 @@ class ilCmiXapiImporter extends ilXmlImporter
         $this->dic = $DIC;
         $this->filesystemWeb = $DIC->filesystem()->web();
         $this->filesystemTemp = $DIC->filesystem()->temp();
+        $this->archives = $DIC->legacyArchives();
         $this->_dataset = new ilCmiXapiDataSet();
         $this->_dataset->_cmixSettingsProperties['Title'] = '';
         $this->_dataset->_cmixSettingsProperties['Description'] = '';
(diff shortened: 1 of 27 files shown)


alex40724 (changes requested):
@chfsx Thanks for the PR.

I added a comment to the ticket. The main problem is, that this is not exercise specific, but a general issue of the unzip method in ilFileUtils. I added a test case, see https://mantis.ilias.de/view.php?id=40540#c101844

chfsx:
@alex40724 thanks for the feedback, see updated initial text in this PR

chfsx:
@alex40724 I see you made already changes, I will update the PR asap

chfsx:
@alex40724 PR updated

mjansenDatabay (approved):
Thx @chfsx , this looks good for `Services/Certificate`.

kergomard (approved):
Thank you very much @chfsx !
I think I can approve this for the Test and the User.

alex40724:
@chfsx Thanks. I removed all ilFileUtils::unzip calls in my code, so these parts should be obsolete.

chfsx:
@alex40724 thanks, I updated the PR and made a change in your ZipAdapter to match the new situation https://github.com/ILIAS-eLearning/ILIAS/pull/7089/commits/f44887f11ef082771ff9aa93e449d0322b45b5f6

alex40724 (approved):

alex40724:
@chfsx Thanks. It's ok to remove processZipFile from the Util class. I will hope to find the time to properly refactor this after the merge. Worst case is, I will keep it as it is internally.

alex40724:
Btw. I will collect some requirements for the Archives interface for use cases I could (if I did not overlook anything) not implement yet. E.g. adding empty directories in "in memory"/temporary zip archives. But this I would address in ILIAS 10.

chfsx:
Friendly reminder to give feedback to this PR until tomorrow’s evening. I‘ll merge this thursday

PR #7090, opened 2024-02-21
UIComponent: File Upload Input: Replace mime type with file extension

lscharmer:
This PR addresses Mantis bug: https://mantis.ilias.de/view.php?id=39689

This PR proposes a change to the `ILIAS\UI\Component\Input\Field\FileUpload` interface.

Currently the file upload can be given a list of mime types to be accepted.
But as the mantis bug requests, a list of file extensions should be shown to the user when rendering the component, to indicate which file types are allowed.
With `ILIAS\FileUpload\MimeType` we can determine the list of file extensions for a given mime type but this list is not viable to show to a user.
For e.g. `text/plain` this gives us a list of ~26 file extensions (bat, c, c++, cc, com, conf, ... etc.) which will be shown to the user.

For this reason this PR proposes to be able to set the file extensions directly.

Additionally to prevent more complexity, this PR suggests to remove the methods to set and get the mime types and use the file extensions only.

@chfsx Please comment if this PR is not in line with the discussion we had for the file upload.

Furthermore if this PR is accepted follow up PR's will be created to implement these changes and adjust the file renderer accordingly.

### DIFF
--- a/components/ILIAS/UI/src/Component/Input/Field/FileUpload.php
+++ b/components/ILIAS/UI/src/Component/Input/Field/FileUpload.php
@@ -43,12 +43,12 @@ interface FileUpload
     public function getMaxFiles(): int;
 
     /**
-     * @param string[] $mime_types
+     * @param string[] $extensions
      */
-    public function withAcceptedMimeTypes(array $mime_types): FileUpload;
+    public function withAcceptedExtensions(array $extensions): FileUpload;
 
     /**
      * @return string[]
      */
-    public function getAcceptedMimeTypes(): array;
+    public function getAcceptedExtensions(): array;
 }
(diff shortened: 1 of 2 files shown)


klees:
Hi @lscharmer,

thanks for you suggestion. We discussed this with the UI coordinators (@Amstutz, @thibsy and me). We see the problem that for some mimetypes the list of possible file extensions is too long to be actually helpful for our users. We still think that a mimetype carries more information and possibilities that are of good use here. For one, it is about the content of the file, not the name only. Keep in mind that this is not necessarily connected for some users (Linux...). So a mimetype could be used to actually check if the content of a file is as expected. Also, with mimetype we can say: We want some `image/*`.

We would thus suggest to move on in a different direction here: Keep the current mimetype functionality, derive file extensions from the mimetype by default, as we currently do. Add the possibility to override the derived file extension if this is not good enough.

Please open a new PR. Please check in advance if this could actually be implemented with the library that we currently use.

Kind regards!

klees:
Hi @lscharmer,

from an interface perspective, I think, it would be best to just add an additional parameter for the extensions to `withAcceptedMimeTypes`.

Kind regards!

PR #7091, opened 2024-02-22
UI/JS: adjust & move js for input, popover, filter

nhaagen:
'fix', i.e. _move_ JS to public section.


klees:
I will take care of this here: https://github.com/ILIAS-eLearning/ILIAS/pull/7019

PR #7092, opened 2024-02-22
T&A: Remove Typo in de.lang

dsstrassner:
https://mantis.ilias.de/view.php?id=40475


matthiaskunkel:
Thanks @dsstrassner for the fix!

PR #7096, opened 2024-02-23
Administration: Fix new item group table array access

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=37799

if approved, this has to be picked to `release_9` and `trunk`.


nicoroeser:
Thank you, @mjansenDatabay!

This patch looks good to me on first sight. @alex40724 is assigned to the Mantis issue report and responsible for that code part, so he’ll probably comment on this, merge, etc.

alex40724:
Thanks, fixed differently.

PR #7097, opened 2024-02-23
Fix of #39992 und #37178

Amstutz:
Currently we have 2 Issues spawning from different A11y Audits, #https://mantis.ilias.de/view.php?id=39992 and https://mantis.ilias.de/view.php?id=37178 .

In its core, both are caused by the counters being missed by screenreaders if they are placed within bulky buttons. Cause of the issue is the role="img" given to glyphs if they are rendered in the context button. role="img" leads to all descendents of the element being "presentational", see: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role .

"Accessibility APIs do not have a way of representing semantic elements contained in an img. To deal with this limitation, browsers, automatically apply role [presentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role) to all descendant elements of any img element as it is a role that does not support semantic children."

The recommendation is: "An accessible name is required. For the HTML [](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img) element, use the alt attribute. For all other elements with the img role, use aria-labelledby if a visible label is present, otherwise use aria-label."

I opted for extending adding an aria-label of the glyph carrying the role="img". So screenreaders should catch the counters again.

### DIFF
--- a/lang/ilias_de.lang
+++ b/lang/ilias_de.lang
@@ -3773,6 +3773,8 @@ common#:#copy_of#:#Kopie von
 common#:#copy_of_suffix#:#- Kopie
 common#:#copy_selected_items#:#Kopieren
 common#:#count#:#Anzahl
+common#:#counter_novelty#:#Neuigkeiten
+common#:#counter_status#:#Status
 common#:#country#:#Land
 common#:#country_free_text#:#Land (Freitext)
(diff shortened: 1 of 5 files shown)


klees:
Picked to trunk.

PR #7098, opened 2024-02-23
T&A 0040488 save and return' returns to question list instead of question page for questions from pool test 

kunzeus:
https://mantis.ilias.de/view.php?id=40488

Change the redirect after sync and sync cancellation.

Tested on: ILIAS 8 
Same code: ILIAS 7, 8, 9


kergomard:
Thank you very much for the PR @kunzeus ! Merged and picked to ILIAS 7, 9 and trunk.

Just one small comment (I changed this myself): Please use `::class` instead of a simple string, to the the name of a class.

Best,
@kergomard

PR #7099, opened 2024-02-23
T&A 0040614: Starting test with "Use Previous Answers" activated leads to fatal type error

kunzeus:
https://mantis.ilias.de/view.php?id=40614

Add a null check to prevent the crash.

tested on: ILIAS 8
Same code on: ILIAS 8, 9


kergomard:
Thank you very much @kunzeus for this PR.

I would have a little change request though: Could you please move the check to the function `assQuestion::authorizedSolutionExists()`? To me this sounds more logical. If there is no pass, no authorizedSolution exists and thus it should also return false there. We should not have to check this first. Clearly the parameter declaration has to change then, too.

Thanks again and best,
@kergomard

PR #7102, opened 2024-02-27
9/UI/legacy-toolbar refactoring, removing hamburger collapse 40080

catenglaender:
# Issue

The toolbar shows visual glitches - one of the most notable ones: https://mantis.ilias.de/view.php?id=40080
While investigating this we discovered (as discussed in the CSS Squad):
* the code of the toolbar is a quite messy chain of overrides
* the hamburger menu that some buttons collapsed into confused many users as it rivals the three dots in the metabar and the mainbar navigation

image

image

# Changes

* Jour Fixe, 19 FEB 2024: We prefer to remove the hamburger menu from the mobile view with 9 and to show all toolbar buttons always. => collapse/expand feature has been removed
* visual glitches have been removed by using the general SCSS layout tool element-bar instead of code unique to the toolbar (this also saves many lines of codes)
* refined the element-bar tool to better handle nested groups

image

image

# Impact

Since we wanted to reduce the amount of custom code for the toolbar, the toolbar inherits new spacings from the elementbar and therefor doesn't look exactly like it did before.

While we checked quite a few toolbar implementations, some side effects may occur. Please report in Mantis if you notice anything.


Services/UIComponent/Toolbar/classes/class.ilToolbarGUI.php
```diff
@@ -333,6 +328,7 @@ public function getHTML(): string
                     $markup_items .= $tpl_separator->get();
                 }
                 foreach ($group as $item) {
+                    $tpl_items->touchBlock("item");
```
catenglaender: To be honest, I do not understand why it needs this touchBlock at this front position to work. If I put the touchBlock last, some empty items are rendered; if I don't put it here, many items are missing. Can someone please explain this to me (@nhaagen)?

catenglaender:
I broke it... @nhaagen needs to help me investigate :grimacing:

catenglaender:
Thanks to @nhaagen for helping me untangle the grouping of toolbar items! :tada: 
This is ready for further reviews/merge now :slightly_smiling_face:

Amstutz:
Thx a lot for this important fix. Further thx for massively reducing our css code being used! Lastly, thx for adapting the unit-tests

PR #7103, opened 2024-02-28
Fix error in rbac log table

sroenchen:
Fixes a missing integer cast in the role log table, that can sometimes result in a hard error.

Mantis: https://mantis.ilias.de/view.php?id=40664


kergomard:
Thank you very much for the PR @sroenchen ! This is ILIAS 8 only.

PR #7104, opened 2024-02-28
Fix failing pdf uploads du to page count preprocessor

sroenchen:
Pdf uploads can currently fail due to either an invalid array key or a type error (if null was returned).

Mantis: https://mantis.ilias.de/view.php?id=40665


chfsx:
Thanks a lot for the PR and the Mantis-issue, i used a lightly different syntax and adapted in other locations as well.

PR #7106, opened 2024-02-28
9/UI/viewcontrols nav arrows shifted for section 40648

catenglaender:
Mantis: https://mantis.ilias.de/view.php?id=40648

# Issue

Viewcontrol section navigation glyph arrows appear shifted in Firefox:

image

This is because two nested vertical-align declarations seem to be rendered differently in Firefox vs. Chromium based browsers.

# Change

Removed the offending vertical-align overrides and discovered that the buttons work fine with just the UI component glyph's default values. 

image

It's not pixel perfect, there could be +/- a pixel here and there, but I would rather tweak the UI component glyph in general to ensure a perfect centered vertical align using flexbox rather than bloating the consumer code with overrides.

### DIFF
--- a/templates/default/070-components/UI-framework/ViewControl/_ui-component_viewcontrol.scss
+++ b/templates/default/070-components/UI-framework/ViewControl/_ui-component_viewcontrol.scss
@@ -72,25 +72,6 @@ $il-vc-pagination-btn-active-bg: $il-main-bg;
 		display: inline-block;
 	}
 }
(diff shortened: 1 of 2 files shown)


nicoroeser:
Thank you, @catenglaender, for your attention to detail!

@Amstutz, can you please take over from here? Thanks!

Amstutz:
Fixing issue by removing lines of CSS. Loving it.

PR #7107, opened 2024-02-28
User/MemberGallery: Fix missing user actions

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40666


kergomard:
Thank you very much @mjansenDatabay for tracking the issue down and for this PR! Merged and picked to trunk.

PR #7108, opened 2024-02-28
AdvMD: redesign form for editing selects for better stability

schmitz-ilias:
This PR makes some fundamental changes to the form for creating and editing Custom Metadata (multi-)select fields. The aim is to make working with these fields more robust and foolproof.

Currently, the entries of select fields can be edited in the Administration (Search And Find > Metadata > Custom Metadata > Edit Fields > Edit) via a multi-text input:

Screenshot from 2024-02-28 16-16-43

This input allows the user to simultaneously move, rename and/or delete as many entries as required, as well as add arbitrarily many new ones. From this wide array of possible simultaneous manipulations it is almost impossible to reliably reverse-engineer user intent, and make the expected changes to the underlying data. In the past, this has contributed to sometimes data-destroying errors (see e.g. [40590](https://mantis.ilias.de/view.php?id=40590) for the newest such occurrence), and has made Custom Metadata select fields more unreliable in general.

This PR keeps the multi-text input when creating a new select field, but replaces it when editing an existing field: instead, a radio input is used, with one radio for each entry. Each radio offers in a subform a text input for changing the entry's value, a dropdown for changing the entry's position, and a checkbox to delete the entry.

Screenshot from 2024-02-28 16-29-50

The final radio allows for adding a new entry, and does not offer the 'delete' checkbox:

Screenshot from 2024-02-28 16-31-12

Restricting manipulations to a single entry at a time takes the guesswork out of processing the form's output, and should contribute to a higher reliability for these kinds of metadata.

This construction is a bit unusual and as far as I can tell largely unprecedented in ILIAS. Due to the unique architecture of the Custom Metadata component, one is fairly restricted in what improvements one can make without refactoring large parts of it. In my opinion, what can be gained in stability with the changes proposed here is well worth a slightly clunky form.

### DIFF
--- a/Modules/Wiki/classes/class.ilPCAMDPageList.php
+++ b/Modules/Wiki/classes/class.ilPCAMDPageList.php
@@ -303,7 +303,7 @@ class ilPCAMDPageList extends ilPageContent
                 }
             }
             $fields = array(
-                "sdata" => array("text", serialize(serialize($updated_data)))
+                "sdata" => array("text", serialize(empty($updated_data) ? '' : serialize($updated_data)))
             );
             $primary = array(
                 "id" => array("integer", $row["id"]),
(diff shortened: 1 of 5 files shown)


matthiaskunkel:
Jour Fixe, 04 MAR 2024: We consider this as an extraordinary solution born out of the necessity to fix a serious problem in the current implementation of the Advanced Metadata, especially because UI changes are made by using legacy UI components we want to get rid off. We accept the PR for 8 and 9 but recommend Tim to contact the UI Clinic to find a solution to fix this issue for trunk by using KS elements.

schmitz-ilias:
merged and picked to 9 and trunk

PR #7109, opened 2024-02-29
LDAP: Fix missing cast when determining "Auth Mode Name"

mjansenDatabay:

Mantis Issue: https://mantis.ilias.de/view.php?id=40676


mjansenDatabay:
I'll merge this because it is a trivial fix.

PR #7112, opened 2024-03-01
[FIX] 0040136: Failed test: Execute `status` command before installation

chfsx:
See comment in https://mantis.ilias.de/view.php?id=40136


klees:
Picked to trunk.

PR #7115, opened 2024-03-01
Auth: Handle race conditions when writing sessions

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40631

### DIFF
--- a/Services/Authentication/classes/class.ilSession.php
+++ b/Services/Authentication/classes/class.ilSession.php
@@ -115,18 +115,13 @@ class ilSession
         return 0;
     }
 
-
-    /**
-    * Write session data
-    *
-    * @param	string		session id
-    * @param	string		session data
-    */
     public static function _writeData(string $a_session_id, string $a_data): bool
     {
         global $DIC;
 
+        /** @var ilDBInterface $ilDB */
         $ilDB = $DIC['ilDB'];
+        /** @var ilIniFile $ilClientIniFile */
         $ilClientIniFile = $DIC['ilClientIniFile'];
 
         if (self::isWebAccessWithoutSessionEnabled()) {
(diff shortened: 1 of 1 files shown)


Services/Authentication/classes/class.ilSession.php
```diff
@@ -143,54 +138,66 @@ public static function _writeData(string $a_session_id, string $a_data): bool
...
-                $fields["context"] = array("text", ilContext::getType());
+            if (class_exists('ilContext')) {
+                $fields['context'] = [ilDBConstants::T_TEXT, ilContext::getType()];
             }
 
-            $ilDB->insert("usr_session", $fields);
+            try {
+                $ilDB->insert('usr_session', $fields);
+            } catch (PDOException $e) {
+                if ($e->getCode() === 23000) {
```
mjansenDatabay, line 176: `23000` is the `SQLState` code for a duplicate key constraint violation. Of course this could be extracted to a PHP constant.

> Error number: 1022; Symbol: [ER_DUP_KEY](https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_dup_key); SQLSTATE: 23000

Services/Authentication/classes/class.ilSession.php
```diff
@@ -25,6 +25,8 @@
 */
 class ilSession
 {
+    private const DUPLICATE_SESSION_DB_ERROR_CODE = 2300;
```
mjansenDatabay: `23000` is the `SQLState` code for a duplicate key constraint violation.

> Error number: 1022; Symbol: [ER_DUP_KEY](https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_dup_key); SQLSTATE: 23000

Services/Authentication/classes/class.ilSession.php
```diff
@@ -143,54 +140,68 @@ public static function _writeData(string $a_session_id, string $a_data): bool
...
+                }
+
+                $ilDB->replace(
+                    'usr_session',
+                    [
+                        'session_id' => [ilDBConstants::T_TEXT, $a_session_id]
+                    ],
+                    $fields
+                );
+            }
```
pascalseeland: Instead of trying to work around an exception in user code, I would prefer to use $ilDB->manipulate with an insert ... on duplicate key clause. Otherwise we could look at extending the $ilDB->insert to facilitate this case.

pascalseeland (changes requested):
Looks fine besides the decision if we want to work around the ilDB interface or not.

- [x] I would  vote for working around it with manipulateF

mjansenDatabay:
@pascalseeland I switched to an `INSERT INTO ... ON DUPLICATE KEY UPDATE ...` approach.

mjansenDatabay:
Merged, with all checklist items completed!

PR #7118, opened 2024-03-01
Bugfix/Improvement: 0040683 Added info text. Changed modal behaviour.

playamongthestars:
https://mantis.ilias.de/view.php?id=40683


components/ILIAS/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -5947,15 +5953,14 @@ private function doHistoryCheck(int $draftId): void
                     IL_CAL_DATETIME
                 ));
                 $this->ctrl->setParameter($this, 'history_id', $history_instance->getHistoryId());
-                $header = $history_date;
+                $header = $history_date.': '. $history_instance->getPostSubject();
```
mjansenDatabay:
```suggestion
                $header = $history_date . ': ' . $history_instance->getPostSubject();
```

PR #7119, opened 2024-03-01
Fix/9/dcl migration issues

chfsx:
@Amstutz this PR fixes to issues related to the DCL file migration:

- https://mantis.ilias.de/view.php?id=39403
- https://mantis.ilias.de/view.php?id=39087


Amstutz:
thx a lot!

PR #7120, opened 2024-03-01
new language variable for removing members from a mailing list

fneumann:
Chris Potter would like to change the 'Are you sure' message for removing members from a mailing list. The language variable, however, is also used for deleting mailing lists themselves. So we need to introduce a separate language variable.


mjansenDatabay (changes requested):
Thx @fneumann , could you please provide a German translation as well?

mjansenDatabay (approved):

PR #7121, opened 2024-03-01
WAC: Send 404 if resource could not be found

mjansenDatabay:
This PR adds a workaround for: https://mantis.ilias.de/view.php?id=40611 . I doubt that it will be merged without changes.

There are two cases when the WAC throws a `ilWACException` with code `ACCESS_DENIED`.

1. If the realpath check in `\ilWACPath::normalizePath` fails
2. If `\ilWebAccessCheckerDelivery::deny` is called and the internal state of `ilWebAccessChecker` is **not** `checked`

So the problem to handle 1. is, that the `ilWACPath` instance is created in the constructor of `ilWebAccessChecker`, which again is created in the constructor of `ilWebAccessCheckerDelivery` (called by `\ilWebAccessCheckerDelivery::run`).

To be able to handle 1. properly, **all** these instances MUST be created (which is not the case if an exception is thrown) and `\ilWebAccessCheckerDelivery::handleRequest` must be called for proper response handling in the context of the existing `ilWebAccessCheckerDelivery` instance.

So my approach is to catch the exception in `\ilWebAccessChecker::__construct` and store it in the internal state of the `WAC`. If `\ilWebAccessChecker::check` is called afterwards and no path is given, the state is checked and and the exception is raised again, so it can be properly handled by `\ilWebAccessCheckerDelivery::handleRequest`.

Furthermore, I added a new method to handle the "Not Found" error by just sending a `HTTP 404` error without the "Dummy Images". Of course this is also debatable.

----

I guess @chfsx will implement another approach. I am sure this will require some restructuring of the component.

### DIFF
--- a/Services/Exceptions/classes/class.ilException.php
+++ b/Services/Exceptions/classes/class.ilException.php
@@ -27,8 +27,8 @@ class ilException extends Exception
     /**
      * A code isn't optional as in build in class Exception
      */
-    public function __construct($a_message, $a_code = 0)
+    public function __construct($a_message, $a_code = 0, Throwable $previous = null)
     {
-        parent::__construct($a_message, $a_code);
+        parent::__construct($a_message, $a_code, $previous);
     }
 }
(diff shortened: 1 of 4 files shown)


chfsx:
Thanks a lot @mjansenDatabay, I merged that locally with small changes and adoptions for PHP syntax in ILIAS 7 up to trunk

PR #7124, opened 2024-03-01
Refinery: Encoding

chfsx:
This PR adds a UTF8 string transformation to the refinery which can be used to replace `utf8_encode` calls in ILIAS, see https://github.com/ILIAS-eLearning/ILIAS/commit/9f1b9b68d97943050fb74c5ca8e4663196f4ac3f

examples how to use that are in https://github.com/ILIAS-eLearning/ILIAS/commit/b3f364b97648fe69c331dc039e1c381646f5192f which fixes the issues:

https://mantis.ilias.de/view.php?id=40543
https://mantis.ilias.de/view.php?id=40542


chfsx:
a second `Transformation` for `ISO-8859-1` should be introduced as well to replace `utf8_decode` calls.

chfsx:
Fixed tests in https://github.com/ILIAS-eLearning/ILIAS/pull/7124/commits/023d88838ebb38db9e04295529f93758eb5e74dc

mjansenDatabay (changes requested):
Hi @chfsx ,

thank you a lot for contributing to the `Refinery`.

- [x] Naming 1: `utf8` sounds more like a "Group" than a transformation. In PR #6314 (it is still open) we suggest to introduce a "Group" named `encoding` (currently it is `escape`). Please move the "UTF-8" related transformations to this group, or, if this PR is urgent, create/introduce such a group with this PR.
- [x] Detection: We would currently like to forego the encoding detection of `$from` unless there are already real use cases for it. Detecting the encoding by using `mb_detect_encoding` is just a heuristics (_... Detects the most likely character encoding ..._). So please remove the class member `$detect_encoding` (it is hard coded anyway) if we don't need this at the moment.
- [x] Naming 2: It is also very implicit that `utf8` by default assumes that a "Latin1" string (`$detect_encoding` is set to `false`) is "UFT-8" encoded. Please rename the method to `latin1ToUtf8`. We should try to aim for names like "$FromTo$To" (or if we indeed need a detection in future: "to$To").

Kindly,
@klees and @mjansenDatabay

chfsx:
Thank you very much for the quick feedback!

yes i think it is a bit urgent, because i would like to remove the deprecations of PHP8.2 before ILIAS 9 is finished. therefore i have made the (very useful!) changes as requested:

- Naming 1: new is encoding a group, two methods are now offered, `latin1ToUtf8` and `asciiToUtf8`, as these are probably the most common according to a short research.
- Detection: is now abolished and you explicitly select from and to by choosing the method on the group
- Naming 2: is adapted accordingly, see above

mjansenDatabay (approved):

klees:
Hi @chfsx,

didn't have this on my radar. Feel free to ping me and/or @mjansenDatabay via comment next time.

Will have a look into this now.

Kind regards!

klees (approved):

klees:
I have merged this manually after resolving conflicts. Thanks a lot! Good stuff.

klees:
Picked to trunk!

PR #7126, opened 2024-03-04
PRG: 40674, relevance as single action / do not change relevance of successful progress

nhaagen:
https://mantis.ilias.de/view.php?id=40674


klees:
Picked to trunk.

PR #7131, opened 2024-03-04
7 mantis ilias 40106, adds dataset namespace default value

chlulei:
Link: https://mantis.ilias.de/view.php?id=40106


smeyer-ilias:
Thanks @chlulei

PR #7138, opened 2024-03-05
Set "current version" after update

iszmais:
https://mantis.ilias.de/view.php?id=40712

### DIFF
--- a/Services/Component/classes/class.ilPluginDBUpdate.php
+++ b/Services/Component/classes/class.ilPluginDBUpdate.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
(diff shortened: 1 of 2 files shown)


klees:
@chfsx Want to have a look?

klees (approved):
LGTM. Solution is a little odd because of the empty method in ilDBUpdate. But this IMO is fine for the moment, as both classes will vanish some time soon.

klees:
Hi @chfsx,

I have just merge this, contact me if we need to change something.

Kind regards!

klees:
Picked to trunk. Thanks @iszmais!

PR #7140, opened 2024-03-06
LSO: 39159, parent type for LSOPages (Intro/Extro), Migration and WAC

nhaagen:
- DB Update inserting type into copg_pobj_def
- Migration to update IDs of PageObjects 
(php setup/setup.php migrate --run ilLearningSequenceMigrationAgent.LSOMigratePageIdsExtro)

https://mantis.ilias.de/view.php?id=39159

### DIFF
--- a/Modules/LearningSequence/classes/Setup/class.LSOMigratePageIdsExtro.php
+++ b/Modules/LearningSequence/classes/Setup/class.LSOMigratePageIdsExtro.php
@@ -20,7 +20,6 @@ declare(strict_types=1);
 
 use ILIAS\Setup;
 use ILIAS\Setup\Environment;
-use ILIAS\Setup\Objective\MigrationObjective;
 
 class LSOMigratePageIdsExtro implements Setup\Migration
 {
@@ -45,10 +44,7 @@ class LSOMigratePageIdsExtro implements Setup\Migration
     public function getPreconditions(Environment $environment): array
     {
         return [
-            new ilDatabaseInitializedObjective(),
-            new MigrationObjective(
-                new  LSOMigratePageIdsIntro()
-            )
+            new ilDatabaseInitializedObjective()
         ];
     }
 
--- a/Modules/LearningSequence/classes/Setup/class.LSOMigratePageIdsExtro.php
+++ b/Modules/LearningSequence/classes/Setup/class.LSOMigratePageIdsExtro.php
@@ -65,7 +65,7 @@ class LSOMigratePageIdsExtro implements Setup\Migration
         $result = $this->db->query(self::QUERY . ' LIMIT 1');
         $row = $this->db->fetchAssoc($result);
         $query = 'UPDATE page_object' . PHP_EOL
-            . "SET page_id = (parent_id * -1), parent_type = 'lso'" . PHP_EOL
+            . "SET page_id = parent_id, parent_type = 'lsoe'" . PHP_EOL
             . "WHERE page_id = " . $row['page_id'] . PHP_EOL
             . "AND parent_id = " . $row['parent_id'];
         $this->db->manipulate($query);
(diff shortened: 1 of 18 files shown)


klees:
Hi @alex40724,

would you be so kind and have a look into this? There are two new migrations in the LSO, but these basically will write into tables that belong to COPage, which is why I think you should have a look.

The state of the art is as such:

* For the LSO we have been reusing the standard container page for the intro and extro pages. To be able to have to these pages (instead of one...) we have synthesized page ids via the id of the parent object (-1 * parent_id, -2 * parent_id).
* This seems to have been creating issues with the WAC, as described in the linked ticket.
* We would like to solve this by introducing a new page type for the LSO, which would require us to relabel the existing pages of the LSO. Luckily, we can identify these by our odd id-scheme.

I will ask @nhaagen to have another look into [your COPage docu](https://github.com/ILIAS-eLearning/ILIAS/blob/release_9/Services/COPage/README-technical.md).

I would still ask you if you would be ok with us writing in the table of COPage via migration. Also, other input or feedback would be appreciated.

Kind regards!

nhaagen:
I'll revise this to separate Intro/Extro by type rather than (negative) id; until then, I set the PR to "draft".

alex40724:
@nhaagen Please have a look at https://github.com/ILIAS-eLearning/ILIAS/commit/7c55ed13b7a226592168ed842d0b2f5685e5177d

I added some docu on setting the parent ID of pages. Setting this to the repo object ID of the context enables a default behaviour for WAC checking in ilObjMediaObject. As far as I can see, LSO already sets this ID correctly, so I assume no additional change is needed here and we can omit the ilObjMediaObject part of this PR.

EDIT: I also added docu on import/export, since this was definitely missing.

Modules/LearningSequence/classes/Setup/class.LSOMigratePageIdsExtro.php
```diff
@@ -0,0 +1,77 @@
...
+     * @throws Exception
+     */
+    public function step(Environment $environment): void
+    {
+        $result = $this->db->query(self::QUERY . ' LIMIT 1');
+        $row = $this->db->fetchAssoc($result);
+        $query = 'UPDATE page_object' . PHP_EOL
+            . "SET page_id = parent_id, parent_type = 'lsoe'" . PHP_EOL
+            . "WHERE page_id = " . $row['page_id'] . PHP_EOL
+            . "AND parent_id = " . $row['parent_id'];
```
alex40724: Please always add the current parent type to the WHERE clause, since this is the important part of the primary key "... AND parent_type='...'".

Modules/LearningSequence/classes/Setup/class.LSOMigratePageIdsIntro.php
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public function step(Environment $environment): void
+    {
+        $result = $this->db->query(self::QUERY . ' LIMIT 1');
+        $row = $this->db->fetchAssoc($result);
+        $query = 'UPDATE page_object' . PHP_EOL
+            . "SET page_id = parent_id, parent_type = 'lso'" . PHP_EOL
+            . "WHERE page_id = " . $row['page_id'] . PHP_EOL
+            . "AND parent_id = " . $row['parent_id'];
+        $this->db->manipulate($query);
```
alex40724, line 68: Please always add the current parent type to the WHERE clause, since this is the important part of the primary key "... AND parent_type='...'".

alex40724 (changes requested):

nhaagen:
Hi @alex40724 , thank you for the hints. 
I incorporated them and rebased to catch up with ilObjMediaObject.

alex40724 (approved):

klees:
Picked to trunk.

PR #7142, opened 2024-03-06
PRG: 40519, no deadline if latest assignment is not in progress

nhaagen:
https://mantis.ilias.de/view.php?id=40519


klees:
Picked to trunk.

PR #7024, opened 2024-02-07
UI: Update Drilldown to Show Filter and Two columns

kergomard:
This PR is a follow up on #6889 after the feedback. It proposes to update the Drilldown Menu to look like shown in the print screens below.
There is actually no change to the interface, it just updates the description shown in the factory in [this commit here](https://github.com/ILIAS-eLearning/ILIAS/commit/50c069eec3b9f0f89bdd72fdba69e7e2fff51918) 

Further:
- It proposes to introduce a general variable column-standard-width to be used to calculate certain width and defines the Modal width accordingly in [this commit here](https://github.com/ILIAS-eLearning/ILIAS/commit/fe901c2431a5b4b10f132b6a14a158f76634e84e). This way it becomes easier to see when more than one column is needed.
- The implementation still has a few hard coded colors and similar issues, it is additionally based on the current Javascript that is not easy to handle, but I would much appreciate a first feedback to see if this is going in the right direction. If you agree, I would then rework the Javascript into ES6 and give it a simpler structure.

Thank you very much and best,
@kergomard 

grafik
----
grafik
----
grafik

PR #7143, opened 2024-03-06
[FEATURE] UI: `Field\TreeSelect` and `Field\TreeMultiSelect` inputs

thibsy:
Hi folks,

It's been a while, but here's the first proposal for the public interface of the resource selector input!

The factory description is derived from the last state of the prototype created in #6320. Since the look and feel has already been discussed there, I won't go over it in detail here. Instead, let's dive into the more logical aspects of the interface:

Following multiple discussions in the UI-Clinic and the implementation of the prototype, we have decided that a `Menu\Drilldown` menu is the most feasible approach for navigating the hierarchical dataset at the moment.

To maintain flexibility, I have kept the interface of the `Field\ResourceSelector` decoupled from the `Menu\Drilldown` menu, so we can easily swap out the implementation to a different approach. This also lays the groundwork for e.g. adding alternative views, tailored for different devices.

For handling large datasets and rendering sup-parts asynchronously, I implemented a monoid-ish approach for nesting `Field\Resource`s: The `Field\ResourceContainer` represents a `Field\Resource` itself, while providing an operation to nest further `Field\Resource`s. In order to indicate what sup-parts of the dataset need to be rendered asynchronously, I have introduced a `withRenderChildrenAsync()` method on the `Field\ResourceContainer`, as it appeared to be the most suitable location.

Using the current approach for rendering sub-parts will need the UI framework to have its own endpoint. This is because unlike other components, the `Field\ResourceSelector`s sub-parts cannot simply be rendered as "HTML over the wire", because this would mean that we need to expose the somewhat special kind of `Menu\Drilldown` menu on the public interface. Besides that, this would again tightly couple usages to the drilldown approach, making it hard to exchange and impossible to automatically handle device-specific views.

Please note the extension of `Field\HasDynamicInputs` in the `Field\ResourceSelector`. I added this as a fallback, because its usage depends on further implementation details and developments in how we want to handle asynchronously rendered sub-parts of a dataset. There is a possibility we want to render sub-parts on the client using a template.

Let me know if you have any adhoc questions, otherwise I am looking to your review.

Kind regards,
@thibsy


thibsy:
Screenshot 2024-03-06 at 14 16 43

Screenshot 2024-03-06 at 14 17 40

thibsy:
Btw, this could also play along nicely with the proposed change to `Menu\Drilldown` components in #7024.

yvseiler:
Thanks a lot for your work @thibsy , it will be very exciting to see the Resource Selector in action. I could follow everything so far except your section here:

> Using the current approach for rendering sub-parts will need the UI framework to have its own endpoint. This is because unlike other components, the Field\ResourceSelectors sub-parts cannot simply be rendered as "HTML over the wire", because this would mean that we need to expose the somewhat special kind of Menu\Drilldown menu on the public interface. 

What are the implications of this? I read in there that the Resource Selector component should not be coupled with the drilldown component (for reasons), but what does "will need the UI framework to have its own endpoint" mean? Thanks for clarification.

thibsy:
Hi @yvseiler, 

Asynchronous rendering currently relies on consumers providing their own endpoints in form of URLs, which are used by the UI framework to fetch HTML content ("HTML over the wire") that is used to modify/replace the existing UI. However, for the resource selector, following the same concept would tightly couple each endpoint to the concept of rendering drilldown menu entries (sub-parts of the dataset), and would also require these components to be available on the public interface - which is not something we want.

Another approach could be to follow the same concept, but instead of HTML the client could fetch sub-parts in form of JSON data, which can be used to render the drilldown menu entries using a template. Since our framework relies on server-side-rendering, I suggested yet another approach, by introducing a dedicated endpoint that could look something like this: `https://domain.com/ui.php?foo=bar`. Having its own endpoint is something we ultimately want, so I suggested to make it happen with this component :). Doing so would also ensure, that the JSON response coming from our endpoint will always be structured the way we want it to be. Using consumer endpoints will never guarantee this.

Hope this helps!

Kind regards,
@thibsy

kergomard (commented):
Thank you very much for the proposal @thibsy !

As discussed on Discord I would suggest to clarify the behavior in the case of a tree containing nodes that are readable, but not visible to a user. I think we should try to keep the behavior consistent across all usages. Thus I propose to add a usage-rule:
`
4: >
    if a node of the hierarchical data set is readable to the current user, but not visible, this node MUST not be shown in the resource selector. If a sub-node of a hidden node is visible to the current user, this node MUST be shown in the resource selector.
`

Thank you very much and best,
@kergomard

Amstutz:
short update here: @klees reviewed it - ill will give this another look and we will ship one review together. In short, we are on it and make progress.

Amstutz:
Hi @thibsy!

Thank you so much for this nice UI element. Looks solid in general, but we could work on some aspects, as most often =) Especially the naming and wording seems to be something that we could really improve. It could be worth a live discussion. Note the @klees and @Amstutz also had some discussion around one and the other question and we are looking forward to your opinion.

- [x] Drilldown a: Is the "Text Input" rival actually correct? The Drilldown Menu recently gained a search functionality. Maybe leave out, since it is a legacy component anyway.
- [x] Drilldown b: If we understand correctly, the Drilldown will be used in the implementation behind the public interface, correct?
- [x] container: The [description of container](https://github.com/ILIAS-eLearning/ILIAS/pull/7143/files#diff-92b2518bdccdd8219e87c90ba2d22077d51c860d5e2f6d2a9edf30766efd6f38R55) is really hard to understand. How does this abstractly represent an object? What even is an object in this context? Maybe this would also benefit from a rewording to tree (see first item in next paragraph) as well. A resource then would be a "leaf" and a container would be a "node" and we could just drop this "represents" wording etc. all together.
- [x] ResourceRetrieval::__construct: Having a constructor on an interface is really weird because a constructor can never be called on an abstract interface but only on concrete classes. Still, the reason why we would need this in the current implementation is comprehensible. Did you try to get rid of that constructor by leveraging the new possibilities of the component revision? This looks as if there should be a way to `contribute` retrieval objects in some way to move that construction back to the component it belongs to.

- [x] purpose: Please go into a little bit more detail what a "resource" is supposed to be. The selector seems to be the right choice when "resources" are structured via a tree that can be traversed to discover "resources". A resource could e.g. be an object in the repository, a file in the filesystems directory structure or a term in a taxonomy. This input would not work well for cases where the "resources" are a longish flat list, where each "resource" is "tagged" in some way. So e.g. selecting a user from a member list would not be a very good use case, right? We think this should be elaborated in the "purpose" description. We are missing "tree" there and an explanation of the vague term "resource".
- [x] accessibility: We would suggest to be a little less explicit regarding the "effect" of selecting a resource. Darkening color to indicate "disabled" could be a problem for accessibilty due to contrast issues. Maybe something like "the actual drilldown menu entry will indicated that it cannot be selected anymore" is enough and leaves some wiggle space for a accessible solution.
- [x] usage.1/2: We are not convinced that these rules are actually helpful. usage.1 uses the vague term "discoverability", thus we won't be able to decide anything based on that rule. usage.2 intends to mirror the "Text Input" rival, right? With regards to the first item in this list, rule 1 might be better phrased 
- [x] factory params: Please put the `ResourceRetrieval` as last parameter. We use that order for the Data Table as well and some consistency is always nice.
- [x] ResourceRetrieval1: Could we name this `DataRetrieval` as well? We will likely be using this pattern and should form some convention here. The current convention would be to use generic `Data`, while this seems to to suggest that the `Retrieval` is named after the thingy it returns. This should be one or the other, but not both.
- [x] Resource::getId: Could we use `string|int` as return type?
- [x] Resource::getIcon: This method indicates that we would get an `Icon` but we would get a `Glyph` instead. Something is off here, either the name or the return type. Seems to be the return type, but we are not 100% sure. The same thing is [in the factory](https://github.com/ILIAS-eLearning/ILIAS/pull/7143/files#diff-92b2518bdccdd8219e87c90ba2d22077d51c860d5e2f6d2a9edf30766efd6f38R75).
- [x] Async: Please promote the split between sync and async containers to the factory, and maybe even to the level of the interfaces that are involved. Currently, I would be asked to provide resources to an async container via the fourth parameter in `ResourceFactory::container`, just to basically discard them by marking the container as async via mutator later on. In that case, if we understand the interface correctly, that resources would need to be loaded again in another process once the front end decides to load them via the `ResourceRetrieval`, right?
- [x] MultiSelect: Please consider splitting single and multiselect selectors via the factory (and classes etc...). For one, I could imagine that we might want to have drastically different components for both cases, since requirements for the user interface seem considerable different. Also, we change the original content of the input via a mutator. So if we have some transformations attached to the input, these could become invalid when using the mutator.

- [x] context: There is some copy pasta there that we should remove. This is not the Rating Input =)
- [x] composition: In its collapsed form, where we only represent the input as a Shy Button, the Input should indicate which resources are currently selected without requiring the user to open the modal to do the selection. This is missing in the "composition" and "effect".
- [x] title: Please rename `Resource::getTitle` to `Resource::getName` (and in other places accordingly). According to [previous research](https://github.com/ILIAS-eLearning/ILIAS/pull/6107

Shall we discuss the current state, our review and the course of action?

@klees and @Amstutz

thibsy:
Hi @Amstutz, Hi @klees

Thanks a lot for reviewing this! A lot of good inputs in there :).

I have changed the overall wording to address several of your inputs, so the tree data structure is better reflected now. I tried to stick with a wording we are already familiar with for the inputs themselves, which are now called `Field\TreeSelect` and `Field\MultiTreeSelect`. These names kind of mirror the already existing `Field\Select` and `Field\MultiSelect` inputs. Please note that we also have a `Field\ColorPicker` input, which IMO could also benefit from a rename to `Field\ColorSelect` someday.

I have put the node-related interfaces into a new namespace, because I did not want to crowd the `Input\Field` namespace too much and it made interface names a bit simpler. Similar things have also been done for e.g. interruptive items for modals. However, in this case I am not sure if `Field\Node` is the right place. Maybe we could also pull this into `Input\Node` or similar. I would have liked to use the existing `Tree\Node` interfaces, but this would have meant a larger refactoring of the tree component as well, which would make this project even bigger than it already is - so I sticked with implementing some new interfaces.

Let me answer some of your questions/remarks now:

- Drilldown a: the text input rival could be outdated, I merely added this because I remembered a discussion we had in one of the workshops about this component, that a tree select inputs like proposed here should not be used if the user exactly knows what they are searching for. However, since we have gained this functionality in the drilldown menu as well, I think we can just leave it out now.
- Drilldown b: yes and no, some version of the drilldown menu will be used to implement the tree select inputs, which should not be on the public interface. There will be a lot of unique features, like breadcrumb navigation and additional select buttons that will require some new internal drilldown component - but most of it can and will be reused.
- `ResourceRetrieval::__construct()`: I think this can be implemented using the new component system, I therefore removed this declaration from the public interface.
- factory params: this is not possible because the byline is an optional argument. I would also like to keep the label and byline argument together, so I sticked with `NodeRetrieval` as a first argument. The same is already done for the file input.

Let me know what you think about this approach.

Kind regards,
@thibsy

klees (approved):
Hi @thibsy,

thanks a lot for the improvements. The purpose description is on point now and the other changes good as well. Well rounded result, it seems. Nice!

Take this to JF please.

Some tangential question came to mind when looking into this. We do not need to discuss these before taking this to the JF and iterating any further, but we may want to derive further actions here.

* Regarding factory params: Shall we change param order in the other according locations in the same way? Would be nice to be consistent here.
* Regarding ResourceRetrieval1: Same question, shall we change other locations accordingly?
* Regarding "Select"-Wording: Shall we rename the ColorPicker to ColorSelect?

For both moves "now" would be the right time, since we will never have less actual usages then now.

Thanks!

matthiaskunkel:
Jour Fixe, 10 JUN 2024: We highly appreciate this suggestion and accept the PR for trunk / ILIAS 10.

Amstutz:
Hi @thibsy 

thx, ill pass this back to you. Feel free to re-assign me/us as soon as implemented.

lg
Timon

thibsy:
Hi folks,

I finally managed it. This thing was a little more complicated than anticipated, but I think all the pieces are in place now. However, this PR is not yet complete and I would like your feedback/help before completing it.

- Most importantly, this PR is missing JavaScript and PHP unit tests, I will implement them once the HTML structure and business logic is signed off.
- The examples do not yet feature a full description, I will do so once the HTML/CSS is signed off.
- The pipelines are failing. I will fix them while implementing the unit tests.

While developing this input field, I also stumbled across the following issues (which still remain):

- the order of assets being loaded is wrong on two accounts: (a) for the drilldown menu, which is loaded after the inputs but required by the tree select, and (b) for the language service, which is being loaded after the UI framework (entirely, I assume). I worked around both of these issues, but I wanted to point them out:
	- [drilldown menu workaround](https://github.com/ILIAS-eLearning/ILIAS/pull/7143/files#diff-ef03ec89ebf29bde4d4f7277b852dd624870adbb45d3d1f08a36c630a4cab2edR846-R847)
	- [language service workaround](https://github.com/ILIAS-eLearning/ILIAS/pull/7143/files#diff-a2e0cc23e8eb0d42480b31d9f1ee426975d2dc297f157b26be45ba9e070a8c9dR488)
- The drilldown menu has an (S)CSS bug, where the width of a leaf node is different from a branch node, leading to an indent of the select-buttons inside the tree select. Unfortunately, I ain't the sharpest tool in the CSS shed, so I could not manage to fix it.

Relevant changes to the prior version of the Tree(Multi)Select input:

- `Field\HasDynamicInputs::getGeneratedDynamicInputs()`: its a while back when I introduced the `HasDynamicInputs` definition, so I was quite confused about what `getDynamicInputs()` was. I clarified this by a rename, hope thats OK. I touched the implementation anyways, to make fit for both Tree(Multi)Select and File inputs.
- `Field\Node\NodeRetrieval::getNode()`: I overlooked the fact that consumers can provide existing values (nodes). For this reason I needed an equivalent to the `Field\UploadHandler::getFileInfoResult()`, which provides the necessary data for rendering an existing value.
- `Field\Factory::node()`: introducing a dedicated endpoint for the UI framework proved to be more a complex step. Therefore I decided to solve the async rendering the old-fashioned way: a consumer endpoint. Therefore we need to expose the Node factory as well on the public interface now.
- `Field\Node\Factory::async()`: now accepts an `Data\URI` parameter, so consumers can provide their async rendering endpoint.
- `TreeMultiSelect::` and `TreeSelect::withValue()`: I am not entirely happy about the handling of existing values. I felt obligated to introduced a `Field\NodeRetrievalCache` trait, so the rendering and client-side value validation on inputs do not cause unnecessary database roundtrips. Maybe you have an idea on how to improve this.

Since this is a somewhat large PR, feel free to contact me so we can discuss the changes on Discord.

Looking forward to your feedback! 

Kind regards,
@thibsy

klees (commented):
Hi @thibsy,

thanks for moving this forward.

Regarding your issues:

* Loading order issues:
  * Could we somehow require drilldown in tree select via js-modules and call it a day? This would mean that we'd include code twice, I guess, but, aside some bytes over the wire if no caching is involved, would that be so horrible? Maybe, overmore, could this be a starting point for a "ui.js" that bundles everything?
  * We could not find (or understand...) the workaround for the language service. Does it still exist?
* SCSS-bug: Please just report it and link it here. We think you could just go on then, right?

Changes:

* `Field\HasDynamicInputs::getGeneratedDynamicInputs()`: Rename is fine in general (although breaking, so we should announce on JF). On the other hand: Why is this even on a public interface? Could we maybe move these specifics to an internal interface when we're at it?
* `Field\Node\NodeRetrieval::getNode()`: So, as far as we understand... There only has been a `NodeRetrieval::getNodes`-method that would have created the complete tree to select from recursively (and maybe even asynchronously...). Since this is a select input after all, it could be the case that the consumer already set some selected node(s) via the ordinary `withValue` from the input field. These node(s) shall be displayed, even before the complete recursive tree is created. So this is why another `getNode` method is required. Did we understand correctly? Is this also why you introduced that `NodeRetrievalCache`? If so:
  * The wording with "already provided values" threw us of.  Maybe state the facts more straight forward, somehow similar like we tried to do above. Maybe put the method below the `getNodes` method, because it seems to be more of a special case.
  * Will `NodeRetrievalCache` really have a big impact on performance? There will only be a few nodes selected most of the time.
  * Why does `getNodes` deserve an `IconFactory` but `getNode` does not? Is this maybe an implementation detail?
* `Field\Factory::node()`: Ok =)
* `Field\Node\Factory::async()`: Ok =D
* `::withValue()`: Caching question from above answered, it seems. It seems as if it could also be solved by introducing just another method to `NodeRetrieval`, something like `getInvalidNodeIds(array $node_ids): array` to produce a list of invalid inputs with one query (potentially). Since `NodeRetrieval` and similar classes are repo pattern in disguise I dont' think this would even feel wrong. Also, we could apply a similar pattern to the other methods, `getNodesById` instead of multiple calls to `getNode` and changing expectation for `getNodes` to already create a complete tree, that might contain "async"-nodes in part. This could drastically reduce DB roundtrips... Or is this already expected? Than maybe point out more clearly in the documentation. Maybe we could call `getNodesById` `getFlatNodesById` then, because we would not want the child nodes there, right?

We hope this makes some sense, maybe ping us for a discussion if not...

Please ping us and assign back when this is ready for the next round.

Kind regards!
@klees and @Amstutz

thibsy:
Hi @klees,

This should be ready for a final round now =).

- The loading issue of the drilldown JavaScript wasn’t addressed as proposed, because we cannot simply include a file. The required file acts as an index and does not export anything. Once all components are merged into a single `ui.js`, this can be handled properly. That change will also eliminate most of the `ComponentRenderer::registerResources()` calls, removing this workaround automatically.
- The `HasDynamicInputs` interface has been removed because it should indeed be internal. Implementations have been adjusted accordingly.
- A new `NodeRetrieval::getNodeAsLeaf()` method has been introduced to improve handling of existing values, which also gets rid of the caching shenanigans.
- The workaround for `il.Language` not being available has been clarified and simplified, without additional `il` usage in `TreeSelect` ([see here](https://github.com/ILIAS-eLearning/ILIAS/pull/7143/files#diff-c3f276a58a17b22bdcddbce0b8f5b6c3f7a03c01576966f4fbcaf0b690d28f11R44)).
- PHPUnit tests have been added to cover most rendering aspects.
- JavaScript unit tests have been added to cover the most important behaviour of the tree select component. These tests were created in compliance with #8787. Since automated JS testing is currently disabled, merge order is not a concern here.
- The JavaScript structure has been refactored to improve its testability.
- Full descriptions have been added to all tree (multi) select input examples.
- SCSS-related issues have been reported in Mantis:
  - https://mantis.ilias.de/view.php?id=44895  
  - https://mantis.ilias.de/view.php?id=44899  
  - https://mantis.ilias.de/view.php?id=44900  

Let me know what you think.

Kind regards,
@thibsy

components/ILIAS/UI/src/examples/Input/Field/TreeSelect/with_required.php
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * ---
+ * description: >
+ *   The example shows how to make a Tree Select Field required. This example does
+ *   not contain any data processing.
+ *
+ * expected output: >
+ *    ILIAS shows the Tree Select Field inside a Standard Form. Its label is displayed
+ *    with an asterisk (*) and the Form will dispaly according definitions above and
```
klees:
```suggestion
 *    with an asterisk (*) and the Form will display according definitions above and
```

klees:
Hi @thibsy,

I'm very excited that this will be in the core sometime soon. Really nice work!

- [ ] walkArray: Why would we use this instead of `forEach`?
- [ ] css: Shall any of the wizards check your CSS?

- [ ] attribute names: There are various attribute names that are used frequently to identify html elements and such... Please move them to according constants for deduplication, robustness and maintainability.
- [ ] `Node`-hierarchy: The hierarchy of `Node` classes feels a little garbled. For one, I was irritated why we would not just return `Node` from every factory method in the `Node`-Factory, to express the fact that consumers really shouldn't peak into the components they get. This was supported by the observation, that `Leaf` and `Async` do not contain any additional methods. Then I wondered, why `Async` does not support a public getter for the URL. I reckon that this is just a case of YAGNI, which is totally fine. Then I wondered about `getChildren` and how they would behave for `Leaf` and `Async`. I think that, on a technical level, this would be easy to solve, just `return []`. From a semantic perspective, it is not quite obvious that this is indeed a good solution. Why would I ask a `Leaf` for children? Wouldn't there be something wrong in the first place? And: Why isn't a `Leaf` just a `Node` with no children? (I know the answer, but I guess someone else will have the same thoughts...) Also: What is meant by calling `Async::getChildren()`? This somehow implies that the children are loaded asychronously, which is not what will happen. All this is not super duper important, but to me feels like some small amount of sand in the gearbox that will produce a regular drag. So, I would be a lot more comfortable if we'd introduce a "correct" hierarchy with a base class (maybe even call `BaseNode` or something, if nothing better comes to mind quickly) with `getId`, `getName`, `getIcon` and then derive `Async`, `Leaf` and `Node` with `getChildren` from there.
- [ ] docu: Docu of `NodeRetrieval::getNodesAsLeaf` says that the method "MUST NOT return instances for invalid node-ids". I guess we should specify what I shall do instead, if invalid node_ids are provided. Skip them silently? Throw an error? Something else? Please add docu accordingly =)
- [ ] `&times;`: We would be using the "multiplication"-glyph as the sign for closing. This seems to be okish, from an accessibility perspective, as the according text is hidden from screenreaders. But also it does feel not quite right and a little complicated. Could we just use a more fitting unicode sign or html-entity? Or is this something copied from elsewhere? Which to me would indicate that rectifying this would be a task for the roadmap.
- [ ] remove label: Could we add the info to the remove-label which node will be removed? Should be doable with the name and would be really nice for accessibility imo. Same tactic might apply to other labels. Currently things become a little blurry and I might be missing something.
- [ ] js unit tests: It seems as if the unit tests are mostly testing multiple things, one after another. Please consider splitting them even more. 

- [ ] `unshift`: Please [`unshift`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/unshift) [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7143/files#diff-24df457c2ad36d1bfd1133cd580b3ddbd4fb5a9c84adb9b4efd20bc86995bdb5R36) so you don't have to `reverse` [there](https://github.com/ILIAS-eLearning/ILIAS/pull/7143/files#diff-24df457c2ad36d1bfd1133cd580b3ddbd4fb5a9c84adb9b4efd20bc86995bdb5R43).
- [ ] lang: The [lang var for the requirement constraint](https://github.com/ILIAS-eLearning/ILIAS/pull/7143/files#diff-4ca148cbe5bf3f6d4f7c59649a399ef7d4bd2efcd9bc34884d3490f4fe3bafd2R71) seems to be incorrect. I think should say that something should be selected, instead of saying "you should select 1 instead of 0"... Looks like copy pasta?

As a side node: I'm still not very happy with that `Async` node that we would introduce here. It feels as if a rather minor implementation detail (will this node be fetched asynchronically or not?) has been lifted to our topmost semantic layer, the names of the components. This feels wrong, but I also think this is the best solution we currently have for that very real performance problem we would have had otherwise.

Second side-node: The `Renderers` suprisingly clean for a component of that size and complexity. Nice =)

Third side-node: JS unit tests look good and also seem to test substantial things. Great!

If there's anything I can help you with, feel free to ping me. @Amstutz and @yvseiler will leave another feedback regarding the frontend, but we felt it was a good idea to publish this already, as this seems to mostly contain aspects that can be tackled independently.

Best regards!

thibsy:
Hi @klees,

Thx for the feedback. I addressed most of your points and rebased onto latest trunk, so a test installation for @yvseiler can be built =).

- [x] `walkArray()`: this function should be used instead of `Array.forEach()` when working with `async`/`await` logic, because `forEach()` does not respect `await`. It does not pause execution for operations inside the callback and therefore leads to unexpected behaviour when the order of operations is important. `for` loops however respect `await` and process the operations sequentially. This function preserves correct flow and avoids race conditions and could save developers much sanity and time =).
- [ ] CSS: yes, please forward these issues to the CSS wizards =).
- [x] attribute names: done. I introduced a `constants.js` where the most relevant attributes, class-names and selectors are stored.
- [x] `Node`-hierarchy: done. I actually introduced a `Branch` node and used the `Node` as a base class. This also plays along with the drilldown menu nicely.
- [x] Docu: I added a comment to `NodeRetrieval::getNodesAsLeaf()` which says that the method MUST return placeholder nodes for invalid node ids. We should probably not silently ignore them, since consumers could be confused about the absence of a node. Instead, a placeholder should be visible that informs a user about the absence accordingly.
- [x] `&times;`: I copied this from the modals. I therefore added this to the roadmap instead, as we should streamline this across all copies of this HTML markup.
- [x] Remove label: done. Also done for select.
- [ ] JS unit tests: I deliberately tested multiple things, as you said, because one action should result in all of these things. E.g. the `updates select button states (can select child nodes)` test case should behave exactly like this. Multiple things _are_ happening at the same time, so I put them into one case. I would rather not separate these cases, because it would result in much more boilerplate code, for a very similar result. But if you have a strong opinion here, I will refactor ofc.
- [x] `unshift`: my version is actually better because every `unshift` call has `O(n)` complexity, whereas every `push` call has `O(1)` complexity and only once `O(n)` when we `reverse`.
- [x] lang: fixed.

Kind regards,
@thibsy

klees:
@catenglaender Could you have a look into the CSS here please...

klees (commented):
Hi @thibsy,

thanks for your changes and explanations. Seems to be all fine, LGTM.

I think we can merge this once @catenglaender and @yvseiler have looked this over.

Kind regards!

Amstutz:
Hi @thibsy

Thank you a lot for contributing to ILIAS. What a great component, thanks a lot for your work.

- [ ] Tree Select Example 1 visual appearance leaf: If a leaf is on the left-hand side, the bulky button for the selection (+) is not aligned with the upper ones, it is slightly indented to the left. I suspect this is because the expand glyph is not present. Is it possible to use the same length of the Bulky button for a leaf as for a branch?
- [ ] Tree Multi Select description "Composition": The text refers to a dropdown in the footer of the modal. This is not visible in the example. Has this not been implemented at all, i.e. the sentence would have to be deleted, or is the dropdown still missing and should still be included? ("This list will be visible inside a dropdown menu which can be opened inside the modal footer, and the surrounding form. The dropdown menu will keep track of the amount of selected nodes by showing an according counter near its trigger.")
- [ ] Tree Multi Select description "Effect": Same like Composition -> the text refers to a dropdown menu. There is no dropdown in the example. Is the dropdown missing or should the sentence be removed? ("When a node is selected or unselected, the counter near the dropdown menu trigger will be updated.")
- [ ] Tree Select & Tree Multi Select Expected Output: After clicking on Save in the form (Branch 1 was selected), an error message appears stating that the form could not be submitted securely. However, the Expected Output states that “the Node-id (<X>) will appear inside the output printed above the Form”. I'm not sure if this is due to my test instance, can you check if the expected output works as described?

- [ ] Tree Select & Tree Multi Select description "Effect" primary-button: You write "Clicking the primary-button of the modal will relay the currently selected node to the surrounding form, where it will be displayed by its name.". But the way I see it, the entry is transferred to the form as soon as you click on the Bulky button (+). Even if I close the modal with “Close” (x), the entry remains in the form. From my point of view, you could remove this sentence and add the following: “When a node is selected, all other nodes inside the tree will indicate that they can no longer be selected. At the same time, the selected node is transferred to the surrounding form.”. Or you could describe it more like in the Expected Output part.
- [ ] Tree Select visual appearance: It would be great if we could adopt the visual concept of the drilldown in the context "Add New Object" (https://github.com/ILIAS-eLearning/ILIAS/pull/9101) for the Tree Select and Tree Multi Select. The drilldown for “Add New Object” has been visually adapted so that it is more visible which menu is currently selected. From my point of view, this would also fit in this use case for tree-select. Please check whether you can adopt the visual pattern for tree-select (see mockup). If this would need too much work, we could also decide to go with the current visual appearance.

- [ ] Tree Select description "Effect" Typo: "When a node us unselected, .." please write "is unselected"
- [ ] Tree Multi Select description "Composition" remove glyph: There is no remove glyph. You write "The selected nodes {...}, where each entry features a removal glyph to unselect the according node again.", I think the "removal glyph" is outdated here. Please write to "The selected nodes will be displayed as an unordered list of names, where each entry features a glyph to unselect the according node again."
- [ ] Tree Multi Select description "Effect" remove glyph: Same as above for Composition, there is no remove glyph. Please remove the word "removal" in the sentence. ("Clicking the removal glyph of a node inside the selection list will unselect this node again").
- [ ] Tree Multi Select description "Context" Typo: The title written here is "Multi Tree Select" but the component is called "Tree Multi Select".
- [ ] Tree Select & Tree Multi Select HTML Validator: There are different errors like Line 5752: "Element div not allowed as child of element button in this context. (Suppressing further errors from this subtree.)" and Line 5758: "The element a must not appear as a descendant of the button element." (glyph Expand Content). Please have a look into the HTML Validator. If I can help to get these results, please let me know.

If you have any questions about the issues above, please get in touch. Thank you very much in advance for looking at these.

kindly,

@Amstutz on behalf of @yvseiler

thibsy:
Hi @Amstutz and @yvseiler

Thx a lot for the feedback. Let me answer / address the points of your review:

- [ ] Tree Select Example 1 visual appearance leaf: I noticed this myself, but unfortunately I could not manage to address this issue (for the life of me). I therefore reported it (tickets are above) so one of the (S)CSS wizards can have a look. I think the width is also broken in the drilldown menu, its just not visible there because leaf nodes don't feature anything in addition.
- [x] Tree Multi Select description "Composition": I decided to skip this feature for the first iteration, since we have been waiting for this thing for a while now. I removed the phrasing accordingly.
- [x] Tree Multi Select description "Effect": removed.
- [x] Tree Select & Tree Multi Select Expected Output: I cannot reproduce this on a local instance.
- [ ] Tree Select visual appearance: I actually used the same variables from the drilldown menu. Could you tell me what exactly looks different / where I should use which variable? When comparing the drilldown menu and tree select components in the Kitchensink I could not find any differences (except for some backgroung-colours due to bein `transparent`). Maybe this creation-screen thingy has some additional CSS?
- [x] Tree Select & Tree Multi Select description "Effect" primary-button: done.
- [x] Tree Select description "Effect" Typo: done.
- [x] Tree Multi Select description "Composition" remove glyph: done.
- [x] Tree Multi Select description "Effect" remove glyph: done.
- [x] Tree Multi Select description "Context" Typo: done.
- [x] Tree Select & Tree Multi Select HTML Validator: done. I had to use the `Glyph\ButtonContextRenderer` for the nodes as well. HTML should be valid now, good catch!

Looking forward to the final round =).

Kind regards,
@thibsy

thibsy:
Hi @Amstutz and @yvseiler,

Turns out I accidentally rolled back the changes from #9101 when rebasing onto the latest trunk - that's why I thought the drilldown looks similar in both contexts. However, thanks to @catenglaender I just made the Tree (Multi) Select work together with the new look of the drilldown menu. Some tweaks will most likely be necessary in the future, but for the time being this should be sufficient.

Kind regards,
@thibsy

yvseiler:
Dear @thibsy 
Thanks a lot for looking into this SCSS magic. I'm not sure if the following little thing can be fixed too, I don't quite understand where it comes from: 

- When the modal is called for the first time, the "+" symbols are coloured with the wrong background (see screenshot 1). It looks like the left add symbol belongs to the right branch, the add symbol on the right side has a white background.
- As soon as you click on a branch, the colour scheme corrects itself so that it looks correct (see screenshot 2).

**Screenshot 1**
Screenshot 1


**Screenshot 2**
Screenshot 2


Otherwise, I think the changes have been successfully made, thank you very much for your efforts!

Kind regards,
@yvseiler

thibsy:
Hi @yvseiler,

Thx for your feedback. I tried looking into the SCSS again, but could not manage to find the right combination of selectors that make the button styles work as expected. I would like to hand this over to the SCSS magicians again, so I have reported the issue on mantis and added it to the list of known issues:

- https://mantis.ilias.de/view.php?id=44895
- https://mantis.ilias.de/view.php?id=44899
- https://mantis.ilias.de/view.php?id=44900
- https://mantis.ilias.de/view.php?id=45433

I hope this is sufficient so we can merge this beast =).

Kind regards,
@thibsy

Amstutz:
I merged to PRs, thx for reporting the bugs. They are on the agenda to be fixed.

matthiaskunkel:
Jour Fixe, 04 AUG 2025: Thibeau notified us about a necessary change for async nodes that still require a target URL due to not yet available changes in the UI framework. We hope that this make it easier for developers to use it.

iszmais:
@thibsy Is stumbled upon this as I encountered some divergence between the `delos.scss` and `delos.css`.

I created a PR to align these: https://github.com/ILIAS-eLearning/ILIAS/pull/10003
Please correct me if I'm seeing something wrong here.

Greetings,
@iszmais

PR #7145, opened 2024-03-06
Ildcl unique f01

sz-tik:
this is the uniqueness check when updating field settings:
more precisely, this means that if a field is initially set to 'non-unique' and there are already records, the field values of all existing records have to be checked. an error message must appear if values occur more than once and the field property is to be changed to 'unique'.


dsstrassner:
hi @sz-tik & @pascalseeland 

Mantis Issue Number or URL is Missing for Bugfix Label?

sz-tik:
@dsstrassner 
Hi Dennis,
 
this code fixed the uniqueness check for field type integer and text.
The following error has led to the uniqueness of all field types being checked
"mantis number 30758: Formula doesn't test uniqueness right away."

Possibly corrected checks for the other field types, esp. the formula field, will follow.

sz-tik:
This pull request has become obsolete, because the changes were covered by the pull request #7152

PR #7147, opened 2024-03-07
9_removes_component_xmlns_from_dataset

chlulei:


smeyer-ilias:
Hi @mjansenDatabay 

we found some issues regarding the namespace declarations used for modules exports which use the dataset implementation. To fix these issues we decided to remove the namespace declaration from  the dataset record element (<ds:Record xmls="http://www.ilias.de/xml/Modules/ContentPage/copa">...). 
We also had to adapt the scheme file. Essentially these were the following changes:
- remove prefixed "copa" namspace
- add prefixed XMLSchema namspace
- prefix all XMLSchema namespace elements/attributes with "xs"

Are these changes ok with you?

smeyer-ilias (approved):
Thanks @chlulei for fixing this.

mjansenDatabay (approved):

PR #7152, opened 2024-03-08
Ildcl unique f02

sz-tik:
Uniqueness check of number and of text field when editing a data record or adding a new data record corrected.
Some improvements made to the code concerning basic field operations of the DC.

### DIFF
--- a/components/ILIAS/DataCollection/classes/Fields/Base/class.ilDclBaseFieldModel.php
+++ b/components/ILIAS/DataCollection/classes/Fields/Base/class.ilDclBaseFieldModel.php
@@ -531,15 +531,17 @@ class ilDclBaseFieldModel
     public function checkValidity($value, ?int $record_id = null): bool
     {
         //Don't check empty values
-        if ($value == null) {
+        if (!isset($value)) {
             return true;
         }
 
         if ($this->isUnique()) {
             $table = ilDclCache::getTableCache($this->getTableId());
             foreach ($table->getRecords() as $record) {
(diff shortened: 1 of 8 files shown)


dsstrassner:
Hi @sz-tik, if you post a Mantis Issue URL, I can label this as Bug fix. Until then, this is an Improvement.

sz-tik:
@dsstrassner
Hi Dennis,

this code fixed the uniqueness check for field type integer and text.
The following error has led to the uniqueness of all field types being checked
"mantis number 30758: Formula doesn't test uniqueness right away."

Possibly corrected checks for the other field types, esp. the formula field, will follow.

f02 also includes the changes from f01, so f01 is actually obsolete. Sry.

Amstutz:
I looked through it, and discussed some questions through discord with @sz-tik . Thx a lot for improving this!

iszmais:
@sz-tik @Amstutz ,

I stumbled upon this during the testing of https://testrail.ilias.de/index.php?/tests/view/77632 since regex and length are no longer checked for ILIAS 10 on text fields.

Obviously, because the check is removed here. 
Normally I would assume this to be a bug, but since this is explicitly removed here (and I could not find any reasoning why) i just want to double-check what the intention of this was.

Greetings,
@iszmais

iszmais:
@sz-tik @Amstutz FIY.

I restored some of the removed code here since it removed required validations: https://github.com/ILIAS-eLearning/ILIAS/pull/9426

Please let me know if this contradicts any of your conceptions.

Greetings,
@iszmais

PR #7154, opened 2024-03-11
ADMS: fix error regarding saving custom tile server input for Open La…

lukastocker:
…yers Maps. (T72450)


dsstrassner:
https://mantis.ilias.de/view.php?id=40686

klees:
Picked to trunk.

PR #7157, opened 2024-03-11
CSS: Fix Add New Item Scrollbars

kergomard:
This removes the scrollbars from the "Add New Item"-dropdown.

See: https://mantis.ilias.de/view.php?id=40773

It should not be picked to trunk as there will be another solution there anyways.

Best,
@kergomard


Amstutz:
Thx a lot for the PR! Note, NOT picked to trunk.

PR #7163, opened 2024-03-12
[FIX] missing procesZipFile-Method

chfsx:
@alex40724 this PR should solve comment in issue https://mantis.ilias.de/view.php?id=40540#c102581
I don't know/when the movement of the method was lost in PR #7089


alex40724:
@chfsx Thanks. Yes, I was expecting this to happen, but did not tackle it either yet.

PR #7164, opened 2024-03-13
0040628: Course Links require Read Access to Target Course

playamongthestars:
This PR is for ticket: https://mantis.ilias.de/view.php?id=40628


Modules/CategoryReference/classes/class.ilObjCategoryReferenceGUI.php
```diff
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
```
mjansenDatabay, line 3: @nmatuschek We should not remove this, but only move this below the license header.

Services/ContainerReference/classes/class.ilContainerReferenceGUI.php
```diff
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
```
mjansenDatabay, line 3: @nmatuschek We should not remove this, but only move this below the license header.

Services/ContainerReference/classes/ilContainerReferenceRepositorySelectorExplorerGUI.php
```diff
@@ -0,0 +1,314 @@
...
+/**
+ * Explorer for selecting repository items.
+ *
+ * The implementation starts as a replacement for the often (ab)used ilSearchRootSelector class.
+ * Clicking items triggers a "selection" command.
+ * However ajax/checkbox/radio and use in an inputgui class should be implemented in the future, too.
+ *
+ * @author Alexander Killing <killing@leifos.de>
+ */
+class ilContainerReferenceRepositorySelectorExplorerGUI extends ilTreeExplorerGUI
```
mjansenDatabay: Please add a `class.` to the filename.

mjansenDatabay: Furthermore it should be sufficient to use `extends ilRepositorySelectorExplorerGUI` and simply **overwrite** `isNodeClickable`

Services/ContainerReference/classes/ilContainerReferenceRepositorySelectorExplorerGUI.php
```diff
@@ -0,0 +1,314 @@
...
+
+/**
+ * Explorer for selecting repository items.
+ *
+ * The implementation starts as a replacement for the often (ab)used ilSearchRootSelector class.
+ * Clicking items triggers a "selection" command.
+ * However ajax/checkbox/radio and use in an inputgui class should be implemented in the future, too.
+ *
+ * @author Alexander Killing <killing@leifos.de>
+ */
```
mjansenDatabay: Please remove this comment.

Services/ContainerReference/Form/class.ilContainerReferenceRepositorySelectorInputGUI.php
```diff
@@ -0,0 +1,123 @@
...
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+use ILIAS\ContainerReference\ilContainerReferenceRepositorySelectorExplorerGUI;
+
+/**
+ * Select repository nodes
+ *
+ * @author Alexander Killing <killing@leifos.de>
```
mjansenDatabay:
```suggestion
```

fsesterh:
Hi @smeyer-ilias , any feedback on this topic?

alex40724:
Hi @nmatuschek 

Thanks for providing a PR for this issue.

- [x] Why having a whole new class to tackle this, couldn't we just add a configure option to ilRepositorySelector2InputGUI which permission to check (read/visible)? I suspect the resulting code change would be much smaller.

- [x] Please add a configure option for ilRepositorySelector2InputGUI unless there is a very good reason not to do so. We should have less code and make the option available for other contexts as well.

kindly,
@alex40724

smeyer-ilias (approved):
Thanks @nmatuschek  for fixing this

PR #7165, opened 2024-03-13
TestQuestionPool: Fix query in lucene definition file

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40815


kergomard:
Thank you very much @mjansenDatabay ! Merged and picked to trunk.

PR #7166, opened 2024-03-13
UI: add required text to top of form (#33699)

lukastocker:
https://mantis.ilias.de/view.php?id=33699

This fix will not only add changes to the UI Form but also to the Legacy Form as it's still used at multiple positions (e.g. course settings).


lukastocker:
Hi @Amstutz, 

please have a look at this implementation especially the design part.

Regards,
Luka

Amstutz:
looks good, thx a lot!

PR #7172, opened 2024-03-13
UI: remove icon's alt txt input if icon label is the same as the node…

lukastocker:
… label. (#36752) 

https://mantis.ilias.de/view.php?id=36752

### DIFF
--- a/src/UI/Implementation/Component/Symbol/Icon/Icon.php
+++ b/src/UI/Implementation/Component/Symbol/Icon/Icon.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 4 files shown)


Amstutz:
Thx a lot. Looking good. Can you provide a Unit Test, making sure, this improvement will be kept for further refactorings?

lukastocker:
> Thx a lot. Looking good. Can you provide a Unit Test, making sure, this improvement will be kept for further refactorings?

Hi @Amstutz ,

I just added some PHPUnit tests:

[x] setting a custom label for an icon
[x] construction of a tree node with an icon and different labels (tree node label, icon label)
[x] testing that different labels (tree node and icon) are not equal
[x] testing rendering of tree node with an icon and different labels (tree node label, icon label) which results to a non-empty alt-attribute.

Note: there's already a rendering test with icons and labels that do not differ. The result is always an empty alt attribute. That test will fail if the alt attribute is not empty in that case.

Greetings
Luka

Amstutz:
thx a lot!

PR #7186, opened 2024-03-14
Align data options between global user management and local user mananagement

iszmais:
https://mantis.ilias.de/view.php?id=40840

PR #7187, opened 2024-03-14
Align data options between global user management and local user mananagement for TRUNK

iszmais:
Sister PR to https://github.com/ILIAS-eLearning/ILIAS/pull/7186


kergomard:
Thank you for this PR @iszmais ! I actually picked your original PR to ILIAS 9.

PR #7189, opened 2024-03-15
Init/PasswordAssistance: Trim form values

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=39869

If approved, this has to be picked to `trunk`.


mjansenDatabay:

mjansenDatabay:
Merged, since the changes are more or less trivial string transformations.

PR #7190, opened 2024-03-15
Init/PasswordAssistance: Trim form values (< ILIAS 9.x)

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=39869

If approved, this has to be picked to `release_8` as well.


mjansenDatabay:

mjansenDatabay:
Merged, since the changes are more or less trivial string transformations.

PR #7191, opened 2024-03-15
9 mantis ilias 39882

chlulei:
Link: https://mantis.ilias.de/view.php?id=39882


chlulei:
Is addressed in https://github.com/ILIAS-eLearning/ILIAS/pull/7224

PR #7192, opened 2024-03-18
T&A Bugfix #0040634: Calling print view in question list of test leads to fatal type error

fhelfer:
https://mantis.ilias.de/view.php?id=40634

### DIFF
--- a/Modules/Test/classes/class.ilObjTestGUI.php
+++ b/Modules/Test/classes/class.ilObjTestGUI.php
@@ -195,7 +195,7 @@ class ilObjTestGUI extends ilObjectGUI implements ilCtrlBaseClassInterface
             $this->ctrl->setParameter($this, 'prev_qid', $this->testrequest->raw('prev_qid'));
         }
 
-        if (!$this->getCreationMode() && $this->testQuestionSetConfigFactory->getQuestionSetConfig()->areDepenciesBroken()) {
+        if (!$this->getCreationMode() && isset($this->testQuestionSetConfigFactory) && $this->testQuestionSetConfigFactory->getQuestionSetConfig()->areDepenciesBroken()) {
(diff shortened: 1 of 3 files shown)


kergomard:
Thank you very much for the PR @fhelfer ! Special thanks for typing this all a little more strictly.

I'm sorry to say though: I don't think this is the right solution:
- Why are asking for the reached points of a print preview?
- Why do we have an `active_id` here that also has a result?
- What feedback are we expecting to be shown here? Feedback for the right answer or the wrong one? (@dsstrassner : Maybe you have an answer for this.)

My feeling is: There is something fundamentally wrong here.
It would be very kind, if you could analyze this a little more deeply. Maybe we should also discuss this amongst @dsstrassner, @mbecker-databay, me, and you?

Thank you very much and best,
@kergomard

dsstrassner:
I am sorry about the delay @fhelfer: 

ILIAS 7 only shows the correct answer(s) and no feedback in the Print View of the Question Tab. The review tab only shows the question.

 Please remove the link between the feedback setting of scoring and results and the print view on the questions tab. The Feedback setting in Scoring and Results should only be relevant for the viewing of answers in the test results. (ILIAS 8 only).

dsstrassner:
I think this all came from the "Einstellungsüberarbeitungsworkshop" and the changes at the scoring and results sub tab ;-)

fhelfer:
Hi @kergomard,

I disabled the calculation of feedback for the print view. There's still result calculations with an active_id = 0 (but most have been disabled with this PR), so there might still be something fundamentally wrong, but this seems to be a good enough fix IMO.

Please give the PR another review.

Best @fhelfer.

kergomard:
Thank you very much again for the PR @fhelfer ! I cherry-picked your changes manually and added a few more of my own, thus I close this.
Best,
@kergomard

PR #7196, opened 2024-03-18
T&A Bugfix #0040378: Copy Paste Gap Matches results in unwanted behaviour

fhelfer:
https://mantis.ilias.de/view.php?id=40378

Could not reproduce the error in 8, seems to work fine


kergomard:
Thank you very much @fhelfer ! This was no problem in ILIAS 8, because I had fixed it there, but not backported the fix to ILIAS 7. I used the same approach, I had used for ILIAS 8 now and close this PR.
Thanks again and best,
@kergomard

PR #7198, opened 2024-03-18
Possible fix for 39455: Failed test: # Input: Container → Filter → Standard validieren

tfamula:
https://mantis.ilias.de/view.php?id=39455

### DIFF
--- a/src/UI/Implementation/Component/Input/Container/Filter/Renderer.php
+++ b/src/UI/Implementation/Component/Input/Container/Filter/Renderer.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 namespace ILIAS\UI\Implementation\Component\Input\Container\Filter;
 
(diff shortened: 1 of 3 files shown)


Amstutz:
Hi @tfamula 

Thx a lot for the fix. 

Can you shortly link the resource, describing, why "heading" is the right choice here? It's not immediately obvious to me. 

Further, tests need fixing, thx!

tfamula:
Hi @Amstutz,
I made a mistake here, thanks for intervening! "heading" is not the right choice. In my eyes, `aria-expanded` should not be used in the affected `span` and the Bulky Button(s) within the `span` should do the job (what they already do, see screenshot). Then, the missing `role` attribute becomes obsolete. 
Please tell me if you agree.


filter

Amstutz:
Hi @tfamula 

To be honest, I am not 100% sure here. However, if I understand https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded correctly, it should be used on the opener, since it shows or hides the filter, right?

So my guess would be, that the aria-expanded is needed, here.

tfamula:
Hi @Amstutz 
for me, the opener is the `Bulky Button`(s). The `span` has in fact "opener" in its class name, but I see it more as as wrapper for the two `Bulky Button`s, of which only one of them is displayed depending on the expanded/collapsed state. IMO, it would be the "more correct" way to let the buttons do the `aria-expanded` job. If the `span` should do it, then the `role` attribute is essential – but I have no clue which role would be suitable here. Beyond that, I'm pretty sure that it is **not** a good way that the `span` as well as the `Bulky Button`s underneath bear the `aria-expanded` attribute at the same time. But I am not an accessibility expert, though.

Amstutz:
I can follow your reasoning. After looking again at the filter mechanics here, I believe your proposal is the best way to take an iteration here to improve a11y in the filters.

Thx for the PR and contributing insight in the discussion. Highly appreciated.

PR #7199, opened 2024-03-18
Test: Get rid of super global variable usage in `ilParticipantsTestRe…

mjansenDatabay:
…sultsGUI`


mbecker-databay:
@mjansenDatabay Please provide this improvement for ILIAS 9, too, if you like. The amount of arbitrary boyscouting changes on the same commit makes the core aspect of the PRs content super hard to port against the next release version.

PR #7200, opened 2024-03-18
PRG: 40853, fix iteration over assignments to acknowledge multiple user/courses

nhaagen:
https://mantis.ilias.de/view.php?id=40853


klees:
Picked to trunk.

PR #7205, opened 2024-03-19
Survey: Potential fix for wrong PDFs in tutor results email

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40861


alex40724:
See my comment in https://mantis.ilias.de/view.php?id=40861

PR #7206, opened 2024-03-19
Block: fix path to component-specific block classes (40822)

schmitz-ilias:
This PR fixes [40822](https://mantis.ilias.de/view.php?id=40822) by adapting a file path to the new directory structure.


klees:
I have merged this, since this is mostly trivial, caused by the changes due to the Component Revision and unmaintained code.

PR #7210, opened 2024-03-19
[FIX] missing tpl.toolbar_itemgroup.html file in trunk

chfsx:
The cherry-pick of #7102 went wrong, at least this change is missing


Amstutz:
thx a lot!

PR #7212, opened 2024-03-19
PRG: 40839, fix null-values in columns for optional user fields

nhaagen:
https://mantis.ilias.de/view.php?id=40839


klees:
Picked to trunk.

PR #7216, opened 2024-03-19
SETUP: fix 'register_nic' for the update process. (#40142)

lukastocker:
This PR contains following adjustments/fixes:

- Before this change "nic_enabled" (true) wasn't set/updated
during the update process in the database table "settings" as it was
missing in the SetupAgent's "getUpdateObjective".

- adjust the "generateNICKey" function in ilNICKKeyStoredObjective.php
as the client id could not be found while trying to create a nic key in
case no key was saved in the database before.

- adjust "isApplicable" to run only when "inst_id" is "0".

- add a query to avoid deactivation of an registered NIC via the update
  process.

- insert "ilInstIdDefaultStoredObjective" to the NICKeyRegisteredObjective
and NICKeyStoredObjective as it the order of the steps were incorrect which
resulted to an emtpy value in the database.

- delete "ilInstIdDefaultStoredObjective" in SetupAgent's "getInstallObjective".

### DIFF
--- a/setup/classes/class.ilNICKeyRegisteredObjective.php
+++ b/setup/classes/class.ilNICKeyRegisteredObjective.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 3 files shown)


klees:
Picked to trunk. Will pick to release_8 once this is tested.

PR #7219, opened 2024-03-19
9/UI/DataTable multi select footer col 40593

catenglaender:
Partly fixes https://mantis.ilias.de/view.php?id=40593

# Issue

Some users have troubles recognizing that the multi-select batch action dropdown below the table belongs to it.

image

# Change

The multi-select batch action dropdown is now part of the table in its own cell and row.
Exact col-span is filled into html template via php. Despite this not purely using css for a visual choice, this solution was discussed and accepted by the weekly CSS Squad in the ILIAS Discord server today.

image

# Outlook

Lable and info text mentioned in the ticket still have to be implemented and are not included in this PR.

Unit tests do not fail despite some changes to the html structure. Maybe at least one test should be added that includes this multi-select batch actions?

### DIFF
--- a/src/UI/Implementation/Component/Table/Renderer.php
+++ b/src/UI/Implementation/Component/Table/Renderer.php
@@ -262,6 +262,8 @@ class Renderer extends AbstractComponentRenderer
                     $component->getMultiActionSignal(),
                     $modal->getShowSignal()
                 );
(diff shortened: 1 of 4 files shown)


klees (approved):

klees:
@Amstutz Would you have a look into this PHP/CSS-kombination?

Amstutz:
We discussed this in the last CSS squad and agreed this to be fine.

PR #7222, opened 2024-03-19
Remove horizontal scrollbar

Dennis1993:
Fix bug https://mantis.ilias.de/view.php?id=40878

Nicht benötigte horizontale Scrollbar entfernen

ilias-onlinelist

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -9991,7 +9991,6 @@ footer {
   box-shadow: 1px 3px 4px rgba(0, 0, 0, 0.1);
   background-color: #fafafa;
   position: absolute;
-  max-width: 500px;
   max-height: 90vh;
   overflow-y: auto;
   right: 0;


Amstutz:
Hi @Dennis1993 thx a lot for looking into this issue and providing a PR! Highly appreciated. However, we decided yesterday in the CSS Squad on Discrod to not merge the PR for the following reasons:

1. We have not been able to reprodue the issue. Is this maybe occuring on a skin?
2. Changes would need to be done in less (from 9 on SCSS), an then be compiled to CSS, see: https://github.com/ILIAS-eLearning/ILIAS/tree/release_8/templates
3. We are not convinced that complitely removing the max width for all slates is a good option due to side effects.

However, we found a maybe related issue due to your report. Maybe while fixing this, we will get rid of this here as well on your system (will be reported shortly).

PR #7223, opened 2024-03-20
0040595: Undefined array key "BlockAfterPassed"

playamongthestars:
This PR fixes this Bug: https://mantis.ilias.de/view.php?id=40595

### DIFF
--- a/Modules/Test/classes/class.ilObjTest.php
+++ b/Modules/Test/classes/class.ilObjTest.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
--- a/Modules/Test/classes/class.ilObjTest.php
+++ b/Modules/Test/classes/class.ilObjTest.php
@@ -8765,42 +8765,42 @@ class ilObjTest extends ilObject implements ilMarkSchemaAware, ilEctsGradesEnabl
         $testsettings = unserialize($test_defaults["defaults"]);
         $this->mark_schema = unserialize($test_defaults["marks"]);
 
-        $this->setTitleOutput((int) ($testsettings['TitleOutput'] ?? 0));
-        $this->setIntroductionEnabled((bool) ($testsettings['IntroEnabled'] ?? false));
-        $this->setIntroduction($testsettings['Introduction'] ?? '');
-        $this->setFinalStatement($testsettings['FinalStatement'] ?? '');
-        $this->setShowInfo((int) ($testsettings['ShowInfo'] ?? 1));
-        $this->setForceJS((int) ($testsettings['ForceJS'] ?? 1));
-        $this->setCustomStyle($testsettings['CustomStyle'] ?? null);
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @nmatuschek for the PR!
This for me looks like a source for future very subtle issues. I would thus kindly ask you to find a more specific solution or to explain why this is not possible.

Thank you very much and best,
@kergomard

mbecker-databay:
I disagree with @kergomard. 
Casting is a viable solution to this issue, especially when dealing with potentially legacy serialized products in $testsettings.

I do not see a real alternative to the approach in the PR.
Instead of smashing this as merge, I will leave this open until tomorrow if anyone has a better idea + the words to explain it.

kergomard:
@mbecker-databay : Could you please explain to me, why this should be?

The approach with casting has the following issues:
- There is no way we can be sure the cast will produce equivalent values. We are thus not fulfilling our promise here: You store your default values and we are going to apply them to produce an equivalent test. The feature becomes useless.
- Should not be picked to trunk, as it is going to break there and it will break worse with every version.

The right approach in my eyes is to check the default values for validity, perform some trivial corrections, if possible, store the corrected values back to the database or inform the user that the values are not recoverable anymore and thus will be removed.

I kindly ask you not to merge this, before we have not discussed this further.

Thanks and best,
@kergomard

mbecker-databay:
I have not yet looked for a trunk solution to this, but the point is here that something was missing in the serialized contents and the PR applies a default.
This catches especially the cases when personal defaults were stored prior to the addition of new options in an older release. So this is a valid fallback mechanism if an item is not stored as part of personal defaults and I still stand on the hill saying that casting/defaulting is a very proper approach when dealing with persisted serialized data.

I put this on the TechSquad agenda for tomorrow and invited @nmatuschek to join us.

kergomard:
Thanks again for the  PR and the discussion @nmatuschek, @mbecker-databay and @mjansenDatabay !

We discussed this at the TechSquad today and came to the following conclusion:
* We all agree that the ideal solution would carefully inform the user about settings that cannot be applied anymore because of changes in ILIAS and about settings that have newly been introduced. Additionally, we could have some migration infrastructure and there might be more. Even if we believe that a solution along these lines will be necessary in the future, as the changes in the settings of the test increase, it is far outside of the scope of a bug-fix.
* In the context of a bug-fix we see two options:
  * The simplest would be to limit casting and null-coalescing to the values we know need it. This would currently be "BlockAfterPassed" plus a backport of the necessary fields in ILIAS 9.
  * A slightly fancier solution would build the values before setting them and only set a value if it is not `null` and possibly pass a minimal plausibility test. If it doesn't this could be used to show the user some information.

I think it is obvious which solution would be preferred, but all depends on available resources.

Thank you very much and best,
@kergomard

playamongthestars:
I reverted the changes and limited the casting to the "BlockAfterPassed" value.  I think the best solution (for future releases)  would be, to get rid of the settings that are currently stored in a serialized value. Of course this could not be part of a bugfix, because this needs more refactoring and a migration.
 
Thank you very much an best wishes 
Nadia

mbecker-databay:
Good with it! Thank you so much!

PR #7224, opened 2024-03-20
9 import validation errmsg xsd missing

chlulei:
Adds a new error message that is shown if the available xsd files of a component do not match the schema version of the export xml.


smeyer-ilias (approved):
Thanks @chlulei

PR #7227, opened 2024-03-20
[FEATURE] UI: improve documentation about rendering tests.

thibsy:
Hi folks,

I was looking in our documentation for some information about unit tests, which I found in the section on how to implement a component.

However, I found that some crucial information is missing, which describes how we want to decouple our rendering tests from other components HTML. I therefore decided to create a new chapter, describing how unit tests, especially rendering tests, of components should be implemented.

Let me know when I should move some parts around, or if this is not actually the practise we want to embrace.

Kind regards,
@thibsy

### DIFF
--- a/components/ILIAS/UI/README.md
+++ b/components/ILIAS/UI/README.md
@@ -230,51 +230,9 @@ If you would like to implement a new component for the framework, you should per
     }
     ```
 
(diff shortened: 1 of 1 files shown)


Amstutz:
thx a lot! highly appreciated.

PR #7229, opened 2024-03-20
[FIX] #40890 Component: streamline plugin environment.

thibsy:
Hi @klees,

I have opened this PR to address the following mantis issue: https://mantis.ilias.de/view.php?id=40890.

As described, there are some inconsistencies between the different plugin objectives when it comes to global constants. Since the problem occurs depending on what objectives are actually performed first, I considered this an ILIAS bug. In this PR I solved the issue by moving the initialisation of `ilObjectDefinition` to the bottom. While at it, I also made sure the same constants are defined in each objective.

Let me know if you think otherwise, or want to address this issue differently - or not at all.

Kind regards,
@thibsy


klees:
Thanks a lot!

klees:

PR #7230, opened 2024-03-21
UI: remove padding of forms if form is set in a modal. (#40675)

lukastocker:
https://mantis.ilias.de/view.php?id=40675


Amstutz:
Again Padding in Forms. @catenglaender your take on this?

catenglaender:
This is an excellent example for why forms shouldn't come with padding: Because a modal brings its own padding (which it should) and then it's too much padding in total.

The ideal implementation of the form would never have outward padding and overriding padding: 0 would never be necessary.

But fixing this for ILIAS 9 is too ambitious. Refactoring it now might make forms elsewhere look a bit broken, so... I am afraid we have to use the fix Luka suggested. I have no idea how to make it any better today without opening the can of worms. :broken_heart:

Amstutz:
@catenglaender I feel you. thx for looking into this.

PR #7232, opened 2024-03-21
9 adds sess 5 4 xsd

chlulei:
Adds missing sess 5 4 xsd, adjusts available sess xsd files and fixes type errors.


smeyer-ilias (approved):
Thanks @chlulei !

PR #7233, opened 2024-03-21
[FIX] 0040897: Whoops\Exception\ErrorException thrown with message "U…

chfsx:
…ndefined array key "multi""

Fixes the issue in https://mantis.ilias.de/view.php?id=40897


smeyer-ilias (approved):
Thanks @chfsx for fixing this and the refactoring.

PR #7238, opened 2024-03-21
fixed 38856: Hilfe-Tooltip - Settings in persönlichen Settings

alex40724:
This fixes https://mantis.ilias.de/view.php?id=38856 and removes help internals from the profile code.


kergomard:
Thank you for the PR @alex40724 !

We have to give this patterns some thought in the future, but for now, I think, this is the best we can do.

Merged and picked to trunk.

PR #7239, opened 2024-03-21
REP: remove content of alt attributes for icons in...

lukastocker:
... the main bar as they are the same as the item title. (#36753)

https://mantis.ilias.de/view.php?id=36753


Amstutz:
Looks good to me. Since the change is in Repository, this is @alex40724 call, to be merged.

PR #7241, opened 2024-03-21
UI: add padding to legacy UI's form buttons positioned in the footer.…

lukastocker:
… (#39609) 

https://mantis.ilias.de/view.php?id=39609

### DIFF
--- a/templates/default/070-components/legacy/Services/_component_form.scss
+++ b/templates/default/070-components/legacy/Services/_component_form.scss
@@ -600,7 +600,6 @@ input[type="checkbox"] {
 
 .ilFormFooter .ilFormCmds {
 	text-align: right;
-	padding:0;
 }
 
 /* jQuery ui autocomplete menu */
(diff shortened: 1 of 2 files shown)


lukastocker:
Hi @Amstutz ,

please have a look at this PR. It works so far, but I am not 100% sure if it will have a negative impact on other legacy forms. Therefore I would love your input.

Greetings
Luka

Amstutz:
@catenglaender mentioned, that he wanted to have a close look at this one. I forwarded it therefore to him, thx a lot!

catenglaender:
We had a look at it today, this is indeed the correct way to do it! :slightly_smiling_face:  One more annoying padding gone! :tada:

(That cursed row outside of it styled with `.form-horizontal .form-group { margin: 0 }` though... we have to get rid of that in ILIAS 10 - it will break the look of all forms temporarily, but it's a good way to move this mess forward)

Amstutz:
thx a lot!

PR #7243, opened 2024-03-22
UI Core: 40906 use ampersand and question mark in preg_replace

shecken:
[Mantis Ticket](https://mantis.ilias.de/view.php?id=40906)

Ampersand and question mark should be used in preg_replace to prevent false replaces.


thibsy (approved):
Thx, good catch! LGTM.

PR #7244, opened 2024-03-22
SETUP: adjust installation and update process with 'register_nic' set…

lukastocker:
…tings (#40140).

As the installation process failed with the previous changes in ilNICKeyRegisteredObjective the update and installation process was adjusted as following:

- remove the error message regarding setting 'register_nic' to false after it was already set to 'true'
- extend 'isApplicable' by checking if 'getRegisterNic' is true
- adjust the list of objectives in the installation step of the 'SetupAgent'

### DIFF
--- a/setup/classes/class.ilSetupAgent.php
+++ b/setup/classes/class.ilSetupAgent.php
@@ -88,9 +88,7 @@ class ilSetupAgent implements Setup\Agent
                 new Setup\Condition\PHPExtensionLoadedCondition("gd"),
                 $this->getPHPMemoryLimitCondition(),
                 new ilSetupConfigStoredObjective($config),
-                $config->getRegisterNIC()
-                    ? new ilNICKeyRegisteredObjective($config)
-                    : new ilNICKeyStoredObjective($config)
+                new ilNICKeyRegisteredObjective($config)
             )
         );
     }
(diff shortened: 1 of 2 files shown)


klees:
Picked to trunk.

setup/classes/class.ilSetupAgent.php
```diff
@@ -88,9 +88,7 @@ public function getInstallObjective(Setup\Config $config = null): Setup\Objectiv
                 new Setup\Condition\PHPExtensionLoadedCondition("gd"),
                 $this->getPHPMemoryLimitCondition(),
                 new ilSetupConfigStoredObjective($config),
-                $config->getRegisterNIC()
-                    ? new ilNICKeyRegisteredObjective($config)
-                    : new ilNICKeyStoredObjective($config)
+                new ilNICKeyRegisteredObjective($config)
```
chfsx, line 91: @klees $config can be null in case of updating, this now leads to a fatal. I suggest to change this to 
```
new ilNICKeyRegisteredObjective($config ?? new Setup\NullConfig())
```
 here

chfsx: not sure if something else is then needed inside `ilNICKeyRegisteredObjective`

klees: @lukastocker Please look into this.

lukastocker: > not sure if something else is then needed inside `ilNICKeyRegisteredObjective`



Hi @chfsx ,

to be sure: the $config can't be null for the installation - only for the update process, correct?

klees: Please have a look into [this PR](https://github.com/ILIAS-eLearning/ILIAS/pull/7259).

chfsx: Thanks a lot, es this seems to be the better solution in the PR. and yes, config can only be null during update

PR #7245, opened 2024-03-22
PRG: 40516, mail translations in recipient's language

nhaagen:
https://mantis.ilias.de/view.php?id=40516


klees (changes requested):
Hi @nhaagen,

LGTM in general, one change request: please cache new `ilLanguage` instances in an array with the language as key. Especially in mass action cases we would recreate `ilLanguage` over and over again, which would be quite expensive.

Thanks!

klees:
Picked to trunk.

PR #7246, opened 2024-03-22
UI: remove box-shadow from radio option's div container. (#40660)

lukastocker:
https://mantis.ilias.de/view.php?id=40660


Amstutz:
thx a lot!

PR #7248, opened 2024-03-22
LP: 40895, key error for failed_users_num

nhaagen:
https://mantis.ilias.de/view.php?id=40895

Issue is in 8, but occurs in 9, too.


smeyer-ilias (approved):
Thanks @nhaagen

PR #7249, opened 2024-03-22
UI: UI: adjust mainbar button icon filter settings... (#40694)

lukastocker:
... to avoid gyph's and icon's color be black on custom mainbar entries.

https://mantis.ilias.de/view.php?id=40694

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -7173,7 +7173,7 @@ fieldset[disabled] .navbar-inverse .btn-link:focus {
   z-index: 300;
 }
 .il-card .il-card-repository-head .icon {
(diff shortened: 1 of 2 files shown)


lukastocker:
Hi @Amstutz ,

please have a look at these changes. 

I also created a PR for v9: https://github.com/ILIAS-eLearning/ILIAS/pull/7250

Amstutz:
thx a lot for the PR. However, do to various skins already in production for 8 with custom icons, I will not merge this down to 8, due to the fact, that I am somewhat unsure, of what side effects could be caused. If somebody is willing to test that thoroughly, I would be willing to reconsider.

PR #7251, opened 2024-03-22
Ildcl unique f03

sz-tik:
mantis#30758 - uniqueness for boolean, datetime field
fixed uniqueness check for boolean field and datetime field.
slightly improved code for checking record input | field editing and the uniqueness check of a number field.


Amstutz:
thx a lot

PR #7253, opened 2024-03-22
Trunk: 39461 - Remove TargetRelease="7.2.0" from style.zip file

mBeym:
Reported in: https://mantis.ilias.de/view.php?id=39461
and Linked to issue: https://mantis.ilias.de/view.php?id=39460
Removes the xml attribute TargetRelease="7.2.0" from the style.zip archive.

This xml attribute prevents creating an individual Style on Forum Objects
(Forum Object => Settings => Style => Create individual Style)
(Probably other objects as well)


alex40724:
Thanks for the PR. This has already been fixed with https://mantis.ilias.de/view.php?id=39882

Note that the old export files should still work, so no need to remove the TargetRelease attribute here.

PR #7255, opened 2024-03-22
PRG: 40853, listen to LP TrackingUpdateStatus for courses instead of course references

nhaagen:
https://mantis.ilias.de/view.php?id=40853
https://mantis.ilias.de/view.php?id=40919

I moved the former ilObjStudyProgramme::setProgressesCompletedFor to the event handler to exit max early and be very clear about the kind of events we want to track.

Also, completed_by now stores the id of a course instead of the reference. Migration included.


klees:
Hello everyone,

to document this change thoroughly I want to give a little history:

* Traditionally the Study Programme never cared about settings of the contained Course References, especially not about "Adopt Learning Progress from Course". Why would someone put a Course Reference into a Study Programme but would not want it to count for the Study Programme. Hence, Study Programme listened to the event system itself and did not rely on the Course Reference to transmit learning progress events.
* In [#37921](https://mantis.ilias.de/view.php?id=37921) it was reported, that Course References in Study Programmes would work anymore if the setting "Adopt Learning Progress from Course" was not set. I should have noticed that this was kind of unlikely, since Study Programme never relied on the Course Reference in the first place, as pointed out before. I didn't notice and we changed the Study Programme to listen to learning progress events from the Course and hence respect the "Adopt Learning Progress from Course" setting.
* During the testing of ILIAS 9, [problems with preexisting Courses that were linked to Study Programmes](https://mantis.ilias.de/view.php?id=40853) were discovered. We found, that we could not solve them with the indirection via the Course Reference that we have introduced. The problem is: The learning progress event will be created in the very moment that "Adopt Learning Progress from Course" is switched on. Hence, Study Programme cannot know that a user in fact had a preexisting learning progress at the Course. Also I understood that the change outlined in the item before was in fact strange and not required.
* Hence, we decided to rollback the change so Study Programme would ignore "Adopt Learning Progress from Course" again.

I will merge the change already to move on with testing of ILIAS 9, but I will bring this to the attention of the Jour Fixe, since the problem was discussed there before.

Kind regards!

klees:
Picked to trunk. I will pick this to  release_8 once we have discussed this in the JF.

klees:
Dear @matthiaskunkel,

I would like to discuss this briefly on the JF.

Thanks!

matthiaskunkel:
Jour Fixe, 15 APR 2024: @klees notified us that he will undo the decided changes about the tracking update status due to a conceptual problem. These changes will be made for 8 and 9 (and trunk) but not for 7.

klees:

klees:
Thx @slowe47 for being so attentive and reminding me.

PR #7257, opened 2024-03-25
PRG: 40877, set back_url for user profile from automations

nhaagen:
https://mantis.ilias.de/view.php?id=40877


klees:
Picked to trunk.

PR #7259, opened 2024-03-25
SETUP: check if  is null for NIC during the update process. (#40140)

lukastocker:


klees:
Picked to trunk.

PR #7260, opened 2024-03-25
Group/Course: Fix Saving of Object Properties

kergomard:
This is not an easy fix as it comes from the interplay of Legacy-Forms and UI-KS-Forms.

See: https://mantis.ilias.de/view.php?id=40928


kergomard:
I close this as I found another fix. Thanks @chfsx for planting the right idea. Fixed with: https://github.com/ILIAS-eLearning/ILIAS/commit/a5668cd62ac061c46a144ab610e63a7c884dfc91

PR #7261, opened 2024-03-25
9/UI/ta 40639 auto save msgbox

catenglaender:
# Issue

Empty auto-save message leaves border pixels on screen, when the message should be completely invisible.
image

It's also not looking like the other alert and message boxes when it shows up:
image

# Change

An :empty message box is now set to display: none; and should no longer show up. (If it shows up anyway it's because JavaScript or php leaves a whitespace " " string behind - this should be reported as a separate bug).

The design is now using uitility classes and shares new setting layer variables from the UI component message box and the UI mode info. Basically, we make a legacy div look like a UI component message box.

image

# Larger context

* We need to discuss in the css-squad: How do we make legacy components look like UI components if refactoring them properly is not feasible (at the moment)?
    * extending legacy components into UI components could be okay as an exception in this case
    * OR using UI component css classes as utility classes in html templates of legacy components

### DIFF
--- a/Modules/Test/ROADMAP.md
+++ b/Modules/Test/ROADMAP.md
@@ -34,6 +34,7 @@ Depending on development ressources, funding and approval, some new features are
 * UI Forms: Length of text fields not shorter than answer text. (see: [Mantis 35243](https://mantis.ilias.de/view.php?id=35243#c87241))
 * Grading: Use of Short Form / Offical Form. Conceptual changes only with the introduction of new UI forms. (see [Mantis](https://mantis.ilias.de/view.php?id=31209#c83984))
 * Streamlining of trimming and removal of white spaces in answer-options (see: [Mantis 35091](https://mantis.ilias.de/view.php?id=35091))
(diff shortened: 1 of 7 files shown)


dsstrassner:
I have added me for statistic purposes. @Amstutz & @kergomard are in charge, to check who will review and who will merge this PR ;-)

Thanks @catenglaender!

Amstutz:
@dsstrassner : We discussed this in the CSS Squad today, we agree that the version that @catenglaender is proposing here is the way to go. However, we also agree that we like to replace that in future with regular UI Components. 

- [x] @catenglaender , please add an according entry to the TA Roadmap.

Kindly,
your CSS Squad.

dsstrassner:
We also decided that @Amstutz will merge this PR when the task is fulfilled. I discussed this with one of the TA code authorities (@kergomard) and he agreed 😁.

catenglaender:
Updating auto-save messsages to UI components has been added to roadmap :slightly_smiling_face:

Amstutz:
thx a lot!

PR #7262, opened 2024-03-25
#40939: Fixing learning modules check for Anchor link to fix.

dmundra:
Fixing https://mantis.ilias.de/view.php?id=40939

This fix is similar to one applied https://github.com/ILIAS-eLearning/ILIAS/commit/9718e733af93371b22f0a779674edfb7710e31e0


alex40724:
Thanks for the PR!

PR #7263, opened 2024-03-26
PRG: 40918, directly mark accredited progress as irrelevant

nhaagen:
https://mantis.ilias.de/view.php?id=40918


klees:
Picked to trunk.

PR #7266, opened 2024-03-26
#40960 fixing check for ilAccordionsInits to use the correct variable name.

dmundra:
Fixes https://mantis.ilias.de/view.php?id=40960


alex40724:
Thanks for the PR. This is fixed since some time.

dmundra:
Ah I see it was fixed in https://github.com/ILIAS-eLearning/ILIAS/commit/1819dcc2f7c9890e8c51c57179025a9a3cde4289. Thanks!

PR #7267, opened 2024-03-26
LegacyForm/FileWizard: Use UI/KS glyphs

mjansenDatabay:
This PR removes the legacy UI glyph usages from `ilFileWizardInputGUI`.

I had to rewrite the JS code, because glyphs inside `a` elements inside `button` elements are **not** valid in `HTML`. `button` elements were used before as trigger element for the different actions like `add`, `remove`, `move up` and `move down`.

> The element a must not appear as a descendant of the button element.

### DIFF
--- a/components/ILIAS/Form/classes/class.ilFileWizardInputGUI.php
+++ b/components/ILIAS/Form/classes/class.ilFileWizardInputGUI.php
@@ -28,6 +28,8 @@ use ILIAS\Filesystem\Util;
 class ilFileWizardInputGUI extends ilFileInputGUI
 {
     protected ilGlobalTemplateInterface $tpl;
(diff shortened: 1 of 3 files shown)


mjansenDatabay:
Hi @mjansenDatabay ,

thx for your pull request. Unfortunately, there is currently no one who feels responsible for this code. We will discuss this kind of PR during the next JF as part of a new agenda point "Looking for Shepherd".

Best regards,
Michael

Amstutz:
@mjansen and @Amstutz are both not pleased with the use of spans for wrapping the glyphs. We looked into using bulky buttons here; however, we came to the conclusion that the resources needed to make the necessary JS changes are not worth it in this particular case, since we want to abandon ilFileWizardInputGUI anyway.

PR #7268, opened 2024-03-27
Tracking: Fix null value passed to `ilTrQuery::executeQueries`

mjansenDatabay:

Mantis Issue: https://mantis.ilias.de/view.php?id=40974


smeyer-ilias (approved):
Thanks @mjansenDatabay

PR #7269, opened 2024-03-27
UIComponent: File Upload Input: Add visible file extension option

lscharmer:
Follow up to PR #7090.

This PR proposes to add an additional parameter to the `withAcceptedMimeTypes` to set the list of file extensions which should be shown to the user. If this is not set or set to `null` the file extensions are derived from the mime types. This only affects which file extensions are shown to the user, not which ones the user can select.

### DIFF
--- a/components/ILIAS/UI/src/Component/Input/Field/FileUpload.php
+++ b/components/ILIAS/UI/src/Component/Input/Field/FileUpload.php
@@ -43,9 +43,15 @@ interface FileUpload
     public function getMaxFiles(): int;
 
     /**
+     * If $visible_extensions is null the file extensions shown to the user will be derived from $mime_types.
+     * If $visible_extensions is not null it will be shown to the user.
+     * The $visible_extensions do not change which files can be selected, they are only a visible indicator.
+     * $visible_extensions must be of the form: ['txt', 'html'].
+     *
      * @param string[] $mime_types
+     * @param null|string[] $visible_extensions
      */
-    public function withAcceptedMimeTypes(array $mime_types): FileUpload;
+    public function withAcceptedMimeTypes(array $mime_types, ?array $visible_extensions = null): FileUpload;
 
     /**
      * @return string[]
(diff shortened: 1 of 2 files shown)


thibsy (changes requested):
Hi @lscharmer

Please note this review will only be focused on the interface.

- [x] Default behaviour: by your description, providing `null`, or no `$visible_extensions`, will automatically derive file extensions from the provided mime types. What happens if this method is not called? And, how are we going to derive these file extensions exactly? If you plan on using `ILIAS\FileUpload\MimeType`, we would expect it to be wrapped, or moved to `ILIAS\Data` in some way.

- [x] `FileUpload::withAcceptedMimeTypes()`: please provide a concise PHPDoc annotation about the type of `$visible_extensions` and how it will be used - when provided and not.
- [x] `FileUpload::getVisibleExtensions()`: please remove this method from the interface again. We try to keep our interfaces minimal and this method can be kept internal.

Kind regards,
@klees and @thibsy

lscharmer:
> * [ ] Default behaviour: by your description, providing `null`, or no `$visible_extensions`, will automatically derive file extensions from the provided mime types. What happens if this method is not called? And, how are we going to derive these file extensions exactly? If you plan on using `ILIAS\FileUpload\MimeType`, we would expect it to be wrapped, or moved to `ILIAS\Data` in some way.

If this method is not called (or to be more specific: If `getAcceptedMimeTypes` returns an empty array), I would suggest to not show any file extensions. Alternatively we could use a language variable which says that all types are allowed.

Yes, currently I plan to derive the file extensions with `ILIAS\FileUpload\MimeType`. But for the implementation I will create another PR. Isn't this a topic for that PR?

thibsy (changes requested):
Hi @lscharmer

Thank you very much for the explanation(s) and your patience on this. 

You can and should definitely open another PR to incorporate the `ILIAS\FileUpload\MimeType` into the `ILIAS\Data` namespace. However, we cannot merge this PR until there is another implementation in place than the "legacy" one.

- [ ] Accepting `null`: I think we can drop the third state of this argument. The current solution would disable this feature by default, and would get rid of the chance to force consumers into providing important metadata about their inputs. We also present the consumer with a design choice here, for which we'd like to have as few options as possible. I therefore suggest to always derive from `$mime_types`, if the argument is present and no extensions are provided (empty array).

Kind regards,
@klees and @thibsy

PR #7271, opened 2024-03-27
IASS: adjust access to members and learning progress tab. (#40930)

lukastocker:
https://mantis.ilias.de/view.php?id=40930

The members tab should only be visible if the user has reading rights. "edit_learning_progress" has nothing to do with this tab as it should be read as "edit_learning_progress_settings".
As result a new function, which will check if the user had permission to edit the learning progress settings, had to be added for the learning progress tab.


klees:
Picked to trunk.

PR #7276, opened 2024-03-28
Test: Add permanent link for question tab

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40931


dsstrassner:
This ILIAS8 only for the test, as the JF decided, that the permanent link in ILIAS9 above should go to the Test-tab / Launcher.

PR #7278, opened 2024-03-28
QuestionPool: Add permanent link for question tab

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40931


dsstrassner:
Hi @mbecker-databay - thanks for merging. Could this PR (and only this PR - not the one for the test 😀 ) be picked to ILIAS 9 and trunk also? This would be nice and helpful. Thanks in advance!

PR #7280, opened 2024-04-02
PRG: 41004, sort by completeness and validity for status info

nhaagen:
https://mantis.ilias.de/view.php?id=41004


klees (changes requested):
Hi @nhaagen,

I think this logic belongs into the repository class, which would also allow to have corresponding information in the status info object and on the dashboard. Is there a specific reason why this cannot go into the repo class?

Kind regards!

klees:
Picked to trunk.

PR #40, opened 2015-06-16
Custom map-data server selection for OpenLayersMaps

dkloepfer:
Implemented changes as suggested in http://www.ilias.de/docu/goto_docu_wiki_wpage_3528_1357.html.

PR #7291, opened 2024-04-02
UI: adjust rendering cases of the required text above formulars. (#40…

lukastocker:
…885)

https://mantis.ilias.de/view.php?id=40885


Amstutz:
thx a lot!

PR #7292, opened 2024-04-02
9/UI/listing restore previous design 40050

catenglaender:
https://mantis.ilias.de/view.php?id=40050

# Issue

image

# Change

A custom design for the test result was conflicting with the listings default look. These lines of code (once written by me, I must confess, to polish the look of the test result) are now removed.

image

@dsstrassner, as there is no way anymore to detect if the presentation table with the listing characteristics is rendered in the context of the test result, the look had to be set to the listing's default design everywhere. However, I am in contact with @yvseiler discussing improvements for the listing characteristics. Right now key and value are too far apart on desktop and we need to fix that globally not just for the test result.

This is how the test result looks now with the characteristics listing's default appearance:

image

### DIFF
--- a/templates/default/070-components/legacy/Modules/_component_test.scss
+++ b/templates/default/070-components/legacy/Modules/_component_test.scss
@@ -121,28 +121,3 @@ $il-test-working-time-font-weight: $il-font-weight-bold;
 #tst_pass_details_overview tr {
     scroll-margin-top: 30px;
 }
-
-//
-// test result
-//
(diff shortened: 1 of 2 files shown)


dsstrassner:
This change is fine for me for the moment 👍  I am sure, that you will find an even better solution over the time.

Amstutz:
I checked this out with @yvseiler and we agree that this is ready to be merged.

PR #7295, opened 2024-04-03
LegalDocuments: Change permission exceptions to `$ilErr->raiseError`

lscharmer:
Fix Mantis Bug: https://mantis.ilias.de/view.php?id=41009


dsstrassner:
Thanks

PR #7298, opened 2024-04-03
Language: Add NullLanguage class

lscharmer:
I'm currently fixing missing properties in the Refinery.
Among others the `ILIAS\Refinery\To\Transformation\StringTransformation` and `ILIAS\Refinery\To\Transformation\ListTransformation`  are missing a `$lng` property (which is used in the `ILIAS\Refinery\ProblemBuilder::getLngClosure()`).

The problem is that these two transformations are used in the templating class `IT` and there it is not always guaranteed (AFAIK) that `$DIC->language()` is available, so I cannot require the `ILIAS\Language\Language` interface in the constructor of those transformations.
Therefore I would like to add a `NullLanguage` which can be used instead.

### DIFF
--- a/components/ILIAS/Language/src/NullLanguage.php
+++ b/components/ILIAS/Language/src/NullLanguage.php
@@ -1,0 +1,33 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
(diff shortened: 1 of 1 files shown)


katringross:
@lscharmer Like I can see this only happens inside of the Refinery. I am not happy to create an extra class for a NullLanguage only for templating classes. 

Wouldn't it be better to create a local language variable within the Closure.

klees:
Hi @lscharmer,

Kind regards!

lscharmer:
Hi @kless,
thanks to bringing this to my attention, as this got lost in time and space.

I would kindly ask if you could review this PR, as I would use this for the Refinery to solve the following problem:

As mentioned in the PR description, some classes in the Refinery have undefined properties.
In an attempt to fix them I noticed that some of them are instantiated directly (e.g. `new ILIAS\Refinery\To\StringTransformation()`).
As AFAIK we cannot guarantee that `$DIC->language()` is available in all of these cases (e.g. in the template classes) I propose a `NullLanguage` class which can be used in those cases.

I discussed this last year with @katringross and we decided to request your conceptional approval before merging this (@katringross feel free to comment if I got something wrong or something changed). If you (@klees) object to this PR you can simply close this PR.

If this PR is merged I will create corresponding followup PRs for the Refinery.

Best regards
Lukas

PR #6539, opened 2023-11-03
Dashboard refactor for ILIAS9 / PHP8.1

iszmais:
(This PR was recreated since its origin fork was deleted: https://github.com/ILIAS-eLearning/ILIAS/pull/5931)

Major Changes:
- [x] Fixed Licencse Header
- [x] Remove/updated trivila/missleading docs
- [x] and `final` and `readonly`
- [x] update constructors
- [x] applied single quitation
- [x] applied short array syntax
- [x] removed trivial local variables
- [x] applied dynamic called classnames
- [ ] removed unused functions
- [ ] enhanced type safety
- [ ] removed `class` prefixes
- [ ] removed `a_` prefixes
- [ ] partially changed names from desktop to dashboard
- [ ] minor bugfixes
- [ ] applied PSR-12

PR #7301, opened 2024-04-03
Apply single quotes and short arrays in dashboard

iszmais:
A partial approach on refactoring the Dashboard (see https://github.com/ILIAS-eLearning/ILIAS/pull/6539):
- applied single quotation
- applied short array syntax

### DIFF
--- a/Services/Dashboard/Access/class.DashboardAccess.php
+++ b/Services/Dashboard/Access/class.DashboardAccess.php
@@ -46,17 +46,17 @@ class DashboardAccess
                 AND object_data.type = %s
                 AND object_reference.ref_id = tree.child
                 AND object_reference.obj_id = object_data.obj_id',
-                array('integer', 'text'),
-                array(SYSTEM_FOLDER_ID, 'dshs')
+                ['integer', 'text'],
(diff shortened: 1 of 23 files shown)


Services/Dashboard/Access/class.DashboardAccess.php
```diff
@@ -46,17 +46,17 @@ protected function getSettingsRefId(): int
                 AND object_data.type = %s
                 AND object_reference.ref_id = tree.child
                 AND object_reference.obj_id = object_data.obj_id',
-                array('integer', 'text'),
-                array(SYSTEM_FOLDER_ID, 'dshs')
+                ['integer', 'text'],
```
lscharmer, line 49: Can you change this as well? As you are changing this line anyway?
```suggestion
                [ilDBConstants::T_INTEGER, ilDBConstants::T_TEXT],
```

lscharmer (approved):

Services/Dashboard/Access/class.DashboardAccess.php
iszmais: Im aware of this.
I will tackle this issue in another PR to keep the overview of the refactoring changes.

PR #7306, opened 2024-04-03
UIComponent: Render File Upload: Show Max nr of files and upload size

lscharmer:
This PR addresses one part of Mantis Issue: https://mantis.ilias.de/view.php?id=39689

This PR displays the max nr of files and the max upload size below the file input field.

The info list may be extended to show the accepted file extensions in a future PR.

### DIFF
--- a/components/ILIAS/UI/src/Implementation/Component/Input/Field/Renderer.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Input/Field/Renderer.php
@@ -797,6 +797,8 @@ class Renderer extends AbstractComponentRenderer
             $file_preview_template
         );
 
+        $this->setHelpBlockForFileField($template, $input);
+
         $input = $this->initClientsideFileInput($input);
         $input = $this->initClientsideRenderer($input, $file_preview_template->get('block_file_preview'));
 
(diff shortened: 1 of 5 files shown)


klees (changes requested):
Hi @lscharmer,

thanks for looking into this.

* [x] Output: Please do not construct strings containing meaningful parts in the renderer but instead via the mechanism we have for that: the template. Introduce separate placeholders for the new labels and values, we will also be able to address possible accessibility or styling changes easily then.

Best regards!
@thibsy, @Amstutz, @klees

thibsy (changes requested):
Hi @lscharmer

Thx a lot for incorporating @klees requested changes. I just looked over this again, and there are still some questions/remarks about this, before we can merge this:

- [x] this is not yet a mandatory, but we will need to limit our `JavaScriptBindable` components to one top-level HTML element in the future. Is it therefore possible, to move the additional info into the current top-level element?
- [x] you are escaping two values of which we certainly know there will be no special characters inside them. Please remove both `htmlspecialchars()` calls.

Kind regards,
@thibsy

thibsy (approved):
Hi @lscharmer,

Thx for implementing the requested changes.

We noticed you are using the `|` character as a separator, which may not be accessibility-friendly and could be flagged in an audit.

Otherwise this LGTM now, so I'll merge this for the time being.

Kind regards,
@thibsy

PR #7310, opened 2024-04-03
WAC/Init: Provide a possible fix for Mantis 39817

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=39817

This PR provides a workaround for WAC issues where the initialisation is executed twice (see: `\ilWebAccessChecker::initAnonymousSession`) and the super global variables (GPRC) are already replaced with `ArrayAccess` implementations.

Instead of booting the system twice, only the session- and user-based initialisation parts are called again for the case, where the WAC has to initialise an anonymous session.

@Uwe-Kohnle I had to change one method signature of your `ilInitialisation` derivative in the `CmiXapi` component.

### DIFF
--- a/Modules/CmiXapi/classes/XapiProxy/DataService.php
+++ b/Modules/CmiXapi/classes/XapiProxy/DataService.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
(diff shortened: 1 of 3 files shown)


chfsx (approved):
The WAC changes are fine, thanks a lot!

chfsx:
@pascalseeland you my merge after your review

PR #7311, opened 2024-04-03
9/UI/TA kiosk header infos

catenglaender:
https://mantis.ilias.de/view.php?id=38857

# Issue

In the kiosk mode of the Test & Assessment, long titles and names can lead to a broken header layout:
image

# Changes

* SCSS tool Elementbar now exposes root CSS variables that can be overridden locally. In this case elementbars in the ILIAS standardpage header get more narrow spacings to make the most of the limited space. This is a proof of concept. We need to discuss in the CSS Squad if and under which conditions we allow SCSS to CSS variable exposure.
* Switched from the Alignment Layout component to a Property Listing UI component in php to follow the Key-Value nature of the information being rendered.
* The following is a temporary solution and should be changed in ILIAS 10 (see outlook): The participant's name was supposed to appear highlighted as during in person exams a supervisor might want to glance over people's shoulders to quickly check the name. The Property Listing doesn't know such a highlighted property, so html tags had to be sneaked into that string in php. A general css class allows overriding through system styles. Not sure if a one line html template might be the better emergency solution.

image

# Outlook

I would suggest the creation of a header component for ILIAS 10 that can handle properties with various priorities and icons similar to the UI entity component. Such a header component could be used to generate not only this system header, but also potentially the header for content pages (this is legacy atm) or even panels and form sections.

### DIFF
--- a/Modules/Test/classes/class.ilTestOutputGUI.php
+++ b/Modules/Test/classes/class.ilTestOutputGUI.php
@@ -960,26 +960,33 @@ abstract class ilTestOutputGUI extends ilTestPlayerAbstractGUI
 
     protected function getTestPlayerTitle(): string
     {
-        $test_title = $this->object->getShowKioskModeTitle() ? $this->object->getTitle() : '';
-        $user_name = $this->object->getShowKioskModeParticipant() ? $this->user->getFullname() : '';
-        $exam_id = '';
+        $titleContent = $this->ui_factory->listing()->property();
+
+        if ($this->object->getShowKioskModeParticipant()) {
+            $testParticipantNameLabel = $this->lng->txt("conf_user_name");
+            // this is a placeholder solution with inline html tags to differentiate the different elements
+            // should be removed when a title component with grouping and visual weighting is available
+            // see:  https://github.com/ILIAS-eLearning/ILIAS/pull/7311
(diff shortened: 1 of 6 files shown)


Amstutz:
Hi @catenglaender,

Thx a lot for the proposal to resolve the issue and the detailed explanation. This helped a lot to follow the reasoning of proposing this temporary solution, and the thoughts behind it, as well as the CSS changes. Highly appreciated.

As you might guess, we especially dislike the use of inline DOM usage. However, we understand that this is the only way to resolve the issue as proposed in the tickets with the tools available. However, before we merge this, please:

- [x] Roadmap: And an entry in the UI Components Roadmap in which you describe the outlook sketched out here, as well a link to the issue in ilTestOutputGUI .

thx a lot.

Amstutz:
@kergomard and/or @dsstrassner feel free to merge as soon as the roadmap entry is added. the UI Component part is fine otherwise.

catenglaender:
UI Component Roadmap edited (adding to Title Section) and inline comment in php. :heavy_check_mark: 
Should be ready for merge now.

dsstrassner:
@kergomard would you be so nice?

kergomard:
Thank you very much @catenglaender  and sorry, lost this out of view!

Merged and picked to trunk

Best,
@kergomard

PR #7315, opened 2024-04-04
User/Password: Don't pass `null` when determining the user's password…

mjansenDatabay:
… encoder

If approved, this must be picked to `trunk` and `release_9`.

Mantis Issue: https://mantis.ilias.de/view.php?id=41030


kergomard:
Again: Sorry @mjansenDatabay ! I was apparently very blind two days ago. Already fixed otherwise.

PR #7316, opened 2024-04-04
Update visibilities properties and constructor in Dashboard

iszmais:
A partial approach on refactoring the Dashboard (see https://github.com/ILIAS-eLearning/ILIAS/pull/6539):
-  add `final` and `readonly`
-  update constructors
- removed trivial local variables

### DIFF
--- a/Services/Dashboard/Achievements/classes/class.ilAchievements.php
+++ b/Services/Dashboard/Achievements/classes/class.ilAchievements.php
@@ -22,16 +22,12 @@ use ILIAS\DI\Container;
 
 class ilAchievements
 {
-    private ilCertificateActiveValidator $validator;
-    protected ilLearningHistoryService $learing_history;
-
     public const SERV_LEARNING_HISTORY = 1;
     public const SERV_COMPETENCES = 2;
     public const SERV_LEARNING_PROGRESS = 3;
     public const SERV_BADGES = 4;
     public const SERV_CERTIFICATES = 5;
-
-    protected array $services = [
+    private const SERVICES = [
         self::SERV_LEARNING_HISTORY,
         self::SERV_COMPETENCES,
         self::SERV_LEARNING_PROGRESS,
@@ -39,16 +35,16 @@ class ilAchievements
         self::SERV_CERTIFICATES
     ];
 
-    protected ilSetting $setting;
-    protected ilSetting $skmg_setting;
+    protected readonly ilCertificateActiveValidator $validator;
+    protected readonly ilLearningHistoryService $learing_history;
+    protected readonly ilSetting $setting;
 
     public function __construct()
     {
         global $DIC;
 
-        $this->setting = $DIC->settings();
         $this->learing_history = $DIC->learningHistory();
-        $this->skmg_setting = new ilSetting('skmg');
+        $this->setting = new ilSetting('skmg');
         $this->validator = new ilCertificateActiveValidator();
     }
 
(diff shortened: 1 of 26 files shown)


iszmais:
@alex40724 @klees The refactoring also contains some minor changes within the depending classes in `RecommendedContent`, `StudyProgramme` and `LearningSequence`. This is just cosmetic/conventional nature and does not interefere with any process. Pls have a short look.

Greetings,
@iszmais

Services/Dashboard/Achievements/classes/class.ilAchievements.php
```diff
@@ -22,44 +22,40 @@
...
         $this->learing_history = $DIC->learningHistory();
-        $this->skmg_setting = new ilSetting('skmg');
+        $this->setting = new ilSetting('skmg');
         $this->validator = new ilCertificateActiveValidator();
     }
 
     public function isActive(int $service): bool
     {
-        switch ($service) {
+        switch (self::SERVICES) {
```
lscharmer: This seems wrong. I don't think you want to match upon the `self::SERVICES` array but on the `$service` integer.

Services/Dashboard/Achievements/classes/class.ilAchievements.php
```diff
@@ -90,7 +86,7 @@ public function isAnyActive(): bool
      */
     public function getActiveServices(): array
     {
-        return array_filter($this->services, function ($s) {
+        return array_filter(self::SERVICES, function ($s) {
```
lscharmer:
```suggestion
        return array_filter(self::SERVICES, $this->isActive(...));
```

lscharmer (changes requested):

iszmais:
@alex40724 @klees 
Just a ping i case this here was overlooked.

The taken changes in your components are quite small:
- changes property from `viewSettings` to `view_settings`
- changed function name from `getCurrentView` to `getView` (since there is only 1 view)
- applied 2 final states to function

Therefore i assume there is not much to review at all, but i would reluctantly merge this without your confirmation in respect to your maintainance/authority.

Greetings,
@iszmais

lscharmer (approved):

iszmais:
:bell:

iszmais:
@klees @alex40724 :bell:

klees (commented):
Hi @iszmais,

I'd rather not use `final`, at least in my components. The feature feels like a good idea at the beginning, but from my experience quickly is in the way of automated testing.

This is more of a preference here, so I leave the decision up to you. If you decide to stay with the current code, I would kindly ask for a small explanation, why these methods should be final.

Feel free to merge.

Kind regards!

iszmais:
@klees Thank you for your feedback.

I used the `final` keyword to "guide" future progress on these providers by promoting a singular (or dual?) structure in this provider system. This should prevent the creation of inheritance clusters and intransparent structures as well as component internal duplication, especially on "critical" functionality like management operations or general identification.
This should further not impede the testing, but, on the contrary, help it by structuring the conceptional requirements on this provider system within the code basis. This way we could prevent numb testing and make sure the test themself doesn't encourage behavior that contradicts the general concept.

However, I'm aware that the `final` keyword only tackles this issue partially and works very one-sided.
Due to that, it's no requirement to me either, just a suggestion.

Therefore, I removed it from your supervised components.

Greetings,
@iszmais

klees (approved):

iszmais:
> Hi @iszmais,
>
> the Technical Board looks into oldish PRs regularly, this is when this PR came up. Is there anything that we can do here to  move this to some definite state, be it a merge or close?
>
> Kind regards!

@alex40724 / @smeyer-ilias All eyes on you. :smile: 
As soon as you can give your approval to the changes in `Repository` (Recommended Content), this will be ready to merge.

Greetings,
@iszmais

iszmais:
@alex40724 / @smeyer-ilias :bell:

iszmais:
@alex40724 / @smeyer-ilias :bell:

alex40724 (approved):

iszmais:
Thank you all for approval!

Unfortunately, this PRs origin target release went out of support over the review process.

Therefore, I will merge this for ILIAS 10+ and close this PR.

Greetings,
@iszmais

PR #7319, opened 2024-04-05
T&A: 39719, Fixed Type Error in TestEvaluationUserData concerning login property

malaja11:
Will be taken care of by @mbecker-databay and @thojou 

Possible solution for Ticket https://mantis.ilias.de/view.php?id=39719 and https://mantis.ilias.de/view.php?id=40938

### DIFF
--- a/Modules/Test/classes/class.ilTestEvaluationUserData.php
+++ b/Modules/Test/classes/class.ilTestEvaluationUserData.php
@@ -34,7 +34,7 @@ class ilTestEvaluationUserData
 {
     private $questionTitles;
     public string $name;
-    public string $login;
+    private string $login;
     public int $user_id;
     protected bool $submitted;
     public float $reached;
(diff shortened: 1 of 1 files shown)


matheuszych:
We could not reproduce the error but we think this is a possible solution when going by the error stacktrace.

mbecker-databay:
I have to check and see how we can get Thomas on the list of assignees here.

dsstrassner:
hi @mbecker-databay, I think we could assign @thojou if he at least provide one Pull Request to the ILIAS Repo. It was possible to assign @matheuszych / Adding as reviewer needs more rights I assume ;-)

dsstrassner:
Hi all, I close this PR, as the linked mantis issues are also closed, since the error seems to be fixed meanwhile. 
This helps me not to lose track of things. Please reopen if certain changes are still needed.

PR #7321, opened 2024-04-08
T&A: 40740, Made reachedPoints getter return float instead of int

matheuszych:
Will be reviewed by @mbecker-databay &  @thojou

Solution for https://mantis.ilias.de/view.php?id=40740#c103286

### DIFF
--- a/Modules/Test/test/ilTestEvaluationUserDataTest.php
+++ b/Modules/Test/test/ilTestEvaluationUserDataTest.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 /**
  * Class ilTestEvaluationUserDataTest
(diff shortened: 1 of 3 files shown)


matheuszych:
The pipeline fails because of the issue mentioned in this PR

https://github.com/ILIAS-eLearning/ILIAS/pull/7323

mjansenDatabay:
> The pipeline fails because of the issue mentioned in this PR
> 
> #7323

This is not an issue anymore ;-).

Modules/Test/test/ilTestEvaluationUserDataTest.php
```diff
@@ -346,13 +346,32 @@ public function testGetQuestionCount(): void
...
+        $pass->setReachedPoints($input);
+        $this->testObj->addPass(0, $pass);
+
+        $this->assertEquals($expected, $this->testObj->getReachedPoints());
+    }
+
+    public function reachedPointsDataProvider(): array
+    {
+        return [
+            "float" => [25.0, 25.0],
```
thojou: To prevent future issues and address the bug description effectively, I recommend adding an additional test case to cover floating point cutoff resulting from integer cast.

```suggestion
            "float X.0" => [25.0, 25.0],
            "float X.Y" => [25.3, 25.3],
```

thojou (approved):

thojou:
Looks good to me! 

@mbecker-databay Since the issue is already resolved in release_9 and trunk, there's no need for a cherry-pick. Can we proceed with merging this?

PR #7322, opened 2024-04-08
T&A: 35998, Test Results in Points now showing correct value for current test pass

matheuszych:
Will be reviewed by @mbecker-databay & @thojou

Solution for https://mantis.ilias.de/view.php?id=35998#c101835

### DIFF
--- a/Modules/Test/classes/class.ilTestEvaluationGUI.php
+++ b/Modules/Test/classes/class.ilTestEvaluationGUI.php
@@ -1008,6 +1008,7 @@ class ilTestEvaluationGUI extends ilTestServiceGUI
         $template->setVariable("PASS_DETAILS", $this->ctrl->getHTML($overviewTableGUI));
 
         $data = &$this->object->getCompleteEvaluationData();
+        $data->getParticipant($active_id)->setPassScoring($pass);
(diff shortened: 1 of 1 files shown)


matheuszych:
@mbecker-databay This bug also persists in release_8 thus this fix should also be added there.

Modules/Test/classes/class.ilTestEvaluationGUI.php
```diff
@@ -1008,6 +1008,7 @@ public function outParticipantsPassDetails()
         $template->setVariable("PASS_DETAILS", $this->ctrl->getHTML($overviewTableGUI));
 
         $data = &$this->object->getCompleteEvaluationData();
+        $data->getParticipant($active_id)->setPassScoring($pass);
```
thojou, line 1011: For clarity:

The issue in release_7 is escalated in Modules/Test/classes/class.ilTestEvaluationUserData.php, specifically in lines 378 - 385:

```php
public function getScoredPass()
{
    if ($this->getPassScoring() == 1) {
        return $this->getBestPass();
    } else {
        return $this->getLastPass();
    }
}
```

Currently, this method always returns the result from `getLastPass` because `getPassScoring` is not set earlier.

This change addresses this issue and should resolve it accordingly.

thojou (approved):
As pointed out by @matheuszych, the issue also exists in the release_8 branch. After merging, we should also check release_9 and trunk branches to ensure the problem is resolved across all versions.

@mbecker-databay could you please also review and merge these changes?

PR #7324, opened 2024-04-08
CategoryLink: Add workaround for undeterminable ref_id

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41063

### DIFF
--- a/Modules/CategoryReference/classes/class.ilObjCategoryReferenceListGUI.php
+++ b/Modules/CategoryReference/classes/class.ilObjCategoryReferenceListGUI.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 use ILIAS\ContainerReference\StandardGUIRequest;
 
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
Hi @smeyer-ilias , do you have any objections here?

alex40724:
@mjansenDatabay The fix looks suspicous and as written in the bug report, other object types may be affected as well. I think this needs more investigation "accordion to prio".

mjansenDatabay:
Hi @alex40724 ,

thanks a lot for your feedback.

> I think this needs more investigation "accordion to prio".

So do you (or Stefan) take care of that (according to your priority, which is fine for me), or should I provide more changes for the other "Reference-Object Types"? The latter would only make sense if you see any chance to integrate/merge this. I could adapt this change for all `ilObj*ReferenceListGUI` classes, introduce a new base class for them, or create a `trait`

Best regards,
Michael

Modules/CategoryReference/classes/class.ilObjCategoryReferenceListGUI.php
```diff
@@ -111,9 +111,9 @@ public function initItem(
         $target_title = ilContainerReference::_lookupTitle($obj_id);
         $target_description = ilObject::_lookupDescription($target_obj_id);
 
-        $this->deleted = $tree->isDeleted($target_ref_id);
+        $this->deleted = !$target_ref_id || $tree->isDeleted($target_ref_id);
```
alex40724, line 114: @mjansenDatabay I am wondering, if this line is correct. Are there other cases where we got not target ref id disregarding the object deletion. But maybe we just go with the assumption that it is ok to treat it like the deletion case.

I think a trait would be the best solution for this.

mjansenDatabay: @alex40724

This is most probably caused by deleted target objects.

AFAIR there are existing mechanisms (relying on event handling) to prevent these orphaned database references in the `ilContainerReference` object types, but it's PHP (means crashes can easily occur depending on memory limits, max. exec. time, error-causing plugins, etc.) and we don't use a RI.

This PR is only meant to be a workaround. Better options could be: Providing a cron job which regularly deletes orphaned data (if existing).

The line itself is IMO correct. If the `$target_ref_id` cannot be evaluated as a boolean `true`, it can be considered (IMO) as `deleted/missing`.

-- UPDATE --

Maybe we could provide a self-healing mechanism here instead. If the target ref_id is deleted, we could delete the entry without the need of a cron job (like suggested above).

alex40724:
I went for this style with the containers that included this line. Thanks for the PR. I hope that we will get foreign key constraints funded some time.

PR #7325, opened 2024-04-08
Fix deletion of sub post in Forum when thread is deleted

iszmais:
https://mantis.ilias.de/view.php?id=40242


iszmais:
@mjansenDatabay this now fixes the event instead of sending multiple. pls have a look.

PR #7327, opened 2024-04-09
T&A: 40962, Adjusted labelling for numeric gap combination in Cloze question

matheuszych:
Will be reviewed by @thojou & @mbecker-databay

As described in the Ticket the select option of a gap combination for a numeric gap shows the gaps value and a "Out of range" option. Since the value is always "In the range" it is cleaner to display "In the range" instead of the value itself.

https://mantis.ilias.de/view.php?id=40962


dsstrassner:
Hi @Chris-Squirrel, is 'in the range' ok? Or should it be rather 'within the range'?

Chris-Squirrel:
Hi [dsstrassner](https://github.com/dsstrassner) - I would go with 'Within range' (without 'the').

dsstrassner:
Thanks, @Chris-Squirrel! @matheuszych please adept your PR accordingly!

matheuszych:
Hi @dsstrassner. I adopted all changes.

thojou:
These adjustments should solve the problem! @mbecker-databay and @kergomard, could you kindly take a look and possibly merge this for us? Thanks a bunch!

PR #7330, opened 2024-04-09
Component: Avoid redundant event handling on plugin update

mjansenDatabay:
This PR fixes a redundant call of `\ilPlugin::readEventListening` when updating a plugin in `\ilPlugin::update`.

Mantis Issue: https://mantis.ilias.de/view.php?id=41085


klees:
Thanks a lot!

PR #7331, opened 2024-04-09
Chatroom: JS Dependency Winston v3 update

lscharmer:
Update code usages for the Winston Logging Dependency for version 3.


mjansenDatabay (approved):
Thanks a lot @lscharmer

PR #7332, opened 2024-04-09
Expandable Standard Panel and Standard Listing Panel

tfamula:
This PR introduces an extension to the `Standard Panel` and `Standard Listing Panel` to make them expandable, as it is need for https://docu.ilias.de/goto_docu_wiki_wpage_7416_1357.html.
The PR includes the interface changes and also the implementation part of it.

The following things are missing and will be added if interface ~~and implementation are~~ is accepted:

- [x] CSS adjustments for the Bulky Buttons in the Panel header (including the requirement in the FR: "[...] the clickable area should be as large as possible so that the user can also click next to the panel title to open or close it.")
- [x] Adding / fixing unit tests

~~Please also consider, that the Javascript part was tested on ILIAS 9, since Javascript is not working yet on current trunk.~~


Amstutz:
Hi @tfamula 

Note, this is just a review of the public interface and the examples, NOT the full implementation. If the public interface is accepted by the JF, a further review will follow. Further note, that Unit tests would also be needed to complete the Review (including the JS part). Finally, note that we encourage to write the unit tests along with the implementation to find potential design issues early in the process.

Some questions:
- [x] Examples: You added things not strictly having a relation to the panels in the examples (such as pagination). Maybe simplify them, but extend the examples on the panels usage somewhat further (such as the async loading or action part).
- [x] Expandable Usage: Why aren't all panels expandable?
- [x] Interface Type: Why does the interface extend "Component"? 

Change or indicate why not:
- [x] Interface Methods: Remove all except "with Expandable" on the public interface. Clients do not need to know those things. Most of it is internal stuff used for rendering.

Please Change:
- [x] Standard Panel Description: Please adapt the description of the Standard Panel
- [x] Interface action parameter type: Do not pass a string as expand_action type. Use URI and Signal, maybe also null instead.


Panel is a component with a weak semantic and a rather broad set of possible usage scenarios. We expect that the support of the panel will be discontinued at some point in the future and ask, whoever is reading this, to use components with stronger semantics, such as presentation table or entities, whenever possible.

@Amstutz, @thibsy, @klees

tfamula:
Hi UI Coordinators!

Thanks for your initial review!
I see the point about units tests. I would like to have at least an OK for the public interface by you and the JF, before I start working on the tests. Thanks also for the reminder regarding the JS part.

> - [x] Examples: You added things not strictly having a relation to the panels in the examples (such as pagination). Maybe simplify them, but extend the examples on the panels usage somewhat further (such as the async loading or action part).

I will try to make a suitable example for the KS docu regarding the action part. Personally, I would like to keep the View Control in this example to visualize that the View Controls will be hidden as soon as the Panel is collapsed (I have added this point to the description, it was missing).

> - [x] Expandable Usage: Why aren't all panels expandable?

I have simply followed the decision of the UI Clinic: "We recommend extending Standard Listing Panel and Standard Panel with the expandable/collapsable behaviour."

> - [x] Interface Type: Why does the interface extend "Component"?

I don't really have a reason for this. I have seen that interfaces like `HasViewControls` and `JavaScriptBindable` also do it, so I did it too because of consistency and just "to be on the safe side".

> - [x] Interface Methods: Remove all except "with Expandable" on the public interface. Clients do not need to know those things. Most of it is internal stuff used for rendering.

You are right. Done!

> - [x] Standard Panel Description: Please adapt the description of the Standard Panel

Please check if I have understood this correctly. In my eyes, it's like a little duplication because I have written something about the expandable behaviour in the interaction part of the rules.

> - [x] Interface action parameter type: Do not pass a string as expand_action type. Use URI and Signal, maybe also null instead.

Seems reasonable. Done!


@tfamula

matthiaskunkel:
Jour Fixe, 29 APR 2024: We highly appreciate the extension of the Standard Panel and Standard Listing Panel and accept the PR for ILIAS 10 / trunk.

Amstutz:
Hi @tfamula 

I unassigned my self for the moment. Please re-assign me, as soon as the implementation is ready for a final review. Thx!

tfamula:
Hi @Amstutz 
The implementation is now ready for final review.
I made the adjustments for the CSS, added / fixed some unit tests (PHP and JS) and also made changes for the `Renderer`s so that the JS part can handle `URI`s as well as `Signal`s.

Regarding the examples: I tried to extend them by the action part. I made one example with `URI`s and one example with `Signal`s. Because the `URI`s are performed asynchronously, I do not really know how to make this behaviour visual to the user in the first example. But of course, she can check the "Network"-part of her browser to see that the actions are executed. Additionally, I could not use the session here because the KS does not provide a solution for this AFAIK. In the second example, Modals are triggered when expanding / collapsing the Panel at the same time, which is way more noticeable for the user.

@tfamula

thibsy (changes requested):
Hi @tfamula

During the review I will refer to the `ILIAS\UI\Implementation\Component` and `ILIAS\UI\Component` namespaces as `I` and `C`.

- [ ] Union types: we need to establish a code-style rule for union types sooner rather than later. You are already using two different conventions for the spacing around the `|` separator, which feels quite odd. You don't need to act on this, I merely added this for documentation and to point out the need for such a ruling.
- [x] `C\Panel\IsExpandable` I: while looking at the implementation of this, I realised it does not quite behave like our other components. It is currently not possible to disable this feature again, instead it can only be activated by providing a state. I suggest to introduce an additional parameter to enable/disable the feature and an optional parameter for the state, like `...(bool $is_expandable, bool $is_expanded = false, ...)`.

- [ ] Accessibility: the solution you provided does not seem to be accessible. I cannot expand/collapse the panels, or focus the corresponding controls using the keyboard at all. There are several issues inside the HTML, which may cause said behaviour:
	- [ ] the title of the panel is actually inside the bulky buttons you are using as expand/collapse actions. Please separate the title from these actions. I also believe the actions should not consist of buttons, please use a glyph component, or link if necessary.
	- [x] it seems like you are trying to reuse some existing functionality to expand/collapse the panel content. Please implement this yourself inside the JavaScript you provided and get rid of the data attributes of `div.panel-opener` again. `panel-opener` also suggests that this thingy only opens something, please rename these occurrences to match the term "toggle".
	- [x] once you assign the `panel-expandable` CSS class to the top-level element, is it still necessary to add more classes? Please try to get rid of `panel-body-expandable` again by using proper query selectors. Maybe the expand-state could also be represented using data attributes, instead of CSS classes.
	- [x] it looks like you are wrapping the expand/collapse buttons inside `div`s with CSS classes to find your action buttons in JavaScript later on. Please use a `span` element with `data-*` attributes instead to search for the glyphs.
- [x] `I\Panel\Renderer` I: please extract the rendering process of the `C\Panel\IsExpandable` feature into its own method. Same goes for rendering the actions, which seem to have much duplicated code as well. You should probably extract this into a trait too, because it seems that the `I\Panel\Listing\Renderer` reuses most of this rendering process as well. Please also extract the HTML into a new template file and inject the output using a template variable instead.
- [x] `I\Panel\Renderer` II: the panel actually did not fully support the characteristics of a `C\JavaScriptBindable` component until now. However, it becomes important now since we also depend on this internally. Please introduce an id attribute to the top-level element of all panels, and always bind and apply the JavaScript code - not only if we initialise an `C\Panel\IsExpandable` one. Same inside the `I\Panel\Listing\Renderer`.
- [x] `I\Panel\Renderer` III: please get rid of the `{BODY_ID}` template variable again (and associated HTML attributes). We should be able to use query selectors for searching HTML elements in JavaScript.
- [x] `I\Panel\Renderer` IV: please get rid of the jQuery usages. Use `document.getElementById(...).addEventListener()` for registering the action event listeners instead. Also, if the `Event.preventDefault()` is necessary, implement it in `il.UI.panel`, not the renderer. We try to keep JavaScript code inside our renderers to a minimum.
- [x] Conflicts: please rebase this onto the latest trunk in order to resolve the conflicts.

I did not review the JavaScript code of this PR yet, because I expect there to be some changes while restructuring the HTML. I will have a look into this in the next iteration.

Kind regards,
@Amstutz, @klees and @thibsy

tfamula:
Hi @thibsy 
thanks for the feedback! I have tried to work through the points as well as possible.

* Union types: Did I? Can you show me where exactly? Because I do not see that I used two different conventions for the spacing around the | separator. But you are right, that there are different conventions in general, so a ruling would be great!

* `C\Panel\IsExpandable` I: Sounds reasonable. I changed it.

* Accessibility: I can not confirm the problem with the accessibility. Focusing on the (visible) Bulky Button using the keyboards works like expected (because <button> always get focus by default). While the focus is on the button, pressing Return or Space also results in collapsing/expanding the Panel. Maybe just try again after I changed a lot things now, especially removing the usages of Bootstrap. But before that, I personally also did not had any problems.

* Accessibility, pt.1: I used the Bulky Button here to meet the requirement of the FR: "[...] the clickable area should be as large as possible so that the user can also click next to the panel title to open or close it." In my understanding, this means, that the Glyph, the title of the Panel and also the area to the right should be clickable.
Using only a Glyph would result in making only the Glyph clickable. That does not fulfil the requirement and also make the functionality less accessible IMO. Moreover, the actions will be removed from Glyphs, see [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7697#issuecomment-2196467915) and also [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7717).
With using a Link you mean a Bulky Link, right? I checked this and this does also not seem suitable. An URI is expected as third parameter, which is not optional. And using only an URI as action would conflict with [Timon's feedback](https://github.com/ILIAS-eLearning/ILIAS/pull/7332#issuecomment-2074327768) before: "Interface action parameter type: Do not pass a string as expand_action type. Use URI and Signal, maybe also null instead."

* Accessibility, pt.2: Yes, I was using Bootstrap functionality here. But I changed the behaviour now to something similar like it is proposed for the [Presentation Table](https://github.com/ILIAS-eLearning/ILIAS/pull/7796). But we now have to live with the fact that there is no "pretty" toggle animation anymore, which was provided by Bootstrap. I guess there are alternatives which can be made by pure CSS. If someone is interested in this, she can do this in a next step after this PR is merged.

* Accessibility, pt.3: I tried to replace as many CSS classes as possible for the expand-state by using data attributes. I could also remove some JS code which became obsolete.

* Accessibility, pt.4: Hopefully done as you imagined it. But still Buttons are used and not Glyphs for the reasons I mentioned above.

A few insights of my own about accessibility:
To be honest, I am not sure if the current solution is accessible enough regarding telling a non-sighted person that the Buttons do expand/collapse the area under them. Since the Buttons do not work like a "toggle" here, adding an "aria-expanded" attribute seems not suitable in my eyes. Maybe we need something descriptive for the Buttons which explains that it either expands or collapses something. But I do not know any specific solution. And this problem is also existent in other UI componets with such an expand-collapse-behaiour, like the Presentation Table, with the difference that no Buttons are used there.

* `I\Panel\Renderer` I: I created a trait for the code duplications. This does not apply to the actions rendering because it's not duplicate code IMO, it's just similar. I have to distungish between expand/collapse respectively URI/Signal. Or perhaps I have misunderstood something.
I also extracted the HTML heading part into a new template file.

* `I\Panel\Renderer` II: Thanks for clarification. Done! But could you maybe explain this sentence in more detail: "However, it becomes important now since we also depend on this internally."?

* `I\Panel\Renderer` III: This was needed for the Bootstrap functionality and is removed now.

* `I\Panel\Renderer` IV: Done!

* Conflicts: Done!

@tfamula 

~~PS: Tests are failing (Copyright). I will have a look at this later.~~

thibsy (commented):
Hi @tfamula,

Thanks for your feedback and for implementing the requested changes!

We looked over the current implementation again and noticed the feature does not fit our accessibility criteria yet. While investigating the accessibility of this mechanism, we have discovered an article about [accessible accordions](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/), which is very similar to what we are trying to achieve here.

The core issues of the current implementation are the missing `aria-*` labels for screen-readers, as well as a loosing focus after using the feature, which happens due to the usage of two buttons. Both issues are addressed in the example of the article above.

We believe it would be beneficial to discuss the current approach and further improvements in a meeting, before we continue with the review process here.

Let us know when you are available.

Kind regards,
@Amstutz, @klees and @thibsy

tfamula:
@thibsy and I had a meeting in which we discussed the accessibility issues. We came to the conclusion that the solution with two Bulky Buttons is not suitable to fit the accessibility criteria. Instead, a solution with only one button (in combination with Glyphs), which is part of the Panel's template, has been developed now. This solves the problem with the loosing focus. Additionally, the needed aria-* attributes has been added.

tfamula:
Hi @thibsy, is there any news on this one?

thibsy (changes requested):
Hi @tfamula,

Thx for implementing this feature according to the provided accessibility guidelines. Also thx for your patience on this!

- [ ] Unified template(s): Since we now have to couple the header of `Panel\IsExpandable` panels to their body, it does not make sense to put them into separate files anymore. However, refactoring the panels so they can mostly share one template file, seems out of scope for this PR, so this is highly optional. If you decide to refactor this, you'd need to work on the CSS, so they can share most of the classes and only attach panel-specific styles where necessary. This could be achieved using a different top-level class for each panel, from which you query-select downwards to override/make adjustments.
- [ ] Panel segregation: the business logic of the different panels is currently mixed up and contained in one file. This applies for both JS and PHP code. It is rather hard to tell, what code belongs to which type of panel. Since this would mean another large refactoring, this point is optional. If you decide to tackle this, feel free to contact me for any advice.

- [ ] ESLint comments: you are using ESLint (thx!), which is good, but please don't write code that requires to disable the linter. Remove all `eslint-disable-*` comments and amend the code accordingly.
- [ ] JS globals: you are using the `document`, `il` and `$` (jQuery) globals in your code. Please import them as [described in our JavaScript bundling guidelines](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js-bundling.md#using-globals). Also note, there is a `JQueryEventDispatcher` you can use to trigger signals.
- [ ] JS null-pointers: There are many possible null-pointers in your code, mostly because you retrieve the panels using `getElementById()`. Please either use optional chaining (`?.`) and/or `if` statements, to check the returned values.
- [ ] JS unit tests: remove the current unit tests and implement one or more functional test. The current tests don't really add any value.
- [ ] `Panel.performAsync()` and `Panel.performSignal()`: these functions check for null. Why? Please expect a valid argument there.
- [ ] JSDoc comments: please provide type-hint annotations for all arguments and returned values.
- [ ] Conflicts: please rebase this onto latest trunk in order to resolve the conflicts.

Kind regards
@klees and @thibsy

tfamula:
Hello @thibsy,
I can understand the change requests. At the same time, I am a bit frustrated that these requests have not been communicated earlier. Im my eyes, most of them could have been even reported in a former iteration step. I firmly assumed that this PR is as good as merged.
Now, we find ourselves in an unfavourable situation. There is simply not enough time to work through your points until "Coding Completed". On the one hand, I had to wait nearly two months for the recent update here. On the other hand, I am on holiday from tomorrow until end of October. So we will not see the Expandable Panels in ILIAS 10, which is a great pity, especially in view of the fact that this PR already started at the beginning of April.

thibsy:
Hi @tfamula,

We totally understand your frustration, this should not have taken so long to review, or we should have notified you about the delay or issues sooner.

We share your goal to include this feature with ILIAS 10, but we also think that the integrity and quality of our code is an important goal, which we try to compromise as little as possible. Would you commit to iterate on this and implement the requested changes after we merge this PR (and once you are back from vacation)?

@alex40724 maybe you can answer our question on his behalf, since according to the last comment he is already on vacation.

We apologise for the inconveniences.

Kind regards,
@Amstutz, @klees and @thibsy

oliversamoila:
Hello everyone.
I requested @thojou to take a review of this PR for use with the Item Group. To update the views in the repository, this topic should receive further attention. We will provide feedback here regarding any changes, requirements and, if necessary, the next steps.

Many thanks and best regards,
@oliversamoila

PR #7336, opened 2024-04-10
Chatroom: Drop upload table. Remove deleted user from ban table

lscharmer:
Drop `chatroom_uploads` table.
Delete the corresponding entry from then `chatroom_ban` table when a user is deleted (add listener).

### DIFF
--- a/components/ILIAS/Chatroom/classes/CleanUpBans.php
+++ b/components/ILIAS/Chatroom/classes/CleanUpBans.php
@@ -1,0 +1,40 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\Chatroom;
+
+use ilDBConstants;
+use Closure;
+
+class CleanUpBans
+{
+    /**
+     * @param Closure(string, string[]): Closure(array): void
+     */
+    public function __construct(private readonly Closure $prepare)
+    {
+    }
+
+    public function removeEntriesOfDeletedUser(int $id): void
+    {
+        $execute = ($this->prepare)('DELETE FROM chatroom_bans WHERE user_id = ?', [ilDBConstants::T_INTEGER]);
+        $execute([$id]);
+    }
+}
--- a/components/ILIAS/Chatroom/classes/class.ilChatroomAppEventListener.php
+++ b/components/ILIAS/Chatroom/classes/class.ilChatroomAppEventListener.php
@@ -1,0 +1,80 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+use ILIAS\Chatroom\CleanUpBans;
+
+class ilChatroomAppEventListener implements ilAppEventListener
+{
+    private static ?ilChatroomAppEventListener $instance = null;
+
+    private readonly CleanUpBans $clean_up;
+
+    /**
+     * @param array<string, mixed>  $a_parameter
+     */
+    public static function handleEvent(string $component, string $event, array $parameter): void
+    {
+        global $DIC;
+        self::$instance ??= new self($DIC->database());
+        self::$instance->handle($component, $event, $parameter);
+    }
+
+    private function __construct(ilDBInterface $database)
+    {
+        $this->clean_up = new CleanUpBans($this->dbCachePrepare($database));
+    }
+
+    private function handle(string $component, string $event, array $parameter): void
+    {
+        if ($event === 'deleteUser') {
+            $this->clean_up->removeEntriesOfDeletedUser($parameter['usr_id']);
+        }
+    }
+
+    /**
+     * @return Closure(string, array): Closure(array): void
+     */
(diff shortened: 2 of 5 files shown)


components/ILIAS/Chatroom/classes/class.ilChatroomAppEventListener.php
```diff
@@ -0,0 +1,37 @@
...
+
+class ilChatroomAppEventListener implements ilAppEventListener
+{
+    /**
+     * @param array<string,mixed>  $a_parameter
+     */
+    public static function handleEvent(string $component, string $event, array $parameter): void
+    {
+        if ($event === 'deleteUser') {
+            global $DIC;
```
mjansenDatabay: IMO we should avoid database communication in listeners (I know there will be many examples where it is done this way).

mjansenDatabay (changes requested):

components/ILIAS/Chatroom/classes/class.ilChatroomAppEventListener.php
lscharmer: Should I add a `static` method in the `ilChatroom` class instead?! (non static doesn't make sense IMO as this is not specific to one room but all of them).

mjansenDatabay: IMO the best approach is to add a new service class.

lscharmer: Done, please also note that I changed the SQL statement.

components/ILIAS/Chatroom/classes/CleanUpBans.php
```diff
@@ -0,0 +1,37 @@
...
+
+class CleanUpBans
+{
+    public function __construct(private readonly ilDBInterface $db)
+    {
+    }
+
+    public function removeEntriesOfDeletedUsers(): void
+    {
+        $this->db->manipulate(
```
mjansenDatabay: The statement is executed whenever the `deleteUser` event was fired/raised. Have you done any performance benchmarking for cases where dozens or even hundreds of users are deleted at once?
And: IMO a `JOIN` instead of the `IN()` + `SUBQUERY` performs better.

Options:

- Only delete the record matching the user id of the fired/raised event, together with the usage of a cached (statically) prepared statement. This could be done in a `CachedCleanupBans` file, or even better (but this would be a bit overengineering) by a separate repository used by `CleanupBans`. In this case, the SQL operation should be moved to a `BanRepository` and decorated this with a `CachedBanRepository` (which simply caches the statement statically).
-  We should use a (optimized, see my comment above) version of your proposal by ensuring the statement is executed only once per request.

mjansenDatabay (changes requested):

components/ILIAS/Chatroom/classes/CleanUpBans.php
lscharmer: AFAIK I cannot use `join` when using `delete`. Also AFAIK, `in` subqueries are optimized when they don't contain any references to the outer query.

I think that it is 1. easier, 2. More robust (it is always ensured that only non existence users are deleted) and 3. Automatically cleans up old bans of users that are already deleted a long time.

I will test the performance of this query. Alternatively we could simply delay this to the execution end with `register_shutdown_function` so only one query is executed even when 10000 users are deleted (which is **definitively** more performant than using prepared statements for every single one).

mjansenDatabay: > AFAIK I cannot use join when using delete.

That's not true, `JOIN` can be used in a `DELETE` statement.

In general, there are IMO three approaches when deleting records of a table based on data of a second relation:

- `IN` or `=` with sub query
- `EXISTS` with sub query
- `JOIN`

As you already did some benchmarking, please use the fastest approach here.

lscharmer: Hi @mjansenDatabay,
this code path now uses prepared statements with the given user id. So the subquery or join discussion isn't relevant anymore.

It was the fastest approach.

components/ILIAS/Chatroom/classes/class.ilChatroomAppEventListener.php
```diff
@@ -0,0 +1,80 @@
...
+     * @param array<string,mixed>  $a_parameter
+     */
+    public static function handleEvent(string $component, string $event, array $parameter): void
+    {
+        global $DIC;
+        self::$instance ??= new self($DIC->database());
+        self::$instance->handle($component, $event, $parameter);
+    }
+
+    public function handle(string $component, string $event, array $parameter): void
```
mjansenDatabay:
```suggestion
    private function handle(string $component, string $event, array $parameter): void
```

components/ILIAS/Chatroom/classes/class.ilChatroomAppEventListener.php
```diff
@@ -0,0 +1,80 @@
...
+
+    private readonly CleanUpBans $clean_up;
+
+    public function __construct(ilDBInterface $database)
+    {
+        $this->clean_up = new CleanUpBans($this->dbCachePrepare($database));
+    }
+
+    /**
+     * @param array<string,mixed>  $a_parameter
```
mjansenDatabay:
```suggestion
     * @param array<string, mixed> $a_parameter
```

mjansenDatabay (changes requested):

components/ILIAS/Chatroom/classes/class.ilChatroomAppEventListener.php
lscharmer: Hm, I don't think that this method should be private.
`handle` is the entry point into the code, even though it is only used in `handleEvent`.
Do you want this method to be private because it isn't used anywhere else? Or because the main entry point into the event handler should be `handleEvent`? IMO this forces to use the Database from the `$DIC`.

If `handle` should be private then `__construct` should also be private, because a class instance would be useless anyway.

mjansenDatabay: IMO `handleEvent` should be the only `public` method. You are right, `__construct` should be private as well.

components/ILIAS/Chatroom/classes/class.ilChatroomAppEventListener.php
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * @template A
+     * @template B
+     *
+     * @param Closure(...A): B
+     * @return Closure(...A): B
+     */
+    private function cache(Closure $proc)
```
mjansenDatabay: I suggest adding a return type here, `\Closure` should be valid.

mjansenDatabay (changes requested):

components/ILIAS/Chatroom/classes/class.ilChatroomAppEventListener.php
lscharmer: Huh, where did that got lost :o :)

mjansenDatabay (approved):

PR #7338, opened 2024-04-10
Fixed Type Error ContainerContentGUI

eozkeskin:
TypeError thrown with message "Argument 4 passed to ilObjectListGUI::initItem() must be of the type string, null given, called in /var/www/html/Services/Container/Content/class.ilContainerContentGUI.php on line 679"

https://mantis.ilias.de/view.php?id=41098


alex40724:
Fixed similar. Thanks.

PR #7339, opened 2024-04-10
[FEATURE] UI: Introduce dedicated image upload field

thibsy:
Hi folks,

I am happy to propose the public interface for a dedicated image upload field, which encourages users to provide alternate texts when necessary. The related FR can be found here: https://docu.ilias.de/goto_docu_wiki_wpage_5882_1357.html

As [discussed in the UI-Clinic](https://docu.ilias.de/goto_docu_dcl_8186_166_10629.html), we have decided to align this input with the existing file input, so we can reuse most of its functionality and reduce maintenance in the future. We also agreed, that the look-and-feel of the input needs some polishing. Therefore, I also updated the factory description of the file input, to match the new mockups more precisely.

I plan on using the `EntityListing` component to build the file-list internally. Maybe we could introduce some concrete `FileUploadEntity` in the future, to wrap up all of the components required to build this (progress bar, accordion, glyphs, ...). Let me know if this is a good idea.

Similar to the `Rating` input, I added an `ImagePurpose` enum to determine how the alternate text input should be handled. We currently know three different scenarios:

- the consumer knows that uploaded images will be informative, so an alt-text input is required.
- the consumer knows that uploaded images will be decorative, so an alt-text input is obsolete.
- the consumer doesn't know what the image purpose is, so an alt-text may be required. For this case a radio group must be used, letting the user make this choice.

I am looking forward to your review!

Kind regards,
@thibsy

### DIFF
--- a/components/ILIAS/UI/src/Component/Input/Field/Factory.php
+++ b/components/ILIAS/UI/src/Component/Input/Field/Factory.php
@@ -22,7 +22,7 @@ namespace ILIAS\UI\Component\Input\Field;
 
 use ILIAS\UI\Component\Input\Container\Form\FormInput;
 use ILIAS\UI\Component\Input\Field\Node\NodeRetrieval;
-use ILIAS\UI\URLBuilderToken;
+use ILIAS\Data\ImagePurpose;
 
 /**
  * This is what a factory for input fields looks like.
@@ -580,27 +580,46 @@ interface Factory
      * ---
      * description:
      *   purpose: >
-     *     A File Input is used to upload one or multiple files, using the native
-     *     filebrowser of a browser or Drag&Drop.
+     *     The File Field can be used to upload one or multiple files from the
+     *     user device storage.
      *   composition: >
-     *     A File Input is composed as a Dropzone and a list of files. The
-     *     Dropzone contains a Shy Button for file selection.
+     *     The File Field consists of a Dropzone and a Shy Button. Uploaded files
+     *     are listed in an EntityListing above the input, featuring the files
+     *     most important information like title, size and type. Optionally, each
+     *     entry may contain an epandable content area, featuring the provided
+     *     metadata input. Among each entry there are several Glyphs, to expand,
+     *     collapse or remove the corresponding entry.
      *   effect: >
-     *     According to configuration, the input will accept files of certain
-     *     types and sizes. Dragging files from a folder on the comuter to the
-     *     Page in ILIAS will highlight the Dropzone.
-     *     Clicking the Shy Button which starts the native browser file selection.
-     *     Droppping the file onto the Dropzone or selecting a file in native
-     *     browser will directly upload the file and add a info-line beneath
-     *     the dropzone with the title and the size of the file and a Remove
-     *     Glyph once the upload has finished.
-     *     Clicking the Remove Glyph will remove the file-info and calls the
-     *     upload-handler to delete the already uploaded file.
-     *     Invalid files will lead to a error message in the dropzone.
+     *     Clicking the Shy Button opens the browsers file dialog, letting the
+     *     user choose files from their device. Depending on the provided mime-
+     *     types, only certain files will be selectable in this dialog.
+     *     Dragging files from the device into the browser window, where ILIAS
+     *     is open, will highlight the Dropzone where they can be dropped.
+     *     Dropping files will create a new entry inside the PropertyListing.
+     *     Clicking the expand/collapse Glyphs will reveal or hide the file
+     *     entry's metadata input.
+     *     Clicking the removal Glyph will delete the file entry from the list.
+     *     The consumer is responsible to handle file deletions properly, if an
+     *     existing file has been removed this way.
+     *     Adding invalid files (e.g. wrong mime-type, too large) will show an
+     *     according message inside the file entry.
+     *     When the surrounding form is submitted, all files are uploaded first.
+     *     During this process every file entry will display a Progress Bar
+     *     representing the upload status.
+     *   rivals:
+     *     Image Field: >
+     *       if only images should be uploaded with this input, the Image Field
+     *       must be used instead.
      *
      * rules:
      *   usage:
-     *     1: The consuming component MUST handle uploads and deletions of files.
+     *     1: The consumer MUST handle uploads and deletions of files.
+     *     2: File entries should not be expandable, if there is no additional input.
+     *     3: >
+     *       The consumer MUST support chunked uploads if the max file size
+     *       exceeds the server limitation.
+     * accessibility:
+     *   1: The input MUST be operable using the keyboard only.
      *
      * context:
      *   - Upload icons for items in the MainBar (https://docu.ilias.de/goto_docu_wiki_wpage_3993_1357.html)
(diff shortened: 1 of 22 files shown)


thibsy:
image-upload-mockup-desktop

Amstutz:
Hi @thibsy 

Thx a lot for the clear description and differentiation of the three scenarios. We appreciate the concept of a Image Field Input to develop further parts of the concept.

Some Question:

- [x] Small Screens: Will and if so, how will the display on small screens differ from the provided mocks.
- [x] Image Field Input: Would it make sense, to understand the Image Field Upload as a special variation of the File Upload? You devised it as a Field on the same level on our Semantic Tree. Why?

Change or indicate why not:

- [x] Error Display: It probably makes sense to display the error message above the field with file name, as is usually the case in forms.

Thx a lot,
@Amstutz and @klees with kind support of @yvseiler

thibsy:
Hi @Amstutz,

Thx for the feedback! Let me answer your questions/remarks:

- Small screens: The file entries will essentially become more "squashed" but will keep the look and feel as much as possible. One notable layout change is the preview section of the new image input though, which will shift above the metadata inputs and will not remain next to them, at some specified breakpoint.
- Image Field Input: The image field is definitely a special variation of the file input. I have changed the inheritance accordingly, which actually simplifies things :).
- Error display: I agree, we could probably reuse the styling of form errors too. I will not adjust the mockups, but I'll keep this in mind.

Hope this helps!

Kind regards,
@thibsy

matthiaskunkel:
Jour Fixe, 29 APR 2024: We highly appreciate this PR and accept it for ILIAS 10 / trunk.

Amstutz:
Looking forward to the implementation, please re-assign us, as soon as we can review this.

klees:
Hi @thibsy,,

Kind regards!

ckozianka:
Hi @thibsy 

Please respond to @klees comment. 

Kind regards

thibsy:
Hi all,

This PR got stuck a while back because we tried to refactor the file input while/before implementing this new variation of a file input. The refactoring was meant to replace the underlying dependency used for client-side file upload handling. This also meant refactoring the JavaScript code to an ES6 module, which turned out to be a much more complicated task than anticipated, so we have decided to postpone the implementation of the image field and search for additional funding. That's what ultimately led us to this indefinite state.

Since @ckozianka and his organisation are funding the implementation of the new image field input (thank you very much!), we have decided to put the refactoring aside and ignore the sunken cost here, so we can ship the image field input for an early stage of ILIAS 11.

About the technicalities of this PR:

- Metadata inputs: this feature heavily relies on the `Field\HasDynamicInputs` implementation of the `Field\File` input, which turns out to have some issues that affect this feature greatly. I have already fixed some of these issues (contained in this PR as well), but there are still some issues left which are not trivial to fix. See the list of issues here:
	- https://mantis.ilias.de/view.php?id=44148
	- https://mantis.ilias.de/view.php?id=44176
	- https://mantis.ilias.de/view.php?id=44192
	- https://mantis.ilias.de/view.php?id=44193
- File upload library: since the refactoring could not be done, we are stuck with the old npm package `dropzone`. I will request this library in a separate PR and label it for discussion at the upcoming JF on Monday.

In summary: the `Field\Image` input of this PR is functional, if the input only allows to upload one image at a time (which is default). The issues mentioned above only affect inputs where multiple inputs are generated on the client. However, since this thing is functional, we could perhaps merge it before the issues are addressed, to make it available for development of ILIAS 11. Let me know what you think :).

The PR still needs some polish in form of improved example descriptions, translations, unit tests, etc.

We apologise to all involved parties for the delay here.

Kind regards,
@thibsy

components/ILIAS/UI/src/Component/Input/Field/Factory.php
```diff
@@ -610,6 +629,78 @@ public function file(
...
+     * description:
+     *   purpose: >
+     *     The Image Field can be used to upload one or multiple images from the
+     *     user device storage, encouraging them to provide an alternate text for
+     *     screen readers depending on the image purpose.
+     *   composition: >
+     *     The Image Field consists of a Dropzone and a Shy Button. Uploaded files
+     *     are listed in an EntityListing above the input, featuring an image
+     *     preview and the files most important information like title, size and type.
+     *     Optionally, each entry may contain an epandable content area, featuring
```
klees:
```suggestion
     *     Optionally, each entry may contain an expandable content area, featuring
```

klees (changes requested):
Hi @thibsy,

thanks for moving forward with this. The implementation looks mostly fine and we have already discussed and approved the interface on the JF.

* [ ] **Switchable Group**: We wonder about [this change](https://github.com/ILIAS-eLearning/ILIAS/pull/7339/files#diff-b26b31fb1a1c4c98cda29014ff2b56e5a419dd97ea68b2991662323c910041d7R74) in SwitchableGroup::isClientSideValueOk. How come that the client might transmit an array value now? Or is it because withValue is called as such? Besides [the ticket](https://mantis.ilias.de/view.php?id=44169) from @iszmais (Databay AG) regarding null and withValue, this might be another hint that we might want to seperate the paths "value set by client interaction" and "value set by consumer in php" from one another. Or isClientSideValueOk is simply outdated. But the sanity check that it was meant to be ("Does the other side indeed send something we know is possible...") seems to get lost more and more. Eitherway, could be something for the roadmap...
* [ ] **+description "Effect"**: You mention that dragging files will highlight the dropzone where they can be dropped. But I do not see any visual effect while dragging the file onto the Dropzone (Select Files). Is the highlight missing or is the description wrong?
* [ ] **example 5**: If I click on “Save” without having selected the mandatory value “-image_purpose_user_defined-”, no error message is displayed. Is that intentional? I would expect to receive a message here informing me that I have to fill in mandatory fields.
* [ ] **example 5 2**: At the moment I see only language variables like -image_purpose_user_defined- or -image_purpose_informative- and -image_purpose_decorative-. Are there already defined labels in the language file or do we still need suggestions from the Language Group? It is quite possible that these are not displayed correctly on my test installation - refreshing the language is unfortunately not possible on ILIAS 11.0alpha at the moment (has no effect).

* [ ] **example 1**: We should be careful when writing "expected outputs". The ["dashed lines"](https://github.com/ILIAS-eLearning/ILIAS/pull/7339/files#diff-07e1bb0fcc1d04204e262ccacaa33be59ea0191b368e872f203e66229980cb4fR30) are some design decision that might change. Maybe something like "next to discernible box" or something like that will be more solid in the long run.
* [ ] **example 5 3**: I can’t click on the «remove» glyph to delete a file. For testing reasons I would prefer to have all interactions available as in an actual implementation in a component. If this is intentional, please explain why we shouldn’t test this behaviour in this example.
* [ ] **mime type**: No example shows which mime type is supported. If this can be restricted (e.g. only jpg or only png), it would be helpful for this to be displayed. At the moment it is common to display it in the byline (where size and number of files are already listed).

* [ ] **example 2**: Please add descriptions and expected outputs to the examples where they are missing.
* [ ] **description "Composition"**: Please add an «x» in «… an epandable content area …» (expandable).
* [ ] **example 1-4**: I can't get any further in testing because the upload of common image file types such as jpg, png, gif is not possible. The error “Files of type ‘image/png’ are not allowed” is displayed after clicking on «Save». Please check which adjustments need to be made so that images can be uploaded.

Best wishes!
@klees, @Amstutz and @yvseiler

thibsy:
Hi all,

* [ ] **Switchable Group**: this is actually a bugfix, we should probably discuss it here: #9024. I will rebase here once the bugfix PRs are merged.
* [x] **+description "Effect"**: this behaviour actually doesn't exist. I phrased this paragraph differently.
* [x] **example 5**: this is because none of the example implements data processing. I have added one where this will be visible.
* [x] **example 5 2**: they were not yet translated, I have added some. The language group can gladly chip in here though.

* [x] **example 1**: agree. done in #9029; I will rebase once the bugfix PRs have been merged.
* [x] **example 5 3**: I fixed this issue, good catch :)!
* [x] **mime type**: I reported and fixed the issue (https://mantis.ilias.de/view.php?id=44978). This behaviour most likely exists in all supported version.

* [x] **example 2**: done.
* [x] **description "Composition"**: for the same reason as your *example 1* remark, I would rather not describe the close Glyph. This could e.g. easily be a minus someday.
* [x] **example 1-4**: I fixed the issue, you can have another look now.

Kind regards,
@thibsy

klees (approved):
Hi @thibsy,

LGTM, I will handover to @yvseiler for frontend stuff.

Thanks a lot!

yvseiler:
The issue mentioned above with missing language variables is a problem with the “Switchable Group” in general. The Mantis issue https://mantis.ilias.de/view.php?id=45569 was created for this purpose.

There is no further need for adjustments on the part of UI/UX or HTML validation. 

I am therefore handing the issue over to you @amstutz & @klees for the remaining bug fixes.

PR #7343, opened 2024-04-11
IASS: fix displaying specific tabs if orgu positions are set. (#40684)

lukastocker:
https://mantis.ilias.de/view.php?id=40684

According to the test case  (linked in the mantis ticket) following should now be possible:

A. Org Unit position permissions "Set learning progress of other users" and "View learning progress of other users" are set:
- display all employees
- display employee's data and enable the possibility to change those data
- display learning progress tab with the data of your employees
- disable the possibility to change learning progress data/settings
- do not display other users and their data. You also can't change their data.

Change A to B. Org Unit position permission "View learning progress of other users" is disabled:
- enable the possibility to assess participants (only your employees - not other users)
- do not display the learning progress tab
- only display your own assessments

Change A to C. Org Unit position permission "Set learning progress of other users" is disabled:
- display data of your employees
- disable the possibility to change data and assessments
- display learning progress data but disable the possibility to change the learning progress data/settings

### DIFF
--- a/Modules/IndividualAssessment/classes/AccessControl/class.ilIndividualAssessmentAccessHandler.php
+++ b/Modules/IndividualAssessment/classes/AccessControl/class.ilIndividualAssessmentAccessHandler.php
@@ -72,9 +72,12 @@ class ilIndividualAssessmentAccessHandler implements IndividualAssessmentAccessH
             );
         }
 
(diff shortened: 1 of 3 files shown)


klees:
Picked to trunk.

PR #7344, opened 2024-04-11
MOB: remove methods for Flash

klees:
Hi @alex40724,

stumble upon this while working on #7019, I think this is long outdated.

Kind regards!


alex40724:
Thanks. I removed most code of this class. There are some references left that need testing though.

PR #7346, opened 2024-04-11
[FIX] #39532,#39260 UI: distinguish example post urls.

thibsy:
Hi folks,

This PR fixes the following mantis issues:

- https://mantis.ilias.de/view.php?id=39532
- https://mantis.ilias.de/view.php?id=39260

The issues are caused by the same problem: the examples don't use proper post URLs and merely check if the current request method is post. Therefore `withRequest()` is called for every example, even though only one of them was actually submitted. This leads to the problem because the `with_additional_input()` example of both dropzones expects another input to be present.

This fix is merely a bandaid, because the root problem is that examples should have access to information like post-URLs via arguments. If you like, I'll add this to the roadmap.

Kind regards,
@thibsy

### DIFF
--- a/src/UI/examples/Dropzone/File/Standard/base.php
+++ b/src/UI/examples/Dropzone/File/Standard/base.php
@@ -11,12 +11,16 @@ function base()
     $factory = $DIC->ui()->factory();
     $renderer = $DIC->ui()->renderer();
     $request = $DIC->http()->request();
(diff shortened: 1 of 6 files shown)


klees:
Picked to trunk.

thibsy:
I added this to the roadmap with d8525b098317e367b61cbf6490060b8525bfa69c

PR #7347, opened 2024-04-11
[FIX] #38079 UI: update Field\Image description

thibsy:
Hi folks,

This PR addresses the following mantis issue: https://mantis.ilias.de/view.php?id=38079

It merely updates the factory description of the `Field\File` input, so it states that `one or multiple files` can be uploaded instead of just `a single file`.

Kind regards,
@thibsy


Amstutz:
thx a lot!

PR #7352, opened 2024-04-11
[FIX] #39474 UI: remove duplicate ids from message boxes

thibsy:
Hi folks,

This PR addresses the following mantis issue: https://mantis.ilias.de/view.php?id=39474

The problem was not exactly related to the `Dropzone\Wrapper` component like the ticket suggests, but rather the `Component\MessageBox` components which were wrapped by them. I noticed there is an `id="il_message_focus"` in the `tpl.messagebox.html` template, which leads to duplicate ids when rendering more than one of them. To fix this issue I simply removed this attribute from the template. I couldn't find any usages of this ID, so I think we're fine.

Since I had to adjust the rendering unit-tests, I also decoupled the `Launcher\Inline` rendering test while at it.

Kind regards,
@thibsy


Amstutz:
thx a lot!

PR #7356, opened 2024-04-12
OrgUnit: Support multibyte characters when shortening path string

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=38382

If approved, this MUST be picked to `release_9` and `trunk` as well.


klees:
Hi @mjansenDatabay,

thanks for identifying the cause of the problem and proposing the fix. I'm not very fond of introducing a dependency on an unmaintained component (`ILIAS/Utilities`...) and increase the burden on this component that is somehow troubled anyway.

I will set the ticket to "Funding required" so we can move this to the `Refinery`.

Kind regards!

PR #7357, opened 2024-04-12
MetaData/Copyright: Handle null values when reading from database

mjansenDatabay:
This PR fixes a missing `null` handling when reading copyright metadata from the database. `title` and `description` might be `null`, but `\ilMDCopyrightSelectionEntry::setTitle` and `\ilMDCopyrightSelectionEntry::setDescription` only accept strings.

Mantis Issue: https://mantis.ilias.de/view.php?id=41120

If necessary, please pick this to `release_9` and `trunk` as well.

### DIFF
--- a/Services/MetaData/classes/class.ilMDCopyrightSelectionEntry.php
+++ b/Services/MetaData/classes/class.ilMDCopyrightSelectionEntry.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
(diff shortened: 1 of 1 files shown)


schmitz-ilias:
Thanks for the fix @mjansenDatabay! Merging to 9 and trunk is not necessary, reading out copyrights was refactored there and should be a bit more robust already.

PR #7358, opened 2024-04-12
News: Fix accessing an undefined array key

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40836


alex40724:
Thanks for the PR!

PR #7361, opened 2024-04-12
set ids to ui tables

iszmais:
Sets the current class to the tables id for table setting persistence in scopes:
- Services/SAML
- Services/Mail
- Services/Contact
- Modules/Chatroom
- Modules/Forum

### DIFF
--- a/components/ILIAS/Chatroom/classes/Bans/BannedUsersTable.php
+++ b/components/ILIAS/Chatroom/classes/Bans/BannedUsersTable.php
@@ -39,6 +39,7 @@ class BannedUsersTable implements UI\Component\Table\DataRetrieval
     private ?array $records = null;
 
     public function __construct(
+        private readonly int $room_id,
         private readonly array $banned_users,
         private readonly ilCtrl $ctrl,
         private readonly ilLanguage $lng,
@@ -57,6 +58,7 @@ class BannedUsersTable implements UI\Component\Table\DataRetrieval
         return $this->ui_factory
             ->table()
             ->data($this->lng->txt('ban_table_title'), $columns, $this)
+            ->withId(self::class . '_' . $this->room_id)
             ->withActions($actions)
             ->withRequest($this->request);
     }
--- a/components/ILIAS/Contact/classes/MailingLists/MailingListsTable.php
+++ b/components/ILIAS/Contact/classes/MailingLists/MailingListsTable.php
@@ -71,6 +71,7 @@ class MailingListsTable implements UI\Component\Table\DataRetrieval
                 $columns,
                 $this
             )
+            ->withId(self::class)
             ->withActions($actions)
             ->withRequest($this->request);
     }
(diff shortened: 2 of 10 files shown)


iszmais:
@mjansenDatabay after some consideration i came to the conclusion that the class name should provide a fitting ID for the following reasons:
- A class should always only provide 1 table. If a class can generate 2 tables it should be splitted into 2 classes.
- A table provided from a class should have the same settings in every view. Even when i acces the data in another view its very likely that i want i to be presented as i defined within another view.
- Even if a table provides multiple modifications of a table (e.g. given to view rights or data votality) the sorting and filter should try to resemble the same view every time. Its is not in the response of the table setting to handle volatile data, but in the table itsself.

Therefore i assume the name of the table to be a good identifier here.
Pls feel free to give me feedback on that.

Greetings,
@iszmais

iszmais:
(The current failed test are result of another commit and already documented [here](https://mantis.ilias.de/view.php?id=41125))

iszmais:
@mjansenDatabay I adapted the ID to be specific. The following are excluded (due to missing context relations):
- BannedUserTable
- MailMemberSearchTable
- ilSamlIdptable
- MailAttachmentTable (differenced by mode, but no unique id)

components/ILIAS/Chatroom/classes/Bans/BannedUsersTable.php
```diff
@@ -57,6 +57,7 @@ public function getComponent(): UI\Component\Table\Data
         return $this->ui_factory
             ->table()
             ->data($this->lng->txt('ban_table_title'), $columns, $this)
+            ->withId(self::class)
```
mjansenDatabay: Interesting... The missing ID (even in ILIAS <= 9.x) can be considered as a bug, because the table is used every chat repository object.

Please add the `ref_id` of the chat room node here.

components/ILIAS/Contact/classes/MailingLists/MailingListsTable.php
```diff
@@ -71,6 +71,7 @@ public function getComponent(): UI\Component\Table\Data
                 $columns,
                 $this
             )
+            ->withId(self::class . '_' . $this->mailing_lists->getCurrentMailingList()->getId())
```
mjansenDatabay: IMO this table does not need a suffix, because for every user there is only one list of the mailing lists.

See: `ilMailingListsTableGUI` in ILIAS 9.x

components/ILIAS/Contact/classes/MemberSearch/MailMemberSearchTable.php
```diff
@@ -59,6 +59,7 @@ public function getComponent(): UI\Component\Table\Data
                 $columns,
                 $this
             )
+            ->withId(self::class)
```
mjansenDatabay: We need a suffix here, see `ilMailMemberSearchTableGUI` in ILIAS 9.x.

mjansenDatabay (changes requested):
Thx @iszmais ,

there are still a few "ID suffix" issues.

Best regards,
Michael

mjansenDatabay (approved):

PR #7363, opened 2024-04-15
Add user defined fields to Shibboleth attribute mapping

iljalukin:
This PR adds user defined fields to Shibboleth attribute mapping

### DIFF
--- a/Services/AuthShibboleth/classes/User/class.shibUser.php
+++ b/Services/AuthShibboleth/classes/User/class.shibUser.php
@@ -93,6 +93,7 @@ class shibUser extends ilObjUser
             $this->setLanguage($this->shibServerData->getLanguage());
         }
         $this->setDescription($this->getEmail());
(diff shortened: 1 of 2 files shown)


chfsx:
Hello @iljalukin 

thanks for the PR, as already mentioned in the mantis-issue, this change must go through the feature process. as a pure bugfix, in my opinion, this is too big a change, for example in the GUI. it also requires other implementations for release_8, release_9 and trunk.  because it is created as a feature in the feature wiki, for example, the views are also defined, the feature is documented and there must be test cases for it. all this would be missing in a "bugfix" with the justification that other authentication mechanisms can do it.

https://mantis.ilias.de/view.php?id=41138

I'm closing the PR for the moment, as mentioned, please contact us so that this can be submitted as a feature-request for ILIAS 10 or, ideally, you can already create the feature request with the corresponding mockups, etc.

PR #7364, opened 2024-04-15
fix error if stale entries in objective table exist

ingoj:


ingoj:
Sometimes old entries may exist in the question assignment table for objective driven courses. This leads to errors stating that an aray key does not exist. To avoid that error, this PR checks the existence of the entry before adding points to the current result.
If stale entries re found, the id of the question is logged to the ILIAS log, so that the database can be cleaned manually.
I could not (yet) figure out, how these entries appeared in our installation.

smeyer-ilias:
Closed in favour of #7487

PR #7365, opened 2024-04-15
Utils/FileUtils: Reset umask after creating directories

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41139


mjansenDatabay:
Hi @chfsx , do you have any objections here? As long as this legacy code is used (which of course the case until all components use the IRSS), this leads to "wrong" permissions.

chfsx:
Sorry for the delay! I ported the changes to release_8, release_9 and trunk. I close this PR since release_7 is no longer supported.

PR #7367, opened 2024-04-15
UI: adjust rendering of header in property forms. (#41142)

lukastocker:
https://mantis.ilias.de/view.php?id=41142

As changes before this commit resulted to missing header content like settings (e.g. learning progress) the rendering of the header in property forms had to be adjusted.

If no required fields are set the header part for required text output won't get rendered (before it rendered an empty <div>).

If at least one required field is set in a formular a required text will get rendered in the header of the formular.

### DIFF
--- a/Services/Form/classes/class.ilPropertyFormGUI.php
+++ b/Services/Form/classes/class.ilPropertyFormGUI.php
@@ -536,13 +536,12 @@ class ilPropertyFormGUI extends ilFormGUI
             }
 
             // required top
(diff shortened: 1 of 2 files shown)


Amstutz:
thx a lot.

Amstutz:
version for trunk is under creation atm.

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41142#c103785

PR #7382, opened 2024-04-18
PRG: 41164, CompletionBy is empty for deleted Courses

nhaagen:
https://mantis.ilias.de/view.php?id=41164


klees:
Picked to trunk.

PR #7383, opened 2024-04-18
fix errors if password field is null

ingoj:
If users without global Admin, but full access to UserAdministration try to save datafields of accounts with admin-role, errors appear because the password field is null (instead of ""). This Commit fixes this issue.  Since the checks in the function  "check input" all test if the password is "",  I chose to set a null password field to "" instead of adding a null-check to each if clause.

This fix may be usable for release_9 and trunk, too


kergomard:
Thank you very much for the PR @ingoj ! I will still close this and fix it differently. The error you see is because the password field was disabled. I don't think this should happen anywhere else and I don't think we should change the signature for this. I simple changed the implementation of the form with https://github.com/ILIAS-eLearning/ILIAS/commit/9181948a1583795245ea088a960f4267d290bc1e .

Thanks again and best,
@kergomard

PR #7384, opened 2024-04-18
PRG: 41159, change langvar 'cancel' in acknowledge courses

nhaagen:
https://mantis.ilias.de/view.php?id=41159


klees:

PR #7386, opened 2024-04-18
Database: Revert behaviour of ilDBPdo::executeMultiple() to documented functionality

tbongers-cat:
The behaviour of `ilDBPdo::executeMultiple()` has changed during the update to PHP8 in https://github.com/ILIAS-eLearning/ILIAS/commit/26f7a22ad87a24539abc4a24506e9f3094584c00 . In the current version, you can use **multiple** statements, but only **one** data entry is executed per statement. Previously, **all** data entries were executed with **one** statement, which seems more useful and is also the way the method is documented in https://github.com/ILIAS-eLearning/ILIAS/tree/release_8/Services/Database#multiple-similar-data-manipulation.

This PR reverts the behaviour to the one used in ILIAS 7 and adjusts the interface accordingly. The method itself doesn't seem to be used in any core components, but the current state might probably break things in plugins (as it did in our case).

### DIFF
--- a/Services/Database/classes/PDO/class.ilDBPdo.php
+++ b/Services/Database/classes/PDO/class.ilDBPdo.php
@@ -1426,10 +1426,10 @@ abstract class ilDBPdo implements ilDBInterface, ilDBPdoInterface
         return $this->manager->dropConstraint($table_name, "PRIMARY", true);
     }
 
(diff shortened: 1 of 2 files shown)


chfsx (approved):
Thanks a lot! it seems as if this method is not used in core, but we leave for other usages. I cherry-pick his up to trunk

PR #7390, opened 2024-04-18
Release 9: Fix missing type cast

mBeym:
Fixes Mantis Ticket: https://mantis.ilias.de/view.php?id=40874


jeph864:
@mBeym , 
thanks for the PR but the fix still fails because the method can't  add group participants. This can be fixed by casting the same variable($fields[3]) in the line following the one you just fixed. is it possible add this change too?

mBeym:
replaced it with a variable used everywhere.

smeyer-ilias:
Thanks to @jeph864  for reviewing and @mBeym fixing this

PR #7392, opened 2024-04-18
Trunk: Fix missing type cast

mBeym:
Fixes Mantis Ticket: https://mantis.ilias.de/view.php?id=40874


smeyer-ilias:
Thanks @mBeym for fixing this.

PR #7393, opened 2024-04-18
TA: 41167, show number of requested hints in detailed results

nhaagen:
https://mantis.ilias.de/view.php?id=41167

### DIFF
--- a/Modules/Test/classes/Results/class.ilQuestionResult.php
+++ b/Modules/Test/classes/Results/class.ilQuestionResult.php
@@ -39,6 +39,7 @@ class ilQuestionResult
         protected string $feedback,
         protected bool $workedthrough,
         protected bool $answered,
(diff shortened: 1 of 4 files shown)


dsstrassner:
Thank you so much for your swift response!

dsstrassner:
hi @thojou and @mbecker-databay please review and merge. From my point of view, this looks good in the GUI.

thojou (approved):
I considered adding the new label and its corresponding value directly into the `$important_fields` array instead of creating a new variable. However, if we don't need to pass the new label into the `setImportantFields` method later on, this change seems suitable.

@mbecker-databay, could you also take a look and proceed with merging?

thojou:
Hey @nhaagen ,

sorry for the slow response. I have know taken this PR and cherry-picked the changes into the trunk.

Thank you.

PR #7396, opened 2024-04-19
[FIX] casting export value

chfsx:
Fix for https://mantis.ilias.de/view.php?id=41183

needs cherry-pick to release_9 and trunk as well


Amstutz:
thx a lot!

PR #7397, opened 2024-04-19
[FIX] 0041147: missing dropzones for files in list-view

chfsx:
@kergomard I don't know exactly what the $this->requested_cmd is, but in any case the dropzone must be rendered at "", "view" and "render" :-)

fixes https://mantis.ilias.de/view.php?id=41147


kergomard:
Thank you for the PR @chfsx !
This is simply the command. The check before rendering the dropzones was introduced as a bug-fix for https://mantis.ilias.de/view.php?id=37125 , but it seems to have been to strict. Reading through the issue, I think the proposed slackening of the check in this PR is plausible. Merged and picked to release_9 and trunk.

Thanks again and best,
@kergomard

PR #7398, opened 2024-04-19
UI: add tabindex to context form template. (#32111)

lukastocker:
And adjust test cases accordingly.

https://mantis.ilias.de/view.php?id=32111

This PR provides the possibility to tab to a label (like the first label in a SwitchableGroupInput) through an set tabindex. As a result screenreaders can detect the label and read them out loud to users before jumping to linked input fields like "Text Input", "Radio Input" etc. and their labels.

The changes apply to multiple UI Framework Input Fields.


Amstutz:
Imo we should use "legend" as recommended by the auditor. This should be the semantically superior solution. I will close this. However, feel free to reopen, if you believe to have good reasons to chose tabindex over legend.

PR #7402, opened 2024-04-19
T&A Bugfix: Missing lng variable TestStatisticsExport

fhelfer:
exchange missing lng var with existing


dsstrassner:
There is no Mantis Issue, but it would be pleasant if @mbecker-databay could merge this and pick to Trunk.

PR #7403, opened 2024-04-20
xapi auth token encoding bugfix

cahidaydin:
Basic auth token has to be encoded and start like 'Basic%20', not like 'Basic+' rawurlencode function encodes space as '%20' but urlencode encodes space char as '+'. With +, authentication to xapiproxy or LRS fails.


klees:
Hi @Uwe-Kohnle,

Kind regards!

PR #7404, opened 2024-04-22
UI: render margin in presentation table important field block only if…

lukastocker:
… important fields are available. (#41126)

https://mantis.ilias.de/view.php?id=41126


Amstutz:
thx a lot!

PR #7407, opened 2024-04-22
T&A 41186: Fix missing solutionhint points after test import

thojou:
This pull request addresses the problem mentioned in the mantis ticket https://mantis.ilias.de/view.php?id=41186. The issue occurred because the XML parsing ignored the solutionhint points.

### DIFF
--- a/Services/QTI/classes/class.ilQTIParser.php
+++ b/Services/QTI/classes/class.ilQTIParser.php
@@ -534,6 +534,9 @@ class ilQTIParser extends ilSaxParser
             case "resprocessing":
                 $this->resprocessingBeginTag($a_attribs);
                 break;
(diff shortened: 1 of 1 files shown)


dsstrassner:
hi @thojou & @mbecker-databay thanks for the quick fix, I will retest this in a few minutes on test9. Is this also a valid solution for ILIAS 8? I have just realized, that ILIAS 8 is also affected.

thojou:
@dsstrassner, Certainly, the changes can also be cherry-picked into release_8. @mbecker-databay, could you manage the process of cherry-picking for release_8?

dsstrassner:
Works for 9. Now ILIAS 8 @mbecker-databay ? 😀

PR #7411, opened 2024-04-22
[FIX] add public and node_module to ignored files

thibsy:
Hi @klees, this should not need much explanation. Kind regards!


klees:
Thanks!

PR #7415, opened 2024-04-23
fixed 37158: A11Y/LoginPage: The A11Y info page cannot be requested o…

alex40724:
…n the login page

See
https://mantis.ilias.de/view.php?id=37158

### DIFF
--- a/Services/Init/classes/class.ilInitialisation.php
+++ b/Services/Init/classes/class.ilInitialisation.php
@@ -1721,7 +1721,7 @@ class ilInitialisation
         }
 
         if (
-            (strtolower($requestCmdClass ?? "") === "ilaccessibilitycontrolconceptgui")
+            (strtolower($requestCmdClass ?? "") === strtolower(ilAccessibilityControlConceptGUI::class))
         ) {
             ilLoggerFactory::getLogger('auth')->debug('Blocked authentication for cmdClass: ' . $requestCmdClass);
             return true;
--- a/Services/Init/classes/class.ilInitialisation.php
+++ b/Services/Init/classes/class.ilInitialisation.php
@@ -1720,6 +1720,13 @@ class ilInitialisation
             return true;
         }
 
+        if (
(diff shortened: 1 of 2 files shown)


Services/Init/classes/class.ilStartUpGUI.php
```diff
@@ -155,6 +154,9 @@ public function executeCommand()
                 require_once("Services/Init/classes/class.ilPasswordAssistanceGUI.php");
                 return $this->ctrl->forwardCommand(new ilPasswordAssistanceGUI());
 
+            case "ilaccessibilitycontrolconceptgui":
```
mjansenDatabay:
```suggestion
            case strtolower(ilAccessibilityControlConceptGUI::class):
```

Services/Init/classes/class.ilInitialisation.php
```diff
@@ -1720,6 +1720,13 @@ protected static function blockedAuthentication(string $a_current_script): bool
             return true;
         }
 
+        if (
+            (strtolower($requestCmdClass ?? "") === "ilaccessibilitycontrolconceptgui")
```
mjansenDatabay:
```suggestion
            (strtolower($requestCmdClass ?? "") === strtolower(ilAccessibilityControlConceptGUI::class))
```

mjansenDatabay (changes requested):
Hi @alex40724 ,

thx for the PR.

I suggest to use class-strings to avoid problems when the classes are renamed in future. If classes are renamed (e.g. in PhpStorm) such class-strings are automatically renamed as well. This is maybe more "robust" than using "Search & Replace".

Best regards,
Michael

alex40724:
@mjansenDatabay Yes, of course. I added your changes.

mjansenDatabay (approved):

mjansenDatabay:
@pascalseeland I'll merge and pick this to all relevant branches.

alex40724:
@mjansenDatabay Thanks!

PR #7417, opened 2024-04-23
NEWS: remove inline margin from news template breadcrump section. (#4…

lukastocker:
…1037)

https://mantis.ilias.de/view.php?id=41037

I just removed the inline style in the news template as it was set to -15px and we do not want to use inline styling. There was no need to adjust the breadcrumb SCSS class.
As a result the breadcrumb is not glued directly to the panels border anymore - it's now aligned with the e.g media cast item as can be seen below:

breadcrumbs_News_mediacast_after

### DIFF
--- a/Services/News/templates/default/tpl.show_news.html
+++ b/Services/News/templates/default/tpl.show_news.html
@@ -1,7 +1,7 @@
 <!-- BEGIN item -->
 <div class="{ITEM_ROW_CSS}">
 <!-- BEGIN context -->
-<div class="small" style="margin-left: -15px;">
+<div class="small">
 {CONTEXT_LOCATOR}
 </div>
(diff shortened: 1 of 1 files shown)


lukastocker:
Hi @Amstutz ,

FYI: the PHP Unit tests seem to be broken since https://github.com/ILIAS-eLearning/ILIAS/commit/7e77741e0abf0f9a9cfaa882ea5bddb73b3f7d88. 
I tested my changes locally without that commit and all is running well. 
As the failing tests are not part of this ticket I did not include a fix to this PR.

Amstutz:
News is not my component, however, since inline style clearly violates rather ancient CSS guidelines, I'll take the liberty to merge this. Failing tests seem unrelated.

PR #7422, opened 2024-04-24
UI: adjust rendering of main action button and dropdown button. (#41033)

lukastocker:
https://mantis.ilias.de/view.php?id=41033

This includes following changes:

- remove 'ilForumInlineCommandContainer > div' with margin-left value from legacy SCSS file _component_forum.scss as it's not needed anymore
- add 'il-item-actions' class for alignment purposes to legacy SCSS file _component_forum.scss
- adjust the 'forum_threads_view' HTML template
- adjust the rendering of the main action button and the dropdown button in the 'ilObjForumGUI' file.

... instead of just changing the margin value in the specific SCSS file. With these changes the template will - at least partly - be build like an UI Framework standard item template.


mjansenDatabay:
Thx @lukastocker

PR #7423, opened 2024-04-25
SCORM: Fix argument type for `number_format` in certificate values / Move SCORM related `Certificate` files to `SCORM`

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=39850

The actual fix can be found here: https://github.com/ILIAS-eLearning/ILIAS/pull/7423/files#diff-d16df26049e855853f421fe21a33e5f92e79da3c91fd5499a6b11cb52950d11bR151

---

With this PR I would also like to move consumer specific classes from `Sevices/Certificate` to the respective component (here: Scorm), like it has been already done for `CmiXapi` and `LITConsumer`.

All related unit tests were moved as well.

There are still parts of tighter coupling to some object type strings in `Services/Certificate` (e.g. to register a consumer), but this will hopefully be removed in future.

---

If approved, please pick this to `release_9` and `trunk` as well.

### DIFF
--- a/Modules/ScormAicc/classes/Certificate/class.ilCertificateScormPdfFilename.php
+++ b/Modules/ScormAicc/classes/Certificate/class.ilCertificateScormPdfFilename.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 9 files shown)


Uwe-Kohnle (approved):
Thanks

Uwe-Kohnle:
conflicts resolved and pushed into 8, 9, trunk

PR #7425, opened 2024-04-25
PRG: 41216, do not throw if ref_id is not found

nhaagen:
https://mantis.ilias.de/view.php?id=41216
PRG throws, if there is no ref_id for an ob_id; this is not wanted in the context of looking up a title / deletion status for user table.


klees:
Picked to trunk.

PR #7141, opened 2024-03-06
DEP: create new package.json and composer. json with approved packages.

lukastocker:
The list about the approved dependency packages can be found here: [JourFixe-2024-01-22](https://docu.ilias.de/ilias.php?baseClass=ilwikihandlergui&cmdNode=16n:r6&cmdClass=ilobjwikigui&cmd=viewPage&ref_id=1357&page=JourFixe-2024-01-22#ilPageTocA2513)

As mentioned in [PR 6653](https://github.com/ILIAS-eLearning/ILIAS/pull/6653#issuecomment-1964466578) I have not included the simplesamlphp package as it's currently not working with the Symfony Console package version 7. 

As you can see a commit with a new phpunit.xml file was also created, as it would not have worked otherwise. Please also see the change log of PHPUnit 10+, as many functions are deprecated: 
- [Change log](https://github.com/sebastianbergmann/phpunit/blob/10.5/ChangeLog-10.5.md) 
- [Deprecrations](https://github.com/sebastianbergmann/phpunit/blob/10.5/DEPRECATIONS.md)

PR #7427, opened 2024-04-25
Composer: add phpunit/phpunit as dependency

klees:
This PR adds `phpunit/phpunit` as composer dependency.

Usage:
* We use it to run unit tests.
* ILIAS/Component uses its mocking framework to analyze component definitions during build phase.

Wrapped By:
* Not applicable, provides the binary and classes for test management.

Reasoning:
* Its the standard PHP testing framework.

Maintenance:
* Won't go away any time soon.

Links:
* Documentation: https://phpunit.de


klees:
Hi JF,

as [@fhelfer points out correctly](https://mantis.ilias.de/view.php?id=41248) we now need this dependency for the build-phase, hence we should add it to `require` (instead of `require-dev` only). If we do not want to do this we could instead pull in the mocking framework of PHPUnit, which to me would be fine as well. If approved, I would kindly ask @lukastocker to replace the previous PHPUnit commit in [his PR](https://github.com/ILIAS-eLearning/ILIAS/pull/7141) by this one here.

Kind regards!

klees:
This felt harmless, but an update to PHPUnit 10 basically destroys our current approach to the bundling of testcases in testsuites. Currently we are using PHPUnits `TestSuite` class in various places, most notably in our global `ILIASSuite`, to bundle our cases. That global suite (and some local suites) are then added to a phpunit.xml to make them available to PHPUnit.

PHPUnit, though, seems to think of the `TestSuite` class as an internal class that is used to process the [organisation of tests](https://docs.phpunit.de/en/9.6/organizing-tests.html) via directories or XML. When using PHPUnit 10 it complains that we directly add a `TestSuite` instance via the XML:

```
PHPUnit 10.5.20 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.15
Configuration: /home/rklees/Code/projekte/trunk/volumes/ilias/scripts/PHPUnit/phpunit.xml
Random Seed:   1714044288

There was 1 PHPUnit test runner warning:

1) Class ILIASSuite declared in /home/rklees/Code/projekte/trunk/volumes/ilias/scripts/PHPUnit/ILIASSuite.php does not extend PHPUnit\Framework\TestCase
```

Maybe I misinterpret all this, but currently it seems as if we need a new approach to organize our testcases.

Kind regards!

matthiaskunkel:
I assume the label 'bugfix' is wrong in this context. It has to be 'dependencies'. I change it accordingly.

klees:
It's both, there is a [ticket](https://mantis.ilias.de/view.php?id=41248) =)

matthiaskunkel:
Okay. Didn't know that... But for the agenda, dependencies would fit well. As long as it is labeled 'Jour Fixe' ;-)

matthiaskunkel:
Jour Fixe, 29 APR 2024: We accept phpunit/phpunit as dependency for ILIAS 10.

klees:
This has been done in https://github.com/ILIAS-eLearning/ILIAS/commit/5056cedad3e7da53169bc48e9f299df272fcdc93

PR #7432, opened 2024-04-25
Repository: redirect swallows parameters needed for tree to load

nhaagen:
https://mantis.ilias.de/view.php?id=41268


klees:
Hi @thibsy,

is this something that could have been caused by some refactoring in `ilCtrl`? This was reported in the context of #7019, but I currently cannot see how this could be connected to the changes in there. So we wonder: What broke this?

Kind regards!

thibsy:
@klees this is most likely caused by the `ilCtrl` refactoring, yes. I cannot put my finger on it though, because there have been many small behavioural changes. If we recognise a pattern however, let me know.

PR #7434, opened 2024-04-25
fix objective results when tests use fractional point values

ingoj:
see mantis 41217, tests which use fractional point values lead to miscalculation of objective results. This is due to some typecasts to int and rounding effects.


ingoj:
may apply to release_9 and trunk, too

smeyer-ilias (approved):
@ingoj Thanks for fixing this and providing a PR.

PR #7435, opened 2024-04-26
Certificate Bugfix #0041252: Certificates aren't issued in the language of the receiving user in any case

fhelfer:
https://mantis.ilias.de/view.php?id=41252

### DIFF
--- a/Services/Certificate/classes/Placeholder/Values/class.ilCoursePlaceholderValues.php
+++ b/Services/Certificate/classes/Placeholder/Values/class.ilCoursePlaceholderValues.php
@@ -131,7 +131,20 @@ class ilCoursePlaceholderValues implements ilCertificatePlaceholderValues
             $placeholders['DATETIME_COMPLETED'] = $this->dateHelper->formatDateTime($completionDate);
         }
 
-        $placeholders['COURSE_TITLE'] = $this->ilUtilHelper->prepareFormOutput($courseObject->getTitle());
+        $lng_code = ilObjUser::_lookupLanguage($userId);
+        $course_translation = $courseObject->getObjectTranslation();
+        $title = $courseObject->getTitle();
+        if ($course_translation instanceof ilObjectTranslation) {
+            $languages = $course_translation->getLanguages();
+            foreach ($languages as $trans) {
+                if ($trans->getLanguageCode() === $lng_code) {
+                    $title = $trans->getTitle();
+                    break;
+                }
+            }
+        }
+
+        $placeholders['COURSE_TITLE'] = ilLegacyFormElementsUtil::prepareFormOutput($title);
 
         return $placeholders;
     }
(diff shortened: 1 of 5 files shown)


Services/Certificate/classes/Placeholder/Values/class.ilDefaultPlaceholderValues.php
```diff
@@ -111,6 +111,8 @@ public function getPlaceholderValues(int $userId, int $objId): array
         if (!$user instanceof ilObjUser) {
             throw new ilException('The entered id: ' . $userId . ' is not an user object');
         }
+        $this->language = new ilLanguage($user->getLanguage());
+        $this->language->loadLanguageModule('certificate');
```
mjansenDatabay: I suggest to use a local variable `$usr_language` for this, instead of overwriting the member variable. We should rename the member variable to `actor_language`.

Services/Certificate/classes/Placeholder/Values/class.ilCoursePlaceholderValues.php
```diff
@@ -156,9 +178,23 @@ public function getPlaceholderValuesForPreview(int $userId, int $objId): array
...
+            'SELECT title' . PHP_EOL
+            . 'FROM object_translation' . PHP_EOL
+            . 'WHERE obj_id = ' . $this->db->quote($objId, 'integer') . PHP_EOL
+            . 'AND lang_code = ' . $this->db->quote($lng_code, 'text') . PHP_EOL;
+        $r = $this->db->query($sql);
+        $row = $r->fetchRow(ilDBConstants::FETCHMODE_OBJECT);
+
+        if ($row) {
+            $title = (string) $row->title;
+        }
```
mjansenDatabay: Do we really need this database query? IMO it should be sufficient to rely on `ilObject` (and maybe additionally on `ilObjectTranslation`.

mjansenDatabay (changes requested):

Services/Certificate/classes/Placeholder/Values/class.ilCoursePlaceholderValues.php
```diff
@@ -131,7 +130,21 @@ public function getPlaceholderValues(int $userId, int $objId): array
             $placeholders['DATETIME_COMPLETED'] = $this->dateHelper->formatDateTime($completionDate);
         }
 
-        $placeholders['COURSE_TITLE'] = $this->ilUtilHelper->prepareFormOutput($courseObject->getTitle());
+        $lng_code = ilObjUser::_lookupLanguage($userId);
+        /** @var ilObjCourse $object */
+        $object = $this->objectHelper->getInstanceByObjId($objId);
```
mjansenDatabay: We do already have `$courseObject`, this should be sufficient, shouldn't it?

mjansenDatabay (changes requested):

Services/Certificate/classes/Placeholder/Values/class.ilCoursePlaceholderValues.php
fhelfer: you are right,thx

Services/Certificate/test/ilCertificateBaseTestCase.php
```diff
@@ -18,12 +18,41 @@
...
+        if (!defined('ANONYMOUS_USER_ID')) {
+            define('ANONYMOUS_USER_ID', 13);
+        }
+
+        global $DIC;
+
+        $this->dic = is_object($DIC) ? clone $DIC : $DIC;
+
+        $DIC = new Container();
+        $DIC['legalDocuments'] = fn() => $this->getMockBuilder(Conductor::class)->disableOriginalConstructor()->getMock();
```
mjansenDatabay: Do we really need this?

mjansenDatabay (changes requested):

mjansenDatabay (approved):

PR #7436, opened 2024-04-26
ComponentsRevision: fix WAC

nhaagen:
while working on https://mantis.ilias.de/view.php?id=41201, I found the reason for content-styles not being loaded was the htacces and WebAccessChecker needed adjusting to the new public/data structure.


klees (approved):

chfsx:
Thanks a lot!

PR #7437, opened 2024-04-26
Remove mailing lists on user deletion

iszmais:
https://mantis.ilias.de/view.php?id=41275

- Delete all of users mailing lists when deleting a user.

### DIFF
--- a/Services/Contact/classes/class.ilContactAppEventListener.php
+++ b/Services/Contact/classes/class.ilContactAppEventListener.php
@@ -37,6 +37,7 @@ class ilContactAppEventListener implements ilAppEventListener
             $user->setId((int) $a_parameter['usr_id']);
 
             $mailingLists = new ilMailingLists($user);
+            $mailingLists->deleteLists();
             $mailingLists->deleteAssignments();
         }
 
--- a/Services/Contact/classes/class.ilMailingLists.php
+++ b/Services/Contact/classes/class.ilMailingLists.php
@@ -144,6 +144,20 @@ class ilMailingLists
         }
     }
 
+    public function deleteLists(): void
+    {
+        $this->db->manipulateF(
+            'DELETE ass FROM addressbook_mlist_ass ass INNER JOIN addressbook_mlist list ON ass.ml_id = list.ml_id WHERE list.user_id = %s',
+            ['integer'],
+            [$this->user->getId()]
+        );
+        $this->db->manipulateF(
+            'DELETE FROM addressbook_mlist WHERE user_id = %s',
+            ['integer'],
+            [$this->user->getId()]
+        );
+    }
+
     public function deleteAssignments(): void
     {
         $this->db->manipulate(


Services/Contact/classes/class.ilMailingLists.php
```diff
@@ -144,6 +144,25 @@ public function deleteTemporaryLists(): void
...
+    {
+        $res = $this->db->queryF(
+            'SELECT * FROM addressbook_mlist_ass as ass INNER JOIN addressbook_mlist as main ON main.ml_id = ass.ml_id WHERE usr_id = %s',
+            [ilDBConstants::T_INTEGER],
+            [$this->user->getId()]
+        );
+
+        $entries = [];
+        while ($row = $res->fetchRow(ilDBConstants::FETCHMODE_OBJECT)) {
+            $entries[] = new ilMailingList(new ilObjUser((int) $row->user_id), $row->ml_id);
```
mjansenDatabay:
```suggestion
            $entries[] = new ilMailingList(new ilObjUser((int) $row->user_id), (int) $row->ml_id);
```

Services/Contact/classes/class.ilMailingLists.php
```diff
@@ -144,6 +144,25 @@ public function deleteTemporaryLists(): void
...
+    public function getAssigned(): array
+    {
+        $res = $this->db->queryF(
+            'SELECT * FROM addressbook_mlist_ass as ass INNER JOIN addressbook_mlist as main ON main.ml_id = ass.ml_id WHERE usr_id = %s',
+            [ilDBConstants::T_INTEGER],
+            [$this->user->getId()]
+        );
+
+        $entries = [];
+        while ($row = $res->fetchRow(ilDBConstants::FETCHMODE_OBJECT)) {
```
mjansenDatabay:
```suggestion
        while ($row = $ilDB->fetchObject($res)) {
```

Services/Contact/classes/class.ilMailingLists.php
mjansenDatabay: Creating user instances here is very expensive in terms of performance. We should avoid this in scenarios where dozens of user accounts can potentially be deleted at once.

Please use simple SQL statements within `ilMailingLists`.

Services/Contact/classes/class.ilContactAppEventListener.php
```diff
@@ -37,7 +37,19 @@ public static function handleEvent(string $a_component, string $a_event, array $
...
 
             $mailingLists = new ilMailingLists($user);
+            foreach ($mailingLists->getAll() as $mailingList) {
+                $mailingList->delete();
+            }
+            $assigned = $mailingLists->getAssigned();
             $mailingLists->deleteAssignments();
+            foreach ($assigned as $ass) {
+                if (ilBuddyList::getInstanceByUserId((int) $a_parameter['usr_id'])->getLinkedRelations()->isEmpty()) {
+                    $related = new ilMailingLists(new ilObjUser($ass->getUserId()));
```
mjansenDatabay: Creating user instances is very expensive, we should avoid this in scenarios where dozens of user accounts can potentially be deleted at once.

Please use simple SQL statements within `ilMailingLists`.

mjansenDatabay (changes requested):

iszmais:
@mjansenDatabay I separated both problems in 2 different mantis issues.

Therefore this now only fix the missing deletion and all concerned scoped are deleted.
I will make a separate PR for the [other](https://mantis.ilias.de/view.php?id=41278)

Services/Contact/classes/class.ilContactAppEventListener.php
```diff
@@ -37,6 +37,9 @@ public static function handleEvent(string $a_component, string $a_event, array $
             $user->setId((int) $a_parameter['usr_id']);
 
             $mailingLists = new ilMailingLists($user);
+            foreach ($mailingLists->getAll() as $mailingList) {
+                $mailingList->delete();
+            }
```
mjansenDatabay: Please introduce a new method `deleteAll` which deletes all these lists with one single SQL statement.

mjansenDatabay (changes requested):

mjansenDatabay (approved):

PR #7439, opened 2024-04-26
UI: Introduce Is Nullable to Certain Column Types

kergomard:
Hi everybody

This PR is first to start a discussion. I would also implement this afterwards, if we think some implementation is needed, but I would then ask you to be a little bit patient with me (this is outside any contract).

While working on the table to show logs in the test I had the requirement to add a column with optional content (in my case a link). This is currently not possible. Right now it is possible to render the link and then insert it as a text column, but I don't believe this to be the right solution either.

This PR thus proposes to add a function `Column::withIsNullable(bool $nullable): void` to certain column types, but not all.

Open questions and explanations:
* I did not add `withIsNullable` to all column-types, because I believe that certain types, e.g. `Status` or `Boolean` profit from being clear: In these cases the options are clear and should always be defined. Other types contain primarily text (`Text`and `EMail`) and here I think it is better to force the use of empty strings for empty values instead of offering two different options to achieve the same thing, although this argument is a little controversial, as there still is a difference between the empty value and a non-existent value. `Number` in my eyes is different and a clear case of the latter, as we have a clear and visual difference between the neutral value, ie. `0`, and a non-existent value, ie. `null`.
* I think the TimeSpan is another interesting case, where my current thinking is that `nullable` should mean that either the left or the right or both values can be `null`. I would argue that this avoids the need to introduce more of these "configuration functions" as all this scenarios are very well possible (and I would argue will come up).
* Finally sorting: I would suggest to add empty values to the end of the list, as I this feels more logical to me (fuzzy argument, I know). The argument for putting it first: Our main database sorts them in first by default (I know: changeable). I have no clear-cut opinion on this.

If I get it right the functions on the interfaces of the table come with little in the way of doc-blocks and I think this should be self explanatory, too, but I could add some information to the `Column\Factory` if desired.

Thank you very much and best,
@kergomard

### DIFF
--- a/components/ILIAS/UI/src/Component/Table/Column/Date.php
+++ b/components/ILIAS/UI/src/Component/Table/Column/Date.php
@@ -25,4 +25,5 @@ use ILIAS\Data\DateFormat\DateFormat;
 interface Date extends Column
 {
     public function getFormat(): DateFormat;
+    public function withIsNullable(bool $nullable): void;
 }
--- a/components/ILIAS/UI/src/Component/Table/Column/Link.php
+++ b/components/ILIAS/UI/src/Component/Table/Column/Link.php
@@ -22,4 +22,5 @@ namespace ILIAS\UI\Component\Table\Column;
 
 interface Link extends Column
 {
+    public function withIsNullable(bool $nullable): void;
 }
--- a/components/ILIAS/UI/src/Component/Table/Column/LinkListing.php
+++ b/components/ILIAS/UI/src/Component/Table/Column/LinkListing.php
@@ -22,4 +22,5 @@ namespace ILIAS\UI\Component\Table\Column;
 
 interface LinkListing extends Column
 {
+    public function withIsNullable(bool $nullable): void;
 }
--- a/components/ILIAS/UI/src/Component/Table/Column/Number.php
+++ b/components/ILIAS/UI/src/Component/Table/Column/Number.php
@@ -27,4 +27,5 @@ interface Number extends Column
 
     public function withDecimals(int $number_of_decimals): self;
     public function withUnit(string $unit, string $unit_position = self::UNIT_POSITION_AFT): self;
+    public function withIsNullable(bool $nullable): void;
 }
--- a/components/ILIAS/UI/src/Component/Table/Column/TimeSpan.php
+++ b/components/ILIAS/UI/src/Component/Table/Column/TimeSpan.php
@@ -25,4 +25,5 @@ use ILIAS\Data\DateFormat\DateFormat;
 interface TimeSpan extends Column
 {
     public function getFormat(): DateFormat;
+    public function withIsNullable(bool $nullable): void;
 }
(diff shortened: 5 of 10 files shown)


kergomard:
This is now partially superseded by https://github.com/ILIAS-eLearning/ILIAS/pull/7566 . There are more column-types that might profit from some kind of "nullable" state (see above), but I don't have much of a horse in the race anymore. Depending on your decision, I'm still willing to further work on this.

klees (commented):
Hi @kergomard,

thanks a lot for this proposal, especially if this was made in you private time. We will of course be patient 🙂

We like this proposal and would really much appreciate if you (or someone else...) would move forward with this.

Regarding your considerations and questions:

* regarding nullable column types: We think that it would be a good idea to add the option to all column types. There might be arguments for or against nullability depending on the type, but we also might miss some argument dependending on the use case. E.g. for a boolean "successfull"/"not successfull": For "not determined yet" we could very well switch to a status column with three possible values, but we could also say nullable boolean would be correct since we just cannot tell the value of that boolean. Same goes for "empty title" as string vs. "no title at all". Let's not go there and support nullability for all column types. This might also simplify handling of nulls in styling, output, etc.
* regarding timespan: The case you are making here would the case for some more specific mutators on that concrete column type. Do we allow missing startdates? Enddates? This is something different than "we do not require a timespan". Let's put these more specific cases aside for the moment.
* regarding sorting: We think we cannot say per se if null would need to be first or last for a given column. This even might be different per use case of one table. Changing that order might become a feature of the complete table, a preference of a user or even an option per nullable column. However, we would need to communicate that to the DataRetrieval, so an according change there and setting it to a fixed value (null-first or null-last) would be fine for the moment.

Feel free to contact us for date to further discuss this proposal, if required. We are looking forward to future iterations.

Kind regards!
@Amstutz, @thibsy and @klees

klees:
@kergomard I have assigned this back to you to get this off my list, feel free to ping me and assign back if there is a next iteration.

kergomard:
Thank you very much for the Feedback @klees, @Amstutz , and @thibsy !

I realized something that I only figured out by a mistake: I don't believe anymore, that this is needed and I'm sorry to have caused you work!

There already is a mechanism for this: You just don't specify the corresponding key in the array passed to the `DataRowBuilder`. It just didn't cross my mind that this was an option. Am I missing something?

If this is the way to go, I would provide a PR to expand the examples correspondingly. If I'm wrong, and we would prefer to have this more verbose approach, I can go on working on this PR.

Sorry again and best,
@kergomard

klees:
I'm currently not sure if we support this deliberately or just by accident. If there is no pointer to that functionality in the examples or descriptions, I would guess this is by accident. I will look into this some time soon and try to give a more informed opinion.

Kind regards!

oliversamoila:
Hello everyone, 
I stumbled across this and the related PR in https://mantis.ilias.de/view.php?id=39326 again. 
As a non-expert on the implementation details, can you give me some info on when we can expect a next step here? I'm just a little concerned that it's assumed that it's the other one's turn ;)
Thank you very much for your work on this topic.

PR #7440, opened 2024-04-26
Fix blank error page and wrong header status

thowlfloen:
Outputs an error message if the database could not be connected and adds the header status 500.

### DIFF
--- a/error.php
+++ b/error.php
@@ -21,11 +21,14 @@ try {
     $tpl->setContent($local_tpl->get());
     $tpl->printToStdout();
 } catch (Exception $e) {
+    header("HTTP/1.1 500 Internal Server Error");
     if (defined('DEVMODE') && DEVMODE) {
         throw $e;
     }
 
     if (!($e instanceof \PDOException)) {
         die($e->getMessage());
+    } else {
+        die('Database Connection Failed.');
     }
 }


error.php
```diff
@@ -21,11 +21,14 @@
     $tpl->setContent($local_tpl->get());
     $tpl->printToStdout();
 } catch (Exception $e) {
+    header("HTTP/1.1 500 Internal Server Error");
```
mjansenDatabay, line 24: (Nowadays) I suggest using the `$DIC['http`]` service to prepare and finally send the response.

nicoroeser, line 24: Mind the effect of putting this _here_ (see the next lines about `DEVMODE`).

nicoroeser: … thinking about it again, won’t be an issue in practice, as user agents will normally also display the body of HTTP 500 responses. Sorry. Should still work in dev mode.

pascalseeland: In general I would agree to use the $DIC, but I think that if we have already hit the error handler and are even failing there, I would not rely on having any kind of $DIC, etc available. I would be fine to merge this PR.

mjansenDatabay:
Hi @thowlfloen ,

`release_8` is not maintained anymore. If you like, you could rebase your PR-branch and change the target to `release_9`, or you can close this PR and create a new one.

Best regards,
Michael

thowlfloen:
No thanks.

PR #7442, opened 2024-04-26
Fix wrong permission check on mail lists

iszmais:
https://mantis.ilias.de/view.php?id=41278

- Allows users to access their mailing lists when they have at least 1, even when they have no contacts.


mjansenDatabay:
Thx @iszmais

PR #7445, opened 2024-04-26
Fix 41280: Modal::getCloseSignal not working

tfamula:
https://mantis.ilias.de/view.php?id=41280

This must be cherry-picked to all maintained versions. Until now, there are no usages of this method. I assume that is the reason it was never noticed.


Amstutz:
you are correct, there is no close in BS Modals, see: https://getbootstrap.com/docs/3.3/javascript/

Amstutz:
Picked to 8 and 9. Please notify if anybody needs this for 7 as well.

PR #7447, opened 2024-04-29
Accessibility: Fix updating documents

mjansenDatabay:
Updating existing documents seems to be broken, if the upload request has been already "processed".

Caused by: https://github.com/ILIAS-eLearning/ILIAS/commit/7e77741e0abf0f9a9cfaa882ea5bddb73b3f7d88

Mantis Issue: https://mantis.ilias.de/view.php?id=41289


tfamula:
Thank you very much @mjansenDatabay and sorry for the delay (was not intentional)! I also picked this to release 9 and trunk.

PR #7449, opened 2024-04-30
Container/Copy: 41174, copying objects to several targets creates too many copies

nhaagen:
https://mantis.ilias.de/view.php?id=41174


kergomard:
Thanks again for the PR @nhaagen !
As we cannot reproduce the issue anymore, I close this PR.

PR #7450, opened 2024-04-30
T&A: 41152, Fixed division by zero exception in assFormulaQuestionResult::convertDecimalToCoprimeFraction()

matheuszych:
Will be reviewed by @thojou & @mbecker-databay

Upon calling the aformentioned method with an empty string or value that could not be cast into a numeric value or that would be cast to 0 the method would throw an exception because at some point it would try to divide by 0.

This is not a perfect solution but it is a possible work around. The issue is also present in ILIAS 7. The actual issue goes much deeper than just the Method that is causing the Exception. The value given to the Method is sometimes invalid and thus the issue is the value beeing wrongly put together before.

Because the code has a very high cyclomatic complexity it is hard to debug. If the core issue shall be fully resolved, a general overhaul of the code is needed first to clear up the complexity.

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assFormulaQuestionResult.php
+++ b/Modules/TestQuestionPool/classes/class.assFormulaQuestionResult.php
@@ -846,6 +846,10 @@ class assFormulaQuestionResult
 
     public static function convertDecimalToCoprimeFraction($decimal_value, $tolerance = 1.e-9)
     {
+        if (empty($decimal_value)) {
+            return '';
+        }
+
         $to_string = (string) $decimal_value;
         $is_negative = strpos($to_string, '-') === 0;
         if ($is_negative) {


thojou (approved):
Ensuring that the `convertDecimalToCoprimeFraction` method doesn't encounter a division by zero is a good idea. This adjustment makes sure it works the same as it does in release_7.

We need to include these modifications in both release_9 and trunk.

@mbecker-databay, @kergomard, could you take a look at this and merge it? LGTM!

kergomard:
I think it is right to be pragmatic here, but, I've still a little but here:
- I don't believe we are checking for `true`/`false` here, so I don't think `!` is the right operator. I know, it is the easiest one, but could we maybe try to give this a little more information? My first reaction would be that `empty()` would be the right check.

If this is 8 only, this would be ok for me, but we are picking this upwards, right?

@kergomard

matheuszych:
Hello @kergomard
I get your point. I adjusted it to use empty instead of !.

It would be optimal to include this workaround in all current versions (7, 8, 9  & 10).

kergomard:
Thanks again @matheuszych !
Merged and picked to ILIAS 7, ILIAS 9, and trunk.

PR #7451, opened 2024-04-30
SOAP: (41174), fix includes

nhaagen:
fixes includes in SoapAdmin; e.g. copy failed.


kergomard:
Hi @nhaagen and @jeph864   

What is the plan with this? This looks like a PR that can be closed, right?

Thank you very much and best,
@kergomard

PR #7454, opened 2024-04-30
UI: Fix In-Accessible Second Level Items Drilldown

kergomard:
Hi everybody

This fixes a small issue with the new drilldown that leads to second level items not being accessible, when no first level item is selected: This happens when you open a drilldown where there is no saved state.

Best,
@kergomard


Amstutz:
thx a lot!

PR #7458, opened 2024-04-30
UI port DataTable: init il, if not present

nhaagen:


thibsy:
Hi @nhaagen

Thx for looking into this. I noticed we have already solved this issue for ILIAS 9 in #6113 (initially #5803), so I have cherry-picked this solution to release_8 instead (25139f0fe323e1ea237b9ea6d2216531a7427ce3).

Kind regards,
@thibsy

PR #7459, opened 2024-04-30
UI Component, ViewControl rollback to state before DataTable 41298

catenglaender:
Mantis https://mantis.ilias.de/view.php?id=41298

# Issue

DataTable port from ILIAS 9 down to ILIAS 8 introduced styling from the UI Input View Controls clashing with the legacy and UI ViewControls.

For example, pagination should look like the one on the left. The DataTable backport introduced a mixed styling seen here to the right:

image

# Change

* Many files (viewcontrols.less, panel.less, button.less) have been rolled back to (almost) exactly the state they had before the backport, which resets the look of the view controls to its previous ILIAS 8 state. This ensures near 100% compatibility with existing ILIAS 8 skins.
* A few lines of patches and fixes now make the UI Input viewcontrols introduced by the DataTable look like ILIAS 8 viewcontrols - they do in fact share most of the code with the legacy viewcontrols.
* DataTable stylecode has been updated to the current state of the ILIAS 9 DataTable which introduces better support for narrow screens. The code is nearly a 1:1 translation from SCSS to LESS - only media queries and the multi-action triggerer margin differ.
* removed the elementbar utility CSS from ILIAS 8 in favor of a few lines of easier to maintain component specific CSS

Here are some screenshots:

Legacy pagination is running on the previous style code:
image

DataTable viewcontrols with ILIAS 8 look:
image

# Left to do in other PRs

Select all toggle is broken, but it was broken when I started working on this
image

ILIAS 8 DataTable doesn't have the footer column that was recently introduced to the ILIAS 9 DataTable which now holds the multi select triggerer.

### DIFF
--- a/src/UI/templates/default/Button/button.less
+++ b/src/UI/templates/default/Button/button.less
@@ -20,7 +20,7 @@ input.btn{
   }
 }
 
-.btn-default {
+.btn-default, .btn-ctrl {
   .button-variant(@il-btn-standard-color; @il-btn-standard-bg; @il-btn-standard-border);
   &:focus-visible {
     position: relative;
--- a/src/UI/templates/default/Panel/panel.less
+++ b/src/UI/templates/default/Panel/panel.less
@@ -51,6 +51,26 @@
 		display: flex;
 		justify-content: space-between;
 		flex-flow: row wrap;
+		.il-viewcontrol-sortation {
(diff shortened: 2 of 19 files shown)


Amstutz:
Checks fail due to Copyright issues. Thx for the quick fix @catenglaender

PR #7460, opened 2024-05-01
[FEATURE] UI: improve context rendering

thibsy:
Hi folks,

I have removed some unnecessary complexity from the UI framework:

- `Renderer::withAdditionalContext()`: this method has been invoked in order to provide a context to the current rendering chain. However, due to the recursive nature of our rendering process, we can always know the current context because all rendering calls will eventually invoke our `DefaultRenderer` again. I have discussed with @klees to replace usages of this method by creating a new instance of the concrete context renderer already. However, I have decided not to do so until we have a proper mechanism to gather assets and inject them into our page component.
- `RendererFactory::getJSBinding()`: since instances are passed by reference, we can simply inject the `JavaScriptBinding` directly into our `DefaultRenderer`. No need to retrieve the `RendererFactory` and/or `ComponentRenderer` again.

Kind regards,
@thibsy

### DIFF
--- a/components/ILIAS/Init/classes/Dependencies/InitUIFramework.php
+++ b/components/ILIAS/Init/classes/Dependencies/InitUIFramework.php
@@ -261,7 +261,8 @@ class InitUIFramework
         };
         $c["ui.renderer"] = function ($c) {
             return new ILIAS\UI\Implementation\DefaultRenderer(
-                $c["ui.component_renderer_loader"]
+                $c["ui.component_renderer_loader"],
+                $c["ui.javascript_binding"],
             );
         };
         $c["ui.component_renderer_loader"] = function ($c) {
--- a/components/ILIAS/OrgUnit/classes/Positions/Authorities/class.ilOrgUnitGenericMultiInputGUI.php
+++ b/components/ILIAS/OrgUnit/classes/Positions/Authorities/class.ilOrgUnitGenericMultiInputGUI.php
@@ -456,10 +456,10 @@ class ilOrgUnitGenericMultiInputGUI extends ilFormPropertyGUI
          * do not render an a-tag around the glyph.
          * should be outdated and removed when Glyphs loose their Clickable
          */
-        $renderer = $this->ui_renderer->withAdditionalContext(
-            $this->ui_factory->button()->bulky($symbol, '', '')
-        );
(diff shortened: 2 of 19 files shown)


klees (changes requested):
Hi @thibsy,

thanks a lot for iterating here. These pieces look a lot more manageable than the huge change we discussed before and also have benefit on their own. This feels good.

- [ ] popContext: A downside of replacing that stateless withAdditionalContext by internal context management is that some out of band control flow (exceptions, mostly, hopefully nobody uses labels ans goto anymore...) could leave the DefaultRenderer in some invalid state. If there is some exception in between [addContext](https://github.com/ILIAS-eLearning/ILIAS/pull/7460/files?diff=split&w=0#diff-5f708921e020d3dc26167c274e247bd9ab78a12867c53a0e533ac22abc66a621R59) and [popContext](https://github.com/ILIAS-eLearning/ILIAS/pull/7460/files?diff=split&w=0#diff-5f708921e020d3dc26167c274e247bd9ab78a12867c53a0e533ac22abc66a621R62), the internal $contexts property will get out of sync with reality. This most likely won't be a problem, since system would be crashing somehow anyway. On the other hand this is exactly the kind of thing that will drive someone nuts when debugging. Please use finally to make sure that popContext will be called anyway, even if exceptions are flying.

The rest LGTU.

Thanks!

thibsy:
Hi @klees

I have rebased this PR onto the latest trunk and implemented your requested change(s).

Please note, in the meantime some external usages of `Renderer::withAdditionalContext()` have been introduced inside the `OrgUnit` and `StudyProgramme` components, which are under your authority. I have commented the usages for the moment and simply returned the component without providing additional context. Could you maybe look into this?

I also noticed that context rendering tests, such as the new filter inputs, are kind of unpractical. To make this a bit easier, I have introduced another argument for the `getDefaultRenderer()` method, which can be used to inject components as context.

Kind regards,
@thibsy

Amstutz:
Hi @thibsy

Some time passed since I last looked into your progress on the improvements of the rendering context mechanism. I am delighted to see that the code looks a lot cleaner and easier to comprehend with your proposed changes. Special thx for updating the readme and working on the test cases. 

Thx @klees for looking into this previously.

Kindly
@amstutz

PR #7461, opened 2024-05-02
MainBar/Performance: Optimize checks in `isCurrentUserAllowedToSeeCus…

mjansenDatabay:
…tomItem`

This PR changes the location of calling
`\ilRbacReview::assignedGlobalRoles`. Currently, it is called even if there is no role-based visibility configuration at all. What this PR does not address is the problem, that `\ilRbacReview::assignedGlobalRoles`
is called for every item, which results in redundant database requests (depending on the number of custom items).

### DIFF
--- a/Services/MainMenu/classes/class.ilObjMainMenuAccess.php
+++ b/Services/MainMenu/classes/class.ilObjMainMenuAccess.php
@@ -87,11 +87,11 @@ class ilObjMainMenuAccess extends ilObjectAccess implements ilMainMenuAccess
     public function isCurrentUserAllowedToSeeCustomItem(ilMMCustomItemStorage $item, Closure $current): Closure
     {
         return function () use ($item, $current): bool {
-            $roles_of_current_user = $this->rbacreview->assignedGlobalRoles($this->user->getId());
             if (!$item->hasRoleBasedVisibility()) {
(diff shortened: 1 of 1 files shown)


Services/MainMenu/classes/class.ilObjMainMenuAccess.php
```diff
@@ -87,11 +87,11 @@ public function getGlobalRoles(): array
     public function isCurrentUserAllowedToSeeCustomItem(ilMMCustomItemStorage $item, Closure $current): Closure
     {
         return function () use ($item, $current): bool {
-            $roles_of_current_user = $this->rbacreview->assignedGlobalRoles($this->user->getId());
             if (!$item->hasRoleBasedVisibility()) {
                 return $current();
             }
             if (!empty($item->getGlobalRoleIDs())) {
+                $roles_of_current_user = $this->rbacreview->assignedGlobalRoles($this->user->getId());
```
chfsx, line 94: @mjansenDatabay  thank you very much! I currently think about reading the global roles once per request and store them staticly to avoid multiple requests. I'll have a look at this asap.

chfsx:
Thanks a lot, I cherry-picked and updated the solution

PR #7462, opened 2024-05-02
UI: Add Info When No Items In Drilldown

kergomard:
Hi everybody
@oliversamoila  noticed that we are currently not showing any information in a `Drilldown` that is empty, so I changed that.

If this PR is merged, if you filter everything from a `Drilldown` you will see:
grafik

Thanks and best,
@kergomard

### DIFF
--- a/components/ILIAS/Init/classes/Dependencies/InitUIFramework.php
+++ b/components/ILIAS/Init/classes/Dependencies/InitUIFramework.php
@@ -182,7 +182,8 @@ class InitUIFramework
         };
         $c["ui.factory.menu"] = function ($c) {
             return new ILIAS\UI\Implementation\Component\Menu\Factory(
-                $c['ui.signal_generator']
+                $c['ui.signal_generator'],
+                $c['lng']
             );
         };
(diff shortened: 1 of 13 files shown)


Amstutz:
Hi @kergomard

Thx a lot for tackling this! 

First a non-actionable note, I personally do not like property promotion. However, they are already in use and my colleagues do not mind. So I won't either. 

Change or shortly explain why not:
- [x] No Item Text: You put the no item text directly on the components object. I would regard this rather as just a rendering detail, and would therefore rather opt to just pack it directly into the rendering. This has the additional advantage, that the lng dep. is already available there.

Thx a lot
@amstutz in consultation with @klees and @thibsy

kergomard:
Hi @Amstutz , hi all
Thank you very much for the review.
You are completely right: I moved text to the renderer.
I also removed the  property promotion as I don't need to touch this class anymore. Personally I'm a big fan, as I think that it improves readability hugely and it reduces the potential for errors.

Thanks again and best,
@kergomard

Amstutz:
Thx a lot!

PR #7464, opened 2024-05-02
fixes mantis 39202

noackorama:


mjansenDatabay:
@jeph864 / @noackorama  You also have to create a PR for `trunk` if this is relevant for ILIAS 10.

PR #7466, opened 2024-05-02
COPage: Fix Content Style for Ordering Question

kergomard:
If accepted this would need to be picked to trunk.
See: https://mantis.ilias.de/view.php?id=41317


kergomard:
Ok, I close this PR again as the corresponding issue has been fixed the same way with bd45baabd26eecd7dda9c2e99e07f7ec51d614e6

PR #7469, opened 2024-05-03
[FEATURE] UI: simplify component renderers

thibsy:
Hi folks,

This is one part of the PR I created to introduce JavaScript hydration to the UI framework (#6948).

It centralises the `ComponentRenderer::render()` function inside our `AbstractComponentRenderer` by using a new abstract method called `renderComponent()` and gets rid of the `AbstractComponentRenderer::checkComponent()` calls. I also tried to streamline the `instanceof` checks a bit, and returned null in methods which threw a `LogicException` or returned an empty string if the component could not be handled.

Kind regards,
@thibsy

### DIFF
--- a/components/ILIAS/UI/README.md
+++ b/components/ILIAS/UI/README.md
@@ -234,18 +234,17 @@ If you would like to implement a new component for the framework, you should per
         /**
          * @inheritdocs
          */
-        public function render(Component\Component $component, RendererInterface $default_renderer) {
-            $this->checkComponent($component);
+        public function render(Component\Component $component, RendererInterface $default_renderer): string 
+        {
+            // if this is not our component, call cannotHandleComponent($component)
+            // to throw a unified exception. 
+            if (!$component instanceof Component\Demo\Demo) {
+                $this->cannotHandleComponent($component);
+            }
+    
             $tpl = $this->getTemplate("tpl.demo.html", true, true);
             $tpl->setVariable("CONTENT",$component->getContent());
             return $tpl->get();
-        }
-
-        /**
-         * @inheritdocs
-         */
-        protected function getComponentInterfaceName() {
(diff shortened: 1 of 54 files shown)


klees (changes requested):
Hi @thibsy,

thanks a lot for this proposal. Again, this feels a lot better to handle.

- [ ] return type: I'm not really convinced that we should exchange `throw` for `return null` (and change the return type accordingly...). I see that we would want to get rid of that `checkComponent` step, it just produces duplicate lists of components, more or less. But changing `string` to `?string` for `renderComponent` to me also implies that we somehow see `null` as a valid and normal return value for that method. Which, in my understanding, is not the case. That method should, with correct code, never in fact return `null` but always a string. So, to not have that `throw`-thingy inline: How about we change the return type back to `string` and introduce a method `cannotHandleComponent` (or something similar) on the `AbstractComponentRenderer`. This method could take a `Component` as argument and then throw an according exception or do whatever in the future. We could then also just keep the name of that method `render`.

- [ ] Remove code: Please remove the implementations of `AbstractComponentRenderer::getComponentInterfaceName` in the single renderers, we don't seem to need that method anymore.

The rest looks plausible, let's do this =)

Best regards!
@klees

thibsy:
Hi @klees

As discussed, I have removed the `renderComponent()` method again and introduced a unified `cannotHandleComponent()` method instead. I also removed the `getComponentInterfaeName()` method, since we can determine the supported components by investigating the namespace(s).

To recap for the public as well: The idea behind this change was to create a place inside the `AbstractComponentRenderer` to implement the JavaScript hydration process. However, we noticed the tooltip rendering has a similar issue like we do with JavaScript, so we have decided to solve the JavaScript hydration process using `Closure`s as well.

Kind regards,
@thibsy

PR #7486, opened 2024-05-06
DBMS: Proposed Version range für MySQL/MariaDB

chfsx:


matthiaskunkel:
Jour Fixe, 13 MAR 2024: We accept the suggested versions for MySQL and MariaDB for ILIAS 10 / trunk.

PR #7488, opened 2024-05-07
UI, ViewControls/DataTable: 41310, add nullControl to reserve input name

nhaagen:
https://mantis.ilias.de/view.php?id=41310

### DIFF
--- a/src/UI/Component/Input/ViewControl/Factory.php
+++ b/src/UI/Component/Input/ViewControl/Factory.php
@@ -105,8 +105,27 @@ interface Factory
      *   effect: >
      *      Each view control will be rendered in the same order as provided.
      * ---
-     * @param \ILIAS\UI\Component\Input\Container\ViewControl\ViewControlInput[] $view_contorls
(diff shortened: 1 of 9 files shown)


klees:
@thibsy / @Amstutz: CaT internal QS is done, this seems to be good for your review.

Amstutz:
Hi @nhaagen

Thx a lot for this enhancement. It was not immediately clear to me, why we would need this instead of living with null. However, I believe to understand the advantage (thx @klees).

Therefore, please change or indicate why not:
- [x] Enhance the purpose description: Maybe drop a phrase that we use this where we otherwise would need to pass null? It might be obvious to most, but could help some.

Please change:
- [x] Add a rendering test: Making sure nothing is rendered.

One last thing: Thx for the code style improvements. However, they make PRs sometimes difficult to read. Maybe put them in separate PRs in the future.

Thx a lot
@thibsy, @klees  and @amstutz

matthiaskunkel:
Jour Fixe, 27 MAY 2024: We highly appreciate the suggestion and accept the PR for all maintained versions (incl. 8).

Amstutz:
Thx a lot for the refinements!

PR #7490, opened 2024-05-07
Tracking/ObjectStatistics: Migrate primary key to UUID

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41330

It is not necessary to migrate existing records in `obj_stat_log`. All records (with `int` values) are just temporary and will be deleted as soon as the synchronization process (to `obj_stat_tmp`) is triggered.

### DIFF
--- a/Services/Tracking/classes/Setup/Migration/MigrateObjectStatisticsPrimaryKey.php
+++ b/Services/Tracking/classes/Setup/Migration/MigrateObjectStatisticsPrimaryKey.php
@@ -1,0 +1,61 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\Tracking\Setup\Migration;
+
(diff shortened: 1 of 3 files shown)


alex40724:
@mjansenDatabay One stupid question: How are the deletion queries on obj_stat_log_seq triggered at all?

mjansenDatabay:
@alex40724 The deletion is done here https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/Services/Database/classes/PDO/class.ilDBPdoMySQL.php#L164

alex40724:
@mjansenDatabay Yes thanks, I just found that, but doesn't this make that an issue of the nextId implementation? This should work "atomQuery" like imo.

mjansenDatabay:
@alex40724 You are right, it is caused by the actual database implementation of `nextId` (which is still similar to the `MDB2` implementation > 10 years ago, see: https://github.com/ILIAS-eLearning/ILIAS/blob/4.3.5/Services/PEAR/lib/MDB2/Driver/mysqli.php#L1094). The question is, and I don't have an answer to this: Why does this happen on multiple installations and **only** with the "Object Statistics" sequence table?

alex40724:
@mjansenDatabay This might be the table with the quickest most frequent nextId() usage in ILIAS. So the probability to get race conditions is higher, once there are multiple entries, the deletion process will slow down (I guess it will be a full table scan).

kergomard:
Hi @mjansenDatabay 

This looks like a PR that should be closed, right?

Thank you very much and best,
@kergomard

mjansenDatabay:
Hi @kergomard ,

I personally don't suggest closing it: There are several problems with the "Object Statistics" subcomponent.

1) Performance: https://mantis.ilias.de/view.php?id=41330
2) Integer Overflow: https://mantis.ilias.de/view.php?id=43591

I doubt that these problems will be solved centrally soon.

Best regards,
Michael

mjansenDatabay:
Hi @smeyer-ilias . If you like, I can provide a PR for the maintained release branches or at least for `trunk`. What do you think?

PR #7492, opened 2024-05-07
UI, Interruptive: 38085, allow for both form methods & provide individual parameters for items

nhaagen:
https://mantis.ilias.de/view.php?id=38085

This lets you 
- construct an Interruptive Modal with GET-method 
- and  provide individual parameter names for its items

Comes in handy, e.g., when using delete-confirmations from a Data Table.

### DIFF
--- a/src/Data/FormMethod.php
+++ b/src/Data/FormMethod.php
@@ -18,24 +18,10 @@
 
 declare(strict_types=1);
 
-namespace ILIAS\UI\Component\Modal\InterruptiveItem;
+namespace ILIAS\Data;
 
-use ILIAS\UI\Component\Component;
-
-/**
- * Interface InterruptiveItem
- *
- * Represents an item to be displayed inside an interruptive modal
- */
-interface InterruptiveItem extends Component
+enum FormMethod: string
 {
-    /**
-     * Return an ID of the item
-     */
-    public function getId(): string;
-
-    /**
-     * Use this as a parameter name when submitting the interruptive modal
-     */
-    public function withParameterName(string $parameter_name): self;
+    case GET = 'GET';
+    case POST = 'POST';
 }
--- a/src/Data/FormMethod.php
+++ b/src/Data/FormMethod.php
@@ -24,4 +24,5 @@ enum FormMethod: string
 {
     case GET = 'GET';
     case POST = 'POST';
+    case DIALOG = 'DIALOG';
 }
--- a/src/UI/Component/Modal/InterruptiveItem/Factory.php
+++ b/src/UI/Component/Modal/InterruptiveItem/Factory.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 namespace ILIAS\UI\Component\Modal\InterruptiveItem;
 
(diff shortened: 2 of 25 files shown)


klees (approved):

thibsy (changes requested):
Hi @nhaagen

Thanks a lot for your contribution to the UI components, this is a great usability improvement!

This PR also addresses the submission problem I was trying to describe in #5773.

- [ ] Form methods: according to [mozilla's specification](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#method) about HTML form methods, another possible value is `dialog`. What happens when we inevitably add this to the enum? The `Interruptive` modal - and all components accepting this enum in the future - would need to be able to react to all three methods. Does this work, or do we need to keep a handle on supported methods? Also, do we need to drop the argument for `Interruptive` modals again once #7045 is merged?

- [ ] Parameter ownership: could you explain why the `InterruptiveItem` has ownership over the parameter name, instead of the `Interruptive` modal? If we compare this to e.g. form containers, the ownership belongs to the form, not the input. It is possible to partially shift this over to an input using `withDedicatedName()`, but never entirely. IMO it would make sense to do the same here and put the parameter name onto the modal.
- [x] Query parameters: you are solving the same issue as `ViewControl` containers. Is it possible to streamline this so the view control renderer and modal renderer could use the same function (of a trait?) to inject the query parameters? This way we can also keep an eye on these use-cases more easily.

- [x] `InterruptiveItem::withParameterName()`: I don't think the parameter name should be supplied using a mutator. Please inject this value into the constructor and introduce an additional factory parameter, which is optional in ILIAS 9 and will be required in ILIAS 10 (we will take care of this when cherry-picking).
- [x] `InterruptiveTest::getDataDactory()`: you return an actual instance of the data factory there. Please use a mock instead, which only stubs the methods you need, so we can control the outcome of this test.
- [x] Examples: I know you already provided an example for this in the data table  example (thx), but please add an example inside the modal namespace as well.

Kind regards,
@Amstutz and @thibsy

nhaagen:
> [ ] Parameter ownership: could you explain why the InterruptiveItem has ownership over the parameter name, instead of the Interruptive modal? If we compare this to e.g. form containers, the ownership belongs to the form, not the input. It is possible to partially shift this over to an input using withDedicatedName(), but never entirely. IMO it would make sense to do the same here and put the parameter name onto the modal.

With forms/inputs, the names are applied by the Form via a NameSource. 
That means that the form has to be rebuilt and data is retrieved after mapping the request.
For the modal, generating a query relaying pre-factored parameters, I really would not take that roundtrip, since there
cannot be any errors on the form itself, and returning a form would mean building the whole modal again anyway.

kergomard:
Hi @nhaagen and @thibsy 

When going through the PRs that have been around without activity, I found this one.

What is the plan with this? Any chance to move this forward?

Thank you very much and best,
@kergomard

thibsy:
Sorry, this was not on my radar because the PR was not assigned to me.

@nhaagen I assume this is still relevant? If so, could you answer the question regarding supported form methods as well? Thx!

oliversamoila:
Hello @nhaagen, 
could you say something more about the relevance and the questions mentioned by Thibeau?

nhaagen:
Hi all,
I've added the dialog-case to the enum for completeness, but restricted the Interruptive to not use it.

At the current state, configuring the Interruptive with form-method 'dialog' will not do anything
W3C says: 
> 	When the form is inside a <dialog> element, it closes the dialog and causes a submit event to be executed on submission, without submitting data or clearing the form

This certainly opens some opportunities in regard of data handling for dialog-based forms, and might still be worth looking into, since it is relevant for several components for sure; but still it is of a more technical nature in order to improve the codebase and offers little direct benefits for the consumer. For me, it feels out of scope of this PR.
Maybe the suggested compromise will help to move forward.

Best regards,
Nils

PR #7493, opened 2024-05-07
T&A 41171 Fix Whoops in detailed Test Results if no user is selected

aaronbidzan:
This PR aims to address the problem reported in Mantis issue https://mantis.ilias.de/view.php?id=41171.

If no user is selected for the problematic multi-command action, a compliant info message is shown. This replicates the behavior of the other multi-commands in the same table.

These changes may also be cherry-picked into trunk. In release_8 the error does not occur.

Kind regards.

### DIFF
--- a/Modules/Test/classes/class.ilParticipantsTestResultsGUI.php
+++ b/Modules/Test/classes/class.ilParticipantsTestResultsGUI.php
@@ -334,9 +334,12 @@ class ilParticipantsTestResultsGUI
     private function showDetailedResultsCmd(): void
     {
         $usr_ids = $this->getUserIdsFromPost();
-        if ($usr_ids !== []) {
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much for the PR @bidzanaaron and sorry it took me a little while to get around to it.

I have two little nitpicks that I fixed myself (in order to not let you do an extra round for this), here just for informational purposes:
- I replaced the `empty($usr_ids)` by `$usr_ids === []`. We know exactly what `empty` we are expecting, as we know that `$usr_ids` has to be an array. So, apart from, I believe, the better readability we gain an additional piece of information.
- And, as we are not using any variable inside the text string I set it in simple quotes (again: we gain a piece of information).

Thanks again and best,
@kergomard

PR #7495, opened 2024-05-08
class.ilLMNavigationRendererGUI: cast time_scheduled_page_activation as boolean

zemplu:
- Argument $show_blocktitle must be of type bool, not null.


alex40724:
Thanks for the PR!

PR #7496, opened 2024-05-08
UI: adjust rendering of symbols in breadcrumbs. (#41308)

lukastocker:
https://mantis.ilias.de/view.php?id=41308

As 'direction: rtl;' is used to shorten the breadcrumbs from right to left we cannot use 'dir/direction ltr;' in spans or bdo/bdi tags as it will destroy the purpose of shortening inputs that are too long in regard to the screen size. It also will change the sortation of the single breadcrumbs (e.g.: repository will be rendered at last instead at first) in case of using '<span dir:'ltr'>'.

Therefore adding a left-to-right html character into the breadcrumb template was necessary.

### DIFF
--- a/src/UI/Implementation/Component/Breadcrumbs/Renderer.php
+++ b/src/UI/Implementation/Component/Breadcrumbs/Renderer.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 namespace ILIAS\UI\Implementation\Component\Breadcrumbs;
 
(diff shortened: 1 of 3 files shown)


lukastocker:
For ILIAS 10 I suggest a refactoring of the breadcrumbs.

I already talked with @catenglaender  and he suggested a fading out for the breadcrumbs (e.g.: https://stackoverflow.com/questions/49941539/how-to-make-text-fade-to-the-left) with a scrolling option for mobile versions.
With this we could remove - if we pull this PR to ILIAS 10 before a refactoring - the changes of this PR and also the "rtl" direction in the breadcrumb SCSS class. 

I will take this suggestion to the next CSS Squad Meeting for discussion.

Amstutz:
Hi @lukastocker 

We discussed the change and decided to take it for the moment.

One thing left todo:

- [x] Unite Test: Please provide a rendering check for this

thx

Amstutz:
Any update on the unit tests? Ill assign @lukastocker , feel free to re-assign if ready for a final review.

Amstutz:
Hi @lukastocker 

Thx a lot for the fix. Highly appreciated and also for adding the TestCases.

However, we have a tpl change in there in a tpl that comes into question in customizing skins. This will lead to fatal, if not properly handled during updates. Since it is November already, 9 being productive for more than half a year... The ticket causing the change is not a major issue. I would therefore rather only fix this in 10. If anybody needs this in 9 also, it should be easy to patch with your commit being provided here, ready for a cherry-pick.

Therefore, can you provide a PR for 10, I would close this one down afterward.

Thx.

lukastocker:
> Hi @lukastocker
> 
> Thx a lot for the fix. Highly appreciated and also for adding the TestCases.
> 
> However, we have a tpl change in there in a tpl that comes into question in customizing skins. This will lead to fatal, if not properly handled during updates. Since it is November already, 9 being productive for more than half a year... The ticket causing the change is not a major issue. I would therefore rather only fix this in 10. If anybody needs this in 9 also, it should be easy to patch with your commit being provided here, ready for a cherry-pick.
> 
> Therefore, can you provide a PR for 10, I would close this one down afterward.
> 
> Thx.

Hey there,

that does sound plausible. I will provide a PR for 10 this week and ping you as soon as it's available. :)

Amstutz:
Will be merged to 10, thx a lot.

PR #5610, opened 2023-02-23
Footer goto link: Use plain link instead of button with JS

lscharmer:
Follow up to PR: #5301
With this PR the footer permanent link will be rendered as a plain link without any JS attached. The Label will show _Permanent Link_.

PR #7498, opened 2024-05-08
Footer: Copy permalink to clipboard

lscharmer:
This PR changes the permanent link in the footer to copy the link to the clipboard instead of navigating to the permanent URL.

This implementation is based on a few other attempts (#4811, #5301, #5610) and a meeting with the members of the "UI Clinic" and some of the UI coordinators where we discussed, among other things, the accessibility for this functionality. This PR is also inspired by GitHub's implementation of it's copy to clipboard function.


thibsy:
Hi @lscharmer

Thx for looking into this topic again.

Before we invest time into reviewing this, could you explain how this proposal is different from the solution using an overlay, which has been denied by JF (#5301)?

Kind regards,
@thibsy

oliversamoila:
Hi @thibsy
After the decision, which was documented with [JF 2023-02-06](https://docu.ilias.de/goto.php?target=wiki_1357_JourFixe-2023-02-06) in #5301, it has become evident over time that the behaviour of the link in the footer implemented with ILIAS 8 does cause problems in user handling.
We had a joint open meeting with you, UI coordinators, some of the UI/UX/Ally experts and other interested people from the developer and user community.
Based on the reviews and requirements of this meeting, we have made a new approach to making the difficult current handling more user-friendly. Of course, we have not reinvented the wheel, but we have, for example, given more consideration to accessibility and the existing development guidelines. 
A more comprehensive list of criteria can be found below.

* Accessibility First (screen reader)
* JS Unit Tests: Unit tests with full test coverage have been written for the code provided.
* JS Modules: The JS code was provided with ES6 modules
* JS Bundling: A rollup config with appropriate bundling was created for the JS code.
* JS Code Style: The JS code corresponds to the code style stored in ILIAS

Kind regards,
@oliversamoila

dsstrassner:
From an it-support perspective, I welcome this improvement for ILIAS 9 and hope, that the a11y and implementation aspects are now according to the UI-coordinator standards.

screenshot of a course in an installation based on this branch

oliversamoila:
If anyone wants to have a try: 
» https://copy-perma-link.invorbereitung.de
» Account: `user`
» Passwort: `ilias.9`

Bye 👋

sheepshaver:
We (DLRG) had also spoken out - solely for support reasons - in favor of replacing the current display with a solution that is understandable to the user and UX-friendly.
The solution shown here fully meets our requirements and can even be used from mobile devices. Great!
Is there a possibility of a backport in ILIAS 8? That would be the cherry on the cake.

ILIASLM:
Worked in three different browsers without a problem. Thank you! 
As there is a lot of text in the footer, my eye reads it as "Copy link to current page to clipboard powered by ILIAS (v9.0_beta3 2024-01-31)". Would it be possible to separate the link text from "powered by" with a link-icon or something?
Linktext
The icon is known, would guide the eye and separate the texts.

I also thought about where users find links in other applications like YouTube or SharePoint. It's a Share-button or a Share-option in a drop-down. Not sure if in a future-version the option Share or Copy Link To Clipboard could be added in the object's drop-down, too.

kseniialf:
Worked perfectly on the computer/mobile devices. I really appreciate how user-friendly/intuitive this solution is. Thank you!

lang/ilias_de.lang
```diff
@@ -3781,6 +3781,7 @@ common#:#copy_all#:#Alle kopieren
 common#:#copy_n_of_suffix#:#- Kopie (%1$s)
 common#:#copy_of#:#Kopie von
 common#:#copy_of_suffix#:#- Kopie
+common#:#copy_perma_link#:#Link zur aktuellen Seite in die Zwischenablage kopieren
```
matthiaskunkel: Shorter is better: 'Link in Zwischenablage kopieren'

lang/ilias_de.lang
```diff
@@ -5260,6 +5261,7 @@ common#:#pd_items_news#:#Einschließlich Neuigkeiten der ausgewählten Objekte
 common#:#pdf_export#:#PDF-Export
 common#:#perm_settings#:#Rechte
 common#:#perma_link#:#Link zu dieser Seite
+common#:#perma_link_copied#:#Link wurde in die Zwischenablage kopiert.
```
matthiaskunkel: Here we have more space for information: 'Link zu dieser Seite wurde in die Zwischenablage kopiert.'

lang/ilias_en.lang
```diff
@@ -3782,6 +3782,7 @@ common#:#copy_all#:#Copy all
 common#:#copy_n_of_suffix#:#- Copy (%1$s)
 common#:#copy_of#:#Copy of
 common#:#copy_of_suffix#:#- Copy
+common#:#copy_perma_link#:#Copy link to current page to clipboard
```
matthiaskunkel: like above: 'Copy link to clipboard'

matthiaskunkel (commented):
I like this suggestion, too. I recommend to shorten the button labels. And there need to be a clear separator between the shy button and the 'powered by...' message. Our default separator is the dot •    And by the way: how about moving the 'powered...' info to the end of the line of links. From the semantic point of view it is something different and should therefore be treated accordingly.

lang/ilias_en.lang
```diff
@@ -5260,6 +5261,7 @@ common#:#pd_items_news#:#Include News of Personal Items
 common#:#pdf_export#:#PDF Export
 common#:#perm_settings#:#Permissions
 common#:#perma_link#:#Permanent Link
+common#:#perma_link_copied#:#Link was copied to clipboard.
```
matthiaskunkel: THis got lost in my first review: Link to this page has been copied to the clipboard.

oliversamoila:
Thank you for your feedback.
Let's wait and see what the UI coordinators have to say about the implementation proposal. 

@matthiaskunkel : Of course we can adapt the language as preferred. We'll do that following the review.
You have already expressed ideas on the topic of separators and order at https://mantis.ilias.de/view.php?id=35759#c88920. However, I would like to move this behind the PR because it is a follow-up topic. I support a change.

@sheepshaver : we will find a solution for your ILIAS 8 or see you soon on 9 anyway. I probably wouldn't do anything in the official ILIAS 8 stable release.

bogenschlag:
Nice. This seems like a step forward again.

Maybe a mid dot (·) between the link text and the “powered by ILIAS …” wouldn’t do harm.

matthiaskunkel:
And maybe a topic for the UI clinic: do we keep the shy button that looks like an ordinary linked text to users? Or do we rely on existing user experience that an icon has been established for sharing links/web pages: 
grafik
Just an idea...

klees (commented):
Hi @lscharmer,

I already talked about this with @oliversamoila via discord, so just a brief feedback here, Oliver will certainly fill in the gaps:

* This will still most certainly be flagged in any accessibility test. The trigger for the action looks much better now. But the feedback that something indeed happened is only available for sighted users. If the feedback is required (I think so and you seem to think so too...) it will also be required for non-sighted users.
* This still contains inline JS-code not in a module and this code is also not covered by unit tests. There might or might not be good reasons for this, we especially do not aim for 100% test coverage. But we would expect an indication why you decided to do so instead of the incorrect claim that code is indeed fully tested and provided via modules.
* The way you attach the tooltip won't be grasped by a screenreader. There is [code to attach tooltips in an accessible way](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/UI/src/Implementation/Render/TooltipRenderer.php) but I doubt that it will fit the bill for the dynamic feedback you intend to give here.
* I'm somehow annoyed that I have again spend a non-negligible amount of my time on a solution that does not meet our (like in "ILIAS'") accessibility goals in a quite obvious way. We can not move on in an iterative way with accessibility here. That component is on virtually every page. It will most certainly be flagged by any accessibility test. The problem will undermine the assumption that our UI process can produce components that are accessible to the best of our knowledge. If we produce problems by accident (and we most certainly will) than be it. But the intended functionality here presents non-trivial challenges from an accessibility perspective that we have pointed out again and again and the flaws in the proposed solution can already be suspected just by looking at the code you present.

I really understand that (some? many?) people want that "copy by click" functionality for the permalink. Please understand that we simply cannot implement that feature in a way functional only for sighted users and leave non-sighted users with a solution that is worse than the current state.

Best regards!

@klees (in agreement with @Amstutz and @thibsy)

lscharmer:
Hi @klees,

thank you for your feedback.

I updated the PR, so that the response message should be comprehensive by screen readers. We tested this with the Apple Screen Reader, "Voice Over" in usual browsers e.g. Chrome, Firefox, Safari. In all tests the button was comprehensible as well as the feedback that the link was successfully copied to the clipboard.

As some glue JS code is needed to connect the UI Component with the copy to clipboard JS code I'm not sure how far you want the glue code to be reduced.
Should this be something like:

src/UI/Implementation/Component/MainControls/Renderer.php:444:
```php
$code = function (string $id) use ($perm_url): string {
    $id = $this->jsonEncode($id);
    $perm_url = $this->jsonEncode((string) $perm_url);

    return "document.getElementById($id).addEventListener('click', () => il.Footer.permalink.copyToClipboardWithTooltip($id, $perm_url));";
};
```

src/UI/Implementation/Component/MainControls/Renderer.php:444:
```php
$code = function (string $id) use ($perm_url): string {
    $id = $this->jsonEncode($id);
    $perm_url = $this->jsonEncode((string) $perm_url);

    return "il.Footer.permalink.copyToClipboardOnClick(document.getElementById($id), $perm_url)";
};
```

or

src/UI/Implementation/Component/MainControls/Renderer.php:444:
```php
$code = function (string $id) use ($perm_url): string {
    $id = $this->jsonEncode($id);
    $perm_url = $this->jsonEncode((string) $perm_url);

    return "il.Footer.permalink.copyToClipboardOnClick($id, $perm_url)";
};
```

If this is the case, any concrete suggestions are highly appreciated.

Best regards
@lscharmer

Amstutz:
Hi @lscharmer

We understand that the current pattern to copy link is difficult to use and learn for parts of your user base. However, we are not ready to take responsibility for the solution you propose here. The affected code is critical, since it appears on every page. A11y issues will appear in every audit performed. Nevertheless, since we will most probably have a viable solution based on the work being done on the new footer, we are willing to hand over the responsibility of this small part of the UI Components to you/Databay. We would merge this request along with PRs for 10 and trunk under the condition, that we may forward you any a11y or other issue concerning the Permalink as well as possible refactorings in the code you are committing. 

Please understand that we are willing to do this in this case, since we trust you and your colleagues regarding a) the capabilities and b) the willingness to cooperate with us shown by e.g. the LUI project manager.

Note, that if we merge this, installations that customize their footer by skin will have a high risk during the next minor update of running into a fatal error due to your change in a footer template variable. We expect that at least some installations will be affected. If this risk is fine with you, please give us the greenlight for merging this to 9, however, in doing so, you accept to handle possible repercussions of this.

Regards
@Amstutz

lscharmer:
Hi @amstutz,

thank you for your response. Your proposed solution is fine for us and we understand your reasoning.

> Note, that if we merge this, installations that customize their footer by skin will have a high risk during the next minor update of running into a fatal error due to your change in a footer template variable. We expect that at least some installations will be affected.

We tested this internally with several skins. No error is thrown but the tooltip is not shown (The link is still copied).

The issue is resolved when the delos skin is merged into the custom skin.

Please note that I added one commit after rebasing this PR as the tooltip has been cut off, which needs to be reviewed.

Best regards
@lscharmer

Amstutz:
Thx @lscharmer 

note that if you want this to be merged to 10 and/or trunk as well, we would need a further PR.

PR #7307, opened 2024-04-03
SOAP: Avoid `client.ini.php` change to make SOAP plugins work

mjansenDatabay:
This is a follow-up PR for https://github.com/ILIAS-eLearning/ILIAS/pull/7116 . Instead of setting cookies based on the `sid` passed in the SOAP request, `session_id` is used to set the current session id to load the user session when `session_start` is called.

Please test this thoroughly before merging.

If approved, this MUST be picked to `release_9` and `trunk` as well.

PR #7500, opened 2024-05-08
8/soap/fixplugins

waddahadel:
I removed some test regarding the assignUser() method, the new signature returns void, which is ill typed here, the test is already redundant so it doesn't matter if we omit it, it won't hurt the logic


Services/Init/classes/class.ilErrorHandling.php
```diff
@@ -37,7 +37,7 @@
  */
 class ilErrorHandling extends PEAR
 {
-    protected ?RunInterface $whoops;
+    protected ?RunInterface $whoops = null;
```
mjansenDatabay, line 40: Why do we need to apply this initial assignment in the `Init` component?

webservice/soap/classes/class.ilSoapRBACAdministration.php
```diff
@@ -114,15 +114,15 @@ public function addUserRoleEntry(string $sid, int $user_id, int $role_id)
         if (!$ilAccess->checkAccess('edit_permission', '', $obj_ref)) {
             return $this->raiseError('Check access failed. No permission to assign users', 'Server');
         }
-
+/* Removal of unnecessary check
```
mjansenDatabay, line 117: Maybe we should just remove the code block.

mjansenDatabay (commented):
Hi @waddahadel ,

thanks for the PR. I added some inline comments. I am primarily interested in the `ErrorHandling` change. Is this change really necessary to make SOAP plugins work, if #7307 gets merged?

Best regards,
Michael

waddahadel:
hey @mjansenDatabay ,

I think @jeph864 can answer the first comment, as for the second one, it's not really something that prevents the web service from working, I think it only throws an error in the log, cause that this method returns void, it used to return true in previous versions, the reporter said that the entry is still made regardless though,,,

waddahadel:
I will open a new pr, I seemed to have messed up with a branch that belongs to @jeph864 in our fork, you can reject this and take a look at the new one @mjansenDatabay

PR #7501, opened 2024-05-10
Component: Make \ilIntroduceComponentArtifactDBUpdateSteps::step_4 mo…

mjansenDatabay:
…re verbose if there are problems


klees:
I'm not willing to deal with @rschenk0|s abusive tone and behaviour.

mjansenDatabay:
> I'm not willing to deal with @rschenk0|s abusive tone and behaviour.

But why did you close this PR with the given reason? Of course, it is based on the Mantis issue where the discussion took place, but IMO that is the only thing it has in common with the discussion you mentioned.

klees:
I started at the issue, the issue links here. Don't want to deal with that abusive behaviour at all and this PR is clearly connected.

PR #7502, opened 2024-05-10
Database/Setup: Add a hint to release "lock" state after failed datab…

mjansenDatabay:
…ase update step

If a database update step fails, the ILIAS setup
(to be concrete: the database update objectives) are in a "locked" state, and administrators often do not know how to "release" this "lock" again after the issue has been fixed manually or within a new Git commit. As long as we have no option/mechanism to automatically clean up this state, I suggest providing an SQL statement which can be manually executed.


klees:
I'm not willing to deal with @rschenk0|s abusive tone and behaviour.

mjansenDatabay:
> I'm not willing to deal with @rschenk0|s abusive tone and behaviour.

But why did you close this PR with the given reason? Of course, it is based on the Mantis issue where the discussion took place, but IMO that is the only thing it has in common with the discussion you mentioned.

klees:
I started at the issue, the issue links here. Don't want to deal with that abusive behaviour at all and this PR is clearly connected.

PR #7505, opened 2024-05-10
T&A: 41172, Fixed none working filtering in Test Participants with or without Answers

matheuszych:
Will be reviewed by @thojou

The issue was quiet straightforward. Wrong Session variable was accessed which did not exist, by default reverting to no filtering.
Since the error is also present in ILIAS 8 I tested this solution and it is also applicable.

I think this is a proper fix for the issue.


thojou:
Hey @matheuszych,

I have taken the PR and cherry-picked the changes into the trunk.

Thank you.

PR #7506, opened 2024-05-10
changed the order of some arguments:  nusoap.php

waddahadel:
there was a deprecated error about mandatory and optional params, changed the order to make it go away


jeph864:
Dear @waddahadel ,
Commit [69abac9](https://github.com/ILIAS-eLearning/ILIAS/pull/7506/commits/69abac90310f75d313dd82eba23e901f6a81c415) removes the necessary part, which assigns a user to a role. So it should not  be removed! The issue seems to be in the assignUser function which does not return anything. A check after calling the function (e.g. rbacreview's isAssigned function would fix this?)! Are all these 4 commits related?

kergomard:
Hi @waddahadel  and @jeph864   

What is the plan with this? Any chance to move this forward?

Thank you very much and best,
@kergomard

PR #7509, opened 2024-05-13
User/UDF: Don't provide NULL values for consumers

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41361

### DIFF
--- a/Services/User/classes/class.ilUserDefinedData.php
+++ b/Services/User/classes/class.ilUserDefinedData.php
@@ -84,9 +84,15 @@ class ilUserDefinedData
         return $this->user_data[$a_field] ?? '';
     }
 
-    public function getAll(): array // Missing array type.
+    /**
+     * @return array<string, string>
+     */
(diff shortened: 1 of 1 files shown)


Services/User/classes/class.ilUserDefinedData.php
```diff
@@ -84,9 +84,15 @@ public function get(string $a_field): string
...
 
-    public function getAll(): array // Missing array type.
+    /**
+     * @return array<string, string>
+     */
+    public function getAll(): array
     {
-        return $this->user_data;
+        return array_map(
+            static fn ($value): string => $value ?? '',
```
mjansenDatabay, line 93: Similar to `get` I suggest transforming the values when retrieved by a consumer, instead of casting them when `__read` is called.

kergomard:
Thank you very much for the PR @mjansenDatabay ! This sounds plausible and looks good to me. Merged and picked to trunk.

PR #7512, opened 2024-05-14
Trash: Reformat output of title and description

tbongers-cat:
Small fix for Mantis #41364, should probably be cherry-picked to other versions as well.

Edit: Added a change for the confirmation dialog as well.


alex40724:
Thanks for the PR!

PR #7515, opened 2024-05-15
Mantis #14367 - fix local user role assignments within org unit

sroenchen:
https://mantis.ilias.de/view.php?id=41367

Fixes php 8.0 compatibility issues

### DIFF
--- a/Modules/OrgUnit/classes/LocalUser/class.ilLocalUserGUI.php
+++ b/Modules/OrgUnit/classes/LocalUser/class.ilLocalUserGUI.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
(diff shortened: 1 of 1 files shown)


klees:
Hi @sroenchen,

just FYI: During the [PHP 8](https://docu.ilias.de/goto_docu_copa_10290.html) project we collectively decided to remove all super global access to the request from inline code in classes. Sadly, the Org-Units didn't make it that far and I took over maintenance only after ILIAS 8. If you ever stumble upon something like this again, maybe consider to make code fit the PHP 8 project requirements as well. Should only be a little extra work, mostly.

Thanks!

klees:
This seems to be fine in >= release_9 already, so no need to pick cherries.

klees:
And also: Thanks for looking into this =)

PR #7518, opened 2024-05-15
Changes implementation of getCurrentMode in class ilPluginLP

chlulei:
If a u_mode entry exists in ut_lp_settings it is used instead of LP_MODE_PLUGIN.

This allows plugins to set their own u_mode.

### DIFF
--- a/Services/Component/classes/class.ilPluginLP.php
+++ b/Services/Component/classes/class.ilPluginLP.php
@@ -74,8 +74,8 @@ class ilPluginLP extends ilObjectLP
     public function getCurrentMode(): int
     {
         $mode = ilLPObjSettings::_lookupDBMode($this->obj_id);
-        if (!is_null($mode)) {
-            return $mode;
+        if ($mode === ilLPObjSettings::LP_MODE_UNDEFINED) {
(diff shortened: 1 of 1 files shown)


Services/Component/classes/class.ilPluginLP.php
```diff
@@ -73,6 +73,11 @@ public function getValidModes(): array
 
     public function getCurrentMode(): int
     {
+        $mode = ilLPObjSettings::_lookupDBMode($this->obj_id);
```
mjansenDatabay: @smeyer-ilias This looks "dangerous" to me if introduced in a stable release. Maybe a bigger audience should be informed about this change.

Is there a documentation of how plugin components can define custom learning progress modes (`u_mode`)? Are plugin specific modes validated by the core? What about possible conflicts with core modes?

smeyer-ilias:
@mjansenDatabay thanks for your feedback. We will create a mantis issue which describes the issues with the current implementation tomorrow. We do not plan to support different LP modes for plugins with this PR.

kergomard:
Hi @smeyer-ilias 

What is the plan with this? It looks very much like a PR that should be closed.

Thank you very much and best,
@kergomard

PR #7520, opened 2024-05-16
New Authorities for Administration Component

fneumann:
See appliation for authorities by e-mail. I recognized that the assignment for issues and security reports is only possible to one person. So I added me there.


matthiaskunkel:
Merged information about changed authorities for component Administration.

PR #7521, opened 2024-05-16
UI: fix backwards compatibility for Data/Table backport

klees:
Hi everyone,

the method `isClientSideValueOk` changing from protected to public made plugins break that implemented their own inputs. When switching back the method to `public` unit tests fail, indicating that we cross class hierarchies when calling the method. This was due to the fact that we indeed had two different methods `isClientSideValueOk`, one from the original input hierarchy, one from a trait for groups. These are reconciled now and thus the method can be protected again.

Kind regards!

### DIFF
--- a/src/UI/Implementation/Component/Input/Field/Group.php
+++ b/src/UI/Implementation/Component/Input/Field/Group.php
@@ -154,4 +154,10 @@ class Group extends FormInput implements C\Input\Field\Group, GroupInternal
     {
         return $this->data_factory;
     }
+
+    /** ATTENTION: @see GroupInternals::_isClientSideValueOk() */
+    protected function isClientSideValueOk($value): bool
+    {
+        return $this->_isClientSideValueOk($value);
+    }
 }
--- a/src/UI/Implementation/Component/Input/Group.php
+++ b/src/UI/Implementation/Component/Input/Group.php
@@ -121,9 +121,11 @@ trait Group
     }
 
     /**
-     * @inheritdoc
+     * ATTENTION: This is not the same as @see Input::isClientSideValueOk(),
+     * even if it had the same name. These are different symbols, as this trait
+     * is not in the hierarchy that defines the original isClientSideValueOk.
      */
-    public function isClientSideValueOk($value): bool
+    protected function _isClientSideValueOk($value): bool
     {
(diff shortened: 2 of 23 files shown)


klees:
I checked this commit with release_9, unit tests are fine. So I think we should pick this to ILIAS 9 and trunk as well.

src/UI/Implementation/Component/Input/Field/Group.php
```diff
@@ -154,4 +154,10 @@ protected function getDataFactory(): DataFactory
     {
         return $this->data_factory;
     }
+
+    /** ATTENTION: see trait Group::_isClientSideValueOk! */
```
thibsy: Could we make this an annotation?
```suggestion
    /** ATTENTION: @see GroupInternals::_isClientSideValueOk() */
```

src/UI/Implementation/Component/Input/ViewControl/Group.php
```diff
@@ -98,4 +98,10 @@ protected function getDataFactory(): DataFactory
     {
         return $this->data_factory;
     }
+
+    /** ATTENTION: see trait Group::_isClientSideValueOk! */
```
thibsy: Could we make this an annotation?
```suggestion
    /** ATTENTION: @see GroupInternals::_isClientSideValueOk() */
```

src/UI/Implementation/Component/Input/Group.php
```diff
@@ -121,9 +121,11 @@ protected function nameInputs(NameSource $source, string $parent_name): array
     }
 
     /**
-     * @inheritdoc
+     * ATTENTION: This is not the same as isClientSideValueOk from the base class,
```
thibsy: Could we make this an annotation?
```suggestion
     * ATTENTION: This is not the same as @see Input::isClientSideValueOk() from the base class,
```

thibsy (approved):
Thx a lot @klees for implementing this! One minor adjustment:

- Could we cross-reference the `isClientSideValueOk()` documentation(s) using PHP annotations?

I have left some inline suggestions you can commit. Afterwards I'll merge and pick for the mentioned versions.

Kind regards,
@thibsy

PR #7522, opened 2024-05-16
Preload GUI Data for Dashboard

iszmais:
https://mantis.ilias.de/view.php?id=40127

### DIFF
--- a/Services/Dashboard/Block/classes/class.ilDashboardBlockGUI.php
+++ b/Services/Dashboard/Block/classes/class.ilDashboardBlockGUI.php
@@ -97,6 +97,14 @@ abstract class ilDashboardBlockGUI extends ilBlockGUI implements ilDesktopItemHa
     {
         $data = $this->loadData();
         $groupedCards = [];
+        $obj_ids = [];
+        foreach ($data as $group) {
+            foreach ($group as $datum) {
+                $obj_ids[] = $datum->getObjId();
+            }
+        }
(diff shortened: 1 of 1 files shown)


Services/Dashboard/Block/classes/class.ilDashboardBlockGUI.php
```diff
@@ -97,6 +97,14 @@ protected function getListItemGroups(): array
     {
         $data = $this->loadData();
         $groupedCards = [];
+        $obj_ids = [];
+        foreach ($data as $title => $group) {
```
lscharmer:
```suggestion
        foreach ($data as $group) {
```

lscharmer: Why do you use a new `foreach` loop when you can use the existing foreach loop below starting in line `108`?

lscharmer (changes requested):

Services/Dashboard/Block/classes/class.ilDashboardBlockGUI.php
iszmais: `preloadListGUIData` has to be called before `getListItemForDataDTO`.
I could call it individually before each DTO call, but that would cause a heavier workload trough multiple SQL calls than the duplicated `foreach`

lscharmer (approved):

PR #7524, opened 2024-05-16
UI/Input: align/unify HTML structure

nhaagen:
This restructures the HTML of Forms, i.e. their input-elements.
Central to this PR is tpl.context_form.html:
```
<fieldset class="c-input" 
	data-il-ui-type="{TYPE}" 
	data-il-ui-name="{NAME}" 
	{DISABLED}>

	<!-- BEGIN error -->
	<div class="c-input__error-msg" aria-describedby="{ERROR_FOR_ID}" role="alert">{ERROR}</div>
	<!-- END error -->

	<legend>
		<label tabindex="0" <!-- BEGIN for -->for="{ID}" <!-- END for -->>{LABEL}<!-- BEGIN required --><span class="asterisk">*</span><!-- END required --></label>
	</legend>

	<div class="c-input__field">
		{INPUT}
	</div>

	<!-- BEGIN byline -->
	<div class="c-input__help-byline">{BYLINE}</div>
	<!-- END byline -->

</fieldset>
```

All Fields are wrapped into this structure, where dependant fields (again wrapped in form context) will show under `<div class="c-input__field">`. 
A little gimmick is the pure CSS implementation of Optional/Switchable Groups.

Currently, the CSS ist missing/broken, but @catenglaender is informed and on it, so I have little doubt the visual impacts will be quite small and ultimately for the better ;)


Amstutz:
Hi @nhaagen

Feedback to the state of the draft; We like the new DOM structure. This could also simplify/resolve a lot of a11y we have left to deal with for forms (mostly relating to the grouping and labelling of things). Further we agree that the inputs should be aligned and a simplified DOM structure has many benefits, such as a simplification of the UI Filter. We are looking forward to performing a complete review, as soon as the CSS is shipped.

Important note: We expect the UI of the forms to deviate from the legacy Forms in the future. To everybody involved, please try to migrate as many forms as possible to keep the consistent look and feel of ILIAS.

thx 
@Amstutz , @thibsy and @klees

kergomard (approved):
I know this is still a draft, but I think that you don't need to wait for any further approval from our side. Your changes make the test cleaner and they then just have to run without failing ;-).

On a side note: Personally, I don't like the "function notation" for `require_once`, but you are not the one who introduced it into the class.

Thanks and best,
@kergomard

Amstutz:
Feel free to re-assign me and @thibsy as soon as the CSS is complete and this is ready for a final review.

dsstrassner:
I added me as an assignee, that I could track this PR for the T&A TechSquad statistics. Please remove me if you think this is inappropriate.

klees (approved):

klees:
Hi @thibsy, hi @Amstutz,

this is good to go. Please take this over!

Kind regards!

Amstutz:
Hi @nhaagen

Thx a lot for this PR. This seems like an excellent and necessary change for a much cleaner structure of our inputs. Highly appreciated.

We have the following questions:
- [ ] Type and Name: You introduce data-il-ui-type data-il-ui-name, why are they needed? Is it SCSS usage, JS usage? Some A11y usage?
- [x] Fieldsets: You use fieldset in the context_form.html, which, we believe, is good. However, you use them again for templates of specific fields such as multiselect.html. There they are used without legend. Why is that? Could they be removed?
- [ ] JavaScript files: there are a lot of changes inside the JavaScript code of inputs. They look mostly look like code-style changes done by ESLint and the actual changes can be narrowed down to the CSS-classes changing. Is that correct? If there is any other notable change inside the JavaScript file, please let us know where.
- [ ] SCSS: We do see many changes, especially usages of the BEMIT class names, thx. the Did @catenglaender checkout the SCSS code? If not please let @catenglaender review this as well.
- [ ] Section: As far as we understand, the context_form.html is also used for sections, correct? If so, is the position of the byline still correct in this special case? Should it not be just bellow the legend in the special case of form sections?

- [ ] Checkbox and Radio: the tpl.checkbox.html and tpl.radio.html seem to be replaced by tpl.optionalgroup_label.html and tpl.switchable_group.html files. Is this correct? If so, please remove them.
- [ ] Template files: there are some new and old template files which instead of one top-level HTML element now consist of two or more. This will lead to problems down the road, when incorporating JavaScript hydration, so we try to avoid components with more than one top-level element. Affected are textarea, optional-group, and switchable-group. Is there a way to avoid this?

Please Change:
- [x] CSS.map: Remove the css.map (see comment in https://github.com/ILIAS-eLearning/ILIAS/pull/7795#issuecomment-2244910383 for the cause of this). 
- [x] Conflicts: There are several, please resolve them. 

We are looking forward to merging those changes soon, kindly
@amstutz and @thibsy

nhaagen:
Hi @Amstutz and @thibsy
thank you for looking so thoroughly through this.

> * [ ]  Type and Name: You introduce data-il-ui-type data-il-ui-name, why are they needed? Is it SCSS usage, JS usage? Some A11y usage?

This PR eventually originated from https://github.com/ILIAS-eLearning/ILIAS/pull/7137, where Inputs are to be detected client side. I somewhat failed with the original HTML-Structure due to two problems: the start of a field could not easily be determined, especially with groups involved, and I had to guess its type according to used tags, e.g. for optional groups and single selects.
From experience with the columns of data-table, I find it quite valuable to unmistakeable state the distinctive type of the input;
the name gives a reference to the "actual" input (and thus the value) as well as the parameter in the POST. 
So, mainly this is for JS usage, but I'm very positive on CSS opportunities as well.

> * [ ]  Fieldsets: You use fieldset in the context_form.html, which, we believe, is good. However, you use them again for templates of specific fields such as multiselect.html. There they are used without legend. Why is that? Could they be removed?

I removed the fieldset around multi-select and substituted it for div in radio.
However, we had to remove the legend from context_form as well, since it behaves quite oddly with styling.

> * [ ]  JavaScript files: there are a lot of changes inside the JavaScript code of inputs. They look mostly look like code-style changes done by ESLint and the actual changes can be narrowed down to the CSS-classes changing. Is that correct? If there is any other notable change inside the JavaScript file, please let us know where.

You're right, it's just the css-classes that changed, and so I had to re-compile.

> * [ ]  SCSS: We do see many changes, especially usages of the BEMIT class names, thx. the Did @catenglaender checkout the SCSS code? If not please let @catenglaender review this as well.

Those changes are from @catenglaender.

> * [ ]  Section: As far as we understand, the context_form.html is also used for sections, correct? If so, is the position of the byline still correct in this special case? Should it not be just bellow the legend in the special case of form sections?

You're probably right. We'll see what we can achieve by CSS only, I'd really like to use the same template overall.
 old: Screenshot from 2024-08-15 08-47-16

new: Screenshot from 2024-08-15 08-47-48


> * [ ]  Checkbox and Radio: the tpl.checkbox.html and tpl.radio.html seem to be replaced by tpl.optionalgroup_label.html and tpl.switchable_group.html files. Is this correct? If so, please remove them.

No, they are still used, i.e. they are now more dedicated; mixing optionalgroup/checkbox  and switchablegroup/radio lead to quite a bit of confusion of labels and values.

> * [ ]  Template files: there are some new and old template files which instead of one top-level HTML element now consist of two or more. This will lead to problems down the road, when incorporating JavaScript hydration, so we try to avoid components with more than one top-level element. Affected are textarea, optional-group, and switchable-group. Is there a way to avoid this?

The components will ultimately be wrapped in rendering, so there is one node only. Maybe I do not understand hydration the full way down, is this an issue? If so, we could probably just add a div around them.

> We are looking forward to merging those changes soon, kindly @Amstutz and @thibsy

Me too ;)

Amstutz:
Hi @nhaagen,

Thanks a lot for answering our questions. A few final things, we are looking forward to merge this.

- [x] `Renderer::wrapInFormContext()`: you are using [`trim()` on the `$dependant_group_html`](https://github.com/ILIAS-eLearning/ILIAS/pull/7524/files#diff-ef03ec89ebf29bde4d4f7277b852dd624870adbb45d3d1f08a36c630a4cab2edR211) inside this method. Could you explain why we need to do this? It seems like the issue may be caused somewhere else.

- [x] `data-il-ui-type` attribute I: you derive the value for this attribute from the components canonical name and supply is as `PascalCase`. Could you transform this value to `kebab-case` instead, to match most of the HTML elements other attribute values? This is however optional, since it does not actually affect anything (yet).

- [x] `SwitchableGroup`/`OptionalGroup` HTML: the HTML of these inputs is invalid at the moment, due to a `<label>` tag nested inside another one, and the `SwitchableGroup` label points to an ID which does not exist. This happens because you provide the input html as the `$label` argument when calling `Renderer::wrapInFormContext()`. I suggest to revert this method signature to the previous state, because the `$default_renderer` is still needed (`FilterContextRenderer`), the `$label` is only necessary due to these two inputs (which can't be provided this way), and the `$wrap` argument is never used. Maybe these two inputs can be handled as a special case inside the `wrapInFormContext()` method using `instanceof` checks. I am open to other suggestions though.
- [x] `data-il-ui-type` attribute II: do you think this attribute could be used for other components in the future as well? If not, and the use-case is specific to inputs, please rename the attribute to `data-il-ui-input`. If so, please rename the attribute to `data-il-ui-component` rather than "type".
- [x] If the attribute could also be used by other components, please move the generation of its value to the `AbstractComponentRenderer`.
- [x] `data-il-ui-name` attribute: please rename this attribute to `data-il-ui-input-name`, since it appears to be specific to inputs only.
- [x] Conflicts: please rebase this onto the latest trunk.

Kind regards,
@Amstutz and @thibsy

thibsy (approved):
Thx @nhaagen for implementing the requested changes.

I have found one small issue, where the `for` attribute of the `Field\Section` should not point to an ID. However, I will fix this myself after merging, since the rest LGTM now and there's really only one line change required.

Again, special thanks for improving the accessibility of our inputs!

Kind regards,
@thibsy

PR #7526, opened 2024-05-16
[FIX] UI: revert input type hints for backwards compatibility.

thibsy:
Hi folks,

The data table backport has introduced several breaking changes in form of type hints which could lead to compile- or fatal errors when plugins implement custom inputs extending from concrete inputs of the UI framework. This commit reverts these changes back to the previous state as good as possible, but there is still one breaking change in `Input::withRequired()`, which gained an additional parameter that cannot not be reverted. However, this should be less problematic because additional optional parameters are possible, so adding this parameter to the custom implementation would still allow it to be compatible for versions above and below 8.11 - nonetheless, it requires action. Maybe we should communicate this to our dev-lists as well.

You can perform a `git diff v8.10 -- src/UI/Component/Input/Field src/UI/Implementation/Component/Input/Field` to check that actually none of the type-hints change in a manner that would break now.

Kind regards,
@thibsy

### DIFF
--- a/src/UI/Component/Input/Container/Form/FormInput.php
+++ b/src/UI/Component/Input/Container/Form/FormInput.php
@@ -66,7 +66,7 @@ interface FormInput extends Input, JavaScriptBindable, OnUpdateable
      * (see getConstraintForRequirement() on Input/Field implementations).
      * A custom constraint SHOULD be explained in the byline of the input.
      */
-    public function withRequired(bool $is_required, ?Constraint $requirement_constraint = null): self;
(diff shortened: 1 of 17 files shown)


klees (approved):

chfsx:
I can confirm that with this change the affected plugins H5P and OpenCast are working again. I have not been able to test any other plugins, as it is unclear which others could still be affected.
Thank you for merging! since many institutions are already reporting problems since the upgrade to ILIAS 8.11, you can refer them to a git pull. nevertheless, we should consider releasing ILIAS 8.12 soon, because probably many check out tags and not branches. what do you think?

PR #7527, opened 2024-05-17
Data: fix Meta/Tag::toHtml (#40110)

klees:
https://mantis.ilias.de/view.php?id=40110

### DIFF
--- a/src/Data/Meta/Html/OpenGraph/Tag.php
+++ b/src/Data/Meta/Html/OpenGraph/Tag.php
@@ -33,7 +33,7 @@ abstract class Tag extends HTMLTag
 
     public function toHtml(): string
     {
-        return "<meta property=\"$this->property_name\" content=\"{$this->getValue()}\" />";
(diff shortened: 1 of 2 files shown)


mjansenDatabay (commented):
Hi @klees ,

thanks for the PR. This LGTM, but one question: May it be problematic as well if the value or property name contains curly braces? Might this be problematic for our current template engine if these HTML fragments are set into the page template? Since we do not pass elaborated objects to the caller but complete HTML meta-tags, we should at least discuss if it is the concern of the `toHtml` method to transform these strings.

Best regards,
Michael

klees:
Hi @mjansenDatabay,

I messed around a little with the curly braces and could not cause any breakage, but also my `{FOO}` in title and description never appeared in the output. So I guess this is fine.

I would not move that problem of the templating engine to the data type, though.

Kind regards!

mjansenDatabay (approved):

thibsy (approved):
LGTM, thx @klees!

PR #7531, opened 2024-05-17
Course Refactor: Move external Course Code out of Certificate

fhelfer:
Hi @smeyer-ilias,

Best, @fhelfer

### DIFF
--- a/components/ILIAS/Certificate/classes/Cron/class.ilCertificateTypeClassMap.php
+++ b/components/ILIAS/Certificate/classes/Cron/class.ilCertificateTypeClassMap.php
@@ -18,6 +18,7 @@
 
 declare(strict_types=1);
 
+use ILIAS\Course\Certificate\CoursePlaceholderValues;
 use ILIAS\StudyProgramme\Certificate\ilStudyProgrammePlaceholderValues;
 use ILIAS\Test\Certificate\TestPlaceholderValues;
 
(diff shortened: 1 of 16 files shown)


mjansenDatabay:

mjansenDatabay (approved):

mjansenDatabay:
The certificate changes are fine for me. @smeyer-ilias Please have a look at the course related changes.

mjansenDatabay:
Hi @smeyer-ilias ,

do you have any objections? Similar to other PRs (e.g. `Test` and `StudyProgramme`, both already approved and integrated into `trunk`) we would like to move the component-specific classes to their origin.

We would like to merge this in a timely manner, means: Within the next 10 days.

Best regards,
Michael

components/ILIAS/Course/classes/Objectives/Setup/class.ilMigrateCourseCertificateProviderDBUpdateSteps.php
```diff
@@ -0,0 +1,40 @@
...
+
+    public function step_1(): void
+    {
+        $this->db->update(
+            'il_cert_cron_queue',
+            ['adapter_class' => [ilDBConstants::T_TEXT, CoursePlaceholderValues::class]],
+            ['adapter_class' => [ilDBConstants::T_TEXT, 'ilCoursePlaceholderValues']]
+        );
+    }
+}
```
alex40724: This is the wrong place, since the Course should not hack into certificate tables. I guess the class names are used as IDs or something? This fact should also be hidden from the consumers. Please move this part to the certificate component.

alex40724 (changes requested):
Hi @fhelfer 

- [x] Remove the access to the certificate table from the course component. This is internal knowledge the course component should not need to have. Future refactorings of this table should not affect code in the course component.

The rest is fine and can be merged. Thanks

kindly,
@alex40724

mjansenDatabay (approved):

mjansenDatabay:
Thx @fhelfer and @alex40724.

PR #7532, opened 2024-05-17
Remove booking pool from resources on delete

iszmais:
https://mantis.ilias.de/view.php?id=41382


alex40724:
Thanks for the PR. This is fixed a little bit differently now.

PR #7533, opened 2024-05-17
T&A Refactor: Move external Test Code out of Certificate

fhelfer:
Hi @kergomard, @mbecker-databay, @dsstrassner 

Best, @fhelfer

### DIFF
--- a/components/ILIAS/Certificate/classes/Cron/class.ilCertificateTypeClassMap.php
+++ b/components/ILIAS/Certificate/classes/Cron/class.ilCertificateTypeClassMap.php
@@ -18,6 +18,8 @@
 
 declare(strict_types=1);
 
+use ILIAS\Test\Certificate\ilTestPlaceholderValues;
+
 /**
  * @author  Niels Theen <ntheen@databay.de>
  */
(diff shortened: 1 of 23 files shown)


kergomard:
Thank you very much @fhelfer for this PR!

I very much believe this to be the right way to go. I will move the files to the right place after merge, but I would kindly ask you to change a few things:
- Please remove all `use`-statements for classes in the global namespace.
- Please remove any "class." from file-names and any "il" from class-names unless completely necessary.
- Please rename variables in snakeCase (at least in the cases you touch them).

I would then take care of this asap.

Thanks again and best,
@kergomard

mjansenDatabay:

kergomard:
Thanks @mjansenDatabay for the clarification!

fhelfer:
Hello @kergomard,

> * Please remove all `use`-statements for classes in the global namespace.

What exactly do you mean by this, 
[here](https://github.com/ILIAS-eLearning/ILIAS/pull/7533/files#diff-a94d7a8b2a1cabef5a072d11bde3e2dbfb5457ba7aa2d6305f544c23df9cc836R23-R34) these cannot be removed, as it will result in class not found error
Everywhere I introduced namespace, I need to use statements for everything outside the namespace

Best, @fhelfer

kergomard:
Hi @fhelfer 

Yes, exactly these ;-) . Please simply add a backslash to wherever the class is used. So `ilCertificateFormRepository` -> `\ilCertificateFormRepository`.

Does the trick and we then know exactly that it is a class in the global space.

Thanks and best,
@kergomard

kergomard:

I merge this now and will take care, if I missed something.

Best,
@kergomard

PR #7534, opened 2024-05-17
StudyProgramme Refactor: Move external StudyProgramme Code out of Certificate

fhelfer:
Hi @klees,

Best, @fhelfer

### DIFF
--- a/components/ILIAS/Certificate/classes/Cron/class.ilCertificateTypeClassMap.php
+++ b/components/ILIAS/Certificate/classes/Cron/class.ilCertificateTypeClassMap.php
@@ -18,6 +18,7 @@
 
 declare(strict_types=1);
 
+use ILIAS\StudyProgramme\Certificate\ilStudyProgrammePlaceholderValues;
 use ILIAS\Test\Certificate\TestPlaceholderValues;
 
 /**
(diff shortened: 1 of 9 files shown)


klees (changes requested):
Hi @fhelfer,

I appreciate the intended move. Its better if other components know the Certificate component instead of the Certificate component knowing all other components.

I find it a little hard to really understand what the exact change is here. This seems to contain a lot of unrelated changes (e.g. introduction of `use` statements, exchange of `"` to `'`) that seem to reflect personal preferences, possibly introduced by some automatism of an IDE.

Could you please strip this down to the changes that are actually required to make the intended change?

Kind regards!

mjansenDatabay:

fhelfer:
Hi @klees,

removed all changes except introduction of a new namespace for Certificates in StudyProgrammes. Please let me know if you also want this to be removed
In addition, based on the hint @mjansenDatabay gave, I added a new UpdateStep changing the adapter class in il_cert_con_queue to according to the new namespace for PlacholderVlaue class.

Best @fhelfer

klees (approved):

klees:
Thx @fhelfer! Still plausible =)

PR #7535, opened 2024-05-17
Update Dependencies after Release v9.1

fwolf-ilias:
Dear @ILIAS-eLearning/technical-board,
This PR updates all dependencies after release v9.1.

**Composer Notices:**
```
Package jasig/phpcas is abandoned, you should avoid using it. Use apereo/phpcas instead.
Package technosophos/libris is abandoned, you should avoid using it. No replacement was suggested
```

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -213,16 +213,16 @@
         },
         {
             "name": "composer/composer",
-            "version": "2.7.4",
+            "version": "2.7.6",
             "source": {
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @fwolf-ilias ! An update of the CS-Fixer gave us two new transient dependencies, things look and work ok, though. We could now run the fixer in parallel.

PR #7537, opened 2024-05-20
[41443] Allow using RSA key when saving LTI Content Selection

FElliger:
Previously, the LTI forms allowed storing a static public key, but the code for processing the content selection response only looked for the keyset URL. With this change, both ways are supported, depending on which key type is stored.

Issue link: https://mantis.ilias.de/view.php?id=41443

### DIFF
--- a/Modules/LTIConsumer/classes/class.ilObjLTIConsumerGUI.php
+++ b/Modules/LTIConsumer/classes/class.ilObjLTIConsumerGUI.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 /**
  * Class ilObjLTIConsumerGUI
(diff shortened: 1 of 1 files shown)


chfsx:
HI! Is this PR related to a bug report or is this - as I assume - rather an additional feature? In both cases, corresponding updates would be required:
- If this fixes a bug, please link the corresponding ticket from https://mantis.ilias.de (if it doesn't exist, please create it)
- if this is a feature, this should go through the regular feature process: [How To Suggest A New Feature](https://docu.ilias.de/goto_docu_wiki_wpage_788_1357.html)

thank you very much!

chfsx:
HI! Is this PR related to a bug report or is this - as I assume - rather an additional feature? In both cases, corresponding updates would be required:
- If this fixes a bug, please link the corresponding ticket from https://mantis.ilias.de (if it doesn't exist, please create it)
- if this is a feature, this should go through the regular feature process: [How To Suggest A New Feature](https://docu.ilias.de/goto_docu_wiki_wpage_788_1357.html)

thank you very much!

FElliger:
I have added a link to the issue: https://mantis.ilias.de/view.php?id=41443

It's a bug form my perspective as well, as ILIAS is offering the option of pasting the RSA key, but fails to use it.

mjansenDatabay:
Hi @ZallaxDev and @Saaweel , I added your accounts to the list of assignees. Please keep in mind that this fix needs to be picked to all maintained branches (`release_9`, `release_10` and `trunk`) if this PR is approved/merged.

PR #7538, opened 2024-05-21
TA 41390, hide user name in anon mode

nhaagen:
https://mantis.ilias.de/view.php?id=41390


kergomard:

I've two little things:
* Could you please add a space after the `if` (e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7538/files#diff-a25e8bc63216ea2d9c6ba4e644b978a00881a7e73ef74a9bb60e4502f9d2c688R939))
* As you need to change this: Could you maybe create a function for this? We have the same block quite a few times and I think we could improve readability.

Thank you very much and best,
@kergomard

kergomard:
Thank you very much for the quick update! Merged and picked to trunk.

PR #7539, opened 2024-05-21
WebDAV: Handover Authorities to Fabian Schmid

kergomard:
Hi all

After I announced that I would actually propose to abandon `WebDAV` @chfsx has expressed interest to take over the authorities for `WebDAV` as it combines well with the `File`-Component he already takes care of. We thus propose to hand over the responsibility for `WebDAV` to @chfsx .

Best,
@kergomard

### DIFF
--- a/docs/development/maintenance.md
+++ b/docs/development/maintenance.md
@@ -1636,14 +1636,14 @@ of ILIAS. The file contains the following fields:
 [//]: # (BEGIN WebDAV)
 
 * **WebDAV**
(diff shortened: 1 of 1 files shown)


mbecker-databay:
Just a small remark: I believe it should be "sign off".

kergomard:
Thank you @mbecker-databay ! Fixed.

matthiaskunkel:
Thanks to @kergomard for taking care of WebDAV component for so many years. And thanks in advance to @chfsx for taking care of the responsibility of this component from now on.

PR #7541, opened 2024-05-21
ErrorHandling: Check for `tpl` existence

mjansenDatabay:
This PR fixes a wrong dependency check in the error handler. The current check of `tpl` causes an error, because there is no method `tpl` in the DI container.


iszmais (approved):
IMO best practice would be to check on the method "mainTemplate" existence, but this solution looks sufficent, too.

PR #7548, opened 2024-05-22
[FIX] UI: revert breaking change in `tpl.standardpage.html`

thibsy:
Hi folks,

@iszmais has brought another [breaking change](https://github.com/ILIAS-eLearning/ILIAS/commit/25139f0fe323e1ea237b9ea6d2216531a7427ce3) to my attention, which has been caused by the data table backport. This makes skins cause JavaScript errors if they are manipulating the `tpl.standardpage.tpl` file of the page component.

I reverted this commit and cherry-picked @nhaagen solution which he proposed in #7458 already.

Kind regards,
@thibsy

### DIFF
--- a/src/UI/templates/default/Layout/tpl.standardpage.html
+++ b/src/UI/templates/default/Layout/tpl.standardpage.html
@@ -15,10 +15,6 @@
 	<title>{VIEW_TITLE}: {SHORT_TITLE}</title>
 	<link rel="icon" href="{FAVICON_PATH}" type="image/x-icon">
 
-	<!-- This is the safest place for now to initialize our global namespace which is -->
-	<!-- used by several scripts included on the way to expose their business logic.  -->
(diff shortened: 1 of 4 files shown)


klees (approved):

klees:
Hi everyone,

I just want to point out that this or similar problems could easily arise due to any change in the rendering of some UI component, e.g. to fix some accessibility issue. The surface we offer to creators of skins is extremely large, every template, css-class name, specific usage of html elements, maybe even internal scss structure, etc. could be subject of some change in some skin and hence cause a bc break when modified.

To me this points out again what we have discussed in JF wrt to the backport of the Data Table: Every change to the system, even bugfixes, even if the change looks benign, could change some behaviour that someone expects to exist. Hence, every change should be subject to the consideration of risks and rewards, may it be implicit or explicit. Something might always break.

Kind regards!

PR #7549, opened 2024-05-22
AccessControl: Fix PHP type error for starting point role id

mjansenDatabay:
AccessControl: Fix PHP type error for starting point role id

Mantis Issue: https://mantis.ilias.de/view.php?id=41408


kergomard:
Thank you very much for the PR. I'm not 100% sure if this is really needed in 9 and above, but I adapted the improvements anyways.

PR #7553, opened 2024-05-22
T&A: Mantisticket 41383, Release 9

malaja11:
@thojou and @mbecker-databay will take care of this

https://mantis.ilias.de/view.php?id=41383

The problem was caused by the method prepareOutput in class.ilObjTestGUI. Within the method, a refID of an ilObjTest was used as a refID of an ilContainer, which caused the exception. 

In consultation with @fhelfer who added this method with https://github.com/ILIAS-eLearning/ILIAS/pull/6333 to fix https://mantis.ilias.de/view.php?id=37993 we noticed that the code now seems unnecessary. 

The same method was also implemented in ilObjCourseGUI and it seems as if it no longer provides any functionality. Instead, the functionality was outsourced into ilMemberViewSettings.

We removed both methods which solved the problem.

### DIFF
--- a/Modules/Course/classes/class.ilObjCourseGUI.php
+++ b/Modules/Course/classes/class.ilObjCourseGUI.php
@@ -2805,21 +2805,6 @@ class ilObjCourseGUI extends ilContainerGUI
     }
 
     /**
-     * @inheritDoc
-     */
-    public function prepareOutput(bool $show_subobjects = true): bool
-    {
-        if (!$this->getCreationMode()) {
-            $settings = ilMemberViewSettings::getInstance();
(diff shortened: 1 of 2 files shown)


thojou:
The changes to the `ilObjTestGUI` appear to be correct. However, a different CA needs to evaluate this for the `ilObjCourseGUI`. 

I also believe that this code is no longer necessary, as the `ilMemberViewSetting` is implemented to deactivate it upon leaving the associated course, so that "switching between courses in MemberView" is no longer possible, which is why this change was introduced in 2009.

alex40724 (approved):
Changes are ok for the course you may merge this in 9/trunk.

thojou:
Hey @malaja11,

I have taken the PR and migrated the changes into the trunk.

Thank you!

PR #7554, opened 2024-05-23
41413: Fix trying to assign string value to int|null property

mBeym:
Fixes an error that occured when trying to import a Course that contained a File object.

The xml of the file object saves the attribute "usr_id" like this: 
```php
"il_" . IL_INST_ID . "_usr_" . $version["user_id"]
//Example: il_0_usr_6 => actual user-id "6"
```

the cast I added will always turn this string into **0**.

I'm uncertain if this is the correct solution, 
considering it's been done like this on other lines in the file, I would assume this is how it should be.
I could not find issues in the imported Objects with this cast added.

Mantis Issue: https://mantis.ilias.de/view.php?id=41413

If accepted this change should be picked into other branches >= release_8.


chfsx:
Thanks a lot, I'll cherry-pick this up to trunk

PR #7555, opened 2024-05-23
41398: Fix mail template loaded from wrong directory

mBeym:
Fixes mail template loaded from wrong (old) path

Mantis Ticket: https://mantis.ilias.de/view.php?id=41398

if accepted these changes should be picked into branches >= release_9


mjansenDatabay (approved):
Thx @mBeym

PR #7556, opened 2024-05-23
Oblige maintainers to demand changes in form of PRs

iszmais:
This PRs propose to enable Maintainers (Coordinators/Authorities) to demand, and self-control, any contribution to their component to be in form of a Pull Request, except security related issues.

This explicitly does **not** create a requirement for contribution via PRs itsself (since that is already the case via the "Rules for Contributors", wich requires PRs, and the "Who is a contributor?" wich includes all individuals involved in the developement process, including Maintainers (Coordinators/Authorities) them self), but strengthens the authority of Maintainers (Coordinators/Authorities) to fulfill their responsibility.

### DIFF
--- a/docs/development/maintenance.md
+++ b/docs/development/maintenance.md
@@ -43,7 +43,8 @@ Responsibilities of a component maintainer:
   by the [Issue-Tracker](https://mantis.ilias.de).
 - Component maintainers are responsible for Pull Requests to their component and get assigned related Pull Requests
   by the Technical Board according to the [Rules for Maintainers and Coordinators assigned to PRs[(Rules for Maintainers and Coordinators assigned to PRs)
-
+- Component maintainer are obliged to demand any contribution to their component in form of a PR and to provide
+  their own contributions in form of PRs, too. Fixes regarding the security of ILIAS are excluded from that rule.
 
 ## Becoming a Maintainer
 


iszmais:
Dear @ILIAS-eLearning/technical-board ,

I flagged that PR since I'm unsure about the current state of the targeted aspect (and if it's really needed). Therefore, I present you my initial conflict:
- It is generally assumed that Pull requests in ILIAS are not mandatory for all instances but just for contributions from external individuals.
- This IMO violates the rules set, withing the `contribution.md` which obliges every contributor to present their changes via a Pull Request and further defines every person involved in the ILIAS development process as a contributor.

In conclusion, my humble assumption would be that every Maintainer (Coordinator/Authority) is already obliged to contribute only in form from PRs.

can you verify or deny that conclusion based on the given documentation?

Greetings,
@iszmais 

PS: If that is the case, this actually even might need to loosen this restriction (e.g. for security duties) via some exceptions

chfsx:
Thank you for your suggestion! We have put the topic on the agenda for today's technical board meeting. For the time being, we would have to close the PR, because changing the documentation alone would unfortunately not be enough. We understand the reasoning behind the PR and want to continue the discussion, but we need to take it to a broader level. The commit rights in ILIAS have historically always been direct, i.e. without prior PRs. This certainly has its origins in earlier version control systems before we switched to Git a few years ago. For example, SVN was used before that, which did not offer anything like pull requests in this form. and even before that - even before my time - I think it was still CVS.

the proposed change with regard to commit/push authorisation must be accompanied by an adjustment of various processes and should also be accepted more or less unanimously among the core developers. this year we have already created further possibilities for people to contribute to the core with the change to authorities. too many changes in a short time can also be a challenge, which is why we cannot/would not simply introduce your proposal, at least for the moment. but we are taking the issue on board.

in terms of security, however, the suggestion could also be counterproductive. we also need to discuss this together with the security group and on a larger scale. because if, for example, security fixes are always pushed directly, but all other changes are pushed via PR, security fixes can also be filtered very easily. 

> can you verify or deny that conclusion based on the given documentation?

to your question: we can confirm that currently the people with the "Authority to Sign off Code Changes" of a component are authorised to push directly into these components without a prior PR. of course, it is also worth working with PRs if you want to make the change visible. or it is necessary to make a PR because you are adapting the "public interface" of your component (if one exists or can be defined). the documentation may not be very precise in this respect.

as i said, we are closing the PR for the moment, which does not mean that we don't like the suggestion, but that we have to postpone the topic and take it up again in a larger group. but this will certainly take some time.


@chfsx on behalf of the @ILIAS-eLearning/technical-board

iszmais:
@chfsx Thank you for your response!

If I understand the current ruling right, I come to the conclusion that this is not represented by the current documentations since.
`All contributors are obliged to provide PRs for any change`
and 
`All active developers (including authorities) are contributors`
concludes into
`All (including authorities) are obliged to provide PRs for any change`

If that should not be the case, you might want to add this exception to the documentation.

Greetings,
@iszmais

chfsx:
your reasoning and conclusion are understandable based on the previous documentation, but this does not correspond to the process that has been in place for many years. we probably have to deal with wording that is not 100% precise in many places, but if you have a suggestion as to how the existing documentation can be supplemented with a sentence that better reflects current practice, we would be pleased to receive a PR.

@chfsx on behalf of the @ILIAS-eLearning/technical-board

PR #7557, opened 2024-05-23
T&A: fixes 40641

okaufman:
This will fix https://mantis.ilias.de/view.php?id=40641 
I switched prepareFormOutput with strip_tags. 
With this fix a complete html-tag (eg. `<h3>`) will be removed from the string. But a simple => will remain.
These characters will also remain in the string with this fix: { } \\ ' " &

### DIFF
--- a/Modules/Test/classes/class.ilObjTestGUI.php
+++ b/Modules/Test/classes/class.ilObjTestGUI.php
@@ -2464,7 +2464,7 @@ class ilObjTestGUI extends ilObjectGUI implements ilCtrlBaseClassInterface
             $max_points += $question_gui->object->getMaximumPoints();
         }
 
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much for the PR @okaufman !

One little change request: Could you please let the tags defined in `ilObjectGUI::ALLOWED_TAGS_IN_TITLE_AND_DESCRIPTION` pass? This way we would get a consistent handling of tags.

Thanks again and best,
@kergomard

okaufman:
I made the changes

kergomard:
Sorry, @okaufman , one more thing: We are lacking spaces after the ',' in the two calls to `strip_tags`. I will then merge immediately.

okaufman:
Sorry, I added the spaces.
Thank you for your help.

kergomard:
Thank you very much @okaufman ! Merged and picked to ILIAS 9 and trunk.

PR #7559, opened 2024-05-23
T&A: Ordering Table for Question List (TestObject, questions)

nhaagen:


dsstrassner:
I assign the other authorities, as soon as this PR leaves the Draft status.

kergomard (changes requested):
Thank you very much for replacing the question tables @nhaagen !

I've had a first look at this. Generally this looks straight forward and I like it. It stil have a few small things and a bigger one.

- [ ] First the biggie: I don't think the way of dealing with table commands is right. The `ilObjTestGUI` knows everything about the internals of the table, the `executeCommand` function gets even more hideously long, and then we do a redirect (this for me generally is a smell). I think we should keep the `showQuestions` command as the jump-off point. We then call a command executing the table action in the table. You can either hand in the table command or read it out in the table (I'm unsure about this right now) and then directly show the table. Theoretically we could also achieve this through `ilCtrl::forwardCommand()`, but I don't think we should as this would make a "real" GUI class of the `ilTestQuestionsTableGUI`. This would also mean that we should set the cmd to 'questions' while executing table commands. For the `LogTable` I went with [this approach](https://github.com/ILIAS-eLearning/ILIAS/pull/7614/files#diff-ddf70ab522d8b2d87a5d9e8d13b382e4f4cedd91ca2bc23296fdaa2847117304R2147), but I would be interested in other points of view. This would also allow us to keep some more functions of `ilTestQuestionsTableGUI` private.
- [x] For me this also goes together with the fact that we are creating an `ilTestQuestionsTableGUI` in the constructor of `ilObjTestGUI`. I don't think we should do this.
- [x] Further I'm not a big fan of the anonymous class in [`ilTestQuestionsTableGUI::getBinding()`](https://github.com/ILIAS-eLearning/ILIAS/pull/7559/files#diff-21243e3a30d57cfb43db1a6c4c0756b5c65763e49339455fe3f67bacfeeedc82R179). I would prefere to have this in a separate class, as it would shorten this one even further and also make the class more visible. `class` in  `functions` also just give me knots in the brain ;-) .
- [ ] I would suggest to at least remove the 'il' as well as the 'GUI' from `ilTestQuestionTableGUI`, I would even propose to go the full 9 yards, namespace it, and move it to `src/Questions/Table.php`, thus reducing the class-name to `Table`. I would then usually use it as `QuestionsTable` and we are clear again.

- [ ] The above leads to this pattern [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7559/files#diff-ddf70ab522d8b2d87a5d9e8d13b382e4f4cedd91ca2bc23296fdaa2847117304R1048). I think we should also get rid of this again. Ideally you don't need to touch these lines at all, as I'm going to change them [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7614/files#diff-ddf70ab522d8b2d87a5d9e8d13b382e4f4cedd91ca2bc23296fdaa2847117304R869) anyways.
- [x] Ok, complete nitpick now: Could we get rid of [this else](https://github.com/ILIAS-eLearning/ILIAS/pull/7559/files#diff-ddf70ab522d8b2d87a5d9e8d13b382e4f4cedd91ca2bc23296fdaa2847117304R3632) and the $title variable?
- [x] And one more of those: Could you please move the namespace part [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7559/files#diff-ddf70ab522d8b2d87a5d9e8d13b382e4f4cedd91ca2bc23296fdaa2847117304R75) to a `use` statement (the other kind).
- [x] Could you please check coding styles? There are a few missing spaces and similar e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7559/files#diff-ddf70ab522d8b2d87a5d9e8d13b382e4f4cedd91ca2bc23296fdaa2847117304R211) or [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7559/files#diff-21243e3a30d57cfb43db1a6c4c0756b5c65763e49339455fe3f67bacfeeedc82R308).

Thanks again and best,
@kergomard

dsstrassner:
Hi @nhaagen, 

I add my comments from Discord to the so that they don't get lost:

- [ ]  Please add a Heading between ViewControl and 'Toolbar', e.g. 'Question List'
- [ ]  Please implement the Confirmation Modal as decided by the JourFixe: [JF 2024-01-22](https://docu.ilias.de/goto_docu_wiki_wpage_8176_1357.html) / The Question Icon is in our [icon set](https://github.com/ILIAS-eLearning/ILIAS/blob/b07e9d437baa61a5f4d245c981825006a0e33f6c/components/ILIAS/UI/resources/images/standard/icon_ques.svg)
- [ ] Please state, if this normal, that the Ordering Table uses Numeric Fields with the up-/down-arrows?
Screenshot from a ordering table numeric field

Thanks and best,

@dsstrassner

dsstrassner (changes requested):
Hi @nhaagen, 

I add my comments from Discord to the so that they don't get lost:

- [x]  Please add a Heading between ViewControl and 'Toolbar', e.g. 'Question List'
- [x]  Please implement the Confirmation Modal as decided by the JourFixe: [JF 2024-01-22](https://docu.ilias.de/goto_docu_wiki_wpage_8176_1357.html) / The Question Icon is in our [icon set](https://github.com/ILIAS-eLearning/ILIAS/blob/b07e9d437baa61a5f4d245c981825006a0e33f6c/components/ILIAS/UI/resources/images/standard/icon_ques.svg)
- [x] Please state, if this normal, that the Ordering Table uses Numeric Fields with the up-/down-arrows?
Screenshot from a ordering table numeric field

Thanks and best,

@dsstrassner

nhaagen:
> Please state, if this normal, that the Ordering Table uses Numeric Fields with the up-/down-arrows?

Yes. The position is a numeric field.

nhaagen:
Hi @kergomard ,
thank you for looking into this; your comments sent me pondering.
I agree on the overly crowded `executeCommand`; the main part with integrating DataTable apparently is not the table itself, but the URLBuilder in competition with ilCtrl cmd. In this special case, there are even more caveats, e.g. when adding questions to a pool.

However, I did not want to rework the entire `ilObjTestGUI` , mainly because of time/budget limitations, but also for quite a lot of public functions and (possible) access points. 

Maybe the last commit is a good compromise - I [hand over a callback for the GUI to the table](https://github.com/ILIAS-eLearning/ILIAS/pull/7559/files#diff-ddf70ab522d8b2d87a5d9e8d13b382e4f4cedd91ca2bc23296fdaa2847117304R209), just to move the command handling there. This also goes with some other dependencies for `QuestionsTable` that I originally wanted to keep out, but it gets quite a bit more "encapsulated".

As for the creation of the table in the constructor : I separated the query from the table as to only create _this_  part, and get the table stuff when needed. 

Please consider this an iteration, and I'm looking forward to your thoughts.
Kind regards,
Nils

dsstrassner:
From my point of view, this is now fine for the Test Question Tab. 
As discussed, the numeric field has a strange behavior if one resizes the browser window. @nhaagen will propose a change at the "numeric field"-component.
@kergomard, please merge if you think, the changes "under the hood" for the test are ok for now!

Thanks @nhaagen!

dsstrassner (approved):

nhaagen:
I hope that the changes to numeric (min-width: 60px) and the adding of Question standard icon may be considered trivial enough to be part of this PR.

kergomard (commented):
Hi @nhaagen 

Thank you for working on this once more!

And the rest are procedural details:
* Could you please put all the changes to classes in `UI` into a separate PR and assign it to the UI-Coordinators. I personally see them all as rather trivial, but this is core library stuff and I would like to have the opinion of the people who at the end are going to be blamed for or stuck debugging errors we introduce. One of my concerns: The magic number [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7559/files#diff-8eb3a5fa8a5dab0c546ef49314c48f65ccf7fa7686c5ea3571d8f2c0483de3c4R2). I think thins look broken without the changes (or actually are as in the case of the actions), but the process seems just better to me. This way I will merge this asap and rebase the logging PR on top of it, thus resolving conflicts becomes my problem not yours ;-).
* Could you please remove all changes in `ilObjTestGUI` unless they are absolutely necessary to make it work? The call structure is going to completely change with the merge of the Logging PR I would like to avoid as many conflicts as possible. By "absolutely necessary". By this I mean changes like [this one](https://github.com/ILIAS-eLearning/ILIAS/pull/7559/files#diff-ddf70ab522d8b2d87a5d9e8d13b382e4f4cedd91ca2bc23296fdaa2847117304L918-R943) or [this one](https://github.com/ILIAS-eLearning/ILIAS/pull/7559/files#diff-ddf70ab522d8b2d87a5d9e8d13b382e4f4cedd91ca2bc23296fdaa2847117304L947) or possibly also [this one](https://github.com/ILIAS-eLearning/ILIAS/pull/7559/files#diff-ddf70ab522d8b2d87a5d9e8d13b382e4f4cedd91ca2bc23296fdaa2847117304L947).

I will then merge asap.

Thank you very much and best,
@kergomard

nhaagen:
Hi @kergomard ,
the UI stuff is separated, current PR is rebased and I hope I didn't break the ilObjTestGUI on the last yard.
There are still failing tests, but I'll have to do that on monday.

kergomard:

Test are running now. I also installed this quickly and it seems to work (as far as it goes).

I thus merge now and then we go on from there.

Thanks again and best,
@kergomard

PR #7560, opened 2024-05-23
[FIX] 0041418: Setup routine does not verify required PHP version

chfsx:
See https://mantis.ilias.de/view.php?id=41418

### DIFF
--- a/setup/classes/class.ilSetupAgent.php
+++ b/setup/classes/class.ilSetupAgent.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -18,6 +16,8 @@ declare(strict_types=1);
  *
  *********************************************************************/
 
+declare(strict_types=1);
+
 use ILIAS\Setup;
 use ILIAS\Refinery;
 use ILIAS\Data;
@@ -31,7 +31,9 @@ use ILIAS\Setup\Config;
  */
 class ilSetupAgent implements Setup\Agent
 {
-    public const PHP_MEMORY_LIMIT = "128M";
+    private const PHP_MEMORY_LIMIT = "128M";
+    private const PHP_MIN_VERSION = "7.4.0";
+    private const PHP_MAX_VERSION = "8.0.999";
 
     protected Refinery\Factory $refinery;
     protected Data\Factory $data;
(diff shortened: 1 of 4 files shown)


chfsx:
And we should introduce this in update objectives as well, I'll update the PR asap

chfsx:
@klees hmm we have the following problem: these objectives do not cause an update to be canceled, do we have such a thing?

klees (changes requested):
Hi @chfsx,

this would be a problem caused by a switch in execution logic of the Setup that we made at some point. The Setup stopped execution at the first unachievable objective before, but tried objectives from other branches in the DAG of objectives after. Hence, the `Conditions` currently do not stop the other objectives anymore. We had the same problem with admin-confirmations and solved it by introducing [`ILIAS\Setup\NotExecutableException`](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/Setup/NotExecutableException.php) that stops setup immediately. I think we can reuse this here:

* [x] Add a parameter `$block_setup = false` to the [`ExternalConditionObjective`](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/Setup/Condition/ExternalConditionObjective.php).
* [x] If this parameter is set, make the Objective throw a `NotExecutableException` instead of an [`UnachievableException`](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/Setup/Condition/ExternalConditionObjective.php#L97).
* [x] Write a test for that behaviour.
* [x] Also put this parameter on the constructor of `PHPVersionCondition`.
* [x] Set the parameter accordingly for our global PHP-constraint.

Also:

* [x] Please make the `$max_version`-parameter optional.

I think we should try not to break interface and behaviour of the general `PHPVersionCondition` class. It is not only useful to check PHP versions for all of ILIAS, but could also be a guard for functionality that only works with specific PHP versions. I don't really know if it is currently used in that way, but I can't rule it out either.

Finally:

* [x] Do you understand why the problem wasn't already caught by composer? We do specific PHP version constraints there...

Thanks for looking into this!

klees:
Hi!

I could have answered my last question as well:

> Do you understand why the problem wasn't already caught by composer? We do specific PHP version constraints there...

The ticket says, that `composer install` was not executed, maybe user just updated the OS.

Kind regards!

chfsx:
Hi @klees 
Thanks for the feedback, I implemented the changes and updated this PR.

> Do you understand why the problem wasn't already caught by composer? We do specific PHP version constraints there...

The problem is the following (and I never thought of this case myself): If you do not install ILIAS with Git and thus have to run composer install but install ILIAS from the bundled ZIP or Tar,  then you do not have to run composer install and thus the PHP version check never takes place there. the reporter of the issue therefore did everything right, we just didn't consider this case.

klees (approved):

klees:
Thanks a lot!

PR #7561, opened 2024-05-24
Timings/Activation: Properly transform object title in legacy form title

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=40996

If approved, this MUST be picked to `release_9` and `trunk` as well.

### DIFF
--- a/Services/Object/classes/class.ilObjectActivationGUI.php
+++ b/Services/Object/classes/class.ilObjectActivationGUI.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
Hi @smeyer-ilias , do you have any objections here? I highly recommend merging this (or providing another solution).

smeyer-ilias:
LGTM, but I assign it to @kergomard, since this is a Services/Object related change.

mjansenDatabay:
Thx for having a look @smeyer-ilias

kergomard:
Thank you very much @mjansenDatabay for looking into this! This is not the the issue in the report, but it also is an issue! As far as I can see this is ILIAS 8 only.

PR #7563, opened 2024-05-24
Session/Object(List)/Info: Ensure proper input/output handling

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41228

### DIFF
--- a/Modules/Session/classes/class.ilSessionDataSet.php
+++ b/Modules/Session/classes/class.ilSessionDataSet.php
@@ -14,8 +14,7 @@
  * https://www.ilias.de
  * https://github.com/ILIAS-eLearning
  *
- ********************************************************************
- */
+ *********************************************************************/
 
 declare(strict_types=1);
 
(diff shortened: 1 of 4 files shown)


kergomard (approved):
Thank you very much @mjansenDatabay ! Changes in `ilObjectListGUI` are fine with me. I remove myself from the list of assignees.

mjansenDatabay:
Hi @smeyer-ilias and @alex40724  , do you have any objections here? I highly recommend merging this (or providing another solution).

smeyer-ilias (approved):
Thanks @mjansenDatabay

mjansenDatabay:
Hi @alex40724 ,

do you have any objections regarding `ilInfoScreenGUI`? I recommend addressing this Mantis issue before the release of 8.13 ;-).

Best regards
Michael

alex40724:
@mjansen Thanks for the PR. We had this topic on one of the recent Jour Fixes (29 Apr 2024). The info screen currently accepts HTML. Various components are using this.

See https://docu.ilias.de/goto_docu_wiki_wpage_7928_1357.html (chapter 4 "A breaking change will be implemented in the addProperty method of ilInfoScreenGUI...")

Not rendering HTML is currently scheduled as breaking change for ILIAS 10.

We cannot introduce this with a bug fix, since it will break several views.

E.g. apply your fix and open the starting screen of a test.

It will look similar to the screenshot below and breaks several test features, e.g. "Verwendung vorheriger Lösungen":

Bildschirmfoto 2024-06-26 um 17 28 49

The consumers need to output escape exploitable keys/values here instead.

alex40724 (changes requested):
See last comment.

alex40724:
Here is a screenshot of the ListGUI patch part, which is broken, too:

Bildschirmfoto 2024-06-26 um 17 52 41

alex40724:
I close this since it breaks too much stuff. This needs a very different approach imo.

PR #7564, opened 2024-05-24
fixes 41041 in trash-cron-job

okaufman:
This should fix https://mantis.ilias.de/view.php?id=41041

### DIFF
--- a/Services/SystemCheck/classes/class.ilSystemCheckTrash.php
+++ b/Services/SystemCheck/classes/class.ilSystemCheckTrash.php
@@ -167,8 +167,8 @@ class ilSystemCheckTrash
         
         $query = 'SELECT child,tree FROM tree t JOIN object_reference r ON child = r.ref_id ' .
                 'JOIN object_data o on r.obj_id = o.obj_id ' .
-                'WHERE tree < ' . $ilDB->quote(0, 'integer') . ' ' .
-                'AND child = -tree ';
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
@okaufman The `release_7` branch is no longer maintained. I suggest to rebase your branch to `release_8`, forc-push the changes and select `release_8` as target branch for your changes.

Services/SystemCheck/classes/class.ilSystemCheckTrash.php
```diff
@@ -167,8 +167,8 @@ protected function readSelectedDeleted()
         
         $query = 'SELECT child,tree FROM tree t JOIN object_reference r ON child = r.ref_id ' .
                 'JOIN object_data o on r.obj_id = o.obj_id ' .
-                'WHERE tree < ' . $ilDB->quote(0, 'integer') . ' ' .
-                'AND child = -tree ';
```
mjansenDatabay, line 171: If we change this, the semantics of `readSelectedDeleted` do not match anymore. As far as I understand, the top node of a deleted sub-tree can be determined by comparing `child` with `tree`. If both values are equal, we can assume that this node is the deleted container object (e.g., selected in the user interface for deletion). Deleted sub-objects of this node have the negative value of `child` of the deleted top node as `tree` value.
 
* Removing the conditions semantically results in: Select all deleted nodes
* Where it was: Only select the top nodes of a deletion process

If the change is required to solve a concrete issue, we should think about renaming the method.

PR #7566, opened 2024-05-24
UI/Link: add withDisabled to Link

nhaagen:
Components like e.g. the LinkColumn require a Link; there are occasions, however, in which the link must not be operable (no access, object has been deleted...).
Instead of opening up parameters to Link|string or falling back to more generic wrappers (... the text column), I think it more desirable to stick with the Link and disable it if need occurs.


Amstutz:
Hi @nhaagen 

Thx a lot for this proposal. We are fine with it, however:

- [x] !important: Please check with @catenglaender if the !important statement is really necessary.
- [x] static return type: Please use it, to better narrow down the class actually returned here.

thx a lot

matthiaskunkel:
Jour Fixe, 10 JUN 2024: We highly appreciate this suggestion and accept the PR for trunk / ILIAS 10.

Amstutz:
thx a lot @nhaagen !

PR #7350, opened 2024-04-11
Transform all properties of symbols that are publicly accessible

iszmais:
This transforms all properties inside the Symbols that:
- can be changed by the consumer via the UI-Factory
- are raw part of the rendered HTML
- not restricted object wise (e.g. size)

This PR changes the following presentations:
- The Name/Title of Symbols
- The Label/Alt of Symbols

_This should use a refinery transformation, when #6314 is merged._

PR #7567, opened 2024-05-24
Transform the title property of cards.

iszmais:
https://mantis.ilias.de/view.php?id=41459

**DEPENDS ON https://github.com/ILIAS-eLearning/ILIAS/pull/7572**

This transforms the title property of cards

This PR changes the following presentations:
- The Title of Objects within a tile view.

_This should use a refinery transformation, when #6314 is merged._

### DIFF
--- a/src/UI/Implementation/Component/Card/Renderer.php
+++ b/src/UI/Implementation/Component/Card/Renderer.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -18,6 +16,8 @@ declare(strict_types=1);
  *
  *********************************************************************/
 
+declare(strict_types=1);
+
 namespace ILIAS\UI\Implementation\Component\Card;
 
 use ILIAS\UI\Implementation\Render\AbstractComponentRenderer;
@@ -27,6 +27,7 @@ use ILIAS\UI\Implementation\Component\Symbol\Icon\Standard as StandardIcon;
 use ILIAS\UI\Component\Button\Shy;
 use ILIAS\UI\Implementation\Component\Button\Button;
 use ILIAS\UI\Component\Link\Link;
+use ilObjectGUI;
 
 class Renderer extends AbstractComponentRenderer
 {
@@ -43,6 +44,9 @@ class Renderer extends AbstractComponentRenderer
 
         $title = $component->getTitle();
         $image_alt = $title;
+        if (is_string($title)) {
+            $title = strip_tags($title, ilObjectGUI::ALLOWED_TAGS_IN_TITLE_AND_DESCRIPTION);
+        }
 
         if ($title instanceof Button || $title instanceof Link) {
             $image_alt = $title->getLabel();
@@ -78,7 +82,7 @@ class Renderer extends AbstractComponentRenderer
 
         if ($component->getImage()) {
             $tpl->setVariable("IMAGE", $default_renderer->render(
-                $component->getImage()->withAlt($this->txt("open")." ".strip_tags($image_alt))
+                $component->getImage()->withAlt($this->txt("open") . " " . $image_alt)
             ));
         }
 
--- a/tests/UI/Component/Card/CardTest.php
+++ b/tests/UI/Component/Card/CardTest.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 require_once(__DIR__ . "/../../../../libs/composer/vendor/autoload.php");
 require_once(__DIR__ . "/../../Base.php");
@@ -181,5 +181,37 @@ class CardTest extends ILIAS_UI_TestBase
             "</div>";
 
         $this->assertHTMLEquals($this->brutallyTrimHTML($expected_html), $html);
+    }
+
+    public function test_render_content_with_allowed_tag_in_title(): void
+    {
+        $r = $this->getDefaultRenderer();
+        $c = $this->getBaseCard();
+        $c = $c->withTitle('<b>name</b>');
+
+        $expected_html =
+            "<div class=\"il-card thumbnail\">" .
+            "   <div class=\"il-card-image-container\"><img src=\"src\" class=\"img-standard\" alt=\"open &lt;b&gt;name&lt;/b&gt;\" /></div>" .
+            "   <div class=\"card-no-highlight\"></div>" .
+            "   <div class=\"caption card-title\"><b>name</b></div>" .
+            "</div>";
+
+        $this->assertHTMLEquals($this->brutallyTrimHTML($expected_html), $this->brutallyTrimHTML($r->render($c)));
+    }
+
+    public function test_render_content_with_prohibited_tag_in_title(): void
+    {
+        $r = $this->getDefaultRenderer();
+        $c = $this->getBaseCard();
+        $c = $c->withTitle('<h1>name</h1>');
+
+        $expected_html =
+            "<div class=\"il-card thumbnail\">" .
+            "   <div class=\"il-card-image-container\"><img src=\"src\" class=\"img-standard\" alt=\"open &lt;h1&gt;name&lt;/h1&gt;\" /></div>" .
+            "   <div class=\"card-no-highlight\"></div>" .
+            "   <div class=\"caption card-title\">name</div>" .
+            "</div>";
+
+        $this->assertHTMLEquals($this->brutallyTrimHTML($expected_html), $this->brutallyTrimHTML($r->render($c)));
     }
 }


iszmais:
@thibsy This is similar to https://github.com/ILIAS-eLearning/ILIAS/pull/7350

This change stays in conflict with an internal ILIAS implementation in `ilObjectListGUI`, which took use of that "loophole".
I removed it, which seems to have no effect on the handling at all, but I can't guarantee for that.

Since there is no authority for this section, this might need a further investigation/appraisal by the jour fixe.

Greetings,
@iszmais

kergomard:
Thank you very much for the PR @iszmais .
Could you explain  what you are doing here and why you believe this to be the right move? Does this come from an issue?
Right now I operate under the assumption that we do (maybe, theoretically) let some tags through for titles and description. It is theoretical because so many changes have been applied here, I'm actually not sure anymore, ...but just to be on the safe side. They are defined in `ilObjectGUI::ALLOWED_TAGS_IN_TITLE_AND_DESCRIPTION`.

I'm also not sure where the function `convertSpecialCharacters()` is defined in ILIAS 8?

Don't hesitate to contact me, if you would like to discuss this.

Best,
@kergomard

Services/Object/classes/class.ilObjectListGUI.php
```diff
@@ -3363,7 +3363,7 @@ public function getAsCard(
         }
 
         $card = $ui->factory()->card()->repositoryObject(
-            $title . '<span data-list-item-id="' . $this->getUniqueItemId(true) . '"></span>',
+            $title,
```
mjansenDatabay: Hmmm, this looks dangerous :-). I guess the `data-list-item-id` is needed for some (for me unknown) functionality.

kergomard: Don‘t geht me wrong: This looks like a really bad idea and it should go. I also trink it could be a good idea to decide that titles and descriptions should not support any tags. I’m just concerned of the side effects this might have right now. Last time I got rid of some data-attributes I didn’t like in this class, somebody was not so happy ;-). I’m not even sure anybody really needs this, but as everybody has some expectations towards this class and as the way back with the change in the ComponentRenderer would be barred, I would prefer a less radical solution for ILIAS8 and 9.

iszmais:
@kergomard @mjansenDatabay Thank you for your concern!

The deleted ID is only used for the "showMore" action in limited containers. I resolved this in a separate PR. unfortunately, this led me down a rabbit hole I have to dig up now since atm `card` components can't render refs in titles at all (even if the entity itself allows shy buttons).
See https://mantis.ilias.de/view.php?id=41432 for more info on that.

This needs to be fixed before we can approach the restriction of html in titles any further.
Therefore, I marked this PR as "draft" until I figured out a solution.

But the final goal imo should still be to not allow raw HTML within the title but to use UI components for that instead.
Especially not tags that aren't on the current whitelist.

@kergomard `convertSpecialCharacters()` is/was a recent placeholder function to provide the security of a [WIP refinery for secure HTML transformations](https://github.com/ILIAS-eLearning/ILIAS/pull/6314). This is now merged I will replace that call asap.

Greetings,
@iszmais

iszmais:
@kergomard @Amstutz 

I fixed this one up and separated the changes into three different PRs:

[The first PR](https://github.com/ILIAS-eLearning/ILIAS/pull/7571) Resolves an issue within the rendering of `Button/Shy` Titles. The usage is not possible atm.

when this is fixed...

[The seconde PR](https://github.com/ILIAS-eLearning/ILIAS/pull/7572/files) Removes the last 2 occurrences of HTML from the card title. Both are withing the same object. I checked all other usages of the card, those are all already string-based.

Therefore, after this last fix...

we can merge this PR to banish HTML, and in conclusion XSS, from the card.

Greetings,
@iszmais

kergomard:
Thank you for your efforts @iszmais !

Please see my comment above once more: Titles MAY contain some formatting, as far as I can see. If we like it or not. So we cannot get simply get rid of this.

Best,
@kergomard

kergomard:
Thanks again at @iszmais 
Here are a couple of thoughts, why don't like this:
* As said above from talk to different people, I came to the conclusion that right now we do allow some TAGS. To make this very clear, I introduced the class constant mentioned above `ilObjectGUI::ALLOWED_TAGS_IN_TITLE_AND_DESCRIPTION`. We thus should not change this in any supported version. Right now there are inconsistencies, I know (we eg. strip all tags from descriptions in `ilObjectListGUI`.
* I might be for doing this and actually not allowing any formatting in titles, but this should be done with ILIAS 10 and consistently across all titles and it should be discussed before simply introducing it. There might also be alternatives, like allowing a title to have markdown in it and then rendering it.
* `ILIASObject` could clearly provide a lot more infrastructure when it comes to titles and descriptions, but we are not there yet. I tried to make a start with the `ObjectProperties`, but I need more time and calm to really get around to this.

Best,
@kergomard

iszmais:
@kergomard Thank you a lot for your detailed explanation!

Even if I still see a need for refactoring this, I agree with the point that this should not be changed in a stable release.
Therefore, I adapted the transformation to the common handling of titles. With that base set, we might take a stricter approach here in future ILIAS versions.

I hope this resolves any major concerns you have. If not, feel free to leave me more feedback.

Still, I want to encourage you to take a look into https://github.com/ILIAS-eLearning/ILIAS/pull/7572 since the added tags inside the object list are also currently not part of the allowed tags and therefore this is still an issue.

Greetings,
@iszmais

kergomard:
Sorry, I was just typing one more thing and then you updated this. I also fell into a rabbit hole here!

A short feedback:
* This is all one big mess (not yours and thanks again for forcing me to look deeper into it), everybody does with titles what s/he thinks is right.
* I don't think making the CardRenderer dependent  on `ILIASObject` is a good idea.

Is it ok with you, if I write an issue and set it to "needs JF decision"?
I will look into https://github.com/ILIAS-eLearning/ILIAS/pull/7572 asap. I think this is a very good change, but there we are also depending on the maintainer of the container.

Best,
@kergomard

iszmais:
@kergomard Thanks for your further investigation!

I agree with the mess :smile: 

We will take this to jour fixe.

Greetings,
@iszmais

klees:
Hi @iszmais,

thanks for caring so much about possible XSS vectors in our code base, but the approach on fixing them by escaping in the renderers of the UI-framework has definitely come to an end here. As @kergomard points out, we know for sure that object titles may contain `<i>` and `<b>` tags, whether we like it or not (we don't...). If we would merge this PR we would immediately get lots of bug reports regarding broken views containing such titles, and these bug reports would be understandable. So we simply cannot do this.

There are some ideas how [we could improve handling of text in ILIAS](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/text-handling.md) in the future to (e.g.) preserve the feature that object titles can indeed contain some formatting while also protect from XSS. But we certainly are not there yet and should address this properly, maybe for ILIAS 10, like @kergomard suggested.


As @kergomard also points out, we also don't want dependencies into ILIAS/Object in UI framework code. Even if we would allow them we most certainly wouldn't use a static.

We don't feel that this PR will lead somewhere and hence close it.

Kind regards!
@thibsy, @Amstutz, @klees

iszmais:
@klees Thank you a lot for your explanation.

I'm quite concerned over the priority here to rate a minor style feature over an actual security issue.

But that is out of my responsibility to decide.

Greetings,
@iszmais

klees:
Hi @iszmais,

I currently do not see any actual vector for an attack here, hence no actual security issue. How would a malicious title enter the system? Instead, this seems to be based on the (correct) observation, that some field might be escaped but isn't. This is certainly something we should change, but not a security issue on its own. If you see an actual vector, please report in the ticket or the tracker.

The styling problem, on the other hand, certainly exists. It is not "minor" in the sense that many institutions would be affected by the problem and would report according problems at our issue tracker or service providers.

So, as often times in software development, this is not a problem of right or wrong, only tradeoffs here. There ultimatively is no single person, group or instance that has the responsibility to decide about the priorities you mention. We, as UI coordinators, do not want to incorporate that change as proposed based on the information we currently have. But that's not a decision about general priorities, as the same issue could be fixed in many ways. If you think we should have taken another decision, feel free to feed us more information or take the issue to the JF to be discussed there.

Kind regards!

iszmais:
@klees Thanks again for the response!

I haven't tested this, but during my investigation here I cam across various handlings of title preservations since most objects do their own thing here (as stated be @kergomard in an earlier comment). Therefore, I assume it's most likely that there are still imports or edit forms left which allow such concerning injections. And IMO the approach shouldn't be if we find one of this inside an object, we fix it for that object, and so on for every future occurrence. This would create an unobservable mess of potential holes which someone has to track on every greater update or change.
Or to put it short: I'm sure there is currently an actual vector somewhere, and I would want to prevent it from being tackled as an object-specific issue when it occurs (because many may not follow our discussion here).

I agree that a change like that would significantly affect the majority of presentations of titles, but I see more flexibility on this side of the change. E.g. a less drastic solution could be the strip of all tags inside the title. With that approach we wouldn't create a mess in the presentation but just remove its styling with no critical effect. And since the allowed tags are only style tags, this would have a very minor effect on the presentation, since the change for the users would only be that they don’t see any bold, italic, etc. styling anymore.

Therefore, we would trade a potential issue with a real heavy effect against a real issue with a very light effect.

Greetings,
@iszmais

klees:
Hi @iszmais,

I think you are arguing a strawman here. I did not see that anybody sustained the position that we should only fix injections "for that object, and so on for every future occurence". Instead, we all seem to be agree that we should tackle that general problem.

E.g.:

> There are some ideas how [we could improve handling of text in ILIAS](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/text-handling.md) in the future to (e.g.) preserve the feature that object titles can indeed contain some formatting while also protect from XSS. But we certainly are not there yet and should address this properly, maybe for ILIAS 10, like @kergomard suggested. [me](https://github.com/ILIAS-eLearning/ILIAS/pull/7567#issuecomment-2137847846)

> I might be for doing this and actually not allowing any formatting in titles, but this should be done with ILIAS 10 and consistently across all titles and it should be discussed before simply introducing it. There might also be alternatives, like allowing a title to have markdown in it and then rendering it. [@kergomard ](https://github.com/ILIAS-eLearning/ILIAS/pull/7567#issuecomment-2136880467)

What we do disagree on is the urgency of said general fix and hence the requirements we have for a solution.

Kind regards!

PR #7571, opened 2024-05-28
Add missing shy button prerender in title

iszmais:
https://mantis.ilias.de/view.php?id=41432


Amstutz:
thx for the PR. special thx for adding the test as well.

PR #7572, opened 2024-05-28
Remove HTML form card titles

iszmais:
This PR is a refactoring response to the [Bugfix of shy titles in cards](https://github.com/ILIAS-eLearning/ILIAS/pull/7571)

and a precondition for the [Secure transformation of the card title property](https://github.com/ILIAS-eLearning/ILIAS/pull/7567)

- Removes title button pre-rendering out of the construction process (Is now handled by the card renderer).
- Moves functional identification of "Show More" ids to the javascript from html.

### DIFF
--- a/Services/Container/js/Container.js
+++ b/Services/Container/js/Container.js
@@ -1,13 +1,10 @@
 il = il || {};
-il.Container = il.Container || {};
 (function($, il) {
 	il.Container = (function($) {
 		var initShowMore = function (id, block, url) {
 			$("#" + id).on("click", function(e) {
 				e.preventDefault();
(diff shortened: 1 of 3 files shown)


iszmais:
@kergomard Since its [dependency](https://github.com/ILIAS-eLearning/ILIAS/pull/7571) is now merge, pls have a look into this.

This refactors the functionality of the "show more" to be more convenient and removes some tags from the title that are not allowed in it.

Greetings,
@iszmais

kergomard:
Thank you for the PR @iszmais . I still close this for now as I don't have the resources for the time being to follow up on this properly. Maybe I get around to it again, but not now and not for ILIAS 8.

PR #7573, opened 2024-05-28
Add "HTMLSpecialCharsAsEntities"-Refinery usage

iszmais:
As mentioned [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7350) with the integration of the refinery, the function `convertSpecialCharacters` is obsolete and should be exchanged.

### DIFF
--- a/components/ILIAS/Init/classes/Dependencies/InitUIFramework.php
+++ b/components/ILIAS/Init/classes/Dependencies/InitUIFramework.php
@@ -277,7 +277,8 @@ class InitUIFramework
                             $c["ui.pathresolver"],
                             $c["ui.data_factory"],
                             $c["help.text_retriever"],
-                            $c["ui.upload_limit_resolver"]
+                            $c["ui.upload_limit_resolver"],
+                            $c["refinery"]->encode()->htmlSpecialCharsAsEntities()
                         ),
                         new ILIAS\UI\Implementation\Component\Symbol\Glyph\GlyphRendererFactory(
                             $c["ui.factory"],
@@ -287,7 +288,8 @@ class InitUIFramework
                             $c["ui.pathresolver"],
                             $c["ui.data_factory"],
                             $c["help.text_retriever"],
-                            $c["ui.upload_limit_resolver"]
+                            $c["ui.upload_limit_resolver"],
+                            $c["refinery"]->encode()->htmlSpecialCharsAsEntities()
                         ),
                         new ILIAS\UI\Implementation\Component\Symbol\Icon\IconRendererFactory(
                             $c["ui.factory"],
@@ -297,7 +299,8 @@ class InitUIFramework
                             $c["ui.pathresolver"],
                             $c["ui.data_factory"],
                             $c["help.text_retriever"],
-                            $c["ui.upload_limit_resolver"]
+                            $c["ui.upload_limit_resolver"],
+                            $c["refinery"]->encode()->htmlSpecialCharsAsEntities()
                         ),
                         new ILIAS\UI\Implementation\Component\Input\Field\FieldRendererFactory(
                             $c["ui.factory"],
@@ -307,7 +310,8 @@ class InitUIFramework
                             $c["ui.pathresolver"],
                             $c["ui.data_factory"],
                             $c["help.text_retriever"],
-                            $c["ui.upload_limit_resolver"]
+                            $c["ui.upload_limit_resolver"],
+                            $c["refinery"]->encode()->htmlSpecialCharsAsEntities()
                         )
                     )
                 )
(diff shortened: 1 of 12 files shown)


thibsy (changes requested):
Hi @iszmais

Thx a lot for incorporating the new refinery transformation. Before we can merge this, could you please:

- [x] only replace the `htmlspecialchars()` call inside `AbstractComponentRenderer::convertSpecialCharacters()` with the new transformation, so we can keep this and possibly future changes minimally invasive?
- [x] for some reason the unit-tests are failing, please look into that as well.

Kind regards,
@thibsy

iszmais:
@thibsy Thanks for your feedback.

The test cases seem to fail because the refinery mock is not usable atm.
I provided a change for that [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7751)

Greetings,
@iszmais

klees:
Please see [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7751#issuecomment-2214097102).

iszmais:
@klees Thank you for your feedback and approach here.

I'm not quite sure what way to resolve this under the new circumstances. If we don't want a dependency on the refinery, neither on the DIC, how we should make use of the new-defined refinery to escape the HTML. I assume it's not a fitting solution to duplicate the refinery code.

Pls leave me a short feedback on how you would like that to be handled conceptionally, and I will provide a PR for this.

Greetings,
@iszmais

klees:
Hi @iszmais,

I think "DRY" is a good rule of thumb, but its only that: a rule of thumb. As one can see in #7785, following DRY naively created a huge liability in form of a dependency. So I rather duplicate one line of trivial code then dealing with the fall out of that liability.

So I guess the only transformation that we would need in the renderers and that is not a trivial line of code would be [`HTMLAttributeValue`](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/Refinery/src/Encode/Transformation/HTMLAttributeValue.php). Then: Is this really a case for DI? Do we want to express that this service here (say: the renderers of the ui framework) is dependent on another service that encodes HTML attributes? Do we want to be able to exchange that service for another service that does something similar?

So depending on the answer to the second question (do we want to inject this?) I would either inject just the thing that we actually require (instead of a complete huge surface...) or we could simply instantiate the thing that we know we will get either way.

Kind regards!

iszmais:
Hi @klees,

thank you for your explanation and further investigation.
I updated the refactoring to add the transformation to the renderers in general just like other "non-rendering" services like translations etc.
Besides this now fits the conventional handling, this might also come handy in future to provide such a service. Therefore, I kept the requirement abstract (it requires just any kind of transformation). This should provide good flexibility for testing and also makes this more accessible for future XSS security fixes.

Feel free to leave me some feedback!

Greetings,
@iszmais

klees:
Hi @iszmais,

I really don't get this. Why shall we make this change?

Kind regards

iszmais:
Hi @klees,

Shortly: To prevent code duplication, centralize the functionality and therefore increase maintainability.
Or just like we centralize translations in one service, we should centralize escaping similar.

Further, I assume that this escaping is not the final state of content-based transformations we will be obliged to handle, therefore, a centralization like this will bring a benefit for future adaptions.

Greetings,
@iszmais

klees:
Hi @iszmais,

but as witnessed by #7785, this change does not increase maintainability, but instead decreases it. We are calling a simple function with three parameters now. It hardly get's better then this. This will be adding 36 lines of code and new dependencies between components. How do you think this increases maintainability?

Greetings!

iszmais:
Hi @klees ,

the centralization of the functionality of a refinement purpose in the target service prevents duplicated functionality and therefore the risk of "missing out" one occurrence on a required update.
Even if this adds more code, this code is way less likely to be changed in future since it is part of the internal handling of rendered components and therefore less affected by external influence. The functionality itself, as a security gateway and with its direct effect on the visual presentation, is more likely to be adapted in future. Because of external influences like updates to HTML, rendering exception handling, security concerns, UI adaptions, browser compatibilities, etc. But also since it handles dynamic user inputs in many cases, which often requires developing robustness.

As mentioned [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7567#issuecomment-2137007860) this handling was tackled from different sides in the past and created "one big mess" of escape approaches.
I would like to prevent this to happen again inside the new UI framework.

Greetings,
@iszmais

klees:
Hi @iszmais,

I don't really know what to add anymore. While I appreciate your argument in principle, I don't see how it fits this case here. I think we should agree to disagree and move on.

Hence, I close this.

Kind regards!

PR #7575, opened 2024-05-28
Transform the title of DataCollections on import

iszmais:
https://mantis.ilias.de/view.php?id=40973

### DIFF
--- a/Modules/DataCollection/classes/class.ilDataCollectionDataSet.php
+++ b/Modules/DataCollection/classes/class.ilDataCollectionDataSet.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -13,8 +14,7 @@
  * https://www.ilias.de
  * https://github.com/ILIAS-eLearning
  *
- ********************************************************************
- */
+ *********************************************************************/
 
 /**
  * DataCollection dataset class
@@ -113,6 +113,9 @@ class ilDataCollectionDataSet extends ilDataSet
         ilImportMapping $a_mapping,
         string $a_schema_version
     ): void {
+        foreach ($a_rec as &$value) {
+            $value = htmlspecialchars($value, ENT_QUOTES | ENT_SUBSTITUTE, 'utf-8');
+        }
         switch ($a_entity) {
             case 'dcl':
                 if ($new_id = $a_mapping->getMapping('Services/Container', 'objs', $a_rec['id'])) {


Modules/DataCollection/classes/class.ilDataCollectionDataSet.php
```diff
@@ -113,6 +113,9 @@ public function importRecord(
         ilImportMapping $a_mapping,
         string $a_schema_version
     ): void {
+        foreach ($a_rec as &$value) {
```
mjansenDatabay, line 116: Hi @iszmais , I suggest to use `\ilDataSet::stripTags` for input filtering (like in all other `DataSet` based XML imports) and known mechanisms for escpaing/encoding when putting user input into HTML templates.

iszmais: @mjansenDatabay Thank you for your feedback!

I oppose the usage of strip tags here for two reasons:
- This would limit the presentation of titles. Especially within a learning management system, I see that as a concern. This would make it hard to create an Object regarding HTML for educational purposes (e.g. my DataCollection "Cool ways to use the \<div\> Tag!" would then only be shown as "Cool ways to use the Tag!" which would be confusing)
- This would make the approach disjunct from the "manual" creation, which currently allows you to save and present tags within the title (even when there are slightly modified).

I may take this concern to a future approach to stricten the input of title, but those should then apply to all ways of creation.

Greetings,
@iszmais

mjansenDatabay: Hi @iszmais ,

I would like to add two thoughts:

1. IMO the input handling in the XML import process should be similar/identical to the input handling when users submit data via the ILIAS graphical user interface (GUI). It is just another application-endpoint. So applying `\ilDataSet::stripTags` would be the most appropriate solution for the problem here.
2. The `title` is a property of the `Object` component, **not** of the `DataCollection` component. Although the title is not handled centrally by the `Object` component in imports (at the moment), the same business logic (restrictions etc.) applied by the `Object` component when handling user input via the GUI **must** apply here. So IMO the "limited presentation of title"-argument is not valid here.

iszmais: Thank you for your response @mjansenDatabay ,

it seems to be a bit of a misunderstanding here:
#### The current common state is that tags are **not** stripped via GUI. They are preserved in a readable form.

I'm not totally if this applies to all types and how the abstract object handles it, but I checked this for the most common types like course, group, category etc. 

Therefore, I support your point, but that resolves into preserved tags via import as well.

Greetings,
@iszmais

iszmais: As said, we might want to change that in the future. But this should be applied simultaneously to all creation processes

kergomard:
Hi @iszmais 
Just one thought on this, as today seems to be the day of me diving ever deeper into cleaning our outputs ;-) : Doesn't this all go sideways the moment this is displayed? From where I stand this should lead to double encodings the moment we move to a more consistent encoding when displaying. As far as I can see, we are very rarely avoiding double encoding in the current output. From where I stand I think it would be correct to let everything through here, ... or we need to change our output encoding.

Best,
@kergomard

iszmais:
@kergomard Thanks for the concern.

Yes, this might be a risk, but I would rate the security of the comfort at first here.
In most normal cases this encoding should do nothing add all. And this makes the rare occasion of this issue less critical and easier trackable.

Those this should not be seen as the perfect solution but to "soften" the issue regarding its prio.

"It's not good, but it's better"

Greetings,
@iszmais

PR #7581, opened 2024-05-30
[FIX] 0041428: _isFileInline function does not exist anmyore

chfsx:
Hi @schmitz-ilias 
This PR fixes https://mantis.ilias.de/view.php?id=41428, I had thought i had adjusted all the posts, this one seems to have been lost. the PR is now catching up!


schmitz-ilias:
Thansk for the PR @chfsx !

PR #7583, opened 2024-05-31
UI, ViewControls/DataTable: 41310, group of nullcontrols to reserve both input names of pagination

nhaagen:
Addendum to https://github.com/ILIAS-eLearning/ILIAS/pull/7488;
The pagination actually has _two_ inputs, and the names are still shifted if only one nullControl is used.

### DIFF
--- a/src/UI/Implementation/Component/Input/ViewControl/Group.php
+++ b/src/UI/Implementation/Component/Input/ViewControl/Group.php
@@ -65,8 +65,12 @@ class Group extends ViewControlInput implements ViewControlGroupInterface, Group
      */
     public function getContent(): Result
     {
-        if (empty($this->getInputs())) {
-            return new Ok([]);
+        $inputs = array_filter(
+            $this->getInputs(),
+            fn($input) => ! ($input instanceof NullControl)
+        );
+        if (empty($inputs)) {
+            return new Ok(null);
         }
         return parent::getContent();
     }
--- a/src/UI/Implementation/Component/Input/ViewControl/Group.php
+++ b/src/UI/Implementation/Component/Input/ViewControl/Group.php
@@ -65,12 +65,8 @@ class Group extends ViewControlInput implements ViewControlGroupInterface, Group
      */
     public function getContent(): Result
     {
-        $inputs = array_filter(
-            $this->getInputs(),
-            fn($input) => ! ($input instanceof NullControl)
-        );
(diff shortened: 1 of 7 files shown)


src/UI/Implementation/Component/Input/ViewControl/Group.php
```diff
@@ -65,8 +65,12 @@ public function withNameFrom(NameSource $source, ?string $parent_name = null): s
      */
     public function getContent(): Result
     {
-        if (empty($this->getInputs())) {
-            return new Ok([]);
+        $inputs = array_filter(
```
klees: Hi @nhaagen,

I think this will bite us at some point, since this adds some element of surprise. I might not grasp the complete problem, so take this with a grain of salt. But I think we would be better of if the `NullControl` has `OK(null)` content (if it does not already have that) and we move that special case for filtering to the usage location where we actually need it.

Maybe it would be good to discuss this.

Kind regards!

klees (changes requested):

src/UI/Implementation/Component/Input/ViewControl/Group.php
nhaagen: Hi @klees ,
you're probably right, the filtering is too radical in the group.
I reverted the changes in Group and added the null-result to the ViewControl..

klees (approved):

klees:
Dear @thibsy and @Amstutz,

I think we can move on here.

Kind regards!

thibsy (changes requested):
Hi @nhaagen

Thx for looking into this! Input issues need somewhat special attention, so I have a few questions/remarks before we can merge this:

- [x] `NullControl::getContent()`: why do we need to override this method from `Input::getContent()`? It looks like this would be an edge case, in which content also needs to be returned if `Input::withInput()` is NOT invoked. By looking at your unit tests, this does not need to be the case though, so I think we should be able to simply remove this method again.

- [x] `ViewControlNullTest`: we believe this test case does not need to be covered explicitly, since the null input should be processed like any other input. I noticed however, that the `ViewControl\Group` isn't covered by any unit test yet - so you may implement it there if you like.

Kind regards,
@thibsy

thibsy (approved):
Thx @nhaagen for implementing the changes and special thanks for adding the missing unit test! This LGTM now, I'll merge and pick for all maintained versions.

PR #7585, opened 2024-05-31
ilObjFile fix type issues

Amstutz:
Some type issues, occured while testing some calendar workflows. See: https://mantis.ilias.de/view.php?id=41456 

Note, there could be more such issues in the affected classes.


chfsx:
I modified your commit a bit and fixes this using https://github.com/ILIAS-eLearning/ILIAS/commit/cb118d463dd63ae9c6b4f3c338a2febb01179e62, thanks a lot for the PR!

PR #7586, opened 2024-05-31
COPage Bugfix #0040624: Use Ilias System Style for TOC Links

fhelfer:
https://mantis.ilias.de/view.php?id=40624

TOC in ILIAS Content Pages has hard coded colors
As a first Bugfix we suggest using the ILIAS Main Color from System Styles.
For future Versions, adding a customizable Wiki Link for Content Styles could be considered
As this is not pickable for release_9 a additional PR will follow

Best, @fhelfer

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -14510,15 +14510,15 @@ li.ilc_page_toc_PageTOCItem {
   white-space: nowrap;
 }
 a.ilc_page_toc_PageTOCLink {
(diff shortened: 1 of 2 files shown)


fhelfer:
~~Do not merge as there are issues with delos.css~~

alex40724:
Thanks for the PR. See my comment in the bug tracker. This is content area, not system area.

PR #7587, opened 2024-05-31
COPage Bugfix #0040624: Use Ilias System Style for TOC Links

fhelfer:
https://mantis.ilias.de/view.php?id=40624

See:  #7586


alex40724:
Thanks for the PR. See my comment in the bug tracker. This is content area, not system area.

Amstutz:
Note, that I just merged a very similar PR: https://github.com/ILIAS-eLearning/ILIAS/pull/7608

@alex40724 : Where do we draw the line between content and system styles? So far I expected things placed int content.css to be content style related and things compiled to delos system style related. If not, where is the difference documented?

PR #7589, opened 2024-05-31
Container: Fix accessing undefined array key in `ItemSetManager`

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41457


alex40724:
Thanks  for the PR.

PR #7592, opened 2024-06-03
Setup: fix skip command while using update or install command. (#41410)

lukastocker:
https://mantis.ilias.de/view.php?id=41410


klees:
Already fixed in >9.

PR #7594, opened 2024-06-03
T&A, 39127: copy/delete comments along with the question

nhaagen:
https://mantis.ilias.de/view.php?id=39127


thojou:
Hallo @nhaagen,

I have reviewed you changes and merged them into release_9. The changes are also cherry-picked into the trunk.

Thank you.

PR #7598, opened 2024-06-04
TA, 41436: allow creation of class for import/unserialize

nhaagen:
https://mantis.ilias.de/view.php?id=41436


kergomard:
Thank you very much and sorry @nhaagen ! Have applied this in a drive-by fix yesterday and not realized that there already was this PR. I think we should move away from serialization, but pragmatically did the same thing for the time being. See: https://github.com/ILIAS-eLearning/ILIAS/commit/a4f7239b2c733eaf6e025cda876bb55c92e5b396

Sorry again and best,
@kergomard

PR #7600, opened 2024-06-04
Bugfix table action for detaching forum moderators

playamongthestars:
Bugfix table action for detaching forum moderators. "ALL_OBJECTS"-Feature is not suitable here.


mjansenDatabay:
Thx a lot @nmatuschek !

PR #7602, opened 2024-06-04
[Docu] Install ILIAS: Removed language section

acgruber:
Referring to [Mantis issue 0041304](https://mantis.ilias.de/view.php?id=41304), I've removed the language section from the install script. The feature to install multiple languages during the installation process has been abandoned in June 2022.

When approved, please cherry-pick to respective branches. Thanks!

### DIFF
--- a/docs/configuration/install.md
+++ b/docs/configuration/install.md
@@ -434,10 +434,6 @@ configuration might look like this afterwards:
 	"http" : {
 		"path" : "http://demo1.cat06.de"
 	},
-    "language" : {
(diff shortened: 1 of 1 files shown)


klees (approved):

klees:
This seems to be trivial, so I'm have merged this. Just want to point out: I do not think that I "maintain" `install.md` or that changes there need to go through me specifically. @acgruber and I will try to make a proposal how we could handle changes for the documentation in the future.

PR #7604, opened 2024-06-04
T&A: 41152, Fix call to undefined method getTestId in ilTestAccess

matheuszych:
When trying to access the detailed overview of a test pass you get an error "Call to undefined method ilTestAccess::getTestId()".

The method `getTestId` was removed some time ago but was not properly replaced. The method in which `getTestId` is called has got an additional parameter `$test_id` which was unused but implemented probably to replace the `getTestId` method call.

Just by simply replacing the `getTestId` call with the `$test_id` parameter we get the expected behaviour.

This issue is only occurs in ILIAS 10.


kergomard:
Thank you very much @matheuszych ! This is what I wanted to do.

PR #7608, opened 2024-06-05
UI: adjust WIKI's link colors. (#41284)

lukastocker:
The following conditions were changed to ILIAS standard colors:
- not visited
- visited
- hover

https://mantis.ilias.de/view.php?id=41284


Amstutz:
thx a lot!

PR #7609, opened 2024-06-05
LDAP: Increase length of password fields

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=31280

If approved, this must be picked to `release_9` and maybe adapted for `trunk` (due to the component revision).


mjansenDatabay:
Hi @pascalseeland ,

any objections here?

Best regards
Michael

PR #7370, opened 2024-04-16
Update Password Form in UserService to KS Components (ILIAS 10)

fhelfer:
Updated the User Password Form in UserService with new KS UI Components.
After consultation with @kergomard, we simply use one input for the new password. No need to duplicate this if the plain text display is available via the eye glyph.

There is also the question @kergomard, whether you still want the code to be moved to Auth for this PR?

Best, @fhelfer

PR #7611, opened 2024-06-05
User Bugfix: Initial Password set for new Users bugs out because of error in PasswordSettings

fhelfer:
This PR fixes a looping redirect because the initial Password set for users throws an error, but the error page always redirects to password change. 

The Bug occurs because of the changes I did in https://github.com/ILIAS-eLearning/ILIAS/pull/7370 Whoops :)
thx @nmatuschek for finding this.

Best, @fhelfer

### DIFF
--- a/components/ILIAS/User/classes/class.ilForcedUserPasswordChangeStartUpStep.php
+++ b/components/ILIAS/User/classes/class.ilForcedUserPasswordChangeStartUpStep.php
@@ -20,6 +20,7 @@ declare(strict_types=1);
 
 use ILIAS\Init\StartupSequence\StartUpSequenceStep;
 use Psr\Http\Message\ServerRequestInterface;
+use ILIAS\Authentication\Password\LocalUserPasswordSettingsGUI;
 
 /**
  * Class ilForcedUserPasswordChangeStartUpStep
@@ -75,8 +76,12 @@ class ilForcedUserPasswordChangeStartUpStep extends StartUpSequenceStep
 
     public function execute(): void
     {
-        $this->ctrl->redirectByClass(
-            ['ildashboardgui', 'ilpersonalsettingsgui'],
+        $this->ctrl->getLinkTargetByClass(
+            [
+                ilDashboardGUI::class,
+                ilPersonalSettingsGUI::class,
+                LocalUserPasswordSettingsGUI::class
+            ],
             'showPassword'
         );
     }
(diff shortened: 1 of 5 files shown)


components/ILIAS/User/classes/class.ilForcedUserPasswordChangeStartUpStep.php
```diff
@@ -75,8 +76,18 @@ public function shouldInterceptRequest(): bool
...
+                ilPersonalSettingsGUI::class,
+                LocalUserPasswordSettingsGUI::class
+            ],
+            'showPassword'
+        );
+        return;
+
         $this->ctrl->redirectByClass(
-            ['ildashboardgui', 'ilpersonalsettingsgui'],
+            ['ildashboardgui', 'ilpersonalsettingsgui', \ILIAS\Authentication\Password\LocalUserPasswordSettingsGUI::class],
```
mjansenDatabay:
```suggestion
            [ilDashboardGUI::class', ilPersonalSettingsGUI::class, \ILIAS\Authentication\Password\LocalUserPasswordSettingsGUI::class],
```

components/ILIAS/User/classes/class.ilForcedUserPasswordChangeStartUpStep.php
```diff
@@ -75,8 +76,18 @@ public function shouldInterceptRequest(): bool
...
     {
+        $this->ctrl->getLinkTargetByClass(
+            [
+                ilDashboardGUI::class,
+                ilPersonalSettingsGUI::class,
+                LocalUserPasswordSettingsGUI::class
+            ],
+            'showPassword'
+        );
+        return;
```
mjansenDatabay: Hmmm, can we just remove the `return;` and the code below?

mjansenDatabay (changes requested):

mjansenDatabay:
@kergomard Should be move the user-related class to the `Auth` component?

mjansenDatabay (approved):

kergomard (approved):
Thank you very much  @fhelfer . The changes LGTM.

@mjansenDatabay : I would think this class should be in `Auth`, yes. ...unless you see a reason for it not to.

Best,
@kergomard

mjansenDatabay (changes requested):
Hi @fhelfer , thx for the PR. Let's rename class `LocalUserPasswordSettingsGUI` to `ilLocalUserPasswordSettingsGUI` (without namespace) as long ist #[7624](https://github.com/ILIAS-eLearning/ILIAS/pull/7624)  is not merged.

mjansenDatabay (approved):

PR #7613, opened 2024-06-05
UI/DateColumn: add user pref to examples

nhaagen:
encourage people to use user-prefs in their tables :)

### DIFF
--- a/src/UI/examples/Table/Data/base.php
+++ b/src/UI/examples/Table/Data/base.php
@@ -19,6 +19,9 @@ function base()
     $df = new \ILIAS\Data\Factory();
     $refinery = $DIC['refinery'];
     $request = $DIC->http()->request();
+    $current_user_date_format = $df->dateFormat()->withTime24(
+        $DIC['ilUser']->getDateFormat()
+    );
 
     /**
      * This is what the table will look like:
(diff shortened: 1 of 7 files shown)


mjansenDatabay:
Thx for providing more examples @nhaagen . Since this PR targets the UI component(s), I change the assignment.

src/UI/examples/Table/Data/base.php
```diff
@@ -227,6 +231,7 @@ protected function getRecords(Range $range = null, Order $order = null): array
             ->data('a data table', $columns, $data_retrieval)
             ->withId('example_base')
             ->withActions($actions)
+            ->withNumberOfRows($current_user_pref_limit)
```
oliversamoila: Hello @nhaagen

The function "withNumberOfRows($current_user_pref_limit)" should not be documented additionally, because it should no longer be available with ILIAS 10. (See "Abandon Hits/Page as a User Setting" https://docu.ilias.de/goto_docu_wiki_wpage_8182_1357.html)

Can you confirm that the following thoughts are correct or correct them?
* Global default for all tables is the row quantity of25
* Developers can define their own default for a specific table (5 - ∞) (by parameterizing the table when using the UI Data Table)
* Users can customize the row specification for a specific table. The information is saved in the session, if implemented.

Remark:
The user preference is abolished for old tables (see "Abandon Hits/Page as a User Setting") 
As old tables still require information about the number of rows, the default value =25 is also transferred here. It is not possible for the user to adjust the default across the platform. The user can determine the number of rows for individual tables, which ILIAS remembers in the session or permanently.

Thanks and best regards
Oliver

Amstutz:
thx a lot @nhaagen for the improvements, IMO we should and will take them. nevertheless, @oliversamoila had some points. Can you answer them. Ill assign you for the moment. feel free to reassign me, as soon as cleared.

nhaagen:
Thanks @oliversamoila for pointing to the deprecation.
I removed the user-prefs row numbers from the example and set the default value to 25.
Consumers my change that default via  "$table->withNumberOfRows()".
Viewcontrols will override this setting; their values are stored in a session.

nhaagen:
Just for documentation: [this PR](https://github.com/ILIAS-eLearning/ILIAS/pull/8106)  asks to remove withNumberOfRows in favor of withRange.

PR #7614, opened 2024-06-05
Introduce New Logging and Refactor

kergomard:
Hi all

First:  I split this into a lot of commits, but they are not really clean as I didn't see another path then to approach this iteratively. I also don't ask anybody for a review, will merge this myself, and take care of the issues arising from this. @dsstrassner and I will do some testing before I merge.
Two:
* @alex40724 : There are changes in the question-specific classes of COPage. I see them as minimal, but I still requested a review of these changes. You can find them in [this commit.](https://github.com/ILIAS-eLearning/ILIAS/pull/7614/commits/67aab57ecea4d7fc7b94a59e078494fce256b0c3)

Now for the beef of the matter:
Ok, I've announced this a few times at the TechSquad now here finally is: This PR introduces the new logging implementation for the test. It has gotten really big, but I don't see much of a way around it.
So, what does it do:
* It removes the old logging implementation and introduces the new logging itself.
* In order to do this properly I needed to change the way questions are edited. Up to now we relied on redirects, many of them based on URI "assembly". This here removes many of them. I left unrelated redirects (partly) in, to at least manage the size issue partially.
* In order to do this in turn I needed to refactor the questions, partially because it was absolutely necessary, but partially also to understand what I was dealing with. So yes, this lead to quite a few changes, but it lead also to the removal of 2200 Lines, even if looking at the complete set of changes and they also contain a lot of shortening  of code lines (ie. introducing line brakes, ie. generating more lines) as well as a hugely more powerful logging system.
* Further I also needed to refactor the whole `Marks` section and I decided to use this to replace the current table implementation with the new `UI`-component as well as to introduced a `repository pattern`, even if the pattern is not perfect yet, in my eyes (there is more potential gain there).

There are a few more not directly  changes:
* I moved classes to the src-folder and introduced namespaces.
* As promised in the TechSquad I removed the `Services`-classes on the `global $DIC` and moved them to a local dependency container. I created a minimal public interface giving access to the `GeneralQuestionProperties` (properties shared by all questions) as well as an `areQuestionsAnsweredByUser`-Function. This is kept small on purpose, as this should still be seen as work in progress, but if I would have started to change things in there even more, this would have gotten completely out of hand.

Best,
@kergomard


alex40724 (approved):
Looks good to me. Thanks.

PR #7619, opened 2024-06-05
Logging: remove Monolog

klees:
Hi everyone,

we (mostly @kergomard and @lukastocker, I only a little bit) are currently working on [updating the dependencies to the versions that have been approved by the JF until now](https://github.com/ILIAS-eLearning/ILIAS/pull/7141).

Sadly, no one seems to feel responsible for Monolog and no PR for the inclusion of Monolog was opened and discussed so far. Various attempts to reach out to @smeyer-ilias have not resulted in a PR until now. Since the Monolog dependency is required on virtually every page of ILIAS (it plays a role in logging...) I hereby propose to remove the dependency from our code for the time being, so we can move on and run and develop the rest of the system with the updated dependencies, with broken logging, sadly.

If someone steps up for Monolog, this commit will be easy to revert. If no one steps up for Monolog, we will have to think about how we want to perform logging in the future. Either way, we should clear the way soon for everyone to work with the freshly approved dependencies.

Kind regards!

### DIFF
--- a/components/ILIAS/Logging/classes/NullLogger.php
+++ b/components/ILIAS/Logging/classes/NullLogger.php
@@ -22,7 +22,6 @@ namespace ILIAS\components\Logging;
 
 use ilLogger;
 use ilLogLevel;
-use Monolog\Logger;
 use Exception;
 
 class NullLogger extends ilLogger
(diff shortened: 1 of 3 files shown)


klees:
Since #7629 exists and has been approved by JF, this is not required anymore.

dsstrassner:
As this not needed, it also does not need the Jour Fixe Label...

PR #7621, opened 2024-06-06
Repository: Remove Creation of Column Separators

kergomard:
Hi @alex40724 

With the merge of the new "Add New Object" solution, we don't need the separators in the `NewItemsGroups` anymore. This removes  the corresponding code and drops the records from the table as well as removing the now unneeded `type`-column.

We do not loose much code, but at least a few lines are gone.

Best,
@kergomard


alex40724:
Thanks!

PR #7160, opened 2024-03-11
T&A 0028645 File Upload Answers missing in Archive Export

kunzeus:
https://mantis.ilias.de/view.php?id=28645

Let me know if anything needs changing or can be improved

Tested on: ILIAS 8
Code looks the same on 7, 8, 9, Trunk

PR #7622, opened 2024-06-06
CI: Check the code style of files in PRs, don't try to fix it

mjansenDatabay:
This PR changes the `php-cs-fixer` when running in a "GitHub Actions" context. Instead of fixing files, this PR suggests to only check them.

I stumbled upon the issue because no code style violations have been detected in PR https://github.com/ILIAS-eLearning/ILIAS/pull/7160 .

If the `php-cs-fixer` actually fixes files, the exit code is `0`, which does not lead to a failing CI pipeline.

### DIFF
--- a/CI/PHP-CS-Fixer/run_check.sh
+++ b/CI/PHP-CS-Fixer/run_check.sh
@@ -19,7 +19,7 @@ else
   	if [ -f ${FILE} ]
   	then
 	  	echo "Check file: ${FILE}"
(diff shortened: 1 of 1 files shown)


thibsy (approved):
Hi @mjansenDatabay

I assigned myself since this was "looking for a shepherd" and I am not totally unfamiliar with the CI workflows and php-cs-fixer. Also, I would really much appreciate this change, since your findings mean our CI currently doesn't check the PHP code-style properly (good catch btw).

I have tested this locally and can confirm that the `--dry-run` option results in an exit code of `1` if the given file does not comply with the code-style and `0` if everything's neat - which should now fail the workflow if something is not compliant.

IMO this can be merged.

Kind regards,
@thibsy

PR #7624, opened 2024-06-06
ilCtrl: Add Exception for baseClass/cmdClass in Removing Slashes from $_GET

fhelfer:
References https://github.com/ILIAS-eLearning/ILIAS/pull/7055/
Adding the possiblity to Use namepaces in Routing
This will most likely not open us up to attacks from outside, because other characters will still be removed
$this->ctrl->setCmdClass() will also check for correct Path and else default to baseClass or null

Best, @fhelfer

### DIFF
--- a/components/ILIAS/Init/classes/class.ilInitialisation.php
+++ b/components/ILIAS/Init/classes/class.ilInitialisation.php
@@ -75,17 +75,28 @@ class ilInitialisation
      */
     protected static function recursivelyRemoveUnsafeCharacters($var)
     {
+        $unsafe_characters = ["\x00", "\n", "\r", "'", '"', "\x1a"];
+
         if (is_array($var)) {
             $mod = [];
             foreach ($var as $k => $v) {
                 $k = self::recursivelyRemoveUnsafeCharacters($k);
-                $mod[$k] = self::recursivelyRemoveUnsafeCharacters($v);
+                if (!in_array($k, [ilCtrlInterface::PARAM_CMD_CLASS, ilCtrlInterface::PARAM_BASE_CLASS], true)) {
+                    $mod[$k] = self::recursivelyRemoveUnsafeCharacters($v);
+                } else {
+                    $mod[$k] = str_replace(
+                        $unsafe_characters,
+                        "",
+                        $v
+                    );
+                }
             }
             return $mod;
         }
+
         return strip_tags(
             str_replace(
-                array("\x00", "\n", "\r", "\\", "'", '"', "\x1a"),
+                array_merge($unsafe_characters, ["\\"]),
                 "",
                 $var
             )


components/ILIAS/Init/classes/class.ilInitialisation.php
```diff
@@ -79,10 +79,19 @@ protected static function recursivelyRemoveUnsafeCharacters($var)
             $mod = [];
             foreach ($var as $k => $v) {
                 $k = self::recursivelyRemoveUnsafeCharacters($k);
-                $mod[$k] = self::recursivelyRemoveUnsafeCharacters($v);
+                if (!in_array($k, [ilCtrlInterface::PARAM_CMD_CLASS, ilCtrlInterface::PARAM_BASE_CLASS], true)) {
+                    $mod[$k] = self::recursivelyRemoveUnsafeCharacters($v);
+                } else {
+                    $mod[$k] = str_replace(
+                        array("\x00", "\n", "\r", "'", '"', "\x1a"),
```
mjansenDatabay: I suggest to put `array("\x00", "\n", "\r", "'", '"', "\x1a")` in a variable (e.g. `$unsafe_chars`) and use this variable in the last `return` statement, e.g. `return strip_tags(str_replace(array_merge($unsafe_chars, ["\\"]), '', $var));`

mjansenDatabay (changes requested):
I am skeptical about these proposed changes. If this PR were integrated, the `Init` component would have very specific knowledge abouth HTTP query parameters of the `ilCtrl` component and makes assumptions on how to sanitise them.

I could imagine two alternatives.

1. The removal of potential problematic characters from the superglobal `$_GET` variable should be moved to the `HTTP` service and should not be part of class `\ilInitialisation` anymore.
One can debate if the retrieval of all `$_GET`-parameters requests should apply a "Default Sanitizing Tansformation" to remove `"\x00", "\n", "\r", "\\", "'", '"', "\x1a"` from the `string` values.
Furthermore, the API could be extended to allow something like `$http->wrapper()->get()->withCustomSanitizingForNextRetrieval($trafo)->retrieve('name', $trafo);` (or similar, I know that the suggested usage is a bit clumsy) to overwrite the "Default Sanitizing Tansformation". This would allow a component to change the problematic characters to be removed for the next (and only for this) retrieval.
But of course, this has to be discussed with the concept authorities of the `HTTP` component.

2) Depending on the current status of the "Component Revision" and the availability of the "Wiring" in the class `\ilInitialisation`, one could also imagine a new "Interface" whose implementations define certain sanitizing rules for query parameters of a specific component.

`\ilInitialisation::recursivelyRemoveUnsafeCharacters` is called in `\ilInitialisation::removeUnsafeCharacters`, which is called in `\ilInitialisation::initCore`, which is called by our main initialisation method `\ilInitialisation::initILIAS`.

So the call stack is:

* \ilInitialisation::initILIAS
* \ilInitialisation::initCore
* \ilInitialisation::removeUnsafeCharacters
* \ilInitialisation::recursivelyRemoveUnsafeCharacters

If the "Wiring" is already completed at this point and contributions to a new "Query Parameter Sanitizing" interface could be retrieved from the components, we could suggest a dynamic sanitizing process instead of replacing values recursively in $_GET here.

There could be a new interface `\ILIAS\(Init|Http)\SuperGlobalSanitizer` with a `apply/execute/sanitize/purify` method (names can be discussed later).

The `Init/Http` component itself could provide an default implementation for this to replace `"\x00", "\n", "\r", "\\", "'", '"', "\x1a"`. Other components could provide other sanitizing modifications for particular parameters. The `ilCtrl` component should then provide an own implementation to handle `baseClass` and `cmdClass`, if PHP namespaces should be allowed in these query parameters.

Of course, there are open questions like conflicts/priorities if multiple components define rules for the same parameter, but maybe this could be handled similar to `\ILIAS\GlobalScreen\Scope\Layout\Factory\LayoutModification`.

pascalseeland:
I'm very much in favor of moving the variable sanitizing into the http component so we have all the handling of the parameters at one location. I'm not sure if we want to have a possibility for individual components to use a custom trafo for the retrieval. I would instead to have the http component to have a set of well defined trafos to choose from so we can ensure we have a minimum level of security when dealing the request parameters.

mjansenDatabay:
I'll close this for now for the reasons mentioned above.

PR #7625, opened 2024-06-06
fix error in listView

ingoj:
under some conditions invaltid objects are found when displaying object-lists. This results in an "array key not found" error. This PR shows a possible fix.


ingoj:
sorry, this PR contained the wrong fixes

PR #7626, opened 2024-06-06
fix undefined array key for some nodes

ingoj:
Sometimes invalid objects are found when displaying object lists. This leads to "array key not found" errors. This PR shows a possible solution

### DIFF
--- a/Services/Object/classes/class.ilObjectActivation.php
+++ b/Services/Object/classes/class.ilObjectActivation.php
@@ -537,6 +537,9 @@ class ilObjectActivation
             }
             // #7571: when node is removed from system, e.g. inactive trashcan, an empty array is returned
             $node = $tree->getNodeData($item_ref_id);
+            if (!isset($node["ref_id"])) {
+                continue;
+            }
             if ((int) $node["ref_id"] !== $item_ref_id) {
                 continue;
             }
--- a/Services/Object/classes/class.ilObjectActivation.php
+++ b/Services/Object/classes/class.ilObjectActivation.php
@@ -537,10 +537,7 @@ class ilObjectActivation
             }
             // #7571: when node is removed from system, e.g. inactive trashcan, an empty array is returned
             $node = $tree->getNodeData($item_ref_id);
-            if (!isset($node["ref_id"])) {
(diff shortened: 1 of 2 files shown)


ingoj:
May apply to release_9 and trunk, too

kergomard:
Thank you very much @ingoj for the PR!

I would have three little requests:
* Could you please provide an issue report in mantis? You don't have to reproduce this on test8, but so we have a reference.
* Could you please combine your changes with the ones in the next `if`-statement, so: `if (!isset($node["ref_id"]) || (int) $node["ref_id"] !== $item_ref_id) {`.
* Might you be so kind and fix the copyright headers as you have to touch this once more anyways?

Thanks again and best,
@kergomard

ingoj:
if- statements are combined
mantis entry will follow in acouple of minutes (not the only one for today)
Sorry, I am not familiar about the changes wirth the copyright headers, do you have some instructions?

kergomard:
Thank you very much @ingoj !

I will take care of the headers after merge, if you want. Sorry, I'm not finding the rules for the copyright, but basically they are that every code-file in ILIAS needs to start with the precise copyright at the top of the file. Nothing apart from the opening php-tag should be above. A correct example would be e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/blob/b07e9d437baa61a5f4d245c981825006a0e33f6c/components/ILIAS/ILIASObject/classes/class.ilObjectActivation.php#L1).

And then I think you now introduced superfluous brackets `) || (` with your commit ;-).

Thanks and best, 
@kergomard

ingoj:
Brackets are removed, and I tried to fix the copyright lines (it semms copyright must start alt line 3, was line 5)

kergomard:
Don't worry about the Copyright. I will figure it out afterwards. The rest is ok now: I  just wait for the issue in Mantis and merge tomorrow.

ingoj:
mantis 41486

kergomard:
Merged and picked to ILIAS 9 and trunk.

You actually fixed the copyright alright, ...just in the wrong file ;-).

Best,
@kergomard

PR #7627, opened 2024-06-06
fix some functoins if optional parameters are omitted

ingoj:
in some functions, $user and $types should be optional parameters. in ILIAS 8 (and posibly 9/trunk) these parameters must be supplied, but the function itself contains checks for these paramaters. 
This PR contains a possible, but untested fix.


ingoj:
Mantis ID: 41487

jeph864:
@ingoj,
Can you add the parameter signatures to inc.soap_functions.php? Without such adjustments there would be a type error

ingoj:
@jeph864,
I updated inc.soap_functions.php accordingly, I hope everything is fine now. Sorry I could not test the changes.

jeph864:
@ingoj ,
thanks for the PR

PR #7628, opened 2024-06-06
DataCollection: Fix import of reference with multi selection

iszmais:
https://mantis.ilias.de/view.php?id=41422


mBeym:
looks good to me

PR #6652, opened 2023-11-23
Composer: Add `phpmailer/phpmailer` as dependency

mjansenDatabay:
This PR adds `phpmailer/phpmailer` as composer dependency.

Usage:
* `components/ILIAS/Mail`

Wrapped By:
* `\ilMailMimeTransportBase`

Reasoning:
* `PHPMailer` is a de facto standard for sending emails with PHP. It supports mutliple transport channels as SMTP, mail or sendmail. Furthermore it supports multiple authentication methods and provides proper APIs for attachment handling (and multipart handling in general), encoding and encryption.
* We should not/do not want to reinvent the wheel in ILIAS and should rely on this library crafted by these email standard specialists.

Maintenance:
* `PHPMailer` is actively maintained by multiple contributors. There is recent activity (even today, see: https://github.com/PHPMailer/PHPMailer/commits/master).
* Security issues are always fixed in a timely manner followed by new releases.

Links:
* Packagist: https://packagist.org/packages/phpmailer/phpmailer
* GitHub: https://github.com/PHPMailer/PHPMailer
* Documentation: https://github.com/PHPMailer/PHPMailer/wiki / https://github.com/PHPMailer/PHPMailer/tree/master/docs / https://phpmailer.github.io/PHPMailer/

PR #7629, opened 2024-06-07
Jour Fixe monolog dependency

smeyer-ilias:
adds monolog dependency.

Monolog logging was integrated into ILIAS with release 5.1. [1]
Monolog is one of the main logging frameworks for PHP and offers features like formating, different appenders  and handlers.
Monolog is actively maintained. See github project [2]

[1] https://docu.ilias.de/goto_docu_wiki_wpage_151_1357.html
[2] https://github.com/Seldaek/monolog

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -191,10 +191,10 @@
 		},
 		"monolog/monolog" : {
 			"source" : "https://github.com/Seldaek/monolog",
-			"used_version" : "v2.6.0",
+			"used_version" : "v2.9.3",
 			"wrapped_by" : "components/ILIAS/Logging",
 			"added_by" : "Stefan Meyer <smeyer.ilias@gmx.de>",
-			"last_update" : "2022-05-11",
+			"last_update" : "2024-06-07",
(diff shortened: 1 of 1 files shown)


kergomard:
Hi @smeyer-ilias 

Thank you very much for the PR. Could you please add the necessary information to your first comment, so that we can keep track of our current reasoning. We need this to be able take an informed decision at the next Jour Fixe and to have handy for next time? See https://github.com/ILIAS-eLearning/ILIAS/pull/6652 for a good example.

Best,
@kergomard

matthiaskunkel:
Jour Fixe, 10 JUN 2024: Dependency is accepted for ILIAS 10 / trunk.

kergomard:
Hi @smeyer-ilias 
This was not the procedure for these changes. The merge should happen by the Technical Board not by you.

And on a side-note: I missed this completely, when looking at your PR, but the procedure was for you to create the PR against `composer_new.json`. Wouldn't we have been so careful and checked all this, your dependency would simply be gone in a few days and we would have needed to investigate.

Best,
@kergomard

PR #7638, opened 2024-06-10
Bugfix table multi action for using 'ALL OBJECTS'

playamongthestars:
Bugfix for multi table action using "ALL OBJECTS"

### DIFF
--- a/components/ILIAS/Contact/classes/class.ilMailMemberSearchGUI.php
+++ b/components/ILIAS/Contact/classes/class.ilMailMemberSearchGUI.php
@@ -256,10 +256,20 @@ class ilMailMemberSearchGUI
         $selected_user_ids = $this->http->wrapper()->query()->retrieve(
             'contact_search_members_user_ids',
             $this->refinery->byTrying([
-                $this->refinery->kindlyTo()->listOf($this->refinery->kindlyTo()->int()),
+                $this->refinery->kindlyTo()->listOf($this->refinery->kindlyTo()->string()),
                 $this->refinery->always([])
             ])
         );
+
+        if ((string) current($selected_user_ids) === 'ALL_OBJECTS') {
+            $selected_user_ids = [];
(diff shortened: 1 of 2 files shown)


components/ILIAS/Contact/classes/class.ilMailMemberSearchGUI.php
```diff
@@ -256,11 +256,21 @@ protected function sendMailToSelectedUsers(): void
         $selected_user_ids = $this->http->wrapper()->query()->retrieve(
             'contact_search_members_user_ids',
             $this->refinery->byTrying([
-                $this->refinery->kindlyTo()->listOf($this->refinery->kindlyTo()->int()),
+                $this->refinery->kindlyTo()->listOf($this->refinery->kindlyTo()->string()),
```
mjansenDatabay, line 259: Hi @nmatuschek ,

thanks for the PR. I suggest keeping the `int` transformation here. You can then add further check:

If `$selected_user_ids` is an empty array, transform the request param kindly to `string` and check if there is exactly 1 element. and the element equals `ALL_OBJECTS`

Best regards
Michael

components/ILIAS/Contact/classes/class.ilMailingListsGUI.php
```diff
@@ -185,7 +185,7 @@ private function getMailingListIdsFromRequest(): array
             $ml_ids = $this->http->wrapper()->query()->retrieve(
                 'contact_mailinglist_list_ml_ids',
                 $this->refinery->byTrying([
-                    $this->refinery->kindlyTo()->listOf($this->refinery->kindlyTo()->int()),
+                    $this->refinery->kindlyTo()->listOf($this->refinery->kindlyTo()->string()),
```
mjansenDatabay, line 188: The same applies here.

mjansenDatabay (changes requested):

mjansenDatabay (approved):

PR #7639, opened 2024-06-10
[FIX] File extension missing when downloaded from data collection (#0041496)

mhergerdt:
Ensure that files downloaded from data collections contain the file extension.

Bug report: https://mantis.ilias.de/view.php?id=41496


iszmais:
Thanks a lot for the PR.
 
In ILIAS the title is different from the filename since the file object has an own, independent title (which is mostly, but not always, the same, except the extension).

Therefore, the transformation of the UI title is not the correct approach here, but I used your approach to correct the selection of the correct filename in [another PR](https://github.com/ILIAS-eLearning/ILIAS/pull/7641). For a normal ILIAS file object this might be a difference, but since the titles inside the DC aren't editable, this makes no different here.

Greetings,
@iszmais

PR #7640, opened 2024-06-10
ilUIPluginRouterGUI: Fix class paths handling

mjansenDatabay:
This PR fixes the `ilCtrl` based routing if `ilUIPluginRouterGUI` is the base class. The `is_file` check does not work because paths returned by `ilCtrlInteface::lookupClassPath` are relative paths starting with `./components/*`.

Mantis Issue: https://mantis.ilias.de/view.php?id=41501


klees:
Thx @mjansenDatabay. I have added a little explanation to the commit message.

PR #7642, opened 2024-06-10
UI: Remove StripTags from Text-Inputs

kergomard:
Hi all

While researching for [Inconsistent Handling of Titles and Description](https://mantis.ilias.de/view.php?id=41447) I came across this transformation in text- and textarea-inputs. I don't think this is the right behavior and it makes implementing the decision above not possible for the new settings form. I'm not adding this as an `Issue` right now, but as an `Improvement`, but I can a Mantis ticket if you prefer.

Thanks and best,
@kergomard

### DIFF
--- a/components/ILIAS/UI/src/Component/Input/Field/Factory.php
+++ b/components/ILIAS/UI/src/Component/Input/Field/Factory.php
@@ -50,6 +50,10 @@ interface Factory
      *     3: >
      *         Text Input MUST NOT be used for letter-only input, an Alphabet Field
      *         is to be used instead.
+     *     4: >
+     *         If withoutStripTags is set, the consumer MUST make sure the value
+     *         is proberly sanitized before outputing it.
+     *
      *   interaction:
      *     1: >
      *         Text Input MUST limit the number of characters, if a certain length
@@ -450,6 +454,9 @@ interface Factory
      *     2: >
      *       Radios MAY also be used to select between two options
      *       where one is not automatically the inverse of the other
+     *     3: >
+     *         If withoutStripTags is set, the consumer MUST make sure the value
+     *         is proberly sanitized before outputing it.
      *   wording:
      *     1: Each option MUST be labeled.
      *     2: The options' labels MUST state something positive.
--- a/components/ILIAS/UI/src/Component/Input/Field/Text.php
+++ b/components/ILIAS/UI/src/Component/Input/Field/Text.php
@@ -36,4 +36,9 @@ interface Text extends FilterInput
      * Gets the max length of the text input
      */
     public function getMaxLength(): ?int;
+
+    /**
+     * Disable stripping tags from user input
+     */
+    public function withoutStripTags(): Text;
 }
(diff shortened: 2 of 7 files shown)


alex40724:
Again: It woul be good to somehow ensure that the "to display titles and descriptions in plain text (incl. HTML strings)" part is in place before removing the stripping. You might get security issues otherwise.

More important: The JF decision is about titles and descriptions. This PR removes the transformation from all text/textarea inputs. Imo you need full output escaping in place before doing so.

klees (changes requested):
Hi @kergomard,

we do not really like that `stripTags` in the general implementation of UI input fields, just like you don't seem to like it 🙂

The specific lines have been introduced as an reaction to some discussion regarding XSS. Many locations, even not in the core, will now depend on `strip_tags` being called internally, although many of us know that there are better solutions to the general problem.

We cannot simply do a rug pull here, as this would compromise security in the aforementioned locations. We thus propose to leave the default behaviour as is and instead provide a possibility to opt out of the `strip_tags` at locations that take proper care themselves.

We hence propose to introduce a method `withoutStripTags` together with a rule that states that clients will need to take care of proper treatment when using that method. Please prepare an according PR and present it on the JF.

Kind regards!

klees:
Hi @kergomard,

I pass this back to you for the time being. Please assign back once you are finished.

Thanks!

kergomard:
Hi all

Thanks for the feedback! I applied the changes and hope this is more or less, what you expected. I decided to not provide a getter for `$strip_tags_from_input` as I think this really should be completely internal.

I'm adding the strip_tags transformation in `withInput()`, from where I stand this is a hell of a side-effects, but I don't see another way of doing this right now. Worst case, if `withInput()` is called multiple times, we end up with multiple trafos. I'm glad, if you have a better idea! Just had an idea: We could also replace `bool $strip_tags_from_input` with `?Transformation $strip_tags_trafo` and set the trafo there in the constructor. We then could null it if `withoutStripTags()` is called. This way we can guarantee only one trafo is set. I think this is better and I just added a commit changing this.

I already added a Jour Fixe tag as that is what I understood from your comment what I should do. Sorry, if I got that wrong!

Looking forward to your feedback and best,
@kergomard

thibsy (changes requested):
Hi @kergomard,

Thanks a lot for implementing the requested changes!

I think you are trying to solve a problem which has already been solved before, with constraints for required inputs.

- [x] `FormInput::getOperations()` approach: If you move [the special handling](https://github.com/ILIAS-eLearning/ILIAS/pull/7642/files#diff-d79e132c56e317e305a3d2ae803c24cf83730ad77a93c13afa68cba960838b08R125-R128) inside `withInput()` into an override of `FormInput::getOperations()`, where you yield an additional transformation if some flag is set, we get rid of the problem you are stating. Maybe it makes sense to set this transformation once but only yield it if some boolean flag is set.

Kind regards,
@thibsy

kergomard:
Hi @thibsy 

Thank you for the hint! ...this makes things a lot easier! Done.

Best,
@kergomard

Amstutz:
Thx a lot. This needs JF approval before merging. However, from our perspective, this can be merged directly after JF.

matthiaskunkel:
Jour Fixe, 17 MAR 2025: We highly appreciate this suggestion and accept the PR for trunk.

Amstutz:
thx a lot!

PR #7654, opened 2024-06-13
UI Components: Refactor/Extend Toast Interfaces

iszmais:
Hello,

this is an alternative approach to integrate the timing functions inside the toast structure by appending it to the container.

The first approach on individual toasts can be found here: https://github.com/ILIAS-eLearning/ILIAS/pull/6164

With that approach we could secure that all toasts are timed the same but still provide the feature of a global configuration.

Therefore, this is crucial for the resolvement of https://mantis.ilias.de/view.php?id=41514

### DIFF
--- a/src/UI/Component/Toast/Container.php
+++ b/src/UI/Component/Toast/Container.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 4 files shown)


iszmais:
@Amstutz @klees @thibsy

 Due to the detailed discussion in the previous PR I don't see any need to put that to the association in any way (also since the applied approach is the suggested solution and this is a requirement to the solution of an actual bug).

But if you think otherwise, please let me know.
I will provide a presentation for the JF then.

Greetings,
@iszmais

PR #7655, opened 2024-06-13
Notification: Implement timing functions on Container (and remove timing functions from toast)

iszmais:
**THIS PR DEPENDS ON https://github.com/ILIAS-eLearning/ILIAS/pull/7654**

https://mantis.ilias.de/view.php?id=41514

In general, this clears up the timing responsibility of toast by appending it to the container.
Further, this removes trivial artifacts from the toast themself.


mjansenDatabay:
The changes in the `Notifications` and `Awareness` components look good to me.

However, there might be objections by Fabian regarding the `GlobalScreen` changes. There are also small HTML and test case changes in the `UI` component that need to be approved.

But first, #7654 needs to be approved and merged as a prerequisite.

PR #7656, opened 2024-06-13
[FEATURE] CI: JavaScript code-style exception

thibsy:
Hi folks,

This PR introduces a new rule to our JavaScript code-style in form of an exception/deviation:

> Airbnb does not allow to manipulate/reassign parameters completely. However, reassigning **parameter properties** of e.g. `HTMLElement`s is a common use-case, which has led to many unnecessary variables being initialised to work around this code-style rule. Therefore we allow the manipulation/reassignment of **parameter properties**.

This partially supersedes [Airbnbs ruling to disallow parameter reassignments](https://github.com/airbnb/javascript?tab=readme-ov-file#functions--reassign-params) and will keep our JavaScript code more compact. Note that reassigning a parameter otherwise is still prohibited.

I have updated the according [ESLint configuration](https://eslint.org/docs/latest/rules/no-param-reassign#props) so these exceptions will no longer be recognised as an error.

Kind regards,
@thibsy

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -54,6 +54,14 @@
       {
         "devDependencies": true
       }
+    ],
+    // allow to reassign argument properties for e.g. altering
+    // HTMLElement styles and state.
+    "no-param-reassign": [
+      "error",
+      {
+        "props": false
+      }
     ]
   }
 }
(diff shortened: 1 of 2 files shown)


matthiaskunkel:
Jour Fixe, 08 JUL 2024: Thibeau notified us about a code style exception. Unlike the original AirBnB code style, we allow the reassignment of parameter properties.

PR #7662, opened 2024-06-14
[FIX] CI: use correct import paths

thibsy:
Hi all,

I have recently [committed a change](https://github.com/ILIAS-eLearning/ILIAS/commit/cca848627e2227ab9e9aecef2f6e8294249a020e) which fixed some invalid paths, and noticed that other scripts are also affected by the same issue.

The following checks have not been checked for pull requests against trunk for some time now:
- Copyright check
- PHP code-style check
- Special characters check
- Language file sort check

This went unnoticed, because these scripts exited with `0` even though some files and functions could not be loaded  - so the workflows always passed. This should be a reminder to always check for the existence of files before using them, and to be careful with exit codes - especially for scripts used in GitHub actions.

Kind regards,
@thibsy

### DIFF
--- a/scripts/Copyright-Checker/copyright-checker.sh
+++ b/scripts/Copyright-Checker/copyright-checker.sh
@@ -173,7 +173,7 @@ function perform_copyright_check() {
 
 # this helper is only required if we are in a GitHub-run.
 if ! [ -z "${GHRUN}" ]; then
-  source "$(pwd)/CI/Import/Functions.sh"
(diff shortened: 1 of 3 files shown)


mjansenDatabay (approved):
Thx @thibsy ,

LGTM!

I found other problems in (e.g.) `scripts/results/report-results.sh` but I handed over this to @gvollbach 

Best regards,
Michael

PR #7664, opened 2024-06-14
User: Fix Resetting Login Attempts

kergomard:
See: https://mantis.ilias.de/view.php?id=33768

### DIFF
--- a/Services/User/classes/class.ilObjUserFolderGUI.php
+++ b/Services/User/classes/class.ilObjUserFolderGUI.php
@@ -516,6 +516,9 @@ class ilObjUserFolderGUI extends ilObjectGUI
                 false
             );
             if ($obj instanceof \ilObjUser) {
+                if (!$obj->getActive()) {
+                    $obj->setLoginAttempts(0);
+                }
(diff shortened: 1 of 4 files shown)


Services/User/classes/class.ilObjUserFolderGUI.php
```diff
@@ -520,6 +520,7 @@ public function confirmactivateObject(): void
                     true,
                     $ilUser->getId()
                 );
+                $obj->setLoginAttempts(0);
```
mjansenDatabay: IMO we should check if the user was inactive. Otherwise, we reset login attempts of active users.

```php
            if ($obj instanceof \ilObjUser) {
                if (!$obj->getActive()) {
                    $obj->setLoginAttempts(0);
                }
                $obj->setActive(
                    true,
                    $ilUser->getId()
                );
                $obj->update();
            }
```

kergomard: Thank you very much @mjansenDatabay ! You are completely right and I didn't think far enough. Will fix and merge.

PR #7666, opened 2024-06-14
Auth: Show "StandardLoginForm" on the login page independently of "CA…

mjansenDatabay:
…S" or "Shibboleth"

Mantis: https://mantis.ilias.de/view.php?id=32755

If approved, this MUST be picked/adapted to `release_9` and `trunk`.

### DIFF
--- a/Services/Init/classes/class.ilStartUpGUI.php
+++ b/Services/Init/classes/class.ilStartUpGUI.php
@@ -790,26 +790,17 @@ class ilStartUpGUI implements ilCtrlBaseClassInterface, ilCtrlSecurityInterface
     {
         global $tpl;
 
-        // @todo move this to auth utils.
-        // login via ILIAS (this also includes ldap)
-        // If local authentication is enabled for shibboleth users, we
-        // display the login form for ILIAS here.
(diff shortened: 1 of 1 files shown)


Services/Init/classes/class.ilStartUpGUI.php
```diff
@@ -790,26 +790,17 @@ protected function showLoginForm(string $page_editor_html, ilPropertyFormGUI $fo
     {
         global $tpl;
 
-        // @todo move this to auth utils.
-        // login via ILIAS (this also includes ldap)
-        // If local authentication is enabled for shibboleth users, we
```
mjansenDatabay, line 795: The comment said: 

```
If local authentication is enabled for shibboleth users, we display the login form for ILIAS here.
```

But exactly this did **not** happen. If CAS or Shibboleth is set as the default authentication method, the standard login form is neither built nor rendered.

thojou (approved):

mjansenDatabay:
The PR for ILIAS >= 9.x can be found here: https://github.com/ILIAS-eLearning/ILIAS/pull/9323

PR #7667, opened 2024-06-14
Repository/Session: include session date in last visited (36691)

schmitz-ilias:
This PR fixes [36691](https://mantis.ilias.de/view.php?id=36691).

Unfortunately, a small change was necessary to Navigation, introducing a different handling for the title specifically for sessions: since Navigation refreshes the title of previously persisted last-visited-entries via `ilObject::_lookupTitle` basically directly from obj_data, there is no way for the Session to interfere and set its title correctly from the outside. I couldn't find an easy way around this, let me know if I missed something.

### DIFF
--- a/Modules/Session/classes/class.ilObjSessionGUI.php
+++ b/Modules/Session/classes/class.ilObjSessionGUI.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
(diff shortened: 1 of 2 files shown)


alex40724 (approved):
@schmitz-ilias  Changes are ok for me. For 10 there might be a better solution in the object service, but afaik this is not official yet (or it was just about icons, not sure).

Thanks. Please merge this to 8 to trunk.

PR #7669, opened 2024-06-17
User/Registration: Don't render empty "Other" section in legacy forms

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41012

Both consumers of `\ilUserProfile::addStandardFieldsToForm` provide an empty array as default for `$custom_fields`, so we should check if it is empty when handling the custom fields while rendering the form.

Please check whether this must be picked to `release_9` and/or `trunk`.

### DIFF
--- a/Services/User/classes/class.ilUserProfile.php
+++ b/Services/User/classes/class.ilUserProfile.php
@@ -16,8 +16,6 @@
  *
  *********************************************************************/
 
-// mjansen@databay.de essential for mail constants, do not remove this include
-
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much for investigating this and for the PR @mjansenDatabay !

The fix for ILIAS 9 and trunk is rather different, but a fix is also needed. See: https://github.com/ILIAS-eLearning/ILIAS/commit/b720d4ed30adef8c6348432162c5f2277fe5694e

Best,
@kergomard

PR #7670, opened 2024-06-17
[FIX] ActiveRecord: PHP 8 dynamic properties

thibsy:
Hi @chfsx,

I noticed a deprecation notice triggered by `ActiveRecord::buildFromArray()` in case of joins, since (if I understood correctly) the joined data is added to the object on which the first join is invoked on, adding the properties dynamically until now. [Since PHP 8.2 this triggers a deprecation notice](https://www.php.net/manual/en/migration82.deprecated.php#migration82.deprecated.core.dynamic-properties) if dynamic properties are created without implementing the corresponding magic methods `__get()` and `__set()`. [According to the manual](https://www.php.net/manual/en/language.oop5.overloading.php#language.oop5.overloading.members) it makes sense to accompany them by the `__isset()` and `__unset()` methods - so I implemented them as well.

Let me know if you want to follow up on this.

Kind regards,
@thibsy

### DIFF
--- a/Services/ActiveRecord/class.ActiveRecord.php
+++ b/Services/ActiveRecord/class.ActiveRecord.php
@@ -24,6 +24,7 @@
  * @description
  * @version 2.0.7
  */
+#[\AllowDynamicProperties]
 abstract class ActiveRecord
 {
     protected bool $ar_safe_read = true;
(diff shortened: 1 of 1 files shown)


chfsx:
Hello @thibsy 

I think the deprecation with PHP8.2 is a good moment to think about the future of ActiveRecortd. Meanwhile, the much more performant repository pattern is used in many places, which also leads to more lightweight classes. i think we should continue on this path.

For ActiveRecord, I would currently rather suggest that we use the attribute 

```
#[\AllowDynamicProperties]
```

thibsy:
@chfsx good point, I agree. Maybe we could deprecate this service sometime?

For the time being, I have remove the "implementation" for dynamic properties and marked the class like you suggested, which should get rid of the deprecation notice as well.

chfsx:
@thibsy yes, we should certainly deprecate the service on occasion, so that at least new implementations do not rely on it. i think we could announce this after the trunk is opened for ILIAS 11.

PR #7672, opened 2024-06-17
8/Learning-Sequence push footer down 41463

catenglaender:
https://mantis.ilias.de/view.php?id=41463

# Issue

Footer is not positioned at the bottom of the content area.
Header overlaps Slate and content area slightly.

image

# Change

Learning Sequence kiosk mode had no content div that actually stretches and pushes the footer down. Used the mainspacekeeper ID for the content area in the html template which comes with the layout styling that was missing in this kiosk mode.

Deleted left overs from a previous implementation of the mode info.

image

# Outlook

Footer issue needs to be solved in 9+ as well.

### DIFF
--- a/Modules/LearningSequence/templates/default/tpl.kioskpage.html
+++ b/Modules/LearningSequence/templates/default/tpl.kioskpage.html
@@ -1,4 +1,4 @@
-<div id="il_center_col" class="col-sm-12">
+<div id="mainspacekeeper" class="col-sm-12">
 
 	<div class="ilLSOKioskModeObjectHeader">
 		<div class="media il_HeaderInner">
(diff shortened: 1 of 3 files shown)


Amstutz:
Hi @catenglaender 

Thx a lot for the fix along with the detailed description. Highly appreciated. However, the change to a prominent ID in the template is critical for a productive release. We never know what JS in core/and or plugins are relaying on this. Therefore, we would like to take this to the JF.

thx.

matthiaskunkel:
Jour Fixe, 08 JUL 2024: We accept the suggested change of the ID for Release_9 and trunk. Please add a short notice to the Release Page of ILIAS 9 -> Changed Behaviour to notify users that have a customised skin.

klees:
Hi @catenglaender,

please create a PR that solves this very issue for ILIAS >9!

Kind regards!

PR #7674, opened 2024-06-19
DataCollection Bugfix Fix Setup failing because DIC is array

fhelfer:
https://mantis.ilias.de/view.php?id=41574

```php
PHP Fatal error:  Uncaught Error: Call to a member function isDependencyAvailable() on array in /var/www/ilias_v2/ILIAS/components/ILIAS/your-component/classes/class.your-stakeholder.php:28
Stack trace:
#0 /var/www/ilias_v2/ILIAS/components/ILIAS/your-component/classes/Setup/class.your-migration.php(62): your-stakeholder->__construct()
#1 /var/www/ilias_v2/ILIAS/components/ILIAS/Setup/src/CLI/MigrateCommand.php(154): your-migration->prepare()
#2 /var/www/ilias_v2/ILIAS/components/ILIAS/Setup/src/CLI/MigrateCommand.php(83): ILIAS\Setup\CLI\MigrateCommand->listMigrations()
#3 /var/www/ilias_v2/ILIAS/vendor/composer/vendor/symfony/console/Command/Command.php(298): ILIAS\Setup\CLI\MigrateCommand->execute()
#4 /var/www/ilias_v2/ILIAS/vendor/composer/vendor/symfony/console/Application.php(1040): Symfony\Component\Console\Command\Command->run()
#5 /var/www/ilias_v2/ILIAS/vendor/composer/vendor/symfony/console/Application.php(301): Symfony\Component\Console\Application->doRunCommand()
#6 /var/www/ilias_v2/ILIAS/vendor/composer/vendor/symfony/console/Application.php(171): Symfony\Component\Console\Application->doRun()
#7 /var/www/ilias_v2/ILIAS/components/ILIAS/Setup/src/CLI/App.php(53): Symfony\Component\Console\Application->run()
#8 /var/www/ilias_v2/ILIAS/artifacts/bootstrap_setup.php(3016): ILIAS\Setup\CLI\App->enter()
#9 /var/www/ilias_v2/ILIAS/cli/setup.php(21): entry_point()
#10 {main}
  thrown in /var/www/ilias_v2/ILIAS/components/ILIAS/DataCollection/classes/class.your-stakeholder.php on line 28
```

### DIFF
--- a/components/ILIAS/DataCollection/classes/class.ilDataCollectionStakeholder.php
+++ b/components/ILIAS/DataCollection/classes/class.ilDataCollectionStakeholder.php
@@ -25,7 +25,7 @@ class ilDataCollectionStakeholder extends \ILIAS\ResourceStorage\Stakeholder\Abs
     public function __construct()
     {
         global $DIC;
-        $this->owner = $DIC->isDependencyAvailable('user')
+        $this->owner = !is_array($DIC) && $DIC->isDependencyAvailable('user')
             ? $DIC->user()->getId()
             : (defined('SYSTEM_USER_ID') ? (int) SYSTEM_USER_ID : 6);
     }


iszmais:
Thank you @fhelfer for your contribution!

However, I'm quite confused that den DIC is an array here since that shouldn't be the case.
Therefore, I would suggest fixing that state instead of resolving it.

Do I miss any previous discussions why this is the desired state?
If I request you to link any Feature Request or Mantis ticket relating to this PR.

Greetings,
@iszmais

fhelfer:
Hey @iszmais,

I had created a Mantis Issue [41574](https://mantis.ilias.de/view.php?id=41574), but Richard closed it.
In the Setup Process, Environment is responsible for providing the Resources.
ATM, I don't know what the best solution is to resolve this error in Setup.
Maybe passing the User through optional constructor parameter and using the environment in prepare() Step of the Setup.

What do you think? Might this need further discussions?

Best, @fhelfer

iszmais:
@fhelfer Thx for the further explanation.

@klees Since the ticket did not state that out clearly: Do you approve such individual approaches in every component to resolve this issue?
In my first impression I would share the view to resolve that more centrally, but my expertise from the setup is not big enough to say if that is feasible or not.

If yes, I would resolve this for my other components the same way. This might even be worth an announcement on the jour fixe.

Pls let me know about your approval (or disapproval).

Greetings,
@iszmais

klees:
Hi @iszmais,

I'm not quite sure what you are asking about here. As outlined in the issue that @fhelfer has linked: During the setup there is no `$DIC` and I do not think that we should or even can have one. Basically, using the  `$DIC` inline as a service locator, as this code shows, is the attempt to simply pull something into existence from thin air. You are relying on something that you did not ask about nor took care to initialize. You just want to have it and use it. Now. This has worked in ILIAS at large because we all rely on the big pile of spaghetti that is `ilInit`. Until it hasn't worked in certain cases because something was not initialized in the expected order or at all. This is something we have begun to tackle using dependency injection in many individual locations in the past and that we will tackle with exactly that mechanism, dependency injection, in the future, as [decided on during the component revision](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/components-and-directories.md#types-of-dependencies-and-integration-strategies).

So I neither approve nor disapprove of this solution. Also, it doesn't have to do anything with the setup. I think the question is posed wrong, because we want to get rid of that integration mechanism in general.

I hope this makes some sense.

Kind regards!

Addendum: We can of course look into possible solutions for the issue, as this might also yield insights about the general direction we want to take. Shall we maybe schedule a meeting in Discord, so interested people can listen as well?

mjansenDatabay:
I agree with @klees . The setup does not (and never did) make any promises to provide a `$DIC` to setup agents/or objectives.

The problem is specific to IRSS-based migrations. The migrations require stakeholders, and a stakeholder requires a "User Id", currently read from the `$DIC`.

I guess the problem was introduced with: https://github.com/ILIAS-eLearning/ILIAS/commit/4cbfef29849b44b25bc83c92e3c633a65850d8ea and (of course) refactorings in the context of the component revision.

So the actual question is IMO: How can the IRSS resource stakeholders retrieve the default owner id in the context of the setup migrations.

alex40724:
"During the setup there is no $DIC"

Then we (better a IRSS helper) could check for isset(). But why is $DIC an array during setup? Imo if it is set, it should work. If it does not work, it should not be set.

klees:
Hi @alex40724,

that's the magic of globals: Everyone can set them everywhere =)

I can neither find nor remember any code _of the Setup_ that sets `$DIC` to some array. Only thing the Setup does with `$DIC` is in NIC-key registration and [that code](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/setup_/classes/class.ilNICKeyRegisteredObjective.php#L74) looks as if it cannot be responsible for the existence of an array in `$DIC`.

So from my POV setting of `$DIC` to some array could happen about everywhere. Even in some code (like plugins...) that we do not currently have in our repository.

Kind regards!

alex40724:
Messing with the $DIC global this way (except when mocking in tests) should be treated as bug, reported and fixed.

klees:
Hi  @alex40724,

I agree in general, but there currently are cases where I do not see any alternatives in the short term.

Kind regards!

iszmais:
Thank you all for your feedback.

This seems to be a bigger topic, but I think we all can mostly agree that an implementation of the `DIC`, if it exists, should follow the common convention.

I will take some time to find the cause of this.
However, since this PR does not "Fix" anything and just displays a workaround, I'm not keen on merging this.

@fhelfer I'm sorry that I cannot utilize your effort here. I will let you know when I find a centralised way to resolve this.

Greetings,
@iszmais

PR #7691, opened 2024-06-20
Fix 41541: 'Enter' key in number filter input field collapses the fil…

tfamula:
…ter content area, instead of applying the filter

https://mantis.ilias.de/view.php?id=41541


Amstutz:
Hi @tfamula 

You directly edited the bundled dist file. So please:

- [x] Edit Source: Edit filter.main.js and use rollup to generate the dist file. See: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js-bundling.md

thx

tfamula:
Hi @Amstutz,
thanks for correction! Done!

PR #7692, opened 2024-06-20
Search: add LSO and all remote objects to search results (38957, 37813)

schmitz-ilias:
This PR fixes [38957](https://mantis.ilias.de/view.php?id=38957) and [37813](https://mantis.ilias.de/view.php?id=37813), by appending the array of allowed object types in search. For LSO everything seems to work fine, but I was not able to test the changes for the remote objects. There could be follow-up issues similar to [38422](https://mantis.ilias.de/view.php?id=38422).

As discussed previously, I will provide a slightly more ambitious PR for trunk, so this only needs to be picked up to 9.


smeyer-ilias:
Thanks @schmitz-ilias I tested this fix on an ECS installation. ECS remote objects are searchable and shown in the search result list.

PR #7694, opened 2024-06-20
Search: build the list of allowed types from object definition (37813, 38957)

schmitz-ilias:
As promised in #7692, here is the PR to build the list of allowed types in the search dynamically.

This has turned out more complicated than originally thought, because `ilObjectDefinition` sometimes gives out quite counterintuitive results: `ilObjectDefinition::isAllowedInRepository` e.g. returns true for `objf`, but false for `rcrs`. Even more strangely, this has behaved this way since at least 2011 (see 7cb31c9). Maybe I'm misunderstanding the semantics of `ilObjectDefinition` here, but this behavior makes it necessary to have a lot of additional safeguarding in place to really only get repository object types.

The result is pretty clunky, and I'm not quite sure whether this is preferable to the fixed list it replaces. Am I missing any possible simplifications?

In this state, the types it returns for me locally are 
```
bibl, blog, book, cat, catr, chtr, cld, cmix, copa, crs, crsr, dcl, exc, file, fold, frm, glo, grp, grpr, htlm, iass, itgr, lm, lso, lti, mcst, mep, prg, prgr, prtt, qpl, sahs, sess, spl, svy, tst, webr, wiki, rcat, rcrs, rfil, rglo, rgrp, rtst, rwik
```
I'm not sure how this behaves with plugins.

Please let me know if there are any improvements to be made, and whether you prefer this solution over just pulling up #7692 to trunk.

### DIFF
--- a/components/ILIAS/Search/classes/Like/class.ilAbstractSearch.php
+++ b/components/ILIAS/Search/classes/Like/class.ilAbstractSearch.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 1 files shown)


smeyer-ilias:
Thanks @schmitz-ilias !

PR #7696, opened 2024-06-21
MediaCast deleted rows in release 8 also delete in class

manuelgerber:
In il_media_cast_data was deleted several rows that not deleted in update and read methode.
Getter and setter are not deleted because they call from outher classes.  That produce other bugs.
That is the solution for calling pages with uploaded audio streams.

### DIFF
--- a/Modules/MediaCast/classes/class.ilObjMediaCast.php
+++ b/Modules/MediaCast/classes/class.ilObjMediaCast.php
@@ -225,9 +225,6 @@ class ilObjMediaCast extends ilObject
             ", def_access" .
             ", sortmode" .
             ", viewmode" .
-            ", autoplaymode" .
-            ", nr_initial_videos" .
-            ", new_items_in_lp" .
             " ) VALUES (" .
             $ilDB->quote($this->getId(), "integer")
             . "," . $ilDB->quote((int) $this->getOnline(), "integer")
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
@manuelgerber Thx for your PR. Could you please report the issue in Mantis (see: https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/contributing.md#rules-for-contributors )?

manuelgerber:
No problem @mjansenDatabay. Hier is the report in mantis.
https://mantis.ilias.de/view.php?id=41583

alex40724:
There seems to be something wrong with your DB. Looks like ILIAS 8 code is running on a ILIAS 7 table here. The fields are not deleted. They have been added with ILIAS 8. This is why the code is needed.

The class ilMediaCastDBUpdateSteps adds these field for ILIAS 8. Not sure what went wrong with your update here. We migrated lots of instances to ILIAS 8 and never had this issue.

manuelgerber:
I'm beginning to understand the mistake. This is once again due to a problem with the Ilias 7 to 8 update. The fields were not transferred during the DB migration. This happens again and again with my Ilias 7 to 8 updates. Some fields, maybe all of them, are not created correctly during the update. I don't know what the problem is.

alex40724:
Mmh, never occured to me. There might be something wrong with your procedure or some setup issue that only occurs under special circumstances. If this is reproducible when performing an update (starting from the same 7 backup), you should file a report for the setup component.

manuelgerber:
The update runs through without any problems. Perhaps there is a connection with an abort on the first attempt. For example, if the plugin compatible with the version was forgotten to be installed. If the correct version is installed and the update is restarted, it runs through without errors.

My assumption is that this interruption is due to a faulty DB migration.

PR #7697, opened 2024-06-21
ILIAS UI: add checked and unchecked to glyphs

iszmais:
This PR proposes to add the custom "checked" and "unchecked" to the glyphs, since both are very commonly used in a wide range of cases.

ATM the symbols can only be added via UI-Factory through a custom icon with...
-  a static path, which ignores skin interference
- `getImagePath()`, which is deprecated.

### DIFF
--- a/components/ILIAS/UI/src/Component/Symbol/Glyph/Factory.php
+++ b/components/ILIAS/UI/src/Component/Symbol/Glyph/Factory.php
@@ -687,7 +687,6 @@ interface Factory
      */
     public function angry(?string $action = null): Glyph;
 
-
     /**
      * ---
      * description:
@@ -1434,7 +1433,6 @@ interface Factory
      */
     public function link(?string $action = null): Glyph;
 
-
     /**
      * ---
      * description:
@@ -1562,7 +1560,6 @@ interface Factory
      */
     public function sort(?string $action = null): Glyph;
 
-
     /**
      * ---
      * description:
@@ -1584,7 +1581,6 @@ interface Factory
      * @return  \ILIAS\UI\Component\Symbol\Glyph\Glyph
      */
     public function columnSelection(?string $action = null): Glyph;
-
 
     /**
      * ---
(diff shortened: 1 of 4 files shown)


klees:
Hi @iszmais,

thanks for this proposal. We do not think that "checked" and "unchecked" are good candidates to be Icons in the sense of the UI framework. The description of Icon states: 

Icons are quickly comprehensible and recognizable graphics. They indicate the functionality or nature of the element they illustrate: Icons will mainly be used in front of object-titles, e.g. in the header, the tree and in repository listing.

"Checked" or "Unchecked" do not seem to indicate a functionality or nature of element, but instead illustrate some status. Although they might be included as Icons in some places currently (which somehow would violate our KS rules...), we think they would fit the Glyph component much better.

Further, as you mentioned, at the moment there is an icon in the image > milos folder for both "checked" and "unchecked" (and an x - we are not sure why), which are used in data tables, among other things. Including this status icon as a glyph would have the advantage that we could display unchecked and checked both changeable (can be changed via Actions > Edit entry) and unchangeable (cannot be edited) by controlling this via CSS (f.e. simple graying out because part of the font). 

So Please:
[ ] Propose as Glyph: Please propose those two legacy icons as Glyphs. You find some help around fonts here: https://github.com/ILIAS-eLearning/ILIAS/blob/release_9/templates/default/fonts/Iconfont/Readme.md. @yseiler offered to provide further assistance for working with ILIAS Fonts if needed.

Question:
[ ] Refactoring: When adding it as a glyph, we should search for the previous icons in the images > milos folder, replace them and remove the images from the images folder. Do you see any option to provide this refactoring as part of this PR? Would you need additional funding for this.

Please move accordingly and reopen a new PR.

Thanks!
@Amstutz, @thibsy, @klees

klees (changes requested):
:point_up:

iszmais:
Thank you for your feedback @klees ,

I totally see the issue in proposing this as an icon and agree with that.

However, I came across some criteria why I think this isn't suitable as a glyph either:
- Glyphs are conventionally bond to an action (like up, down, close, back, forward) and also semantically display actions. As you said, the proposed icons are majorly used to display a status and therefore imply to be without any action.
- Their usage is widely spread and ambiguous. While a "Close" Glyph means close in every context, a "Check-mark" Glyph can have multiple meanings depending on its context. (e.g. a check-mark inside the T&A mostly means finished/approved/completed while a check-mark inside the administration mostly means set/active/enabled)
- Currently, those are implemented as icons and therefore are fully customisable via the skin. This is an important factor since these icons often display important key points for accessibility. An implementation as glyph would hardly stricten the possibility of customization.

In conclusion, I'm having a struggle with both implementations. Therefore, I would suggest implementing an additional scope with the symbols to provide a place for such "icons" (since I assume this won't be the only ones). Perhaps something accessible as `ui()->factory()->symbol()->icon()->abstract()`.

Pls let me know your opinion on that.

Greetings,
@iszmais

iszmais:
@kjansen2608 FYI

iszmais:
Meeting Protocol with @klees :
"Glyphs are not obliged to provide any action anymore. Therefore, the consideration of glyph for status representation does not violate their concept. I general glyph should provide abstract (not ILIAS specific) symbolism, which is the case here"

Thx for the feedback!
I will change the PR to provide the icons as a glyph

Greetings,
@iszmais

iszmais:
@klees Pls have another look. I will follow soon with an implementation PR.

klees:
Hi @iszmais,

this is good for the JF.

Thanks!

matthiaskunkel:
Jour Fixe, 22 JUL 2024: PR is highly appreciated and accepted for trunk.

klees:
Hi @iszmais,

please ping me once implementation is ready.

Thanks!

iszmais:
EDIT: Within the integration of the new item we stumbled upon a correlation with an already existing glyph within the il-icon font with the name "**check**".

Therefore, we adapted the naming of the here integrated glyphs to **checked** (instead of **check**) and **unchecked** (instead of **uncheck**)

Greetings,
@iszmais

iszmais:
@klees the implementation is ready. Pls have a look.

(The glyphs are introduced [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9521))

Greetings,
@iszmais

matthiaskunkel:
Jour Fixe, 12 MAY 2025: Ingmar notified us about the modified titles of the glyphs to distinguish these glyphs from the check icon.

Amstutz:
Thx a lot, note that this is merged along with the implementation here https://github.com/ILIAS-eLearning/ILIAS/pull/7722/files

PR #7699, opened 2024-06-24
41582: Fix uninitialized array key access in booking process gui

mBeym:
Fixes: https://mantis.ilias.de/view.php?id=41582


alex40724:
Thanks for the PR!

PR #7702, opened 2024-06-24
fix type error

ingoj:
possible fix for mantis 41587

$ref_id was already typecasted to int and null-checked, si I used it in line 616 instead of $id[0] to fix type error

Copyright notice is adjusted, too


smeyer-ilias:
Thanks @ingoj 
I 've already fixed this some days after your PR:
https://github.com/ILIAS-eLearning/ILIAS/commit/2033857d520725e47ec2a427da548dcc0576eb39

PR #7706, opened 2024-06-25
T&A Bugfix #0027980: Re-Import of test with skills fails

fhelfer:
https://mantis.ilias.de/view.php?id=27980


kergomard:
I fixed this a little differently with https://github.com/ILIAS-eLearning/ILIAS/commit/933bee57034e14941391e5e33bdf90ddc3d42731. Your approach would have introduced a clear side effect (not that it would have been the only one here) and I don't think this is clean.
Sorry I did this directly! It was just such a small fix.

Best,
@kergomard

PR #7712, opened 2024-06-25
DataCollection: Remove legacy table from Table "Tables"

iszmais:
This is part of "Removing LUI".

Removes the legacy table "Tables" from the DataCollection and replaces it with an item list of UI-Framework-Components.

clipboard-202407011232-hjbf7


iszmais:
@oliversamoila FYI

PR #7713, opened 2024-06-26
TA 41608 Bugfix: Apply default setting to test when extra confirmation is needed didn't work

malaja11:
When a user wants to apply a default setting which uses "Fixed Questions" to a test that uses "Random Questions" or vice versa, all questions must be deleted. Therefore the user is asked for a confirmation.
The problem was that during this process, the id of the default setting was not saved correctly, so that when the user confirmed, the selected setting is not applied. Instead, there was an error message "please select a default setting".

I fixed this bug by saving the id during the forwarding process.

cc @thojou

### DIFF
--- a/Modules/Test/classes/class.ilObjTestGUI.php
+++ b/Modules/Test/classes/class.ilObjTestGUI.php
@@ -2478,12 +2478,15 @@ class ilObjTestGUI extends ilObjectGUI implements ilCtrlBaseClassInterface, ilDe
      */
     public function applyDefaultsObject($confirmed = false)
     {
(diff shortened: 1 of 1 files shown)


dsstrassner:
Thank you, @malaja11 - there is no Mantis report at the time being, or do I miss something?
Please create one, so I can add the bug fix-label.

malaja11:
No, sorry, I forgot to create one, but here it is now: https://mantis.ilias.de/view.php?id=41608

thojou:
Hey @malaja11 

I took you PR for release_9.

Tank you!

PR #7714, opened 2024-06-26
UI: JS for dropdown

klees:
Hi everyone,

this [fills in some missing js-code](https://mantis.ilias.de/view.php?id=41580).

@thibsy: This currently violates some rules from our js-coding-style. Could you be so kind and advice me how to fix these? Also: This is missing js-unit tests. Since this is 99% side effects, I find it hard to think of useful tests here. Any idea?

Kind regards!


thibsy (changes requested):
Hi @klees

Thanks a lot for your contribution to the UI components and for filling in the missing JavaScript!

For checking/fixing the code-style using ESLint you can generally follow the [instructions in the code-style document](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/js-coding-style.md#applying-code-style). However, there are some issue which ESLint cannot fix automatically.

Please consider the following suggestions. You do not need to follow them, they are highly optional:

- [x] `querySelectorAll()`: you are currently retrieving all buttons and lists of the dropdown element inside the constructor. You later on check if there is exactly one of each element. Is this actually necessary, or could we simply use [`querySelector()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector) and take the first element (if any)? This would also get rid of the weird array destructuring :).
- [x] `function` properties: you are assigning an arrow function to the `#hideOnClick`, `#showOnClick` and `#hideOnEscape` properties, so the property can be supplied as an event handler. This works, but IMO it is cleaner to implement them [as private methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties), which makes documenting them using JSDoc a lot easier too (e.g. `@type {function(type, ...): type}` vs `@param`, `@return`, `@throws`).
- [x] `Document` injection: you inject the `Document` into the constructor, which is actually not necessary because we can access the property using [`HTMLElement.ownerDocument`](https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument).
- [x] Global namespace: we don't have established some convention for this, but IMO we should populate the dropdown business logic under `il.UI.dropdown.init` or similar, not directly as `dropdown`.
- [x] File names: the code-style actually states, that file names should exactly match the default export of a file, so please rename `dropdown.class.js` to `Dropdown.js` and due to conflicts `dropzone.js` to `index.js` or similar.

- [x] Event listeners: currently you mix up two approaches of where an event listener is implemented. E.g. the "enter" listener is implemented inside the `addEventListener()` statement, but the "escape" listener is implemented in a dedicated property. Please go in either of these two directions for all listeners.
- [x] JSDoc annotations: you are missing some annotations for the concrete [`KeyboardEvent` events](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent). Please add them.
- [x] `DOMDocument`: please change these JSDoc annotations to `HTMLDocument` or simply `Document`. I think [`DOMDocument` comes from PHP](https://www.php.net/manual/en/class.domdocument.php), but JavaScript doesn't know this object type.
- [x] Copyright headers: there seems something wrong with the copyright license header in some files. Could you look into this?

Kind regards,
@Amstutz and @thibsy

klees:
Moin @thibsy,

thanks for the review. Some remarks:

* `function` properties: If I turn these to actual functions, `this` does not point to the class instance anymore once we are in the event handling. Hence `this.show is not a function` is produced as an error on clicking etc.
* `Document` injection: I think we actually *should* use the document from the element, because we want to mount the event handler on the document that contains the button.

The rest should be as requests.

Kind regards!

thibsy (changes requested):
Hi @klees,

Thx a lot for incorporating all of the changes, especially the optional ones! Also, good point about using the associated `Document` of the button - we should definitely keep this in mind for future developments!

Before we can merge this, I still have some minor changes regarding JSDoc:

- [x] `@param` annotations: since you are using [private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties#private_fields), you need to document the assigned value like `@type {function(KeyboardEvent)}`, so code completion can pick it up correctly.
- [x] Obsolete annotation: please remove the `DOMDocument` annotation from the constructor, we don't need it anymore.

About the `function` properties: it is true that `this` would be bound incorrectly, but you could provide an inline arrow-function which delegates the call to `this` without loosing scope. This is also done in [the markdown editor](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/UI/resources/js/Input/Field/src/Markdown/markdown.class.js#L64-L82).

Kind regards,
@thibsy

klees:
Hi @thibsy,

both done.

We cannot use the approach from the markdown editor here because it doesn't allow to remove event handlers later on. This is why the named functions are required in the first place...

Kind regards!

nhaagen:
Hi @klees ,
we need that on the ViewControls as well: https://github.com/klees/ILIAS/pull/62

thibsy (approved):
Hi @klees,

> We cannot use the approach from the markdown editor here because it doesn't allow to remove event handlers later on. This is why the named functions are required in the first place...

Thx for clarifying, I did not catch that. IMO this is good to be merged now.

Kind regards,
@thibsy

PR #7715, opened 2024-06-27
UI, DataTable: do not order for one record

nhaagen:
https://mantis.ilias.de/view.php?id=41513

### DIFF
--- a/components/ILIAS/UI/src/Implementation/Component/Table/Data.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Table/Data.php
@@ -146,7 +146,7 @@ class Data extends AbstractTable implements T\Data
     {
         $view_controls = [
             self::VIEWCONTROL_KEY_PAGINATION => $this->getViewControlPagination($total_count),
-            self::VIEWCONTROL_KEY_ORDERING => $this->getViewControlOrdering(),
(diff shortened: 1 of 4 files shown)


nhaagen:
Here's the fix for the failing test: https://github.com/ILIAS-eLearning/ILIAS/pull/7716

klees (changes requested):
Hi @nhaagen,

is this enough? Does this also remove the little arrows from the headers of the rows?

Kind regards!

nhaagen:
> is this enough? Does this also remove the little arrows from the headers of the rows?

yes, the headers are built with the internal signal from the ViewControl.
if it's not there - which will be the case with a null-control  - the headers are rendered without any link:

https://github.com/ILIAS-eLearning/ILIAS/blob/1f47355707b6d7814fe663240ef860f90ee2f354/components/ILIAS/UI/src/Implementation/Component/Table/Renderer.php#L296

klees (approved):

klees:
@Amstutz/@thibsy: This LGTM. Please have a look as well =)

Thanks!

Amstutz:
@nhaagen

thx a lot for this neat improvement.

one detail:
- [x] $total_cout: make it mandatory in getViewControlOrdering

thx a lot,
@thibsy and @amstutz

Amstutz:
thx a lot!

PR #7717, opened 2024-06-28
UI/Button: allow Glyphs as label #6922

nhaagen:
was (for 9): #6922

Allow Glyphs as labels for Buttons. In a second step, Actions will be removed from Glyphs.

This sharpens the separation and essential character of both components: Glyphs represent a concept, while Buttons trigger actions.


klees (approved):

klees:
Hi @thibsy and @Amstutz,

please go on here, LGTM.

Kind regards!

nhaagen:
taking up from #6922, there still was an issue with missing/double labels, especially in bulkies:
> - [ ] A11y: Currently there are A11y issues possible. E.g. if a Label is given along with a Glyph as a symbol with the same aria label (would be considered a redundant label). However, simply matching it to the bulky buttons implementation could cause an issue if no label is given at all (since the Glyph renderer of Bulkies does not render an aria label). This seems to be a bit of a can of worms. Maybe we should just always require a label to be set for buttons (and perhaps not always show it, e.g. if space is scarce)?

Please see examples and tests in this PR, e.g. the exmaple of Bulky Buttons
Screenshot from 2024-08-14 15-11-29

will now read 
- 'Icon, button', 
- 'Glyph, button'
- 'Background Tasks, button' (because "Background Tasks" is the translation for the briefcase aria label)
- 'Example, button' (since "Example" was given as alt to the icon)

Amstutz:
thx @nhaagen this seems ready for jf now.

matthiaskunkel:
Jour Fixe, 16 SEP 2024: We highly appreciate this suggestion and accept the PR for ILIAS 10.

PR #7719, opened 2024-06-28
[FIX] #41560 UI: Footer Link Colours

thibsy:
Hi folks,

This PR introduces two less variables `@il-footer-link-color` and `@il-footer-link-hover-color` which can be used to style links differently inside the footer. I defaulted them to the `@il-link-color` and `@il-link-hover-color` variables. This solutions is needed to address [a mantis issue](https://mantis.ilias.de/view.php?id=41560), which describes contrast issues with dark footer background colours.

While adding these variables I also noticed, the `@il-link-hover-color` was never assigned, so I have added this with the PR as well.

Kind regards,
@thibsy

### DIFF
--- a/src/UI/templates/default/MainControls/footer.less
+++ b/src/UI/templates/default/MainControls/footer.less
@@ -24,7 +24,10 @@ footer{
 			display: inline-block;
 			margin-right: @il-margin-xs-horizontal;
 			a, button {
-				color: @il-link-color;
+				color: @il-footer-link-color;
+			}
+			a:hover, button:hover {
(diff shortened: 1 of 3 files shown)


Amstutz:
Hi @thibsy 

Thx a lot. I accept this for 8. However, for 9 and beyond we switched to SCSS and we heavily reduced to amount of publicely changeable variables. Could you provide a seperate PR for 9 in which we along with @catenglaender discuss if we want to take this new variables also to the reduced settings set?

Thx a lot!

PR #7720, opened 2024-06-28
Init: Fix opening a permanent link from a wiki leads to error message

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41618


mjansenDatabay:
I'll, merge this, because it is a trivial port from `release_9`.

PR #7724, opened 2024-06-28
UI/Input/Numeric: add min-width so arrows and number are always visible

nhaagen:
https://mantis.ilias.de/view.php?id=41669


klees (approved):
LGTM in general, I have no idea if matches how these widths should be handles with ITCSS, though.

klees:
@catenglaender Could you look into this? From my POV this is mostly some CSS issue, hence I assigned you and @Amstutz.

Amstutz:
For me good as well. @catenglaender feel free to merge, if fine with you.

catenglaender:
Long term we should investigate if using `ch` as a unit for input field width might be something we would like to use and maybe a tool could manage different width patterns. For now, it's totally in line with the other input fields.

PR #7725, opened 2024-06-28
UI/DataTable: do not double actions with same key for multi-dropdown

nhaagen:
Actions happend to appear more than once in the MultiAction Dropdown.

### DIFF
--- a/components/ILIAS/UI/src/Implementation/Component/Table/AbstractTable.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Table/AbstractTable.php
@@ -188,7 +188,7 @@ abstract class AbstractTable extends Table implements JSBindable
      */
     public function getAllActions(): array
     {
-        return array_merge($this->actions_single, $this->actions_multi, $this->actions_std);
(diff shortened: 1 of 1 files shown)


klees:
@nhaagen Is there any ticket in mantis describing the faulty behaviour?

klees:
Hi @nhaagen,

is there a ticket for this?

Kind regards!

klees:
Hi @thibsy and  @Amstutz,

ticket or not, the change looks plausible. Please have a look as well...

Kind regards!

Amstutz:
thx @nhaagen , can you 

- [ ] add a unit test, making sure, this issue is not introduced in future refactorings as well? 

if done, this is ready to be merged.

nhaagen:
I cannot really reproduce this anymore; if you give the actions a key, they are not doubled, and if you don't and add an action as both single and multi, e.g., it's either a mistake or on purpose. 
However, theres hardly a good way to compare actions besides their label, and that's too weak. 
I'll close.

PR #7726, opened 2024-06-28
Docs: add tutorial for setup

klees:
Hi @acgruber,

I would appreciate some feedback if this is what we are aiming at from the perspective of style and information density. I will try to find someone that looks into this from the perspective of the content.

Kind regards!

### DIFF
--- a/docs/development/devguide/tutorial/02-tools/03-setup.md
+++ b/docs/development/devguide/tutorial/02-tools/03-setup.md
@@ -1,3 +1,119 @@
-This file is stored here to show the tutorial's future structure.
+# The ILIAS Setup 
 
-Its content is currently work in progress and will be available asap.
+## What is This About?
+
+The ILIAS Setup makes sure that ILIAS can be run on your system. Only downloading
+ILIAS' code won't be enough. There are various things that need to be checked and
+prepared to make your system actually run ILIAS: the environment is checked for
+required dependencies, folders and files need to be created from the raw sources,
+the database needs to be installed and filled with the settings for your installation.
+
+This guide briefly explains the necessary steps to install ILIAS and gives pointers
+to additional resources for the actual installation and to understand the ILIAS setup
+in depth.
+
+## Why do I Need This in ILIAS?
+
+If you want to get involved with ILIAS in general or with coding in ILIAS specifically:
+you won't be able to achieve anything without a running ILIAS installation. Setting
+up a local installation will help tremendously when working on ILIAS, even if you just
+fix a simple bug.
+
+If you want to code elaborate features with ILIAS you will likely want to hook into
+the mechanism of the Setup at some point. This is where an more in-depth understanding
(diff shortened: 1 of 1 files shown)


docs/development/devguide/tutorial/02-tools/03-setup.md
```diff
@@ -1,3 +1,119 @@
...
+ILIAS requires to function. Move to the ILIAS folder
+
+```sh
+cd ILIAS
+```
+
+and start by installing the JavaScript dependencies:
+
+```sh
+npm clean-install
```
PurHur: Please add --ignore-scripts

acgruber:
Hi @klees,

This is exactly what we're aiming for both in terms of style and density. Great example of how to use enganging language to guide a newbie through an article as well. Thanks a lot for this contribution!

Cheers!

docs/development/devguide/tutorial/02-tools/03-setup.md
klees: Done. Thanks!

docs/development/devguide/tutorial/02-tools/03-setup.md
```diff
@@ -1,3 +1,119 @@
...
+
+## What do I Need to Watch Out For? (Dos & Dont's)
+
+* Please be aware that exposing a web application to the internet is not something
+  to be done lightly. This tutorial covers the basics to create a running installation
+  on a local system for development purpose. Make sure to configure your system
+  appropriatly to make sure that no one illegitematly uses your local installation.
+* Certain changes to the code will require to rebuild artifacts. If you e.g. create
+  new GUI classes or install a plugin, parts of the artifacts will need to be
+  refreshed. If in doubt, run `composer install`. This will rebuild static artifacts
```
tfamula, line 118: Is there a reason why you are using `composer install` here and not `composer dump-autoload`?

klees: I think `composer install` is more general in the sense that it does all the required activities to get required code in place. People would need to understand what has changed in some code update to understand if `dump-autoload` is enough or `install` is required. Would not want to go there in this tutorial. Is there something where `install` would have a disadvantage over `dump-autoload`?

tfamula: Not really, I was asking more out of curiosity. There are only two really small ones which come to my mind. `composer dump-autoload` is a bit faster and it is not that strict regarding PHP versions. When switching between different ILIAS versions locally, I have to pay more attention using the right PHP version when executing `composer install`. But your argument prevails here of course – thanks for the explanation!

klees: Thanks for the feedback =)

acgruber:
Merged into trunk. Thanks again @klees for the contribution and also @PurHur and @tfamula for your inputs!

PR #7732, opened 2024-07-01
DataCollection: Remove legacy table from Table "Views"

iszmais:
This is part of "Removing LUI".

Removes the legacy table "Views" from the DataCollection and replaces it with an item list of UI-Framework-Components.

image


iszmais:
@oliversamoila FYI

oliversamoila:
Thank you @iszmais , 
I think that's a really good solution. We can take another look at the labeling of the properties and their status. 
Thank you very much.

iszmais:
We might consider striping the entry in such lists (this is possible now).
But I would do that on demand since I do not expect such lists to have more than a few entries in most cases.

PR #7733, opened 2024-07-01
Maps: 36490, refresh map until rendered

nhaagen:
https://mantis.ilias.de/view.php?id=36490


klees:
There are failing tests in some class that seems to be unrelated. I reported the issue as https://mantis.ilias.de/view.php?id=41689 but still merge this to move on with this issue.

klees:
I have merged this for release_9. @nhaagen Could you create another fix for trunk? Cherry-pick was not straightforward =(

PR #7739, opened 2024-07-02
T&A: 37677, Fix wrong order of units in formula question

matheuszych:
When creating a formula question you can use units and select a sequence of these units.
If the sequence of these units is changed, the user is still shown the orginal creation order.

I believe this is a valid solution for the issue.

[Mantis: 37677](https://mantis.ilias.de/view.php?id=37677)


thojou:
Hey @matheuszych,

I took you PR and integrated the changes into release_9 and trunk.

Thank you.

PR #7742, opened 2024-07-02
Add Useraccount to TA Testers

dsstrassner:
I have added the Docu.ilias.de Account of one of the Test & Assessment testers to the maintenance.md to correct the entry here and on the docu.ilias.de.


kergomard:
Thank you very much @dsstrassner !

PR #7743, opened 2024-07-02
User: Fix building query when using "Custom User Field" filter

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=29032

If approved, please pick this to other maintained releases if necessary.


kergomard:

I went for another solution, as I had the feeling that this fix was tackling symptoms not causes. I hope I didn't miss anything.

Thanks again and best,
@kergomard

PR #7745, opened 2024-07-02
UI/DataTable (8): #38099, pass additional parameters

nhaagen:
https://mantis.ilias.de/view.php?id=38099


klees:
Picked to release_9, seems to be fine in trunk already.

PR #7746, opened 2024-07-02
[FIX] UI #37494: distinguish and apply array-parameters (#6279)

nhaagen:
https://mantis.ilias.de/view.php?id=38100

* UI, URLBuilder: distinguish and apply array-parameters
* UI, URLBuilder: store array-params as array, repeat on output
* UI, URLBuilder: align PHPparts to not index array-params
* UI, URLBuilder: do not encode brackets of array-params
* UI, URLBuilder: add some docu concerning array-params

### DIFF
--- a/src/UI/templates/js/Core/dist/core.js
+++ b/src/UI/templates/js/Core/dist/core.js
@@ -162,6 +162,93 @@ var il = il || {};
      ********************************************************************
      */
 
+
+    const URLBuilderTokenSeparator = '_';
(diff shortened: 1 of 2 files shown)


klees:
@nhaagen Explained to me that he has picked to required commits, this is what's left.

klees:
I'll close this, since ILIAS 8 is in the security phase now. I did not receive any actual bug reports that request these changes. Does not seem to matter much...

PR #7750, opened 2024-07-03
[FIX] `ilTreeAdminNodeAddedObjective` database instance

thibsy:
Hi folks,

I noticed trunk cannot be fresh installed at the moment due to a failed setup objective `ilTreeAdminNodeAddedObjective`. The issue occurs because the database instance is not properly populated in the total abomination (`$DIC`) and cannot be service-located when creating an instance of the `ilTree` using `$DIC->database()` somewhere.

I fixed this by populating an instance of the pimple container and isolate it for this objective only, by reseting the state after the objective is achieved. The same thing is done in many other objectives.

Kind regards,
@thibsy

### DIFF
--- a/components/ILIAS/Tree/classes/Setup/class.ilTreeAdminNodeAddedObjective.php
+++ b/components/ILIAS/Tree/classes/Setup/class.ilTreeAdminNodeAddedObjective.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -18,8 +16,11 @@ declare(strict_types=1);
  *
  *********************************************************************/
 
+declare(strict_types=1);
+
 use ILIAS\Setup;
 use ILIAS\Setup\Environment;
+use ILIAS\DI\Container;
 
 class ilTreeAdminNodeAddedObjective implements Setup\Objective
 {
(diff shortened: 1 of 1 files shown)


mjansenDatabay (commented):
Hi @thibsy ,

thx for your pull request. Using a simple array for the `$DIC` lead exactly to issues like https://mantis.ilias.de/view.php?id=41574

Full Discussion: https://github.com/ILIAS-eLearning/ILIAS/pull/7674

@klees stated:

> During the setup there is no $DIC and I do not think that we should or even can have one. 

So my questions are:

- Do we really need the `$DIC` here? If not, could we get rid of it completely in this class?
- If we still need the `$DIC`, should we really fall back to an empty array?

Best regards,
Michael

thibsy:
Hi @mjansenDatabay

Thx for looking into this. I understand why we don't want to have a global `$DIC` anywhere, however, addressing this issue using proper DI would have meant a somewhat larger refactoring here. The problem is, that `ilTree` creates an instance of some logger using the `ilLoggerFactory` singleton, which does not provide an option to inject the database. Retrieving the logger will also at some point create an instance of `ilSetting`, which has the same problem again.

Therefore I only changed the null-coalesce to assign `null` instead of an empty array now.

Kind regards,
@thibsy

klees:
Hi both,

I think we should also wrap that code in a `try/finally` block to make sure, that the original `$DIC` is restored in every case.

The thing that @thibsy mentions is the exact cause for the "abomination" in some objectives. Would be really nice if we could gently move the system towards more dependency injection to get rid of these...

Kind regards!

thibsy:
@klees thx for your input. Makes sense, done.

mjansenDatabay:
LGTM now, thanks @thibsy and @klees for fixing/providing feedback.

mjansenDatabay (approved):

PR #7751, opened 2024-07-03
Refactor test base

iszmais:
Add partial refinery and remove broken/trivial `use` calls

### DIFF
--- a/components/ILIAS/UI/tests/Base.php
+++ b/components/ILIAS/UI/tests/Base.php
@@ -23,7 +23,6 @@ require_once(__DIR__ . '/../../Language/classes/class.ilLanguage.php');
 
 use ILIAS\UI\Component as C;
 use ILIAS\UI\Component\Component as IComponent;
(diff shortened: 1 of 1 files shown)


iszmais:
@Amstutz @klees 

I assume the refinery will be used more often in the test in future (just like [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7573)) and therefore should be functional inside them.

Greetings,
@iszmais

klees:
Hi @iszmais,

I'm sorry that you have invested some of your time here and it now will be spend in vain, but @thibsy and I had a deeper look into the issue only after we saw this additional change. We wondered: Why would we need the complete Refinery in all our UI testcases? And it turns out: We really don't... The only original reason we had used the `Refinery` before was to turn an `URI` into a string. But we already have a method to do so. Hence we want to remove the dependency to the `Refinery` from the complete rendering of the UI framework for good: #7785 

Hence I close this PR.

Thanks for your understanding.

PR #7752, opened 2024-07-03
9/UI/metabar slates not visible in Safari 40968

catenglaender:
https://mantis.ilias.de/view.php?id=40968

# Issue

(Some) Metabar slides do not appear on mobile breakpoint in Safari/Webkit.

metbar-slate-not-visible

# Change

The issue seems to be caused by `overflow-x: clip;` on the header, which no longer seems to be needed as currently no unwanted scrollbars seem to form. Strangely, `overflow-x: clip;` is actually supported by Safari/Webkit, just something along the chain of `position: sticky;` and `position: absolute;` causes a broken rendering (y expansion also gets clipped).

metbar-slate-visible-again

# Impact

Removing `overflow-x: clip;` which was likely introduced to combat an unwanted scrollbar might cause said scrollbar to show up again under conditions that I can currently not reproduce. If there now is an unwanted scrollbar appearing, it's probably wise to directly fix the cause of the scrollbar rather than to clip the header.

### DIFF
--- a/templates/default/070-components/UI-framework/Layout/_ui-component_standardpage.scss
+++ b/templates/default/070-components/UI-framework/Layout/_ui-component_standardpage.scss
@@ -316,7 +316,6 @@ footer {
 	header {
 		position: -webkit-sticky;
 		position: sticky;
-		overflow-x: clip;
 		width: 100%;
 		top: 0;
 	}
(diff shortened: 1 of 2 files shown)


Amstutz:
thx a lot for fixing this and for providing the detailed analysis. highly appreciated!

oliversamoila:
Thank you so much !!!  🎉 😊

PR #7753, opened 2024-07-03
Fix copyright checker for examples

iszmais:
https://mantis.ilias.de/view.php?id=41651

This PR fixes the `copyright-checker.sh` to enforce explicitly no copyright inside the examples and removes current copyrights.

### DIFF
--- a/scripts/Copyright-Checker/copyright-checker.sh
+++ b/scripts/Copyright-Checker/copyright-checker.sh
@@ -62,20 +62,35 @@ function is_copyright_valid() {
   local file_extension="${file##*.}"
   local offset=1
 
-  # since PSR-12 the php files will contain the copyright license as
-  # document-level comment, which starts on line 3.
-  if [ "php" = "${file_extension}" ]; then
-    offset=3
-  fi
+  is_ui_example "${file}"
+  local is_example=${?}
 
-  for copyright_line in "${COPYRIGHT_LINES[@]}"; do
-    local line_to_check="$(sed "${offset}q;d" "${file}")"
-    if ! [ "${copyright_line}" = "${line_to_check}" ]; then
-      return 1
+  if [ 1 -eq ${is_example} ]; then
+    if [ "php" = ${file_extension} ]; then
+      offset=3
     fi
+    for copyright_line in "${COPYRIGHT_LINES[@]}"; do
+      local line_to_check="$(sed "${offset}q;d" "${file}")"
+      if ! [ "${copyright_line}" = "${line_to_check}" ]; then
+        return 1
+      fi
 
-    offset=$((1 + ${offset}))
-  done
+      offset=$((1 + ${offset}))
+    done
+  else
+    local lines=$(wc -l < "${file}");
+    while [ ${offset} -lt ${lines} ]
+    do
+      local line_to_check="$(echo "$(sed "${offset}q;d" ${file})" | xargs)"
+      if echo "$line_to_check" | grep -q "^namespace "; then
+        return 0
+      fi
+      if echo "$line_to_check" | grep -q -e '^/\*' -e '^//' -e '^#'; then
+        return 2
+      fi
+      offset=$((1 + ${offset}))
+    done
+  fi
 
   return 0
 }
@@ -156,15 +171,12 @@ function perform_copyright_check() {
     is_copyright_valid "${file}"
     local is_valid="${?}"
 
-    is_ui_example "${file}"
-    local is_example="${?}"
-
-    # invert the copyright-check for UI examples, because we
-    # don't want them to have too much content.
-    if ([ 0 -eq ${is_example} ] && [ 0 -eq ${is_valid} ]) ||
-      ([ 1 -eq ${is_example} ] && [ 1 -eq ${is_valid} ]); then
+    if [ 1 -eq ${is_valid} ]; then
       printf "copyright is not as expected in %s\n" "${file}"
       exit_status=1
+    elif [ 2 -eq ${is_valid} ]; then
+      printf "copyright is not allowed in %s\n" "${file}"
+      exit_status=2
     fi
   done
 
--- a/components/ILIAS/UI/src/examples/Card/RepositoryObject/base.php
+++ b/components/ILIAS/UI/src/examples/Card/RepositoryObject/base.php
@@ -4,8 +4,6 @@ declare(strict_types=1);
 
 namespace ILIAS\UI\examples\Card\RepositoryObject;
 
-/* Copyright (c) 2016 Timon Amstutz <timon.amstutz@ilub.unibe.ch> Extended GPL, see docs/LICENSE */
-
 function base()
 {
     //Init Factory and Renderer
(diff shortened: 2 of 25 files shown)


iszmais:
@lscharmer 

with your advanced expertise on UNIX-shell-scripts, I would highly appreciate if you have a look at the file `scripts/Copyright-Checker/copyright-checker.sh`.

Greetings,
@iszmais

scripts/Copyright-Checker/copyright-checker.sh
```diff
@@ -62,20 +62,35 @@ function is_copyright_valid() {
   local file_extension="${file##*.}"
   local offset=1
 
-  # since PSR-12 the php files will contain the copyright license as
-  # document-level comment, which starts on line 3.
-  if [ "php" = "${file_extension}" ]; then
-    offset=3
-  fi
+  is_ui_example ${file}
```
lscharmer:
```suggestion
  is_ui_example "${file}"
```

If you don't include the quotes this will exapnd into multiple arguments if the `$file` variable contains spaces.
E.g:
```sh
foo()
{
   echo ">$1<, >$2<, >$3<"
}

file='a b c'
foo $file
 # >a<, >b<, >c<
foo "$file" 
# >a b c<, ><, ><
```

scripts/Copyright-Checker/copyright-checker.sh
```diff
@@ -62,20 +62,35 @@ function is_copyright_valid() {
...
+    for copyright_line in "${COPYRIGHT_LINES[@]}"; do
+      local line_to_check="$(sed "${offset}q;d" "${file}")"
+      if ! [ "${copyright_line}" = "${line_to_check}" ]; then
+        return 1
+      fi
+
+      offset=$((1 + ${offset}))
+    done
+  else
+    local lines=$(wc -l < ${file});
```
lscharmer:
```suggestion
    local lines=$(wc -l < "${file}");
```
Same as 1. comment.

scripts/Copyright-Checker/copyright-checker.sh
```diff
@@ -62,20 +62,35 @@ function is_copyright_valid() {
...
+    while [ ${offset} -lt ${lines} ]
+    do
+      local line_to_check="$(echo "$(sed "${offset}q;d" ${file})" | xargs)"
```
lscharmer, line 84:
```suggestion
      local line_to_check= "$(sed "${offset}q;d" ${file})"
```
The `xargs` command seems unnecessary. It would convert newlines to spaces but there are no newlines because we picked a single line with `sed`.
Also in line 73 we do the same thing and it is also used without xargs:
```sh
73: local line_to_check="$(sed "${offset}q;d" "${file}")"
```

scripts/Copyright-Checker/copyright-checker.sh
```diff
@@ -62,20 +62,35 @@ function is_copyright_valid() {
...
+      if [[ "${line_to_check}" == "namespace "* ]]; then
```
lscharmer:
```suggestion
      if [[ "${line_to_check}" == "namespace "* ]]; then
      if echo "$line" | grep -q "^namespace "; then
```

This is just a suggestion, please just close this if you don't agree:
1. `[[` is bash specific. For a more POSIX compliant shell script I would remove this.
2.  `grep` is more known than `[[`
3.  Avoid confusion between `[` and `[[` and only use `[` (which s POSIX compliant).

Contra:
Usage of other commands than `[` or `[[` in an `if`, `while` or other blocks is not so common.

scripts/Copyright-Checker/copyright-checker.sh
```diff
@@ -62,20 +62,35 @@ function is_copyright_valid() {
...
+        return 0
+      fi
+      if [[ "${line_to_check}" == "/*"* || "${line_to_check}" == "//"* || "${line_to_check}" == "#"* ]]; then
```
lscharmer:
```suggestion
      if echo "$line" | grep -q -e '^/\*' -e '^//' -e '^#'; then
```
Same as above, please ignore if you disagree.

scripts/Copyright-Checker/copyright-checker.sh
```diff
@@ -156,15 +171,11 @@ function perform_copyright_check() {
...
-
-    # invert the copyright-check for UI examples, because we
-    # don't want them to have too much content.
-    if ([ 0 -eq ${is_example} ] && [ 0 -eq ${is_valid} ]) ||
-      ([ 1 -eq ${is_example} ] && [ 1 -eq ${is_valid} ]); then
+    if [ 1 -eq ${is_valid} ]; then
       printf "copyright is not as expected in %s\n" "${file}"
       exit_status=1
+    elif [ 2 -eq ${is_valid} ]; then
+      printf "copyright is not allowed in %s\n" "${file}"
```
lscharmer, line 178:
```suggestion
      printf "copyright is not allowed in %s\n" "${file}"
      exit_status=2
```
Missing exist status. Or should this error not cause the whole script to return an error?
You can also simply call `exit_status="$is_valid"` and remove both lines from the `if` blocks (like I suggested in the previous comment with the `case` example)

scripts/Copyright-Checker/copyright-checker.sh
```diff
@@ -62,20 +62,35 @@ function is_copyright_valid() {
...
-    if ! [ "${copyright_line}" = "${line_to_check}" ]; then
-      return 1
+  if [ 1 -eq ${is_example} ]; then
+    if [ "php" = ${file_extension} ]; then
+      offset=3
     fi
-
-    offset=$((1 + ${offset}))
-  done
+    for copyright_line in "${COPYRIGHT_LINES[@]}"; do
```
lscharmer, line 72: Please ignore this comment if you think is is out of the scope of this PR, I just want to point this out for completeness:
`$foo[@]` is a bash specific feature I would change the `COPYRIGHT_LINES` to
``` sh
COPYRIGHT_LINES=`cat <<EOF
/**
...
EOF`
```
and then
```suggestion
    echo "$COPYRIGHT_LINES" | while read copyright_line; do
```

This would also have the benefit that the license header is written literally in the file and makes it easy to copy from.

lscharmer (commented):
Thank you for your trust in my POSIX shell script skills ;)

matthiaskunkel:
Jour Fixe, 08 JUL 2024: Ingmar notified us about the fixed copyright checker for examples. Change is highly appreciated and accepted for trunk. It would be great if a developer from the community could check the shell code script and give feedback.

scripts/Copyright-Checker/copyright-checker.sh
iszmais: `xargs` is used here to trim the line to also match docs that are indented.
Feel free to propose a better solution for this.

iszmais:
Thank you a lot @lscharmer for looking into this!
I resolved your targeted issues.

To keep this PR open for contribution, I would suggest (as said inside the JF) keeping this open at least a week to any developer to pick out of the shepherdless.

If none states her/his interest, feel free to review the whole PR (rest are just doc changes) and merge this.

Greetings,
@iszmais

PR #7754, opened 2024-07-03
Auth Feature: Configurable LogOut Behavior

fhelfer:
Implemented according to Feature wiki: https://docu.ilias.de/goto_docu_wiki_wpage_8261_1357.html

This lets the user configure the Destination after logout of a session.
Options are: login page, logout page, internal resource or external resource.

### DIFF
--- a/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutHandler.php
+++ b/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutHandler.php
@@ -1,0 +1,71 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\components\Authentication\Logout;
+
+use ilLink;
+use ilSetting;
+use ilStartUpGUI;
+use ILIAS\Data\URI;
+use ilCtrlInterface;
+
+class ConfigurableLogoutHandler
+{
+    public const LOGIN_SCREEN = 'login_screen';
+    public const LOGOUT_SCREEN = 'logout_screen';
+    public const INTERNAL_RESSOURCE = 'internal_ressource';
+    public const EXTERNAL_RESSOURCE = 'external_ressource';
+    private ilSetting $settings;
+    private ilCtrlInterface $ctrl;
+
+    public function __construct()
+    {
+        global $DIC;
+        $this->ctrl = $DIC->ctrl();
+        $this->settings = new ilSetting('auth');
+    }
+
+    public function afterLogout(): URI
+    {
+        switch ($this->settings->get('logout_behaviour', '')) {
+            case self::LOGIN_SCREEN:
+                return new URI(
+                    ILIAS_HTTP_PATH .
+                    '/' . $this->ctrl->getLinkTargetByClass(ilStartUpGUI::class, 'showLoginPage')
+                );
+            case self::INTERNAL_RESSOURCE:
+                $ref_id = (int) $this->settings->get('logout_behaviour_ref_id', '0');
+
+                return new URI(ilLink::_getStaticLink($ref_id));
+            case self::EXTERNAL_RESSOURCE:
+                $url = $this->settings->get('logout_behaviour_url', '');
+                if ($url) {
+                    return new URI($url);
+                }
+                // no break
+            case self::LOGOUT_SCREEN:
+            default:
+                return new URI(
+                    ILIAS_HTTP_PATH .
+                    '/' . $this->ctrl->getLinkTargetByClass(ilStartUpGUI::class, 'showLogout')
+                );
+        }
+    }
+}
--- a/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutService.php
+++ b/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutService.php
@@ -21,25 +21,27 @@ declare(strict_types=1);
 namespace ILIAS\components\Authentication\Logout;
 
 use ilLink;
+use ilAccess;
+use ilObject;
 use ilSetting;
 use ilStartUpGUI;
 use ILIAS\Data\URI;
 use ilCtrlInterface;
+use InvalidArgumentException;
 
-class ConfigurableLogoutHandler
+class ConfigurableLogoutService
 {
     public const LOGIN_SCREEN = 'login_screen';
     public const LOGOUT_SCREEN = 'logout_screen';
     public const INTERNAL_RESSOURCE = 'internal_ressource';
     public const EXTERNAL_RESSOURCE = 'external_ressource';
-    private ilSetting $settings;
-    private ilCtrlInterface $ctrl;
 
-    public function __construct()
-    {
-        global $DIC;
-        $this->ctrl = $DIC->ctrl();
-        $this->settings = new ilSetting('auth');
+    public function __construct(
+        private readonly ilCtrlInterface $ctrl,
+        private readonly ilSetting $settings,
+        private readonly ilAccess $access,
+        private readonly string $http_path = ILIAS_HTTP_PATH,
+    ) {
     }
 
     public function afterLogout(): URI
@@ -47,25 +49,53 @@ class ConfigurableLogoutHandler
         switch ($this->settings->get('logout_behaviour', '')) {
             case self::LOGIN_SCREEN:
                 return new URI(
-                    ILIAS_HTTP_PATH .
+                    $this->http_path .
                     '/' . $this->ctrl->getLinkTargetByClass(ilStartUpGUI::class, 'showLoginPage')
                 );
+
             case self::INTERNAL_RESSOURCE:
                 $ref_id = (int) $this->settings->get('logout_behaviour_ref_id', '0');
-
-                return new URI(ilLink::_getStaticLink($ref_id));
+                if ($this->isValidInternalResource($ref_id)) {
+                    return new URI(ilLink::_getStaticLink($ref_id));
+                }
+                // no break
             case self::EXTERNAL_RESSOURCE:
                 $url = $this->settings->get('logout_behaviour_url', '');
-                if ($url) {
+                if ($url && $this->isValidExternalResource($url)) {
                     return new URI($url);
                 }
                 // no break
             case self::LOGOUT_SCREEN:
             default:
                 return new URI(
-                    ILIAS_HTTP_PATH .
+                    $this->http_path .
                     '/' . $this->ctrl->getLinkTargetByClass(ilStartUpGUI::class, 'showLogout')
                 );
         }
+    }
+
+    public function isValidInternalResource(int $ref_id): bool
+    {
+        return $this->isInRepository($ref_id) && $this->isAnonymousAccessible($ref_id);
+    }
+
+    public function isInRepository($ref_id): bool
+    {
+        return ilObject::_exists($ref_id, true) && !ilObject::_isInTrash($ref_id);
+    }
+
+    public function isAnonymousAccessible(int $ref_id): bool
+    {
+        return $this->access->checkAccessOfUser(ANONYMOUS_USER_ID, 'read', '', $ref_id);
+    }
+
+    public function isValidExternalResource(string $url): bool
+    {
+        try {
+            $uri = new URI($url);
+        } catch (InvalidArgumentException) {
+            return false;
+        }
+        return filter_var($url, FILTER_VALIDATE_URL) !== false;
     }
 }
--- a/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutService.php
+++ b/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutService.php
@@ -58,20 +58,21 @@ class ConfigurableLogoutService
                 if ($this->isValidInternalResource($ref_id)) {
                     return new URI(ilLink::_getStaticLink($ref_id));
                 }
-                // no break
+                break;
             case self::EXTERNAL_RESSOURCE:
                 $url = $this->settings->get('logout_behaviour_url', '');
                 if ($url && $this->isValidExternalResource($url)) {
                     return new URI($url);
                 }
-                // no break
+                break;
             case self::LOGOUT_SCREEN:
             default:
-                return new URI(
-                    $this->http_path .
-                    '/' . $this->ctrl->getLinkTargetByClass(ilStartUpGUI::class, 'showLogout')
-                );
+                break;
         }
+        return new URI(
+            $this->http_path .
+            '/' . $this->ctrl->getLinkTargetByClass(ilStartUpGUI::class, 'showLogout')
+        );
     }
 
     public function isValidInternalResource(int $ref_id): bool
@@ -96,6 +97,7 @@ class ConfigurableLogoutService
         } catch (InvalidArgumentException) {
             return false;
         }
+
         return filter_var($url, FILTER_VALIDATE_URL) !== false;
     }
 }
--- a/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutTarget.php
+++ b/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutTarget.php
@@ -69,6 +69,7 @@ class ConfigurableLogoutTarget
             default:
                 break;
         }
+
         return new URI(
             $this->http_path .
             '/' . $this->ctrl->getLinkTargetByClass(ilStartUpGUI::class, 'showLogout')
--- a/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutTarget.php
+++ b/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutTarget.php
@@ -29,7 +29,7 @@ use ILIAS\Data\URI;
 use ilCtrlInterface;
 use InvalidArgumentException;
 
-class ConfigurableLogoutService
+class ConfigurableLogoutTarget
 {
     public const LOGIN_SCREEN = 'login_screen';
     public const LOGOUT_SCREEN = 'logout_screen';
@@ -44,7 +44,7 @@ class ConfigurableLogoutService
     ) {
     }
 
-    public function afterLogout(): URI
+    public function asURI(): URI
     {
         switch ($this->settings->get('logout_behaviour', '')) {
             case self::LOGIN_SCREEN:
@@ -80,7 +80,7 @@ class ConfigurableLogoutService
         return $this->isInRepository($ref_id) && $this->isAnonymousAccessible($ref_id);
     }
 
-    public function isInRepository($ref_id): bool
+    public function isInRepository(int $ref_id): bool
     {
         return ilObject::_exists($ref_id, true) && !ilObject::_isInTrash($ref_id);
     }
--- a/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutTarget.php
+++ b/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutTarget.php
@@ -47,6 +47,7 @@ class ConfigurableLogoutTarget implements LogoutTarget
         switch ($this->settings->get('logout_behaviour', '')) {
             case LogoutDestinations::LOGIN_SCREEN->value:
                 return LogoutDestinations::LOGIN_SCREEN->asURI($this->ctrl, $this->http_path);
+
             case self::INTERNAL_RESSOURCE:
                 $ref_id = (int) $this->settings->get('logout_behaviour_ref_id', '0');
                 if ($this->isValidInternalResource($ref_id)) {
@@ -54,13 +55,14 @@ class ConfigurableLogoutTarget implements LogoutTarget
                 }
 
                 break;
+
             case self::EXTERNAL_RESSOURCE:
                 $url = $this->settings->get('logout_behaviour_url', '');
                 if ($url && $this->isValidExternalResource($url)) {
                     return new URI($url);
                 }
-
                 break;
+
             case LogoutDestinations::LOGOUT_SCREEN->value:
             default:
                 break;
--- a/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutTarget.php
+++ b/components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutTarget.php
@@ -24,15 +24,13 @@ use ilLink;
 use ilAccess;
 use ilObject;
 use ilSetting;
-use ilStartUpGUI;
 use ILIAS\Data\URI;
 use ilCtrlInterface;
 use InvalidArgumentException;
+use ILIAS\Authentication\Logout\LogoutDestinations;
 
-class ConfigurableLogoutTarget
+class ConfigurableLogoutTarget implements LogoutTarget
 {
-    public const LOGIN_SCREEN = 'login_screen';
-    public const LOGOUT_SCREEN = 'logout_screen';
     public const INTERNAL_RESSOURCE = 'internal_ressource';
     public const EXTERNAL_RESSOURCE = 'external_ressource';
 
@@ -47,33 +45,28 @@ class ConfigurableLogoutTarget
     public function asURI(): URI
     {
         switch ($this->settings->get('logout_behaviour', '')) {
-            case self::LOGIN_SCREEN:
-                return new URI(
-                    $this->http_path .
-                    '/' . $this->ctrl->getLinkTargetByClass(ilStartUpGUI::class, 'showLoginPage')
-                );
-
+            case LogoutDestinations::LOGIN_SCREEN->value:
+                return LogoutDestinations::LOGIN_SCREEN->asURI($this->ctrl, $this->http_path);
             case self::INTERNAL_RESSOURCE:
                 $ref_id = (int) $this->settings->get('logout_behaviour_ref_id', '0');
                 if ($this->isValidInternalResource($ref_id)) {
                     return new URI(ilLink::_getStaticLink($ref_id));
                 }
+
                 break;
             case self::EXTERNAL_RESSOURCE:
                 $url = $this->settings->get('logout_behaviour_url', '');
                 if ($url && $this->isValidExternalResource($url)) {
                     return new URI($url);
                 }
+
                 break;
-            case self::LOGOUT_SCREEN:
+            case LogoutDestinations::LOGOUT_SCREEN->value:
             default:
                 break;
         }
 
-        return new URI(
-            $this->http_path .
-            '/' . $this->ctrl->getLinkTargetByClass(ilStartUpGUI::class, 'showLogout')
-        );
+        return LogoutDestinations::LOGOUT_SCREEN->asURI($this->ctrl, $this->http_path);
     }
 
     public function isValidInternalResource(int $ref_id): bool
--- a/components/ILIAS/Authentication/classes/Logout/class.ilAuthLogoutBehaviourGUI.php
+++ b/components/ILIAS/Authentication/classes/Logout/class.ilAuthLogoutBehaviourGUI.php
@@ -220,9 +220,10 @@ class ilAuthLogoutBehaviourGUI
             $mode = $data['logout_behaviour']['logout_behaviour_settings'][0];
 
             switch ($mode) {
-                case LogoutDestinations::LOGIN_SCREEN:
-                case LogoutDestinations::LOGOUT_SCREEN:
+                case LogoutDestinations::LOGIN_SCREEN->value:
+                case LogoutDestinations::LOGOUT_SCREEN->value:
                     break;
+
                 case ConfigurableLogoutTarget::INTERNAL_RESSOURCE:
                     $this->settings->set(
                         'logout_behaviour_ref_id',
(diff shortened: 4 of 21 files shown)


components/ILIAS/Init/classes/class.ilStartUpGUI.php
```diff
@@ -1316,7 +1316,31 @@ private function doLogout(): void
...
+                $url = $settings->get('logout_behaviour_url', '');
+                if ($url) {
+                    $this->ctrl->redirectToURL($url);
+                }
+                // no break
+            case 'logout_screen':
+            default:
+                $this->ctrl->redirect($this, 'showLogout');
+                break;
+        }
```
mjansenDatabay: Please extract this code into a new class. The called method should return an URL and MUST NOT depend on `ilCtrlInterface`.

Please ensure, that `$ref_id` is valid during execution time. If it is not valid, the default should be executed.

components/ILIAS/Authentication/classes/Logout/class.ilAuthLogoutBehaviourGUI.php
```diff
@@ -0,0 +1,177 @@
...
+            $this->lng->txt('destination_logout_screen')
+        );
+
+        $login_group = $this->ui_factory->input()->field()->group(
+            [],
+            $this->lng->txt('destination_login_screen'),
+            $this->lng->txt('destination_login_screen_info')
+        );
+
+        $ref_id = $this->ui_factory->input()->field()->numeric(
```
mjansenDatabay: We should add a "Custom Refinery Constraint" which validates the `$ref_id`. It MUST be a valid repository tree node which is NOT trashed, and it must be accessible by the anonymous user (which you already checked in `saveFormCmd`).

When populating the form with the data from the persistence layer, please add  a hint/info in case the `$ref_id` is NOT valid anymore, because the configured `$ref_id` got trashed/or completely deleted in the meantime.

components/ILIAS/Authentication/classes/Logout/class.ilAuthLogoutBehaviourGUI.php
```diff
@@ -0,0 +1,177 @@
...
+                case 'internal_ressource':
+                    // check access
+                    if ($this->access->checkAccessOfUser(ANONYMOUS_USER_ID, 'read', '', $data['logout_behaviour']['logout_behaviour_settings'][1]['ref_id'])) {
+                        $this->settings->set('logout_behaviour_ref_id', (string) $data['logout_behaviour']['logout_behaviour_settings'][1]['ref_id']);
+                    } else {
+                        $this->tpl->setOnScreenMessage($this->tpl::MESSAGE_TYPE_FAILURE, $this->lng->txt('logout_behaviour_invalid_ref_id'));
+                        $this->tpl->setContent($this->ui_renderer->render($form));
+
+                        return;
+                    }
```
mjansenDatabay: Can we put this into a "Custom Refinery Constraint"?

components/ILIAS/Authentication/classes/Logout/class.ilAuthLogoutBehaviourGUI.php
```diff
@@ -0,0 +1,177 @@
...
+                    $this->settings->set('logout_behaviour_ref_id', (string) ($data['logout_behaviour']['logout_behaviour_settings'][1]['ref_id'] ?? ''));
+
+                    break;
+                case 'external_ressource':
+                    /** @var URI $url */
+                    $url = $data['logout_behaviour']['logout_behaviour_settings'][1]['url'] ?? '';
```
mjansenDatabay, line 235: Can we put this into a "Custom Refinery Constraint"?

lang/ilias_en.lang
```diff
@@ -2129,13 +2129,25 @@ auth#:#auth_saml_username_claim_info#:#Define the attribute ILIAS uses to genera
 auth#:#auth_sync#:#User Synchronisation
 auth#:#auth_sync_cas#:#Automatic generation of user accounts
 auth#:#auth_sync_cas_info#:#Automatically create ILIAS user accounts for users who have successfully authenticated themselves using CAS, but do not yet have an ILIAS account.
+auth#:#destination_after_logout#:#Destination after logout
+auth#:#destination_external_ressource#:#Show External Ressource
+auth#:#destination_external_ressource_url#:#Externe URL
```
mjansenDatabay:
```suggestion
auth#:#destination_external_ressource_url#:#External URL
```

lang/ilias_en.lang
```diff
@@ -2129,13 +2129,25 @@ auth#:#auth_saml_username_claim_info#:#Define the attribute ILIAS uses to genera
...
+auth#:#destination_logout_screen#:#Show Logout Screen
 auth#:#err_auth_ldap_failed#:#Authentication failed. Please contact your ILIAS administrator.
 auth#:#err_auth_saml_failed#:#Authentication failed. Please contact your system administrator.
 auth#:#err_auth_saml_no_ilias_user#:#Authentication failed. Please contact your system administrator.
 auth#:#login_page_activate#:#Activate Editor for Selected Languages
 auth#:#login_page_editor_switched#:#The active page editor has been changed
 auth#:#login_page_switch_ipe#:#Use ILIAS page editor
 auth#:#login_page_switch_rte#:#Use Rich-text editor
+auth#:#logout_behaviour#:#Logout Behaviour
+auth#:#logout_behaviour_invalid_ref_id#:#Please make sure Anonymous User has Access to the ressource.
```
mjansenDatabay, line 2148:
```suggestion
auth#:#logout_behaviour_invalid_ref_id#:#Please make sure the "Anonymous" user has access to the defined ressource.
```

mjansenDatabay (changes requested):

components/ILIAS/Init/classes/class.ilStartUpGUI.php
fhelfer: @mjansenDatabay, please take a look at the changes I made to here: https://github.com/ILIAS-eLearning/ILIAS/pull/7754/files#diff-f0a2089fcdfd5ddae416a3afc1ab425b6261788d70f2df72dfdbd1cc34df56d5R48-R67
Is there any way to get rid of `ilCtrlInterface` ? I haven't found a solution that works.

Best, @fhelfer

components/ILIAS/Authentication/classes/Logout/class.ilAuthLogoutBehaviourGUI.php
```diff
@@ -0,0 +1,213 @@
...
+use ILIAS\UI\Factory as UIFactory;
+use ILIAS\UI\Renderer as UIRenderer;
+use ILIAS\HTTP\Services as HttpService;
+use ILIAS\Refinery\Factory as Refinery;
+use Psr\Http\Message\ServerRequestInterface;
+use ILIAS\components\Authentication\Logout\ConfigurableLogoutHandler;
+use ILIAS\UI\Component\Input\Container\Form\Standard as StandardForm;
+
+/**
+ * @ingroup           ServicesAuthentication
```
mjansenDatabay:
```suggestion
```

components/ILIAS/Authentication/classes/class.ilObjAuthSettingsGUI.php
```diff
@@ -850,6 +851,15 @@ public function executeCommand(): void
...
 
+            case strtolower(ilAuthLogoutBehaviourGUI::class):
+                $this->setSubTabs('authSettings');
+                $this->tabs_gui->setTabActive('authentication_settings');
+                $this->tabs_gui->setSubTabActive('logout_behaviour');
+
+                $gui = new ilAuthLogoutBehaviourGUI($this->object->getRefId());
+                $this->ctrl->forwardCommand($gui);
+
+                break;
```
mjansenDatabay, line 890:
```suggestion
                break;
                
```

lang/ilias_en.lang
```diff
@@ -2129,13 +2129,26 @@ auth#:#auth_saml_username_claim_info#:#Define the attribute ILIAS uses to genera
 auth#:#auth_sync#:#User Synchronisation
 auth#:#auth_sync_cas#:#Automatic generation of user accounts
 auth#:#auth_sync_cas_info#:#Automatically create ILIAS user accounts for users who have successfully authenticated themselves using CAS, but do not yet have an ILIAS account.
+auth#:#destination_after_logout#:#Destination after logout
```
mjansenDatabay:
```suggestion
auth#:#destination_after_logout#:#Destination After Logout
```

components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutHandler.php
```diff
@@ -0,0 +1,71 @@
...
+    public const INTERNAL_RESSOURCE = 'internal_ressource';
+    public const EXTERNAL_RESSOURCE = 'external_ressource';
+    private ilSetting $settings;
+    private ilCtrlInterface $ctrl;
+
+    public function __construct()
+    {
+        global $DIC;
+        $this->ctrl = $DIC->ctrl();
+        $this->settings = new ilSetting('auth');
```
mjansenDatabay: Can we pass these dependencies as constructor args?

```php
public function __construct(private ilCtrlInterface $ctrl, private ilSetting $settings, private string $http_path = ILIAS_HTTP_PATH)

File: components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutHandler.php [hunk 11]
```diff
@@ -0,0 +1,71 @@
...
+use ILIAS\Data\URI;
+use ilCtrlInterface;
+
+class ConfigurableLogoutHandler
+{
+    public const LOGIN_SCREEN = 'login_screen';
+    public const LOGOUT_SCREEN = 'logout_screen';
+    public const INTERNAL_RESSOURCE = 'internal_ressource';
+    public const EXTERNAL_RESSOURCE = 'external_ressource';
+    private ilSetting $settings;
```
mjansenDatabay:
```suggestion
    public const EXTERNAL_RESSOURCE = 'external_ressource';

    private ilSetting $settings;
```

components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutHandler.php
```diff
@@ -0,0 +1,71 @@
...
+
+namespace ILIAS\components\Authentication\Logout;
+
+use ilLink;
+use ilSetting;
+use ilStartUpGUI;
+use ILIAS\Data\URI;
+use ilCtrlInterface;
+
+class ConfigurableLogoutHandler
```
mjansenDatabay: We should do some brainstorming for a proper naming. IMO this is not a "Handler", because this class does not handle something, consumers can retrieve an URL based on a given configuration.

components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutHandler.php
```diff
@@ -0,0 +1,71 @@
...
+        switch ($this->settings->get('logout_behaviour', '')) {
+            case self::LOGIN_SCREEN:
+                return new URI(
+                    ILIAS_HTTP_PATH .
+                    '/' . $this->ctrl->getLinkTargetByClass(ilStartUpGUI::class, 'showLoginPage')
+                );
+            case self::INTERNAL_RESSOURCE:
+                $ref_id = (int) $this->settings->get('logout_behaviour_ref_id', '0');
+
+                return new URI(ilLink::_getStaticLink($ref_id));
```
mjansenDatabay:
```suggestion
                return new URI(ilLink::_getStaticLink($ref_id));

```

components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutHandler.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public function afterLogout(): URI
+    {
+        switch ($this->settings->get('logout_behaviour', '')) {
+            case self::LOGIN_SCREEN:
+                return new URI(
+                    ILIAS_HTTP_PATH .
+                    '/' . $this->ctrl->getLinkTargetByClass(ilStartUpGUI::class, 'showLoginPage')
+                );
```
mjansenDatabay:
```suggestion
                );

```

components/ILIAS/Authentication/classes/Logout/class.ilAuthLogoutBehaviourGUI.php
```diff
@@ -0,0 +1,213 @@
...
+
+        return $form;
+    }
+
+    public function showForm(): void
+    {
+        $this->tpl->setContent($this->ui_renderer->render($this->getForm()));
+    }
+
+    public function saveFormCmd(): void
```
mjansenDatabay: Maybe we should remove the `Cmd` suffix, because `showForm` does not have such.

components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutHandler.php
```diff
@@ -0,0 +1,71 @@
...
+            case self::INTERNAL_RESSOURCE:
+                $ref_id = (int) $this->settings->get('logout_behaviour_ref_id', '0');
```
mjansenDatabay: We need further checks here:

1. An existence check
2. A "Trash" check
3. A permission check for the `read` operation (for `ANONYMOUS_USER_ID`)

If one of the the checks fails, we should return the default URL.

It would be nice if we could have **centralized** checks/constraints when trying to retrieve the URL and when saving the form.

components/ILIAS/Authentication/classes/Logout/class.ilAuthLogoutBehaviourGUI.php
```diff
@@ -0,0 +1,213 @@
...
+        if ($request) {
+            $form = $form->withRequest($request);
+        }
+
+        return $form;
+    }
+
+    public function showForm(): void
+    {
+        $this->tpl->setContent($this->ui_renderer->render($this->getForm()));
```
mjansenDatabay: If we show the form with an existing configuratrion and the entered `$red_id` is not valid anymore, we should provide a `MessageBox > Failure` here.

mjansenDatabay (changes requested):

components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutService.php
```diff
@@ -0,0 +1,103 @@
...
+            '/' . $this->ctrl->getLinkTargetByClass(ilStartUpGUI::class, 'showLogout')
+        );
+    }
+
+    public function isValidInternalResource(int $ref_id): bool
+    {
+        return $this->isInRepository($ref_id) && $this->isAnonymousAccessible($ref_id);
+    }
+
+    public function isInRepository($ref_id): bool
```
mjansenDatabay:
```suggestion
    public function isInRepository(int $ref_id): bool
```

components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutService.php
```diff
@@ -0,0 +1,103 @@
...
+
+    public function __construct(
+        private readonly ilCtrlInterface $ctrl,
+        private readonly ilSetting $settings,
+        private readonly ilAccess $access,
+        private readonly string $http_path = ILIAS_HTTP_PATH,
+    ) {
+    }
+
+    public function afterLogout(): URI
```
mjansenDatabay: Maybe we should rename this to `asUrl`?

mjansenDatabay (approved):

oliversamoila:
I have approved it in the FeatureWiki. Thank you all.

Please merge if it works for you.
Best regards
Oliver

mjansenDatabay (changes requested):
Hi @fhelfer ,

IMO we'll have to use the logout URL in `\ilStartUpGUI::doLogout` when redirecting to the `saml.php` and providing the `logout_url` query parameter.

Beset regards
Michael

components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutTarget.php
```diff
@@ -0,0 +1,103 @@
...
+                $url = $this->settings->get('logout_behaviour_url', '');
+                if ($url && $this->isValidExternalResource($url)) {
+                    return new URI($url);
+                }
+                break;
+            case self::LOGOUT_SCREEN:
+            default:
+                break;
+        }
+        return new URI(
```
mjansenDatabay:
```suggestion

        return new URI(
```

components/ILIAS/Init/classes/class.ilStartUpGUI.php
```diff
@@ -1342,19 +1339,34 @@ private function doLogout(): void
...
         $client_id = CLIENT_ID;
         ilUtil::setCookie('ilClientId', '');
 
-        $this->ctrl->setParameter($this, 'client_id', $client_id);
         $this->ctrl->setParameter($this, 'lang', $user_language);
-        $this->ctrl->redirect($this, 'showLogout');
+        $this->ctrl->setParameter($this, 'client_id', $client_id);
+
+        $this->mainTemplate->setOnScreenMessage(
+            $this->mainTemplate::MESSAGE_TYPE_FAILURE,
```
mjansenDatabay: Is this really a "Failure"?

components/ILIAS/Init/classes/class.ilStartUpGUI.php
```diff
@@ -1304,10 +1305,6 @@ private function showLogout(): void
 
         $tpl->setVariable('LPE', $this->getLogoutPageEditorHTML());
         $tpl->setVariable('TXT_PAGEHEADLINE', $this->lng->txt('logout'));
-        $tpl->setVariable(
-            'TXT_LOGOUT_TEXT',
-            $this->lng->txt('logout_text') . $this->dic['legalDocuments']->logoutText()
```
mjansenDatabay, line 1312: Where do we want to show this text?

`$this->dic['legalDocuments']->logoutText()`

components/ILIAS/LegalDocuments/classes/Conductor.php
```diff
@@ -89,14 +74,19 @@ public function loginPageHTML(string $id): string
     public function logoutText(): string
     {
         try {
```
lscharmer, line 76: The try catch block is only for `...->query->retrieve(...)` not the whole method.

components/ILIAS/LegalDocuments/classes/Conductor.php
```diff
@@ -89,14 +74,19 @@ public function loginPageHTML(string $id): string
...
-            $id = $this->container->http()->wrapper()->query()->retrieve('withdraw_from', $this->container->refinery()->to()->string());
-        } catch (Exception) {
-            return '';
-        }
+            $id = $this->container->http()->wrapper()->query()->retrieve('withdraw_consent', $this->container->refinery()->to()->string());
 
-        $logout_text = $this->internal->get('logout-text', $id);
+            $logout = $this->internal->get('logout', $id);
+            if (null !== $logout) {
+                $this->container->ctrl()->setParameterByClass(ilStartUpGUI::class, 'withdraw_from', $id);
```
lscharmer: Can be removed.

lscharmer (changes requested):
There is also another thing that needs to be changed in the LegalDocuments. Another `get` parameter is set in the  `WithdrawProcess` which must be returned instead.

mjansenDatabay (changes requested):
Hi @fhelfer ,

as discussed, we'll have to introduce a "Decorator" for the `ConfigurableLogoutTarget`, which is returned by the "LegalDocuments" component in case a user withdrew from the agreements.
This new implementation should always return a "Logout Page" URL if the "LegalDocuments" decide to show the "Withdrew/Re-Login to Delete Account" message in case an "External URL" is configured as logout URL in the ILIAS administration.

Example:

I suggest introducing a new "Backed Enum" which provides a method (e.g. again `asURI(ilCtrlInterface $ctrl, string $http_path): URI`) to retrieve the URL string from this new "Enum". The "Enum" should have the cases `LoginUrl` and `LogoutUrl`. Why? These URL types are "static", the other types are dynamic and should be also known/buildable by `ConfigurableLogoutTarget`.

New interface:
```php
interface LogoutTarget
{
    public function asURI(): URI;
}

```

`ilStartupGUI::doLogout`:
```php
// ...
$target = $DIC['legalDocuments']->logoutTarget($target);
// ...
```

New interface impl. of the `LegalDocuments` component:
```php
class ForcedLogoutTarget implements LogoutTarget
{
    public function __construct(private LogoutTarget $origin, ilCtrlInterface $ctrl)
    {
    }


    public function asURI(): URI
    {
        if ($user_withdrew_legal_docs) { // <-------------- This has to be decided by the `LegalDocuments` component
            return new URI(
               $this->http_path .
                    '/' . $this->ctrl->getLinkTargetByClass(ilStartUpGUI::class, 'showLogout')
            );
        }
        
        return $this->origin->asURI();
    }
}

```

The new "Enum" type could be then used in `ConfigurableLogoutTarget` (`Authentication`) and `ForcedLogoutTarget` (`LegalDocuments`) to retrieve the login (only relevant for `ConfigurableLogoutTarget`) and logout URLs.

Best regards,
Michael

components/ILIAS/Authentication/classes/Logout/class.ilAuthLogoutBehaviourGUI.php
```diff
@@ -219,8 +220,8 @@ public function saveForm(): void
             $mode = $data['logout_behaviour']['logout_behaviour_settings'][0];
 
             switch ($mode) {
-                case ConfigurableLogoutTarget::LOGOUT_SCREEN:
-                case ConfigurableLogoutTarget::LOGIN_SCREEN:
+                case LogoutDestinations::LOGIN_SCREEN:
+                case LogoutDestinations::LOGOUT_SCREEN:
```
mjansenDatabay: Don't we need the `->value` here?

components/ILIAS/LegalDocuments/classes/Conductor.php
```diff
@@ -99,6 +88,24 @@ public function logoutText(): string
...
+            $id = $this->container->http()->wrapper()->query()->retrieve(
+                'withdraw_consent',
+                $this->container->refinery()->to()->string()
+            );
+        } catch (Exception) {
+            $id = null;
+        }
+
+        if ($id !== null) {
+            return new ForcedLogoutTarget($this->container->ctrl());
```
lscharmer: This return can be moved into the catch block (and the if can be removed).

components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutTarget.php
```diff
@@ -0,0 +1,97 @@
...
+            case LogoutDestinations::LOGIN_SCREEN->value:
+                return LogoutDestinations::LOGIN_SCREEN->asURI($this->ctrl, $this->http_path);
+            case self::INTERNAL_RESSOURCE:
+                $ref_id = (int) $this->settings->get('logout_behaviour_ref_id', '0');
+                if ($this->isValidInternalResource($ref_id)) {
+                    return new URI(ilLink::_getStaticLink($ref_id));
+                }
+
+                break;
+            case self::EXTERNAL_RESSOURCE:
```
mjansenDatabay, line 59:
```suggestion

            case self::EXTERNAL_RESSOURCE:
```

components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutTarget.php
```diff
@@ -0,0 +1,97 @@
...
+        private readonly string $http_path = ILIAS_HTTP_PATH,
+    ) {
+    }
+
+    public function asURI(): URI
+    {
+        switch ($this->settings->get('logout_behaviour', '')) {
+            case LogoutDestinations::LOGIN_SCREEN->value:
+                return LogoutDestinations::LOGIN_SCREEN->asURI($this->ctrl, $this->http_path);
+            case self::INTERNAL_RESSOURCE:
```
mjansenDatabay, line 51:
```suggestion

            case self::INTERNAL_RESSOURCE:
```

components/ILIAS/Authentication/classes/Logout/ConfigurableLogoutTarget.php
```diff
@@ -0,0 +1,97 @@
...
+                }
+
+                break;
+            case self::EXTERNAL_RESSOURCE:
+                $url = $this->settings->get('logout_behaviour_url', '');
+                if ($url && $this->isValidExternalResource($url)) {
+                    return new URI($url);
+                }
+
+                break;
```
mjansenDatabay, line 64:
```suggestion
                break;

```

components/ILIAS/Authentication/classes/Logout/class.ilAuthLogoutBehaviourGUI.php
```diff
@@ -0,0 +1,241 @@
...
+
+            $this->http->close();
+        }
+        if (isset($data['logout_behaviour']['logout_behaviour_settings'][0])) {
+            $mode = $data['logout_behaviour']['logout_behaviour_settings'][0];
+
+            switch ($mode) {
+                case LogoutDestinations::LOGIN_SCREEN:
+                case LogoutDestinations::LOGOUT_SCREEN:
+                    break;
```
mjansenDatabay, line 225:
```suggestion
                    break;

```

components/ILIAS/Authentication/classes/Logout/class.ilAuthLogoutBehaviourGUI.php
```diff
@@ -0,0 +1,241 @@
...
+                case ConfigurableLogoutTarget::INTERNAL_RESSOURCE:
+                    $this->settings->set(
+                        'logout_behaviour_ref_id',
+                        (string) ($data['logout_behaviour']['logout_behaviour_settings'][1]['ref_id'] ?? '')
+                    );
+                    break;
+                case ConfigurableLogoutTarget::EXTERNAL_RESSOURCE:
```
mjansenDatabay, line 233:
```suggestion
                case ConfigurableLogoutTarget::EXTERNAL_RESSOURCE:

```

components/ILIAS/LegalDocuments/classes/ForcedLogoutTarget.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    public function __construct(
+        private readonly ilCtrlInterface $ctrl,
+        private readonly string $http_path = ILIAS_HTTP_PATH
+    ) {
+    }
+
+    public function asURI(): URI
+    {
+        return LogoutDestinations::LOGIN_SCREEN->asURI($this->ctrl, $this->http_path);
```
mjansenDatabay: Actually, this is a little different than my suggested approach.

In my approach, the original "Logout Target" should be passed to the `__construct(...)` of **this** class.

The `asURI` method of **this** class should than decide (with the help of "indicators", also passed to `__construct(...)`), if the URI of the original "Logout Target" is returned, or `LogoutDestinations::LOGIN_SCREEN->asURI($this->ctrl, $this->http_path);`.

mjansenDatabay (changes requested):
Hi @fhelfer ,

I think we are close, but I again added some remarks :-).

Best regards
Michael

lscharmer (commented):
LGTM. This is not something that prevents this merge but I'm wondering why the LegalDocuments need the LogoutTarget Stuff. IMO This is just a boolean query instead, which would remove the need for the ForcedLogoutTarget class etc.

components/ILIAS/Authentication/classes/Logout/class.ilAuthLogoutBehaviourGUI.php
fhelfer: Unfortunately yes

mjansenDatabay:
> LGTM. This is not something that prevents this merge but I'm wondering why the LegalDocuments need the LogoutTarget Stuff. IMO This is just a boolean query instead, which would remove the need for the ForcedLogoutTarget class etc.

The "Logout" stuff was already part of the "Withdrawal Process", and it is a business rule of the "Legal Documents", that the withdrawal should be acknowledged with a hint in a "Message Box" on an ILIAS screen.

The "Legal"Documents" already hook into the "Logout Process", so IMO consuming an interface for the "Logout Target" use case and providing an alternative/forced logout target whenever the withdrawal process requires presenting important information to our users, is reasonable from my point of view. Of course the "Logout Target" is a concept of the "Authentication" component, but as there are already methods like `\ILIAS\LegalDocuments\Conductor::onLogout` and `\ILIAS\LegalDocuments\Conductor::logoutText`, a new method to into the "LogoutTarget" is very similar, especially if it only depends on an interface to optionally provide the default logout page URL.

Of course we now have dependencies (`ilStartupGUI` calls a method of `LecalDocuments` and `LegalDocuments` depends on an interface of `Authentication`) on both sides.
The best approach would probably be to use a kind of "Collector/Provider" pattern with some priority management since components might want to force a specific logout URL (like "Saml" or "LecalDocuments") and/or consume the current logout URL (like "Saml"). Furthermore, there must be clear rules defined how the logout process and the URL providers interact with each other, especially concerning something like "Order" etc..

components/ILIAS/LegalDocuments/classes/Conductor.php
```diff
@@ -99,6 +88,23 @@ public function logoutText(): string
         return null === $logout_text ? '' : $this->container->ui()->renderer()->render($logout_text());
     }
 
+    public function logoutTarget(LogoutTarget $target): LogoutTarget
+    {
+        $user_withdrew_legal_docs = false;
```
lscharmer: You don't need this variable (`$user_withdrew_legal_docs`). You can just return the `$target` in the `catch` block and else return `ForcedLogoutTarget`. You don't need the `if` in the `ForcedLogoutTarget`.

I also noticed that we can make this even simpler: As we don't use `$id` we can just check for existence with `$this->container->http()->wrapper()->query()->has('withdraw_consent');` instead of the try catch block.

mjansenDatabay (approved):

components/ILIAS/LegalDocuments/classes/ForcedLogoutTarget.php
lscharmer: I disagree with the notion that the `ForcedLogoutTarget` should decide whether the original `$target` should be used or not, the class name does not reflect this. If I see a `ForcedLogoutTarget` I would assume that this **is** the decision that we want to force a specific logout target and not that it would decide if a page should be forced or not.

Regardless, as this is just a minor thing, the current approach is ok for me.

lscharmer (approved):

PR #7757, opened 2024-07-04
Fix: Missing Radio Buttons in ECS Course Allocation

iljalukin:
https://mantis.ilias.de/view.php?id=41657


pascalseeland:
Merged outside of github as one fix was missing. Merged as 06a5f8a.

PR #7758, opened 2024-07-04
Certificate Bugfix #0041637:  Certificates aren't issued in the language of the receiving user in any case (month names)

fhelfer:
https://mantis.ilias.de/view.php?id=41637
depends on #7759

**!Important: all these have to be picked to 9 as well**

### DIFF
--- a/components/ILIAS/Certificate/classes/Helper/ilCertificateDateHelper.php
+++ b/components/ILIAS/Certificate/classes/Helper/ilCertificateDateHelper.php
@@ -26,7 +26,7 @@ class ilCertificateDateHelper
     /**
      * @param string|int $date
      */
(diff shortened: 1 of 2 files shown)


mjansenDatabay (changes requested):
Hi @fhelfer ,

can we make `$user` the last optional(!) parameter for `formatDate` (here and in the related PRs)? IMO I will not be able to merge this PR otherwise, because all other certificate providers don't pass a `ilObjUser` instance as 2nd argument, and we have no guarantee that the other PRs are merged soon.

-- UPDATE --

No consumer/caller uses the 2nd parameter, yet.

Best regards,
Michael

mjansenDatabay (approved):

PR #7759, opened 2024-07-04
Calendar Bugfix: Calendar Global Usage despite having local Services

fhelfer:
Calendar Uses Global $DIC language even if Languages are set static to a different Language.
This language is not respected when Months are translated.

This is a huge issue for Certificates, because from the Certificate Component it's not possible to set a Date in the language of the Certificate Recipient

Best, @fhelfer

### DIFF
--- a/components/ILIAS/Calendar/classes/class.ilDatePresentation.php
+++ b/components/ILIAS/Calendar/classes/class.ilDatePresentation.php
@@ -32,6 +32,7 @@ class ilDatePresentation
 {
     public static bool $use_relative_dates = true;
     private static ?ilLanguage $lang = null;
+    private static ?ilObjUser $user = null;
 
     public static ?ilDateTime $today = null;
     public static ?ilDateTime $tomorrow = null;
@@ -73,6 +74,19 @@ class ilDatePresentation
         return self::$lang ?: $lng;
     }
 
+    public static function setUser(ilObjUser $user): void
+    {
+        self::$user = $user;
+    }
+
+    public static function getUser(): ilObjUser
+    {
+        global $DIC;
+
+        $user = $DIC->user();
+        return self::$user ?: $user;
+    }
+
     /**
      * reset to defaults
      */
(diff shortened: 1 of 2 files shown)


components/ILIAS/Calendar/classes/class.ilDatePresentation.php
```diff
@@ -85,13 +99,15 @@ public static function resetToDefaults()
         self::setUseRelativeDates(true);
     }
 
-    public static function formatDate(ilDateTime $date, bool $a_skip_day = false, bool $a_include_wd = false, bool $include_seconds = false): string
-    {
+    public static function formatDate(
+        ilDateTime $date,
+        bool $a_skip_day = false,
+        bool $a_include_wd = false,
+        bool $include_seconds = false,
```
alex40724, line 86: Why not adding an optional language user object and deriving the appropriate language object from that?

alex40724 (commented):
Thanks for the PR. Would it be possible to pass an optional ilObjUser instance to formatDate() instead of using the static setters which mess with a "global state" like mechanism?

components/ILIAS/Calendar/classes/class.ilDatePresentation.php
fhelfer: The static Method setLanguage is already in use in many components. Some of which have not a clear direct formatDate() call. Without knowing the side effects, it's difficult to now change that.
I agree that passing this is a much better practice, but as it is already used, I cannot predict what effects that may have if we remove setLanguage(). And having both setLanguage and a parameter seems to open us up to race condition and confusion for developers.

Let me know your thoughts to this. Also, maybe start a discussion with other maintainers if that seems like a greater deal.

Best @fhelfer

fhelfer:
> Thanks for the PR. Would it be possible to pass an optional ilObjUser instance to formatDate() instead of using the static setters which mess with a "global state" like mechanism?



> Thanks for the PR. Would it be possible to pass an optional ilObjUser instance to formatDate() instead of using the static setters which mess with a "global state" like mechanism?

As for passing a User, that would be not that much of a problem, but it felt like mixing different concepts. So maybe we wait the discussion on that.

mjansenDatabay:
Hi @alex40724 and @smeyer-ilias ,

any further feedback or a merge of this PR would be highly appreciated, since six other PRs depend on this.

Best regards and thanks for caring,
Michael

components/ILIAS/Calendar/classes/class.ilDatePresentation.php
```diff
@@ -154,59 +171,101 @@ public static function formatDate(ilDateTime $date, bool $a_skip_day = false, bo
...
+                            );
+                }
+            } else {
+                switch ($user->getTimeFormat()) {
+                    case ilCalendarSettings::TIME_FORMAT_24:
+                        return $date_str . $sep .
+                            $start->get(
+                                IL_CAL_FKT_DATE,
+                                'H:i',
+                                $user->getTimeZone()
```
alex40724, line 237: @fhelfer Please have a look at these kind of changes. This costs really lot of time to check, if this code "has no changes at all". Please, if you provide bug fix PRs, do not include so many code style improvements.

alex40724 (approved):
Thanks. We checked the PR in detail. Please note that these kind of PRs are really hard to read, since it includes so many code style related changes. It would be better so separate only the bugfix related changes.

Please provide a separate PR for ILIAS 9 without the coding style changes.

PR #7778, opened 2024-07-05
Fix `explode`-ing `null` instead of `string` in `\ilObjFileInfo…

mjansenDatabay:
…Repository::initInlineSuffixes`

If approved, this should be picked to all maintained releases, if necessary.


chfsx:

PR #7779, opened 2024-07-05
Mail: include markdown in mails

lukastocker:
- add markdown rendering. Also always escape HTML in markdown output.
- use UI form in ilMailFormGUI and add Markdown for Message.
- enable markdown for free text mails.
- adjust PHPUnit Tests
- include new functions in form and buttons to avoid a redirect of the page if the user clicked onto the placeholders to insert them into the mail.

### Notes regarding commits in this PR:

**Mail: decode HTML in mail body of free text mails else ... =>** This commit is optional. If needed I can remove this one.
**Mail: Replaces all known HTML tags with... =>** This commit is optional too, but I would recommend to include the commit to this PR.

Inn addtion I created two more PRs you could check out:
1. [Underline Extension for Markdown ](https://github.com/ILIAS-eLearning/ILIAS/pull/7780)
2. Adjusting the [Markdown's Headline configuration](https://github.com/ILIAS-eLearning/ILIAS/pull/7781), which is also needed for the underline extension PR


components/ILIAS/Init/classes/class.ilInitialisation.php
```diff
@@ -688,6 +688,27 @@ protected static function setSessionCookieParams(): void
         }
     }
 
+    protected static function initMail(\ILIAS\DI\Container $c): void
```
mjansenDatabay: Please don't reintroduce this method, instead ut new registrations to `\ILIAS\Mail\Service\MimeMailService`.

components/ILIAS/Mail/classes/BackgroundTask/class.ilMailDeliveryJob.php
```diff
@@ -56,20 +56,12 @@ public function run(array $input, Observer $observer): Value
...
-        }
-
         $mail->setSaveInSentbox((bool) $input[8]->getValue());
         $mail = $mail
             ->withContextId((string) $input[9]->getValue())
-            ->withContextParameters($context_parameters);
+            ->withContextParameters((array) unserialize($input[10]->getValue(), ['allowed_classes' => false]));
 
-        $mail_data = new MailDeliveryData(
+        $mail->sendMail(
```
mjansenDatabay: I suggest to stick with `MailDeliveryData` introduced by @shecken in ILIAS 9.x

components/ILIAS/Mail/classes/BackgroundTask/class.ilMailDeliveryJob.php
```diff
@@ -56,20 +56,12 @@ public function run(array $input, Observer $observer): Value
             $mail = new ilFormatMail((int) $input[0]->getValue());
         }
 
-        if (isset($context_parameters['auto_responder'])) {
```
mjansenDatabay: We still need this Code in ILIAS 10.x

See: https://docu.ilias.de/goto_docu_wiki_wpage_8142_1357.html

components/ILIAS/Init/classes/class.ilInitialisation.php
```diff
@@ -1270,6 +1291,7 @@ protected static function initClient(): void
         self::initGlobal("https", "ilHTTPS", "./components/ILIAS/Http/classes/class.ilHTTPS.php");
         self::initSettings();
         self::setSessionHandler();
+        self::initMail($GLOBALS['DIC']);
```
mjansenDatabay: Please don't reintroduce this method, put new registrations to `\ILIAS\Mail\Service\MimeMailService`.

components/ILIAS/Mail/classes/BackgroundTask/class.ilMassMailDeliveryJob.php
```diff
@@ -70,7 +70,7 @@ public function run(array $input, Observer $observer): Value
                 )
             );
 
-            $mail_data = new MailDeliveryData(
+            $mail->sendMail(
```
mjansenDatabay: I suggest to stick with `MailDeliveryData` introduced by @shecken in ILIAS 9.x

mjansenDatabay (changes requested):
Hi @lukastocker ,

first of all thanks for your PR.

I just had a quick look and added a few remarks regarding the changes in `Init` and `Mail`. 

For me it looks like there were some issues when rebasing your feature branch based on `trunk`, since some relevant changes we made for ILIAS `10.x` (like the "Auto Responder" feature, the changes related to the "Mail Signature", and the `$DIC` registrations) are partly reverted or removed by your changes.

Best regards,
Michael

components/ILIAS/Mail/classes/Mime/Sender/class.ilMailMimeSenderUser.php
```diff
@@ -79,19 +79,13 @@ public function getFromName(): string
...
-            'FIRSTNAME' => $this->user->getFirstname(),
-            'LASTNAME' => $this->user->getLastname(),
-        ];
-
-        $template = $from;
-        $interpolated = $this->mustache_factory->getBasicEngine()->render($template, $placeholders);
-
-        if ($template !== $interpolated) {
-            return $interpolated;
+        $name = str_ireplace('[FULLNAME]', $this->user->getFullname(), $from);
```
mjansenDatabay: This looks wrong. With ILIAS 10.x, we need the "Mustache" handling for placeholders in the "Mail Signature".

components/ILIAS/Mail/classes/class.ilMimeMail.php
```diff
@@ -247,9 +247,9 @@ protected function build(): void
 
     private function removeHTMLTags(string $maybeHTML): string
     {
-        $maybeHTML = str_ireplace(['<br />', '<br>', '<br/>'], "\n", $maybeHTML);
+        $maybeHTML = str_ireplace(['<br />', '<br>', '<br/>', '</p>'], "\n", $maybeHTML);
```
mjansenDatabay: Why do we need to replace the closing `p`, but not the opening tag?

components/ILIAS/Mail/classes/class.ilMailTemplateContextService.php
```diff
@@ -105,7 +105,7 @@ public static function getTemplateContexts(?array $a_id = null): array
         if (is_array($a_id) && count($a_id)) {
             $where[] = $DIC->database()->in('id', $a_id, false, 'text');
         }
-        if ($where !== []) {
+        if (count($where)) {
```
mjansenDatabay: I suggest to leave this as it is, since this is not related to the actual change, right?

components/ILIAS/Mail/classes/class.ilMailTemplateContextService.php
```diff
@@ -130,8 +130,7 @@ protected static function getContextInstance(
         if (!$a_path) {
             $a_path = $a_component . '/classes/';
         }
-        $class_file = $a_path . 'class.' . $a_class . '.php';
-
+        $class_file = $_SERVER['DOCUMENT_ROOT'] . $a_path . 'class.' . $a_class . '.php';
```
mjansenDatabay: This looks dangerous to me. Maybe we should skip the `file_exists` check and assume that `class_exists` does the trick (by relying that all ILIAS files are part of the "Composer Class Map").

lukastocker:
> Hi @lukastocker ,
> 
> first of all thanks for your PR.
> 
> I just had a quick look and added a few remarks regarding the changes in `Init` and `Mail`.
> 
> For me it looks like there were some issues when rebasing your feature branch based on `trunk`, since some relevant changes we made for ILIAS `10.x` (like the "Auto Responder" feature, the changes related to the "Mail Signature", and the `$DIC` registrations) are partly reverted or removed by your changes.
> 
> Best regards, Michael

Hi @mjansenDatabay ,

thanks for your quick feedback. I will look into the remarks and come back to you soon.

components/ILIAS/Mail/classes/class.ilMailAttachmentTableGUI.php
```diff
@@ -0,0 +1,99 @@
...
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+/**
+ * @author  Michael Jansen <mjansen@databay.de>
+ * @ingroup ServicesMail
+ */
+class ilMailAttachmentTableGUI extends ilTable2GUI
```
mjansenDatabay: This table has been abandoned with ILIAS 10.x, there is already a new "UI Table" base implementation.

See: https://docu.ilias.de/goto_docu_wiki_wpage_8069_1357.html

components/ILIAS/Mail/classes/class.ilAccountMail.php
```diff
@@ -173,16 +174,17 @@ private function addAttachments(array $mailData): void
...
     public function send(): bool
     {
         $user = $this->getUser();
-        if (!$user instanceof ilObjUser) {
+        if (null === $user) {
             throw new RuntimeException('A user instance must be passed when sending emails');
         }
 
-        if ($user->getEmail() === '') {
+        if (!$user->getEmail()) {
```
mjansenDatabay: Do we really need this change? Is it related to the feature?

components/ILIAS/Mail/classes/class.ilAccountMail.php
```diff
@@ -173,16 +174,17 @@ private function addAttachments(array $mailData): void
...
      * If no field values found the defaults are used.
      * Placehoders will be replaced by the appropriate data.
+     * @return bool
      * @throws RuntimeException
      */
     public function send(): bool
     {
         $user = $this->getUser();
-        if (!$user instanceof ilObjUser) {
+        if (null === $user) {
```
mjansenDatabay: Do we really need this change? Is it related to the feature?

components/ILIAS/Mail/classes/class.ilAccountMail.php
```diff
@@ -173,16 +174,17 @@ private function addAttachments(array $mailData): void
      * It first tries to read the mail body, subject and sender address from posted named formular fields.
      * If no field values found the defaults are used.
      * Placehoders will be replaced by the appropriate data.
+     * @return bool
```
mjansenDatabay: This is not needed because we have an explicit return type here.

components/ILIAS/Mail/classes/class.ilAccountMail.php
```diff
@@ -296,13 +300,13 @@ public function replacePlaceholders(string $a_string, ilObjUser $a_user, array $
...
         ) {
             $target = $this->http->wrapper()->query()->retrieve('target', $this->refinery->kindlyTo()->string());
-            $tarr = explode('_', (string) $target);
-            if ($this->repositoryTree->isInTree((int) $tarr[1])) {
+            $tarr = explode("_", $target);
+            if ($tree->isInTree((int) $tarr[1])) {
                 $obj_id = ilObject::_lookupObjId((int) $tarr[1]);
                 $type = ilObject::_lookupType($obj_id);
-                if ($type === $tarr[0]) {
+                if ($type == $tarr[0]) {
```
mjansenDatabay: Do we really need this change? Is it related to the feature?

components/ILIAS/Mail/classes/class.ilAccountMail.php
```diff
@@ -296,13 +300,13 @@ public function replacePlaceholders(string $a_string, ilObjUser $a_user, array $
...
                     $replacements["TARGET_TITLE"] = ilObject::_lookupTitle($obj_id);
-                    $replacements["TARGET"] = ILIAS_HTTP_PATH . '/goto.php?client_id=' . CLIENT_ID . '&target=' . $target;
+                    $replacements["TARGET"] = ILIAS_HTTP_PATH . "/goto.php?client_id=" . CLIENT_ID . "&target=" . $_GET["target"];
```
mjansenDatabay: Do we really need this change? Is it related to the feature? Why do we introduce a new access to `$_GET` here?

components/ILIAS/Mail/classes/Setup/Migration/class.ilUpdateMailTemplatesForMustache.php
```diff
@@ -29,7 +29,7 @@ class ilUpdateMailTemplatesForMustache implements Migration
 
     public function getLabel(): string
     {
-        return 'ilUpdateMailTemplatesForMustache';
+        return "ilUpdateMailTemplatesForMustache";
```
mjansenDatabay: Do we really need this change? Is it related to the feature?

And: IMO we can remove the complete migration for ILIAS 10.x (I could change this separately, if you like).

mjansenDatabay (changes requested):

components/ILIAS/Mail/classes/class.ilMailAttachmentStakeholder.php
```diff
@@ -0,0 +1,42 @@
...
+use ILIAS\ResourceStorage\Stakeholder\AbstractResourceStakeholder;
+
+class ilMailAttachmentStakeholder extends AbstractResourceStakeholder
+{
+    public function __construct()
+    {
+    }
+    /**
+     * @inheritDoc
+     */
```
mjansenDatabay:
```suggestion

```

components/ILIAS/Mail/classes/class.ilMailAttachmentStakeholder.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return 'mail_attachments';
+    }
+
+    public function getOwnerOfNewResources(): int
+    {
+        return 74;
+    }
+
+
```
mjansenDatabay:
```suggestion
```

components/ILIAS/Mail/classes/Setup/Migration/class.ilUpdateMailTemplatesForMustache.php
lukastocker: We def. don't need this change. It must be a relict of some changes I did before my holidays. I will remove this change.
Regarding the removal of the migration: I think it should be a new PR as it has nothing to do with my implementation. You can do this change if you want, else I gladly will do it. :)

components/ILIAS/Mail/classes/class.ilFormatMail.php
```diff
@@ -65,6 +70,8 @@ public function formatReplyRecipient(): string
 
     /**
      * @param string[] $a_names
+     * @param string $a_type
+     * @return array
```
mjansenDatabay: The added docs don't provide any added value. We should remove them.

components/ILIAS/Mail/classes/class.ilFileDataMail.php
```diff
@@ -340,6 +346,9 @@ public static function getStorage(int $a_mail_id, int $a_usr_id): ilFSStorageMai
 
     /**
      * Save attachment file in a specific mail directory .../mail/<calculated_path>/mail_<mail_id>_<user_id>/...
+     * @param int $a_mail_id
+     * @param string $a_attachment
+     * @return bool
```
mjansenDatabay: The added docs don't provide any added value. We should remove them.

components/ILIAS/Mail/classes/class.ilFileDataMail.php
```diff
@@ -307,6 +311,8 @@ public function unlinkFile(string $a_filename): bool
     /**
      * Resolves a path for a passed filename in regards of a user's mail attachment pool,
      * meaning attachments not being sent
+     * @param string $fileName
+     * @return string
```
mjansenDatabay: The added docs don't provide any added value. We should remove them.

components/ILIAS/Mail/classes/class.ilFileDataMail.php
```diff
@@ -243,25 +245,27 @@ public function storeAsAttachment(string $a_filename, string $a_content)
...
         );
-
-        return $filename;
     }
 
     /**
      * Copy files in mail directory. This is used for sending ILIAS generated mails with attachments
+     * @param string $a_abs_path
+     * @param string $a_new_name
+     * @return bool
```
mjansenDatabay: The added docs don't provide any added value. We should remove them.

components/ILIAS/Mail/classes/class.ilAccountMail.php
```diff
@@ -149,6 +149,7 @@ private function readAccountMail(string $a_lang): array
 
     /**
      * @param array{lang?: string, subject?: string, body?: string, sal_f?: string, sal_g?: string, sal_m?: string, type?: string, att_file?: string} $mailData
+     * @return void
```
mjansenDatabay: The added docs don't provide any added value. We should remove them.

components/ILIAS/Mail/classes/Preview/class.ilStudyProgrammeMailTemplateContextPreview.php
```diff
@@ -0,0 +1,84 @@
...
+    public function resolveSpecificPlaceholder(
+        string $placeholder_id,
+        array $context_parameters,
+        ilObjUser $recipient = null,
+        bool $html_markup = false
+    ): string {
+        switch ($placeholder_id) {
+            case ilStudyProgrammeMailTemplateContext::TITLE:
+                $string = 'programme title';
+                break;
```
mjansenDatabay, line 41: Could we add a new line between each case (after the `break;`) for a better readability?

And: Shouldn't this class be moved to the `StudyProgramme` component?

components/ILIAS/Mail/classes/Preview/class.ilMailPreviewGUI.php
```diff
@@ -0,0 +1,89 @@
...
+        $this->not_translated = false;
+        $this->preview_factory = $preview_factory;
+        $this->placeholder_resolver = $DIC["mail.template.placeholder.resolver"];
+        $this->tpl = $DIC->ui()->mainTemplate();
+    }
+
+    /**
+     * Render mail preview
+     *
+     * @return string
```
mjansenDatabay: We could make the return type explicit and remoe the docs.

components/ILIAS/Mail/classes/Preview/class.ilCourseMailTemplateTutorContextPreview.php
```diff
@@ -0,0 +1,94 @@
...
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+class ilCourseMailTemplateTutorContextPreview extends ilCourseMailTemplateTutorContext
```
mjansenDatabay, line 21: Shouldn't this class be moved to the `Course` component?

Furthermore, I suggest removing obsolete docs (without any added value) and introducing a blank line between the `case` statements of the `swtich` in `resolveSpecificPlaceholder`.

components/ILIAS/Mail/classes/class.ilMimeMail.php
lukastocker: You can ignore this change. It's also a relict from a test I did. Will remove this.

components/ILIAS/Mail/classes/BackgroundTask/class.ilMassMailTaskProcessor.php
```diff
@@ -80,7 +83,7 @@ public function __construct(
      * @param string $contextId - context ID of the Background task
      * @param array $contextParameters - context parameters for the background tasks
      * @param int $mailsPerTask - Defines how many mails will be added before a background task is executed
-     * @throws ilMailException
+     * @throws ilException
```
mjansenDatabay: Do we really need to switch from `ilMailException` to `ilException` here (and in other places)? IMO this should be globally discussed (if we all agree on removing component-specific exceptions or even switch to `SPL` exceptions, which I totally support in general) and if approved, put into a separate PR.

components/ILIAS/Mail/classes/Mime/Transport/class.ilMailMimeTransportBase.php
```diff
@@ -90,11 +90,10 @@ final public function send(ilMimeMail $mail): bool
...
 
-        if ($mail->getFrom()->hasReplyToAddress() && !$this->getMailer()->addReplyTo(
-            $mail->getFrom()->getReplyToAddress(),
-            $mail->getFrom()->getReplyToName()
-        )) {
-            ilLoggerFactory::getLogger('mail')->warning($this->getMailer()->ErrorInfo);
+        if ($mail->getFrom()->hasReplyToAddress()) {
+            if (!$this->getMailer()->addReplyTo($mail->getFrom()->getReplyToAddress(), $mail->getFrom()->getReplyToName())) {
+                ilLoggerFactory::getLogger('mail')->warning($this->getMailer()->ErrorInfo);
+            }
```
mjansenDatabay: We don't need these changes for the feature, right? We should keep the PR minimal.

components/ILIAS/Mail/classes/Mime/Transport/class.ilMailMimeTransportBase.php
```diff
@@ -116,7 +115,7 @@ final public function send(ilMimeMail $mail): bool
             }
         }
 
-        if ($mail->getFinalBodyAlt() !== '') {
+        if ($mail->getFinalBodyAlt()) {
```
mjansenDatabay: We don't need these changes for the feature, right? We should keep the PR minimal.

components/ILIAS/Mail/classes/class.ilMailTemplateContextService.php
lukastocker: To be save I would also recommend to skip "file_exists" as the class always existed in the class map in my tests and the correct mail was sent. 

File_exists does  have some issues - I read somewhere that it won't work all the time - especially if you have an http adress (e.g. when a developer has a local machine running without SSL).

Would you be okay with me removing 'file_exists'?

components/ILIAS/Mail/classes/Preview/ilPreviewFactory.php
```diff
@@ -0,0 +1,32 @@
...
+    protected function getTutorContextPreview()
+    {
+        return new ilCourseMailTemplateTutorContextPreview();
+    }
+
+    protected function getPRGContextPreview()
+    {
+        return new ilStudyProgrammeMailTemplateContextPreview();
+    }
+}
```
mjansenDatabay: IMO we should not further introduce hard dependencies. Could this be solved with the "Component Revision" mechanisms?

mjansenDatabay (changes requested):

components/ILIAS/Mail/classes/class.ilMailTemplateContextService.php
mjansenDatabay: Yes, you could remove this.

components/ILIAS/Mail/classes/class.ilMailTemplateGUI.php
```diff
@@ -120,63 +136,73 @@ protected function insertTemplate(): void
...
-                ilMailTemplateContextService::getTemplateContextById($form->getInput('context'))->getId(),
-                $form->getInput('title'),
-                $form->getInput('m_subject'),
-                $form->getInput('m_message'),
-                $form->getInput('lang')
-            );
-
-            $this->tpl->setOnScreenMessage('success', $this->lng->txt('saved_successfully'), true);
-            $this->ctrl->redirect($this, 'showTemplates');
-        } catch (\ILIAS\Mail\Templates\TemplateSubjectSyntaxException) {
```
mjansenDatabay: IMO we should not remove the exception handling.

In some cases the entered `Mustache` code could not be parsed correctly during the "Mail Processing", so we tried to provide meaningful error messages when administrating the "Mail Templates", to help the administrator to provide valid text (which is supported by `Mustache`).

If we switch the implementation to the "UI/KS Form", we should provide similar "Constraints/Transformations" for this (I did not find such, but maybe I am just blind :-)).

components/ILIAS/Mail/classes/class.ilMailTemplateGUI.php
```diff
@@ -196,60 +222,42 @@ protected function updateTemplate(): void
...
+        }
+        $context = ilMailTemplateContextService::getTemplateContextById($context_id);
+        $form = $this->getTemplateForm(null, null, $context)->withRequest($this->request);
+        $result = $form->getInputGroup()->getContent();
 
-                $this->tpl->setOnScreenMessage('success', $this->lng->txt('saved_successfully'), true);
-                $this->ctrl->redirect($this, 'showTemplates');
-            } catch (OutOfBoundsException) {
-                $this->tpl->setOnScreenMessage('failure', $this->lng->txt('mail_template_missing_id'));
-            } catch (\ILIAS\Mail\Templates\TemplateSubjectSyntaxException) {
```
mjansenDatabay: IMO we should not remove the exception handling.

In some cases the entered `Mustache` code could not be parsed correctly during the "Mail Processing", so we tried to provide meaningful error messages when administrating the "Mail Templates", to help the administrator to provide valid text (which is supported by `Mustache`).

If we switch the implementation to the "UI/KS Form", we should provide similar "Constraints/Transformations" for this (I did not find such, but maybe I am just blind :-)).

components/ILIAS/Mail/tests/ilMailTaskProcessorTest.php
```diff
@@ -41,16 +41,16 @@ protected function setUp(): void
         parent::setUp();
 
         $this->languageMock = $this->getMockBuilder(ilLanguage::class)
-            ->disableOriginalConstructor()
-            ->getMock();
+                                   ->disableOriginalConstructor()
```
mjansenDatabay: We should try to avoid "Style" changes here and everywhere else to keep the PR smaller.

mjansenDatabay (changes requested):

components/ILIAS/Mail/classes/Preview/class.ilStudyProgrammeMailTemplateContextPreview.php
lukastocker: I can do that.

As for the location: it could be either way as it's part of the StudyProgramme component but also part of the mail/preview component. If we move this we should also move "class.ilCourseMailTemplateTutorContextPreview.php".

What is your opinion here?

components/ILIAS/Mail/tests/ilMailTaskProcessorTest.php
lukastocker: This change did  not happen manually but through the recommended php fixer.

components/ILIAS/Mail/classes/Preview/class.ilStudyProgrammeMailTemplateContextPreview.php
mjansenDatabay: We should move the course preview as well as the study programme preview.

components/ILIAS/Mail/tests/ilMailTaskProcessorTest.php
mjansenDatabay: Hmmm, when I execute `vendor/composer/vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix --config=./scripts/PHP-CS-Fixer/code-format.php_cs --using-cache=no --dry-run -vvv components/ILIAS/Mail/classes` I only get one violation the class `ilMail`.

lukastocker: That's odd. I will check what might have happened here.

mjansenDatabay:
@lukastocker I will be on vacation for a few weeks. I'll have time to take another/more detailed look at the PR after my return at the beginning of August.

lukastocker:
> @lukastocker I will be on vacation for a few weeks. I'll have time to take another/more detailed look at the PR after my return at the beginning of August.

Hi @mjansenDatabay ,

I hope you had a wonderful holiday!

I will also be away during August - I will probably be back at the beginning of September. If those plans change suddenly @klees will contact you.

I just pushed some changes accordingly to your comments. 

Greetings
Luka

components/ILIAS/Mail/classes/class.ilMailAppEventListener.php
```diff
@@ -27,7 +27,8 @@
  */
 class ilMailAppEventListener implements ilAppEventListener
 {
-    private readonly Container $dic;
+    private Container $dic;
```
mjansenDatabay: Why do we need to remove the `readonly` flag here?

components/ILIAS/Mail/classes/class.ilMailAppEventListener.php
```diff
@@ -68,7 +69,7 @@ public function withParameters(array $parameters): self
 
     private function isRelevantEvent(): bool
     {
-        return $this->component === 'components/ILIAS/User'
+        return $this->component === 'Services/User'
```
mjansenDatabay: If I compare this to other event listeners in the current `trunk`, this looks not correct. The schema for the event component seems to be `components/ILIAS/[COMPONENT]`.

components/ILIAS/Mail/classes/BackgroundTask/class.ilMailDeliveryJob.php
```diff
@@ -67,7 +67,7 @@ public function run(array $input, Observer $observer): Value
         $mail->setSaveInSentbox((bool) $input[8]->getValue());
         $mail = $mail
             ->withContextId((string) $input[9]->getValue())
-            ->withContextParameters($context_parameters);
+            ->withContextParameters((array) unserialize($input[10]->getValue(), ['allowed_classes' => false]));
```
mjansenDatabay: Why can't we use `$context_parameters` here?

components/ILIAS/Mail/classes/BackgroundTask/class.ilMassMailDeliveryJob.php
```diff
@@ -32,8 +32,8 @@
  */
 class ilMassMailDeliveryJob extends AbstractJob
 {
-    private readonly ILIAS\DI\Container $dic;
-    private readonly ilMailValueObjectJsonService $mailJsonService;
+    private ILIAS\DI\Container $dic;
+    private ilMailValueObjectJsonService $mailJsonService;
```
mjansenDatabay: Why did you remove the `readonly` flags?

components/ILIAS/Mail/classes/BackgroundTask/class.ilMassMailTaskProcessor.php
```diff
@@ -28,20 +28,21 @@
...
-    private readonly TaskFactory $taskFactory;
-    private readonly ilLanguage $language;
-    private readonly ilLogger $logger;
-    private readonly ilMailValueObjectJsonService $objectJsonService;
+    private TaskManager $taskManager;
+    private TaskFactory $taskFactory;
+    private ilLanguage $language;
+    private ilLogger $logger;
+    private ilMailValueObjectJsonService $objectJsonService;
+    private int $anonymousUserId;
```
mjansenDatabay: Why did you remove the `readonly` flags?

components/ILIAS/Mail/classes/BackgroundTask/class.ilMassMailTaskProcessor.php
```diff
@@ -92,11 +95,11 @@ public function run(
         $objectsServiceSize = count($mailValueObjects);
 
         if ($objectsServiceSize <= 0) {
-            throw new ilMailException('First parameter must contain at least 1 array element');
+            throw new ilException('First parameter must contain at least 1 array element');
```
mjansenDatabay: Why did you change the exception class here?

components/ILIAS/Mail/classes/BackgroundTask/class.ilMassMailTaskProcessor.php
```diff
@@ -92,11 +95,11 @@ public function run(
...
         }
 
         if ($mailsPerTask <= 0) {
-            throw new ilMailException(
+            throw new ilException(
```
mjansenDatabay: Why did you change the exception class here?

components/ILIAS/Mail/classes/BackgroundTask/class.ilMassMailTaskProcessor.php
```diff
@@ -106,9 +109,11 @@ public function run(
 
         foreach ($mailValueObjects as $mailValueObject) {
             if (!($mailValueObject instanceof ilMailValueObject)) {
-                throw new ilMailException('Array MUST contain ilMailValueObjects ONLY');
+                throw new ilException('Array MUST contain ilMailValueObjects ONLY');
```
mjansenDatabay: Why did you change the exception class here?

components/ILIAS/Mail/classes/Mime/Transport/class.ilMailMimeTransportBase.php
```diff
@@ -90,12 +90,6 @@ final public function send(ilMimeMail $mail): bool
 
         $this->getMailer()->Subject = $mail->getSubject();
 
-        if ($mail->getFrom()->hasReplyToAddress() && !$this->getMailer()->addReplyTo(
```
mjansenDatabay: Why did you remove this logging here?

components/ILIAS/Mail/classes/Mime/Transport/interface.ilMailMimeTransport.php
```diff
@@ -24,4 +24,5 @@
 interface ilMailMimeTransport
 {
     public function send(ilMimeMail $mail): bool;
+    public function getErrorInformation(): string;
```
mjansenDatabay: This method seems not be called anywhere.

components/ILIAS/Mail/classes/Mime/Transport/class.ilMailMimeTransportNull.php
```diff
@@ -31,4 +31,12 @@ public function send(ilMimeMail $mail): bool
 
         return true;
     }
+
+    public function getErrorInformation(): string
```
mjansenDatabay: This method seems not be called anywhere.

components/ILIAS/Mail/classes/Mime/Transport/class.ilMailMimeTransportBase.php
```diff
@@ -189,4 +183,9 @@ final public function send(ilMimeMail $mail): bool
 
         return $result;
     }
+
+    public function getErrorInformation(): string
```
mjansenDatabay: This method seems not be called anywhere.

components/ILIAS/Mail/classes/Object/class.ilMailValueObject.php
```diff
@@ -23,28 +23,45 @@
...
+        string $recipients,
+        string $recipientsCC,
+        string $recipientsBCC,
+        string $subject,
+        string $body,
         array $attachments,
-        private readonly bool $usePlaceholders = false,
-        private readonly bool $saveInSentBox = false
+        bool $usePlaceholders = false,
+        bool $saveInSentBox = false
```
mjansenDatabay, line 50: Why did you remove the `readonly` flag for alle the member variables?

components/ILIAS/Mail/classes/Service/MimeMailService.php
```diff
@@ -42,6 +46,29 @@ public function __construct(protected Container $dic)
...
         }
+
+        if (!isset($this->dic['mail.texttemplates.service'])) {
+            $this->dic['mail.texttemplates.service'] = static function (Container $c): ilMailTemplateService {
+                return new ilMailTemplateService(
+                    new ilMailTemplateRepository($c->database()),
+                    $c["mail.mustache.factory"]
+                );
+            };
+        }
```
mjansenDatabay: We already have `\ILIAS\Mail\Service\MailService::textTemplates` to retrieve the text templates, so I suggest to not add a duplicate registration to the `$DIC` here.

components/ILIAS/Mail/classes/Service/MimeMailService.php
```diff
@@ -42,6 +46,29 @@ public function __construct(protected Container $dic)
...
+
+        if (!isset($this->dic['mail.template.placeholder.resolver'])) {
+            $this->dic['mail.template.placeholder.resolver'] = static function (Container $c): ilMailTemplatePlaceholderResolver {
+                return new ilMailTemplatePlaceholderResolver(
+                    $c["mail.mustache.factory"]->getBasicEngine()
+                );
+            };
+        }
```
mjansenDatabay: We already have `\ILIAS\Mail\Service\MailService::placeholderResolver`, so I suggest to not add a duplicate registration to the `$DIC` here.

components/ILIAS/Mail/classes/Service/MimeMailService.php
```diff
@@ -42,6 +46,29 @@ public function __construct(protected Container $dic)
...
+
+        if (!isset($this->dic['mail.mustache.factory'])) {
+            $this->dic['mail.mustache.factory'] = static function (Container $c): ilMustacheFactory {
+                return new ilMustacheFactory();
+            };
+        }
```
mjansenDatabay: We already have `\ILIAS\Mail\Service\MailService::mustacheFactory`, so I suggest to not add a duplicate registration to the `$DIC` here.

components/ILIAS/Mail/classes/Setup/Migration/class.ilMailTemplateConvertHtmlToMarkdown.php
```diff
@@ -0,0 +1,333 @@
...
+ *
+ * Replaces all known HTML tags with their markdown equivalent
+ * and removes other HTML for fields that have been converted
+ * to markdown fields
+ */
+class ilMailTemplateConvertHtmlToMarkdown implements Migration
+{
+    public const NUMBER_OF_STEPS = 10000;
+    public const TRANLSATION_TABLE = "mail_man_tpl";
+    protected $tags = [
```
mjansenDatabay:
```suggestion
    public const TRANLSATION_TABLE = "mail_man_tpl";

    /**
     * @var list<string>
     */
    private $tags = [
```

components/ILIAS/Mail/classes/Setup/Migration/class.ilMailTemplateConvertHtmlToMarkdown.php
```diff
@@ -0,0 +1,333 @@
...
+        '</b>', '</strong>', '</i>', '</em>', '</u>',
+        '</h1>', '</h2>', '</h3>', '</h4>', '</h5>', '</h6>',
+        '</ul>', '</ol>', '</a>', '<img',
+        '<br>', '<br/>', '<br />', '</p>'
+    ];
+
+    protected ilDBInterface $db;
```
mjansenDatabay:
```suggestion
    private ilDBInterface $db;
```

components/ILIAS/Mail/classes/Setup/Migration/class.ilMailTemplateConvertHtmlToMarkdown.php
```diff
@@ -0,0 +1,333 @@
...
+        return (int) $row["count"];
+    }
+
+    /**
+     * Gets the next dataset that contains any of the
+     * tags in ilMailTemplateConvertHtmlToMarkdown::tags
+     *
+     * @returns [string|int]
+     */
+    protected function getNextObject(): array
```
mjansenDatabay:
```suggestion
    private function getNextObject(): array
```

components/ILIAS/Mail/classes/Setup/Migration/class.ilMailTemplateConvertHtmlToMarkdown.php
```diff
@@ -0,0 +1,333 @@
...
+            'm_message' => (string) $row["m_message"]
+        ];
+    }
+
+
+    /**
+     * Converts all HTML tags in ilMailTemplateConvertHtmlToMarkdown::tags to markdown
+     * and removes all other HTML tags
+     */
+    protected function convert(string $html): string
```
mjansenDatabay:
```suggestion
    private function convert(string $html): string
```

components/ILIAS/Mail/classes/Setup/Migration/class.ilMailTemplateConvertHtmlToMarkdown.php
```diff
@@ -0,0 +1,333 @@
...
+        if ($hasChildren === false) {
+            return $node->textContent;
+        }
+        return $this->convert($this->innerHTML($node, false));
+    }
+
+    /**
+     * Remove empty lines in lists for better readability
+     */
+    protected function remove_empty_lines($string): string
```
mjansenDatabay:
```suggestion
    private function remove_empty_lines($string): string
```

components/ILIAS/Mail/classes/Setup/Migration/class.ilMailTemplateConvertHtmlToMarkdown.php
```diff
@@ -0,0 +1,333 @@
...
+        $lines = array_filter($lines, function ($value) {
+            return $value !== '';
+        });
+        return implode("\n", $lines) . "\n";
+    }
+
+    /**
+     * Get the inner HTML of a DOMNode
+     */
+    protected function innerHTML(\DOMNode $n, $include_target_tag = true): string
```
mjansenDatabay:
```suggestion
    private function innerHTML(\DOMNode $n, $include_target_tag = true): string
```

components/ILIAS/Mail/classes/Setup/Migration/class.ilUpdateMailTemplatesForMustache.php
```diff
@@ -40,6 +40,7 @@ public function getDefaultAmountOfStepsPerRun(): int
     public function getPreconditions(Environment $environment): array
     {
         return [
+            new \ilIniFilesLoadedObjective(),
```
mjansenDatabay: IMO we do not need this file anymore. There is also an existing PR which removes the `Mustache` migration, see: #7827

components/ILIAS/Mail/classes/class.ilAccountMail.php
```diff
@@ -32,11 +32,11 @@
...
-    private readonly GlobalHttpState $http;
-    private readonly ilSetting $settings;
-    private readonly Refinery $refinery;
-    private readonly ilTree $repositoryTree;
-    private readonly ilMailMimeSenderFactory $senderFactory;
+    private GlobalHttpState $http;
+    private ilSetting $settings;
+    private Refinery $refinery;
+    private ilTree $repositoryTree;
+    private ilMailMimeSenderFactory $senderFactory;
```
mjansenDatabay: Why did you remove the `readonly` here?

components/ILIAS/Mail/classes/class.ilAccountMail.php
```diff
@@ -296,8 +298,8 @@ public function replacePlaceholders(string $a_string, ilObjUser $a_user, array $
             $this->http->wrapper()->query()->retrieve('target', $this->refinery->kindlyTo()->string()) !== ''
         ) {
             $target = $this->http->wrapper()->query()->retrieve('target', $this->refinery->kindlyTo()->string());
-            $tarr = explode('_', (string) $target);
-            if ($this->repositoryTree->isInTree((int) $tarr[1])) {
+            $tarr = explode("_", $target);
+            if ($tree->isInTree((int) $tarr[1])) {
```
mjansenDatabay: Do we really need this change here?

components/ILIAS/Mail/classes/class.ilFileDataMail.php
```diff
@@ -154,6 +155,7 @@ public function getAttachmentPath(string $a_filename, int $a_mail_id): string
     /**
      * Adopt attachments (in case of forwarding a mail)
      * @param string[] $a_attachments
+     * @param int $a_mail_id
```
mjansenDatabay:
```suggestion
```

components/ILIAS/Mail/classes/class.ilFileDataMail.php
```diff
@@ -30,14 +29,15 @@
  */
 class ilFileDataMail extends ilFileData
 {
+    public int $user_id;
```
mjansenDatabay: IMO we do not need this change for the `Markdown` support, please revert this change here (and in the lines below).

components/ILIAS/Mail/classes/Setup/Migration/class.ilMailTemplateConvertHtmlToMarkdown.php
```diff
@@ -0,0 +1,333 @@
...
+        }
+        $row = $this->db->fetchAssoc($res);
+        return (int) $row["count"];
+    }
+
+    /**
+     * Gets the next dataset that contains any of the
+     * tags in ilMailTemplateConvertHtmlToMarkdown::tags
+     *
+     * @returns [string|int]
```
mjansenDatabay:
```suggestion
     * @return array{tpl_id: int, lang: string, m_message: string}|array{}
```

components/ILIAS/Mail/classes/class.ilFileDataMail.php
```diff
@@ -568,7 +569,7 @@ public function deliverAttachmentsAsZip(
         if (!$isDraft) {
             $path = $this->getAttachmentPathByMailId($mailId);
             if ($path === '') {
-                throw new ilMailException('mail_download_zip_no_attachments');
+                throw new ilException('mail_download_zip_no_attachments');
```
mjansenDatabay: Why did you change the exception class here?

components/ILIAS/Mail/classes/class.ilFileDataMail.php
```diff
@@ -555,7 +556,7 @@ public function onUserDelete(): void
     /**
      * @throws ILIAS\Filesystem\Exception\FileNotFoundException
      * @throws ILIAS\Filesystem\Exception\IOException
-     * @throws ilMailException
+     * @throws ilException
```
mjansenDatabay: Why did you change the exception class here?

components/ILIAS/Mail/classes/class.ilFileDataMail.php
```diff
@@ -532,7 +533,7 @@ public function onUserDelete(): void
                     }
                 }
                 @rmdir($path);
-            } catch (Exception) {
+            } catch (Exception $e) {
```
mjansenDatabay: We do not nee the variable IMO.

components/ILIAS/Mail/classes/class.ilFileDataMail.php
```diff
@@ -496,7 +497,7 @@ public function onUserDelete(): void
                     @unlink($file->getPathname());
                 }
             }
-        } catch (Exception) {
+        } catch (Exception $e) {
```
mjansenDatabay: We do not nee the variable IMO.

components/ILIAS/Mail/classes/class.ilFileDataMail.php
```diff
@@ -30,14 +29,15 @@
  */
 class ilFileDataMail extends ilFileData
```
mjansenDatabay: IMO all changes of this class can be reverted.

components/ILIAS/Mail/classes/class.ilMailAttachmentGUI.php
```diff
@@ -18,71 +18,57 @@
 
 declare(strict_types=1);
 
+use ILIAS\HTTP\GlobalHttpState;
```
mjansenDatabay: It seems you reverted all changes relevant for the implementation of the ILIAS 10.x feature https://docu.ilias.de/go/wiki/wpage_8069_1357 .

components/ILIAS/Mail/tests/ilMailTaskProcessorTest.php
mjansenDatabay: Thanks for caring. There still seem to be many changes files due to code style changes.

components/ILIAS/Mail/classes/class.ilMail.php
```diff
@@ -32,91 +30,94 @@ class ilMail
...
+        ilLanguage $lng = null,
+        ilFileDataMail $mailFileData = null,
+        ilMailOptions $mailOptions = null,
+        ilMailbox $mailBox = null,
+        ilMailMimeSenderFactory $senderFactory = null,
+        callable $usrIdByLoginCallable = null,
+        int $mailAdminNodeRefId = null,
+        ilObjUser $actor = null,
+        ilMailTemplatePlaceholderResolver $placeholder_resolver = null,
+        ?Conductor $legal_documents = null
```
mjansenDatabay: Can we add the constructor property promotion again? It is tough to identify the necessary changes. Of course the class is still too big, but IMO a lot of your changes are not related to the `Markdown` functionality.

components/ILIAS/Mail/classes/class.ilMail.php
```diff
@@ -32,91 +30,94 @@ class ilMail
...
     ) {
         global $DIC;
         $this->logger = $logger ?? ilLoggerFactory::getLogger('mail');
-        $this->mail_address_type_factory = $mail_address_type_factory ?? new ilMailAddressTypeFactory(null, $logger);
-        $this->mail_address_parser_factory = $mail_address_parser_factory ?? new ilMailRfc822AddressParserFactory();
-        $this->event_handler = $event_handler ?? $DIC->event();
+        $this->mailAddressTypeFactory = $mailAddressTypeFactory ?? new ilMailAddressTypeFactory(null, $logger);
```
mjansenDatabay: Why did you rename the variables to `camelCase` here, above and below?

components/ILIAS/Mail/classes/class.ilMail.php
```diff
@@ -1364,17 +1402,42 @@ public static function _getIliasMailerName(): string
...
+            $DIC['ilClientIniFile']->readVariable('client', 'description'),
+            $signature
+        );
+        $signature = str_ireplace('[ILIAS_URL]', $clientUrl, $signature);
+
+        if (!preg_match('/^[\n\r]+/', $signature)) {
+            $signature = "\n" . $signature;
+        }
+
+        return $signature;
```
mjansenDatabay: These changes seem to revert https://docu.ilias.de/go/wiki/wpage_8142_1357 .

components/ILIAS/Mail/classes/class.ilMail.php
```diff
@@ -125,12 +126,12 @@ public function withContextParameters(array $parameters): self
...
         $clone = clone $this;
 
-        $clone->context_parameters = $parameters;
+        $clone->contextParameters = $parameters;
 
         return $clone;
     }
 
-    private function isSystemMail(): bool
+    protected function isSystemMail(): bool
```
mjansenDatabay: Why do we need all these access modifier changes?

components/ILIAS/Mail/classes/class.ilMail.php
```diff
@@ -144,7 +145,7 @@ public function existsRecipient(string $newRecipient, string $existingRecipients
 
         $diffedAddresses = $list->value();
 
-        return $diffedAddresses === [];
+        return count($diffedAddresses) === 0;
```
mjansenDatabay: Please revert this.

components/ILIAS/Mail/classes/class.ilMail.php
```diff
@@ -157,7 +158,7 @@ public function getSaveInSentbox(): bool
         return $this->save_in_sentbox;
     }
 
-    private function readMailObjectReferenceId(): void
+    protected function readMailObjectReferenceId(): void
```
mjansenDatabay: Please revert this.

components/ILIAS/Mail/classes/class.ilFormatMail.php
```diff
@@ -24,6 +24,11 @@
  */
 class ilFormatMail extends ilMail
 {
+    public function __construct(int $a_user_id)
```
mjansenDatabay: IMO all changes in this class could be reverted.

components/ILIAS/Mail/classes/class.ilMailAttachmentStakeholder.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public function getId(): string
+    {
+        return 'mail_attachments';
+    }
+
+    public function getOwnerOfNewResources(): int
+    {
+        return 74;
```
mjansenDatabay: Hmmmm, if `42` is the answer of all questions, what is `74` :-)?

components/ILIAS/Mail/classes/class.ilMailFormGUI.php
```diff
@@ -75,7 +88,14 @@ public function __construct(
...
         $this->mbox = new ilMailbox($this->user->getId());
         $this->purifier = $bodyPurifier ?? new ilMailBodyPurifier();
-        $this->ui_factory = $DIC->ui()->factory();
+        $this->ui_factory = $ui_factory ?? $DIC->ui()->factory();
+        $this->request = $DIC->http()->request();
+        $this->ui_renderer = $ui_renderer ?? $DIC->ui()->renderer();
+        $this->post = new ArrayBasedRequestWrapper($this->request->getParsedBody());
+        $this->query = new ArrayBasedRequestWrapper($this->request->getQueryParams());
+        $this->upload_handler = new ilMailFormUploadHandlerGUI();
+        $this->storage = $DIC->resourceStorage();
```
mjansenDatabay: The "IRSS" seems to be not used in this class.

mjansenDatabay:
Hi @lstocker,

thanks a lot for your latest changes.

After a further review, I still noticed a lot of (in most cased trivial) issues.

- [x] Please revert all changes that are not really necessary for the feature implementation. I added inline comments for a lot of these issues, but of course not for all. These changes make it tough to spot the actual changes required for the `Markdown` support.
- [x] It seems this PR affects/reverts/breaks features already implemented and approved/integrated for ILIAS 10.x. Please have a look at this.
    1. Improvement of Mail Signature: https://docu.ilias.de/go/wiki/wpage_8142_1357 (see: `\ilMail::_getInstallationSignature`)
    2. Manage Attachments: https://docu.ilias.de/go/wiki/wpage_8069_1357 (see: primarily `\ilMailAttachmentGUI`)
- [x] The migration code in `ilMailTemplateConvertHtmlToMarkdown::convert` looks dangerous to me when manipulating existing mail templates, and if we don't have any tests which cover the DOM manipulations, I am terrified to approve these changes :). Please provide proper tests for these DOM manipulations (and maybe extract the algorithms to new testable classes).
- [x] The form for composing/creating new mails seems to be migrated to the "UI/KS > Inputs". Besides the collision with the "Manage Attachments" feature (mentioned above), I noticed that no "Autocomplete/Type ahead" is available anymore for TO/CC/BCC fields. I believe our user base is not aware of this change in ILIAS 10.x, and I could not find any comment about this in the feature request. I fear this has to be discussed as "Development Issue" if we really have to to without it.
- [x] What I do not get yet is the partial introduction of "IRSS" based file handling. A new stakeholder is introduced but seems to be only used during the uploading handling. When sending/submitting the message, the mails are stored in the filesystem again in `\ilMailFormGUI::handleAttachments` (and deleted afterwards). Please elaborate a bit why this introduction was necessary and how this relates to the concept of the "Attachment Stage" (see: `ilMailAttachmentGUI` and `ilFileDataMail`) and the final mail attachments.
- [ ] I could not find any differentiation between existing internal messages and messages created with `Markdown` in the output/rendering for the stored mail records. In my understanding, we do need either a setup migration or different implementation for `Markdown`-enabled mails and legacy mail records of database table "mail". The latter could contain raw HTML, so `htmlspecialchars` was used to ensure a secure output handling. Please elaborate on this.
- [x] When trying to create text templates in the global mail administration, ILIAS responds with PHP errors. Please fix them.
- [x] The message content when composing/creating new mails is marked as a required field. Please make the message content optional again (like it was in former ILIAS versions and in most email clients).
- [ ] We should move the `UI` and `Refinery/Markdown/CommonMarkExtension` component changes to a separate PR.
- [ ] Please ensure that the attachment handling works for all scenarios. I noticed unexpected behavour in the workflows below.
    1. Create/Compose a new mail. Save the form with 1-N attachments as a draft. Open the draft. All attachments should be listed. At the moment It seems that data is lost during navigation.
    2. Create/Compose a new mail. Add 1-N attachments. Add further recipients by using the search buttons above the form. Return to the mail form. All previously attached files should be listed. At the moment the attachment-indicating glyph is rendered in the draft folder table, but no attachments are listed when the draft is edited again.

I suggest proceeding as follows:

1. Revert all code style changes and other irrelevant changes to **really provide the minimum change set**.
2. Fix the PHP errors (like these in the text template administration) and attachment-related issues.
3. Address the collision of your changes and the "Manage Attachment" and "Improvement of Mail Signature" functionalities.
4. Add unit tests for the text template migration.

Afterwards we should have a chat about the UI/KS > Inputs, especially regarding the missing "Autocomplete/Type ahead" feature add see how we proceed from this state.

Best regards,
Michael

mjansenDatabay (changes requested):

components/ILIAS/Mail/classes/Setup/Migration/class.ilUpdateMailTemplatesForMustache.php
lukastocker: That's good to know. Thank you 👍🏻 :)

components/ILIAS/Mail/classes/Object/class.ilMailValueObject.php
lukastocker: With the changes coming soon I will set 'readonly'  above the constructor for the member variables instead of inside of the constructor for regularity purposes.

components/ILIAS/Mail/classes/class.ilMailAttachmentGUI.php
lukastocker: Thanks for pointing that out - that def. should not have been happened. I will revert this!

lukastocker:
Hi @mjansenDatabay ,

I have pushed some changes regarding our discussion.

Please note following:
1. Saving Attachments in Drafts is still not working. This issue is connected to the IRSS Upload functions. 
2. Editing Attachments is now available through the "Edit Attachments" Button in the toolbar. But if you change an attachment no name and size will be shown as the contents get lost through the IRSS process. There is currently no solution for this. 
3. BCC, TO and CC fields are now handled as a Tag Input field instead of a Text Input field. Searching a user will show all available matching user for selection.
4. I have solved most conflicts but not pushed the changes because I want to wait for your feedback regarding the mail markdown changes. As soon as those are good to go I will check if more conflicts were created through some updates in ILIAS from the last days.

Greetings
Luka

mjansenDatabay:
Hi @lstocker,

thanks for providing the changes.

> Saving Attachments in Drafts is still not working. This issue is connected to the IRSS Upload functions.

Please elaborate on this. What exactly is technically preventing us from doing this?

> ... no name and size will be shown as the contents get lost through the IRSS process. There is currently no solution for this.

I could not really test this (the file upload in the "Manage Attachments" view did not work, I guess it will work if you rebase your branch on the latest "trunk" changes).

But: What do you exactly mean by this? I am sure we can somehow retrieve "File Size" and "Filename" for a resource from the IRSS interfaces.

Have a look at the global IRSS resource overview.

Resource-Overview-File-Services-ILIAS-10-07-2024_09_14_AM

The IRSS seems to still now the original name and the size of a resource.

> BCC, TO and CC fields are now handled as a Tag Input field instead of a Text Input field. Searching a user will show all available matching users for selection.

This looks much better now, but (of course) we cannot put a JSON encoded list of all users into the rendered HTML page. As we discussed in our meeting, the "Tag Input" component has to provide an "Async. Data Source" option to retrieve the list of matches via the `XHR` object asynchronously. `\ilMailForm::getRecipientAsync` has to be then used as "Data Source".

With the current state of the feature branch, I am even not able to request the HTML source code of the "Mail Compose" page in my browsers (Firefox, Chrome), just because the JSON string for the "Tag Input" is so huge. There are ~ 4000 user accounts on my local test installation.

---

One further issue I noticed:

- [ ] Please re-introduce the option to let the user decide whether he/she wants to send the mail as a "Serial Letter". Always enabling the placeholder interpolation (implicitly) for every mail would have a serious performance drawback (it would enforce each recipient to retrieve an individual message, either only internally, only externally, or for both channels). It should be still possible to send a message which physically results in one external email only (by using multiple BCC recipients). The "Serial Letter" state should therefore be disabled per default like in ILIAS 9.x, and it should be possible to activate it optionally.

Erstellen-Mail-ILIAS-10-07-2024_09_24_AM

For me it would be okay to have a "Mode" component (in the toolbar) to switch between both views. Pleas keep in mind that switching the mode should not result in a any loss of the data the displayed form is already populated with.
And: When coming from a context supporting a "Mail Text Template" (e.g. a course), the mode should be "Serial Letter" with the option to let the user decide which "Text Template" he/she likes to use for the given "Mail Template Contect".

Best regards,
Michael

klees:
Hi everyone,

I discussed this with @mjansenDatabay and we decided not to push this for ILIAS 10 but instead try to integrate it with ILIAS 11 early on. This is on a good way, but the remaining issues seem to big to be tackled in the last 2 weeks of this development cycle. @lukastocker will keep on working on this once other ILIAS 10 stuff is on its way.

Kind regards!

lukastocker:
Suggestion of this implementation was moved to ILIAS11 (trunk). => https://github.com/ILIAS-eLearning/ILIAS/pull/9423

PR #7780, opened 2024-07-05
Markdown: add underline extension.

lukastocker:
This PR is a suggestion and should not get reviewed and merged before this [PR](https://github.com/ILIAS-eLearning/ILIAS/pull/7779).

This PR includes the possibility to use 'underline' as markdown as it's not included in the core.
It's also linked to PR [Adjust Headline Config](https://github.com/ILIAS-eLearning/ILIAS/pull/7781) which is needed for this PR. Else there might be some errors during the rendering process of the markdown.


lukastocker:
Suggestion of this implementation was moved to ILIAS11 (trunk).

PR #7781, opened 2024-07-05
Markdown: adjust markdown's headline configuration

lukastocker:
This PR removes '-' and '=' in the headline configuration. As a result only '#' can be used as a markdown headline.
This avoids rendering of false headlines, e.g.:
- the user uses '-' to separate a headline or an email => 'Topic 1 - How to create a course'
- the user uses '-' to separate parts of an email like this '-----------'

This is also needed for the suggested [underline extension PR](https://github.com/ILIAS-eLearning/ILIAS/pull/7780) as it uses '--'.


lukastocker:
Suggestion of this implementation was moved to ILIAS11 (trunk).

PR #7783, opened 2024-07-08
OrgU: 41493, move temporary table into subquery

nhaagen:
https://mantis.ilias.de/view.php?id=41493

### DIFF
--- a/components/ILIAS/OrgUnit/classes/PathStorage/class.ilOrgUnitPathStorage.php
+++ b/components/ILIAS/OrgUnit/classes/PathStorage/class.ilOrgUnitPathStorage.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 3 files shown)


klees:
Hi @nhaagen,

thanks for looking into this and providing this code. I've given this some thought and decided to keep that temporary table as is. If I understand everything correctly, the temporary table is a hack to improve performance. If there are subsequent requests into the memberships of orgunits in one request, we effectively create that table once and query it multiple times, which saves some work. Also, this sometimes causes errors, as described by our customers.

So, this is a decision where we could either make everyone slower with these requests and have no errors that appear sometimes. Or keep speed for everyone as is, but error sometimes.

This surely is a hack, I don't think we use temp tables anywhere else and, also, they are not part of our db-abstraction in any way. Which is sad. But I also expect people to yell at me if the system gets slower because we simply replace the stuff with inline queries as proposed.

A correct solution, from my POV, would be to implement a proper repository for requests into the org units and memberships, which then could be used to implement appropriate caching strategies. But I doubt that we will get there any time soon.

If there is someone that could provide resources to improve the state of the org units, I would be very glad to help. Maybe, at some point in the future, we are enoyed enough to provide resources ourselves. But for the moment I think this need to stay as is.

Kind regards!

PR #7785, opened 2024-07-08
10/UI/refinery/uri

klees:
Dear colleagues,

prompted by these two PRs from @iszmais (#7751, #7573) @thibsy and me had a look into the Refinery-dependency of the rendering in the UI framework. And we found: We do not really need it. Hence we can remove some surface from our internals, which we consider a good thing.

This is the according code. Also, I found some duplicated code in `Data\URI` / `Refinery`, which I removed as well (ping @mjansenDatabay).

Kind regards!


klees:
Hi @mjansenDatabay,

I think the change in the `Refinery` is trivial, but please have a look anyway =)

Kind regards!

mjansenDatabay (approved):
LGTM @klees , this transformation looked exactly like the impl. in `\ILIAS\Data\URI::__toString` :-).

thibsy (approved):
Hi @klees,

Thx a lot for this PR. Nice to see how one usage of refinery transformations could get rid of so many dependency injections :).

The changes LGTM are somewhat trivial, so I will merge right away.

Kind regards,
@thibsy

PR #7787, opened 2024-07-09
T&A 41620: Fix pasting crash after copying via bulk actions

aaronbidzan:
This PR aims to address the problem reported in Mantis issue https://mantis.ilias.de/view.php?id=41620.

Copying questions from a question pool using the “Actions for Entire Table” button found under Bulk Actions and trying to paste them now results in the correct behavior. This has been accomplished by adding a check for ALL_OBJECTS which inserts all the needed IDs accordingly.

Kind regards.


thojou:
Hi @bidzanaaron,

I reworked your changes slightly to ensure they are applied after an is_null verification. 
I've merged your PR into release_9 and trunk.

Thank you!

PR #7788, opened 2024-07-09
UI/Modals: display modals using dialog-tag

nhaagen:
https://mantis.ilias.de/view.php?id=41670


nhaagen:
There are still issues with the lightbox modals, they add js in the template which is currently not executed.

Amstutz:
Hi @nhaagen,

Thanks a lot for your contribution to the UI components and for getting rid of jQuery-UI here! Its also nice to see how few PHP changes were required to refactor this :).

- [x] `data-dismiss` attributes: there are still some of these attributes inside the HTML templates. I think they are obsolete now. Please check this and remove them if so.
- [x] `tpl.lightbox.html`: there is JavaScript code inside the inline `<script>` tag, which listens to some jQuery-UI modal events, I think. If so, please remove this code as well.
- [x] Code-style I: please rename the `modal.class.js` file to `Modal.js` to match the files default export. Due to naming collision, please also rename `modal.js` to `index.js` or similar.
- [x] SCSS: You added a new pattern folder in 030-tools. We might add such a folder, but are not sure yet. Please remove it ftm and add the file into 030-tools directly.

- [x] `id`-parameter: please only accept the corresponding `HTMLElement` instead of their ids. This way, we can get rid of any global `document` usage, and use the elements associated [`HTMLElement.ownerDocument`](https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument) instead. If you need to perform `instanceof` checks to check for `<dialog>` elements, please do so using [`HTMLElement.ownerDocument.defaultView.<object>`](https://developer.mozilla.org/en-US/docs/Web/API/Document/defaultView).
- [ ] Code-style II: `Modal.closeModal()` and `Modal.replaceFromSignal()` violate the code-style because they are not using `this` in a class context. Please extract these methods 
  - [ ] (a) into their own dedicated files, which default export a function and accept all state as arguments (`document`, `il.UI.core`), or
  - [x] (b) into `modal.js` as inline functions when you create the global `il.UI.modal` object.
- [x] Null-pointers: there are several possible null-pointers inside `modal.class.js` where you invoke functions chained to `Document.getElementById()`. Please either check the returned value  and/or use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

Kind regards,
@Amstutz and @thibsy

klees:
Hi @Amstutz and @thibsy,

I think you could move on here...

Kind regards!

Amstutz:
Hi @nhaagen 

Thx a lot for the PR and for the iteration of improvement. Please resolve the conflict, then we will directly merge.

Thx

nhaagen:
🐿 shipit?

Amstutz:
thx!

PR #7790, opened 2024-07-09
Deps: Remove Inline Libraries

klees:
Hi JF,

these libraries have never been approved by the JF via the process that we use for this purpose:

https://github.com/ILIAS-eLearning/ILIAS/tree/trunk/vendor#adding-a-new-dependency-to-ilias

Also, the libraries are not added via the package manager. Hence we should remove them for the time being.

Kind regards!


matthiaskunkel:
Jour Fixe, 22 JUL 2024: Decision about PR is postponed. Product manager wants to discuss procedure with TB first. Jour Fixe label is kept.

klees:
This is on the way to a proper integration. Thx @jcopado and team.

PR #7791, opened 2024-07-09
8/UI/mejs-videoplayer keyboard focus ILIAS style 34185

catenglaender:
https://mantis.ilias.de/view.php?id=34185
PR für ILIAS 9: https://github.com/ILIAS-eLearning/ILIAS/pull/7794

# Issue

Focus in video player is low contrast and not uniform with the rest of ILIAS.

videoplayer_focus-incorrect

# Change

Our il-focus mixins are now used to highlight player elements.

videoplayer_focus-correct

# Outlook

Maybe we need an il-focus-thin mixin with a 1px white + 1 px blue + 1px white line that leaves selected icons more readable? UPDATE: Implemented in IL 9 version of this PR: https://github.com/ILIAS-eLearning/ILIAS/pull/7794

### DIFF
--- a/templates/default/delos.css
+++ b/templates/default/delos.css
@@ -16317,6 +16317,50 @@ span.bibl_text_inline_Emph {
   width: 100px;
   cursor: pointer;
 }
+.mejs__overlay-button:focus-visible,
+.mejs__container:focus-visible,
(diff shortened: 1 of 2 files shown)


Amstutz:
thx a lot. I thinkg we take this as is. If there is further feedback, we can make an other iteration.

PR #7793, opened 2024-07-10
T&A 35591: Add blue message boxes in sub-tabs Taxonomy of qpl, glo and cat to describe the function

matheuszych:
A blue message box with additional information about taxonomies is missing in the question pool view.

[Mantis: 35591](https://mantis.ilias.de/view.php?id=35591)


thojou (approved):
hey @dsstrassner,

Looks good to me :)

alex40724:
Thanks for the PR.

Please note that the taxonomy service must not be used to output consumer related messages this way. This must be solved on the consumer side. There is an interface for that, see https://github.com/ILIAS-eLearning/ILIAS/blob/release_9/Services/Taxonomy/README-technical.md#forwarding-to-the-taxonomy-settings

The second parameter of getSettingsGUI allows to pass a message.

PR #7794, opened 2024-07-10
9/UI/mejs-videoplayer keyboard focus ILIAS style 34185

catenglaender:
Mantis: https://mantis.ilias.de/view.php?id=34185
PR für ILIAS 8: https://github.com/ILIAS-eLearning/ILIAS/pull/7791

# Issue

Focus in video player is low contrast and not uniform with the rest of ILIAS.

videoplayer_focus-incorrect

# Change

Our il-focus mixins are now used to highlight player elements.
il-focus mixins now take parameters so a thinner focus line can be constructed for the small player icons.
HTML element button now has focus applied by default so dependencies should take ILIAS focus outlines unless they override them.

videoplayer_focus-correct_il9


Amstutz:
thx a lot!

PR #7796, opened 2024-07-11
UI PresentationTable: 41643, revise and simplify expanding/collapsing of rows

nhaagen:
https://mantis.ilias.de/view.php?id=41643


klees (approved):
LGTM

klees:
@nhaagen This would also apply to trunk, right?

nhaagen:
> @nhaagen This would also apply to trunk, right?

yes, the differences are in the CSS for 8/9, i.e. LESS vs. SASS.

Amstutz:
Picked manually due to conflicts in the delos.css.map

PR #7800, opened 2024-07-11
T&A 40598: Instant Feedback in Kprim Question

malaja11:
Mantisticket: https://mantis.ilias.de/view.php?id=40598

In Mantis the proposed solution was to  remove the feedback option "Show answer specific feedback for all chosen/selected answers." in KPrim-Questions. 
We determined that this feedback option is also in use in Single and Multiplechoice Question. In agreement with @dsstrassner we decided to try and fix the bug rather than removing the functionality.


thojou:
Hey @dsstrassner,

Since there is a decision made at the Jour Fixe about removing this feature only for the KPrim question, do we need to add the "jour fixe" label to this PR and wait for approval?

dsstrassner:
Hi @malaja11 thanks for your investigation and the fix. @thojou please proceed and merge / cherry-pick this fix for ILIAS 8/9 and Trunk. 
We will inform the JourFixe within the Mantis Issue, but as we have a better solution as formerly proposed, I think we do not need to wait until then with merging!

cc @yvseiler ;-)

thojou:
hey @malaja11,

I took your PR and integrated the changes into trunk.

Thank you.

PR #7801, opened 2024-07-12
UI, Object: remove dropzones for items in manage-view

nhaagen:
https://mantis.ilias.de/view.php?id=41294

currently, there may not be any < form >s in the content of repository's "manage"-view.
This removes dropzones from items (which generate form tags in their Modals).


kergomard:
As this is an issue for all currently supported versions and your solution is only viable for ILIAS 8, I went with a more direct approach in 70161a77ae69f324ea9782525004467c9e0c55b2 .

Thanks again and best,
@kergomard

PR #7808, opened 2024-07-15
Remove ilObjGroup `getSubItems` implementation

iszmais:
https://mantis.ilias.de/view.php?id=41719

This function does nothing indifferent to the parent function, except calling an undefined method which results into an error.

Through this the `Group` will use the parent/container `getSubItem()` function just like any other container


alex40724:
@iszmais Sounds reasonable. Thanks for the PR.

PR #7810, opened 2024-07-16
T&A 41371: Fix question skill tree assignment table

thojou:
This PR is a follow-up to #7730 and the Mantis ticket https://mantis.ilias.de/view.php?id=41371.

As discussed in the previous PR, the SCSS/CSS has been modified to apply extra spacing only to the checkbox if it is inside the first column of a row. This should a) maintain existing behaviors in the legacy UI tables and b) fix the representation for tables without bulk actions.

Additionally, the competence path has been changed to display the tree name at the beginning of the path. This clarifies where the competences are coming from.

Bildschirmfoto vom 2024-07-16 15-01-33

@catenglaender,
Could you please verify if I meet the requirements? I checked that I used the latest version of Sass, but I still got plenty of changes that are not related to the changes in my commit.


catenglaender (approved):
This is the perfect solution SCSS wise for the legacy table. It's very much appreciated that you went for the clean solution. Thank you very much!

About compiling SCSS to CSS... this is very strange... @Amstutz have you seen something like this before? Something happened with the view control buttons .btn-ctrl mixin styling. If you don't have an inkling what this is I could investigate tomorrow as the viewcontrol styling is my code.

Last but not least:

- [ ] please remove the map.css file from the commit (why on earth is this still here? Didn't we put it into the .gitignore?)

That's it CSS wise. Thank you for your continued effort. My apologies for the nitpicking, but these really are the little strange things that can be so annoying when working on a custom skin for ILIAS. One little detail less to worry about.

catenglaender:
UPDATE: the generated scss is correct. I think the css before is the result of a typo that was corrected in the scss but then css wasn't compiled.

thojou:
Hello @catenglaender,

I appreciate your feedback. For me, this was a good learning experience, so I'm happy about the nitpicking. I have removed the delos.css.map file from my commit.

If a file was added to the Git history once and then added to the ignore file afterward, changes to this file will still be tracked by Git. Therefore, you have to remove it from the entire history with git rm --cached <FILE>, which I just did. In the future, this file should not be included in any commits anymore.

oliversamoila (approved):
Looks good to me.

dsstrassner (approved):
@oliversamoila and I looked on this branch on a local installation and from our point of view this looks fine.

PR #7811, opened 2024-07-16
Fix PublicAssetManager's path validation to also accept dots within path naming

thojou:
Currently, we are working on the initial setups based on the ILIAS trunk for development and testing purposes and have encountered an issue during the setup.

In our deployment, we use path-naming conventions that include "dots," e.g., "/var/www/demo10.ilias.de." We heavily rely on this naming convention, but currently, the "PublicAssetManager" does not allow such naming.

With this PR, I adjusted the regex to also allow dots inside the path. I also adjusted the regex to only allow dots and dashes inside the path if they are surrounded by other characters. Dots and dashes at the beginning or end, or without any other characters between the slashes, are still disallowed.

I also think that this issue is only a side effect. In my understanding, the main purpose of validating the path is to ensure that the user has inserted a correct configuration. With that said, as long as the path is valid, this regex should not complain about special characters like dots or dashes.

### DIFF
--- a/components/ILIAS/Component/src/Resource/PublicAssetManager.php
+++ b/components/ILIAS/Component/src/Resource/PublicAssetManager.php
@@ -25,7 +25,7 @@ namespace ILIAS\Component\Resource;
  */
 class PublicAssetManager
 {
-    public const REGEXP = '%^(/[\w-]+)+$%';
+    public const REGEXP = '%^(/\w+|\w+[\-.]\w+)+$%';
 
     public const DONT_PURGE = ["data", "Customizing"];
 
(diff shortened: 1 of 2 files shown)


klees (changes requested):
Hi @thojou,

thanks a lot. And yes, you are right, this is basically a sanity checks that this is indeed something like a valid path and not some gibberish =) I would much prefer if we had according data types for paths, but we currently don't have any and I skipped building one for the moment. So this is fine, I guess.

One minor change request: I don't think `testInvalidFolderPaths` does what you think it does: It will only really test the first case because that one will throw an exception. Please use a data provider for the various cases.

Since I'm in vacation for about a week, I kindly ask @mjansenDatabay to merge this if this is somehow timecritical.

Kind regards!

thojou:
Hey @klees,

Thank you for your feedback. I changed the test case for invalidPath to use a data provider. I'm not sure why I missed that, but I completely forgot that even though the tests pass, an exception is still thrown, which stops the following code from executing.

Thank you for taking the time to point that out. Have a great vacation.

PR #7812, opened 2024-07-17
Introduce keys method to RequestWrapper

thojou:
During the refactoring of legacy global accesses to `$_GET` and `$_POST`, we discovered some special use cases that are currently not implementable with the existing state of the request wrapper.

There are parts of the code that require looping over global variables to access multiple values based on a key pattern or to forward parameters to another request.

To support these special use cases via the new `RequestWrapper` API, a new method, `keys`, has been introduced. This method returns all parameter names from the respective `$_GET`, `$_POST`, or  `$_COOKIE` requests.

Using this method, the user can loop over the list of keys, apply their conditions, and still use the `retrieve` method to get the actual value, possibly applying some sort of transformer.

The most simplistic use case is shown in the following example:

**Before**

```php
foreach ($_POST as $key => $value) {
   // access $value
}
```

**After**

```php
global $DIC;

$postWrapper = $DIC->http()->wrapper()->post();
$refinery = $DIC->refinery();

foreach ($postWrapper->keys() as $key) {
    $value = $postWrapper->retrieve($key, $refinery->identity());
    // access $value
}
```

It should be noted that using the "identity" transformation is not ideal and should be replaced with something more meaningful that adds real validation to the transformation.

This PR is also related to #7803.

### DIFF
--- a/components/ILIAS/HTTP/src/Wrapper/ArrayBasedRequestWrapper.php
+++ b/components/ILIAS/HTTP/src/Wrapper/ArrayBasedRequestWrapper.php
@@ -58,4 +58,12 @@ class ArrayBasedRequestWrapper implements RequestWrapper
     {
         return isset($this->raw_values[$key]);
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function keys(): array
+    {
+        return array_keys($this->raw_values);
+    }
 }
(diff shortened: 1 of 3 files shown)


chfsx:
Thanks a lot, I merged this locally rebased

PR #7815, opened 2024-07-18
Phpunit code coverage

thojou:
This PR enhances the current phpunit.xml file with a source directive to allow PHPUnit to generate code coverage during test execution.

Furthermore, a new entry has been added to the .gitignore file because the path for the PHPUnit cache results has changed following the recent migration to the new XML structure.

### DIFF
--- a/scripts/PHPUnit/phpunit.xml
+++ b/scripts/PHPUnit/phpunit.xml
@@ -1,8 +1,29 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.5/phpunit.xsd" bootstrap="../../vendor/composer/vendor/autoload.php" colors="true" beStrictAboutOutputDuringTests="true" beStrictAboutTestsThatDoNotTestAnything="true" failOnRisky="true" failOnWarning="true" backupGlobals="true" executionOrder="random" cacheResultFile=".phpunit.cache/test-results">
-  <testsuites>
-    <testsuite name="all">
-        <directory>../../components/*/*/tests/</directory>
-    </testsuite>
-  </testsuites>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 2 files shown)


scripts/PHPUnit/phpunit.xml
```diff
@@ -1,8 +1,29 @@
...
+        </testsuite>
+    </testsuites>
+    <source>
+        <include>
+            <directory suffix=".php">../../components/*/*/</directory>
+        </include>
+        <exclude>
+            <directory>../../components/*/*/tests/</directory>
+            <directory>../../components/*/*/templates/</directory>
+            <directory>../../components/*/*/resources/</directory>
```
mjansenDatabay, line 26: Maybe we should add `libs` and `lib` here (e.g. for `components/ILIAS/Wiki/libs` and `components/ILIAS/UICore/lib`) as well as`artifacts`.

klees:
Hi @thojou,

thanks a lot! I would not do as @mjansenDatabay recommends, as these folders are in fact _our_ code (as we do not pull it via composer and hence do not get any updates) and we do indeed need to take care about them.

Just a question for clarification: This won't create code-coverage reports per se, just open up the possibility, right? Code coverage would decrease speed a lot, hence the question.

Kind regards!

thojou:
Hi @klees,

thanks for the question!

To clarify: adding the `<source>` directive to `phpunit.xml` does not, by itself, generate code coverage reports during test execution - it simply enables the possibility of generating coverage if you explicitly request it.

Whether coverage is actually collected depends on how you run your tests and your PHP environment:

* In IDEs like *PhpStorm*, you explicitly need to choose *Run with Coverage* - normal test runs without that option won’t generate coverage, even though the `<source>` directive is present.
* On the CLI, you need to ensure your PHP is configured for code coverage. If you’re using Xdebug, for example, you must have the environment variable `xdebug.mode=coverage` defined when running PHPUnit. Other coverage engines (like `pcov`) work slightly differently, but the principle is the same: coverage must be enabled in the environment explicitly.

So, test execution performance will remain unaffected unless your PHP is configured to generate coverage at runtime. In other words, having the `<source>` in `phpunit.xml` alone does not slow down tests - it just makes coverage collection possible when requested.

Hope that clears it up!

Kind regards!

klees:
I have merged this manually and applied the suggestions from @mjansenDatabay as well.

Thanks for your patience.

PR #7817, opened 2024-07-18
RessourceStorage: Add primary to ressource table

iszmais:
https://mantis.ilias.de/view.php?id=41750

This adds a missing primary to increase the performance.
The key seems to be set in past as an index (ILIAS 6 maybe?) but is unset on newly installed/not migrated installations.


chfsx:
Thanks a lot, strange that those keys are not present in newly installed installations

iszmais:
@chfsx As far as I can tell, the table is a reuse of a different table, and therefore the keys are not intentionally in the current Ilias versions.

Therefore, I assume this also needs to be merged into ILIAS 9 and trunk.

Greetings,
@iszmais

PR #7827, opened 2024-07-18
Mail: Remove legacy migration for mail templates

iszmais:
Removing of legacy migration inside the mail.

Softly depends on https://github.com/ILIAS-eLearning/ILIAS/pull/7926


iszmais:
@kergomard conceptionaly i would strongly assume that this also affects the similar migration inside the user service.

Feel free to approve or to correct me on that.

mjansenDatabay (changes requested):
Thx @iszmais , please create a separate PR for the "User" component.

iszmais:
@mjansenDatabay I split the parts for review purposes.

However, since this is conceptionally entwined, I would suggest merging both PRs in dependency to another.

Greetings,
@iszmais

mjansenDatabay (approved):
LGTM!

PR #7836, opened 2024-07-20
Fix skin path in html exports

padvincenzo:
Hi, currently in learning modules HTML exports, all the images, fonts, etc. are copied into the `templates/default` path of the export, but exported modules uses the skin path, and so they are never found.

Before modification:

Screenshot 2024-07-20 125646

After:

Screenshot 2024-07-20 125700

I know it might not be the best solution, I would just like to point out this problem.

### DIFF
--- a/Services/Style/System/classes/class.ilSystemStyleHTMLExport.php
+++ b/Services/Style/System/classes/class.ilSystemStyleHTMLExport.php
@@ -65,10 +65,14 @@ class ilSystemStyleHTMLExport
 
     public function export(): void
     {
+        $location_stylesheet = ilUtil::getStyleSheetLocation('filesystem');
+
+        // Fix skin path
+        $this->style_dir = dirname($this->style_dir, 2) . DIRECTORY_SEPARATOR . dirname($location_stylesheet);
+
         $this->createDirectories();
 
         // export system style sheet
-        $location_stylesheet = ilUtil::getStyleSheetLocation('filesystem');
         $iterator = new RecursiveIteratorIterator(
             new RecursiveDirectoryIterator(dirname($location_stylesheet), FilesystemIterator::SKIP_DOTS),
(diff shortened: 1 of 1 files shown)


klees:
Hi @padvincenzo,

is there a mantis ticket that reports this? If so: Please link and we could bump the label to `bugfix`.

Kind regards!

Amstutz:
thx a lot! is this also valid for 9 and 10 (trunk)?

padvincenzo:
> Hi @padvincenzo,
> 
> is there a mantis ticket that reports this? If so: Please link and we could bump the label to `bugfix`.
> 
> Kind regards!

I think there are some, I also opened a mantis in this subject some time ago. All of our clients suffer from this bug, that we solved with a different customization. 
Once I noticed these lines of code I opened the PR.

padvincenzo:
> thx a lot! is this also valid for 9 and 10 (trunk)?

I'm not sure since none of our clients upgraded to ilias 9 yet, but I think so

padvincenzo:
This PR solves only half of the problem about the skin in LM exports: exporting the skin the custom templates are exported too (and they are not needed in an export).

@Amstutz if you want any changes in this PR let me know, I'll be happy to help

Amstutz:
@padvincenzo , if there is an issue in mantis, please link it, as @klees requested.

@alex40724 : Can you have a look a this, I really do not have much insight in LM exports and their workings? Might we move this code closer to LM modules?

alex40724:
@Amstutz There is nothing learning module specific in this code. This is system style related. Note that this is re-used by other components, like glossaries, blogs or wikis. To  me the change looks feasible. Thanks @padvincenzo.

alex40724 (approved):

Amstutz:
thx for the feedback.

PR #7837, opened 2024-07-22
ks: section dropdown

alex40724:
This PR changes the Section View Control to allow Dropdowns as middle items.
Bildschirmfoto 2024-07-22 um 13 45 21

Use cases are navigational elements like here:
https://docu.ilias.de/goto_docu_wiki_wpage_7518_1357.html

Docu Changes:
https://github.com/ILIAS-eLearning/ILIAS/pull/7837/files#diff-4cefcd552bb8e4cb0bf88461b0ea86defb16d82fe86477c3738bc0efcaea6697

Currently the Section View Control documentation includes "Split Buttons" which do not exist, so this part is removed. However the Month Buttons are already allows, so this is added.

This PR already includes an implementation. The correspdoning methods already allow to pass any component, the following check has been extended to check for the Dropdowns.


Amstutz:
Hi @alex40724 , thx for the proposal. We will look into this and will attach the JF label if our first review considers this ready for JF.

Amstutz:
Hi @alex40724
 
Thx for proposing this enhancement of the Section View Control. We discussed your change with our UI Expert @yvseiler and the conceptual changes authority @catenglaender.
 
For the Public Interface Part to be discussed with the JF. Please:
 
- [ ] Usage Rule: Please narrow down the purpose of the usage of Dropdowns here. "Dropdowns MUST NOT be used for any other purpose than skipping a section of the navigation (e.g. jumping from Chapter 1 to Chapter 5)".

For the implementation. Please change as follows or explain why not:
- [ ] Default behavior: Use a min (e.g. 100px) and max width (e.g. 400px) to prevent the change of the size of the button for all cases below 100px and have a fixed max width (as already the case). 
- [ ] Responsive behavior: Please change the behavior for everything below md to 95vw. We believe that making the control larger on a smaller screen will improve its usability.

Please change:
- [ ] Test Case: Add one to make sure, Dropdowns can be added now and in the future.
- [ ] Union-types: please change the type-hint of the middle button to a union type Button|Dropdown|Month from factory all the way down to the corresponding properties and getters. This way we can also get rid of the instanceof checks.

thx 
@Amstutz, @thibsy, @catenglaender, @yvseiler

klees:
Hi @alex40724,

I have assigned you here to get this of my list. Please assign back to @thibsy/@Amstutz/me once you have finished working on the feedback.

Thanks!

matthiaskunkel:
Hi @alex40724 
I highly appreciate this PR and would be delighted if this UI change could at least be implemented for ILIAS 10. It would be a significant gain in usability. The current element is clearly too short in many cases, so that it is not clear what is written on it.

JohnnySepp:
Hi @alex40724,
Since I have also encountered problemd regarding accessibility (https://mantis.ilias.de/view.php?id=45381 and https://mantis.ilias.de/view.php?id=45380), it would be nice if the PR could be merged with the suggested changes from @Amstutz to have a propper KS-Element that will solve the issues in Learning Module and Blog (https://mantis.ilias.de/view.php?id=40710) and the mediacast (https://mantis.ilias.de/view.php?id=40718)

alex40724:
Follow up https://github.com/ILIAS-eLearning/ILIAS/pull/10298

PR #7840, opened 2024-07-23
[FIX] make trash cron job more error-resistant

chfsx:
Deleting the recycle bin is very error-prone and leads to the fact that, for example, deleting a plugin object whose plugin has been uninstalled causes the entire recycle bin deletion job to be cancelled. see also the mantis ticket https://mantis.ilias.de/view.php?id=40265, where the same case occurs with a data collection object.

my suggestion is that the cleanup process handles such cases a little more graciously and at least tries to delete the most necessary data from object_data, but then continues. maybe it needs further adjustments in `removePartial`, you can probably judge that best.

### DIFF
--- a/Services/SystemCheck/classes/class.ilSystemCheckTrash.php
+++ b/Services/SystemCheck/classes/class.ilSystemCheckTrash.php
@@ -142,15 +142,44 @@ class ilSystemCheckTrash
             $sub_nodes = $this->readDeleted((int) ($deleted_info['tree'] ?? 0));
 
             foreach ($sub_nodes as $sub_num => $subnode_info) {
-                $ref_obj = ilObjectFactory::getInstanceByRefId((int) ($subnode_info['child'] ?? 0), false);
+                $tree_id = (int) ($subnode_info['tree'] ?? 0);
+                $node_id = (int) ($subnode_info['child'] ?? 0);
+                try {
+                    $ref_obj = ilObjectFactory::getInstanceByRefId($node_id, false);
+                } catch (Throwable $t) {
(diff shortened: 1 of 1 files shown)


smeyer-ilias:
Hi @fschmid
I don't think that this is the right way to solve problems in the deletion process:
- deleting the object_data, -reference and tree entries means that the problems are no longer recognized in the deletion routines and furthermore unreferenced data remains in the database and in the file system, which can only be cleaned up with great effort in the future.

A few months ago I deleted the recycle bin (approx. 80,000) entries on a test installation (copy of a very old ILIAS installation) and proceeded in this order:
1) Deleting "leaf nodes" (file, blog, ...)
2) Deleting folders
3) Deleting groups and courses
4) Deleting categories

At least the following errors occurred:
- https://mantis.ilias.de/view.php?id=40764
- https://mantis.ilias.de/view.php?id=40763
- https://mantis.ilias.de/view.php?id=40762

After these were resolved, the deletion on the production installation ran without errors.

For repo plugins, I agree that something similar to your suggestion should be implemented. Ideally, a plugin has its delete methods programmed out, but if this is not the case, the repo plugin slot must "step in" and delete the relevant data. But this would be an extension of the repo plugin slot.

chfsx:
thanks for the feedback, of course i see the point that the components always have to ensure a clean deletion. our experience shows that it rarely all works, even after the last fixes you mentioned. i myself hardly know of any installation where deleting the recycle bin simply runs through.

the plugin issue remains in any case. i understand you to mean that you see this as a feature request? or does it affect another component according to your assessment, then we'll try it that way.

PR #7841, opened 2024-07-23
User Feature: Session Reminder and Configuration as a User Default Settings 

fhelfer:
This PR adds a new User Default Setting to enable/disable the option to configure a Session Reminder time in User Settings.

Feature Wiki: https://docu.ilias.de/goto_docu_wiki_wpage_8270_1357.html


components/ILIAS/Authentication/classes/class.ilSessionReminder.php
```diff
@@ -53,16 +52,54 @@ public static function byLoggedInUser(): self
...
+    public static function isLocallyActivated(int $user_id): bool
+    {
+        return self::getLocalSessionLeadTime($user_id) > 0;
     }
 
+    public static function getMaxLeadTime(): int
+    {
+        $expires = ilSession::getSessionExpireValue();
+        return max(self::MIN_LEAD_TIME, ($expires / 60) - 1);
+    }
```
mjansenDatabay:
```suggestion
    }

```

components/ILIAS/Authentication/classes/class.ilSessionReminder.php
```diff
@@ -73,10 +110,15 @@ public function __construct(ilObjUser $user, ClockInterface $clock)
 
     private function init(): void
     {
+        /** @var ilSetting $ilSetting */
+        global $DIC;
```
mjansenDatabay: Can you please use "Dependency Injection" in `\ilSessionReminder::byLoggedInUser` instead?

components/ILIAS/Authentication/classes/class.ilSessionReminder.php
```diff
@@ -53,16 +52,54 @@ public static function byLoggedInUser(): self
         return $reminder;
     }
 
-    public static function isGloballyActivated(): bool
+    public static function getGlobalSessionReminderLeadTime(): int
     {
-        /** @var ilSetting $ilSetting */
         global $DIC;
```
mjansenDatabay: If we touch the code, maybe we could move the dependency to `\ilSessionReminder::byLoggedInUser` and inject it to `__construct` instead.

We should then try to make all `static` methods non-`static` (exception: `\ilSessionReminder::byLoggedInUser`).

components/ILIAS/Authentication/classes/class.ilSessionReminder.php
```diff
@@ -104,13 +146,12 @@ private function isEnoughTimeLeftForReminder(): bool
...
-        return (
-            self::isGloballyActivated() &&
+        return
             !$this->getUser()->isAnonymous() &&
             $this->getUser()->getId() > 0 &&
-            (int) $this->getUser()->getPref('session_reminder_enabled') &&
+            self::isLocallyActivated($this->getUser()->getId()) &&
             $this->isEnoughTimeLeftForReminder()
-        );
+        ;
```
mjansenDatabay: For my personal taste the `;` in a single line looks a bit ugly :-). I prefer the `(...);`-notation, even if this introduces two further characters :-).

mjansenDatabay (changes requested):
Thx @fhelfer for your contribution.

I added some minor remarks for the changes in the the `Authentication` component.

We should have a chat about the naming of the `*LeadTime`-methods, which I am not so happy with.

1. `getGlobalSessionReminderLeadTime`
2. `getAcceptableLeadTime`
3. `getLocalSessionLeadTime`
4. `getMaxLeadTime`

The first and last names are more or less self-explanatory, but: What is the concept of "acceptable"? "acceptable" for whom? Do you mean something like "effective", and what do you mean by "local"? Do you mean something like `individual`?

Best regards,
Michael

components/ILIAS/Authentication/classes/class.ilSessionReminder.php
```diff
@@ -104,13 +143,11 @@ private function isEnoughTimeLeftForReminder(): bool
...
     {
-        return (
-            self::isGloballyActivated() &&
+        return
             !$this->getUser()->isAnonymous() &&
             $this->getUser()->getId() > 0 &&
-            (int) $this->getUser()->getPref('session_reminder_enabled') &&
-            $this->isEnoughTimeLeftForReminder()
-        );
+            $this->isIndividuallySet() &&
```
mjansenDatabay: Why not `$this->getEffectiveLeadTime() > 0`? We could then remove `isIndividuallySet`.

mjansenDatabay (approved):

components/ILIAS/User/classes/Setup/Migration/class.ilUserDB10UpdateSteps.php
```diff
@@ -0,0 +1,46 @@
...
+
+use ILIAS\Setup;
+use ILIAS\Refinery;
+use ILIAS\Setup\Environment;
+use ILIAS\Setup\Objective;
+use ILIAS\UI;
+
+/**
+ * @author Fabian Schmid <fabian@sr.solutions>
+ */
```
mjansenDatabay:
```suggestion
```

components/ILIAS/User/classes/Setup/Migration/class.ilUserDB10UpdateSteps.php
```diff
@@ -0,0 +1,46 @@
...
+
+    protected ilDBInterface $db;
+
+    public function prepare(ilDBInterface $db): void
+    {
+        $this->db = $db;
+    }
+
+
+
```
mjansenDatabay:
```suggestion

```

components/ILIAS/User/classes/Setup/Migration/class.ilUserDB10UpdateSteps.php
```diff
@@ -0,0 +1,46 @@
...
+    public function step_1(): void
+    {
+        $query = 'DELETE FROM settings WHERE module="common" AND keyword="session_reminder";';
```
mjansenDatabay: Please use proper `ilDBInterrface`-quoting (with `ilDBInterrface::quote` and `ilDBConstants::T_TEXT`) and remove the `;`.

components/ILIAS/User/templates/default/tpl.std_fields_settings_row.html
```diff
@@ -61,7 +61,10 @@
 		    <option value="{OPTION_VALUE}" {OPTION_SELECTED}>{OPTION_TEXT}</option>
 			<!-- END def_sel_option -->
         </select>
-		<!-- END def_selection -->		
+		<!-- END def_selection -->
+		<!-- BEGIN def_input -->
+		<input type="number" class="form-control" name="input[{PROFILE_OPTION_DEFAULT_VALUE}]" value="{CURRENT_OPTION_VISIBLE}" max="{CURRENT_OPTION_MAXIMUM}" />
```
mjansenDatabay: Should we provide a `min` attribute as well?

components/ILIAS/User/classes/Setup/class.ilUserSetupAgent.php
```diff
@@ -61,6 +61,9 @@ public function getUpdateObjective(Setup\Config $config = null): Setup\Objective
             ),
             new ilDatabaseUpdateStepsExecutedObjective(
                 new ilUserDB90()
+            ),
+            new ilDatabaseUpdateStepsExecutedObjective(
+                new ilUserDB10UpdateSteps()
```
mjansenDatabay: Shouldn't we use similar names here? `ilUserDB90` <-> `ilUserDB10UpdateSteps`. But of course, Stephan has to decide upon this.

components/ILIAS/User/classes/Setup/Migration/class.ilUserDB10UpdateSteps.php
```diff
@@ -0,0 +1,46 @@
...
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
```
mjansenDatabay:
```suggestion

declare(strict_types=1);

```

components/ILIAS/User/classes/Setup/Migration/class.ilUserDB10UpdateSteps.php
```diff
@@ -0,0 +1,46 @@
...
+        $this->db = $db;
+    }
+
+
+
+    public function step_1(): void
+    {
+        $query = 'DELETE FROM settings WHERE module="common" AND keyword="session_reminder";';
+        $this->db->manipulate($query);
+    }
```
mjansenDatabay: As far as I can see, we had the following setting in ILIAS 9.x:

- `session_reminder_enabled`

Maybe I am wrong, but IMO we need a migration to enable the session reminder for ILIAS 10.x if it was already enabled for ILIAS 9.x. Afterwards, the setting `session_reminder_enabled` could be deleted from the `settings` database table.

So if I understand the changes for ILIAS 10.x correctly, we should set the global default to the "Suggested Lead Time".

components/ILIAS/User/classes/Setup/Migration/class.ilUserDB10UpdateSteps.php
```diff
@@ -0,0 +1,46 @@
...
+    protected ilDBInterface $db;
+
+    public function prepare(ilDBInterface $db): void
+    {
+        $this->db = $db;
+    }
+
+
+
+    public function step_1(): void
```
mjansenDatabay: I could not find any database records for the query ...

```sql
SELECT * FROM settings WHERE module = 'common' AND keyword = 'session_reminder';
```

... even with an enabled "Session Reminder". So the question is, what are you trying to delete here :-)?

mjansenDatabay (changes requested):

components/ILIAS/User/classes/Setup/Migration/class.ilUserDB10UpdateSteps.php
```diff
@@ -0,0 +1,57 @@
...
+
+    public function prepare(ilDBInterface $db): void
+    {
+        $this->db = $db;
+    }
+
+    public function step_1(): void
+    {
+        $query = 'SELECT * FROM settings WHERE module = %s AND keyword = %s';
+        $result = $this->db->queryF($query, ['text', 'text'], ['common', 'session_reminder_enabled']);
```
mjansenDatabay:
```suggestion
        $result = $this->db->queryF($query, [ilDBConstants::T_TEXT, ilDBConstants::T_TEXT], ['common', 'session_reminder_enabled']);
```

mjansenDatabay (approved):

kergomard (changes requested):
Hi @fhelfer 

Thank you very much for this PR! I will wait with merging until this feature pr is merged #8116, this will also mean, that you will need to rebase to consolidate the user update steps. So you might want to wait until the merge, before working on this again. When you consolidate this, please make sure to not add `use` statements for classes in the global namespace.

I would like the following changes, but let me know, if you see something wrong with them:
- I don't like [this getter here](https://github.com/ILIAS-eLearning/ILIAS/pull/7841/files#diff-deba6d6232c13de03871455819a34d2d015baedaa2465608ecab708385066bddR140). "getSelect" has an abysmal semantic value already, but "getInput" is even worse. I would kindly ask you to change this to something more specific and to only retrieve the 'default_session_reminder', this way at least I get, what we are doing here.
- On the other hand [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/7841/files#diff-b6d0a6d4a6ae931e3eb0d5996e0de46a5e4207f214a54300d544a1c578533f3aR190) seems oddly over specific: We have "selection", and then "session_reminder_lead_time". Shouldn't this rather be "numeric"?

One very small additional change:
- Could you please remove the camelCase [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7841/files#diff-e3e44e320dc8a96875a37aa2387d45785c29d253ff2d05a2c484723b6ea18d45R116) as you are touching this variable anyway.

Other then this I think things are fine. I will make a little smoke test after the changes and the rebase.

Thanks again and best,
@kergomard

kergomard:
Thank you very much for the fixes @fhelfer !

I now did some smoke tests and sadly there is still something wrong (sorry for missing it, when reading through the code):
-Please fix the update step, as you are inserting 'session_reminder_lead_time' twice, if 'session_reminder_enabled' has a value, this throws a duplicate primary key error. Please also wrap the whole step into an if-statement to only be executed if there is not already a 'session_reminder_lead_time' value, this way it is made easier to rerun the step on failure.

I assume the changes in `Auth` to be covered by the thumbs-up from @mjansenDatabay and will thus merge as soon as you have fixed this,
@kergomard

mjansenDatabay:
> I assume the changes in `Auth` to be covered by the thumbs-up from @mjansenDatabay

This is right

fhelfer:
Hey @kergomard,

would you have another look at it. I also added dbupdatestep to remove session_reminder_enabled from usr_pref.
If the changes are fine please go ahead and merge

Thx and Best @fhelfer

kergomard:
This looks fine now.
For the record: Failing tests are not because of this PR.

kergomard (approved):

oliversamoila:
Thank you very much, @kergomard, @fhelfer and @mjansenDatabay 😊

PR #7844, opened 2024-07-23
T&A: Remove unnecessary getCorrectedLangVarForKey function

aaronbidzan:
When writing unit tests, I encountered the getCorrectedLangVarForKey function, which when observed had no usages and returned the provided key parameter back with no modifications.

Due to this, I have made this pull request which removes the function.

Kind Regards.


kergomard:
Thank you very much @bidzanaaron ! You are right, I never finished this. The idea here was to already provide a facility in case a lang_var changes (as we use them as keys). I now finalized the facility with [this commit](https://github.com/ILIAS-eLearning/ILIAS/commit/0345c5e36165993846286862c953ffd0dfc7a457) and thus close this PR.

Thanks again and best,
@kergomard

PR #7846, opened 2024-07-23
Prevent calling of undefined constants

iszmais:
https://mantis.ilias.de/view.php?id=41775

This replaces the call of the `ANONYMOUS_USER_ID` by a static.
Since this is already in a context where we don't have a user dependency, it is quite risky to assume the constant to be set IMO.

### DIFF
--- a/Modules/File/classes/class.ilObjFileStakeholder.php
+++ b/Modules/File/classes/class.ilObjFileStakeholder.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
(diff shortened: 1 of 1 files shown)


chfsx:
Thank you very much! of course, this only works under the assumption that the ANONYMOUS ID is 13, which has always been the case, and at least I've never come across any other ID... but I think we'll have to find an even better solution. but I've adopted this for the moment and also merged it to trunk.

PR #7847, opened 2024-07-24
UI: disable displaying divider label in sub-menu lists if placed in m…

lukastocker:
…ain-menu. (#41666)

https://mantis.ilias.de/view.php?id=41666


Amstutz:
thx.

PR #7851, opened 2024-07-24
T&A: Fix return of wrong instance in method buildTestQuestionAdministrationInteraction of Ilias\Test\Logging\Factory

malaja11:


kergomard:
Thank you very much @malaja11 ! This is indeed an error, ...but as the function is actually only there for the sake of completeness and is never used I didn't notice.

PR #7853, opened 2024-07-24
T&A: fix call of TestError constructor in method createTestErrorFromContext of class Ilias\Test\Logging\TestLogger

malaja11:

### DIFF
--- a/components/ILIAS/Test/src/Logging/TestLogger.php
+++ b/components/ILIAS/Test/src/Logging/TestLogger.php
@@ -39,7 +39,6 @@ class TestLogger implements LoggerInterface
         private readonly Factory $interaction_factory,
         private readonly AdditionalInformationGenerator $additional_information,
         private readonly \ilComponentLogger $component_logger,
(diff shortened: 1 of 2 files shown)


kergomard:
Again: Thank you very much @malaja11 !  Could you please do me a favor and also change the creation of the TestLogger in the [TestDIC](https://github.com/ILIAS-eLearning/ILIAS/blob/378554afd787151da216a3d3a0288c0cafa387f0/components/ILIAS/Test/src/TestDIC.php#L125) to the new signature.

Thanks again and best!
@kergomard

malaja11:
No problem :)
I just changed the call here https://github.com/ILIAS-eLearning/ILIAS/blob/378554afd787151da216a3d3a0288c0cafa387f0/components/ILIAS/Test/src/TestDIC.php#L125 too, I hope this is fine now.

kergomard:
Yes it is! Now things should not go boom anymore ;-).

Thanks again and best,
@kergomard

PR #7855, opened 2024-07-24
Add stacking to Bar Charts

tfamula:
Hello!
This PR adds [stacking](https://www.chartjs.org/docs/latest/samples/bar/stacked.html) to Bar Charts. This can also be done with [multiple stacks](https://www.chartjs.org/docs/latest/samples/bar/stacked-groups.html). We need the functionality for [this FR](https://docu.ilias.de/go/wiki/wpage_7662_1357).

chart.js is currently broken in trunk, so the examples will not be shown in the KS docu. There is already a PR to fix this: https://github.com/ILIAS-eLearning/ILIAS/pull/7854

@schmitz-ilias and @tfamula 

horizontal
vertical


klees:
Hi @tfamula and @schmitz-ilias,

thanks for discussing this with me today. We have looked into the interface that we want to present to our users. The general problem is that we do not know values of the data that should be displayed at implementation time. We only know potential shape of the datasets. Take the poll for example: We do not know how many options administrators might give users and we also do not know how users in a given installation have responded to a certain poll. Hence we need to try to model the data and directions to display it in a way that is robust to different shapes and value of data that is actually presented. Also, we should model the data in a way that makes it explorable for users, as we have begun to do for the bar charts where people can activate and deactivate certain parts of the charts. Finally, we need try to get as much insights into the structure of the data that should be displayed to be able to prevent problems in the implementation, make good choices in the UI framework and give interesting options to users.

We came to the following conclusions:

The fact that certain parts of a point in a dataset can be used to derive additional information is something that should become part of the definition of a Dataset. [`Chart\Dataset`](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/Data/Chart/Dataset.php#L38) and [`Dimension`](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/Data/Dimension/Factory.php) should be extended by the notion of "derived dimensions" that can express these information. For the stacking we would need a `sum`-dimension (when we discussed this, this was named `added`, but I guess `sum` would make more sense...) and the code could look something like this:

```php
$c_dimension = $df->dimension()->cardinal();

$dataset = $df->dataset([
    "Dataset 1" => $c_dimension,
    "Dataset 2" => $c_dimension,
    "Dataset 3" => $c_dimension,
],[ 
    "Derived Dataset" => $df->dimension()->sum("Dataset 1", "Dataset 2");
]);
```

Another example for a derived dimensions could be `average`.

This would then allow to do some plausibility checks in the Dataset (e.g.: it probably wouldn't make sense to add a range to a cardinal) but also only add points by giving values of the original dimension.

To display a sum dimension, consumers could then supply a `StackedBarConfig` like this:

```php
    $b1 = new BarConfig();
    $b1 = $b1->withColor($df->color("#d38000"));
    $b2 = new BarConfig();
    $b2 = $b2->withColor($df->color("#307C88"));
    $b3 = new BarConfig();
    $b3 = $b3->withColor($df->color("#557b2e"));
    $s1 = new StackedBarConfig($b1, b2);

    $bars = [
        "Derived Dataset" => $s1,
        "Dataset 3" => $b3
    ];
 
    //Generating and rendering the vertical chart
    $bar = $f->chart()->bar()->vertical(
        "A vertical bar chart",
        $dataset,
        $bars
    );
```

This would show the two dimensions together in one stacked bar along with the data from the third dimension in a separate bar. Users then could then also do this

```php
    $b1 = new BarConfig();
    $b1 = $b1->withColor($df->color("#d38000"));
    $b2 = new BarConfig();
    $b2 = $b2->withColor($df->color("#307C88"));

    $bars = [
        "Derived Dataset" => $b1,
        "Dataset 3" => $b2
    ];
 
    //Generating and rendering the vertical chart
    $bar = $f->chart()->bar()->vertical(
        "A vertical bar chart",
        $dataset,
        $bars
    );
```

to show the derived dataset with just a single ordinary bar.

I hope this is similar to what you remember and you can move on along these lines. Feel free to ping me if there are any doubts or we need further discussion or exploration for other reason.

Thanks for your time and patience!

klees:
Hi @tfamula,

please assign back and ping me if this is ready for the next iteration.

Thanks!

tfamula:
HI @klees,
we have completed a solution now, which is close to the draft of our joint meeting, but nevertheless a bit different since we have given it some more thought.
The Dataset will accept a new parameter which is filled with `DimensionGroup`s. With this we can semantically group several Dimensions. At that point, we do not have to know what "kind of" group it is. This becomes crucial in the next step when it comes to the configuration of the (grouped) Datasets. We decided to introduce a new `GroupConfig` which contains the characteristics of a Grouped Dataset. These GroupConfigs will be passed as a new parameter to the Charts, since they should be treated independently of the BarConfigs for single Datasets. In case of Stacked Bars, we still need to define an own BarConfig for each Bar, which e.g. contains information about the color of a single Bar. At the moment, `GroupConfig::withStacked()` is available for grouped Datasets. As you can also see, we decided to rename things a bit. We have neither used "added", nor "sum", but "stacked", because that's exactly what we do here. We just stack multiple Datasets, but they still remain own independent Bars for the end user. That's why there is also no "derived" Dataset in our eyes and we decided for the more general name "group".

In the future, the DimensionGroups could be extended by methods like `DimensionGroup::withDerivedSum()` and `DimensionGroup::withDerivedAverage()`. Then, this "newly created" Dataset, which is derived from given Datasets, can be handled internally (probably in the Renderer) for a suitable presentation in the context of Charts.


@schmitz-ilias and @tfamula 

PS: I will fix/extend the unit tests once the solution is approved by you.

klees:
Hi @tfamula and @schmitz-ilias,

sounds plausible and looks good too. Please present the interface on the JF.

Thanks!

matthiaskunkel:
Jour Fixe, 30 SEP 2024: We highly appreciate this suggestion and accept the PR for ILIAS 10/trunk.

tfamula:
@klees I would take care of the unit tests now.

schmitz-ilias:
Hi @klees, unit tests are done

yvseiler:
Dear @schmitz-ilias 

We have noticed that we have not yet paid any attention to the fact that the colour values currently used in the charts are difficult to distinguish from an accessibility perspective. This applies to all charts in the KS.

With the stacked bar introduced here, however, the problem becomes even more visible: the stacked bars are difficult to tell apart (blue and green values are very similar). For the time being, we recommend inserting a small space between the bars in this stacked bar variant to make it easier to distinguish the stacked bars.

f.e. graphic on the right side (source: https://flourish.studio/blog/accessible-chart-design/)
Bildschirmfoto 2024-10-10 um 13 57 10

Thank you for keeping an eye on this recommendation with the small space at stacked bar. If you have any questions, please get in contact with us.

Best regards
Kristina, Kendra, Yvonne

PS. In future, the charts should be examined to see how they could be made more distinguishable from each other so that ‘users don't have to rely on just colour to understand which category is which.’. 
_Update 14.10._: We should tackle this later with a suitable concept.

klees:
Hi Kristina, Kendra and Yvonne,

graphs and colors in the context of the UI framework are a hard problem imo.

Let me explain a little by starting with the current state of the art: The colors of the bars are not controlled by the CSS, like other colors. Instead they are defined by the consumers of the UI components, i.e. at the usage locations of the charts. So @schmitz-ilias effectlively can only improve the examples and maybe add an according accessibility rule if we continue to use the current implementation. Both is something which should be achievable with a decent amount of effort and would already improve the situation a little, so we should probably do this.

From my understanding, there are two reasons for the current implementation where colors are not picked via CSS:

1. The library we use renders the bars as an HTML canvas and we cannot reach into this canvas with CSS. For the canvas, we basically pick crayons with a defined color and draw stuff on it, so no css attributes, boxes, etc. We could, in principle, device some mechanism to connect the CSS with the canvas, e.g. by some build step, some javascript or any other creatives means. But it is not something we have out of the box and also a task where various non-obvious design decisions need to be made. We could also, again in principle, switch the library we use or even implement some bar rendering logic ourselves. But these two approaches are not super straight forward or even simple as well.
2. For some use-cases it even would be *correct* to let the consumer of the UI component pick a color. In the UI framework we want to build generic UI components for ILIAS and its plugins. So imagine a use case where I want to visualize some learning progresses with the bar chart. I would want to gurantee, that the "success" bar uses some green from the learning progress icon and that "failure" uses some red from the other icon. So I would need to assign named colors to each bar ("lp-success-green", "lp-failure-red"), even if these don't need to be resolved to actual color codes at that moment. One step further: I want to visualize some results of various parties in a political education plugin I'm creating. I would want to make sure that the bar for the SPD uses the exact SPD red, the bar for the greens has the correct green and the NoAfD gets a feces brown. I would probably need to assign actual color codes to the bars there, because my skin most certainly won't contain a named color for the correct brown.

Even if we overcome 1. and ignore 2., we would need to come up with a scheme to automatically pick colors that would roughly need to implement the following requirements:

* have a set with an unknown number of different colors
* ...where individual colors are visually as far appart as possible
* ...and they are preferable w.r.t. accessibility considerations
* ...and they have something to do (or explicitly not to do?) with this or that color in the skin

This set of requirements to me feels *uff*. And then again: Colors may not be enough, even "accessible" ones. They are not suppose to convey meaning, although there might be exceptions that I'm currently not aware of. So we would need to think again about these problems, but this time for something like various shading or patterns that we would want to put on the bars to make them work without colors.

So basically, we currently offload these problems to our consumers. On the one hand this is just a little mean. On the other hand the problem is at least a little easier to solve at consumer side, because some contingencies that we need to consider in the UI framework may be decided at concrete usage locations, such as: how many colors do I need? And also: I simply cannot see an easy way to solve these problems, for whatever meaning of solve.

I hope this makes a little sense. I'm absolutely in when it comes to iterate in this space, look into solutions or try to be creative. But currently I cannot think of any better point in the design space, especially regarding the current amount of usages of the component. I would be really interested to hear from anyone who can come up with ideas here.

Kind regards!

schmitz-ilias:
Hi @yvseiler, @klees,

I added an accessibility rule to bar charts:
```
When displaying a multi-dimensional dataset in a Bar Chart, the colors of adjacent bars MUST have high contrast between each other.
```
I also changed the color palette of bar chart examples with multi-dimensional datasets to one conforming to the new rule (taken from [analysisfunction.civilservice.gov.uk/policy-store/data-visualisation-colours-in-charts](https://analysisfunction.civilservice.gov.uk/policy-store/data-visualisation-colours-in-charts/#section-5)):
new_palette_bar_chart

It would of course be better to be a little bit more specific in the rule, but I think together with the updated example it should suffice as a guideline for developers for now. @yvseiler, if you have any suggestions for the rule, or feedback concerning the color palette, I'll gladly include them here (but please be aware of the ever-looming coding complete deadline, I'd like to include stacked bar charts in the Poll in ILIAS 10).

@klees, I'm not sure why the checks are suddenly failing. It looks like building an artifact in `ilObjectTypeSpecificPropertiesArtifactObjective` fails, but this branch makes no changes anywhere near that. For me locally `composer du` runs through fine, and the checks for the individual commits are also not complaining. I'm guessing this is an unrelated problem from trunk?

Let me know if there is anything else that needs to be done here.

Cheers, Tim

matthiaskunkel:
Thanks, @schmitz-ilias, for these improvements. 
Just one remark: we already have a colour scheme in ILIAS. And it would be great if the bar charts would follow this scheme. These colours are used in the content style, for example. Based on the linked website and the offered tools I suggest to slighly modify the colours and give them some more contrast. But they still look familiar then.
We could reuse our '!brandprimary' which is #4C6586. '!brandsecondary' should get less lightness to keep a 1:4,5 ratio: #598131. Also '!brandwarning' needs less lightness: #C25705. And as a fourth colour we could use '!extendedcolor6' which is #82639E. And more colours are still available among the extended colours.
And a wish concerning the order of colours: first bar is blue, second orange and third green. It's much better to distinguish between blue and orange when only having two bars and to have orange between blue and green when three bars are shown in a row.

yvseiler:
Dear @schmitz-ilias , dear @klees 

Thank you very much for your comments, helpful background information and the quick response and customisations from you @schmitz-ilias. I find your adjustments a good first approach for the implementation with ILIAS 10 for the time available and the little we have conceptually at hand here at the moment. Many thanks for that.

We UI/UX/Ally experts are well aware that we here as a community have to think conceptually about which paths we can pursue - the three of us are not yet in a state of presenting THE ONE solution here. Your list @klees is a helpful catalogue of measures, thanks a lot for your thoughts on that.

**Question concerning color hex-codes**
Regarding @matthiaskunkel 's comment: I'm not sure, but I understood @klees' comment to mean that manually defined colour values are not used, but are derived from where, right? Or can we ‘only’ not transmit them via SCSS? I'm a little bit confused now, sorry. I thought, that with exact colour specifications, the problem would be that we might need more than 5 colours that we don't even have in our branding colours. Or would it be helpful for you @schmitz-ilias to receive 5 first matching colour values from us as suggested by Matthias? 

For me it would also be fine if we outsource the discussion and decision of an accessible chart colour palette here from the StackedBar feature and tackle it for ILIAS 11. This would at least give us a reasonably accessible representation for 10 with the adjustments Tim made already. Colours matching the branding (if this is the way with a manual colour palette) could then be implemented in the next release as shortly before Coding Complete. Would that also be a viable option for you @matthiaskunkel ?

mkunkel:
@yvseiler you mean to mention @matthiaskunkel 

I have nothing to do with this project.

yvseiler:
> @yvseiler you mean to mention @matthiaskunkel
> 
> I have nothing to do with this project.

I'm so sorry, thanks for your hint!

klees:
Hi @matthiaskunkel and @yvseiler,

the users of the bar chart have to define colors for the bars explicitly. Have a look [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7855/files#diff-77c97c189635fb09eca84dbe7c1edc63034b045fb324e9377afd3a13e0e2d076L147). Neither can users reference colors via variables from CSS or SCSS (like @matthiaskunkel mentions) nor can the UI framework assign colors from CSS, SCSS or any other source.

The colors for the bars (at the current state of the art...) will be defined separatly for every single usage location and they will not react to changes in the skin. @schmitz-ilias|s change _only_ targets the examples, nothing more.

Kind regards!

klees:
Hi @schmitz-ilias,

you can have a look into the [lists of actions for check](https://github.com/ILIAS-eLearning/ILIAS/actions/workflows/checks.yml), scroll down until you see the first failing action for `trunk` and then find this commit here that breaks the build:

https://github.com/ILIAS-eLearning/ILIAS/commit/454e212eb1a80b6bd3dd57964747df2db50d6e5d

@alex40724, would you be so kind and look into this?

Kind regards!

schmitz-ilias:
@yvseiler, @matthiaskunkel, a "recommended" accessible color palette would definitely be helpful for me, not necessarily for this PR as @klees explained above, but definitely for the implementation of [stacked bar charts in Poll](https://docu.ilias.de/go/wiki/wpage_7662_1357). Maybe we can use the FW article for further discussion of this?

As a side note, the palette you recommended above @matthiaskunkel does not seem to have high enough contrast:
https://webaim.org/resources/contrastchecker/?fcolor=4C6586&bcolor=C25705
https://webaim.org/resources/contrastchecker/?fcolor=C25705&bcolor=598131
https://webaim.org/resources/contrastchecker/?fcolor=598131&bcolor=4C6586
Or am I looking at the wrong metric?

schmitz-ilias:
@klees, trunk is fixed, checks are passing, have at it.

yvseiler:
> but definitely for the implementation of [stacked bar charts in Poll](https://docu.ilias.de/go/wiki/wpage_7662_1357). Maybe we can use the FW article for further discussion of this?

That's a good idea @schmitz-ilias.

klees:
Thanks y'all!

PR #7857, opened 2024-07-25
Block: Preserve block title on slice

iszmais:
https://mantis.ilias.de/view.php?id=41806

This is just a minor fix to preserve numeric titles in blocks.
Since PHP casts every numeric string key into an int and `array_slice` resets all integer keys, this causes that every numeric title ist displayed incremented from 0 instead of its number.

### DIFF
--- a/Services/Block/classes/class.ilBlockGUI.php
+++ b/Services/Block/classes/class.ilBlockGUI.php
@@ -550,7 +550,7 @@ abstract class ilBlockGUI
     {
         $data = $this->getData();
         $this->max_count = count($data);
-        $data = array_slice($data, $this->getOffset(), $this->getLimit());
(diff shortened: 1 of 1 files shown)


iszmais:
@alex40724 @smeyer-ilias @lscharmer 

Since this directly affects the "My courses and groups" and "Dashboard" view (by resolving the bug there), I would appreciate if one of you could take care of this here.

If needed, I can provide the distribution. Therefore, just the approval would be enough.

Greetings,
@iszmais

lscharmer (approved):

mjansenDatabay (approved):

iszmais:
Thank you for the review!
I resolved the conflicts

Greetings,
@iszmais

PR #7864, opened 2024-07-26
T&A: fix return in method buildScoringInteractionFromId of class ILIAS\Test\Logging\TestLoggingDatabaseRepository

malaja11:


kergomard:
And thank you once more @malaja11 !

PR #7866, opened 2024-07-26
Export: do not export tail dependencies if there are not entities

schmitz-ilias:
This PR makes it so that tail dependencies are only exported if there are actual entities to export. In this past, this has lead to imports failing the validation via xsd. In current ILIAS 9 this does not seem to be a problem anymore, but the changes proposed here might make export/import a bit more robust in the future.

### DIFF
--- a/Services/Export/classes/class.ilExport.php
+++ b/Services/Export/classes/class.ilExport.php
@@ -602,6 +602,10 @@ class ilExport
         $this->log->debug("process tail dependencies of " . $a_entity);
         $sequence = $exp->getXmlExportTailDependencies($a_entity, $a_target_release, $a_id);
         foreach ($sequence as $s) {
+            if (empty((array) $s["ids"])) {
+                continue;
+            }
+
             $comp = explode("/", $s["component"]);
             $exp_class = "il" . $comp[1] . "Exporter";
             $s = $this->processExporter(
--- a/Services/Export/classes/class.ilExport.php
+++ b/Services/Export/classes/class.ilExport.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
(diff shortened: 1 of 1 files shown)


Services/Export/classes/class.ilExport.php
```diff
@@ -602,6 +603,10 @@ public function processExporter(
         $this->log->debug("process tail dependencies of " . $a_entity);
         $sequence = $exp->getXmlExportTailDependencies($a_entity, $a_target_release, $a_id);
         foreach ($sequence as $s) {
+            if (empty((array) $s["ids"])) {
```
smeyer-ilias: Please make this more robust against "undefined array key" failures.

smeyer-ilias (changes requested):
Thanks @tschmitz for fixing this. Please have a look on my comment.

Services/Export/classes/class.ilExport.php
tschmitz: @smeyer-ilias Wish I could take credit but I think you're looking for someone else 😇

smeyer-ilias: Sorry and thanks for your reply

schmitz-ilias: Done, thanks @smeyer-ilias (and also to our intercontinental consultant)

schmitz-ilias:
Just as a side note, I realized why this is not a problem in 9 currently: the xsd for Metadata does exist, but is not yet taken into account to actually validate imports (it's not in the `SchemaValidation` folder). So my guess is that this fix will not be enough, and we should think about being more lenient about empty export files. Otherwise Exports from 8 will become mostly unusable in 10.

smeyer-ilias (approved):
Thanks @schmitz-ilias 

and best regards to Boston!

PR #7868, opened 2024-07-29
Test: Fix Import of Test from IMSm ...

fneumann:
See https://mantis.ilias.de/view.php?id=41819


kergomard:
Thank you for the PR  @fneumann .

This would be fixing symptoms and not causes. Fixed with https://github.com/ILIAS-eLearning/ILIAS/commit/4c33958b84c4206aa09f540027e46393495cdea3 and picked to trunk.

Best,
@kergomard

PR #7870, opened 2024-07-30
T&A Mantis #41829: ilSession filenames not set in Test Import

fhelfer:
https://mantis.ilias.de/view.php?id=41829

nur trunk


kergomard:
Thank you very much @fhelfer ! You are most obviously right, that this can't work otherwise.

PR #7874, opened 2024-08-01
T&A: fix inconsistent use of ServerRequestInterface and RequestInterface

malaja11:
While writing unit tests I noticed the inconsistent use of the ServerRequestInterface and RequestInterface in the class ILIAS\Test\Logging\TestLogViewer and wanted to change the type of the member variable $request to ServerRequestInterface. This change lead to several failing unit tests. The reason for this is the inaccuracy/ inconsistency of the used types in the ILIAS\HTTP\Services and ILIAS\http\RawHTTPServices classes.

As the member variable $request in ILIAS\http\RawHTTPServices has the type ServerRequestInterface, I decided to change the associated return types to this interface too.

### DIFF
--- a/components/ILIAS/HTTP/src/RawHTTPServices.php
+++ b/components/ILIAS/HTTP/src/RawHTTPServices.php
@@ -87,7 +87,7 @@ class RawHTTPServices implements GlobalHttpState
     /**
      * @inheritDoc
      */
-    public function request(): \Psr\Http\Message\RequestInterface
+    public function request(): \Psr\Http\Message\ServerRequestInterface
     {
         if ($this->request === null) {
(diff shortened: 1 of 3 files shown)


kergomard:
Thanks for the PR @malaja11 !
This has caused me some trouble too when programming, but then I worked around it and forgot. Sorry.
The changes in Test & Assessment are right from where I stand and thus I assign this to you @chfsx . I think, you are the right one for the other changes, right?

Thanks again and best,
@kergomard 

FYI @dsstrassner , @thojou , and @mbecker-databay .

chfsx:
Thanks a lot for the PR!

PR #7877, opened 2024-08-02
Bibliographic: switch to LOM API

schmitz-ilias:
This PR replaces all usages of the old `MetaData` classes in `Bibliographic` with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md).

Due to the current state of the trunk I was not able to test the changes, but since the footprint here is pretty small, the risk of breaking things further should be pretty low.

Cheers, @schmitz-ilias


chfsx:
Thanks a lot for changing this!

PR #7880, opened 2024-08-02
DidacticTemplate: switch to LOM API

schmitz-ilias:
This PR replaces all usages of the old `MetaData` classes in `DidacticTemplate` with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md).

Cheers, @schmitz-ilias


smeyer-ilias (approved):
skål

PR #7881, opened 2024-08-02
LearningSequence: switch to LOM API

schmitz-ilias:
This PR replaces all usages of the old `MetaData` classes in `LearningSequence` with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md).

Due to the current state of the trunk I was not able to test the changes, but since the footprint here is pretty small, the risk of breaking things further should be pretty low.

I tried to stick to what was already there in terms of dependency management, let me know if there is anything you want done differently.

Cheers, @schmitz-ilias

### DIFF
--- a/components/ILIAS/LearningSequence/classes/Player/class.ilLSViewFactory.php
+++ b/components/ILIAS/LearningSequence/classes/Player/class.ilLSViewFactory.php
@@ -18,6 +18,8 @@ declare(strict_types=1);
  *
  *********************************************************************/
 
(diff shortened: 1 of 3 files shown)


klees (changes requested):
Hi @schmitz-ilias,

thanks a lot, looks nice and clean =)

Just one change request (same for the other PR): Please use array-access in the local DIC.

Thanks!

schmitz-ilias:
Hi @klees

thanks for the quick feedback, the local DIC now uses array access.

PR #7883, opened 2024-08-02
MainMenu: switch to LOM API

schmitz-ilias:
This PR replaces all usages of the old `MetaData` classes in `MainMenu` with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md).

I would have liked to stay closer to what was already there in terms of dependency management, but since your `DIC`-trait is in `OrgUnits` and not `MainMenu`, I grabbed the API directly from the global DIC. I hope that's alright, let me know if there is anything you want done differently.

Cheers, @schmitz-ilias


chfsx:
Thanks a lot! this DIC-trait usage will be removed, this is some legacy thing...

PR #7885, opened 2024-08-02
Repository: switch to LOM API

schmitz-ilias:
This PR replaces all usages of the old `MetaData` classes in `Repository` with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md).

Also included is the introduction of the API to `GlobalDICDomainServices`. This is not used in `Repository` itself, but is necessary to be able to use your dependency management when using the API in your other components (PRs will be provided in the future).

Cheers, @schmitz-ilias


alex40724:
Thanks!

PR #7889, opened 2024-08-05
Object: switch to LOM API

schmitz-ilias:
This PR replaces all usages of the old `MetaData` classes in `Object` with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md).\

The new API is a bit more explicit and a bit more strict in certain places, which turned out to be particularly noticeable in the creation process of objects. For simplicity, the changes proposed here stay as close as possible to how things currently work. With  further restructurings of Object in mind, it's probably better to make the switch to the new API earlier rather than later, so that you know what you'll be working with.

Cheers, @schmitz-ilias

### DIFF
--- a/components/ILIAS/ILIASObject/classes/class.ilObjectGUI.php
+++ b/components/ILIAS/ILIASObject/classes/class.ilObjectGUI.php
@@ -953,13 +953,23 @@ class ilObjectGUI implements ImplementsCreationCallback
 
         $class_name = 'ilObj' . $this->obj_definition->getClassName($this->requested_new_type);
 
+        /** @var ilObjectPropertyTitleAndDescription $title_and_description_prop */
+        $title_and_description_prop = $data['title_and_description'];
+
         $new_obj = new $class_name();
         $new_obj->setType($this->requested_new_type);
         $new_obj->processAutoRating();
+        /*
+         * Title and description needs to be set before calling ilObject::create
+         * at least preliminarily, because the creation of a new LOM set requires
+         * the object to have a non-null title.
+         */
+        $new_obj->setTitle($title_and_description_prop->getTitle());
+        $new_obj->setDescription($title_and_description_prop->getDescription());
         $new_obj->create();
 
         $new_obj->getObjectProperties()->storePropertyTitleAndDescription(
-            $data['title_and_description']
+            $title_and_description_prop
         );
         $new_obj->setTitle($new_obj->getObjectProperties()->getPropertyTitleAndDescription()->getTitle());
         $new_obj->setDescription($new_obj->getObjectProperties()->getPropertyTitleAndDescription()->getDescription());
--- a/components/ILIAS/ILIASObject/classes/class.ilObjectGUI.php
+++ b/components/ILIAS/ILIASObject/classes/class.ilObjectGUI.php
@@ -953,26 +953,16 @@ class ilObjectGUI implements ImplementsCreationCallback
 
         $class_name = 'ilObj' . $this->obj_definition->getClassName($this->requested_new_type);
 
-        /** @var ilObjectPropertyTitleAndDescription $title_and_description_prop */
-        $title_and_description_prop = $data['title_and_description'];
-
         $new_obj = new $class_name();
         $new_obj->setType($this->requested_new_type);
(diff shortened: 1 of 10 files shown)


components/ILIAS/ILIASObject/classes/class.ilObjectGUI.php
```diff
@@ -953,13 +953,23 @@ public function saveObject(): void
...
+
         $new_obj = new $class_name();
         $new_obj->setType($this->requested_new_type);
         $new_obj->processAutoRating();
+        /*
+         * Title and description needs to be set before calling ilObject::create
+         * at least preliminarily, because the creation of a new LOM set requires
+         * the object to have a non-null title.
+         */
+        $new_obj->setTitle($title_and_description_prop->getTitle());
```
kergomard: Hi @schmitz-ilias 

Thank you for the PR. This is not a review yet, even if the changes look ok on a first glance. Just one reaction, as I think this is something bigger: I don't think setting the title and description here works. This is done directly in the ObjectProperties. Currently there are no `ObjectProperties`, before an object is created in the database, so this should go sideways. I did not test this, but the code and my experience suggests it. This has to change one day, but I don't have financing to change this right now and it is more then just a couple of minutes.

Thanks again and best,
@kergomard

schmitz-ilias: Hi @kergomard 

Thanks for the quick feedback. Setting title and description here via `ilObjectProperties::storePropertyTitleAndDescription` would indeed go sideways, but I'm just setting title and description in the `ilObject` itself as local properties, nothing is persisted. This is needed, because that is where `ilObject::createMetadata` grabs title and description from.

As it's currently implemented, one could also just set the title to some dummy value, any non-empty string would do. It is overwritten again a few lines lower anyways, that time actually via `ilObjectProperties`. Using the raw title and description entered by the user seemed more robust to me.

For what it's worth, I tested the creation of objects on trunk at least for Courses and Groups, and that seems to work fine. As you already said, this all should be untangled properly at some point, but for now this seems like the slimmest solution.

kergomard: Hi @schmitz-ilias

Thanks, you are right, we are getting a default set of properties back from `getObjectProperies` even if the object does not exist.

So, this should work. My bad.

Best,
@kergomard

kergomard (changes requested):
Hi @schmitz-ilias 

Thanks again, this looks good and I've only a few minor requests for changes. If you think, they are not right, please provide a quick rational:
- Could you please remove the intermediary variable and the comment [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7889/files#diff-b8998dad1df666b2bd7d4bead66771499aa0b8d9d925704b11f6690e5a7acf4aR956-R967). I don't think they add anything.
- I think [these two lines](https://github.com/ILIAS-eLearning/ILIAS/pull/7889/files#diff-b8998dad1df666b2bd7d4bead66771499aa0b8d9d925704b11f6690e5a7acf4aL964-L965) are now superfluous, too. Please remove them. ...or am I missing something.
- Would it be possible to provide an array-access for the use in `ilObjectDIC`? For the other accesses, I have less of an opinion, as they are not what they should be anyways.

Thank you very much and best,
@kergomard

schmitz-ilias:
Hi @kergomard 

I worked in your feedback, please have a look. 

Regarding setting title/description, I originally thought that this would be necessary in case some modifications were done to title/description on storing (cutting off strings that are too long, etc.), but on closer inspection these modifications are done directly in `ilObjectPropertyTitleAndDescription`. So, setting twice is indeed unecessary.

Let me know if there is anything else!

Cheers, @schmitz-ilias

kergomard:
Thank you very much @schmitz-ilias !

I'm now fine with all the changes. Could you please fix the failing tests though? They are due to the dependency now in `ilObjectDIC`. Sorry, I missed that before! I will merge then instantly.

Thanks again and best,
@kergomard

schmitz-ilias:
Done, thanks @kergomard !

PR #7891, opened 2024-08-05
MediaObjects: switch to LOM API

schmitz-ilias:
This PR replaces all usages of the old `MetaData` classes in `MediaObjects` with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md).

To not ruin your dependency management, I introduced a new `MetadataManager` in the `InternalDomainService`. I did not wrap all the used functionality, but instead decided to also offer the API itself through the `MetadataManager`. To me the fully wrapped version looked less readable (but I might of course be a bit biased). Further, because the new API is a bit more specific and a bit more strict in some places, a very slight rejiggering of the creation process of `MediaObjects` was necessary. 

Cheers, @schmitz-ilias


alex40724:
Thanks!

PR #7892, opened 2024-08-05
Forum: ilObjForumGUI set correct default values for sorting

gvollbach:
Should fix: https://mantis.ilias.de/view.php?id=41852

@mjansenDatabay Could you please validate the settings? Thx.

### DIFF
--- a/Modules/Forum/classes/class.ilObjForumGUI.php
+++ b/Modules/Forum/classes/class.ilObjForumGUI.php
@@ -44,6 +44,7 @@ class ilObjForumGUI extends ilObjectGUI implements ilDesktopItemHandling, ilForu
 {
     use ilForumRequestTrait;
 
+    const DEFAULT_THREAD_SORTING = 4;
     private array $viewModeOptions = [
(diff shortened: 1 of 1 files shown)


Modules/Forum/classes/class.ilObjForumGUI.php
```diff
@@ -85,7 +85,7 @@ class ilObjForumGUI extends ilObjectGUI implements ilDesktopItemHandling, ilForu
     protected \ILIAS\Style\Content\Object\ObjectFacade $content_style_domain;
     protected \ILIAS\Style\Content\GUIService $content_style_gui;
     private array $modal_collection = [];
-    protected int $thread_sortation = 1;
+    protected int $thread_sortation = 4;
```
mjansenDatabay: Can we use a constant instead of this magic number (here and below)?

mjansenDatabay:
Fixed with https://github.com/ILIAS-eLearning/ILIAS/commit/8a9f6220447b2e0ce5cea7577e60bb43bef0372e

PR #7893, opened 2024-08-05
T&A 41706: Fix title max length inconsistency

thojou:
This PR is related to Mantis https://mantis.ilias.de/view.php?id=41706 and adds a maximum length validation of 255 to Test, QuestionPool, and LearningModule. This should fix the inconsistency because the general title limit is 255, while the default ilTextInputGUI maximum length is set to 200.


alex40724 (approved):

alex40724:
Thanks!

dsstrassner:
@thojou good to go or do you want an extra review from @mbecker-databay or @kergomard?

thojou:
I've merged this PR into release_8. The code related to the issue has been rewritten in release_9, so cherry-picking isn't applicable. We also need to verify whether the issue still exists in ILIAS 9.

PR #7896, opened 2024-08-06
T&A 41725: Fix Test Results "Access From" Minutes default to number of selected month

thojou:
This PR is related to the Mantis ticket https://mantis.ilias.de/view.php?id=41725.

Previously, an incorrect formatting character was used during DateTime formatting, causing the minutes in the time to display the month value from the date. This has now been corrected to ensure the DateTime is displayed as expected.

This fix can also be cherry-picked into ILIAS 9. In the Trunk version, this behavior has already been reworked and functions as expected.


dsstrassner:
@thojou Could it be, that ILIAS 9 is already refactored?

thojou:
Yes, @dsstrassner, you're right. I would have checked this during the merging process, but since @mbecker-databay already took care of it, all maintained versions and the trunk should work as expected.

PR #7897, opened 2024-08-06
trunk StatusCommand: implement a lazy filter

daniwe4:
Metric now uses a callable as the third parameter. This means that only the requested values are queried when filtering in the status command.

To use the filter in the status command, simply add the -f option. This is followed by the filter string separated by a period.
For example:

php setup/setup.php status -f config.common


dsstrassner:

PR #7907, opened 2024-08-07
[FEATURE] UI: Introduce `Footer` sections

thibsy:
Hi folks,

I have picked up @chfsx's proposal for the new `MainControls\Footer` interface he made in #6458 and incorporated the changes / remarks we have discussed in our meeting on Discord.

The initial proposal talked about a sharing-link-, link-group-, and technical-info-section. After our discussion, we agreed about separating the permanent-link from all other sections, thus introducing a fourth section only for this link. We also talked about links which are represented as an Icon, that could e.g. be used to link social-media accounts. I therefore added this as a fifth section now as well. The factory description lists them in the order I intend to display them - please feel free to weigh in on this. Maybe @yvseiler can provide some insight here? 

Let me go over the changes now:

- Section documentation: I have documented the new sections on the `MainControls\Factory::footer()` description and improved the definition a bit. I was trying to be as abstract as possible, let me know if something is missing or unclear.
- `Button\Shy` rule: to enforce the usage of `Link\Standard` components when providing links, I have added a usage rule that restricts using `Button\Shy` components to only be used as a trigger for opening internal dialogs. We could also check the action inside the implementation in order to be sure.
- `Symbol\Icon` action: the action of `withAdditionalIcon()` is optional because I wanted to enable adding custom icons, such as certificate badges or sponsorships, which do not always require an underlying action.
- Deprecation of `withAdditionalModalAndTrigger()`: We agreed that we do not want consumers to inject modals into this component. I therefore deprecated this method. If you like, I can create a roadmap entry which describes the problem we have here. Removing the method and usages altogether seemed out of scope for this iteration.
- `self` return-type: I have implemented the new methods and refactored the old ones to return `self` now. I did not use `static` here, because there are currently no different versions of the footer (if there ever will be).
- Removal of getters: I removed all getters from the `MainControls\Footer` interface, because they are only for internal use.

I am looking forward to your review!

Kind regards,
@thibsy


klees (changes requested):
Hi @thibsy,

thanks a lot for iterating on this.

* [x] Icons: Would we allow Icons in the Footer to trigger other things then the shy-buttons? Why? If not: Add Icons to rule usage.4.

* [x] wording: We should use "modal" [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7907/files#diff-77f662b9aa891d63305aef829594fdaea425b59ee9c5e256e21556943ca4e854R349), instead of "dialog". We might even consider to strengthen the wording further here: Do we want any inputs hidden behind buttons in the footer or would we just like to show information in the modals? We would suggest the latter...
* [x] a11y: Please add a rule along the lines of "Each section and component in the footer MUST be operationable by keyboard and properly be labelled. This is the case for all parts of ILIAS, however, here it is of special importance, since the Footer is carrying critical information appearing on every page in ILIAS."

Best regards!
@Amstutz and @klees

klees:
Hi @thibsy,

please pass this back once changes are included.

Thanks!

thibsy:
Hi @klees,

Thx for your feedback!

- Icons: I have added them to the usage rule 4 as well.
- A11y: I have added a rule among these lines, but I left out the explanation. It does not seem relevant on the factory.

Kind regards.
@thibsy

yvseiler:
Dear @thibsy 
Sorry for the late feedback and thanks for your work on this. I think it's good that the icon section is separate and not a combination in a link group with icons and without icons, which could make the footer appear confusing. From my point of view, there is a limited use for such icon links to emphasise links, e.g. for social media links or to replace text for icons that have already been learned (e.g. email, telephone,...). 

Regarding the placement of such an icon section (e.g. social media links): These are often more to the right/at the end of a footer, so I agree with you about adding it as a fourth/fifth section.

Best regards
Yvonne

matthiaskunkel:
Jour Fixe, 16 SEP 2024: We highly appreciate this suggestion and accept the PR for ILIAS 10.

thibsy:
Hi @Amstutz and @klees, this should ready for technical review now. Once I have your go-ahead, I will amend the failing unit tests and make our pipelines happy.

Amstutz:
This would be ready to be merged, however, there seems to be a problem with the CS Fixer:

image

If resolved, @klees feel free to merge.

klees:
@Amstutz Wrong PR for your comment, right? This would be for https://github.com/ILIAS-eLearning/ILIAS/pull/7045?

PR #7909, opened 2024-08-07
Glossary: switch to LOM API

schmitz-ilias:
This PR replaces all usages of the old `MetaData` classes in `Glossary` with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md).

I stumbled on some export related methods that seemed unused. I didn't feel comfortable ripping out all the wiring myself, so I only took out the LOM part, and left a hopefully helpful comment so you can follow up at some point (see `ilObjGlossary::exportXMLMetaData`). I also removed a last remnant of Terms having had LOM in `ilGlossaryTerm::_copyTerm`.

I also noticed that the export mechanism that is actually in use does not include LOM in the export (but import mapping for LOM is already in place). I added the last little part needed so that LOM is imported and exported with Glossaries. It might be worthwhile to also include this change in the productive releases. I can provide a separate PR for that if you want, let me know.

Cheers, @schmitz-ilias

### DIFF
--- a/components/ILIAS/Glossary/Export/class.ilGlossaryExporter.php
+++ b/components/ILIAS/Glossary/Export/class.ilGlossaryExporter.php
@@ -123,6 +123,18 @@ class ilGlossaryExporter extends ilXmlExporter
                 "entity" => "common",
                 "ids" => $a_ids);
 
+            $md_ids = [];
+            foreach ($a_ids as $crs_id) {
+                $md_ids[] = $crs_id . ":0:glo";
+            }
+            if ($md_ids !== []) {
+                $deps[] = [
+                    "component" => "components/ILIAS/MetaData",
+                    "entity" => "md",
+                    "ids" => $md_ids
+                ];
+            }
+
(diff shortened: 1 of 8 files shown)


tfamula:
Thank you very much for the PR, @schmitz-ilias!
Special thanks for identifying the unused export related methods. I removed them in a separate commit: https://github.com/ILIAS-eLearning/ILIAS/commit/687f6532a8c46f0630e8c2be92ec486d1ef9df25. I could not detect any broken functionality after removing all the allegedly unused code and hope that I caught all the places.
I also added the missing LOM part for the export in release 8 and 9: https://github.com/ILIAS-eLearning/ILIAS/commit/5d77d1c4a3f97c26d2ff93f650081e4b8351e922 and https://github.com/ILIAS-eLearning/ILIAS/commit/ebf7ef129a27ed1b9a1b368acc48a736d52a29f5

@tfamula

PR #7911, opened 2024-08-08
CI: Potential fix for missing problems in `php-cs-fixer` run

mjansenDatabay:
Currently the `php-cs-fixer` check returns the correct result when detecting violations, but unfortunately 1 cannot see any detailed results, e.g. which files are violating our code style.

I noticed that the progress bar seems to be displayed in an odd way, instead of a modified line the bar seems to be rendered on multiple lines (I guess overlaying the reported problems).

See:

41874-Cron-php-does-not-work-anymore-when-account-has-IP-restriction-·-ILIAS-eLearning-ILIAS-a8894d2

### DIFF
--- a/CI/PHP-CS-Fixer/run_check.sh
+++ b/CI/PHP-CS-Fixer/run_check.sh
@@ -19,7 +19,7 @@ else
   	if [ -f ${FILE} ]
   	then
 	  	echo "Check file: ${FILE}"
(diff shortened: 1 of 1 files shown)


pascalseeland (approved):
LGTM

mjansenDatabay:
Hopefully fixed with #7955

PR #7912, opened 2024-08-08
Weblink: switch to LOM API, export LOM as tail dependency

schmitz-ilias:
This PR replaces most usages of the old `MetaData` classes in `WebResource` with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md). As part of this, LOM of Weblinks is now exported as a tail dependency instead of directly in the Weblink XML.

There still remain some usages of the old `MetaData` classes to ensure that exports from previous releases are still imported properly with their LOM. The current plan is to remove these usages with ILIAS 11.

Due to the current state of the trunk I was not able to test the changes, but at least on 9 the export/import part seems to work.

Cheers, @schmitz-ilias

### DIFF
--- a/components/ILIAS/WebResource/classes/class.ilObjLinkResource.php
+++ b/components/ILIAS/WebResource/classes/class.ilObjLinkResource.php
@@ -18,18 +18,25 @@
 
 declare(strict_types=1);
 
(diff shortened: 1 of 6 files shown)


smeyer-ilias (approved):
Thanks @schmitz-ilias

PR #7913, opened 2024-08-08
Object: Fix default command target for SAHS list items

iszmais:
https://mantis.ilias.de/view.php?id=31690

This restores a fix from 2022 (https://github.com/ILIAS-eLearning/ILIAS/pull/4622) where this was resolved.
Unfortunately, the used parameter then seems to be not in use anymore due to the refactoring of the component.

### DIFF
--- a/Services/Object/classes/class.ilObjectListGUI.php
+++ b/Services/Object/classes/class.ilObjectListGUI.php
@@ -3162,7 +3162,7 @@ class ilObjectListGUI
             $modifySAHS = $this->modifySAHSlaunch($def_command['link'], $def_command['frame']);
             $def_command['link'] = $modifySAHS[0];
             $def_command['frame'] = $modifySAHS[1];
-            $new_viewport = !in_array($this->getDefaultCommand()['frame'], ['', '_top', '_self', '_parent'], true); // Cannot use $def_command['frame']. $this->default_command has been edited.
+            $new_viewport = !in_array($def_command['frame'], ['', '_top', '_self', '_parent'], true);
             $link = $this->ui->factory()
                              ->link()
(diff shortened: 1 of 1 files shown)


iszmais:
Hi @lscharmer,

since you provided the initial fix and since this affects mostly the dashboard (favourites), I would appreciate if you could take care of the review here.
I will provide PRs for the other ILIAS Versions.

Greetings,
Ingmar Szmais

kergomard:
Hi @iszmais 

I'm pretty sure we do not need a shepherd here, as this is a change in `ILIASObject` and I'm responsible for that.

This looks good and right. I will merge your PRs.

Best,
@kergomard

iszmais:
@kergomard 

Even better, thank you very much!

I’ve been quite confused since the "Object" Section in mantis is flagged as without authority.
Maybe we want to change that to prevent future miscommunications.

Greetings,
@iszmais

kergomard:
Can you explain what you mean by "the "Object" Section in mantis is flagged as without authority"?

iszmais:
@kergomard When report a bug in mantis and select the category "Object" this is prefixed with a "¥" which I assumed means it is currently not maintained.

kergomard:
Hi @iszmais 

I was suspecting something like this. No the "¥" means "This object has no simply discernible surface, it could confuse the normal ILIAS user in the list of categories, and thus it should be shown at the end of the list".  I know, convoluted explanation, but it this way the more knowledgeable users find all possible categories and others will never scroll that far and not get confused by the more technical entries.

Best,
@kergomard

iszmais:
Ah ok!
Thank you for the elaborate explanation!

Greetings,
@iszmais

PR #7916, opened 2024-08-08
Course: switch to LOM API

schmitz-ilias:
This PR replaces most usages of the old `MetaData` classes in `Course` with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md).

I was not able to test the changes related to ECS, but everything else seems to work fine.

There still remain some usages of the old `MetaData` classes to ensure that exports from ILIAS 8 and older are still imported properly with their LOM. The current plan is to remove these usages with ILIAS 11.

Cheers, @schmitz-ilias


smeyer-ilias (approved):
Thanks @schmitz-ilias

PR #7917, opened 2024-08-08
Group: remove usages of ilMDUtils

schmitz-ilias:
This PR removes all usages of the outdated method `ilMDUtils::ilMDUtils::_fillHTMLMetaTags` from `Group`.

The only occurrences of the old `MetaData` classes in `Group` are there to ensure that exports from ILIAS 8 and older are still imported properly with their LOM. The current plan is to remove these usages with ILIAS 11.

Cheers, @schmitz-ilias


smeyer-ilias:
Thanks @schmitz-ilias

PR #7919, opened 2024-08-09
Calendar/Mail: Use (more or less) unique filenames for "ics" mail att…

mjansenDatabay:
…achments

See: https://mantis.ilias.de/view.php?id=41884


smeyer-ilias:
Thanks for fixing this

PR #7923, opened 2024-08-09
Test/TestQuestionPool: switch to LOM API

schmitz-ilias:
This PR replaces all usages of the old `MetaData` classes in `Test` and `TestQuestionPool` not related to export/import with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md). The usages related to import/export will be covered in a different PR, see #7924.

Cheers, @schmitz-ilias

### DIFF
--- a/components/ILIAS/Test/classes/class.ilObjTest.php
+++ b/components/ILIAS/Test/classes/class.ilObjTest.php
@@ -48,6 +48,7 @@ use ILIAS\Test\Export\CSVExportTrait;
 use ILIAS\Refinery\Factory as Refinery;
 use ILIAS\Filesystem\Filesystem;
 use ILIAS\Filesystem\Stream\Streams;
+use ILIAS\MetaData\Services\ServicesInterface as LOMServices;
 
 /**
  * Class ilObjTest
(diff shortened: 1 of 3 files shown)


kergomard (commented):
Thank you very much for these improvements @schmitz-ilias !

I would have two very small change requests:
- Could you please rename the variable `lom_services` to `lo_metadata`. My rational: I don't think the `_service`-part adds any information, but on the other hand "lom" is too cryptic to be easily understood. `learning_object_metadata` seems slightly too verbose. The most important part, I think, is `metadata` and so we can shorten the "learning_objective" to "lo", showing that this is a special type of metadata. I'm understanding this right that the part of the `MetaData`-component exposed here only deals with metadata compliant with the LOM standard, right?
- And if you could be so kind to change `strlen($author) == 0` to `$author === ''` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7923/files#diff-377cf2c64a9111fe133f461ef77079e17cd9132640dc33f5c5e0ef927c46de64R4348) I would be gratefull. 

Other then this I'm fine with the changes and would merge as soon as you changed this.

Thanks again and best,
@kergomard

schmitz-ilias:
Hi @kergomard,
correct, only Learning Object Metadata is available through the API, Custom Metadata is a different can of worms. Thanks for your feedback, I've made the suggested changes.

kergomard:
Thank you very much @schmitz-ilias !

PR #7924, opened 2024-08-09
Test/TestQuestionPool: LOM as tail dependency in export

schmitz-ilias:
This PR replaces almost all usages of the old `MetaData` classes in `Test` and `TestQuestionPool` related to export/import. For everything else, see #7923.

From what I could gather from [28891](https://mantis.ilias.de/view.php?id=28891) and a bit of testing, LOM of `Tests` and `TestQuestionPools` is currently directly included in the XML of the Test(QuestionPool) itself, but is not actually imported. When importing Tests, the LOM seems completely unused, and TestQuestionPools only use title and description.

This PR includes LOM as a tail dependency in the exports, and removes it from the XML of the objects themselves. Further, it adds fields for title and description to the XML for TestQuestionPools, and makes sure they are read out on import. Unfortunately, tail dependencies do not apply everywhere. Exports of TestQuestionPools via the question table, and exports of Tests with results of participants sidestep the `Export` infrastructure somewhat, so this PR will simply remove LOM from those types of exports.

If you prefer consistency between export types, I could also rewrite this PR to only remove the LOM from exports, without adding it back in as a tail dependency. Keeping the LOM for some or all types of exports does not have much value in my opinion: it is currently not imported, and is also not in a format that could be imported in the future.

Be aware that I tested most of these changes in 9, because import/export is currently broken in trunk.

With this PR and #7923, the only usage of old `MetaData` classes in `Tests` and `TestQuestionPools` are `ilMD` and `ilMDSaxParser` in `ilQuestionPageParser`. As far as I can see, the only purpose their inclusion serves is the import of LOM of MediaObjects that might be included in Question Pages. `ilMD` and `ilMDSaxParser` are deprecated (as of a few weeks ago), and will be removed with ILIAS 11, but even if they are not replaced in `ilQuestionPageParser` I think the loss of functionality is bearable. Let me know if you have a strong opinion about this, I will also put it up for discussion in an upcoming JF along with a few similar cases.

Cheers, @schmitz-ilias

### DIFF
--- a/components/ILIAS/Test/classes/class.ilObjTest.php
+++ b/components/ILIAS/Test/classes/class.ilObjTest.php
@@ -4127,9 +4127,6 @@ class ilObjTest extends ilObject
     {
         $this->mob_ids = [];
 
-        // MetaData
-        $this->exportXMLMetaData($a_xml_writer);
-
         // PageObjects
         $expLog->write(date("[y-m-d H:i:s] ") . "Start Export Page Objects");
         $this->bench->start("ContentObjectExport", "exportPageObjects");
@@ -4150,20 +4147,6 @@ class ilObjTest extends ilObject
         $this->exportFileItems($a_target_dir, $expLog);
         $this->bench->stop("ContentObjectExport", "exportFileItems");
         $expLog->write(date("[y-m-d H:i:s] ") . "Finished Export File Items");
-    }
-
-    /**
-    * export content objects meta data to xml (see ilias_co.dtd)
-    *
-    * @param	object		$a_xml_writer	ilXmlWriter object that receives the
-    *										xml data
-    */
-    public function exportXMLMetaData(&$a_xml_writer)
-    {
-        $md2xml = new ilMD2XML($this->getId(), 0, $this->getType());
-        $md2xml->setExportMode(true);
-        $md2xml->startExport();
-        $a_xml_writer->appendXML($md2xml->getXML());
     }
 
     /**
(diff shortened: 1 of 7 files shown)


dsstrassner:
hi @kergomard I added you as a reviewer, as you already looked into #7923. If you think @thojou or @mbecker-databay should also have a look or if you do not have the time, please add them also or reassign the review task.

kergomard (changes requested):
Hi @schmitz-ilias 

Thank you very much for the PR and sorry it took me so long to get around to this! This looks good actually, I would just have one question/request:
- Why are you appending XML as a string [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7924/files#diff-62c27507778f872232b5d011252656b37408b865ed8f479f2be3525c70ab058bR406-R408) instead of using `ilXmlWriter->xmlElement()`? If there is no good reason: Might I ask you to change this?

Thank you very much and best,
@kergomard

schmitz-ilias:
>     * Why are you appending XML as a string [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7924/files#diff-62c27507778f872232b5d011252656b37408b865ed8f479f2be3525c70ab058bR406-R408) instead of using `ilXmlWriter->xmlElement()`? If there is no good reason: Might I ask you to change this?

No good reason, using `ilXmlWriter->xmlElement()` is definitely nicer. Fixed!

kergomard:
Thank you once more! Merged and picked to ILIAS 10.
@kergomard

PR #7929, opened 2024-08-13
Tree/PathGUI: Properly handle HTML attributes

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=22578


fwolf-ilias:

PR #7932, opened 2024-08-13
T&A Bugfix #0029216: Missing Participant Results in Test Export/Import

fhelfer:
This bug was initially reported in https://mantis.ilias.de/view.php?id=29216. During the bug-fixing process, however, we discovered in #6335 that the issue was more complex than we originally thought.

As a result, we began working on a solution to ensure that the feature would not be abandoned. Before I arrived at a solution involving significant refactoring or a new implementation, I came across this bug fix, which appears to resolve the issue satisfactorily.

Further analysis is ongoing, and we are exploring additional ways to address the problem. However, this bug fix provides a quick solution that prevents the feature from being removed in version 10 and allows it to function properly in other versions as well.

Unfortunately, the solution impacts ilExportGUI, meaning this is not a localized bug fix in T&A. The core issue was that the import of the questions was not handled correctly because a shortcut through ilImport and ilExport was chosen. As a result, the export did not include a manifest file in the zip, leading to an exception in ilImport::doImportObject during the import process, which requires the manifest to function correctly.

This step is crucial to ensure that the questions are not lost. I have removed the shortcut, so the Test Import/Export now runs the entire way through ilImport/ilExport.

This is open for discussion and will be addressed further in our next T&A meeting. After we've decided on the best solution, I will implement fixes for all other versions.

Best @fhelfer

### DIFF
--- a/Modules/Test/classes/class.ilObjTestGUI.php
+++ b/Modules/Test/classes/class.ilObjTestGUI.php
@@ -1340,43 +1340,29 @@ class ilObjTestGUI extends ilObjectGUI implements ilCtrlBaseClassInterface
         $map = $imp->getMapping();
         $map->addMapping('Modules/Test', 'tst', 'new_id', $newObj->getId());
 
+        $fileName = ilSession::get('tst_import_subdir') . '.zip';
+        $fullPath = ilSession::get('tst_import_dir') . '/' . $fileName;
+
         if (is_file(ilSession::get("tst_import_dir") . '/' . ilSession::get("tst_import_subdir") . "/manifest.xml")) {
             $newObj->saveToDb();
 
             ilSession::set('tst_import_idents', $_POST['ident'] ?? '');
             ilSession::set('tst_import_qst_parent', $questionParentObjId);
 
-            $fileName = ilSession::get('tst_import_subdir') . '.zip';
-            $fullPath = ilSession::get('tst_import_dir') . '/' . $fileName;
-
             $imp->importObject($newObj, $fullPath, $fileName, 'tst', 'Modules/Test', true);
         } else {
-            $qtiParser = new ilQTIParser(ilSession::get("tst_import_qti_file"), ilQTIParser::IL_MO_PARSE_QTI, $questionParentObjId, $_POST["ident"] ?? '');
-            if (!file_exists(ilSession::get("tst_import_results_file"))
-                && (!isset($_POST["ident"]) || !is_array($_POST["ident"]) || !count($_POST["ident"]))) {
-                $qtiParser->setIgnoreItemsEnabled(true);
-            }
-            $qtiParser->setTestObject($newObj);
-            $qtiParser->startParsing();
-            $newObj = $qtiParser->getTestObject();
-            $newObj->saveToDb();
-
-            $questionPageParser = new ilQuestionPageParser($newObj, ilSession::get("tst_import_xml_file"), ilSession::get("tst_import_subdir"));
-            $questionPageParser->setQuestionMapping($qtiParser->getImportMapping());
-            $questionPageParser->startParsing();
-
             $test_importer = new ilTestImporter();
-            $test_importer->addTexonomyAndQuestionsMapping($qtiParser->getQuestionIdMapping(), $newObj->getId(), $map);
-            $test_importer->importRandomQuestionSetConfig($newObj, ilSession::get("tst_import_xml_file"), $map);
+            $test_importer->setImport($imp);
+            $test_importer->setInstallId(IL_INST_ID);
+            $test_importer->setImportDirectory(ilSession::get("tst_import_dir") . '/' . ilSession::get("tst_import_subdir"));
+            $test_importer->init();
 
-            if (file_exists(ilSession::get("tst_import_results_file"))) {
-                $results = new ilTestResultsImportParser(ilSession::get("tst_import_results_file"), $newObj);
-                $results->setQuestionIdMapping($qtiParser->getQuestionIdMapping());
-                $results->startParsing();
-            }
-
-            $newObj->saveToDb();
-            $newObj->update();
+            $test_importer->importXmlRepresentation(
+                '',
+                0,
+                '',
+                $map,
+            );
         }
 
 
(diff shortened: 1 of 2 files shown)


Services/Export/classes/class.ilExportGUI.php
```diff
@@ -271,7 +271,7 @@ public function createExportFile(): void
                 } elseif ($this->getParentGUI() instanceof ilContainerGUI) {
                     $this->showItemSelection();
                     return;
-                } elseif ($format == "xml") {        // standard procedure
+                } elseif ($format == "xml" || $format === 'xmlres') {        // standard procedure
```
mjansenDatabay: This introduces a dependency in the `Export` component to a concrete identifier of an export provider (here: the `Test`).

fhelfer: Yes, also not happy with this

fhelfer (commented):
.

Modules/Test/classes/class.ilTestExporter.php
```diff
@@ -41,7 +49,15 @@ public function getXmlRepresentation(string $a_entity, string $a_schema_version,
...
         $expFactory = new ilTestExportFactory($tst);
-        $testExport = $expFactory->getExporter('xml');
+        if ($this->http->wrapper()->post()->has('format')) {
+            $format = $this->http->wrapper()->post()->retrieve(
+                'format',
+                $this->refinery->kindlyTo()->string()
+            );
+        } else {
+            $format = 'xml';
+        }
```
fhelfer: Same here, introduces dependency on http

Modules/Test/classes/class.ilObjTestGUI.php
```diff
@@ -1348,25 +1348,6 @@ public function importVerifiedFileObject()
...
-            $questionPageParser = new ilQuestionPageParser($newObj, ilSession::get("tst_import_xml_file"), ilSession::get("tst_import_subdir"));
-            $questionPageParser->setQuestionMapping($qtiParser->getImportMapping());
-            $questionPageParser->startParsing();
-
-            if (file_exists(ilSession::get("tst_import_results_file"))) {
-                $results = new ilTestResultsImportParser(ilSession::get("tst_import_results_file"), $newObj);
-                $results->setQuestionIdMapping($qtiParser->getQuestionIdMapping());
-                $results->startParsing();
-            }
-            $newObj->update();
```
fhelfer, line 1369: same code here
https://github.com/ILIAS-eLearning/ILIAS/blob/53b9cc075cb47c676e3ede50a15077f4fc8f433e/Modules/Test/classes/class.ilTestImporter.php#L124-L130

Services/Export/classes/class.ilExportGUI.php
mjansenDatabay: Please add the coding authorities for the `Export` component to decide upon the required changes as well.

fhelfer: Cannot edit assignees myself. Insufficient rights
That would be @smeyer-ilias right?

Modules/Test/classes/class.ilObjTestGUI.php
```diff
@@ -1348,25 +1348,6 @@ public function importVerifiedFileObject()
...
             $map->addMapping('Modules/Test', 'tst', 'new_id', $newObj->getId());
             $imp->importObject($newObj, $fullPath, $fileName, 'tst', 'Modules/Test', true);
-        } else {
-            $qtiParser = new ilQTIParser(ilSession::get("tst_import_qti_file"), ilQTIParser::IL_MO_PARSE_QTI, $questionParentObjId, $_POST["ident"] ?? '');
-            if (!file_exists(ilSession::get("tst_import_results_file"))
-                && (!isset($_POST["ident"]) || !is_array($_POST["ident"]) || !count($_POST["ident"]))) {
-                $qtiParser->setIgnoreItemsEnabled(true);
-            }
-            $qtiParser->setTestObject($newObj);
-            $qtiParser->startParsing();
```
fhelfer, line 1358: FYI @mjansenDatabay, @mbecker-databay, @dsstrassner, @kergomard, @thojou 
Neue Erkenntnis, das ganze ist sogar noch simpler als gedacht. Das ist bereits Code der aus ilTestImporter kopiert wurde und diese wichtige Zeile fehlt. Damit funktioniert das ganze auch wieder und würde den PR zu einem Einzeiler machen
Ich finde trotzdem den Ansatz besser, die Manifest Datei mit in die Zip zu packen und sich so auch die Duplikation von Code zu sparen.

```php
$newObj` = $qtiParser->getTestObject();
````

dsstrassner:
I spoke with @mbecker-databay at the ILIAS DevConf in Graz last week and he and @fhelfer will investigate the dev-guide regarding the media files to provide a modern export solution for trunk which also refactors the code and helps the T&A for the future.
The bug fix could be applied to ILIAS 8 and ILIAS 9 if @kergomard agrees.

dsstrassner:
@fhelfer Please proceed with the one line solution for ILIAS 8 and 9. ILIAS 10 needs the further analysis as discussed in the TechSquad.

fhelfer:
Wasn't quite the one-liner, I could not figure out how I managed to get it working the first time. But with the changes copied from [ilTestImporter](https://github.com/ILIAS-eLearning/ILIAS/pull/7932/files#diff-1cb659c07cfa02fb1a744e11d5b0cf83003aa4d4a493c7f0943005f089d4cfffR1391-R1396) I managed to get it running.
Not quite the best solution, but as this is not the goal, it will do for now until my analysis provides a better way of fixing.

Best @fhelfer

kergomard:

We are getting there!

Just one more thing:
- Couldn't we instead of copying the code from ilTestImporter, move the first part of the code to a function `public function addTexonomyAndQuestionsMapping(array $question_id_mapping, ilImportMapping $mapping): ilImportMapping;`, make `ilTestImporter::importRandomQuestionSetConfig()` `public` and move `$newObj->questions = []` inside it? This would avoid some duplicate code and the only thing we need to do is to instantiate an `ilTestImporter` which should be rather cheap.

What do you think?

Thanks again and best,
@kergomard

fhelfer:
Ty for your review @kergomard,

I reduced the duplicated code a bit more, as you can see.
Please let me know if you are satisfied with this solution.

Best @fhelfer

kergomard:
Thank you very much @fhelfer for the update ! Merged and picked to release_9 and ported to ILIAS 10 and trunk.

PR #7935, opened 2024-08-13
Fix rollup config for Bar Chart JS

tfamula:
While I was working on #7691, I have found out by the way that the rollup config for the Bar Chart JS is faulty.
Please also cherry-pick to release 9 and trunk.


Amstutz:
thx a lot!

Amstutz:
picked to 9 and trunk as well.

PR #7940, opened 2024-08-14
Fix: Keep filter data during ui data table rendering

thojou:
While refactoring LUI to use UI DataTables, I discovered an issue in the current implementation related to a TODO item. The $filter_data array passed to the DataRetrieval methods is always empty. This happens because it gets overwritten with an empty array during the rendering process, which seems unintended given the TODO comment.

I modified the code to keep filter data already passed to the table like in the following snippet:

```php
$filter = $this->ui_service->filter()->standard(/*..*/)
    ->withRequest($request);

$this->ui_factory->table()->data(/*...*/)
    ->withRequest($request)
    ->withFilter($filter->getData());
            
$this->ui_renderder->render([
   $filter,
   $table
]);
```

With this usage and the adjustments from the commit, the `$filter_data` is filled inside the DataRetreival as expected.

### DIFF
--- a/components/ILIAS/UI/src/Implementation/Component/Table/Data.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Table/Data.php
@@ -137,7 +137,7 @@ class Data extends AbstractTable implements T\Data
         }
 
         return [
-            $table->withFilter($filter_data),
+            $table,
             $view_controls
         ];
     }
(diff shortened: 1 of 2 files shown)


klees (approved):
Hi @thojou,

LGTM, but a little added context:

We are looking to integrate filters directly with the `Data Table`, just like we do with the `View Controls`. We are not there yet (obviously...) and the problem you discovered here is indeed a problem, even if we had added filters already. But the actual processing of the filter values etc. will become internalized at some point. Stay tuned =)

Best regards!
@thibsy and @klees

klees:
Picked to release_9 and release_8.

PR #7948, opened 2024-08-15
Chatroom: Remove unused dependency node-mysql

lscharmer:
This PR removes the unused node dependency `node-mysql` from the dependency list of the Chatroom.


mjansenDatabay:
Yeah, `-1` libs/deps :-)

PR #7955, opened 2024-08-17
[FIX] start only one PHP-CS-Fixer run in GitHub actions

thibsy:
Hi all,

This PR improves the PHP-CS-Fixer GitHub action, so **all** changed files will be checked in **one** run. 

Until now a separate run has been started for every file - which is unnecessary - and did not display what file is invalid. Using one PHP-CS-Fixer run will output the invalid files nicely now.

Kind regards,
@thibsy


mjansenDatabay:
Thx @thibsy for the fix, I also noticed recently that the actual problems are not shown in the build log. Great to have a solution for this.

PR #7716, opened 2024-06-27
T&A, fix Test for QuestionpoolExport

nhaagen:
Seed: 1719492809

1) ilQuestionpoolExportTest::testConstruct
Error: Undefined constant "IL_INST_ID"
/home/runner/work/ILIAS/ILIAS/components/ILIAS/TestQuestionPool/classes/class.ilQuestionpoolExport.php:62
/home/runner/work/ILIAS/ILIAS/components/ILIAS/TestQuestionPool/tests/ilQuestionpoolExportTest.php:41

PR #7957, opened 2024-08-19
Fixes questions in question pool being saved on every cmd call regardless of cmd

matheuszych:
When trying to edit a question inside of a question pool, regardless of the cmd, the question is always saved. This leads to partially empty requests which reset the questions content such as title or answer possibilities (example Multiple Choice Single Answer Question).

I believe the fix is a missing return statement which would prevent the saving of the question if the cmd is not ment to do so.

### DIFF
--- a/components/ILIAS/Test/tests/ilTestBaseTestCaseTrait.php
+++ b/components/ILIAS/Test/tests/ilTestBaseTestCaseTrait.php
@@ -53,6 +53,9 @@ trait ilTestBaseTestCaseTrait
         if (!defined("ILIAS_LOG_FILE")) {
             define("ILIAS_LOG_FILE", '/var/log/ilias.log');
         }
+        if (!defined("IL_INST_ID")) {
+            define("IL_INST_ID", '0');
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


dsstrassner:
@kergomard please have a look. Seems to be similar to #7716?

kergomard:
Thank you very much  @matheuszych for the PR and @dsstrassner for pushing this to the top of my stack.
You are obviously right @matheuszych , I had to cherry-pick your commit manually though, as the fix to the tests were already applied. Ideally you would keep these kind of fixes separate in the future.

Fixes are here: https://github.com/ILIAS-eLearning/ILIAS/commit/d078b14497d8ada3f49654311af7f3e7da3bc734

Thanks again and best,
@kergomard

matheuszych:
Hello @kergomard,

I am unsure whether the constant issue is resolved. The constant was [added](https://github.com/ILIAS-eLearning/ILIAS/pull/7716) to the bass class of `TestQuestionPool` component but it is also needed in the `Test` component. By defining it in the trait, as I did here, it would be available in both cases since the trait is used in both the `assBaseTestCase.php` and `ilTestBaseTestCase`.

P.S. I implemented this fix in this PR because otherwise, the pipeline would fail. (I also did that in some other PRs for the same reason.)

Best regards,
@matheuszych

kergomard:
Hi @matheuszych 

It was added to the trait, just a few lines up. The Pull Request you referenced was closed, as I had added the constant before.

Best,
@kergomard

PR #7960, opened 2024-08-19
CS Fixer: Prevent multiple filenames to be given as 1 argument

lscharmer:
This PR changes the CS Fixer, so that if multiple files are to be checked, they are not given to the underlying CS Fixer command as one argument but as one argument for each file.

### DIFF
--- a/scripts/PHP-CS-Fixer/run_check.sh
+++ b/scripts/PHP-CS-Fixer/run_check.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if [[ -z ${GHRUN} ]]; then
+if [[ -z "${GHRUN}" ]]; then
   RUNCSFIXER=$(vendor/composer/vendor/bin/php-cs-fixer fix --using-cache=no --dry-run -vvv --config=./scripts/PHP-CS-Fixer/code-format.php_cs $@)
   RESULT=$?
   if [[ ${RESULT} -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


thibsy (changes requested):
Hi @lscharmer,

Thx for looking into this!

If I unterstand correctly, this change is necessary because the changed files would be fed to the PHP-CS-Fixer command in the form of `"file1 file2 file3"`. Could this issue therefore be resolved by simply removing the quotations around the variable, so we do not pass the argument as one string?

Otherwise, please change:

- [x] There are some other changes to the `if` statements which do not seem necessary. Do I overlook something? If not, please remove them.
- [ ] You currently only `echo` one file from the array, so the PHP-CS-Fixer will only be run for the first changed file. Please use `${CHANGED_FILES[@]}` there.

Kind regards,
@thibsy

lscharmer:
Hi @thibsy,

thank you for your feedback.

> If I unterstand correctly, this change is necessary because the changed files would be fed to the PHP-CS-Fixer command in the form of `"file1 file2 file3"`.

Yes

> Could this issue therefore be resolved by simply removing the quotations around the variable, so we do not pass the argument as one string?

Yes it will, but if any of these files contain spaces they are treated as separate arguments and not one. The solution with `xargs` ensures that they are separated by newlines. AFAIK we currently don't have the case where spaces are in the file name but I think that scripts should be able to handle any valid filename (With spaces within the filename that script would break).

> * [ ]  There are some other changes to the `if` statements which do not seem necessary. Do I overlook something? If not, please remove them.

No they are not necessary, just simplifications (Within this script the flags used with `[[` work the same with `[` but `[[` is exclusive to bash). I try to keep the used syntactic features to a minimum so that they are better to understand. As this is subjective I will remove this from the PR.

> * [ ]  You currently only `echo` one file from the array, so the PHP-CS-Fixer will only be run for the first changed file. Please use `${CHANGED_FILES[@]}` there.

`$CHANGED_FILES` is not an array. It's a string which will be interpreted as an array by splitting on each space if used as an array (only if you define a var as an array in bash then `$CHANGED_FILES` will expand to the first element of the array and not the whole).

Best regards
@lscharmer

thibsy (approved):
Hi @lscharmer

Thx for the clarification! Using `xargs` seems reasonable here, this LGTM now.

Kind regards,
@thibsy

PR #7963, opened 2024-08-19
OSC: Move HTML escaping from DB to frontend

lscharmer:
This PR changes that messages aren't escaped when writing them to the database but are saved unescaped in the database and are escaped when outputting them in HTML instead.

### DIFF
--- a/components/ILIAS/Chatroom/chat/SocketTasks/ConversationMessage.js
+++ b/components/ILIAS/Chatroom/chat/SocketTasks/ConversationMessage.js
@@ -1,5 +1,20 @@
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
 var Container  = require('../AppContainer');
-var HTMLEscape = require('../Helper/HTMLEscape');
 var UUID = require('node-uuid');
 
 module.exports = function(conversationId, userId, message) {
@@ -31,7 +46,7 @@ module.exports = function(conversationId, userId, message) {
 			var messageObj = {
 				conversationId: conversationId,
 				userId: userId,
-				message: HTMLEscape.escape(message),
+				message: message,
 				timestamp: (new Date).getTime(),
 				id: UUID.v4()
 			};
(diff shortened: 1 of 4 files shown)


components/ILIAS/OnScreenChat/resources/onscreenchat.js
```diff
@@ -1022,16 +1038,23 @@
 
 			let messageDate = new Date();
 			messageDate.setTime(messageObject.timestamp);
+			const n = 'm' + new Date().getTime();
```
mjansenDatabay: We should find another variable name for `n`.

components/ILIAS/OnScreenChat/classes/Setup/UpdateSteps.php
```diff
@@ -0,0 +1,55 @@
...
+    public function step_1(): void
+    {
+        $replace = [
+            '&lt;' => '<',
+            '&gt;' => '>',
+            '&amp;' => '&',
+            '&quot;' => '"',
+        ];
+
+        $s = 'message';
```
mjansenDatabay: Can we find a better name fror `$s`?

mjansenDatabay (changes requested):
Hi @lscharmer ,

please have a look at `components/ILIAS/OnScreenChat/classes/Provider/OnScreenChatProvider.php` and check whether or not we have to introduce proper output escpaing for the main bar entry when presenting conversation messages in the slate.

Best regards,
Michael

components/ILIAS/OnScreenChat/resources/onscreenchat.js
lscharmer: Maybe `m` ;)

mjansenDatabay: :facepalm:

components/ILIAS/Chatroom/chat/SocketTasks/ConversationMessage.js
```diff
@@ -1,3 +1,19 @@
...
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
 var Container  = require('../AppContainer');
 var HTMLEscape = require('../Helper/HTMLEscape');
```
mjansenDatabay: Can we remove this if `HTMLEscape.escape` is not used anymore?

lscharmer:
> Hi @lscharmer ,
> 
> please have a look at `components/ILIAS/OnScreenChat/classes/Provider/OnScreenChatProvider.php` and check whether or not we have to introduce proper output escpaing for the main bar entry when presenting conversation messages in the slate.
> 
> Best regards, Michael

Hi @mjansenDatabay,

messages shown in the slate are rendered with `$DIC->ui()->factory()->item()->shy()` and the `withDescription` method is used for displaying the chat message. This method html escapes the given description already. So currently messages are displayed double escaped in the menu slate. With this PR they will be only escpaed once.

Best regards
@lscharmer

PR #7970, opened 2024-08-20
Changes to the install.md suggested by Christian Suessenguth 

fwolf-ilias:
Thank you for looking into it @acgruber.

### DIFF
--- a/docs/configuration/install.md
+++ b/docs/configuration/install.md
@@ -111,20 +111,20 @@ is set to `COMPACT`.
 
 The ILIAS Testserver (https://test8.ilias.de) is currently configured as follows:
 
-| Package        | Version                     |
-|----------------|-----------------------------|
-| Distribution   | Ubuntu 20.04 LTS            |
-| MariaDB        | 10.0.38                     |
-| PHP            | 8.0                         |
-| Apache2        | 2.4.18                      |
-| zip            | 3.0                         |
-| unzip          | 6.00                        |
-| JDK            | OpenJDK 8                   |
-| Node.js        | 10.23.0                     |
-| wkhtmltopdf    | 0.12.5                      |
-| Ghostscript    | 9.26                        |
-| Imagemagick    | 6.8.9-9 Q16                 |
-| MathJax        | 2.7.9                       |
+| Package        | Version          |
+|----------------|------------------|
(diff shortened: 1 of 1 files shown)


acgruber:
To document the background of this PR, here's the original feedback we've received from CS:

> 1. The Apache2 snippet shows the following:
> Options FollowSymLinks -Indexes
> 
> There is a + or a - missing in front of FollowSymLinks otherwise Apache2 is not (re)starting. This is also missing in the install instructions for ILIAS 7.
> 
> 2. Below the section "Install other Dependencies"
> openjdk-7-jdk and openjdk-8-jdk
> 
> are being listed. Apart from the confusion that a user does not know if the first or the second line is correct those packages are only available for Debian sid – which is veeeery old. The newest release for Debian 12 (bookworm) is openjdk-17-jdk
> 
> 3. Those backticks should not be there:
> "chown www-data:www-data \`/var/www/html"
> "chown www-data:www-data \`/var/www/files"
> 
> 4. In the example configuration file the directory "errorlog_dir" : "/var/www/logs/" is mentioned but it has not been created during the previous instructions. Maybe a hint regarding this would be helpful.
> 
> In the example configuration theres also a full name and email address which can easily be found/parsed by SPAM bots. Maybe an address like [example@domain.de](mailto:example@domain.de) would be better.

acgruber (commented):
Hi @fwolf-ilias,

Thank you for looking into the recommendations and suggesting the changes accordingly.

I have nothing to add from my editor's point of view, but would like to kindly ask for a small feedback from @klees:

Richard, it has been suggested to remove your personal data from this document to avoid spam mails. Would you agree on that or prefer keeping your contact details?

Thanks & cheers!

klees:
Hi @acgruber,

I have never received any spam mail caused by this, I guess, and if so, it's to late now, my contact data is out there =)

Initially I have put my contact there for just a little publicity, but I'm fine either way, so I leave the decision up to you.

Kind regards!

acgruber:
Hi @klees,

Thanks for the quick reply. I suspected as much, so let's keep that bit of publicity going :)

@fwolf-ilias: Could you please leave Richard's details in the document and merge the remaining changes?

Thanks & cheers!

PR #7971, opened 2024-08-21
Replacement of the legacy modal in assLongMenuGUI

matheuszych:
These changes are part of the [Legacy-UI refactoring](https://docu.ilias.de/go/wiki/wpage_7320_1357).
I replaced the legacy ilModalGUI with the newer Kitchen Sink interruptive Modal.

### DIFF
--- a/components/ILIAS/TestQuestionPool/classes/class.assLongMenuGUI.php
+++ b/components/ILIAS/TestQuestionPool/classes/class.assLongMenuGUI.php
@@ -16,8 +16,8 @@
  *
  *********************************************************************/
 
+use ILIAS\UI\Factory as UIFactory;
 use ILIAS\UI\Renderer as UIRenderer;
-use ILIAS\UI\Component\Symbol\Glyph\Factory as GlyphFactory;
 
 /**
  * @version		$Id$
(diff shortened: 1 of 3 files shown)


dsstrassner:
@kergomard will look into it and approve ---> merge.

kergomard:
Thank you very much @matheuszych for this PR!

First:
- Could you please rebase this on the current trunk and see that everything works? I tried a quick rebase and was unable to make it work.

Then I would just have one very small requests, to keep consistency:
Could you please remove the `glyph_factory`, the intermediate `$ui`-variable (and  if you would be so kind also the `$lng` intermediate variable) and move to array-access for the `$DIC` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7971/files#diff-0e8f9097de4106b1b8b53cd80b674e7318950a98d1c12a93764ff24bd575dc42L47-R54).

Thank you very much  again and best,
@kergomard

matheuszych:
Hey @kergomard,

I addressed the issues you mentioned and i believe this PR can be re-reviewed.

Kind regards,
@matheuszych

kergomard:
Thanks again @matheuszych !
I'm now fine with what the code looks like ...but not really with what it does ;-). It now shows the modal when clicking on "Add Answers", ...but I was unable to save anything: When clicking "save" in the modal it closes and nothing changes.

Best,
@kergomard

matheuszych:
Hey @kergomard,

From what I understand the "Save" button inside the modal is only supposed to save the gap options of that single gap, which is currently the case, but not the whole question. To save the whole question, you still need to use the outer "Save" or "Save and Return" buttons. The same behaviour can be observed in ILIAS 9. Correct me if i am wrong thou.

Update: I believe I found what you ment. For some reason you need to double open the modal to apply the changes. I will look into this.

Kind regards,
@matheuszych

matheuszych:
@kergomard It should now work as expected.

kergomard:
Thanks @matheuszych ! Now it does, what it should do. Could you please fix the complaints of the CSS-Fixer and then I will merge.
Best,
@kergomard

matheuszych:
@kergomard All done 😃

kergomard:

PR #7972, opened 2024-08-21
UI: Render Actions in Reporting Panel

kergomard:
Dear ui-coordinators

This PR adds rendering for actions to the reporting panel. Up to now, you were able to add actions, but they were not rendered.

I label this as an "improvement" and not a "bugfix" as nobody complained for older versions.

Thanks and best,
@kergomard


Amstutz:
Thx a lot. Ill merge this change for trunk. If needed in former versions, we could discuss picking it down. However, Ill stick with trunk to avoid any issues in productive releases since we have no ticket here.

PR #7973, opened 2024-08-21
UI: Fix Typo in StandardPanel Rules

kergomard:
Dear ui-coordinators

Looking at the examples and usages, this looks like a small typo in the rules of the standard panels. Maybe we also want to allow both, I'm not sure though.

Best,
@kergomard

### DIFF
--- a/components/ILIAS/UI/src/Component/Panel/Factory.php
+++ b/components/ILIAS/UI/src/Component/Panel/Factory.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 namespace ILIAS\UI\Component\Panel;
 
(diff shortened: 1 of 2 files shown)


klees:
Hi @kergomard,

I don't think this is a typo. AFAIK there indeed are panels that use Section View Control. And all View Controls indeed change the presentation of data =)

Is this a change that you require?

Kind regards!

kergomard:
Hi @klees 
No, I don't. There just is a slight inconsistency that I thought we should resolve:
- The example called "With all view controls" for the standard panel contains a mode view control, but no section view control, which should not be the case following its composition rules. I didn't go through the usages, but my feeling was, that the example corresponds to what I've seen and that the composition rules are probably wrong/imprecise. I just thought we should make sure, the two correspond.

Thanks for the fast reply and best,
@kergomard

thibsy (changes requested):
Hi @kergomard

Thx a lot for the feedback and bearing with us. Highly appreciated. You are right. However, we feel that we best heavily reduce the composition rules here, since they are already given through the public interface. 

Therefore, please change or indicate why not:

- [x] Reduce Composition Rules: How about just leave one rule; " ... MAY contain one or more view control."

Note, this would already be set through the public interface, so this change would not needed to be accepted by the JF.

Kind regards,
@Amstutz, @klees and @thibsy

kergomard:
Hi @thibsy , @klees , and @Amstutz 
Thank you very much for the feedback! I was not in a hurry, so perfectly fine with me ;-) .

I applied your changes (sticking to the Upper Case spelling used in the rest of the document), but then also renamed the corresponding example from "With all view controls" to "With view controls". This will prevent us from needing to update it on any change of the View Controls. Clearly, I can role it back, if you do not agree.

Best,
@kergomard

thibsy (approved):
Thx @kergomard for implementing the requested change.

I agree with you on renaming the example accordingly. Thx!
This LGTM now, so I'll merge right away.

Kind regards,
@thibsy

PR #7974, opened 2024-08-21
10/UI/allow view controls in reporting panels

kergomard:
Dear coordinators

This PR would introduce the possibility to add view controls to reporting panels. I built it on top of #7972 as this gave the opportunity to remove some code duplication, but please let me know, if you would prefere me to not do that and simply add the duplicate code.

I added the same constraints on view controls as we have in the standard panels and I think there are cases where all three types could be used reasonably. I don't think there is a justification for also adding the section view contro, though.

Thank you very much in advance and best,
@kergomard 

PS: The failing tests are from the copyright checker, I will fix them in one go.

### DIFF
--- a/components/ILIAS/UI/src/Component/Panel/Factory.php
+++ b/components/ILIAS/UI/src/Component/Panel/Factory.php
@@ -109,9 +109,7 @@ interface Factory
      *         Report Panels SHOULD be used when user generated content of two sources (i.e results, guidelines in a template)
      *         is to be displayed alongside each other.
      *   composition:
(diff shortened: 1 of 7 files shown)


klees:
Hi @kergomard,

please present this on the JF.

Thanks!

matthiaskunkel:
Jour Fixe, 30 SEP 2024: We highly appreciate this suggestion and accept the PR for ILIAS 10/trunk.

Amstutz:
Thx a lot @kergomard

PR #7977, opened 2024-08-21
Deletion of the legacy overlay in TestRandomQuestionSetPoolDefinitionFormGUI

aaronbidzan:
These changes are part of the [Legacy-UI refactoring](https://docu.ilias.de/go/wiki/wpage_7320_1357).
This PR deletes the unused legacy overlay. After testing, no differences in functionality could be identified.


kergomard:
Ok, this cost me some brain-power to understand what `ilOverlays` does. As far as I can see we really do not need this anymore.

Best,
@kergomard

PR #7978, opened 2024-08-21
MyStaff: Check feature activation before expensive database operations

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=31352

If approverd, this MUST be picked to `trunk` and SHOULD be picked to `release_8` as well.

### DIFF
--- a/Services/MyStaff/classes/class.ilMyStaffAccess.php
+++ b/Services/MyStaff/classes/class.ilMyStaffAccess.php
@@ -56,6 +56,10 @@ class ilMyStaffAccess extends ilObjectAccess
         if (self::$instance === null) {
             self::$instance = new self();
 
+            if (!self::isMyStaffActive()) {
+                return self::$instance;
+            }
+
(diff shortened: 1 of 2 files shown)


Services/MyStaff/classes/class.ilMyStaffAccess.php
```diff
@@ -75,11 +79,18 @@ public function __construct()
     {
     }
 
+    public static function isMyStaffActive(): bool
```
mjansenDatabay, line 82: I know ... `static` methods are bad, feel free to push changes to the PR-branch.

mjansenDatabay:
Hi @tfamula ,

any objections here? These fixes will really help a lot of ILIAS installations concerning database-related performance problems. This affects every request, all these `DROP TABLE` statements are unnecessary if `MyStaff` is disabled.

Best regards,
Michael

tfamula:
Hi @mjansenDatabay!
Sorry for the long delay here and thank you very much for this PR! Nice to see that such a big improvement can be done with a relatively small change (at least when Staff is not in use). I picked this also to trunk and release 8.
And as regards the static method: So be it.

PR #7981, opened 2024-08-21
UI: Expose Toggle Signals on Presentation Table

kergomard:
Dear ui-coordinators

Ok, this is going to be the last PR for the moment. To implement the feature request [Revise of Test Tabs](https://docu.ilias.de/go/wiki/wpage_8317_1357), I would like to propose a way to expand and collapse rows in multiple presentation tables. I would have liked it to be a simple handing out of a signal, that could then be triggered, but it aint that easy. My first proposal would be to hand out a signal for expanding and one for collapsing with corresponding options. Right now, this means that the consuming component needs to trigger the signal itself and thus needs to use jQuery to do so (and will most probably need to be migrated soon, when we get rid of jQuery for this). Alternatively, we could hand out the js as a string, that then just needs to be incorporated in a js-function running it. But I didn't find any precedence for this procedure and thus didn't go with it. This could be implemented in the presentation, but also in `Signal`. The advantage of this approach would be, that it would keep the consumer further away from any implementation details. I'm looking forward to your opinions.

There actually isn't much code here. Most of it is just a new example. I would write a couple of small tests if and as soon as we find a way to proceed with this.

Thanks for all your work and best,
@kergomard


thibsy (changes requested):
Hi @kergomard,

Generally speaking, we do not advise to use multiple tables on the same page, especially since they ship with view-controls and other features out of the box and crowd the page really fast.

Nonetheless, we acknowledge that you are in a situation where you cannot handle this use-case with a semantically strong UI component yet, which is something we should work on on in the future.

Therefore, please implement the following changes:

- [x] `getExpandAllSignal()` and `getCollapseAllSignal()`: please remove these methods from the interface again. You can leave the implementation as is, we simply want to keep this an internal functionality, so our (other) consumers are not encouraged to use it.
- [x] Example: please remove the example again, because we do not want to encourage consumers to use more than one table on the same page. (thx for providing one though!)

Kind regards,
@Amstutz, @klees and @thibsy

kergomard:
Thank you very much @thibsy , @Amstutz , and @klees !

You are completely right. If you want I can mark these changes as a stopgap measure and I can label the corresponding functions @deprecated with ILIAS 11? I asked for a slot at the next SIG-Meeting and will find a way to clarify and improve the presentation with ILIAS 11 so that we do not need this anymore.

Thanks again and best,
@kergomard

thibsy (approved):
Hi @kergomard,

Thx for implementing the requested changes. This LGTM now, so I will merge right away.

Since this is an internal functionality which is not exposed on the public interface, there is no need to add deprecations.

Kind regards,
@Amstutz, @klees and @thibsy

PR #7985, opened 2024-08-22
Object: Remove Usages of initCreationForms

kergomard:
Hi all

When changing the file creation dialogs, somewhere in the process the removal of `ilObjectGUI::initCrtaionForms()` got lost. Thanks to @mjansenDatabay for pointing this out to me. I now removed the code and it mostly just is a removal as the function was used to remove one or the other options that where there previously ('create', 'import', or 'clone' an object). Now there are two special cases though:
- `ilObjBibliographicGUI` added a file object here. I just moved the corresponding code to `initCreateForm` and copied the `saveObject()`-function for the legacy form here, so that it would still save. With these changes (and requiring the missing LibRIS-library) things work again. Please have a look at the changes @alex40724 .
- The second problem child is `ilObjLTIConsumerGUI`. There additional forms are added. I made room for that, but will deprecate support for these "hacks" with ILIAS 11. @Uwe-Kohnle this should be moved to the new forms and a switchable group. I could not really test this, as things go sideways for independent reasons. Please have a look at these changes @Uwe-Kohnle .

Then corresponding issue is here: https://mantis.ilias.de/view.php?id=41943

Thanks and best,
@kergomard

### DIFF
--- a/components/ILIAS/Bibliographic/classes/class.ilObjBibliographicGUI.php
+++ b/components/ILIAS/Bibliographic/classes/class.ilObjBibliographicGUI.php
@@ -101,7 +101,7 @@ class ilObjBibliographicGUI extends ilObject2GUI implements ilDesktopItemHandlin
         $DIC->language()->loadLanguageModule('obj');
         $DIC->language()->loadLanguageModule('cntr');
 
-        if (is_object($this->object)) {
+        if ($this->object instanceof ilObjBibliographic) {
             /** @var ilObjBibliographic $obj */
             $obj = $this->object;
             $this->facade = new ilBiblFactoryFacade($obj);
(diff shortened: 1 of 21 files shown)


chfsx (approved):
Changes in Bibliographic are fine, thanks a lot @kergomard!

kergomard:
Sorry @alex40724 and @chfsx : I just had in mind that `Bibliographic` was in your domain @alex40724 ! I changed the review and the text above.

@chfsx : I would personally suggest to actually move to KS forms for the creation, but this would have ment implementing the upload handler ...work I avoided ;-). ...for  you this should be a very quick change, so I can also simply remove my changes in `Bibliography` if you prefer.

Thanks and thanks @dsstrassner for letting me know of my mistake!

chfsx:
@kergomard yes we will move to KS forms in Bibl. as well, but I take your changes for the moment and will start refactoring after the whole object creation process in trunk works again. at this point I will replace the missing LibRIS library as well

kergomard:
Just to leave it somewhere. Object creation as well as many other modals work again with the changes in: #7788

kergomard:
Hi @Uwe-Kohnle 

Could you please have a look at this? I would like to merge until the middle of next week.

Thanks and best,
@kergomard

kergomard:
Hi @Uwe-Kohnle 

I had to change the implementation for the LTI Consumer once more as my previous idea had a flaw ;-). You are now the only one using Accordions here, ...so it would definitely be a good idea to move to the solution I suggested above. I tested this and things seem to work (I also added a trivial fixed for an `Iterator`), if I don't hear anything back from you until Wednesday October 2nd, I will merge this.

Best,
@kergomard

PR #7988, opened 2024-08-26
Course: Fix default value in LO course

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41738


smeyer-ilias:
Thank for analyzing and fixing this issue.

PR #7989, opened 2024-08-26
GS: Add workaround in access checks if user id equals `0`

mjansenDatabay:

If approved, this MUST be picked to all maintained branches.


mjansenDatabay:
Maybe not required anymore if #7991  got merged, because `gs_content.php` sets `\ilContext::init(\ilContext::CONTEXT_WAC);`, which is an exception in `\ilInitialisation::blockedAuthentication`.

chfsx:
Thanks a lot, i merged locally and picked to relevant branches

PR #7990, opened 2024-08-26
Improvement for file data handling. Removed unused code.

playamongthestars:
Improvement for file data handling. Removed unused code.

### DIFF
--- a/Modules/Forum/classes/Files/class.ilFileDataForum.php
+++ b/Modules/Forum/classes/Files/class.ilFileDataForum.php
@@ -149,4 +149,13 @@ class ilFileDataForum implements ilFileDataForumInterface
         $this->setPosId($posting_id);
         $this->rc_implementation->importFileToCollection($path_to_file, $this->getCurrentPosting());
     }
+
+    public function getDraftId(): int
+    {
+        return 0;
+    }
+
+    public function setDraftId(int $draft_id): void
+    {
+    }
 }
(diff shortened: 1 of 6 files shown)


Modules/Forum/classes/Files/class.ilFileDataForum.php
```diff
@@ -149,4 +149,13 @@ public function importPath(string $path_to_file, int $posting_id): void
...
     }
+
+    public function getDraftId(): int
+    {
+        return 0;
+    }
+
+    public function setDraftId(int $draft_id): void
+    {
+    }
```
mjansenDatabay, line 160: Do we really need these methods?

There seems to be no interface (acting as a contract for consumer) which requires these two methods. You could create a new interface extending `ilFileDataForumInterface` and let the draft classes implement this new interface (with the two methods), if really needed.

Modules/Forum/interfaces/class.ilFileDataForumInterface.php
```diff
@@ -29,6 +29,10 @@ public function getPosId(): int;
 
     public function setPosId(int $posting_id): void;
 
+    public function getDraftId(): int;
+
+    public function setDraftId(int $draft_id): void;
```
mjansenDatabay, line 34: IMO these two methods (if we still really need them) don't belong to the interface, but maybe to an extended one.

mjansenDatabay (changes requested):
Thx @nmatuschek 

I added some inline comments, but also added changes with https://github.com/ILIAS-eLearning/ILIAS/commit/cadb06d83b5e7f5b200eb773a67d8ba49097ed5b

IMO this PR is obsolete with the latest changes, can you confirrm this?

PR #7991, opened 2024-08-26
Auth: Init global user for "blocked authentication" state (fix for `$DIC->user()->getId() === 0`)

mjansenDatabay:

If approved, this MUST be picked to all maintained branches.

This will init/set the global user object in case `\ilInitialisation::blockedAuthentication` returns `true`, so the user id of a non-logged-in user will be `13` and not `0` for these cases.

### DIFF
--- a/Services/Init/classes/class.ilInitialisation.php
+++ b/Services/Init/classes/class.ilInitialisation.php
@@ -1407,7 +1407,7 @@ class ilInitialisation
         }
 
         if (
-            !$DIC['ilAuthSession']->isAuthenticated() or
+            !$DIC['ilAuthSession']->isAuthenticated() ||
             $DIC['ilAuthSession']->isExpired()
         ) {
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
I merged the sister PR for ILIAS >= 10.x .

-- UPDATE --

And reverted this (reasons can be found in the issue).

Services/Init/classes/class.ilInitialisation.php
```diff
@@ -1417,6 +1417,7 @@ public static function resumeUserSession(): void
             $current_script = substr(strrchr($_SERVER["PHP_SELF"], "/"), 1);
             if (self::blockedAuthentication($current_script)) {
                 ilLoggerFactory::getLogger('init')->debug('Authentication is started in current script.');
+                self::initUserAccount();
                 // nothing todo: authentication is done in current script
```
thojou, line 1421: Just a few cents. As we are initializing an anonymous user know, we may want to remove the comment about "nothing todo"

PR #7992, opened 2024-08-26
Auth/RepoTree: Add "blocked auth." exception

mjansenDatabay:

This PR adds an exception within `\ilInitialisation::blockedAuthentication` for the context of the asynchronous repository tree explorer rendering.

`\ilInitialisation::blockedAuthentication` is from hell, IMO the different command handling endpoints need a way to communicate such exceptions to the "Init" component (maybe the mechanisms of the component revision will do the trick in future).

If approved, this has to be merged to all maintained branches.

### DIFF
--- a/Services/Init/classes/class.ilInitialisation.php
+++ b/Services/Init/classes/class.ilInitialisation.php
@@ -1759,6 +1759,14 @@ class ilInitialisation
             ilLoggerFactory::getLogger('auth')->debug('Blocked authentication for goto target: ' . $target);
             return true;
         }
+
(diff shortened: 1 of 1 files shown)


thojou (approved):
Hey,

Together with PR #7991, I tested this PR.

The problem mentioned in https://mantis.ilias.de/view.php?id=41601 appears to be fixed if both adjustments are made. LGTM.

As ILIAS8 has reached end of live, we may want to move this PR to ILIAS9 for further processing?

Best,
@thojou

PR #7993, opened 2024-08-26
CI/StyleToRepo: adjust workflow for 9 and (maybe) fail on git

nhaagen:
CI/Style-To-Repo no longer ran with release 9; also, the repo-token expired without anybody to notice.

https://mantis.ilias.de/view.php?id=41557


klees:
Picked to trunk.

PR #7997, opened 2024-08-27
remove data fragments when deleting an Exercise

nhaagen:
https://mantis.ilias.de/view.php?id=41968


alex40724:
Thanks, the fix looks differently. ILIAS 9 takes care of hopefully everything, most of the code is back ported to 8 as well. Still this needs more repo like code.

klees:
Hi @alex40724,

could you be so kind and hint to the actual change, e.g. a commit id? We need to solve this privacy issue in ILIAS 8 as well and would like to backport the fix.

Kind regards!

PR #7998, opened 2024-08-27
9/UI/setting split page bg color from main bg color

catenglaender:
# Issue

Page background color cannot be easily targeted when making a simple skin from an scss entrypoint in Customizing/global/skin.

# Change

Added page background as variable on the settings layer.


Amstutz:
thx

PR #7999, opened 2024-08-27
File/Finder: Fix return type of `key` iterator method

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41967


chfsx:

PR #8004, opened 2024-08-29
[FIX] incompatibly implementation of ArrayAccess

chfsx:
@klees I first thought about committing this directly as a trivial fix, but maybe there was a reason for this implementation? from my point of view this doesn't work, because on the one hand the ArrayAccess interface requires something else and also because null or false are not valid return types, right?

### DIFF
--- a/components/ILIAS/Component/src/Dependencies/NullDIC.php
+++ b/components/ILIAS/Component/src/Dependencies/NullDIC.php
@@ -30,12 +30,12 @@ class NullDIC implements \ArrayAccess
     {
     }
 
-    public function offsetGet($id): null
+    public function offsetGet($id): mixed
     {
         return null;
     }
 
-    public function offsetExists($id): false
+    public function offsetExists($id): bool
     {
         return false;
     }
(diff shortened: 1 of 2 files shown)


chfsx:
another location here: https://github.com/ILIAS-eLearning/ILIAS/pull/8004/commits/47a9aa5ddc82ba424fa19364d7ec7ff6e8f7f9b2

klees:
Hi @chfsx,

is there any indication that this does not work indeed? Or is this just flagged by your IDE?

Rationale here is: `null` is strictly contained in `mixed` (as `false` is in `bool`), so I can indeed specialize return types for an interface implementation from `mixed` to `null` (and from `bool` to `false`). Interface requires me to always return `mixed`, I promise to always return `null`, so I fulfil the requirement. More formally, this is known as [covariance and contravariance](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)).

I specialized the types to indicate that this is indeed a null implementation that never does anything.

Greetings!

chfsx:
It' break in a Fatal in my env, but I'm currently checking if it uses a wrong PHP version in my case (could be 8.1):

```
PHP Fatal error:  Null can not be used as a standalone type in /Users/fschmid/Development/ILIAS/Core/trunk/components/ILIAS/Component/src/Dependencies/NullDIC.php on line 33
```

klees:
AFAIR the `null` type has indeed been added with `8.2`.

chfsx:
yes it was that, sorry. I actually also assumed that this must work, so I was surprised why it broke off for me...

PR #8007, opened 2024-08-29
UI: Add standard icons for read and unread mail

fneumann:
In preparation for FR "Revise Content-Tab »Folder« in User Mail System" we need to add new icons for read and unread mails to be shown in the new UI table of mail folders. These icons are added here as new standard icons.


yvseiler:
Thanks for adding both icons, looking forward to see them in action.

mjansenDatabay (approved):
Thx @fneumann !

PR #8010, opened 2024-08-30
WIP: Vcard wip for #41987

Amstutz:
See: https://mantis.ilias.de/view.php?id=41987

Note, this is WIP, and more to help document the issue, not for merging. We introduced this as a quick fix for our installation, but this needs some more investigation on what is causing this ilctrl loop.

### DIFF
--- a/Services/User/classes/class.ilPublicUserProfileGUI.php
+++ b/Services/User/classes/class.ilPublicUserProfileGUI.php
@@ -152,17 +152,17 @@ class ilPublicUserProfileGUI implements ilCtrlBaseClassInterface
 
         $tpl->loadStandardTemplate();
 
+        if($next_class == "ilobjportfoliogui" && $cmd != "deliverVCard") {
+            $portfolio_id = $this->getProfilePortfolio();
+            if ($portfolio_id) {
+                $gui = new ilObjPortfolioGUI($portfolio_id); // #11876
+                $gui->setAdditional($this->getAdditional());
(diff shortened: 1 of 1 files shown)


Services/User/classes/class.ilPublicUserProfileGUI.php
```diff
@@ -152,17 +152,17 @@ public function executeCommand(): string
 
         $tpl->loadStandardTemplate();
 
+        if($next_class == "ilobjportfoliogui" && $cmd != "deliverVCard") {
```
mjansenDatabay:
```suggestion
        if ($next_class === strtolower(ilObjPortfolioGUI::class) && $cmd !== 'deliverVCard') {
```

I always suggest/recommend using the `class-string` instead. If the class is renamed in future, modern IDEs will find usages like this more reliable.

Amstutz: Fully agree. I do not think this is relevant in showing the issue here. this is not intended for merging. please see the related ticket. we need to find out, why this is even needed and fix the cause.

kergomard:
Thank you very much @Amstutz for pointing me in the right direction and providing the PR! This gave me the necessary context to look deeper into it.

I opted for a different solution, implemented with https://github.com/ILIAS-eLearning/ILIAS/commit/117c11c42e512ec84233888dcee93c233836e8f2 and picked to ILIAS 9 and trunk.

PR #8011, opened 2024-08-30
Init/PasswordAssistance: Refactor session handling

mjansenDatabay:
This PR improves the code to handle "Password Assistance" sessions by introducing a proper "Repository" and
some entity/value object classes, so we can finally get rid of the `inc.pwassist_session_handler.php` script.


mjansenDatabay:
I'll merge this into `trunk` for ILIAS 11 (only), so we do not interfere with the testing phase of ILIAS 10.

PR #8012, opened 2024-08-30
Test & Assessment: Enhance unit tests and expand coverage for improved code quality

matheuszych:
This PR sums up the collective work of @malaja11, @bidzanaaron, @thojou  and @matheuszych on expanding the unit test coverage of the `Test & Assessment` component. Our initiative was not motivated by any feature wiki request or mantis issue but rather by the will to improve the overall quality of the `Test & Assessment` code base and increase our knowledge base about ILIAS Core.

We not only added many new tests and test cases but also improved already existing ones. While doing that we also refactored some of source code that we came across.

To increase the efficiency and quality of our unit test development, we built some helper tools that helped us to accomplish just that.


matheuszych:
I will look into why the pipeline does fail here.

kergomard (commented):
Thank you very much for your effort @malaja11, @bidzanaaron , @thojou, and @malaja11 !

I would have the following suggestions:
- I would suggest to order `use`-statements by there source, so all in Test related uses in a first block, then if there are uses from QuestionPool in a second block, then a third block with all other uses from ILIAS, and finally uses from dependencies. If wou agree, I would suggest to start implementing when ever we touch the `use`-statements.
- I would also suggest to discuss a few things about test organisation at the next TechSquad (I'm sorry, I would have liked to do this earlier, but I kept kicking the can down the road). These are all simply observations and I'm very much interested in a discussion. I tried to provide examples, they are not to be nasty, they are just to get a discussion started:
    - The use of the `DIC` in tests. I know, I started using it more havily in the `setUp`-function, but now I think I was wrong to do it even there. Let us see if we can develop  some comon ground/guidelines. I was triggered by lines like [this one](https://github.com/ILIAS-eLearning/ILIAS/pull/8012/files#diff-0390f4bca70e8e516dfbb884300f7f650c9a5208e50269788e098b7f47016b76R44) where the `DIC` is used in a test-function.
    - The naming of test functions. I think we should be precise what the function actually test so `test_Getter`is not a good idea. I'm not even certain, what you try to test here? Is the test necessary in any form? Additionally you test two rather independent things in one test. Do we really want to test every fully typed setter and getter? I'm not sure we should. So, better naming would lead to clearer tests, in my eyes.
    - Personally I also started thinking that it might be a good idea to move tests to their own sub-namespace `ILIAS\Test\Tests` and then to use the same namespace structure we have in the Test itself.
    - And finally, and maybe most importantly: We should discuss what we want to test and how. My feeling while going through the tests is that many of the proposed tests here are not very helpful, e.g. why should we test that the `CertificateTestObjectHelper` returns `null` on any value [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8012/files#diff-0390f4bca70e8e516dfbb884300f7f650c9a5208e50269788e098b7f47016b76R44)? On the other hand I think that the tests for the `AdditionalInformationGenerator` add something, but then again you didn't test the most important function, were things can go wrong in a gazillion intresting ways (`parseForExport`). I  would suggest to have less tests, but better ones and remove everything  that is too trivial. Personally I think that for one all tests that only test construction should go, the only thing they do is create work until we have tweaked the `DIC`and the `TestDIC` to make them work. An example of a test I like the direction it goes into, but I don't think it gets there is `Logging\FactoryTest::test_buildQuestionAdministrationInteractionFromDBValues($type)`. First I think this should be two tests (`testBuildQuestionAdministrationInteractionFromDBValuesThrowsOnInvalid($type)` and `testBuildQuestionAdministrationInteractionFromDBValuesReturnsCorrectInteractionTypeOnValid($type)`. Then I would expand the test to include all values you set, not just that the return value is of a certain kind. But make sure the values set are really and absoutely positively of the type they come from the database? Is there a value in the db that is nullable? Check what the function does with null (I didn't check this). Another  example in my eyes is [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8012/files#diff-19491316ad7a30f3d70290386eb5fb313b683cedea3b80d2a3f9c087bf7f8f27R53). Why are we checking what the function does with a negativ number? Do we believe this is sensible that you get a negative number? As far as I unerstand it, we accept negative numbers because `int`, but we honestly don't care if everything goes sideways when the number is negativ.
    - As stated I would suggest to keep all conventions in the test folder: No camelCase-Variables, no snake_case-functions, ... . I think you do have multiple conventions in your tests.

- Please remove all `use`-statements for all clases in the global space (e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8012/files#diff-08aad04f801bbf06c06c4ded6d017bdc3d3e56fefae487075e50e1fb64054382R23) and many more (also in the tests-folder).
- Please don't add `@throws`-doc-blocks (e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8012/files#diff-08aad04f801bbf06c06c4ded6d017bdc3d3e56fefae487075e50e1fb64054382R74)).
- Please convert camelCase-variable names to snake_case, when you touch them (e.g. [here](https://docu.ilias.de/go/xmvc/14524)).
- Please don't turn [these type of statments around](https://github.com/ILIAS-eLearning/ILIAS/pull/8012/files#diff-7063aad69bd78d0160c541906de4f0bb89565022da50b8b24c1b90015efeb370R82) or [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8012/files#diff-90cdd897460f9c715c41db3985a58302f2ca0ee47bf812a45eaaeaeb178d18baR114). They are easier to read the other way around as the statement inside the  condition is shorter and it treats the special case first.
- Please don't remove line brakes like [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8012/files#diff-7063aad69bd78d0160c541906de4f0bb89565022da50b8b24c1b90015efeb370R119), the codes get harder to read. The change [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8012/files#diff-90cdd897460f9c715c41db3985a58302f2ca0ee47bf812a45eaaeaeb178d18baR36) from where I stand is ok though.
- Please do not replace `=== null` with `is_null()` it is far harder to read (see: https://github.com/ILIAS-eLearning/ILIAS/pull/8012/files#diff-90cdd897460f9c715c41db3985a58302f2ca0ee47bf812a45eaaeaeb178d18baR46).
- Please use `array<type>` in Doc-Blocks. I've no strong preference, but I think we are mostly moving in this direction, as this notation offers more possibilities to be clear about the content of the array, e.g. with `array<string, string>` (and thanks for typing them, much appreciated!).
- Please simply revert all changes you made in [`class.SettingsMainGUI`](https://github.com/ILIAS-eLearning/ILIAS/pull/8012/files#diff-aed1bb9e990db7e5e403940f4d75ce40b6f431ec1b79e1c4d2723401475e1dd2) (except, if you want to be kind, the two added `void` return types).
- Please remove the `CSVExportTraitTest`. The trait will be gone soon.
- pl includes for classes in the global space.

Ok, I will stop here. Again: I do really not claim to have the answers to the questions in the  first section, I'm just trying to organize my thoughts and to get things slowly sorted out. I spent a lot of time fixing tests that in the end just test that a constructor constructs in a complete irrealistic environment. I don't think they are worth the effort, on the other hand, good tests have regularly saved my ass.

Thanks again and  best,
@kergomard

mjansenDatabay:
It seems this PR modifies code outside the "T&A" component -> "Administration". Please revert these changes and focus on the "T&A" only.

thojou:
Hey @kergomard,

Thank you so much for your detailed feedback. We took the time to discuss it, and @matheuszych  has already started implementing some changes. We've realized that some of our adjustments were overly aggressive, and we’ll be reversing them accordingly.

> I would suggest ordering use statements by their source, grouping Test-related uses, then QuestionPool, followed by other ILIAS uses, and finally dependencies. If you agree, we could implement this when we touch the use statements.

We definitely agree that keeping use statements organized is important, but introducing custom code conventions such as grouping by context might create unnecessary overhead and be prone to errors, especially when auto-imports or IDE hotkeys handle these tasks. A more efficient approach could be adding the [ordered_imports](https://mlocati.github.io/php-cs-fixer-configurator/#version:3.64%7Cfixer:ordered_imports) rule to php-cs-fixer, ensuring this is managed automatically.

> The naming of test functions. I think we should be precise about what the function actually tests, so test_Getter isn't ideal...

Agreed. Improving test function naming for better clarity and ensuring each test has a clear purpose is a good practice. We’ll work on this going forward.

> Personally, I also started thinking it might be a good idea to move tests to their own sub-namespace...

We completely agree! Organizing tests into their own namespace like ILIAS\Test\Tests makes sense. But we think this should be done in a separate PR.

> Lastly, we should discuss what we want to test and how... many proposed tests here aren't very helpful.

You make a valid point. It’s crucial that test cases reflect real-world scenarios and provide meaningful validation. After reviewing the CertificateTestObjectHelper, we found that the class itself didn’t offer much value—it was essentially just a wrapper with no real functionality. For this reason, instead of writing more meaningful tests, we’ve decided to remove both the class and its tests. That said, we still believe smaller tests, like those for getters and setters, can be useful in preventing simple mistakes like typos.

**Regarding your specific requests:**

> Please remove all use statements for classes in the global space.

We believe that including global namespace classes in the use statements makes the dependency list more complete. We don’t have a strong stance on this, but we’d appreciate further insight if we're breaking any conventions here.

> Please don't add @throws doc-blocks.

Could you clarify why @throws tags are discouraged? They help annotate methods that might throw exceptions, making it easier to identify where custom error handling is needed, especially within an IDE.

Thank you again for your time and thoughtful feedback. We look forward to discussing this further at the next TechSquad meeting.

Best regards,
@thojou

kergomard:
Hi @thojou 

Thanks for taking this up!

> I would suggest ordering use statements by their source, grouping Test-related uses, then QuestionPool, followed by other ILIAS uses, and finally dependencies. If you agree, we could implement this when we touch the use statements.

>> We definitely agree that keeping use statements organized is important, but introducing custom code conventions such as grouping by context might create unnecessary overhead and be prone to errors, especially when auto-imports or IDE hotkeys handle these tasks. A more efficient approach could be adding the [ordered_imports](https://mlocati.github.io/php-cs-fixer-configurator/#version:3.64%7Cfixer:ordered_imports) rule to php-cs-fixer, ensuring this is managed automatically.

This creates overhead, yes, but it also forces us to keep our headers clean (the amount of unused `use`-statements I've seen just because we rely on IDEs to add them is a sign for me, that it is good to check them from time to time). I don't think that we can delegate this to any automation, as this is about semantics (not syntax) and automation (as the cs-fixer as I understand it) operate on syntax (it just orders them alphabetically or by length). Additionally the cs-fixer rule would be for everyone and I don't think we are going to agree on that for all components. So, I would be in favor to clean this up manually, whenever one looks at them and if they are not perfect it doesn't matter too much, it is just for a little bit more structure.

On the additional points:

>  Please remove all use statements for classes in the global space.

>> We believe that including global namespace classes in the use statements makes the dependency list more complete. We don’t have a strong stance on this, but we’d appreciate further insight if we're breaking any conventions here.

Two answers to this:
1. I think you are the only ones (one of very few) I've seen doing this in the ILIAS code, so this would be kind of a convention, but I think this is the lesser argument.
2. More importantly: Everything we do, as I see it, is about making the code as uncluttered/easy to read, semantically dense, and free of lies as we can. So I would ask: What information are we trying to transmit with the global includes: The only answer I can think of, is to have a list of all types referenced. I don't think this information has real value in this case. Additionally: If we do this, we would, to be complete, also need to include the classes in the same namespace, and I think we should definitely not do that. On the other hand we create a lot of noise and more work to keep things clean and we loose the backslash as a marker in the code that we are addressing a global type. This I would also not consider to be valuable information, though. So, from where I stand there is only one strong argument: We create clutter and work. This is why I would kindly ask you to remove them again and to not add them.

> Please don't add @throws doc-blocks.

>> Could you clarify why @throws tags are discouraged? They help annotate methods that might throw exceptions, making it easier to identify where custom error handling is needed, especially within an IDE.

This is the same argument as before: I don't think this information helps us, as most of the errors will be thrown outside of our component and we mostly don't intend to react to them. Exceptions in ILIAS are mostly about things going sideways in an unrecoverable way and we want them to just bubble. In the few cases they are not, as far as I can see, we would need to deal with them as early as possible and we should not let them bubble. So there is no interesting information there as far as I can see. Additionally the exceptions are mostly thrown outside of our code and thus they could also change without us noticing and we would need to invest quite some work to avoid introducing lies into our code. And finally they make the code more cluttered and less readable. All of this for me points very strongly at not including the statements.

I hope this is understandable otherwise I'm at your disposal.

Best,
@kergomard

dsstrassner:
This is on hold because of the other tasks before coding complete.

dsstrassner:
hi @kergomard & @thojou - is this PR still a thing now or only after the beta phase of ILIAS 10?

kergomard:
Hi all

I think this can be merged whenever, but first, from where I stand, it needs a little more love ;-) .

Best,
@kergomard

matheuszych:
Hi all,

this PR is currently not compatible anymore. Due to other priorities we decided to close this for now. We do not want to throw away these changes completly, but they need major rework. A new PR will be made when everything gets sorted out.

Best regards
@matheuszych

PR #8013, opened 2024-09-02
MediaObject: Make `WAC` checking instance more robust against indeter…

mjansenDatabay:
…minable usages

See: https://mantis.ilias.de/view.php?id=41981


alex40724:
@mjansenDatabay Just by looking at the code: This means, if the code finds any invalid usage, it will immediately return false and not show the media? However there might by other valid usages later in the iteration, which give a valid oid and an access of true. So imo we should just do

```
-->              if (!is_null($oid) && $this->checkAccessMobUsage($usage, $oid)) {
                        return true;
                    }
```
? So it would continue with the iteration and try the next usage.

alex40724 (changes requested):
See comment above.

mjansenDatabay:
> See comment above.

Hi @alex40724 , the code continues with the next iteration (it is a switch statement within a loop), it does not return `false`.

The `continue 2;` inside the switch statement ensures no other code in the current iteration will be executed, and the loop continues with the next iteration.

PR #8014, opened 2024-09-04
mantis #41510 plugin type is ignored, when no plugins installed

sz-tik:
three functionalities are affected:

when no plugins are installed
- ignore plugin field in existing table (with warning)
- for an already existing field of type plugin: choosing a plugin is disabled
- for an additional field the type 'plugin' is not offered

### DIFF
--- a/Modules/DataCollection/classes/Fields/class.ilDclFieldEditGUI.php
+++ b/Modules/DataCollection/classes/Fields/class.ilDclFieldEditGUI.php
@@ -266,6 +266,15 @@ class ilDclFieldEditGUI
             foreach (ilDclDatatype::getAllDatatype() as $datatype) {
                 $model = new ilDclBaseFieldModel();
                 $model->setDatatypeId($datatype->getId());
+                //check plugin field type for validity (no plugins - the type must not be displayed)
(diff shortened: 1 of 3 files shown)


iszmais:
@sz-tik Thank you a lot for this approach to the issue!

I appreciate the detail handling of edge cases but would rather prefer a more subtle approach to this, since most of the changes made here seem more fitting for a feature request than an actual bug fix.
Especially the implied warning will affect multiple scenarios which never came in contact with DC-Plugins and will therefore decrease the user experience for the great majority.
Further, this may cause some contradictions with the [related Feature Request](https://docu.ilias.de/go/wiki/wpage_8372_1357)

Therefore, I would suggest keeping the bugfix strictly tight to solve the origin problem: to remove the selection on creation without any message displayed. The case of updated fields is non-critical for ILIAS 8 and was already taken care of for ILIAS 9+ by not showing any other fields. And since "empty" Plugin fields don't serve any purpose, I wouldn’t cover those either since their deletion is inevitable (expect their plugin is restored, but in this case a warning would just be misleading).

I will provide a PR for this implementation within the refactoring for ILIAS 10 soon.

Greetings,
@iszmais

iszmais:
I provided a fix [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8062)

PR #8016, opened 2024-09-05
UI/Input/Container: add JS to retrieve values clientside

nhaagen:
This adds some JS to the UI Form to retrieve its values client side.
The JS builds an internal tree over the container's elements by their names rather than their position in the DOM.
Consumers may search for a FormNode and retrieve values from all elements below.
Optional or Switchable Groups will only relay values if selected.

Please note that this is a basic addition without any visible effect, yet, but it will be very handy in folding forms/groups/filters and visualize current selections of otherwise invisible inputs.

(was: https://github.com/ILIAS-eLearning/ILIAS/pull/7137)


components/ILIAS/UI/resources/js/Input/Container/src/container.js
```diff
@@ -0,0 +1,45 @@
...
+import LinkTransforms from './transforms/link.transform';
+import SelectTransforms from './transforms/select.transform';
+import MultiSelectTransforms from './transforms/multiselect.transform';
+
+import filter from './filter.main';
+import ContainerFactory from './container.factory';
+
+const transforms = {
+  'switchable-group-field-input': new SwitchableGroupTransforms(),
+  'optional-roup-field-input': new OptionalGroupTransforms(),
```
klees:
```suggestion
  'optional-group-field-input': new OptionalGroupTransforms(),
```

components/ILIAS/UI/resources/js/Input/Container/src/container.class.js
```diff
@@ -0,0 +1,221 @@
...
+    this.#container = container;
+    this.#nodes = new FormNode('form', 'FormContainerInput', container.getAttribute('id'));
+
+    Array.from(container.querySelectorAll(SEARCH_FIELD))
+      .filter((domFieldNode) => domFieldNode.parentNode === domFieldNode.closest('form'))
+      .forEach((domFieldNode) => this.#register(this.#nodes, domFieldNode));
+  }
+
+  /**
+   * @param {FormNode} pointer
```
klees:
```suggestion
   * @param {FormNode} current
```

klees:
Hi @nhaagen,

really looking forward to have this!

Could you please:
* Rebase this (we have removed `novalidate` for forms in the meantime...)
* Join me for me a meeting to talk this over, after you have looked into this again?

Thanks!

klees:
Hi @Amstutz and @thibsy,

please have a look into this. This is a precondition to iterate on filters.

Thanks!

thibsy (commented):
Hi @nhaagen,

Before I do a full technical review, I wanted to discuss some architectural changes first, because they may affect the current implementation greatly.

What you propose is one generic `FormNode`, which will be used to represent any kind input field (to rule them all!). I believe this structure does not work well together with dedicated implementations of input fields yet, which would be the case for the `Markdown` or `Textarea` field at the moment.

I think we should aim to achieve something where both of these worlds are fused together. Having a generic `FormNode` is definitely a great fallback for simple inputs, like numeric or text, but I think it should not be used for all. The current proposal also couples the implementation details of certain fields, like the multi-select or radio, to the corresponding value transformation, in a not so obvious way.

My suggestion is to create a base `FormNode` (with a name like `Field`), from which dedicated input fields like the ones above can extend. This way some complex value transformations can be implemented directly on a specific instance. With your proposal, some of the input fields need to maintain their client-side functionality implicitly anyways, so I suggest to implement a client-side component for all of them. This way we can also initialise them visibly in the rendering process. However, doing so would require a way to lookup input instances after their initialisation, because the container - due to the recursive nature of our rendering process - is rendered and initialised after the input fields are. Some registry would be necessary, not necessarily public though.

If you prefer it, we could also talk about this on Discord in a meeting. Feel free to contact me.

Kind regards,
@thibsy

thibsy:
@nhaagen I pass this back to you for the time being.

klees:
Hello @thibsy,

as the current architecture mostly evolved under my guidance, I think it is on me to address your concerns.

So, the general goal for this implementation, not very carefully worded, is something like this: There is a tree of fields that constitutes the form. These fields contain values that we want to look into, also forming a tree. Note that these two trees are not the same, per se. There might be less nodes in the tree of values than there are in the tree of fields, as e.g. an optional group can be unchecked. Also note that these values represent the input of the user, not the thing we later process as "content" of the form on the server.

For every node in that tree of values we want to be able to get some minimized representation of the value that is currently contained in that node. This is kind of simple for e.g. a datetime field, but we shall also be able to deal with longer input fields, such as textareas or nested inputs like a duration field or a general group. Also, duration is a case that is created from two fields in a group, but for the sake of a minimized representation of the contained value, we shall better treat it as one value (because we can then say "04. - 08.06.2025" instead of "start: "04.05.2025, end: 08.06.2025" or something like that). Since we want to be able to "expand" or "collapse" single fields to recreate the current filter behaviour or build new similar behaviour for complex inputs, such as the tree selector, we need to be able to dig into these values as well. So there is a notion of "children", but also these are not the contained fields per se, e.g. for the unchecked optional group.

Hence, we roughly have these requirements:

* There is a nested structure of fields in the form that implies some nested structure of input values in that form.
* On every level of that nested structure, we want to be able to derive a minimized representation of the value.
* There are fields with values where a minimal representation can be naively derived from the actual value the user has put into the field. There also are fields where this is not possible.
* There are fields that have values we can dig into and there are fields where we cannot do so.
* There are fields that are build by combining existing fields. Yet, the minimized representation may or may not be a straight forward combination of the minimized representation of the constituents.
* There is an extension point: We want to be able to add new types of fields, simple or nested, in the core or via plugins and the like.

If I understand correctly, you would propose to implement these requirements by a kind of straight forward representation of the fields in some object graph, where every field gets a class that matches its type. The classes would then use some inheritance pattern to distribute similar functionality to the different field types.

We, on the other hand, decided to implement the requirements by using the strategy pattern, where only minimal parts of behaviour are encapsulated in exchangeable parts. Similar functionality is distributed via one class and can only be modified in strictly defined locations.

We also thought about that class/inheritance based approach some time, but in the end decided against it (as you can see =) ). As often in engineering, this is no strict "right or wrong" decision, but instead governed by pros and cons and tradeoffs. So lets try to list these for both solutions:

#### Class/Inheritance based Approach

* **(+)** more obvious and common
* **(+)** straight forward mapping from code to "things"
* **(-)** very open for modification (in JS...)
* **(-)** inheritance is very weird in JS (might just be me...)
* **(-)** uses the "is-a" relationship of inheritance to distribute code, its not very clear if one taxonomy will actually be enough to faithfully represent "fields" while distributing functionality correctly (might also just be me...)

#### Strategy Pattern based Approach

* **(-)** needs to be explained (as witnessed here...)
* **(+)** narrow extension point for outsiders
* **(-)** might be unintuitive how to add new types
* **(+)** "algorithm" is in one place
* **(+)** favor composition over inheritance

So after all, we feel that not using classes and inheritance will turn out to be easier to understand and modify in the long run, especially once code from others comes into play. I feel that the hard to understand structure of the inputs that we have on the server side, which definitely partly results from inheritance, is a warning sign to use the same approach to represent fields on the client side.

As a final remark: We also think that we probably would not want to present a class-based interface to the form to consumers. Instead we think that there should be a single point, which shall be the container, that provides an imperative interface to the contained thingies. We are not quite there yet, though, but the general idea is, that many interactions with some part of a form will require coordinated behaviour of other parts in the form. E.g., when expanding one part of the form, collapsing the other will be required. That behaviour seems to arise on the level of the container, which also seems to be the main carrier of the state. This would be somehow similar to the main bar. From that POV, both above approaches would rather be internal implementation details.

I hope this makes some sense. We can still schedule a meeting ofc, I would appreciate some indication if you think so.

Kind regards!

thibsy (commented):
Hi @klees and @nhaagen,

Thx a lot for your explanation and the discussion we had on Discord about this topic. I also apologise for the delay here - I partially blame the Swiss Military.

I had a hard time following this PR, which also made it difficult to review. This was until I realised I was focusing too much on making the strategy pattern work.

After some time I noticed that each strategy will only ever work with exactly one type of form node (and its hierarchy), or at least they are designed to. It would e.g. work to apply the duration strategy to the link field, but it would not make much sense. There are also strategies like for optional-groups, which would likely even break if applied to e.g. a password field without children.

This means we essentially have a one-to-one relationship between a strategy and a field type, and thus loose some benefits of using the strategy pattern in the first place. For instance, one problem we could have solved using this pattern is the retrieval of formatted vs raw values (for which you introduced dedicated functions). Benefitting from this pattern, one would simply switch the algorithm used to generate these values **at runtime**. However, this is not possible at the moment because one needs knowledge about the internals of your code and what strategy can be applied to which field types.

Embracing the strategy pattern would result in creating groups of strategies, where each group is bound to a set of field types. Only then strategies of the same group become interchangeable with one another and the benefits of this pattern could be used. This would also result in additional complexity, because we'd have to make sure each strategy is only ever applied to the right context.

Whether we want to continue down this road or not depends on the answer to the following questions:

- [ ] Do we expect strategies to be interchangeable at runtime, or are they only ever configured during construction?
- [ ] Are there ever going to be multiple strategies for the same (set of) field types?

If the answer to both of these questions isn't yes, we may not gain much from this abstraction. In that case a different architecture will probably result in a more maintainable solution. The [visitor pattern](https://refactoring.guru/design-patterns/visitor-double-dispatch) would come to mind (fields visited by concrete field-type visitors).

Please note I appreciate the architectural thinking that went into this very much. And if we really do anticipate growth in complexity in terms of strategies, an improved strategy pattern could still be a good foundation.

Kind regards,
@thibsy

klees:
Hi @thibsy,

thanks for the reply, even if it is late =) Much appreciated as well...

What you say about the strategies is correct: They will only work with one type of field (mostly...). They cannot simply be exchanged for one another. Also, I currently don't expect strategies to be exchanged on runtime. As you wrote, they will only be assigned on construction. I can't see why that would ever change.

Also, I currently don't strongly expect there to be multiple strategies per field type. This might be a little more interesting, though, as the shortened representations are somehow related to the locations where we wan to display them. So we might end up with the requirement to show some different shortened value in the filter or in a minimized field in a form. 

But also, these are not the reasons why we chose to implement the stuff as is. This  also is why we did not call that approach "strategy pattern", because the design goals are not that of a strategy pattern. The core reason, instead, is: We wanted to minimize the possible extension point, for us and others, in the face of a class/object-system that is highly flexible. If we, say, would simply use a base class `FieldValue` and then fill in some blanks for the different behaviour (that we currently put in the "strategies"), we would also open up all other behaviour for modification. Which would be behaviour the overall algorithm would intimately rely on. We have no programmatic way to say that "this behaviour should be modified" and "this behaviour shouldn't be modified" (at least to my knowledge...). Which would be not so bad for internal usage (as we would review, can write according documentation) etc., but will introduce a point where someone will cause havoc at some point (either we, after we forgot this discussion etc. or someone else in a plugin or...). It might be overcautious to introduce complexity to prevent these only potential fuckups, though.

I think we (as in @Amstutz/you/me) should make a call and the proceed accordingly. I will not die on the cautious hill, I'm not in the military... =)

Kind regards!

PR #5971, opened 2023-06-07
Component: Docu of Directories and Namespaceing

klees:
Hello everyone,

this is a first concrete proposal to implement [the goals from the Component Revision 2022](https://docu.ilias.de/goto_docu_wiki_wpage_7295_1357.html): A new structure for directories and introduction of Namespacing in general ILIAS. This contains the results of our workshop of the Dev-Conf in Bremen.

I hope you like it. I still expect that there will be concerns and question. I would kindly ask you to add them to this PR. I will try to answer them here and adjust the paper accordingly. If it feels to much or if someone explicitly asks, I will invite to a workshop to discuss this proposal.

If this is accepted there will be a follow up regarding the integration of the components. I already have proposals regarding that topic, but I do not want to discuss too much at the same time and I also think that we could very well just make the change proposed here without moving any further. If someone wants a preview: Feel free to ask, I'm happy for sparring partners to discuss my current ideas.

Finally, I added a short section regarding how I propose to implement this. I totally expect that this will not missing some details or that some maintainers want to be involved more at some point. This will become even more complicated once we discuss integration of components. If you have input regarding the roadmap already I'll be happy to hear it. I will invite maintainers to discuss this in more detail once we have moved forward with the discussion of the integration strategies.

Best wishes!

PR #8018, opened 2024-09-06
Plugins: for 10, move plugins to public/Customizing/

nhaagen:
This moves the plugin directory to /public/Customizing/plugins; 
'Services' and 'Modules' are removed from the path.

so, e.g., the TestPageComponent will now reside in 
`/public/Customizing/plugins/COPage/PageComponent/TestPageComponent/`

- [ ] docs still have to be updated

### DIFF
--- a/components/ILIAS/Component/classes/Setup/class.ilComponentBuildPluginInfoObjective.php
+++ b/components/ILIAS/Component/classes/Setup/class.ilComponentBuildPluginInfoObjective.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
--- a/components/ILIAS/Component/classes/Setup/class.ilComponentBuildPluginInfoObjective.php
+++ b/components/ILIAS/Component/classes/Setup/class.ilComponentBuildPluginInfoObjective.php
@@ -19,7 +19,7 @@ use ILIAS\Setup;
 
 class ilComponentBuildPluginInfoObjective extends Setup\Artifact\BuildArtifactObjective
 {
-    protected const BASE_PATH = "./Customizing/global/plugins/";
+    protected const BASE_PATH = "./public/Customizing/plugins/";
     protected const PLUGIN_PHP = "plugin.php";
     protected const PLUGIN_CLASS_FILE = "classes/class.il%sPlugin.php";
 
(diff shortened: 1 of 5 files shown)


klees:
Dear colleagues,

this fixes some oversight of mine in the initial version of [`docs/development/components-and-directories.md`](https://github.com/ILIAS-eLearning/ILIAS/pull/5971). I did not realize then, that we should move the `Customizing` folder to the `public` folder for various reasons:

* It will contain skins that will contain assets that need to be delivered by the webserver.
* Although ILIAS does not provide a mechanism for endpoints in plugins, there will sure be plugins that contain endpoints that need to be open for webaccess.
* Just like skins, plugins might also contain assets that need to be delivered by the webserver.

In principle we could solve all this via the [same mechanisms for the public folder](https://github.com/ILIAS-eLearning/ILIAS/pull/7019) that we also use for the ILIAS components now. But on the one hand, this will require work for the skins (most probably: turn them into components...) that we didn't do until now. On the other hand, we will abandon old style plugin with ILIAS 11 anyway and turn them into ~just normal components from other vendors~, at least according to our current plans. Why should we require a preliminary change for old-style plugins if we move on with ILIAS 11 anyway...

So we propose to make this change for skins and old style plugins so they can mostly continue to work as always for ILIAS 10.

Kind regards!

thojou:
Hey @klees,

Thank you for the detailed explanation in this PR. After going through it, I have a few thoughts that I’d like to contribute to the discussion.

From my understanding, as you mentioned, plugins and skins are already largely handled as components, and while I recognize that skins would require more extensive refactoring to fully align with the new component structure, I wonder if moving all plugins (in addition to skins) to the public directory is the best approach.

I can see the necessity of moving skins to the public/Customizing directory for now, given the complexities around their refactoring. However, applying the same change to plugins raises some concerns for me. My primary concern is that this could compromise the principle of maintaining a clean, encapsulated document root, preventing unnecessary code from being exposed. By moving plugins to the public directory, we may inadvertently leak parts of the codebase that shouldn’t be publicly accessible, and additional .htaccess rules or other protections would be needed to prevent access to non-entrypoint or non-asset files in these plugins.

Would it make sense to limit this change to skins, given the immediate needs, and continue handling plugins under the existing mechanisms? This could avoid potential risks while still addressing the issue with skins until we can undertake the larger refactoring effort in the future.

Looking forward to your thoughts on this!

Kind regards,
@thojou

matthiaskunkel:
Jour Fixe, 16 SEP 2024: We follow Richard's suggestion and postpone the change of the directory where plugins are placed to ILIAS 11. Plugins in ILIAS 10 have still be placed in the /Customizing/global directory. PR accepted for ILIAS 10.

klees:
Hi @thojou,

thanks for the feedback and questions.

Let me first clarify one thing regarding the componification of plugins and skins: It is indeed true that I could put some vendor component (or plugin, as we call them now...) in, e.g. `component/CaT/CustomComponent` and the system would pick them up just like the components in `component/ILIAS`. But there are only little hooks that currently are available via the new component mechanism, so we currently can not use the new component mechanism for just the same things that old style plugins can provide. So for ILIAS 10 we can not simply fully replace all current plugins with new components.

I see your concern regarding security, but I think this change will just make the current situation (plugins are fully exposed if no other measures are taken) last one release longer. On the other hand I simply cannot see that we really have another option here. 


Hope this makes some sense =)

Kind regards!

PR #8019, opened 2024-09-09
HTTP: Fix sending `Set-Cookie` HTTP header

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41226


chfsx:
Thanks a lot, I picked this to 9 and trunk as well!

PR #8022, opened 2024-09-09
Tree: Create Valid Path for Administration

kergomard:
Hi @fwolf-ilias 

This PR fixes an issue when generating a path through `ilPathGUI`. It introduces a special-casing for nodes in Administration.

See: https://mantis.ilias.de/view.php?id=41897

Best,
@kergomard


fwolf-ilias:
Thank you @kergomard for providing a fix!

PR #8027, opened 2024-09-10
Composer: Upgrade `phpspreadsheet` and remove patch

mjansenDatabay:
This upgrades `phpspreadsheet` and removes the "Composer Path" applied a few weeks ago.


kergomard:
Thank you very much for this PR @mjansenDatabay and for getting rid of the patch! This is ILIAS 8 only so I just merge.

PR #8039, opened 2024-09-10
Data/LanguageTag/UnitTests: Fix test provider

mjansenDatabay:
This PR fixes the following error with PHP 8.3:

```
There was 1 PHPUnit error:

1) ILIAS\Tests\Data\LanguageTagTest::testRisky
The data provider specified for ILIAS\Tests\Data\LanguageTagTest::testRisky is invalid
Using $this when not in object context
```


klees:
Test fails because of wrong copyright. Will fix in separate commit.

PR #8041, opened 2024-09-10
Unit Tests: Show details for skipped tests

mjansenDatabay:
This PR changes the "PHPUnit" configuration to display details about skipped tests. Having details about skipped tests is useful to determine the file/test class of the skipped tests (e.g. for CI reports at the ILIAS JourFixe).

### DIFF
--- a/scripts/PHPUnit/phpunit.xml
+++ b/scripts/PHPUnit/phpunit.xml
@@ -13,6 +13,7 @@
          displayDetailsOnTestsThatTriggerWarnings="true"
          displayDetailsOnTestsThatTriggerDeprecations="true"
          displayDetailsOnTestsThatTriggerNotices="true"
+         displayDetailsOnSkippedTests="true"
          backupGlobals="true"
          executionOrder="random"
          cacheResultFile=".phpunit.cache/test-results">


klees (commented):
Hi @mjansenDatabay,

I'm not convinced that we would want this in the default config. For one, this can be turned on easily with `--display-skipped` when running the test via `./scripts/PHPUnit/run_tests.sh --display-skipped` if the detailed info is required. Because this basically moves the meat of the tests, the failed cases, up on my screen this somehow obfuscates the information we need for the most common use case of the tests ("Did I introduce any failures...").

Any opinion?

Kind regards!

mjansenDatabay:
Hi @klees ,

thanks for looking into this.

> Any opinion?

Of course, otherwise I would not have created this PR :-). 

Currently, the list of skipped tests looks lile this:

Bildschirmfoto vom 2024-09-27 15-20-33

Of course this is caused by/depends on the reason texts used for the skipped tests, but I currently don't see any added value by just listing them here. If skipped tests are listed, we should IMO provide the information where to find those test files.

But: Displaying the details is not decisive for me. Feel free to close the PR.

Best regards,
Michael

klees:
Hi @mjansenDatabay,

this is strange. My benchmark is: run `./scripts/PHPUnit/run_tests.sh`. I then get this:

grafik

followed by some more lines of `........ x/y (z%)` and then this

grafik

Why are these skipped tests printed in your screenshot? You don't use the standard script, but the only differences that might matter is that `--teamcity` which, for me, produces waaaaay more output then what you are showing in the screenshot. Could it be that this is some output of the tools as interpreted and presented by PHPStorm?

Kind regards!

mjansenDatabay:
Hi @klees ,

you are right, this is some "PhpStorm" magic which seems to be related to the `--teamcity` argument (automatically appended).

Best regards,
Michael

PR #8042, opened 2024-09-10
Saml: Fix table. Fix config paths

lscharmer:
This PR fixes the table on the SAML Administration page and the paths when copying the config files for the first time.


mjansenDatabay:
Thx @lscharmer

PR #8045, opened 2024-09-10
Container: switch to LOM API

schmitz-ilias:
This PR replaces all usages of the old `MetaData` classes and direct queries to the `MetaData` database tables in `Container` with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md).

Since in the filter bespoke database queries were replaced with calls to a more generic API, there might be a drop in performance when filtering large containers by copyright. Feel free to let me know if you notice something like that, and I'll try to improve performance on the API side.

I tried to stick to your dependency management as much as possible, but I'm not sure whether I fully got the idea. Let me know if you want things reshuffled anywhere.

Cheers, @schmitz-ilias


alex40724:
Thanks!

PR #8046, opened 2024-09-10
BibliographicListItem: Fix upload handling

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41559

If approved, please pick this to all maintained versions (if necessary).


chfsx:
Thanks a lot, I merged to release_9, trunk was already fixed

PR #8050, opened 2024-09-11
LearningModule: switch to LOM API

schmitz-ilias:
This PR replaces the leftover usages of the old `MetaData` classes in `LearningModule` with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md).

As far as I can tell, the usages of `ilMD2XML` all lead to dead ends, so I commented them out. I did not remove them outright, since they lead to a whole network of unused export-related methods. Those should all be looked at carefully, and ripped out all at once.

Note that because of the current state of the trunk, I was not able to test everything related to export IDs.

Let me know if you want anything done differently.

Cheers, @schmitz-ilias

### DIFF
--- a/components/ILIAS/LearningModule/Presentation/class.ilLMPresentationGUI.php
+++ b/components/ILIAS/LearningModule/Presentation/class.ilLMPresentationGUI.php
@@ -16,6 +16,8 @@
  *
  *********************************************************************/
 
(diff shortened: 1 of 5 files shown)


alex40724:
@schmitz-ilias Thanks for the comments on the unused code. I will check this later and I have to get an idea how to get rid of the static calls, but this is ok for now.

PR #8052, opened 2024-09-11
switch to LOM API

schmitz-ilias:
This PR replaces all usages of the old `MetaData` classes in `File` with the new [LOM API](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/api.md).

Due to the current state of the trunk I was only able to test that the copyright selection input on file creation is generated properly. For what it's worth, that seems to be working fine.

Cheers, @schmitz-ilias

### DIFF
--- a/components/ILIAS/File/classes/Versions/class.ilFileVersionsGUI.php
+++ b/components/ILIAS/File/classes/Versions/class.ilFileVersionsGUI.php
@@ -650,13 +650,13 @@ class ilFileVersionsGUI
         );
 
         // return form at this point if copyright selection is not enabled
-        if (!$this->getLOM()->copyrightHelper()->isCopyrightSelectionActive()) {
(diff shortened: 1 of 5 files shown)


chfsx (changes requested):
Hi @schmitz-ilias, thank you very much for the PR (and your presentation at the devconf!). 

The changes are basically all great, I just have one more request. could you please remove all of the
```
protected function getLOM(): LOMServices
```
and access $this->lom_services directly? maybe this is just a question of style, but I would actually like to avoid and reduce such internal getters in order to keep the classes (in the sense of lines of code) smaller and reduce them further.

thanks a lot!

schmitz-ilias:
Hi @chfsx, thanks for the quick feedback, done!

chfsx:
Many thanks, I rebased and merged locally!

PR #8053, opened 2024-09-11
DataCollection: introduce LOM

schmitz-ilias:
This PR introduces LOM for `DataCollection` as described in [Learning Object Metadata for Data Collection](https://docu.ilias.de/go/wiki/wpage_8114_1357). For details on what this entails, see the [documentation](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/MetaData/docs/enabling_lom.md).

It also adds an extremely bare-bones PRIVACY.md. If you prefer to add one properly at some point in the future let me know, then I'll remove it from this PR.

Cheers, @schmitz-ilias


components/ILIAS/DataCollection/classes/class.ilDataCollectionExporter.php
```diff
@@ -141,15 +142,25 @@ public function getXmlExportTailDependencies(string $a_entity, string $a_target_
...
+                'ids' => $page_object_ids
             ];
         }
 
-        return [];
+        $md_ids = [];
+        foreach ($a_ids as $id) {
+            $md_ids[] = $id . ':0:dcl';
+        }
+        if (!empty($md_ids)) {
```
iszmais:
```suggestion
        if ($md_ids !== []) {
```

iszmais (approved):
Thank you a lot for your contribution!
LGTM, I just have to minor change requests:
- Resolve the opened thread for a bit more type strictness (we want to prevent mixed value functions if possible).
- Currently, there is no LOM Subtab implemented as it is described within the FR. Pls add this implementation or update the FR (Not the concept itself but add a note in implementation where you state why you did not implement it). I'm fine with both solutions.

Greetings,
@iszmais

oliversamoila:
Thank you @schmitz-ilias
In general, I agree that the DataCollection should support metadata implementation to the same level as other objects. Thanks for that.
Nevertheless, as we have already discussed, I believe that we should think about the central deactivation of the service and such options for the next release. It's yet another complex function block on this object - although I think the number of people who absolutely need it is very limited. Let us talk about this for ILIAS 11.

I would like to ask you to remove the PRIVACY.md part.
The content regarding LOM is certainly correct. However, nothing else is ‘work in progress’ at the moment and so I don't want to give that impression either.

Greetings and many thanks again
@oliversamoila

schmitz-ilias:
Hi all,

thanks for the quick response! I've made the suggested changes.

@iszmais, the subtab is handled by `ilObjectMetaDataGUI::setSubTabs`, no need to implement it in consuming components. That being said, for Data Collections it is only actually displayed when the object is in a Category with Taxonomies enabled (otherwise it is the only subtab of 'Metadata', and thus not shown). I'll clarify this in the FW along with the rest of 'Implementation' as soon as this PR is merged.

@oliversamoila, I've removed the PRIVACY.md. Regarding disabling LOM for objects, I'm sure we can work something out for ILIAS 11. There are some conceptual questions that need to be answered first (what exactly should be disabled?), I'm open to discuss options.

Cheers, @schmitz-ilias

iszmais:
@pascalseeland Due to the reach of this change I would like you to have another look add this and would appreciate your additional approval.

schmitz-ilias:
Hi @pascalseeland, @iszmais, @oliversamoila,

just a quick reminder that this PR still requires your collective approval. Let me know if there is anything else I can do to hurry this along.

(By the way, as far as I can tell the failing checks are due to an unrelated issue in the trunk.)

Cheers, Tim

iszmais:
@schmitz-ilias Thank you for your contribution!

Since it's already approved by myself and no one stated something against it I see no issue in merging this.

Greetings,
@iszmais

schmitz-ilias:
Thanks a lot @iszmais !

PR #7525, opened 2024-05-16
Abandon Session Mode «Load-Dependent Session Settings»

mBeym:
Removes everything related to the load-dependent session settings

Feature Wiki Page: https://docu.ilias.de/goto_docu_wiki_wpage_8263_1357.html

* @Uwe-Kohnle: We had to change one line in the SCORM components
* @kergomard: We removed the "Session Mode" setting from the user component

A quick review is highly appreciated.

PR #8056, opened 2024-09-12
Bugfix Authentication:0041779 add a session max idle set objective

daniwe4:
https://mantis.ilias.de/view.php?id=41779

Hello everyone, this PR ensures that the PHP ini variable 'session_max_idle' is controlled via our setup. The current values ​​are obtained from a phpinfo file that has currently been added to the repo. This process could also be made dynamic by creating the file on the fly and deleting it again after reading it. This would also prevent the .htaccess file from being adjusted. Please let me know what you think about it.

Greetings Daniel

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -23,6 +23,10 @@
 	XSendFile On
 </IfModule>
 
+<Files info.php>
+  Require ip 127.0.0.1
+</Files>
+
 AddType video/ogg .ogv
 AddType video/mp4 .mp4
 AddType video/webm .webm
--- a/Services/Authentication/classes/Setup/class.ilAuthenticationSetupAgent.php
+++ b/Services/Authentication/classes/Setup/class.ilAuthenticationSetupAgent.php
@@ -25,23 +25,45 @@ class ilAuthenticationSetupAgent implements Setup\Agent
 {
     use Setup\Agent\HasNoNamedObjective;
 
+    protected const DEFAULT_SESSION_EXPIRE = 30;
+
+    public function __construct(
+        protected Refinery\Factory $refinery
+    ) {
+    }
+
     public function hasConfig(): bool
     {
-        return false;
+        return true;
     }
 
     public function getArrayToConfigTransformation(): Refinery\Transformation
     {
-        throw new LogicException('Agent has no config.');
+        return $this->refinery->custom()->transformation(function ($data): \ilAuthenticationSetupConfig {
+            return new ilAuthenticationSetupConfig($data["session_max_idle"] ?? self::DEFAULT_SESSION_EXPIRE);
+        });
     }
 
     public function getInstallObjective(Setup\Config $config = null): Setup\Objective
     {
-        return new Setup\Objective\NullObjective();
+        if ($config !== null) {
+            return new ilSessionMaxIdleIsSetObjective($config);
+        }
+
+        return new ilSessionMaxIdleIsSetObjective(new ilAuthenticationSetupConfig(self::DEFAULT_SESSION_EXPIRE));
     }
 
     public function getUpdateObjective(Setup\Config $config = null): Setup\Objective
     {
+        if ($config !== null) {
+            return new Setup\ObjectiveCollection('Setup Authentication and Sessions', true, ...[
+                new ilDatabaseUpdateStepsExecutedObjective(
+                    new ilAuthenticationDatabaseUpdateSteps8()
+                ),
+                new ilSessionMaxIdleIsSetObjective($config)
+            ]);
+        }
+
         return new ilDatabaseUpdateStepsExecutedObjective(
             new ilAuthenticationDatabaseUpdateSteps8()
         );
--- a/Services/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
+++ b/Services/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
@@ -1,0 +1,101 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+use ILIAS\Setup;
+
+class ilSessionMaxIdleIsSetObjective implements Setup\Objective
+{
+    public function __construct(
+        protected ilAuthenticationSetupConfig $config
+    ) {
+    }
+
+    public function getHash(): string
+    {
+        return hash("sha256", self::class);
+    }
+
+    public function getLabel(): string
+    {
+        return "Ensures 'session_max_idle' is set properly";
+    }
+
+    public function isNotable(): bool
+    {
+        return false;
+    }
+
+    public function getPreconditions(Setup\Environment $environment): array
+    {
+        return [
+            new \ilIniFilesPopulatedObjective(),
+            new ilDatabaseInitializedObjective()
+        ];
+    }
+
+    public function achieve(Setup\Environment $environment): Setup\Environment
+    {
+        $client_ini = $environment->getResource(Setup\Environment::RESOURCE_CLIENT_INI);
+        $admin_interaction = $environment->getResource(Setup\Environment::RESOURCE_ADMIN_INTERACTION);
+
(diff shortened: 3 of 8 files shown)


.htaccess
```diff
@@ -23,6 +23,10 @@
 	XSendFile On
 </IfModule>
 
+<Files info.php>
```
mjansenDatabay, line 26: Would you be so kind and also add a hint in `docs/configuration/secure.md`? An example configuration for `nginx` could also help "our" administrators.

Services/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    public function isApplicable(Setup\Environment $environment): bool
+    {
+        return true;
+    }
+
+    protected function getPHPIniPathForApache(string $url): string
+    {
+        return exec("wget -q -O - ${url} | grep 'Loaded Configuration File' | cut -d '<' -f5 | cut -d '>' -f2");
```
mjansenDatabay, line 99: I guess this will not work on Windows systems, but I am also not 100% sure if we'll have to support it. Maybe we can (extend and) use the `Environment` component to determine the operating system an executre a different code here.

setup/README.md
```diff
@@ -554,3 +554,10 @@ are printed bold**, all other fields might be omitted. A minimal example is
     * *deletion_unit* (type: string) possible values are `days`, `weeks`, `months`, `years`
     * *deletion_value* (type: string or number) depending on `deletion_unit` possible values are `days max 31`, `weeks max 52`, `months max 12`, `years no max`
     * *deletion_time* (type: string) with format `HH:MM e.g. 23:30`
+* *authentication* (type: object) see also [Chat Server Setup](/Modules/Chatroom/README.md), eg.:
```
mjansenDatabay: Why do we refer to the chat server README here?

Services/Authentication/classes/Setup/class.ilAuthenticationSetupAgent.php
```diff
@@ -25,23 +25,41 @@ class ilAuthenticationSetupAgent implements Setup\Agent
...
+        });
     }
 
     public function getInstallObjective(Setup\Config $config = null): Setup\Objective
     {
-        return new Setup\Objective\NullObjective();
+        if ($config !== null) {
+            return new ilSessionMaxIdleIsSetObjective($config);
+        }
+        return new ilSessionMaxIdleIsSetObjective(new ilAuthenticationSetupConfig(30));
```
mjansenDatabay:
```suggestion

        return new ilSessionMaxIdleIsSetObjective(new ilAuthenticationSetupConfig(30));
```

Services/Authentication/classes/Setup/class.ilAuthenticationSetupAgent.php
```diff
@@ -25,23 +25,41 @@ class ilAuthenticationSetupAgent implements Setup\Agent
...
     public function getUpdateObjective(Setup\Config $config = null): Setup\Objective
     {
+        if ($config !== null) {
+            return new Setup\ObjectiveCollection('Setup Authentication and Sessions', true, ...[
+                new ilDatabaseUpdateStepsExecutedObjective(
+                    new ilAuthenticationDatabaseUpdateSteps8()
+                ),
+                new ilSessionMaxIdleIsSetObjective($config)
+            ]);
+        }
```
mjansenDatabay, line 65:
```suggestion
        }

```

Services/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
```diff
@@ -0,0 +1,103 @@
...
+                "Do you want to continue anyway?";
+
+            if (!$admin_interaction->confirmOrDeny($message)) {
+                throw new Setup\NoConfirmationException($message);
+            }
+        }
+
+        $settings = $settings_factory->settingsFor("common");
+
+        $settings->set("session_max_idle", (string) $session_max_idle);
```
mjansenDatabay: IMO we'll have to set the `expire` value in the `client.ini.php`, not the `session_max_idle` in our `settings` key/value storage.

The `session_max_idle` setting is only used by the "Load-Dependent Session"-mode, which is already removed from the ILIAS 10.x code base (see: https://docu.ilias.de/go/wiki/wpage_8263_1357 / PR #7525 )

mjansenDatabay (changes requested):
Thx @daniwe4 for the PR.

I added some inline comments for my remarks. 

Best regards,
Michael

Services/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
daniwe4: Done!

Services/Authentication/classes/Setup/class.ilAuthenticationSetupAgent.php
daniwe4: Done!

Services/Authentication/classes/Setup/class.ilAuthenticationSetupAgent.php
daniwe4: Done!

setup/README.md
daniwe4: Done!

.htaccess
daniwe4: Done!

Services/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
daniwe4: I think that we don't support Windows in the setup natively. But we could of course install some kind of monitoring that checks whether the installation is on Windows and sets the values ​​differently accordingly.
Maybe you have ideas for a possible implementation.

mjansenDatabay: I will ask @mbecker-databay as a "Windows Power User" ;-).

mjansenDatabay: @daniwe4 Maybe we should try to solve this with PHP-only.

IMO we could use `file_get_contents` for the HTTPS call. Of course, fetching URLs with `file_get_contents` could be disabled, and other exceptions might occur as well (SSL-related, Proxy, etc.).
We could also try to check if `curl` (https://www.php.net/manual/en/book.curl.php) is available, and use `file_get_contents` as fallback.

mbecker-databay: What @mjansenDatabay says.
I know y'alls are all for dropping windows support but in this case an internal solution should be preferred over exec.

smeyer-ilias:
@mjansen: "The use of file_get_contents requires that allow_url_fopen is enabled—at least in the CLI context. In my opinion, this setting should remain disabled for security reasons."

Passing "localhost" to wget will lead to incorrect results if there is no vhost configuration - which is not unusual - for localhost or if different PHP versions are used for different host headers (e.g., for testx-ilias.de). It might be possible to use the ILIAS HTTP path instead of localhost. However, in that case, the IP restriction from secure.md might no longer apply (in a reverse proxy scenario). Maybe a sort of maintenance mode is enabled while performing the setup. In this case the info.php will not be accessible.

For these reasons, among others, I am wondering whether this feature would be better placed in the trunk."

mjansenDatabay:
Hi @ all,

a big "Thank you" for the advice, @smeyer-ilias . 

> The use of file_get_contents requires that allow_url_fopen is enabled

Yes, that's what I meant with:

> Of course, fetching URLs with file_get_contents could be disabled, and other exceptions might occur as well (SSL-related, Proxy, etc.).

---

> For these reasons, among others, I am wondering whether this feature would be better placed in the trunk."

To be honest, I hadn't really noticed the “target release” of the PR :eyeglasses: .

And you are right, we should not put this into a stable release.

@daniwe4 Would it be possible to switch the target branch to `trunk` for the reasons mentioned in the discussions?

---

> Passing "localhost" to wget will lead to incorrect results if there is no vhost configuration - which is not unusual - for localhost or if different PHP versions are used for different host headers (e.g., for testx-ilias.de). It might be possible to use the ILIAS HTTP path instead of localhost. However, in that case, the IP restriction from secure.md might no longer apply (in a reverse proxy scenario).

 I will spend some time thinking about this once more and also consult with colleagues, if there are any alternative solutions. 

Best regards,
Michael

daniwe4:
Hello, I moved this PR to Trunk. We are welcome to continue the discussion there about how we read the Apache-PHP.ini from the CLI. 
New PR https://github.com/ILIAS-eLearning/ILIAS/pull/8114
Greetings Daniel

mjansenDatabay:
The `trunk` version has been merged, I close this for `release_9` due to the reasons above.

PR #8061, opened 2024-09-12
Data: Add @template type annotations for `ILIAS\Data\Result`

lscharmer:
This PR adds type annotations for the `ILIAS\Data\Result` inteface as well as it's implementors `ILIAS\Data\Result\Ok` and `ILIAS\Data\Result\Error` to support a generic type. These type annotations are understood by phpstan (https://phpstan.org/blog/generics-in-php-using-phpdocs) and phpstorm.

With this change the following can be checked via static analysis:
```php
/**
 * @param Result<Baz> $bar
 */
function foo(Result $bar): void
{
    // ...
}
```

### DIFF
--- a/components/ILIAS/Data/src/Result.php
+++ b/components/ILIAS/Data/src/Result.php
@@ -1,8 +1,22 @@
 <?php
 
-declare(strict_types=1);
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
 
-/* Copyright (c) 2017 Richard Klees <richard.klees@concepts-and-training.de> Extended GPL, see docs/LICENSE */
+declare(strict_types=1);
 
 namespace ILIAS\Data;
 
@@ -10,6 +24,8 @@ namespace ILIAS\Data;
  * A result encapsulates a value or an error and simplifies the handling of those.
  *
  * To be implemented as immutable object.
+ *
+ * @template-covariant A
  */
 interface Result
 {
@@ -21,7 +37,7 @@ interface Result
     /**
      * Get the encapsulated value.
      *
-     * @return mixed
+     * @return A
      * @throws \Exception    if !isOK, will either throw the contained exception or
      *                      a NotOKException if a string is contained as error.
      */
@@ -43,8 +59,10 @@ interface Result
     /**
      * Get the encapsulated value or the supplied default if result is an error.
      *
-     * @param mixed $default
-     * @return mixed
+     * @template B
+     *
+     * @param B $default
+     * @return A|B
      */
     public function valueOr($default);
 
@@ -53,7 +71,10 @@ interface Result
      *
      * Does nothing if !isOK.
      *
-     * @param callable $f mixed -> mixed
+     * @template B
+     *
+     * @param callable(A): B $f
+     * @return Result<B>
      */
     public function map(callable $f): Result;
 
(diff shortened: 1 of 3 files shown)


klees (commented):
Hi @lscharmer,

thanks for the proposal. Although I like statically typed languages with parametrized types and strong type checkers just like the next guy, I'm not super convinced that we would want to do this.

On the one hand: What do we actually gain here? We don't mandate phpstan usage. Even if so, this would only allow to check this teeny tiny part of the code base. On the other hand: New syntax. New stuff for people to trip up over. New inconsistencies, because lot of locations won't document accordingly.

This doesn't mean I oppose this, but to me this somehow lacks a bigger picture. If someone asks @mjansenDatabay or me "Why do we do this? Shall I do this? Why?", say on a dev-conf. What shall we answer? "So Lucas can check this location with PHPStan. Maybe, maybe not. Don't really know." just won't cut it =)

So: Is there a plan here? Or at least an idea of a plan? Or at least a general idea or direction regarding PHPStan or this kind of documentation for generics?

Kind regards!

klees:
Please pass back if you have some feedback!

lscharmer:
Hi @klees,
thank you for your feedback.

Just to be clear: With new syntax you mean the `@template*` annotations?
The `Foo<Bar>` syntax is not new, as we are using this with arrays too (e.g. array<int, string>) and as they are generic types too.
This PR doesn't require users of this API to write `@template*` annotations, just the `Foo<Bar>` syntax which is already known (but this PR does **not** require this, it's optional). But I agree that the `@template*` annotations are something that people may trip over.

> So Lucas can check this

*Lukas

> What shall we answer? "So Lucas can check this location with PHPStan. Maybe, maybe not. Don't really know." just won't cut it =)

I created this PR as I talked with @mjansen about this as he also thought about adding annotations to the result class.
As mentioned in the PR description: This is not just for PHPStan. For IDE's e.g. PHPStorm this provides autocompletion just like it provides autocompletion for arrays, which are annotated accordingly. IMO this doesn't force people to use these annotations, nor understand them (in regards to `@template*`). The result class can be used just the same as bevor but for people who want to be more specific with their types they now can be. The problem of wrong documentations exists, but it exists just the same as with wrong `array<Foo>` annotations.

> So: Is there a plan here? Or at least an idea of a plan? Or at least a general idea or direction regarding PHPStan or this kind of documentation for generics?

No, there is no plan here. This is just a simple addition which is a general improvement IMO as I regard these annotations as general knoweledge just the same as annotations like `@param array<int, string> $bar`. If you disagree you can simply close this PR :)

Best regards
Lukas

mjansenDatabay:
Hi @klees , hi all,

@lscharmer answered (as usual) faster than a politician changes his mind. Okay, in this case, I was just way too slow (~ 2 weeks) :-).

But to be transparent: I send @lscharmer to suggest these changes for the `Result` type. It is not only an optimization for "PHPStan", it's also "Psalm" and "Phpstorm" which support (have agreed on) this syntax.

Of course, everytime the IDE (to be concrete: "PhpStorm") parses the file with "PhpStan" being enabled, problems are shown in the corresponding IDE area. Furthermore, such issues could be listed in gathered "PhpStan" bulk reports (e.g. in the CI, local scripts triggered by Git hooks, etc.).
But even if "PhpStan" is disabled - which is maybe still the case for many ILIAS developers, although I hope this is not true anymore, especially since we all together defeated the PHP 8 daemon a few years ago with the help of "PhpStan" - types can be resolved with "PhpStorm" out of the box, which makes navigation between files/classes/... in a huge code base such as ILIAS very easy. Autocompletion feature/intellisense in "PhpStorm" highly depends on this syntax when using "Generics" in userland code.

This syntax is widely known in the PHP community and part of ongoing discussions regarding "Generics", e.g. here https://thephp.foundation/blog/2024/08/19/state-of-generics-and-collections/#static-analysis .

But as Lukas already said: It's okay for us to close this PR.

Best regards,
Michael

mjansenDatabay:
This could be a topic for a "Dev. Conf" workshop: "Generics in PHP using PHPDocs", @lscharmer volunteers to lead it :-).

dsstrassner:
Some side note: this is the second idea for a DevConf Workshop / Presentation I saw in the past days. I would propose that @matthiaskunkel or @ILIAS-eLearning/technical-board opens the DC for the next DevConf in Duisburg early, so we can start collecting these ideas, so they don't vanish from our minds.

klees:
Hi @lscharmer,

as outlined above, I don't see a real benefit here.

Kind regards!

PR #8066, opened 2024-09-13
hun - 20240913

KissDaniGH:


matthiaskunkel:
Thanks to @KissDaniGH for his PR and the maintenance of the Hungarian language support.

PR #8070, opened 2024-09-16
Add jstree as dependency

alex40724:
Larger trees like the repository or org unit tree must load subtrees asynchronously due to performance reasons. The KS tree component currently does not support this. Thus we still need the legacy implementation for ILIAS 10 which uses the js tree lib.

github: https://github.com/vakata/jstree

Latest version is two weeks old. 74 contributors.

Wrapped by UIComponent.


matthiaskunkel:
Jour Fixe, 16 SEP 2024: Dependency is accepted for ILIAS 10. Please merge PR to trunk (after successful checks).

alex40724:
Merged manually.

PR #8074, opened 2024-09-16
Test export plugins cause error during ILIAS setup (42070)

fneumann:
Edit by @dsstrassner: Issue Link is https://mantis.ilias.de/view.php?id=42070

### DIFF
--- a/Modules/Test/classes/class.ilTestExportPlugin.php
+++ b/Modules/Test/classes/class.ilTestExportPlugin.php
@@ -37,19 +37,12 @@ abstract class ilTestExportPlugin extends ilPlugin
         'csv'
     );
     private \ilGlobalTemplateInterface $main_tpl;
-    private \ilLanguage $lng;
-    private \ilCtrl $ctrl;
 
     public function __construct(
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @fneumann .
Ok, this actually is understandable even if I hate that we have to do this ;-) . As this is a bug-fix:
- Could you please link the corresponding mantis issue in the comment @fneumann (or create one, if there is none yet), so that we have a clean traceability. I will then take care of merging this.

Thanks again and best,
@kergomard

dsstrassner:
Do we need the second review from @mbecker-databay because of some internal  Databay agreements? If not, please proceed @kergomard …

kergomard:
Thank you very much for the  information @dsstrassner ! I will merge this, to have it off my plate. I don't see much of a reason not to.
Thanks again @fneumann for the PR.

thojou:
Hey @dsstrassner,

I also looked into this and discussed the issue with @mjansenDatabay. As @kergomard mentioned, it seems there is no better solution available at this point. Thanks for merging.

PR #8078, opened 2024-09-16
Fix typo and clarify db-install in README.md

ingoj:
arbuilder: the file offered for download has an extention of .php (not .tet). I added a note to clarify that arBuilder only creates a step to install a db-table.


ingoj:
may apply to release_9 as well and somewhere in trunk

acgruber:
Hello @chfsx,
Since you're listed in the respective maintenance.json, could you please have a look at these changes and confirm? Also, I would appreciate your advise on how we should merge this. My usual approach would be to merge docu related stuff into trunk only, but maybe not in this case? What do you think?
Thanks & best regards!

chfsx:
Thanks a lot! I pick that to the other branches as well

PR #8079, opened 2024-09-17
Fixes uri input throwing error when left empty but not required

matheuszych:
When using a URI input and setting it to not required you can not leave it empty because you get an error stating that "" is not a valid URI. This would be true but setting the input to not required should probably allow me to leave it empty so no validity checks are performered in that case.

I believe this a viable fix for the issue mentioned.

https://mantis.ilias.de/view.php?id=42077

### DIFF
--- a/Services/Form/classes/class.ilUriInputGUI.php
+++ b/Services/Form/classes/class.ilUriInputGUI.php
@@ -45,16 +45,19 @@ class ilUriInputGUI extends ilTextInputGUI
     {
         $lng = $this->lng;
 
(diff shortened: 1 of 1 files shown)


klees:
Hi @dsstrassner,

this is not a fix in the UI framework but in the legacy forms, so the UI framework coordinators are not the correct assignees. We still had a look into this and take this over as shepherds.

Kind regards!

klees (approved):
Looks plausibel, thanks!

klees:
Picked to trunk.

PR #8080, opened 2024-09-18
Dependencies: Remove `rector` files from `scripts`

mjansenDatabay:
`rector` is currently not suggested/approved as dev-dependency for ILIAS 10.x. Furthermore, our custom rectors are not compatible to the latest version/API of rector.

Therefore, I suggest to remove the rector files from our codebase.


matthiaskunkel:
Hi @mjansenDatabay, this is a PR against 10 and trunk, right? Has nothing to do with the Dependency JF next week.

mjansenDatabay:
Hi @matthiaskunkel , you are right. This PR suggests removing the `rector`-based files from our code base, since `rector` is not part of our dependencies anymore (in ILIAS > = 10.x).

nicoroeser:
Dependencies label removed as discussed in Jour Fixe just now.

PR #8085, opened 2024-09-18
Filter uploaded files from hasUploads if ERR_NO_FILE detected

thojou:
As outlined in https://mantis.ilias.de/view.php?id=42106, we encountered an issue with file upload handling in the assFileUpload component. Specifically, when a form with a file upload field is submitted without selecting a file, it results in an empty UploadedFile object. Below is a sample dump illustrating the issue:

file upload dump

The problem arises because the hasUploads method still returns true, as it counts any available entries from the HTTP request, regardless of whether a file was actually uploaded.

We modified the hasUploads method to filter out files with the error "ERR_NO_FILE", ensuring that they are not counted as valid uploads.

If this change is likely to introduce side effects—for example, if the method is expected to return all files, regardless of their error state—an alternative solution could be to implement a custom hasValidUploads method. This would specifically count only successfully uploaded files.

### DIFF
--- a/src/FileUpload/FileUploadImpl.php
+++ b/src/FileUpload/FileUploadImpl.php
@@ -358,6 +358,10 @@ final class FileUploadImpl implements FileUpload
         }
 
         $uploadedFiles = $this->flattenUploadedFiles($this->globalHttpState->request()->getUploadedFiles());
+        $uploadedFiles = array_filter(
+            $uploadedFiles,
+            fn (UploadedFileInterface $uploadedFile) => $uploadedFile->getError() !== UPLOAD_ERR_NO_FILE
+        );
 
         return ($uploadedFiles !== []);
     }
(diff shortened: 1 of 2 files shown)


dsstrassner:
Hi @chfsx, any updates on this PR? I ask in my role as authority for T&A,  the Mantis issue 42106 was first assigned to me. As the end of full support for ILIAS 8 is near, would it be possible to merge this PR or needs the PR some changes from @thojou? 

Thanks and best regards

@dsstrassner

kergomard:
Hi @chfsx 
I don't know if this  is needed and correct. At least in the test I think the current behavior without the changes is correct and the issue in the test needed a different fix.

Best,
@kergomard

mjansenDatabay:
Hi @chfsx  , we stumbled over this PR while checking old PRs in our TB meeting.

Please decide if you want to close the PR, request changes, request a new PR targeting a maintained ILIAS branch, or adopt the changes yourself for a maintained release.

Providing feedback for the author is highly appreciated.

thojou:
Hi all,

I'm closing this PR to remove it from my list. Since this was originally created for release_8 and the core has undergone many iterations, if the issue persists, we will need a different approach.

Kind regards,
@thojou

PR #8086, opened 2024-09-18
Fix assFileUpload to prevent error message if files are deleted

thojou:
As outlined in https://mantis.ilias.de/view.php?id=42106, a false-positive error message appears when deleting already uploaded files in an assFileUpload question.

This PR modifies the file upload handling logic in the assFileUpload component to prevent the error message from being displayed when files are deleted.

This PR requires #8085 to be merged beforehand.

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assFileUpload.php
+++ b/Modules/TestQuestionPool/classes/class.assFileUpload.php
@@ -1177,11 +1177,7 @@ class assFileUpload extends assQuestion implements ilObjQuestionScoringAdjustabl
      */
     protected function isFileUploadAvailable(): bool
     {
-        if (!isset($_FILES['upload'])) {
-            return false;
-        }
-
-        if (!isset($_FILES['upload']['tmp_name'])) {
+        if (!$this->file_upload->hasUploads()) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


Modules/TestQuestionPool/classes/class.assFileUpload.php
```diff
@@ -1177,18 +1177,14 @@ protected function isFileReuseHandlingRequired(): bool
      */
     protected function isFileUploadAvailable(): bool
     {
-        if (!isset($_FILES['upload'])) {
-            return false;
-        }
-
-        if (!isset($_FILES['upload']['tmp_name'])) {
+        if(!$this->file_upload->hasUploads()) {
```
mjansenDatabay:
```suggestion
        if (!$this->file_upload->hasUploads()) {
```

thojou: Thanks for pointing that out, @mjansenDatabay. I've made the necessary adjustments to the code.

kergomard:
Thank you very much @thojou and sorry, I hadn't seen the "requires" part before merging, ...but as this equalizes behavior across ILIAS 8 and ILIAS 9 (more or less), I leave it in, we can then add additional checks depending on the reaction in #8085.

Best,
@kergomard

PR #8089, opened 2024-09-18
Saml: Fix wrong baseurlpath in Saml README.md

lscharmer:
Sister PR to #8088.

### DIFF
--- a/components/ILIAS/Saml/README.md
+++ b/components/ILIAS/Saml/README.md
@@ -103,7 +103,7 @@ The key `debug` is now not a boolean but an array value instead. New installatio
 The key `baseurlpath` must be changed to the following. New installations will have this value as default.
 ```diff
 - 'baseurlpath' => 'simplesaml/',
-+ 'baseurlpath' => 'Services/saml/lib/',
++ 'baseurlpath' => 'Services/Saml/lib/',
 ```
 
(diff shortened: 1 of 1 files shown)


components/ILIAS/Saml/README.md
```diff
@@ -103,7 +103,7 @@ The key `debug` is now not a boolean but an array value instead. New installatio
 The key `baseurlpath` must be changed to the following. New installations will have this value as default.
 ```diff
 - 'baseurlpath' => 'simplesaml/',
-+ 'baseurlpath' => 'Services/saml/lib/',
++ 'baseurlpath' => 'Services/Saml/lib/',
```
mjansenDatabay, line 106: Hmmmm, in `trunk` we do not have `Services` anymore.

lscharmer: Thats correct, but this is just to be consistent. This is the ILIAS 9 section, for ILIAS 10 there is also a section where this is addressed. So this place is just changed so it is consistent with ILIAS 9 README even though the documentation doesn't apply to ILIAS 10.

Here is the ILIAS 10 config change:
https://github.com/ILIAS-eLearning/ILIAS/blob/783271bb1e66663a244a08a60ad4bcff59765a31/components/ILIAS/Saml/README.md?plain=1#L144

PR #8090, opened 2024-09-19
Notes fix of #41936

Amstutz:
Fix for #https://mantis.ilias.de/view.php?id=41936

Note that the issue is the wrong primary type (was integer, should be string). This leads to duplicate entries in the db, such as:
image

Since only the first is read and the other is written, comments in migrated objects can not be activated/deactivated.

Note that this PR is not without issues. This does not remove wrong entries, it just fixed the wrong type and then always sticks to reading the one with a type if available. 

The Proper way to fix this would be to remove the wrongly generated entries and copying their values to the right ones as well as adding the obj_type for all new entries generated since the update.


alex40724:
Fixed manually.

PR #8091, opened 2024-09-19
Remove Dependence on Custom TinyMCE

kergomard:
This removes all dependencies to the custom TinyMCE we are using in previous versions of ILIAS and uses the default TinyMCE instead.

I also propose to update TinyMCE to the most current version 7.3.0 as support for 6.8.4 will end in 2025 far before ILIAS 10 will be out of support.

Best,
@kergomard


matthiaskunkel:
Jour Fixe, 30 SEP 2024: We highly appreciate this suggestion and accept the PR for ILIAS 10/trunk, incl. the update of TinyMCE to version 7.3.0.

kergomard:
Thank you very much for the positive feedback!

PR #8092, opened 2024-09-19
LearningModule / T&A Feature: Disable ImageMap Question Answers after Solution provided

fhelfer:
https://mantis.ilias.de/view.php?id=33539#c105407

Disabling answers in Learning Module after the solution has been selected or shown because of too many failed attempts. Hover and Click Event Listeners are being disabled on the ImageMapping Areas.


kergomard (approved):
Thank you very much for the PR @fhelfer ! Not beautiful, but does the trick. One more reminder that refactoring questions should really, really be next on the list.

PR #8093, opened 2024-09-20
Repository/PluginSlot: Add ability to use next_class in the derived class

shecken:
There is no way for plugins to route from their GUI to another one via the next_class system. The only chance so far is to read the next_class from the control structure again in the performCommand() and then use it. But that is not the purpose of this function. That is why I would like to add a performNextClass() that gives the plugins an option to react to the value in next_class.

The advantage is that you no longer have to write everything in the plugin's GUI and you get a clearer code.

### DIFF
--- a/components/ILIAS/Component/classes/class.ilPluginConfigGUI.php
+++ b/components/ILIAS/Component/classes/class.ilPluginConfigGUI.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -13,8 +14,7 @@
  * https://www.ilias.de
  * https://github.com/ILIAS-eLearning
  *
- ********************************************************************
- */
+ *********************************************************************/
 
 /**
  * Parent class for all plugin config gui classes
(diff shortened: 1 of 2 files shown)


components/ILIAS/Component/classes/class.ilPluginConfigGUI.php
```diff
@@ -85,8 +85,17 @@ public function executeCommand(): void
             );
         }
 
+        $next_class = $ilCtrl->getNextClass();
+        if($next_class && $this->performNextClass($next_class)) {
```
mjansenDatabay:
```suggestion
        if ($next_class && $this->performNextClass($next_class)) {
```

components/ILIAS/Repository/PluginSlot/class.ilObjectPluginGUI.php
```diff
@@ -159,6 +159,11 @@ public function executeCommand(): void
                 $this->ctrl->forwardCommand($gui);
                 break;
             default:
+                if($next_class && $this->performNextClass($next_class))
```
mjansenDatabay:
```suggestion
                if ($next_class && $this->performNextClass($next_class))
```

klees (approved):
LGTM.

@alex40724 I pass on to you, from my perspective this can be merged.

klees:
Addendum: I asked @shecken to fix CS, of course.

PR #8096, opened 2024-09-23
TextHandling: implement a tool to solve string-based text handling problems

lukastocker:
- create three layers (Markup, Shape and Text) in `Data/src/TextHandling` --- include classes accordingly
  - declare possible structures via enum
  - include checks for compliant strings for Markdown inputs
- write PHPUnit Tests

You can read more in the [Text Handling Development Documentation](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/text-handling.md)

### DIFF
--- a/components/ILIAS/Data/src/Factory.php
+++ b/components/ILIAS/Data/src/Factory.php
@@ -34,13 +34,13 @@ use ILIAS\Data\Meta;
  */
 class Factory
 {
-    /**
-     * cache for color factory.
-     */
+    // TODO: move this to proper dependency_injection
     private ?Color\Factory $colorfactory = null;
     private ?Dimension\Factory $dimensionfactory = null;
     private ?Meta\Html\Factory $html_metadata_factory = null;
     private ?Meta\Html\OpenGraph\Factory $open_graph_metadata_factory = null;
+    private ?Text\Factory $text_factory = null;
+
 
     /**
      * Get an ok result.
@@ -221,5 +221,20 @@ class Factory
     public function languageTag(string $language_tag): LanguageTag
     {
         return LanguageTag::fromString($language_tag);
+    }
+
+    public function text(): Text\Factory
+    {
+        if ($this->text_factory === null) {
+            $md_format = new \ILIAS\Refinery\String\MarkdownFormattingToHTML();
+            $this->text_factory = new \ILIAS\Data\Text\Factory(
+                new Text\MarkdownFactory(
+                    new Text\Shape\Markdown($md_format),
+                    new Text\Shape\SimpleDocumentMarkdown($md_format),
+                    new Text\Shape\WordOnlyMarkdown($md_format)
+                )
+            );
+        }
+        return $this->text_factory;
     }
 }
--- a/components/ILIAS/Data/src/Text/Factory.php
+++ b/components/ILIAS/Data/src/Text/Factory.php
@@ -1,0 +1,37 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\Data\Text;
+
+class Factory
+{
+    public function __construct(
+        protected MarkdownFactory $markdown_factory
+    ) {
+    }
+
+    public function markdown(?string $markdown = null): MarkdownFactory|Markdown
+    {
+        if ($markdown !== null) {
+            return $this->markdown_factory->generic($markdown);
+        }
+        return $this->markdown_factory;
+    }
+}
--- a/components/ILIAS/Data/src/Text/HTML.php
+++ b/components/ILIAS/Data/src/Text/HTML.php
@@ -1,0 +1,34 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\Data\Text;
+
+/**
+ * This class currently is a stub only. It will need a proper implementation, just
+ * like Text\Markdown has.
+ */
+class HTML
+{
+    public function __construct(
+        protected string $html_text = ""
+    ) {
+        $this->html_text = $html_text;
+    }
+}
--- a/components/ILIAS/Data/src/Text/Markdown.php
+++ b/components/ILIAS/Data/src/Text/Markdown.php
@@ -1,0 +1,33 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\Data\Text;
+
+use ILIAS\Data\Text\Shape\Markdown as MarkdownShape;
+
+class Markdown extends TextBase
+{
+    public function __construct(
+        MarkdownShape $markdown_shape,
+        string $raw
+    ) {
+        parent::__construct($markdown_shape, $raw);
+    }
+}
--- a/components/ILIAS/Data/src/Text/MarkdownFactory.php
+++ b/components/ILIAS/Data/src/Text/MarkdownFactory.php
@@ -1,0 +1,46 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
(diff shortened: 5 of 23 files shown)


klees:
Hi everyone,

I made some further improvements here and I think this is good to go now. I kindly ask for reviews and feedback from every interested developer, as this shall become the toolbox to handle text in ILIAS programmatically, as described [in the paper](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/docs/development/text-handling.md).

Kind regards!

klees:
Unit tests currently fail for unrelated reasons...

components/ILIAS/Data/src/TextHandling/Shape/SimpleDocumentMarkdown.php
```diff
@@ -0,0 +1,72 @@
...
+
+namespace ILIAS\Data\TextHandling\Shape;
+
+use ILIAS\Data\TextHandling\Structure;
+use ILIAS\Data\TextHandling\Text;
+
+class SimpleDocumentMarkdown extends Markdown
+{
+    /**
+     * @return mixed[] consts from Structure
```
mjansenDatabay:
```suggestion
     * @return Structure[]
```

components/ILIAS/Data/src/TextHandling/Shape/WordOnlyMarkdown.php
```diff
@@ -0,0 +1,72 @@
...
+
+namespace ILIAS\Data\TextHandling\Shape;
+
+use ILIAS\Data\TextHandling\Structure;
+use ILIAS\Data\TextHandling\Text;
+
+class WordOnlyMarkdown extends SimpleDocumentMarkdown
+{
+    /**
+     * @return mixed[] consts from Structure
```
mjansenDatabay:
```suggestion
     * @return Structure[]
```

or simply remove the docs because of the definition in the paren class.

mjansenDatabay (commented):
Hi @lstocker, hi @klees,

Thank you for the pull request and your work. After studying the paper back then, the pull request now starts to "complete the picture" for me.

I only have the following comments:

- For my brain, it was/is challenging ("personal preference/fail") when class names (not fully qualified) are, for example, simply called `Markdown`. The same applies to `SimpleDocumentMarkdown` and `WordOnlyMarkdown`. Since some class names were chosen this way two or three times, it made it (for me) more difficult to keep track, even though the number of files isn’t overwhelmingly large.
- The implementations for `PlainText` and `HTML` texts are still missing. Are you planning to include them as part of this PR? If not, do you see a medium-term need for them? There are probably some challenges here: on one hand, we would certainly need proper/corresponding "converters/transformations," and on the other hand, it might be difficult to define "structures" given the proliferation of "HTML" text usage in ILIAS. Do you see it the same way?
I guess since strategically, ILIAS is moving in the direction of "Markdown", "HTML" will be less relevant in the future.

Best regards,
Michael

Attached is a simplified class diagram in which I’ve omitted associations, aggregations, and compositions for clarity. Having a graphical overview might be helpful for some people, and it could save time when reading the paper and studying the source code.

```mermaid
classDiagram
direction BT

class Text {
    <<interface>>
    + getShape(): Shape
    + getMarkup(): Markup
    + getSupportedStructure(): Structure[]
    + toHTML(): Text\HTML
    + toPlainText(): Text\PlainText
    + getRawRepresentation(): string
}

class Shape {
    <<interface>>
    + fromString(text: string): Text
    + isRawStringCompliant(text: string): bool
    + toHTML(text: Text): Text\HTML
    + toPlainText(text: Text): Text\PlainText
    + getMarkup(): Markup
}

class Markup {
    <<interface>>
}

class Structure {
    <<enum>>
    + HEADING_1
    + HEADING_2
    + HEADING_3
    + HEADING_4
    + HEADING_5
    + HEADING_6
    + BOLD
    + ITALIC
    + UNORDERED_LIST
    + ORDERED_LIST
    + LINK
    + PARAGRAPH
    + BLOCKQUOTE
    + CODE
}

%% Classes start here
class Base {
    <<abstract>>
    + __construct(shape: Shape, raw: string)
    + getShape(): Shape
    + getMarkup(): Markup
    + getSupportedStructure(): Structure[]
    + toHTML(): Text\HTML
    + toPlainText(): Text\PlainText
    + getRawRepresentation(): string
}

class HTML {
    + __construct(html_text: string)
}

class PlainText {
    + __construct(plain_text: string)
}

class Markdown {
    + __construct(markdown_shape: Shape\Markdown, raw: string)
}

class SimpleDocumentMarkdown {
    + __construct(simple_document_markdown_shape: Shape\SimpleDocumentMarkdown, raw: string)
}

class WordOnlyMarkdown {
    + __construct(word_only_markdown_shape: Shape\WordOnlyMarkdown, raw: string)
}

class ShapeMarkdown {
    + __construct(markdown_to_html_transformation: Refinery\Transformation)
    + toHTML(text: Text): Text\HTML
    + toPlainText(text: Text): Text\PlainText
    + getMarkup(): Markup\Markdown
    + fromString(text: string): Text\Markdown
    + isRawStringCompliant(text: string): bool
}

class ShapeSimpleDocumentMarkdown {
    + getSupportedStructure(): Structure[]
    + fromString(text: string): Text\SimpleDocumentMarkdown
    + isRawStringCompliant(text: string): bool
}

class ShapeWordOnlyMarkdown {
    + getSupportedStructure(): Structure[]
    + fromString(text: string): Text\WordOnlyMarkdown
    + isRawStringCompliant(text: string): bool
}

class MarkupMarkdown

%% Relationships
Base <|.. Text
Markdown <|-- Base
SimpleDocumentMarkdown <|-- Markdown
WordOnlyMarkdown <|-- SimpleDocumentMarkdown
ShapeMarkdown <|.. Shape
ShapeSimpleDocumentMarkdown <|-- ShapeMarkdown
ShapeWordOnlyMarkdown <|-- ShapeSimpleDocumentMarkdown
MarkupMarkdown <|.. Markup
```

klees:
Hi @mjansenDatabay,

thanks for the positive feedback, suggesstion, questions and, last but not least, the diagram. I have added it to the PR.

Regarding your comments:

* at classnames: Yeah, I see your point. I was torn here. On the one hand, I could not figure better classnames, as these  thingies point at the same contexts, so `Markdown` it is for `Shape` and `Text`. On the other hand I did not want to simply add `Text` and `Shape` at the end, that's weird for my brain (`Text\MarkdownText`, `Shape\MarkdownShape`). I also thought about `Markdown\Text`, `Markdown\Shape`... but that won't help you, I guess, and also feels harder to keep track. The overwhelming amount of code, though, will use the instances of `Text` directly, hinting on `Text\Markdown` etc. feels good to me. The overwhelming amount of code in the TextHandling-component, on the other hand, will happen in `Shape`, and the thingies in `Text` are only wrappers used for type hinting. So, if no stronger resentment on the naming scheme comes up (either from you or someone else...) I would keep it as is, at least for the moment.
* at implementations: Yes, these implementations are missing. No, I do not plan to include them in this PR. I think we have kind of immediate need for them, as currently the most text handling issues arise in HTML and plain text (since markdown isn't used much as of now...). I also see that we'll have some bigger troubles with them, as we'll need these converters. Regarding the structure, I would stick to the structures that we have defined now until proven that we need something more. As described in the paper, I don't think that we want to encode all of or only much of HTML, only the stuff that we need for "texts". This might be a little to narrow now, but I also don't think we will need to add a lot. I think `PlainText` and `HTML` will be mostly used as targets in the future, not so much as input. So, in essence: Lets see how this goes and move this along...

Hopee this helps, thanks for looking into this!

Kind regards!

thibsy (changes requested):
Hi @lukastocker,

Thanks a lot for this great contribution! This is much needed :).

I discussed this PR with @klees today, who also asked me to look into it, since I am somewhat invested in the markdown aspects (see e.g. #7578).

- [x] Caching: the `Shape::isRawStringCompliant()` method call could possibly be quite expensive, or not? Since calling the `to<format>()` methods will always create new instances, do you think it would make sense to cache their result?

- [x] Factory pattern: you are providing a well documented entry point to the system, but I think the implementation is not quite ready for consumption yet. Creating `Text` instances needs knowledge about the internals right now (what transformation for markdown? what shape for which text?) and there is no wiring going on - consumers would need to do all this. I suggest to implement a factory pattern here as well, which lets consumers create `Text` instances more easily. This would also solve some other "issues" you are having, where e.g. a new `Markup` instance is created on each call, which could be prevented using DI.
- [x] Format methods: I think it would be more appropriate, if the `Shape::to<format>()` methods would throw a `LogicException` instead of an `InvalidArgumentException`. If a wrong instance is ever passed, the `Shape` is not being used properly and it would thus be an invalid state - but correct me if I'm wrong.
- [x] `Structure` enum: could you make this a `string` enum? The cases could e.g. be the tag-name of the corresponding HTML elements (I think all options are representable). The UI markdown field would directly benefit from this, because it needs to provide actions for the supported structures.
- [x] Inline Regex: you are currently declaring inline Regex patterns for validating the raw markdown strings. Could you transfer them into a `string` enum as well? This will probably get rid of some redundancy and provides a better place to document the individual patterns, all while improving readability =).

- [x] Namespace: the `ILIAS\Data` namespace contains additional datatypes for ILIAS, whereas the namespace `ILIAS\Data\TextHandling` suggests that more is happening there. Please choose a simpler name like e.g. `ILIAS\Data\Text`, to indicate we are not in fact "handling" text there, but simply represent it.
- [x] Fatal error: `Text::toHTML()` and `::toPlainText()` will currently cause an error because you are passing them a `string` instead of a `Text` instance.

Please note, I think now is the best time to implement the factory pattern, because once this PR is merged and developers start using these classes, refactoring becomes more complex than a simple search-and-replace.

Thx for your time!

Kind regards,
@thibsy

klees:
Hi  @thibsy,

thanks for the feedback, input and interesting discussion via discord. To the meat of it:

* Caching: I think we would need to go a loooong way until this will become interesting. As far as I know, ILIAS performance currently is mostly bound to I/O, more specifically roundtrip time to the database. Every call to ILIAS makes many requests to the DB, these determine the overall performance. `isRawStringCompliant` on the other hand works completely on some existing data in memory. If we will ever notice we could look into performance improvements, e.g. caching.
* Factory pattern: Done! 
* Format methods: Yes, this is indeed correct. Since you already have a `Text` instance anyway when calling `toPlainText` or `toHTML` you should rather use that instance directly. Having that code on `Shape` is just a code organisation thing. I changed accordingly.
* `Structure` enum: Sure can do.
* `Inline` Regex: Sure can do.
* Namespace: I have renamed the topmost namespace from `TextHandling` to `Text` and moved all the instances in the previously nested namesspace `TextHandling\Text` into that topmost namespace. These are the classes that will mostly be hinted at I would assume. Also, I looked for "handling" in the docu and replaced it by "representation". This will explicitly not "handle" the text =) Also, I think this might help a little with @mjansenDatabay|s problem regarding naming.
* Fatal error: Got it. Thanks!

I would appreciate another look...

Kind regards!

thibsy (approved):
Thx a lot @klees for implementing the factory! I like the new naming and structure. I look forward to using it.

matthiaskunkel:
Jour Fixe, 03 FEB 2025: Richard notified us about this PR and asks all developers for feedback or questions.

klees:
Thanks for the feedback so far. I'm merging this now and going to present it on the Dev-Conf.

Kind regards!

PR #8097, opened 2024-09-23
Exercise: Fix uninitialised GUI services

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=41992


alex40724:
This does not seem to be the complete fix, I am on it...

PR #8101, opened 2024-09-23
Export: Remove One More Check For Type Match

kergomard:
Hi @smeyer-ilias 

This might just have been forgotten. See: https://mantis.ilias.de/view.php?id=42140

Best,
@kergomard


smeyer-ilias (approved):
Thanks @kergomard for finxing this

PR #8104, opened 2024-09-24
LSO: 42026, init class vars in import

nhaagen:
https://mantis.ilias.de/view.php?id=42026


klees:
This seems to be fine in ILIAS 9+ already, feel free to open a new report if that's not the case.

PR #8114, opened 2024-09-27
trunk Authentication:0041779 add a session max idle set objective

daniwe4:
based on https://github.com/ILIAS-eLearning/ILIAS/pull/8056

### DIFF
--- a/components/ILIAS/setup_/README.md
+++ b/components/ILIAS/setup_/README.md
@@ -561,3 +561,10 @@ are printed bold**, all other fields might be omitted. A minimal example is
     * *deletion_unit* (type: string) possible values are `days`, `weeks`, `months`, `years`
     * *deletion_value* (type: string or number) depending on `deletion_unit` possible values are `days max 31`, `weeks max 52`, `months max 12`, `years no max`
     * *deletion_time* (type: string) with format `HH:MM e.g. 23:30`
+* *authentication* (type: object)
+  ```
+  "authentication" : {
+    "session_max_idle": 1800
+  }
+  ```
+  * *session_max_idle* (type: number) maximum session idle (in minutes)
--- a/components/ILIAS/App/tests/RootFolderTest.php
+++ b/components/ILIAS/App/tests/RootFolderTest.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 namespace ILIAS\App\tests;
 
--- a/components/ILIAS/Authentication/classes/Setup/class.ilAuthenticationSetupAgent.php
+++ b/components/ILIAS/Authentication/classes/Setup/class.ilAuthenticationSetupAgent.php
@@ -25,23 +25,49 @@ class ilAuthenticationSetupAgent implements Setup\Agent
 {
     use Setup\Agent\HasNoNamedObjective;
 
+    protected const DEFAULT_SESSION_EXPIRE = 30;
+
+    public function __construct(protected Refinery\Factory $refinery)
+    {
+    }
+
     public function hasConfig(): bool
     {
-        return false;
+        return true;
     }
 
     public function getArrayToConfigTransformation(): Refinery\Transformation
     {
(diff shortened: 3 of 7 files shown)


daniwe4:
Hello,

to continue the discussion from https://github.com/ILIAS-eLearning/ILIAS/pull/8056.

Would it be an idea to iterate over the file system and use the path to decide whether it is the Apache/Nginx php.ini. Then we could read it using PHP and we don't need a curl/wget call.

Unfortunately I don't know exactly where the Windows php.ini is located. I would need your knowledge.

Code perhaps, can look like this
```php
function rsearch($folder, $regPattern) {
    $dir = new RecursiveDirectoryIterator($folder);
    $ite = new RecursiveIteratorIterator($dir);
    $files = new RegexIterator($ite, $regPattern, RegexIterator::MATCH);
    $fileList = array();
    foreach($files as $file) {
        $fileList[] = $file->getPathName();
    }
    return $fileList;
}

$pathes = rsearch("/etc/php/", "#php\.ini#");
``

**Comment by mjansenDatabay** [comment]
Hi @daniwe4 ,

IMO this will not work because of ...

1) `open_basedir`: Limit the files that PHP can be accessed by PHP to the specified directory-tree, including the file itself...
2) I am afraid that it cannot be assumed that the "Web php.ini" can be found in a defined/known location in different OS (Windows, macOS, Ubuntu, Debian, ...) and different/multiple PHP versions.

Maybe we could go on with Stefan's suggestion in #8056:

> It might be possible to use the ILIAS HTTP path instead of localhost. However, in that case, the IP restriction from secure.md might no longer apply (in a reverse proxy scenario).

IMO the `ilSessionMaxIdleIsSetObjective` already defines the necessary precondition to ensure the `http_path` is set/defined in our `ilias.ini.php`. I suggest to also take `ilHttpConfigStoredObjective` into consideration as a further precondition, since our "Proxy Settings" are read/stored in this objective.
Maybe we could then use **this** `http_path` for the HTTP(S) communication and make the code in our objective as robust as possible by using "PHP curl" (see: `ilCurlConnection`) and, if it is not available, `file_get_contents` as a fallback. Of course this might fail as well, as already mentioned in #8056.

```php
// ...
        if ($we_cannot_reliably_determine_cookie_liftime) {
            $message = "ILIAS could not determine the value for 'session.cookie_lifetime' in your php.ini to check whether it comlies with our expection to ensure a proper session handling. Dou you like to continue, anyway?";

            if (!$admin_interaction->confirmOrDeny($message)) {
                throw new Setup\NoConfirmationException($message);
            }
        }
// ...
```

What I personally cannot confirm/falsify is Stefan's doubt regarding the IP restrictions:

> However, in that case, the IP restriction from secure.md might no longer apply (in a reverse proxy scenario).

I personally cannot find any conflicting guidelines/rules in our "secure.md".

As long as this is all "optional" and we give administrators the option to continue without this value, I don't see huge problems for a `trunk`-based integration.

---

As I said in the other PR, I will also ask my administrator colleagues.

Best regards,
Michael

daniwe4:
Hello @mjansenDatabay ,

Thank you very much for the detailed explanation. I'll try to implement the whole thing like this. 

Kind regards,
Daniel

daniwe4:
Hello @mjansenDatabay, @pascalseeland  and @klees ,

i implemented your suggestions and made 'ilProxySettings.php' and 'ilCurlConnection' fit for DI so that I could use them in the setup process. 
I hope it fits like that. 

Kind regards,
Daniel

mjansenDatabay (approved):

components/ILIAS/setup_/README.md
```diff
@@ -561,3 +561,10 @@ are printed bold**, all other fields might be omitted. A minimal example is
     * *deletion_unit* (type: string) possible values are `days`, `weeks`, `months`, `years`
     * *deletion_value* (type: string or number) depending on `deletion_unit` possible values are `days max 31`, `weeks max 52`, `months max 12`, `years no max`
     * *deletion_time* (type: string) with format `HH:MM e.g. 23:30`
+* *authentication* (type: object)
+  ```
+  "authentication" : {
+    "session_max_idle": 1500
```
mjansenDatabay:
```suggestion
    "session_max_idle": 1800
```

mjansenDatabay (approved):
Hi @daniwe4 ,

LGTM! I just added one more inline comment (default value in the docs).

I like the chosen approach with the dynamically created PHP script.

Best regards
Michael

components/ILIAS/setup_/README.md
daniwe4: Done!

PR #8117, opened 2024-09-30
LSO: 421621 fix control structure

nhaagen:
https://mantis.ilias.de/view.php?id=42162

remove setCmd, re-wire GUI calls


klees:
Picked to trunk.

PR #8118, opened 2024-09-30
PRG: 42037, check for target_ref_id as well as for access

nhaagen:
https://mantis.ilias.de/view.php?id=42037


klees:
@nhaagen Thanks! Please provide fixes for 8 and 9, if applicable.

PR #8122, opened 2024-09-30
T&A 42182 Fix scoring datetime handling by using the default server timezone

lukas-heinrich:
This PR adresses a bug where a UTC DateTime is written into the database but needs to be handled as same as configured server timezone

 This is related to mantis issue: https://mantis.ilias.de/view.php?id=42182

### DIFF
--- a/Modules/Test/classes/class.ilObjTest.php
+++ b/Modules/Test/classes/class.ilObjTest.php
@@ -1337,7 +1337,7 @@ class ilObjTest extends ilObject implements ilMarkSchemaAware, ilEctsGradesEnabl
             $this->setTestId($data->test_id);
 
             if ($data->author) {
-                if(strlen($this->getAuthor()) == 0) {
+                if (strlen($this->getAuthor()) == 0) {
                     $this->saveAuthorToMetadata($data->author);
                 }
                 $this->setAuthor($data->author);
@@ -1978,17 +1978,14 @@ class ilObjTest extends ilObject implements ilMarkSchemaAware, ilEctsGradesEnabl
     /**
     * Gets the reporting date of the ilObjTest object
     *
-    * @return string The reporting date of the test of an empty string (=FALSE) if no reporting date is set
+    * @return DateTimeImmutable|null The reporting date of the test or null (=FALSE) if no reporting date is set
     * @access public
     * @see $reporting_date
+    * @depracated Use Score Settings instead
     */
-    public function getReportingDate(): ?string
+    public function getReportingDate(): ?DateTimeImmutable
     {
-        $date = $this->getScoreSettings()->getResultSummarySettings()->getReportingDate();
-        if ($date) {
-            $date = $date->format('YmdHis'); //legacy-reasons ;(
-        }
-        return $date;
+        return $this->getScoreSettings()->getResultSummarySettings()->getReportingDate();
     }
 
     /**
@@ -6121,14 +6118,9 @@ class ilObjTest extends ilObject implements ilMarkSchemaAware, ilEctsGradesEnabl
     {
         $total = $this->evalTotalPersons();
         if ($total > 0) {
-            if ($this->getReportingDate()) {
-                if (preg_match("/(\d{4})(\d{2})(\d{2})(\d{2})(\d{2})(\d{2})/", $this->getReportingDate(), $matches)) {
-                    $epoch_time = mktime($matches[4], $matches[5], $matches[6], $matches[2], $matches[3], $matches[1]);
-                    $now = time();
-                    if ($now < $epoch_time) {
-                        return true;
-                    }
-                }
+            $reporting_date = $this->getScoreSettings()->getResultSummarySettings()->getReportingDate();
+            if ($reporting_date !== null) {
+                return $reporting_date <= new DateTimeImmutable('now', new DateTimeZone('UTC'));
             }
             return false;
         } else {
--- a/Modules/Test/classes/class.ilObjTestGUI.php
+++ b/Modules/Test/classes/class.ilObjTestGUI.php
@@ -2814,9 +2814,7 @@ class ilObjTestGUI extends ilObjectGUI implements ilCtrlBaseClassInterface
                 $info->addProperty(
                     $this->lng->txt('tst_score_reporting_date'),
                     ilDatePresentation::formatDate(new ilDateTime(
-                        $reporting_date
-                            ->setTimezone(new DateTimeZone($this->user->getTimeZone()))
-                            ->format('YmdHis'),
+                        $reporting_date->format('YmdHis'),
                         IL_CAL_TIMESTAMP,
                         $reporting_date->getTimezone()->getName()
                     ))
--- a/Modules/Test/classes/class.ilObjTestSettingsScoringResultsGUI.php
+++ b/Modules/Test/classes/class.ilObjTestSettingsScoringResultsGUI.php
@@ -106,7 +106,7 @@ class ilObjTestSettingsScoringResultsGUI extends ilTestSettingsGUI
 
         if ($templateId) {
             $this->settingsTemplate = new ilSettingsTemplate(
-                (int)$templateId,
+                (int) $templateId,
                 ilObjAssessmentFolderGUI::getSettingsTemplateConfig()
             );
         }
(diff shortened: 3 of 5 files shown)


dsstrassner:
@thojou do you need an external Review or will you merge this yourself as Lukas has assigned the Mantis Report to you?

thojou:
Hi @mjansenDatabay,

Could you kindly review these changes? They are intended to align with the adjustments you made in a previous commit.

If everything looks good to you, I'll proceed with merging the PR.

Thanks!

mjansenDatabay:
Hi all,

using the server time zone is (IMO) exactly as wrong as storing the date/time information as a "UTC"-based datetime-string, possibly a mu better ;-).

Why? We lose the time zone information in both approaches. I have already seen (multiple) productive environments where the server time zone has changed over time.
I used "UTC" in PR #6278 to solve https://mantis.ilias.de/view.php?id=37823 , similar to other ILIAS components (e.g. the calendar component), even if "UTC" datetime-strings or "UTC"-based unix timestamps are (most of the time) only appropriate for "logging use cases", not for storing future points in time.

The best way (IMO) would be to add a new "timezone" field to the database schema and store the time zone for the set datetime string in the time zone of the user which defines the access to the rest results (`reporting_date`).

But because we are not in an ideal world a (I guess there is not much funding, if at all), I suggest normalizing the comparisons with the existing "UTC" datetime-strings, so both operands use the "UTC"-time zone.

So instead of changing the time zone in `\ilObjTestScoreSettingsDatabaseRepository::doSelect` and in `\ilObjTestSettingsResultSummary::toStorage`, I suggest to change the checks. I found three of them.

- \ilObjTestSettingsScoringResultsGUI::isScoreReportingAvailable
- \ilObjTest::canEditMarks
- \ilTestPassesSelector::isReportingDateReached

Of course you don't need to follow my suggestions.

Best regards,
Michael

kergomard:
Thanks @lukas-heinrich for the PR and @mjansenDatabay for the feedback!
I agree with what Michael said for the pragmatic side of things. I wonder though about the "ideal world": From my point of view using the server time zone is much, much worse then using UTC. At least in theory (I know the vagaries of DateTime and computers) UTC is a standardized point of reference. As far as I understand it all Linux computers always run on UTC the rest is just some sugar on top for us users. Windows is really not our problem. So from where I stand internally we should always work with values bound to UTC and only change time zone once we hand it out to the user. I understand the approach to add an explicit time zone, but from where I stand we should then set UTC explicitly and never anything else. What am I missing?

Best,
@kergomard

thojou:
Hi @mjansenDatabay, @kergomard,

Thank you for your feedback!

I completely agree that storing date information with time zone data would be a significant improvement. Currently, for reporting_date, we're not even using a UNIX timestamp but a formatted string like YmdHis.

Here's the issue: There are multiple places where a system administrator can change the time zone settings in PHP. For example, it can be configured in the server's "php.ini" file or directly in the ilias.ini.php.

In the scenario where the issue occurs, the ilias.ini.php is set to Europe/Berlin. When the reporting_date is checked (e.g., in \ilTestPassesSelector::isReportingDateReached), it's compared against a time() UNIX timestamp. The problem is that while the UNIX timestamp is in the "Europe/Berlin" time zone, the reporting_date is in "UTC". To address this, we decided to read the time zone setting from the configuration instead of relying on a hardcoded value.

I agree with @mjansenDatabay's suggestion of implementing a more robust solution, but as mentioned, funding may be required for that.

Best regards,
@thojou

kergomard:
Now I also got the ideal solution an @mjansenDatabay is completely right, we actually should save the timezone in an ideal world. I even think that we should probably provide a field to choose a timezone and prefill it with the current users timezone. The example use case would be an international organization with a centralized training unit that plans ahead for longer times and e.g. has to plan trainings in Venezuela, where the state has a certain tendency to change daylight savings impulsively.

lukas-heinrich:
Hi, thank you very much for the feedback!

I have implemented a new approach to standardize the handling of the reporting date and to ensure a correct display for the user.

kergomard:
I'm very, very sorry, to be complicated, but we should strive to be consistent across our usages in the test. Please remove all Dependence on `ilDate*` and replace it with `DateTimeImmutable`. There is no need to use the ILIAS-specific functions (anymore). As we are at it we should also get rid of all usages of `ilObjTest::getReportingDate()` (and with trunk of the function itself, but I can take care of that when porting), you can use the ScoreSettings directly (maybe with time we can reduce the dependence on the whole test object).

Sorry again for being complicated and best,
@kergomard

lukas-heinrich:
Hi @kergomard ,

Thank you for the tip. I have adjusted the relevant places so that `DateTimeImmutable` is used for the comparisons. In the two places where the time is displayed to the user, `reporting_date` is then formatted via an `ilDateTime` object and `ilDatePresentation::formatDate` displays it in the correct time zone for the user.

Please let me know if the implementation should be adapted.

kergomard (changes requested):

Now, I think, we are getting to the details ;-) , please, let me know, if I'm wrong on any of this:

- If I'm not mistaken [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/8122/files#diff-852ff402e4f9a1f0125a36bb16167adabe95713f43631bac7c7530209d161f3fR263) is going to give you the Date in the Timezone of the server. I'm pretty sure this is not, what we want. I think we will want to specify that we want the time in UTC, so `new DateTimeImmutable('now', new DateTimeZone('UTC'))`.
- I also don't think [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/8122/files#diff-1cb659c07cfa02fb1a744e11d5b0cf83003aa4d4a493c7f0943005f089d4cfffR2819) is right. The timezone of the reporting date should always be UTC, so why setting it here. I implemented a slightly different solution this morning for the info page [here](https://github.com/ILIAS-eLearning/ILIAS/commit/d879eeabe8059f76a1d48598bc935ccce83c5582#diff-1cb659c07cfa02fb1a744e11d5b0cf83003aa4d4a493c7f0943005f089d4cfffR2816).
- A nitpick, I know: Could you please change all double quotes (") to single quotes(') unless you are quoting a string containing variables or escape-sequences? This gives the double quotes a clear signalling value, as they signify: Take care there is something interesting happening in this string.
- The remaining uses of  `ilDateTime` for ILIAS 8 are ok, I will remove them when porting to newer versions.

Thanks again and best,
@kergomard

lukas-heinrich:
Many thanks for the feedback @kergomard !

Regarding the first point, we can implement it in such a way that we explicitly specify the time zone. However, this is not absolutely necessary, as the operators of DateTimeImmutable can correctly compare objects with different time zones. So for us it doesn't matter which time zone the server has (e.g. UTC or Europe/Berlin, I have also tested it with time zones from Asia).

I have set the time zone [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8122/files#diff-1cb659c07cfa02fb1a744e11d5b0cf83003aa4d4a493c7f0943005f089d4cfffR2819) because otherwise it would be lost when converting to the 'YmdHis' format. If we do not specify the time zone explicitly at this point, ilDateTime creates an object with the server's default time zone (and if this is not UTC, we end up with the original problem again). We could specify UTC explicitly here, I decided against it and used a getter so that the timezone is not hardcoded in too many places. Please correct me if I am wrong or if a different implementation is desired here.

I will replace the quotes, I forgot to check them again before committing.

Thanks and best regards,
@lukas-heinrich

kergomard:
Thank you very much for the Information @lukas-heinrich !
And clearly: You are right and I will adapt my usage above accordingly!

- I would still suggest to explicitly set the timezone [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8122/files#diff-852ff402e4f9a1f0125a36bb16167adabe95713f43631bac7c7530209d161f3fR263). The reason for it being that in this way we can be sure to always deal with UTC as long as we are inside the program (not generating output). I know, in this case the DateTime variable doesn't go far, but by doing this we still form the habit of working with UTC times internally (unless, as @mjansenDatabay points out above, we are dealing with appointment and we can ensure that we always have a user-defined timezone).
- I would then leave [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/8122/files#diff-1cb659c07cfa02fb1a744e11d5b0cf83003aa4d4a493c7f0943005f089d4cfffR2819) as it is now (ie. getting the timezone out of the DateTime-variable).

@kergomard

lukas-heinrich:
Hi @kergomard ,

I totally understand the handling of UTC at the point and have explicitly specified the time zone at the relevant sections. I also squashed the git commits to keep a clean history.

Best,
@lukas-heinrich

PR #8123, opened 2024-09-30
Background Tasks: Async background tasks fail silently if soap client return false

fneumann:
See https://mantis.ilias.de/view.php?id=42183

This PR ads a fallback to SyncTaskManager if the soap client returns false.

### DIFF
--- a/src/BackgroundTasks/Implementation/TaskManager/AsyncTaskManager.php
+++ b/src/BackgroundTasks/Implementation/TaskManager/AsyncTaskManager.php
@@ -23,6 +23,7 @@ use ILIAS\BackgroundTasks\Implementation\Bucket\State;
 use ILIAS\BackgroundTasks\Implementation\Tasks\UserInteraction\UserInteractionRequiredException;
 use ILIAS\BackgroundTasks\Implementation\Tasks\UserInteraction\UserInteractionSkippedException;
 use ILIAS\BackgroundTasks\Task\UserInteraction;
+use ILIAS\Export\ImportStatus\Exception\ilException;
 
 class AsyncTaskManager extends BasicTaskManager
 {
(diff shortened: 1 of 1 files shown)


src/BackgroundTasks/Implementation/TaskManager/AsyncTaskManager.php
```diff
@@ -55,9 +55,14 @@ public function run(Bucket $bucket): void
             )
             : '';
         try {
-            $soap_client->call(self::CMD_START_WORKER, array(
+            $result = $soap_client->call(self::CMD_START_WORKER, array(
                 $session_id . '::' . $client_id,
             ));
+            if ($result === false) {
+                $DIC->logger()->root()->info("[BT] Calling Webserver returned false, fallback to sync version");
```
mjansenDatabay: Shouldn't we use `warning` or `error` in this case?

fneumann: This would create many entries, if soap is not active. @chfsx may decide this.

chfsx: I suggest to use the component logger for BackgroundTask instead of root and then use warning.

mjansenDatabay:
One thing that I discovered in projects with time consumer async. background tasks.

1. It could be worth setting a response timeout of `0` by calling `$soap_client->setResponseTimeout(0);` (in `\ILIAS\BackgroundTasks\Implementation\TaskManager\AsyncTaskManager::run`).
2. A `ignore_user_abort(true);` could be set in `\ilSoapBackgroundTasksAdministration::runAsync`

As far as I can see, the SOAP call is "Fire & Forget" anyway. 

This approach will ensure some of the `false` or `Exception` cases will not occur anymore and the client/caller (ILIAS itself) does not unnecessarily wait.

fneumann:
I reverted the fallback to sync execution if the soap call returns false. This is the case for time-consuming tasks and and would result in a double execution. The initial reason for this PR (failed call for root user due to missing acceptance of legal documents) is solved by https://github.com/ILIAS-eLearning/ILIAS/commit/38d246e153c2f13be49ac9c8ddc46b934d25e0b0.

Nevertheless, I think this PR is still useful regarding the change of the logger and the pre-check for activated SOAP.

chfsx:
Thanks a lot, I squashed to one commit and picked to other branches as well

PR #8130, opened 2024-10-01
Container: Fix Clipboard

kergomard:
See: https://mantis.ilias.de/view.php?id=4214


alex40724:
@kergomard It would be good to fix the real cause here asap, which is the way the dropzones trying to identify their forms, and their activation in the object service. Otherwise this will produce more issues in the future.

PR #8132, opened 2024-10-01
IAss: Migration, do not throw on missing file

nhaagen:
https://mantis.ilias.de/view.php?id=42099


nhaagen:
missed @klees comment on https://mantis.ilias.de/view.php?id=42099#c108057 , I close this.

PR #8134, opened 2024-10-01
LegalDocuments: Fix wrong Soap check

lscharmer:
This PR fixes Mantis Bug: https://mantis.ilias.de/view.php?id=42168
The boolean query to check if the ToS or Dpro need to be accepted via Soap is fixed.


mjansenDatabay (approved):
Thx  @lscharmer

PR #8138, opened 2024-10-01
PRG: abandon tree in manage view

nhaagen:
https://docu.ilias.de/go/wiki/wpage_7476_1357


klees:
Failing checks are due to PHP-CS-Fixer trying to check missing files...

PR #8141, opened 2024-10-01
[FIX] 0042151: User is shown whoops instead of Dashboard after witching to tile view

chfsx:
@kergomard this PR makes the generation of tile image flavours more error proof and fixes mantis issue 0042151


kergomard:
Thank you very much @chfsx ! Merged and picked to trunk.

PR #8143, opened 2024-10-01
Content Style: Remove Paddings on Question Content

kergomard:
This gets rid of unnecessary padding and margins in questions.

See: https://mantis.ilias.de/view.php?id=42120


Amstutz:
Hi @kergomard. this Looks fine with me from a CSS perspective. However, I pass it on to @alex40724 since this is also content style related.

PR #8145, opened 2024-10-02
Course/SOAP: Fix that title defaults to `dummy` when using `(add|update)Course`

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=42079


swischniak:
Hi @smeyer-ilias. Any info on the timeline here? Looking forward to a merge/fix.

PR #8147, opened 2024-10-04
BackgroundTasks: Use "Fire & Forget" approach when starting workers as…

mjansenDatabay:
…ynchronously

This PR suggests making the start of asynchronous workers for background tasks really asynchronous to ensure a "Fire & Forget" approach. This is done by setting the "Response Timeout" to `0` and by adding a `ignore_user_abort(true);`) on the server-side (SOAP).


chfsx:
Thanks a lot! I seen no problem to merge this in trunk, but maybe this could be helpful in 8-9 as well. do you have any experience with this in productive releases?

mjansenDatabay:
> Thanks a lot! I seen no problem to merge this in trunk, but maybe this could be helpful in 8-9 as well. do you have any experience with this in productive releases?

Hi @chfsx ,

thanks for the question: We use this in an ILIAS 6.x installation (productive system) where this solves issues with a high usage of async. background tasks, each handling complex "Learning Progress Re-calculations".

Best regards
Michael

PR #8148, opened 2024-10-04
UI: Fix Missing min in modal.js

kergomard:
Dear coordinators

If I'm getting it right, there is a small error in the renderer for the table as the include for the modal.js lacks a min ;-) .

Best,
@kergomard


Amstutz:
thx a lot @kergomard

PR #8158, opened 2024-10-08
Dashboard: Fix key of sliced `$data` page

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=42238


iszmais:
@mjansenDatabay Thank you for the contribution!
There is already a PR for that change: https://github.com/ILIAS-eLearning/ILIAS/pull/7857

Unfortunatly, this is not enough to ensure the type safety of the array, this just preserves the value.
I will provide a solution without array slice and take this as an announcement to the jour fixe.

Greetings,
@iszmais

PR #8161, opened 2024-10-08
10/UI/Input error-message accessibility

catenglaender:
# Issue

The form and input fields don't point to corresponding error messages with aria-describedby but the other way around. This leads to screen reader not reading out the error messages when it is expected by the user.

# Changes

Adjusted the html templates so both field and form error messages are referred to with aria-describedby by the parent.

Added an "Error:" label only visible to screen readers before error text as done in other very accessible ui frameworks.

image

### DIFF
--- a/components/ILIAS/UI/src/Implementation/Component/Input/Container/Form/Renderer.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Input/Container/Form/Renderer.php
@@ -116,7 +116,13 @@ class Renderer extends AbstractComponentRenderer
     {
         if (null !== ($error = $component->getError())) {
             $tpl->setCurrentBlock("error");
+            $error_id = $this->createId();
             $tpl->setVariable("ERROR", $error);
+            $tpl->setVariable("ERROR_ID", $error_id);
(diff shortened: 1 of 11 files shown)


Amstutz:
Hi @catenglaender

There is just one question:
- [ ] Role Alert: Why did you remove it? Is this not needed anymore? Did this cause bad side effects?

Thx a lot
@thibsy and @Amstutz

catenglaender:
Hey @Amstutz, well spotted.
This was indeed part of an investigation and discussion with @klees. Our interpretation of the specs is that the alerts of our form are not an real-time interruption as a result of an immediate interaction and therefor don't warrant the alert status. See: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role
> It should not be used on HTML that the user hasn't interacted with. For example, if a page loads with multiple visible alerts scattered throughout, the alert role should not be used, as the messages were not dynamically triggered.

So we would recommend to not use the alert role unless we give immediate alerts without a page reload. If we would alert e.g. on blur event this would be a different story.

Amstutz:
thx a lot for the explanation!

PR #8165, opened 2024-10-09
LH: new data table

c-knof:


oliversamoila:
Hey @c-knof and @katringross,
thank you very much for the improvements and the removal of the LegacyUI.

oliversamoila:
Hello @c-knof and @katringross,
Shouldn't the ‘Page translation enabled’ label either be integrated into the status column or be displayed in a separate column?
> image

PR #8166, opened 2024-10-09
[FIX] Auth: pass arguments as correct data-type.

thibsy:
Hi @pascalseeland,

I have noticed there are some missing type-casts, which lead to a fatal error on login requests.

Kind regards,
@thibsy


mjansenDatabay (approved):

mjansenDatabay:
Thx @thibsy !

PR #8167, opened 2024-10-09
Refactoring login/logout screen PageEditor usage

playamongthestars:
This PR removed code duplicates  for the login/logout screen design feature


components/ILIAS/Authentication/classes/class.ilAuthPageEditorSettings.php
```diff
@@ -18,7 +18,7 @@
 
 declare(strict_types=1);
 
-class ilAuthLogoutPageEditorSettings
+class ilAuthPageEditorSettings
 {
     private static ?self $instance = null;
```
mjansenDatabay: If we support multiple contexts, we need an array of `$instances`.

```suggestion
/** @var array<value-of<AuthPageEditorContext>, self> */
private static array $instances = [];
```

components/ILIAS/Authentication/classes/class.ilAuthPageEditorSettings.php
```diff
@@ -29,19 +29,20 @@ class ilAuthLogoutPageEditorSettings
...
+        $this->storage = new ilSetting($context->value);
 
         $this->read();
     }
 
-    public static function getInstance(): self
+    public static function getInstance(AuthPageEditorContext $context): self
     {
-        return self::$instance ?? (self::$instance = new self());
+        return self::$instance ?? (self::$instance = new self($context));
```
mjansenDatabay:
```suggestion
        return self::$instances[$context->value] ?? (self::$instances[$context->value]  = new self($context));
```

components/ILIAS/Authentication/classes/class.ilObjAuthSettingsGUI.php
```diff
@@ -846,23 +846,13 @@ public function executeCommand(): void
                 $this->ctrl->forwardCommand($cas_settings);
                 break;
 
-            case 'ilauthloginpageeditorgui':
+            case 'ilauthpageeditorgui':
```
mjansenDatabay:
```suggestion
            case strtolower(ilAuthPageEditorGUI::class):
```

components/ILIAS/Authentication/classes/class.ilObjAuthSettingsGUI.php
```diff
@@ -1001,24 +991,30 @@ public function setSubTabs(string $a_tab): void
...
-                    $this->ctrl->getLinkTargetByClass(ilAuthLogoutBehaviourGUI::class, ''),
-                    ''
+                    $this->ctrl->getLinkTargetByClass(
+                        ilAuthPageEditorGUI::class,
+                        'showLoginEditor'
+                    )
                 );
             }
 
             if ($this->access->checkAccess('write', '', $this->object->getRefId())) {
```
mjansenDatabay: IMO we could remove this `if` statement and put the code into the `if` statement above.

components/ILIAS/Init/classes/class.ilStartUpGUI.php
```diff
@@ -959,7 +959,7 @@ private function substituteLoginPageElements(
 
     private function getLoginPageEditorHTML(): string
     {
-        $lpe = ilAuthLoginPageEditorSettings::getInstance();
+        $lpe =ilAuthPageEditorSettings::getInstance(AuthPageEditorContext::LOGIN);
```
mjansenDatabay:
```suggestion
        $lpe = ilAuthPageEditorSettings::getInstance(AuthPageEditorContext::LOGIN);
```

components/ILIAS/Init/classes/class.ilStartUpGUI.php
```diff
@@ -1379,7 +1379,7 @@ private function doLogout(): void
         ilUtil::setCookie("ilClientId", "");
 
         $this->ctrl->setParameter($this, 'lang', $user_language);
-        $this->ctrl->setParameter($this, 'client_id', $client_id);
+        $this->ctrl->setParameter($this, 'client_id', CLIENT_ID);
```
mjansenDatabay: Do we need this change?

components/ILIAS/Authentication/classes/class.ilObjAuthSettingsGUI.php
```diff
@@ -846,23 +846,13 @@ public function executeCommand(): void
...
-                break;
-
-            case 'ilauthlogoutpageeditorgui':
-
-                $this->setSubTabs('authSettings');
-                $this->tabs_gui->setTabActive('authentication_settings');
-                $this->tabs_gui->setSubTabActive('logout_editor');
-
-                $lpe = new ilAuthLogoutPageEditorGUI($this->object->getRefId());
+                $lpe = new ilAuthPageEditorGUI($this->object->getRefId(), AuthPageEditorContext::LOGIN);
```
mjansenDatabay: Is it correct to always pass a static`AuthPageEditorContext::LOGIN` context into the GUI class here, where we had a separation of these two contexts before?

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
```diff
@@ -153,9 +187,60 @@ private function forwardToPageObject(): void
...
+            case 'deactivate':
+                $this->deactivate();
+                break;
+
+            case 'activate':
+                $this->activate();
+                break;
+
+            case 'edit':
+                $this->ctrl->setParameter($this, 'loginpage_languages_key', (string) current($keys));
```
mjansenDatabay: I suggest to rename `loginpage_languages_key` to `authpage_languages_key` (or something similar).

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
```diff
@@ -126,20 +147,33 @@ private function forwardToPageObject(): void
 
         $this->lng->loadLanguageModule('content');
 
-        if (!ilLogoutPage::_exists('aout', $this->key)) {
-            $new_page_object = new ilLogoutPage();
-            $new_page_object->setParentId($this->key);
-            $new_page_object->setId($this->key);
-            $new_page_object->createFromXML();
+        if ($context->value == AuthPageEditorContext::LOGIN->value) {
```
mjansenDatabay:
```suggestion
        if ($context === AuthPageEditorContext::LOGIN) {
```

components/ILIAS/Authentication/classes/PageLanguages/PageLanguagesOverviewTable.php
```diff
@@ -18,23 +18,39 @@
...
 use ilLanguage;
 use ilCtrlInterface;
-use ilAuthLogoutPageEditorSettings;
+use AuthPageEditorContext;
+use ilAuthPageEditorSettings;
 
-class LogoutPageLanguagesOverviewTable implements UI\Component\Table\DataRetrieval
+class PageLanguagesOverviewTable implements UI\Component\Table\DataRetrieval
 {
+    private static array $context_lng_vars =
```
mjansenDatabay: We should add a PhpDoc describing the array shape:

`/** @var array<value-of<AuthPageEditorContext>, array{'pages' string, 'page': string, 'param_namspace' string, 'table_action': string}}>*/`

components/ILIAS/Authentication/classes/PageLanguages/PageLanguagesOverviewTable.php
```diff
@@ -18,23 +18,39 @@
...
+                'page' => 'login_page',
+                'param_namespace' => 'loginpage',
+                'table_action' => 'handleLoginPageActions'
+            ],
+            AuthPageEditorContext::LOGOUT->value => [
+                'pages' => 'logout_pages',
+                'page' => 'logout_page',
+                'param_namespace' => 'logoutpage',
+                'table_action' => 'handleLogoutPageActions'
+            ]
```
mjansenDatabay: And: Instead of using this array, you could add a public method to the `Enum`, e.g.

```php
public function getLanguageIdentifierFor(string $case): string
{
    // Use a match/switch and return the language key based on the $case
}
```

With this approach, we could remove the `static` variable.

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
```diff
@@ -39,8 +39,9 @@ class ilAuthLogoutPageEditorGUI
     private ?int $key = null;
     private GUIService $content_style_gui;
     private int $ref_id;
+    private AuthPageEditorContext $context;
 
-    public function __construct(int $a_ref_id)
+    public function __construct(int $a_ref_id, AuthPageEditorContext $context)
```
mjansenDatabay: Do we really need this as a parameter if we overwrite the variable in `executeCommand` below?

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
```diff
@@ -104,10 +120,15 @@ public function executeCommand(): void
         }
     }
 
-    private function forwardToPageObject(): void
+    private function forwardToPageObject(AuthPageEditorContext $context): void
     {
+        $request_value = match ($context) {
+            AuthPageEditorContext::LOGIN => 'loginpage_languages_key',
+            AuthPageEditorContext::LOGOUT => 'logoutpage_languages_key',
+        };
```
mjansenDatabay: Again, we could/ should think about moving this into the `Enum`, like the language variable retrieval.

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
```diff
@@ -153,9 +187,60 @@ private function forwardToPageObject(): void
...
+        $action = $this->http->wrapper()->query()->retrieve(
+            'loginpage_languages_action',
+            $this->refinery->byTrying([
+                $this->refinery->kindlyTo()->string(),
+                $this->refinery->always('')
+            ])
+        );
+
+        $keys = $this->http->wrapper()->query()->retrieve(
+            'loginpage_languages_key',
```
mjansenDatabay: I suggest to rename `loginpage_languages_key` to `authpage_languages_key` (or something similar).

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
```diff
@@ -153,9 +187,60 @@ private function forwardToPageObject(): void
...
+    private function showLogoutEditor(): void
+    {
+        $this->context = AuthPageEditorContext::LOGOUT;
         $this->showIliasEditor();
     }
 
+    private function handleLoginPageActions(): void
+    {
+        $action = $this->http->wrapper()->query()->retrieve(
+            'loginpage_languages_action',
```
mjansenDatabay: I suggest to rename `loginpage_languages_action` to `authpage_languages_action` (or something similar).

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
```diff
@@ -199,8 +284,13 @@ private function handleLogoutPageActions(): void
      */
     private function getLangKeysToUpdate(): array
     {
+        $request_value = match ($this->context) {
```
mjansenDatabay: Again, we could/ should think about moving this into the `Enum`, like the language variable retrieval.

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
```diff
@@ -126,20 +147,33 @@ private function forwardToPageObject(): void
...
+                $new_page_object = new ilLoginPage();
+                $new_page_object->setParentId($this->key);
+                $new_page_object->setId($this->key);
+                $new_page_object->createFromXML();
+            }
+
+            $this->ctrl->setReturnByClass(ilLoginPageGUI::class, 'edit');
+            $page_gui = new ilLoginPageGUI($this->key);
+        } else {
+            if (!ilLogoutPage::_exists('aout', $this->key)) {
```
mjansenDatabay: This should be an `elseif` statement.

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
```diff
@@ -83,15 +85,29 @@ public function __construct(int $a_ref_id)
...
+            case strtolower(ilLoginPageGUI::class):
+                $this->tabs->clearTargets();
+                $this->tabs->setBackTarget(
+                    $this->lng->txt('back'),
+                    $this->ctrl->getLinkTarget($this, 'show')
+                );
+
+                $this->context = AuthPageEditorContext::LOGIN;
+                if (strtolower($this->redirect_source ?? '') !== strtolower(ilInternalLinkGUI::class)) {
+                    $this->forwardToPageObject($this->context);
```
mjansenDatabay: Do we really need to pass this as an argument, if we already have the context stored in `$this->context`?

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
```diff
@@ -83,15 +85,29 @@ public function __construct(int $a_ref_id)
...
                 $this->tabs->clearTargets();
                 $this->tabs->setBackTarget(
                     $this->lng->txt('back'),
                     $this->ctrl->getLinkTarget($this, 'show')
                 );
 
+                $this->context = AuthPageEditorContext::LOGOUT;
                 if (strtolower($this->redirect_source ?? '') !== strtolower(ilInternalLinkGUI::class)) {
-                    $this->forwardToPageObject();
+                    $this->forwardToPageObject($this->context);
```
mjansenDatabay: Do we really need to pass this as an argument, if we already have the context stored in `$this->context`?

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
```diff
@@ -104,10 +120,15 @@ public function executeCommand(): void
         }
     }
 
-    private function forwardToPageObject(): void
+    private function forwardToPageObject(AuthPageEditorContext $context): void
```
mjansenDatabay: Do we really need to introduce a new parameter if we already have the context stored in `$this->context`?

mjansenDatabay (changes requested):
Hi @nmatuschek ,

thanks a lot for this improvement, removing three files is really a step forward.

Nevertheless, I left some inline remarks.

- [x] Please run the `php-cs-fixer` for the changed files.

Best regards,
Michael

components/ILIAS/Init/classes/class.ilStartUpGUI.php
playamongthestars: Yes, we need this change, because $client_id is undefined.

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
playamongthestars: No, this would cause an unwanted behaviour

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
```diff
@@ -39,8 +39,10 @@ class ilAuthLogoutPageEditorGUI
     private ?int $key = null;
     private GUIService $content_style_gui;
     private int $ref_id;
+    /** @var array<value-of<AuthPageEditorContext>, array{'pages' string, 'page': string, 'param_namspace' string, 'table_action': string}}> */
```
mjansenDatabay: This seems to be wrong here.

components/ILIAS/Authentication/classes/class.ilAuthPageEditorGUI.php
mjansenDatabay: You are right!

mjansenDatabay (approved):

PR #8171, opened 2024-10-09
[FIX - 42249] fixed display issue in detailed results view of cloze questions

lzehnder:
This pull request contains a code change provided by a customer of ours to solve the presentation issue of cloze question fields being too wide in the results view.

This issue has been reported in the following mantis ticket:
https://mantis.ilias.de/view.php?id=42249


kergomard (commented):
Thank you very much for the PR @lzehnder !
Even if I don't think this is how it should be, let us try for the time being. This also introduces additional accessibility errors, but as these are the same we have in the presentation of the question itself, we need to fix them for both places. We will need another solution in the future, but as this would make things work equally as in the longmenu question, we should try.

Merged and picked to ILIAS 9 and trunk.

Best,
@kergomard

PR #8173, opened 2024-10-09
Add `@rollup/plugin-commonjs` npm package

thibsy:
This PR adds `@rollup/plugin-commonjs` as `npm` package.

Usage:
- None yet, file input to come.

Wrapped By:
- Not applicable.

Reasoning:
- This package handles: 
	- It converts CommonJS, or JavaScript code prior to ES6, into ES6 compatible code. It is essentially the reversed process of the `@babel/preset-env` package.
- Its required because:
	- Rollup.js, the module bundler we are using, demands that the code which should be bundled is compatible with ES6. However, since we are using third party libraries in our code, it sometimes happens that a library does not provide a ES6 compatible format. This is when we need to transform the code before we can bundle it. This plugin is usually only used together with the `@rollup/plugin-node-resolve` package.

Maintenance:
- The package is part of Rollup.js' [plugin repository](https://github.com/rollup/plugins), which is actively maintained and receives yearly major releases and ~monthly to ~weekly minor/patch updates.

Links:
- Package manager: https://www.npmjs.com/package/@rollup/plugin-commonjs
- GitHub: https://github.com/rollup/plugins/tree/master/packages/commonjs/#readme
- Documentation: https://rollupjs.org/introduction/#importing-commonjs

### DIFF
--- a/docs/development/js-bundling.md
+++ b/docs/development/js-bundling.md
@@ -135,7 +135,7 @@ compatible with `rollup.js` by default. To be able to transpile CommonJS module 
 
 ```javascript
 import copyright from './scripts/Copyright-Checker/copyright';
-import resolve from '@rollup/plugin-node-resolve';
+import nodeResolve from '@rollup/plugin-node-resolve';
 import commonjs from '@rollup/plugin-commonjs';
 
 export default {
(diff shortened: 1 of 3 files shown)


matthiaskunkel:
Jour Fixe, 14 OCT 2024: Dependency is accepted for ILIAS 10. Note: we do not want new dependencies that require this one.

thibsy:
I think this is not (yet) needed.

PR #8174, opened 2024-10-09
Components: Remove Artifacts

kergomard:
See: https://mantis.ilias.de/view.php?id=42254


klees:
PHP-CS-Fixer-Errors are fine, can't check non-existing files.

klees:
Thanks!

PR #8175, opened 2024-10-10
Changes function signature ilDataset::setExportDirectories(string, string) to ilDataset::initByExporter(ilXMLExporter)

schmitz-ilias:
This PR makes changes to the method ilDataset::setExportDirectories: it is renamed to ilDataset::initByExporter, and instead of accepting the export directories directly as strings, it now takes an ilXMLExporter (from which the export directories can then be obtained).

The changes by themselves do not affect functionality in any way, they basically just shift around some responsibility from consuming components to Export/Dataset. This is necessary for the forthcoming rework of [Export to use IRSS](https://docu.ilias.de/go/wiki/wpage_7758_1357): the more explicit interface between Export and consumers via Dataset requires a more explicit dependency between the two, facilitated by this signature change.

Since the changes affect ~30 components and are largely trivial, we would like to merge them directly without waiting for approval of every single involved authority.

### DIFF
--- a/components/ILIAS/Bibliographic/classes/class.ilBibliographicExporter.php
+++ b/components/ILIAS/Bibliographic/classes/class.ilBibliographicExporter.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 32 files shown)


matthiaskunkel:

PR #8177, opened 2024-10-10
UI/Fields: JS binding on wrapping fieldset

nhaagen:
https://mantis.ilias.de/view.php?id=42235

As @catenglaender points out in #8161, we use(d) an id-attribute somewhere in the template, usually on the actual form-element, to bind labels, aria and JS, which is not sufficient in all cases.

This PR strives to separate ids acording to their meaning, i.e. the label should usually point to the form-element, while the JS-binding does not necessarily have to. Also, we would like to point the whole fieldset to it's error or byline with aria-describedby (this is missing in here, but dealt with in  #8161).

Finally, after having quite some time with the screen-reader and forms, tabbing around,  I'm certain that some elements _do not_ want to point to the form-element via their label: When having a "label for=...", the tab and reader will jump to the form-element directly. This is quite nice for single inputs like text or numeric, but for grouped elements, the label and group-nature are totally omitted.
I added a tabindex to those grouped elements to make tabbing aware of their upmost level, while leaving the label-for on the more simple/single ones.

For JS-binding, the id now _always_ sits on the outer fieldset, and the JS may be aligned accordingly.

Commit [77b088a](https://github.com/ILIAS-eLearning/ILIAS/pull/8177/commits/77b088a7cfa743f0685db14392ecfdf40ba9ecb6) fixes the bug that initially started this :)


schmitz-ilias:
Hi @nhaagen,

thanks for the fix, much appreciated! Earlier this week I merged a feature that required changes in `ilMetaCopyrightListener.js`, which is causing (some of) the merge conflicts here, sorry about that.

I had a look today and adapted my changes from the feature to those you are making here, but now I'm not sure where to put that fix. Strictly speaking they should go into this PR since they depend on your KS changes to work, but that seems a bit messy. Alternatively, and maybe more convenient for you, I could just commit the fix to trunk and check in again after this PR has been merged, then you could just take `ilMetaCopyrightListener.js` out of this PR.

Any preferences?

nhaagen:
Hi @schmitz-ilias ,
I integrated the existing changes here (I hope), and would just do that again; on the other hand, I'm really looking forward to the comments concerning this PR here, and am not sure about desired changes.
Anyways, what I did to ilMetaCopyrightListener is only about detection/querying the initial objects, and that should not be hard to adopt, so probably it is good advice to just go on with your changes and make it work in the current state and adjust later? I can do that, too,  when conflicts arise.

klees (approved):
LGTM.

klees:
@Amstutz/@thibsy: Could you please move on here?

Amstutz:
thx a lot. special thx for the explanation. The failing lng check seems unrelated.

PR #8183, opened 2024-10-10
Search: do not show relevance in search results (41910)

schmitz-ilias:
As already accepted by the JF with [41910](https://mantis.ilias.de/view.php?id=41910), this PR removes ilProgressBar from Search by not showing relevance in search results anymore, and introducing a sortation view control such that one can still order the results by relevance:
object_search_results

What was not discussed with [41910](https://mantis.ilias.de/view.php?id=41910) is that relevance is not only shown when searching for objects, but also when searching for users and in Wikis and Forums.

The latter are simple enough. When searching in Wikis and Forums, the results are anyways always ordered by relevance, reordering is not possible. This PR then simply removes the column 'Relevance' (in the screenshot in a Forum, analogously for Wikis):
forum_search_results

In user searches however, the solution accepted for object searches can only be applied somewhat inelegantly. Since in user search results, all user fields (including custom ones) can be displayed, the sortation view control has to accommodate an effectively arbitrary number of fields:
user_search_results

We still believe this to be an acceptable solution, since it is self-imposed by users. When a reasonable number of columns is made visible, the view control is also in a manageable state, the issue only presents itself when a lot of columns are selected.

### DIFF
--- a/lang/ilias_en.lang
+++ b/lang/ilias_en.lang
@@ -15232,7 +15232,7 @@ search#:#search_sort_creation_date_asc#:#Oldest
 search#:#search_sort_creation_date_desc#:#Latest
 search#:#search_sort_generic_asc#:#%s, asc.
 search#:#search_sort_generic_desc#:#%s, desc.
-search#:#search_sort_relevance#:#by Relevance
+search#:#search_sort_relevance#:#By Relevance
 search#:#search_sort_title_asc#:#Alphabetically: A-Z
 search#:#search_sort_title_desc#:#Alphabetically: Z-A
 search#:#search_term_combination#:#Combination
--- a/lang/ilias_en.lang
+++ b/lang/ilias_en.lang
@@ -15156,7 +15156,6 @@ search#:#lucene_port#:#Port
 search#:#lucene_prefix_wildcard#:#Prefix Wildcard Queries
 search#:#lucene_prefix_wildcard_info#:#Supports prefix wildcard queries: ‘*LIAS’ finds ‘ILIAS’
 search#:#lucene_relevance#:#Relevance of Search Results
-search#:#lucene_relevance_short#:#Relevance
 search#:#lucene_settings_tab#:#Lucene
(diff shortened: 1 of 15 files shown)


lang/ilias_en.lang
```diff
@@ -15228,6 +15227,14 @@ search#:#search_show_inactive_user#:#Show Inactive Users
 search#:#search_show_inactive_user_info#:#If enabled, the user search will show inactive users, too.
 search#:#search_show_limited_user#:#Show Users with Limited Access
 search#:#search_show_limited_user_info#:#If enabled, the user search will also show users with limited access outside their allowed time period.
+search#:#search_sort_by#:#Sortation: %s
+search#:#search_sort_creation_date_asc#:#Oldest
+search#:#search_sort_creation_date_desc#:#Latest
+search#:#search_sort_generic_asc#:#%s, asc.
+search#:#search_sort_generic_desc#:#%s, desc.
+search#:#search_sort_relevance#:#by Relevance
```
matthiaskunkel, line 15235: Please change to 'By Relevance' because 'by' is the first word in the line.

mjansenDatabay (approved):

lang/ilias_en.lang
schmitz-ilias: Fixed, thanks for the feedback!

oliversamoila:
Hello @schmitz-ilias,
thank you very much for the changes. I have one more suggestion regarding the sorting: could you use the same values as we use in the new UI data table? 
This would a) be consistent and b) would also be effective in all places with language improvements.
For example, the identifiers used address problems in which the sorting designations are not always clearly understandable. We therefore use, for example:
* "Erstellt am, Älteste zuerst" | "Created on, Earliest first"
* "Erstellt am, Neueste zuerst" | "Created on, Most Recent first"
* "Last Name, A to Z"
* "Last Name, Z to A"

Best regards
@oliversamoila

matthiaskunkel:

PR #8184, opened 2024-10-10
SOAP: Fix extracting NULL when parsing XML

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=42056


jeph864:
@mjansenDatabay thanks

PR #8186, opened 2024-10-11
PRG: raise event 'userReAssigned'

nhaagen:
https://mantis.ilias.de/view.php?id=42104


klees:
Picked to release_9 and release_10. @nhaagen: Would you be so kind an open a version for release_8? Had some conflicts when cherry picking there...

PR #8188, opened 2024-10-11
T&A: Implement participant table revision with new UI

thojou:
[Test - Revision of the Participant Tab using New Data Table](https://docu.ilias.de/go/wiki/wpage_7708_1357)

This PR implements the feature described in the wiki, and the functionality is now complete and ready for review. 

However, there are a few differences compared to the wiki description:

- **Status of Attempt:** ILIAS does not provide sufficient information to display all the states mentioned. Currently, only `Not yet started`, `Running`, and `Finished` statuses are supported.
- **Disinvite User Method:** The wiki does not mention a `disinviteUser` method. A decision is needed on whether to implement this.
- **Autocomplete for User Selection:** The autocomplete function doesn't work in the legacy UI due to a missing jQuery UI dependency.


kergomard:
Thank you very much @thojou ! I close this PR as we are continuing the implementation here: https://github.com/ILIAS-eLearning/ILIAS/pull/8235

PR #8194, opened 2024-10-16
Fixes roundtrip modal issue with multiple overlaping forms

matheuszych:
This PR regards the following Mantis Ticket: https://mantis.ilias.de/view.php?id=42306

In my opinion this is a viable fix for issue. The issue was a few redundant HTML forms, that caused HTML restructuring when loaded by the browser, trying to clean up the HTML.

### DIFF
--- a/components/ILIAS/UI/src/templates/default/Modal/tpl.roundtrip.html
+++ b/components/ILIAS/UI/src/templates/default/Modal/tpl.roundtrip.html
@@ -2,7 +2,7 @@
 	<div class="modal-dialog" role="document" data-replace-marker="component">
 		<div class="modal-content">
 			<div class="modal-header">
-				<form><button formmethod="dialog" class="close" aria-label="{CLOSE_LABEL}"><span aria-hidden="true">&times;</span></button></form>
+				<button formmethod="dialog" class="close" aria-label="{CLOSE_LABEL}"><span aria-hidden="true">&times;</span></button>
 				<h1 class="modal-title">{TITLE}</h1>
 			</div>
 			<div class="modal-body">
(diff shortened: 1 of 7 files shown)


dsstrassner:
~~Hi @kergomard, as this was assigned to me in Mantis, I assign this to you, to decide how to proceed.~~
Some wisdom has emerged and I reassign accordingly.

klees:
Hi @matheuszych,

this won't work. The form that you have deleted there actually exists to close the modal. Which will simply be broken after this change. Please have a look into the Kitchen Sink Documentation to try this out. The Roundtrip Modal does not vanish anymore.

So we need some other idea how to solve this.

Also, I think this might be understood to be a problem of the Ordering Table. What seems to be the problem here is the "form in form" problem, which is not allowed by HTML and handled by browsers accordingly. Since the Ordering Table provides space for the modals and the outer form as well, it might be the correct target for a fix as well.

On the other hand, there could be other contexts, maybe with legacy UI code, where the Modal also ends up in a surrounding form. So we might also solve this by some JS in the modal.

Anyway, I don't think this is an actual problem of the test but some problem of the UI components. Shall we ask @nhaagen to have a look into this?

Since this change here won't be it, I close this PR.

Kind regards!

matheuszych:
Hello @klees,

first of all thank you for your swift response.

In the specific scenario mentioned in the Mantis Ticket this did resolve the issue but i also understand that this might have created many others in different places. To be fair thou, closing the modal with the x symbol or cancel button did still work (at least in this specific case).

As you said this is most probably a "form in form" issue.

@nhaagen could you please take a look at the underlining issue?

Best regards
@matheuszych

klees:
@matheuszych Yeah, I did not want to mention that you did not try your fix for the issue at hand. I merely assumed that you might not be aware that the Kitchen Sink is another location to check the components that you have modified.

klees:
@matheuszych Could you try this one: https://github.com/ILIAS-eLearning/ILIAS/pull/8197?

PR #8196, opened 2024-10-16
AdvancedMetadata: Fix problem with missing `decimals`

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=42308

If approved, this should be picked to all maintained releases.


smeyer-ilias:
Thanks @mjansenDatabay

PR #8198, opened 2024-10-16
10/UI/modal min-width 42293

catenglaender:
https://mantis.ilias.de/view.php?id=42293

# Issue

Modals collapse inwards and don't have a consistent width.

image

# Changes

New width system based on svw (smallest viewport width). svw doesn't push into browser and system interface elements cutting into the width e.g. on smartphones with a camera hole inside the screen held in landscape orientation. Using vw as a fallback.

image

On mobile the modal now takes 100 svw:

image

modal-lg and modal-sm modifiers are still being respected. This is a modal lg:

image

# Big picture

Beyond this issue
* in some edge cases the drilldown in the modal behaves weird, overflows, height collapses... had to give up fixing that today, issue lies in drilldown

I am afraid we will not get around some refactoring/cleaning work after the legacy modal is removed:
* legacy and ui component modal are annoyingly interwoven - changing one might have negative consequences for the other as the inner class names are still identical
* scrollbar handling is not optimal for long modals on long pages

### DIFF
--- a/templates/default/070-components/UI-framework/Modal/_ui-component_modal.scss
+++ b/templates/default/070-components/UI-framework/Modal/_ui-component_modal.scss
@@ -22,8 +22,8 @@ $modal-header-border-color:   #e5e5e5 !default;
 //** Modal footer border color
 $modal-footer-border-color:   $modal-header-border-color !default;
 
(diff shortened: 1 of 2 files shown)


Amstutz:
thx a lot! special thx for the screenshot and explanations!

PR #8199, opened 2024-10-17
LTI: Fix Tests

kergomard:
Hi @jcopado & @Uwe-Kohnle 

Could you please have a look at this asap, as the Unit-Tests in ILIAS 9 do not work anymore since today. The commit where the build of the autoload-classes starts failing has nothing to do with the failing.

A way  to make them work again, would be this PR, the other would be to remove the namespace from this file. I don't know which one is right, but the current state is actually wrong and I've no idea why it worked for so long.

Issue here: https://mantis.ilias.de/view.php?id=42317

The failing tests now are because of the copyright checker.

Best,
@kergomard


jcopado:
Hi Stephan! We will have a look on monday
Thanks for the PR

ZallaxDev:
Hi @kergomard!
We have reviewed your PR and it seems to be correct.

Thank you very much.
Best Regards!

PR #8200, opened 2024-10-17
Export: renames ilObjectExporter to ilILIASObjectExporter to adjust to changes made by the component revision

chlulei:
Since the Object component was renamed to ILIASObject in the component revision the export currently fails because it can not find an `ilILIASObjectExporter`.

This PR simply renames the class.


kergomard:
Thank you very much for the PR @chlulei !
Once more a sign that we need to stop relying on naming conventions!

PR #8201, opened 2024-10-17
SOAP: Fix component (ILIAS 10.x)

mjansenDatabay:
SOAP is currently not usable at all in ILIAS 10.x

See: https://mantis.ilias.de/view.php?id=42320

**!!! Important: There is still an open issue !!!**

When container objects are copied, cascading SOAP requests are used within the SOAP context itself. In this case, the URL determined in `\ilSoapClient::init` (./components/ILIAS/WebServices/SOAP/classes/class.ilSoapClient.php) is not correct.

So if you don't configure a static SOAP WSDL path, this won't work.

Any ideas or contributions to this PR are highly appreciated.

 I'm sure I could find a solution, but I already spent 4 hours to analyse and fix all this mess.


jeph864:

PR #8204, opened 2024-10-17
Forum: export option

chlulei:
This PR reimplements the non standard export option of Forum to conform with the new export infrastructure. This also includes changes to the `ilObjForumGUI`. For details see [the documentation](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/Export/README-CustomExportOptions.md).

The export option is "HTML". The implementation of the option is fairly trivial because no files are stored on the server. They are only offered as a direct download.


mjansenDatabay (approved):
LGTM!

PR #8205, opened 2024-10-17
Container: Fix undefined logger in `cloneObject`

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=42319


alex40724:
Thanks, this is already fixed.

PR #7983, opened 2024-08-21
Replacement of the legacy modals in ilTestPlayerConfirmationModal

matheuszych:
These changes are part of the [Legacy-UI refactoring](https://docu.ilias.de/go/wiki/wpage_7320_1357).
I replaced the legacy ilModalGUI with the newer Kitchen Sink interruptive Modal.

PR #8211, opened 2024-10-17
UI: Replace legacy modals in ilTestPlayerAbstractGUI

lukas-heinrich:
These changes are part of the [Legacy-UI refactoring](https://docu.ilias.de/go/wiki/wpage_7320_1357). This PR revises and brings together #7983 and #7984.

@matheuszych and I replaced the legacy ilModalGUI with the newer Kitchen Sink interruptive and roundtrip Modal. The control of the modal via DOM elements in JavaScript has been replaced by using signals. Class `ilTestPlayerConfirmationModal` and associated template files have been removed as they are no longer needed.

In addition, the commit [cf00c9d](https://github.com/ILIAS-eLearning/ILIAS/commit/cf00c9d610396958957e1d1f73cf054f4a5ca74e) was reverted to fix `lObject::read(): Object with obj_id: -1 (tst) not found!`. The behavior described in the corresponding Mantis ticket has not reappeared here.


kergomard (changes requested):

First:
- I was completely unable to get the FeedbackModal to show, when testing.

The code is mostly ok and I've just three small requests:
- Please rename the variable [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8211/files#diff-eb107de08e0a9134113438fa1319c62882481cccd58d471723e39bad9d904f4eR1317) to snake_case.
- Please do not revert my commit for [this line here](https://github.com/ILIAS-eLearning/ILIAS/pull/8211/files#diff-b7c00144f7830b5f9b62afdbc7dc093fe4847eacbf87f5c2c1de1fb6c0560572R2914), but explain precisely when you run into the issue that the `obj_id` is -1. For the moment, I believe that my solution is actually the right one and we need to figure  out, how we can get here without having a valid test (and then wanting a test). Please create an issue, if necessary.
- Please fix the coding style.

Finally (this nothing you introduced here, but as you are working on this):
- Could you please have a look at why the buttons below the question do not work.

Thanks again and best,
@kergomard

kergomard:
Hi again @lukas-heinrich 
One more little thing: There is an issue with the

kergomard:
Sorry @lukas-heinrich ! Didn't want to do that: Neither the comment nor closing the PR!

lukas-heinrich:
Hi @kergomard !

Thank you very much for the feedback. I will implement the suggestions and update them in the PR.

I also tried to find out why the buttons in the bottom navigation bar are not working. This seems to be due to the browser placing them in the wrong place in the DOM.  I was able to trace the behavior back to the fact that modals are rendered in the player's form, which themselves contain other forms. This leads to incorrect behavior because HTML forms are not allowed to have other HTML forms. 
I have adjusted my relevant parts, but unfortunately I could not fix the issue completely. The modal `.il-copg-mob-fullscreen` is still in the form element and I have currently found no way to render it outside the form. I created a Mantis ticket for this specific modal: https://mantis.ilias.de/view.php?id=42364

Best regards,
@lukas-heinrich

lukas-heinrich:
Hi @kergomard !

I renamed the variable and moved the modals out of the form as described.

With regard to [this line of code](https://github.com/ILIAS-eLearning/ILIAS/pull/8211/files#diff-b7c00144f7830b5f9b62afdbc7dc093fe4847eacbf87f5c2c1de1fb6c0560572R2914), I can't tell which solution is the right one. I get the problem that `$this->test_id` is -1 when I delete an already answered question. After I click on the confirm button in the modal, the script exits with an error. Unfortunately, I can't tell how the program should behave correctly at this point. If desired, I can undo the revert and open a Mantis ticket for the issue.

I have also noticed that the pipeline has failed again. I don't really understand why, any hints are welcome.

Best,
@lukas-heinrich

kergomard:

Now it does work. There is one last thing: The answer feedback modal has a funky background-color.

And: Yes, please revert the change and add an issue. I will look into that.

After these two changes, I will merge.

Best,
@kergomard 

grafik

lukas-heinrich:
Hi @kergomard !

Thank you for your support. I removed the changes and opened a [ticket](https://mantis.ilias.de/view.php?id=42393) for the issue.

The formatting seems to be due to the fact that the feedback modal is an `interruptive` modal. If I interpret the documentation correctly, you can also use a _roundtrip_ modal here and the modal is displayed in a better way (see screenshot). 
I have this the change in a separate commit so that you can ignore it while merging if you do not want to modify the feedback modal.

image

kergomard:
Thank you very much @lukas-heinrich .

The idea with the roundtrip modal was the right one, as the text  in the interruptive modal is always wrapped in an info-box and this box will have a yellow background. ...but there are more issues here:
- With the interruptive modal, the "continue" action, doesn't do what it should, i.e. it does not continue.
- Additionally, right now, if you set the question to be locked when navigating, this doesn't actually happen.
- What is more, you have some feedback info in the question, that part should have a colored background, and this doesn't happen in the interruptive modal either. For this to work we need to change some CSS and this we need to do through a clean PR for Timon, the maintainer, to review.

All this we are not going to fix in a spell and we also don't need to have this done by tomorrow. Could you maybe see with @thojou when you two would have the time, so we can have a look at this together in a call? Maybe @fhelfer and/or @matheuszych would like to join, too? I think this would be the best way to go forward. I would be available tomorrow after around 11ish or then next week.

Thanks again and best,
@kergomard

kergomard:
Hi @lukas-heinrich 
When trying to figure out the whole functionality with @dsstrassner , I finally understood that the current behavior actually conforms to the specifications (the case I had found, that from where I stand is still not correct, none the less is as specified: If you set the locking to "Lock Answers After Moving to Next Question" and you have the feedback turned on, clicking away the modal will lead to a state that still lets you change your answer. You have to set locking to "Lock Answers After Feedback or Moving to Next Question" and only then the answer will really be locked, when clicking away the Modal). We sadly cannot get rid of the modal right now, as there are too many states that will be unclear to the user. We need to find a solution for this, but we are not going to find it now for ILIAS 10. Sorry, that I didn't grasp the full picture!

Could you please:
- See if you can find a solution for the display problems.
- Remove all camelCase names of properties you introduced.
- Remove the @throws.
- Rebase (it is trivial, as far as I can see as the only conflict should be in the test-class that is remove anyway).

Thanks and sorry again,
@kergomard

lukas-heinrich:
Hi @kergomard !

Thank you for the feedback! I have visually revised the code and I found a solution for locking the answer while the dialog is already shown. It should now work as expected, I've tried tests with different combinations of with/without feedback and locking modes.

Best,
@lukas-heinrich

kergomard:

I merge this and will remove some of the comments myself to not torture you any longer with this. Merged and picked to trunk.

Best,
@kergomard

PR #8212, opened 2024-10-18
Init/PasswordAssistance: Fix `auth_mode` equals `12_[id]` (Saml)

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=42327

If approved, this has to be picked to `relase_9` and `trunk`.


mjansenDatabay:
I'll merge this, cause it's trivial.

PR #8181, opened 2024-10-10
Trunk LTI 001 - Adding dependencies & LTI Provider folder renaming

jcopado:
The composer.json file has been modified to include the php-jwt and LTI-PHP dependencies, required in the LTI module to stop including external libraries inside the module itself.

The LTI folder has been renamed to LTIProvider to maintain better consistency in component names.

PR #8216, opened 2024-10-18
LTI: Adding the php-jwt dependency to the composer.json file and update the composer.lock file

ZallaxDev:
We integrate the php-jwt dependency as it is required for the LTI and LTIConsumer components to work.
This dependency is secure as it is officially supported by the team behind Firebase at Google. It is kept up to date, allowing compatibility with modern versions of PHP, requiring the current version to be at least PHP 8 or higher.
No better alternatives have been found with this support, maintenance and compatibility.
The inclusion of these dependencies will allow the removal of the 'src' folder included in the LTI component and make it easier to upgrade these components in the future.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,6 +49,7 @@
 		"ext-xml": "*",
 		"ext-zip": "*",
 		"ext-imagick": "*",
+		"firebase/php-jwt": "*",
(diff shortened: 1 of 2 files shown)


dsstrassner:
JF Decision is in #8181 and was at 14. OCT 2024

PR #8218, opened 2024-10-18
LTI: Adding LTI-PHP dependency to composer.json and updatd composer.lock

ZallaxDev:
We integrate the LTI-PHP dependency as it is required for the operation of the LTI and LTIConsumer components.
This dependency is secure as it is supported by a foundation composed by different prestigious institutions (JISC-Funded).
It has extensive support for all available LTI versions and guarantees compatibility with future versions due to its commitment to the world of eLearning.
We believe that the choice of this dependency over others is ideal for several factors, among which are the simplicity for the functionality required of them within the LTI component and their background with a team coming directly from the world of eLearning and open-source.
Its development is active and has very recent updates.
The inclusion of these dependencies will allow the removal of the 'src' folder included in the LTI component and will facilitate the updating of these components in the future.


ZallaxDev:
Missclick, sorry

PR #8220, opened 2024-10-18
LTI: Renamed the LTI folder to LTIProvider and removed the src folder from inside LTIProvider.

ZallaxDev:
Renamed the LTI folder to LTIProvider to maintain a better consistency in the component names.
Removed the 'src' folder included in the LTI component (now LTIProvider) since its content has been explicitly integrated as dependencies inside composer as of Ilias 10.


dsstrassner:
JF Decision is in #8181 and was at 14. OCT 2024

PR #8222, opened 2024-10-21
UI 42336: Fixes being unable to sort table records by columns that are not show…

matheuszych:
When trying to sort the entries of a table using the viewcontrols you get only shown the columns that are visible by default. Even when enabling optional columns you are unable to sort by them because the according entries in the dropdown menu are missing. This applies to the new KS tables.

I believe this could be a possible fix.

[Mantis](https://mantis.ilias.de/view.php?id=42336)

### DIFF
--- a/components/ILIAS/UI/src/Implementation/Component/Table/Data.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Table/Data.php
@@ -125,8 +125,8 @@ class Data extends AbstractTable implements T\Data
         $view_controls = $this->getViewControls($total_count);
 
         if ($request = $this->getRequest()) {
-            $view_controls = $this->applyValuesToViewcontrols($view_controls, $request);
-            $data = $view_controls->getData();
+            # This retrieves container data from the request
+            $data = $this->applyValuesToViewcontrols($view_controls, $request)->getData();
             $range = $data[self::VIEWCONTROL_KEY_PAGINATION];
             $range = ($range instanceof Range) ? $range->croppedTo($total_count ?? PHP_INT_MAX) : null;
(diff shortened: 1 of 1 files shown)


nhaagen (approved):
Hi @matheuszych ,
I've been pondering about this for a bit; as you figured, the ViewControls are interdependent and have to be re-initialized.

For the interested readers of this: FieldSelection provides the selected columns, and Sortation reads $table->getVisibleColumns.
When retrieving the Viewcontrols [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8222/files#diff-1894392fb23ad26ef97a486053bf64261ac5dfd8dd7c8d8ce24e29bfa705b06dL128), the table is not yet updated.

What happens (and why this is a good solution): first call on the controls is just to retrieve container-data from the request, while the second call - provided with this PR -is to actually get and display the ViewControls.

I was a little unsure at first because of (my own) naming and the seemingly redundant call; maybe it will also help others if we made this a little more obvious, e.g. by
`$data = $this->applyValuesToViewcontrols($view_controls, $request)->getData();`
instead of setting $view_controls in l128 already?
@matheuszych , would you be so kind as to bear with me and change those lines, too, perhaps even add a comment?

Anyway, thank you a lot, and best regards,
Nils

matheuszych:
Hello @nhaagen,
first of all thank you for your review. I adjusted the code accordingly and added comments.
Best regards
@matheuszych

nhaagen:
Thank you very much, @matheuszych

klees:
Thanks!

klees:
Picked to release_8 and release_9.

PR #8224, opened 2024-10-21
A space was missing in lang var for Notifications

dsstrassner:
I will refrain from writing a Mantis report.


mjansenDatabay:
Thx Denis

PR #8227, opened 2024-10-21
UI: 42318 Fixes date column output in user timezone

lukas-heinrich:
This PR addresses the problem that a date in a DataTable is not displayed in the user's time zone. The problem is solved here by adding a method for defining the time zone to be displayed. If the time zone is not explicitly defined, the date is displayed in the user's time zone.

See: https://mantis.ilias.de/view.php?id=42318

### DIFF
--- a/components/ILIAS/UI/UI.php
+++ b/components/ILIAS/UI/UI.php
@@ -36,6 +36,7 @@ class UI implements Component\Component
         $define[] = UI\Renderer::class;
         $define[] = UI\HelpTextRetriever::class;
         $define[] = UI\Storage::class;
+        $define[] = UI\UserTimezone::class;
         $define[] = UI\Component\Progress\AsyncRefreshInterval::class;
         $define[] = UI\Component\Input\Field\PhpUploadLimit::class;
         $define[] = UI\Component\Input\Field\GlobalUploadLimit::class;
@@ -380,6 +381,7 @@ class UI implements Component\Component
         $internal[UI\Implementation\Component\Table\Column\Factory::class] = static fn() =>
             new UI\Implementation\Component\Table\Column\Factory(
                 $use[\ILIAS\Language\Language::class],
+                $use[UI\UserTimezone::class],
             );
         $internal[UI\Implementation\Component\Table\Action\Factory::class] = static fn() =>
             new UI\Implementation\Component\Table\Action\Factory();
--- a/components/ILIAS/UI/src/Implementation/Component/Table/Column/Date.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Table/Column/Date.php
@@ -29,7 +29,8 @@ class Date extends Column implements C\Date
     public function __construct(
         Language $lng,
         string $title,
-        protected DateFormat $format
+        protected DateFormat $format,
+        protected \DateTimeZone $timezone
     ) {
         parent::__construct($lng, $title);
     }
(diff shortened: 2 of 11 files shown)


klees (changes requested):
Hi @lukas-heinrich,

thanks for looking into this.

* Dependency on ilObjUser: Why do you think we need to introduce a dependency from the UI framework to that object with the huge interface?

Kind regards!
@klees, @thibsy and @Amstutz

Please assign back to me when we can go on here.

oliversamoila:
Hallo @lukas-heinrich,
arbeitet ihr an diesem PR noch weiter? Gib sehr gern noch eine Rückmeldung auf die Frage von Richard.

Viele Grüße und besten Dank.
Oliver

lukas-heinrich:
Hi @klees , @thibsy  and @Amstutz ,

Thank you for your review and feedback on this proposal.

In my implementation, I followed the structure of the Init-classes, where I understood only dependency injection should occur. I avoided directly calling `ilObjUser::getTimeZone()` to prevent side effects and maintain proper SoC in the Init-class. However, I'm struggling to find an alternative way to pass the user's time zone information elsewhere.

If I've misunderstood something or overlooked an alternative approach, I would welcome suggestions for a better implementation.

Best regards,
@lukas-heinrich

lukas-heinrich:
Hi everyone,

As discussed, I have adjusted the implementation to remove the dependency on `ilObjUser`. The approach of using a closure that passes the user's timezone to the factory was not possible in this case. Instead, I use an interface and the implementation in the user component to integrate it within the new component bootstrap mechanism.

I look forward to your feedback.

Kind Regards,
@lukas-heinrich

klees:
Hi @lukas-heinrich,

thanks for looking into this again. This is fine now from my POV. Could you explain a little, why the approach with the closure did not work out? Sounds interesting.

I kindly ask @kergomard to look into this as well.

Thanks!

thibsy:
Hi @klees,

@lukas-heinrich actually asked me for directions on how to achieve this and I wasn't sure if its possible to simply declare a closure like this. So I pointed him into the direction of `UI\Component\Progress\AsyncRefreshInterval`, which I believe is the appropriate way to ask other components for configuration like this.

I just checked now, a closure would have worked just fine:
```php
$internal['closure_works'] = static fn() => static function (): mixed {
    global $DIC;
    return $DIC->something();
};
$internal['foo'] = new Foo(
    static function () { // closure works
        global $DIC;
        return $DIC->something();
    },
);
$internal['bar'] = new Bar(
    $internal['closure_works'],
);
```

Kind regards,
@thibsy

klees:
Thanks for the explanation =)

@lukas-heinrich seems to have tripped some complexity wire here, we have discussed this again today and even found another angle to look onto this problem.

I will elaborate very soon...

kergomard:
Hi all

I de assign myself from this PR to get this off my list for the time being.

I do actually not agree with the changes in the `User` proposed here as they would lead to a tangle of knots: You are introducing the use of the `$DIC` into the new integration mechanisms proposed by the Component-Revision. This cannot be right. Looking more deeply into it, you are introducing it as a workaround, as otherwise you can't access the necessary infrastructure at the moment the UI-Framework is initialized.
The correct way of doing this would be to move the necessary dependencies for the user to the new integration mechanism and then you could initialize the user properly there. but this is first a lot of work and second you would make the interface depend on the User and transitively on the Database and other things. Now we are back at tangling ourselves in knots: We need the db to initialize the interface, that would allow us to show errors when initializing other dependencies.
I think  the current solution is actually correct: The timezone is set at the moment of presentation and it is up to you to do so. This could be provided by a function in whatever presentation of dates and times you are intending to show (in this case it would be the table column), but personally I do not see much profit from doing this compared to the current solution. Maybe we can device a mechanism by which `User` can parametrize the UI-Renderer (I think it would need to be the renderer, not the factory) once it is initialized? ...but this would need more conceptual work and I don't see this high on my list of priorities.

Thank you for the effort and best,
@kergomard

dsstrassner:
Hi @oliversamoila, I think you should discuss with @thibsy how to proceed in the future with this PR :see_no_evil:

PR #8228, opened 2024-10-21
UI: 39326 Allow empty cell in data table by providiung a null value

lukas-heinrich:
This PR extends the `Column` class with the `withEmptyPlaceholder` method and introduces a new `ExmptyCell` component. The idea is that data tables can now handle null values instead of throwing an error due to an incorrect parameter (see https://mantis.ilias.de/view.php?id=39326).

- A placeholder can now be specified for each column, which is displayed if the value of the column is empty. If no placeholder is defined, the cell remains without any content. 
- A new component has been defined for empty cells. In this implementation, the message is displayed in italics. In the future, the component will make it possible to adapt the display via style sheets (e.g. to define the `text-align`).
- A cell is treated as empty if the format method of the respective column returns an `EmptyCell` component. Depending on the type of column, values are recognized differently as empty:
   - an empty string, zero numeric values, false, null and empty arrays are considered empty by default (see [empty()](https://www.php.net/manual/en/function.empty.php)) 
   - in columns `Boolean` and `Number` only null values are considered empty

### DIFF
--- a/components/ILIAS/UI/src/Component/Table/Column/Column.php
+++ b/components/ILIAS/UI/src/Component/Table/Column/Column.php
@@ -39,6 +39,11 @@ interface Column extends \ILIAS\UI\Component\Component
         string $desc_label = null
     ): self;
 
+    /**
+     * you may specify a placeholder to be displayed if the value is empty
+     */
+    public function withEmptyPlaceholder(string $empty_placeholder): self;
+
     public function withIsOptional(bool $is_optional, bool $is_initially_visible = true): self;
     public function isOptional(): bool;
     public function isInitiallyVisible(): bool;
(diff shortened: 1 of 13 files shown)


kergomard:
@lukas-heinrich : You are aware that we are already having the same discussion here: https://github.com/ILIAS-eLearning/ILIAS/pull/7439 
Temporarily assign this to you, but will change the assignment, if you keep this open.

oliversamoila:
Hello everyone, 
I stumbled across this and the related PR in https://mantis.ilias.de/view.php?id=39326 again. 
As a non-expert on the implementation details, can you give me some info on when we can expect a next step here? I'm just a little concerned that it's assumed that it's the other one's turn ;)
Thank you very much for your work on this topic.

kergomard:
Not really. I think we could actually close this one for the time being and use the linked PR for further discussions.

The matter is not urgent, as there is a way to deal with this, the question is rather if we want to this more verbosely.

Best,
@kergomard

lukas-heinrich:
Hi everyone,

I also think we can close this PR and use the linked PR for further progress and discussion.

Best
@lukas-heinrich

PR #8229, opened 2024-10-21
Legacy UI: Remove `Tooltip` (with usages) from code base

mjansenDatabay:
This PR removes the legacy `ilTooltipGUI` usages and the class itself.

If left tooltip-related "interfaces" in the classes to ensure no consumer code breaks.


klees:
Checks fail because the linter tries to lint removed files...

klees:
@mjansenDatabay Could you rebase? I will merge then... Thanks!

mjansenDatabay:
Hi @klees ,

thanks for the review. I can/will rebase, but I fear the checks will still fail. The problematic file reported in the CI process is the one deleted in this PR. This is a known issue in the shell script. When files are deleted or moved, the php-cs-fixer receives this deleted file as an argument.

PR #8230, opened 2024-10-21
T&A: Refactor ilListOfQuestions according to data table component

lukas-heinrich:
Refactors `ilListOfQuestionsTable` by using the `DataTable` component instead of ilTable2GUI inheritance. Removes unnecessary code and adjusts the unit tests:

- Creates a new Table class that implements the creation and formatting of columns and rows
- Replaces getter and setter with queries to the test object
- Removes legacy code that is always false
- Removes the template of the table rows


kergomard (changes requested):
Thank you very much for this PR @lukas-heinrich 

Here I would ask for a few more changes (I didn't test it yet, so :

- [ ] I would kindly ask you to remove the whole `ilListOfQuestionsTableUI` class and move everything to `ListOfQuestionsTable`.
- [ ]  Please remove all setters from `ListOfQuestionsTable` and move everything to the constructor. I don't think they have any added value: The table does not have any state right now, where it could function usefully without data or a test. ...if I'm not mistaken. 
- [ ] Please remove all doc-blocks unless they add something significant otherwise they will start lying sooner or later.
- [ ] Please remove everything concerning `obligations`. We don't need them anymore. For testing you can rebase on this PR [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8235).
- [ ] Please move the `ListOfQuestionsTable` to the namespace `ILIAS\Test\Questions\Presentation` (see PR above). We should actually also rename this class, as it doesn't tell us anything about it's use, maybe `QuestionsOfAttemptTable`? The reason I would like to rename it is, that we already have other tables of questions in other contexts and I would like to avoid confusion. I'm open for suggestions, if you have better ideas!
- [ ] Please (re)name all variables you touch or introduce in snake_case.

I quickly opened it and everything looks fine otherwise.

Thank you very much again and best!
@kergomard.

lukas-heinrich:
Hi @kergomard !

Thank you very much for the review and tips. I implemented all of your suggested changes. 

I have rebased my branch locally and here only cherry-picked my commit, because I was unable to rebase it back to trunk. Therefore, I think it is likely that there will be broken code if this branch is merged before #8235. I hope this doesn't cause too much trouble.

Best Regards,
@lukas-heinrich

kergomard:
I picked this to "the big one" [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8235/commits/20256c15e2a084119f9afd4432dcc5d10475b136) and will merge all together. This way we avoid some rebase issues. I thus close this PR.

Thanks again and best,
@kergomard

PR #8233, opened 2024-10-21
T&A: Refactor ilTestTopList according to data table component

lukas-heinrich:
Refactors `ilTestTopListTable` by using the `DataTable` component instead of ilTable2GUI inheritance. Removes unnecessary code and adjusts the unit tests:

- Outsources database query to new repository class and use a generator instead of returning the complete array
- Creates a new Table class that implements the creation and formatting of columns and rows
- Adapts unit tests to the new structure and creates test case for formatting the time span
- Streamlines the code by passing queries with parameters for sorting and display type to general functions


kergomard (changes requested):
Thank you very much for this PR @lukas-heinrich 

This actually already looks really good.

The usual set of small improvements, sorry:

- [ ] Please remove all doc-blocks unless they add something significant otherwise they will start lying sooner or later.
- [ ] Please move the const values for order_by and table_type to `Enums`, we are using them in parameters and this allows us to type them correctly.
- [ ] Please Rename `TestTopListTable` to simply `DataRetrieval`. It is not actually the table, simply adding `DataRetrieval` makes the name too long and anyhow: All other Information is already there in the namespace, so no need for it.
- [ ] Please move the Toplist-classes to the Results-namespace `ILIAS\Test\Results\Toplist` (the parent namespace will be added with: https://github.com/ILIAS-eLearning/ILIAS/pull/8235).They are not about scoring (Notengebung), but about the presentation of results.

Thanks again and best,
@kergomard

lukas-heinrich:
Hi @kergomard !

Thank you very much for the review and feedback. I have implemented the changes. If there are any further suggestions, I will be happy to look at them.

Best Regards,
@lukas-heinrich

kergomard:

This looks right. Just a little request for the next time: Please don't force-push, but keep the history (add a new commit). This makes my life a lot easier, as I can then just have a look at your additional changes, instead of having to search for them in the whole code.

Thanks and best,
@kergomard

lukas-heinrich:
Hi @kergomard 

Thanks for merging. And thanks also for the hint! I didn't have that on my radar -  but I will make sure from now on that the history is preserved.

Best regards
@lukas-heinrich

PR #8234, opened 2024-10-21
[FIX] 0042349: UI Tooltips (via Topics) won't shown on slate-triggerers in MainBar

chfsx:
as mentioned in https://mantis.ilias.de/view.php?id=42349, Slates currently are unable to receive UI-Topics which are needed for e.g. the Tooltips of the Help-Package to be rendered correctly (@alex40724 this might be of interest for you).

I implemented the HasHelpTopics interface in Slates with this PR. Please be aware that this might break some layouts when using Topcis, but as far as I can see the e.g. Buttons inside a CombinedSlate inside the MetaBar are broken anyway (look strange since some time)., I sure it will be easier for people which are fitter than me in (S)CSS to takle this in one step.

A central question is, of course, which component receives the tooltip in the case of a slate (the same question probably arises for all components that internally use another component, for example, as a trigger, i.e. something like Modals). From this point of view, tooltips can be said to be at the triggerer. For further use cases (such as extensive explanatory texts), further `Purposes` should be implemented for the `Topics`, which then also allow a different form of presentation in a component.

Is there anything needed beside this?

### DIFF
--- a/components/ILIAS/UI/src/Component/MainControls/Slate/Slate.php
+++ b/components/ILIAS/UI/src/Component/MainControls/Slate/Slate.php
@@ -25,11 +25,12 @@ use ILIAS\UI\Component\JavaScriptBindable;
 use ILIAS\UI\Component\Signal;
 use ILIAS\UI\Component\Symbol\Symbol;
 use ILIAS\UI\Component\Triggerer;
+use ILIAS\UI\Component\HasHelpTopics;
 
 /**
  * This describes a Slate
  */
-interface Slate extends Component, JavaScriptBindable, Triggerer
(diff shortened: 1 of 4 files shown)


klees:
Hi @chfsx,

this looks absolutely plausible and we think that it can be presented on the JF =)

Thanks!

oliversamoila:
[JourFixe], 09 DEC 2024 : We follow the suggestion of Fabian and accept the mentioned changes for ILIAS 10 and Trunk. 
Many Thanks

klees:
I'll merge even though tests fail. They seem to be unrelated to the original changes and I expect the problem to go away once this is rebased.

klees:
Picked to release_10.

PR #8235, opened 2024-10-21
10/test/feature restructure tabs on new pax table

kergomard:
Ok, this PR is once more of an informational nature. It contains the following changes:
- The changes from #8188 by @thojou .
- The restructuring of tabs in test (https://docu.ilias.de/go/wiki/wpage_8317_1357).
- The removal of mandatory questions (https://docu.ilias.de/go/wiki/wpage_8397_1357).
- The removal of excel exports in question pools (https://docu.ilias.de/go/wiki/wpage_8462_1357).
- The removal of question type specific settings (https://docu.ilias.de/go/wiki/wpage_8401_1357).
- It moves the interface for vertical and horizontal ordering questions to a vanilla html/js approach.

It still lacks the changes to move the matching questions to vanilla html/js. This might still become part of this PR or it might arrive separately.


templates/default/070-components/legacy/Modules/_component_test_legacy.scss
```diff
@@ -3,6 +3,8 @@
 @use "../../../070-components/legacy/component_alert" as alert;
```
kergomard, line 3: @Amstutz 
If you get a chance to look at the changes in the templates here in the next couple of day, then I'm grateful. If you don't: Don't worry, I take them out and provide a separate PR. ...and if you would like a separate PR: I can also simply provide it.

Thanks and best,
@kergomard

Amstutz (changes requested):
Hi @kergomard 

That is some big PR. I was initially shocked to look into that, but then immediatly glad, that I only had to cover this one file.

Please answer:
- [x] dropzone bg color: Why do we need a new variable here, is there no existing one?
- [ ] margins: is it possible to use the variables from templates/default/050-layout/basics/_layout_spacing-variables.scss ?
- [x] dropzone bg color hover: Same question as above, however, if we need this, please also introduce a variable als you dit with the bg-color (this makes it easy to see from the top, that there are special colors at work here).

thx.

kergomard:
Thank you very much for the fast reply @Amstutz !
You are right and I was not careful enough: I use `$il-main-dark-bg` and `$il-main-darker-bg` for the background colors now.
I did not change the margins though, even if I would love to:

- The surrounding elements are defined in the content styles. I would like to change this, but for now they are and I would very much prefer not touching them, so I need them to be exactly 5px and 10px and we don't have those (and would also provoke unexpected effects if somebody decided to change them). I hope this is ok with you.

Thanks again and best,
@kergomard

Amstutz (approved):
Hi @kergomard 

Thx for the changes and reply. Can you drop a short comment above the margin explaining that. Otherwise I might change it during the next revision.

thx a lot. Feel free to proceed with the PR as you feel fits.

kergomard:
Thanks all for all the work!

PR #8236, opened 2024-10-21
Registration: Fix confirmation of code deletion

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=42350


mjansenDatabay:
I'll merge this trivial change.

PR #8242, opened 2024-10-22
Imprint: Fix `ilCtrl` issue that I keep constantly preaching about fi…

mjansenDatabay:
…xing

See: https://mantis.ilias.de/view.php?id=42355


klees:
Unit tests fail, but problem was introduced in trunk before.

PR #8250, opened 2024-10-23
Bugfix: 42377 &  42369: UI/Fields: Fix undefined input in file inputs 

lukas-heinrich:
This PR fixes an issue caused by #8177 which means that the id of the input field can no longer be accessed. I fixed the issue by using the id of the associated fieldset element as described in the PR. 

I am looking forward to feedback or hints if it can be implemented differently.

### DIFF
--- a/components/ILIAS/UI/resources/js/Input/Field/file.js
+++ b/components/ILIAS/UI/resources/js/Input/Field/file.js
@@ -252,7 +252,7 @@ il.UI.Input = il.UI.Input || {};
 		 */
 		let removeFileManuallyHook = function (event) {
 			let removal_glyph = $(this);
-			let input_id = removal_glyph.closest(SELECTOR.file_input).attr('id');
+			let input_id = getFileInputID(removal_glyph.closest(SELECTOR.file_input));
 			let dropzone = dropzones[input_id];
 			current_form.errors = false;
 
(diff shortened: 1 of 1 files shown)


kergomard:
Hi all

I assigned you all, because I was unsure if it was the ui-coordinators or @chfsx who felt responsible for this.

You were just a little bit quicker than me @lukas-heinrich ;-).

This has quite a few consequences:
- Settings with new forms cannot be saved.
- Imports don't work.

This would be the issue that triggered my look into it: https://mantis.ilias.de/view.php?id=42369

Thank you all and best,
@kergomard

dsstrassner:
T&A issue for this: https://mantis.ilias.de/view.php?id=42377

klees:
@nhaagen Review this please.

Thanks @lukas-heinrich

nhaagen (commented):
Hi @lukas-heinrich ,
first of all: sorry for the inconvenience caused by #8177.
Then: thank you for the fix!

This seems to work, however: did you try to just modify the selector [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8250/files?diff=split&w=0#diff-940f26b7073bfdb0d6c3458b61892e735a23bf5be04c26551c430fb9bcdeace0L37)?
I think that `SELECTOR.file_input =  '[data-il-ui-component="file-field-input"]' `would suffice.
Would you give that a shot?

Thanks and kind regards,
Nils

lukas-heinrich:
Hi @nhaagen !

Thank you for the feedback. This is of course an even simpler solution. I have customized the selector and tested it, it works as it should.

Thanks and best regards.
@lukas-heinrich

nhaagen (approved):
LGTM, thank you.

Amstutz:
thx big time @lukas-heinrich and @kergomard as well as @nhaagen for the review!

PR #8257, opened 2024-10-24
T&A: Replace legacy ilTable2GUI inheritance in ilTestRandomQuestionSet classes

lukas-heinrich:
Hi everyone, 

this PR refactors `ilTestRandomQuestionSetSourcePoolDefinitionListTable` and `ilTestRandomQuestionSetNonAvailablePoolsTableGUI` in `ilTestRandomQuestionSetConfigGUI` by removing ilTable2GUI inheritance:

- creates a new Ordering-Table class that implements the creation and provides order handling
- creates a new Data-Table class that implements the creation and formatting of columns and rows
- adjusts the table action handling in both GUI classes by using `URLBuilder`
- provides the global http request in `ilTestRandomQuestionSetConfigGUI` constructor
- removes the template of the table rows


kergomard (changes requested):

Sorry, I didn't get around to having a look at this yesterday.

- [ ] As always, sorry: Please get rid of all unnecessary doc-blocks. This includes ALL `@throws` as we do not plan to do anything with them, we don't care.
- [ ] Please remove the `ilTestRandomQuestionSetSourcePoolDefinitionListTableGUI` and please get rid of all Setters in the table-classes. Unless I'm wrong, we can move all of them to the constructor of `RandomQuestionSetSourcePoolDefinitionListTable`. We might even be able to load the data directly on `getRows`, but I didn't look to much into the details here.
- [ ] I think we can also then get rid of quite a few public getters in `RandomQuestionSetNonAvailablePoolTable`.
- [ ] Please insert line brakes in long lines with parameters like [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8257/files#diff-bfe24e18b8bed2928f7c125e9e7c002fb6842df34fd919a4c79feb365479220cR73).
- [ ] Do we really need [this](https://github.com/ILIAS-eLearning/ILIAS/pull/8257/files#diff-b5b536325950a92c7d526c6e5f6ce77a8ccbd29f0b36f6385fb18d2b57f7ed9dR190-R192)? I don't like this solution and hope we can get through only using `ilCtrl` or the `URLBuilder`, but no `parse_url()` + `DataFactory` shenanigans ;-) .
- [ ] If you get around to it, could you please do me a little favor and rename the "ilTestQuestionFilterLabelTranslater" into a "ilTestQuestionFilterLabelTranslator" and the corresponding Test-class? This hurts my  sensibilities ;-) .
- [ ] You will also need to rebase, sorry!

I've not tested the functionality yet. I will do some smoke testing, once you have made your changes. If you are not able to do this today (and if I'm getting it right, you won't be), we will merge this after the branch for ILIAS 10 has been split. It is not the end of the world, if this change arrives one version later. I will not port this back to ILIAS 10 after Coding Complete, unless somebody else is absolutely adamant about absolutely wanting that. I think it is good practice not to do that.

Thanks again and best,
@kergomard

lukas-heinrich:
Hi @kergomard 

Thank you very much for your feedback! 
I have rebased the branch and made the following changes:

- Removed the `ilTestRandomQuestionSetSourcePoolDefinitionListTableGUI`-class and also all setters and getters
- Also removed the setter in `RandomQuestionSetNonAvailablePoolTable`
- I renamed `ilTestQuestionFilterLabelTranslater` and corresponding local variables :)
- Revised the handling with action targets, using implementations from `QuestionsTableActions` and `GroupsTable` as references
- Also adjusted the fetching of the ID parameters [here](https://github.com/ILIAS-eLearning/ILIAS/blob/bbcbb947baba310baab5e9381d71ec33d331510f/components/ILIAS/Test/classes/class.ilTestRandomQuestionSetConfigGUI.php#L685). The reason for this is that the IDs are always transmitted by the table as an array, but by the form as a single integer. However, since the ID is always passed as an array by the table actions, I was able to simplify the delete method
- General style revising: removed doc blocks and insertion of wrapped lines

Best Regards
@lukas-heinrich

kergomard:
Hi @lukas-heinrich 

Thank you very much for updating this!

From a code-point of view this looks mostly fine now, just one very small thing:
- Please search quickly for the variable `$definitionIds`, it should be `$definition_ids`, e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8257/files#diff-530b8baeee0e8f48d0d1d73cbe3cdd8300315e4a11aece215ee33f70c6f4b549R708)

And then there are some usability-issues:
- Please remove the input field for the number of questions from the table. They are not allowed and shouldn't be there.
- If I import a test with random questions from ILIAS 9 (the export of such tests in ILIAS 10 is currently broken, as far as I can see, and for importing it from ILIAS 9, I need to change this line [here](https://github.com/ILIAS-eLearning/ILIAS/blob/release_10/components/ILIAS/Export/xml/SchemaValidation/ilias_exp_manifest_10_0.xsd#L59) to `<xs:pattern value="(components/[a-zA-Z0-9_]+|Services|Modules)/[a-zA-Z0-9_]+"/>`, I've created a corresponding ticket in Mantis), I find the question pool twice, once as a pool in trash, once as the actual missing pool, that needs to be derived).

Then a question, probably to @dsstrassner for when he comes back:
- Do we know, why we are only linking the first entry of the list to the corresponding question pool and are setting the rest to inactive links? this doesn't look right to me.

Thanks again and best,
@kergomard

lukas-heinrich:
Hi @kergomard !

I fixed the variable naming and removed the input field in the table cell. I did not change the mentioned line in the manifest file, because I think this was already done [here](https://github.com/ILIAS-eLearning/ILIAS/commit/38cccf7cd42dbf71ac1603aaa25394481728a851).

Best Regards,
@lukas-heinrich

kergomard:
Thank you very much @lukas-heinrich ! I merged and (contrary to my previous statements) picked it back to release_10. It is early enough and this way we keep the two systems close together to make fixing easier.

The unlinked titles are simply the way they are for now. Neither @dsstrassner nor myself remember how this came about.

Best,
@kergomard

PR #8264, opened 2024-10-25
Feature: New taxonomy filters to use taxonomies in the question selection for fixed tests (2)

matheuszych:
This is the implentation of the Feature Wiki entry [New taxonomy filters to use taxonomies in the question selection for fixed tests](https://docu.ilias.de/go/wiki/wpage_8158_1357).

I believe all the requested features are implemented according to the Request.

PR 2!


kergomard:
Thank you very much @matheuszych et al!

PR #8265, opened 2024-10-28
UI/Dropdowns: 42414, close all other Dropdowns

nhaagen:
https://mantis.ilias.de/view.php?id=42414


klees (approved):

thibsy (changes requested):
Hi @nhaagen,

Thx a lot for your contribution to the UI framework and solving the issue at hand.

We want this issue to be addressed soon, therefore we accept your proposal. However, we are not convinced that exposing a global array of dropdown menus is the best way to do so. Hence please:

- Roadmap: add a roadmap entry that describes we want a better solution for this. E.g. an event listener attached to the document, which either calls the dropdown's show method if the event target is the desired dropdown, or the hide method if the target is something else.

Kind regards,
@Amstutz and @thibsy

nhaagen:
Hi @thibsy ,
I changed to only store one dropdown, but of course, the issue is still the same.
Roadmap-entry is in.

thibsy (approved):
Thx a lot @nhaagen for improving the fix already. This LGTM, so I'll merge right away.

PR #8266, opened 2024-10-28
TA: 42251, result presentation uses full width if no best solution

nhaagen:
https://mantis.ilias.de/view.php?id=42251


kergomard:
Thank you very much @nhaagen ! Merged and picked to ILIAS 9 and 10.

PR #8269, opened 2024-10-28
Styles/TestRail Cases: separate objectives, build update or new import

nhaagen:
Apparently, we have to distinguish exisiting and new cases;
This creates separate files.

run `php setup/setup.php achieve style.buildUIFrameworkExampleTestCases.new -y` 
or respectively `php setup/setup.php achieve style.buildUIFrameworkExampleTestCases.update -y`
to build the XML.

### DIFF
--- a/components/ILIAS/Style/classes/Setup/TestRailXMLWriter.php
+++ b/components/ILIAS/Style/classes/Setup/TestRailXMLWriter.php
@@ -34,16 +34,32 @@ class TestRailXMLWriter
 
     protected SimpleXMLElement $xml;
 
+    protected int $no_found_ids = 0;
+    protected int $no_expexted_ids = 0;
+    protected int $no_new_ids = 0;
+    protected int $no_components = 2;
+
+    protected array $expected_by_caseids = [];
+
     public function __construct(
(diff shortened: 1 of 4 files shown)


thibsy (commented):
Hi @nhaagen,

Thanks a lot for your contribution to the UI components and your work on the automated test case generation!

- [x] `TestRailXMLWriter::$no_components`: is there a specific reason this value starts at `2` instead of `0`, like the other ones? Also, does `$no_*` refer to "number of"? If so, maybe clarify this.

- [x] Boolean checks: since `TestRailXMLWriter::$only_new_cases` is always a `bool`, you can simplify [these checks here](https://github.com/ILIAS-eLearning/ILIAS/pull/8269/files#diff-98a4fda8ff8dc9327c7c885c9240975074f86abfb36a725c6ca50751ebb15230R253-R254).

Kind regards,
@thibsy

nhaagen:
Hi @thibsy ,
the check if simplified and I added a comment concerning the strange "2"-number.

thibsy (approved):
Thx @nhaagen for clarification! This LGTM now, so I'll merge right away.

PR #8270, opened 2024-10-28
[FEATURE] ILIASObject: remove `ilProgressBar` usage

thibsy:
Hi @kergomard,

I have replaced the legacy `ilProgressBar` usage as fast and good as I could. I am not quite sure, if this code is actually reachable, because when copying multiple (nested) objects, the page simply loads without ever showing the `ilObjectCopyProgressTableGUI` table. Maybe you can try this out yourself?

Also note, I have replaced the legacy usage as is. I believe we should not render multiple progress bars on this page in the future, but instead use one progress bar to cover the progress of _all objects being copied_. I also needed to provide the `_max_steps` as an URL parameter, because there is (currently) no way to ping-pong the max-value of the progress bar.

Further note, this PR contains an additional commit which needs to be merged first. see #8268

Kind regards,
@thibsy

### DIFF
--- a/components/ILIAS/ILIASObject/classes/class.ilObjectCopyGUI.php
+++ b/components/ILIAS/ILIASObject/classes/class.ilObjectCopyGUI.php
@@ -27,8 +27,10 @@ use ILIAS\Style\Content\Container\ContainerDBRepository;
 use Psr\Http\Message\ServerRequestInterface;
 use ILIAS\UI\Factory as UIFactory;
 use ILIAS\UI\Renderer as UIRenderer;
+use ILIAS\Data\Factory as DataFactory;
 use ILIAS\UI\Component\Input\Container\Form\Standard;
 use ILIAS\Object\ImplementsCreationCallback;
+use ILIAS\HTTP\GlobalHttpState;
 
 /**
  * GUI class for the workflow of copying objects
@@ -73,6 +75,7 @@ class ilObjectCopyGUI
     protected ServerRequestInterface $request;
     protected UIFactory $ui_factory;
     protected UIRenderer $ui_renderer;
+    protected GlobalHttpState $http;
 
     protected ContainerDBRepository $container_repo;
 
@@ -112,6 +115,7 @@ class ilObjectCopyGUI
         $this->ui_factory = $DIC['ui.factory'];
         $this->ui_renderer = $DIC['ui.renderer'];
         $this->retriever = new ilObjectRequestRetriever($DIC->http()->wrapper(), $this->refinery);
+        $this->http = $DIC->http();
 
         $this->container_repo = new ContainerDBRepository($DIC['ilDB']);
 
@@ -1007,6 +1011,9 @@ class ilObjectCopyGUI
         );
 
         $progress = new ilObjectCopyProgressTableGUI(
+            new DataFactory(),
+            $this->ui_renderer,
+            $this->ui_factory,
             $this,
             'showCopyProgress',
             $ref_id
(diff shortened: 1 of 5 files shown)


kergomard (changes requested):
Hi @thibsy 

I finally got around to looking into this. I started by having a look at the functionality and sadly, this is not working as expected yet, so I would kindly ask you to have another look at it.

To be able to test it.
- You need a category (or a course) with multiple objects in it.
- You need to fix the soap-endpoint [here](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/WebServices/SOAP/classes/class.ilSoapClient.php#L89) (for me it was enough to just remove '/public' from the endpoint, ...but there might be more issues than that.

What isn't working (as far as I can see):
- The progress bar is never updated, but disappears right away.
- There should be a redirect once the copying has finished. We either never get there or it is not working.

I only read cursorily through the code and thus right now only have two minor, purely formal change requests:
- Could you please use the standard comparison order [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8270/files#diff-1b765c2f0199df0f5d37f344a634ea223bb388be576783eb0f9967917023a95cR1028). From where I stand code becomes harder to read this way around.
- Please move the usages of namespaces [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8270/files#diff-450e808ba299dc3219e4c06e28adcbeb412750ec02f98be0f2e0a4a1c9ae8a0cR30-R40) with aliases (DataFactory, UIRenderer, UIFactory).

Thanks again and best,
 @kergomard

kergomard:
Hi @thibsy 
Will you have a chance to have another look at this?

Thanks and best,
@kergomard

thibsy:
Hi @kergomard,

Yes, I will be able to have another look at this. However, I am working on other projects with higher priority right now, so it could still take a while.

I assigned myself and pass this back to you once I reworked the PR.

Thx and kind regards,
@thibsy

oliversamoila:
Dear @thibsy , 
after discussion with @kergomard , we will no more integrate this PR into the release of ILIAS 10. 
Please put the PR against the trunk again, take up the questions/feedback from Stepjhan and give me a sign for the realisation date. 
Many thanks and best regards
@oliversamoila

oliversamoila:
Dear @thibsy ,
I would like to ask how things are going with a version of this PR against the trunk?
Many thanks and best regards
@oliversamoila

thibsy:
Hi @kergomard,

I am pretty sure the new progress bar usage _would_ be working, but there seems to be something wrong with the copying process which I couldn't quite figure out. For some reason it always gets stuck at 50% of the required steps. And thats only after you avoid `ilSoapUtils:149` entirely. Probably a separate issue.

Redirect seems to be working, you can try this out by rendering a success state inside the `updateProgress()` endpoint manually after some time. I simply refactored the old approach a bit (much whitespace changes, sry), but this probably needs some more sophisticated solution sometime =). Its currently not possible to listen to some "finished" event on the progress bar(s), which might be a possible solution for things like this.

Let me know how you want to proceed. IMO we could also merge this and if there is something wrong besides the copying mechanism I will work on it again. Just so we can maybe ship this with ILIAS 11.

Kind regards,
@thibsy

kergomard:
Hi @thibsy 

Thank you very much for the update. I will apply a few changes to the code once this is finished, so this is ok for me for now.

There is one functional issue though:
- [ ] The current version does not redirect once the process has finished, it just stays on the page. I see the progress bar go to 100% and even have a check-mark, but it doesn't redirect.

For me copying did only work after a rebase.

Best,
@kergomard

kergomard:
Ok, as all plattforms seem to be broken in a different color: I merge this and then we fix issues, when and if they come up.

Thank you very much again @thibsy .

PR #8281, opened 2024-10-29
Issue #42137: All pre tags gets font-family monospace.

yvseiler:
https://mantis.ilias.de/view.php?id=42137


Amstutz:
thx a lot. @yvseiler , PRs for 9 and 10 would be highly appreciated.

PR #8283, opened 2024-10-30
COPage: set caption and text representation to...

lukastocker:
...default if empty while saving other media object settings like style class or alignment (#41399).

https://mantis.ilias.de/view.php?id=41399


alex40724:
Thanks for the PR. Fixed differently.

PR #8285, opened 2024-10-30
TA: 42371, check for authorized answer before setting status to progress

nhaagen:
https://mantis.ilias.de/view.php?id=42371


dsstrassner (approved):
Hi @nhaagen, thanks for your PR. I looked at it within a doil instance and from a concept aspect everything looks okay and as we discussed at the TechSquad Meeting.

@kergomard would you be so kind, to look into the code and merge and cherry pick to ILIAS 10 and Trunk?

Thanks in advance and best regards!

@dsstrassner

kergomard (approved):
This here gives me the heebie-jeebies ...and it is getting very expensive! ...but it is what it is and it looks right from a functional point of view, so: Merged and picked to ILIAS 10 and trunk.

PR #8286, opened 2024-10-30
UI/Modal/Lightbox: 42416, carousel navigation

nhaagen:
https://mantis.ilias.de/view.php?id=42416

small js to re-enable modal's carousel navigation


thibsy (changes requested):
Hi @nhaagen,

Thanks a lot for your contribution to the UI components, and special thanks for getting rid of another third-party library!

Before we can merge, please implement the following changes:

- [x] `class` usage: your methods don't really rely on `this`, therefore please move them outside of the `class` scope, into their own files, or introduce a new `Lightbox` `class` which wraps this logic.
- [x] Null-pointer(s): your JavaScript currently throws an error if the Lightbox is only rendered with one page, due to missing controls etc. Try to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) and/or proper checks on returned values in general.
- [x] Selector: please remove the `h1` from the [modal title selector](https://github.com/ILIAS-eLearning/ILIAS/pull/8286/files#diff-1d5f60cc2fc6cb7aa28969a06417304d425ccbd993e2b6316f074e1c5966b45cR131) to make it more robust to changes.

Kind regards,
@Amstutz and @thibsy

thibsy (approved):
Thx @nhaagen for implementing the requested changes and extracting the Lightbox-specific logic. This LGTM now, so I'll merge and pick right away.

PR #8287, opened 2024-10-30
LegalDocuments: Only Show Link When Accessible

kergomard:
Hi @mjansenDatabay 

When going through this mantis ticket: https://mantis.ilias.de/view.php?id=42485, I had the feeling that fixing the user-side was actually not enough. I would propose the following changes for Legal-Documents.

Best,
@kergomard


mjansenDatabay:
Thx Stephan.

PR #8292, opened 2024-10-30
T&A: Participant Table Revision Adjustments

thojou:
This PR introduces the following updates:

- **Client-IP Range Validation**: Ensures that the lower IP is less than or equal to the higher IP in the range.
- **German Translation Update**: Adjusts the translation for "Client-IP-Range."
- **Error Message Improvements**: Clarifies error messages for invalid participant selection on actions such as "Finish Test Attempt," "Delete Participant," and "Delete Results."

### DIFF
--- a/components/ILIAS/Test/src/Participants/ParticipantRepository.php
+++ b/components/ILIAS/Test/src/Participants/ParticipantRepository.php
@@ -22,7 +22,6 @@ namespace ILIAS\Test\Participants;
 
 use ILIAS\Data\Order;
 use ILIAS\Data\Range;
-use ILIAS\Test\Results\Data\StatusOfAttempt;
 
 class ParticipantRepository
 {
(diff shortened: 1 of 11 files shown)


kergomard:
Thank you very much @thojou !

This looks mostly good and I would have a single change request:
- Please add `TableAction::getSelectionErrorMessage(): ?string` to the interface, implement it on all Actions and remove the check for the existence of the function. If there is no message we should return `null` from where I stand and then only check for `null`.

Thanks again and best,
@kergomard

thojou:
Hey @kergomard,

thanks for the feedback. I’ve adjusted the code accordingly.

Best,  
@thojou

kergomard (approved):

kergomard:
Thank you very much @thojou ! Merged and picked to trunk.

PR #8293, opened 2024-10-30
Administration/Benchmark: Fix undefined array indexes

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=42494

If approved, this MUST be picked to `trunk` (and maybe also `release_9` and `release_8`).


fneumann:
Merged and tested successfully. Thx to @mjansenDatabay.

PR #8294, opened 2024-10-30
Auth/OpenIDConnect: Use `id_token` for claims

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=39471

This PR reverts a significant OpenID Connect change (at least for Microsoft providers) of PR #6306 .

The problem is: At least with Microsoft providers, not all claims are correctly provided in the response of the `UserInfo` endpoint. For instance we noticed that lists of values (e.g. role strings in a custom "roles" claim) are JSON-encoded **twice**.

Example `UserInfo` JSON Response:

```
{..., "roles":"[\\"Mitarbeitende\\"]", ...}
```

```php
var_dump(json_decode('{"roles":"[\\"Mitarbeitende\\"]"}'));
```

A JSON-decoded object of the response string looks like this:

```
object(stdClass)#1 (1) {
  ["roles"]=>
  string(17) "["Mitarbeitende"]"
}
```
The result has to be decoded **again** to finally retrieve the list of roles. This is **unexpected** and **not specified**.

```php
var_dump(json_decode(json_decode('{"roles":"[\\"Mitarbeitende\\"]"}')->roles));
```

```
array(1) {
  [0]=>
  string(13) "Mitarbeitende"
}
````
It seems this cannot be changed in Microsoft providers. Therefore I hereby propose to use the `id_token` like it was the case before the change applied with PR #6306 .

See also:

* https://learn.microsoft.com/en-us/entra/identity-platform/userinfo#consider-using-an-id-token-instead

### DIFF
--- a/Services/OpenIdConnect/classes/class.ilAuthProviderOpenIdConnect.php
+++ b/Services/OpenIdConnect/classes/class.ilAuthProviderOpenIdConnect.php
@@ -96,7 +96,7 @@ class ilAuthProviderOpenIdConnect extends ilAuthProvider
             $oidc->authenticate();
             // user is authenticated, otherwise redirected to authorization endpoint or exception
 
-            $claims = $oidc->requestUserInfo();
+            $claims = $oidc->getVerifiedClaims();
             $this->logger->dump($claims, ilLogLevel::DEBUG);
             $status = $this->handleUpdate($status, $claims);
 

### CODE REVIEW

**Comment by matthiaskunkel** [comment]
Jour Fixe, 11 NOV 2024: We highly appreciate this suggestion and accept the PR for 9 and higher.

**Comment by felix-itz** [comment]
And this change broke it again for OIDC provided by our Shibboleth server which does not send a suitable user name during authentication as `sub` and needs the extra tour with `requestUserInfo()`. Maybe a misconfiguration of our ShibbolethIDP. Who knows. I will patch this in our ILIAS instance again to requestUserInfo.

```
client_auth.ERROR: ilAuthProviderOpenIdConnect::doAuthentication:183 Could not determine valid external account, value is empty or not a string.
client_auth.ERROR: ilAuthProviderOpenIdConnect::handleUpdate:106 stdClass Object
```

mjansenDatabay:
Hi @felix-itz ,

> And this change broke it again for OIDC provided by our Shibboleth server which does not send a suitable user name during authentication as `sub` and needs the extra tour with `requestUserInfo()`. Maybe a misconfiguration of our ShibbolethIDP. Who knows. I will patch this in our ILIAS instance again to requestUserInfo.
> 
> ```
> client_auth.ERROR: ilAuthProviderOpenIdConnect::doAuthentication:183 Could not determine valid external account, value is empty or not a string.
> client_auth.ERROR: ilAuthProviderOpenIdConnect::handleUpdate:106 stdClass Object
> ```

Okay, then it is quite clear that we need a configuration option to decide whether to use th `id_token` or `requestUserInfo`.

If anyone (for instance, you or your organisation) is interested in pushing this forward (e.g. with a bit of :dollar:), don't hesitate to contact me (please not here).

PR #8298, opened 2024-10-31
Notifications: Don't delete all time-based settings on deactivation

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=42517

If approved, this could be merged down to `release_10`, `release_9` and also `release_8` .


iszmais:
Thank you for your contribution!

Greetings,
@iszmais

PR #8299, opened 2024-10-31
LegacyUI/Table: Fix usage of `YAHOO` library

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=42387


kergomard:
Thank you very much @mjansenDatabay and sorry ...just finished the YAHOO part in the table. ...columns still do nothing, will have a look at it now.

PR #8301, opened 2024-10-31
User: Fix SQL query in `_getUserIdsByInactivationPeriod`

mjansenDatabay:


kergomard:

PR #8308, opened 2024-10-31
OrgU: 42512, init genericMultiLineInput and contextualize renderer for glyph

nhaagen:
https://mantis.ilias.de/view.php?id=41621
https://mantis.ilias.de/view.php?id=42512


klees:
Picked to trunk.

PR #8314, opened 2024-11-04
T&A Fix ilAssQuestion list duplicated join

thojou:
This PR is related to Mantis https://mantis.ilias.de/view.php?id=42475.
During code review, i detected a duplicated code snippet in the `ilAssQuestionList`.
The duplicated join was likely introduced during previous rebasing and merging.
Removing this redundancy may increase performance.


kergomard:
Thank you very much @thojou ! I merge this and pick to trunk. Would you please then quickly check if it helped and close the issues if this is the case?

PR #8315, opened 2024-11-04
TA: 40076, check for actual contents of feedback to set flag in QPL table

nhaagen:
https://mantis.ilias.de/view.php?id=40076


kergomard (changes requested):
Thank you very much @nhaagen for this PR!

This looks good with one little change needed, I think:
- The column "feedback" is nullable with a default value of null in both feedback-tables. So [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/8315/files#diff-8348fef9ba500f473b75a6576bf8310d71df03acedaa247cc62441843c8e4b66R664) is not enough.

Thanks again and best,
@kergomard

nhaagen:
Yes, you're right - this is not actually failing, but it is a deprecation; trim will actually still accept null:
```
php > var_dump( trim(null) );
string(0) ""
```

I ensured the string, though

PR #8318, opened 2024-11-04
Certificate Bugfix #0042529: Certificate Standard Icon signWAC Error

fhelfer:
https://mantis.ilias.de/view.php?id=42529

cherry-pick to 10 as well

since icon_cert.svg is located outside of public/data/root in public/assets it cannot be signed


mjansenDatabay:
Thx @fhelfer

PR #8319, opened 2024-11-04
Test: Move Matching Question to Vanilla-JS

kergomard:
This PR move the matching question to a pure JS approach.

It is here so that @Amstutz can have a look at the changes in the scss-files of the test [here](https://github.com/ILIAS-eLearning/ILIAS/compare/trunk...kergomard:10/test/fix_matching_question?expand=1#diff-fd8ea3febcb3753d26f543e007bddf981690e2aa6b3879314ad0ec6d4b85c97eR629-R650).

I also remove one more fixed value from the previous dropzone definitions and moved around some colors, to make this work for all cases.

It would be very kind @Amstutz if you could have a look at this. Thank you very much in advance. Edit: Sorry, forgot the link to the lines that actually are of interest to you: https://github.com/ILIAS-eLearning/ILIAS/pull/8319/files#diff-fd8ea3febcb3753d26f543e007bddf981690e2aa6b3879314ad0ec6d4b85c97e

Best,
@kergomard


Amstutz:
Hi @kergomard 

Sry, I missed this one. The SCSS changes are fine. Thx for introducing/expanding the local variables. We already discussed the margin changes via discord. Thx for adding the comment explaining on why you are using the uncommon values for margins. 

The SCSS/CSS Changes are fine with me. Thx a lot!

Amstutz (approved):

kergomard:
Thank you very much @Amstutz !

PR #8320, opened 2024-11-04
T&A/8: 34365 Cloze Question Import (QPL) from 5.4 to 7 moves text from Question to Cloze Text

lukas-heinrich:
See https://mantis.ilias.de/view.php?id=34365

The issue is caused by an error in the interpretation of the version number. The method is designed for the version scheme of Ilias 5 ('x.x.x -dd.mm.yyyy'), but since version 6 only ('x.x - dd.mm.yyyy') is used and since version 8 only ('x.x').

I found a similar bug reporting and used it as a reference:

- https://mantis.ilias.de/view.php?id=39073
- https://github.com/ILIAS-eLearning/ILIAS/commit/81bc7b15d7362ad5fae7a4fac10a0b7ba611a352#diff-8e45538b112fbd1b4a72c93852ae30bc2ae4d8f5e7935d7f54aaeb288229ef8c

### DIFF
--- a/Modules/TestQuestionPool/classes/import/qti12/class.assClozeTestImport.php
+++ b/Modules/TestQuestionPool/classes/import/qti12/class.assClozeTestImport.php
@@ -48,43 +48,32 @@ class assClozeTestImport extends assQuestionImport
         ilSession::clear('import_mob_xhtml');
         $presentation = $item->getPresentation();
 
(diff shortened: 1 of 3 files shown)


kergomard:

I merge this as is, but please add the copyright checker to your Git-Hooks (or check the error messages from Github, which ever you prefer), this way we don't get failing tests because of failing copyright checks.

Thanks and best,
@kergomard

PR #8321, opened 2024-11-04
T&A/9: 34365 Cloze Question Import (QPL) from 5.4 to 7 moves text from Question to Cloze Text

lukas-heinrich:
This PR contains the fix for [34365](https://mantis.ilias.de/view.php?id=34365) for Ilias versions >= 9.

See #8320:

> The issue is caused by an error in the interpretation of the version number. The method is designed for the version scheme of Ilias 5 ('x.x.x -dd.mm.yyyy'), but since version 6 only ('x.x - dd.mm.yyyy') is used and since version 8 only ('x.x').

If I see this correctly, the commit can also be cherry-picked to `release_10` and `trunk`.

Thanks for the review and best regards,
@lukas-heinrich

### DIFF
--- a/Modules/TestQuestionPool/classes/import/qti12/class.assClozeTestImport.php
+++ b/Modules/TestQuestionPool/classes/import/qti12/class.assClozeTestImport.php
@@ -16,29 +16,29 @@
  *********************************************************************/
 
 /**
-* Class for cloze question imports
-*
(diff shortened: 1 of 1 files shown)


kergomard (commented):
Thank you very much @lukas-heinrich for the PR.

This looks right I merged and picked to ILIAS 10 and trunk.

Best,
@kergomard

PR #8322, opened 2024-11-04
T&A 41680: Fix sql data too long error by validating input parameter

lukas-heinrich:
Hi everyone,

This PR addresses an error including data loss when an answer text exceeds the length of the corresponding table column. 

See: https://mantis.ilias.de/view.php?id=41680
> I have implemented a fix that checks that the text does not exceed the maximum length before saving, so that no data is lost due to the error. However, this does not solve the problem that you cannot save the answer text. Therefore, the question here would be whether a limit of 1000 characters makes sense at this point?

This could also be cherry-picked to `release_9`, `release_10` and `trunk`

Thanks for the review and feedback in advance,
@lukas-heinrich

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assKprimChoiceGUI.php
+++ b/Modules/TestQuestionPool/classes/class.assKprimChoiceGUI.php
@@ -286,7 +286,6 @@ class assKprimChoiceGUI extends assQuestionGUI implements ilGuiQuestionScoringAd
         $kprimAnswers = new ilKprimChoiceWizardInputGUI($this->lng->txt('answers'), 'kprim_answers');
         $kprimAnswers->setInfo($this->lng->txt('kprim_answers_info'));
         $kprimAnswers->setSize(64);
-        $kprimAnswers->setMaxLength(1000);
         $kprimAnswers->setRequired(true);
         $kprimAnswers->setAllowMove(true);
         $kprimAnswers->setQuestionObject($this->object);
--- a/Modules/TestQuestionPool/classes/class.assMultipleChoiceGUI.php
+++ b/Modules/TestQuestionPool/classes/class.assMultipleChoiceGUI.php
@@ -812,7 +812,6 @@ class assMultipleChoiceGUI extends assQuestionGUI implements ilGuiQuestionScorin
         if ($this->object->getSelfAssessmentEditingMode()) {
             $choices->setSize(40);
         }
-        $choices->setMaxLength(800);
         if ($this->object->getAnswerCount() == 0) {
             $this->object->addAnswer("", 0, 0, 0);
         }
--- a/Modules/TestQuestionPool/classes/class.assSingleChoiceGUI.php
+++ b/Modules/TestQuestionPool/classes/class.assSingleChoiceGUI.php
@@ -692,7 +692,6 @@ class assSingleChoiceGUI extends assQuestionGUI implements ilGuiQuestionScoringA
         if ($this->object->getSelfAssessmentEditingMode()) {
             $choices->setSize(40);
         }
-        $choices->setMaxLength(800);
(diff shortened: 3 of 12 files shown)


kergomard (changes requested):
This is an important fix, thank you very much for the PR @lukas-heinrich !

I would have a few little change request:
- The same happens in KPRIM and single choice questions and they basically use the same input at its core. Could you please expand this solution to all of them.
- Could you please make sure we use the whole 1000 characters (right now it is set to 800 in multiple choice questions). I think we should set a fixed length in the base-class of the input and then remove all setting of the maxvalue (unless we have a very good reason for it).
- [This](https://github.com/ILIAS-eLearning/ILIAS/pull/8322/files#diff-f0e4d2345d5da8ea5606b40394a4ceb0e186878130b52b783d0b98316a9fe02dR79) is just for the case that somebody decides to do something funny, but I think we should use `mb_strlen()` here to at least make sure we do not cut the string off too early.

Thanks again and best,
@kergomard

lukas-heinrich:
Hi @kergomard 

Thank you very much for the review and the hint, I hadn't realized that the problem also exists in other classes. I have now made the following changes:

- The maximum length of 1000 characters is set in the super class `ilSingleChoiceWizardInputGUI` and I removed the setting of the value where it is no longer required. I have checked in the database tables that the fields maxlength for this question types is actually 1000 characters. 
- In the parent class (SC-Question) and in the child classes (MC-Question, Kprim-Question) I added the setting of the attribute in the template and the server-side validation by using `mb_strlen()`

When examining the child classes, I noticed that the fields in the Essay/Text-Question type also behave differently than expected:

- The issue here is that although the maxlength attribute is set in the HTML template, its value is only 200. I was also able to check in the database that the answer can contain up to 1000 characters. So I added the maxlength setting manually
- In addition, a server-side validation of the maximum length was missing at this point. I think it's a good idea to check the values also here, so I updated the `checkInput` method in the `ilTextWizardInputGUI.` I did this in a separate commit, since I believe that this part may be outside T&A.

Best Regards,
@lukas-heinrich

kergomard:
Merged and picked to ILIAS 9, ILIAS 10, and trunk.

lukas-heinrich:
Hi again @kergomard ,

Unfortunately, I made a  mistake and didn't test the cherry-picking locally well enough. There is a problem with the pick in the `release_10` branch because the local variable `$lng` was removed there during a refactoring. Therefore, this line needs to be adjusted :

```
diff --git a/components/ILIAS/TestQuestionPool/classes/class.ilSingleChoiceWizardInputGUI.php b/components/ILIAS/TestQuestionPool/classes/class.ilSingleChoiceWizardInputGUI.php
--- a/components/ILIAS/TestQuestionPool/classes/class.ilSingleChoiceWizardInputGUI.php	(revision 3a92dc66cefd306ea9ead74e7f927f8e49fca31c)
+++ b/components/ILIAS/TestQuestionPool/classes/class.ilSingleChoiceWizardInputGUI.php	(date 1730904003911)
@@ -246,7 +246,7 @@
                     }
 
                     if (mb_strlen($answervalue) > $this->getMaxLength()) {
-                        $this->setAlert($lng->txt("msg_input_char_limit_max"));
+                        $this->setAlert($this->lng->txt("msg_input_char_limit_max"));
                         return false;
                     }
                 }

```

The other branches and other GUI classes do not seem to be affected.

Sorry for the inconvenience!
@lukas-heinrich

kergomard:
Thank you very much @lukas-heinrich for testing this once more! That was my mistake not yours, ...and it will not have been the last one of this kind! Applied your fix with https://github.com/ILIAS-eLearning/ILIAS/commit/1e38bd88f3ccef5bebad2eea21376b2ba38246d8 and ported something similar to trunk.

PR #8329, opened 2024-11-05
ORGU: 42501, use multi byte string when writing the path into the sto…

lukastocker:
…rage. https://mantis.ilias.de/view.php?id=42501


klees:
Picked to release_9, release_10, trunk.

PR #8331, opened 2024-11-05
42321: Save standard search page number (release_9)

padvincenzo:
The page number is saved in Lucene search, but not in standard (direct) search.


smeyer-ilias:
Thanks for the PR. 
Unfortunately I saw it too late and your R8 PR is already merged.

PR #8332, opened 2024-11-05
42321: Save standard search page number (release_8)

padvincenzo:
The page number is saved in Lucene search, but not in standard (direct) search.


smeyer-ilias (approved):
Thanks @padvincenzo for fixing this

PR #8335, opened 2024-11-05
IASS: 42482, render the record template content from the iass setting…

lukastocker:
…s if member's record is empty. https://mantis.ilias.de/view.php?id=42482


klees:
Picked to trunk.

PR #8336, opened 2024-11-05
[Bugfix 42501]ORGU: use multi byte strlng for comparing the actual middle lengths a…

lukastocker:
…nd the fixed max middle length.


dsstrassner:
https://mantis.ilias.de/view.php?id=42501

klees:
Picked to release_9, release_10 and trunk.

PR #8337, opened 2024-11-05
UI Footer: 42441, adjust footer styling like following:

lukastocker:
- add top-border to maincrontrol footer
- remove border from footer grids
- add padding to footer grids

https://mantis.ilias.de/view.php?id=42441


Amstutz:
thx a lot!

PR #8339, opened 2024-11-05
T&A 41683/41684: remove superglobal access in question actions

lukas-heinrich:
This PR addresses an 'undefined array key' problem caused by accessing the Superglobal variables `$_GET` and `$_POST` directly. 

See: https://mantis.ilias.de/view.php?id=41683, https://mantis.ilias.de/view.php?id=41684

### Problem

The error occurs when the [Drop-In-Replacements for Superglobals](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/src/HTTP/README.md#dropinreplacements) is deactivated. In this case, the arrays are accessed directly, which leads to an error message, as the IDs are not always in `$_POST`, but sometimes also in `$_GET`. If Drop-In-Replacement is activated (default behavior), the issue does not occur because the wrapper class intercepts this error.

### Reproduction

To reproduce the behavior as described in the ticket, Drop-In-Replacements must be deactivated explicitly:

```ini
# client.ini.php

[server]
prevent_super_global_replacement = "1"
```

### Fix

I have fixed the issue by completely removing access to the request arrays in the relevant methods. Instead, I use the request object (`InternalRequestService`) which is already available in the class. This object provides methods for querying the required parameters, which already select the correct request array and safely access the array index. In addition, typecasting already takes place there.

I also refactored the methods for inserting, as there was a lot of duplicate code here.

### DIFF
--- a/Modules/Test/classes/class.ilObjTestGUI.php
+++ b/Modules/Test/classes/class.ilObjTestGUI.php
@@ -91,6 +91,9 @@ class ilObjTestGUI extends ilObjectGUI implements ilCtrlBaseClassInterface
 
     protected array $ui;
 
+    protected const INSERT_MODE_BEFORE = 0;
+    protected const INSERT_MODE_AFTER = 1;
+
     /**
      * Constructor
      * @access public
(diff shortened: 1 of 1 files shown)


thojou (approved):
Hey @lukas-heinrich,

thanks for the changes.
This looks good to me.

@kergomard Could you please have a second final look.

Best,
@thojou

kergomard (approved):
This is ILIAS 8 and 9 only, right? Then I'm all for it and thank you very much!

For 10 and above I would have some nitpicks, but I would think they would already be fixed there.

thojou:
Thanks @lukas-heinrich,

i've merged your changes into ILIAS 8 and cherry-picked the changes into ILIAS 9

Best,
@thojou

PR #8341, opened 2024-11-06
UI: 42431, remove margin to align action's contents evenly regardless…

lukastocker:
… of the object's level. 

https://mantis.ilias.de/view.php?id=42431


Amstutz:
thx!

PR #8342, opened 2024-11-06
UI: 41827, adjust tag input field height to auto resize the field on …

lukastocker:
…the basis of the content. 

https://mantis.ilias.de/view.php?id=41827


Amstutz:
thx a lot for the fix

- [x] can you recompile the css with the npm version of sass to avoid the rgb color encoding, thx.

PR #8347, opened 2024-11-06
fix Mantis #40887: Update LTI paths

c-knof:
It looks like LTI has been renamed to LTIProvider, but some paths have been forgotten to be renamed as well. 
Please check if I have found them all or if I have changed too many.


Saaweel:
Thank you very much for your pull request, but we have just corrected this bug a few hours ago.

https://github.com/ILIAS-eLearning/ILIAS/pull/8346/commits/5d4d6737d66532b0be3174bb695f249c8ff58664

PR #8349, opened 2024-11-06
Chatroom: Fix invite to chatroom for MyStaff and Meta Bar

lscharmer:
This PR fixes Mantis Bug: https://mantis.ilias.de/view.php?id=42546

Currently the link routes over the `ilRepositoryGUI` which redirects to the corresponding chatroom GUI, but the redirect doesn't keep the `usr_id` parameter. This PR directly calls the corresponding chatroom GUI instead.


mjansenDatabay (approved):

kergomard (commented):
Thank you very much @lscharmer ! 
Meged and picked to trunk.

PR #8360, opened 2024-11-07
Legacy Form: fix date range comparison for combination inputs

iszmais:
https://mantis.ilias.de/view.php?id=42053

This PR resolves an issue within the date range comparison of legacy forms.

##  Initial Problem
Within date range fields (combination fields of datetime with comparison mode) the comparison value of such datetime fields is not formatted correctly, causing the comparison to fail.
- e.g. `25.10.2024` is defined greater than `24.11.2036` since `25 > 24.`

The dates should be formatted into timestamps to be compared semantically correctly.

## Solution
The comparison provider method should use the serialized data since this is already transformed into a timestamp.
The solution must be provided for ILIAS 9+ as well.

### DIFF
--- a/Services/Form/classes/class.ilDateTimeInputGUI.php
+++ b/Services/Form/classes/class.ilDateTimeInputGUI.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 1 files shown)


iszmais:
@matthiaskunkel 
this is a technical change of behaviour and therefore would need a jour fixe approval.
However, I would strongly assume that no one ever really attempted to filter dates by `day > month > year` and that this was just overseen most of the time.

Therefore, I leave this decision to you. If obliged i will provide a presentation (or annotation) for the JF.

Greetings,
@iszmais

matthiaskunkel:
Jour Fixe, 11 NOV 2024: We highly appreciate this suggestion and accept the PR for 8 and higher.

mbecker-databay (approved):
Looks good to me, I approve.
I will leave this open until feb 5th in case anyone wishes to make additional remarks.

PR #6936, opened 2024-01-19
Block Removal of obsolete Code regarding old rendering

fhelfer:
Since old getHTML has been abandoned and force switched to getHTMLNew() for all Blocks, the old code fragments scattered in various components can be deleted.

PR #8365, opened 2024-11-07
Tutorial Support: Remove deprecated property

iszmais:
Due to the changes made in https://github.com/ILIAS-eLearning/ILIAS/pull/6936

this property is no longer used, since all renderings are now "new renderings". This might throw an error depending on the PHP Version due to the deprecation of dynamic properties.

(I recommend removing this in ILIAS 10, too)


smeyer-ilias:
Thanks @iszmais for fixing this. 
I 've pushed this to R10 and trunk

PR #8366, opened 2024-11-07
8 mantis ilias 42091, readds removed cache code in ilCalenderAuthenticationToken

chlulei:
Link: https://mantis.ilias.de/view.php?id=42091


smeyer-ilias (approved):
Thanks @chlulei

PR #8368, opened 2024-11-07
ContainerReference: Add SaveObject for LegacyForm

kergomard:
This adds the saveObject-function for all References.

See: https://mantis.ilias.de/view.php?id=42572


schmitz-ilias:
Thanks @kergomard !

PR #8372, opened 2024-11-08
Calendar: fix css of recurrence inputs (42329)

schmitz-ilias:
This PR fixes [42329](https://mantis.ilias.de/view.php?id=42329) by introducing a new css class for the recurrence input, such that the changes to legacy forms can be reverted 'locally'.

Are there any requirements on css style in legacy UI (naming conventions, etc.)?

### DIFF
--- a/Services/Calendar/templates/default/tpl.recurrence_input.html
+++ b/Services/Calendar/templates/default/tpl.recurrence_input.html
@@ -1,4 +1,4 @@
-<div class="form-inline">
+<div class="form-inline calrecurrenceinput">
 	{FREQUENCE}
 	<!-- BEGIN daily_block -->
 	<div id="sub_DAILY_1" style="padding-left: 40px;">
@@ -21,8 +21,8 @@
 		<label class="control-label" for="count_WEEKLY">{TXT_EVERY}</label>
 		<input type="text"
 			   class="form-control"
-				name="count_WEEKLY" 
-				value="{COUNT_WEEKLY_VAL}" 
+				name="count_WEEKLY"
+				value="{COUNT_WEEKLY_VAL}"
 				size="2"
 				maxlength="3"
 				style="text-align:right" />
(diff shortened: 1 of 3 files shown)


Amstutz:
@schmitz-ilias 

Thx for this fix. The fix is fine in general. For CSS Style in legacy UI, you mostly followed everything I would expect. Namely, no inline styling. No CSS classes by variables, placing the code into the right file etc... The naming guidelines for ITCSS (https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/templates/Guidelines_SCSS-Coding.md) are relatively difficult to expect / adapt for such cases. They are expected for UI Components.

However, there is one major question:

- [ ] Template Changes: You are changing the HTML template in a productive release. This is a breaking change that will lead to a fatal in any skin customizing this HTML template and not adopting it during the minor. I do not like to do this in a major release. So, would it be possible to achieve the fix with only CSS for 9? Would it be possible to only publish for 10 and trunk? Luckily, the template in question is certainly not the most commonly being overridden.

Thx for the feedback regarding this question, thx
@Amstutz

schmitz-ilias:
Hi @Amstutz, thanks for your feedback!

Without the change in the template, this can't be fixed I'm afraid. The template only uses pretty generic CSS classes (e.g. `control-label` and `form-control`), so without adding the additional class in the enclosing `div` it's not possible to make changes to recurrence inputs without making a mess out of all legacy forms. All alternatives I can think of (like tying the changes to IDs of inputs) are at least somewhat hacky and far less robust.

Limiting the fix to 10 and above would certainly be possible, but unfortunate. The current styling of recurrence inputs in 9 is ugly to the point of being a usability issue. How severe this is when weighed against the compatibility issues for skins I cannot say.

Can you elaborate a bit on how this would break skins? If I understand correctly, a skin could replace `tpl.recurrence_input.html` as a whole, right? Wouldn't the worst case then be that the replacement template does not contain the new class `calrecurrenceinput`, meaning the fix wouldn't apply?

Amstutz:
yeah you are right, I thought I saw I new variable in the template, but it is all just white spaces. Sry. In this case, there will not be much of an issue. If there is any new variable (or block), it would lead to a fatal in php, if the variable is not found in the template.

Amstutz:
picked to 10 and trunk as well.

PR #8373, opened 2024-11-08
Dependencies: Add `jumbojett/openid-connect-php` v. `1.0.2`

mjansenDatabay:
This PR adds `jumbojett/openid-connect-php` v. `1.0.2` as composer dependency for ILIAS 10.

Usage:
* `components/ILIAS/OpenIDConnect`

Reasoning:
* The library is used for `OpenID Connect`-based logins in ILIAS.
* When searching the web, it is the most recommended library.

Maintenance:
* There are 83 contributors.
* The library is actively maintained, there are recent commits and releases.
* While there seemed to be pauses in development and some specification impl. issues during the 0.* releases, development has **resumed** since end of 2023 with the first major release. Version 1.0.2 was recently released (mid of September 2024, the 1.0.1 release was published in early September 2024). New unit tests have been added.
* It basically consists of one file and has 1 composer dependency (smallish), so I see myself in a position to maintain it if the actual maintainers stop their activities.

Links:

Alternatives:
* Actually, I don't see any real alternatives for ILIAS 10.
* Many of the packages listed when searching (via https://packagist.org/ or search engines) are derivatives of the suggested library. Other libraries (like the ones checked below) have other drawbacks and are more complicated in their usage.
* https://github.com/socialconnect/auth : Worse maintenance, fewer contributors, huge package (the code and structure look indeed cleaner) with a lot of (unnecessary) code and more transitive dependencies.
* https://github.com/facile-it/php-openid-client/commits/master/ : Only 5 contributors, huge package with more transitive dependencies, little activity, but the code and structure look cleaner.
* https://bitbucket.org/PEOFIAMP/phpoidc/src/master/ No comment :-)
* https://packagist.org/packages/league/oauth2-client : We will probably suggest this for ILIAS 11, but this requires more refactorings (it is actually an `oauth` library, and we'll have to do much more `oidc` related stuff on our side) and therefore some funding. It introduces `paragonie/random_compat` as a transitive dependency. Using this for ILIAS 10.x would be too risky as well (in regard to security testing etc.).
* Of course we can spend weeks to implement our own OpenID Connect client, but this will further or later result in other libraries being suggested like `\Firebase\JWT\JWT` etc.pp. To be honest, I don't want that.

**Of course, it will be re-evaluated on the next dependency Jour Fixe in December 2024.**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -73,7 +73,8 @@
 		"phpunit/phpunit": "^10.5",
 		"monolog/monolog": "^2.9.3",
 		"phpoffice/phpspreadsheet": "^2.2",
-		"celtic/lti": "^5.0.0"
+		"celtic/lti": "^5.0.0",
+		"jumbojett/openid-connect-php": "^1.0"
 	},
 	"require-dev": {
 		"friendsofphp/php-cs-fixer": "^3.40",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "a76d50e264734da04e87304c7c061beb",
+    "content-hash": "beeb13326a9130db6f008b1e446fd528",
     "packages": [
         {
             "name": "apereo/phpcas",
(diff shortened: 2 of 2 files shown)


mjansenDatabay:
Bildschirmfoto vom 2024-11-11 19-47-06

PR #8374, opened 2024-11-08
UI: 41308, adjust rendering of symbols in breadcrumbs.

lukastocker:
https://mantis.ilias.de/view.php?id=41308

As 'direction: rtl;' is used to shorten the breadcrumbs from right to left we cannot use 'dir/direction ltr;' in spans or bdo/bdi tags as it will destroy the purpose of shortening inputs that are too long in regard to the screen size. It also will change the sortation of the single breadcrumbs (e.g.: repository will be rendered at last instead at first) in case of using '<span dir:'ltr'>'.

Therefore adding a left-to-right html character into the breadcrumb template was necessary.


Amstutz:
thx!

PR #8378, opened 2024-11-08
LSO: 42456, check child_ref_id before searching for full path in lear…

lukastocker:
…ning sequence event handler to avoid error message during the copy process.

https://mantis.ilias.de/view.php?id=42456


klees:
Already good in 9+.

PR #8385, opened 2024-11-11
Questions: Fix Questions LM

kergomard:
See: https://mantis.ilias.de/view.php?id=42548

While trying to find the right place for the different js- and css-files needed to render questions in LMs, I came up with this solution. I searched for usages of this classes all over ILIAS, but this is what I could find. I think this also starts clarifying responsibilities.

@qualitus-dahme and @alex40724 could you please have a look at this until the end of the week?

Thanks and best,
@kergomard


kergomard:
The error in the copyright checker is just a reminder that this must go.

alex40724 (approved):

PR #8386, opened 2024-11-11
UI: 42443, adjust padding for centering the filter's labels.

lukastocker:
https://mantis.ilias.de/view.php?id=42443


lukastocker:
After a short discussion with @yvseiler  we decided to center the label.
For this a "padding: x y" was added to the filter's label as the general label padding-right setting conflicted with centering the label.

Amstutz:
thx!

PR #8393, opened 2024-11-11
11/UI: Sequence Navigation

nhaagen:
Screenshot from 2024-11-11 13-58-35

This adds to the UI:
- a "Navigation" section
- a sequence player

The Sequence Navigation uses the same builder-pattern known from ui-tables: consumers will have to provide a binding
defining possible positions of the sequence (that has to be done at one point, anyways) and build a view/page/... = Segment with allowed contents.

Operating the sequnce will then display the segments one by one, also allowing for view controls as well as global and segment-specific actions.

This PR includes https://github.com/ILIAS-eLearning/ILIAS/pull/8436

### DIFF
--- a/components/ILIAS/UI/src/Component/Legacy/Factory.php
+++ b/components/ILIAS/UI/src/Component/Legacy/Factory.php
@@ -72,4 +72,28 @@ interface Factory
      */
     public function latexContent(string $content): LatexContent;
 
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *     The legacy segment is used as container for visible content in
+     *     the context of sequence navigations.
+     *     We currently lack quite a lot of UI components for describing the
+     *     actual contents of the page, e.g. questions, combinations of text/table/form, etc.
+     *     Until these components exist an enable us to better describe the
+     *     correlations to navigations, the legacy segment is used to contain
+     *     rendered HTML.
+     *   composition: >
+     *     The legacy segment contains html (or any other content) as string;
+     *     it also has a title.
+     *
+     * context:
+     *   - A segment is the content affected by operating the sequence navigation.
+     * ---
+     * @param string $title the title of the legacy segment
+     * @param string $content the content of the legacy segment
+     * @return \ILIAS\UI\Component\Legacy\Segment
+     */
+    public function segment(string $title, string $content): Segment;
+
 }
--- a/components/ILIAS/Init/Init.php
+++ b/components/ILIAS/Init/Init.php
@@ -122,6 +122,7 @@ class Init implements Component\Component
                 $pull[\ILIAS\UI\Implementation\Component\Progress\State\Bar\Factory::class],
                 $pull[\ILIAS\UI\Implementation\Component\Input\UploadLimitResolver::class],
                 $use[\ILIAS\Setup\AgentFinder::class],
+                $pull[\ILIAS\UI\Implementation\Component\Navigation\Factory::class],
             );
     }
 }
--- a/components/ILIAS/Init/src/AllModernComponents.php
+++ b/components/ILIAS/Init/src/AllModernComponents.php
@@ -94,6 +94,7 @@ class AllModernComponents implements \ILIAS\Component\EntryPoint
         protected \ILIAS\UI\Implementation\Component\Progress\State\Bar\Factory $ui_progress_state_bar_factory,
         protected \ILIAS\UI\Implementation\Component\Input\UploadLimitResolver $ui_upload_limit_resolver,
         protected \ILIAS\Setup\AgentFinder $setup_agent_finder,
+        protected \ILIAS\UI\Implementation\Component\Navigation\Factory $ui_factory_navigation,
     ) {
     }
 
@@ -170,6 +171,7 @@ class AllModernComponents implements \ILIAS\Component\EntryPoint
         $DIC['ui.factory'] = fn() => $this->ui_factory;
         $DIC['ui.renderer'] = fn() => $this->ui_renderer;
         $DIC['setup.agentfinder'] = fn() => $this->setup_agent_finder;
+        $DIC['ui.factory.navigation'] = fn() => $this->ui_factory_input_field;
     }
 
     public function getName(): string
--- a/components/ILIAS/UI/UI.php
+++ b/components/ILIAS/UI/UI.php
@@ -169,6 +169,9 @@ class UI implements Component\Component
             $internal[UI\Implementation\Component\Prompt\State\Factory::class];
         $provide[UI\Implementation\Component\Input\UploadLimitResolver::class] = static fn() =>
             $internal[UI\Implementation\Component\Input\UploadLimitResolver::class];
+        $provide[UI\Implementation\Component\Navigation\Factory::class] = static fn() =>
+            $internal[UI\Implementation\Component\Navigation\Factory::class];
+
         // =================================================================================
 
         $internal[UI\Implementation\Factory::class] = static fn() =>
@@ -203,6 +206,7 @@ class UI implements Component\Component
                 $internal[UI\Implementation\Component\Launcher\Factory::class],
                 $internal[UI\Implementation\Component\Entity\Factory::class],
                 $internal[UI\Implementation\Component\Prompt\Factory::class],
+                $internal[UI\Implementation\Component\Navigation\Factory::class],
             );
 
         $internal[UI\Implementation\Component\Counter\Factory::class] = static fn() =>
@@ -461,6 +465,13 @@ class UI implements Component\Component
             );
         $internal[UI\Implementation\Component\Prompt\State\Factory::class] = static fn() =>
             new UI\Implementation\Component\Prompt\State\Factory();
+
+        $internal[UI\Implementation\Component\Navigation\Factory::class] = static fn() =>
+            new UI\Implementation\Component\Navigation\Factory(
+                $pull[Data\Factory::class],
+                $pull[Refinery\Factory::class],
+                $use[UI\Storage::class],
+            );
 
         $internal[UI\Implementation\DefaultRenderer::class] = static fn() =>
             new UI\Implementation\DefaultRenderer(
--- a/components/ILIAS/UI/docs/ROADMAP.md
+++ b/components/ILIAS/UI/docs/ROADMAP.md
@@ -283,6 +283,15 @@ as "close" sign for modals. This character is not suitable for expressing this i
 we are actually using a "multiplication" sign. We should search for a better alternative and
 streamline these usages in order to avoid any A11y implications.
 
+### Move Components into Navigation (beginner, 2h)
+The top sections "breadcrumbs" and "menu" should both be moved into navigation.
+
+### Trait for usage of ILIAS\UI\Storage in Table and Sequence Navigation
+Storage of paramters in DataTable and SequenceNavigation look very much alike;
+in favor of those and further/future components the implementation should be
+realized as a trait to be used by several components.
+
+
 ## Long Term
 
 ### Mark Some Components as Internal
--- a/components/ILIAS/UI/src/Component/Legacy/Segment.php
+++ b/components/ILIAS/UI/src/Component/Legacy/Segment.php
@@ -1,0 +1,35 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\UI\Component\Legacy;
+
+use ILIAS\UI\Component\Component;
+use ILIAS\UI\Component\Navigation\Sequence\Segment as NavigationSegment;
+use ILIAS\UI\Component\Button;
+
+interface Segment extends Component, NavigationSegment
+{
+    /**
+     * Segments MAY add actions to the sequence.
+     * Those actions MUST target the actually displayed contents rather
+     * than changing context entirely (i.e. breaking the sequence).
+     */
+    public function withSegmentActions(Button\Standard ...$actions): static;
+}
--- a/components/ILIAS/UI/src/Component/Navigation/Factory.php
+++ b/components/ILIAS/UI/src/Component/Navigation/Factory.php
@@ -1,0 +1,75 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\UI\Component\Navigation;
+
+interface Factory
+{
+    /**
+     * ---
+     * description:
+     *   purpose: >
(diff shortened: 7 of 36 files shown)


klees:
Hi @nhaagen,

please rebase.

Kind regards!

components/ILIAS/UI/src/Component/Legacy/Factory.php
```diff
@@ -47,4 +47,28 @@ interface Factory
      */
     public function content(string $content): Content;
 
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *     The legacy segment is used as container for visible content in
+     *     the context of sequence nvigations.
```
klees:
```suggestion
     *     the context of sequence navigations.
```

klees:
Hi @Amstutz and @thibsy,

could you please move this forward? This is good to go from our perspective...

Kind regards!

yvseiler:
Hi @nhaagen 

I am very happy that this component has found its way into the UI framework. Many thanks for your efforts and those of others involved here.

Note: The following statements focus on visual and interaction design, as well as the descriptive texts of the “Expected Output” and a short HTML Validation test (the same as we use in core testing). The statements refer to the version of the CaT test platform provided to me (ILIAS v11.0_alpha). Many thanks for providing this platform.

- [x] Why are the buttons of the primary navigation and ViewControls displayed one below the other? I seem to remember that @catenglaender had worked out a suggestion here with a styled sticky primary navigation. Should that be part of this or was it or the styling intentionally omitted?
Bildschirmfoto 2025-04-07 um 10 02 12


- [x] The HTML Validator returns a few errors, but not all of them affect the component. I tried to extract the ones that belongs to the Sequence Navigation. Please check which of these apply to the Sequence Navigation component and can be corrected.

```
Error: Duplicate ID .

[From line 5590, column 21; to line 5590, column 69](https://validator.w3.org/nu/#l5590c69)

ss="well"><div class="c-sequence c-sequence--linear" id="">
```

```
Error: Bad value for attribute id on element [div](https://html.spec.whatwg.org/multipage/#the-div-element): An ID must not be the empty string.

[From line 5590, column 21; to line 5590, column 69](https://validator.w3.org/nu/#l5590c69)

ss="well"><div class="c-sequence c-sequence--linear" id="">
```

- [x] The text “expected output” refers to the representation of a "Characteristic Value Listing". In the meantime, a highlighted placeholder is displayed. Please adapt the text accordingly, e.g. “ILIAS shows a group of buttons and a placeholder for the segment”.
- [x] The button “a segment action for pos x” is missing in the text of expected outputs. Please add this button to the description. It could be helpful to add the sentence that the label adjusts at position 2 to support the fact that this button belongs to the corresponding segment.
- [x] Small linguistic adjustment in the text of expected output: Please change “vieww control” to “view control”.

As soon as the above question about the display of the primary navigation has been clarified, I will be happy to provide further feedback. Thanks!

kindly,
@yvseiler

thibsy (changes requested):
Hi @nhaagen,

Note: In addition to @yvseiler's review, this review will focus on the public interface so we can present this on JF.

- [ ] `C\Navigation\Sequence` namespace: could you explain why this level of hierarchy is necessary or beneficial?
- [ ] `Binding::getSequencePositions()`: could you explain why this method is necessary? Why not introduce a `::getSegments()` method that returns a `Generator<Segment>`? 

- [ ] Request coupling: we need to accept an `URLBuilder` and `URLBuilderToken` in the constructor of a `C\Navigation\Sequence\Sequence` component, for generating the URLs. You currently require consumers to always provide a request because you need to derive actions from the current request URL. I think this is something we should avoid in general, because consumers should be able to control URLs without having to rely on the current request URL. I understand we still need the `withRequest()` call for applying view controls, but we should do as little with the request as possible here.
- [ ] `C\Navigation\Sequence\Binding`: please rename this interface to `SegmentRetrieval` and maybe move it up one hierarchy level, depending on question above.
- [ ] `C\Legacy\Segment::withSegmentActions()`: it looks like this method rather belongs to `C\Navigation\Sequence\Segment`. Please move the method there and remove the redundant "segment" wording.

Kind regards,

UI-coordinators

catenglaender:
This is how the component looks now - if it looks any different on dev servers and environments, delos.scss needs to be recompiled and placed in the public/assets/css folder (my latest commit includes an up to date compile in templates/default/)

image

* Following up on some conversations I had both in-house and at the DevConf, the navigation has been moved back up to the header. This is to strengthen the Sequence Navigators main purpose of stringing together content without triggering an action. (Future child components and players might have a different focus and therefor make different decisions about the visual hierarchy.)
* several accessibility markers have been set to improve the UX with screen readers - this includes description texts which required new translation strings.
* screen readers have a general problem with our sr-only class and mixin. When space is too limited for the invisible div to spread, Orca on Ubuntu stops reading after a forced linebreak. Setting 'white-space: nowrap' on sr-only remedies this.
* tasks from @yvseiler's feedback are done

So with regards to rendering, I hope we are near the finish line now. :slightly_smiling_face:

nhaagen:
Hi @thibsy and coordinators,
thanks for digging through this.

> * [ ]  `C\Navigation\Sequence` namespace: could you explain why this level of hierarchy is necessary or beneficial?

I think it is very likely that other navigation-elements will follow; as examples, I added Breadcrumbs and Menu to the roadmap, but I expect even more to come.

> * [ ]  `Binding::getSequencePositions()`: could you explain why this method is necessary? Why not introduce a `::getSegments()` method that returns a `Generator<Segment>`?

Segments might be quite huge, with a lot of dependencies and queries to the db;  for the navigation itself, this is irrelevant but for the actual to-be-rendered segment.
I admit that the example suggests otherwise, however, only an array of identifiers is actually needed.

> * [ ]  Request coupling: we need to accept an `URLBuilder` and `URLBuilderToken` in the constructor of a `C\Navigation\Sequence\Sequence` component, for generating the URLs. You currently require consumers to always provide a request because you need to derive actions from the current request URL. I think this is something we should avoid in general, because consumers should be able to control URLs without having to rely on the current request URL. I understand we still need the `withRequest()` call for applying view controls, but we should do as little with the request as possible here.

We had some internal discussion about this. I understand the issue, but would still refrain:
The navigation itself should stay "in place", i.e. just operate on the view it was rendered in.
I don't see any reason, though, to determine that place during construction, and so pin the navigation to
a single view. As you say, at least when view controls are involved, the request is needed anyways,
and it is sort of unnecessarily cumbersome to manually create the URLBuilder.
The Segments, however, may still be delivered entirely unrestricted.

> * [ ]  `C\Navigation\Sequence\Binding`: please rename this interface to `SegmentRetrieval` and maybe move it up one hierarchy level, depending on question above.

I'll do that :)

> * [ ]  `C\Legacy\Segment::withSegmentActions()`: it looks like this method rather belongs to `C\Navigation\Sequence\Segment`. Please move the method there and remove the redundant "segment" wording.

Actually,  Navigation\Sequence\Segment only needs to _read_ actions; I'd reckon, they are usually built internally upon already existing ones. 
The Legacy\Segment, however, might well add some, since there is no actual exisiting content.
In the end, exisiting components should implement Navigation\Sequence\Segment.


Best regards,
Nils

thibsy (changes requested):
Hi @nhaagen,

Thx a lot for your explanation.

- [x] `C\Navigation\Sequence` namespace: that makes sense, thx!
- [x] `Binding::getSequencePositions()`: I think we understand the necessity of this method now, thx. Due to the rename to `SegmentRetrieval`, it might make sense to rename this method as well, e.g. simply `getAllPositions()`.
- [x] Request coupling: your explanation makes sense. However, currently rendering a sequence navigation without request will throw an exception. Therefore, please either add a usage rule to the factory description, or accept e.g. an `Data\URI` in the constructor / factory to pin-point the view, as you suggest.
- [x] `C\Legacy\Segment::withSegmentActions()`: that makes sense, thx!

After this we can present the interface on JF.

Kind regards,
@Amstutz and @thibsy

matthiaskunkel:
Jour Fixe, 28 APR 2025: We highly appreciate this suggestion and accept this new UI element for trunk.

nhaagen:
Not even merged yet, but already a change on the interface:
When I was using the Sequence with a Form in a Segment, there were certainly several ways to get the request to the Retrieval somehow, but it felt quite weird to do so. Also, this seems to be a basic shortcoming that will occur every time we build segments with some sort of interaction. E.g, considering a form failing its validation, the form should be displayed again, highlighting the errors. The binding has to take care of that when building the Segment.
Considering the wide range of other possible usecases, it's probably best to just hand over the request.

klees:
Hi @thibsy and @Amstutz,

we are currently in the process to actually use this UI component for something (the manual correction in the test). When working on this, it became appearant to @nhaagen, that he would require the `Request` in a segment, since he wanted to build a form in that sequence.

Since we currently haven't codified any non-legacy-segments, we first wondered if this was an accident due to the use of the legacy segment. This is not the case: Even if, say, the form would be a proper segment, it would require the request to perform processing and, e.g., show errors.

We then discussed options, how the request could be made available in the `SegmentBuilder`:

1. Have it as a property on the `SegmentBuilder` and access it via `$this`.
2. Pass it via a construct such as `$additional_params` as we do with the DataRetrieval for tables.
3. Pass it as a distinct parameter.

We discarded 1. because the Sequence Navigation itself has a `withRequest`-method. If the `DataRetrieval` has a property for the request, these two could get out of sync. Something similar goes for 2., since parameters would need to be bound manually in that case, again with the potential problem of out of sync. With 3., the Seguence Navigation could pass the received Request on its own, which would guarantee they are in sync. Also: Other then e.g., for the Data Table, we expect that the Segments contain some interactivity that is only loosely structured, which would require the Request anyway. Passing the Request then is just the way to make that interactivity generally possible.

Hope this makes some sense.

Kind regards!

thibsy (changes requested):
Hi @nhaagen,

Thanks a lot for your follow-up on the request handling - your explanation sounds reasonable.

Please note I have abbreviated the class names and prefixed them with a `C` for the public interface and an `I` for the implementation. Hope they make sense.

- [x] `C\Legacy\Segment::withSegmentActions()`: I was still confused in this iteration why this method exists exactly. The `C\Legacy\Segment` inherits from `C\Sequence\Segment`, which already provides a `withActions()` function with a seemingly similar purpose. If we really need two distinct versions of this method for the two components, I think some documentation and a different name on the `C\Legacy\Segment` would be beneficial.
- [x] `I\Sequence\Sequence::withViewControls()`: could you explain why the custom transformation is necessary? Could you also check/explain, why this works? IMO providing this transformation _last_ will cause problems if there already are custom transformations, provided by the consumers. If this is true, we should probably not add this transformation.

- [x] `C\Sequence\Segment`: the getters of this interface should be internal. Either move them to an internal interface for segments, or simply drop them on the public interface. Additional question: did you name them with a redundant "segment" intentionally, so they will not conflict with components that may inherit this? If this won't be a problem in your opinion, please remove the word from the method names.
- [x] `C\Sequence\Segment::getSegmentActions()`: do we really need the distinction between an empty `array` and `null`? I think we can always return an `array` here to simplify the handling of the actions state (reduce from 3 to 2 states).
- [x] `ILIAS\UI\Storage` logic: the implementation for using the storage looks very similar to the one used for data tables. Could we move this into a trait, where an abstract method for the prefix must be implemented by usages? If this is out of scope, you may add this to the roadmap instead.

- [ ] Covariance: please double check that every public interface (inside `ILIAS\UI\Component`) return-type-hints interfaces and implementations return-type-hints concrete implementation (inside `ILIAS\UI\Implementation\Component`) - there are wrong declarations. You can read more about [interface segregation and covariance in the UI framework](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/UI/docs/segregation-and-covariance.md). Renderers should e.g. also check for the implementation instead of the interface.
- [x] `I\Sequence\Sequence::withRequest()`: I think we can reduce the amount of side-effects this function call has upon its instance by moving the logic of `initFromRequest()` to the renderer. I don't think we really need to calculate the current position and generate URL's on the sequence. Therefore, please:
	- [ ] move the position handling to the renderer and drop the `getCurrentPosition()`, `withCurrentPosition()` and `getNext()` methods.
	- [ ] remove the URL-builder and -token properties.
	- [ ] and merge the remaining `initFromRequest()` method into `withRequest()`, which should only leave you with setting the request and applying it to the view controls.
- [x] `C\Sequence\SegmentRetrieval`: please accept the `ServerRequestInterface` as first argument. We tend to accept complex objects before the more trivial ones in the UI framework. 
- [x] `UI\FactoryInternal`: you need to add the new factory to this interface too. You also need to `$provide` the new factory in the `UI.php` and inject it into the `Init\AllModernComponents` bridge.
- [x] `I\Sequence\Sequence`: please type-hint to the `ILIAS\UI\Storage` instead of `ArrayAccess` in the constructor.
- [x] `\ilAchievementsGUI`: I think there's an unrelated change in there, please remove it.
- [x] `ROADMAP.md`: there is some leftover from a git conflict in there, please remove it.
- [x] Unit tests: please implement some rendering and other relevant PHPUnit tests for this component and adhere our [testing guideline](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/UI/README.md#How-to-write-unit-tests-for-a-Component).

Kind regards,
@Amstutz and @thibsy

nhaagen:
I moved the position handling to the renderer, but cannot say that it looks better than before; especially the missing refinery on the renderer is bugging me.

thibsy:
@nhaagen are you waiting for feedback here? Just making sure we are not in a deadlock. If this is ready for the next round, you can reassign me.

klees:
Hi @thibsy,

regarding `I\Sequence\Sequence::withRequest()`: I think I see where you coming from, but I would still prefer if we could leave things as they have been before. I had a look into the actual [change that was spawned by the change request](https://github.com/ILIAS-eLearning/ILIAS/pull/8393/commits/e2452e00afa5572dd9ef607bbd2483297ec0318b) with @nhaagen.

Maybe you had something different in mind than that, but I think we can observe that the `Renderer`  gains some weight regarding code and knowledge about data processing of the component. To me this is too much. Why, e.g., should the `Renderer` know how to reach in the query? Code like 

```php
        $position = $query->retrieve(
            $token_position->getName(),
            $refinery->byTrying([
                $refinery->kindlyTo()->int(),
                $refinery->always(0)
            ])
        );
```

looks super accidental regarding the initial task of the `Renderer`: Turn a component into according HTML. I'd rather have that code in the component itself.

Two other things speak to the previous solution: It is super similar to the way we do things in the forms. Data retrieval and processing is also handled inside the component there. And: Having that code in the component makes it easier to write test for that code.

So, can we put things back the way that they have been before? Even more, code like

```php
        if ($position >= count($positions) || $position < 0) {
            $position = 0;
        }
```

would also belong into the component from my POV.

Ping me,if we shall discuss this personally.

Kind regards!

thibsy:
Hi @klees,

Thx for your feedback here. I agree, this is probably too much for a renderer. 

While thinking about the side-effects again, I also noticed that they may not be implemented in the component anymore, but when rendering the component they still occur. So, I guess getting rid of the side-effects only shifted them to a less obvious location now. Good ol' "Verschlimmbesserung" =).

@nhaagen sorry for the extra roundtrip, you can revert the according change.

Kind regards,
@thibsy

nhaagen:
Hi @thibsy ,
I think this is ready for the next round :)
Thanks in advance,
Nils

thibsy (changes requested):
Hi @nhaagen,

Thx for implementing the changes, and sorry for the extra roundtrip. Also thx for implementing the unit tests using our described technique.

Please consider the following suggestions. We do not enforce them at the moment, so they are highly optional - I just wanted to point it out:

- [x] `I\Legacy\Renderer`: we should check the `instanceof` the implementation, not the interface. Otherwise we might accidentally render custom components.
- [x] `I\Navigation\Sequence\Renderer` same here.

- [x] `C\Navigation\Sequence\Sequence::withActions()`: there is a type-hint missing for what kind of actions can be provided. Probably `C\Button\Standard`?
- [x] `I\Legacy\Factory::segment()`: the return type-hint must be `Segment` (implementation not interface).
- [x] `I\Navigation\Sequence\Sequence::getViewControls()` here on the other hand we return something we received from the consumer, which might be custom, so we cannot treat this as an internal component later on. Change the return type-hint to `IViewControlContainer`.

We can merge this directly afterwards. I labeled this for JF presentation, because there have been some small interface changes.

Kind regards,
@thibsy

thibsy (approved):
Perfect, thx @nhaagen for implementing the last changes. I will merge this right away.

matthiaskunkel:
Jour Fixe, 01 SEP 2025: The PR has already been merged. But @thibsy wanted to notify us about the latest interface changes in this PR, see statement by Richard from 20MAY.

PR #8359, opened 2024-11-07
[FIX] Add Buttons and Modals separately to Footer

chfsx:
This PR changes the interface at the updated `Footer` component. Currently, it is not possible to add buttons with modals to a link group. The current method `withAdditionalModalAndTrigger` always inserts such buttons in a separate section (the same as `withAdditionalLink`).
Therefore, I suggest removing the specific method `withAdditionalModalAndTrigger` and only offering a method `withAdditionalModal`. This is only necessary until there is a general way to add modals to the final page in ILIAS.
Please note the following:
- In the case of an emrge, trunk would currently fail because GlobalScreen still uses the old method, so I would like to deliver a second commit that uses the new method before the merge. We are welcome to coordinate with each other on this.
- The changes also have to be picked in release_10.

image

PR #8396, opened 2024-11-11
LegalDocuments: Use updated footer api to provide links & modals

lscharmer:
This PR changes the `LegalDocuments` to use the new footer API based on this PR: https://github.com/ILIAS-eLearning/ILIAS/pull/8359.

This PR fixes Mantis Bug: https://mantis.ilias.de/view.php?id=42339 (Show goto links in the footer for anonymous user)

### DIFF
--- a/components/ILIAS/UICore/classes/Screen/PageContentProvider.php
+++ b/components/ILIAS/UICore/classes/Screen/PageContentProvider.php
@@ -171,11 +171,6 @@ class PageContentProvider extends AbstractModificationProvider
         return $subtab_title . $tab_title;
     }
 
-    public function getFooterModification(CalledContexts $screen_context_stack): ?FooterModification
-    {
-        return $this->globalScreen()->layout()->factory()->footer()->withModification(fn(?Footer $footer): ?Footer => $this->dic['legalDocuments']->modifyFooter($footer));
-    }
-
     /**
      * @deprecated this is needed as long as the PageContentProvider is the only place which stores the permalink
      */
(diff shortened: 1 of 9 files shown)


components/ILIAS/UICore/classes/Screen/PageContentProvider.php
```diff
@@ -173,7 +173,7 @@ private function buildTabTitle(): string
 
     public function getFooterModification(CalledContexts $screen_context_stack): ?FooterModification
     {
-        return $this->globalScreen()->layout()->factory()->footer()->withModification(fn(?Footer $footer): ?Footer => $this->dic['legalDocuments']->modifyFooter($footer));
+        return $this->globalScreen()->layout()->factory()->footer();
```
chfsx: return `null` or remove the whole method `getFooterModification` would be better here.

chfsx (changes requested):
LGTM in general, only one small change needed in `PageContentProvider`, I'd recommend to remove the method completely since a `FooterModification` is no longer needed.

components/ILIAS/UICore/classes/Screen/PageContentProvider.php
lscharmer: You are right. I removed the method.

chfsx:
Thanks a lot! @thibsy I hope the change in components/ILIAS/UICore/classes/Screen/PageContentProvider.php is fine for you as well since this file is in UICore.

chfsx (approved):

thibsy (approved):

PR #8400, opened 2024-11-11
Exercise: fix setting of property lp_user_id (42651)

schmitz-ilias:
This PR fixes [42651](https://mantis.ilias.de/view.php?id=42651), by setting a property in `ilObjExerciseGUI` properly. There have been no changes in those lines of code since ILIAS 8, not sure why it's triggered an error now and not before.


schmitz-ilias:
@smeyer-ilias, had a closer look, this broke because of `ilCtrl` refactorings in `ilLearningProgressBaseGUI`. During `ilLearningProgressBaseGUI::executeCommand` there is a redirect that loses parameters like `user_id`. This could have other side effects. (I still think the fix here in Exercise is also worthwhile.)

alex40724:
This is fixed in the meantime. Thanks.

PR #8401, opened 2024-11-12
UI/DataTable: with additional view control

nhaagen:
Consumers may add more view controls to the data table;
example and tests are based on https://github.com/ILIAS-eLearning/ILIAS/pull/8350, so this will be failing for now.


nhaagen:
PR #8350 is included, now, this one here does not really make much sense without it.
Furthermore, there are changes in the namesourcing of ViewControls, which of course will affect all VCs.
Basically, it comes down to "a VC will have fixed parameter names". The container's name, however, might still be numbered.
While this of course shortens and prettifies the URL, the main reason is another one: VCs are singletons for their controlled view - it simply does not make sense to have more than one pagination, e.g.
Fixed paramter names make NullControls widely obsolete - there will be no more shifting in param-names when a viewcontrol is omitted due to runtime circumstances (no pagination needed for very little records, no sorting option available, etc...)
Actually, I'm thinking of removing the null control alltogether.

klees:
Hi @thibsy and @Amstutz,

please have a look into this, LGTM to me now. Please note the comment from @nhaagen, that this already contains #8350.

If you need some more intro to the code we can maybe discuss this tomorrow as well.

Kind regards!

thibsy (commented):
Hi @nhaagen,

Although this review already covers some implementations details, this is not yet a full technical review; we will do so to a latter point.

- [ ] Dedicated names: you introduce a new mechanism which does not share the same features with other inputs, because dedicated names will not be enumerated. Can you explain why we need this? Could you also explain, why every view control input receives a dedicated name right in the factory? It feels like this drifts form inputs and view controls farer apart, and I think we should try to avoid this.
- [ ] `mixed` parameters: could you explain why exactly the parameters of the `DataRetrieval` used for view controls become `mixed`? Same for `Data::withAdditionalParameters()`. Do we really need to expect any kind of value there? Why don't we pass this data as an array anymore? If I understood correctly, these parameters should be somewhat similar to data retrieved by `Form::getData()`, no?
- [ ] `Data::withAdditionalParameters()`: can you explain what this method is used for briefly? Maybe a comment would help.

Kind regards,
@Amstutz and @thibsy

nhaagen:
Hi @thibsy ,
actually glad you're asking, these were some discussion points with @klees , too

> * [ ]  Dedicated names: you introduce a new mechanism which does not share the same features with other inputs, because dedicated names will not be enumerated. Can you explain why we need this? Could you also explain, why every view control input receives a dedicated name right in the factory? It feels like this drifts form inputs and view controls farer apart, and I think we should try to avoid this.

We currently have enumerated params just like in form-fields; however, the internal fields of the view controls are fixed and not open to consumers; there is no need to enumerate them, and besides the shorter url, expectable parameters are simpler to handle. Foremost with the session storage: if there was, e.g., no pagaination, we filled up the now empty parameter-spaces with null-controls, just to shift to the correct parameter name. This seems more brittle than just dedicating names. Enumeration takes place on the container, though; we now have something like "vc0/pag/offset", and that's quite speaking, without a chance of collision. Well, it _would_ collide, if there were more than one pagination - but that in itself would be wrong on so many levels that I do not expect it to happen.

> * [ ]  `mixed` parameters: could you explain why exactly the parameters of the `DataRetrieval` used for view controls become `mixed`? Same for `Data::withAdditionalParameters()`. Do we really need to expect any kind of value there? Why don't we pass this data as an array anymore? If I understood correctly, these parameters should be somewhat similar to data retrieved by `Form::getData()`, no?

Actually, with transformations attached to the form(fields), there might be just anything in the parameters; "mixed" stresses this fact and allows for higher level objects than arrays in the future.

> * [ ]  `Data::withAdditionalParameters()`: can you explain what this method is used for briefly? Maybe a comment would help.

The original thought was to already construct the table with some environment vars the dataRetrieval might need, and just pass it around with "batteries included". It's just a way of attaching something to the table.
But you are right, a comment would help :)

Best regards,
Nils

thibsy (changes requested):
Hi @nhaagen,

Thanks a lot for your explanations.

Please note, before you take any action, we also reviewed the PR included here, which may cause some changes too. So it's probably best to make an iteration on #8350 first.

- [x] `Table\Data::withAdditionalViewControl()`: could you explain how the handling of this input works exactly? Maybe provide a dedicated example. I am not quite sure, what `withAdditionalViewControlData()` is used for. Why is this necessary? Maybe specify in a comment as well.
- [x] `Table\Data::withAdditionalViewControl()`: how do we ensure to not accept duplicate view controls? I know that names will not collide, but it would still not be logical to accept a view control which already exists, right?
- [ ] `Table\Data::getViewControls()`: is there a specific reason why you are [setting the table ID as dedicated name](https://github.com/ILIAS-eLearning/ILIAS/pull/8401/files#diff-1894392fb23ad26ef97a486053bf64261ac5dfd8dd7c8d8ce24e29bfa705b06dR191-R193)?

- [x] `mixed` type hints: I understand the logic behind this; however, this PR is quite intrusive while the actual benefit of it is low. Maybe we should put some more thought into this and apply this change to all of the input-related things to a latter point. Therefore, please revert this change and add the `$additional_viewcontrol_data` parameter in the `DataRetrieval` as last one and make it optional (default value).
- [x] View control names: please document the behaviour of the applied names somewhere, maybe the `Input\Factory`. Please also add a usage rule, describing how each individual view control should only be used once for its context.
- [x] `Input\ViewControl\Factory`: please set the dedicated names inside the constructor of view controls, instead of the factory. There are already some duplicates now.
- [x] `Container\ViewControl\Factory`: please use the `FormInputNameSource` when creating view control containers. Otherwise, we will run into naming collisions when they are provided with dedicated names. Maybe rename `ViewControlNameSource` to `ViewControlInputNameSource` for clarity.
- [x] Conflicts: please rebase onto the latest trunk to resolve conflicts. Make sure the fix of #8829 is properly applied here too.

Kind regards,

@Amstutz and @thibsy

klees:
Hi @thibsy and @Amstutz,

regarding "`mixed` type hints": How strong is your resentment here? We have made the change for two reasons:

* The new parameter order seems to be more natural regarding the position of the various controls in the table.
* `?array` is actually wrong, at least for the filter, since it can not respect any transformations that are attached to the filter.

Then, the best moment to change an interface often is "now", since the usages will only increase from now on. The second reason even is a borderline bugfix. So, can we seize the day?

Kind regards!

thibsy:
We (UI coordinators and @nhaagen) discussed the `DataRetrieval` interface change and agreed that it will actually prevent errors in the future, because the return type of `Input\Container::getData()` already is `mixed`. This is due to `Refinery\Transformation` which could be applied to the container itself, making `getData()` return something other than an `array`.

@nhaagen we have just merged #8350, could you rebase this onto the latest trunk and check if #8829 is properly applied? I was referencing the wrong PR in my last comment. Thx!

nhaagen:
> @nhaagen` we have just merged #8350, could you rebase this onto the latest trunk and check if #8829 is properly applied? I was referencing the wrong PR in my last comment. Thx!

I was wondering already - but, yes, they are _both_ applied :)

thibsy:
@nhaagen is this ready for another round? If so, please also try and answer our questions.

nhaagen:
Hi @Amstutz and @thibsy,
I rebased and think this is good for another iteration.

> * [x]  `Table\Data::withAdditionalViewControl()`: could you explain how the handling of this input works exactly? Maybe provide a dedicated example. I am not quite sure, what `withAdditionalViewControlData()` is used for. Why is this necessary? Maybe specify in a comment as well.

There is an [example](
https://github.com/nhaagen/ILIAS/blob/11/UI/DataTable/additionalVCs/components/ILIAS/UI/src/examples/Table/Data/with_additional_viewcontrols.php) now showing the additional view control(s);
withAdditionalViewControlData is used to relay the data to DataRetrieval, so getRows may evaluate on it; however, I switched it to protected - no need to set it from outside.

> * [ ]  `Table\Data::withAdditionalViewControl()`: how do we ensure to not accept duplicate view controls? I know that names will not collide, but it would still not be logical to accept a view control which already exists, right?

In the example, I added some mode-controls and another field selection; that might be used to control details of shown data or to separate certain aspects into different controls to work in parallel, as not to extend lists with every possible combination. So, I'd really not llimit to one kind of VC.

> * [ ]  `Table\Data::getViewControls()`: is there a specific reason why you are [setting the table ID as dedicated name](https://github.com/ILIAS-eLearning/ILIAS/pull/8401/files#diff-1894392fb23ad26ef97a486053bf64261ac5dfd8dd7c8d8ce24e29bfa705b06dR191-R193)?

I had two tables on one view; without further separation via id, the request targeted both.

Best regards,
Nils

thibsy (commented):
Hi @nhaagen,

Thanks a lot for the clarification!

The problem with multiple tables on one page you mention:

> I had two tables on one view; without further separation via id, the request targeted both.

is a conceptual issue of view controls. I have outlined the problem [in mantis just now](https://mantis.ilias.de/view.php?id=44478). The way you address the issue wouldn't work, because table IDs are optional. I'm afraid until we have came up with a solution for the conceptual issue, we cannot support multiple data tables on the same page.

Beside this, I am still somewhat concerned about the view control naming changes of this PR. Let me explain why:

You are opening the possibility to provide any kind of view control input to the data table. While the data table is in full control over the "original" view controls, it is not in control over the "custom" ones. This alone is not a problem, but it becomes problematic due to your changes to the naming of view control inputs:

Assume a consumer provides "custom" view control of the same type as an "original" view control. This will work and can be rendered, but it will lead to unexpected results when they are operated, because they will be rendered and operated under the same name. You also have to think about the future development of the data table. What if we'd like to offer some default mode controls someday? We'd have to assume consumers are already providing their own mode controls, which either prevents us from adding one or will lead to errors for these consumers once this feature is rolled out. Of course, this is also true without the naming issue, but it would at least stay functional. I'd also argue that fixed names will increase the odds of conflicting with existing query parameters which are relayed using the current request.

I therefore ask you to revert the changes regarding the naming of view controls. Maybe we can address this while looking for solutions to the problem I have outlined.

Since this is an implementation detail, we can still go to JF with this.

Kind regards,
@thibsy

thibsy:
@nhaagen I pass this back to you for the time being.

matthiaskunkel:
Jour Fixe, 31 MAR 2025: We highly appreciate this PR and accept it for trunk.

klees:
Hi all,

I have discussed this with @thibsy today. We have agreed on the following

* The code regarding naming and the code regarding additional view controls shall be separated into two PRs. The problem with naming turns from latent to acute once we introduce additional view controls, but it exists anyway. So we shall solve and discuss this in two PRs, which in fact shouldn't influence each other (too much). The changes regarding additional view controls should be good to go then.
* For the code regarding naming, we would like to introduce a method on the `NameSource`-interface, that is supposed to spawn a new `NameSource` of a similar type then the old one, which is reset. This would replace `$source = new (get_class($source))();` then. We probably will need this for the filters as well. The problem we try to solve here is as such: Suppose the "same" view control exists on a page twice, for whatever reason. Say its the sortation of a table and we have two tables on the page. If we would use the same `NameSource` for both tables, the sortations would be named: `table_1.sortation_1` and `table_2.sortation_2`. This is, in fact, how forms work right now. But other than with forms, these names are visible for view controls, as they appear in the URL. And this looks odd, because the we only have one sortation per table. So better would be a naming scheme where `table_1.sortation`  and `table_2.sortation` would be created, which is what the recreating of the `NameSource` in fact does.

Dear @nhaagen, please proceed accordingly =)

Thanks a lot @nhaagen and @thibsy!

nhaagen:
This now only adds the addtional ViewControls (and amends consumers);
Dedicated naming of ViewControls is [here](https://github.com/ILIAS-eLearning/ILIAS/pull/10267),
and conflicting input names are [here](https://github.com/ILIAS-eLearning/ILIAS/pull/10269).
There is also an issue with Signals when there are more ViewControls in Containers - [here](https://github.com/ILIAS-eLearning/ILIAS/pull/10271).

components/ILIAS/UI/src/Component/Input/Factory.php
```diff
@@ -154,6 +154,7 @@ public function container(): Container\Factory;
      *      1: View Controls MUST reside in a View Control Container.
      *      2: View Controls MUST be visually close to the visualization their operation will have effect upon.
      *      3: View Controls MUST effect one visualization only.
+     *      4: View Controls allready have dedicated names. Those names SHOULD NOT be overridden.
```
thibsy: This change belongs to the view-control naming change, please remove it from this PR.

components/ILIAS/UI/src/examples/Table/Data/with_additional_viewcontrols.php
```diff
@@ -0,0 +1,157 @@
...
+use ILIAS\UI\Implementation\Component\Table as T;
+use ILIAS\UI\Component\Table as I;
+use ILIAS\Data\Range;
+use ILIAS\Data\Order;
+use ILIAS\UI\URLBuilder;
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * ---
+ * expected output: >
```
thibsy, line 19: The description is missing from this example.

thibsy (changes requested):
Hi @nhaagen, thx for separating the different kinds of changes.

I have noticed two nitpicks which I commented inline. In addition, please:

- [ ] `sortation.js`: I am not entirely sure why these changes are necessary. Do they belong to this PR? If not, please remove them.
- [ ] Conflicts: please rebase this onto latest trunk.

Kind regards,
@thibsy

nhaagen:
> * [ ]  `sortation.js`: I am not entirely sure why these changes are necessary. Do they belong to this PR? If not, please remove them.
Options are from either the viewControl or the column header and differ slightly in their format. this allows for both.

> * [x]  Conflicts: please rebase this onto latest trunk.

thibsy (approved):
Thx @nhaagen for implementing the changes so quickly.

This LGTM now, so I will go ahead and merge this, before some data retrieval causes conflicts again =).

Kind regards,
@thibsy

PR #8402, opened 2024-11-12
UI/Modal/LightBox: 42429, always show modal before initializing carousel

nhaagen:
https://mantis.ilias.de/view.php?id=42429


Amstutz:
thx a lot!

PR #8408, opened 2024-11-12
Style/TestCases: update path to KS Docu and drop references-tag

nhaagen:
Fabian asked me to amend the path to the UI docu in the testrail-cases and leave out the references-tag (both relevant not for this season, but the next round in 11).


Amstutz:
thx!

PR #8412, opened 2024-11-12
T&A Bugfix #0042396: Fix Test Scoring Settings for Import/Export

fhelfer:
https://mantis.ilias.de/view.php?id=42396

Similar to https://github.com/ILIAS-eLearning/ILIAS/commit/e05188483564bedf5f4e176713b060185e5456fb

### DIFF
--- a/Modules/Test/classes/class.ilObjTest.php
+++ b/Modules/Test/classes/class.ilObjTest.php
@@ -69,7 +69,6 @@ class ilObjTest extends ilObject implements ilMarkSchemaAware, ilEctsGradesEnabl
      */
     public $mark_schema;
     public int $sequence_settings;
-    public int $score_reporting;
     public int $instant_verification;
     public int $answer_feedback_points;
     public $reporting_date;
@@ -303,7 +302,6 @@ class ilObjTest extends ilObject implements ilMarkSchemaAware, ilEctsGradesEnabl
         $this->introduction = "";
         $this->questions = array();
         $this->sequence_settings = TEST_FIXED_SEQUENCE;
-        $this->score_reporting = self::SCORE_REPORTING_FINISHED;
         $this->instant_verification = 0;
         $this->answer_feedback_points = 0;
         $this->reporting_date = "";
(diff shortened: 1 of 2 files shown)


Modules/Test/classes/class.ilObjTest.php
```diff
@@ -5254,6 +5254,7 @@ public function fromXML(ilQTIAssessment $assessment)
         }
 
         $this->saveToDb();
+        $result_summary_settings = $result_summary_settings->withShowPassDetails($result_details_settings->getShowPassDetails());
```
fhelfer, line 5236: By default, the value of `results_presentation` in the `tst_tests` table is set to `3` if it is not specified. This results in `showPassDetails()` always being true.

Subsequently, `result_presentation` will be set to the corresponding import value (in my case, `0`). However, due to this line 
https://github.com/ILIAS-eLearning/ILIAS/blob/a54812f6c900046a80f48f2675620d54ce182b6f/Modules/Test/classes/ScoreReporting/class.ilObjTestScoreSettingsDatabaseRepository.php#L126-L128
, it will always be set to `1`. The reason for this is that PassDetails is true because it inherits the still-existing default value of 3 from the database.

@nhaagen, you seem to be the author of this feature. Please take a look at this and see if there is a better solution.

Best,
@fhelfer

fhelfer: This seems to also be an issue for newer releases

nhaagen (changes requested):
Hi @fhelfer,

I wonder why those fields were left out  (i.e. if there was a reason) when moving the  settings into a repo, but then I cannot remember or think of any, so apparently  - they are just missing.

As a result, there are different ways of altering the values, as setters on the object might not be made in the repo and otherwise round.

Exactly as you did  [here](https://github.com/fhelfer/ILIAS/blob/802f6622be0832266f45ae0371c27540b44e2a7d/Modules/Test/classes/class.ilObjTest.php#L5097-L5098), the repo should be the single source of truth.

- [x]  However, please also align the [getter](https://github.com/fhelfer/ILIAS/blob/802f6622be0832266f45ae0371c27540b44e2a7d/Modules/Test/classes/class.ilObjTest.php#L1803)  to read from the repo, akin to ::getScoreCutting or ::getCountSystem,

- [x]  and use it [here](https://github.com/fhelfer/ILIAS/blob/802f6622be0832266f45ae0371c27540b44e2a7d/Modules/Test/classes/class.ilObjTest.php#L5475)

- [x] Also, please remove the [setter](https://github.com/fhelfer/ILIAS/blob/802f6622be0832266f45ae0371c27540b44e2a7d/Modules/Test/classes/class.ilObjTest.php#L1698-L1708) and all references to it entirely.

This is a little bit more complicated for the results_presentation (since there are different settings combined in it), but imo the best way, still.

Thank you and best regards,
Nils

nhaagen (approved):

kergomard (approved):
Thank you very much @fhelfer ! This looks good, yes.

This has already been fixed in ILIAS 9 and above.

PR #8416, opened 2024-11-13
OrgU: 42682, orgu_ref must be of type int

nhaagen:
https://mantis.ilias.de/view.php?id=42682
please pick to 9 and 8, too.


klees:
Picked to release_8 and release_9 *and trunk too* =)

PR #8421, opened 2024-11-13
T&A/10 40791: Replaces superglobal usages in Test component with RequestDataCollector

lukas-heinrich:
Hi everyone, 

this is the revised and rebased version of #7803 by @matheuszych:

> Superglobal variables such as $_GET or $_POST should not be accessed directly. Instead, use the request and refinery mechanisms. Currently, there are four instances where this change is not feasible due to limitations in the request and refinery systems, but these will be addressed in the future.

[Mantis: 40791](https://mantis.ilias.de/view.php?id=40791)

We are looking forward to feedback!
Best,
@lukas-heinrich

### DIFF
--- a/components/ILIAS/Test/classes/class.ilTestSkillLevelThresholdsGUI.php
+++ b/components/ILIAS/Test/classes/class.ilTestSkillLevelThresholdsGUI.php
@@ -18,6 +18,8 @@
 
 declare(strict_types=1);
 
+use ILIAS\Test\RequestDataCollector;
+
 /**
  * @author		Björn Heyser <bheyser@databay.de>
  * @version		$Id$
@@ -39,6 +41,7 @@ class ilTestSkillLevelThresholdsGUI
         private readonly ilGlobalTemplateInterface $tpl,
         private readonly ilLanguage $lng,
         private readonly ilDBInterface $db,
+        private readonly RequestDataCollector $request_data_collector,
         private readonly int $test_id
     ) {
     }
@@ -86,20 +89,20 @@ class ilTestSkillLevelThresholdsGUI
 
     private function saveSkillThresholdsCmd(): void
     {
-        if (strtolower($_SERVER['REQUEST_METHOD']) == 'post') {
+        if (strtolower($_SERVER['REQUEST_METHOD']) === 'post') {
             $assignment_list = $this->buildSkillQuestionAssignmentList();
             $assignment_list->loadFromDb();
 
             $valid = true;
 
             $table = $this->getPopulatedTable();
-            $elements = $table->getInputElements((array) ($_POST['rendered'] ?? []));
+            $elements = $table->getInputElements($this->request_data_collector->retrieveArrayOfStrings('rendered'));
             foreach ($elements as $elm) {
                 if (!$elm->checkInput()) {
                     $valid = false;
                 }
 
-                $elm->setValue($_POST[$elm->getPostVar()]);
+                $elm->setValue($this->request_data_collector->strVal($elm->getPostVar()));
             }
 
             if (!$valid) {
@@ -111,9 +114,9 @@ class ilTestSkillLevelThresholdsGUI
             $threshold = [];
             foreach ($elements as $elm) {
                 $key = $elm->getPostVar();
-                $value = $_POST[$key];
                 $matches = null;
                 if (preg_match('/^threshold_(\d+?):(\d+?)_(\d+?)$/', $key, $matches) && is_array($matches)) {
+                    $value = $this->request_data_collector->int($key);
                     $threshold[$matches[1] . ':' . $matches[2]][$matches[3]] = $value;
                 }
             }
@@ -129,7 +132,7 @@ class ilTestSkillLevelThresholdsGUI
                 $thresholds_by_level = [];
 
                 foreach ($levels as $level) {
-                    if (isset($threshold[$skillKey]) && isset($threshold[$skillKey][$level['id']])) {
+                    if (isset($threshold[$skillKey][$level['id']])) {
                         $skill_level_threshold = new ilTestSkillLevelThreshold($this->db);
 
                         $skill_level_threshold->setTestId($this->getTestId());
--- a/components/ILIAS/Test/classes/class.ilTestSkillLevelThresholdsGUI.php
+++ b/components/ILIAS/Test/classes/class.ilTestSkillLevelThresholdsGUI.php
@@ -96,7 +96,7 @@ class ilTestSkillLevelThresholdsGUI
             $valid = true;
 
             $table = $this->getPopulatedTable();
-            $elements = $table->getInputElements($this->request_data_collector->retrieveArrayOfStrings('rendered'));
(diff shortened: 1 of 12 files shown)


components/ILIAS/Test/classes/class.ilTestSkillLevelThresholdsGUI.php
```diff
@@ -86,20 +89,20 @@ public function getTestId(): int
 
     private function saveSkillThresholdsCmd(): void
     {
-        if (strtolower($_SERVER['REQUEST_METHOD']) == 'post') {
+        if (strtolower($_SERVER['REQUEST_METHOD']) === 'post') {
```
thojou: To further encapsulate global access, we could consider implementing an `isMethod(string $method)` in the `RequestDataCollector`. 

What are your thoughts on this approach, @kergomard?

kergomard (changes requested):

An special thank you for the added typing!

- [ ] Please keep the name of the functions `RequestDataCollector::retrieveArrayOf*FromPost()` exactly as they were. The reason for this naming: It is wrong that we don't know where the data comes from. A value from the query should only be used for navigation for example. The name was explicitly chosen to make this clear. We should expand this to all functions, i.e. we should make clear for all where the data comes from and there are different ways to do that, but I think we should take one step at a time. If you have a better way to do that: Please do! Otherwise I would kindly ask you to keep the naming.
- [ ] Please deprecate the function `RequestDataCollector::getPostKeys()` and remove `RequestDataCollector::getQueryKeys()`. This is nothing that anybody needs to know as far as I can see. The current usages in any case have clearly a very strong smell to them.
- [ ] I very much dislike the method `RequestDataCollector::isMethod()` and I'm currently rather sure that we should simply proxy the `ServerRequest::getMethod()` function. The important argument for this is that we cannot apply ANY changes to the method (eg. `strtolower`). The RFC is very clear: The request method is case sensitive (See: https://www.rfc-editor.org/rfc/rfc7230 and this is still the same in HTTP/2 https://www.rfc-editor.org/rfc/rfc7540) and the methods we use are currently called 'POST' and 'GET' (not 'post', not 'Post'). We could create an `Enum` for the methods and thus give some more meaning to the `isMethod()` function. But I think, we shouldn't do that, as I think if we need it, the`Enum` should come to us and not be created by us.
- [ ] Please change [this](https://github.com/ILIAS-eLearning/ILIAS/pull/8421/files#diff-62d7c1a79c7015945e5e0e22ae4d83beef9fdfad30be6baa23803e6d58e266f6R630) to `ilObjTest::_getResultPass()` thus making very clear that we are using a static here. And: Thank you very much for correcting this!
- [ ] Please do not use none `bool` values as `bool`'s (e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8421/files#diff-eb107de08e0a9134113438fa1319c62882481cccd58d471723e39bad9d904f4eR2765)).

Thanks again and best,
@kergomard

kergomard:
Hi @lukas-heinrich 

As I just saw yoour commit while following the Jour Fixe and you are surely still working on this:
- Could you please remove the `strlen()` from [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8421/commits/38b2db5c002e72180497d3a604b975fbc30ab505#diff-c269786e11c8b44db07d34a714ce820dea8a2b10f4fe2bc8e5ce79dee4005410R135). This should be `!== ''`. You don't want to know how long this string is, you just want to know, that it is not an empty string.
- In general, to make me happy, use simple comparisons wherever you can. So no `empty()` if you are sure to only check for `null` or `''` no `strlen()` if you want to know that something is `!== ''`. The reason for this is quick readability and making things as explicit and as specific as possible.

Thank you and best,
@kergomard

thojou:
Hi @kergomard,

Thanks for your insights! I have a few thoughts to share:

> * [ ]  I very much dislike the method `RequestDataCollector::isMethod()` and I'm currently rather sure that we should simply proxy the `ServerRequest::getMethod()` function. The important argument for this is that we cannot apply ANY changes to the method (eg. `strtolower`). The RFC is very clear: The request method is case sensitive (See: https://www.rfc-editor.org/rfc/rfc7230 and this is still the same in HTTP/2 https://www.rfc-editor.org/rfc/rfc7540) and the methods we use are currently called 'POST' and 'GET' (not 'post', not 'Post'). We could create an `Enum` for the methods and thus give some more meaning to the `isMethod()` function. But I think, we shouldn't do that, as I think if we need it, the`Enum` should come to us and not be created by us.

I also think, that introducing an Enum for HTTP methods, could add clarity and type safety to the isMethod() function. But this shouldn't happen on our side.

That said, I have a slightly different take on handling HTTP method casing, with the following reasoning:

* While HTTP methods are case-sensitive per specification, in practical scenarios, it's highly unlikely for a developer to intentionally use something like `Post` instead of `POST`.
* At a lower level, libraries like `GuzzleHttp\Psr7\Request` already normalize HTTP methods to uppercase (`strtoupper`). Therefore, any comparison to a non-uppercase method will inherently fail.

To address this, I suggest retaining the transformation within `RequestDataCollector::isMethod()` but modifying it to use `strtoupper` instead of `strtolower`. This would ensure consistency with the underlying API while maintaining compliance with the HTTP standards.

Looking forward to your thoughts!

Best,
@thojou

kergomard:
Hi @thojou 

> At a lower level, libraries like GuzzleHttp\Psr7\Request already normalize HTTP methods to uppercase (strtoupper). Therefore, any comparison to a non-uppercase method will inherently fail.

Yep, and it should. And what guzzle does is actually wrong (please see [here](https://github.com/php-fig/http-message/blob/master/src/RequestInterface.php#L72-L85)). It is only a "SHOULD NOT", but still. So, please proxy the `Request::getMethod()`-function and do the checks locally.

And one more little thing:
- The changes  in your last commit @lukas-heinrich go in the right direction, but not quite far enough ;-). What does counting characters give us (logically I mean)? We don't need the `strlen()` at all, what you want is `$message !== ''`. This makes it very easy to read and avoids me having to know how a function behaves exactly and what changes go with with which version (Could you tell me from the top of your head in which version `strlen()` stopped accepting `null` or since when `count()` expects a countable?). Sorry if I'm very neurotic about these things. The goal is to not introduce code that is more taxing to understand than absolutely necessary.

Best,
@kergomard

lukas-heinrich:
Hi @kergomard ,

I fully agree, the readability is much better this way than when comparing length. Apparently I wasn't attentive enough when revising yesterday.

I have also removed the conversion to lower case in `isMethod`. In the places where it is necessary to check the HTTP method, I strongly recommend doing this with capital letters. On the one hand because, as @thojou  mentioned, the library already converts to upper case. And on the other hand because RFC 7231 explicitly defines the standard HTTP methods in upper case.

Best,
@lukas-heinrich

kergomard:
Ok, I merge this now. I will quickly change `isMethod()` myself. I thought of another variant that you might like more and also works. I will rename it to `isPostRequest()`. This way we get rid of the part I don't like (I need to know a string to check for a method without any discoverability what valid strings would be) and I think we can keep what you do like about it (a function that does the checking, if I'm getting it right) ;-). I will pick this to trunk as well.

Thanks for all the efforts!
@kergomard

PR #8423, opened 2024-11-13
[FIX] #42452 UI: select `File\Dropzone` file input correctly.

thibsy:
Hi folks,

This fixes an issue where the `File\Dropzone` does not delegate files properly to the actual `Field\File` input, by amending the DOM-search to the new inputs HTML structure.

See https://mantis.ilias.de/view.php?id=42452

Kind regards,
@thibsy


klees:
Picked to trunk.

PR #8424, opened 2024-11-13
TA: 39183, insert marker for error-text positions w/o correct text

nhaagen:
https://mantis.ilias.de/view.php?id=39183
Screenshot from 2024-11-13 14-45-05


dsstrassner:
Thanks for your PR @nhaagen, but I agree with @kergomard. We should not do this. That the issue 39183 was not closed by me, is the result of the aftermath of the ILIAS 9 release. The PM asked me to postpone some issues, that we reach the goals for the release, and I simply forgot about this issue.

PR #8430, opened 2024-11-13
42176: Shuffling Answer options does not work in Matching Questions in ILIAS Learning Modules

fneumann:
https://mantis.ilias.de/view.php?id=42176

The relevant change is at the bottom in the function toJSON. The shuffler has to be initialized again for the definitions. Otherwise they would be shuffled in the same way as the terms before.  I guess this could be a general problem of the shuffle transformation.


kergomard:
Thank you very much @fneumann for the PR!

This looks right, just please remove the comment. If you want to add a note, add it to the commit message and link more clearly to the Mantis issue there. I will then merge asap.

Thank you very much and best,
@kergomard

fneumann:
Thanks for the review @kergomard 
I removed the comment from the code.

kergomard:
Thanks for the update @fneumann. Merged and picked to ILIAS 9, ILIAS 10, and trunk.

PR #8432, opened 2024-11-14
User/Fields/Registration: Make `email` mandatory for active passwo…

mjansenDatabay:
…rd generation

See: https://mantis.ilias.de/view.php?id=42689

TestRail Test Case: https://testrail.ilias.de/index.php?/tests/view/79500

If approved, this must be picked to `release_10`.

There will be a second PR for the `Registration` componet to show a "Messagebox > Info" if the "email" is not a required/visible field for the self-registration context.

### DIFF
--- a/components/ILIAS/User/classes/Profile/class.ilUserProfile.php
+++ b/components/ILIAS/User/classes/Profile/class.ilUserProfile.php
@@ -19,7 +19,6 @@
 declare(strict_types=1);
 
 use ILIAS\User\Profile\DefaultFields;
-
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
Just one idea @kergomard : If this class is "touched" again in a bigger future refactoring (with funding, of course), it could be worth implementing a kind of "Visitor Pattern" approach to get rid of these context-specific modifications here.

kergomard:
Thank you for the PR @mjansenDatabay ! Merged and picked to release_10.

PR #8437, opened 2024-11-14
Form/Calendar: move datetime and duration inputs to native html (42458)

schmitz-ilias:
This PR fixes [42458](https://mantis.ilias.de/view.php?id=42458) by using native html `datetime-local`/`date` inputs in `ilDateTimeInputGUI` and `ilDateDurationInputGUI`. This removes the dependency on bootstrap for the datetime picker, and allows us to get rid of a bunch of js and surrounding php code. Some js is still needed to restore previous functionality of duration inputs (setting minimum for the end date dynamically to the value of the start date, shifting the end when changing the start to keep the difference, full day toggle), this PR rebuilds that part of it for the new input types.

Note that some configurations of the inputs then don't apply anymore: since this relies on the browser-specific implementations for the picker, custom configs like for `ilBirthdayInputGUI` are not possible. I also removed the unused option to show seconds in `ilDateTimeInputGUI` (supporting this would technically be possible, but is a bit fiddly).

Lastly, step size for minutes is still configurable, but behaves quite a lot differently than in the bootstrap implementation. The new behavior should be discussed, but I don't think it's worth delaying this fix as a whole because of this. I will open a separate Mantis ticket (**Edit:** [42740](https://mantis.ilias.de/view.php?id=42740)) explaining this issue further.

Cheers, @schmitz-ilias

### DIFF
--- a/components/ILIAS/Calendar/classes/class.ilCalendarUtil.php
+++ b/components/ILIAS/Calendar/classes/class.ilCalendarUtil.php
@@ -512,158 +512,36 @@ class ilCalendarUtil
         $tpl = $DIC->ui()->mainTemplate();
 
         if (!self::$init_datetimepicker) {
-            $tpl->addJavaScript("assets/js/moment-with-locales.min.js");
+            //$tpl->addJavaScript("assets/js/moment-with-locales.min.js");
             // unminified version does not work with jQuery 3.0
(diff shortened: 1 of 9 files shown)


alex40724:
I tested several fields and did not experience any issues. Thanks!

alex40724:
Merged to trunk as well.

PR #8440, opened 2024-11-15
T&A 42760 Fix has answered questions for scored attempt 

thojou:
This PR addresses https://mantis.ilias.de/view.php?id=42760.

During a recent refactoring, a new method was introduced to check for scored attempts. However, it appears that a typo occurred in the implementation. Specifically, instead of calling `getAttemptOverviewInformation` to verify if the reference exists,
 the code mistakenly used `hasAttemptOverviewInformation`.

To resolve this, I refactored the code to use the null-coalescing operator, ensuring the proper method is utilized and the reference is correctly checked.

Best,
@thojou


kergomard:
Thank you very much @thojou ! Merged and picked to trunk.

PR #8441, opened 2024-11-15
TA: 38995, sort taxonomies in table and filter

nhaagen:
Use taxonomy sorting in question pool table and filter
https://mantis.ilias.de/view.php?id=38995

### DIFF
--- a/Modules/TestQuestionPool/classes/QuestionTable.php
+++ b/Modules/TestQuestionPool/classes/QuestionTable.php
@@ -112,7 +112,7 @@ class QuestionTable extends ilAssQuestionList implements Table\DataRetrieval
                 $tax = new ilObjTaxonomy($tax_entry['tax_id']);
                 $tax_tree = $tax->getTree();
                 $sortfield = $tax->getSortingMode() === ilObjTaxonomy::SORT_ALPHABETICAL ? 'title' : 'order_nr';
-                $children = $this->tax_node_reader($tax_tree, $sortfield, $tax_tree->readRootId());
+                $children = $this->taxNodeReader($tax_tree, $sortfield, $tax_tree->readRootId());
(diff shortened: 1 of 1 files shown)


dsstrassner:
@thojou & @kergomard Is this also relevant for the Test in ILIAS 10: Add from Pool?

kergomard (changes requested):
Hi @nhaagen 

A few changes:
- Could you please rename [this function](https://github.com/ILIAS-eLearning/ILIAS/pull/8441/files#diff-1fef9ee18782e953d630b40abf539092486df75b29d0fc3c88bd6e8073998da7R204) to camelCase.
- Could we please remove the use of `ilArrayUtil`, the first assignment to the `$children`-variable, the `array()` (should be `[]`), and the double quotes [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8441/files#diff-1fef9ee18782e953d630b40abf539092486df75b29d0fc3c88bd6e8073998da7R207-R211).
- I think we can get rid of quite some nesting [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8441/files#diff-1fef9ee18782e953d630b40abf539092486df75b29d0fc3c88bd6e8073998da7R236-R254). This is unreadable from where I stand (4 nested `foreach`es with two `if`s and then topped of with an `array_map`, this is simply too much for my brain; a few `continue`s, but maybe we can also find a better way to structure this, it looks uglier every time I look at it). Please also have another look at what is happening to the `$taxonomies`-variable. It looks completely unnecessary. As does the `$listing`-variable.

Thanks again and best,
@kergomard

kergomard:
Thank you so much @nhaagen ! Now I'm smiling ;-) . Merged and picked to ILIAS 10 and trunk.

PR #8445, opened 2024-11-15
[FIX] UI: set form-level error on `Container\Container::withRequest()`.

thibsy:
Hi folks,

I stumbled upon an inconsistency when setting the errors of forms. Until now, the `Container\Container::getData()` call has set the form-level error message, whereas for inputs this already happened in `Input\InputInternal::withInput()` triggered by `Container\Container::withRequest()`. I therefore suggest to combine this logic and set the form-level error message there as well. If approved, this commit should also be picked for 8 and 9 as well.

Kind regards,
@thibsy


klees:
Picked to trunk.

PR #8446, opened 2024-11-16
36955: Different appearance of xln and https:// links

alex40724:
https://mantis.ilias.de/view.php?id=36955


mjansenDatabay:
Thx @alex40724 , LGTM! I'll pick this to all relevant branches.

PR #8447, opened 2024-11-18
RBAC: Document types of array of `ilRbacReview`

mjansenDatabay:
This PR documents the types of the `array` usages in `ilRbacReview`.

If approved, this should be merged to `trunk` as well.

### DIFF
--- a/components/ILIAS/AccessControl/classes/class.ilRbacReview.php
+++ b/components/ILIAS/AccessControl/classes/class.ilRbacReview.php
@@ -198,6 +198,7 @@ class ilRbacReview
     /**
      * Returns a list of assignable roles in a subtree of the repository
      * @todo move tree to construct. Currently this is not possible due to init sequence
+     * @return list<int>
      */
     public function getAssignableRolesInSubtree(int $ref_id): array
     {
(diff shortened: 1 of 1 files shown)


kergomard:

Generally I think this is a good idea, but do you maybe have a better insight than me, if these two changes [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8447/files#diff-0f7bab3b19bb235b2978b6050b53b7b93028a113b01ecb6da8457b18815c8ed4R670-R673) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8447/files#diff-0f7bab3b19bb235b2978b6050b53b7b93028a113b01ecb6da8457b18815c8ed4R719) are not going to have side-effects that are very subtle and hard to track? Right now I think the direct usages are all save, i.e. they use loose comparison, ...but the return values are also assigned to properties and then I stopped going further down that rabbit-hole. On the other hand, I think from a certain point on, the consumers will simply assume they are receiving `int`s, the logical choice here, but they actually won't, ...so they will most probably simply have started bug-fixing by cast everything anyway.

I'm absolutely certain we should apply these changes to trunk, ... I'm a little less sure, if we should do the same for ILIAS 10.

Thanks again and best,
@kergomard

mjansenDatabay:
Hi @kergomard ,

I reverted the `intval`-casts .

Best regards,
Michael

kergomard:
Thank you very much @mjansenDatabay ! I merge this like this into ILIAS 10 and will add the casts in trunk.

PR #8448, opened 2024-11-18
LSO: Fix Missing Type Property

kergomard:
See: https://mantis.ilias.de/view.php?id=42722

This PR removes the custom class property and moves all usages to the default property of the parent class.


klees:
Hi @kergomard,

looks extremly plausible, I think we maybe just didn't catch up with ilObject-property-naming conventions...

Thanks!

klees:
Picked to trunk.

PR #8450, opened 2024-11-18
Mantis 42805: Release 8 accordion

ingoj:
Idea for a fix for Mantis 42805


ingoj:
Details are shown within the commits

PR #8451, opened 2024-11-18
Mantis 42807: Fix error if user has no initial test but final results

ingoj:
Possible fix for mantis 42807

I used the same syntax as used in ILIAS 8 class.ilContainerObjectiveGUI.php, but I changed the replacement value for limit to 100 (%). In the (not to be expected) case the limit is not set, this uses the maximum amount (100%) as replacement, so that an objective is not accidentally fulfilled.

the same should apply to ILIAS 10 and trunk, but the files are moved (due to the component revision) in 10 and slightly changed in trunk.


smeyer-ilias (approved):
Thanks @ingoj for fixing this!

PR #8461, opened 2024-11-19
CI: ignore deleted files in pipeline checks

lukas-heinrich:
Hi everyone, 

in some PR-Checks, I noticed that the pipeline fails because the cs-fixer and copyright check tries to read files that have been deleted. Locally I could not reproduce this, because a git command is used to get the changed files locally, but in the pipeline an API call is made. The local command gets the mask `--diff-filter=ACMRT` to exclude deleted files. I have adapted the function so that the [API call](https://docs.github.com/de/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files) gets a filter to exclude deleted files too.

I look forward to feedback,
@lukas-heinrich

### DIFF
--- a/scripts/Import/Functions.sh
+++ b/scripts/Import/Functions.sh
@@ -11,7 +11,7 @@ function get_changed_files() {
     CHANGED_FILES=$(git diff-tree --name-only --diff-filter=ACMRT --no-commit-id -r ${GH_SHA} | grep -e '.php$' -e '.js$')
   else
     URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/files"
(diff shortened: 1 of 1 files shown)


fhelfer:
Hi @lukas-heinrich,

thank you very much for the PR. I tested your changes for a failing PR https://github.com/ILIAS-eLearning/ILIAS/pull/8476 
in my own repo https://github.com/fhelfer/ILIAS/pull/7 and the pipeline works for the same changes.
@mjansenDatabay Please go ahead and merge.

Best @fhelfer

mjansenDatabay (approved):

mjansenDatabay:
Seems to work in all cases! Thanks a lot @lukas-heinrich !

PR #8466, opened 2024-11-19
Notifications: Always show stored sub-form values

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=42517


iszmais:
Thank you @mjansenDatabay for your implementation.

After some second thoughts, I'm not sure if we really should sustain this information even if the main setting is deactivated.
IMO, the most clear data state should be that there are no "artifacts" of the settings left after the general deactivation.
Therefore, I proposed a PR in which the settings are removed on deactivation: https://github.com/ILIAS-eLearning/ILIAS/pull/8469
This further introduces an example value to the empty fields to ensure useful settings.

Greetings,
@iszmais

PR #8467, opened 2024-11-19
TA: 33607 factor ilTestEvaluationData

nhaagen:
https://mantis.ilias.de/view.php?id=33607

The ticket is about performance in the statistics tab, and I had quite a hard time to figure out what was actually going on.
This [diagram](https://mantis.ilias.de/file_download.php?file_id=19529&type=bug) helped a lot, though: clearly, there are multiple roundtrips to the DB on several occasions.

This PR does not eleminate all of them, but I hope it helps to make the inner workings of EvaluationData more visible;
consider it a starting point for further improvements.

First of all, the factory allows us to remove the DB from data classes as ilTestEvaluationData as well as weeding out ilObjTest a little.
Second, I tried to isolate calls to the Test and state them explicitly, while also avoiding some consecutive calls.

One of the bigger changes is this:
There was a [query retrieving the users ](https://github.com/ILIAS-eLearning/ILIAS/blob/a77278b6474a7528fa4fe217648bafb2c5c0eee5/Modules/Test/classes/class.ilObjTest.php#L2601), and then, iterating over the results, [another query to fill up passes and questions](https://github.com/ILIAS-eLearning/ILIAS/blob/a77278b6474a7528fa4fe217648bafb2c5c0eee5/Modules/Test/classes/class.ilObjTest.php#L2654).

I switched the logical order of those to combine both queries and instantiated [participants](https://github.com/nhaagen/ILIAS/blob/029ab68465fde15bcf72a5c146bde34f88491252/Modules/Test/classes/Evaluation/class.ilTestEvaluationFactory.php#L143-L146)  
 and [questions for passes](https://github.com/nhaagen/ILIAS/blob/029ab68465fde15bcf72a5c146bde34f88491252/Modules/Test/classes/Evaluation/class.ilTestEvaluationFactory.php#L187-L189) over the resulting records.

Another change is the (old)[checking for participant access](https://github.com/ILIAS-eLearning/ILIAS/blob/a77278b6474a7528fa4fe217648bafb2c5c0eee5/Modules/Test/classes/class.ilTestEvaluationData.php#L82-L89); the call stood quite out in the diagram mentioned above.
Now, the list of valid user ids is generated but once, and the entries are (new)[incorporated into the db-query](https://github.com/nhaagen/ILIAS/blob/029ab68465fde15bcf72a5c146bde34f88491252/Modules/Test/classes/Evaluation/class.ilTestEvaluationFactory.php#L123).

I did not run a real performance test, however, the outlined changes alone should improve speed quite a bit.


kergomard:
Hi @nhaagen !

Thank you very much for the PR! I had only a very quick look into the code and was wondering how we should proceed with this. ...so I decided to send you this note quickly.

Could you please have a look at the current state in ILIAS 10 please, before proceeding. I would love to improve this (and ILIAS 10 will still have these same issues), ...but I don't think we should open a completely incompatible development. Please look especially into `ILIAS\Test\Results`.

Thanks and best,
@kergomard

nhaagen:
Hi @kergomard ,
it's not that incompatible, I think: ILIAS\Test\Results\Data::retrieveResultData still relies heavily on ilObjTest::getCompleteEvaluationData / ilTestEvaluationData.
I'll adjust and propose a version for 10.

kergomard:
Thank you very much @nhaagen and sorry it took me so long until I got around to it.

I picked this manually with https://github.com/ILIAS-eLearning/ILIAS/commit/0dc7d2f479858e1d6cd81f056582aaed93230aab and fixed a problem, where the results where empty, if no questions had been answered (`il_test_results` then contains no corresponding records) with https://github.com/ILIAS-eLearning/ILIAS/commit/94bff3f6e73a716e502d24713b3fb26319aa558e.

Best,
@kergomard

PR #8468, opened 2024-11-19
Maps: 42771, remove OpenStreetMaps Label and Link from within the map itself. …

lukastocker:
…It should only be visible below the map. 

https://mantis.ilias.de/view.php?id=42771


klees:
Picked to trunk.

PR #8474, opened 2024-11-19
TA: 33607, factory for evaluation data

nhaagen:
v10 für https://github.com/ILIAS-eLearning/ILIAS/pull/8467


components/ILIAS/Test/classes/Evaluation/class.ilTestEvaluationFactory.php
```diff
@@ -0,0 +1,350 @@
...
+            $this->db->quote($this->test_obj->getTestId(), 'integer'),
+            $this->db->in('tst_active.active_id', $active_ids, false, 'integer'),
+        );
+
+        $ret = [];
+        $result = $this->db->query($query);
+        while ($row = $this->db->fetchAssoc($result)) {
+            $ret[] = $row;
+        }
+        return $ret;
```
thojou: For the current implementation, the graphs show a total of ~450,000 calls to the fetchAssoc method, as it iterates through the result set to build the array. Since we’re collecting all the data into an array anyway, we might consider switching to fetchAll to retrieve all rows at once, which could simplify the code.

Alternatively, if memory usage is a concern, we could explore converting this method into a generator by yielding each row.

components/ILIAS/Test/classes/Evaluation/class.ilTestEvaluationFactory.php
```diff
@@ -0,0 +1,350 @@
...
+                    usr_data.usr_id,
+                    usr_data.firstname,
+                    usr_data.lastname,
+                    usr_data.title,
+                    usr_data.login
+
+        FROM        tst_test_result, qpl_questions, tst_active
+
+        LEFT JOIN tst_pass_result ON tst_active.active_id = tst_pass_result.active_fi
+        LEFT JOIN usr_data ON tst_active.user_fi = usr_data.usr_id
```
thojou, line 74: Is it possible to have data in tst_active without an existing user?
Could we use `INNER JOIN` for usr_data?

thojou (commented):
Thank you, @nhaagen, for this PR! I reviewed the changes and found just a few minor tweaks that could be applied. 

Overall, the PR looks great to me. 
One thing to consider: For ILIAS 11 it might be worth discussing whether the database interactions currently in the Factory could be moved into a dedicated Repository for better separation of concerns and streamlining patterns.

Best,
@thojou

components/ILIAS/Test/classes/Evaluation/class.ilTestEvaluationFactory.php
nhaagen: Yes, deleting a user will not remove it from tst_active.

components/ILIAS/Test/classes/Evaluation/class.ilTestEvaluationFactory.php
nhaagen: This here should already have cut the calls in half :)
I switched to Generator, though, thank you for pointing that out.

components/ILIAS/Test/classes/Evaluation/class.ilTestEvaluationFactory.php
thojou: Sure! Thanks for pointing that out.

kergomard:
Hi all

Thanks for all the work @nhaagen !

I also picked this one manually and added the same changes as in https://github.com/ILIAS-eLearning/ILIAS/pull/8467 and then I picked the changes to trunk.

One comment: We do not need a Repository for this, we need to get rid of it. This should be simplified and then fully integrated into the code for Results.

Best,
@kergomard

PR #8477, opened 2024-11-20
CSS/Maps: 42840, restore missing styling for OpenLayersMap and adjust…

lukastocker:
… references to the pointer. 

The pointer images are now part of the public folder.

The missing styles were part of the node_module "ol" but as this module is not part of ILIAS anymore the styling was put into a new OpenLayers SCSS file. The content is identical to "ol.css" in ILIAS 9.

https://mantis.ilias.de/view.php?id=42840


Amstutz (commented):
Hi @lukastocker 

Two Questions:

- [x] Maybe check if the pngs could be replaced by svgs. 
- [x] SCSS: Rather extensive. Is this all necessary? Can it be shortened?

kindly
@amstutz

lukastocker:
Hi @Amstutz ,

I just checked the available SVG's. We could use icon_pin.svg (https://github.com/ILIAS-eLearning/ILIAS/blob/release_10/components/ILIAS/UI/resources/images/standard/icon_pin.svg). I would remove the "mm_20_shadow.png" reference in the code as it's only rendering a shadow to the original pin (mm_20_blue.png). 

What's your opinion about this solution?

yvseiler:
Thanks @lukastocker, yes please use the icon_pin.svg image, that should be the right one.

At the next meeting with the Iconset Task Force, I will see whether we really want to use the symbol of a golf flag here, as we did with Learning Objectives, or whether we could use a classic map pin here. If so, we would replace the icon file separately, but it would have exactly the same name and path and would therefore be updated.

Amstutz:
Thx @yvseiler 

@lukastocker I assigned this to you, feel free to reassign if the icon is replaced.

thx

lukastocker:
**Note:**
@yvseiler will talk about the icon following monday (2th of december 2024) with the Iconset Task Force as the icon_pin.svg's color is barely visible when used on a map. 

Depending on the result of the discussion I will implement the new solution.

yvseiler:
Dear @lukastocker 
I now have an icon here that can be used as a replacement for the pin on the map. Thank you for your patience. Because the original from the icon designer is not yet available, please embed this one here. If necessary, we will provide an updated version of the file later. 

Please add it to the images > standard folder (https://github.com/ILIAS-eLearning/ILIAS/tree/trunk/components/ILIAS/UI/resources/images/standard). 

icon_mapm

Thank you very much!

lukastocker:
Hi @Amstutz ,

I have implemented the new icon and shortened the .scss file as much as possible. 

Please have a look and merge if everything's okay.

Amstutz:
cool, thx a lot!

PR #8478, opened 2024-11-20
LegalDocuments: Fix modal displayed in ilObjUserGUI

lscharmer:
This PR fixes https://mantis.ilias.de/view.php?id=42746

The issue is that the KS Modal uses a form for the Modal close button which results in a form within a form which is invalid HTML (The Modal is rendered inside a `ilNonEditableValueGUI`).
This means that the Modal must be rendered outside of the form. Sadly I don't see another solution than the proposed one (which is to call into `LegalDocuments` whenever the form is rendered) but I'm open to other solutions.


kergomard (approved):
Changes in User are fine with me.

mjansenDatabay (approved):

PR #8481, opened 2024-11-20
LegacyUI/Modals: Remove legacy modals

mjansenDatabay:
This PR removes the legacy modal implementation from the ILIAS code base.


oliversamoila:
Thank you very much !!

PR #8484, opened 2024-11-20
DataCollection: Init copy field

iszmais:
https://mantis.ilias.de/view.php?id=41549


matthiaskunkel:
Jour Fixe, 25 NOV 2024: PR accepted for all maintained versions.

PR #8485, opened 2024-11-20
LSO: 42591, copy intro and extro content page if learning sequence is…

lukastocker:
… copied.

https://mantis.ilias.de/view.php?id=42591


klees:
Picked to release_9, release_10 and trunk.

PR #8434, opened 2024-11-14
42733: Restore CAS auth functionality

Rillke:
- ~Allow `phpCAS` to `unset` variables~
- do not use deprecated `phpCAS::setDebug()`
- fix call to `phpCAS::client()` by providing ILIAS base URL
- Add more debug logging

https://mantis.ilias.de/view.php?id=42733

PR #8489, opened 2024-11-21
42733: Update CAS use instructions in settings GUI

Rillke:
see https://mantis.ilias.de/view.php?id=42733
see https://github.com/ILIAS-eLearning/ILIAS/pull/8434#discussion_r1842090761

image

### DIFF
--- a/Services/CAS/classes/class.ilCASSettingsGUI.php
+++ b/Services/CAS/classes/class.ilCASSettingsGUI.php
@@ -1,20 +1,22 @@
 <?php
 
-declare(strict_types=1);
-
-/******************************************************************************
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
  *
- * This file is part of ILIAS, a powerful learning management system.
- *
- * ILIAS is licensed with the GPL-3.0, you should have received a copy
- * of said license along with the source code.
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
  *
  * If this is not the case or you just want to try ILIAS, you'll find
  * us at:
- *      https://www.ilias.de
- *      https://github.com/ILIAS-eLearning
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
  *
- *****************************************************************************/
+ *********************************************************************/
+
+declare(strict_types=1);
 
 /**
  * @author Stefan Meyer <meyer@leifos.com>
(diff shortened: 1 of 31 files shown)


mjansenDatabay:
Hi @Rillke , AFAIK we (usually) only modify the EN and DE lang files, but @matthiaskunkel may correct me here.

matthiaskunkel:
Thanks for this PR. It's great that you also adapted the other languages. In this case – I guess – it was possible to do this without knowing these languages. The advantage of this approach is that the modified language entry in English does not overwrite the existing ones in other languages when we synchronise the languages for a release. 
So, the changes in the language files are fine for me. But it still needs a review of the changes in class.ilCASSettingsGUI.php by a responsible authority.

Rillke:
Hi @mjansenDatabay @matthiaskunkel

Thanks for the review.

> So, the changes in the language files are fine for me. But it still needs a review of the changes in class.ilCASSettingsGUI.php by a responsible authority.

Thanks, good to know. ~What's your opinion on those languages where I do not feel comfortable adding the~

> and the server setting prevent_super_global_replacement in client.ini.php: %s

~part? The roman and germanic languages should be mostly fine, but I have no clue about Chinese or Arabic.~

Where to read about i18n procedure? [The Internationalization guide on docu.ilias](https://docu.ilias.de/go/pg/42472_42) appears to be empty.

Services/CAS/classes/class.ilCASSettingsGUI.php
```diff
@@ -87,7 +92,14 @@ protected function initFormSettings(): ilPropertyFormGUI
...
         $form->setTitle($this->lng->txt('auth_cas_auth'));
-        $form->setDescription($this->lng->txt("auth_cas_auth_desc"));
+
+        $dropInReplacementsUrl = 'https://github.com/ILIAS-eLearning/ILIAS/tree/release_8/src/HTTP#dropinreplacements';
+        $dropInReplacementsLink = $this->factory->link()->standard($dropInReplacementsUrl, $dropInReplacementsUrl);
+        $form->setDescription(sprintf(
+            $this->lng->txt("auth_cas_auth_desc"),
+            '7.1.0',
+            $this->renderer->render($dropInReplacementsLink)
+        ));
```
mjansenDatabay: * We should use the `snake_case`-style for variables (for added code).
* As far as I can see `7.1.0` is meant to be the min. PHP version. For ILIAS 8.x, this is 7.4 . You could use `\ilSetupAgent::PHP_MIN_VERSION`.
* Usually we always refer to the documentation in the `trunk` branch. Of course, this might be a problem if the settings described here are not valid in a later release anymore. Nevertheless, I suggest referring to the `trunk`-based document.

Rillke: @mjansenDatabay The `\ilSetupAgent::PHP_MIN_VERSION` is private. I guess we can drop the text about the requirements about PHP and cURL altogether because composer takes care of that.

mjansenDatabay (approved):

mjansenDatabay:
Thx for your contribution @Rillke !

mjansenDatabay:
Hi @Rillke ,

just FYI: I created a feature wiki page to suggest abandoning the CAS authentication for ILIAS 11.x: https://docu.ilias.de/go/wiki/wpage_8560_1357

Best regards,
Michael

PR #8490, opened 2024-11-21
User Bugfix #0042870: UserAutocomplete Limit is being ignored

fhelfer:
https://mantis.ilias.de/view.php?id=42870

`ilUserAutoComplete::$limit` is being ignored since `$cnt` isn't incremented in `getList()`
Also present in newer releases

Best, @fhelfer


kergomard:
Thank you very much @fhelfer ! This looks very plausible ;-) . Merged and picked to ILIAS 9, ILIAS 10, and trunk.

PR #8491, opened 2024-11-21
User: Fix accessing undefined key `default_hits_per_page`

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=42866


kergomard:
Thank you very much @mjansenDatabay for the PR! This look very plausible. This only needed picking to ILIAS 8.

PR #8492, opened 2024-11-21
User: Fix accessing undefined key `default_mail_incoming_mail` (and o…

mjansenDatabay:
…thers)

See: https://mantis.ilias.de/view.php?id=42867

### DIFF
--- a/Services/User/classes/class.ilObjUserFolderGUI.php
+++ b/Services/User/classes/class.ilObjUserFolderGUI.php
@@ -2596,7 +2596,7 @@ class ilObjUserFolderGUI extends ilObjectGUI
             }
         }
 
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much  @mjansenDatabay ! I picked this manually with https://github.com/ILIAS-eLearning/ILIAS/commit/5cd103851694f7761d18c4072349d1b1ae42ec91 and picked it then to ILIAS 8, ILIAS 10, and trunk. I removed the casts as we are already taking care of an array of strings, so if the value exists, it is a string. If I missed something this is clearly my oversight, not yours!

mjansenDatabay:
Hi @kergomard , removing the casts should be okay.

PR #8493, opened 2024-11-21
User: Fix self registration crashes if text fields are not required

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=42869


kergomard:
Thank you very much @mjansenDatabay for the PR! I would say this actually is an error in `ilTestInputGUI`, but for the sake of simplicity, your fix at least avoids crashes in the User. So: Merged and picked to ILIAS 10, and trunk.

mjansenDatabay:
@kergomard I agree that the actual issue is the type-handling of `value` in `ilTextInputGUI`.

PR #8494, opened 2024-11-21
T&A/10 40791: Replaces superglobal usages in TestQuestionPool component with RequestDataCollector

lukas-heinrich:
Hi everyone, 

this PR introduces a major refactoring in which @matheuszych , @bidzanaaron, @FBartz28 and I replaced all direct `$_POST` and `$_GET` accesses in TQPL by using `RequestDataCollector`. Sorry that the PR is that large, we have summarized the changes here.

### Coding Style Refactoring

At all places we came across we:

- Converted CamelCase notation of class properties to snake_case
- Added data types to method signatures
- Removed unnecessary `@return` and `@param` annotations from PHP-Docs when types are recognizable in the method signature
- Removed empty constructors that only call the parent constructor
- Removed local `$lng` variables loaded from the DIC, as they are provided as a property by the parent class in all cases

### Replacement of Superglobals

- All accesses to the $_POST array were replaced with accesses to the wrapper through `RequestDataCollector`
- For safer access to arrays, we adjusted the methods in the `RequestDataCollector` so that nested arrays can be accessed as well
- Renamed property with the name `$request` to clarify that it's not the request itself but the wrapper
- At places where it was recognizable to us through method signatures or doc comments, we directly converted the data into the expected data type. At all other places, they are processed without transformation
- At places where the type cannot be determined with certainty, we implemented the check for whether the value is set using the `empty()` function instead of a null check

### Refactoring of the `setValue` Functions

The problem here was that in these two functions, the data from the requests are processed directly. The entire logic for validation is implemented simultaneously in many places. This is a problem because not all checks were always implemented here. Examples:

- In some of these functions, it was checked whether the keys are contained in the request array, in others it was simply assumed
- In some functions, it was ensured that the points are correctly converted to `floats` (via the `str_replace` function), in other places this was missing
- In some functions, the answers were converted by type casts, in others by the Refinery, and in others not at all

Therefore, we have outsourced these conversion functions, which work similarly for most question types and GUI classes, into a separate helper class. Here, for example, the `transformPoints` method ensures that points are always converted to `float` - regardless of whether they were transmitted as a string or numeric value.

### Refactoring of the `checkInput` Functions

The same applies here as for the `setValue` function. Additionally, missing type conversions pose a big problem when strict type checks with `===` are performed, but it wasn't ensured beforehand that the data is also in the expected format. That's why we have also outsourced the validation functions.

The validation functions handle type conversion and checking simultaneously. They work in such a way that upon successful conversion without errors, they return the array with the converted values. If an error occurs, a string is returned containing the language identifier for the error message.

### Minor Bugfixes / Code Smells

- **[class.assClozeTest.php:1053](https://github.com/ILIAS-eLearning/ILIAS/blob/ca090615444fb6b71c2e5da6b71f4c4970eb36b1/components/ILIAS/TestQuestionPool/classes/class.assClozeTest.php#L1054C13-L1054C81):** Type checking doesn't make sense here as the input can be present as both a string and a number
- **[class.assClozeTestGUI.php:1677](https://github.com/ILIAS-eLearning/ILIAS/blob/ca090615444fb6b71c2e5da6b71f4c4970eb36b1/components/ILIAS/TestQuestionPool/classes/class.assClozeTestGUI.php#L1677C8-L1677C49)**: Removed self-assignment of local variables
- **[class.assKprimChoice.php `removeAnswerImage`](https://github.com/ILIAS-eLearning/ILIAS/blob/ca090615444fb6b71c2e5da6b71f4c4970eb36b1/components/ILIAS/TestQuestionPool/classes/class.assKprimChoice.php#L604)**: Method must be public
- **[class.assOrderingQuestion.php:1192](https://github.com/ILIAS-eLearning/ILIAS/blob/ca090615444fb6b71c2e5da6b71f4c4970eb36b1/components/ILIAS/TestQuestionPool/classes/class.assOrderingQuestion.php#L1192)**: Fixed undefined constants
- **[class.ilObjQuestionPoolGUI.php:512](https://github.com/ILIAS-eLearning/ILIAS/blob/ca090615444fb6b71c2e5da6b71f4c4970eb36b1/components/ILIAS/TestQuestionPool/classes/class.ilObjQuestionPoolGUI.php#L512)**: Replaced undefined variable `$ilAccess` with `$this->access`
- **[tpl.il_as_cloze_gap_builder.html](https://github.com/ILIAS-eLearning/ILIAS/blob/ca090615444fb6b71c2e5da6b71f4c4970eb36b1/components/ILIAS/TestQuestionPool/templates/default/tpl.il_as_cloze_gap_builder.html#L21)**: Replaced incorrect CSS rule separator

### DIFF
--- a/components/ILIAS/TestQuestionPool/classes/class.assClozeTest.php
+++ b/components/ILIAS/TestQuestionPool/classes/class.assClozeTest.php
@@ -1050,7 +1050,7 @@ class assClozeTest extends assQuestion implements ilObjQuestionScoringAdjustable
                 continue;
             }
 
-            if ($gap->getType() === assClozeGap::TYPE_SELECT && $value === -1) {
+            if ($gap->getType() === assClozeGap::TYPE_SELECT && $value == -1) {
                 continue;
             }
 
@@ -1071,26 +1071,20 @@ class assClozeTest extends assQuestion implements ilObjQuestionScoringAdjustable
     {
         $solution_submit = [];
 
-        foreach ($this->test_request->getPostKeys() as $key) {
-            if (preg_match("/^gap_(\d+)/", $key, $matches)) {
-                $value = $this->test_request->retrieveArrayOfStringWithFilter($key, static function ($key) {
-                    return preg_match("/^gap_(\d+)/", $key);
-                });
-
-                if ($value === null || $value === ''
-                    || !preg_match('/^gap_(\d+)/', $key, $matches)) {
-                    continue;
-                }
-                $gap = $this->getGap((int) $matches[1]);
-                if ($gap === null
-                    || $gap->getType() === assClozeGap::TYPE_SELECT && $value === -1) {
-                    continue;
-                }
-                if ($gap->getType() === assClozeGap::TYPE_NUMERIC) {
-                    $value = str_replace(',', '.', $value);
-                }
-                $solution_submit[trim($matches[1])] = $value;
+        foreach ($this->questionpool_request->getPostKeys() as $post_key) {
+            $value = $this->questionpool_request->string($post_key);
+            if ($value === '' || !preg_match('/^gap_(\d+)/', $post_key, $matches)) {
+                continue;
             }
+            $gap = $this->getGap((int) $matches[1]);
+            if ($gap === null
+                || $gap->getType() === assClozeGap::TYPE_SELECT && $value === '-1') {
+                continue;
+            }
+            if ($gap->getType() === assClozeGap::TYPE_NUMERIC) {
+                $value = $this->questionpool_request->float($post_key);
+            }
+            $solution_submit[trim($matches[1])] = $value;
         }
 
         return $solution_submit;
--- a/components/ILIAS/TestQuestionPool/classes/class.assClozeTest.php
+++ b/components/ILIAS/TestQuestionPool/classes/class.assClozeTest.php
@@ -23,7 +23,6 @@ use ILIAS\TestQuestionPool\Questions\QuestionAutosaveable;
 use ILIAS\TestQuestionPool\Questions\QuestionPartiallySaveable;
 use ILIAS\Test\Logging\AdditionalInformationGenerator;
 use ILIAS\Refinery\Random\Group as RandomGroup;
-use ILIAS\TestQuestionPool\RequestDataCollector;
 
 /**
  * Class for cloze tests
@@ -79,7 +78,6 @@ class assClozeTest extends assQuestion implements ilObjQuestionScoringAdjustable
     public ilAssQuestionFeedback $feedbackOBJ;
     protected $feedbackMode = ilAssClozeTestFeedback::FB_MODE_GAP_QUESTION;
     private RandomGroup $randomGroup;
-    private readonly RequestDataCollector $request_data_collector;
 
     public function __construct(
         string $title = "",
@@ -89,11 +87,9 @@ class assClozeTest extends assQuestion implements ilObjQuestionScoringAdjustable
         string $question = ""
     ) {
         global $DIC;
-
         parent::__construct($title, $comment, $author, $owner, $question);
         $this->setQuestion($question); // @TODO: Should this be $question?? See setter for why this is not trivial.
         $this->randomGroup = $DIC->refinery()->random();
-        $this->request_data_collector = new RequestDataCollector($DIC->http(), $DIC->refinery(), $DIC->upload());
     }
 
     /**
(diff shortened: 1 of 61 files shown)


kergomard (changes requested):
Thank you very much for all the work @lukas-heinrich,@matheuszych , @bidzanaaron, and @FBartz28 and especially for the clean-up and the many small improvements!

I went through the hole thing, but I'm sure I missed things. It doesn't matter, we should merge this as quickly as possible to avoid any issues and to allow for a thorough testing.

I have one fundamental change request. The `RequestValidationHelper` is simply and fundamentally wrong. This is not your fault, but I think it has consequences. I would kindly ask you to:
- [ ] Move it to the `classes`-folder
- [ ] Rename it to `ilTestLegacyFormsHelper`
- [ ] Deprecate it right away
- [ ] Remove it from the TestDIC
- [ ] To  make things easier: Do not inject the `Refinery`, but take it directly from the $DIC.

And then a few peanuts. I'm sure we could improve a lot more, but this already is an improvement and we should not burden it more:
- [ ] Why is [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/8494/files#diff-7770a8c709030dfc438a2a684e3477d65d5fc82715a5d9622b5ad7d7a46b8460R1053) a loose comparison with a value of the wrong type? This is always going to be a string, you make sure of it above, so it should be `$value === '-1'`.
- [ ] [These here](https://github.com/ILIAS-eLearning/ILIAS/pull/8494/files#diff-4b5753f7ecfc5dfbe31ec2ff6200018449b7cbc8c4ebd7189dc2efe092d0fd02R48) should  be strict comparisons too.
- [ ] Please do not use logical operations on values that are not `bool`s (e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8494/files#diff-45027f39f03759654228c0705d008b05b70ba76818948175c445e408a1c9d4ceR909)). I know, it is comfortable, but this hides a lot of information and makes a lot of assumptions.
- [ ] As always: Please remove the `@throws` you have added [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8494/files#diff-0c215e70eeea47756692539fe283dfb2d0cf612e518b63adcc1ec8db3de03a65R632-R634).
- [ ] Please avoid places like [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/8494/files#diff-0c215e70eeea47756692539fe283dfb2d0cf612e518b63adcc1ec8db3de03a65R670-R671) and condense them into one single assignment. Use new lines to clarify the different commands, as it makes reading easier. In general: Avoid double assignments. Sometimes you can condense them like in the case above and sometimes a function is the right solution. This for me would be the case [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8494/files#diff-e03ea2247e3cfec3e1ccdd03dc62e4a3814fa78c5edfbbcfa75ff63d1fd4d991R1396). You don't need to investigate this case further though, because I'm rather sure there is something fishy going on there and it is not your fault.
- [ ] Please revert these two changes [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8494/files#diff-5a64a2ac8ae4ca33ed87b9a68c51a744e59115dc96cdc7643b4ad60a224f5163R161) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8494/files#diff-5a64a2ac8ae4ca33ed87b9a68c51a744e59115dc96cdc7643b4ad60a224f5163R169). You may also change them to `list<int>`. I actually do not care about what we use, but everything I saw lately we are generally moving to the `array<int>`/`list<int>` notation and we shouldn't make any changes in the other direction.
- [ ] Please delete [these lines](https://github.com/ILIAS-eLearning/ILIAS/pull/8494/files#diff-c0a3ed9ea4d6c4f9ae708ffbfdae5ef1271865c7fe05f53660ed3908284000a9R153-R156). As a general rule: We never want any code that is commented out. We have a git log, if we need to go back in time.
- [ ] Could you please turn [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/8494/files#diff-901a5601543d776a88d57d20542c035711b4d70d1a84e3d5bbfa9d9699990321R50-R55) around again to avoid one level of nesting.
- [ ] Please remove all null-save-operators you added in this block [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8494/files#diff-03175ce6c21b34e70d76b8ac9dec241a77127fbd6ee363f9ec971109ca85b20eR56-R166). If there is any way we can get there and not have an object, we have had a real issue previously and things should go boom. Theoretically we could instantly return if it was null, but even if `getInstanceByObjId()` can return null, this is nothing that should ever happen here as far as I can see.
- [ ] Please check the errors of the Copyright-Checker below.

I would leave it here.

@kergomard

lukas-heinrich:
Hi @kergomard ,

Thank you very much for the detailed review and feedback. I have made the changes, tested them again and pushed them here.

Best,
@lukas-heinrich

kergomard:
Thank you very much once more @lukas-heinrich ! I merged to avoid to get conflicts. Picked to trunk as well.

PR #8497, opened 2024-11-21
UI/Mainbar: 42574, keep context and focus slate entries

nhaagen:
https://mantis.ilias.de/view.php?id=42574

### DIFF
--- a/components/ILIAS/UI/resources/js/MainControls/dist/mainbar.js
+++ b/components/ILIAS/UI/resources/js/MainControls/dist/mainbar.js
@@ -1,3 +1,17 @@
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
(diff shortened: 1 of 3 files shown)


klees (approved):

klees:
Hi @thibsy and @Amstutz,

this replaces some functionality (from jquery-ui?) that we implicitly have relied upon. Looks straight forward to me...

Kind regards!

thibsy (changes requested):
Hi @nnhaagen,

Thx a lot for your contribution to the UI framework.

Before we can merge this, please implement the following changes:

- [x] Docu: this code looks very specific to jQuery-UI. Could you please provide a comment that links to the [`:focusable` selector](https://api.jqueryui.com/focusable-selector/), or check if this complex logic is really necessary for our use-case. Maybe the `tabindex` is sufficient?
- [x] `tagName`: please use `instanceof` checks instead of this attribute and hard-coded strings. You can e.g. use `element.ownerDocument.defaultView.HTMLButtonElement`.

Kind regards,
@thibsy

thibsy (approved):
Hi @nhaagen, sorry for the delay here. This mostly LGTM now, so I will go ahead and merge this. For future reference, please try to avoid using globals like this and access them via the element.

thibsy:
@nhaagen, actually you need to rebase first. But I'll merge right after :)

PR #8499, opened 2024-11-21
Repository: re-add paramter for loc redirect (42783)

schmitz-ilias:
This PR fixes [42783](https://mantis.ilias.de/view.php?id=42783) by re-adding a parameter on redirect that got lost because of ilCtrl refactorings. Course and Test use Repository as a go-between, so the issue has to be fixed there. An alternative would be to set the relevant links differently in `ilObjTestListGUI::modifyTitleLink`.

EDIT: This should also fix [43045](https://mantis.ilias.de/view.php?id=43045).


schmitz-ilias:
The issue was fixed (in a more elegant way) with 79daa3fb191f5ccda40f696c1051e1273880a7b1 so this PR is no longer needed.

PR #8501, opened 2024-11-22
CRS/PRG: 42860, set function to public to avoid errors while trying to de…

lukastocker:
…lete a learning objective.

https://mantis.ilias.de/view.php?id=42860


klees:
Picked to trunk.

PR #8505, opened 2024-11-25
IASS: 42882, adjust check of RBAC for users who only have read and wr…

lukastocker:
…ite permissions.

https://mantis.ilias.de/view.php?id=42882


klees:
Picked to trunk.

PR #8508, opened 2024-11-25
T&A fix 42872: CharSelectorConfig::setDefinition TypeError

fneumann:
See https://mantis.ilias.de/view.php?id=42872

No merge to release_9 required since CharSelector is removed there.


kergomard:
Ok, this looks right and I simply merge it.

PR #8511, opened 2024-11-26
Certificate Bugfix #0042876: Fix Saving Certificate with <br> is no valid XHTML

fhelfer:
https://mantis.ilias.de/view.php?id=42876

This Fixes the error message for <br> Tags being invalid XHTML
And also the removal of empty lines as well as the <p> tags adding unpleasing empty lines


mjansenDatabay:
Thx @fhelfer

PR #8512, opened 2024-11-26
UI Form: 42530, remove width 100% from form-control to resize …

lukastocker:
…input fields according to their content.

https://mantis.ilias.de/view.php?id=42530


Amstutz:
thx a lot!

PR #8513, opened 2024-11-26
Services/Component: Fix error message when downgrading plugins (#42931)

corro:
See [#42931](https://mantis.ilias.de/view.php?id=42931)


klees:
Thanks @corro!

klees:
Picked to 9, 10 and trunk.

PR #8514, opened 2024-11-26
Init webui-popover.min.js

tfamula:
After https://github.com/ILIAS-eLearning/ILIAS/pull/8377 has been merged, the node module must be initialized, so the Popovers can work properly.


klees:
Thanks @tfamula!

PR #8517, opened 2024-11-26
UI: 42600, set cursor to pointer for content in tree.

lukastocker:
https://mantis.ilias.de/view.php?id=42600


Amstutz:
thx a lot!

PR #8519, opened 2024-11-26
Certificate Bugfix #0042878: Fix Certificate Deleting Image does not save with message nothing changed

fhelfer:
https://mantis.ilias.de/view.php?id=42878


mjansenDatabay:
Thx a lot @fhelfer !

PR #8520, opened 2024-11-26
TA: 37918, show autosave content

nhaagen:
https://mantis.ilias.de/view.php?id=37918
This adds autosaved answers to results and scoring.

### DIFF
--- a/components/ILIAS/Test/src/Results/Data/Factory.php
+++ b/components/ILIAS/Test/src/Results/Data/Factory.php
@@ -249,20 +249,21 @@ class Factory
                 $show_inline_feedback
             );
 
-            if ($test_obj->getAutosave() &&
-                $type === 'assTextQuestion'
-            ) {
-                $usr_solution .= $question_gui->getAutoSavedSolutionOutput(
+            $autosave_output = null;
+            $show_autosave_title = false;
+            if ($test_obj->getAutosave()) {
+                $autosave_output = $question_gui->getAutoSavedSolutionOutput(
                     $active_id,
                     $attempt_id,
-                    false,
-                    false,
-                    false,
-                    false,
-                    false,
-                    false,
-                    false,
-                    true
+                    $graphical_output,
+                    $result_output,
+                    $show_question_only,
+                    $show_feedback,
+                    $show_correct_solution,
+                    $show_manual_scoring,
+                    $show_question_text,
+                    $show_autosave_title,
+                    $show_inline_feedback
                 );
             }
 
@@ -306,7 +307,8 @@ class Factory
                 $workedthrough,
                 $answered,
                 $requested_hints,
-                $recapitulation
+                $recapitulation,
+                $autosave_output
             );
         }
 
--- a/components/ILIAS/Test/src/Results/Data/QuestionResult.php
+++ b/components/ILIAS/Test/src/Results/Data/QuestionResult.php
@@ -38,7 +38,8 @@ class QuestionResult
         private readonly bool $workedthrough,
         private readonly bool $answered,
         private readonly int $requested_hints,
-        private readonly ?string $content_for_recapitulation
+        private readonly ?string $content_for_recapitulation,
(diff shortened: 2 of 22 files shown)


kergomard (changes requested):

Special thanks for splitting up functions and the additional clean-up!

I think we can run with this, I've just a few peanuts to change, no need to backport any of this to the PR for ILIAS 9 (I will look at it separately), as this is only a little bit of hygiene:

- [x] Please do not check strings for truth, they don't have any ;-). See [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8520/files#diff-8b0db21826e3df03970c6a420a87f2ba6f7885ddbb504f7cab26cd97354d5d37R426) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8520/files#diff-906cacbf12c01a6882c25ff5f819c4232deceef83170b39f0b720ec4220a48ddR376). Personally I would probably go for the explicit solution `!== null && !== ''`, but `empty()` would already make things more explicit.

- [x] Please remove [this superfluous lines](https://github.com/ILIAS-eLearning/ILIAS/pull/8520/files#diff-906cacbf12c01a6882c25ff5f819c4232deceef83170b39f0b720ec4220a48ddR367) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8520/files#diff-4b5753f7ecfc5dfbe31ec2ff6200018449b7cbc8c4ebd7189dc2efe092d0fd02R861) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8520/files#diff-3819f2ac4a55bafe36b885ed52fd7c7868d6cd8fbd9dbf22b44f10dfda34e450R801) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8520/files#diff-3819f2ac4a55bafe36b885ed52fd7c7868d6cd8fbd9dbf22b44f10dfda34e450R955-R956). 

- [x] Please remove the int-casts [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8520/files#diff-5dbb581327300fd409ad3d66337d0b5eb7bc2c2e6051e6f009f2f64a2f99baceR239) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8520/files#diff-5dbb581327300fd409ad3d66337d0b5eb7bc2c2e6051e6f009f2f64a2f99baceR275). We have typed properties here, they already have to be `int`s (I know, this line comes from previously).

- [x] Could you please change the double quotes to single quotes and the long array notation to short one in [these lines here](https://github.com/ILIAS-eLearning/ILIAS/pull/8520/files#diff-c2f4ca76fb60a3bc6fc0fd5b19b30f595397ef8720f126ea1d2482e696879dceR1437-R1454)? I would have loved to have a simpler solution for this, but I think you are right, this is good enough for now.

- [x] Please change [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/8520/files#diff-3819f2ac4a55bafe36b885ed52fd7c7868d6cd8fbd9dbf22b44f10dfda34e450R816) to short array notation.

- [x] Thank you very much for adding the `$pass` parameter [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8520/files#diff-45027f39f03759654228c0705d008b05b70ba76818948175c445e408a1c9d4ceR829), but could you please change the order `$active_id, $pass, $user_post_solution`.

As said: All peanuts. Thank you very much for improving the code with this fix!

Best,
@kergomard

nhaagen:
Hi @kergomard ,
thank you for looking through this.
I hope it helps in the direction of separating rendering and data-handling of questions, too, as I'd really like to have that sometime :)

kergomard:
Thank you very much for also providing a PR for ILIAS 10 @nhaagen and sorry for having you wait for so long.

To avoid another roundtrip I picked this manually with https://github.com/ILIAS-eLearning/ILIAS/commit/ebfd14afedc0fc8cc048239e23ad29fc6abe672e .

Best,
@kergomard

PR #8521, opened 2024-11-26
TA: 37918, show autosave content

nhaagen:
https://mantis.ilias.de/view.php?id=37918
This adds autosaved answers to results and scoring.
for 10: https://github.com/ILIAS-eLearning/ILIAS/pull/8520

### DIFF
--- a/Modules/Test/classes/Results/class.ilQuestionResult.php
+++ b/Modules/Test/classes/Results/class.ilQuestionResult.php
@@ -40,7 +40,8 @@ class ilQuestionResult
         protected bool $workedthrough,
         protected bool $answered,
         protected int $requested_hints,
-        protected ?string $content_for_recapitulation
+        protected ?string $content_for_recapitulation,
+        protected ?string $autosaved_answer
     ) {
     }
 
@@ -110,4 +111,9 @@ class ilQuestionResult
     {
         return $this->requested_hints;
     }
+    public function getAutosavedAnswer(): ?string
+    {
+        return $this->autosaved_answer;
+    }
+
 }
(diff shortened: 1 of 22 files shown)


kergomard (changes requested):
Thank you very much for this PR @nhaagen !

As said: No need for further clean-up here.

One question though:
- I'm a little scared of typing `getSolutions()`, ...specially if we backport this to ILIAS 8. How sure are you that this doesn't go sideways because the Test-Object is (partially) strictly typed? Right now, I would prefer not to do it, but maybe you have a better insight then me.
- [This here](https://github.com/ILIAS-eLearning/ILIAS/pull/8521/files#diff-498f6047f032462823f878b0200920c74b32188507aed6136fbd6d26ae59a855R343-R346) also peaked my interest. It looks like something that should have gone wrong for a long time and this just looks wrong. Do you know under which circumstances this happens? Could we fix this more thoroughly?

Thanks again and best,
@kergomard

dsstrassner:
I would not backport this to ILIAS 8 and only see this as an intermediate solution for ILIAS 9 & 10. I hope that the "test-end"-workshop results present us a way out of this saving problematic at the end of tests.

oliversamoila:
[JourFixe], 09 DEC 2024: We follow the suggestion of Nils and Stephan and accept the mentioned changes for ILIAS 9, ILIAS 10 and Trunk. 
The authorities are requested to make a corresponding note in Changed Behavior.
Many Thanks

dsstrassner:
I removed the JourFixe Label, because this is decided. Will you merge this @kergomard?

kergomard:
Hi @nhaagen and @dsstrassner 

Sorry, for having left this for so long. I went once more through the whole code and that looks so I merged it  as is.

I've one small doubt though:
For me in the Show Detailed Results View the autosave is hard to differentiate. I would like it, if there was a clearer distinction.

grafik

Maybe we find a clearer solution for this in one of the next iterations.

Thanks again and best,
@kergomard

PR #8527, opened 2024-11-27
[Bugfix 42976]Typo in Lang Var for TA

dsstrassner:
Should fix https://mantis.ilias.de/view.php?id=42976


kergomard:
Thank you very much @dsstrassner ! Merged and picked to trunk.

PR #8528, opened 2024-11-27
[Bugfix]Byline 42975

dsstrassner:
adding Participants to the byline instead of Dashboard

https://mantis.ilias.de/view.php?id=42975


kergomard:
Thank you very much for the PR @dsstrassner ! Merged and picked to trunk.

PR #8532, opened 2024-11-28
LegacyForm: Fix Error on Big File

kergomard:
See: https://mantis.ilias.de/view.php?id=42862

There is a problem in all ILIAS-Versions using the file-upload in Legacy-Forms: If a file is uploaded that is bigger than the `post_max_size` in php, the `$_POST` array is going to be empty. As the command is in the array, the system has no clue where to go and where you land simply depends on the default command and on a big of luck for it not to go completely sideways.

This fix is hacky at best, but it avoids the biggest issue as it moves the check for the filesize to the moment of uploading the file (as it is in the new forms). There surely are far better solutions, but as this is in an unmaintained component, this is the best I can propose with my limited time.

This should be ported to all accepted versions, if accepted.

Best,
@kergomard

### DIFF
--- a/components/ILIAS/Form/classes/class.ilFileInputGUI.php
+++ b/components/ILIAS/Form/classes/class.ilFileInputGUI.php
@@ -338,6 +338,8 @@ class ilFileInputGUI extends ilSubEnabledFormPropertyGUI implements ilToolbarIte
             );
         }
 
+        $f_tpl->setVariable('MAX_SIZE_WARNING', $this->lng->txt('form_msg_file_size_exceeds'));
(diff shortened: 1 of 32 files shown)


nicoroeser:
[JourFixe], 09 DEC 2024:
We appreciate this change and accept it for all supported versions. We ask Kergomard, Stephan @kergomard to check if the `$DIC['ui.upload_limit_resolver']`.

kergomard:
Ok,  I decided to not pick this back to ILIAS 8 as there are quite a few more changes needed and the support of ILIAS 8 will finish soon, so this might introduce errors that cannot be fixed afterwards. Thus: Merged and picked to ILIAS 9 and trunk.

I've changed the retrieval of the max-size to the corresponding resolver.

PR #8536, opened 2024-11-28
Use linkify-element in ilExtLink

lscharmer:
This PR implements the change from the dependency `linkify-jquery` to `linkify-element` in ilExtLink.js, which was added in the PR #8169.

### DIFF
--- a/components/ILIAS/Chatroom/classes/gui/class.ilChatroomHistoryGUI.php
+++ b/components/ILIAS/Chatroom/classes/gui/class.ilChatroomHistoryGUI.php
@@ -213,7 +213,8 @@ class ilChatroomHistoryGUI extends ilChatroomGUIHandler
         $roomTpl->setVariable('PERIOD_FORM', $durationForm->getHTML());
 
         if ($room && $messages !== []) {
-            $this->mainTpl->addJavaScript('assets/js/socket.io.js');
+            ilLinkifyUtil::initLinkify($this->mainTpl);
+            $this->mainTpl->addJavaScript('assets/js/socket.io.min.js');
             $this->mainTpl->addJavaScript('assets/js/Chatroom.min.js');
             $roomTpl->setVariable('CHAT', (new ilChatroomViewGUI($this->gui))->readOnlyChatWindow($room, array_column($messages, 'message'))->get());
         } else {
--- a/components/ILIAS/Chatroom/classes/gui/class.ilChatroomViewGUI.php
+++ b/components/ILIAS/Chatroom/classes/gui/class.ilChatroomViewGUI.php
@@ -83,6 +83,7 @@ class ilChatroomViewGUI extends ilChatroomGUIHandler
      */
     private function setupTemplate(): void
     {
+        ilLinkifyUtil::initLinkify($this->mainTpl);
         $this->mainTpl->addJavaScript('assets/js/socket.io.min.js');
         $this->mainTpl->addJavaScript('assets/js/Chatroom.min.js');
         $this->mainTpl->addJavaScript('assets/js/AdvancedSelectionList.js');
--- a/components/ILIAS/Link/resources/ilExtLink.js
+++ b/components/ILIAS/Link/resources/ilExtLink.js
@@ -1,3 +1,19 @@
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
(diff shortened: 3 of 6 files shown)


components/ILIAS/Chatroom/classes/gui/class.ilChatroomHistoryGUI.php
```diff
@@ -213,7 +213,10 @@ private function showMessages(
         $roomTpl->setVariable('PERIOD_FORM', $durationForm->getHTML());
 
         if ($room && $messages !== []) {
-            $this->mainTpl->addJavaScript('assets/js/socket.io.js');
+            $this->mainTpl->addJavaScript('assets/js/linkify.min.js');
+            $this->mainTpl->addJavaScript('assets/js/linkify-element.min.js');
+            $this->mainTpl->addJavaScript('assets/js/ilExtLink.js');
```
mjansenDatabay: Hi @lscharmer ,

please remove these calls and instead try to address adding the JS files in `\ilLinkifyUtil::initLinkify` (`\ilLinkifyUtil::getLocalJsPaths must be changed)`.

Best regards,
Michael

components/ILIAS/Chatroom/classes/gui/class.ilChatroomViewGUI.php
```diff
@@ -83,6 +83,9 @@ public function joinWithCustomName(): void
      */
     private function setupTemplate(): void
     {
+        $this->mainTpl->addJavaScript('assets/js/linkify.min.js');
+        $this->mainTpl->addJavaScript('assets/js/linkify-element.min.js');
+        $this->mainTpl->addJavaScript('assets/js/ilExtLink.js');
```
mjansenDatabay: The same applies here.

mjansenDatabay (changes requested):

components/ILIAS/OnScreenChat/classes/class.ilOnScreenChatGUI.php
```diff
@@ -309,15 +309,17 @@ public static function initializeFrontend(ilGlobalTemplateInterface $page): void
             iljQueryUtil::initjQueryUI($page);
             ilLinkifyUtil::initLinkify($page);
 
+            $page->addJavaScript('assets/js/linkify.min.js');
+            $page->addJavaScript('assets/js/linkify-element.min.js');
```
mjansenDatabay: And here :-)

components/ILIAS/Link/resources/ilExtLink.js
```diff
@@ -8,11 +24,14 @@ il.ExtLink = {
...
 	 */
 	autolink: function (selector, link_class) {
-		$(selector).linkify({
+		const options = {
 			validate: {
 				url: (val) => /^https?:\/\//.test(val), // only allow URLs that begin with a protocol
 				email: false // don't linkify emails
 			}
+		};
+		$(selector).each(function () {
```
mjansenDatabay, line 33: One last question: Could we easily get rid of the jQuery calls in this file? If yes, it would be nice if we could use vanilla JS.

lscharmer: I kept the jQuery call to be fully backwards compatible. As far as I can see `il.ExtLink` is used in the `Chatroom`, `OnScreenChat` and `COPage`. Which need support for a single `HTMLNodeElement` and for a `selector`. If the jQuery call should be replaced I would rather support these 2 cases but nothing more. If this should be fully backwards compatible, this would need support for `HTMLNodeElement`, `selector`, `jQuery Node`, `Arrays`, etc.
Do you think we should drop the unused cases?

components/ILIAS/Chatroom/classes/gui/class.ilChatroomHistoryGUI.php
lscharmer: Thx I didn't new this existed.

components/ILIAS/Link/resources/ilExtLink.js
mjansenDatabay: @alex40724 Any opinion here, cause you are one of the consumers. If you don't have the time to answer in a timely manner, it's okay. We will then leave it as itr is for now.

I personally would like to drop the jQuery support and only support the use cases of the current consumers.

mjansenDatabay (approved):

PR #8538, opened 2024-11-29
OrgU: #42977, move calls from global to object-specific settings

nhaagen:
Settings for single repository objecs used to be in ilOrgUnitGlobalSettings; they rather belong to ilOrgUnitObjectPositionSetting.

Then, initialization of the settings is not necessary, they work in a tristate allowing to fall back on global defaults if no explicit value was set.

Please see https://mantis.ilias.de/view.php?id=42977#c110034

### DIFF
--- a/components/ILIAS/OrgUnit/classes/Settings/class.ilOrgUnitObjectPositionSetting.php
+++ b/components/ILIAS/OrgUnit/classes/Settings/class.ilOrgUnitObjectPositionSetting.php
@@ -26,7 +26,7 @@ class ilOrgUnitObjectPositionSetting
 {
     protected ilDBInterface $db;
     private ?bool $active = null;
-    private array $settings_cache = [];
+    private ?ilOrgUnitObjectTypePositionSetting $settings_cache = null;
 
     public function __construct(
         private int $obj_id
@@ -51,12 +51,6 @@ class ilOrgUnitObjectPositionSetting
         }
 
         if ($this->active === null) { //implicit: isGloballyEnabled && isChangeable
-            //the object was not properly initialized with position settings;
-            //init?
-            //$this->initFromGlobalSettings();
-            //throw?
-            //throw new \LogicException('no position settings for obj '.$this->obj_id);
-            //silently ignore?
             return (bool) $this->getGlobalSettingsForThisObjectType()->getActivationDefault();
         }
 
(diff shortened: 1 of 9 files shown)


components/ILIAS/OrgUnit/classes/Settings/class.ilOrgUnitObjectPositionSetting.php
```diff
@@ -16,93 +17,109 @@
...
         $this->db = $GLOBALS['DIC']->database();
-        $this->obj_id = $a_obj_id;
         $this->readSettings();
     }
 
-    /**
-     * Lookup activation status
-     */
-    public function lookupActive(int $a_obj_id): bool
+    public static function getFor(int $obj_id): self
```
klees, line 39: Please add static caching for this method. We will have a bunch of lookups into orgu settings per object, so this shall prove beneficial for performance.

components/ILIAS/OrgUnit/classes/Settings/class.ilOrgUnitObjectPositionSetting.php
```diff
@@ -16,93 +17,109 @@
...
+
+        return $this->active;
     }
 
     /**
-     * Check if position access is active. This returns true or false if it is object specific or null if the object has no setting.
+     * convenience wrapper for specific objects to check on global settings
      */
-    public function isActive(): ?bool
+    public function isGloballyEnabled(): bool
```
klees: This naming throw me off when reading `isActive`. I thought this was about whether the org units are activated at all globally... So please rename to `isActiveForOwnType`.

klees (changes requested):
Hi @nhaagen,

LGTM in general, I have left two change requests inline.

Kind regards!

klees:
Hi @alex40724, hi @kergomard,

could you please have a quick look here? Changes are basically editorial, shouldn't be much effort.

Thanks!

kergomard (approved):
Thank you very much for the changes @nhaagen and @klees . I'm fine with them for RBAC and Object.

Best,
@kergomard

kergomard:
[JourFixe], 09 DEC 2024 : We have discussed the pros and cons and would like to follow the process of creating a new object.
This means that the global definition of the setting is also taken into account when copying and importing an object (in this case: Test) with positions - the global settings override individual settings on the source instance of the object.

klees:
@alex40724 Could you please have a look here?

klees:
Hi @alex40724,

would you be so kind a have a look here?

Kind regards!

klees:
Hi @alex40724,

could you look into this?

Kind regards!

klees:
Hi @alex40724,

would you be so kind and look into this?

Kind regards!

dsstrassner:
Currently there is no authority for OrgUnits.

matthiaskunkel:
We have an authority for OrgUnits again. I have removed the Shepherd tag and assigned the PR to @maalers . Thanks to Marcel for having a look at it. Feedback from @alex40724 is still pending, if I am not wrong.

PR #8555, opened 2024-11-29
Fix functionality and accessibility for expand/collapse behaviour in UI Filter

tfamula:
Hi UI Coordinators!

in ILIAS 10, the expand/collapse behaviour in Filters does not work anymore due to the absence of Bootstrap (see [42827)](https://mantis.ilias.de/view.php?id=42827).
I decided to not adopt the old Bootstrap behaviour, but to refactor the expand/collapse behaviour, based on https://github.com/ILIAS-eLearning/ILIAS/pull/7332. At the same time, this is also a big improvement regarding accessibility.
Please have a look. If accepted, this must be picked to trunk, too. One can consider if this could also be picked to release 9 and 8 (because of the A11y improvements), but I leave that up to you.

@tfamula 

PS: The unit tests are failing because I have not adjusted the rendering tests yet. I would do that as a last step.

### DIFF
--- a/components/ILIAS/UI/resources/js/Input/Container/dist/filter.js
+++ b/components/ILIAS/UI/resources/js/Input/Container/dist/filter.js
@@ -1,3 +1,18 @@
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ */
+
 var filter = function($) {
 
   //Init the Filter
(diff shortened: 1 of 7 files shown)


thibsy (commented):
Hi @tfamula,

Thx for your contribution to the UI framework and getting rid of some bootstrap remnants!

About the concrete changes, please answer:

- [x] `onAjaxCmd`: can you explain why you remove this method from the filter's client-side public interface? Are there no usages anymore?

Kind regards,
@thibsy

tfamula:
Hi @thibsy,
correct! This has been used formerly for the Bulky Buttons ([expand button](https://github.com/ILIAS-eLearning/ILIAS/pull/8555/files#diff-b587a5279f811b4bca3f4632fd807154551130cc98813ce024d201e44ff44132L105) and [collapse button](https://github.com/ILIAS-eLearning/ILIAS/pull/8555/files#diff-b587a5279f811b4bca3f4632fd807154551130cc98813ce024d201e44ff44132L116)). Now, the Bulky Buttons are not needed anymore, because I implemented the `<button>` element directly into the template and visibility of elements is handled internally now in the JS file.
The Ajax requests are still needed to save the current expand/collapse state. That's why there is still the method [`performAjaxCmd`](https://github.com/ILIAS-eLearning/ILIAS/pull/8555/files#diff-ebb3bc21a4bb713ec55d79cdb24b6f839b821711beaf9b9a8b0ebffebe947152R348). It works quite similar to the old method, but I renamed it to make clear it is no part of the filter's client-side public interface anymore.

thibsy:
@tfamula thx for the explanation. This LGTM then. Please go ahead and amend the unit tests, so we can merge this :).

tfamula:
@thibsy Ready to merge!

PR #8558, opened 2024-11-29
Certificate: skip empty ul elements when translating to fo (43008)

schmitz-ilias:
This PR is a possible fix for [43008](https://mantis.ilias.de/view.php?id=43008). It makes it so that empty `ul`-elements in the certificate text are ignored when saving the certificate configuration. This of course does not affect already generated certificates in `il_cert_user_cert`, those stay broken.

Please triple check this change, I'm very close to completely clueless about the whole xhtml to fo mechanism.


mjansenDatabay:
Thx for your contribution, @schmitz-ilias !

PR #8560, opened 2024-11-29
10/UI/view controls/42965

nhaagen:
https://mantis.ilias.de/view.php?id=42965

buddy_system.js clears dropdowns; if the document-event is registered with "useCapture", VCs cannot stop propagation.


klees:
Seems to be resolved in the meantime. Thanks @nhaagen.

PR #8568, opened 2024-11-29
Contact: Remove dependencies on jQuery

mjansenDatabay:
This PR removes all jQuery dependencies in the "Contact" component.

This can be considered as a first refactoring step. It does not re-structure the code compltetely or split up the (relevant/basic) file into several ES6 modules, but it already contains some improvements beside the jQuery-removal.

### DIFF
--- a/components/ILIAS/Contact/BuddySystem/templates/default/tpl.buddy_system_relation_table_listener.html
+++ b/components/ILIAS/Contact/BuddySystem/templates/default/tpl.buddy_system_relation_table_listener.html
@@ -1,34 +1,56 @@
-<script type="text/javascript">
-document.addEventListener("DOMContentLoaded", function() {
-    (function ($, $scope) {
-        var table = $("#{TABLE_ID}");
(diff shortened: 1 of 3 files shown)


mjansenDatabay:
Hi @kergomard ,

maybe you could have a quick look at the change here `components/ILIAS/User/classes/Gallery/class.ilUsersGalleryGUI.php`: https://github.com/ILIAS-eLearning/ILIAS/pull/8568/files#diff-0353e6dd3573ca5a672ae4bb6550cb5d8bcc7d2beab5616c5d039fcab3c83b32

Thanks and best regards,
Michael

mjansenDatabay:
Hi @kergomard ,

another question: Would you have any objections to approving this for 10.x as well?

In my opinion, the risk here is not very high. I just want to get rid of jQuery.

Best regards
Michael

kergomard (approved):
Thank you very much @mjansenDatabay for this PR and sorry I lost it from view!

One more place we can get rid of jQuery! Yay!

PR #8570, opened 2024-11-29
NPM: Add `moment`

lscharmer:
This PR adds moment as NPM dependency.

General Information:
- [X] License: MIT

Usages:

- Calendar: 
  - components/ILIAS/Calendar/Calendar.php
  - components/ILIAS/UI/src/Implementation/Component/Button/Renderer.php
- Legacy Forms: Implicitly by ilCalendarUtil in the ilDateTimeInputGUI field type
- OnScreenChat: ./components/ILIAS/OnScreenChat/js/moment.js
  It is mainly (but not only) used to display relative dates and especially times (locale aware) in the conversation windows. These date/time presentations are dynamically updated (without expensive server round-trips) every 60 seconds.

Wrapped by:

- OnScreenChat: components/ILIAS/OnScreenChat/resources/moment.js

Reasoning:

`moment` is a JavaScript library to parse, validate, manipulate and display dates and times in JavaScript.
    It has been initially introduced by the Calendar component and is still used there. I guess it has been added as dependency because at that time JavaScript did not support any locale aware date/time formatting. As a project we could agree upon a migration to the native JavaScript Intl.DateTimeFormat and Intl.RelativeTimeFormat packages for client-side date.

Maintenance:

`moment` is in maintenance mode, the last release was in December 2023. This could be a potential risk for new releases of Node.js in the future or potential security issues. If we could not rely on this library anymore, we will have to check if there are good alternatives.

Links:

- NPM: https://www.npmjs.com/package/moment
- GitHub: https://github.com/moment/moment
- Documentation: https://momentjs.com/docs/

### DIFF
--- a/package_new.json
+++ b/package_new.json
@@ -8,6 +8,7 @@
     "test": "tests"
   },
   "dependencies": {
+    "moment": "^2.29.4"
   },
   "devDependencies": {
   },


schmitz-ilias:
Hi @lscharmer, I don't think moment is used in ilCalenderUtil or ilDateTimeInputGUI anymore. If I'm not missing anything, it was made redundant by the recent switch to html-native datetime inputs.

mjansenDatabay:
@schmitz-ilias See `components/ILIAS/Calendar/Calendar.php`

@lscharmer It seems `components/ILIAS/UI/src/Implementation/Component/Button/Renderer.php` also includes the file.

kergomard:
We reject this Dependency for ILIAS 11 as there seems to be only one usage in the chat and the library is not well maintained.

PR #8571, opened 2024-11-29
Notificaion: Remove jQueryUI selector from group rendering

iszmais:
https://mantis.ilias.de/view.php?id=42938


klees:
Hi @iszmais,

this won't survive the next compilation of that target file you have modified there.

Hence, I close this.

Kind regards!

PR #8582, opened 2024-12-02
Export: instead of loading all schema files, load only schema files that are used, when they are used

chlulei:
Link: https://mantis.ilias.de/view.php?id=42536


smeyer-ilias (approved):
Thanks @chlulei !

PR #8583, opened 2024-12-02
10/UI/Input load section styling correctly 42739

catenglaender:
https://mantis.ilias.de/view.php?id=42739

# Issue

Sections titles are positioned in column to the left of their content.

image

# Change

The section scss file was never imported (maybe as a result of an incomplete merge at some point). With "ui-component_section" loaded by _ui-component_input, the section title is shown on top of the section and not beside it.

image


Amstutz:
Nice! Sometimes, the little things make all the difference...

PR #8584, opened 2024-12-02
User Bugfix #0043016: Fix Soap User prefs check rejecting imported values

fhelfer:
https://mantis.ilias.de/view.php?id=43016

If approved, this must be picked to all maintained branches.


kergomard:
Thank you very much for the PR @fhelfer ! Merged and picked to all supported versions.

PR #8588, opened 2024-12-02
[FEATURE] UI: use `<template>` tags for `Field\HasDynamicInputs` inputs.

thibsy:
Hi all,

I refactored the client-side rendering process of `Field\HasDynamicInputs` inputs by using a [`<template>` element](https://html.spec.whatwg.org/#the-template-element). This reduces the complexity of our server-side rendering, because we do not need to bother about duplicate IDs anymore, since templates isolate their content inside a [`DocumentFragment`](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment). We also don't need to count our inputs anymore, because we can simply traverse the DOM using the `[id]` selector. One tradeoff we have here is, that attributes which reference an existing ID need special care. E.g. a labels `for` attribute needs to be updated explicitly.

Let me know what you think. I will amend the broken pipelines once this passed your review.

Kind regards,
@thibsy


thibsy:
@klees I internalised the `createId()` function (inside `TemplateRenderer`) and amended the unit tests.

PR #8594, opened 2024-12-02
XAPI: Fix Creation of New Object

kergomard:
As you are using legacy forms you  cannot rely on the object to save your data. I hope this is correct, but I couldn't test it, so please test before merging. Theoretically this looks right, but I've no idea how to enable the option to create xAPI-objects locally.

See: https://mantis.ilias.de/view.php?id=42577

Best,
@kergomard


kergomard:
Apparently the corresponding issue has been fixed, so this PR is not needed anymore and can be closed.

PR #8596, opened 2024-12-02
9/UI/MainControls metabar keep open fix 42143

catenglaender:
https://mantis.ilias.de/view.php?id=42143

# Issue

When clicking anywhere inside a metabar slate that's not a form element that can take focus, the slate will close. Clicking some links and buttons becomes impossible in Safari and GNOME Web (Epiphany). This is not the behavior in 8 or 10.

Expected behavior: Clicking inside an open metabar slate should never close it. Clicking next to it or tabbing out of it should close it.

metabar-closing-unexpectedly

# Changes

Weirdly, 9 shares identical code with 10, where the behavior is as expected. Something else in 9 fires a focusout event when clicking anywhere with nextFocusTarget (event.relatedTarget) = null that is not fired in 10.

A "legitimate" keyboard focus out does not have relatedTarget = null, but a specific div given. I used this to filter if the event was the misfire that causes the unexpected closing.

metabar-closing-correctly

# Bigger Picture

Maybe someone with more time and expertise could find exactly what causes the focusout in 9? Whatever it is, it doesn't seem to happen the same way in 10.


catenglaender:
# Update

Turns out the problem exists in 10 as well, but didn't show up until recent changes were made.

What's happening is this:
- Slate opens, focusInEngagedSlates() sets focus on first focusable input element inside
- any click inside the slate but on no clickable element removes the focus, this is counted as a focusout event with relatedTarget = null
- focusout event with no relatedTarget that's a DOM element inside the slate triggers onCLickDisengageAll()

Therefor, discarding focusout events with relatedTarget = null seems to be a more decent approach than I initially suspected.

Amstutz:
tht x lot.

PR #8616, opened 2024-12-04
IASS: 42941, check if irss resource id is null before running 'remove'.

lukastocker:
https://mantis.ilias.de/view.php?id=42941


klees:
Picked to trunk.

PR #8624, opened 2024-12-04
Fix TestExport using own file system instead of irss

fhelfer:
Fixes https://mantis.ilias.de/view.php?id=42988

This was a hard one, essentially the new Export-Infrastructure uses IRSS to store export files. The test component has some files of its own to add to the export. It does so by storing them locally in data/client_id/qpl_data/qpl_obj_id/export_zip/...
Files in this directory are then added to the IRSS storage. But the table still shows the incomplete local Zip which doesn't contain all the files (e.g. TailDependencies). IRSS Zip does contain everything, though.

@kergomard Since I fundamentally changed this class, can you see any side effects that i don't know of?
Code seems to work for all Table Actions.

Best @fhelfer


kergomard:
Thank you very much @fhelfer 
Ok, this looks like we could get rid of most of the class, I think there are a few open questions for me:
- What you  write sounds to me as if we are creating two zip-files of which at least one never gets deleted. Is this right? If it is: We have to make sure to delete all leftovers.
- I would kindly ask you to test this well with all potential dependencies: Taxonomies, Competences, Images (in Answers, in Texts). The most problematic will be the images in Texts as they cannot be added anymore and as they are broken in a gazillion different ways anyway, so: If you do not find a good way to test that, just state it quickly and we will ignore them for the time being.

These are the side-effects I can see right now, as this class seems to only do the management of the files and not much more.

As you are on it: Could you check the behavior in the test? If we have the same issue there it might be useful to find a common solution.

Thanks again and best,
@kergomard

dsstrassner:
Hi @fhelfer maybe you could check this as first mission of 2025, if you can't do this week. 😁

fhelfer:
Hey @kergomard,

I am sorry that this PR has not been addressed by me for so long. 
I made sure leftover zips get deleted.
This is only the case for normal xml export. Archive and xmlres still do their own thing, since they are not really using the correct way through ExportHandler.
But that is a general issue from my in-depth Analysis.
I tested it with XML Export, everything is exported as expected. For the Results Export, these dependencies are missing since they are only written to the IRSS ZIP.

Not many side effects expected. In the Table, there are now files from different sources listed. This is not a problem since they handle themselves differently.

Please have another look. There are still many errors left, so export/import might not be working properly due to some issues that are not part of this PR.

Best @fhelfer

dsstrassner:
@kergomard you are on it?

kergomard:
Hi @dsstrassner and @fhelfer 
Yes, I'm on it. I was actually a little bit hesitant because of [this](https://github.com/ILIAS-eLearning/ILIAS/pull/8841).

Thank you very much for the PR @fhelfer ! I would have the following two little requests:
- Could you please try, if the changes in the above PR would also work for the Test (remove the special casing and the xml export)? For me this looks like it should be possible, if it actually is: Could you incorporate the changes from the PR above in this PR and amend it correspondingly, please.
- The second thing is a little bit more mundane: I don't see a possible universe where [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/8624/files#diff-377cf2c64a9111fe133f461ef77079e17cd9132640dc33f5c5e0ef927c46de64R1661) makes sense, as the only thing this does is lead to another call to `ilObjTest::getActiveIdOfUser()` in `ilObjTest::loadQuestions()`. I would suggest to remove [this call to `lObjTest::getActiveIdOfUser()`](https://github.com/ILIAS-eLearning/ILIAS/pull/8624/files#diff-377cf2c64a9111fe133f461ef77079e17cd9132640dc33f5c5e0ef927c46de64R1660) and just always pass `null` to it. Please also change the signature of `lObjTest::getActiveIdOfUser()` and change the check for `$active_id` to `empty($active_id)`.

I would then merge this.

Best,
@kergomard

kergomard:
Thank you very much @fhelfer and sorry I didn't grasp the whole picture there. I did a quick check and the exports do at least not fail with these changes. Merged and picked to trunk.

PR #8626, opened 2024-12-04
11/UI/MainControls metabar keep open fix 42143

catenglaender:
Clicks inside a metabar slate close it. This makes clicks on some buttons and links impossible. Expected behavior is that they stay open.

Cherry-pick of this ILIAS 9 fix https://github.com/ILIAS-eLearning/ILIAS/pull/8596 for 11


klees:
Hi @catenglaender,

please look for a solution that doesn't need jquery.

Thanks!

catenglaender:
Hi @klees,

I removed jquery from the very small scope of changed lines.

Since I was already there I couldn't resist to try to remove a few more lines of jquery around it... which I had to give up and completely revert though because the slate heavily interacts with the metabar and expects jquery objects everywhere. So does the event management.

However, I was surprised to find that modern ES6 JavaScript has all the features that we are widely using jquery for already built in. Contains, children etc. nowadays are just as easy one-function-operations as they are in jquery. Once we have a non-jquery event manager, I suspect that moving away from jquery might be less daunting than one might think.

klees:
@catenglaender Rebase please!

thibsy (approved):
Thx @catenglaender. I assume this is the exact same fix which has already been merged for ILIAS 9, so I approve this and merge directly after your rebase.

catenglaender:
Rebase is done :heavy_check_mark:

PR #8629, opened 2024-12-04
Mail: Hide BCC in internal messages (ILIAS >= 9)

mjansenDatabay:
This PR suggests to ignore the BCC recipients when storing the database records (for "Internal Mails") for all (TO/CC/BCC) recipients.
The sender will (of course) still see the BCC recipients in the user interface.

See: https://mantis.ilias.de/view.php?id=43029#c110260

If approved, this must be picked to all mtainined branches.


mjansenDatabay:
I reverted this change, hiding the CC recipients is not a good idea :-). The issue was only given in ILIAS <= 8.x.

PR #8631, opened 2024-12-05
ContentStyle: Fix missing `max` key in schema version for exports

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=43090

If approved, this must be be picked to `trunk` as well (and maybe to 9.x, but I did not check this).


schmitz-ilias:
Sorry, I missed your PR.... if it's any consolation, in the meantime @chlulei has fixed the issue in basically the same way.
Thanks anyways!

PR #8632, opened 2024-12-05
Export: Don't expect `max` key, export provides might not have define…

mjansenDatabay:
…d it

See: https://mantis.ilias.de/view.php?id=43090

If approved, this must be be picked to `trunk` as well (and maybe to 9.x, but I did not check this).


schmitz-ilias:
Same as #8631, sorry again!

PR #8633, opened 2024-12-05
Footer: Copy permalink to clipboard (ILIAS 10)

lscharmer:
Sister PR to: #7498

As the Footer has been reworked from 10 upwards, the permanemt link does not have a specific template block for itself anymore.

So this PR adds a new template file: `components/ILIAS/UI/src/templates/default/MainControls/tpl.permanent-link.html` for the copy permanent link HTML part.

Alternatively I could add a new permanent link specific block in the existent template `components/ILIAS/UI/src/templates/default/MainControls/tpl.footer.html` if that would be preferred. I'm also open to other solutions.


Amstutz:
thx for the PR. Sry, this slipped from my radar.

PR #8644, opened 2024-12-06
OrgU: 42998, ctrl-structure for tabs when called from ilAdministratioGUI

nhaagen:
https://mantis.ilias.de/view.php?id=42998


klees:
Picked to trunk.

PR #8653, opened 2024-12-06
Chatroom: Replace dependency async with Promises

lscharmer:
This PR replaces the dependency `async` with nodes builtin Promises.


mjansenDatabay (approved):
Thx Lukas.

PR #8664, opened 2024-12-09
Add `rollup` as npm Dependency for ILIAS 11

thibsy:
Assessment:

Rollup is a JavaScript module bundler, which is used to compile small pieces of code into one big file that should be used or delivered to the client.

General Information:
- Name of the dependency: `rollup`
- Version: `4.28.1`
- [x] this dependency was already used in ILIAS.

Type of dependency:
- [ ] composer
- [x] npm

Usage:

* `components/ILIAS/UI/resources/js/*` (many usages)

Reasoning:

* We need a module bundler to create bundled JavaScript assets which are shipped to the client.
* Rollup has a plugin eco-system, which provides some plugins we need to properly create our bundles.

Maintenance:

* The package is actively maintained.
* The package has an active commit history.
* The package is used by 15Mil.

Links:
* Packagist/NPM: https://www.npmjs.com/package/rollup
* GitHub: https://github.com/rollup/rollup
* Documentation: https://rollupjs.org/

Alternatives:

There are several alternatives. However, for the time being, it would be too much of an effort to exchange the module bundler since we maintain many small bundles instead of one big one, so many config files would need to be rewritten. Besides, the differences between module bundlers are very small.
* `webpack`
* `parcel`
* `vite`

### DIFF
--- a/package_new.json
+++ b/package_new.json
@@ -10,6 +10,7 @@
   "dependencies": {
   },
   "devDependencies": {
+    "rollup": "^4.28.1",
   },
   "scripts": {
     "test": "mocha --recursive"


kergomard:

lscharmer:
Merged with #9086

PR #8670, opened 2024-12-09
Add `@rollup/plugin-commonjs` as npm Dependency for ILIAS 11

thibsy:
Assessment:

This package converts CommonJS, or JavaScript code prior to ES6, into ES6 compatible code at runtime.

General Information:
- Name of the dependency: `@rollup/plugin-commonjs`
- Version: `28.0.1`
- [ ] this dependency was already used in ILIAS.

Type of dependency:
- [ ] composer
- [x] npm

Usage:

* `components/ILIAS/UI/resources/js/Input/Field/file.js` (to be refactored)

Reasoning:

* Rollup.js, the module bundler we are using, demands that the code which should be bundled is compatible with ES6. However, since we are using third party libraries in our code, it sometimes happens that a library does not provide a ES6 compatible format. This is when we need to transform the code before we can bundle it. This plugin is usually only used together with the @rollup/plugin-node-resolve package.

Maintenance:

* The package is part of Rollup.js' [plugin repository](https://github.com/rollup/plugins), which is actively maintained and receives yearly major releases and ~monthly to ~weekly minor/patch updates.

Links:
- Packagist/NPM: https://www.npmjs.com/package/@rollup/plugin-commonjs
- GitHub: https://github.com/rollup/plugins/tree/master/packages/commonjs/#readme
- Documentation: https://rollupjs.org/introduction/#importing-commonjs

Alternatives:

There are no alternatives. This package is specific to our JavaScript module bundler.

### DIFF
--- a/package_new.json
+++ b/package_new.json
@@ -10,6 +10,7 @@
   "dependencies": {
   },
   "devDependencies": {
+    "@rollup/plugin-commonjs": "^28.0.1",
   },
   "scripts": {
     "test": "mocha --recursive"


kergomard:
Dependency is rejected for ILIAS 11, as this would open the door to include code that doesn't conform to our coding standards. If it should be absolutely necessary to use this, don't hesitate to propose it again.

PR #8671, opened 2024-12-09
T&A: Fix Question Hints.

thojou:
This PR addresses multiple issues described in the following mantis tickets:

* https://mantis.ilias.de/view.php?id=42843
* https://mantis.ilias.de/view.php?id=42855
* https://mantis.ilias.de/view.php?id=43058
* https://mantis.ilias.de/view.php?id=43060

This may be cherry picked into trunk.

Best,
@thojou

### DIFF
--- a/components/ILIAS/TestQuestionPool/classes/class.ilAssQuestionHintGUI.php
+++ b/components/ILIAS/TestQuestionPool/classes/class.ilAssQuestionHintGUI.php
@@ -136,8 +136,8 @@ class ilAssQuestionHintGUI extends ilAssQuestionHintAbstractGUI
             }
 
             if ($hintJustCreated && $this->question_obj->isAdditionalContentEditingModePageObject()) {
-                $this->ctrl->setParameterByClass(self::class, 'hint_id', $questionHint->getId());
-                $this->ctrl->redirectByClass(self::class, 'edit');
+                $this->ctrl->setParameterByClass(ilAssHintPageGUI::class, 'hint_id', $questionHint->getId());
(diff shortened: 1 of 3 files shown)


kergomard:
Thank you very much @thojou for the PR!

This looks right, I have just one request:
- Do we really need these two saved parameters [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8671/files#diff-f7c925d424f8ebaa2f9cfcee6d467689e2519152a25fcd1a0be705decd4a6dedR199-R200)? They usually cause issues because they show up in places we don't really want them to show up. I would be very thankful if we could solve this by setting the parameter when we really need it.

Thanks and best,
@kergomard

thojou:
Hey @kergomard,  

Thanks for the feedback!  

I also considered allocating the parameter only when needed. However, this approach would require handling the parameter across multiple locations within the *Hint*GUI classes. While debugging, I discovered that the issue does not occur in the TestPlayer.  

Based on this observation, I opted to save the parameter within the `ilObjQuestionPoolGUI` class, as it is required inside this class for further processing. Additionally, this approach avoids interfering with test-related code and eliminates the need to check for the question's ID existence before forwarding it to the relevant actions.  

I'm happy to hear your thoughts on that. If you still prefer your approach, I'm happy to adjust the code as requested.

Best,  
@thojou

kergomard:
Hi @thojou 

Thank you very much for the explanation!

Prefer in an abstract way: Yes, make it a condition: No. Merged and picked to trunk.

Best,
@kergomard

PR #8683, opened 2024-12-09
[FEATURE] UI: align `Table\Factory` parameter order

thibsy:
Hi all,

We have noticed in #7143 that the order of factory parameters for `Table\Factory::data()` and `Table\Factory::ordering()` are not aligned with the order of input fields, such as for the new (proposed) `Field\TreeMultiSelect` and `Field\TreeSelect` inputs.

This PR therefore proposes to change the order of parameters, so the `Table\*Retrieval` instance comes as first argument, not the last. This also amends all usages inside ILIAS.

Kind regards,
@thibsy

### DIFF
--- a/components/ILIAS/AccessControl/src/Log/Table.php
+++ b/components/ILIAS/AccessControl/src/Log/Table.php
@@ -93,6 +93,7 @@ class Table implements DataRetrieval
         $cf = $this->ui_factory->table()->column();
 
         return $this->ui_factory->table()->data(
+            $this,
             $this->lng->txt('rbac_log'),
             [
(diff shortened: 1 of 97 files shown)


thibsy:
I think the copyright checker might be broken. The copyright license headers look valid to me. Can you confirm?

matthiaskunkel:
Jour Fixe, 03 FEB 2025: We highly appreciate this suggestion and accept the PR for trunk.

thibsy:
I have rebased onto latest trunk and performed another search-and-replace. The failing checks have nothing to do with the changes in this PR. Since this is time-sensitive, I'll merge this right away now.

PR #8684, opened 2024-12-09
Tracking/Forum: use KS form in lp settings

schmitz-ilias:
This PR moves the Learning Progress settings form in objects to KS, as part of [this feature](https://docu.ilias.de/go/wiki/wpage_7382_1357).

Tracking offers an interface that allows other components to add to the form. That interface also had to be changed accordingly, along with its usage in consumers. Currently, the only such consumer is the Forum.

Cheers, @schmitz-ilias

### DIFF
--- a/components/ILIAS/Forum/classes/class.ilForumLP.php
+++ b/components/ILIAS/Forum/classes/class.ilForumLP.php
@@ -18,48 +18,58 @@
 
 declare(strict_types=1);
 
+use ILIAS\UI\Component\Input\Input;
+
 /**
  * Class ilForumLP
  * @author Michael Jansen <mjansen@databay.de>
  */
 class ilForumLP extends ilObjectLP
 {
-    public function appendModeConfiguration(int $mode, ilRadioOption $modeElement): void
+    /**
+     * @return Input[]
+     */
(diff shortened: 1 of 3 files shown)


mjansenDatabay (approved):

mjansenDatabay:
Thx @schmitz-ilias , LGTM!

components/ILIAS/Forum/classes/class.ilForumLP.php
```diff
@@ -18,48 +18,58 @@
...
+
 /**
  * Class ilForumLP
  * @author Michael Jansen <mjansen@databay.de>
  */
 class ilForumLP extends ilObjectLP
 {
-    public function appendModeConfiguration(int $mode, ilRadioOption $modeElement): void
+    /**
+     * @return Input[]
```
mjansenDatabay, line 30: @schmitz-ilias I will change the return type here, cause it is not a list, but `array<string, Input>`.

schmitz-ilias: ~~Yeah that is much nicer, I'll do the same in Tracking. Thanks!~~

Just saw the PR, you were already two steps ahead ^^ Thanks!

PR #8690, opened 2024-12-10
T&A Fix redirect issue when accessing tests from favorites.

thojou:
This PR is related to https://mantis.ilias.de/view.php?id=42858.

Contrary to the issue description, the bug is already present in ILIAS 9.
A separate PR will address the issue for ILIAS 10 and Trunk.

Best regards,
@thojou


kergomard (approved):
Ok, this is painful ;-)

PR #8691, opened 2024-12-10
Fix #43138: Cannot delete Shibboleth role assignment rules

fneumann:
https://mantis.ilias.de/view.php?id=43138

Thx to hoyer for the hin on mantis.


chfsx:
Thanks @fneumann , I picked the commit to release_9, release_10 and trunk

PR #8697, opened 2024-12-11
PRG: 38422, have ObjectListGUI create CommandLink

nhaagen:
https://mantis.ilias.de/view.php?id=38422


klees:
Picked to trunk, release_8 and release_9.

PR #8699, opened 2024-12-11
UI: 43025 add padding to .c-form.c-form-horizontal if parent is a acc…

lukastocker:
…ordion content class.

https://mantis.ilias.de/view.php?id=43025


Amstutz:
thx.

Amstutz:
One Question:

- [x] Use Variables: Isn't 15px one of this standard padding sizes, that we have variables for? 

if so, please use them

Amstutz:
thx a lot @lukastocker !

PR #8701, opened 2024-12-11
IASS: 42933, remove duplicated colon from IASS Members GUI.

lukastocker:
https://mantis.ilias.de/view.php?id=42933


klees:
Picked to trunk.

PR #8703, opened 2024-12-11
CSS: Small Addition to Avoid Dragging Subelements

kergomard:
Hi @Amstutz 

This is a very small improvement for draggable elements in the test and assessment. This way images in them do not get separated from their parent elements (that should be dragged).

I added two small line-brake-fixes. I thought the change was small enough, so that they wouldn't bother. I hope this is ok.

If accepted, I would kindly ask you to cherry-picked to trunk.

Thank you very much,
@kergomard


Amstutz:
thx a lot!

PR #8708, opened 2024-12-11
UI: 43155, fix tooltip issue for toggle buttons.

lukastocker:
https://mantis.ilias.de/view.php?id=43155

**NOTE:** 
The extra <div> container around label and button is corrupting the handling of core.Tooltip.js.. 
The JS adds another class to the parent div of the element (here: button). Therefore the new class was added to <div class="il-toggle-item"> instead of <div class="c-tooltip__container">. See following:

```
Regular and working:
<div class="c-tooltip__container">
	<button>...</button>
	<div class="c-tooltip c-tooltip--hidden"></div>
</div>


(Old) case and not working:
<div class="c-tooltip__container">
	<div class="il-toggle-item">
		<label>...</label>
		<button>...</button>
	<div>
	<div class="c-tooltip c-tooltip--hidden"></div>
</div>```
 ```

**UPDATE:**
In [this comment](https://github.com/ILIAS-eLearning/ILIAS/pull/8708#issuecomment-2879515852) below I explained why the changes mentioned above were reverted.

### DIFF
--- a/components/ILIAS/UI/resources/js/Core/dist/core.js
+++ b/components/ILIAS/UI/resources/js/Core/dist/core.js
@@ -70,7 +70,7 @@
      ********************************************************************
      */
 
-     /**
+    /**
      * This represents one tooltip on the page.
      */
     class Tooltip {
@@ -110,7 +110,7 @@
       #main = null;
 
       constructor(element) {
-        this.#container = element.parentElement;
+        this.#container = element.closest('.c-tooltip__container');
         this.#element = element;
         this.#document = element.ownerDocument;
         this.#window = this.#document.defaultView || this.#document.parentWindow;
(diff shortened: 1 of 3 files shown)


Amstutz:
Hi @lukastocker 

Hi, took a brief look at this, without passing this through my colleagues. What immediately comes to mind:

- [x] JS Fix: is there no JS fix possible here. We like to avoid having tpl files without a single parent. 

Thx a for a quick feedback. If there are good reason for not fixing the JS here, we will look into this as is.

Thx.

lukastocker:
Hi @Amstutz ,

I just reviewed the first solution in this PR and reverted all changes to add a more simple change to the JS file instead of changing the template file. With the new changes we still have a parent div in the template file.

I did test this with multiple tooltip buttons on one page - they each are showing the correct tooltip.
The tooltip for a e.g. standard button also still works with this implementation.

thibsy (approved):
LGTM, thx!

PR #8709, opened 2024-12-11
Init/LoginPage: Don't show login item in "MetaBar" on "index" action

mjansenDatabay:
This PR suggest to hide the "Login" item in the "MetaBar" in case the "Index" action ("/") is requested.

If approved, this PR must be picked to all maintained branches.

### DIFF
--- a/Services/Init/classes/Provider/StartUpMetaBarProvider.php
+++ b/Services/Init/classes/Provider/StartUpMetaBarProvider.php
@@ -128,7 +128,7 @@ class StartUpMetaBarProvider extends AbstractStaticMetaBarProvider
 
     private function isUserOnLoginPage(UriInterface $uri): bool
     {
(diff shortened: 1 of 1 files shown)


Services/Init/classes/Provider/StartUpMetaBarProvider.php
```diff
@@ -128,7 +128,7 @@ private function isUserLoggedIn(): bool
 
     private function isUserOnLoginPage(UriInterface $uri): bool
     {
-        return preg_match('%^.*/login.php$%', $uri->getPath()) === 1;
+        return preg_match('%^.*(/login.php|/)$%', $uri->getPath()) === 1;
```
thojou:
```suggestion
        return preg_match('%^.*(?:/login\.php|/)$%', $uri->getPath()) === 1;
```

Just a few cents:

* Add escaping for `.`
* Use a non-capture-group

thojou (approved):

PR #8721, opened 2024-12-13
T&A 43131: Adds fix for vanishing formula question result and wrong reached points calculation

matheuszych:
When trying to submit a formula question result, either by going to the next question or exiting the test, the answer given by the   user is removed from the input field and not persisted into the database.

The vanishing issue seems to be related to a if condition being negated. Reached points not being properly calculated, looks like a result of changing regex strings from double to single quotes. Double quotes need often extra escaping which is not needed in single qotes causing the issue.

We decided to streamline all the regex in the file to use single quotes, to prevent the issue from potentially happening in the future.

[Mantis: 43131](https://mantis.ilias.de/view.php?id=43131)


kergomard:
Thank you very much @matheuszych ! This looks good. Merged and picked to trunk.

PR #8722, opened 2024-12-13
T&A 43151: Fixes not being able to choose no answer in multiple choice question check

matheuszych:
An exception was being thrown when trying to check an multiple choice question where no answer was selected. This would put null into a variable that later in code must be an array.

A possible fix is to check if the variable is null and if so set it to an empty array.

[Mantis: 43151](https://mantis.ilias.de/view.php?id=43151)


kergomard:

This looks right. Merged and picked to trunk.

PR #8723, opened 2024-12-13
UI: Set Autofocus on Filter in Drilldown

kergomard:
Hi all

This PR changes the behavior according to https://mantis.ilias.de/view.php?id=42093

There are drawbacks though:
- This will might lead mobile device to display the on-screen keyboard when the drilldown becomes visible.
- This will yank screenreaders to this position, when the drillown becomes visible, when it is inside a dialog-tag or when the input is visible from the start (ie: doesn't happen in mainbar, does happen in object-creation-modal).

Alternatively I could do this in the object-creation-modal through js, but I think it is a good idea to have a broader discussion before I do that.

Best,
@kergomard 

@yvseiler :  I added you as a reviewer as maybe you (or somebody else of the UI/UX/A11x-experts has a better grip on this and a more founded opinion. Thanks!

### DIFF
--- a/components/ILIAS/UI/src/templates/default/Menu/tpl.drilldown.html
+++ b/components/ILIAS/UI/src/templates/default/Menu/tpl.drilldown.html
@@ -4,7 +4,7 @@
         <div></div>
         <div class="c-drilldown__filter">
             <label for='{ID_FILTER}' class="control-label">{LABEL}</label>
-            <input id='{ID_FILTER}' type="text" name='{NAME}' class="form-control" />
+            <input id='{ID_FILTER}' type="text" name='{NAME}' class="form-control" autofocus />
         </div>
         <div class="c-drilldown__backnav">
             {BACKNAV}
--- a/components/ILIAS/UI/tests/Client/Menu/Drilldown/drilldown.test.js
+++ b/components/ILIAS/UI/tests/Client/Menu/Drilldown/drilldown.test.js
@@ -20,7 +20,7 @@ const parsedHtml = `<section class="c-drilldown" id="id_2">
         <div></div>
         <div class="c-drilldown__filter">
             <label for="id_3" class="control-label">filter_nodes_in</label>
-            <input id="id_3" type="text" name="" class="form-control">
+            <input id="id_3" type="text" name="" class="form-control" autofocus >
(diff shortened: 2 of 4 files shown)


yvseiler:
Thank you for your comprehensible list of the concerns and consequences of this adaptation.

I fully understand the need expressed in the issue. For expert users, it is an advantage that the focus is directly in the filter field.

However, the mentioned drawbacks give me a bit of a stomach ache:
- Small screens: I tried it on iPhone and there the modal is also zoomed in when the focus is in the filter input field, so that you have to scroll horizontally to see the entire modal content. This makes the content in the modal more difficult to use.
- The focus for screen readers should be on the modal title, not directly in the filter field, which would be more helpful for people with screen readers to understand where they are.

I am therefore in favor of not implementing this adaptation at the moment, but keeping it up our sleeve for a next ILIAS version. 

Reason: Since we are introducing a changed behavior here with ILIAS 10, we should stick with the current implementation in a first iteration, where an additional tab causes the cursor to land in the filter field and does not penalize the behavior on small screens. It makes more sense to collect structured feedback on this new way of adding objects in order to then reconsider with the next ILIAS version whether it makes more sense to introduce the focus directly in the filter field.

thibsy:
Hi @kergomard,

I have some concerns about the [accessibility implications](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus#accessibility_concerns) of the `autofocus` attribute too. While it might make sense for the object creation dialog, it is arguable that it would also make sense for the main menu, or (coming soon) the tree select input. It would be more suitable, if the surrounding components (context) handle this manually.

I therefore close this and suggest to pursue your alternative approach: introduce this functionality on the client-side interface of the drilldown, so other components can set the focus programatically. By implementing this inside the drilldown, we can also react to possible future states, where the filter may may not be available for some reason.

Kind regards,
@Amstutz, @klees and @thibsy

kergomard:
Thank you very much for the feedback @yvseiler , @thibsy , @Amstutz , and @klees . I will then for the time being simply close this issue and keep the behavior as it is right now. We might then, as @yvseiler suggests, iterate over the solution!

Happy holidays!
@kergomard

PR #8727, opened 2024-12-14
WebDAV: Fix Uploading Files

kergomard:
See: https://mantis.ilias.de/view.php?id=41677


kergomard:
Hi @chfsx : I close this PR as it cannot be merged anyway as ILIAS 8 is security maintenance only.

PR #8729, opened 2024-12-15
Certificate: Set Default Bday Format il_CAL_DATE

kergomard:
See: https://mantis.ilias.de/view.php?id=43240

We could also always pass the full list of parameters to the class, but as there is the option...

Best,
@kergomard


mjansenDatabay (approved):

mjansenDatabay:
Thx @kergomard

PR #8732, opened 2024-12-16
T&A 43220: fix Excel Export by adding file extension

lukas-heinrich:
Hi everyone,

This PR is related to https://mantis.ilias.de/view.php?id=43220 and addresses an issue when exporting as Excel file without specifying a file extension.

The issue is caused by the fact that the ilExcel class revises the file name without returning the new path before writing to an output. When exporting a file without an extension, it will be appended automatically. Therefore, the path in the export classes is no longer correct and operations like copying the file will fail.

Best,
@lukas-heinrich

### DIFF
--- a/components/ILIAS/Test/src/ExportImport/ResultsExportExcel.php
+++ b/components/ILIAS/Test/src/ExportImport/ResultsExportExcel.php
@@ -44,12 +44,16 @@ class ResultsExportExcel implements Exporter
         private readonly \ilLanguage $lng,
         private readonly \ilObjUser $current_user,
         private readonly \ilObjTest $test_obj,
-        private readonly string $filename = '',
+        private string $filename = '',
         private readonly bool $scoredonly = true,
     ) {
         $this->user_date_format = $this->current_user->getDateTimeFormat();
         $this->aggregated_data = $test_obj->getAggregatedResultsData();
         $this->worksheet = new \ilExcel();
+
+        if (!str_contains($this->filename, '.xlsx')) {
+            $this->filename .= '.xlsx';
+        }
     }
 
     public function withFilterByActiveId(int $active_id): self
--- a/components/ILIAS/Test/src/ExportImport/ResultsExportExcel.php
+++ b/components/ILIAS/Test/src/ExportImport/ResultsExportExcel.php
@@ -44,16 +44,12 @@ class ResultsExportExcel implements Exporter
         private readonly \ilLanguage $lng,
         private readonly \ilObjUser $current_user,
         private readonly \ilObjTest $test_obj,
-        private string $filename = '',
+        private readonly string $filename = '',
         private readonly bool $scoredonly = true,
     ) {
         $this->user_date_format = $this->current_user->getDateTimeFormat();
         $this->aggregated_data = $test_obj->getAggregatedResultsData();
         $this->worksheet = new \ilExcel();
-
-        if (!str_contains($this->filename, '.xlsx')) {
-            $this->filename .= '.xlsx';
-        }
     }
 
     public function withFilterByActiveId(int $active_id): self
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much for the PR @lukas-heinrich 

As far as I can see though, this cannot be right. The E-Mail is sent [here](https://github.com/ILIAS-eLearning/ILIAS/blob/55fdd7f83f1f1b95f1363c3d7550fd6af9e3382f/components/ILIAS/Test/classes/class.ilObjTest.php#L6478-L6504), if I'm not mistaken, and there we do already have an extension. I think the solution should be to remove the extension [here](https://github.com/ILIAS-eLearning/ILIAS/blob/55fdd7f83f1f1b95f1363c3d7550fd6af9e3382f/components/ILIAS/Test/classes/class.ilObjTest.php#L6492) and then add it [here](https://github.com/ILIAS-eLearning/ILIAS/blob/55fdd7f83f1f1b95f1363c3d7550fd6af9e3382f/components/ILIAS/Test/classes/class.ilObjTest.php#L6495).

Thank you very much and best,
@kergomard

lukas-heinrich:
Hi @kergomard ,

thank you very much for the feedback and for reviewing my suggestion. I may have been mistaken in my approach, but let me try to explain what I discovered during debugging. I believe I didn't describe this very well in the introductory text.

- The process begins in `\ILIAS\Test\ExportImport\Factory::getExporter`, where we create a new instance of the export class `ResultsExportExcel` and provide a filename without an extension
- With the call to the `write` method [here](https://github.com/ILIAS-eLearning/ILIAS/blob/55fdd7f83f1f1b95f1363c3d7550fd6af9e3382f/components/ILIAS/Test/classes/class.ilObjTest.php#L6490C13-L6490C23), the method `\ilExcel::writeToFile` is called. This in turn uses the method `\ilExcel::prepareStorage` (see [here](https://github.com/ILIAS-eLearning/ILIAS/blob/55fdd7f83f1f1b95f1363c3d7550fd6af9e3382f/components/ILIAS/Excel/classes/class.ilExcel.php#L410)) to modify the file path again before the file is written to disk. Remember, the filename had no extension, so one is added here automatically.
- However, we are not aware of this modifying, as the method does not return the modified path. The `$path` variable in `\ilObjTest::sendAdvancedNotification` still contains the file path without the extension.

In my opinion, this fact is the problem that causes the copying in `\ilFileDataMail::copyAttachmentFile` to fail. I was able to verify this by temporarily removing the suppression operator `@` here: https://github.com/ILIAS-eLearning/ILIAS/blob/55fdd7f83f1f1b95f1363c3d7550fd6af9e3382f/components/ILIAS/Mail/classes/class.ilFileDataMail.php#L284

I hope this helps to better understand what I observed. I'm very open to corrections, suggestions for changes, or questions :)

Thank you and best regards,
@lukas-heinrich

kergomard:
Hi @lukas-heinrich 

You are completely right about the problem and I really had missed the point about copying, but I do still not agree with the solution ;-).

I would like to keep the filename without any extension. Please move your appending of the filename to `ResultsExportExcel::write()` and then also explicitly set the format of the Export to `ilExcel::FORMAT_XML` with `$this->worksheet->setFormat()`. This way we ensure that 'xlsx' really is the right extension and we do it for exactly this singular case where it actually matters.

Thanks again and best,
@kergomard

lukas-heinrich:
Hi @kergomard,

Thank you very much for the feedback. I totally agree that it's better to do it at the point where the file is actually written. So I made the changes and pulled it again.

Best
@lukas-heinrich

thojou (approved):

kergomard (approved):
Now me is a very happy camper!

Thank you very much @lukas-heinrich ! Merged and picked to trunk.

PR #8733, opened 2024-12-16
T&A 43168/43170: fix: 'Argument 2 ($points) must be of type float, string given'

lukas-heinrich:
Hi everyone,

this PR is related to [43168](https://mantis.ilias.de/view.php?id=43168) and [43170](https://mantis.ilias.de/view.php?id=43170), which are both caused by a missing type cast. 

I use the Refinery at this point, because the conversion from separator `',' => '.'` also has to happen at this point, and I think using the Refinery is a better approach than using a `str_replace`. 

I look forward to Feedback.
Best,
@lukas-heinrich


kergomard (approved):

Merged and picked to trunk.

PR #8736, opened 2024-12-16
Cron: Support KS/UI forms

mjansenDatabay:
With this commit cron job implementations
are allowed to migrate from legacy forms
to the UI/KS forms.

An instance of `ilCronJob` can provide a
boolean `false` when overriding `usesLegacyForms`
and then simply provide own implementations of
`getCustomConfigurationSections` and
`saveCustomConfiguration`.

### DIFF
--- a/components/ILIAS/Cron/classes/class.ilCronJob.php
+++ b/components/ILIAS/Cron/classes/class.ilCronJob.php
@@ -248,10 +248,45 @@ abstract class ilCronJob
         return false;
     }
 
+    /**
+     * @deprecated
+     */
+    #[\Deprecated('Will be removed without any alternative, KS/UI forms will be expected', since: '13.0')]
+    public function usesLegacyForms(): bool
+    {
+        return true;
+    }
+
+    public function getCustomConfigurationInput(
+        \ILIAS\UI\Factory $ui_factory,
+        \ILIAS\Refinery\Factory $factory,
+        ilLanguage $lng
+    ): \ILIAS\UI\Component\Input\Container\Form\FormInput {
+        throw new RuntimeException('Not implemented');
+    }
+
+    /**
+     * @deprecated
+     */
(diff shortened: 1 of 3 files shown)


mjansenDatabay:
Since this (kind of) adds new methods to the public interface (here: abstract base class) and contains an announcement regarding future removals of methods, I'll tag this PR with the JF label.

klees:
Hi @mjansenDatabay,

really looking forward to this!

I have two questions:

* For my curiosity about that new `#[\Deprecated(...)]` construction: Is the second parameter a version?
* Regarding the design: For the [Activities](https://github.com/klees/ILIAS/blob/11/activities/components/ILIAS/Component/src/Activities) we are currently trying to solve a somewhat similar problem. We [decided to just use FormInput there](https://github.com/klees/ILIAS/blob/11/activities/components/ILIAS/Component/src/Activities). Due to the monoidal structure of the inputs, we can simply treat a group of (sections of groups of ...) inputs as a simple input. This makes the interface simpler but also the implementation on our end, while still offering a great deal of flexibility to our users. Did you consider this option too? Is there a strong reason to use an array of sections here?

Kind regards!

mjansenDatabay:
Hi @klees ,

> For my curiosity about that new #[\Deprecated(...)] construction: Is the second parameter a version?

Yes, that's the ILIAS version. I used the syntax shown here https://wiki.php.net/rfc/deprecated_attribute .

> Regarding the design

To answers this, I'll have to first study the references you mentioned above. I will, but not before January (tomorrow is my last working day in 2024).

Best regards,
Michael

matthiaskunkel:
Jour Fixe, 06 JAN 2025: Michael notified us about changes at the central cron component and the removal of methods. PR is not final yet. But all developers should know about the impact of this PR. Michael would like to get rid of the support of the legacy cron implementation with ILIAS 13 at the latest.

mjansenDatabay:
I changed the approach by accepting a single `FormInput` provided by the concrete job implementation.

PR #8740, opened 2024-12-17
Setup: Change documentation for objectives and migrations

tbongers-cat:
Some additions to the Setup documentation, as discussed with @klees


klees:
Picked to trunk.

PR #8746, opened 2024-12-17
UI: add Input\Container::withInput for awesome flexibility

klees:
Hi all,

this is something that came up when @jeph864 and I were discussing [Activitities](https://github.com/klees/ILIAS/tree/11/activities/components/ILIAS/Component/src/Activities) and webservices.

The current plan is to use the inputs from the UI framework as a description for the input to Activities. @jeph864 is looking how these Activities can be published via a json-via-http API and, for this purpose, will need a way to provide values to the input. At first we thought that we might want to teach the UI framework how to deal with JSON input in a `ServerRequestInterface`, but then we found this solution to be more flexible and also simpler. This could open up the inputs from the UI framework to even more forms of user interface, such as the CLI.

The implementation should be quite obvious, but I'll wait until I have approval from @Amstutz and @thibsy here.

What do you think?

Kind regards!

### DIFF
--- a/components/ILIAS/UI/src/Component/Input/Container/Container.php
+++ b/components/ILIAS/UI/src/Component/Input/Container/Container.php
@@ -24,6 +24,7 @@ use ILIAS\UI\Component\Component;
 use ILIAS\Refinery\Transformation;
 use Psr\Http\Message\ServerRequestInterface;
 use ILIAS\UI\Component\Input\Input;
+use ILIAS\UI\Component\Input\InputData;
 
 /**
  * This describes commonalities between all Containers for Inputs, such as Forms.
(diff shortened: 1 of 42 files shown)


thibsy (changes requested):
Hi @klees,

Thx a lot for your contribution to the UI components!

Introducing a flexible way on our input containers to provide input data seems plausible and should interact seamlessly with our existing logic.

We have one minor change, which should not stop this from going to JF or postpone your implementation. However, please implement the following changes:

- [x] `withInputData()`: please rename this method to `withInput()` so it is similar to the `Input::withInput()` method. `withInputData()` may be more appropriate, but since it essentially does the same thing, we should try to be consistent here.

Kind regards,
@Amstutz and @thibsy

klees:
@thibsy/@Amstutz: Done. I also added the implementation now.

matthiaskunkel:
Jour Fixe, 06 JAN 2025: We highly appreciate this suggestion and accept the PR for trunk.

thibsy (approved):
LGTM, thx a lot!

PR #8755, opened 2024-12-18
[FIX] GS: use legacy initialisation adapter for endpoints.

thibsy:
Hi @chfsx,

I noticed that the endpoints of the global screen are still using the legacy initialisation directly, instead of the new adapter. Since #7969 we can no longer call `ilInitialisation::initILIAS()` directly, because already bootstrapped components would be missing (such as UI). This should fix some errors with the main menu.

Kind regards
@thibsy


chfsx:
Thanks a lot for the PR!

PR #8756, opened 2024-12-19
Fix #43147: LTI 1.3 Cosumer authentication fails with provider on different domain

fneumann:
https://mantis.ilias.de/view.php?id=43147

Set an addititional PHPSESSID cookie for LTI 1.3 authentication
 - The standard cookie does not work with LTI tools on different domains because it has samesite = Lax
- The additional cookie sets samesite to None so that it is sent with the POST request from the LTI tool
 - Limit the additional cookie to the path of the ltiauth.php script
 - Force secure to require HTTPS (needed for samesite = None)

See https://web.dev/articles/samesite-cookie-recipes?hl=en#unsafe-requests

### DIFF
--- a/Modules/LTIConsumer/classes/class.ilLTIConsumerContentGUI.php
+++ b/Modules/LTIConsumer/classes/class.ilLTIConsumerContentGUI.php
@@ -84,6 +84,23 @@ class ilLTIConsumerContentGUI
                     $this->dic->toolbar()->addText($this->getStartButtonTxt11());
                 }
             } else {
+
+                // Set an addititional PHPSESSID cookie for LTI 1.3 authentication
+                // - The standard cookie does not work with LTI tools on different domains because it has samesite = Lax
+                // - The additional cookie sets samesite to None so that it is sent with the POST request from the LTI tool
+                // - Limit the additional cookie to the path of the ltiauth.php script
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
Hi @Saaweel,
Please decide whether to close this PR, request a new PR targeting a maintained ILIAS branch, or adopt the changes yourself for a maintained release.

Saaweel:
@mjansenDatabay I agree! We're making a decision on what to do, thanks!

ZallaxDev:
Hi @mjansenDatabay,
We were reviewing these changes to incorporate them also in ILIAS 10. With this procedure, we discovered that although this solves the problem in some scenarios, it does not do it in all of them. The solution that we were able to implement in ILIAS 10, although it goes through the Cookie handling, requires other changes in different files.
We need to study how to implement the same solution in the current LTI code of ILIAS 9, which at the moment is quite different from ILIAS 10 and fix other problems that currently prevent the correct functioning of LTI in ILIAS 9, so we will work to find the necessary funding and solve all these problems as soon as possible.

We take this opportunity to thank @fneumann as this PR has given us valuable information on how to completely fix the problem in ILIAS 10.

We will mark this PR as closed and will try to integrate into ILIAS 9 when possible.

PR #8759, opened 2024-12-19
Obj: 43324, _lookupDeletedDate will return null for missing entry

nhaagen:
https://mantis.ilias.de/view.php?id=43324


kergomard:
Thank you very much for the PR @nhaagen ! Merged and picked to ILIAS 9 and trunk.

PR #8767, opened 2024-12-19
PRG: 43301, mail: use specific dic to keep context

nhaagen:
https://mantis.ilias.de/view.php?id=43301

The bug is originally in ilPRGPermissionsHelper having a context (prg_ref_id) of -1 when constructed w/o a current PRG.
it then continues over ilOrgUnitPositionAccess::filterUserIdsByPositionOfUser and ilOrgUnitPositionAccess::getTypeForRefId - which is empty.


klees:
Seems to be good ILIAS >9 already...

PR #8769, opened 2024-12-19
LSO: 43208, fix sending a globalScreen message after completing a lea…

lukastocker:
…rning sequence.

https://mantis.ilias.de/view.php?id=43208


klees:
Hi @lukastocker,

the change of order in the calls to `withNewAmount` and `withClosedCallable` seems the code that solves the issue. Could you explain what is going on here? The code looks as if it should not make any difference.

Kind regards!

klees:

PR #8777, opened 2024-12-31
T&A Bugfix: Download Certificates & Logger error

fhelfer:
Hey @kergomard,

please look into changes of this PR.
My IDE ran into these two undefined errors.
It's just a quick fix, and there are better solutions for it. Please let me know what are the best practices to do so.

Best @fhelfer


kergomard:
Thank you very much for the PR @fhelfer ! And sorry I missed it: Fixed in the meantime.

By-the-by: The whole `->deliver()` part never failed, because this code was never reached ;-). It could simply be removed.

Best,
@kergomard

PR #8778, opened 2024-12-31
[FIX] Docs: enforce file extensions for ES6 import specifiers.

thibsy:
Hi folks,

Technically, it is not possible to create JavaScript bundles using Rollup.js right now, because our code-style disallows file extensions in ES6 import specifiers, but the module bundler requires them. This is due to the [Node.js specification of import specifiers](https://nodejs.org/api/esm.html#esm_import_specifiers), which states that a file extension is always necessary for relative specifiers. Since we are maintaining a repository and cannot use absolute specifiers, we need to make a code-style adjustment.

This PR therefore changes:

* the ESLint configuration to enforce file extensions in import statements, except for specifiers that are resolved by other tools like Node.js itself or Rollup.js
* the JavaScript code-style to document why the change is necessary

Kind regards,
@thibsy

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -55,6 +55,9 @@
         "devDependencies": true
       }
     ],
+    // enforce file extensions for Node.js compatibility
+    // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md
+    "import/extensions": ["error", "ignorePackages"],
     // allow to reassign argument properties for e.g. altering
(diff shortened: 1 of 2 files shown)


matthiaskunkel:
Jour Fixe, 06 JAN 2025: Thibeau notified us about this change in the JavaScript handling for 10 and trunk.

thibsy:
- picked for release_10: 6b2749cafefc548670df73c6af8f2cc109b23d21

PR #8784, opened 2025-01-06
UI/Form: 43303, describedby not allowed on form, role='form' superflous

nhaagen:
https://mantis.ilias.de/view.php?id=43303
Removes attributes 'role' and 'decribedby' from form, which are unnecessary/forbidden according to w3c.
I also could not see any difference using a screenreader.


Amstutz:
thx a lot!

PR #8785, opened 2025-01-06
Obj (OrgU): 43252, load lang-module 'obj' in constructor of ilObjectGUI

nhaagen:
https://mantis.ilias.de/view.php?id=43252

Apparently, lang-module 'obj' is not always loaded; I placed it on ilObjectGUI, that seemed most fitting.


kergomard:
Thank you very much @nhaagen ! Merged and picked to trunk.

PR #8787, opened 2025-01-06
[FIX] UI: migrate JavaScript unit tests to Node.js

thibsy:
Hi folks,

This PR supersedes #7979, which migrates the JavaScript unit tests from `mocha` to `jest`, and migrates them to the native Node.js test runner and assertion library instead. You can find the documentation at:

- https://nodejs.org/api/test.html, and
- https://nodejs.org/api/assert.html

This gets rid of the `mocha` (or `jest`) and `chai` npm packages, however, we still need `jsdom` for simulating some browser functionality. Furthermore, this gets rid of the `@babel/preset-env` and `@babel/register` packages, which have been required to run unit tests in the Node.js environment until now. So, I think this is a huge win :).

The migration was quite easy to perform, since most of the code could be migrated using different imports, but there are some issues left:

- the toast UI component is not yet implemented as an ES6 module, which makes it untestable. "Fixing" the unit tests would mean to migrate this component to ES6 syntax - so I simply skipped them for the time being. I added this to the roadmap.
- the drilldown UI component tries to compare two HTML strings in `dom is correct after init`. While looking over the test case, I noticed that the unit test checks against HTML that is IMO not the result we are looking for. Maybe @kergomard would be so kind to look into this?
- the notification-item and counter UI component unit tests depend on a real instance of jQuery to be loaded. I could not figure out how to achieve this, since the environment is different from browsers and jQuery is not provided as an ES6 module. Since we don't want to use jQuery anyways, I skipped these unit tests for the time being and added this to the roadmap instead.

Some notable changes to the assertion library are:

- many of the unit tests, like e.g. `.to.be.instanceOf()` or `.to.be.string()` need to be implemented using a simple `equal()` assertion now. This means, e.g. `expect(val).to.be.string()` becomes `strict.equal(typeof val === 'string', true)` now. However, this simply removes some sugar since the underlying assertion looked somewhat like this anyways.
- there have been some `.not.to.deep.include(someValue)` assertions, which are not possible anymore. However, IMO this can easily be changed to an assertion which checks against the true nature of the value - which is what I did for this migration.

Please note that the unit tests are currently failing, because I added the `jsdom` package to `package_new.json`, instead of adding it to the actual `package.json`. I am not sure about the timing here, should I add the package directly, so we can merge once the dependencies are in? Otherwise I might need to disable the JS unit test workflow again.

P.S. this needs to be picked for ILIAS 10 as well.

Kind regards,
@thibsy

### DIFF
--- a/.babelrc.json
+++ b/.babelrc.json
@@ -1,7 +1,0 @@
-{
-  "presets": [
-    [
-      "@babel/preset-env",
-    ]
-  ]
-}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,14 +9,7 @@
     // globals from the browser environment.
     "document": "readonly",
     "window": "readonly",
-    "il": "writable",
-    // globals from the mocha testing framework.
-    "beforeEach": "readonly",
-    "afterEach": "readonly",
-    "describe": "readonly",
-    "before": "readonly",
-    "after": "readonly",
-    "it": "readonly"
+    "il": "writable"
   },
   // minified and bundled scripts are exempt from the
   // code-style.
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -44,7 +44,7 @@ jobs:
           GHRUN: "yes"
 
       - name: JS Unit Test
-        run: exit 0 # npm test
+        run: npm test
         env:
           GHRUN: "yes"
 
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         php: [8.3, 8.4]
-        nodejs: [ 20.x ]
+        nodejs: [ 23.x ]
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,8 +1,0 @@
-{
-  "require": [
-    "@babel/register"
-  ],
-  "spec": [
-    "components/ILIAS/**/tests/**/*.js"
-  ]
-}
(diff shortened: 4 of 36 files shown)


thibsy:
Assessment:

This package is a JavaScript implementation of various web standards, for use with the Node.js environment.

General Information:
- Name of the dependency: `jsdom`
- Version: `25.0.1`
- [x] this dependency was already used in ILIAS.

Type of dependency:
- [ ] composer
- [x] npm

Usage:

* various JS unit tests in `components/ILIAS/UI`

Reasoning:

* this package implements the web-standard of browsers and provides objects like `window` and `document`, which are unknown to the Node.js environment. These objects can be injected into our components to perform/test desired actions.

Maintenance:

* The package is actively maintained, several releases last year (2024).
* The package has an active commit history.
* The package is widely used, 14+m according to GitHub.

Links:
* Packagist/NPM: https://www.npmjs.com/package/jsdom
* GitHub: https://github.com/jsdom/jsdom
* Documentation: https://github.com/jsdom/jsdom/blob/main/README.md

Alternatives:

* `linkedom`: however, its not as widely used (~15K) and [does not do the exact same thing.](https://github.com/WebReflection/linkedom?tab=readme-ov-file#how-does-it-work)

matthiaskunkel:
Jour Fixe, 03 FEB 2025: We highly appreciate this suggestion and accept the PR for 10 and trunk, incl. the acceptance of the dependency jsdom for 10 and 11. PM will add information to the related dependency lists.

klees:
@thibsy Please rebase and then ping me for immediate merge. Thanks a lot!

klees:
Aaah, thick thump....

thibsy:
@klees I have rebased. However, maybe we should wait until #9086 is merged. This PR requires `jsdom` to be installed, which is not the case yet. Until then pipelines would fail. Also, it would allow me to amend the commit and install the package directly, making it easier for cherry-picking it to release_10 (package is missing there too).

thibsy:
@Amstutz and @klees, this can be merged now. Let me know if you have trouble picking to ILIAS 10.

klees:
Picked `[FEATURE]` commits to release_10...

PR #8792, opened 2025-01-06
CSS: Set image max-height

iszmais:
https://mantis.ilias.de/view.php?id=43464

This might not be the desired implementation, but should just visualize the desired change.


Amstutz:
Thx. This fits the description: 

```
     * description:
     *   purpose: >
     *     A responsive image is to be used if the image needs to adapt to changing
     *     amount of space available.
     *   composition: >
     *     Responsive images scale nicely to the parent element
```

Ill therefore merge this.

PR #8793, opened 2025-01-06
CopyWizard: Fix duplicate copies (in 2nd - nth target) when selecting…

mjansenDatabay:
… multiple targets

See: https://mantis.ilias.de/view.php?id=42234


kergomard:

PR #8795, opened 2025-01-07
TA/FormulaQuestion: 43313, results presentation and calculation of po…

nhaagen:
https://mantis.ilias.de/view.php?id=43313


kergomard:
Thank you for the PR @nhaagen !

There were so many side-effects from the changes in the building of the user_solutions-array, that I decided to role back the whole thing. See: https://github.com/ILIAS-eLearning/ILIAS/commit/a254ec18358735dfbaa461cabae9ace065fbe93b

Best,
@kergomard

nhaagen:
ja, I saw that, sorry for causing the trouble.
But it hints to the need for some proper handling of user answers...

PR #8797, opened 2025-01-07
changes to security.md for patch application

rfalkenstein:
This should specify the git format-patch command. I think the current documentation is not correct since git can't know which commits are supposed to be chosen if you only input the branch name as a parameter.

### DIFF
--- a/docs/development/security.md
+++ b/docs/development/security.md
@@ -51,10 +51,10 @@ to the corresponding patch files to the issue manager through a secure channel.
 The following process MUST be followed to hand in a fix for a security issue:
 * Create one patch file per supported branch needing the fix. The patch file MUST
 be named "<ilias_version>_<title_of_fix>.patch" (e.g. "9_my-very-important-fix.patch").
-To create the patch use `git format-patch` with the option `--stdout`
-(e.g. `git format-patch latest-not-included-commit-id-of-current-branch --stdout > 9_my-very-important-fix.patch`)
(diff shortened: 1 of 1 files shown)


rfalkenstein:
@klees  @mjansenDatabay

klees (commented):
Hi @rob0403,

thanks for the proposal. I think it heavily relies on the individual approach to branching if the one or the other command is correct. At [CaT](https://www.concepts-and-training.de) (and I think others do so as well...), we create new branches for every individual issue, based on the release-branch. I think that the current command would be the correct command for that case, since effectively all changes required to fix the issue and only these changes are included in the individual branch.

If branches are used for other purposes (e.g. to contain patches for some individual installations) and the fix is pushed on top of that branch, the command you have proposed would be correct. 

I would propose to present the two variations of the command as options. Maybe point out the crucial thing a bit more: The patch file should contain the required changes to fix the issue, nothing more, nothing less.

Kind regards!

rfalkenstein:
Okay, I got it. It was not clear to me that the current doc meant the release-branch of the release you are patching against when saying "my-bug-fixing-branch". Then it works as suggested. I interpreted this as the name of the newly checked-out branch you are doing the bugfixes in.

rfalkenstein:
@klees  I tried to make it clearer. Thanks for your hint. If anyone has a better idea for a more general wording please feel free :-)

klees (approved):

acgruber:
Thank you @rob0403 for your contribution and @klees for your review. Merged into trunk.

PR #8798, opened 2025-01-07
LSO: 42703, create new obj seems to be missing the target

nhaagen:
https://mantis.ilias.de/view.php?id=42703


klees:
Picked to trunk.

PR #8800, opened 2025-01-08
OrgU: 42554, param handling when deleting position

nhaagen:
https://mantis.ilias.de/view.php?id=42554

besides adjusting the paramter handling (its a post, no ids in query), I also removed the unused and apparently wrong lang_var msg_position_u**dp**ated. Finally, not really in OrgU, but closely related, the StudyProgrammeMembershipSourceReaderOrgu is triggered via event when deleting the position and threw due to getChildren beeing private on orguTree


klees:
Picked to trunk.

PR #8801, opened 2025-01-08
LegalDocuments: Fix label & dpro column in user administration table

lscharmer:
This PR fixes Mantis Bug: https://mantis.ilias.de/view.php?id=41010

Beside the language variable change, the column for the data protection was missing. As this is saved on the `usr_pref` table, this PR adds a special case to join this table, similar to the `online_time` field.


kergomard:
Thank you very much for the PR @lscharmer ! Merged and picked to ILIAS 10 and trunk.

mjansenDatabay:
Thx @kergomard

PR #8805, opened 2025-01-08
[FEATURE] UI: improve segregation of public and internal interfaces.

thibsy:
Hi folks,

This PR may seem huge, because it affects almost all factory implementations and related unit tests, but the actual changes can be summarised to this:

* Adds documentation about the public and internal interface segregation, see [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8805/files#diff-d6d390b074675b9a85fb650b7a76f4a7643d4edd8d6f9847b6d09957f1ec6959).
* Adds DI for ALL nested UI factories and adds them to `Init\AllModernComponents`. There have been several factories which were created dynamically on each call.
* Changes all UI factory implementations to return concrete implementations rather than the interface. The reason for this becomes clear when reading the document about change 1.
* Removes all unnecessary PHPDoc annotations of UI factories. I needed to remove all `@inheritDoc` annotations in order for intelisense to suggest the concrete implementation only. While at it I also removed some other unnecessary comments.

What the PR does at its essence: create a clear boundary between public and internal context, so we can start implementing internal components and only expose them internally.

I am happy to discuss this PR with you :).

Kind regards,
@thibsy


Amstutz:
thx a lot for the large clean-up, and adaptation of the docu! Highly appreciated. You are welcome to merge, as soon as the conflicts are resolved.

@klees and @Amstutz

PR #8812, opened 2025-01-09
introduce catamorphism for ui components

nhaagen:
UI Components inductively built upon each other, i.e. components will contain other components.
It is, for some components, possible to access their substructure, but always in a specific manner; there is no generic way to get from the containing component to its children.

Now, @jeph864 is working on Activities and the idea arose to use forms/inputs to publish APIs (see https://github.com/ILIAS-eLearning/ILIAS/pull/8746). What's missing right now is a simple way to fold down the inputs into e.g. some json-object, and while at it, this is true for all components.

Therefore, I implemented a mechanism on the componentHelper to walk down the object tree and apply operations to the components, storing the result in a property. This result does not need to stay a component, and you can see its effect quite clearly in the examples included here.  Also, modification of the components' instances are perfectly feasible.

Components need to implement `getSubStructure(): array ` to participate in the catamorphism; if they don't or return an empty array, recursion just stops and the component is considered a leaf-node.

This implementation comes with Container and Field\Group for now, but is explicitely meant for all UI Components.
I fiddled around wit the Panel, e.g., and it behaves just alike.

### DIFF
--- a/components/ILIAS/UI/src/Component/Component.php
+++ b/components/ILIAS/UI/src/Component/Component.php
@@ -30,4 +30,32 @@ interface Component
      * Get the canonical name of the component.
      */
     public function getCanonicalName(): string;
+
+    /**
+     * This implements the catamorphism (https://en.wikipedia.org/wiki/Catamorphism)
+     * for components, which is a clever way to implement a generalized fold over
+     * data structures.
+     *
+     * The scheme starts at the leaves of the structure and applys the function to
+     * each leave and moves up the tree recursively. The return value of the function
+     * is put into the "sub structure" to be consumed when the function is applied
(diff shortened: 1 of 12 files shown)


klees:
Hi @thibsy and @Amstutz,

please have a look into this. I changed @nhaagen|s initial implementation slighlty: `Component`s that use `ComponentHelper` now can implement `getSubComponents` to announce their substructure.

This is missing implementations of `getSubComponents` for many components. There are components components (such as `Link`...) that can use the default implementation (for "no substructure"), but many components would need a better implementation. I would propose to add these implementations afterwards, once we have agreed on the general mechanism.

Kind regards!

jeph864:
Hey all, I just accidentally merged the PR. I will revert it

PR #8815, opened 2025-01-10
Docs/Authorities: Remove `PerPascalSeeland` from authorities for "Log…

mjansenDatabay:
…in, Auth & Registration


pascalseeland:
Thanks @mjansenDatabay for taking over the authorities since I will be on parental leave until Jan. 2027.

PR #8816, opened 2025-01-10
Object: Move Translation Table To System Folder

kergomard:
Hi @alex40724 

I started to look into the translations table in `ILIASObject` and found this class that is only used in `SystemsFolder`. I would like to move it there. I renamed the table and the corresponding template to what it actually does.

Best,
@kergomard


kergomard:
Hi @alex40724 

A little ping to see, if we can get this through the door.

Thank you very much and best,
@kergomard

alex40724:
Thanks!

PR #8818, opened 2025-01-10
UI: move UI\Component\Table\Storage up to UI\Storage

nhaagen:
Storage is quiet generic; as discussed, I'll move it up some levels so other Components will use it, too.


thibsy (changes requested):
Hi @nhaagen,

- [x] Interface description: while we are at it, could you also provide a brief interface description, which describes the nature of this storage? this way, other components and also implementors will know, how they need to treat values stored in this object.

Kind regards,
@Amstutz and @thibsy

thibsy (approved):
Thx a lot @nhaagen for the brief explanation. This LGTM, so I will merge this now and we can announce this at JF afterwards (due to the minimal interface change).

PR #8823, opened 2025-01-10
T&A: 43407: Fix missing StatusOfAttempt

lukas-heinrich:
Hi everyone, 

this PR refers to the ticket [43407](https://mantis.ilias.de/view.php?id=43407), which describes an incorrect display of the Status Of Attempts.

The behavior described can be traced back to the fact that the object's properties are not set at any point, which I have added. Setting `StatusOfAttempt::RUNNING` at this point may look irritating. When debugging, however, I was able to determine that the `getAttemptOverviewInformation()` method is not executed if there is no pass. Accordingly, it can be assumed that the status at this point is at least _Running_ or that the test run has already been completed.

I look forward to your feedback and am open to suggestions or questions. This PR was co-authored and internally reviewed by @thojou.

Best,
@lukas-heinrich

### DIFF
--- a/components/ILIAS/Test/classes/Evaluation/class.ilTestEvaluationFactory.php
+++ b/components/ILIAS/Test/classes/Evaluation/class.ilTestEvaluationFactory.php
@@ -18,6 +18,7 @@
 
 declare(strict_types=1);
 
+use ILIAS\Test\Results\Data\StatusOfAttempt;
 use ILIAS\Test\Scoring\Marks\MarkSchema;
 
 /**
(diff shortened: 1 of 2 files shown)


kergomard:
Hi @lukas-heinrich 

Thank you very much for the PR! I fixed this a little bit differently with https://github.com/ILIAS-eLearning/ILIAS/commit/ade2a77ea8a6b2b85bc42d75da21fbca1ed8f31b on a drive-by (sorry for not realizing you had already provided a PR) and then picked your cleanup manually.

Best,
@kergomard

PR #8825, opened 2025-01-10
T&A: Ghost Attempts & correct detection of unfinished test passes

lukas-heinrich:
Hi everyone,

with this PR, I propose a change to correctly recognize unfinished test attempts. Originally this is a fix for [28447](https://mantis.ilias.de/view.php?id=28447), but I was able to find some other places that should be fixed with this change:

- https://mantis.ilias.de/view.php?id=42856
- https://mantis.ilias.de/view.php?id=27077
- Incorrect number of test attempts in the results view
- Blank test results in the detailed test view and man scoring
image

### Reproduction
To reproduce the behavior, make sure that the test has at least the following settings:

- **Overview of Answers Given**: checked
- Set Time Limit for Completing Test \> **Maximum Time Available**: e.g. 1 min
- Set Time Limit for Completing Test \> **Reset Time Limit for All Test Attempts**: checked

### Cause of the behavior

- The fundamental problem is that when resetting the Processing Time, the pass is incremented by one ([`\ilObjTest::isExecutable`](https://github.com/ILIAS-eLearning/ILIAS/blob/release_8/Modules/Test/classes/class.ilObjTest.php#L7467C1-L7467C58))
- This is why `self::_getPass($active_id)` returns a pass that doesn't exist
- When finishing the attempt via the button or cron job, the pass is incremented again (see ` \ilTestPassFinishTasks::performFinishTasks`)

### Tests

To verify this change, I tested the following scenarios and checked whether the number of passes and the unfinished passes are shown as expected.

- Attempt completed
- Attempt not completed
- First attempt completed, second attempt completed
- First attempt completed, second attempt not completed

I ran these four cases both for a test without a time limit and for a test with a limit, both with and without a reset. I also checked the database to see if there are any ghost passes.

I am looking forward to your feedback and comments. This PR was internally reviewed by @thojou .
Best,
@lukas-heinrich

### DIFF
--- a/Modules/Test/classes/class.ilObjTest.php
+++ b/Modules/Test/classes/class.ilObjTest.php
@@ -5621,16 +5621,12 @@ class ilObjTest extends ilObject implements ilMarkSchemaAware
             && $active_id > 0
             && ($starting_time = $this->getStartingTimeOfUser($active_id)) !== false
             && $this->isMaxProcessingTimeReached($starting_time, $active_id)) {
-            if ($allow_pass_increase
-                    && $this->getResetProcessingTime()
-                    && (($this->getNrOfTries() === 0)
-                || ($this->getNrOfTries() > (self::_getPass($active_id) + 1)))) {
+            if ($allow_pass_increase && $this->getResetProcessingTime()) {
                 // a test pass was quitted because the maximum processing time was reached, but the time
                 // will be resetted for future passes, so if there are more passes allowed, the participant may
                 // start the test again.
                 // This code block is only called when $allowPassIncrease is TRUE which only happens when
                 // the test info page is opened. Otherwise this will lead to unexpected results!
-                $test_session->increasePass();
                 $test_session->setLastSequence(0);
                 $test_session->saveToDb();
             } else {
--- a/Modules/Test/classes/class.ilObjTest.php
+++ b/Modules/Test/classes/class.ilObjTest.php
@@ -5621,18 +5621,8 @@ class ilObjTest extends ilObject implements ilMarkSchemaAware
             && $active_id > 0
             && ($starting_time = $this->getStartingTimeOfUser($active_id)) !== false
             && $this->isMaxProcessingTimeReached($starting_time, $active_id)) {
-            if ($allow_pass_increase && $this->getResetProcessingTime()) {
-                // a test pass was quitted because the maximum processing time was reached, but the time
-                // will be resetted for future passes, so if there are more passes allowed, the participant may
-                // start the test again.
-                // This code block is only called when $allowPassIncrease is TRUE which only happens when
-                // the test info page is opened. Otherwise this will lead to unexpected results!
-                $test_session->setLastSequence(0);
-                $test_session->saveToDb();
-            } else {
-                $result["executable"] = false;
-                $result["errormessage"] = $this->lng->txt("detail_max_processing_time_reached");
-            }
+            $result["executable"] = false;
+            $result["errormessage"] = $this->lng->txt("detail_max_processing_time_reached");
             return $result;
         }
 


lukas-heinrich:
The branch is now updated and works for the `release_9` branch.

kergomard (commented):
Hi @lukas-heinrich 

Thank you very much for diving deeply into this issue. I'm sure this was all, but easy!

I stared at this for a while, but I'm not sure what to do with it:
1. This looks wrong. It looks like there could be cases we would be resetting the sequence to 0 on a running attempt.
2. Then I looked at the `$allow_pass_increase`-parameter, ...and saw that in ILIAS 10 it will never be set, in 9 it is set in one place (`ilObjTestGUI::buildPageViewToolbar()`), ...which again looks all kinds of wrong.
3. There are simply too many ways through a test, so that I would either have to merge this blindly or have to spend a lot of time figuring this out.

...as you have already stared at this for a while:
- Why are you sure that my doubt in point one above is unfounded?
- What does this mean for ILIAS 10?

Sorry, to hand this back to you like this, but I didn't follow my guts a couple of times lately, and my guts where always right and I should have followed them ;-).

Thanks again and best,
@kergomard

lukas-heinrich:
Hi @kergomard ,

Thank you very much for looking into this and providing feedback. I have indeed spent considerable time with this issue, so I would like to try to describe my findings and conclusions.

I hardly believe it is an error that the number of passes is increased by one at this point. This results in database values like `last_finished_pass` referencing a pass that doesn't actually exist. This leads to various errors as described above. Some of these issues are visible to users (e.g. the error message "*WARNING: the processing time for this user is not over yet! You should only end the test run if there is a compelling reason (e.g. exclusion from the test).*" when I try to end an expired attempt), while others (like the ghost attempts) are not visible directly.

Since I haven't found any scenario where the sequence influences the detection of unfinished attempts, I didn't question or modify this functionality. In my opinion, the outer branching ensures that the code is only executed when the processing time has been exceeded. I removed the two lines from the inner branching because I couldn't find any scenario where this condition is not true. However, since I understand the concerns, I could revert the changes to the branching to definitely rule out any undesired behavior.

The fact that the parameter `allow_pass_increase` is not set in ILIAS 10 further strengthens my assumption that changing the sequence at this point has no impact on processing time-exceeded passes. In ILIAS 10, this inner branching is never executed, and lo and behold, I couldn't reproduce the problem described in this PR.

Therefore, I conclude that at least the call to `increasePass` should be removed at this point. In my opinion, it should also be possible to remove the complete inner branching, leaving only the part from the else block. I am looking forward to opinions on this suggestions.

Best regards,
@lukas-heinrich

kergomard:
Hi  @lukas-heinrich 

Sorry, I lost this out of sight for so long!

Your thinking leads me to believe that we should actually do exactly that: Remove the inner branching. For me this feels far more comfortable. My Reasoning:
- Resetting the sequence only makes sense, if we are also starting a new attempt, otherwise we just throw the user back to the beginning. If we remove this though, saving the session becomes also unnecessary, as we didn't change anything, thus we can simply get rid of the whole block and I have the feeling, I again understand what is going on. I will need to trust your research on the side-effects, as I see so many places where this could go sideways.

What do you think? Would you adapt the PR accordingly?

Thanks again and best,
@kergomard

lukas-heinrich:
Hi @kergomard ,

thank you for the feedback! 
I fully agree, which is why I removed this inner branching. I have tested the change with various parameters and passes and could not detect any issues.

Best,
@lukas-heinrich

kergomard:
Thank you very much @lukas-heinrich ! Ok, we will try this, I don't think this is completely risk free, but as far as I can see any unwanted consequences shouldn't be catastrophic.

Merged and picked to trunk.

PR #8828, opened 2025-01-10
Customizing: remove directory /Customizing

nhaagen:
https://mantis.ilias.de/view.php?id=43496

This removes the /Customizing - directory.

There are 

- public/Customizing/plugins
- public/Customizing/skin and 
- /lang/Customizing 

for the respective files.


klees:
Hi @matthiaskunkel,

most of the changes here reflect what we have decided [on the JF](https://github.com/ILIAS-eLearning/ILIAS/pull/8018#issuecomment-2352849934) some time ago. There simply are locations that we have missed in the initial PR.

However, there is one thing we did not properly think about: What shall we do with custom lang files? If we'd move them along with the skins and plugins, they would end up in the web-folder, hence they could be delivered to the world via HTTP. This felt wrong, so @nhaagen proposed to use `lang/Customizing` for these files.

Do you have Any strong feelings here?

Kind regards!

matthiaskunkel:
Moving the customised language files to /lang/Customizing is fine for me.

matthiaskunkel:
How about taking the opportunity and rename the directory to 'customizing'?

klees:
Picked to trunk.

PR #8829, opened 2025-01-10
Fix offset not in array if on last page and filter is applied.

thojou:
Hey,

This PR addresses the issue described in Mantis Bug Tracker https://mantis.ilias.de/view.php?id=42959.

During debugging, it became clear that the problem was not limited to a specific table implementation, but affected all table implementations. Here’s a scenario to illustrate the problem:

If a table has multiple pages and the user navigates to the last page, applying a filter that reduces the total number of displayed pages by at least one will trigger the issue.
The root cause lies in how the data table range is cropped—this happens too late in the execution flow. To address this issue, the PR introduces two adjustments:

1. The `findCurrentPage` method now defaults to `0` instead of throwing a `LogicException` if no valid range is defined for the given offset.
2. Previously, the `croppedTo` method returned a zero-length range, which seemed counterintuitive. This is adjusted so that the method now returns to the interval of 0 to smaller user selection or standard configuration.

These changes are directed to resolve the data tables as a whole. 
Looking forward to your feedback and comments.

Best
@thojou

### DIFF
--- a/components/ILIAS/UI/src/Implementation/Component/Input/ViewControl/Renderer.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Input/ViewControl/Renderer.php
@@ -255,6 +255,7 @@ class Renderer extends AbstractComponentRenderer
 
         list(Pagination::FNAME_OFFSET => $offset, Pagination::FNAME_LIMIT => $limit) = array_map('intval', $component->getValue());
         $limit = $limit > 0 ? $limit : reset($limit_options);
+        $offset = $offset > $total_count ? 0 : $offset;
 
         if (! $total_count) {
             $input = $ui_factory->input()->field()->numeric('offset')->withValue($offset);
--- a/components/ILIAS/UI/src/Implementation/Component/Table/Data.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Table/Data.php
@@ -128,9 +128,14 @@ class Data extends AbstractTable implements T\Data
             $view_controls = $this->applyValuesToViewcontrols($view_controls, $request);
             $data = $view_controls->getData();
             $range = $data[self::VIEWCONTROL_KEY_PAGINATION];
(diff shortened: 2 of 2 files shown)


klees (changes requested):
Hi @thojou,

thanks for looking into this. I'm not really convinced that this the way to move forward.

Regarding the change in Data\Range: I understand that you feel the current behaviour is counterintuitive. But the current behaviour is the current behaviour, so changing it just like that could lead to all kinds of surprises for other people. The idea is that the method limits the range to the upper bound given as `$max`, while being as faithful to the lower bound as possible. We're already in some edge case there, where we cannot keep the lower bound as is. But moving the lower bound to 0 while keeping the length just feels arbitrary in the light of the regular behaviour of the method. The method might be documented better, I guess, but the behaviour seems consistent to me. Why would we need that change?

Regarding the change in `Renderer::findCurrentPage`: Defaulting to some page if we are out of range of available pages seems to be fine. Why would we jump to the first page (instead of the last...)? No strong oppinion here, just an honest question if there are any arguments to be repeated sometimes 🙂

Kind regards
@klees (in agreement with @thibsy  and @Amstutz)

Please assign back to me when I can move on with this.

thojou:
Hey @klees,

thanks for you feedback!
As discussed in the recent TechSquad meeting, I’ve tried a different approach to address the out-of-bounds issue.

Now, if the starting index is greater than the total number of items, the range is resetted to start from 0 within the `applyViewControls` method.
For pagination rendering, the offset is also resetted before attempting to find the current page. With this approach, the `LogicException` is not thrown.

Also, this solution avoids modifying the `croppedTo` and `fundCurrentPage` method.
Looking forward to your feedback and comments.

Best,
@thojou

klees:
Thanks @thojou!

klees:
Picked to trunk...

PR #8833, opened 2025-01-13
User: init js for actions in user gallery (43271)

schmitz-ilias:
This PR fixes the last little bit of [43271](https://mantis.ilias.de/view.php?id=43271) by loading the required js for actions in the user gallery.

Note that this is only an issue when 'Add to Group' is deactivated for the 'Who is online' tool, otherwise the js already gets loaded there.


kergomard:
Thank you very much @schmitz-ilias for this PR. I think this would be more expensive than strictly necessary though. I added the call to `init` to the constructor so it should actually only be called once (see: https://github.com/ILIAS-eLearning/ILIAS/commit/a2610d7578226672d6f2cb505b540cefa5388f1d).

Best,
@kergomard

PR #8834, opened 2025-01-13
Remove phpmailer patch

mBeym:
I'm unable to find a reason for this patch to still (or ever to) exist.

- the patch only ever "fixed" this for mails send through php's **mail()** method.
- phpmailer already adds **undisclosed-recipients:;** if no **to** address is set when send through **mail()**
- Trying to send a mail through phpmailer with an empty string **to** address set makes the send fail

Also see:
https://github.com/PHPMailer/PHPMailer/issues/2789


mjansenDatabay:
Thx @mBeym

PR #8836, opened 2025-01-13
OrgUnit: fix saving of adv md dates (43385)

schmitz-ilias:
This PR fixes [43385](https://mantis.ilias.de/view.php?id=43385), by basically swapping two lines. Not sure why that works exactly, legacy date inputs seem to be a bit finicky. It looks like this is also an issue in ILIAS 8 and 9, so this is not limited to the html-native datetime pickers.


klees:
Hi @schmitz-ilias,

thanks a lot! Did you check if this actually is a problem in ILIAS 9? So, can I pick this? Or is this more like a hunch =)

Kind regards!

schmitz-ilias:
Hi @klees, this is a problem in 9, feel free to pick down.

PR #8837, opened 2025-01-14
Plugins/Language: 43587, return path in public folder from ilPluginSlotInfo

nhaagen:
https://mantis.ilias.de/view.php?id=43587

### DIFF
--- a/components/ILIAS/Component/classes/class.ilComponentRepository.php
+++ b/components/ILIAS/Component/classes/class.ilComponentRepository.php
@@ -24,7 +24,7 @@ declare(strict_types=1);
  */
 interface ilComponentRepository
 {
-    public const PLUGIN_BASE_PATH = "public/Customizing/plugins";
+    public const PLUGIN_BASE_PATH = "Customizing/plugins";
 
     /**
      * Check if a component exists.
--- a/components/ILIAS/Component/classes/class.ilPluginSlotInfo.php
+++ b/components/ILIAS/Component/classes/class.ilPluginSlotInfo.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
--- a/components/ILIAS/Component/classes/class.ilPluginSlotInfo.php
+++ b/components/ILIAS/Component/classes/class.ilPluginSlotInfo.php
@@ -65,7 +65,11 @@ class ilPluginSlotInfo
 
     public function getPath(): string
     {
-        return ilComponentRepository::PLUGIN_BASE_PATH . "/" . $this->getQualifiedName();
+        return implode('/', [
+            ilComponentRepository::PLUGIN_BASE_PATH,
+            $this->component->getName(),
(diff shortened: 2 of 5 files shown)


klees (commented):
Hi @nhaagen,

thanks a lot for looking into this. I'm a little anxious that this damages other functionality, since we change that base path. Did you check:

* Where `ilComponentRepository::PLUGIN_BASE_PATH` is used overall?
* Where `ilPluginSlotInfo::getPath` is used?
* I any test plugin keeps working after this change?

If all "yes": **Phew** :sweat_smile:  If not: Please check and report back. Thanks!

Kind regards!

nhaagen:
I was reluctant, too, and then had to remove the "public/" part in ilPluginLanguage (see the first comimt of this PR)
In the according ticket, @christof756 pointed to some other misbehavior, and I changed the constant.

PLUGIN_BASE_PATH is really being used by ilPluginSlotInfo only, and ilPluginSlotInfo::getPath ist sort of hard to trace: It's used by ilPluginInfo, and that is involved every time the plugin is accessed.
However, TestSetupPlugin and EditorAsMode both work fine with the changes suggested here.

klees:
Picked to trunk.

fwolf-ilias:
@klees @nhaagen  

I need some clearification, because I already test some plugins with ILIAS 10 both in the testing phase and in the plugin development.

As I understand the component revision, the plugins are moving from `./Customizing/global/plugins/[Services/Modules]/` to `./public/Customizing/plugins/`. I had some difficulties to get the dbupdatesteps running because said ilPluginSlotInfo::getPath path was buggy. (It used the QulifiedName of the plugin slot component which broke the path). Now the path is missing the `public` bit because of this PR and isn't able to find the `./sql/dbupdate.php` of the plugin again. 

I was testing if the update was performed via cli/setup.php. Maybe the webinterface of the plugin administration is reacting to this path in a different way. At the moment i can't update dbupdatesteps for plugins with the setup cli.

klees:
Hi @nhaagen,

can you please look into this issue? This kind of sounds like the problems I was expecting...

Maybe it's better to roll this change back and instead add some special-case for the languages. Sounds like the old "web-relative-path" vs. "code-relative-path" problem to me.

Kind regards!

klees:
Hi @fwolf-ilias,

thanks for bringing this up. To understand you correctly: You had problems to trigger db_update.php from the GUI?

Kind regards!

nhaagen:
I'll look into this; @klees , you're right, it wepath vs FSpath allright.

fwolf-ilias:
Hi @rklees long story short: 

I wasn't able to trigger dbupdate.php from CLI before and after this PR due to two different reasons.
**Before PR:** The path to dbupdate.php is wrong because of the used QulifiedName
**After PR:** The path to dbupdate.php is wrong because ./public/ is missing

In both cases the steps are simply ignored, without error, because ILIAS thinks that this plugin has no dbupdates.

Can't tell if the GUI behaves differently.

klees:
Hehe, lolled literally. So many ways to break something =)

PR #8841, opened 2025-01-14
TestQuestionPool: Remove custom xml export option from export tab, the option is no longer needed

chlulei:
This PR fixes an issue with the export of Test Question Pools, see [43108](https://mantis.ilias.de/view.php?id=43108), by removing the custom XML export option in Test Question Pool in favor of a standard XML export.

There are however a few follow up issues we stumbled on while investigating the problem, which seem to run a bit deeper. There are two different conflicting export mechanisms in Test Question Pool (one on the export tab and one in the question table). The content of both these types of exports is subtly different, but in ILIAS 9 they are not differentiated in the UI. In ILIAS 10 on the other hand, because of the changes in the export mechanism (both internally in Test Question Pool and in the Export component), the latter is not shown in the Export tab anymore, but remains as a file in the data folder of ILIAS.

The situation seems on the whole a bit muddled, and we think there is a good opportunity for a bit of clean up here, since so much has changed in the Export with 10 anyways. It would for example be possible to add the 'pure' question-export as a custom export option, such that it shows up in the Export tab. Maybe it would be best to get together in the near future and figure out how Test (Question Pool) and Export can work together a bit more neatly going forward.

What do you think?

Best,
@schmitz-ilias and @chlulei


schmitz-ilias:
Ah, I see you are already thinking along similar lines in #8624...

fhelfer:
Thx @schmitz-ilias,

I would suggest we close this PR as I will copy your changes to #8624 and also delete ilTestExportOptionXML.php along the way. With that, we can get rid of the special handling of Test-Component in ilExportGUI.

Best, @fhelfer.

schmitz-ilias:
Yeah that works for us, thanks @fhelfer

kergomard:
Thank you very much for the PR @chlulei and @schmitz-ilias ! Yes, clean-up is needed and if everything goes well it will be @fhelfer or one of his colleagues who will take care of this, so somebody will be in touch. We will need to see how we move on with the single question export, I'm not sure it makes much sense and thanks for the hint on the file being left in the data folder!

PR #8844, opened 2025-01-14
Possible Fix for Mantis 43614: Fix duplicate global roles for FILTER_ALL

ingoj:
and eliminate duplicate declare strict_types


kergomard:
Thank you very much @ingoj ! Merged and picked to ILIAS 10 and trunk.

Best,
@kergomard

PR #8846, opened 2025-01-14
Badge/User: Improve toggling hidden badges in public profile

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=43589

If approved, this has to be picket to `trunk`.


mjansenDatabay:
@kergomard A review of the changes in "User" is highly appreciated. Of course it is still ugly that the "Badge" component has to ...

1. Modify a template in the "User" component
2. Add PHP code to `ilPublicUserProfileGUI`

But as long as the public profile did not offer any kind of "Slot System" for other compoents, this PR at least ...

* Fixes the linked issue
* Removes depenencies on "jQuery"
* Moves the JS code from the HTML template to an ES6 class

kergomard (approved):
Thank you very much @mjansenDatabay ! This looks far cleaner even though you are right, that there is still quite some room for improvement, ...but for this I would need to put some work in, so it is not on you.

PR #8848, opened 2025-01-15
Quick update of italian language

padvincenzo:


matthiaskunkel:
Thanks to @padvincenzo for the update.

PR #8852, opened 2025-01-15
[FIX] Component: return `Component\EntryPoint` status

thibsy:


klees:
Picked to release_10.

PR #8853, opened 2025-01-15
COPage Editor: Add entrypoint for custom functions

iszmais:
In dependency of the implementation of 
https://docu.ilias.de/go/wiki/wpage_8375_1357 
this PR proposes an inherited function through which specific PageObject Editor can provide custom selectable placeholder function for their scope.

This change allows specific scopes to integrate their custom CO features into the Page editor itself instead of injecting them into the rendered HTML.

---

The features’ implementation can be found [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8939) for better visualization

### DIFF
--- a/components/ILIAS/COPage/Editor/js/src/components/paragraph/actions/paragraph-action-types.js
+++ b/components/ILIAS/COPage/Editor/js/src/components/paragraph/actions/paragraph-action-types.js
@@ -12,7 +12,7 @@
  * https://www.ilias.de
  * https://github.com/ILIAS-eLearning
  *
- ******************************************************************** */
+ *********************************************************************/
 
 const ACTIONS = {
 
@@ -57,5 +57,6 @@ const ACTIONS = {
   SPLIT_PARAGRAPH: 'par.split',
   MERGE_PREVIOUS: 'merge.previous',
   SECTION_CLASS: 'sec.class', // section format
+  INSERT_TEXT_TEMPLATE: 'insert.text.template',
 };
 export default ACTIONS;
--- a/components/ILIAS/COPage/Editor/js/src/components/paragraph/ui/paragraph-ui.js
+++ b/components/ILIAS/COPage/Editor/js/src/components/paragraph/ui/paragraph-ui.js
@@ -1,9 +1,3 @@
-import ACTIONS from '../actions/paragraph-action-types.js';
-import PAGE_ACTIONS from '../../page/actions/page-action-types.js';
-import TinyWrapper from './tiny-wrapper.js';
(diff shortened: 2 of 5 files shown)


alex40724 (changes requested):
Hi @iszmais 

I understand the need to extend the editor. Unfortunately it is not well prepared to be extended this way yet. The menu related functions are not well located in ilPageObject, I am thinking about moving them to the Paragraph, but they are used by other components like the tables, too.

I can't approve this kind of change currently, here are some concrete issues I see:

On the one hand it provides a very generic naming ("custom" event, "getCustomFunctions") in a very central place (ilPageObjectGUI). On the other hand it seems to just allow to add simple text templates within the Paragraph component.

Having this in ilPageObjectGUI just makes the problems of it bigger. I am trying to reduce the things this class does.

If we just add the possibility to add text templates "on click" we need another solution.

- no custom event, more something like INSERT_TEXT_TEMPLATE
- no involvement of ilPageObjectGUI if possible. I would really like to refactor the location of the menu related functions somehow.
- rather than getCustomFunctions in ilPageObjectGUI, we should use the PageConfig classes. They are meant to configure the behaviour for different contexts.

So before extending this some refactoring would be obligatory. Main goal is to reduce the ilPageObjectGUI code.

kindly,
@alex40724

iszmais:
@alex40724 Thank you for your feedback!

I added your smaller requested changes. However, I agree that this should be implemented after some refactoring, therefore, I will postpone the final refurbishing after this is done.

For the conceptional part:
While this PR also adds the functionality to insert text templates (which I renamed from custom), this is not the only functionality the custom additions could provide. They have full access to all basic functionalities of the Editor and can make use of them. E.g. they could also add predefined links for their context, for example.
Therefore, I would still see this at a general entrypoint and even moved it to affect the whole menu if needed, so custom PageObjects may provide other sections of tools as well. Besides that, I totally agree that this should be done in the config more than the Object itself. Therefore, I reimplement a stump in the config which I will adapt after the refactoring as soon as the requirements on it are cleared.

Greetings,
@iszmais

alex40724:
@iszmais The $menu array is not a proper interface that I will expose this way. I would like to use a better (internal) structure for this in the future. I thought more of a function getTextTemplates which gets all the text templates as key value pair. The key concept of the interface are the text templates, not the menu itself.

iszmais:
Due to the changes made [here](https://github.com/ILIAS-eLearning/ILIAS/commit/365d0613a368abfb2843ccd23f13232809a7ca30) this PR is no longer needed.

Thanks @alex40724 for taking care of this!

PR #8436, opened 2024-11-14
UI/Legacy: factory->legacy() will return Legacy\Factory, not Component

nhaagen:
ILIAS\UI\Factory::legacy used to build the Legacy component directly.
This proposes a [Factory for legacy components](https://github.com/ILIAS-eLearning/ILIAS/pull/8436/files#diff-fffae9cf25b57b2a183fcfe906fdc503176d04a58f3e2c4935ef79f3b8fd90b9R728-R741) to open up Legacy for further components.

When working on the [SequenceNavigation](https://github.com/ILIAS-eLearning/ILIAS/pull/8393), we had the need to define a depency for the actual component (the navigation) without previously defining all the inner parts of the dependency (in this case, the Segment).
Right now, there are still a lot of proper components missing, especially for the "inner parts", the content. 
Often a simple legacy content is OK, too, but I consider it quite helpful to mark interim compositions with interfaces as not  to use "Legacy" as type in the new code. I bet, I'm not all alone with this.

Due to the UI-crawlers I also had to rename Legacy\Legacy to Legacy\LegacyContent. 
I tried my best to change all the relevant occurences along, that's why this PR is so huge.

PR #8857, opened 2025-01-16
UI/Legacy: legacy() returns factory, more fixing after merge

nhaagen:
following the merge of https://github.com/ILIAS-eLearning/ILIAS/pull/8436, there are several occurencs of legacy w/o calling content.


thibsy (approved):
Thx @nhaagen a lot for following up on this. There really are many legacy usages, unfortunately :(.

PR #8860, opened 2025-01-16
T&A 43537: Fixes permission error when editing a question in page editor

matheuszych:
When clicking on a question in the page editor the user gets a "Permission denied" despite having correct permissions.

I believe this is a simple routing error, that can be fixed by removing a GUI class from the redirect chain.

[Mantis: 43537](https://mantis.ilias.de/view.php?id=43537)


kergomard:
Thank you very much @matheuszych ! I don't like this solution much, because I think we should have clear paths through our code (especially as our RBAC is what it is), but for now, this is good enough. Merged and picked to trunk.

PR #8861, opened 2025-01-16
T&A 43267: Fixes being able to save a question with no positive point answers (Matching question)

matheuszych:
When editing a Matching question and setting all matches to give negative (incl. 0) points, the question get successfully saved but when looking in the question overview it says that the question is incomplete.

There error seems to be affected by the way how it is determined, whether the question has enough positive points assigned.

By checking what the highest point value is, you could determine if the question is solveable and if correctly solved even gives any points. Correctly solved questions should always give more than 0 points.

[Mantis: 43267](https://mantis.ilias.de/view.php?id=43267)

### DIFF
--- a/components/ILIAS/TestQuestionPool/classes/forms/class.ilAssMatchingPairCorrectionsInputGUI.php
+++ b/components/ILIAS/TestQuestionPool/classes/forms/class.ilAssMatchingPairCorrectionsInputGUI.php
@@ -54,8 +54,7 @@ class ilAssMatchingPairCorrectionsInputGUI extends ilMatchingPairWizardInputGUI
             return false;
         }
 
-        $sum = array_sum(array_filter($result, fn($points) => $points > 0));
-        if ($sum <= 0) {
+        if (max($result) <= 0) {
             $this->setAlert($this->lng->txt('enter_enough_positive_points'));
             return false;
         }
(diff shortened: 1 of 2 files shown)


kergomard:
Thank you very much @matheuszych for this PR!

As we are already touching this: I think we should also make sure that the total points are positive. Right now we are only checking that at least one of the points is positive. Do I miss something?

Thanks again and best,
@kergomard

matheuszych:
Hello @kergomard,
from what I understand. We should check that if the user answers fully correctly, that means not only matches all correct pairs but does also not match any incorrect ones. Leaving two nodes unconnected, might aswell be the correct answer. The author needs to decide, whether connecting two nodes is valued more than leaving them unconnected.

Let's say there are three connections. Only one would be correct giving you a point. The author might want to deduct points in case the user decides to do more connections. For example deduct a point for every additional connection. So if the user connected all three, they would get a overall of negative one point. This might be expected behaviour by the test author.

I hope this makes somewhat sense.

Best regards
Matheus

kergomard (approved):
Thank you for the explanation @matheuszych ! For me this is one more sign that the `ilTestLegacyFormsHelper` is actually conceptually wrong: You are right in case of the Matching-Question, you would be wrong e.g. in case of the Multiple-Choice-Question, where the total points also need to be positive, but we check both with one single function `checkPointsEnoughPositive`, thus treating different things as the same. We need to get rid of it and we need to make sure that we do never treat different things as if they were the same, ...but this is for ILIAS 11. For now: I merge this and pick to trunk.

PR #8863, opened 2025-01-16
T&A 43576: Allows sorting of a tests participant list by matriculation number and test id

matheuszych:
I believe this will allow the sorting by matriculation number and test id in the test participant list view.
[Mantis: 43576](https://mantis.ilias.de/view.php?id=43576)

### DIFF
--- a/components/ILIAS/Test/src/Participants/ParticipantTable.php
+++ b/components/ILIAS/Test/src/Participants/ParticipantTable.php
@@ -237,7 +237,7 @@ class ParticipantTable implements DataRetrieval
             'id_of_attempt' => static fn(
                 Participant $a,
                 Participant $b
-            ) => $a->getMatriculation() <=> $b->getMatriculation()
+            ) => $a->getAttemptOverviewInformation()?->getExamId() <=> $b->getAttemptOverviewInformation()?->getExamId()
         ];
     }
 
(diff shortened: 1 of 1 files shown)


kergomard:

The sorting by the attempt_id will not work as you are comparing the matriculation again [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8863/files#diff-a60cd7209785d84379d439efa3a8d2f51434aa5931f970ffcf630e2c54e64dbeR240). ...but to be honest: Do we really want to allow that? I don't see any use of it what so ever. What do you think @dsstrassner ? ...the student id ("matriculation"), I can understand.

Best,
@kergomard

kergomard (changes requested):

The sorting by the attempt_id will not work as you are comparing the matriculation again [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8863/files#diff-a60cd7209785d84379d439efa3a8d2f51434aa5931f970ffcf630e2c54e64dbeR240). ...but to be honest: Do we really want to allow that? I don't see any use of it what so ever. What do you think @dsstrassner ? ...the student id ("matriculation"), I can understand.

Best,
@kergomard

dsstrassner:
(Ich kommentiere mal auf deutsch: es wäre vollständig, wenn wir auch nach Test-ID sortieren könnten. Ich weiß das Institutionen sich das Schema der ID anpassen haben lassen, für die eine Sortierung noch sinnvoller ist, als ggf. test10 z.B.)

kergomard:
Thanks for the feedback from your holidays @dsstrassner ! Ok, then let's just fix the sorting and this can be merged.

matheuszych:
@kergomard All done :)

kergomard:
Thank you very much for the fast reaction @matheuszych ! Merged and picked to trunk.

PR #8864, opened 2025-01-16
DataCollection: removed page objects for `createviewtable` and `editviewtable`

iszmais:
There seem to be page objects for every tableview within a datacollection for the tabs create and edit.
Since those are never shown anywhere and are not editable I removed them.


iszmais:
@oliversamoila I might be wrong here (if so, pls call me out on that), but as far as I know there is no way to edit and display a customized CO Page for each view that is displayed when an entry is created/edited.
 
I'm unsure since this seems to be a pretty big change back then. Otherwise this is mostly copied from the detailed view and partially incomplete with wrongly labeled translations, errors, etc.

If my assumption is correct, I would propose to remove this.

Greetings,
@iszmais 

(Also I could not refer to its creator tt@studer-raimann.ch since I could not find him on GitHub anymore. So if you know anyone related to this, we might ask about the idea behind it)

oliversamoila:
Sorry, but I'm confirming this in German so that it's as clear as it needs to be.

Es existiert maximal eine Seitengestaltung zur Aufbereitung/Präsentation von Daten einer Tabelle je Ansicht. Eine Tabelle kann n Ansichten haben. Eine Ansicht einer Tabelle hat entweder keine Einzelansicht (engl. 'Single') oder aber genau eine Einzelansicht. Eine Einzelansicht dient als Vorlage für die Anzeige von Daten, die über Platzhalter eingebunden werden. Eine Einzelansicht ist daher nicht weiter individualisierbar.

I hope this helps. It is not necessary to support an advanced option to design several single views per view of a table.

Greetings,
@oliversamoila

iszmais:
@oliversamoila thanks for your confirmation.

I will remove this then. We might want to restore it if there is a usecase thats gone missing with the removal, but i assume there is not.

Greetings,
@iszmais

PR #8868, opened 2025-01-17
T&A 43575: Removes additional time, when removing participant data

matheuszych:
When deleting the participant data for a speicifc test, the additinal time a user has is not deleted. This causes the user to retain that additional time when doing the test again.

Whenever the delete participant data method is called, it is also checked whether the user has any entries in the additional time table. If that is the case, they also get delted.

INFO: If you give a anonymous user additional time, all anonymous users in that test get the extra time. To keep in mind. If you remove a anonymous users data from the test, the extra time gets deleted for all anonymous users, since there is a single entry for all of them (user 13).

Also, the property `anonymous_id` in `Participant` probably should be a string instead of int. Otherwise I got an error when trying to start the test as an anonymous user.

[Mantis: 43575](https://mantis.ilias.de/view.php?id=43575)

### DIFF
--- a/components/ILIAS/Test/classes/class.ilObjTest.php
+++ b/components/ILIAS/Test/classes/class.ilObjTest.php
@@ -1178,7 +1178,7 @@ class ilObjTest extends ilObject
                 static fn($active_id) => $participant_data->getUserIdByActiveId($active_id),
                 $participant_data->getAnonymousActiveIds(),
             );
-            $this->removeAdditionalTimesByUserIds($user_ids);
+            $this->participant_repository->removeExtraTimeByUserId($this->getTestId(), $user_ids);
         }
 
         if ($participant_data->getUserIds() !== []) {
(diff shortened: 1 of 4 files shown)


kergomard:
Thank you very much for this improvement @matheuszych !

I would have to change requests:
- Could you please move [`ilObjTest::removeAdditionalTimeByUserId`](https://github.com/ILIAS-eLearning/ILIAS/pull/8868/files#diff-377cf2c64a9111fe133f461ef77079e17cd9132640dc33f5c5e0ef927c46de64R1220-R1229) to the `ParticipantRepository` and rename it to `removeExtraTimeByUserId`. This makes things more homogeneous.
- Adding time to an anonymous user is completely bonkers. Please remove this option from the table with a corresponding check in the action.

Thanks again and best,
@kergomard

matheuszych:
Hello @kergomard,
thank you for your feedback. I moved and renamed the method. I also removed the option to add/remove extra time for anonymous users. I left in the part where the extra time gets deleted for anonymous users, as a sort of passive database migration, in case someone has already given anonymous users extra time.

If you want me to adjust anything please let me know :)

Best regards
Matheus

kergomard:
Thank you very much for the quick update @matheuszych ! I merge this and cherry-pick to trunk.

One little thing that I will fix myself quickly: Please do not introduce use-statements for classes in the global name space like [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8868/files/5714db5ea6ffc37cc196f69aaace568843ef661c..a807ab2ce8c6f059b23b3852f6a049f782e7282a#diff-a319d365487348c278d69311e874d51d814fd0ddb4a0fdf3bb6354c98a4bc498R23).

PR #8871, opened 2025-01-20
Type error fixed

qualitus-dahme:
https://support.iliasnet.de/view.php?id=9387
Type error fixed, returning string only


qualitus-dahme:
Closed, wrong merge branches

PR #8873, opened 2025-01-20
Component/Plugins: 43587, addendum - use absolute path for PLUGIN_BASE_PATH

nhaagen:
following https://github.com/ILIAS-eLearning/ILIAS/pull/8837, https://mantis.ilias.de/view.php?id=43587
There is a difference in using path from GUI and cli; using absolute path resolves this.


klees:
Picked to trunk.

PR #8878, opened 2025-01-21
T&A 42475: Fixes performance issues in question table

matheuszych:
[Mantis: 42475](https://mantis.ilias.de/view.php?id=42475)

There are a couple of performance issues regarding the "Add from Pool" & "Add from other Test" tables. With many records the loading times become very noticeable.

I analyzed the underlying code and believe to have found the possible causes. One problem seems to be that the ordering and limiting of the records happens in PHP instead of in the database. This means that no matter how many records should be displayed to the user (partially filtered) all records get returned from the database and get ordered and limited in PHP which could get quite slow. Another problem is that for every record returned more queries are sent to the database to check whether the question has feedback, hints or taxonomies. This can cause an exponential growth of queries. Also the main query is executed multiple times throughout the execution process. For example once to count all found records (pagination) and once with all found records limited to the page size. Overall this approach results in an overhead of queries.

The idea was to move the filtering, ordering and limiting components directly to the database. The database is much more efficient at doing these tasks. This also simplifies the handling on the PHP side, so the found records can be (more or less) directly passed to the user. Due to the higher complexity of some of the relations between tables, the current query increased in size and so did the overall performance.

> To put the code up to a test I duplicated a question inside a test 4600+ times. In another test I went to the "Add from other Test" table and tested the loading times.
> 
> Without the code it takes about 6.2 sec with no filtering, no sorting and unlimited elements per page.
> With the code it takes about 5 sec with no filtering, no sorting and unlimited elements per page.
> Average of 24% performance increase.
> 
> Without the code it takes about 6.8 sec with some filtering, sorting and unlimited elements per page.
> With the code it takes about 5.8 sec with some filtering, sorting and unlimited elements per page.
> Average of 17% performance increase.
> 
> Please keep in mind that this test is meant to show the extreme test case. The regular user will probably limit the amount of elements significantly. This would bring the loading times well below 1 sec.

If this general idea gets welcomed, may I propose going forward to try to put the filtering, ordering and limiting onto the database side? This should increase performance and lower the code complexity.  Although I do not know, if this would overwhelm the databases for bigger ILIAS Installations, in the long run.

I would be glad to know your opinions on this matter.

A not directly related issue: 
[Mantis: 43093](https://mantis.ilias.de/view.php?id=43093)
As a quick note. If ordering logic would to be moved to the database, the Collation would automatically perform the correct Ordering of German (An some others) special characters.

### DIFF
--- a/components/ILIAS/Test/classes/Tables/class.ilTestQuestionBrowserTableGUI.php
+++ b/components/ILIAS/Test/classes/Tables/class.ilTestQuestionBrowserTableGUI.php
@@ -126,6 +126,9 @@ class ilTestQuestionBrowserTableGUI
 
     private function getQuestionsBrowserTable(string $parent_title = ''): QuestionsBrowserTable
     {
+        $question_list = new ilAssQuestionList($this->db, $this->lng, $this->refinery, $this->component_repository);
+        $question_list = $this->addModeParametersToQuestionList($question_list);
+
         return new QuestionsBrowserTable(
             (string) $this->test_obj->getId(),
             $this->current_user,
@@ -134,10 +137,9 @@ class ilTestQuestionBrowserTableGUI
             $this->lng,
             $this->ctrl,
             $this->data_factory,
-            new ilAssQuestionList($this->db, $this->lng, $this->refinery, $this->component_repository),
+            $question_list,
             $this->test_obj,
             $this->tree,
-            $this->testrequest,
             $this->taxonomy,
             $parent_title
         );
@@ -160,8 +162,11 @@ class ilTestQuestionBrowserTableGUI
         }
 
         if (in_array('ALL_OBJECTS', $selected_array, true)) {
-            $filters = $this->ui_service->filter()->getData($this->getQuestionsBrowserFilterComponent()) ?? [];
-            $selected_array = array_keys($this->getQuestionsBrowserTable()->loadRecords($filters));
+            $selected_array = array_keys(
+                $this->getQuestionsBrowserTable()->loadRecords(
+                    $this->ui_service->filter()->getData($this->getQuestionsBrowserFilterComponent()) ?? []
+                )
+            );
         }
 
         array_map(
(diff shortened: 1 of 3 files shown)


kergomard (changes requested):

Yes, we would like to have all ordering and limiting done by the database. That is where we are heading, we are just not there yet.

I like this very much, ...although it is quite a bit of juggling with the queries.

I have a few things, ...some of them really small. Please make the following changes (as always: unless I missed something):
- Remove the `ilAssQuestionList::getOrder()` and `ilAssQuestionList::getRange()` and access the properties directly. We don't need them to be exposed publicly as far as I can see.
- As we are looking at this more closely: Please check all other public getters in `ilAssQuestionList`, if they are actually needed and remove all that are not. Please make sure all setters are typed.
- Please rename [`$CONDITIONS`](https://github.com/ILIAS-eLearning/ILIAS/pull/8878/files#diff-8348fef9ba500f473b75a6576bf8310d71df03acedaa247cc62441843c8e4b66R533) to lower case.
- Please rename [`$selectFields`](https://github.com/ILIAS-eLearning/ILIAS/pull/8878/files#diff-8348fef9ba500f473b75a6576bf8310d71df03acedaa247cc62441843c8e4b66R584) to snake-case and rename subquery-functions to either all be `SubQuery` or `Subquery`, but not a mix.
- Please remove `QuestionBrowserTable::testrequest` and move `QuestionBrowserTable::setModeParametersToQuestionList(): void` to `ilTestQuestionBrowserTableGUI::addModeParametersToQuestionList(ilAssQuestionList $question_list): ilAssQuestionList`. 
- This is a nitpick: Could you please again remove the intermediate variable you introduced [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8878/files#diff-b52fc96f0f686417990aae433e2b0034f14037de1ad0bc79c0738a5087201dd9R161). I know, this is a matter of taste, but in my opinion it makes reading harder not easier.
- Could you please add curly braces around variable names in strings (e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8878/files#diff-8348fef9ba500f473b75a6576bf8310d71df03acedaa247cc62441843c8e4b66R651).
- Could we get rid of `AND $page_object_table.lang = '-'` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8878/files#diff-8348fef9ba500f473b75a6576bf8310d71df03acedaa247cc62441843c8e4b66R605)? I don't think this check is needed and the '-' in lang-column looks really random to me, as if it could change at any moment.
- [This here](https://github.com/ILIAS-eLearning/ILIAS/pull/8878/files#diff-8348fef9ba500f473b75a6576bf8310d71df03acedaa247cc62441843c8e4b66R664-R667) looks like we have absolutely no idea what we have in our hands and then we just cast things around and test randomly. Please make sure we do not need the string cast and we do not need to check for empty, but can check for something more specific.
- Please consolidate `ilAssQuestionList::getTaxonomyFilterExpression()` and `ilAssQuestionList::getTaxonomyFilterExpressions()` into one function. We either have a singular or a plural, but not both.
- Could yo please change [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/8878/files#diff-8348fef9ba500f473b75a6576bf8310d71df03acedaa247cc62441843c8e4b66R717-R718) to a comparison operation to get the boolean.
- Could you please remove the brackets [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8878/files#diff-8348fef9ba500f473b75a6576bf8310d71df03acedaa247cc62441843c8e4b66R722-R723) again. I know, everybody learns that this is a good idea, I believe it makes life harder, because a bracket tells me: "This is not in the normal order, so be careful", ...but then it still is.
- We do not need the `(int) casts [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8878/files#diff-8348fef9ba500f473b75a6576bf8310d71df03acedaa247cc62441843c8e4b66R659), right? This would make things easier to read. You could then even get everything inside a string.

Thanks a bunch again, I'm looking very much forward to this!

Best,
@kergomard

matheuszych:
Hello @kergomard,
thank You for your feedback. I believe to have resolved all issues you mentioned. In the `ilAssQuestionList` class I changed a couple of other things as well, since I already have touched most of it. I also found a slight bug when trying to filter by `feedback` or `hints` and fixed it.

Best regards
Matheus

kergomard:
Hi @matheuszych 

Thank you very much for the update!

I've one question: Why did you move the properties of the `QuestionBrowserTable` back to being `protected` and not `private`? If there is no absolute necessity: Could you please move them back to being private, I don't want nobody to get the idea that this table could be reused in any way never ever.

And, you are not going to like me for this, as you followed my wish to the letter, I just didn't formulate my wish precisely enough, sorry: I still don't like the function [`ilAssQuestionList::getTaxonomyFilterExpressions()`](https://github.com/ILIAS-eLearning/ILIAS/pull/8878/files#diff-8348fef9ba500f473b75a6576bf8310d71df03acedaa247cc62441843c8e4b66R248). The naming is correct now, ...but the function just got too long. I think we should split out the lines 251 to 287 into a function `getFilterByAssignedTaxonomyIdsExpression()` called from within `getTaxonomyFilterExpressions()`. We could then also revert back to early returns and we could get rid of the `$forceBypass`-variable, if I'm not completely dreaming right now this could be written as `if ($taxNodes === []) { continue; }`, right?

@kergomard

matheuszych:
Hello @kergomard
i believe the properties were made protected by accident when merging. I reverted that. The method `ilAssQuestionList::getTaxonomyFilterExpressions()` was indeed a bit to long. It is now split. From what it looks like `$forceBypass` can be replaced.

Everything should now be correct.

Best regards
Matheus

kergomard:
Thank you very much @matheuszych ! ...might it be though that something went wrong when you pushed your changes? ...I don't see them.

matheuszych:
@kergomard
Pushing the changes might help 😄

kergomard:
Thank you very much! Merged and picked to trunk.
I moved the check for `$taxNodes` to before the `foreach` and changed a few variable-names.

PR #8880, opened 2025-01-22
Setup: I want to allow plugins to have an Setup\Agent in src/ directory

fwolf-ilias:
I'm not quite sure why this is not allowed due to the ignore list in Setup\AbstractOfFinder.  

We want to gradually adapt our plugin to the ILIAS core and therefore move all classes that are in namespaces to the src/ folder. The SetupAgent is not found there due to an ignored path pattern.


fwolf-ilias:
For clearification: I added @chfsx as Reviewer because he added the ignore pattern in the first place and maybe add some reasoning.

chfsx:
I'm honestly no longer sure what the reason for this was... But from my point of view, there's no reason to keep it.

klees:
Picked to trunk.

PR #8881, opened 2025-01-22

iszmais:
Resolves https://mantis.ilias.de/view.php?id=38015
Resolves https://mantis.ilias.de/view.php?id=38017

### Initial Problem

Values of a multi input fields are transmitted to the Form/JS but string concatination and are therefore susceptible for a lot of context related issues of the form value.
This causes all input field of that kind to misfunction on specific input value and makes this feature unusable for some input types (event if it is enabled for those atm).

### Solution

This PR encodes the values to a context unrelated format (base64) before transmitting them to the JS where they are decoded and proper handled. Through that the inputs are preserved similar in every case, not matter of their semantically value.
Further this increases the support of that feature to all input types as long as their value is a multi scalar type and therefore makes it usable not only for all current types where its enabled, but for upcoming and custom types of any kind, too.

### Conflicts

- Performance: The transformation causes a slight increase in workload and space. However, these changes are negligible in relation to the overall performance. A form where this increase would be noticeable would already struggle with displaying such large information via a user interface.
- Current persisted values: The internal persistence of input values is not affected by those changes therefore all current occurrences will stay the same. Also this feature has only 1 current usage wich isn't scalar (ilOrgUnitPositionFormGUI =>  ilOrgUnitGenericMultiInputGUI)

If there is any struggle or concern with the proposed changes please leave me a feedback.

Greetings,
@iszmais 

(Ported from ILIAS7: https://github.com/ILIAS-eLearning/ILIAS/pull/6538)

### DIFF
--- a/Services/Form/js/ServiceFormMulti.js
+++ b/Services/Form/js/ServiceFormMulti.js
@@ -1,6 +1,19 @@
 /**
- * Handle multi values for select and text input fields
- */
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
 var ilMultiFormValues = {
 	
 	// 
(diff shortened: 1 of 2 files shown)


fneumann:
I am offering to do a review and would start in two weeks. If anyone in the community outside Databay AG would like to take over the review, I would be happy to hand it over.

Services/Form/js/ServiceFormMulti.js
```diff
@@ -260,14 +269,20 @@ var ilMultiFormValues = {
...
+			if(value !== undefined && value != '') {
+				$(this).find('option[value="' + il.Form.escapeSelector(String(value)) + '"]').attr('selected', true);
+				if (this.tagName === 'INPUT') {
+					$(this).attr('value', value);
+				}
+			} else {
+				if (this.tagName === 'INPUT') {
+					$(this).val('');
 		}
+			}
```
lscharmer, line 284: Please indent this correctly.

lscharmer (changes requested):

lscharmer (approved):

fneumann (approved):
Checked code and tested interactively. The former problem with special character '~' is solved.

PR #8882, opened 2025-01-22
MANTIS 43667: Change frm_posts is_author_moderator to non nullable (bool)

mBeym:
Fixes: https://mantis.ilias.de/view.php?id=43667

If accepted should be picked into trunk


mjansenDatabay:
Thanks for your work, Marvin. I decided against the schema change an only modofied the XSD schema file for ILIAS 9 - trunk.

PR #8883, opened 2025-01-22
[FIX] CI: remove composer-autoload import from PHP-CS-Fixer config.

thibsy:


klees (approved):
I just confirmed locally that this should be fine. Maybe something in PHP-CS-Fixer changed, I'm pretty much confident that this was required once. Might check later...

PR #8884, opened 2025-01-22
[Improvement] Check skipped major objective early on before updating the database

fwolf-ilias:
Before executing update steps, check if a major version was skipped. Terminate early on so the database is not updated and the user can update to the next viable version.


klees:
Thanks @fwolf-ilias, extremely plausible =)

klees:
Picked to trunk.

PR #8888, opened 2025-01-23
UI Components #0043810 Adjustment Text Expected Output

yvseiler:
Text Expected Results Example 2 extended to make it easier to understand how to test.
https://mantis.ilias.de/view.php?id=43810


Amstutz:
Thx a lot for the improvement.

Please further enhance by:

- [x] Specify Screen with behaviour: The selected version of the image depends on the space available, meaning on very large screens, a large image version is almost always selected, even if there is only 10% of the space available.

thx.

yvseiler:
Thanks @Amstutz for your feedback. I changed the description and added more information about screen behaviour. Have another look at it, thanks a lot.

PR #8889, opened 2025-01-23
10/UI/form/section headings html tags

catenglaender:
https://mantis.ilias.de/view.php?id=42226

# Issue

Sections no longer have heading tags.

image

# Changes

Sections now assess their nesting level and html tags are rendered accordingly. ~~I also strengthened the visual difference of differently deep nesting levels.~~

UPDATE: @yvseiler and I decided against visual indents for now. The headline size already does a sufficient job to convey nesting and we don't have deeply nested sections yet. We might revisit this once we have the first use cases of triple+ nested sections (if ever)

UPDATED IMAGE:
image

# ~~Advice needed~~

~~Unit tests can currently only handle the first level expecting a h2. I will discuss a better solution with my colleagues.~~

Unit Tests now account for both the general form tests and a specific triple nested scenario.

### DIFF
--- a/components/ILIAS/UI/src/Implementation/Component/Input/Field/Renderer.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Input/Field/Renderer.php
@@ -739,7 +739,18 @@ class Renderer extends AbstractComponentRenderer
     protected function renderSection(F\Section $section, RendererInterface $default_renderer): string
     {
         $inputs_html = $default_renderer->render($section->getInputs());
-        return $this->wrapInFormContext($section, $section->getLabel(), $inputs_html);
+
+        $headline_tpl = $this->getTemplate("tpl.headlines.html", true, true);
+        $headline_tpl->setVariable("HEADLINE", $section->getLabel());
+        $nesting_level = $section->getNestingLevel() + 2;
+        if ($nesting_level > 6) {
+            $nesting_level = 6;
+        };
+        $headline_tpl->setVariable("LEVEL", $nesting_level);
+
+        $headline_html = $headline_tpl->get();
+
+        return $this->wrapInFormContext($section, $headline_html, $inputs_html);
     }
 
     protected function renderUrlField(F\Url $component, RendererInterface $default_renderer): string
--- a/components/ILIAS/UI/src/Implementation/Component/Input/Field/Section.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Input/Field/Section.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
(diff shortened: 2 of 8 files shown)


klees (commented):
Hi @catenglaender,

LGTM in general, plain, simple and functional.

* [x] Question 1: Did you check with @yvseiler et al regarding the visuals (indentation etc.).
* [x] Question 2: How will this look on small screens?
* [x] Advice: Please check back with @nhaagen to look into unit tests.

Kind regards!

catenglaender:
I contacted @yvseiler and we will have a look at it

@nhaagen gave me guidance on how to do the unit tests. The general getFormWrappedHtml() will only be able to test forms with unnested sections for now. I added another Unit Test to test nested headline creation specifically.

UPDATED mobile screenshot with no indents
image

catenglaender:
@yvseiler and I decided against visual indents for now. The headline size already does a sufficient job to convey nesting and we don't have deeply nested sections yet. We might revisit this once we have the first use cases of triple+ nested sections (if ever)

Failing unit tests are not related to this PR.

With that, I think, the work here is done for now. Ready for review/merge :slightly_smiling_face:

klees:
Hi @thibsy and @Amstutz,

would you please look into this? Tthis is not the most elegant solution (since immutability is broken), but also the violation of the principle seems to be local. Some like the [catamorphism](https://github.com/ILIAS-eLearning/ILIAS/pull/8812) can produce a nicer solution, but then again we are looking to ship a release. So from my POV this solution is ok for the time being. 

Kind regards!

Amstutz:
Hi @catenglaender 

Thx a lot for this PR. We believe this to be ready to be merged, except for one minor detail:

- [x] Snake case: Please use snake_case for all variables e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8889/files#diff-e193eb78940e9c0dc5787b86593505e07ffac4f4242b9d1bee8b1fff730a0cfbR65).

And one thing you can change or explain why not:
- [x] Counting: You start $nesting_level with 2. That seems a bit odd. We understand, why this is done like so. But if done like that, better rename to heading_level. If kept as nesting level, start at 0 and translate in rendering (+2).

Thx a lot

catenglaender:
I made the changes :slightly_smiling_face: 
* variables use snake case (hope I got them all)
* nesting level starts with 0
* translation from nesting level to headline level happens in the renderer, so does the cut off at 6 levels deep.

Amstutz:
thx a lot!

klees:
@Amstutz Is this picked to trunk?

thibsy:
@klees it is, depending on where you click in the "referenced commit message" the proper commit is shown :)

klees:
image

This one! Was tricked by the commit message starting with `10/`.

PR #8892, opened 2025-01-23
Move soap internal config into ilias ini

shecken:
Our suggestion is to move the configuration for the internal SOAP communication to the ilias ini. This means that a complete initialization of ILIAS is no longer necessary.
In addition, we have changed the paths to absolute paths.


GitHamo (approved):
@shecken Thank you very much for your efforts. Very good work. Looks good to me.

Note: should be cherry-picked in `release_11`, `trunk`.

PR #8893, opened 2025-01-23
Fix undefined Variable

kergomard:
Hi @chfsx 

There is an undefined variable in `StreamDelivery`, the corresponding definition seems to have been removed by accident with [this commit](https://github.com/ILIAS-eLearning/ILIAS/commit/95a12d6c3f8ceb33e113d8a08c8da66e32863e9b).

The corresponding issue is here: https://mantis.ilias.de/view.php?id=43818

Best.
@kergomard


chfsx:
Thanks a lot!

PR #8310, opened 2024-11-03
KS: 41633: Video Player not working

alex40724:
See https://mantis.ilias.de/view.php?id=41633

Major changes:
- No mediaelment.js lib is used.
- Native video/audio tag and youtube/vimeo iframes are being used.
- Consumers need to provide correct embed src for vimeo and youtube.

I added this to the Jour Fix agenda, since this fix includes major changes that affect styling, accessibility and consumer code.

PR #8897, opened 2025-01-23
KS: 41633: Video Player not working (2nd pr)

alex40724:
This is a follow up of

https://github.com/ILIAS-eLearning/ILIAS/pull/8310

Please read the discussion and remarks in the old PR.

Additionally this PR contains a lot of color related changes. I am using the latest version of dart sass (1.83.4). It seems that older versions don't use the rgb() notation this way. Using expanded or prevision options did not change the output for me.

### DIFF
--- a/components/ILIAS/UI/src/Implementation/Component/Player/Renderer.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Player/Renderer.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 namespace ILIAS\UI\Implementation\Component\Player;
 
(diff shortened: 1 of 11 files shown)


klees:
@yvseiler /  @catenglaender / @Amstutz Any idea why the output of delos changed so much?

We really should look how to integrate that step into build.

catenglaender:
delos.css changed so much because the latest SASS version processes color operations differently.

The SASS version in our node_modules folder is now pinned. So if you use the SASS version that comes with our npm packages, the output will have a much smaller diff:

```bash
node_modules/.bin/sass templates/default/delos.scss templates/default/delos.css
```

Admittedly, the current READMEs and guidelines should probably state this a lot more clearly. :sweat_smile: Maybe composer could execute SASS to compile and then move the generated css to public/assets?

alex40724:
> Admittedly, the current READMEs and guidelines should probably state this a lot more clearly. 

It is not that its unclear in the guidelines, it just states explicitely to use the latest version (multiple times):

> When contributing style code to the ILIAS repository, you MUST first compile the Sass code to CSS using the latest released version of Dart Sass: https://github.com/sass/dart-sass/releases
...
> CSS is obtained by using the latest Dart Sass compiler on delos.scss, e.g. like so:

https://github.com/ILIAS-eLearning/ILIAS/blob/release_10/templates/Guidelines_SCSS-Coding.md#guidelines

alex40724:
@klees I recompiled using the older version.

klees:
@alex40724 Thanks!
@catenglaender Please adjust the documentation accordingly.

catenglaender:
Thanks @alex40724 for pointing out the incorrect docs. I've made some changes to the style code documentation: #8955

klees:
Picked manually after rebase and css-recompile.

klees:
Hi @alex40724,

could you be so kind a provide a version of this for trunk?

Thanks a lot!

PR #8898, opened 2025-01-23
TinyMCE: Fix license warning when using TinyMCE editor

mjansenDatabay:
This PR fixes a "TinyMCE" v7 warning which I noticed in the browser console.

> TinyMCE is running in evaluation mode. Provide a valid license key or add license_key: 'gpl' to the init config to agree to the open source license terms. Read more at https://www.tiny.cloud/license-key/


kergomard:
Thank you very much @mjansenDatabay ! I looked at the licensing and for me everything points to this being right. Failing tests are due to missing license headers in the files. Merged and picked to trunk.

PR #8900, opened 2025-01-24
COPage: 0042698, flagged drag-and-drop operation as move operation

klein-zqs:
Solution for 0042698: "Drag-and-drop is not a move operation"
https://mantis.ilias.de/view.php?id=42698


alex40724:
Merged to 10/trunk. Thanks for the PR!

PR #8901, opened 2025-01-24
[Dependencies] Fix "lock file does not contain a compatible set of packages"

fwolf-ilias:
Fix lock file does not contain a compatible set of packages regarding composer-plugin-api, simplesamlphp/composer-module-installer and simplesamlphp/simplesamlphp-assets-base

**Error message:**
```
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - simplesamlphp/composer-module-installer is locked to version v1.4.0 and an update of this package was not requested.
    - simplesamlphp/composer-module-installer v1.4.0 requires composer-plugin-api ^2.6 -> found composer-plugin-api[2.1.0] but it does not match the constraint.
  Problem 2
    - simplesamlphp/composer-module-installer v1.4.0 requires composer-plugin-api ^2.6 -> found composer-plugin-api[2.1.0] but it does not match the constraint.
    - simplesamlphp/simplesamlphp-assets-base v2.0.9 requires simplesamlphp/composer-module-installer ^1.3.2 -> satisfiable by simplesamlphp/composer-module-installer[v1.4.0].
    - simplesamlphp/simplesamlphp-assets-base is locked to version v2.0.9 and an update of this package was not requested.
```

Fixed by `php8.1 composer update`

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -213,16 +213,16 @@
         },
         {
             "name": "composer/composer",
-            "version": "2.8.4",
+            "version": "2.8.5",
             "source": {
                 "type": "git",
                 "url": "https://github.com/composer/composer.git",
-                "reference": "112e37d1dca22b3fdb81cf3524ab4994f47fdb8c"
+                "reference": "ae208dc1e182bd45d99fcecb956501da212454a1"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/composer/composer/zipball/112e37d1dca22b3fdb81cf3524ab4994f47fdb8c",
-                "reference": "112e37d1dca22b3fdb81cf3524ab4994f47fdb8c",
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you for the PR @fwolf-ilias ! I merge it.
If I'm getting this right, this is only a problem under php8.1 and from now on all updates of the dependencies have to be made on php8.1, right?

Best,
@kergomard

fwolf-ilias:
This was a problem for both php 8.1 and php 8.2. I don't know why this is suddenly going wrong. When the dependencies were updated after the last release, everything worked fine.

Nevertheless, after an update of the dependencies it works again. There was an downgrade of `simplesamlphp/composer-module-installer`, which seemed to remove the inconsistencies regarding `composer-plugin-api`. Perhaps there was an adjustment of the simplesamle dependencies in the background, if that is possible at all.

PR #8902, opened 2025-01-24
PRG: set repository selection to async in content automation

nhaagen:
https://mantis.ilias.de/view.php?id=43817


klees:
Picked to trunk.

PR #8906, opened 2025-01-24
Search/Init: Modify position of "MetaBar" items to prevent varying or…

mjansenDatabay:
…dering

See: https://mantis.ilias.de/view.php?id=43756

If approved, this has to be merged to `trunk`.


smeyer-ilias:
Thanks @mjansenDatabay for fixing this.

PR #8907, opened 2025-01-24
Bugfix define $filtered variable

padvincenzo:
Bug: `Whoops\Exception\ErrorException thrown with message "Undefined variable $filtered"`


klees:
Hi @padvincenzo,

is there a report for this problem in our [bugtracker](https://mantis.ilias.de)? If so: please link. If not: Please open one and link it here =) I will then set the label to "bugfix".

Having an issue in the bug tracker to point at makes it easier to understand the change and how it can be tested.

Thanks for helping!

PR #8908, opened 2025-01-27
ilServer: Fix custom fonts for PDF documents

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=43546


smeyer-ilias:
@mjansenDatabay Sorry for the late reply and thanks for the PR! In the current version (FOP 2.11), I wasn't able to get this PR working. I also consider the necessity of generating a separate fop.xml to include custom fonts too complicated. In #11293 there is an alternative proposal.

mjansenDatabay:
So let's close this one @smeyer-ilias , thanks for reviewing!

PR #8909, opened 2025-01-27
Mantis 38554: Forum: Get rid of ilForumThreadObjectTableGUI

mBeym:
https://mantis.ilias.de/view.php?id=38554


components/ILIAS/Forum/classes/Thread/ForumThreadTableSessionStorage.php
```diff
@@ -0,0 +1,176 @@
...
+use ILIAS\HTTP\Wrapper\WrapperFactory;
+use ILIAS\Refinery\Factory;
+use ilObjForum;
+use ilSession;
+use ThreadSortation;
+
+class ForumThreadTableSessionStorage
+{
+    public const KEY_THREAD_SORTATION = 'thread_sortation';
+    public const KEY_THREAD_PAGE = 'thread_page';
```
mjansenDatabay: With PHP 8.3 you can use typed constants.

components/ILIAS/Forum/classes/Thread/ForumThreadTableSessionStorage.php
```diff
@@ -0,0 +1,176 @@
...
+    ) {
+        $this->forum_ref_id = $forum_ref_id;
+        $this->is_moderator = $is_moderator;
+
+        global $DIC;
+        $this->http_wrapper = $http_wrapper ?? $DIC->http()->wrapper();
+        $this->refinery = $refinery ?? $DIC->refinery();
+    }
+
+    public function fetchData(ilForum $forum, ForumDto $topicData): array
```
mjansenDatabay: Can we introduce a simple object instead of the array? I suggest something like `ThreadsPage`?

We could then also get rid of the `if (array_key_exists('thread', $thread)) {` checks in class `ilObjForumGUI`.

components/ILIAS/Forum/classes/Thread/ForumThreadTableSessionStorage.php
```diff
@@ -0,0 +1,176 @@
...
+        );
+    }
+
+    private function getKeyValueFromSession(int $refId, string $key, mixed $default = null): mixed
+    {
+        $session_value = ilSession::get($this->buildSessionKey($refId, $key));
+        if ($session_value === null) {
+            $session_value = $default;
+        }
+        return $session_value;
```
mjansenDatabay, line 158:
```suggestion

        return $session_value;
```

components/ILIAS/Forum/classes/Thread/ForumThreadTableSessionStorage.php
```diff
@@ -0,0 +1,176 @@
...
+    public function getThreadPage(): int
+    {
+        $query_thread_page = $this->getKeyValueFromQuery(self::KEY_THREAD_PAGE, 0);
+
+        if ($query_thread_page !== 0) {
+            $this->setSessionKeyValue($this->forum_ref_id, self::KEY_THREAD_PAGE, $query_thread_page);
+            return $query_thread_page;
+        }
+
+        return $this->getKeyValueFromSession(
```
mjansenDatabay, line 133: I might be wrong, but if the queried page is `0`, why do you read the page from the session storage? `0` translates to "First Page", which is valid. I think we need a different default value when executing `$this->getKeyValueFromQuery(self::KEY_THREAD_PAGE, 0);`, e.g. `null` or `-1`.

components/ILIAS/Forum/classes/Thread/ForumThreadTableSessionStorage.php
```diff
@@ -0,0 +1,176 @@
...
+            return ThreadSortation::tryFrom(
+                $query_thread_sortation
+            );
+        }
+
+        $session_thread_sortation = $this->getKeyValueFromSession(
+            $this->forum_ref_id,
+            self::KEY_THREAD_SORTATION,
+            ThreadSortation::DEFAULT_SORTATION->value
+        );
```
mjansenDatabay, line 117: I might be wrong, but if the queried sortation is `4` ("Last Posting DESC"), why do you read the page from the session storage? I think we need a different default value when executing `$this->getKeyValueFromQuery(self::KEY_THREAD_SORTATION, ThreadSortation::DEFAULT_SORTATION->value);`, e.g. `null` or `-1`.

mjansenDatabay (changes requested):

components/ILIAS/Forum/classes/Thread/ForumThreadTableSessionStorage.php
```diff
@@ -0,0 +1,171 @@
...
+    private WrapperFactory $http_wrapper;
+    private Factory $refinery;
+
+    public function __construct(
+        private readonly int  $forum_ref_id,
+        private readonly bool $is_moderator,
+        ?WrapperFactory       $http_wrapper = null,
+        ?Factory              $refinery = null
+    )
+    {
```
mjansenDatabay:
```suggestion
    ) {
```

PR #8910, opened 2025-01-27
Object: Respect item permissions when changing availability

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=43859


kergomard:
Thank you very much @mjansenDatabay for the PR! I was already fixing this when I saw it and decided to finalize what I was doing. Fixed slightly differently with: https://github.com/ILIAS-eLearning/ILIAS/commit/0816f97255d84eba431bef83a9681b62d6c32974

PR #8912, opened 2025-01-28
SOAP: Fix user is logged out when copying courses (in fallback mode)

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=43882

If approved, this has to be picked to all maintained branches.


jeph864:

PR #8914, opened 2025-01-28
PRG: 43817, do not prepareOutput for async modal responses

nhaagen:
https://mantis.ilias.de/view.php?id=43817


klees:
Picked to trunk.

PR #8916, opened 2025-01-28
Fix Missing Flavour in Tile

kergomard:
Hi @chfsx 

This was a fix that you applied originally to the Tile-Property in Object itself, so I just ported it here without investigating further how this can possibly happen.

See: https://mantis.ilias.de/view.php?id=43885

Best,
@kergomard


chfsx:
Thanks a lot!

PR #8919, opened 2025-01-29
trunk ilSessionMaxIdlesSetObjective:0043901 avoid execution if no server is reachable on port 80

daniwe4:
Mantis: https://mantis.ilias.de/view.php?id=43901

### DIFF
--- a/components/ILIAS/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
+++ b/components/ILIAS/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
@@ -23,7 +23,7 @@ use ILIAS\Setup;
 class ilSessionMaxIdleIsSetObjective implements Setup\Objective
 {
     public function __construct(
-        protected ilAuthenticationSetupConfig $config
+        protected Setup\Config $config
     ) {
     }
 
@@ -71,14 +71,16 @@ class ilSessionMaxIdleIsSetObjective implements Setup\Objective
         $url = $ini->readVariable('server', 'http_path');
         $filename = uniqid((string) mt_rand(), true) . '.php';
         $url .= '/' . $filename;
-        $key = bin2hex(random_bytes(32));
-        $this->generateServerInfoFile($key, $filename);
+        $token = bin2hex(random_bytes(32));
+        $this->generateServerInfoFile($filename, $token);
 
         try {
+            $curl = null;
             if (ilCurlConnection::_isCurlExtensionLoaded()) {
-                $result = $this->getPHPIniValuesByCurl($settings, $key, $url);
+                $curl = $this->getCurlConnection($settings, $url, $token);
+                $result = $curl->exec();
             } else {
-                $result = $this->getPHPIniValuesByFileGetContents($key, $url);
+                $result = $this->getPHPIniValuesByFileGetContents($url, $token);
             }
         } catch (Throwable $e) {
             $io->inform(
@@ -92,6 +94,9 @@ class ilSessionMaxIdleIsSetObjective implements Setup\Objective
 
             return $environment;
         } finally {
+            if (!is_null($curl)) {
+                $curl->close();
+            }
             unlink("public/$filename");
         }
 
@@ -136,10 +141,43 @@ class ilSessionMaxIdleIsSetObjective implements Setup\Objective
 
     public function isApplicable(Setup\Environment $environment): bool
     {
+        $factory = $environment->getResource(Setup\Environment::RESOURCE_SETTINGS_FACTORY);
+        /** @var ilSetting $settings */
+        $settings = $factory->settingsFor('common');
+        /** @var ilIniFile $ini */
+        $ini = $environment->getResource(Setup\Environment::RESOURCE_ILIAS_INI);
+        /** @var Setup\CLI\IOWrapper $io */
+        $io = $environment->getResource(Setup\Environment::RESOURCE_ADMIN_INTERACTION);
+
+        $url = $ini->readVariable('server', 'http_path');
+
+        if (ilCurlConnection::_isCurlExtensionLoaded()) {
+            try {
+                $curl = $this->getCurlConnection($settings, $url);
+                $curl->exec();
+                $result = $curl->getInfo(CURLINFO_HTTP_CODE);
+                if ($result !== 200) {
+                    throw new \Exception();
+                }
+            } catch (\Exception $e) {
+                $this->infoNoConnection($io);
+                return false;
+            } finally {
+                $curl->close();
+            }
+        } else {
+            try {
+                $this->getPHPIniValuesByFileGetContents($url);
+            } catch (Exception $e) {
+                $this->infoNoConnection($io);
+                return false;
+            }
+        }
+
         return true;
     }
 
-    private function generateServerInfoFile(string $key, string $filename): void
+    private function generateServerInfoFile(string $filename, string $token): void
     {
         $content = <<<TEXT
 <?php
@@ -147,7 +185,7 @@ if (!isset(\$_GET['token'])) {
     return "";
 }
 
-if (\$_GET['token'] !== "$key") {
+if (\$_GET['token'] !== "$token") {
     return "";
 }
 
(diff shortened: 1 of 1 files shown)


components/ILIAS/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
```diff
@@ -136,7 +136,12 @@ public function achieve(Setup\Environment $environment): Setup\Environment
 
     public function isApplicable(Setup\Environment $environment): bool
     {
-        return true;
+        $conn = @fsockopen("localhost", 80);
```
mjansenDatabay: Hi @daniwe4 ,

thanks for your PR.

I personally don't like the `@` operator.

I suggest temporarily modifying the error handler and use `try/catch/finnaly` blocks and reset the error handler in the `finally` block.

```php
        set_error_handler(static function (int $severity, string $message, string $file, int $line): never {
            throw new ErrorException($message, $severity, $severity, $file, $line);
        });

        try {
// ...
        } catch (Exception $e) {
// ...
        } finally {
            restore_error_handler();
        }
```

Best regards,
Michael

mjansenDatabay (changes requested):

components/ILIAS/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
daniwe4: Thank you @mjansenDatabay ,

it's done.

components/ILIAS/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
```diff
@@ -136,7 +136,21 @@ public function achieve(Setup\Environment $environment): Setup\Environment
 
     public function isApplicable(Setup\Environment $environment): bool
     {
-        return true;
+        set_error_handler(static function (int $severity, string $message, string $file, int $line): never {
+            throw new ErrorException($message, $severity, $severity, $file, $line);
+        });
+
+        try {
+            $conn = fsockopen("localhost", 80);
```
mjansenDatabay: Thanks @daniwe4 for the change.

One more remark: You suggest to create a socket connection to "localhost" here, but in `achieve`, we read the URL (with the host we want to connect to) from the INI file. Should we extract the hostname from the INI value?

daniwe4: I just tried the URL. Unfortunately that doesn't seem to work. Do you know how that works? 
Otherwise it's just a check whether a web server is running while the setup is being carried out. We could still pull http or https from the URL and then adjust the port.

daniwe4: Hello @mjansenDatabay ,

I revised everything again.

Greetings Daniel

mjansenDatabay (approved):

daniwe4:
Hi, we noticed that in some cases curl is present but still throws an error when executed. That's why I'm extending the code again.

mjansenDatabay (approved):

mjansenDatabay:
> Hi, we noticed that in some cases curl is present but still throws an error when executed. That's why I'm extending the code again.

Thanks @daniwe4 

@thojou Could you have a look here, please?

thojou (changes requested):
Hey @daniwe4,  

Thanks for the PR! I appreciate your effort in improving this.  

I have some reservations about changing the `isApplicable` method. If it returns `false`, the client ini variable `session.expire` won’t be set during configuration.

That said, I do see that `isApplicable` introduces some redundant code, which is already present in `getPHPIniValuesByCurl` and `getPHPIniValuesByFileGetContents`.  

To streamline this, I’d kindly suggest the following changes:  

* Keep `isApplicable` always returning `true`.  
* Move the `set_error_handling` into `getPHPIniValuesByFileGetContents`.  
* Wrap the code inside `getPHPIniValuesByCurl` with:  
  ```php
  try { ... } finally { $curl->close(); }
* Consider checking the HTTP status code inside getPHPIniValuesByCurl. If it's not OK, throw an exception so that the parent try-catch block can properly handle the failure.

With these changes, adding custom error handling to `getPHPIniValuesByFileGetContents` should help achieve the objective, since both `getPHPIniValuesByCurl` and `getPHPIniValuesByFileGetContents` are already wrapped in a try-catch block.  

Let me know what you think!
Best,
@thojou

daniwe4:
Hello @thojou,

Thank you very much for your review.
Very good points you raise there.

**Regarding the point 'isApplicable':**

I think that the function is there precisely to check whether the environment is suitable. If it is not, False must also be returned here.
In case of installation and return of false from 'isApplicable', the default value for 'session.expire' is taken from the 'client.master.ini'.
In case of an update and 'isApplicable' returns false, the old value is retained.
I've added outputs to explain this to the end user.

**Regarding the point 'redundant code':**

I completely agree with you and I have changed the code so that there should no longer be any redundant positions.


I've also put the HTTP status code check into the try catch block.

So hopefully you like these changes.

Best regards,

@daniwe4

components/ILIAS/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
```diff
@@ -134,20 +144,59 @@ public function achieve(Setup\Environment $environment): Setup\Environment
         return $environment;
     }
 
+    /**
+     * @throws ilCurlConnectionException
+     * @throws ErrorException
+     */
```
thojou:
```suggestion
```

components/ILIAS/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
```diff
@@ -134,20 +144,59 @@ public function achieve(Setup\Environment $environment): Setup\Environment
...
+        $ini = $environment->getResource(Setup\Environment::RESOURCE_ILIAS_INI);
+        /** @var Setup\CLI\IOWrapper $io */
+        $io = $environment->getResource(Setup\Environment::RESOURCE_ADMIN_INTERACTION);
+
+        $url = $ini->readVariable('server', 'http_path');
+
+        $curl = $this->getCurlConnection($settings, $url);
+        if (!is_null($curl)) {
+            try {
+                $curl->exec();
```
thojou:
```suggestion
        if (ilCurlConnection::_isCurlExtensionLoaded()) {
            try {
                $curl = $this->getCurlConnection($settings, $url);
                $curl->exec();
```

components/ILIAS/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
```diff
@@ -163,28 +212,54 @@ private function generateServerInfoFile(string $key, string $filename): void
...
+    private function getCurlConnection(ilSetting $settings, string $url, ?string $token = null): ?ilCurlConnection
     {
+        if (!ilCurlConnection::_isCurlExtensionLoaded()) {
+            return null;
+        }
+
+        if (!is_null($token)) {
+            $url = $url . "?token=" . $token;
+        }
+
```
thojou:
```suggestion
    private function getCurlConnection(ilSetting $settings, string $url, ?string $token = null): ilCurlConnection
    {
        if (!is_null($token)) {
            $url = $url . "?token=" . $token;
        }

```

components/ILIAS/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
```diff
@@ -163,28 +212,54 @@ private function generateServerInfoFile(string $key, string $filename): void
...
-
         $curl->setOpt(CURLOPT_FOLLOWLOCATION, 1);
         $curl->setOpt(CURLOPT_MAXREDIRS, 1);
 
-        $result = $curl->exec();
-        $curl->close();
+        return $curl;
+    }
+
+    private function getPHPIniValuesByFileGetContents(string $url, ?string $token = null): string
```
thojou:
```suggestion
    /**
     * @throws ErrorException
     */
    private function getPHPIniValuesByFileGetContents(string $url, ?string $token = null): string
```

components/ILIAS/Authentication/classes/Setup/class.ilSessionMaxIdleIsSetObjective.php
```diff
@@ -163,28 +212,54 @@ private function generateServerInfoFile(string $key, string $filename): void
...
+        set_error_handler(static function (int $severity, string $message, string $file, int $line): never {
+            throw new ErrorException($message, $severity, $severity, $file, $line);
+        });
+
+        if (!is_null($token)) {
+            $url = $url . "?token=" . $token;
+        }
 
-        return $result;
+        return file_get_contents($url);
```
thojou:
```suggestion
        try {
            return file_get_contents($url);
        } catch (ErrorException $e) {
            restore_error_handler();
            throw $e;
        }
```

thojou (changes requested):
Hey @daniwe4,

Thanks again for your changes and for explaining the default behavior—for me, it makes a lot more sense now! :smiley: 

I just have a few small suggestions that I’d kindly like to ask you to consider:

* We should avoid throwing exceptions in `isApplicable`, as it would cause the objective to fail.
* The extra bool variable `restore_error_handler` can be avoided if the restore is handled directly inside the getPHPIniValuesByFileGetContents method.
* If we move the check for `ilCurlConnection::_isCurlExtensionLoaded()` outside of `getCurlConnection`, then the call to `getCurlConnection` can be placed inside the try-catch block for better handling.

I really appreciate your effort on this—thanks again for your great work! Let me know what you think. :smiley: 

Best,
@thojou

daniwe4:
Hi @thojou ,

thank you for your tips. All are now in code. I hope that's how it works.

Best regards,
@daniwe4

thojou (approved):

thojou:
Hey @daniwe4,

Thanks again for your changes. this looks good to me now.
I've merged it into trunk.

Best,
@thojou

mjansenDatabay:
@thojou / @daniwe4 I picked this to `release_10`.

PR #8927, opened 2025-01-29
Fix typo in ilias_de.lang for reminder email

Hufschmidt:
Minor typo, see changes. 

Should ideally also be applied to **v9** and **v10**.


dsstrassner:
hi @Hufschmidt, ILIAS 8 Branch is closed (End of Maintenance - Phase / we are now in the security phase)- so there is no possibility to merge your PR. If the problem also in ILIAS 9 and above exits, a change of the branch is maybe a solution?

PR #8940, opened 2025-01-30
SAML: KS Form for saml settings

mBeym:


components/ILIAS/Saml/classes/class.ilSamlSettingsGUI.php
```diff
@@ -494,26 +495,19 @@ private function showUserAttributeMappingForm(?ilPropertyFormGUI $form = null):
...
-        return $form;
+        return $this->ui_factory->input()->container()->form()->standard(
+            $this->ctrl->getFormAction($this, self::CMD_SAVE_SETTINGS),
+            [
+                self::LNG_LOGIN_FORM => $this->ui_factory->input()->field()->checkbox(
+                    $this->lng->txt('auth_saml_login_form'),
+                    $this->lng->txt('auth_saml_login_form_info')
+                )
+                    ->withValue((bool) ($values[self::LNG_LOGIN_FORM] ?? true))
+                    ->withDisabled(!$this->access->checkAccess(self::PERMISSION_WRITE, '', $this->ref_id)),
```
mjansenDatabay: It was wrong before, but we need to use `rbac_system` here to check the access. The form is located in the "Administratiuon" area, so we must not use `ilAccessHandler` here (which perform "Tree Checks").
Please also fix the other usages of `$this->access->` and remove this class property afterwards.

Furthermore. Please rename the "Submit Button Label" to `$this->lng->txt('refresh')` in case the user has no `write` permission.

mjansenDatabay (changes requested):

components/ILIAS/Saml/classes/class.ilSamlSettingsGUI.php
```diff
@@ -497,17 +497,24 @@ private function showUserAttributeMappingForm(?ilPropertyFormGUI $form = null):
 
     private function getSettingsForm(array $values = []): StandardForm
     {
-        return $this->ui_factory->input()->container()->form()->standard(
+        $access = $this->rbac_system->checkAccess(self::PERMISSION_WRITE, $this->ref_id);
+        $form = $this->ui_factory->input()->container()->form()->standard(
             $this->ctrl->getFormAction($this, self::CMD_SAVE_SETTINGS),
```
mjansenDatabay: @mBeym We should change the action as well (to the one relevant for displaying the form) if no `write` permission is granted.

components/ILIAS/Saml/classes/class.ilSamlSettingsGUI.php
```diff
@@ -126,7 +128,7 @@ public function __construct(private readonly int $ref_id)
         $this->ctrl = $DIC->ctrl();
         $this->tpl = $DIC->ui()->mainTemplate();
         $this->lng = $DIC->language();
-        $this->access = $DIC->access();
+        $this->rbac_system = $DIC->rbac()->system();
```
mjansenDatabay: You already have the complete RBAC services assigned to a class member one line below.

components/ILIAS/Saml/classes/class.ilSamlSettingsGUI.php
```diff
@@ -544,27 +544,29 @@ private function saveSettings(): void
...
             $this->tpl->setOnScreenMessage(self::MESSAGE_TYPE_SUCCESS, $this->lng->txt(self::LNG_SAVED_SUCCESSFULLY));
         }
 
-        $form->setValuesByPost();
-
         $this->showSettings($form);
     }
 
-    private function showSettings(?ilPropertyFormGUI $form = null): void
+    private function showSettings(StandardForm $form = null): void
```
mjansenDatabay:
```suggestion
    private function showSettings(?StandardForm $form = null): void
```

mjansenDatabay (changes requested):

components/ILIAS/Saml/classes/class.ilSamlSettingsGUI.php
```diff
@@ -494,23 +494,23 @@ private function showUserAttributeMappingForm(?ilPropertyFormGUI $form = null):
         $this->tpl->setContent($form->getHTML());
     }
 
-    private function getSettingsForm(): ilPropertyFormGUI
+    private function getSettingsForm(array $values = []): StandardForm
```
mjansenDatabay, line 500:
```suggestion
    /**
     * @param array<string, mixed> $values
     */
    private function getSettingsForm(array $values = []): StandardForm
```

mjansenDatabay (changes requested):

PR #8941, opened 2025-01-30
SAML: KS Form for saml idp settings

mBeym:


components/ILIAS/Saml/classes/class.ilSamlSettingsGUI.php
```diff
@@ -613,74 +646,103 @@ private function getIdpSettingsForm(): ilPropertyFormGUI
...
         }
-        $form->addCommandButton(self::DEFAULT_CMD, $this->lng->txt(self::LNG_CANCEL));
 
-        return $form;
+        $this->ctrl->setParameter($this, self::REQUEST_PARAM_SAML_IDP_ID, $this->idp->getIdpId());
+        return $this->ui_factory->input()->container()->form()->standard(
+            $this->ctrl->getFormAction($this, self::CMT_SAVE_IDP_SETTINGS),
+            $inputs
+        );
     }
```
thojou: I ran into a syntax error at this point when checking out the PR. Could you verify if this section is correct? There might be something off here.

`ParseError: syntax error, unexpected token "private"`

image

components/ILIAS/Saml/classes/class.ilSamlSettingsGUI.php
```diff
@@ -568,31 +569,63 @@ private function showSettings(?StandardForm $form = null): void
...
 
-        $uid_claim = new ilTextInputGUI($this->lng->txt('auth_saml_uid_claim'), 'uid_claim');
-        $uid_claim->setInfo($this->lng->txt('auth_saml_uid_claim_info'));
-        $uid_claim->setRequired(true);
-        $form->addItem($uid_claim);
+        /** @var FormInput[] $inputs */
+        $inputs = [
+            $ui_field->text(
+                $this->lng->txt('auth_saml_idp')
+            )->withValue($values['entity_id'] ?? "")->withDisabled(true),
```
thojou, line 580: Is it intentional that this field doesn't have a specific name?

components/ILIAS/Saml/classes/class.ilSamlSettingsGUI.php
```diff
@@ -568,31 +569,63 @@ private function showSettings(?StandardForm $form = null): void
         ]));
     }
 
-    private function getIdpSettingsForm(): ilPropertyFormGUI
+    private function getIdpSettingsForm(array $values = []): StandardForm
```
thojou, line 572: The doc type could be more specific, `@param array<string, mixed>` instead of a generic array.

components/ILIAS/Saml/classes/class.ilSamlSettingsGUI.php
```diff
@@ -568,31 +569,63 @@ private function showSettings(?StandardForm $form = null): void
...
+                    'account_migr_status' => (bool) ($values['account_migr_status'] ?? true)
+                ]
+                : null
+            )
+        ];
 
-        $sync = new ilCheckboxInputGUI($this->lng->txt('auth_saml_sync'), 'sync_status');
-        $sync->setInfo($this->lng->txt('auth_saml_sync_info'));
-        $sync->setValue('1');
+        $write_access = $this->rbac->system()->checkAccess(self::PERMISSION_WRITE, '', $this->ref_id);
```
mjansenDatabay:
```suggestion
        $write_access = $this->rbac->system()->checkAccess(self::PERMISSION_WRITE, $this->ref_id);
```

thojou (changes requested):

PR #8943, opened 2025-01-30
UI Components, fix validate issue, remove dialog from modal role 43609

Amstutz:
Another issue related to https://mantis.ilias.de/view.php?id=43609.

The role dialog is not needed if the element is a dialog. This should not be shown as an error, but a warning from my understanding. However, https://validator.w3.org/nu/ shows this as error, which is what most of our testers use. 

See also: https://github.com/validator/validator/issues/1299

I would therefore propose to just remove it, since I do not see any immediate benefit.


klees:
Picked to trunk.

PR #8945, opened 2025-01-31
T&A 43538: highlight a tab

okaufman:
This will fix https://mantis.ilias.de/view.php?id=43538, where the tab for "Content for Recapitulation" is not highlighted.
This problem also happened when selecting "File download". 

Best wishes 
Olivia


kergomard:
Thank you very much for the PR @okaufman ! Merged and picked to trunk.

PR #8948, opened 2025-02-03
T&A 43838: Redirects user to test tab when test attempt gets finished by administrator

matheuszych:
[Mantis: 43838](https://mantis.ilias.de/view.php?id=43838)

When a administrator finishes a test attempt, while the user ist still taking the test, the user gets redirected to the info tab of the test. The user should get redirected to the test tab instead.


matheuszych:
Already reviewed by @thojou.

kergomard:
Thank you very much for the PR @matheuszych ! Merged and picked to trunk.

PR #8949, opened 2025-02-03
T&A 43749: Fixes checkboxes of multiple choice question (multiple answers) not being checked in detailed evaluation view

matheuszych:
[Mantis: 43749](https://mantis.ilias.de/view.php?id=43749)

In the detailed evaluation view the checkboxes of multiple choice (multiple answers) questions are partially shown incorrectly. This seems to be related to a variable being named differently in one spot, not passing the value further. Since it is correctly working in `release_10`, i copied the code from there and everything seems to work as expected.


matheuszych:
Already reviewed by @thojou.

matheuszych:
This should also resolve [Mantis: 43932](https://mantis.ilias.de/view.php?id=43932)

kergomard:
Thank you very much @matheuszych ! I don't like the nomenclature here (`$user_solution` should actually simply be `$solution` as it is not always the answer of the user), but your correction makes this consistent across question types, so it is right for now. Merged and picked to ILIAS 10 and trunk.

PR #8953, opened 2025-02-03
T&A 43837: allow multi select to finish pass only if processing time resett…

lukas-heinrich:
Hi everyone,

this PR is related to [43837](https://mantis.ilias.de/view.php?id=43837). With this change, it is no longer possible to end multiple test passes simultaneously if the test only allows a single test pass.

I look forward to your feedback, thank you in advance! These changes were reviewed in an internal process by @thojou  .

Best,
@lukas-heinrich

### DIFF
--- a/components/ILIAS/Test/src/Participants/ParticipantTableFinishTestAction.php
+++ b/components/ILIAS/Test/src/Participants/ParticipantTableFinishTestAction.php
@@ -126,6 +126,20 @@ class ParticipantTableFinishTestAction implements TableAction
             $this->user
         );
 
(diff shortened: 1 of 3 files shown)


kergomard (changes requested):

I would  have two nit-picks:
- Could you please remove the comment. It just states what the code below does so it doesn't add much and it is not even easier to rid than the code.
- Could you please change the order and move the variable assignment to `$test_session_factory` below your new code? It is only needed later on. This way we have a clearer order.

Thanks and best,
@kergomard

kergomard:
Thank you very much @lukas-heinrich ! Merged and picked to trunk.

PR #8955, opened 2025-02-03
_10/UI/scss-docs clarifications about compiling and custom system styles

catenglaender:
# Issue

While fixing other issues (e.g. #8897 ) it came up that the instructions towards compiling delos from SASS scss to CSS are not up to date.

# Changes

This PR suggests updates to the SCSS Coding Guidelines and the Custom System Style Instructions to reflect the most recent best practices.

While the documentation for sure could be optimized and clarified even further, I feel like this is at least a good step towards more clarity and correctness.

### DIFF
--- a/templates/Guidelines_SCSS-Coding.md
+++ b/templates/Guidelines_SCSS-Coding.md
@@ -1,7 +1,6 @@
 # General
 
-*Please note, that you need to provide a Pull Request for all CSS related changes done in this folder. The CSS Coordinators
-will then review the PR and give you feedback and/or merge your changes.*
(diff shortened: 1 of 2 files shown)


dsstrassner:
Not sure about the assignment of the ui coordinators or if I should add @acgruber or only @amstutz

Amstutz:
thx a lot!

Amstutz:
@catenglaender , note that we still need branches for 10 and trunk here: https://github.com/ILIAS-eLearning/delos/tree/release_9 . Currently we only do 9 afaik.

PR #8956, opened 2025-02-03
LearningModule: propagate change in title of page/chapter to LOM (43861)

schmitz-ilias:
This PR fixes [43861](https://mantis.ilias.de/view.php?id=0043861), by having the `EditSubObjectsGUI` also update the title of pages and chapters in LOM.

`ilLMObject::update` (as well as `create` and `delete` and so on) already take care of keeping LOM up to date, but for changing the title `EditSubObjectsGUI` uses the static `ilLMObject::_writeTitle`.


alex40724:
This is fixed differently. Thanks for the PR.

PR #8960, opened 2025-02-04
Trash: remove termination on common error

iszmais:
https://mantis.ilias.de/view.php?id=43992


alex40724:
Thanks for the PR. I hope I find some time to investigate on this, since it would be better to find the initial cause of the issue.

PR #8966, opened 2025-02-04
UI: remove indifferent aria label from viewcontrol

iszmais:
https://mantis.ilias.de/view.php?id=43994

Alternatively, IMO an implementation which allows the provision of own buttons (with set aria-labels) would also be a sufficient solution.


Amstutz:
Hi @iszmais 

Thx a lot for this improvement! We believe it to be implemented correctly. Ill merge this, however since there is also so spec change in the rule, we will also notify the JF about this merge.

Thx a lot
@thibsy , @klees , @Amstutz

PR #8978, opened 2025-02-05
T&A 30750: Poor performance using "Corrections"

lukas-heinrich:
Hi everyone,

This PR addresses the issue described in [30750](https://mantis.ilias.de/view.php?id=30750) where the corrections process for adding an answer option in ILIAS is very slow. With this PR, I propose several changes to improve the process performance.

For reproduction, I worked with a test containing 11 questions (10 of them being Cloze Questions) that was completed by 500 participants to achieve approximately the same scale as described in the ticket and to conduct a better comparison.

As can be traced in the commit history, I implemented and tested each change separately step by step. Using comparison checks, I verified that my changes do not alter the expected processing results. This PR includes the following changes:

1. The "Add Answer" operation always explicitly refers to a specific question. However, during point recalculation, every question in the test is updated in the database. I could not identify any scenario where recalculating all questions affects the overall result calculation.
2. Updating the total results (*tst_result_cache*) only after the final recalculation instead of after each update
3. Creation of question objects once; I found that this method is called for each test attempt and rebuilds the same data each time. As far as I can tell, the object has no user state, so it's possible to create the object once and reuse it subsequently
4. Optimized method for updating points and querying all participant results only calculates information relevant to the post-correction process, thereby drastically reducing database queries
5. Optimization of database queries by caching parameters for logging and caching combinations for Cloze-Gaps

Through analysis using external tools, I measured the performance as follows:

- Current status in `release_9`: 103 seconds, ~235,000 database queries, 570MB peak memory
- Status after the last commit in this PR: 1.8 seconds, ~12,000 database queries, 52MB peak memory

The changes and measurement results have been documented in more detail in internal documentation. We can provide this document if needed.

Thank you in advance for the review, I look forward to feedback on these changes.

Best,
@lukas-heinrich

### DIFF
--- a/Modules/Test/classes/class.ilTestCorrectionsGUI.php
+++ b/Modules/Test/classes/class.ilTestCorrectionsGUI.php
@@ -286,10 +286,12 @@ class ilTestCorrectionsGUI
         $this->main_tpl->parseCurrentBlock();
     }
 
-    protected function showAnswerStatistic()
+    protected function showAnswerStatistic($participant_results = null)
     {
         $questionGUI = $this->getQuestion((int) $this->testrequest->raw('qid'));
-        $solutions = $this->getSolutions($questionGUI->object);
+        $solutions = $participant_results
+            ? $this->getSolutionsByParticipantResults($questionGUI->object, $participant_results)
+            : $this->getSolutions($questionGUI->object);
 
         $this->setCorrectionTabsContext($questionGUI, 'answers');
 
@@ -346,10 +348,10 @@ class ilTestCorrectionsGUI
         $scoring = new ilTestScoring($this->testOBJ, $this->database);
         $scoring->setPreserveManualScores(true);
         $scoring->setQuestionId($question_id);
-        $scoring->recalculateSolutions();
+        $results = $scoring->recalculateSolutions();
 
         $this->main_tpl->setOnScreenMessage('success', $this->language->txt('saved_successfully'));
-        $this->showAnswerStatistic();
+        $this->showAnswerStatistic($results);
     }
 
     protected function confirmQuestionRemoval()
@@ -526,6 +528,24 @@ class ilTestCorrectionsGUI
         }
 
         return $solutionRows;
+    }
+
+    /**
+     * @param array<int, ilTestEvaluationUserData> $participant_results
+     */
+    protected function getSolutionsByParticipantResults(assQuestion $question, array $participant_results): array
+    {
+        $solutions = [];
+
+        foreach ($participant_results as $active_id => $result) {
+            foreach ($result->getPasses() as $pass) {
+                foreach ($question->getSolutionValues($active_id, $pass->getPass()) as $row) {
+                    $solutions[] = $row;
+                }
+            }
+        }
+
+        return $solutions;
     }
 
     /**
--- a/Modules/Test/classes/class.ilTestCorrectionsGUI.php
+++ b/Modules/Test/classes/class.ilTestCorrectionsGUI.php
@@ -345,6 +345,7 @@ class ilTestCorrectionsGUI
 
         $scoring = new ilTestScoring($this->testOBJ, $this->database);
         $scoring->setPreserveManualScores(true);
+        $scoring->setQuestionId($question_id);
         $scoring->recalculateSolutions();
 
         $this->main_tpl->setOnScreenMessage('success', $this->language->txt('saved_successfully'));
--- a/Modules/Test/classes/class.ilTestCorrectionsGUI.php
+++ b/Modules/Test/classes/class.ilTestCorrectionsGUI.php
@@ -286,7 +286,10 @@ class ilTestCorrectionsGUI
         $this->main_tpl->parseCurrentBlock();
     }
 
-    protected function showAnswerStatistic($participant_results = null)
+    /**
+     * @param null|array<int, ilTestEvaluationUserData> $participant_results
+     */
+    protected function showAnswerStatistic(?array $participant_results = null)
     {
         $questionGUI = $this->getQuestion((int) $this->testrequest->raw('qid'));
         $solutions = $participant_results
--- a/Modules/Test/classes/class.ilTestCorrectionsGUI.php
+++ b/Modules/Test/classes/class.ilTestCorrectionsGUI.php
@@ -287,7 +287,7 @@ class ilTestCorrectionsGUI
     }
 
     /**
-     * @param null|array<int, ilTestEvaluationUserData> $participant_results
+     * @param null|list<ilTestEvaluationUserData> $participant_results
      */
     protected function showAnswerStatistic(?array $participant_results = null)
     {
@@ -534,7 +534,7 @@ class ilTestCorrectionsGUI
     }
 
     /**
-     * @param array<int, ilTestEvaluationUserData> $participant_results
+     * @param list<ilTestEvaluationUserData> $participant_results
      */
     protected function getSolutionsByParticipantResults(assQuestion $question, array $participant_results): array
     {
--- a/Modules/Test/classes/Evaluation/class.ilTestEvaluationFactory.php
+++ b/Modules/Test/classes/Evaluation/class.ilTestEvaluationFactory.php
@@ -123,6 +123,62 @@ class ilTestEvaluationFactory
         return $ret;
     }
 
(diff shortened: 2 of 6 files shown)


dsstrassner:
Hi @lukas-heinrich, please provide the document to @mjansenDatabay & @kergomard & me via e-mail.

Modules/Test/classes/class.ilTestCorrectionsGUI.php
```diff
@@ -286,10 +286,12 @@ protected function showSolution()
         $this->main_tpl->parseCurrentBlock();
     }
 
-    protected function showAnswerStatistic()
+    protected function showAnswerStatistic($participant_results = null)
```
mjansenDatabay:
```suggestion
    /**
     * @param null|array<int, ilTestEvaluationUserData> $participant_results
     */
    protected function showAnswerStatistic(?array $participant_results = null)
```

kergomard (changes requested):
Thank you very much for the PR a for tackling the performance issue @lukas-heinrich !

I've the following immediate concerns:
- [x] As far as I can see `ilTestScoring::recalculateSolution()` is still needed in `ilTestScoringGUI::saveManScoringParticipantScreen()` and cannot be removed.

Some small things:
-  [x] Please do not remove typing like [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8978/files#diff-4e9b80468abe4288bf99750d58dd525cc4a2824f679b9d957e9e4c33fa163072R136). If there is a problem it is clearly not the typing, but the parameters passed.
- [x] Please restructure this if - else. It looks bonkers.
- [x] Please use consistent doc-blocks for arrays of the type `array<type>` (or `list<type>`).
- [x] [This check](https://github.com/ILIAS-eLearning/ILIAS/pull/8978/files#diff-4e9b80468abe4288bf99750d58dd525cc4a2824f679b9d957e9e4c33fa163072R126) is useless: We always return an array. Please remove it.
- [x] Please remove [the intermediate variable here](https://github.com/ILIAS-eLearning/ILIAS/pull/8978/files#diff-4e9b80468abe4288bf99750d58dd525cc4a2824f679b9d957e9e4c33fa163072R128) and use strict typing below. We do know what type this variable has, don't we? If we don't: Please fix the problem where it originates.
- [x] Please remove [the intermediate variable here](https://github.com/ILIAS-eLearning/ILIAS/pull/8978/files#diff-4e9b80468abe4288bf99750d58dd525cc4a2824f679b9d957e9e4c33fa163072R144). You can keep things readable by using line breaks in the following line.
- [x] Please add a return-type to `ilTestScoring::updateReachedPoints` and use strict comparisons within it.
- [x] Single Quotes are enough [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8978/files#diff-4e9b80468abe4288bf99750d58dd525cc4a2824f679b9d957e9e4c33fa163072R197).
- [x] As far as I can see the float-cast [here](https://github.com/ILIAS-eLearning/ILIAS/pull/8978/files#diff-4e9b80468abe4288bf99750d58dd525cc4a2824f679b9d957e9e4c33fa163072R201) can go.

Please create a PR for this against ILIAS 10, with the following changes. I might have missed something, so all of this is conditioned on it being possible.
- [ ] Remove `assQuestion::_setReachedpoints()` completely.
- [ ] Remove the duplication of code in `ilTestEvaluationFactory::getEvaluationData()` and `ilTestEvaluationFactory::getCorrectionsEvaluationData()` by restructuring the functions correspondingly.
- [ ] Make the parameter in `ilTestCorrectionsGUI::showAnswerStatistic()` compulsory (maybe you also need to add one more that hands in the question-object). If it isn't possible, it isn't, but I deeply dislike this type of optional parameters. From where I stand they are a smell.
- [ ] Make sure to not introduce any dependence on the `$DIC` in `TestScoring`.

Thanks again for all the work! I did not test the functioning of it, but I will give it a quick spin once we have the PR for ILIAS 10.

Thanks again and best,
@kergomard

lukas-heinrich:
Hi @kergomard  ,

thank you very much for the review and the detailed feedback. I am preparing a PR for Ilias 10, I have just pushed the proposed changes for the release_9.

Best
@lukas-heinrich

kergomard:
Thank you very much again @lukas-heinrich !

I will wait for the PR for ILIAS 10 and merge all in one, ...just for me not loosing the overview.

There is one more little thing I missed before, sorry: Could you have a look at the failing Content-Style-Checks?

Thanks and best,
@kergomard

rfalkenstein:
Hi guys, we are currently testing this patch for ILIAS 8 and it works well. The loading times for adding a new answer in the corrections tab was reduced from 8 minutes to 3 seconds. In the next step we could offer to introduce these fixes into our ILIAS 9 staging installation. 
@lukas-heinrich @thojou  do you consider the current state of the branch "bugfix/9/30750" ready for testing with ILIAS 9? Then we'll introduce this into our ILIAS 9 staging installation and test (again, but for ILIAS 9) against tests with a lot of users.

thojou:
Hey @rfalkenstein,

Thanks for your feedback and for testing the changes in ILIAS 8.
This branch is complete as well. 
We’d really appreciate if you could test these changes on your ILIAS 9 staging installation.

Best,
@thojou

lukas-heinrich:
Hi @kergomard ,

I had to push another commit on this branch for Release 9 and rebase it. The background is that the `getEvaluationData` method has been changed in the meantime. To prevent changes to the method from being implemented in two places in the future, I have restructured and refactored the method call.

I would appreciate if you could take another look at the latest commit. We will provide a PR for Ilias 10 soon.

Thank you very much and best
Lukas

kergomard:
Hi @lukas-heinrich 

Thank you very much for your changes.

I'm going to be honest: I don't like this pattern at all:
- It doesn't simplify, but makes the code more convoluted.
- The function is still far too long.
- We have to "adaptor"-functions that just add an indirection.

I decided to quickly give it a go, as I think it is easier to demonstrate in what direction I would go, then to try to explain it. I couldn't create a PR against your branch (don't ask me why), but you can find a proposal here: https://github.com/kergomard/ILIAS/commit/9bd9068c40d3774aac1eaccab4e2052faebfb9fc . If you decide to use this: Please check it thoroughly! I think it works and I clicked on all buttons on the surface, but I didn't test it thoroughly.

Thanks again and best,
@kergomard

lukas-heinrich:
Hi @kergomard ,

thank you for the feedback and the implementation. I agree, this one is a much more elegant solution for structuring the methods. I have pushed the changes from your commit and thoroughly tested them - I haven't found any issues.

Thanks again and Best
@lukas-heinrich

kergomard:
Thank you for catching this error @lukas-heinrich !

Just that we are not blocking one another here: It is still correct that I'm waiting for a PR against ILIAS 10 here, right?

And one very small  request (at least for ILIAS 10):
- [ ] Could you please remove the reassignment to `$attempt` and pass `$this->addQuestionToAttempt($attempt, $row)` directly to `$user_eval_data->addPass()` (`$user_eval_data->addPass($row['pass'], $this->addQuestionToAttempt($attempt, $row))` - you need to split it on multiple lines, though). I do really believe we should treat our variables in function-scope as constants as much as we can (maybe one day we will be able to declare them as such).

Best,
@kergomard

lukas-heinrich:
Hi @kergomard ,

thank you for the hint, I will adjust it accordingly. 
It is correct that the revision for Ilias 10 is still in progress. However, I assume that this will be completed by the end of this week and that the PR can be provided after thorough testing.

Best
@lukas-heinrich

kergomard:
Hi @lukas-heinrich 

The PR against ILIAS 10 is still missing, right?

Thank you very much and best,
@kergomard

thojou:
Hey @kergomard,

We are currently conducting internal reviews on the ILIAS 10 PR.
As soon as we're ready, we will publish the branch.

Best,
@thojou

lukas-heinrich:
Hi @kergomard 

The PR against ILIAS 10 is now ready: #9277 

Best,
@lukas-heinrich

thojou:
Hey,

We've also received feedback from testing this branch on production datasets. No further issues were reported. From our perspective, this PR is ready to go! 🙂

Best,
@thojou

kergomard:
Thank you very much @lukas-heinrich 

As we now have the corresponding PR for ILIAS 10, I've merged this.

Best,
@kergomard

PR #8982, opened 2025-02-05
[MathJax] Fix "Read only" für Administration

fneumann:
Resolves https://mantis.ilias.de/view.php?id=44014

This disables all properties on the MathJax settings form if a user has only read access to the administration node. 

Unfortunately, the "Save" button of a UI form can't be omitted. So with "Read only" access, the button is renamed to "Reload" and the form is shown again.


mjansenDatabay (approved):

lscharmer (approved):

fneumann:
@lscharmer @mjansenDatabay Thanks for the reviews!

PR #8984, opened 2025-02-05
RTE/TinyMCE: Fix font size and font family selection + block formats

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=44044


mjansenDatabay:
@kergomard It would be great if you could review this. Matthias informed me via email that Florian Suittenpointner discovered during testing that the font and font size are no longer selectable.

Thanks in advance :pray: .

kergomard:
Thank you very much @mjansenDatabay for fixing this issue!

This looks right and I did a quick test. Merged and picked to trunk.

Best,
@kergomard

PR #8991, opened 2025-02-06
[FIX] #43712, #43711 UI: `Dropzone\File` test case descriptions.

thibsy:
Hey guys,

This PR fixes two issues:
- https://mantis.ilias.de/view.php?id=43712, and 
- https://mantis.ilias.de/view.php?id=43711

I bundled them together because they should actually fix the test case descriptions. I assume we need to generate another import file for this? Or would this be too much overhead? We can also adjust the test cases directly.

Let me know how to proceed here.

Kind regards,
@thibsy


Amstutz:
Hi @thibsy, 

Thx a lot. I'll merge this and pick to the trunk. For 11 and upwards, this will fix the TestCases. For 10, I would just change manually (if even necessary), since exporting and re-importing seems like an overkill for the small changes

thx

PR #8998, opened 2025-02-06
UI/Input/Rating: 0044003, validate HTML/missing aria-describedby

nhaagen:
https://mantis.ilias.de/view.php?id=44003


Amstutz:
thx!

PR #8999, opened 2025-02-06
UI component Adjust Expected Results of tree node key value #43714

yvseiler:
https://mantis.ilias.de/view.php?id=43714

Expected results text has been corrected to match the example shown.

Thanks for looking into this and merge if ok.


Amstutz:
Note the PR also covers, Expected Results Example 2 extended to make it easier to understand how to test

PR #9000, opened 2025-02-06
UI/ViewControls: 44034, add dropdown.js to registry

nhaagen:
https://mantis.ilias.de/view.php?id=44034


klees:
Picked to trunk.

PR #9001, opened 2025-02-07
LegacyForm: Implement Autocomplete in Legacy Forms

kergomard:
Hi all

Ok, this is a first version for some javascript to make the autocomplete work again. I would be glad if somebody would have a look at it, ...otherwise I would suggest to simply merge it. The situation should not get worth then currently ;-).

@Amstutz : There are a few changes in CSS, that is why I assigned you and added you as a reviewer. They should be small enough. ...although I think the scss was compiled with a newer version of `sass.js`, so that again there are quite a few changes in `delos.css`. I used the version in our node_modules.

@mjansenDatabay : Maybe you can have a look how well this works in the mail? I tested it, but you surely have a better eye for it.

See: https://mantis.ilias.de/view.php?id=42460,
https://mantis.ilias.de/view.php?id=42386

Thanks and best,
@kergomard


mjansenDatabay:
Hi @kergomard ,

I will check this.

I already found an issue regarding the event handling. When typing, ILIAS is flooded with HTTP requests. We'll need something like a deferred callback execution pattern.

When the respective events (`keyup` and `keydown`) are fired,  `window.setTimeout` can be used to let a callback fetch the data from the endpoint by using e.g. a timeout value of `500`.

```js
this.handle = window.setTimeout(..., 500);
```

If the same event happens again and again, you re-use the handle to clear the timeout.

```js
if (this.handle) {
 window.clearTimeout(this.handle);
}

this.handle = window.setTimeout(..., 500);
```

Best regards,
Michael

mjansenDatabay:
Hi @kergomard ,

Furthermore, I suggest:

- [x] Using an `autocomplete="off"` attribute for the input element.
- Adding some aria labels and other A11Y attributes:
  - [x] Use an `aria-live="asertive"` and `aria-relevant="additions"` attribute for the container element (of the suggestion list).
  - [x] Use a `role="status"` attribute for the container element (of the suggestion list).
- [x] Adding an additional space when the user selects a suggestion in "multi"-mode and the input field is not empty (1:1 as in `release_9`).

---

Regarding the keyboard navigation, your approach is (IMO) even better than the `jQueryUI` based approach.

Best regards,
Michael

kergomard:
Thank you very much for the feedback @mjansenDatabay !

I implemented the suggested changes. Concerning the aria-attributes I opted for `role="status"` and did not set `aria-live="assertive"` as `aria-live="polite"` is actually implied. I think this is the better option. As we do not know the  context, this is done through js..

Thanks again and best,
@kergomard

Amstutz (approved):
CSS changes are fine. Thx a lot for providing this solution!

matthiaskunkel:
Jour Fixe, 17 FEB 2025: We highly appreciate this suggestion and accept the PR for ILIAS 10 and trunk. There is no longer a need to re-introduce jQuery as suggested in PR8928 (which therefore will be closed).

kergomard:
Thanks @Amstutz and again @mjansenDatabay for the feedbacks and the improvements! Merged and picked to trunk.

PR #9003, opened 2025-02-07
LSO: fix component names in export tail dependencies (44084)

schmitz-ilias:
This PR fixes [44084](https://mantis.ilias.de/view.php?id=44084), looks like a component revision related mixup.


klees:
Thanks a lot!

klees:
Picked to trunk.

PR #9004, opened 2025-02-07
Mail: Use text for column header

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=44101

If approved, this has to be merged to `trunk` as well.


oliversamoila:
Hello @mjansenDatabay 
according to the last CSS-Squad meeting on 01 April I would like to ask you to merge the PR. 
We will discard the integration of the glyph as a column title for the time being.

Thank you very much !
Best regards
Oliver

PR #9005, opened 2025-02-10
UI: 43428, adjust rendering of breadcrumb title.

lukastocker:
Only decode special chars if the breadcrumb entry is the last entry, else don't decode the special chars as this step is not necessary in this case and to avoid loosing entries with double and single quotes.

https://mantis.ilias.de/view.php?id=43428


Amstutz:
Hi @lukastocker 

The solution still seems a bit shaky. Probably a further iteration is needed. However, it does fix the problem at hand, so we decided to take it untill we improved further here.

Thx

PR #9006, opened 2025-02-10
Export: Build export target with export object instead of obj id for standard and container export

chlulei:
Some Components change the object data with setters bevore the export and expect the export to honor these changes.

Bevore this PR the object data was read by using the object id of the export object. After this PR the object data is aquired from the export object directly.


dsstrassner:
Hi @smeyer-ilias, @tfamula says in https://mantis.ilias.de/view.php?id=42898, that this PR will also help to resolve Mantis 42898. As this Mantis issue blocks a Test Case, I kindly ask when the merge of this PR is planned.

smeyer-ilias (approved):
Thanks @chlulei

PR #9008, opened 2025-02-10
ObjectService: fix return type of gui factory

iszmais:
https://mantis.ilias.de/view.php?id=44116


kergomard:
Thank you very much for the PR @iszmais . This factory should never have been there. I just haven't gotten around to removing it. It is gone now.

Best,
@kergomard

PR #9012, opened 2025-02-11
Export: Default to current ILIAS version if major version is missing in export

chlulei:
Link: https://mantis.ilias.de/view.php?id=41417


smeyer-ilias (approved):

smeyer-ilias:
Thanks @chlulei for fixing this

PR #9013, opened 2025-02-11
PRG: 44130, 'completion_by' should point to course, not courselink

nhaagen:
https://mantis.ilias.de/view.php?id=44130


klees:
Picked to release_10 and trunk.

PR #9014, opened 2025-02-11
Mantis 43926:Fix thumbnail of certificate background iamge cropped incorrectly in object settings

mBeym:
https://mantis.ilias.de/view.php?id=43926

If accepted should be picked into trunk.


fhelfer:
Thank you very much for the PR @mBeym, this looks good, I tested it, it works and fixes the issue, and I cannot see any side effects
Ready for merge @mjansenDatabay 

Best, @fhelfer

PR #9015, opened 2025-02-11
10/UI/DataTable-Ordering mobile style 44049

catenglaender:
# Issue

https://mantis.ilias.de/view.php?id=44049

Ordering Table has no styled mobile view.

# Change

Applied data table mobile styling to ordering table
Using translation string for the word "order"
Added a point grid handle to communicate the affordance for drag and drop without words.

ordering-table_design


Amstutz:
Hi @catenglaender 

Thx a lot for this improvement. In general it is fine, however there is one thing:

- [ ] Symbol as Glyph: The new Symbol used (see screenshot) is implicitly in css hard coded without explicit semantics. A new glyph for it with a corresponding small description would be better. If done like this, it could also be reused in any context fitting the description and also easily be replaced in skins if necessary.

Thx.

catenglaender:
Hi @Amstutz,

here is the change with a draggable glyph. Adhering to the rules I have written into the factory it also has to be visible on desktop xD

image

image

Amstutz:
Thx a lot. 

Note, the public interface changed, so this needs JF approval. However, since the changes are most probably not controversial (glyph added), we (UI Coordinators) decided to fast-lane this, since there is some urgency due to the upcoming release.

Nevertheless, this needs to be announced at the upcoming jf, even if already merged.

thx

matthiaskunkel:
Jour Fixe, 04 AUG 2025: @yvseiler notified us about the new glyph that visualises drag&drop of table rows (as shown in the animation above). Description is available in the factory text.

PR #9019, opened 2025-02-11
Administration: Raise intead of throw error on permission denied

lscharmer:
This PR fixes Mantis Bug: https://mantis.ilias.de/view.php?id=43548

With this change the user will be redirected to a public page and shown the error message instead of throwing a hard error.


fneumann (approved):
LGTM and successfully tested

PR #9024, opened 2025-02-12
[FIX] #44148 UI: `Field\SwitchableGroup::withValue()` in all contexts

thibsy:
Hi folks, 

This PR fixes https://mantis.ilias.de/view.php?id=44148, by making `Field\SwitchableGroup::withValue()` call `isClientSideValueOK()` as well. Up till now this only happened when the input was nested in a `Field\Group` or similar, which lead to different behaviour depending on its context.

I let you decide how far down you want to pick this fix; the issue exists in all maintained versions.

Kind regards,
@thibsy


thibsy:
Merged with #7339

PR #9025, opened 2025-02-12
Mantis #44126 ILIAS 9: Only show a generic error message on error after auth failure

mBeym:
This PR changes the error message presented on the login page.

See: https://mantis.ilias.de/view.php?id=44126


thojou (approved):
Hey @mBeym,

thanks for your changes.
This looks good to me.

Best,
@thojou

PR #9026, opened 2025-02-13
Session: disallow timing-action for objects on info screen (43327)

schmitz-ilias:
This PR 'fixes' [43327](https://mantis.ilias.de/view.php?id=43327) by removing the broken action from the object list. It also removes `ilObjSessionGUI::renderObject`, since that method was only added to make the action work.


smeyer-ilias (approved):
Thanks @schmitz-ilias

PR #9028, opened 2025-02-13
[Course] Mantis #0043615: Initialize LOMServices

Akumatic:
lom_services was not initialized, thus throwing an exception when releasing a course.


dsstrassner:
Not sure which assignment is correct.

smeyer-ilias (approved):
Thanks @Akumatic for fixing this.

PR #9029, opened 2025-02-13
[FIX] #44176 UI: add support for all `Field\File` metadata inputs.

thibsy:
Hi folks,
    
This PR fixes https://mantis.ilias.de/view.php?id=44176, by removing the mechanism that merged the `Field\Hidden` dynamic inputs template with existing nested inputs of `Field\Group` inputs provided for metadata. This leads to a change in behaviour when handling file uploads, where a metadata input is provided, because it will be nested an additional time now. I have also decided not ti map the file-id value to the `Field\UploadHandler::getFileIdentifierParameterName()` key anymore, since this was only necessary because due to the merger of said inputs.

@chfsx this PR also contains a commit to address the changed behaviour in file upload handling. The `ilObjFile` was the only place where I could find usages of the `Field\File` input with a metadata input.

Kind regards,
@thibsy

### DIFF
--- a/components/ILIAS/File/classes/class.ilObjFileGUI.php
+++ b/components/ILIAS/File/classes/class.ilObjFileGUI.php
@@ -422,6 +422,7 @@ class ilObjFileGUI extends ilObject2GUI
         $inputs[self::PARAM_FILES] = $this->ui->factory()->input()->field()->file(
             $this->upload_handler,
             $this->lng->txt('upload_files'),
+            null,
             $this->ui->factory()->input()->field()->group([
                 self::PARAM_TITLE => $this->ui->factory()->input()->field()->text(
                     $this->lng->txt('title')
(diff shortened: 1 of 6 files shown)


thibsy:
The pipelines are failing because they are currently broken in trunk.

chfsx (approved):
LGTM, thanks a lot

Amstutz (changes requested):
Hi @thibsy

Thank you a lot for contributing to ILIAS and the improvements to the file input.

We checked the new "Example 3: with metadata". There is only one question, no further comments from a UI/UX perspective, and the HTML validation did not reveal any errors in this area. Many thanks for that.

- [ ] After clicking on "Save" button in example 1 and 2 an array feedback code is displayed. This is not the case for example 3. Is this intended?

kindly,
@Amstutz with kind support of @yvseiler

thibsy:
@Amstutz this was intentional, but if all others display a result, I can add this to the example as well.

Amstutz:
@thibsy I do not have a strong opinion for one or the other.

Just one thing:

- [x] Clear description: The human readable description of the example that is pushed to testrail must make it obvious for the tester, what the example is supposed to do. Feel free to adapt the description or the example itself, as long as it is immediately clear, what the intended behavior is, we are fine.

thx.

thibsy:
@Amstutz done. Pipeline fail due to a wrong copyright check, see #9554

thibsy:
Merged with #7339

PR #9031, opened 2025-02-13
Test: Remove Option to Disable Taxonomies from QPL

kergomard:
Hi all

The option to disable taxonomies in the question pool is actually very problematic and we would like to remove it. The problem is that taxonomies could be used in tests as a criterion to select questions. Currently this criterion is independent of the setting in the pool, but this leads to a situation where you might use taxonomies as a criterion to select questions from a pool where taxonomies are disabled and this does not make much sense. So we could actually make the selection by taxonomies depend on the taxonomies being enabled in the question pool, but this again would lead to problems when the taxonomies are enabled or disabled in a linked question pool.

We thus suggest to simply remove this option from ILIAS 10 and above. Taxonomies will thus always be enabled.

Best,
@kergomard


dsstrassner:
I create an Abandon FR and change the testcase, if the JF agrees on this PR:

- https://testrail.ilias.de/index.php?/cases/view/37

matthiaskunkel:
Jour Fixe, 17 FEB 2025: We highly appreciate this suggestion and accept the PR for 10 and trunk.

kergomard:
Thank you all for the positive answer! Merged and picked to trunk.

dsstrassner:
[Abandon Page in the Feature Wiki](https://docu.ilias.de/go/wiki/wpage_8601_1357)

PR #9032, opened 2025-02-13
Mantis 44181: add missing typecasts in class.ilCustomuserFieldsGUI

ingoj:
possible fix for mantis 44181, typecasts are missing in 3 of 4 calls to getPluginForType


ingoj:
I updated my repo so that it mathces the fixes already done for ILIAS 10

kergomard:
Thank you very much for the PR. Merged. No pick needed as it is fixed in all newer versions.

PR #9033, opened 2025-02-13
Transfer authorities for ECS Interface

Akumatic:
Due to @pascalseeland's parental leave, members of the University of Stuttgart and University of Freiburg would take over his Authorities for the ECS Interface.

### DIFF
--- a/docs/development/maintenance.md
+++ b/docs/development/maintenance.md
@@ -572,13 +572,13 @@ of ILIAS. The file contains the following fields:
 [//]: # (BEGIN ECSInterface)
 
 * **ECS Interface**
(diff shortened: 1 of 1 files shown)


dsstrassner:
hi @Akumatic, perhaps it would make sense to leave Pascal as a authority in  'Authority to (De-)Assign Authorities'.

pascalseeland:
@dsstrassner we discussed this issue internally and I would prefer it like we have written in the PR. This way people actively working on ILIAS will be responsible for keeping things moving. I did choose the take the 2 year parental level and I will not keep track of any ILIAS development or other things happening in the community which I than would still be required to do.

PR #9034, opened 2025-02-13
Update class.ilObjRemoteCourseGUI.php

jheim2:
Cast availability type to string before setting value


dsstrassner:
Hi @jheim2, is there a Mantis report, which we could link to?

jheim2:
https://mantis.ilias.de/view.php?id=44189

smeyer-ilias (approved):
Thanks @jheim2 for fixing this.

smeyer-ilias:
@jheim2 I remove my assignment, since the Remote*-components are part of the ECS-Service.

PR #9035, opened 2025-02-13
Mantis 44187: add excludePattern to RoleFilter

ingoj:
This is an approach for af Fix for Mantis 44187. 
From the documentation in ILIAS docu platform I saw, that includePatterns may occur multiple times, but exclude can only be used once. 
This approach allows that no pattern is present (I found no way to allow multiple includes in another way)
I could not yet test this modification (sorry), so this is just a suggestion.


smeyer-ilias:
Thanks @ingoj for the PR. 
I close this PR since it is fixed with https://github.com/ILIAS-eLearning/ILIAS/pull/9131/files

Please see my comment in Mantis 44187

PR #9045, opened 2025-02-14
[PROPOSAL] deprecation of active record

chfsx:
We propose to deprecate ActiveRecord and suggest to remove it with ILIAS 12 or 13 (depending on how quickly users can replace their implementations). The ActiveRecord was a simple way of handling database operations and object mapping in its day. However, the pattern is not very lightweight and has a very large overhead. It also leads to many database queries and high memory consumption. For several releases now, the “Repository Pattern” has been propagated as a better alternative.

### DIFF
--- a/components/ILIAS/ActiveRecord/class.ActiveRecord.php
+++ b/components/ILIAS/ActiveRecord/class.ActiveRecord.php
@@ -17,12 +17,11 @@
  *********************************************************************/
 
 /**
- * Class ActiveRecord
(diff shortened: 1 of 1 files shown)


matthiaskunkel:
Jour Fixe, 17 FEB 2025: Fabian notified us about the plans to remove active records from ILIAS. A final deprecation might take place in ILIAS 12. Please check your components and refactor them accordingly to get rid of the usage of active records.

PR #9053, opened 2025-02-17
Blog/MediaCast/MediaPool: create LOM sets for pre-existing objects (44066)

schmitz-ilias:
This PR fixes [44066](https://mantis.ilias.de/view.php?id=44066) by adding a migration that creates LOM sets for pre-existing objects of those types that support LOM since ILIAS 10. Empty LOM sets for these objects leads to problems in export/import, and probably also in other places.


alex40724:
@schmitz-ilias Please add the migrations at the bottom of the release 10 page, if not done yet.

PR #9054, opened 2025-02-17
DataCollection: add migration to create LOM sets for pre-existing Data Collections (44066)

schmitz-ilias:
This PR adds a migration that creates LOM sets for pre-existing Data Collections. Empty LOM sets for Data Collections could lead to problems in export/import, and probably also in other places, see [44066](https://mantis.ilias.de/view.php?id=44066).


iszmais (approved):

iszmais:
Thx Alot!

PR #9055, opened 2025-02-17
Survey: Reintroduce third state when checking `_lookupFinished`

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=44214


alex40724:
Thanks for the PR!

PR #9056, opened 2025-02-17
Math: Throw `DomainException` instead of `E_USER_ERROR` (PHP 8.4)

mjansenDatabay:
This PR suggests to throw an exception when exponential expression cannot be converted, instead of using the deprecated `E_USER_ERROR`.

See: https://www.php.net/manual/en/migration84.deprecated.php

### DIFF
--- a/components/ILIAS/Math/classes/class.ilMathBaseAdapter.php
+++ b/components/ILIAS/Math/classes/class.ilMathBaseAdapter.php
@@ -126,7 +126,9 @@ abstract class ilMathBaseAdapter implements ilMathAdapter
             if ((($exp_sign = $exp[0]) === '+') || ($exp_sign === '-')) {
                 $exp = substr($exp, 1);
             } else {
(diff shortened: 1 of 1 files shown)


mjansenDatabay:
@dsstrassner / @kergomard / @thojou / @maxbecker 

This can have an impact on the "Test" component as the only consumer.

kergomard:
Thanks for the fix and heads-up @mjansenDatabay ! We will clearly take care of any consequences in the "Test".

Best,
@kergomard

thojou:
Hey all,

as this may effect the T&A Module, I will shepherd this PR. 
I would give other ILIAS core developers the chance to intervene, otherwise I will merge the PR after 14 days.

matthiaskunkel:
According to his post Thomas takes care of the PR. So I removed the Shephard tag.

thojou:
Hey @mjansenDatabay,

Thanks for your PR! As mentioned earlier, I've merged the changes into the trunk.

Best,
@thojou

PR #9059, opened 2025-02-17
mantis 44139: Release_10 lang quotes

ingoj:
Ported some changes from pr# 9016 to release_10
refers to mantis #44139


matthiaskunkel:
Thanks to @ingoj for this PR.

PR #9061, opened 2025-02-18
T&A 25064: Fixes conceptual issue where users can estimate their result when viewing their learning progress in a test

aaronbidzan:
Mantis: https://mantis.ilias.de/view.php?id=25064

This PR attempts to solve the conceptual issue mentioned. Previously, participants of a test with the learning progress feature enabled could estimate their result by looking at the learning progress percentage, even if the date after which results should be shown is in the future.

I have now added a check to see if the participant trying to access their learning progress can view their test result before showing them their learning progress. This now gives the same feedback to the user as if they were trying to view their results under the “Result” tab.

Kind regards,
@bidzanaaron

### DIFF
--- a/Modules/Test/classes/class.ilObjTestGUI.php
+++ b/Modules/Test/classes/class.ilObjTestGUI.php
@@ -452,8 +452,25 @@ class ilObjTestGUI extends ilObjectGUI implements ilCtrlBaseClassInterface, ilDe
                 if ((!$this->access->checkAccess("read", "", $this->testrequest->getRefId()))) {
                     $this->redirectAfterMissingRead();
                 }
+
                 $this->prepareOutput();
                 $this->addHeaderAction();
+
+                $test_object = $this->getTestObject();
+                if ($test_object === null) {
+                    $this->redirectToInfoScreenObject();
+                    break;
+                }
+
+                $test_session = $this->test_session_factory->getSessionByUserId($this->user->getId());
+                if (!$test_object->canShowTestResults($test_session)) {
+                    $this->tpl->setOnScreenMessage(
+                        'info',
+                        $this->lng->txt('tst_res_tab_msg_no_lp_access'),
+                    );
+                    break;
+                }
+
                 $this->tabs_gui->activateTab(ilTestTabsManager::TAB_ID_LEARNING_PROGRESS);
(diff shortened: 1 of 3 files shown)


kergomard (changes requested):

Ok, I think we first need to clarify something conceptual:
- What makes the case of having a date so special? We have the same problem if no access to the test results is granted or when the access is dependent on passing the test. Your PR actually already does block access in all cases where the use does not have access, as `ilObjTest::canShowTestResults()` does check for all cases as far as I can see, ...but the message then is nonsensical. So, I would have the following change requests:

- [ ] If I'm understanding this right the `!$test_object->hasAnyTestResult($test_session)` check [in this line](https://github.com/ILIAS-eLearning/ILIAS/pull/9061/files#diff-1cb659c07cfa02fb1a744e11d5b0cf83003aa4d4a493c7f0943005f089d4cfffR1062) is here only as accessing the learning progress without any results is unproblematic. If I'm right, I would ask you to remove this function and just check directly in `executeCommand()` for `$test_object->canShowTestResults()`. It is enough information.
- [ ] Please create a different language variable for this as the variable you are currently using is unclear. To keep this simple and not have to implement a different message for every case I would suggest "You are currently not allowed to access your Learning Progress, as you are not allowed to see your test results.", "Sie dürfen aktuell nicht auf Ihren Lernfortschritt zugreifen, da der Zugriff auf die Testresultate nicht freigegeben ist."
- [ ] This will also allow you to remove `ilObjTestGUI::getFormattedDateFormat()` and the `use` statement for the Data-Factory.

Thank you very much and best,
@kergomard

aaronbidzan:
Hey @kergomard,

I have taken your feedback into account and I fully agree with your suggestions.
As you asked, I have added in the English and German language variables to make it more clear and removed the functions as they are not needed.

If there is anything else I haven't considered or other things that you would like changed, be sure to let me know.

Kind regards,
@bidzanaaron

kergomard:
Thank you very much @bidzanaaron ! I merged and picked it to ILIAS 10 and trunk.

After looking at it again, I wondered whey we needed to whole `if getTestObject === null` and I removed it in 10 and above. Additionally, and this is just for me ;-) : Please avoid introducing intermediate variables for trivial things (in this case `$test_object`. I just removed it again in ILIAS 10 and above. I would personally also like to avoid `$test_session = ...`, but here I clearly see the advantage in readability.

Thanks again and best,
@kergomard

PR #9062, opened 2025-02-18
T&A 28142: prevent skill assignment for questions from a question pool (ILIAS 9)

lukas-heinrich:
Hi everyone,

this PR relates to ticket [28142](https://mantis.ilias.de/view.php?id=28142). It prevents questions from a Question Pool in a test from having their competency schema modified.

I am looking forward to feedback.
Best
@lukas-heinrich

PR #9063, opened 2025-02-18
T&A/10 28142: prevent skill assignment for questions from a question pool (ILIAS 10)

lukas-heinrich:
Hi everyone, 

this is an adjusted version of #9062  (28142) for Ilias 10.

Best
 @lukas-heinrich


dsstrassner:
PR #9062 is for ILIAS 9

kergomard (changes requested):

This is for both PRs.

I think we can mostly run with this as this is all legacy anyway. I would have a couple of small changes though:
- [x] Please remove the function `ilAssQuestionSkillAssignmentsTableGUI::areManipulationsEnabled()` and access the variable directly.
- [x] Please make the new function `areManipulationsPossible()` to private.
- [x] Please do not introduce Variables in camelCase, even if the whole class uses camelCase (so it should be `$manipulation_allowed_list`).

Thank you very much and best,
@kergomard

PR #9065, opened 2025-02-18
T&A 43423: Addition of optional working time column in participants table

aaronbidzan:
This PR aims to implement the proposed feature in https://mantis.ilias.de/view.php?id=43423.

An optional but not initially visible working time column has been added to the participants table in tests. It shows the time that a participant's attempt took in a human-readable format.

Kind regards,
@bidzanaaron

### DIFF
--- a/components/ILIAS/Test/src/Participants/ParticipantTable.php
+++ b/components/ILIAS/Test/src/Participants/ParticipantTable.php
@@ -106,6 +106,7 @@ class ParticipantTable implements DataRetrieval
                 'name' => $this->test_object->buildName($record->getUserId(), $record->getLastname(), $record->getFirstname()),
                 'login' => $record->getLogin(),
                 'matriculation' => $record->getMatriculation(),
+                'total_time_on_task' => $record->getAttemptOverviewInformation()?->getHumanReadableTotalTimeOnTask() ?? '',
                 'status_of_attempt' => $this->lng->txt($status_of_attempt->value),
                 'id_of_attempt' => $record->getAttemptOverviewInformation()?->getExamId(),
                 'ip_range' => $record->getClientIpTo() !== '' || $record->getClientIpFrom() !== ''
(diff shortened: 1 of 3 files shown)


dsstrassner:
@thojou Sure about this solution? If I took the PR and install from it, I got a column working time. But the contents of the column "Status of Attempt" is wrong, as I finished the attempt as participant: 

image

kergomard (changes requested):

I've just one, but rather all encompassing change request:
- [ ] Please do not use `$test->getCompleteEvaluationData()` here, but add the information you need (including the retrieval in human readable time) to the Participant-Information. Also, I don't think this is the working-time of the attempt, it should be the total working time (I usually call this "total time on task").

Thanks and best,
@kergomard

aaronbidzan:
Hey @kergomard,

I have implemented your requested changes. I have changed the column key to “total_time_on_task”, would you also want a new language variable to go with it, or is the “working_time” language variable fine here?

Furthermore, I have also removed the `getCompleteEvaluationData` since this is unnecessary here.

Kind regards,
@bidzanaaron

kergomard:
Thank you very much @bidzanaaron !

Don't worry about the language variable: My reference to "time on task" was more for you, when you go looking.

This looks cleaner, but you will need to make it worse again: The time you are showing is the time spent on a single pass, what we want is the time spent on all passes. I think you will need to introduce some infrastructure to do this. The `ParticipantRepository` might help you though.

Thanks again and best,
@kergomard

aaronbidzan:
Hello @kergomard,

I have taken your feedback into account and pushed a new version. I have found the `getTimeOnTask()` function in the Participant Data, which already does the lifting in terms of adding up all the working times.

Is this fine, or should I do anything different in handling this?

Kind regards,
@bidzanaaron

dsstrassner:
@kergomard Please have another look ;-)

kergomard:
Thank you very much @bidzanaaron !

Ok, this is not perfect, but ok. The problem we have is that the information about the "Total Time on Task" does actually not belong here, as it is not information about the attempt. But the class also contains information of the total attempts, that doesn't belong here either. We need to fix this in a next iteration.

Merged and picked to trunk.

Thank you very much and best,
@kergomard

PR #9070, opened 2025-02-19
Release 10 paths

qualitus-dahme:
Changed Paths in ILIAS10 because of components revision


components/ILIAS/WebAccessChecker/classes/class.ilWACPath.php
```diff
@@ -245,9 +245,9 @@ protected function normalizePath(string $path): string
         $real_data_dir = realpath("./" . self::DIR_DATA);
         $realpath = realpath("." . $original_path);
 
-        if (!str_starts_with($realpath, $real_data_dir)) {
-            throw new ilWACException(ilWACException::NOT_FOUND, "Path is not in data directory");
-        }
+//        if (!str_starts_with($realpath, $real_data_dir)) {
+//            throw new ilWACException(ilWACException::NOT_FOUND, "Path is not in data directory");
+//        }
```
mjansenDatabay, line 250: Hi @qualitus-dahme . Please revert this change.

chfsx: @qualitus-dahme I reverted this change now, please do not push such changes for several reasons:

- changes in components of other authorities needs a PR assigned to the authority (unless trivial changes, but removing an exception isn't trivial)
- throwing exceptions usually has a reason, you cannot just remove this since this could lead to a security issue, especially in places where security is the key.

PR #9072, opened 2025-02-19
Test/FinishPassesCron: Fix object existence check + Migrate field type

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=44237

I pushed **two** commits to the branch because there were two different issues here:

- Database field `enable_processing_time` was of type `varchar`, which resulted in `!== 1` checks being `true`. This PR suggests to migrate to `tiynint` for this boolean state.
- Aborting the process if the test object **existed** was simply wrong (no offense, it's just a forgotton `!`). It has to be aborted if the object does **not** exist.

If approved, these two commits must be picked to all other relevant branches.


kergomard:

This is not an offense, it is just a simple fact! ...I hesitated a second because of the type conversion of the column, but as far as I can see this should run perfectly and it makes things just this little bit better that makes a big difference!

Merged and picked to trunk.

Best,
@kergomard

PR #9076, opened 2025-02-19
Fix static URL handler

mjansenDatabay:

If approed, this must be picked to all relevant branches.


chfsx:
Thanks a lot! Merged and picked

PR #9077, opened 2025-02-19
Fix wrong base class for workspace

mjansenDatabay:


chfsx:
Thanks a lot, merged and picked

PR #9081, opened 2025-02-20
Fix capability permission handling in workspace context

mjansenDatabay:

If approved, this must be picked to all relevant braches.

### DIFF
--- a/Modules/File/classes/Capabilities/Check/BaseCheck.php
+++ b/Modules/File/classes/Capabilities/Check/BaseCheck.php
@@ -38,17 +38,19 @@ abstract class BaseCheck implements Check
         Context $context,
         Permissions ...$permission
     ): bool {
-        $permission_string = implode(
-            ',',
-            array_map(static fn(Permissions $permission) => $permission->value, $permission)
-        );
-        if ($context->getContext() === Context::CONTEXT_WORKSPACDE) {
-            return $helpers->workspace_access_handler->checkAccess(
-                $permission_string,
-                '',
-                $context->getCallingId(),
-                'file'
-            );
+        if ($context->getContext() === Context::CONTEXT_WORKSPACE) {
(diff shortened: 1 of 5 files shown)


Modules/File/classes/Capabilities/Check/BaseCheck.php
```diff
@@ -38,17 +38,19 @@ protected function hasPermission(
...
-            array_map(static fn(Permissions $permission) => $permission->value, $permission)
-        );
-        if ($context->getContext() === Context::CONTEXT_WORKSPACDE) {
-            return $helpers->workspace_access_handler->checkAccess(
-                $permission_string,
-                '',
-                $context->getCallingId(),
-                'file'
-            );
+        if ($context->getContext() === Context::CONTEXT_WORKSPACE) {
```
mjansenDatabay, line 41: As far as I can see the workspace access handler cannot handle a `,`-separated list of permissions. For the `Info` capability in the workspace context the permission string looked like this:

```
visible,read,view_content,read_learning_progress,edit_learning_progress,edit_permission,write,delete,copy,edit_file
```

An access check with this permission will always result in a boolean `false`.

Modules/File/classes/class.ilObjFileGUI.php
```diff
@@ -319,7 +319,12 @@ public function executeCommand(): void
                     $this->addHeaderAction();
                     $ilTabs->clearTargets();
 
-                    parent::executeCommand();
+                    if (empty($cmd) || $cmd === 'render') {
+                        $cmd = Capabilities::INFO_PAGE->value;
+                        $this->$cmd();
```
mjansenDatabay, line 324: If we don't do this, the parent class will automatically call a `render` function, which is not given in `ilObjFileGUI` and lead to a crash.

PR #9083, opened 2025-02-20
[FIX] use revision title instead of filename

chfsx:
@alex40724 I suggest to use the revision title instead of the filename to avoid the issue https://mantis.ilias.de/view.php?id=44135


alex40724:
Thanks for the PR. I set the zip name now correctly. The code changes of this PR lead to changes in the XML export file listings.

PR #9085, opened 2025-02-20
Test: Fix Removal of Inline Images

mBeym:
The fix https://github.com/ILIAS-eLearning/ILIAS/commit/fa0282e08e6f41311d4d2a85dc5b0b94242d6dbc added in the past by @kergomard is missing in ILIAS versions >= 9.

Causing the same issue as reported in the past: https://mantis.ilias.de/view.php?id=36264

If accepted, should be picked into ILIAS versions 10 & trunk as well.

### DIFF
--- a/Modules/TestQuestionPool/classes/class.ilAssHtmlPurifier.php
+++ b/Modules/TestQuestionPool/classes/class.ilAssHtmlPurifier.php
@@ -41,6 +41,15 @@ abstract class ilAssHtmlPurifier extends ilHtmlPurifierAbstractLibWrapper
         $config->set('HTML.Doctype', 'XHTML 1.0 Strict');
         $config->set('HTML.AllowedElements', $this->getAllowedElements());
         $config->set('HTML.ForbiddenAttributes', 'div@style');
+        $config->autoFinalize = false;
+        $config->set(
+            'URI.AllowedSchemes',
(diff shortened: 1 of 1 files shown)


dsstrassner:
Is there an actual mantis report @mBeym? 

Strange that this commit is missing in ILIAS 9, as @kergomard stated, that he has picked it to trunk?¿?

mBeym:
> Is there an actual mantis report @mBeym?
> 
> Strange that this commit is missing in ILIAS 9, as @kergomard stated, that he has picked it to trunk?¿?

not for this issue still being present in ILIAS 9+ no.

I can create one if you want.

dsstrassner:
> > Is there an actual mantis report @mBeym?
> > Strange that this commit is missing in ILIAS 9, as @kergomard stated, that he has picked it to trunk?¿?
> 
> not for this issue still being present in ILIAS 9+ no.
> 
> I can create one if you want.

I think, we can reopen the old one, with a link to a test on test9, that the issue still exists in ILIAS 9 and change the product version.

kergomard:

Yes, this looks simply like something went wrong when picking and I didn't notice.

Merged and (this time really, really) picked to ILIAS 10 and trunk.

I added a comment in the old issue and change the version.

Best,
@kergomard

PR #9090, opened 2025-02-21
Dashboard: fix minor bugs within the administration

iszmais:
https://mantis.ilias.de/view.php?id=43951
https://mantis.ilias.de/view.php?id=42751
https://mantis.ilias.de/view.php?id=44050

### DIFF
--- a/components/ILIAS/Dashboard/Administration/classes/class.ilDashboardSortationTableGUI.php
+++ b/components/ILIAS/Dashboard/Administration/classes/class.ilDashboardSortationTableGUI.php
@@ -28,7 +28,7 @@ class ilDashboardSortationTableGUI extends ilTable2GUI
     private ilPDSelectedItemsBlockViewSettings $viewSettings;
     private ilDashboardSidePanelSettingsRepository $side_panel_settings;
 
-    public function __construct($a_parent_obj, $a_parent_cmd)
+    public function __construct($a_parent_obj, $a_parent_cmd, bool $disable = false)
     {
         global $DIC;
         $this->uiFactory = $DIC->ui()->factory();
@@ -39,13 +39,15 @@ class ilDashboardSortationTableGUI extends ilTable2GUI
         $this->lng->loadLanguageModule('mme');
         $this->initColumns();
         $this->setFormAction($this->ctrl->getFormAction($this->parent_obj));
-        $this->addCommandButton('saveSettings', $this->lng->txt('save'));
+        if (!$disable) {
+            $this->addCommandButton('saveSettings', $this->lng->txt('save'));
+        }
         $this->setRowTemplate(
             'tpl.dashboard_sortation_row.html',
             'components/ILIAS/Dashboard'
         );
         $this->setEnableNumInfo(false);
-        $this->initData();
+        $this->initData($disable);
     }
 
     public function initColumns(): void
@@ -55,7 +57,7 @@ class ilDashboardSortationTableGUI extends ilTable2GUI
         $this->addColumn($this->lng->txt('topitem_active'));
     }
 
-    public function initData(): void
+    public function initData(bool $disable = false): void
     {
         $data[] = [
             'position' => $this->uiRenderer->render(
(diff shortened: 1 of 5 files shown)


components/ILIAS/Dashboard/Administration/classes/class.ilDashboardSortationTableGUI.php
```diff
@@ -55,7 +57,7 @@ public function initColumns(): void
         $this->addColumn($this->lng->txt('topitem_active'));
     }
 
-    public function initData(): void
+    public function initData($disable = false): void
```
lscharmer: Missing type for parameter `$disable`.

components/ILIAS/Dashboard/Administration/classes/class.ilDashboardSortationTableGUI.php
```diff
@@ -71,12 +73,13 @@ public function initData(): void
             $presentation_cb->setChecked($this->viewSettings->isViewEnabled($presentation_view));
             $presentation_cb->setValue('1');
             $presentation_cb->setDisabled(
-                $presentation_view === ilPDSelectedItemsBlockConstants::VIEW_RECOMMENDED_CONTENT
+                $disable ?? ($presentation_view === ilPDSelectedItemsBlockConstants::VIEW_RECOMMENDED_CONTENT)
```
lscharmer: Do you mean `?:` instead of `??`?. If I see this correctly `$disable` cannot be nullable, as `ilNumberInputHUI::setDisabled` requires a non nullable boolean.
This means `$disable ?? 'foo'` will never result in `'foo'`, because `??` only "reacts" on `null`

lscharmer (changes requested):

components/ILIAS/Dashboard/Administration/classes/class.ilDashboardSortationTableGUI.php
```diff
@@ -71,12 +73,13 @@ public function initData(): void
             $presentation_cb->setChecked($this->viewSettings->isViewEnabled($presentation_view));
             $presentation_cb->setValue('1');
             $presentation_cb->setDisabled(
-                $presentation_view === ilPDSelectedItemsBlockConstants::VIEW_RECOMMENDED_CONTENT
+                $disable ?: ($presentation_view === ilPDSelectedItemsBlockConstants::VIEW_RECOMMENDED_CONTENT)
```
lscharmer, line 76:
```suggestion
                $disable || ($presentation_view === ilPDSelectedItemsBlockConstants::VIEW_RECOMMENDED_CONTENT)
```

Just noticed this now :o

lscharmer (changes requested):

components/ILIAS/Dashboard/Administration/classes/class.ilDashboardSortationTableGUI.php
iszmais: I would still keep the `?:` due to the minor consideration that those 2 values are not logically connected but more functionally resulting in each other. E.g. would it be very unintuitive to flip the values which whould be a possibility the `||` would imply.

But as said thats more of a preference

iszmais: Or, for example, the `$presentation_view` logically is not required to be set, when a input is disabled anyways. This way i ensure this does not resolve into any conflict for such requests.

Or in other words: `?:` is more "strict" than `||`

lscharmer: I disagree, `||` does not imply that you can switch them, because the later is never evaluated if the former is `true`, the same with `&&`: You cannot switch `$foo !== null && $foo->bar()` just as you cannot switch `$foo === null || $foo->bar()` around.

iszmais: Yes, im aware that those are not implicitly flippable.

What I tried to say is that most code assistent software provides optional suggestions for flipping those, which it does not for "shortened ifs".
This has its origin within the mathematically translation which differs between logical combinations (in which both combining parts are equal in their weight) and logical conclusions (in which 1 value results into another, but not obliged otherwise).

And since this is a logical conclusion ("Set Disable to XY, if the flag 'disabled' is not set") and no logical combination (Set disabled to the flag 'disabled' or to XY) i would prefer to use a conclusion operator here.

lscharmer: I still disagree, as this is not the promise of `||` and `&&`, this would apply for `&` and `|` (**logical** `or` and `and`) but as this is just nit-picking, it's really no blocker

lscharmer (approved):

PR #9093, opened 2025-02-21
Object: fix id mapping on import (44072)

schmitz-ilias:
Just a small fix for the ID mapping in `ilObjectDataSet`. In specific circumstances it failed, leading to object properties not getting imported, see [44072](https://mantis.ilias.de/view.php?id=44072).


kergomard:
Thank you very much @schmitz-ilias ! Merged and picked to trunk.

PR #9095, opened 2025-02-22
Proposal: raise events for any tree table

padvincenzo:


mjansenDatabay:
This will have an impact on other components. There are many derivates of `ilTree`, other components use a wrapper for `ilTree`. If the event is raised for all the other tree operations in the future, the code in `\ilAppEventHandler::raise` is executed additonally for all the other components.

padvincenzo:
Hi @mjansenDatabay , thanks for your reply. This change will sure affect many components, and that's the proposed behaviour:
* currently none of the other Tree based classes raise events;
* the deleteNode event is raised without the `__isMainTree()` check;
* all the three Tree events has a parameter that represent the tree table, which listeners can use to decide if they actually are interested in the raised event.

Also, but it's not the subject of this PR, the check on eventHandler is different among the three events, is that intended?

mjansenDatabay:
Hi [padvincenzo](https://github.com/padvincenzo),

> * currently none of the other Tree based classes raise events;

This is true.

> * the deleteNode event is raised without the `__isMainTree()` check;

This is also true. `deleteNode` seems to be added later/retrospectively by this [commit](https://github.com/ILIAS-eLearning/ILIAS/commit/b054299207e56f9bc90d76cff4626dfdea69ca62) and does not apply the same checks as the other methods.

> * all the three Tree events has a parameter that represent the tree table, which listeners can use to decide if they actually are interested in the raised event.

This is also true. But when removing the `$this->__isMainTree()` condition, the events will be raised every time (for instance) a folder is added in the mail system (this will result in 5 events and 5 calls to `\ilAppEventHandler::raise` on account creation). At least the loop over and the delegation to the event hook plugins has to be done everytime.

I am not against your suggestion, nor am I the responsible component authority. I only like to emphasize that this suggestion has a "footprint".

> Also, but it's not the subject of this PR, the check on eventHandler is different among the three events, is that intended?

I am not the person who maintains the "Tree" component. But I guess that the events were not introduced at the same time by the same people by the same process.

Best regards,
Michael

fwolf-ilias:
Hi @padvincenzo ,

the inconsistencies aside, is there a use case where those events are needed for a tree outside of the repository tree? 

A change in behaviour due to the complex effects should be carefully considered. Without an issue or a new use case, I wouldn't try to change it. 

Bests
Fabian Wolf

PR #9096, opened 2025-02-24
T&A9 43729: Fixes error when trying to apply personal settings to a test

matheuszych:
[Mantis: 43729](https://mantis.ilias.de/view.php?id=43729)

This should resolve the issue for now. A more indepth fix is not planned, since the feature will be probably removed in ILIAS 11.

### DIFF
--- a/Modules/Test/classes/MainSettings/class.ilObjTestSettingsMainGUI.php
+++ b/Modules/Test/classes/MainSettings/class.ilObjTestSettingsMainGUI.php
@@ -539,13 +539,18 @@ class ilObjTestSettingsMainGUI extends ilTestSettingsGUI
 
     private function saveAvailabilitySettingsSection(array $section): void
     {
-        $timebased_availability = $section['timebased_availability'];
-        if ($this->test_object->participantDataExist()) {
-            $timebased_availability['is_activation_limited'] = $this->test_object->isActivationLimited();
-            $timebased_availability['activation_starting_time'] = $this->test_object->getActivationStartingTime();
-        }
+        $time_based_availability = $section['timebased_availability'];
 
-        $this->test_object->storeActivationSettings($timebased_availability);
+        $this->test_object->storeActivationSettings(
+            $this->test_object->participantDataExist()
+                ? $this->test_object->isActivationLimited()
+                : $time_based_availability['is_activation_limited'],
+            $this->test_object->participantDataExist()
+                ? $this->test_object->getActivationStartingTime()
+                : $time_based_availability['activation_starting_time'],
+            $this->test_object->getActivationEndingTime(),
+            $this->test_object->getActivationVisibility(),
+        );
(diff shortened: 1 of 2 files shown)


matheuszych:
Already reviewed by @thojou.

kergomard:

I don't think this is the right place to solve this issue though. This is a problem in the personal default settings not in `ilObjTest::storeActivationSettings()`. Please move the corresponding checks there. I would also kindly ask you to change the parameters of `ilObjTest::storeActivationSettings()` to accept all values separately and not as an array (mind the usage in `SettingsMainGUI`). This way, this is less confusing in the future.

Thanks again and best,
@kergomard

matheuszych:
Hello @kergomard,
which class do you mean with
>Please move the corresponding checks there.

kergomard:
I mean the function `ilObjTest::applyDefaults()`

matheuszych:
@kergomard 
I adjusted the code. Can you please re-review the changes?

kergomard:

We are getting there ;-): The `?? false` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9096/files#diff-a293cd08c2c060fd8ed019c010e376e3314cbd4c65f85d2e54fbd4488176937eR547) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9096/files#diff-a293cd08c2c060fd8ed019c010e376e3314cbd4c65f85d2e54fbd4488176937eR550) are unnecessary/wrong as far as I can see. We are using a transformation to ensure these fields always exist. This is done on purpose, so we don't need these kinds of dangerous checks (as shown in your case where suddenly a time is `false`).

Thanks and best,
@kergomard

matheuszych:
Ah ok. I just like to be really cautious when accesing any array keys. In this case, the intention was to use the default value that `storeActivationSettings` expects for the according parameter in case the key was not set.

I removed the check, since the transformation already ensures that they are always set.

kergomard:

I very much understand your mistrust of arrays, that is why they should, in most places, leave us as quickly as possible. Why I don't like the approach to just protect all array access:
- You are introducing places where "defaults" are set. These defaults are not authoritative though and would need to be changed, if we change the real default.
- We should always know what we are dealing with. the protection is just a very strong smell, that you have no idea, what you are dealing with. In this case it is worth the bother to look around and see, if we can figure it out. So the protections in `ilObjTest::applyDefaults()` is correct, as we really do not have any idea, what we are dealing with. The one in `ilSettingsMainGUI` on the other hand isn't, as here it would introduce a smell, where it isn't strictly needed (an improvement here would be to create an `Activation`-object in the trafo that we then can pass around, but this would need some more thought and is probably not a thing the test should take care of, but me in my role as responsible for `ilObject`. I hope this all makes sense.

One more small side note: Please do not force push, but keep the history intact. It makes my reviews easier.

Thanks again and best,
@kergomard

PR #9101, opened 2025-02-24
10/UI/tweaks drilldown

kergomard:
Hi @Amstutz and all

I assign you for the time being, as this is mostly css. I also added a small tweak in the js of the drilldown adding a resize listener to recalculat fixed element height and width when the window is resized, but this should be trivial enough.

So, there are three commits here:
1. @yvseiler has added some improvements to the presentation of the drilldown that I've reviewed and expanded to cover some edge cases.
2. The mentioned listener in drilldown. I did a little bit of boyscouting, hopefully easy enough to spot, as the the `drilldown` was called "dropdown" in multiple places. I hope that is still ok.
3. A fix for a problem where the leaves of the menu-tree where displayed smaller than the branch-nodes (see: https://mantis.ilias.de/view.php?id=44278).

I attache a before and an after video to showcase the changes.
Before:
https://github.com/user-attachments/assets/a751846a-25da-4b00-878b-81ff57a95cb5

After:
https://github.com/user-attachments/assets/fb94cd9b-5128-4a6a-a136-2c1d394f1fbe

Thank you very much and best,
@kergomard


Amstutz:
thx alot @yvseiler and @kergomard for the improvements!

PR #9109, opened 2025-02-25
T&A Bugfix: Import of Tests with Images takes exponentially longer than normal

fhelfer:
Solves the issue that came along with https://mantis.ilias.de/view.php?id=42016

The Problem with Test Imports and Images was that the thumbnails that were generated were always generated with Maximum Thumb Size. This would make all Images in the imported test scaled to much higher than 8k (8192x8192)

```php
        if ($size < $this->object->getMaximumThumbSize() // 8192) {
            return $this->object->getMaximumThumbSize();
        }
```

@kergomard solved this unintentionally with commit https://github.com/ILIAS-eLearning/ILIAS/commit/8b66c0cdf2127cecd8caf5b39cace0f37e1d2665 which removed the line.
But since this line was faulty and could be fixed by removing the logic typo i re-added the line (now correctly)
which should also still work with the Bugfix addressed by the commit.

Best @fhelfer

### DIFF
--- a/components/ILIAS/TestQuestionPool/classes/import/qti12/class.assQuestionImport.php
+++ b/components/ILIAS/TestQuestionPool/classes/import/qti12/class.assQuestionImport.php
@@ -374,6 +374,10 @@ class assQuestionImport
             return $this->object->getThumbSize();
         }
 
+        if ($size < $this->object->getMinimumThumbSize()) {
+            return $this->object->getMinimumThumbSize();
+        }
+
         if ($size > $this->object->getMaximumThumbSize()) {
             return $this->object->getMaximumThumbSize();
         }


fhelfer:
please also cherry pick to release_8 & release_9 and trunk as well

kergomard:

I removed the line very intentionally, but you are right, we can check for minimum size and things get better ;-). Merged to ILIAS 9 and trunk.

Best,
@kergomard

dsstrassner:
> please also cherry pick to release_8 & release_9 and trunk as well

@kergomard couldn't pick this to ILIAS 8. But let me ask @fhelfer: Interested users may test this in their ILIAS 8 installations if they use 8?

PR #9111, opened 2025-02-25
ErrorHandling/SOAP: Use SOAP exception handler for HTTP POST only

mjansenDatabay:
The SOAP exception handler should only be used for HTTP POST requests, even if `ilContextSoap` is given

See: https://mantis.ilias.de/view.php?id=44363

If approved, this must be picked to `release_10` and `trunk`.


jeph864:

PR #9112, opened 2025-02-25
SOAP: Fix use of deprecated PHP filter

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=44362

If approbed, this must be picked to `trunk` as well.


jeph864:
thanks @mjansenDatabay

PR #9123, opened 2025-02-26
UI: Remove leftover parameter in Form\Factory test

tbongers-cat:
This just removes a leftover parameter from the Form\Factory test that is not needed anymore.


Amstutz:
thx a lot

PR #9126, opened 2025-02-27
UI LatexContent

fneumann:
This PR proposes to add a new component [Legacy\LatexContent](https://github.com/fneumann/ILIAS/blob/dev11-streamline-latex-ui-modification/components/ILIAS/UI/src/Component/Legacy/LatexContent.php) to the UI framework. It is related to the corresponding feature request [Streamline LaTeX usage](https://docu.ilias.de/go/wiki/wpage_5614_1357) . It relpaces the [closed former PR](https://github.com/ILIAS-eLearning/ILIAS/pull/8261).

`Legacy\LatexContent` is used as an intermediate solution for legacy content with mathematical formulas that are rendered by the JavaScript library MathJax. The LaTeX code of these formulas must be wrapped in the delimiters `[tex]` and `[tex]`.

The modification of page content by MathJax is controlled by CSS classes. The class `tex2jax_default_ignore` is added to the body element of a page to prevent a modification outside `Legacy\LatexContent`. These contents are wrapped in elements with a class `tex2jax_process` to enable the modification inside.

`Legacy\LatexContent` registers a JavaScript file `mathjax.js` on the page. This file configures and loads MathJax asynchronously. MathJax scans the page for content to be processed and subsequently loads additional modules. All resources that are needed by MathJax are defined by the class `DefaultLatexResources` implementing the interface `LatexResources`. It is used by the UI framework to populate the public folder with these resources, and it is offered to other Components by a 'Define/Use' integration to support the export of the resources for offline content.

The `Legacy\Factory` is added to provide separate renderers for the legacy components `Content` and `LatexContent`. This way the resources for MathJax are only added to pages that use `LatexContent`. However, these are stilly many pages because the content of the ILIAS page editor needs to be wrapped in a `LatexContent` component. 

Compared to the [previous pull request](https://github.com/ILIAS-eLearning/ILIAS/pull/8261), this new one does not inject a configuration to the UI framework by ILIAS. That means there is no setup switch to prevent the inclusion of the MathJax library on the page once a `LatexContent` is included, event if it does not have a LaTeX formula at all.


fneumann:
Dear UI Coordinators, please find here the new approach for moving MathJax to the UI framework. How should we proceed? Would you like to have a look at this PR before I put the feature Request [Streamline LaTeX Usage](https://docu.ilias.de/go/wiki/wpage_5614_1357) on the JourFix Agenda?

Amstutz:
Hi Fred,

We just discussed the among the UI Coordinators. The Public Interface is ready to be discussed in the upcoming JF. Thx a lot. The label is attached.

Please note, that we are also working on the review of the implementation, there are some observations and point to do.

Further note, that issues connected to this Latex Content component will need to be processed by you according to the Coordinator Model.

Thx for you work

@klees, @thibsy and @Amstutz

matthiaskunkel:
Jour Fixe, 31 MAR 2025: We highly appreciate this PR and accept it for trunk.

fneumann:
Dear UI coordinators, the JF acceptance is only related to the public interface of the new component. I'm still waiting for your review of the implementation and need the final acceptance to create separate PRs for the components that use MathJax. Especially a provision of interface LatexResources would be important to implement to use these resources in an HTML export of learning modules.

thibsy (changes requested):
Hi @fneumann,

We will refer to the public interface `ILIAS\UI\Component` as `C` and to the implementation `ILIAS\UI\Implementation\Component` as `I` and `ILIAS\UI\Implementation\Render` as `R`.

- [x] `mathjax.js` I: we highly discourage to load assets like this. Please check if (a) the Mathjax asset MUST be loaded in `async` and (b) if we cannot simply register these assets in the according order in our `R\RessourceRegistry`, so some `mathjax.js` with the configuration (only) is added before the Mathjax assets are.

- [x] `R\LatexResources`: please remove this interface and its default implementation. We either fully embrace latex or not at all. Therefore, please contribute all assets explicitly as `ILIAS\Component\Resource\NodeModule`s in `UI.php` and register them in the `R\ResourceRegistry` in the renderer individually as well. We understand the problem you are trying to solve here, but this solution only works for `C\Legacy\LatexContent` components. We therefore recommend to search for an alternative solution, e.g. extending the `R\ResourceRegistry`, which works for all of the components. Addressing this issue is out of this PR's scope, so we suggest to open a mantis issue for the time being.
- [ ] `I\Legacy\LatexRenderer` I: please merge this renderer with the existing `I\Legacy\Renderer` and remove the factory again. This measure is unnecessary, since we strive to bundle all of our JavaScript logic into one file in the future anyways. `withLatexResources()` is no longer needed as well.
- [x] `I\Legacy\LatexRenderer` II: please introduce a template file for the `C\Legacy\LatexContent` components instead of using inline HTML and set their content using an `R\Template`.
- [ ] `mathjax.js` II: if we understand the documentation of Mathjax correctly, it should be possible to add the `<body>` HTML tag to `skipHtmlTags` and get the same result, but without an additional CSS class on every page.
- [x] `mathjax.js` III: please adhere to our [SCSS guidelines](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/templates/Guidelines_SCSS-Coding.md) and use e.g. `c-legacy__content--latex` as enabling class.
- [x] PHP code-style: some of your files do not correspond with our code-style because they do not end with a line-break.

Kind regards,
@Amstutz, @klees and @thibsy

thibsy:
@fneumann please reassign me once this is ready for another round.

fneumann:
Dear @Amstutz, @klees and @thibsy,

thanks for your in-depth review. I was able to address most of your points, see my comments below. The code is revised and ready for a re-review.

- `mathjax.js` I: The combination of mathjax configuration and async loading in one script was recommended in the [MathJax documentation](https://docs.mathjax.org/en/latest/web/configuration.html#configuring-and-loading-in-one-script) because the main MathJax script (`tex-chtml-full.js`) has a minified size of 304 kB. It is now added separately with synchronous loading. Please note that MathJax loads additional assets (see note on `R\LatexResources`).

- `R\LatexResources`: The `node_modules/mathjax/es5` directory has over 106 files. We support only a standard configuration provided by `tex-chtml-full.js` which is registered as a resource. Other assets are loaded by MathJax on demand when Latex expressions are rendered. We don't know which of them will be needed, so we have to provide all sub directories of es5 as public assets. This is now done in `UI.php` with an anonymous function. `R\ResourceRegistry` can't be used since it can't handle directories or the file types .json and .woff. The interface `I\LatexResources` and the class `R\DefaultLatexResources` were an attempt to provide the selection for the export of ILIAS learning modules. I deleted them. The learning module component will have to copy `mathjax.js` and everything under `public/node_modules/mathjax/es5` to create a self-contained HTML export with MathJax support.

- `I\Legacy\LatexRenderer` I: in the JF on 31 March 2025 we decided not to optionally deactivate MathJax in the setup. This decision was drawn based on the assumption that MathJax will only be added to pages that use the `C\Legacy\LatexContent` component, respectively pages from the ILIAS page editor or pages displaying test question. Since the `registerResources` function does not get the rendered component as a parameter, we can only make that difference by a separate renderer. Currently there are four other UI components using an own renderer factory, so this is not an exotic pattern. Are there actual plans for ILIAS 11 to bundle all Javascript in one file? Otherwise I recommend to use a separate renderer for `C\Legacy\LatexContent` as long as we have to process Latex for the legacy content in in the page editor and test.

- `I\Legacy\LatexRenderer` II: done.

- `mathjax.js` II: This is not possible. The `skipHtmlTags` generally blocks rendering for certain tags. It cannot be changed with an enabling class for sub elements. So we have to provide a disabling class (see next point).

- `mathjax.js` III:  I now use `c-layout__page--latex` to ignore latex on page level and `c-legacy__content--latex` to proces latex on component level. Both are fixed attributes in the template files. 

- PHP code-style: applied the PHP CS Fixer to all modified PHP files.

Best regards
Fred

fneumann:
Dear @Amstutz, @klees and @thibsy,

the PR is ready for a next round. Could one of you have alook?

Thanks in advance
Fred

thibsy (changes requested):
Hi @fneumann,

Thanks a lot for removing some of the complexity and for looking into the package configuration. It's a pity that we cannot get rid of the class on our body tag.

Please consider the following suggestions. You do not need to follow them, but please indicate why you still prefer to do otherwise:

- [x] `C\Legacy\LegacyRendererFactory`: there are plans to bundle all of our JavaScript code into one file. We noticed again in #7143 the need for such a bundle. Therefore, we highly suggest to remove this renderer factory and provide the Mathjax assets on every page. This way we can get a good feeling on the actual performance implications. We could still address them in a usability bugfix later.

- [x] Unit tests: please implement some relevant PHPUnit test cases and stick to our [guidelines](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/UI/README.md#How-to-write-unit-tests-for-a-Component).
- [x] `c-layout__page--latex`: please remove the `--latex` modifier from this class name. This implies that latex is allowed, when in fact, it is prohibited.
- [x] Example: please try to be more descriptive in your example description. This will be the basis for our Testrail test cases. Assume testers don't know how this should look like.

Kind regards,
@Amstutz, @klees and @thibsy

fneumann:
Dear @Amstutz, @klees and @thibsy,

I implemented your requested changes. Regarding the `C\Legacy\LegacyRendererFactory` I can't add something new to what I alredy explained. To not further block the creation of PRs for the consuming ILIAS components, I removed the separate renderer and the factory. Mathjax will now be added to every page having a legacy component. We well see if someone complains.

Please have a hopepully final review of the changes.  

Best regards
Fred

thibsy (approved):
Thx @fneumann for bearing with us and implementing all of the changes now.

Changes LGTM, so I will go ahead and merge this.

Kind regards,
@thibsy

fneumann:
Thanks @thibsy for merging this PR. This brings us a big step forward. Now the way is open to use the UI element and to remove the old MathJax service from the system.

PR #9127, opened 2025-02-27
T&A Bugfix: 44352 ErrorText Questions accept 0 for points

okaufman:
This PR will fix https://mantis.ilias.de/view.php?id=44352

If a user enters '0' for points, the message "Please enter only positive numbers." will pop up when trying to save.


kergomard:
Thanks @okaufman for the PR! Merged and picked to trunk.

PR #9131, opened 2025-02-27
DidacticTemplate: Adjust xsd to include 'excludePattern'

chlulei:
Link: https://mantis.ilias.de/view.php?id=44187


smeyer-ilias (approved):
Thanks @chlulei

PR #9136, opened 2025-02-28
T&A Fix missing variable in TestOutputGUI

thojou:
This PR is related to an issue report via the Discord channel.
The user posted the following screenshot:

image

Best,
@thojou


kergomard (approved):
Thank you very much for the PR @thojou 

As this is a picking error I introduced two days ago, I will merge without a mantis ticket.

No picking to other versions needed as this is ILIAS 9 only.

Thanks again and best,
@kergomard

PR #9138, opened 2025-02-28
T&A Fix instant_response picking error

thojou:
Hey,

This PR is related to the Mantis Ticket https://mantis.ilias.de/view.php?id=44411
It's a followup for PR #9136 and Mantis Ticket https://mantis.ilias.de/view.php?id=44409

I adjusted the code according to the implementation in release_10

Best,
@thojou


kergomard (approved):
Thank you very much @thojou ! This is ILIAS 9 only then and needs no picking. Merged.

PR #9139, opened 2025-02-28
Exercise: Fix ZIP uploads for Windows clients

mjansenDatabay:
Mantis Issue: https://mantis.ilias.de/view.php?id=41722

If approved, this should be picked to all relevant branches.


alex40724:
Thanks for the PR. Fixed it a little bit different.

PR #9142, opened 2025-03-03
PRG: 44216, get selected standard userfields during assignment

nhaagen:
https://mantis.ilias.de/view.php?id=44216


klees:
Picked to trunk.

PR #9145, opened 2025-03-03
OrgU: 44010, check createable subobjects for empty

nhaagen:
https://mantis.ilias.de/view.php?id=44010
https://mantis.ilias.de/view.php?id=44188


klees:
Picked to trunk.

PR #9146, opened 2025-03-04
OrgU: 44453, get _all_ orgus for getOrgUnitsByUserAndPosition

nhaagen:
https://mantis.ilias.de/view.php?id=44453


klees:

PR #9155, opened 2025-03-05
issue #44466

sroenchen:
Fix evaluation failing if no usr_id is given and add fallback to user title


kergomard:
Thank you very much for the PR @sroenchen !
I close this though as `usr_id === null`  is used to detect deleted users in the function and the function does accept null as a parameter value.

Thanks again and best,
@kergomard

PR #9158, opened 2025-03-05
Calling ilObjTest->getActiveIdOfUser in a cron/cli enviroment as anonymous user no activeId is returned

PurHur:
In a cron/cli enviroment without explicit user context the  `$this->user` is the anonymouos user not the passed user of the passed user_id. The fallback user_id is fetched in the first lines of the method so this `$user_id` should be used.

Mantis https://mantis.ilias.de/view.php?id=44475


kergomard:
Thank you for the PR  @PurHur 

Please fix the code-style violation. I think in general the change is correct though.

Best,
@kergomard

kergomard:
Thanks for the fix @thojou !

PR #9164, opened 2025-03-05
10/contentPage/right-column panel breakpoints 44276

catenglaender:
https://mantis.ilias.de/view.php?id=44276
https://mantis.ilias.de/view.php?id=43744
https://mantis.ilias.de/view.php?id=44394

# Issue

Break behavior in panels inside sidebars is not optimal. Viewcontrols and titles extend over panel borders. Side panels do not stack under content on mobile.

# Changes

Stringed container query up higher on the center_col, so both header and viewcontrols are inside of the container. Thinned out code.

Dashboard_Content-Breaks

# Lessons learned

Container Query containers should never be expected to push elements around them or grow and shrink with the content inside them. They don't do either because they are their own independent DOM context. Neither DOM context can effect anything outside or inside the other.

The solution is to define containers as high up as possible (e.g. main content and sidebar are fine as containers but panel-viewcontrols are not).

### DIFF
--- a/templates/default/050-layout/_layout_container-query.scss
+++ b/templates/default/050-layout/_layout_container-query.scss
@@ -14,10 +14,8 @@
 /// @prop {Numbers} wide-card switch to slightly condensed card/item design (e.g. for 2 column grid) below
 /// @prop {Numbers} toolbar switch an entire toolbar to a condensed design below
 $container-query-contexts: (
-    compact-card: 300px,
     sidebar: l.$il-standard-page-slate-width,
-    wide-card: 600px,
-    toolbar: 600px,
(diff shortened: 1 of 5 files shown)


Amstutz:
Hi @catenglaender 

Thx you very much for fixing the issues here. We already discussed the need for the important statements here in the CSS squad, so they are a valid exception.

There is one thing:

- [x] Doc Comment: You adapt the params in the $container-query-contexts map. Please also adapt the doc comment just above the map. Thx.

If done, this can be merged and picked to trunk as well.

Amstutz:
Ill did it after merging.

PR #9165, opened 2025-03-06
PRG: 44477 (44216): get userfields during assignment

nhaagen:
https://mantis.ilias.de/view.php?id=44477


klees:
Picked to trunk.

PR #9166, opened 2025-03-06
Mantis 44504: No write access to Users via Position Access

ingoj:
Possible Solution for mantis 44504


kergomard:
Thank you very much for the PR @ingoj !
I fixed it a little bit differently though:
- I don't think it is wrong to give write access to the users in LocalUserAdministration.
- I avoided a class property (`$this->access`), ...that would have been undeclared.
- I did a little bit of boy-scouting. 

Fixed with https://github.com/ILIAS-eLearning/ILIAS/commit/88b46ec610e745cb9e72b4a3fd26e4910f578918 and picked to ILIAS 10 and trunk.

Thanks again and best,
@kergomard

PR #9169, opened 2025-03-07
UI Mantis #0044501: Fix Breadcrumb Arrow being bold

fhelfer:
https://mantis.ilias.de/view.php?id=44501


yvseiler:
Hi @fhelfer 

We need to be aware that with this change there will be an adjustment in all browsers, not just Safari. However, Safari is the one that displays the glyph bolder than Chrome, for example (the glyph will also be narrower there due to the change). 

I have looked at the visual customization for Safari, Firefox and Chrome and it works for me.

kindly,
@yvseiler

PR #9173, opened 2025-03-07
ECS: Fix building "Agree Date" when creating new user

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=43632


Akumatic:
Thanks for the PR. Pascal has made a few corrections for our local instance, which I pushed earlier - this line was also part of it & is already merged (6cb13a8) and picked for 10 and Trunk.

PR #9174, opened 2025-03-10
Export: Add function to ilExport to get path to 'expDir_' in export container, adjust ilDataset to use the new function

chlulei:
Link: https://mantis.ilias.de/view.php?id=44445


smeyer-ilias (approved):
Thanks @chlulei !

PR #9175, opened 2025-03-10
ECS: Fixing accessing undefined key when reading assignments in cronjob

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=44511


Akumatic:
Thanks for fixing. I will incorporate changes from 9 later that have not yet been transferred to 10 and trunk and then cherry-pick the commit

PR #9178, opened 2025-03-11
SystemCheck/Repository/Tree: Fix PHP 8.x issues

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=44533


alex40724:
Thanks for the PR.

PR #9180, opened 2025-03-11
_10/UI/Chart/ProgressMeeter-match-goal-arrow-and-bar_43602

catenglaender:
Mantis: https://mantis.ilias.de/view.php?id=43602

# Issue

Meter needles and bars do not match. For example, a 50% bar extends beyond a 52% needle.

image

# Changes

Wow, this was much more tricky than I initially thought and the code looks like someone was already battling with this before me. The problem is that the mathematically correct position of the bar and the needle arrow are both off for two reasons:
- line ends in vector svgs are perfectly straight and when the line ends they are capped off with a half circle that extends beyond their actual length. This buttcap adds another +4% visually.
- the meter is not a circle but slightly squashed. This means that just rotating the needle is off the further you get away from 0 degrees.

So I had to add two compensations:
1. subtract 4% from the length value of the bar (monocircle only)
2. the further left and right you go from 50%, some degrees of rotation are added or subtracted respectively. This needed to be done for both the needle arrow and the mini meter pointer.

image

### DIFF
--- a/components/ILIAS/UI/src/Implementation/Component/Chart/ProgressMeter/Renderer.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Chart/ProgressMeter/Renderer.php
@@ -113,6 +113,8 @@ class Renderer extends AbstractComponentRenderer
         $hasComparison = false
     ): Template {
         $main_percentage = $component->getMainValueAsPercent();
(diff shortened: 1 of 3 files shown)


Amstutz:
thx a lot.

PR #9181, opened 2025-03-11
Form: make autocomplete work for multi-text inputs (44429)

schmitz-ilias:
This PR fixes the new legacy (weird how that works) autocomplete for multi-text inputs. This is relevant e.g. for creating new inputs in the calendar, see [44429](https://mantis.ilias.de/view.php?id=44429).

Changes are fairly slight. The `autocompleteHandler` assumes that the selection list is the direct next sibling to the text input. However, the list is placed via `appendSibling` in the parent container. For multi-text inputs that parent doesn't only contain the input, but also a few buttons, so the list ends up being placed as the last sibling behind the buttons. 

I made it so that the list is placed explicitly directly behind the input. An alternative fix would have been to replace all usages of `previousElementSibling` and `previousElementSibling` with more elaborate selectors, but that would have had a somewhat larger footprint.

Also, I adapted the js for multi inputs to the new legacy autocomplete. Beware, most of the changes there are whitespace and automatic formatting, relevant are lines 190-197 and 281-283.


schmitz-ilias:
@kergomard, would you mind taking a look here? The failing test is unrelated as far as I can see.

kergomard:
Hi @schmitz-ilias 

Thanks for taking care of this.

The changes in `legacyAutocomplete`, are definitely ok. I've no idea about the changes in `SurfaceFormMulti` as it is very hard to find the changes. Personally I would suggest to remove all code that is commented out and to change to vanilla javascript as this is going to break again with ILIAS 11, if we don't.

I tried to test this at the surface and couldn't make it work, though, so there might still be something wrong somewhere.

Best,
@kergomard

components/ILIAS/Form/resources/autocomplete/src/autocompleteHandler.js
```diff
@@ -137,11 +137,12 @@ async function fetchListItemsAndBuildSelector(fullUrl, inputField, config) {
     list.addEventListener('click', (e) => { onSelectHandler(e, config); });
     const activeElementValue = document.activeElement.dataset.value;
     removeList(inputField);
-    inputField.parentNode.appendChild(list);
+    inputField.parentNode.insertBefore(list, inputField.nextElementSibling);
```
iszmais, line 140: By what is it ensured that there is a `nextElementSibling`?
I could not find any proof of that.

iszmais (commented):
@schmitz-ilias Thank you for your contribution.

I just have one minor reassurance question.
(Also, please resolve the merge conflicts and rerun the checks to ensure they are working in the current state.)

Greetings,
@iszmais

components/ILIAS/Form/resources/autocomplete/src/autocompleteHandler.js
schmitz-ilias: There being no `nextElementSibling` is fine, no special handling needed: `insertBefore` can take `null` as its second parameter, and will then append the element to the end of the list.

schmitz-ilias:
Hi @iszmais, thank you very much for having a look at this PR!

I rebased to the current ILIAS 10, please have another look and let me know if there is anything else!

iszmais (approved):
Thank you @schmitz-ilias for the fix and the additional refactoring!

LGTM, therefore I will merge this.

schmitz-ilias:
Thank you @iszmais!

PR #9182, opened 2025-03-12
#44548 - Fix tracking issue preventing lp update

sroenchen:
Fix an issue that can prevent learning progress being recorded when unloading scormplayer. Mantis: https://mantis.ilias.de/view.php?id=44548


smeyer-ilias (approved):
Thanks @sroenchen

PR #9190, opened 2025-03-12
PRG: 44335, align placeholder description with entries in lang-file

nhaagen:
https://mantis.ilias.de/view.php?id=44335


klees:
@nhaagen Do we need to pick this to 9? If not, I wonder what changed. I so, I wonder why the problem didn't occur earlier.

klees:
Picked to trunk.

klees:
This in fact already existed in ILIAS 9. Changed picked to release_9.

PR #9192, opened 2025-03-12
Fix 44459: Cast code's type in the Initialization of SoapFault

jeph864:
See: https://mantis.ilias.de/view.php?id=44459

### DIFF
--- a/webservice/soap/classes/class.ilSoapAdministration.php
+++ b/webservice/soap/classes/class.ilSoapAdministration.php
@@ -196,7 +196,7 @@ class ilSoapAdministration
             case self::NUSOAP:
                 return new soap_fault($a_code, '', $a_message);
             case self::PHP5:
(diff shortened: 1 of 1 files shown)


webservice/soap/classes/class.ilSoapAdministration.php
```diff
@@ -196,7 +196,7 @@ protected function raiseError(string $a_message, $a_code)
             case self::NUSOAP:
                 return new soap_fault($a_code, '', $a_message);
             case self::PHP5:
-                return new SoapFault($a_code, $a_message);
+                return new SoapFault((string)$a_code, $a_message);
```
mjansenDatabay:
```suggestion
                return new SoapFault((string) $a_code, $a_message);
```

jeph864:
@mjansenDatabay thanks!

PR #9193, opened 2025-03-13
UI: 43797, adjust breadcrumb width for mobile views. 

lukastocker:
https://mantis.ilias.de/view.php?id=43797

Note: in this case ".open" is not used anymore


Amstutz:
thx a lot!

PR #9194, opened 2025-03-13
Fix error calling LTI 1.3 tool due to typo in ilObjLTIConsumer

fneumann:
https://mantis.ilias.de/view.php?id=44571


ZallaxDev:
The change is correct, we proceed to integrate it.

Thank you very much for your PR @fneumann 

Best Regards!

PR #9196, opened 2025-03-13
Test: Prevent "LOCK" situation for fixed questions

mjansenDatabay:
If questions are fixed (e.g. when requesting instant feedback) and there are multiple user solutions (one authorized, one not) for the same question in the same test pass, it might occur that the learner is caught in a "LOCK" situation, where she/he is not able to navigate to the next/previous question and not able to finish the test.
Every interaction results in a server request with `cmd=submitIntermediateSolution`, and the server responds with `ilTestException`.

This PR suggests a new state `isAnswerFixed` which is passed to the client/browser.
If the question is considered "answered" (authorized solution exists), "changed" (intermediate solution exists) and "fixed", the client-side states `answerChanged` and `stickyChanged` are set to `false`.
Doing this, the client does not set the `submitIntermediateSolution` command when interacting with the question navigation.

See: https://mantis.ilias.de/view.php?id=44467


kergomard:

Not beautiful and just a bandage, but I think this is the best we can and should do right now. Merged and picked to ILIAS 10 and trunk.

Best,
@kergomard

PR #9200, opened 2025-03-13
T&A Mantis #0044579: Participant Data are skipped during Excel Statistic Export

fhelfer:
https://mantis.ilias.de/view.php?id=44579

The Statistic Export for Random Tests has an additional row for each pass, containing the question titles.
 These lines are printed bold, which requires the if-conditions in ilExcelTestExport. This part was incorrect and caused participant data to be replaced with empty strings in lines of the export for fixed tests.

In addition to this, the Export for Random Tests had the lines for questions titles and participant results not in order with. The changes to ilTestExportAbstract switch the two arrays, so Participant data is the first line matched with the question titles to the first pass.

Let me know if you see any errors to this

Best @fhelfer


kergomard:

This should be ILIAS 9 only, as the whole export has changed anyway in ILIAS  10. So I do not pick.

Best,
@kergomard

PR #9201, opened 2025-03-13
Object: read out timings from crs timing view correctly (41946)

schmitz-ilias:
This PR fixes [41946](https://mantis.ilias.de/view.php?id=41946) by reading out timings correctly in `ilObjectActivation`. `ilTimingsUser` is relevant when the timings can vary user by user. This includes changeable timings, but also relative timings, and the latter are currently ignored.

As this code is used for building object lists, performance is a concern, so I resorted to using static methods directly wired to the database. It's not the prettiest fix, but I don't have any better ideas that don't involve rewriting the whole activation/timing mess.

Cheers, @schmitz-ilias

### DIFF
--- a/Services/Object/classes/class.ilObjectActivation.php
+++ b/Services/Object/classes/class.ilObjectActivation.php
@@ -235,8 +235,12 @@ class ilObjectActivation
 
         $item['timing_type'] = $item_array['timing_type'] ?? 0;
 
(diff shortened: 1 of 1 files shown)


kergomard:
Hi @schmitz-ilias 

After fighting with me for a while, I accept this for the time being.

Best,
@kergomard

PR #9202, opened 2025-03-14
OrgU: 44587, add/remove staff only with edit permission

nhaagen:
https://mantis.ilias.de/view.php?id=44587


klees:
Will fix CS afterwards.

klees:
Picked to trunk.

PR #9207, opened 2025-03-14
UI: 44226, adjust progress-bar styling.

lukastocker:
https://mantis.ilias.de/view.php?id=44226

Note:
As we do not want inline styling in the template stlyes I removed them for the progress-bar part and moved and added the styling to the SCSS file.


Amstutz:
thx a lot!

PR #9210, opened 2025-03-14
Certificate Bugfix #0044595: Import Dir has wrong rights

fhelfer:
https://mantis.ilias.de/view.php?id=44595

This might also be a problem for ILIAS 10 & 11
I will test immediately


mjansenDatabay (changes requested):
We should not re-introduce legacy function calls.

If there is an issue with `createDir`, we should suggest a change for the `FileSystem` component.

PR #9211, opened 2025-03-14
UI: 44051, remove hover effect from disabled glyph symbol in pagination

lukastocker:
view control.

https://mantis.ilias.de/view.php?id=44051

Please note that the pseudo class ":has()" works since December 2023 across all regular browser versions and devices. It might not work for older versions. You can check the supported versions [here](https://developer.mozilla.org/en-US/docs/Web/CSS/:has#browser_compatibility)

As the first fully supported browser/device versions (e.g. Chrome 105, Safari 15.4, Firefox 121, Edge 105,...)  got a lot of updates until now and therefore are not "new" I think it's valid to use the pseudo class ":has()".  

Tested on Vivaldi: 7.1.3570.58 
Tested on Chrome: 134.0.6998.35 
Tested on Firefox: 136.0.1


Amstutz:
thx a lot

PR #9213, opened 2025-03-17
KS-Form for Apache Auth Settings

mBeym:
Changes the apache auth settings to use the ks-form.

### DIFF
--- a/components/ILIAS/Authentication/classes/Form/ApacheAuthSettingsForm.php
+++ b/components/ILIAS/Authentication/classes/Form/ApacheAuthSettingsForm.php
@@ -1,0 +1,222 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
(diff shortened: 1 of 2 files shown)


components/ILIAS/Authentication/classes/Form/ApacheAuthSettingsForm.php
```diff
@@ -0,0 +1,222 @@
...
+    private string $show_command;
+    private string $save_command;
+    private array $values;
+
+    private UIFactory $ui_factory;
+    private ilLanguage $lng;
+    private ilRbacSystem $rbac_system;
+    private ilRbacReview $rbac_review;
+    private ilCtrlInterface $ctrl;
+    private FieldFactory $ui_field;
```
mjansenDatabay, line 50: We should use constructor property promotion if possible. Please use a rector which migrates the code to PHP 8.3/8.4.

mjansenDatabay (changes requested):

mjansenDatabay:
Thx @mBeym ,

I left an inline comment.

Furthermore, please check whether the input validation is robust against XSS attacks. It might be the case that we need further transformations (in general, I suggest an additional `trim` and, as long as we have no alternative, a `ilUtil::stripSlashes` transformation, **if necessary**).

Best regards,
Michael

components/ILIAS/Authentication/classes/Form/ApacheAuthSettingsForm.php
mBeym: 1. But then I can't retrieve them from $DIC if not passed as parameters?

2. No rector in trunk

mjansenDatabay:
Integrated with https://github.com/ILIAS-eLearning/ILIAS/commit/285dacfe19284ebd608671f0f66f46f011ae454b

PR #9214, opened 2025-03-17
Object Bugfix #0044603: Custom Icons are being saved with private access rights

fhelfer:
https://mantis.ilias.de/view.php?id=44603

### DIFF
--- a/Services/Object/classes/Properties/AdditionalProperties/Icon/class.ilObjectCustomIcon.php
+++ b/Services/Object/classes/Properties/AdditionalProperties/Icon/class.ilObjectCustomIcon.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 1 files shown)


Services/Object/classes/Properties/AdditionalProperties/Icon/class.ilObjectCustomIcon.php
```diff
@@ -114,6 +114,7 @@ public function saveFromTempFileName(string $tempfile_name): void
         }
 
         rename(ilFileUtils::getDataDir() . '/temp/' . $tempfile_name, $this->getFullPath());
+        $this->filesystem->setVisibility($relative_path, 'public');
```
mjansenDatabay, line 117: Is this only applied for the next operation, or does it have further site-effects within an ILIAS request @fhelfer ?

fhelfer: setVisibility() will only set Public Visibility for the currently newly uploaded Custom Icon. 
This will have no effect on the directory or any other uploaded files in the request

kergomard:
Thank you very much for the PR @fhelfer ! merged and picked to ILIAS 10 and trunk.

PR #9218, opened 2025-03-18
User Bugfix #0044535: Autofill of "Maximum Login Attempts" is invalid

fhelfer:
https://mantis.ilias.de/view.php?id=44535


fhelfer:
This could also be solved by allowing '0' as an input for login_max_attempts
Please let me know which option is preferred and I will change it accordingly

Best @fhelfer

kergomard:

I think your solution is the right one. I find the 0 there very disturbing as it is hard to understand what it actually means. Merged and picked to ILIAS 10 and trunk.

Best,
@kergomard

PR #9219, opened 2025-03-18
T&A 44486: Fixes umlaute not being displayed correctly in correction tab of text subset question

matheuszych:
[Mantis: 44486](https://mantis.ilias.de/view.php?id=44486)

I believe removing the `htmlentities` function call allows the data to be correctly persisted into the database.

### DIFF
--- a/Modules/Test/classes/class.ilTestCorrectionsGUI.php
+++ b/Modules/Test/classes/class.ilTestCorrectionsGUI.php
@@ -132,9 +132,7 @@ class ilTestCorrectionsGUI
 
         $this->setCorrectionTabsContext($questionGUI, 'question');
 
-        if ($form === null) {
-            $form = $this->buildQuestionCorrectionForm($questionGUI);
-        }
+        $form ??= $this->buildQuestionCorrectionForm($questionGUI);
 
         $this->populatePageTitleAndDescription($questionGUI);
         $this->main_tpl->setContent($form->getHTML());
@@ -574,17 +572,11 @@ class ilTestCorrectionsGUI
      */
     protected function allowedInAdjustment(\assQuestionGUI $question_object): bool
     {
-        $setting = new ilSetting('assessment');
-        $types = explode(',', $setting->get('assessment_scoring_adjustment'));
-        $type_def = array();
-        foreach ($types as $type) {
+        $type_def = [];
+        foreach (explode(',', (new ilSetting('assessment'))->get('assessment_scoring_adjustment', '')) as $type) {
             $type_def[$type] = ilObjQuestionPool::getQuestionTypeByTypeId($type);
         }
 
-        $type = $question_object->getQuestionType();
-        if (in_array($type, $type_def)) {
-            return true;
-        }
-        return false;
+        return in_array($question_object->getQuestionType(), $type_def);
     }
 }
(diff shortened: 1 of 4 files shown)


matheuszych:
Already reviewed by @thojou.

kergomard:

This is the right approach, ...but did you check all places where this input is used as an output and verified that we properly sanitize our outputs? If not: Would you please be so kind and check.

Best,
@kergomard

matheuszych:
@kergomard

I have looked everywhere where this input/output is used. All rendering of the value seems to be done here:
https://github.com/ILIAS-eLearning/ILIAS/blob/719b07ef2eab1afdeeb3c196f621cf30147f4a7a/Modules/TestQuestionPool/classes/class.assTextSubsetGUI.php#L242

I added output sanitization there.

kergomard:
Hi @matheuszych !

Thanks for checking.

I also gave this a second more thorough look. The problem actually is a double encoding. Would we simply do, what you propose, existing answers would still break, as we would now double encode them everywhere.

I still believe your move to move encoding from input to output is correct, we just need to do a little bit more:
- [ ] The encoding you have (added with your last commit)(https://github.com/ILIAS-eLearning/ILIAS/pull/9219/files#diff-5f1b21852b129d6151b8f2d7fb8640833674ca5017c913939b6d76179e2925d4R242) should be changed to `$template->setVariable('SOLUTION', $this->escapeTemplatePlaceholders(htmlspecialchars($user_solutions[$i]['value1'], ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401, null, false))`. This way things wouldn't get double encoded.
- [ ] We will also need to change [this line here](https://github.com/ILIAS-eLearning/ILIAS/blob/release_9/Modules/TestQuestionPool/classes/forms/class.ilAssAnswerCorrectionsInputGUI.php#L133). I would suggest to internalize the function `ilLegacyFormElementsUtil::prepareFormOutput()` into `ilAnswerwidzardInputGUI`(I'm pretty sure we are going to need this in other places), remove all comments from it and change the `htmlspecialchars...` line to also avoid double encoding (see above).

I think this way we get the cleanest right now possible solution. What do you think?

Thanks again and best,
@kergomard

matheuszych:
Hello @kergomard 
I agree with you. Double encoding might be a problem here. I adjusted the code accordingly. This seems to be the correct way of doing it right now.

Best regards
@matheuszych

kergomard:
Hi @matheuszych 

Thank you very much for updating the PR. I think the function can more, than we need, but lets keep it this way for the time being. But:
- The first parameter of `ilAnswerWizardInputGUI::prepareFormOutput(int|string $input, bool $strip = false): string` is not correct, as the points, where we are already using this, should be `float`s not `int`s. I think we could only accept strings here and cast before passing it to the function, but I could also live with adding `float` to the parameter, I would then though think, that we should use the refinery for the cast int the function.

Thanks again and best,
@kergomard

matheuszych:
Hello @kergomard

As of right now we would be using it in two spots. One where we pass a `float` and one where we pass a `string`. I would suggest we accept `float`s and `string`s and, as you suggested, use the refinery to cast them to `strings`. One "issue" with the refinery would be, that we either would need to make the method `non-static` to access the classes refinery property (probably defeats the whole purpose of such helper method) or we need to access the Container directly inside the method.

I changed it for now to do so.

Best regards
@matheuszych

kergomard:
Thank you @matheuszych !

Sorry, I was apparently not awake enough this morning:
- This function should clearly neither be `static` nor `public` (I completely missed it). If we need this pattern in many places we need to figure out how to deal with it, but it is right now just a solution for this one place and it is a strong smell. Please make it `protected` and not `static` and use the refinery from the class.

Sorry again, I should have caught this.

Best,
@kergomard

matheuszych:
Hello @kergomard
No Problem 😄.  I pushed the changes.

Best regards
@matheuszych

kergomard:
Thank you very much for updating this again @matheuszych !

Best,
@kergomard

PR #9220, opened 2025-03-18
T&A 44580: Fixes wrong learning progress determination when passing second - Nth attempt/pass

aaronbidzan:
This PR attempts to solve the issue described in [Mantis #44580](https://mantis.ilias.de/view.php?id=44580).

When a random test with one question synced from a question pool is completed for the 2nd - Nth time, the learning progress is incorrectly calculated and shows `failed`, even if the user scores 100%.

This happens because the learning progress is calculated based on cached test results. Following the code execution order, the cache is updated **after** the learning progress is calculated. As a result, the learning progress is determined using outdated data that still has a `failed` status.

To fix this, I moved the `updateTestResultCache` function into the `performFinishTasks` function, placing it between setting the test as finished and updating the learning progress.

Kind regards,  
@bidzanaaron

### DIFF
--- a/Modules/Test/classes/class.ilCronFinishUnfinishedTestPasses.php
+++ b/Modules/Test/classes/class.ilCronFinishUnfinishedTestPasses.php
@@ -185,7 +185,7 @@ class ilCronFinishUnfinishedTestPasses extends ilCronJob
                 if ($this->test_ending_times[$test_id]['enable_processing_time'] == 1) {
                     $this->log->info('Test (' . $test_id . ') has processing time (' . $this->test_ending_times[$test_id]['processing_time'] . ')');
                     $obj_id = $this->test_ending_times[$test_id]['obj_fi'];
-                    if(ilObject::_exists($obj_id)) {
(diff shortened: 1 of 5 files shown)


mjansenDatabay:
One addition: The issue also occurs with more than one question.

kergomard (changes requested):
Thank you  very much for the PR @bidzanaaron !

I've one little request:
- [ ] `ilTestEvaluationGUI::finishTestPass(int $active_id, int $obj_id)` is only called from within `ilTestEvaluationGUI`. And it is always called as `$this->finishTestPass($participant->getActiveId(), $this->object->getId())`. This means that we do not need the second parameter and you can pass `$this->object` to `ilTestPassFinishTasks` instead of reading a whole new object from db.

...as always: If I didn't miss anything.

Thanks again and best,
@kergomard

kergomard:
Thank you very much @bidzanaaron !

On a side note: Please don't force push on changes, but keep the history intact, this makes my life far easier to rereview.

Thanks again and best,
@kergomard

PR #9221, opened 2025-03-18
T&A 44619: Fixes error on kprim question correction statistics page

matheuszych:
[Mantis: 44619](https://mantis.ilias.de/view.php?id=44619)

A possible fix is to access the `language` property directly since it is already available because of the parent class.


matheuszych:
Already reviewed by @thojou.

kergomard:

Merged and picked to ILIAS 10 and trunk (just for the removal of the intermediate variable).

Best,
@kergomard

PR #9223, opened 2025-03-19
T&A 44644: Fixes taxonomy filtering issue in question pool table

matheuszych:
[Mantis: 44644](https://mantis.ilias.de/view.php?id=44644)

These changes aim to fix an issue caused when removing, readding and than submitting the taxonomy filter form.  I believe that the `input_element.val("");` call causes the value of the select to be removed, when removing the filter instead of just hiding it. For other input types, it is an expected behaviour but not for a select.

### DIFF
--- a/src/UI/templates/js/Input/Container/dist/filter.js
+++ b/src/UI/templates/js/Input/Container/dist/filter.js
@@ -1,3 +1,21 @@
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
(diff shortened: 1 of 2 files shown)


matheuszych:
Already reviewed by @thojou.

thojou:
Hey @Amstutz,

As you were assigned to the related Mantis ticket, I wanted to kindly ask for your review of these code changes.

Thank you.
Best,
@thojou

Amstutz:
Hi @thojou 

Thx a lot! The assignment is correct. All changes to the UI Componentns (src/UI) need to be reviewed and merged by the UI Coordinators. 

thx for contributing! we will look into it.

@Amstutz

Amstutz:
Thx a lot for the PR 

We looked into it. From our point of view this is ready to go. However, since the filters are implemented by @tfamula we need a short thumbs up from him as well.

Thx

Amstutz:
I @tfamula , please re-assign me, once you looked into this, thx!

tfamula (approved):
Good catch, thank you very much!

Amstutz:
@matheuszych and @tfamula , thx a lot! note, I merged the PR to 9. However, for 10 and trunk the filter js code has been refactored. if the fix needs to be applied to those versions as well, another PR would be highly appreciated.

tfamula:
The error also appears for 10 and trunk

PR #9224, opened 2025-03-19
Cache: ILIAS 9 add TTL back to set function

gvollbach:
@chfsx This pull request re-adds the possibility to set the TTL if adding an entry to the cache. If nothing is set, the default TTL from the cache config will be used.

### DIFF
--- a/src/Cache/Container/ActiveContainer.php
+++ b/src/Cache/Container/ActiveContainer.php
@@ -211,16 +211,20 @@ final class ActiveContainer implements Container
         return $this->buildFinalTransformation($transformation)->transform($unpacked_values);
     }
 
-    public function set(string $key, string|int|array|bool|null $value): void
+    public function set(string $key, string|int|array|bool|null $value, int $ttl = null): void
     {
         if ($this->isLocked()) {
             return;
         }
+        if ($ttl === null) {
+            $ttl = $this->config->getDefaultTTL();
+        }
+
         $this->adaptor->set(
             $this->request->getContainerKey(),
             $key,
             $this->pack($value),
-            $this->config->getDefaultTTL()
+            $ttl
         );
     }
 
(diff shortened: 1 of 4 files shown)


src/Cache/Container/ActiveContainer.php
```diff
@@ -211,16 +211,20 @@ public function get(string $key, Transformation $transformation): string|int|arr
         return $this->buildFinalTransformation($transformation)->transform($unpacked_values);
     }
 
-    public function set(string $key, string|int|array|bool|null $value): void
+    public function set(string $key, string|int|array|bool|null $value, int $ttl = null): void
     {
         if ($this->isLocked()) {
             return;
         }
+        if ($ttl === null) {
```
chfsx: Can we strip that down to a `$ttl = $ttl ?? $this->config->getDefaultTTL();` here?

src/Cache/Container/ActiveContainer.php
```diff
@@ -211,16 +211,20 @@ public function get(string $key, Transformation $transformation): string|int|arr
         return $this->buildFinalTransformation($transformation)->transform($unpacked_values);
     }
 
-    public function set(string $key, string|int|array|bool|null $value): void
+    public function set(string $key, string|int|array|bool|null $value, int $ttl = null): void
```
chfsx: The parameter MUST be `?int $ttl = null` since in PHP 8.4 this will be mandatory

chfsx (changes requested):
@gvollbach thanks a lot for the PR, could you please adjust the two small thing I mentioned in the code?

src/Cache/Container/ActiveContainer.php
chfsx: This applies to all places where this new parameter is introduced

gvollbach:
@chfsx Thx for the review, I changed the code according to your request.

chfsx:
Thanks a lot! I rebased and squashed locally and pushed to all relevant branches

PR #9227, opened 2025-03-19
T&A 44512: Fixes issue with pool_title column in tst_rnd_quest_set_qpls not being able to fit all data in some cases

matheuszych:
[Mantis: 44512](https://mantis.ilias.de/view.php?id=44512)

The question pool title is stored in (atleast) two tables. The `object_data` table column `title` allows up to `255` characters and the  `tst_rnd_quest_set_qpls` table column `pool_title` currently only allows up to `128` characters. This can causes issues on updated platforms when opening older random tests.

Since we have already allowed the title to be up to `255` characters we cannot decrease it and probably should increase it on the other side to match that limit.


matheuszych:
Already reviewed by @thojou.

kergomard:

I agree with the solution in general, but please do not change `setup/sql/ilias3.sql`. It is none of our business.

Best,
@kergomard

matheuszych:
@kergomard Thank you for your feedback.

I adjusted the code accordingly.

kergomard:
Thank you very much @matheuszych ! Merged and picked to ILIAS 10 and trunk.

PR #9228, opened 2025-03-19
T&A 44463: Fixes test exporting dates using incorrect timezone

matheuszych:
[Mantis: 44463](https://mantis.ilias.de/view.php?id=44463)

When exporting a test, the starting and ending time where exported in the systems timezone. Importing it in a system with a different timezone causes the time to not be alligned correctly.

By always exporting timestamps in the UTC-0 timezone and displaying it to the user in their prefered timezone this issue can be mitigated and timestamps can be directly compared to eachother.

### DIFF
--- a/Modules/Test/classes/MainSettings/ilObjTestSettingsAccess.php
+++ b/Modules/Test/classes/MainSettings/ilObjTestSettingsAccess.php
@@ -70,31 +70,15 @@ class ilObjTestSettingsAccess extends TestSettings
         array $environment = null
     ): Group {
         $constraint = $refinery->custom()->constraint(
-            static function (array $vs): bool {
-                if ($vs['start_time'] === null
-                    || $vs['end_time'] === null) {
-                    return true;
-                }
-                return $vs['start_time'] < $vs['end_time'];
-            },
+            static fn (array $vs) =>
+                $vs['start_time'] === null || $vs['end_time'] === null || $vs['start_time'] < $vs['end_time'],
             $lng->txt('duration_end_must_not_be_earlier_than_start')
         );
 
         $trafo = $refinery->custom()->transformation(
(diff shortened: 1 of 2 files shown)


matheuszych:
Already reviewed by @thojou.

kergomard:
Thank you very much @matheuszych for the PR !

I've two questions/change request:
- [ ] Why are you converting the `ImmutableDateTime` to a `Timestamp` and then back to a `DateTime`? With a few small changes (instead of getting `$this->getStartingTime()` and `$this->getEndingTime()` get the dates directly from the settings) you will always have a `DateTimeImmutable` in your hands and you can then change `ilObjTest::buildIso8601PeriodFromUnixTimeForExportCompatibility()` to accept this as a parameter and then only need to set the timezone on it.
- [ ] Do we really need to remove the leading zeros from minutes and seconds? Removing this would make this more compact and readable.

Thanks again and best,
@kergomard

matheuszych:
Hello @kergomard

You are right. We can directly use the `DatetimeImmutable` here instead of doing a transformation to timestamp first.

Technically we do not needs this. I just copied the format from how it was before. Previously all the datetime components did not have a leading zero. The `DateTime::format` method does provide a format character for each component, with and without a leading zero, except for minutes and seconds. That is why i choose a "custom" approach to these two.

Best regards
@matheuszych

kergomard:

Sorry, I didn't make this explicit: We need to rename the function now, as it doesn't do anything with unix-timestamps anymore. I would suggest `ilObjTest::buildIso8601PeriodForExportCompatibility()`. I think this is enough.

Thanks again and best,
@kergomard

kergomard:
Thanks once more @matheuszych ! Merged and picked to ILIAS 10 and trunk.

PR #9229, opened 2025-03-19
T&A 43482: Fixes test export with user results and adds user file upload question file uploads to export and import

matheuszych:
[Mantis: 43482](https://mantis.ilias.de/view.php?id=43482)

When exporint a test with participant data, the file uploads from file upload questions do not get exported. This causes the files to be missing on import and preventing a manual correction if the question.

I believe this should fix this issue, by including the participants file uploads in the export and importing them on import into the systems storage.

### DIFF
--- a/Modules/Test/classes/class.ilTestExport.php
+++ b/Modules/Test/classes/class.ilTestExport.php
@@ -18,6 +18,8 @@
 
 declare(strict_types=1);
 
+use ILIAS\ResourceStorage\Services as ResourceStorage;
+
 require_once './Modules/Test/classes/inc.AssessmentConstants.php';
 
 /**
@@ -52,6 +54,7 @@ abstract class ilTestExport
     protected ILIAS $ilias;
 
     protected string $inst_id;
+    private ResourceStorage $irss;
 
     public function __construct(
         public ilObjTest $test_obj,
@@ -63,6 +66,7 @@ abstract class ilTestExport
         $this->db = $DIC['ilDB'];
         $this->lng = $DIC['lng'];
         $this->bench = $DIC['ilBench'];
+        $this->irss = $DIC['resource_storage'];
 
         $this->inst_id = (string) IL_INST_ID;
 
@@ -243,7 +247,13 @@ abstract class ilTestExport
         $this->bench->stop("TestExport", "buildExportFile_dumpToFile");
 
         if ($this->isResultExportingEnabledForTestExport()) {
-            $resultwriter = new ilTestResultsToXML($this->test_obj->getTestId(), $this->db, $this->test_obj->getAnonymity());
+            $resultwriter = new ilTestResultsToXML(
+                $this->test_obj->getTestId(),
+                $this->db,
+                $this->irss,
+                $this->export_dir . "/" . $this->subdir . "/objects",
+                $this->test_obj->getAnonymity()
+            );
             $resultwriter->setIncludeRandomTestQuestionsEnabled($this->test_obj->isRandomTest());
             $this->bench->start("TestExport", "buildExportFile_results");
             $resultwriter->xmlDumpFile($this->export_dir . "/" . $this->subdir . "/" . $this->resultsfile, false);
--- a/Modules/Test/classes/class.ilTestImporter.php
+++ b/Modules/Test/classes/class.ilTestImporter.php
@@ -18,6 +18,8 @@
 
 declare(strict_types=1);
 
+use ILIAS\ResourceStorage\Services as ResourceStorage;
+
 /**
  * Importer class for files
  *
@@ -34,12 +36,14 @@ class ilTestImporter extends ilXmlImporter
 
     private ilLogger $log;
     private ilDBInterface $db;
+    private ResourceStorage $irss;
 
     public function __construct()
     {
         global $DIC;
         $this->log = $DIC['ilLog'];
         $this->db = $DIC['ilDB'];
+        $this->irss = $DIC['resource_storage'];
 
         parent::__construct();
     }
@@ -121,7 +125,7 @@ class ilTestImporter extends ilXmlImporter
         $results_file_path = ilSession::get("tst_import_results_file");
         // import test results
         if ($results_file_path !== null && file_exists($results_file_path)) {
-            $results = new ilTestResultsImportParser($results_file_path, $new_obj, $this->db, $this->log);
+            $results = new ilTestResultsImportParser($results_file_path, $new_obj, $this->db, $this->log, $this->irss);
             $results->setQuestionIdMapping($a_mapping->getMappingsOfEntity('Modules/Test', 'quest'));
             $results->setSrcPoolDefIdMapping($a_mapping->getMappingsOfEntity('Modules/Test', 'rnd_src_pool_def'));
             $results->startParsing();
--- a/Modules/Test/classes/class.ilTestResultsImportParser.php
+++ b/Modules/Test/classes/class.ilTestResultsImportParser.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +18,9 @@
 
 declare(strict_types=1);
 
+use ILIAS\Filesystem\Stream\Stream;
+use ILIAS\ResourceStorage\Services as ResourceStorage;
+
 class ilTestResultsImportParser extends ilSaxParser
 {
     private $table;
@@ -27,6 +31,7 @@ class ilTestResultsImportParser extends ilSaxParser
     private bool $user_criteria_checked = false;
 
     protected $src_pool_def_id_mapping;
+    private string $import_directory;
 
     /**
     * Constructor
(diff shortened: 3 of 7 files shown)


matheuszych:
Already reviewed by @thojou.

kergomard:
Thank you very much @matheuszych for the PR!

I think we need to find a solution, that doesn't require any changes in `ilTestExport` or `ilTestExportGUI` as they would not survive the port to ILIAS 10. 

Over the weekend I moved the export with results to the default export pattern. It now should export everything including Head- and Tail-Dependencies. You should now simply be able to add `ResourceStorage` to `ilTestExporter` and then all other changes would be in `ilTestResultsToXML` and then on the importer side. Don't wonder about the slightly wonky pattern, where I move the information around in a parameter, it was the only way I could find that didn't involve using the session and I think using the session for this is plain wrong. We will need to find a better solution, but not as a bugfix.

One small nitpick:
- Please do not add trailing commas to existing parameter-lists that you do not touch. If you change a parameter list, I won't say anything, but I will just remove them the next time I touch it. There are no rules for that in our coding styles and I see this actually as being a "fault tolerance"-feature. The grammar is still wrong (a comma indicates that there is at least one more element following), but the interpreter will be kind with you. I will adapt, if and when we set general rules for this ;-) .

Thank you very much and best,
@kergomard

matheuszych:
Hello @kergomard
I restructered the code a bit. I tried to keep it a simple as possible. Could you please re-review the changes?

kergomard (changes requested):
Thank you very much for updating the PR @matheuszych !

Please change the following or explain, why we shouldn't:
- [ ] Please remove the usage of the `$DIC` from `ilTestResultsImportParser` and move `ResourceStorage` to parameters of the constructor.
- [ ] Please declare the return type of the closure [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-84dff1fb73f8439476fe28c004986be04ddf8cf211aaf021ccb1ac5bcd0f7fb6R342).
- [ ] I think it is not correct to use `UploadResult` and `UploadedFile` here and also `ilObjFileStakeholder()` is wrong. Please create a stream directly from the path and use `assFileUploadStakeholder`.
- [ ] Please do not change `$a_attribs` in `ilTestResultsImportParser::handleParticipantUploadedFiles()`, but use class-variables. The `$a_attribs` should contain a very specific information: The content of the attributes of the corresponding node and should thus be seen as unmutable. By changing them you introduce a completely unexpected side-effect.
- [ ] We should be able to get rid of the `else` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-84dff1fb73f8439476fe28c004986be04ddf8cf211aaf021ccb1ac5bcd0f7fb6R375) one way or another (see previous point).
- [ ] Please remove the usage of the `$DIC` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-4c0c9752016e9033a1e229adb5d7ed7ded9fea043d88c11de4ad9cfe1a0c06d0R236-R237) and move it to parameters of the constructor.
- [ ] This check [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-4c0c9752016e9033a1e229adb5d7ed7ded9fea043d88c11de4ad9cfe1a0c06d0R247-R251) looks plain wrong. We were able to get a stream from the file, so the file exists. I think we should be able to simply remove it.
- [ ] [Here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-4c0c9752016e9033a1e229adb5d7ed7ded9fea043d88c11de4ad9cfe1a0c06d0R252-R255), as we are using legacy anyway, please call `ilFileUtils::makeDirParents($path_with_rid)`.
- [ ] Please do not `copy()` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-4c0c9752016e9033a1e229adb5d7ed7ded9fea043d88c11de4ad9cfe1a0c06d0R256), but put the stream in the file directly.

Sorry, for the long list of requests and I have even a final one more: Please port this to ILIAS 10 and provide a corresponding PR as this will be more than a simple cherry-pick.

Thank you very much and best,
@kergomard

dsstrassner:
hi @matheuszych - any Updates on this?

matheuszych:
Hello @kergomard,

I implemented the required changes and rebased them onto the current release_9 branch. Could You please take a look again?

Best regards
@matheuszych

matheuszych:
Hello @kergomard ,

I tried to replicate the pipeline issue with PHP 8.1 and PHP 8.2 by running php-cs-fixer locally on the seven files, but in both cases php-cs-fixer runs successfully. Could you please take a look at why this might be happening?

Best regards
@matheuszych

kergomard (changes requested):
Hi @matheuszych !

Thank you very much for the PR !

First, I would have a few questions/change requests:
- We do we need the separate `ilTestResultsToXML::setObjectsExportDirectory()`? We are always setting this anyway and things should go sideways, if we don't, so this should go into the constructor.
- Please use array-access to access the `$DIC` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-fe80b7670b0fa43a225eb4043ffc0f64874838b754934bfc981b3100ac2b4345R69) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-a32f2039d1a46415faadbf247c4345a022c0e8dcb876a0a5d20e4a7c4eff4057R46) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-1a8f703bfa69245f11e138d8abd3d39baca77aa1845723fda18e3474a27a034bR35) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-7f7dee0a52a5cf705a98ff60bedd62d1a9a7248119818be1c4059bb872cb0afbR37).
- Please give an alias of 'ResourceStorage' to 'ILIAS\ResourceStorage\Services'.
- I don't like how `ilTestResultsImportParser::handleParticipantUploadedFiles()` is called and also the internal structure of it: Please make sure you don't need to recheck `$value1` after assigning a value [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-84dff1fb73f8439476fe28c004986be04ddf8cf211aaf021ccb1ac5bcd0f7fb6R234-R235). From where I stand all the logic should be inside the`handleParticipantUploadedFiles()`-function. The function should also be called something else as the current name does not tell us anything about what it does. If I'm getting this right what the function does is `importParticipantsUploadedFiles()`. The function should always return `$a_attribs['value1']` or the new rid and may return `null`. Change it do do a early return of `$value1`, if `$value2` is not equal to 'rid' and remove the `array_filter()` + `array_shift()` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-84dff1fb73f8439476fe28c004986be04ddf8cf211aaf021ccb1ac5bcd0f7fb6R339-R343) in favor of a function that just returns the path to the file.
- Change `ilTestResultsToXML::exportUploadedFiles()` to `continue` if `$uploaded_file['value2'] is not equal to 'rid'.
- We also don't need the `fopen()` + `fwrite` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-4c0c9752016e9033a1e229adb5d7ed7ded9fea043d88c11de4ad9cfe1a0c06d0R240-R253) a `file_put_contents()` should do things directly from the stream.

Second: Please provide a separate PR for ILIAS 10 as this will not be a simple cherry-pick.

Third: I'm actually also not sure, why the checks fail. Let us see if something changes, when you change the above.

@kergomard

matheuszych:
Hello @kergomard,

thank you for your review. I implemented the requested changes and will provide a PR for ILIAS 10 soon™.

The php-cs-fixer issue seems to persist. Looks like the same exact spot as before. Might be something with the `Modules/Test/classes/class.ilTestResultsImportParser.php` or `Modules/Test/classes/class.ilTestResultsToXML.php` but not sure what.

Best regards
@matheuszych

kergomard:

Ok, I think we now have a very small change, that we should revert:
- `$value1` can now be an empty `string` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-84dff1fb73f8439476fe28c004986be04ddf8cf211aaf021ccb1ac5bcd0f7fb6R234). I think we should keep the current behavior and make sure that the empty string is transformed to `null`.

For the cs-fixer issue:
- For the `ilTestResultsImportParser` the issue might be a missing space between `<?php` and the copyright, but this is not going to be the issue.

I would suggest that I simply squash your commits and cherry-pick them locally in the hopes that this solves the issue. But I will leave this for when your PR for ILIAS 10 is ready.

Thanks again and best,
@kergomard

matheuszych:
Hello @kergomard,

i added the change [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-84dff1fb73f8439476fe28c004986be04ddf8cf211aaf021ccb1ac5bcd0f7fb6R332) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9229/files#diff-84dff1fb73f8439476fe28c004986be04ddf8cf211aaf021ccb1ac5bcd0f7fb6R338). I also added the space beteween the php tag and the copyright block.

The pipeline works now! 🥳 

Best regards
@matheuszych

kergomard (approved):
Thank you very much @matheuszych ! I wait then for the PR for 10 and merge both of them together (makes tracking for me easier).

matheuszych:
https://github.com/ILIAS-eLearning/ILIAS/pull/9587

PR #9230, opened 2025-03-19
Mantis #32086: Fix deleted objects (trashed) shown in role auto complete

mBeym:
Fixes https://mantis.ilias.de/view.php?id=32086.

### DIFF
--- a/Services/AccessControl/classes/class.ilRoleAutoComplete.php
+++ b/Services/AccessControl/classes/class.ilRoleAutoComplete.php
@@ -1,6 +1,5 @@
 <?php
 
-declare(strict_types=1);
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -16,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 /**
  * Auto completion class for user lists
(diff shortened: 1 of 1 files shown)


Services/AccessControl/classes/class.ilRoleAutoComplete.php
```diff
@@ -38,8 +38,10 @@ public static function getList(string $a_str): string
 //            "JOIN object_reference obr ON ref_id = t1.parent " .
             "JOIN object_reference obr ON ref_id = fa.parent " .
             "JOIN object_data o2 ON obr.obj_id = o2.obj_id " .
+            "INNER JOIN tree t ON t.child = obr.ref_id AND t.tree = 1 " .
             "WHERE o1.type = 'role' " .
             "AND assign = 'y' " .
+            "AND obr.deleted IS NULL " .
```
mjansenDatabay: I suggest moving this condition to the JOIN object_reference ... statement.

kergomard:

I've one question:
- [ ] If I'm understanding this right `AND obr.deleted IS NULL` does the same thing as `INNER JOIN tree t ON t.child = obr.ref_id AND t.tree = 1`. Do we really need to incur the performance penalty for the second join? My feeling is that we should prefer the simpler first check and just be happy with what it filters. What am I missing?

Thanks again and best,
@kergomard

mBeym:
@kergomard => I just wanted to make sure only fully deleted objects are filtered. Just in case there is ever a reason the id in the tree stays at "1".

"Better safe than sorry"

kergomard:
Ok, let us see. I can always remove this later on, if needed.

Thanks again and best,
 @kergomard

PR #9231, opened 2025-03-19
Potential Fix mantis: 44591 Course members from campus management are only adde…

jheim2:
…d to courses but not removed

Course members were not being removed from ILIAS when they no longer appeared in the CMS JSON. This was due to a mismatch in the SQL query condition, which only checked for cms_sub_id IS NULL, while in our case, cms_sub_id was 0. The fix extends the condition to also check for cms_sub_id = 0, ensuring proper removal of members.

### DIFF
--- a/Services/WebServices/ECS/classes/Course/class.ilECSCourseMemberAssignment.php
+++ b/Services/WebServices/ECS/classes/Course/class.ilECSCourseMemberAssignment.php
@@ -113,7 +113,7 @@ class ilECSCourseMemberAssignment
         $ilDB = $DIC['ilDB'];
 
         if (is_null($a_cms_sub_id)) {
-            $cms_sub_id_query = 'AND cms_sub_id IS NULL ';
+            $cms_sub_id_query = 'AND cms_sub_id IS NULL OR cms_sub_id = 0 ';
         } else {
             $cms_sub_id_query = 'AND cms_sub_id = ' . $ilDB->quote($a_cms_sub_id, 'integer') . ' ';
         }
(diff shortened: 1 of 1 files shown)


Akumatic:
Thanks for the PR. Will take a closer look to the behavior.

If `a_cms_sub_id` is not zero but an integer, the `else` block should work - even for 0 as a value.

On a quick glance the whole function is only called once in [ilECSCmsCourseMemberCommandQueueHandler.refreshAssignmentStatus](https://github.com/ILIAS-eLearning/ILIAS/blob/3bd0835b5062650b6338400db52a08b61b3fb03b/Services/WebServices/ECS/classes/Course/class.ilECSCmsCourseMemberCommandQueueHandler.php#L266) - and even if the sub_id is null, it's casted to an int, so null should never be passed to `lookupUserIds`:
```
$usr_ids = ilECSCourseMemberAssignment::lookupUserIds(
    $course_id,
    is_null($sub_id) ? $sub_id : (int) $sub_id,
    $obj_id
);
```

iljalukin:
Sorry for jumping in here, but isn't the exact opposite the case? If $sub_id is null, then it is passed on without change, otherwise it is cast to int.
Is is_null even necessary at this point if it never happens that cms_sub_id=NULL in the database? Couldn't (int) $sub_id simply be used as an argument in lookupUserIds and in lookupAssignment?

Akumatic:
> Sorry for jumping in here, but isn't the exact opposite the case? If $sub_id is null, then it is passed on without change, otherwise it is cast to int.

You're absolutely right, that was a mistake on my part at that short Glance.

> Is is_null even necessary at this point if it never happens that cms_sub_id=NULL in the database? Couldn't (int) $sub_id simply be used as an argument in lookupUserIds and in lookupAssignment?

It could be simplified like that if it is ensured that there are no NULL in the database & no entries are overlooked by the cast

Akumatic:
Underlying issue should be fixed with #9482

PR #9235, opened 2025-03-19
ECS: Fix missing typecast

iljalukin:
https://mantis.ilias.de/view.php?id=42370


Akumatic:
Thanks for the pr and sorry for the late reply. Merged & picked to 10 and trunk

PR #9236, opened 2025-03-19
Test: Prevent navigation before control script is initialized

fneumann:
See https://mantis.ilias.de/view.php?id=44467

With a slow network connection the script ilTestPlayerQuestionEditControl.js might not be loaded or initalized before navigation buttons on a test page are clicked. This may result in POST requests to the fallbackCmd "saveIntermediateSolution" with empty data, which is wrong. 

This PR prevents such requests by setting the navigation buttons to disabled state initially. The state is removed once their onLoadCode is executed. The click handler whch is added by the onLoadCode prevents the default action and checks whether ilTestPlayerQuestionEditControl is ready before its checkNavigation() function.


kergomard (changes requested):
Thank you very much for the PR @fneumann !

I think this is a good idea in general, I've just one issue with the changes in Javascript:
- I don't think most of them are needed. The only one that actually is, is adding `. "document.getElementById('{$id}').removeAttribute('disabled');";` twice. The `e.preventDefault()` just adds one more place we do that (there  are also some in the `ilTestPlayerQuestionEditControl.js` and there are a few `return false;` there that should have the same effect. And also setting `ready` in `ilTestPlayerQuestionEditControl.init()` seems redundant to me. If you see a good reason, why we should absolutely do that, the check should be within `ilTestPlayerQuestionEditControl.checkNavigation()`, but I would by far prefer it, if this goes.

Thanks again and best,
@kergomard

fneumann:
Hi @kergomard 

Thanks for your review although I don't understand it completely. Let me explain my motivation:

The various actions in Javascript are necessary to ensure that no incorrect navigation takes place until ilTestPlayerQuestionEditControl is fully initialized. In the solution, I take into account that we do not know whether the OnLoadCode for the button or for ilTestPlayerQuestionEditControl is executed first. 

1. With `removeAttribute` the button is activated after the click handler has been added. This prevents a standard action without a handler.
2. `preventDefault` in the handler suppresses the standard action from being executed on click if the call to `checkNavigation` fails due to a not yet existing ilTestPlayerQuestionEditControl.
3. the test on “ready” ensures that `checkNavigation` is only executed when it has the correct configuration.

You can simulate the behavior by commenting out the last line in `\ilTestPlayerAbstractGUI::populateQuestionEditControl`. Then no navigation should hapen when the button is clicked, even if it is already activated. In the real case of a slow connection, you can try clicking several times until the initialization is complete. Keeping the button visually inactive until the initialisation is finished would require the use of a timer based loop for checking. I thinks that is too much for the rare case.

kergomard:
Hi @fneumann 
Can you tell me how to provoke a case where the onloadcode for `ilTestPlayerQuestionEditControl.init()` would be executed after the onload for the `button`? I actually do not think this is possible, as internally in `il.Util.addOnLoad()` `$().ready()` is used and there the order is given (https://api.jquery.com/ready).

The rest follows from there.

Best,
@kergomard

fneumann:
I don't mean the order is random, but it depends on whether \ilTestPlayerAbstractGUI::populateQuestionEditControl is executed before e.g. \ilTestPlayerAbstractGUI::buildNextButtonInstance or \ilTestNavigationToolbarGUI::retrieveFinishTestButton. I think this dependency in order is quite hidden in the code and may be broken with further refactorings.

kergomard:
Hi @fneumann 

Yes, and we will need to make sure, that we do not break that. We should not introduce unneeded complexity and ugly hacks in JS to save us from our own mistakes that as far as I can see we didn't make yet. So, unless there is an absolute need to do this that I don't see right now,  I stick to my points above.

Thanks and best,
@kergomard

fneumann:
Hi @skergomard

It's good to read that you will keep the necessary order of the onload codes in mind when it comes to further refactorings. So I removed the checks for TestPlayerQuestionEditControl being loaded and ready when the navigation buttons are activated.

kergomard:
Thank you very much for adapting this @fneumann  and sorry this got stuck a while with me.

PR #9238, opened 2025-03-20
Group(-Reference): Bugfix 043739: Group Links in Dashboard Error

fhelfer:
https://mantis.ilias.de/view.php?id=43739


smeyer-ilias:
Thanks @fhelfer

PR #9242, opened 2025-03-20
Mantis #38284: Force local auth mode in local registration when not the default

mBeym:
https://mantis.ilias.de/view.php?id=38284

### DIFF
--- a/Services/Registration/classes/class.ilAccountRegistrationGUI.php
+++ b/Services/Registration/classes/class.ilAccountRegistrationGUI.php
@@ -358,6 +358,9 @@ class ilAccountRegistrationGUI
         }
 
         $this->userObj = new ilObjUser();
+        if ((int) $this->settings->get('auth_mode') !== ilAuthUtils::AUTH_LOCAL) {
+            $this->userObj->setAuthMode('local');
+        }
 
         $this->user_profile->skipGroup("preferences");
         $this->user_profile->skipGroup("settings");


mjansenDatabay (approved):

Services/Registration/classes/class.ilAccountRegistrationGUI.php
```diff
@@ -358,6 +358,9 @@ protected function createUser(int $a_role): string
         }
 
         $this->userObj = new ilObjUser();
+        if ((int) $this->settings->get('auth_mode') !== ilAuthUtils::AUTH_LOCAL) {
+            $this->userObj->setAuthMode('local');
+        }
```
thojou, line 363: Hey @mBeym,

Thanks for the changes.

Do we really need the condition around the `setAuthMode` call? As I understand it, at this point, we always want to enforce local auth mode, regardless of the configured setting. If that's the case, we could simplify the logic by directly setting it.

Best,
@thojou

mBeym: I don't know how this should be handled conceptually. Since "default" and "local" are different settings but "default" can be "local" I think this is the correct(er) way.

mjansenDatabay: The code @mBeym provides is correct. If the default auth mode is "local", we should use "default", otherwise explicitly "local".

Of course, this could be questioned in general, but this would probably have side effects, where logics would then only detect users with “default” authentication mode.

thojou: Thanks for the explanation. In that case, I’m okay with the changes.

Best,
@thojou

PR #9247, opened 2025-03-21
Certificate Bugfix #0044668: Import deleteDir throws Error with NFS runnning

fhelfer:
Although this is a step back from abandoning ilFileUtils and using flysystem, the certificate import doesn't seem to work with NFS. There has to be a FileStream that isn't properly being closed. Until the actual error cannot be found this is the only fix I found so far for https://mantis.ilias.de/view.php?id=44668


Services/Certificate/classes/Template/Action/Import/class.ilCertificateTemplateImportAction.php
```diff
@@ -110,6 +110,7 @@ public function import(
         );
 
         $unzipped = $unzip->extract();
+        $unzip->close();
```
mjansenDatabay: According to ILIAS 9 source code, there is no `close` method.

mjansenDatabay:
Fixed with https://github.com/ILIAS-eLearning/ILIAS/commit/3f208136d8f1853a38458fad04efe55cad03c7b1

PR #9248, opened 2025-03-24
UI/Input/Pagination: 38501/38505, use buttons with symbols for rocker navigation

nhaagen:
https://mantis.ilias.de/view.php?id=38501
https://mantis.ilias.de/view.php?id=38505

Replace glyphs in the pagination's nav with buttons + symbol.

(Does not apply for 9, where symbols  cannot be used on buttons, yet)


klees:
Hi @thibsy and @Amstutz,

could you have a look here?

Kind regards!

Amstutz:
thx a lot!

PR #9249, opened 2025-03-24
PRG: 44473, do not offer 'relevant' for draft nodes

nhaagen:
https://mantis.ilias.de/view.php?id=44473


klees:
Picked to trunk.

PR #9255, opened 2025-03-24
User/Import: Use `isset(...)` instead `is_null(...)`

mjansenDatabay:
This PR makes the XML parser for user XML documents more robust against missing "Id" and "Action" attributes.

The same attributes are checked with `isset(...)` in other methods of this class, so I suggest streamlining the checks.


kergomard:
Thank you very much @mjansenDatabay ! Merged and picked to ILIAS 10 and trunk.

PR #9256, opened 2025-03-25
Mantis 44504 (reopened): fix position write access

ingoj:
fix permission check for userForm (chekWriteAccess) and replace check for uploadUserPicture by call to this function

### DIFF
--- a/Services/User/classes/class.ilObjUserGUI.php
+++ b/Services/User/classes/class.ilObjUserGUI.php
@@ -1431,18 +1431,8 @@ class ilObjUserGUI extends ilObjectGUI
      */
     public function uploadUserPictureObject(): void
     {
-        if ($this->usrf_ref_id == USER_FOLDER_ID and
-            !$this->rbac_system->checkAccess('visible,read', $this->usrf_ref_id)) {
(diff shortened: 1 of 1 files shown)


kergomard:
Thank you very much @ingoj for the PR !

It is very much appreciated.

While looking through this I realized one more thing: We can get rid of the checks in `ilObjUserGUI::uploadUserPictureObject()` completely. There are just two calls to this function one from `ilObjUserGUI::saveObject()` and the other from `ilObjUserGUI::editObject()`. In both cases the rights have already been checked. To be sure that this function cannot be called directly from the interface, we need to remove the "Object" from its name. So, I would kindly ask you to:

- [ ] Rename the function `ilObjUserGUI::uploadUserPictureObject()`  to `ilObjUserGUI::uploadUserPicture()` and also adapt its usages.
- [ ] Make the function `ilObjUserGUI::uploadUserPictureObject()` `private`.
- [ ] Completely remove the right-checks from the function.

Am I missing something?

Thanks again and best,
@kergomard

ingoj:
changes to uploadUserPictures() are implemented as discussed.

kergomard:
Thank you very much @ingoj !

Best,
@kergomard

PR #9262, opened 2025-03-26
UI: introduce catamorphism for UI components

nhaagen:
reopen https://github.com/ILIAS-eLearning/ILIAS/pull/8812


thibsy (commented):
Hi @nhaagen and @klees,

- [ ] Terminology: you are talking about a `Component` "sub structure". I recently introduced the term [composite UI components](https://github.com/ILIAS-eLearning/ILIAS/blob/trunk/components/ILIAS/UI/docs/segregation-and-covariance.md#crossing-the-line-with-composite-ui-components), which are described as "components that accept other UI components, either during construction or through mutators". Are we talking about the same thing here or is the sub structure more? Could it possibly feature other UI components which are used to render a component?

- [x] `Closure` vs. `callable`: I was looking into the difference between those types and started wondering, if we should stick to one of those options for exposing things on our public interface. I prefer a `Closure` because they enforce using anonymous functions and make the parameter less ambiguous. A `callable` can be passed as `string` or `array` too, whereas `Closure` will always be an instance of this object. If you have no strong opinion, please change the type-hint accordingly.
- [x] Catamorphism I: the implementations - or anonymous functions - provided to `foldWith()` are not quite intuitive, I believe. I think it would feel more natural if previous results would be passed along as an argument - similar to the `array_reduce()` catamorphism. In addition, this will get rid of the instance property and the additional method on the public interface. Please try to implement something among the lines of the example below.

Kind regards,
@Amstutz and @thibsy

---

```php
/**
 * @param Closure(Component, array): mixed $fn
 */
public function foldWith(Closure $fn): mixed
{
    $clone = clone $this;
    $results = [];
    foreach ($clone->getComponents() as $component) {
        $results[] = $component->foldWith($fn);
    }
    return $fn($clone, $results);
}
```

klees:
Hi @thibsy,

since I was idea guy here, I'm the one who answers =)

* Terminology: What we are talking about goes further then your composite component. Every "composite component" would have a sub structure, but we would also expose components contained inside other components that are not passed by consumers during construction or otherwise, e.g. the ones that are only created during rendering.
* Closure vs. Callable: Fine for me, @nhaagen please change.
* Anonynmous Function: Regarding the natural feeling, this is different to me =) My view on this is: The nested construction is broken down along its structure. Hence, the broken down pieces are contained within the currently broken down structure. Also, with the proposed signature, some magic will never be unlocked: The feeling that the structure can indeed be changed in place. With the current implementation this isn't so, though, but passing the results along in a second structure will prevent that forever... Final remark: This is not like `array_reduce`, which will only break down a sequence, but also goes down the nesting. Otherwise: If you still want, I guess we can also work with the proposed structure of the callback. The other considerations I mentioned would probably stay theoretical for quite some time anyway. But _never_ take an `array_reduce` for a proper catamorphism =)

Kind regards

thibsy (changes requested):
Hi @klees,

Thx for your explanation and the brief conversation about this topic yesterday. 

I quickly repeat for @nhaagen and the general public as well: @klees and I discussed, that its actually _desired_ to have side-effects inside the structure - like storing the calculated result inside a `Component` - because we are _transforming_ the structure itself into something else. However, this is very theoretical and will not reach its full potential in the near future - if ever. Its also quite mind-boggling to understand this concept, so we fear this will not be used intuitively. Therefore we decided to stick with my proposal instead of a true catamorphism, which behaves more like an `array_reduce()` would.

Please implement this according to my previous review and rename the method to `reduceWith()`.

Kind regards,
@thibsy

nhaagen:
Here's the "reduceWith" instead of  catamorph-transforms

thibsy (commented):
Thx @nhaagen for implementing the changes!

There are some very minor nit-picks - while we wait for JF approval :).

- [x] `Component::reduceWith()`: the parameter name and the PHPDoc annotation do not match (`$f`/`$fn`).
- [x] `catamorph.php` Files: we should rename them to `reduce_with.php` for consistency.

Otherwise this LGTM, thx.

Kind regards,
@thibsy

thibsy (approved):
LGTM now, thx!

matthiaskunkel:
Jour Fixe, 14 APR 2025: We highly appreciate this suggestion and accept the PR for ILIAS 11.

PR #9263, opened 2025-03-26
Shibboleth: set user title instead of object title

iljalukin:
https://mantis.ilias.de/view.php?id=44713


chfsx:
Thanks a lot for the PR!

PR #9265, opened 2025-03-26
Release 10 SCORM paths

qualitus-dahme:


components/ILIAS/Scorm2004/templates/default/tpl.scorm2004.player.html
```diff
@@ -70,7 +70,7 @@
 			</div>
 		</div>
 
-		<script type="text/javascript" src="./Modules/Scorm2004/scripts/rteconfig.js"></script>
+		<script type="text/javascript" src="./components/ILIAS/Scorm2004/scripts/rteconfig.js"></script>
```
mjansenDatabay, line 73: IMO this will not work, since the browser will not have access to this script in ILIAS 10.x.

components/ILIAS/Scorm2004/classes/class.ilSCORM13PlayerGUI.php
```diff
@@ -487,7 +490,7 @@ public function getPlayer(): void
         if ($this->slm->getCacheDeactivated()) {
             $this->tpl->setVariable('JS_SCRIPTS', 'ilias.php?baseClass=ilSAHSPresentationGUI' . '&cmd=getRTEjs&ref_id=' . $this->ref_id);
         } else {
-            $this->tpl->setVariable('JS_SCRIPTS', '../components/ILIAS/Scorm2004/scripts/buildrte/rte-min.js');
+            $this->tpl->setVariable('JS_SCRIPTS', 'components/ILIAS/Scorm2004/scripts/buildrte/rte-min.js');
```
mjansenDatabay, line 493: IMO this will not work, since the browser will not have access to this script in ILIAS 10.x.

components/ILIAS/Scorm2004/classes/class.ilSCORM13PlayerGUI.php
```diff
@@ -457,7 +460,7 @@ public function getPlayer(): void
         // include_once("../components/ILIAS/YUI/classes/class.ilYuiUtil.php");
         // $this->tpl->setVariable('YUI_PATH', ilYuiUtil::getLocalPath());
         // $this->tpl->setVariable('TREE_JS', "../components/ILIAS/UIComponent/NestedList/js/ilNestedList.js");
-        $this->tpl->setVariable('TREE_JS', "../components/ILIAS/Scorm2004/scripts/ilNestedList.js");
+        $this->tpl->setVariable('TREE_JS', "components/ILIAS/Scorm2004/scripts/ilNestedList.js");
```
mjansenDatabay, line 463: IMO this will not work, since the browser will not have access to this script in ILIAS 10.x.

components/ILIAS/Scorm2004/templates/default/tpl.scorm2004.player.html
qualitus-dahme: We checked this, actually the script is accessable. This script has been copied to the public folder by the component init function.

components/ILIAS/Scorm2004/classes/class.ilSCORM13PlayerGUI.php
qualitus-dahme: We checked this, actually the script is accessable. This script has been copied to the public folder by the component init function.

components/ILIAS/Scorm2004/classes/class.ilSCORM13PlayerGUI.php
qualitus-dahme: We checked this, actually the script is accessable. This script has been copied to the public folder by the component init function.

mjansenDatabay:
HI @qualitus-dahme  ,

Okay, I see: You are using the same subfolder structure `components/ILIAS/Scorm2004/scripts` in the public folder.

Hmmmm, I also understood how you were able to achieve this, by using the anonymous class ... I am not sure if this is supported, even if it works. The JF decided to use a flat structure for all assets.

See:
- https://github.com/ILIAS-eLearning/ILIAS/pull/7019#issuecomment-2055732880
- and the final JF decision: https://github.com/ILIAS-eLearning/ILIAS/pull/7019#issuecomment-2056365041

IMO it is okay to make exceptions if they are really necessary for the SCORM component, but we should discuss them publicly with others.

Kind regards,
Michael

QualitusHartwig:
@qualitus-dahme FYI picked it to trunk to prevent future conflicts. We should also check if there are any other PRs (at least for 10) that were not picked to trunk yet to ensure trunk is functional.

PR #9266, opened 2025-03-26
Components revision SCORM paths

qualitus-dahme:


QualitusHartwig:
@qualitus-dahme Picked to 11 and trunk
See https://mantis.ilias.de/view.php?id=46826

PR #9268, opened 2025-03-27
Fix page editor plugin export in T&A

fneumann:
See https://mantis.ilias.de/view.php?id=44719

This pull request solves the issue by providing the handling of page editor plugins from the COPage service also in the Test and TestQuestionPool components. 

To avoid code duplication, functions from ilCOPageExporter and ilCOPageImporter are moved to a new class `ilPageComponentPluginExportImportStore`, which is used by the export and import in the CoPage, TestQuestionPool and Test components. 

The static functions provided in `ilPageComponentPluginExporter` and `ilPageComponentPluginImporter` are rewritten to use the new class as a store, but currently they must be kept for compatibility with existing plugins.

### DIFF
--- a/Modules/Test/classes/class.ilObjTest.php
+++ b/Modules/Test/classes/class.ilObjTest.php
@@ -4170,6 +4170,7 @@ class ilObjTest extends ilObject implements ilMarkSchemaAware
             $page_object = new ilAssQuestionPage($question_id);
             $page_object->buildDom();
             $page_object->insertInstIntoIDs((string) $inst);
+            ilPageComponentPluginExportImportStore::getInstance()->extractPluginProperties($page_object);
             $mob_ids = $page_object->collectMediaObjects(false);
             $file_ids = ilPCFileList::collectFileItems($page_object, $page_object->getDomDoc());
             $xml = $page_object->getXMLFromDom(false, false, false, "", true);
--- a/Modules/Test/classes/class.ilTestExporter.php
+++ b/Modules/Test/classes/class.ilTestExporter.php
@@ -35,6 +35,7 @@ class ilTestExporter extends ilXmlExporter
     private ilCtrl $ctrl;
     private ilComponentRepository $component_repository;
     private QuestionInfoService $questioninfo;
+    private ilPageComponentPluginExportImportStore $pc_plugin_store;
 
     public function __construct()
     {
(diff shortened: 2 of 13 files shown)


fneumann:
@alex40724 
As far as I understood @kergomard in the TA Tech Squad, he would accept the changes in T&A if you accept them for the COPage service. Could you have a look?

kergomard:
I remove my assignment here. Please add me again, if there is anything I will need to do.

Best,
@kergomard

alex40724:
@fneumann Thanks for the PR and sorry for the delay. I had a second look at the code today. It all looks feasible, but it is too much code to tell, if everything works as expected without doing detailed tests with real data. I did not have the time for this yet.

fneumann:
Hi @alex40724 

Thanks for your reply. For ILIAS 9 I built a workaround in the plugin https://github.com/DatabayAG/PCLimitedMediaPlayer, and I hope that will work in ILIAS10, too. I can rewrite the PR for ILIAS 11 to give you more time for testing and without the danger of affecting productive instances yet.

alex40724 (approved):
@fneumann  Ok, to move forward I suggest the following. I read the code in detail today, again. I do not see any general issues with this, so I will just review this and assign it to you.

You can commit this to ILIAS 9 and merge this to all upper branches, if you test the behaviour with a plugin in these versions and if you volunteer to fix issues that may arise due to this change.

The code is fine. I am not a big fan of static factory calls like

> ilPageComponentPluginExportImportStore::getInstance()

but I might refactor this in the trunk at a later time.

Thanks for your patience.

fneumann:
Thank you @alex40724 for the acceptance. Meanwhile, we found an ugly solution for our plugin in ILIAS 9 and 10 by overloading a core class. So at the moment there is no need to risk any problem in the production branches.

I propose to update this PR for the release branch of ILIAS 11 once it is created. Certainly, I volunteer to fix issues related to this change.

@kergomard would you agree with that procedure?

kergomard (commented):
Hi @fneumann 

I'm unsure if it is worth the bother for ILIAS 11 as in ILIAS 12 we will hopefully have a better import/export in the test and if we then can do things the right way, there would be more to gain. This is something to be discussed with @thojou , though.

I can personally explain very well, why we do not support page plugins in the test, so I see no urgent need for this.

Implementation wise this to me looks all wrong. I don't know if @alex40724 sees this as officially part of the public interface of the Page Editor, if it is it should go into the `$DIC` and the new integration mechanism. If it isn't, then I would not implement this. It looks to me like a way to obscure the availability of a functionality.

Thanks and best,
@kergomard

PR #9269, opened 2025-03-27
Shibboleth: allow tags in login page instructions

iljalukin:
https://mantis.ilias.de/view.php?id=44725


chfsx:
Thanks a lot for the PR!

PR #9270, opened 2025-03-28
T&A 35875: Prevents a multiple choice question to give negative points when answered fully correctly

matheuszych:
[Mantis: 35875](https://mantis.ilias.de/view.php?id=35875)

A possible fix for the generall issue is to prevent a multiple choice question to have a negative total of maxium points. In case you answer a question perfectly (highest possible scores) you shouldn't be able to get negative points. On the other hand if you do not answer it quiet "perfect" you should be able to receive negative points for the question.

The inputs cannot be validated on their own, because there combination is what has to be evaluated.

For future reference: When the forms get refactored to the new UIForms this should be changed into a transformation/validation.

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assMultipleChoice.php
+++ b/Modules/TestQuestionPool/classes/class.assMultipleChoice.php
@@ -483,7 +483,7 @@ class assMultipleChoice extends assQuestion implements ilObjQuestionScoringAdjus
      */
     public function getMaximumPoints(): float
     {
-        $total_max_points = 0;
+        $total_max_points = 0.0;
         foreach ($this->getAnswers() as $answer) {
             $total_max_points += max($answer->getPointsChecked(), $answer->getPointsUnchecked());
         }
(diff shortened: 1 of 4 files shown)


matheuszych:
Already reviewed by @thojou.

kergomard:
Hi @matheuszych 

Thank you very much for the PR !

I've only one little issue:
- [ ] I don't like the naming combined with the return type of the function `assMultipleChoiceGUI::handleMaxPointsNotNegative(ilPropertyFormGUI $form): bool`. I don't think this is correct. I would kindly ask you to change it to `assMultipleChoiceGUI::checkMaxPointsNotNegative(ilPropertyFormGUI $form): bool`. The return-value should be `true`, if the check is passed and `false`, when it isn't.
- [ ] Could you please also set `$total_max_points` to a `float` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9270/files#diff-a2b169b5bf9f80a17c6424d0472445235a33f80d5790027560f083f608d9d35eR486).

Thanks again and best,
@kergomard

matheuszych:
Hello @kergomard
Thank you for the review!

I adjusted the code accordingly.

Best regards
@matheuszych

kergomard:
Thank you very much for updating the PR @matheuszych !

PR #9271, opened 2025-03-28
[Bugfix 44735]Add prefix to cache key to prevent language mix-up across ILIAS tenants

jheim2:
Adding prefix in the cache key prevents cache conflicts when multiple ILIAS tenants run on the same PHP-FPM server. This avoids incorrect language assignments caused by shared cache entries.

### DIFF
--- a/src/Cache/Adaptor/BaseAdaptor.php
+++ b/src/Cache/Adaptor/BaseAdaptor.php
@@ -1,4 +1,4 @@
-<?php
+    <?php
 
 /**
  * This file is part of ILIAS, a powerful learning management system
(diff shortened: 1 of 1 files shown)


jheim2:
https://mantis.ilias.de/view.php?id=44735

dsstrassner:
I assume @chfsx is the correct assignee. If I am wrong, please reassign.

src/Cache/Adaptor/BaseAdaptor.php
```diff
@@ -1,4 +1,4 @@
-<?php
+    <?php
```
mjansenDatabay: Please remove the whitespace characters.

jheim2: Sorry, I did this after an intense social event following DevConf, using the frequently crashing WiFi on ICE on my way back home.

chfsx:
The changes would have to be adjusted a little bit so that we can save some performance. `buildContainerPrefix` is called very often, actually with every read or write access to a value in the cache. This was not really a problem before, because it was simply a text string. With your change, a `getcwd()` would now be executed every time, which is of course unnecessary because it returns the same value every time. This means that it should only be done once.
I am therefore closing the PR and solving this directly. Here is the link to the changes: https://github.com/ILIAS-eLearning/ILIAS/commit/0562a86eb8981e3c97afca1d1a847ad1d1363211

PR #9272, opened 2025-03-28
Glossary: remove some unused code, old public access export (bt-42313)

schmitz-ilias:
This PR removes the old public access export mechanism from Glossary, see [42313](https://mantis.ilias.de/view.php?id=42313), as well as some other unused code related to the legacy settings. I didn't include DB update steps to remove the now unused columns, let me know if you want me to add those.


tfamula:
Thank you very much, @schmitz-ilias! I already picked this to trunk.
Could you be so kind and provide a PR with DB update steps to remove the now unused columns? IMO, this is sufficient for trunk.

PR #9273, opened 2025-03-31
Events: use Component-bootstrap to 'register' event definitions

nhaagen:
This initializes the eventhandling's ArtifactObjective with event definitions from components' bootstrap.
Anouncement of events (listen/raise) moved from the XML files into component.php


klees:
Hi JF,

just want to announce this change. It moves the system towards the goals of the [Component Revision](https://docu.ilias.de/go/wiki/wpage_7295_1357), specifically "1.1 Modernize Services/Component".

This is a mere 1 to 1 translation of the previous XML-parsing to the new component integration approaches without touching the innards of the event handling. The rest of the architecture there still relies on static calls and instantiation by name of event handlers. So this is only an incremental step to make event handling compliant with the target architecture of the Component Revision. But: Steady wins the race...

Kind regards!

components/ILIAS/EventHandling/classes/Definition.php
```diff
@@ -0,0 +1,41 @@
...
+
+declare(strict_types=1);
+
+namespace ILIAS\EventHandling;
+
+class Definition
+{
+    public function __construct(
+        protected string $component,
+        protected string $type,
```
thibsy, line 27: Could we make this an `enum` =)?

components/ILIAS/EventHandling/classes/Definition.php
```diff
@@ -0,0 +1,41 @@
...
+        protected string $type_specification
+    ) {
+    }
+
+    public function toArray(): array
+    {
+        return [
+            'component' => str_replace('ILIAS\\', 'components/ILIAS/', $this->component),
```
thibsy, line 35: This only works for ILIAS components. Do you already have some idea how plugins or non-ILIAS components can hook into this?

thibsy (commented):
Hi @nhaagen and @klees,

Glad to see the component revision moving forward =). I have just looked over the changes out of curiosity and have left two comments / questions for you.

Thx for this proposal and kind regards,
@thibsy

matthiaskunkel:
Jour Fixe, 28 APR 2025: We highly appreciate this suggestion and accept the PR for trunk.

components/ILIAS/EventHandling/classes/Definition.php
nhaagen: you can register events like

```
        $contribute[\ILIAS\EventHandling\Definition::class] = static fn() =>
            new \ILIAS\EventHandling\Definition(
                'XXX/myspecialevent/myClass',
                'listen',
                'yyy/myspecialevent/something'
            );
        $contribute[\ILIAS\EventHandling\Definition::class] = static fn() =>
            new \ILIAS\EventHandling\Definition(
                'Plugins/myPlug/myPluginClass',
                'listen',
                'yyy/myspecialevent/something2'
            );
```
and 
`->raise('yyy/myspecialevent/something', $event, $parameter);` or
`->raise('yyy/myspecialevent/something2', $event, $parameter);`


EventHandler will then look for class ilmyClassAppEventListener  / ilmyPluginClassAppEventListener
and call it with $component='yyy/myspecialevent/something';

As far as I can see there is no further adjustement necessary; the replace() is merely a convenience for using self::class in the Component.php

components/ILIAS/EventHandling/classes/Definition.php
nhaagen: sure :)

nhaagen: The type? I guess so, but I'd rather not blow up this PR further right now.

dsstrassner:
It appears that this PR didn't make it into ILIAS 11.

PR #9275, opened 2025-03-31
LearningSequence: Use ENUM type instead of constants for page types

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=44630


klees:
JS-tests seem to fail for [an unrelated reason](https://github.com/ILIAS-eLearning/ILIAS/commit/7d5a213f617977dced9676be341ae5e40d56045c#commitcomment-154795624), so I merge this anyway. Thanks @mjansenDatabay!

klees:
Picked to trunk and release_10.

PR #9278, opened 2025-03-31
[FIX] 0042136: Members no longer able to download multiple files

chfsx:
This fixes the issue described in https://mantis.ilias.de/view.php?id=42136


alex40724:
Thanks for the PR!

PR #9286, opened 2025-04-01
Fix page editor plugin copy in T&A

fneumann:
https://mantis.ilias.de/view.php?id=44761


kergomard:
Thank you very much for the PR @fneumann !

I didn't test this, but it looks very plausible. Merged and picked to ILIAS 10 and trunk.

Best,
@kergomard

PR #9288, opened 2025-04-01
PRG: resolve specific placeholders

nhaagen:
https://mantis.ilias.de/view.php?id=44336
also relevant for 9, fixes regression by https://github.com/ILIAS-eLearning/ILIAS/pull/9190


klees:

PR #9297, opened 2025-04-02
Fix missing method getTitleForHTMLOutput on ilObject in ilObjTestGUI (#44772)

Hufschmidt:
See https://mantis.ilias.de/view.php?id=44772

TL;DR: Fixes copy-paste issue in https://github.com/ILIAS-eLearning/ILIAS/commit/7fa3eaaa98cca0e167470a0c87839096fcfa9ff7


kergomard:
Thank you very much for the PR @Hufschmidt !

I don't think this is right though as this would be sending a string, that has no tags to strip tags, so I fixed this and the other occurrence directly with: https://github.com/ILIAS-eLearning/ILIAS/commit/1597a8c6b600a0cca248547896462448f83612f1

Thanks again and best,
@kergomard

PR #9298, opened 2025-04-02
Fix: LTI icon in the menu for adding new repository objects.

ZallaxDev:
Currently, when a user creates a new LTI Consumer Object, the icon is not displayed. With this change, ILIAS will correctly display the LTI icon in this menu.


klees:
Picked to trunk.

klees:
Dear @ZallaxDev,

just  a quick info regarding our processes: The labels for PRs are explained [here](https://github.com/ILIAS-eLearning/ILIAS/blob/release_9/docs/development/contributing.md#list-of-labels). I removed "bugfix" and added "improvement", since there is no mantis issue linked here.

The general idea is: PRs that link an issue in mantis are more solid and maybe even more serious, because they solve a known issue that can be referenced in release notes, can be checked according to the steps to reproduce and such. Hence, different labels for "bugfix" and stand alone improvements, where criterion is "does this PR link a mantis issue".

No problem, though =)

Kind regards!

PR #9300, opened 2025-04-03
MediaCast/CustomSkins: Make JS code extraction more robust

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=43066


alex40724:
Thanks for the PR.

PR #9069, opened 2025-02-18
UICore: 44208, reset parameter to relative path for Template

nhaagen:
https://mantis.ilias.de/view.php?id=44208

with https://github.com/ILIAS-eLearning/ILIAS/pull/8873,  $plugin->getDirectory() will return an absolute path,
ilTemplate expects a relative path, though.
This has ilTemplate accept both.

PR #9302, opened 2025-04-03
Plugins: adjust path to public/Customizing/global/Services

nhaagen:
https://mantis.ilias.de/view.php?id=44596

includes https://github.com/ILIAS-eLearning/ILIAS/pull/9069 (for https://mantis.ilias.de/view.php?id=44208)
and fixes https://mantis.ilias.de/view.php?id=44425

Main issue was, that the Slot-providing component is in components/ILIAS whith no reference to "Services" or "Modules", while the plugin needs to know this path. 
I removed ilPluginSlotInfo::getPath and re-added the type to ilPluginInfo.


klees:
I will ask a colleague to walk through plugin testcases again.

klees:
Picked to trunk.

PR #9303, opened 2025-04-03
11/object/move translation table to ks

kergomard:
Hi all

This is the implementation of https://docu.ilias.de/go/wiki/wpage_8066_1357 . I used this chance to consolidate namespaces and to "integrate" the translations into the `ObjectProperties` and correspondingly this contains a lot of changes all around the code-base.

Three important points to note:
- I removed the option to disable the description on the translation screen. This was only done so that the Object-List wouldn't need to show the field. I don't think it is right to create such an exception. Objects have descriptions and this will pop up in different contexts now: there e.g. is no option to add a title only to the ks forms, it is always title and description.
- I changed "Master Language" to "Base Language" (as it is in German). I think we really should get rid of any language of domination.
- I had to move the property `ILIAS\HTTP\GlobalHttpState $http` to `ilObjectGUI` and so I needed to change child classes implementing this property themselves.

I add the label "jour fixe" to make everybody aware of this and I add you @alex40724  as reviewer as many changes concern your components and so you might become aware of this before. If you might have a look at `LearningModule`, `COPage`, and maybe `Wiki` this would be kind.

Thank you very much and best,
@kergomard


klees (approved):

matthiaskunkel:
Jour Fixe, 14 APR 2025: Stephan gave a short presentation of this PR and asks all developers to have a look at the suggested changes and give feedback. We will discuss and decide upon the PR on the next JF at 28 APR 2025.

matthiaskunkel:
Jour Fixe, 28 APR 2025: We highly appreciate this suggestion and accept the PR for trunk.

kergomard:
Thank you very much ! After announcing this twice at the JF and having it accepted for trunk, I merge.

PR #9305, opened 2025-04-03
Logging: Fix monolog 3.9 compatibility issues

lscharmer:
This PR fixes Monolog 3.9 incompatibilities:
- The processors are given a `Monolog\LogRecord` instead of an array now.
- Move `suid` record entry to `extra` as you cannot add arbitrary keys to the `Monolog\LogRecord` directly.

Mantis Issue: https://mantis.ilias.de/view.php?id=44782


smeyer-ilias (approved):
Thanks @lscharmer

PR #9306, opened 2025-04-03
11/Components/Plugins: adjust composer to mkdir public/Customizing/global/plugins

nhaagen:
Actually, plugins should be just components, but for now and to ease stepping up: 
allow for plugins to reside in public/Customizing/global/plugins (https://mantis.ilias.de/view.php?id=44596)

adds 89e5bc7605759ef7bc31fdf6d6f8d1f90e4beec8 from 10


klees:
This is on me, cherry-pick was partial only...

PR #9307, opened 2025-04-04
Add `jquery` as npm Dependency for ILIAS 11

thibsy:
This PR adds `jQuery` as NPM dependency.

Usage:
* In many components, searching for `$(` results in 3700+ occurrences

Wrapped By:
* Not applicable

Reasoning:
* `jQuery` has been introduced many years ago as a DOM abstraction and utility (XMLHttpRequest, Event Handling etc.) library. Nowadays, many of this functions could be replaced by the "Vanilla JS" API. Nevertheless, I doubt that we could tackle the migration in a timely manner for ILIAS 11 and most likely 12 either. It is unlikely to remove this dependency for all components outside of a big project with according funding and organisational resources.

Maintenance:
* `jQuery` is actively maintained and has an existing security policy (https://github.com/jquery/jquery/blob/e06ff08849057cd099365bf43598c8952fe9956d/SECURITY.md).

**Important:** It might be necessary to also include `"jquery-migrate": "^3.0.0",`, maybe the developer who introduced this can explain this further.

Links:
* NPM: https://www.npmjs.com/package/jquery
* GitHub: https://github.com/jquery/jquery
* Documentation: https://api.jquery.com/

### DIFF
--- a/components/ILIAS/jQuery/jQuery.php
+++ b/components/ILIAS/jQuery/jQuery.php
@@ -36,8 +36,8 @@ class jQuery implements Component\Component
         $contribute[Component\Resource\PublicAsset::class] = fn() =>
             new Component\Resource\NodeModule("jquery-migrate/dist/jquery-migrate.min.js");
         */
-        /* $contribute[Component\Resource\PublicAsset::class] = fn() =>
-            new Component\Resource\NodeModule("jquery/dist/jquery.js"); */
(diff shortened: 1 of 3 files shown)


matthiaskunkel:
Jour Fixe, 14 APR 2025: Even though we would like to get rid of jquery, we also need this dependency in ILIAS 11 and therefore accept this PR and the dependency for ILIAS 11.

PR #9308, opened 2025-04-04
T&A 44733: Fixes invalid data types in assFormulaQuestionResult::getReachedPoints method

matheuszych:
[Mantis: 44733](https://mantis.ilias.de/view.php?id=44733)

In some cases the values passed to the `assFormulaQuestionResult::getReachedPoints` method did have a incorrect (unusual) datatype. To resolve this issue, I tried to streamline the inputs to use more suitable datatypes.

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assFormulaQuestion.php
+++ b/Modules/TestQuestionPool/classes/class.assFormulaQuestion.php
@@ -933,11 +933,12 @@ class assFormulaQuestion extends assQuestion implements iQuestionCondition, ilAs
 
         $points = 0;
         foreach ($this->getResults() as $result) {
+            $unit_id = $user_solution[$result->getResult() . '_unit'] ?? null;
             $points += $result->getReachedPoints(
                 $this->getVariables(),
                 $this->getResults(),
                 $user_solution[$result->getResult()] ?? '',
-                $this->unitrepository->getUnit($user_solution[$result->getResult() . '_unit'] ?? 0),
+                $unit_id !== null ? $this->unitrepository->getUnit($unit_id) : null,
                 $this->unitrepository->getUnits()
             );
         }
--- a/Modules/TestQuestionPool/classes/class.assFormulaQuestion.php
+++ b/Modules/TestQuestionPool/classes/class.assFormulaQuestion.php
@@ -892,7 +892,7 @@ class assFormulaQuestion extends assQuestion implements iQuestionCondition, ilAs
         }
         $solutions = $this->getSolutionValues($active_id, $pass, $authorizedSolution);
         $user_solution = array();
-        foreach ($solutions as $idx => $solution_value) {
+        foreach ($solutions as $solution_value) {
             if (preg_match("/^(\\\$v\\d+)$/", $solution_value["value1"], $matches)) {
                 $user_solution[$matches[1]] = $solution_value["value2"];
                 $varObj = $this->getVariable($solution_value["value1"]);
@@ -906,7 +906,9 @@ class assFormulaQuestion extends assQuestion implements iQuestionCondition, ilAs
                 if (!array_key_exists($matches[1], $user_solution)) {
                     $user_solution[$matches[1]] = array();
                 }
-                $user_solution[$matches[1]]["unit"] = $solution_value["value2"];
+                $user_solution[$matches[1]]["unit"] = $this->unitrepository->getUnit(
+                    $this->refinery->kindlyTo()->int()->transform($solution_value["value2"]),
+                );
             }
         }
         //vd($this->getResults());
@@ -917,7 +919,7 @@ class assFormulaQuestion extends assQuestion implements iQuestionCondition, ilAs
                 $this->getVariables(),
                 $this->getResults(),
                 $user_solution[$result->getResult()]["value"] ?? '',
-                $user_solution[$result->getResult()]["unit"] ?? '',
+                $user_solution[$result->getResult()]["unit"] ?? null,
                 $this->unitrepository->getUnits()
             );
         }
@@ -931,14 +933,11 @@ class assFormulaQuestion extends assQuestion implements iQuestionCondition, ilAs
 
         $points = 0;
         foreach ($this->getResults() as $result) {
-            $v = isset($user_solution[$result->getResult()]) ? $user_solution[$result->getResult()] : null;
-            $u = isset($user_solution[$result->getResult() . '_unit']) ? $user_solution[$result->getResult() . '_unit'] : null;
-
             $points += $result->getReachedPoints(
                 $this->getVariables(),
                 $this->getResults(),
-                $v,
-                $u,
+                $user_solution[$result->getResult()] ?? '',
+                $this->unitrepository->getUnit($user_solution[$result->getResult() . '_unit'] ?? 0),
                 $this->unitrepository->getUnits()
             );
         }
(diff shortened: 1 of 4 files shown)


matheuszych:
Already reviewed by @thojou.

kergomard:
Hi @matheuszych 

Thank you very much for this PR. 
I started looking at this and fell and fell and fell always deeper into the rabbit-hole. So: Sorry, I will have quite a few things here.

First and foremost, what got me started: I wouldn't have started looking deeper into this, were it not for [this line](https://github.com/ILIAS-eLearning/ILIAS/pull/9308/files#diff-0f3fcdefc4dddc4a1ef07b59eb6e2bb049fc26e0118ac2d306bebe9fd65f62f1R169). This looks plain wrong. If this would be what we need to do, to get this solution to work: I think we should roll it back. We would be introducing an unclear amount of database round trips with all the associated costs. Now, luckily I don't think it is necessary.  I only found one place where this class is created in `ilTestSkillEvaluation::determineReachedSkillPointsWithSolutionCompare()`. I would thus suggest to initialize the repository in `ilTestSkillEvaluation::__construct()` and then hand it in. This is really lightweight, as loading happens later, so nothing to worry about, if it isn't used. So, I think this is solvable ;-).

As a next step I looked into `assFormulaQuestionResult::getReachedPoints()` and first wrote this nit-pick:
- Please change the check [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9308/files#diff-b8b4411675745afd110e35777522dc8a3ba491c2581e8bcc3eb8974f70bd3ca4R525) from `is_numeric()` to `!== null`. I know: Same difference, but not the same semantics: The latter is more precise as this is, what you are actually checking. You absolutely know, that it is numeric, if it isn't `null`. 

...and realized that the whole function is just flaky anyways:
- We run through roughly a hundred lines of code where `$value` is a `string` and then, suddenly, it becomes a `?float`? ...and only then the value is checked if it is a number? So I looked deeper and found that I can technically understand why this is right, it just is very hard to parse and as we need to touch this, we should make this clearer, otherwise we will run into further issues in a spell.

I've a suggestion (line numbers refer to your new code):
- On a trivial level: Remove the `else`-block starting on line 428 to get one less indentation.
- Split the retrieval of `$result` in its own function (I think this should be lines 430 - 454 and lines 500 - 508).
- Split the retrieval of `$frag_value` into its own function (lines 457 - 497). In here we are also setting a variable `$check_fraction` of which I never see a use, just assignments, ...and there are more weird things happening in here as we may or may not have a `$frac_value`, which then leads to your newly added `isset($frac_value)` on line 511. I think we should make sure to always have this value (or just to not need it). This function will need to be called before the previous one, if I'm not completely mistaken.
- Split the calculation of `$points` from `$value` and potentially `$frac_value` to its own function (naming is not so easy here, I would suggest `calculatePointsForAnswerValue(string $value/$frac_value)`. Before calling this, points will always be `0`, so the line 429 (`$points = 0`) can go.
- And finally split a function `addPointsFromAnswerUnit(float $previous_points, assFormulaQuestionUnit $unit): float`(please also rename `$base1` and `$base2` to `$answer_base_unit` and `$solution_base_unit` respectively). Additionally we should not do the reassignments to `$base1` and `$base2` if we have an object. Why are we expecting `$base1` and `$base2` to be anything useful, if they are not `assFormulaQuestionUnit`s? If we need to do this checks, we should actually stop evaluation at this point and assume something is wrong here.
- Please make sure to not change types of variables in the middle of things. I know, PHP just swallows it, but please make these changes, if they are absolutely unavoidable, very obvious by creating a new $variable.

I know, this is quite a bit to ask, but I think we should do this and test it thoroughly. My feeling is: We are not really sure, what we are actually doing here and this is a bad sign. If you do not feel like doing this for ILIAS 9, please simply just move the construction of the repository up in this PR and do the rest in a PR for ILIAS 10.

What do you think?

Thanks again and best,
@kergomard

matheuszych:
Hello @kergomard

First of all thank you for your thorough review. No need to be sorry. I believe we both have the same intention to write as highly qualitative code and this is only possible thru such reviews.

About the first point. I realised that we can access the `ilUnitConfigurationRepository` directly from the question object. So there was never the need to inject it to begin with.

I changed the `is_numeric` to a not null check for the `$value`. I like to use checks like that since I believe it be slightly more exact. You can be more sure about the values content even when you do not have the outer context. But that is just me.

I am not a huge fan of this method either but I initialy did not want to refactor it completely as it would probably be out of scope for this fix. There a multiple things there that do not make sense such as dead variables or variables with seamingly random data types. Please take another look at the refactored changes.

Best regards
@matheuszych

kergomard (changes requested):
Thank you very much for the update  @matheuszych !

Still there are a few things:
- [ ] I don't think we should do [this](https://github.com/ILIAS-eLearning/ILIAS/pull/9308/files#diff-498f6047f032462823f878b0200920c74b32188507aed6136fbd6d26ae59a855R940). The problem is we are assuming that we can just ask for unit `0` and nothing will come back. This might be true or not, but we shouldn't even be asking. So: If the value is set, we call the function, if it isn't, we don't and just pass `null`.
- [ ] Please rename the function `assFormulaQuestionResult::evaluateResultAsFromula()` to `assFormulaQuestionResult::caculateResult()`. The result is not a formula, so we cannot evaluate it as a formula, what we can do though, is calculate it from the formula. After staring at it for a while: This is actually what we are doing. By the way (and there will be one more of those later): If you use words like "evaluate" or "handle" or "manage" in function or class names you can be more or less sure to get it back from me, as they don't tell us what the function/class actually does. What are you evaluating? If it is good or evil? Right or Wrong? Executable? Something that "handles" something is equivalent to "a thingy that does something, but I've no idea what".
- [ ] Please move the addition of units to the result on lines [432-438](https://github.com/ILIAS-eLearning/ILIAS/pull/9308/files#diff-b8b4411675745afd110e35777522dc8a3ba491c2581e8bcc3eb8974f70bd3ca4R432-R438) into the previous function.
- [ ] I don't think the function `assFormulaQuestionResult::handleValueByResultType()` does, what you think it does ;-). When no result-type is set (last arm of the `switch`-statement), `$frac_value` will be an empty string and the corresponding multiplication on line 523 will go sideways and I assume, without looking into the `ilMath::_mul()`-function that `$value` will be an empty string from then on. The line 499 is unneeded as we are never going to use the variable. And please rename the function `assFormulaQuestionResult::handleValueByResultType()` to `assFormulaQuestionResult::transformResultAccordingToType()` and make sure it does always returns `?float`, there is no reason to leave this for later.
- [ ]  As said: Please use `!== null` and not `is_float()`. I understand why you are doing it, but I do not agree and actually think it is a bad habit, as I think that you achieve exactly the opposite of what you are trying to achieve. You are removing information. If I get there I ask myself: "Why do we need to check, if the variable `$float_value` is a `float`?" and I will need to look around to see: "Ohh, yes, it is not simply lying, it might also be `null`." While the `!== null` tells me that instantly, without needing to look at any other line. Additionally: If this variable contains anything that is not `?float` all bets are off and things should go sideways.

I think, I can live with it then. The level of certainty that we know what we are doing is still not satisfactory, but at least it is acceptable.

Thank you for all the efforts!

Best,
@kergomard

matheuszych:
Hello @kergomard 
Thank you for your further review. I implemented the requested changes.

Best regards
@matheuszych

kergomard:
I found a few things that I didn't like yet, but thought it would be easier, if I would fix them directly, ...and then I found a few errors and a few hours passed. There was still quite a bit I didn't understand, ...and there still is. The biggest problem I didn't fully resolve: We have a lot of duplicate code in there.

I did some additional changes with: https://github.com/ILIAS-eLearning/ILIAS/commit/e1d0f98c2b39639179c2e7f64d972dd395062109

This only matters if complex answers are enabled and I'm pretty sure it has been a while since they functioned as expected, ...but let us see, if I got something wrong.

Best,
@kergomard

PR #9309, opened 2025-04-04
T&A 44764: Fixes end datetime of availability period in test settings being rese…

matheuszych:
[Mantis: 44764](https://mantis.ilias.de/view.php?id=44764)

The `activation_ending_time` setting in a test was always being overwritten with the date `01.01.1970` (timestamp 0). The error seems to be caused by never applying the data from the form.

### DIFF
--- a/Modules/Test/classes/MainSettings/class.ilObjTestSettingsMainGUI.php
+++ b/Modules/Test/classes/MainSettings/class.ilObjTestSettingsMainGUI.php
@@ -539,17 +539,23 @@ class ilObjTestSettingsMainGUI extends ilTestSettingsGUI
 
     private function saveAvailabilitySettingsSection(array $section): void
     {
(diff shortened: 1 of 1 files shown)


matheuszych:
Already reviewed by @thojou.

kergomard:
Hi @matheuszych 

Thank you very much for the PR! Please remove the array deconstruction here, and just use the corresponding keys as it was before. Straight forward is better: With your change I need twice as much time to read and understand the same thing.

Thank you very much and best,
@kergomard

matheuszych:
Hello @kergomard 
Thank you for your review. I removed the array deconstruct.

Best regards
@matheuszych

kergomard:
Thank you very much @matheuszych ! Merged and picked to ILIAS 10 and trunk.

PR #9310, opened 2025-04-04
IASS: 44574, migrate files according to db-entry

nhaagen:
https://mantis.ilias.de/view.php?id=44574

I checked in r8: there should be but one file in the users's directory.
On migration, use that file, even if the actual filename differs from the DB-entry.


klees:
No need to pick to release_10 and trunk, this is about a migration that should have run by then.

PR #9313, opened 2025-04-04
BackgroundTasks Bugfix #0044856: Character Limit Bucket Descriptions

fhelfer:
Fixes an Error when setting a description for Buckets longer than 255 chars.

Since there are no formal requirements specified for the description field, the consumer cannot know this constraint.
For example, changing the language variable for feedback via File in Exercises, creates a task with a description more than 255 chars, if language variable + exercise title are too long.
This approach ensures consistency for all components by explicitly truncating the input to 255 characters.

See https://mantis.ilias.de/view.php?id=44856


chfsx:
Thanks a lot, I picked to all relevant branches as well

PR #9314, opened 2025-04-04
[FIX] respect upload policies in MediaObjects (and maybe other locations)

chfsx:
@alex40724 From my point of view, you can drop the explicit setting of the maximum size here, because the upload_limit_resolver already calculates this correctly. Furthermore, this would enable the upload policies (custom limits), especially in the MediaObjects, where they probably make a lot of sense.


alex40724:
Thanks, I removed this line and added a comment to keep this info.

PR #9315, opened 2025-04-04
9/object/fix icons

kergomard:
Hi @chfsx 

As discussed this is the backport of the fix for object specific types back to ILIAS 9. Changes in Object are clearly fine with me. Feel free to merge, no picking needed.

See: https://mantis.ilias.de/view.php?id=42146

Best,
@kergomard


chfsx:
Thanks a lot @kergomard !

PR #9316, opened 2025-04-05
fixed 44739: When exercises are deleted, submissions from deleted use…

alex40724:
…rs are not deleted

The bug adresses the deletion of user related data in exercises when a user is being deleted.

There are two options:
a) adding more method calls to ilObjUser->delete (pretty sure we do not want that).
b) using events.

There is already a deleteUser event in place in the user service. However this event is called when most of the user data is already deleted. It is not possible to create instances of ilObjUser with the correspoding IDs at this time anymore.

This makes it impossible to re-use and domain level logic that relies on these instances. It makes it also impossible to create foreign key constraints to usr_data imo. (I did not try that)

Manytimes you want to delete data in dependent components, before the user "object" is finally deleted.

Thus this PR suggests to add a new event "beforeUserDeletion".

### DIFF
--- a/Modules/Exercise/Service/classes/class.InternalDomainService.php
+++ b/Modules/Exercise/Service/classes/class.InternalDomainService.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
@@ -17,6 +15,8 @@ declare(strict_types=1);
  * https://github.com/ILIAS-eLearning
  *
  *********************************************************************/
+
+declare(strict_types=1);
 
 namespace ILIAS\Exercise;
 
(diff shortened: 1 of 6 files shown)


alex40724:
@kergomard Please check this change in the user service:

https://github.com/ILIAS-eLearning/ILIAS/pull/9316/files#diff-e6c81eef5e6261f6eba82ec6de2d7aeecfae82d154fdef8c383b6aa36938a3e6

kergomard:
Hi Alex

I don't think though, we should introduce a new event. I don't think there is a justification for that and additionally the naming becomes very confusing. I would kindly ask you to simply move the current `deleteUser` event to where you now trigger `beforeUserDeletion`. I think this is the correct behavior anyway: We delete all dependent data of the user, before we actually delete the user itself.

Thanks again and best,
@kergomard

kergomard (changes requested):
Hi Alex

I don't think though, we should introduce a new event. I don't think there is a justification for that and additionally the naming becomes very confusing.
- [ ] Please simply move the current `deleteUser` event to where you now trigger `beforeUserDeletion`. I think this is the correct behavior anyway: We delete all dependent data of the user, before we actually delete the user itself.

Thanks again and best,
@kergomard

PR #9318, opened 2025-04-07
T&A Bugfix #0044803: Missing isset ilTestScoring.php

fhelfer:
https://mantis.ilias.de/view.php?id=44803


kergomard:
Thank you very much for the PR @fhelfer ! This is ILIAS 9 only, so no pick needed.

Best,
@kergomard

PR #9319, opened 2025-04-07
Auth Bugfix #0044801: Auth/User/SessionReminder: Building instanced with "logged in user" will not work without PHP session

fhelfer:
https://mantis.ilias.de/view.php?id=44801

### DIFF
--- a/components/ILIAS/Authentication/classes/class.ilSessionReminderCheck.php
+++ b/components/ILIAS/Authentication/classes/class.ilSessionReminderCheck.php
@@ -95,7 +95,7 @@ class ilSessionReminderCheck
             return $this->toJsonResponse($response);
         }
 
-        $expiration_time = $data['expires'] ?? null;
+        $expiration_time = $data['expires'];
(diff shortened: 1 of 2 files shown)


components/ILIAS/Authentication/classes/class.ilSessionReminderCheck.php
```diff
@@ -93,11 +95,12 @@ public function handle(): ResponseInterface
             return $this->toJsonResponse($response);
         }
 
-        $expiration_time = (int) $data['expires'];
+        $expiration_time = $data['expires'] ?? null;
```
mjansenDatabay: The key does always exists, so the `?? null` has no added value.

```suggestion
        $expiration_time = $data['expires'];
```

mjansenDatabay (changes requested):
Thx for the PR @fhelfer , I just added one inline comment.

mjansenDatabay (approved):

PR #9320, opened 2025-04-07
T&A 44781: Fallback to first page and reset offset and range if out of bounds after filter. (Backport)

matheuszych:
[Mantis: 44781](https://mantis.ilias.de/view.php?id=44781)

This error was already reported to ILIAS 10 and fixed by @thojou in https://github.com/ILIAS-eLearning/ILIAS/pull/8829.
I just ported the changes to ILIAS 9.


dsstrassner:
Habe mich nur wegen der Statistik zugeordnet.

klees:
Thx  @matheuszych!

PR #9321, opened 2025-04-08
[FIX] 0037444: Drop zones can be activated with subtabs, container ttles and others

chfsx:
This fixes Issue https://mantis.ilias.de/view.php?id=37444.

The dropzones react to anything that can be “dragged around”, for example, selected text, icons, etc. The PR adjusts the dropzones so that they only react to files that are dragged onto the browser.


thibsy:
@Amstutz and @klees LGTM, please continue :).

klees:
Picked to release_10 and trunk.

PR #9324, opened 2025-04-08
[FIX] Notes: replace legacy `setCloseSignal()` usages.

thibsy:
Hi @alex40724, could you have a look? Should be a trivial fix - some remnants of the old function. Thx!
Must be picked for `trunk` as well.


alex40724:
Thanks for the PR!

PR #9325, opened 2025-04-08
[FIX] UI: use `Item\Notification` JavaScript ID as string.

thibsy:


klees:
Picked to trunk.

PR #9326, opened 2025-04-08
[FIX] UI: call to `il.UI.page.isSmallScreen()` may be undefined.

thibsy:


klees:
Picked to trunk.

PR #9329, opened 2025-04-08
[FIX] UI: remove `bootstrap-datetimepicker.min.js` includes.

thibsy:
(Fixes https://mantis.ilias.de/view.php?id=44787)


klees:
Picked to trunk.

PR #9341, opened 2025-04-09
[FIX] Style: remove `GeSHi` usage and syntax-highlighting.

thibsy:
Hi @Amstutz,

When proposing the `geshi/geshi` composer package I noticed it was EOL. The code also may not be PHP 8.4 compatible - or at least their website is not working because of it anymore :). This PR simply removes this usage and therefore syntax-highlighting from the Kitchensink Docu.

Unfortunately, there is no alternative exactly like `geshi/geshi`. They all require to load additional assets (js/css) in order to work, thus adding unneeded/unwanted complexity. Let me know how you feel about having no syntax-highlighting in the browser anymore.

Kind regards,
@thibsy


Amstutz:
thx a lot!

PR #9342, opened 2025-04-09
StaticURL: Fix null-`base_path`

mjansenDatabay:


mjansenDatabay:
Was already fixed with ad1b41b in `release_10`, I picked it to `release_9`, FYI @chfsx

PR #9343, opened 2025-04-10
Export: Fixes in migration of export files to IRSS, ...

chlulei:
mark export file info table entry as migrated if the corresponding export files no longer exists. Also remove namespaces from the setup agent and the migration, otherwise executing the migration leads to an error.

LINK: https://mantis.ilias.de/view.php?id=44427


smeyer-ilias (approved):
Thanks @chlulei

PR #9347, opened 2025-04-10
Mantis #44839: Fix non-existent cmd used in form action

mBeym:
https://mantis.ilias.de/view.php?id=44839


kergomard:

Best,
@kergomard

PR #9349, opened 2025-04-10
T&A:  remove the feature "question hints"

nhaagen:
https://docu.ilias.de/go/wiki/wpage_8398_1357


kergomard (changes requested):

I looked through this and it looks good (even if these biggish removal PRs are hard to review, I'm sure I missed things).

I think there is one more step we should do:
- [x] Hints can be COPages. We should make sure we delete those, too.

Thank you very much and best,
@kergomard

kergomard:
Thank you very much for the PR and adapting it once more @nhaagen !

PR #9351, opened 2025-04-10
Course/Group/WebLink: abandon LOM in SOAP administration

schmitz-ilias:
This PR removes the dependency on the `ilMDSaxParser` from Course, Group and WebResource (see [Abandon Setting LOM of Courses/Groups/Weblinks via SOAP](https://docu.ilias.de/go/wiki/wpage_8349_1357)), along with some cleanup.


smeyer-ilias (approved):
Thanks @schmitz-ilias !

PR #9353, opened 2025-04-11
T&A 44808: Fixes type mismatch when creating a test object inside a question

matheuszych:
[Mantis: 44808](https://mantis.ilias.de/view.php?id=44808)

The issue seems to be that the wrong identifier is used. In the specific case mentioned in the Mantis issue, it was a `QuestionPool` that had the certain identifier but I believe it could have been any other object as well. By strictly using a `Test` identifier we should not have the problem.

### DIFF
--- a/Modules/TestQuestionPool/classes/class.assQuestion.php
+++ b/Modules/TestQuestionPool/classes/class.assQuestion.php
@@ -701,11 +701,8 @@ abstract class assQuestion
         }
 
         // update test pass results
-        $test = new ilObjTest(
-            $this->getObjId(),
-            false
-        );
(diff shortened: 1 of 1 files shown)


matheuszych:
Already reviewed by @thojou.

kergomard:

I'm pretty sure this is not the right solution. Please find out why there is a wrong obj_id set on the question and correct that.

Thank you very much and best,
@kergomard

matheuszych:
Hallo @kergomard
Thank you for your review! We found the issue in the data source. Therefore this PR is obsolete.

Best regards,
@matheuszych

PR #9354, opened 2025-04-11
#44852 Log missing language module without halting the import process

padvincenzo:
If I export the global language file from an ILIAS installation with some plugins, and I try to import it on another ILIAS that misses one of those plugins, I get the error message "Trying to access array offset on value of type null" with no logs anywhere else.
With this PR those modules (plugins) will be skipped from the import process, adding the information in the ILIAS log file.

### DIFF
--- a/Services/Language/classes/class.ilObjLanguageExt.php
+++ b/Services/Language/classes/class.ilObjLanguageExt.php
@@ -459,6 +459,10 @@ class ilObjLanguageExt extends ilObjLanguage
                 $ilDB->quote($module, "text")
             ));
             $row = $ilDB->fetchAssoc($set);
+            if (!$row) {
+                $DIC->logger()->lang()->warning("Language module '{$module}' not found for language {$a_lang_key}.");
+                continue;
+            }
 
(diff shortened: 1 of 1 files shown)


kevenClausenKPG (commented):
Could you please change the following:
`$DIC->logger()->lang()->warning("Language module '{$module}' not found for language {$a_lang_key}.");
`
in
`$DIC->logger()->root()->warning("Language module '{$module}' not found for language {$a_lang_key}.");
`
Once that's done, we can merge it. Thank you!

padvincenzo:
Hi @kevenClausenKPG , I've now changed `lng()` logger to `root()`

kevenClausenKPG (commented):
Let's run the CS Fixer with the following command to ensure the tests pass cleanly in GitHub:
`libs/composer/vendor/bin/php-cs-fixer fix --using-cache=no --diff --config=./CI/PHP-CS-Fixer/code-format.php_cs -vvv Services/Language/classes/class.ilObjLanguageExt.php
`

PR #9355, opened 2025-04-11
Glossary: add migration to delete LOM remnants of definitions

schmitz-ilias:
This PR adds a migration to delete leftover LOM sets of Glossary definitions from the database. All the heavy lifting is done in a reusable abstract migration supplied by MetaData, see [here](https://github.com/ILIAS-eLearning/ILIAS/blob/release_10/components/ILIAS/MetaData/classes/Setup/DeleteLOMForObjectTypeMigration.php).


alex40724:
Thanks for the PR!

PR #9086, opened 2025-02-20
Integrate all approved NPM dependencies for ILIAS 11

lscharmer:
Hi,
this PR merges the following npm dependency PRs:
- #8539
- #8559
- #8561
- #8564
- #8567
- #8569
- #8572
- #8646
- #8661
- #8664
- #8665
- #8667
- #8668
- #8749
- #9039
- #8615

### eslint + eslint-config-airbnb-base

The list doesn't include the accepted Dependency #8669 (eslint-config-airbnb-base).
We require `eslint` v9.16 which conflicts with `eslint-config-airbnb-base`'s requirement for `eslint` v8.2. But as this version is already EOL (see: https://eslint.org/version-support/) I opt for dropping `eslint-config-airbnb-base` instead of downgrading `eslint` until this is decided. @thibsy As you suggest both dependencies: Can you provide a way forward?

### Updated asset paths

I adjusted asset paths in the following `Component` classes, so that `composer install` doesn't fail:

- `ILIAS\Calendar`: Dropped package `moment` (@smeyer-ilias)
- `ILIAS\UI`: Renamed path in package `@yaireo/tagify` (@Amstutz / @klees)
- `ILIAS\UIComponent`: Dropped package `jstree`
- `ILIAS\jQuery`: Dropped package `jquery`

If these changes should be made in another PR (not by me) I can drop these instead.

PR #9361, opened 2025-04-11
[FIX] npm `rollup` cross-platform usage

thibsy:
Hi all,

I have stumbled upon a pickle, which may be relevant for the new dependency process in ILIAS. With `npm` packages are - to the best of my knowledge/understanding - installed according to the current Node.js [`process.arch`](https://nodejs.org/api/process.html#processarch) and [`process.platform`](https://nodejs.org/api/process.html#processplatform) values. From what I have gathered, this has effect on what transitive optional / peer dependencies are included when installing a package.

@lscharmer (so kindly) installed all ILIAS 11 `npm` packages on linux, apparently (#9086). This led to issues on macos and possibly other platforms when using `npm clean-install`, which installs the exact dependency tree from the `package-lock.json` file, which now only contains the platform-specific transitive dependencies of `rollup` for linux. To address this issue, I have updated this package and "let" `npm` install _all_ optional dependencies of `rollup`, so the tool becomes usable on _any_ platform - not just macos. (Extra-Pickle: I tried updating the package using `npm update rollup --omit=optional` which installed all of the optional dependencies anyways. No idea why.)

This may be a relevant take away for the dependency process, because we need to treat `npm` packages more carefully. We may also need some information in the pull-request template that suggests how the dependency has to be installed - with optional / peer dependencies.

Kind regards,
@thibsy

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -314,30 +314,284 @@
         }
       }
     },
-    "node_modules/@rollup/rollup-linux-x64-gnu": {
-      "version": "4.38.0",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-linux-x64-gnu/-/rollup-linux-x64-gnu-4.38.0.tgz",
-      "integrity": "sha512-vPvNgFlZRAgO7rwncMeE0+8c4Hmc+qixnp00/Uv3ht2x7KYrJ6ERVd3/R0nUtlE6/hu7/HiiNHJ/rP6knRFt1w==",
+    "node_modules/@rollup/rollup-android-arm-eabi": {
+      "version": "4.39.0",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.39.0.tgz",
+      "integrity": "sha512-lGVys55Qb00Wvh8DMAocp5kIcaNzEFTmGhfFd88LfaogYTRKrdxgtlO5H6S49v2Nd8R2C6wLOal0qv6/kCkOwA==",
+      "cpu": [
+        "arm"
+      ],
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "os": [
+        "android"
+      ]
+    },
+    "node_modules/@rollup/rollup-android-arm64": {
+      "version": "4.39.0",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm64/-/rollup-android-arm64-4.39.0.tgz",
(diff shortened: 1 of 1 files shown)


lscharmer:
Hi @thibsy,
thank you for figuring this out. I did not know about this :o 
I also agree that this is useful to be included in the dependency process.
We also already have the existing `"extra"` sections for each NPM dependency in the `package.json` so it is not just documented in a PR (e.g. useful when updating a dependency version mid release).

The changes look good to me, also tested locally.

lscharmer (approved):

thibsy:
@lscharmer +1 for your idea on the `extra` section in the `package.json`.

thibsy:
Since this has become increasingly inconvenient for ILIAS development on macOS, and as one could argue that @lscharmer approved this in their role as shepherd, I will go ahead and merge this PR now. I will present the underlying issue at the upcoming JF for further discussion, since this is relevant for the dependency process.

matthiaskunkel:
Jour Fixe, 23 JUN 2025: Thibeau notified us about this PR that is already merged. This PR helps to improve the dependency handling on macOS.

thibsy:
Hi all,

I did some more research:

- we cannot tell `npm` to install only dev dependencies, so we cannot use separate "installation techniques" for these two kinds of dependencies.
- we cannot simply tell developers to run `npm install` instead of `npm clean-install` because **both** installation techniques are looking at the `package-lock.json` file - for different purposes though. Platform-specific optional dependencies are, however, for both techniques read from the lock file.
- we _could_ tell developers to run `npm install --package-lock=false`, which would then ignore the `package-lock.json` file and install platform-specific optional dependencies as well. BUT, this will result in non-deterministic platforms.
- there is a [`--os`](https://docs.npmjs.com/cli/v9/commands/npm-install#os) and [`--cpu`](https://docs.npmjs.com/cli/v9/commands/npm-install#cpu) flag can be used to override the [`process.arch`](https://nodejs.org/api/process.html#processarch) and [`process.platform`](https://nodejs.org/api/process.html#processplatform) values (apparently).
- there is also a [`os`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#os) and [`cpu`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#cpu) configuration for the `package.json` file.
- there is also a [`engine`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines) and [`devEngines`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines) configuration for the `package.json` file, which lets you define separate environments for prod and dev - this may be what we are looking for.

I may create a follow-up for this.

Kind regards,
@thibsy

PR #9362, opened 2025-04-11
SCORM: Fix date/time handling in raw data import

mjansenDatabay:
See: 

- https://mantis.ilias.de/view.php?id=44862
- https://mantis.ilias.de/view.php?id=44858


qualitus-dahme (approved):

qualitus-dahme:
@mjansenDatabay do we need this fix picked to 10 and trunk?

mjansenDatabay:
@qualitus-dahme The chances are high, that the issue also occurs in `release_10` and `trunk`, so you'll have to pick and test the changes for both versions as well.

PR #9363, opened 2025-04-11
OrgU: 44175, hide inputs at first

nhaagen:
Not proud of this, but I think it's the least invasive.
https://mantis.ilias.de/view.php?id=44175


klees (changes requested):
Hi @nhaagen!

Thanks a lot. I think we should add this directly after that other on-load-code, in the JS. This should strictly be executed in that order, right? `genericMultiLineInit`, them remove `style`-attribute, right? This is a guarantee that we don't really get from `addOnLoadCode` afaik.

Kind regards!

nhaagen:
Like this?

klees (approved):

klees:
@Amstutz @thibsy Please be so kind and move this forward!

thibsy:
@klees I think this is @Amstutz only (DC not UI).

klees:
No, sorry, it's just for me...

PR #9364, opened 2025-04-14
UI/Modal/Lightbox: Use proper event listener to ensure DOM is ready

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=44748


thibsy (approved):
LGTM, thx @mjansenDatabay

thibsy:
@mjansenDatabay this error does not occur in ILIAS 10 and 11, correct?

mjansenDatabay:
> @mjansenDatabay this error does not occur in ILIAS 10 and 11, correct?

Hi @thibsy , yes!

PR #9369, opened 2025-04-14
UICore: Fix implicit nullable parameter in unit test

mjansenDatabay:
This PR fixes an error with PHP 8.4. Implicit NULL-able parameters are not allowed anymore.


thibsy (approved):
Thx!

PR #9371, opened 2025-04-14
LearningModule: fix public access download link for html exports (43475)

schmitz-ilias:
Just a small fix for [43475](https://mantis.ilias.de/view.php?id=43475).


alex40724:
Thanks!

PR #9379, opened 2025-04-14
add missing call to upload handler (44850)

schmitz-ilias:
This PR fixes [44850](https://mantis.ilias.de/view.php?id=44850). Not sure why File is involved at all, something with dropzones I guess.


chfsx:
Thanks a lot, I picked to all relevant branches as well.

PR #9382, opened 2025-04-14
PRG: 44845, acknowledge courses from sub-programmes

nhaagen:
https://mantis.ilias.de/view.php?id=44845


klees:
Picked to trunk.

PR #9389, opened 2025-04-14
9/UI/Optional-Switchable-Group fix expanded glitch 36406

catenglaender:
https://mantis.ilias.de/view.php?id=36406

# Issue

When rendering a page with a Switchable or Optional Group UI components, the collapsed elements that should be hidden are visible for a split second before jQuery has a chance to hide them.

https://github.com/user-attachments/assets/f4d0e41a-86ba-4a1c-a629-ebe975a67655

# Change

This is just a band-aid for release_9. The new form structure in release_10 fixes this. Do not cherry pick this to any other ILIAS version.

Since we already re-did everything for release_10 with a CSS only solution I didn't touch the JavaScript in 9. (Wow, have we come a long way... the forms in 10 are so amazing to target with CSS.)

I made forms nested under unchecked checkboxes or radios invisible by default - it's basically doing what the JavaScript does, but using CSS which loads before the DOM is rendered.

It's also based on a class given by the Switchable and Optional Group Renderer. I think this makes it safe enough to roll out as a bugfix. What do you guys think?


klees (approved):

kergomard (approved):

Changes in Test are ok. I remove @thojou and @dsstrassner as reviewers.

Amstutz:
thx a lot!

PR #9391, opened 2025-04-15
Math: Fix `phpunit` issues

mjansenDatabay:
This PR suggests to remove deprecated PHPDoc
annotations for `phpunit`. Instead, native
PHP attributes are used.


mjansenDatabay:
Can be closed, the test cases have been already fixed.

PR #9394, opened 2025-04-15
Mantis 44891: Bugs and usability issues when copying role permissions

ingoj:
possible fixes for mantis 44891

buttons which start copying of permissions are relabeled
on-screen-message additionally for selecting target roles displays source role
copy behaviour form has a new on-screen-message that shows copy source and up to three targets (if more targets are selected, the number of selected targets is displayed)

this fix introduces 4 new language variables, which are provided in german and english, all in rbac:
   rbac_copy_role_behaviour_info
   rbac_copy_multi_targets
   rbac_copy_no_targets
   rbac_select_roles


ingoj:
in class.ilObjRoleFolderGUI copyright notice is adjusted, too

ingoj:
Probably this PR needs to be discussed in JF

kergomard (changes requested):
Thank you very much @ingoj for the PR!

There are a few little things concerning code-styles:
- [ ] Could you please fix the code-style violations (use `php ./libs/composer/vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix --config=./CI/PHP-CS-Fixer/code-format.php_cs --using-cache=no -vvv <path to file>` from the root of your ILIAS installation to do so).
- [ ] Please rename the variables you introduced from camelCase to snake_case (e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9394/files#diff-e757573375fadb0c900fb0c008759c7cbb108fe9ac2ca81353590e059739d54dR299)).
- [ ] Please remove the variable `$sourceName` and move the corresponding retrieval-call to its single use.
- [ ] I don't like the `switch`-statement. I think it makes things very hard to read. Please create a function `buildTargetNamesString(array $targets): string` with an early-return structure with `if`s.
- [ ] Please copy the content of the language variable 'user_role_selection' in all languages to the variable 'rbac_select_roles'. That avoids that everybody needs to retranslate this value.

I hope I didn't miss anything and sorry if I'm not precise enough here.

I think we can do this without going to the JF.

Thanks again and best,
@kergomard

ingoj:
I updated my repo according to your requested changes. Just as a remark: Only in 2 languages (beside english and german) translations were preset. Nevertheless I copied the variable for all languages, I hope this was right.

kergomard:
Thank you very much @ingoj for the PR!

To avoid more forth and back I took your changes, added a few of my own and committed them in https://github.com/ILIAS-eLearning/ILIAS/commit/e87cc1e01bde39458983d9dce511af5f8012f177 . I picked the changes to ILIAS 10 and trunk.

Thanks again and best,
@kergomard

PR #9395, opened 2025-04-15
Release 9 lng studyprogramme

ingoj:
Possible fix for mantis #44892 for german and english language

### DIFF
--- a/lang/ilias_de.lang
+++ b/lang/ilias_de.lang
@@ -5011,8 +5011,8 @@ common#:#obj_poll_duplicate#:#Abstimmung kopieren
 common#:#obj_prg#:#Studienprogramm
 common#:#obj_prg_duplicate#:#Studienprogramm kopieren
 common#:#obj_prg_select#:#-- Bitte wählen Sie ein Studienprogramm --
-common#:#obj_prgr#:#Link zum Studienprogramm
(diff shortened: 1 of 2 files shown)


matthiaskunkel:
Thanks @ingoj for the PR. But to be honest, I do not like these long German terms. Kurslink and Gruppenlink are okay. But Studienprogrammlink is a bit like Donaudampfschifffahrtskapitänsmütze ... I understand the wish to streamline terms and to use patterns. But if we want a unified pattern here, I would go for "Link zum Kurs" and "Link zur Gruppe".

ingoj:
Hi @matthiaskunkel,
I fully understand your dislike of these long words. Of course we could switch to a "Link to.." pattern (in english, too). But this would result in all Container-Links in the permission-list of roles to move from their associated object types (except for the link to studyprogramme). We would have all 4 link types listed together somewhere near the learning sequence. 
To shorten the words wo also could change the wordings to "Kurs Link", "Studienprogramm Link",  a.s.o.. Which variant (including to leave everything as it is) would you prefer? I can adjust this PR accordingly, if wanted. 
Do we want/need a JF discussion for this?

matthiaskunkel:
I prefer to keep the wording as it is – even if this is inconsistent.

PR #9399, opened 2025-04-15
Export: The public access column can be deactivated when the export gui is used

chlulei:


smeyer-ilias (approved):
Thank you very much @chlulei

PR #9400, opened 2025-04-15
10/UI/footer/increase spacing to page 42597

catenglaender:
https://mantis.ilias.de/view.php?id=42597

# Issue

Footer is perceived to be too close to content.

image

# Change

Increased page padding

image

# Outlook

This needs to be done much better next page layout revision. There are many views that add their own padding already.


Amstutz:
thx a lot!

PR #9401, opened 2025-04-15
CoPage Bugfix #0044902: Fix undefined call to child_nodes() CoPage

fhelfer:
https://mantis.ilias.de/view.php?id=44902


alex40724:
Thanks for the fix!

PR #9404, opened 2025-04-15
Search: abandon advanced search

schmitz-ilias:
This PR implements [Abandon Advanced Search](https://docu.ilias.de/go/wiki/wpage_8577_1357), along with some clean up in Search.

I had to keep the classes around that do the actual searching (`ilAdvancedSearch` and `ilLikeAdvancedSearch`), because they are still used in Container filters, see `FilterManager`. We should have a look at this again at some point along with `ilMetaDataSearch`, since those classes still contain direct accesses to MetaData db tables, but that is not a super high priority.


smeyer-ilias (approved):
Thanks @schmitz-ilias

PR #9405, opened 2025-04-15
[Mantis 44906] ilCalendarCategories should only cache instances that it…

Hufschmidt:
See https://mantis.ilias.de/view.php?id=44906

**TL;DR:** `ilCalendarCategories::initialize` replaces cache on each invocation (for `$mode self::MODE_REPOSITORY`, `self::MODE_CONSULTATION` and `self::MODE_PORTFOLIO_CONSULTATION`) because the cache is written without checking whether it can be loaded, since `$mode` is part of the cache lookup `entry_id`.

See mantis issue for details.

### DIFF
--- a/Services/Calendar/classes/class.ilCalendarCategories.php
+++ b/Services/Calendar/classes/class.ilCalendarCategories.php
@@ -326,14 +326,18 @@ class ilCalendarCategories
         }
 
         if ($a_use_cache) {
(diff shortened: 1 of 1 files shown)


smeyer-ilias (approved):
Thanks @Hufschmidt for analyzing and sharing your results!

PR #9413, opened 2025-04-16
Exercise: Make migrations more robust if no step is left

mjansenDatabay:
This PRs make the `step` method for exercise-related setup migrations more robust. We received errors by community members (which will hopefully create a corresponding issue) where the migrations try to access properties of `NULL`. The database query simply does not return any object/stdClass in these cases. The amount of steps is zero when executing the migrations again.


alex40724:
Thanks for the PR!

PR #9416, opened 2025-04-16
10/UI/popover-scroll-and-mobile-fix 43704

catenglaender:
https://mantis.ilias.de/view.php?id=43704

Note that webuipopover as a dependency will be removed in 11, as I was informed by my colleague @nhaagen . The new and improved UI Popover will probably use a system closer to the Button tooltip which doesn't rely on fixed positioning and simply cannot have the same problems to begin with. Therefor this fix is more of a band-aid than a future proof solution.

# Issue

Popover UI component fails test because
* popover tooltip doesn't follow scroll
* popvers aren't visible on mobile screen sizes

# Changes

Oh boy was this fiddly.
I couldn't pin point exactly why, but something inside the dependency does not handle scrolling in flexbox or css grids elements well and fails to update the popover's position. This is what I ended up doing:
* passing either "body" on mobile screen sizes OR ".il-layout-page-content" on desktop screen sizes as the scroll container
* attaching a scroll event listener to the currently relevant container
* a debounce reduces the number of scroll events fired
* scroll event reaches into webuipopover's internal methods by using $triggerer.data('plugin_webuiPopover') to detect open popovers and refresh their position

I am aware that the following edge cases are not handled:
* changing the window size between mobile and desktop does not refresh the event listeners. Popovers therefor no longer scrolled until closed with a click and reopened.
* opening and closing the slate does not update the popovers position until scrolled.

Furthermore ESLint has much to complain about the rest of the code in this file.

Since I didn't want to add too many event listeners and those glitches disappear when re-opening the the popover, I decided against fixing those. The Popover in 11 will handle these edge cases.

# Attention

To make this work I had to change the mobile scroll container from html to body. Otherwise the popover would be appended to the html element outside of the body which is not allowed by web standards.

Scrolling body elements are not completely unheard of, but it's unusual. It's problematic as this change effects all pages in ILIAS, which is dangerous for a release that has already been tested so thoroughly.

Maybe someone here has a better idea?

### DIFF
--- a/components/ILIAS/UI/resources/js/Popover/popover.js
+++ b/components/ILIAS/UI/resources/js/Popover/popover.js
@@ -1,109 +1,177 @@
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ */
+
 var il = il || {};
 il.UI = il.UI || {};
 
-(function($, UI) {
-
-    UI.popover = (function ($) {
-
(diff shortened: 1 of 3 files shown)


Amstutz:
thx a lot for this partial fix! We believe it to be good enough for the use case given.

PR #9417, opened 2025-04-16
UI/URLBuilder: Use correct type declaration for `acquireParameter`

mjansenDatabay:
The current return type declaration in the PHPDoc comment results in false positives in SCA (e.g. in PHPStorm).

Bildschirmfoto vom 2025-04-16 13-43-42

With the change this PR suggests, calls to `withParameter` are **not** reported as "method not found" anymore.


thibsy (approved):
LGTM, thx!

PR #9424, opened 2025-04-17
Markdown: adjust markdown's headline configuration.

lukastocker:
This PR removes '-' and '=' in the headline configuration. As a result only '#' can be used as a markdown headline.
This avoids rendering of false headlines, e.g.:

the user uses '-' to separate a headline or an email => 'Topic 1 - How to create a course'
the user uses '-' to separate parts of an email like this '-----------'
This is also needed for the suggested Markdown Underline PR as it uses '--'.

### DIFF
--- a/components/ILIAS/CommonMarkExtension/CommonMarkCoreExtension.php
+++ b/components/ILIAS/CommonMarkExtension/CommonMarkCoreExtension.php
@@ -1,0 +1,103 @@
+<?php
+
+/**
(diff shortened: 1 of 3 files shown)


dsstrassner:
Hi @thibsy I hope this is the correct assignment, at least at the present time.

thibsy:
Hi @lukastocker, I close this to get it off my list. I don't think we should pursue adjustments of our markdown flavour like that.

PR #9423, opened 2025-04-17
Mail: include markdown in mails

lukastocker:
- add markdown rendering. Also always escape HTML in markdown output
- use UI form in ilMailFormGUI and add Markdown for Message
- enable markdown for free text mails
- include new functions in form and buttons to avoid a redirect of the page if the user clicked onto the placeholders to insert them into the mail
- adjust template rendering in 'withAdditionalOnLoadCode' if coming from a context
- adjust PHPUnit Tests
- use ResourceCollection for uploading/showing file information

PR #9425, opened 2025-04-17
Markdown: add underline extension

lukastocker:
This PR is a suggestion and should not get reviewed and merged before this https://github.com/ILIAS-eLearning/ILIAS/pull/9423.

This PR includes the possibility to use 'underline' as markdown as it's not included in the core.
It's also linked to PR https://github.com/ILIAS-eLearning/ILIAS/pull/9424 which is needed for this PR. Else there might be some errors during the rendering process of the markdown.


dsstrassner:
Hi @thibsy I hope this is the correct assignment, at least at the present time.

thibsy:
Hi @lukastocker, I close this to get it off my list. I don't think we should pursue adjustments of our markdown flavour like that.

PR #9435, opened 2025-04-22
T&A Bugfix #0044944: Manual Scoring in Random Test doesnt update Reached Points

fhelfer:
https://mantis.ilias.de/view.php?id=44944

### DIFF
--- a/Modules/Test/classes/class.ilTestScoringGUI.php
+++ b/Modules/Test/classes/class.ilTestScoringGUI.php
@@ -269,7 +269,8 @@ class ilTestScoringGUI extends ilTestServiceGUI
                     $maxPointsByQuestionId[$questionId],
                     $pass,
                     true,
-                    $this->object->areObligationsEnabled()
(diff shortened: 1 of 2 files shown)


kergomard:
Thank you very much for the PR @fhelfer .

I don't think this is the right solution though: Please always hand in the `test_id` and do not allow empty values for it. There are exactly two usages of this function (and a third one in `assFileUpload` that is never accessed and was thus removed in ILIAS 10 and can also go in ILIAS 9 as far as I can see).

Thanks again and best,
@kergomard

kergomard:
Hi @fhelfer 

Sorry, I apparently didn't read this right. I still don't like the solution, so I went with a different approach for ILIAS 10 and above (see: https://github.com/ILIAS-eLearning/ILIAS/commit/4fea97c7fab530c7e0a0bfca1c835d146ac88d04), but for ILIAS 9, I think this is good enough.

Best,
@kergomard

PR #9436, opened 2025-04-22
UICore: Fix str_starts_with Does Not Accept Null

kergomard:
Hi @thibsy 

There is an error in `ilTemplate` where `$a_in_module` might get set to `null` which then is not accepted by `str_starts_with`. This would be a possible solution, I think.

See: https://mantis.ilias.de/view.php?id=44919

...the failing tests come from the superfluous `Modules`-folder introduced in a previous commit.

Thanks and best,
@kergomard

### DIFF
--- a/components/ILIAS/UICore/classes/class.ilTemplate.php
+++ b/components/ILIAS/UICore/classes/class.ilTemplate.php
@@ -335,7 +335,7 @@ class ilTemplate extends HTML_Template_ITX
     /**
      * @throws ilSystemStyleException
      */
-    protected function getTemplatePath(string $a_tplname, string $a_in_module = null): string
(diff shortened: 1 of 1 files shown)


thibsy (changes requested):
Hi @kergomard,

Thx a lot for the fix. I think this should be fine, however, you will have to use a null-coalesce for the usage on `ilTemplate:228`, which implicitly accepts null and would cause a probable type error.

Also, could you check if the failing unit tests have something to do with this change?

I will merge directly afterwards.

Kind regards,
@thibsy

kergomard:
Thanks @thibsy for the pointer!
I changed the PR correspondingly. The failing tests where due to other reasons. A rebase solved the issue.

Best,
@kergomard

thibsy (approved):
LGTM now, thx for following up!

PR #9442, opened 2025-04-22
UI: 44638, adjust color styling of dropdown breadcrumbs for small scr…

lukastocker:
…eens.

https://mantis.ilias.de/view.php?id=44638


lukastocker:
Note: 
The faulty PHPUnit test run has nothing to do with this implementation.

Amstutz:
thx a lot!

PR #9447, opened 2025-04-23
UI: adjust dropdown menu width and multi select checkboxes in dropdow…

lukastocker:
…n menus.

https://mantis.ilias.de/view.php?id=44689

@catenglaender and myself decided to set the width of the dropdown menu to 320px to avoid breaking content too early but  limiting the length per line as sometimes content can be quite large.
Additionally the flex-wrap in multi select label was removed to avoid splitting the checkbox from the label.

### DIFF
--- a/templates/default/070-components/UI-framework/Dropdown/_ui-component_dropdown.scss
+++ b/templates/default/070-components/UI-framework/Dropdown/_ui-component_dropdown.scss
@@ -61,6 +61,7 @@ $cursor-disabled: not-allowed !default;
   
   // The dropdown menu (ul)
   .dropdown-menu {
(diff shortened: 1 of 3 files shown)


dsstrassner:
Sorry, I added me as assignee (for statistical purposes), but forgot to also add you @Amstutz

Amstutz:
@catenglaender I would suggest do wrap the width in a local scss variable, what do you think? I leave the final decision here to you. Assign this back to me if ready to be merged, or merge it (and pick to trunk). Thx a lot!

Amstutz:
I merge this since it is a good improvement and open for some time.

PR #9449, opened 2025-04-23
T&A: ilTestAccess Refactoring

lukas-heinrich:
Hi everyone,

With this PR, I propose a refactoring of the ilTestAccess class, which takes place in the context of [Using test results as prerequisites](https://docu.ilias.de/go/wiki/wpage_6929_1357). The refactoring focuses on extracting functions for determining participant status, as well as standardizing the storage and retrieval of test results. Through the use of caching, the querying of participant status has also been optimized. Additionally, as part of these changes, DTO classes are being introduced to describe test results in a type-safe manner. The changes in this PR have been further verified through the introduction of new unit tests.

Furthermore, there is a change in the last commit that I additionally propose: we noticed that the table `tst_result_cache` is already being updated after each saved question during a test attempt. I consider this to be an error, as it can lead to inconsistencies in the table. For example, the values for the *mark schema* provide contradictory values to what is stored in the *passed* flag. During the review, please take a closer look at this last commit to see if you agree with me on this point or if this might have even been intended behavior.

As always, these changes have already been approved in an internal review by @thojou. I look forward to your feedback on these extensive changes. Particularly regarding the naming of the class TestResultManager, we are open to suggestions that more precisely describe the functions of this class.

Thank you in advance.
Best,
@lukas-heinrich


kergomard (changes requested):
Thank you very much for all the work @lukas-heinrich !

...and thank you already for bearing with me as we are moving forward, but I don't think we are there yet.

There will be quite a long list here, sorry and we will need multiple iterations until we get over the finishing line, but lets get started:

- [ ] Please make sure that no naming contains the word 'pass'. It should be 'attempt'.
- [ ] Please rename `TestResultManager` to `Repository` and the corresponding value on the `$DIC` to `results.data.repository`.
- [ ] Why  do we now also have a `TestPassResult` when we already have an `AttemptResult`? Please consolidate.
- [ ] We should streamline naming here. So I think we should have `AttemptResult` and `ParticipantResult`.
- [ ] Please do not set the `ilObjTestAccess::$test_result_manager` or then `results_repository` as `static`. You are using it in `static` methods, I know, but these static methods rely on in it having been initialized at least once. I don't think we should do that. Please remove as many of the `static` methods as possible (also if this means initializing `ilObjTestAccess` in a few additional places. This should actually be rather cheap.
- [ ] Please do not set any functions in `ilObjTestAccess` or `ilObjTest` to 'deprecated', but remove them directly. If there are any usages somewhere outside of the test, please move them to the new pattern and we then need to tag the corresponding maintainers (we will simply announce this at the JF with a merge deadline).
- [ ] Please do not add unnecessary comments like e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-1cb1cfbe6a65a6c0f2e353698a244437f9b9a8a8d82027d069802046e3c9693cR29) and do not use constructor property promotion if you then modify a parameter in the constructor.
- [ ] In `TestResult` I don't like the two ways marks are treated. Please always hand in an actual `Mark` and also store it as an actual `Mark`.
- [ ] I do not like that we are adding one more complication to the whole thing by also using `ILIAS\Cache`. Is there an absolute need for it? Are we storing something across requests? If there isn't: Please remove its use and find another solution. I see a gazillion issues coming our way. If we would really, really need it, we would need to find a better way of integrating it. As far as I can see you can cache things in a instance-property of the repository.
- [ ] [This](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-4e3299e0891c0ed089ed6a8490747c046a7332ab1d5e30ad58a64e38c5401904R139) shouldn't be a call to a `static`, right?
- [ ] Why are we checking if a result exists [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-4e3299e0891c0ed089ed6a8490747c046a7332ab1d5e30ad58a64e38c5401904R149-R152)? We are doing a replace afterwards anyways.
- [ ] I don't think this [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-4e3299e0891c0ed089ed6a8490747c046a7332ab1d5e30ad58a64e38c5401904R86-R98) is right: We are needing two calls to the database for something that could be done in one and additionally `buildTestResultObject()` doesn't do what it says it does, as it also fetches data from the database.

Ok, as said, I'm sure this is only the first iteration, but I think this is enough for a first round.

Thanks again and best,
@kergomard

lukas-heinrich:
Hi @kergomard ,

Thank you very much for the review and detailed feedback. I have incorporated the suggested changes and adjusted my implementation in many places. I have also improved the naming to follow conventions and avoid confusion.

Regarding the use of `ILIAS\Cache`, I consider my implementation to be useful. By using the APCU cache, we can share the cached results between multiple server requests, which will have a very positive impact on performance. The test results are needed at least once in many places (e.g., in `ilObjTestAccess`). By storing them in the APCU cache, we can save ourselves these database queries. I think the increased complexity is a necessary trade-off for the improved performance.

As always, I look forward to your feedback and further comments.

Best
@lukas-heinrich

kergomard (changes requested):

This looks far clearer to me now and I like it!

My only bigger ask would be:
- [ ] I'm starting to be ok with the use of `ILIAS\Cache`. There is just one thing that has to change, for me to be happy with it: Any usage of `ILIAS\Cache` has to be completely encapsulated by the Repository, so the function `ILIAS\Test\Results\Data\Repository::invalidateStatusCache()` has to go from public-view. I don't think this is a big problem either, as it is used in a function that, from where I stand, should be part of the Repository anyways. There is an additional reason for this: We do already have a `cache` and it is a database-table. You see the problem all through the repository, where it is never completely clear what we are referring to. I would thus also suggest to refer to the table as `result_cache_table` e.g. [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-cb642d90ca5896833f8e074f8e17865aaf603c2a0032e550f79a955892f4f6d5R147) where it would be `bool $update_result_cache_table`.

The rest are rather small things as far as I can see:
- [ ] Why did you split [this here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-289c62ed5f4528b15627b43fd04297b268ea7cedeb89af254899261b7faa4f37R267-R268) up again? We don't need the `$pass_finisher`-variable anywhere, so we do not need it. Soon we could even get rid of the brackets around the `new ilTestPassFinishTask()`, so please role this change back.
- [ ] Please do not call [this variable](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-377cf2c64a9111fe133f461ef77079e17cd9132640dc33f5c5e0ef927c46de64R1797) `$result_cache`. The "cache"-part is an internal implementation detail, from the point of view if `ilObjTest` it holds the results and shouldn't care if it comes from a cache or not.
- [ ] Please  never use the `TestDIC` as a service locator like [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-e914f2583e4f3f424a922f15832292886da9b1c7764c70a3977fd356ec19300bR620). Please move the dependence to the constructor and hand it in as a `readonly` parameter.
- [ ] Please remove the CamelCase of the variable [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-ff08467a3b32f4c2134caeabfcd346f5cd728801005ee87770f67fbf160348beR105).
- [ ] Please pass the parameter [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-c616a5a6d397a783ddba52bc6e531fdb147e9e30d5f6dde3b4dfd3615da7eb0bR647) without creating a variable.
- [ ] Please remove the comment [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-597490a9cf34dedb4656d7b2bc9f80bc2e503e4efec63613efc0003f44364a0dR23-R25). Stating the provenience of the data seems very error prone to me and it is very easy to see this by taking a quick look at the corresponding repo.
- [ ] Could we please keep the properties private [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-597490a9cf34dedb4656d7b2bc9f80bc2e503e4efec63613efc0003f44364a0dR29-R38), [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-ad4c39b83c4dc721c851f6c67447032a928a6b038d989957a2e371651e3cdbb4R31-R37) and [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-cb642d90ca5896833f8e074f8e17865aaf603c2a0032e550f79a955892f4f6d5R34-R36). We do not want anybody to have the grandiose idea to extend from this class.
- [ ] Please do not call this `$object` [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-cb642d90ca5896833f8e074f8e17865aaf603c2a0032e550f79a955892f4f6d5R154) this does not tell us anything. Call it `$result_object`. Same [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-cb642d90ca5896833f8e074f8e17865aaf603c2a0032e550f79a955892f4f6d5R218), even if in this case I would call the parameter `$test_attempt_result_array` and the `$object` `$test_attempt_result`.
- [ ] A little bit more clean-up: Could you please move parameters of functions with more than two parameters to their own line? It increases readability immensely.
- [ ] Please use single-quotes for strings that do not contain any variables or formatting (this is mostly in the repo). Concatenate if multiline is needed for readability.
- [ ] Please do not separate [this](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-cb642d90ca5896833f8e074f8e17865aaf603c2a0032e550f79a955892f4f6d5R358-R359) into two variables and use curly-braces around variables in strings.
- [ ] Could you please readd the removed tab from [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-751de2293484db1002175e5f247dbeb6b3a26863ffdfab05dcf7daeb68316dbbR387) as it makes things more readable. ...as far as I can see this should be allowed by our style.
- [ ] [This](https://github.com/ILIAS-eLearning/ILIAS/pull/9449/files#diff-10c30c6f6958a0e7e25cb8882aa0f2793a1294acc83c0e5cf99c8977fd03e190R418-R430) looks wrong. I was wondering were it was used and couldn't figure that out.

I hope I didn't miss too many things and thus do not need to come back to you for too many more changes. Has this somehow been tested by somebody?

@kergomard

lukas-heinrich:
Hi @kergomard,

Thank you very much for your feedback! I have incorporated the suggested changes and improved the readability of the code in some places.
During development, the changes were tested and adjusted repeatedly. However, we will still carry out some detailed internal tests to ensure that nothing breaks. This is expected to take place next week.

Best,
@lukas-heinrich

lukas-heinrich:
Hi @kergomard,

A quick update: We have tested the changes from this PR internally. Several tests were created with questions, answered and the results checked, as well as corrections and manual scoring were performed. No problems were found during the tests.

Best,
@lukas-heinrich

kergomard:

Could you please take care of the failing tests and the conflicts and I will then merge asap.

Thanks for all the work and best,
@kergomard

lukas-heinrich:
Hi @kergomard,

I have now resolved the problem with the failing tests, and the PR should now be ready to merge. Thank you again for the detailed review and feedback.

Best
@lukas-heinrich

kergomard:

There is one more thing you will need to do, sorry:
- I would like to keep this in separate Commits and not have to squash it as the commits provide quite a bit of modularity/information, but I cannot merge and rebase until you remove the merge commit 7b8ab8a359b3b686b8683788ec3e31d5d591d707 from the middle.

Best,
@kergomard

lukas-heinrich:
Hi @kergomard,

I have rebased the branch and thereby removed the merge commit.

Best,
@lukas-heinrich

kergomard:

PR #9450, opened 2025-04-23
HTMLLearningModule: fix export/import of files (44138, 44462)

schmitz-ilias:
Hi @mbecker-databay, this PR fixes some issues with export/import of HTML Learning Modules, see [44138](https://mantis.ilias.de/view.php?id=44138) and [44462](https://mantis.ilias.de/view.php?id=44462). The files of HTML LMs were migrated to the IRSS, but some export/import stuff hadn't been adapted accordingly so far.

A few caveats: I tried to keep this as slim as possible, so I reused a method from the IRSS wrapper of the Repository component. HTMLLearningModule already uses some Repository stuff for dependency management anyways. Also, the `ilHTLMExportOptionHTML` uses some legacy file stuff, since as far as I can tell the HTML export files are not yet migrated to the IRSS.

@alex40724, I've also assigned you here because of a small fix in your IRSS wrapper. Without the flag I added, `RecursiveDirectoryIterator` iterates over `SplFileInfo`, which doesn't have `isDot`. The flag makes it so that it behaves more like `DirectoryIterator`. As far as I can see, this only matters for this new usage (all the old ones set recursive to false).

### DIFF
--- a/components/ILIAS/HTMLLearningModule/Service/class.InternalRepoService.php
+++ b/components/ILIAS/HTMLLearningModule/Service/class.InternalRepoService.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /**
  * This file is part of ILIAS, a powerful learning management system
  * published by ILIAS open source e-Learning e.V.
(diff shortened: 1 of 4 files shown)


mbecker-databay (approved):
Looks good to me, though I set @chfsx to look over it as well, as I am not perfectly sure about the implications of your remarks. Thank you so much!

chfsx (approved):
LGTM, thanks a lot!

alex40724 (approved):

PR #9453, opened 2025-04-24
#44703 / Paths in rCopy do not inherit directory permissions

mbecker-databay:
Following a report from a client about copied resources being inaccessible, here's a fix for https://mantis.ilias.de/view.php?id=44703
The makeDirParents method inherits the first existing parents permissions, so that works here.

### DIFF
--- a/Services/FileServices/classes/class.ilFileUtils.php
+++ b/Services/FileServices/classes/class.ilFileUtils.php
@@ -151,6 +151,7 @@ class ilFileUtils
                     $item->getPath(),
                     strlen($sourceDir)
                 );
+                self::makeDirParents(dirname(ILIAS_WEB_DIR . '/' . CLIENT_NAME .'/' . $itemPath));
                 $stream = $sourceFS->readStream($item->getPath());
                 $targetFS->writeStream($itemPath, $stream);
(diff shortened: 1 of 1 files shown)


chfsx:
Thanks a lot! I'll cherry pick this to the other branches as well

Services/FileServices/classes/class.ilFileUtils.php
```diff
@@ -151,6 +151,7 @@ public static function rCopy(
                     $item->getPath(),
                     strlen($sourceDir)
                 );
+                self::makeDirParents(dirname(ILIAS_WEB_DIR . '/' . CLIENT_NAME .'/' . $itemPath));
```
mjansenDatabay, line 154: IMO this looks strange @mbecker-databay / @chfsx .

Why should we always create directories (recursively) within the `ILIAS_WEB_DIR`? What happens if `rCopy` is called in the context of the `ILIAS_DATA_DIR`?

The directories are derived based on the source and target paths in the upper section of this method. Using the constnat `ILIAS_WEB_DIR` looks not correct IMO.

chfsx:
Valid point, I did not think about this, sorry. I reverted the commit for the moment and will create a new with a general solution

chfsx:
The problem seems to be the default visibilty for directories in FlySystem (which may have changes in an update). see #9458

PR #9454, opened 2025-04-24
Auth/SessionReminder: Get rid of `jQuery`

mjansenDatabay:
This PR suggests to get rid of `jQuery` in the session reminder. The commit also contains the following changes:

- Use ES6 classes and `rollup` for bundling, instead of having one large JS file
- Don't use cookies anymore as storage mechanism, use the `localStorage` instead (with scoped keys to avoid conflicts with other components)
- Change hash algorithm and add further attributes als salt


components/ILIAS/UICore/classes/class.ilGlobalPageTemplate.php
```diff
@@ -105,7 +107,8 @@ protected function prepareBasicJS(): void
         $sessionReminder = new ilSessionReminderGUI(
             ilSessionReminder::byLoggedInUser(),
             $this,
-            $this->lng
+            $this->lng,
+            $this->logger_factory
```
mjansenDatabay, line 111: @thibsy I had to change this file to retrieve the logger factory from the DIC and inject it to the session reminder instance.

thibsy (approved):
LGTM, thx!

PR #9460, opened 2025-04-24
Export: fix error on import when using symlinks

chlulei:
Link: https://mantis.ilias.de/view.php?id=43708


smeyer-ilias (approved):
Thanks @chlulei

PR #9467, opened 2025-04-25
Tracking: remove unused table (44940)

schmitz-ilias:
This PR removes the unused table `catch_write_events` (see [44940](https://mantis.ilias.de/view.php?id=44940)), but leaves related methods in `ilChangeEvent` as stubs. In a separate PR for trunk (#9470), those stubs (and all usages of the write-only method `_catchupWriteEvents`) will be removed.

Note that I'll also provide a sister PR to this for 10 and trunk, which also puts the new `ilTrackingSetupAgent` into `Tracking.php`: #9468


smeyer-ilias (approved):
Thanks @schmitz-ilias

PR #9468, opened 2025-04-25
Tracking: remove unused table (44940) for 10

schmitz-ilias:
Sister PR to #9467 for 10 and trunk.


smeyer-ilias (approved):
Thanks @schmitz-ilias

PR #9471, opened 2025-04-25
FileSystem: Allow output options for `LegacyImages::convertToFormat`

mjansenDatabay:
This PR suggests making "Output Options" injectable when converting and image with `src/Filesystem/Util/Convert/LegacyImages.php`.

See discussion: https://discord.com/channels/819582183290437642/1022408960302653481/1365248247102443565

### DIFF
--- a/src/Filesystem/Util/Convert/LegacyImages.php
+++ b/src/Filesystem/Util/Convert/LegacyImages.php
@@ -166,13 +166,15 @@ class LegacyImages
         string $path_to_output,
         string $output_format,
         ?int $width = null,
-        ?int $height = null
+        ?int $height = null,
+        ?ImageOutputOptions $output_options = null
     ): string {
         $converter = $this->image_converters->convertToFormat(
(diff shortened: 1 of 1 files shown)


chfsx:
I would actually prefer not to offer further options in the legacy implementations. Would it be possible for you to simply switch to the other implementation at these points? 

https://github.com/ILIAS-eLearning/ILIAS/blob/91a54643b5a34f81da6c754129f535a39c898f1a/components/ILIAS/Filesystem/src/Util/Convert/Converters.php#L38

As far as I remeber, you can already set the options yourself if you want to:

https://github.com/ILIAS-eLearning/ILIAS/blob/91a54643b5a34f81da6c754129f535a39c898f1a/components/ILIAS/Filesystem/src/Util/Convert/Images.php#L163

mjansenDatabay:
This should be possible @chfsx .

mjansenDatabay:
@chfsx See #9495 , thanks a lot.

PR #9478, opened 2025-04-28
TA: 44901, single-encode questionlist titles

nhaagen:
https://mantis.ilias.de/view.php?id=44901


dsstrassner:
Could we use this PR as well for ILIAS 9 & Trunk? ;)
ILIAS 9 is also broken.

nhaagen:
Actually yes, this applies to 9, 10 and trunk.
However, there was a conflict with 9, so here is another PR: https://github.com/ILIAS-eLearning/ILIAS/pull/9486

kergomard:
Thank you very much for both PRs @nhaagen ! Merged both and picked to trunk.

PR #9481, opened 2025-04-28
11/data/mail address

catenglaender:
This is an EmailAddress Data Type for more consistent and accurate validation.

Since ILIAS is used internationally, we are allowing many more characters than most Email Validators do. Umlauts ä, ö, ü and accented vowels become more and more common since updated standards technically allow them. Sadly, most existing checkers only let a subset of ASCII characters through. For this strict validation you can use isAscii() to verify that the email address is safe by old standards.

If isAscii() returns false, but the object did not throw an error on construction, it's a valid Unicode email address. However, most email servers (and php itself) cannot handle those optimally by default. One day when this Data Object is actually used for Email Addresses, PHP itself and the web server still need to be configured to handle Unicode strings correctly.

### DIFF
--- a/components/ILIAS/Data/src/EmailAddress.php
+++ b/components/ILIAS/Data/src/EmailAddress.php
@@ -1,0 +1,229 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\Data;
+
+/**
+ * An Email Address is a common personal address for people online.
+ * This class ensures the address follows common format requirements.
+ * It also splits the address into some of the parts outlined in RFC 2822 that can then be retrieved separately.
+ * Non-ASCII characters are technically valid in the local (RFC 6531) and the domain (RFC 3490) parts, but...
+ *      - in the local part they are only supported by a small amount of mailservers
+ *      - in the domain part they are converted to ASCII punycode
+ * Because of this limited support, PHP's own FILTER_VALIDATE_EMAIL does not allow Unicode characters. However, we don't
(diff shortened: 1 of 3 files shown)


components/ILIAS/Data/src/EmailAddress.php
```diff
@@ -0,0 +1,223 @@
...
+
+        $this->address_full = $address;
+        return $address;
+    }
+
+    protected function digestDomainPart(string $address): string
+    {
+        [, $domain] = explode('@', $address, 2);
+
+        $this->is_domainpart_ascii = $this->checkAscii($address);
```
mjansenDatabay: Do we want to check the full `$address` or only the extracted `$domain` part here?

catenglaender: Oh... You are absolutely right, it should be only the domain part. I will change that.

Right now, these checks have no real effect on anything, but maybe (in the future) admins could somewhere tell the instance what the email server can handle and require ASCII in only domain part, only in local parts or require ASCII in both, so the check has to be targeted.

catenglaender:
:heavy_check_mark: Rebased
:heavy_check_mark: Corrected domain part ASCII check
:heavy_check_mark: Corrected PHPStan check error in factory
Remaining failing unit tests are deprecation warnings not related to this PR

components/ILIAS/Data/tests/EmailAddressTest.php
```diff
@@ -0,0 +1,242 @@
...
+use ILIAS\Data\EmailAddress;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Tests the Link Datatype
+ */
+class EmailAddressTest extends TestCase
+{
+    #[DataProvider('validEmailProvider')] public function testEmailObjectConstruction(
```
klees: Please move the `DataProvider` attributes to their own lines.

klees (approved):
Hi @catenglaender,

only a minor and maybe personal formatting request. The rest LGTM.

Kind regards!

klees:
Hi @mjansenDatabay,

would you be so kind and take care of a final review and approval, to follow the "not the same SP" rule? What's your stance on the formatting issue?

Thanks a lot!

components/ILIAS/Data/tests/EmailAddressTest.php
mjansenDatabay: I agree, the PHP attribute should be moved to it's own line.

mjansenDatabay:
Hi @klees / @catenglaender ,

yes, I will merge this after the formatting issue (and the merge conflict) has (have) been solved.

Best regards
Michael

catenglaender:
Thank you everyone! :slightly_smiling_face: 

formatting fixed
merge conflict solved

mjansenDatabay:
Thx @catenglaender

PR #9348, opened 2025-04-10
Mantis Issue 44650

Hufschmidt:
See https://mantis.ilias.de/view.php?id=44650 for full details and explanation.

**TL;DR:**
ECS Member Assignment fails to fetch valid list of ECS members and reassigns participants on each update, also triggering Notification mails each time.

Accidentally got broken with a type cleanup commit [here](https://github.com/ILIAS-eLearning/ILIAS/commit/78042fe653ccfd16964a474ee4d9b6aa09f9274f). This one re-introduces the required check against 0 (in addition to NULL) and adds some logging to more easily identify this issue.

PR #9482, opened 2025-04-28
ECS: Fix Mantis #44591

Akumatic:
https://mantis.ilias.de/view.php?id=44591


Hufschmidt:
Thanks, this should also improve the issue in https://mantis.ilias.de/view.php?id=44650 and with the DB-Update step might make this PR (mostly) irrelevant: https://github.com/ILIAS-eLearning/ILIAS/pull/9348 (At least the first commit)

PR #9483, opened 2025-04-28
Add support for role-based upload-limit to legacy code

Hufschmidt:
## TL;DR
This set of patches extend the awesome _role-based upload limit_ feature implemented within [Customizable Maximum File Upload Size](https://docu.ilias.de/ilias.php?baseClass=ilwikihandlergui&cmdNode=16g:rq&cmdClass=ilobjwikigui&cmd=viewPage&ref_id=1357&wpg_id=6387) for the rest of the ILIAS code base, effectively making all core components (excluding maybe some plugins) support this new role based upload limit. Note though that this does not magically enable chunking support for all those components, but still only works within the limits of the PHPs `upload_max_filesize` and `post_max_size` settings as before.

## History
Here in Marburg we have had our own set of custom patches for a customizable upload limit since rougly ILIAS 5.4. Initially using user-defined-fields instead of having a role-based approach, which worked well but we never felt like our patches were implemented well enough to share with the world. 

With the release of ILIAS 9 and the new *Customizable Maximum File Upload Size* feature we had hoped to be able to fully drop our custom patches as they where quite extensive. Sadly this new feature so far was only implemented for the new KitchenSink UI elements, missing support in places like:
- Bibliography (Re-Upload)
- Question Types (Mostly related to image upload)
  - K-Prim
  - Matching
  - Multiple-Choice
  - Single Choice
  - Image Map
  - File
- SCORM File-Upload
- TinyMCE
- Mail Attachement Upload
- Classes using FileInputGUI, ilFileWizardInputGUI or ImageFileImport
  - Objekt import Wizard
- Plugins:
  - InteractiveVideo

None the less we'd like to thank TH Köln, Fabian Schmid, Thibeau Fuhrer and Lukas Zehnder for funding and implementing much of the required groundwork to get this into the ILIAS core and allowing us to try and update our internal patches into something that might be viable for the rest of the ILIAS community.

## About the patches
The commits should be mostly self-explanatory, although there might be a few places where additional comments are in order:
- There are two commits (285d8793682a248c3142255be507998d9842042a, a2acf45b646da446de4899d58c121a4407e2d7ad) that drop (deprecated) methods that have been fully replaced in the core, but might still be used by plugins. This PR does not require these to be kept but I thought it a good idea to cleanup the codebase while at it.
- A new helper class `UploadFileLimits` has been implemented as a (legacy support) analog to `UploadLimitResolver` for code that has not implemented the new KitcheSink UploadHandler-Interface UI-Element (or maybe never will, like WebDAV). Ideally most places that now use this class can/should eventuelly be ported to the "real" `UploadLimitResolver`, the only exception being maybe places that are not related to UI, like the WebDAV code.
- The new helper class has been added to the DIC (as `upload_file_limits`) much like the current `UploadLimitResolver` is available as `ui.upload_limit_resolver`. 
- Since I wanted this DIC entry to be always available, regardless of the ILIAS user context and since it ideally should be available before `UploadLimitResolver` (_Since it needs it to read the current PHP limits_), I created a new entry for it in `ilInitialisation`. I'm not 100% sure this **really** required, but I wanted to error on the side of caution since there are a lot of places that need to know the currently max/best/role-based upload limit.
  - Happy to get feedback on this.
- I replaced the instances of `UploadLimitResolver` inside most ILIAS test question classes with the new `UploadFileLimits`, since no class was currently using the new KitchenSink UploadHandler-Interface UI-Element anyway. Once the test question classes are updated to use the new Kitchensink UI elements implementing the UploadHandler-Interface this should of course be changed again.
- Otherwise the commits try to:
  - Replace any non-centralized access to ini_get(`upload_max_filesize`) and ini_get(`post_max_size`) towards `UploadFileLimits` or `UploadLimitResolver`, whatever makes more sense, eg. whether a UploadHandler-Interface handler is available or not.
  - Replaces `ilFileUtils::getFileSizeInfo()` and `ilFileUtils::getPhpUploadSizeLimitInBytes()` with role-based upload-limit aware methods on `UploadFileLimits` or `UploadLimitResolver`.
  - Changes any non-centralized calculation/conversion of filesize from bytes to string into (new) helper methods on `UploadFileLimits` or `UploadLimitResolver` using the new DataSize data structure. (_This saves having to import/use DataSize in some places where `UploadFileLimits` or `UploadLimitResolver` are available anyway._)
  - Minor code cleanup and improvements
  - Dropping some now unused class methods
  
I sincerely hope that these changes can find their way into release_9 and ideally release_10 as well.
On first glance not much in relation to these patches has been changed in ILIAS 10 here (yet).

## WIP
- ~~These patches have been tested and are currently live since today (28.04.2025) on our production system, but another review and some user-facing testing might be useful.~~ Patches have been in production use for about 6 months without issues.
- Feedback is more then welcome. :wink:

### ilInitialisation Order
I might still change the `ilInitialisation` order, putting `$DIC['upload_file_limits']` setup back into the user-based context (again).

Looking at the changes that have been made, I can see the following categories:
1. Places that should only be reachable by the GUI and where I've added `$DIC['upload_file_limits']` access in their constructors. (Commits: a6daabcf779ed3de26954b4318718ae9761a6696)
2. Places that previously where already using the pre-existing `$DIC['ui.upload_limit_resolver']` helper. (Commits: b6a3371d42382903c4a4bf505e16c0f2294ccc52, a4d24e82f3b3f331680f0a25af6027856b396c2a, 7f18881076c7fa9a25752c9e8e7431047d0930b1)
3. Places where the pre-existing `$DIC['ui.upload_limit_resolver']` was replaced by `$DIC['upload_file_limits']` until a new KitchenSink UI element will be used here. (Commit: b25118e753970f94d6865066d30a325e09dac9e1)
4. Places that replace the static `ilFileUtils` methods that have been marked as deprecated. (Commits: e9d37e8dd0719c2366f16890d6674640248c0edb, 0cf35afa5abdb60708cc541ee08f607329458e4d)

Of those places only point **4.** might have issues with the ILIAS init order as the static methods were available/initialized once the `ilFileUtils` class was first auto-loaded. Both commits seemingly involve only GUI classes and WebDAV which uses a context with `hasUser()`, `hasHTML()` and `initClient()` active, so using a user-context **should** be fine.

Both point **1.** and **4.** might cause *undefined array key errors* if the `upload_file_limits` instance is not available in certain contexts when their constructor is called, but seeing as these are all GUI-Class constructors (except WebDav) this **should** be fine.

### DIFF
--- a/Modules/Bibliographic/classes/class.ilObjBibliographicGUI.php
+++ b/Modules/Bibliographic/classes/class.ilObjBibliographicGUI.php
@@ -432,9 +432,9 @@ class ilObjBibliographicGUI extends ilObject2GUI implements ilDesktopItemHandlin
         $bibl_upload_handler = new ilObjBibliographicUploadHandlerGUI($rid);
 
         $max_filesize_bytes = $this->upload_limit->getBestPossibleUploadLimitInBytes($bibl_upload_handler);
-        $max_filesize_mb = round($max_filesize_bytes / 1024 / 1024, 1);
-        $info_file_limitations = $this->lng->txt('file_notice') . " " . number_format($max_filesize_mb, 1) . " MB <br>"
-            . $this->lng->txt('file_allowed_suffixes') . " .bib, .bibtex, .ris";
+        $max_filesize_mb = $this->upload_limit->getBestPossibleUploadLimitInfo($bibl_upload_handler);
+        $info_file_limitations = $this->lng->txt('file_notice') . " " . $max_filesize_mb . " <br>" . $this->lng->txt('file_allowed_suffixes') . " .bib, .bibtex, .ris";
+
         $section_replace_bibliographic_file = $this->ui_factory
             ->input()
             ->field()
--- a/Modules/Bibliographic/classes/class.ilObjBibliographicGUI.php
+++ b/Modules/Bibliographic/classes/class.ilObjBibliographicGUI.php
@@ -431,7 +431,7 @@ class ilObjBibliographicGUI extends ilObject2GUI implements ilDesktopItemHandlin
         $rid = $bibl_obj->getResourceId() ? $bibl_obj->getResourceId()->serialize() : "";
         $bibl_upload_handler = new ilObjBibliographicUploadHandlerGUI($rid);
 
-        $max_filesize_bytes = $this->upload_limit->getPhpUploadLimitInBytes();
+        $max_filesize_bytes = $this->upload_limit->getBestPossibleUploadLimitInBytes($bibl_upload_handler);
         $max_filesize_mb = round($max_filesize_bytes / 1024 / 1024, 1);
         $info_file_limitations = $this->lng->txt('file_notice') . " " . number_format($max_filesize_mb, 1) . " MB <br>"
             . $this->lng->txt('file_allowed_suffixes') . " .bib, .bibtex, .ris";
--- a/Modules/File/classes/Versions/class.ilFileVersionFormGUI.php
+++ b/Modules/File/classes/Versions/class.ilFileVersionFormGUI.php
@@ -17,7 +17,6 @@
  *********************************************************************/
 
 use Psr\Http\Message\RequestInterface;
-use ILIAS\Data\DataSize;
 use ILIAS\UI\Implementation\Component\Input\UploadLimitResolver;
 
 /**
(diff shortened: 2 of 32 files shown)


rfcmaXi:
I would very much welcome it if this significant change could be implemented in the core.

In my opinion, we currently have an unfinished status of the role-based upload limit, which means that it cannot actually be used productively if it is not conceptually effective at all cases.

klees:
Hello everyone,

@FraenzePacita will get in touch and take care of this.

Kind regards!

FraenzePacita:
Unfortunately, time has flown by... As ILIAS 9.x is now only receiving security updates, I am closing this PR so that the issue can be investigated again for newer versions of ILIAS.

PR #9492, opened 2025-04-30
Include autoloader

qualitus-dahme:
https://mantis.ilias.de/view.php?id=44484


QualitusHartwig:
@qualitus-dahme FYI I picked it to trunk, however it's not enough and both files still return 500 in trunk.

PR #9494, opened 2025-04-30
Component: 42340, add AgentFinder to ModernComponents

nhaagen:
https://mantis.ilias.de/view.php?id=42340
for 10: https://github.com/ILIAS-eLearning/ILIAS/pull/9493


klees (changes requested):
Hi @nhaagen,

thanks a lot, looks plausible. One change request:

* [x] Let's treat `AgentFinder` as a service, so `$define[\ILIAS\Setup\AgentFinder]` and then implement it as well. We do not really care about any specific implementation here, its just that we want to talk about any service that may give agents.

Kind regards!

klees:
Hehe, really nice to see that this saves some lines of code.

PR #9495, opened 2025-04-30
Certificate: Use best quality when uploading background images

mjansenDatabay:
See: #9471


chfsx:
@mjansenDatabay of course this looks a bit like an overkill but I think this comes from the fact, that this place uses "legacy" storage by writing the file itself to the FS. thanks for using this instead of adding options to the legacy converters

mjansenDatabay:
> @mjansenDatabay of course this looks a bit like an overkill but I think this comes from the fact, that this place uses "legacy" storage by writing the file itself to the FS. thanks for using this instead of adding options to the legacy converters

Thanks for the feedback @chfsx . We don't need this code for ILIAS >= 10.x (with 10.x the IRSS is used), so this "overkill" is just temporary.

PR #9496, opened 2025-04-30
Handle window current Externallinks

jheim2:


jheim2:
sorry wrong repository!

PR #9500, opened 2025-04-30
ContainerSettings: Limit mail prefix for subjects

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=45003

If approved, this has to be picked to `release_10` and `trunk`.


kergomard:

PR #9504, opened 2025-04-30
10/UI/Table Data & Ordering accessibility validation 44005 44065

catenglaender:
# Issue

DataTable and OrderingTable UI Component did not pass W3C Validator.

image

# Changes

image

I went on a mysterious journey into the depth of W3C and MDN :mage_man: 

There is a bug in the W3C validator: If you use `aria-colindex` on `<td>`, everything explodes with very unhelpful errors for the table header. The actual problem was that you can't have `aria-colindex` on `<td>` if you also have it on `<th>`, but none of the errors said that. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-colindex

Furthermore many, many roles are not necessary and even discouraged on `<table>`, `<tr>` and `<td>` in a `display: table;` context, so I simplified some tags. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/table_role

`aria-colindex` and `aria-colcount` and the colspan for the multi select footer were calculated incorrectly as
1. `aria-colindex`starts counting with 1 and doesn't allow 0.
1. we add many columns that are not in the actual table data (multi-select & drag and drop column, position column, single select action column).

The logic is a bit hacky, as the OrderingTable uses the same column for multi-select and the drag and drop handle.

Input/Viewcontrol Fieldselection had a button in an `<ul>` not wrapped in `<li>`.

# Not quite done

I will do the Unit Tests after code review, when we are sure the html structure will not change.

The Kitchen Sink page of the Ordering Table still fails to pass because something is not right with the UI form input select-field-input setting selected twice. That's another issue though.

### DIFF
--- a/components/ILIAS/UI/src/Implementation/Component/Table/Renderer.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Table/Renderer.php
@@ -217,10 +217,23 @@ class Renderer extends AbstractComponentRenderer
             $component->getAdditionalParameters()
         );
 
+        // these column counts and index numbers are meant for aria attributes in the html tpl
+        $compensate_col_index = 1; // aria-colindex expects counting to start with 1, not 0
+        if ($component->hasMultiActions()) { // adds column with checkbox before first data column which is numbered 1.
+            $compensate_col_index += 1;
+        }
+        $compensate_col_count = 0; // count starts with 1, only needs to be compensated for special columns
+        if ($component->hasMultiActions()) {
+            $compensate_col_count += 1;
+        }
+        if ($component->hasSingleActions()) { // adds column with action dropdown at the very end
+            $compensate_col_count += 1;
+        }
+
         $id = $this->bindJavaScript($component);
         $tpl->setVariable('ID', $id);
         $tpl->setVariable('TITLE', $component->getTitle());
-        $tpl->setVariable('COL_COUNT', (string) $component->getColumnCount());
+        $tpl->setVariable('COL_COUNT', (string) $component->getColumnCount() + $compensate_col_count);
         $tpl->setVariable('VIEW_CONTROLS', $default_renderer->render($view_controls));
 
         $sortation_signal = null;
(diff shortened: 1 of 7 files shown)


dsstrassner:
I hope the assignments are correct.

https://mantis.ilias.de/view.php?id=44065 and https://mantis.ilias.de/view.php?id=44005

Amstutz:
Thx Denis, afaik macht bei CaT Arbeiten @klees immer zuerst das interne Review.

klees (dismissed):
Hi @catenglaender,

aaah, look at us, not taking care or knowing s**t about accessibility ^^

- [x] `grid`-role: If I remember correctly, that role was the key to allow for movement via arrow-keys in the table. Did you check for that feature after removal of the `grid` role?

The rest looks good, thanks so much for the investigation and improvements.

Best regards!

catenglaender:
Thanks for the input! :slightly_smiling_face: 

* I restored the grid role (gridcell is not needed for `<td>` as far as I understand the best practices)
* Unit tests are updated
* @lukastocker will commit an even better fix for the viewcontrol dropdown in a separate PR related to another issue

Amstutz:
thx for the improvements. @klees is this ready for the coordinators to take a look at?

Amstutz:
Hi @catenglaender 

thx a lot!

- [x] I am afraid a recent merge generated a conflict. Can you resolve this?

Thx a lot

catenglaender:
:heavy_check_mark: Resolved merge conflicts
:exclamation: Failing Unit Tests are unrelated to the files in this PR
:exclamation: @kergomard I took out a change of yours, where you fixed the alignment of the header and cells using the template block `header_rowselection_ordering_no_multi_actions`, because the amount of columns and content shown when there are no multi actions or the drag and drop is disabled has changed a bit. It's correct now as far I can tell from the examples (see screenshots below) but if you know of other cases I should test, please let me know.
image
image

kergomard:
Hi @catenglaender 
I didn't check your changes out and test them, but as far as I understood the code your changes make my changes unnecessary.

Best,
@kergomard

catenglaender:
Thanks for your input! Very much appreciated.

Alright, this one is good to go then :)

PR #9505, opened 2025-05-02
PRG: 44366, do not show completed/accredited assignments when using f…

nhaagen:
https://mantis.ilias.de/view.php?id=44366


klees:
Picked to trunk.

PR #9507, opened 2025-05-02
UI: Add Empty Header Cell in Ordering-Table

kergomard:
Possible fix for https://mantis.ilias.de/view.php?id=45020

### DIFF
--- a/components/ILIAS/UI/src/Implementation/Component/Table/Renderer.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Table/Renderer.php
@@ -643,6 +643,10 @@ class Renderer extends AbstractComponentRenderer
             $tpl->setVariable('POS_INPUT_TITLE', $this->txt('table_posinput_col_title'));
         }
 
+        if (!$component->hasMultiActions()) {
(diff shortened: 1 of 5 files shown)


Amstutz:
Hi @kergomard 

Thx a lot for improving a11y with this fix. Can you provide a unit, making sure this issue will not slip our attention again. After this, we consider this ready to be merged.

Thx

kergomard:
Thank you very much for the quick feedback @Amstutz !

I added a Unit Test and hope that this is more or less, what you expected. I moved things a around a little bit as there seemed to be no corresponding TestCase, but let me know if I should revert something or move things around.

Thank you very much and best,
@kergomard

Amstutz:
Hi @kergomard 

Thx for adding the test. Just as a note for future cases. We try to avoid rendering tests with huge strings of rendered content, since they tend to be expensive to maintain (need to be changed, if one component involved in the rendering is changing). Often a suitable approach is to check if the rendered string contains a certain element of interest (e.g. with string contains).

Thx a lot for your work!

PR #9511, opened 2025-05-05
UI: 44004, fix HTML validation error in ViewControl.

lukastocker:
https://mantis.ilias.de/view.php?id=44004

According to the W3 HTML Validator it is not allowed to have "button" as a child in a "ul" element. It has to be wrapped with an "li" tag.

Before:
ViewControl_Button_HTML_Validation_beforeFix

After:
ViewControl_Button_HTML_Validation_afterFix

The button functionality was successfully tested on my local instance after implementing the changes.


Amstutz:
Hi @lukastocker 

Thx a lot for improving a11y!

One question:

- [x] would it not be better to place the button outside of the listing?

One thing todo:

- [x] please write a unit test, making sure, this solution stays intact.

thx a lot

lukastocker:
> Hi @lukastocker
> 
> Thx a lot for improving a11y!
> 
> One question:
> 
> * [ ]  would it not be better to place the button outside of the listing?
> 
> One thing todo:
> 
> * [ ]  please write a unit test, making sure, this solution stays intact.
> 
> thx a lot

Hi @Amstutz ,

the dropdown list opens up after clicking a toggle button. If you implement the apply button outside of the ul/li list it will be positioned incorrectly and therefore look like it does not belong to the list anymore.

The test "testViewControlFieldSelectionRendering" in the "ViewControlFieldSelectionTest.php" file was changed accordingly.

Amstutz:
HI @lukastocker 

But that would be fixable by CSS, right? So one question still remains:

- [x]  Button in List: Is this semantically correct?

This is a question, not an opinion, I am really not sure :-) Thx for some explanation.

Kindly,
@Amstutz

lukastocker:
Hi @Amstutz ,

it is possible to put a button in a list but after some consideration semantically I would not agree. Therefore I discussed this issue with @catenglaender and updated this PR with a new solution.

A div is now added around ul and the apply button, while the apply button is also moved outside of ul. JavaScript and SCSS had to be adjusted accordingly but the adjustments were rather small.

I tested other ViewControls  like sortation to check if they still function on my local instance and the style is still the same as before. 

ViewControl_button_Validation

Amstutz:
hi @lukastocker 

Thx a lot for the reply and the improvements. This is almost good to go, however, 

- [x] SCSS: We have variables for margin and paddings, right? Please use them or indicate why not.

Thx a lot
@amstutz

lukastocker:
Hi @Amstutz ,

there where no variables for 2 and 5px regarding margin/padding. Therefore I used the variables for 3 and 6px. Visually the 1px more does not stand out. I did change it in both .dropdown-menu references in the SCSS file to have an consistent style.

Amstutz:
thx a lot!

Amstutz:
I just observed while picking this for trunk, that there have been some minor changes in tpl.viewcontrol_fieldselection.html can you, preventing a easy pick. Can you check them out, make sure that the js also works for trunk and then, provide a pr for trunk as well? would be highly appreciated. thx!

PR #9517, opened 2025-05-06
UI: Tagify.js No Without min

kergomard:
Dear UI coordinators

There is a small change in the tagify-library, so that it is now saved without the "min" in its name.

Best,
@kergomard


klees:
Thx @kergomard!

PR #9519, opened 2025-05-06
[FIX] 0044037: Repository object plugins are not listed in "add new object" modal

chfsx:
This fixes mantis issue https://mantis.ilias.de/view.php?id=44037

I assigned both @klees and @alex40724 since I made changes in two components for this. maybe there is a better solution

### DIFF
--- a/components/ILIAS/Component/classes/class.ilComponentInfo.php
+++ b/components/ILIAS/Component/classes/class.ilComponentInfo.php
@@ -14,8 +14,7 @@
  * https://www.ilias.de
  * https://github.com/ILIAS-eLearning
  *
- ********************************************************************
- */
+ *********************************************************************/
 
 declare(strict_types=1);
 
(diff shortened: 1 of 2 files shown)


klees:
Hi @chfsx,

thanks a lot for looking into this. What I don't get: Can't we simply remove that test `if (!in_array($mod->getType(), ilComponentInfo::TYPES, true)) {..` for the type? If there is no object type in the object definition for that component (which should be true for all non-repo-object components...) we will simply be ignored later on...

Kind regards!

chfsx:
I didn't want to make any major changes here, but yes, the check can probably be removed entirely. Should I adjust that or would you prefer to do it as part of the merge?

klees:
Hi @chfsx,

I think we should wait for @alex40724 opinion some more days...

Kind regards!

alex40724 (approved):
Thanks for the PR. The changes in "Repostory" look good.

klees:
Picked to trunk.

PR #9521, opened 2025-05-07
ILIAS UI: add new glyphs to il-icon font

iszmais:
https://mantis.ilias.de/view.php?id=45001

Add the following glyphs to the il-icon font:
- unselect
- select
- clear
- checked
- unchecked


Amstutz:
Thx a lot for adding the glyph to the font!

PR #9527, opened 2025-05-08
T&A Use htmlspecialchars for answer output in correction statistics gui

thojou:
This PR is related to Mantis-Ticket https://mantis.ilias.de/view.php?id=44305

This PR replaces the use of `ilHtmlPurifierFactory` with the native `htmlspecialchars` function. The change ensures that HTML tags are preserved in the output rather than being stripped, aligning with the intended display behavior.

If merged, this must be cherry-picked into release_10 and trunk

Best,
@thojou


kergomard:

Best,
@kergomard

PR #9528, opened 2025-05-08
Survey: Create Object on Import

kergomard:
See: https://mantis.ilias.de/view.php?id=43826


alex40724:
@kergomard Thanks for the PR. Fixed with similar code.

PR #9530, opened 2025-05-08
WAC: Prevent sending HTTP cookies with invalid domain

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=45047


chfsx:
Many thanks, I picked and adapted to release_10 and trunk as well

PR #9533, opened 2025-05-08
Tree: Fix ilPathGUI Modifies ref_id

kergomard:
Hi @fwolf-ilias 

I know, this is not very beautiful, but it is actually the most localized I could come up with.

See: https://mantis.ilias.de/view.php?id=44882

If accepted this needs to be picked to ILIAS 10 and trunk.

Thank you very much and best,
@kergomard


kergomard:
The reference to the correct mantis ticket is: https://mantis.ilias.de/view.php?id=44952

PR #9536, opened 2025-05-09
edited description for alt skin creation from templates

fsesterh:
Edited the description for the alternative way to create skins by either copying from templates/default or by cloning the git repo. Currently the readme tells users to copy templates/default to Customizing/global/skin, but it should be placed in Customizing/global/skin/STYLENAME. Consequently users then have to move the template.xml up one directory and process with renaming delos.css and delos.scss.


Amstutz:
Thx a lot for improving the docu @fsesterh, highly appreciated!

PR #9537, opened 2025-05-09
T&A: fixes 44936

okaufman:
This will fix https://mantis.ilias.de/view.php?id=44936


kergomard:
Thank you very much for the PR @okaufmann !

This is ILIAS 9 only, so no picking needed.

Best,
@kergomard

PR #9538, opened 2025-05-09
Calendar: do not include course info files in downloads of manually generated appointments (43866)

schmitz-ilias:
This PR fixes [43866](https://mantis.ilias.de/view.php?id=43866).

An alternative would be to walk back the decision from [21523](https://mantis.ilias.de/view.php?id=21523), and show course information also for manually generated appointments.


smeyer-ilias (approved):
Thanks @schmitz-ilias

PR #9539, opened 2025-05-09
do not pass empty format to LOM (43236)

schmitz-ilias:
This PR fixes [43236](https://mantis.ilias.de/view.php?id=43236) by not passing empty strings as format to the LOM service. I should probably have a look at making the API a bit less fussy in this regard, but that is not super straightforward.

The root cause of the issue is that `ilObjFile::getFileType` gives back an empty string (at least at the time when the object is updated). If that is not the expected behavior, then this should probably be fixed differently.

Cheers, @schmitz-ilias


chfsx:
Thanks a lot, I picked to trunk as well

PR #9541, opened 2025-05-09
45060 Fix undefined array key obj_link

padvincenzo:


alex40724:
Thanks for the PR. This is fixed similarly in the code now.

PR #9542, opened 2025-05-09
UI: 43719, adjust language variable for closing...

lukastocker:
... a prompt.

Also fix the amount of links shown after using the back button in a prompt.

https://mantis.ilias.de/view.php?id=43719

**Note:** 
The language variable does not need to be named "close_prompt" as there's is already the standard variable "close" for this specific button.


Amstutz:
thx!

PR #9548, opened 2025-05-12
SCORM: Format datetime instance before database update

mjansenDatabay:


mjansenDatabay:
Hi @qualitus-dahme ,

this is a trivial fix, do you have any objections?

Best regards,
Michael

QualitusHartwig:
@mjansenDatabay Hi Michael, I have no objections and will create a new PR for 10+ since this one can't be merged to 9 anymore.

PR #9550, opened 2025-05-12
Course/Membership: Fix user selects no option in "select" course user…

mjansenDatabay:
… data

See: https://mantis.ilias.de/view.php?id=44984


smeyer-ilias:
Thanks @mjansenDatabay , merged in 10, 11 trunk

PR #9551, opened 2025-05-12
44970: Fix Shuffle for Cloze Question in ILIAS Learning Module

thojou:
This PR is related to Mantis Ticket https://mantis.ilias.de/view.php?id=44970

This update modifies the JavaScript logic to use the shuffle attribute from individual cloze question gaps instead of reading it globally from the entire question. This allows each gap to independently control whether its options are shuffled. Previously, the shuffle state was hardcoded to true for all gaps, limiting flexibility.

Best
@thojou


dsstrassner:
Hi @kergomard & @thojou,

 is there a need to review this by others? Do we need the approval of @qualitus-dahme?

kergomard:

No, I don't think an review by anybody else is necessary. The file is in completely the wrong place and we need to clean this up one day, ...but this is for later.

Best,
@kergomard

PR #9552, opened 2025-05-12
[FIX] #45068 T&A: Fatal Error When Copying Courses Containing Certain…

jheim2:
… Tests


kergomard:
Thank you for the pull request @jheim2 

This is surely not the best solution in the world (we should figure out, why this key is/can be missing), but it doesn't make things worse as far as I can see.

Best,
@kergomard

PR #9553, opened 2025-05-12
10/Plugins: 44502, template not found for plugins

nhaagen:
https://mantis.ilias.de/view.php?id=44502

ilTemplate needs to support a full path to the template (from ilPlugin) and must not amend further parts to the path.

### DIFF
--- a/components/ILIAS/UICore/classes/class.ilTemplate.php
+++ b/components/ILIAS/UICore/classes/class.ilTemplate.php
@@ -80,7 +80,7 @@ class ilTemplate extends HTML_Template_ITX
 
         $fname = $this->getTemplatePath($file, $in_module);
         if (!file_exists($fname)) {
-            throw new ilTemplateException("Template '$fname' was not found.");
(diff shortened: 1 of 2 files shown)


klees (approved):
Hi @nhaagen, hi all,

since we prove with unit tests, that the previous state of the art is preserved and the case from the ticket that was not processed correctly is now included as well, I think this is fine.

The ways in which templates are discovered is a hot mess anyway. I don't think that this adds much to it. The additional test on the other hand adds some safety.

Kind regards!

(@Amstutz or @thibsy, would you maybe play shepherd here?)

nhaagen:
🐑

thibsy (approved):
Thx for looking into this @nhaagen. Since you have tested this with various plugins, I will go ahead and merge this - changes LGTM.

thibsy:
Just stumbled over this. I picked this manually now because of the failed pipeline, but thx for the PR and fix!

PR #9554, opened 2025-05-13
[FIX] CI: apply copyright check to UI examples.

thibsy:
Hi all,

This should be a no-brainer. We have decided to streamline our UI examples, so the copyright license header is contained in every example as well. This adjusts the copyright check of the CI pipelines, so the examples are checked normally as well.

Kind regards,
@thibsy


Amstutz:
thx a lot!

PR #9555, opened 2025-05-13
WebLink: Add check for `updateLink` command

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=44299


corro:
LGTM

PR #9556, opened 2025-05-13
LegacyUI: Fix Missing JS on LastTab

kergomard:
Hi all

This PR is just for visibility. I'm going to just merge it without requiring a shepherd. I thinks this is small enough and it only concerns legacy-ui as it fixes the last tab, when not all tabs-fit the screen. There surely are more beautiful solutions.

See: https://mantis.ilias.de/view.php?id=44817

Best,
@kergomard


kergomard:
Merged and cherry-picked to trunk.

PR #9558, opened 2025-05-13
Mantis 45070 \ILIAS\FileDelivery\Services::getBaseURI swallows Port

mbecker-databay:
Adding the port


chfsx:
Thanks a lot, I picked to trunk as well

PR #9560, opened 2025-05-13
Dependencies: Add `jumbojett/openid-connect-php` v. `1.0.2`

mjansenDatabay:
This PR adds `jumbojett/openid-connect-php` v. `1.0.2` (from `master` branch as `dev-master` package with PHP 8.4 support) as composer dependency for ILIAS 11.

General Information:
* Name of the dependency: `jumbojett/openid-connect-php`
* Version: `v1.0.2`(from `master` branch as `dev-master` package with PHP 8.4 support)
* [X] the dependency's license is compatible with ILIAS' license: Apache-2.0 License

Type of dependency:
* [X] composer
* [ ] npm

Usage:
* `components/ILIAS/OpenIDConnect`

Reasoning:
* The library is used for `OpenID Connect`-based logins in ILIAS.
* When searching the web, it is the most recommended library.

Maintenance:
* There are 87 contributors.
* While there seemed to be pauses in development and some specification implementation issues during the 0.* releases, development has **resumed** since end of 2023 with the first major release. Version 1.0.2 was released in mid of September 2024.
* The library is actively maintained, there are recent commits and releases (e.g., to ensure the PHP 8.4 compatibility and to fix specification issues)
* It basically consists of one file and has 1 composer dependency (smallish), so we see ourselves in a position to maintain it if the actual maintainers stop their activities.

Links:

Alternatives:
* Actually, we don't see any real alternatives for ILIAS 11.
* Many of the packages listed when searching (via https://packagist.org/ or search engines) are derivatives of the suggested library. Other libraries (like the ones checked below) have other drawbacks and are more complicated in their usage.
* https://packagist.org/packages/league/oauth2-client - We had considered proposing this library for ILIAS 11 and even explored funding options among some of our customers. However, there were no volunteers to support the integration or a new implementation. Since it's primarily an OAuth library, significant OIDC-specific functionality would still need to be implemented on our side. As such, funding is required. Note that it also introduces `paragonie/random_compat` as a transitive dependency.
* https://github.com/socialconnect/auth : Worse maintenance, fewer contributors, huge package (the code and structure look indeed cleaner) with a lot of (unnecessary) code and more transitive dependencies.
* https://github.com/facile-it/php-openid-client/commits/master/ : Only 5 contributors, huge package with more transitive dependencies, little activity, but the code and structure look cleaner.
* https://bitbucket.org/PEOFIAMP/phpoidc/src/master/ No comment :-)
* Of course we can spend weeks to implement our own OpenID Connect client, but this will further or later result in other libraries being suggested like `\Firebase\JWT\JWT` etc.pp. To be honest, we don't want that.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,7 +70,8 @@
 		"seld/jsonlint": "^1.11",
 		"simplesamlphp/simplesamlphp": "^2.3.7",
 		"symfony/console" : "^6.4",
-		"psr/http-message": "^2.0"
+		"psr/http-message": "^2.0",
+		"jumbojett/openid-connect-php": "dev-master#bc719cc9930990a4a9916cc127b839b4ed1c89ad"
 	},
 	"require-dev": {
 		"captainhook/captainhook": "^5.24",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "1a3cfcd49d0476b7477b42bd3a8c0716",
+    "content-hash": "60882785b9edd9338f71a4b759c9eab5",
     "packages": [
         {
             "name": "brick/math",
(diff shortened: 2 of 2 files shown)


matthiaskunkel:
Jour Fixe, 26 MAY 2025: Thanks a lot for the comprehensive search for an alternative library. We agree with Michael's suggestion and accept Jumbojett as dependency for ILIAS 11.

thojou (approved):
Hey @mjansenDatabay,

I've reviewed the changes and as already discussed, with your latest changes applied, we can continue integrating these changes into the trunk.

Best Regards,
@thojou

PR #9569, opened 2025-05-13
Give correct list of file formats for errors in ImageOutputOptions

Hufschmidt:
Should be self-explanatory.

Currently throws error-message:
> InvalidArgumentException thrown with message "Format must be either jpg or png, gif given."

Which is missing webp.

This fix uses the actually array that is checked against to list supported file formats like:
> InvalidArgumentException thrown with message "Format must one of jpg, png, webp, keep, but gif was given."


chfsx:
Thanks a lot, I picked to release_10 and trunk as well

PR #9570, opened 2025-05-14
Dashboard: Add cleanup for deleted users

lscharmer:
Fix Mantis Bug: https://mantis.ilias.de/view.php?id=44731


iszmais:
@lscharmer Thank you! LGTM.

PR #9575, opened 2025-05-14
10/UI/ViewControls/Pagination next previous button instead glyph link 41562

catenglaender:
https://mantis.ilias.de/view.php?id=41562
https://mantis.ilias.de/view.php?id=38505

# Issue

In the older UI component ViewControls/Pagination (NOT Input/ViewControls/Pagination) only the next/previous glyphs are clickable and not the entirety of what looks like a button.

# Changes

This is because the deprecated glyphs with actions have been used. Now the classic ViewControls/Pagination Renderer uses a button with a glyph inside instead (like the Input variant) which is clickable all the way across.

image

Visibility of disabled state matches the global disabled button color settings now, which are a bit too bold for my taste but better for accessibility.

image

# By the way

Not directly related to the ticket, but just a note: I would prefer not showing a next/previous button if there is no such page. Why show a disabled state at all? The pages communicate sufficiently that there are pages.

### DIFF
--- a/components/ILIAS/UI/src/Component/Button/Factory.php
+++ b/components/ILIAS/UI/src/Component/Button/Factory.php
@@ -64,9 +64,9 @@ interface Factory
      *          sticky (stay visible on small screens).
      *   accessibility:
      *       1: >
-     *          Standard buttons MAY define aria-label attribute. Use it in cases
-     *          where a text label is not visible on the screen or when the label
-     *          does not provide enough information about the action.
+     *          Standard buttons MAY define an aria-label attribute. Use it in cases where a text label is not visible
+     *          on the screen, when the button doesn't include a glyph that comes with a fitting aria-label or when the
+     *          label does not provide enough information about the action.
      *       2: >
      *          Some Buttons can be stateful; when engaged, the state MUST be
      *          reflected in the "aria-pressed"-, respectively the "aria-checked"-attribute.
@@ -125,6 +125,10 @@ interface Factory
      *           The loading animation rules of the Standard Button MUST be respected.
      *   accessibility:
      *       1: >
+     *           Primary buttons MAY define an aria-label attribute. Use it in cases where a text label is not visible
+     *           on the screen, when the button doesn't include a glyph that comes with a fitting aria-label or when the
+     *           label does not provide enough information about the action.
+     *       2: >
      *          Some Buttons can be stateful; when engaged, the state MUST be
      *          reflected in the "aria-pressed"-, respectively the "aria-checked"-attribute.
      *          If the Button is not stateful (which is the default), the
--- a/components/ILIAS/UI/src/Implementation/Component/ViewControl/Renderer.php
+++ b/components/ILIAS/UI/src/Implementation/Component/ViewControl/Renderer.php
@@ -333,9 +333,12 @@ class Renderer extends AbstractComponentRenderer
 
         $f = $this->getUIFactory();
 
+        $back_btn = $f->button()->standard("", "");
+        $forward_btn = $f->button()->standard("", "");
+
         if ($component->getTriggeredSignals()) {
-            $back = $f->symbol()->glyph()->back('')->withOnClick($component->getInternalSignal());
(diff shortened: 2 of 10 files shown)


klees (dismissed):
Hi @catenglaender,

thanks for looking into this!

- [x] automated tests: Fix them!
- [x] labels: Buttons shall have some. This might imply that we would need to hide them with CSS. Please check the other view controls as well and adjust accordingly...

The rest looks fine!

Thanks!

klees:
@catenglaender Please pass back to me when you are done. Thanks!

catenglaender:
Hey, thanks for the feedback... the labels are indeed a tricky part. However, I have a slight lean to keep it how it is, because...

* Using withSymbol on a button with an empty label string renders a glyph with an aria label into it - in this case the aria-labels of the glyphs are the words "Back" and "Next".
* When I focus with a screen reader on a button the content of direct children will be read out and that includes aria-labels. I just tested it with Orca and it correctly reads "Back push-button" and "Next push-button" on focus. In this case the symbol provides the button label.
* these default glyph names are translation strings, so the glyphs in German correctly read as "zurück" and "weiter".
* kitchen sink docs could be tweaked to mention this edge case (bold text added): Standard buttons MAY define aria-label attribute. Use it in cases where a text label is not visible on the screen, **the button doesn't include a glyph that comes with a fitting aria-label** or when the label does not provide enough information about the action.

Of course, this only works if glyphs are guaranteed to always render with an aria-label, but I would argue that for the foreseeable future our glyph ui component will always come with something that can be read by screen readers whether it be an aria-label or screen-reader only text.

What do you think?

klees:
Hi @catenglaender,

sounds plausible, please propose the addition to the docu that you mention.

Thanks!

oliversamoila:
Hello @catenglaender, 
many thanks for your work on this topic. 
Both the shape and the handling with these elements seem to be of great interest. :)

I would like to make two comments: 

1 ) To your question: 
> Why show a disabled state at all? The pages communicate sufficiently that there are pages.

I remember that when we last had this discussion, we decided in favour of the permanent display so that there would be less movement on the screen. 
Otherwise, if there was no back glyph on page 1, you started on ‘page 1’ and selected ‘page 2’ after loading page 2, you would be on page 1 with the cursor position. On the other hand, such ‘jumps’ can never be completely intercepted anyway. There are also such cases when the forward and backward navigation is maintained during page selection.

2 ) I would very much like to argue that the deactivated element within the view control pagination should not be given such a visual emphasis as is currently the case in second screenshot. In my opinion, the only non-operable area actually receives the most attention.
That would be a real reason for me not to display the disabled button. 

Thank you and best regards
Oliver

catenglaender:
@oliversamoila, thanks for your input!

No worries, I won't hide the button.

Going by taste, I very much prefer the less bold design of the disabled button, but we already have had quite a few failed accessibility tests and complaints because of the disabled button's appearance in the past.

Disabled buttons are commonly being overlayed in gray or noticeably transparent - very hard to see for anyone who has problems with low contrast. That is why the color values for the disabled buttons are the way they are: very clearly gray, text/glyph in a very readable black. No transparency, no overlay, but clear, high contrast gray. Unfortunately the high contrast nature messes with the information architecture.

Maybe we get to experiment with this more in ILIAS 11, because I would highly recommend to introduce a secondary element that signifies a "disabled" state... that could be a striped background or a dithered/dotted line or filter. Maybe this case even justifies the creation of specific glyphs like back-disabled, next-disabled, where the arrow itself is dotted or broken up. If shapes unmistakeably communicate the disabled state, we can focus a bit more on aesthetics with the colors.

As it is, for 10, I would recommend using the disabled button color variables even if they are not exactly pretty. But that is what they are there for.

catenglaender:
Tests are fixed.

Ready to merge, I think.

klees:
Hi @thibsy and @Amstutz,

please carry this on...

Kind regards!

Amstutz:
Thx a lot! I merge this in the fast track, however, due to the rule-change we need to announce that in the upcoming jf.

matthiaskunkel:
Jour Fixe, 04 AUG 2025: @klees notified us about the substitution of a glyph button for pagination.

PR #9576, opened 2025-05-14
UI: Add new column type `Breadcrumb` for the UI Table

lscharmer:
This PR proposes a new column `Path` for the data table.
This addition to the UI framework became necessary as part of the Removing of Legacy-UIComponents-Service and Table project.
@oliversamoila had asked me to provide a PR so that the new UI data table could be used in various places where paths are displayed as information.

For example, a column for mapping paths is required in the following contexts:
* Contacts/Mail : Communication > Contacts > My Courses > : Mapping of paths of courses in the system
* Statistics : Administration > Learning success > Statistics and learning progress > Statistics : Paths of objects
* Roles : Administration > Roles > e.g. local roles : Paths of contexts of roles

### DIFF
--- a/components/ILIAS/UI/src/Component/Table/Column/Breadcrumb.php
+++ b/components/ILIAS/UI/src/Component/Table/Column/Breadcrumb.php
@@ -1,0 +1,25 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\UI\Component\Table\Column;
+
+interface Breadcrumb extends Column
+{
+}
--- a/components/ILIAS/UI/src/Component/Table/Column/Factory.php
+++ b/components/ILIAS/UI/src/Component/Table/Column/Factory.php
@@ -142,4 +142,16 @@ interface Factory
      * @return \ILIAS\UI\Component\Table\Column\LinkListing
      */
     public function linkListing(string $title): LinkListing;
+
+    /**
+     * ---
+     * description:
+     *   purpose: >
+     *      The Breadcrumb Column is used to show a breadcrumb path.
+     *   composition: >
+     *      The column field requires a Breadcrumbs Component.
+     * ---
+     * @return \ILIAS\UI\Component\Table\Column\Breadcrumb
+     */
+    public function breadcrumb(string $title): Breadcrumb;
 }
--- a/components/ILIAS/UI/src/Implementation/Component/Table/Column/Breadcrumb.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Table/Column/Breadcrumb.php
@@ -1,0 +1,33 @@
+<?php
+
+/**
+ * This file is part of ILIAS, a powerful learning management system
+ * published by ILIAS open source e-Learning e.V.
+ *
+ * ILIAS is licensed with the GPL-3.0,
+ * see https://www.gnu.org/licenses/gpl-3.0.en.html
+ * You should have received a copy of said license along with the
+ * source code, too.
+ *
+ * If this is not the case or you just want to try ILIAS, you'll find
+ * us at:
+ * https://www.ilias.de
+ * https://github.com/ILIAS-eLearning
+ *
+ *********************************************************************/
+
+declare(strict_types=1);
+
+namespace ILIAS\UI\Implementation\Component\Table\Column;
+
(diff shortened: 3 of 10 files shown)


klees:
Hi @lscharmer,

please assign @Amstutz, @thibsy  and me once this is not WIP anymore.

Kind regards!

iszmais (approved):

oliversamoila:
Dear UI Coordinators, 
as already announced in the last Removing LegacyUI meeting, we would like to propose an extension of the column types for the UI data table and its derivatives.
A path column with use of the already known breadcrumb.
This should enable the conversion of further legacy tables. See above in the opening paragraph by Lukas.

We would be delighted to receive initial feedback from you and would be happy to bring the topic to the JourFixe afterwards.
The PR has already received internal reviews. We have also carried out successful tests with the screen reader within examples.

Best regards and thank you.
@oliversamoila

components/ILIAS/UI/src/Implementation/Component/Table/Column/Path.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace ILIAS\UI\Implementation\Component\Table\Column;
+
+use ILIAS\UI\Component\Table\Column\Path as PathInterface;
+use ILIAS\UI\Component\Component;
+use ILIAS\UI\Component\Breadcrumbs\Breadcrumbs;
+
+class Path extends Column implements PathInterface
+{
+    public function format($value): Breadcrumbs
```
klees: Hi @lscharmer,

just noticed this here while reading: Can we simply hint on `Breadcrumbs` here instead of using that function below? 

Kind  regards!

klees:
Hi @lscharmer and @oliversamoila,

we have discussed via our UI-Coordinator channels. We would kindly ask you to rename this to `Breadcrumb Column` (instead of `Path` column...) because we already have a name for that thingy we want to display there and do not need another one... =)

Please take it to the JF afterwards...

Kind regards!

components/ILIAS/UI/src/Implementation/Component/Table/Column/Path.php
lscharmer: The function check is removed.

lscharmer:
> Hi @lscharmer and @oliversamoila,
> 
> we have discussed via our UI-Coordinator channels. We would kindly ask you to rename this to `Breadcrumb Column` (instead of `Path` column...) because we already have a name for that thingy we want to display there and do not need another one... =)
> 
> Please take it to the JF afterwards...
> 
> Kind regards!

The component is renamed to `Breadcrumb`.

klees:
Hi @lscharmer,

I removed us coordinators from the assignees to get this of our immediate to do lists. Please reassign us once this is through JF and ready for the review of the implementation.

Kind regards!

matthiaskunkel:
Jour Fixe, 23 JUN 2025: We highly appreciate this suggestion and accept the PR for trunk.

klees (changes requested):
Hi @lscharmer,

thanks a lot for this contribution! Really nice to see how the existing component fits in here...

- [x] examples: The text you have put under "expected output" above the example seems to rather be a "description". Wording might be bad here, as output could be read as "output of example". But that is not what is meant here: "description" shall describe the things we see in the example, so people would know what we are trying to show. "expected output" on the other hand is to be understood in the context of testrail. I guess it is called "expected results" there, so we might have the wording issue there (please report in mantis, if applicable!). Anyway, the thing you have put in "expected output" would be the "description". Please change accordingly. Bonus points: Is there some specific behaviour to be tested here? Add it to "expected output".
- [x] examples: Please remove padding-left for the breadcrumb in tables. Since both the breadcrumb and the cell contain padding, we lose space on the left. In addition, this means that the column title and the beginning of the breadcrumb are not aligned on a vertical line. My suggestion is to remove the “padding-left: 20px” from the breadcrumb in the table. This allows us to make optimal use of the space even with a narrow column.

Also, as this contains some CSS, I ping @catenglaender to look into that part.

Best regards!

klees:
Hi @lscharmer,

please pass back to me and ping me once changes are included.

Thanks!

lscharmer:
Hi @klees,
thank you for your review.

> * [x]  examples: Please remove padding-left for the breadcrumb in tables. Since both the breadcrumb and the cell contain padding, we lose space on the left. In addition, this means that the column title and the beginning of the breadcrumb are not aligned on a vertical line. My suggestion is to remove the “padding-left: 20px” from the breadcrumb in the table. This allows us to make optimal use of the space even with a narrow column.

I also removed the 6px padding around the other sides of the breadcrumbs component as it applies to them as well.

klees:
Hi @lscharmer,

thanks a lot!

Kind regards!

klees:
Hi @lscharmer,

just one additional note, which also might be interesting for everyone else:

We need to carefully consider the wording of these descriptions above the examples, as they will become testcases. We need to be specific enough to make sure that testers understand what they are supposed to look into. But we also should try to not be too specific so testcases won't need to be adapted for every change. Basically similar to out automated tests...

Interesting thing here is the explicit mentioning of `>` as delimiter in the breadcrumb. This feels like some design decision that could change at the whim of someone pretty quickly. So this might be something that breaks these tests soon. I will add a commit that makes this a little more generic.

Maybe something to consider next time...

Kind regards!

klees:
For future reference: [ea3a51fe9322500fd23e4ab809c5cb70afa5714b](https://github.com/ILIAS-eLearning/ILIAS/commit/ea3a51fe9322500fd23e4ab809c5cb70afa5714b)

PR #9577, opened 2025-05-15
Object/TileImageMigration: Make deletion of directories more robust

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=45072


kergomard:

You are clearly right that this is more beautiful and that I was being a little bit lazy here ;-) . I changed the sentence to not state so clearly that the migration has worked as there are other things that can go wrong see: https://github.com/ILIAS-eLearning/ILIAS/commit/3a8f4e3cf0cd4c548352f4b9380cf94676ad0c4c . No picking needed.

Thanks again and best,
@kergomard

PR #9578, opened 2025-05-15
style docu: added note for suffix-specific icons

fsesterh:
Added note under step 4 for suffix-specific icons that have to be managed via the admin gui. See: https://docu.ilias.de/go/wiki/wpage_7496_1357

This could also be merged with release_9 branch as far as I can tell.


Amstutz:
thx a lot!

PR #9579, opened 2025-05-15
Privacy/Security: Fix min/max values when requesting random value

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=45100

If approved, this must be picket to `release_10` and `trunk`.

### DIFF
--- a/Services/PrivacySecurity/classes/class.ilSecuritySettingsChecker.php
+++ b/Services/PrivacySecurity/classes/class.ilSecuritySettingsChecker.php
@@ -252,11 +252,9 @@ class ilSecuritySettingsChecker
                 ? $security->getPasswordMinLength()
                 : 6;
             $max = ($security->getPasswordMaxLength() > 0)
-                ? $security->getPasswordMaxLength()
-                : 10;
-            if ($min > $max) {
-                $max = $max + 1;
-            }
+                ? max($security->getPasswordMaxLength(), $min)
+                : max($min, 10);
+
             $random = new ilRandom();
             $length = $random->int($min, $max);
             $next = $random->int(1, 2);


klees:
I will look into this!

klees (commented):
Hi @mjansenDatabay,

the change looks as if it would indeed make the immediate error message disappear. But it also looks strange, because to me it looks as if we might end up in a situation where `$min == $max` and then we would loose randomness in length... This might not be a problem per se, though, but still strange. But yet again, also this `$min > $max` condition looks odd already...

In general, it looks as if we would be fixing at the wrong location and we should instead impose the constraint `$min <= $max` in the settings class already instead of dealing with the fallout. This is not a strict requirement, though, as we are basically looking into some code that neither you or me are actually responsible for.

Thoughts?

Kind regards!

mjansenDatabay:
Hi @klees ,

you are absolutely right.

I took another look, and it turns out that a constraint is already applied in the settings form via `\ilSecuritySettings::validate`:

```php
...
        if ($this->getPasswordMinLength() > 0 && $this->getPasswordMinLength() < $password_min_length) {
            $code = $password_min_length_error_code;
            if (!$a_form) {
                return $code;
            } else {
                $a_form->getItemByPostVar('password_min_length')
                       ->setAlert(sprintf(ilObjPrivacySecurityGUI::getErrorMessage($code), $password_min_length));
            }
        }
        if ($this->getPasswordMaxLength() > 0 && $this->getPasswordMaxLength() < $this->getPasswordMinLength()) {
            $code = self::SECURITY_SETTINGS_ERR_CODE_PASSWORD_MAX_LENGTH_LESS_MIN_LENGTH;
            if (!$a_form) {
                return $code;
            } else {
                $a_form->getItemByPostVar('password_max_length')
                       ->setAlert(ilObjPrivacySecurityGUI::getErrorMessage($code));
            }
        }
...
```

So according to the configuration form, `$min` is always smaller than or equal to `$max`, **if** both settings are set.

In my opinion, the edge case can still occur when:

* the minimum password length is explicitly set to a value `≥ 12`, and
* the maximum password length is disabled (set to `0`), which defaults to `10`

```php
            $min = ($security->getPasswordMinLength() > 0)
                ? $security->getPasswordMinLength()
                : 6;
            $max = ($security->getPasswordMaxLength() > 0)
                ? $security->getPasswordMaxLength()
                : 10;
            if ($min > $max) {
                ++$max;
            }
```

In that case, the `$min` ends up greater than `$max`, and although the code tries to adjust by incrementing `$max`, it may not be enough, e.g., `12 > 11`.

Best regards,
Michael

klees:
Hi @mjansenDatabay,

who would set a maximum length for a password anyway? :face_with_peeking_eye:

So, shall we default to `max($min, 10)` then, if no maximum length is set? I now also understand why the location for the fix is right indeed: We probably won't go into generating arbitrary length password for checking purposes, if max isn't set.

Kind regards!

mjansenDatabay:
Hi @klees 

> who would set a maximum length for a password anyway? 🫣

```
¯\_(ツ)_/¯
```

> shall we default to max($min, 10) then, if no maximum length is set

Done. I updated the PR anf forced-push the changes.

Best regards,
Michael

klees:
Thx @mjansenDatabay. Picked to release_10 and trunk.

PR #9581, opened 2025-05-15
Remove glyph action testing obligation

iszmais:
This removes glyph actions from the content of testing since this feature is deprecated for ILIAS 10+. (as mentioned [here](https://github.com/ILIAS-eLearning/ILIAS/pull/7697#issuecomment-2196467915))
This way new implemented glyphs are not obliged to implement a deprecated feature.

**Actions are still supported by established glyphs**
Test just don't fail anymore if new glyphs don't support them (to not conflict the deprecation).


Amstutz:
Thx a lot for the fix. Note, I created a ticket to remove those instances still relaying on the usage of the actions. Thx for pointing this out here! https://mantis.ilias.de/view.php?id=45115

PR #9582, opened 2025-05-15
UI: 45099, add newly compiled delos css file to include already merge…

lukastocker:
…d changes.

https://mantis.ilias.de/view.php?id=45099

This is a fix in regard of  https://github.com/ILIAS-eLearning/ILIAS/pull/9416 as the wrongly compiled delos css was merged there.


Amstutz:
thx a lot!

PR #9587, opened 2025-05-19
T&A10 43482: Fixes test export with user results and adds user file upload question file uploads to export and import

matheuszych:
[Mantis: 43482](https://mantis.ilias.de/view.php?id=43482)

When exporint a test with participant data, the file uploads from file upload questions do not get exported. This causes the files to be missing on import and preventing a manual correction if the question.

I believe this should fix this issue, by including the participants file uploads in the export and importing them on import into the systems storage.

### DIFF
--- a/components/ILIAS/Test/classes/class.ilTestImporter.php
+++ b/components/ILIAS/Test/classes/class.ilTestImporter.php
@@ -18,6 +18,7 @@
 
 declare(strict_types=1);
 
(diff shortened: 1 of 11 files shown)


kergomard:
Thank you very much @matheuszych for porting this to ILIAS 10!

I tested this quickly on a local test and everything works as expected. Merged and picked to trunk.

Best,
@kergomard

PR #9588, opened 2025-05-19
StudyProgramme: Fix undefined ilObjectCustomIconFactory

fhelfer:
For Study-programmes linked on the dashboard, an error is thrown.


klees:
Thanks. Looks plausible, but any idea what this is about and why this change was required?

kergomard:
Sorry, @klees ! This was my bad as I missed this usage when I namespaced the icon factory.

Best, @kergomard

klees:
Hehe, no problem =)

I suspected something like that. Thanks for improving stuff!

PR #9595, opened 2025-05-20
UI Tag Input: 45126 remove encoding as it is already encoded through …

lukastocker:
…the Tag Input code. 

Also strip tags to prevent e.g. JavaScript injections.

https://mantis.ilias.de/view.php?id=45126

Before fix:
TagInput_encoding_special_chars_before

Now with fix:
TagInput_encoding_special_chars_after


Amstutz:
thx a lot!

PR #9476, opened 2025-04-25
Remove deprecated MetaData classes, last usages

schmitz-ilias:
This PR removes all deprecated classes in MetaData. It also removes the last few usages of these classes, all of them related to export/import. The consequences of that were discussed [last year in the JF](https://docu.ilias.de/go/wiki/wpage_8457_1357).

Changes to the different components are in separate commits, Test and MediaObject are grouped together since they are related.

Note that it takes bit of patching up remaining export bugs in Survey and Test to be able to test these changes. That being said, with a few bandaid-patches everything seems to work fine.

`ilQuestionPageParser` required the most amount of surgery. I took out everything related to MetaData, and rebuilt what was necessary for the import to function smoothly. I would have liked a smaller footprint there, but that wasn't really possible, at least not without leaving a bunch of dead code lying around.

Let me know if you want anything done differently.

PR #9597, opened 2025-05-21
[Metadata] performance issue with unnecessary query on il_meta_rights

fneumann:
https://mantis.ilias.de/view.php?id=45131

The usage count for a license is only executed when the usage number of a license is needed.


schmitz-ilias:
Thank you very much for the improvement @fneumann!

Merged and picked to 10. Merging to trunk is not necessary, `ilMDCopyrightSelectionEntry::read` is already unused in 10, and will be removed with ILIAS 11 (see #9476).

PR #9605, opened 2025-05-22
UI: Do Not Double Encode Values

kergomard:
Dear UI-Coordinators

This PR is one way to solve the problem with mustache-placeholders in text-fields (See: https://mantis.ilias.de/view.php?id=45141). Let me know, if you would prefer another solution.

Thanks and best,
@kergomard

### DIFF
--- a/components/ILIAS/UI/src/Implementation/Component/Input/Field/Renderer.php
+++ b/components/ILIAS/UI/src/Implementation/Component/Input/Field/Renderer.php
@@ -256,7 +256,7 @@ class Renderer extends AbstractComponentRenderer
             // with declare(strict_types=1) in place,
             // htmlspecialchars will not silently convert to string anymore;
             // therefore, the typecast must be explicit
-            return htmlspecialchars((string) $v, ENT_QUOTES);
+            return htmlspecialchars((string) $v, ENT_QUOTES, 'utf-8', false);
         };
     }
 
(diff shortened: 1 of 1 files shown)


klees:
Hi @kergomard,

we (as: the UI coordinators) needed to ponder this a little to understand what this is actually doing, why this is necessary and what could be potential fallout. Thanks for [your additional explanation](https://mantis.ilias.de/view.php?id=45141#c116175).

To clarify, for all interested parties, what this is about, I want reiterate in my own words:

* The initial problem is a flaw in our general rendering process. Our templating engine `ilTemplate` uses curly braces for placeholders, just like mustache does. It can be called in a way that makes unused placeholders disappear. This behaviour is controlled by parameter in the call to the class and is also what triggers the initially reported bug. There is no easy fix, though, as locally changing that parameter is not enough. Somewhere in our call stack, there is a call to `ilTemplate` that uses that parameter. The call is in a location so generic, that we simply cannot know if it is ok to change the parameter there. If I remember correctly, that call is somewhere in `ilRepositoryGUI`, but I might be wrong, its been some time I last looked into this issue. **So fixing the root cause is not possible here.**
* **Stefan wants to work around the root cause.** The general idea is to include the curly braces that he needs for mustache into the value of a textarea as html-entities. This will render them correctly and they can later be retrieved as curly braces in the data processing part of the form as well. This does not work, because for the textarea we sprinkle some `htmlentities` over the values to avoid potential security issues caused by injections. The entities for the curly braces are encoded a second time and hence appear in their entity-form in the value. Hence, this PR instructs the escaping to skip already existing entities in the value.
* So, this change should not make any difference regarding security. Problematic chars like `<` are transformed into html entities before and after.
* This would only make a difference for people that explicitly had html-entities in their values before, which, after the change, would be displayed as the characters they represent. These people, if they exist at all, would need to call `htmlentities` on their value now (or file a ticket in mantis...).

So, not pretty, but possibly the bEsT way to solve the issue atm...

Kind regards!

klees:
I will merge later...

PR #9608, opened 2025-05-22
(trunk) Forum Bugfix #0045142: Fix ForumNotification overwritten by setting from all_users

fhelfer:
https://mantis.ilias.de/view.php?id=45142

Please also cherry-pick to release_10


mjansenDatabay:
Picked to `release_10`, thx Fabian

PR #9615, opened 2025-05-22
CoPage: Initialize `string`-typed properties

mjansenDatabay:
See: https://mantis.ilias.de/view.php?id=45146


alex40724:
Thanks for the PR!

PR #9619, opened 2025-05-26
UI: 44004, fix HTML validation error in ViewControl.

lukastocker:
https://mantis.ilias.de/view.php?id=44004


lukastocker:
Hi @Amstutz ,

as requested [here](https://github.com/ILIAS-eLearning/ILIAS/pull/9511) I tested the changes on a trunk branch and it works fine. Therefore I created this PR without any additional changes.

The failing PHPUnit tests do not result from this implementation - they already happened before this PR.

Greetings
Luka

Amstutz:
thx a lot!

iszmais:
@lukastocker @Amstutz im not fully certain about it but this pr seems to have broken the 8.3. pipeline. (At least the topic seem to be matching)

image

Can you look into this?

Greetings,
@iszmais

klees:
Hi @lukastocker and @Amstutz,

@alex40724 just pointed me to the same. Looks as if there is a deviation between test und template. The template says, that the `button` is inside the `ul`, the test says otherwise. I would say that the test is correct.

Any opinion?

I'm in here anyway, just give me a hint an I'll fix if you like.

Kind regards!

lukastocker:
Hi @Amstutz and @klees ,

the test is indeed correct. But the button should have been outside of the ul tag. I fixed it and opened a PR:

https://github.com/ILIAS-eLearning/ILIAS/pull/9624

Tests are running fine and the validation works also.

PR #9621, opened 2025-05-26
Init: Don't use absolute path to client.ini if `CLIENT_WEB_DIR` is de…

mjansenDatabay:
…fined

See: https://mantis.ilias.de/view.php?id=42368

If approved, this must be picked to `trunk`.


klees:
Picked to trunk.

PR #9625, opened 2025-05-27
Fix: Correct declarations of LTI symbols

ZallaxDev:
The "LTIS" const has been modified to redefine the "ltis" icon and a new "LTI" const has been added to define the "lti" icon.
With this change, the LTI configuration icon in the administrative panel appears correctly.


klees:
Hi @ZallaxDev,

that "s" at the end of the icon name: Is it for "Administration" like in many other cases? So, "LTI" is for the object and "LTIS" is for LTI administration? If so: Please document accordingly. If not: Do you know why we have that "s" there?

I know that the answer to this question might elude you as you wasn't part of the initial implementation of the functionality, but this caught might attention so I needed to ask anyway =)

Thanks!

klees:
Thanks!

klees:
Picked to trunk.

PR #9628, opened 2025-05-27
Mantis 45178: Data in loc_user_results is stored wrong for qualifying tests, if multiple initial and qualifying tests are used 

ingoj:
Wrong test type was returned if test was qualifying and multiple initial tests were used. 
Added a check tor type for multiple qualifying tests, too 
Adjusted copyright notice


schmitz-ilias:
Thank you very much for the analysis and the fix @ingoj!

schmitz-ilias:
Merged and picked to 10 and trunk.

PR #9637, opened 2025-05-28
T&A Prevent reset of submitted passes inside tst_active if extra time is added

thojou:
This PR is related to Mantis-Ticket https://mantis.ilias.de/view.php?id=45186.

It fixes an issue where already submitted tests are incorrectly shown as "not finished" after extra time is granted. The underlying problem was caused by a database query that reset submission data, which should no longer be necessary in ILIAS 9. The query has been removed to ensure correct status display and prevent user confusion.

This PR must only be merged into release_9, as the behaviour has been reworked from ILIAS >= 10.

Best,
@thojou

### DIFF
--- a/Modules/Test/classes/class.ilObjTest.php
+++ b/Modules/Test/classes/class.ilObjTest.php
@@ -7619,12 +7619,6 @@ class ilObjTest extends ilObject implements ilMarkSchemaAware
             }
 
             $this->db->manipulateF(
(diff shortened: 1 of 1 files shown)


kergomard:
Ok, it is a little tricky to be really sure that this is what we want. But I think it is.

Thank you very much @thojou and best,
@kergomard

PR #9638, opened 2025-05-28
Centralize datetime format fetching

iszmais:
This PR remove manual datetime switches to the centralized user function.


mjansenDatabay:
Thx @iszmais

